PR #2111, opened 2024-09-23
gpu: nvidia: Refactor to native parameters for matmul

ShanoToni:
# Description
This refactor is required to address current issues in nvidia backend matmul. Currently the use of runtime dimensions requires native-specific member objects to be initialised, which are specific to the dimensions (Tensor descriptors, matrix layouts). This is handled currently by cleaning up the members and re-initializing on execution. This is not thread safe.

The refactor proposes abstraction of the members to separate structs, which in the non-runtime dimension case would be initialised with the primitive descriptor and passed to the primitive on its respective creation. When not using runtime dimensions the parameters are cleaned up when the primitive is destroyed.

In the runtime dimension case, the parameters are initialised only with the non native specific members. Inside the execute call a copy is created of the new params struct and only then are the native specific structs initialised for the copy. When using runtime dimensions the params are cleaned up at the end of the host_task running the matmul.

# Checklist

## General


src/gpu/nvidia/cudnn_matmul.hpp
```diff
@@ -116,24 +122,25 @@ struct cudnn_matmul_t : cudnn_matmul_base_t {
             }
             return true;
         }
+
+        std::shared_ptr<cublas_params> params_;
```
densamoilov, line 126: Is there any reason why we want to have shared ownership over the params?

src/gpu/nvidia/cudnn_matmul_executor.hpp
```diff
@@ -82,22 +85,24 @@ struct cudnn_matmul_base_exec_t {
...
+            matmul_impl_->execute(cublas_handle, cudnn_handle, params, weights,
+                    src, dst, bias, reorder_scratch, src_scale, wei_scale,
                     dst_scale);
 
-            free_runtime_scratch(matmul_impl_->has_runtime_params(),
-                    cublas_handle, cuda_stream, bias_scratch_ptr);
+            if (params.has_runtime_params_) {
+                cudaStream_t streamId;
+                cublasGetStream(cublas_handle, &streamId);
+                cudaStreamSynchronize(streamId);
```
densamoilov, line 95: I think using the new host task with any synchronization is undefined behavior. Should we do this synchronization conditionally?

src/gpu/nvidia/cudnn_matmul_executor.hpp
```diff
@@ -380,35 +417,40 @@ struct cudnn_matmul_lt_runtime_args_exec_t final
...
-            std::size_t block_a_scratch_size, std::size_t block_b_scratch_size,
-            std::size_t block_c_scratch_size,
-            std::size_t src_scale_scratchpad_size,
-            std::size_t wei_scale_scratchpad_size) {
+            std::shared_ptr<cublas_lt_params> params,
+            const memory_desc_wrapper &src_d,
+            const memory_desc_wrapper &weights_d,
+            const memory_desc_wrapper &dst_d) {
+
+        cublas_lt_params matmul_params = *params;
```
densamoilov, line 425: How big are the params structures? Given that we have to copy the whole structure during each execution we may want be cautions about their size. if it's heavy then we could split it into two parts: creation time and runtime and recreate only the runtime part in the case of runtime dimensions.

If the structures are not big and we don't anticipate them to grow in the future then I guess it's fine to copy the whole structure.

densamoilov (commented):
The refactoring looks very nice to me, a long-awaited one to be honest. Thanks!

src/gpu/nvidia/cudnn_matmul.hpp
ShanoToni: It needs to be created in the primitive descriptor and it is used by the implementation, attempting to pass a unique_ptr back and forward from impl to primitive might not be desirable.

src/gpu/nvidia/cudnn_matmul_executor.hpp
ShanoToni: Rebased on the new changes and now uses the utility func `device_sync`

src/gpu/nvidia/cudnn_matmul_executor.hpp
ShanoToni: There is only so much we can reduce the copy, a small refactor(another one) ensures we only allocate a new local params pointer for runtime dims cases and in any other we pass a shared_ptr.

spalicki (approved):

PR #2112, opened 2024-09-23
attributes: remove obsolete quantization internals

dzarukin:
I'd like to start the process of improving attributes internals responsible for quantization.
There were two major changes affected that piece of functionality:
* v3.0 moved from static to dynamic quantization.
* v3.4 introduced data types and groups for quantization parameters.

Both changes left a pile of technical debt towards a better solution, and it start biting in attempts to extends groups and data type support further.

To succeed with refactor, some obsolete entries required to be removed, and later to be replaced with modern analogue of that functionality in affected implementations.

This PR removes `oscale` abstraction and everything related to it as there's no way to invoke it through a public API. It also removes `defined()` method as unneeded. The change touches all teams as there are old pieces of code laying there. Please confirm you are fine with the change and will update quantized support in the correspondent backend as and if needed.

### DIFF
--- a/src/common/gemm.cpp
+++ b/src/common/gemm.cpp
@@ -85,7 +85,7 @@ std::string get_descriptor(dim_t M, dim_t N, dim_t K) {
         if (!is_src_ab && lda != M) ss << "lda:" << lda << " "; \
         if (is_wei_ab && ldb != N) ss << "ldb:" << ldb << " "; \
         if (!is_wei_ab && ldb != K) ss << "ldb:" << ldb << " "; \
-        if (alpha != 1.f) ss << "attr-oscale:common:" << alpha << " "; \
+        if (alpha != 1.f) ss << "attr-scales:src:common:" << alpha << " "; \
         if (beta != 0.f) ss << "attr-post-ops:sum:" << beta << " "; \
         ss << ",," << get_descriptor(M, N, K); \
         VPROF(start_ms, primitive, exec, VERBOSE_profile, ss.str().c_str(), \
--- a/src/common/primitive_attr.cpp
+++ b/src/common/primitive_attr.cpp
@@ -40,7 +40,7 @@ const runtime_scales_t &default_runtime_scale() {
     return default_runtime_scale_instance;
 }
 
-void scales_t::set_single_scale(float scale) {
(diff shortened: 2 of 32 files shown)


dzarukin:
@oneapi-src/onednn-cpu-aarch64 @oneapi-src/onednn-gpu-nvidia Kind ping to check on the changes and confirm you are fine with them.
Please respond till EoW (the earlier the better), otherwise, I'll assume changes are good to go.

mgouicem:
@theComputeKid @t4c1 
Could you take a look at this PR? It removes output scaling in quite a few implementations (though those should not have been reachable under normal use of the API).
Reviewing this PR might help you re-enable this feature under the new API.

t4c1 (commented):
SYCL/AMD/Nvidia changes look fine, but they are relatively small. The main question I have is do we actually want to remove a feature without having a replacement ready?

mgouicem:
> SYCL/AMD/Nvidia changes look fine, but they are relatively small. The main question I have is do we actually want to remove a feature without having a replacement ready?

We actually do have the replacement, see [this RFC](https://github.com/oneapi-src/oneDNN/tree/rfcs/rfcs/20220201-quantization-scaling#3-unifying-scaling-in-long-term).

The output_scales mechanism was replaced with per-argument scales. In a nutshell, it allows:
- to align scale with zero-point mecanism
- to get scales at runtime instead of creation time, improving primitive cache hit-rate, and reducing caching overheads
- to better match user storage of scales, and avoid them overheads of output_scales pre-computation.

Internally, you have two options:
- wait for Dima to do the refactor of internals to simplify working with quantization parameters
- move to new arg_scales mecanism, right now. If you want an example of the arg_scales mecanism, a good example is in src/cpu/matmul/ref_matmul.{h,c}pp.

theComputeKid:
@mgouicem we are currently in the process of making some changes in aarch64 for quantization and will check how it affects us. Will get back to you shortly.

cc: @renato-arantes

theComputeKid:
Just checked and all good from my side. From an aesthetic perspective, could you please trigger the new MacOS pipelines with Werror enabled to make sure it passes build? (You might need to rebase to run them).

theComputeKid (approved):

PR #2117, opened 2024-09-24
github: workflows: Add macos smoke tests

theComputeKid:
Adding smoke tests for macos-14, for clang.

Unfortunately, we have test failures (per config) and warnings (per compiler), so we cannot turn on Werror and we also need to attach test failures to specific configurations.

We want to remove all test failures and warnings. This is a work in progress to at least remove some warnings (#2116). If we can get ci in as fast as possible for at least one OS, even in a primitive, messy state, we can stop more test failures and warnings from being added.

I also applied formatting to the YAML file.

There are some problems with GCC testing that needs some investigation from myself but I didn't want GCC to stop me from starting testing in ci for clang.

### DIFF
--- a/.github/workflows/ci-aarch64.yml
+++ b/.github/workflows/ci-aarch64.yml
@@ -35,18 +35,24 @@ permissions: read-all
 
 jobs:
   macos:
-    name: macOS
     runs-on: macos-14
     strategy:
       matrix:
-        compiler: [
-          {CC: clang, CXX: clang++},
-          {CC: gcc-14, CXX: g++-14}
-        ]
-        config: [
-          {CMAKE_BUILD_TYPE: Debug, ACL_WITH_ASSERTS: '1'},
-          {CMAKE_BUILD_TYPE: Release, ACL_WITH_ASSERTS: '0'}
-        ]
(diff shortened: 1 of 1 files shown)


theComputeKid:
Hmm.. Not ready for review yet. I need to clean the job name string for every matrix combo.

vpirogov:
> Hmm.. Not ready for review yet. I need to clean the job name string for every matrix combo.

It would be great to move out as much logic as possible from actions to build.sh/test.sh scripts like it's [done for Azure CI](https://github.com/oneapi-src/oneDNN/blob/main/.github/automation/.azure-pipeline.yml). This would help with with debugging and reusability of the code. Also with the job names.

theComputeKid:
@vpirogov I fixed the names at least, now we can see clean names:

```
CI AArch64 / macOS (clang, Release)
CI AArch64 / macOS (gcc-14, Debug)
```

I 100% agree with your sentiment, I was waiting to have at least 2 OS's to make sure I have a cross platform build script, so that I only need to write it once. I'll bring forward my plans and do it as part of this change then.

vpirogov (approved):

PR #2119, opened 2024-09-24
[WIP] aarch64: add b_is_vnni member to brgemm_attr_t same as for x64

ankalinin:
This is aarch64 build error fix ("error: no member named 'b_is_vnni' in 'dnnl::impl::cpu::aarch64::brgemm_attr_t'")


theComputeKid (approved):
Thanks for the quick response. I'm happy that the ci caught this, and it shows that the ci is doing its job.

PR #2121, opened 2024-09-25
gpu: nvidia: Added support for native host task extension

ShanoToni:
# Description

Adding support for the extension `ext_codeplay_enqueue_native_command` [LINK](https://github.com/intel/llvm/blob/2f0abc61d551b773e0497276c61fbcc576475a23/sycl/doc/extensions/experimental/sycl_ext_codeplay_enqueue_native_command.asciidoc#L2) to documentation.

Apart from improved dependencies for native commands, the native command provides less execution overhead, improving performance.

Example:
```
# old host task
    total perf: min(ms):38.256 avg(ms):97.7698
# native host task
    total perf: min(ms):33.4952 avg(ms):33.7441
```

Results generated running on A100 with command
`--eltwise --engine=gpu --mode=P --skip-impl=dpcpp:ref:any --batch=tests/benchdnn/inputs/eltwise/test_eltwise_gpu`

# Checklist

## General

## Performance improvements

- [ x ] Have you submitted performance data that demonstrates performance improvements?

### DIFF
--- a/src/gpu/nvidia/sycl_cuda_compat.hpp
+++ b/src/gpu/nvidia/sycl_cuda_compat.hpp
@@ -35,9 +35,13 @@ T get_native_mem(const interop_handle &ih, U acc) {
             ih.get_native_mem<::sycl::backend::ext_oneapi_cuda>(acc));
 }
 
-template <typename T>
-void host_task(::sycl::handler &cgh, const T &task) {
+template <typename HandlerT, typename FnT>
+void host_task(HandlerT &cgh, const FnT &task) {
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+    cgh.ext_codeplay_enqueue_native_command(task);
+#else
     cgh.host_task(task);
+#endif
 }
 
 template <typename native_object_t, typename sycl_object_t,
--- a/src/gpu/nvidia/cudnn_batch_normalization_executor.hpp
+++ b/src/gpu/nvidia/cudnn_batch_normalization_executor.hpp
@@ -192,14 +192,12 @@ protected:
             xpu::sycl::interop_memory_arg_t<::sycl::access::mode::write>
                     arg_scale,
             float val, const size_t n) const {
-        cuda_stream->interop_task([&](::sycl::handler &cgh) {
-            T *scale_ptr = static_cast<T *>(arg_scale.get_native_pointer(ih));
-            CUDA_EXECUTE_FUNC(cuMemsetD32Async,
-                    reinterpret_cast<CUdeviceptr>(scale_ptr),
-                    reinterpret_cast<int &>(val), n,
-                    cuda_stream->get_underlying_stream());
-            cudaDeviceSynchronize();
-        });
+        T *scale_ptr = static_cast<T *>(arg_scale.get_native_pointer(ih));
+        CUDA_EXECUTE_FUNC(cuMemsetD32Async,
+                reinterpret_cast<CUdeviceptr>(scale_ptr),
+                reinterpret_cast<int &>(val), n,
+                cuda_stream->get_underlying_stream());
+        sync_device();
     }
 
     // Handle the cases when mean and var are read-only accessors or nullptr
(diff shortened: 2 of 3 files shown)


src/gpu/nvidia/sycl_cuda_compat.hpp
```diff
@@ -35,11 +35,16 @@ T get_native_mem(const interop_handle &ih, U acc) {
...
+void host_task(HandlerT &cgh, const FnT &task) {
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+    cgh.ext_codeplay_enqueue_native_command([=](sycl::interop_handle ih) {
+#else
+    cgh.host_task([=](sycl::interop_handle ih) {
+#endif
+        task(std::move(ih));
+    });
 }
-
```
mgouicem, line 42: (nit) maybe restore the newline for readability?

src/gpu/nvidia/sycl_cuda_compat.hpp
```diff
@@ -35,11 +35,16 @@ T get_native_mem(const interop_handle &ih, U acc) {
...
-    cgh.host_task(task);
+template <typename HandlerT, typename FnT>
+void host_task(HandlerT &cgh, const FnT &task) {
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+    cgh.ext_codeplay_enqueue_native_command([=](sycl::interop_handle ih) {
+#else
+    cgh.host_task([=](sycl::interop_handle ih) {
+#endif
+        task(std::move(ih));
+    });
```
mgouicem, line 46: (nit) would it work for readability, to not split braces with macro conditionals and do something like:

```suggestion
#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
    cgh.ext_codeplay_enqueue_native_command(task);
#else
    cgh.host_task(task);
#endif
```

mgouicem (approved):

kala855 (approved):

src/gpu/nvidia/sycl_cuda_compat.hpp
ShanoToni: Line has been added back in.

src/gpu/nvidia/sycl_cuda_compat.hpp
ShanoToni: Agreed, changes made.

t4c1: Do you need to wrap the task in another lambda compared to what Mourad suggested?

ShanoToni: Removed the unnecessary lambda, thanks for catching my miss @t4c1!

t4c1 (approved):

src/gpu/nvidia/sycl_cuda_compat.hpp
```diff
@@ -35,9 +35,13 @@ T get_native_mem(const interop_handle &ih, U acc) {
             ih.get_native_mem<::sycl::backend::ext_oneapi_cuda>(acc));
 }
 
-template <typename T>
-void host_task(::sycl::handler &cgh, const T &task) {
+template <typename HandlerT, typename FnT>
```
densamoilov, line 38: The spec says:
> The call to interopCallable must not submit any synchronous tasks to the native backend object, and it must not block waiting for any tasks to complete. The call also must not add tasks to backend objects that underly any other queue, aside from the queue that is associated with this handler. If it does any of these things, the behavior is undefined.

And we have a few places where we do explicitly synchronize, [here](https://github.com/oneapi-src/oneDNN/blob/51b489fe514cb827de37ba28246ea435bc81cce7/src/gpu/nvidia/cudnn_batch_normalization_executor.hpp#L201), for example.

It seems that it may become an issue when `ext_codeplay_enqueue_native_command` is used.

densamoilov: In validation the following case hung and the process was killed by timeout: `--mode-modifier=P --bnorm --engine=gpu --inplace=true ic23n"bnorm_ci_0d:1"`, so there seems to be an issue indeed. Given that we enable the new host task extension conditionally we should also introduce a a function that should do the synchronization under the same condition.

ShanoToni: I have made some changes and added a utility `sync_device` which would synchronize only when using default host_task.
I wanted to additionally ask, what is the reason for the additional interop_task? If my understanding is correct the current situation is ```q.submit{ host_task { q.submit { memset/device sync } } }``` in the current state I do not think that the internal submit will wait on any prior events, as that would be handled by the external submit, and no following dependencies will come out from the internal submit as we synchronize the device in the `host_task` case, while the `enqueue_native_command` would resolve any dependencies.

I have removed the internal interop task and as I am not seeing failures without it, let me know if this change does not make sense/ is wrong, and I can revert it.

densamoilov: I don't know why there was the additional interop_task and I can't think of any reason why we may need it if we block the entire device anyway.

densamoilov (approved):

PR #2126, opened 2024-09-26
cpu: aarch64: improve memory management safety for jit_uni_softmax

Permanence-AI-Coder:
# Description

The code changes in this PR were generated automatically by the Permanence AI Coder and reviewed by @fwph & @jweese. This PR updates memory allocation for `softmax_driver_ ` objects in two structs. Note that in the macro `DNNL_DISALLOW_COPY_AND_ASSIGN` was missing; it has been added to be consistent with other objects. 

on-behalf-of: @permanence-ai <github-ai@permanence.ai>

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/jit_uni_softmax.cpp
+++ b/src/cpu/aarch64/jit_uni_softmax.cpp
@@ -16,6 +16,7 @@
 *******************************************************************************/
 
 #include <assert.h>
+#include <memory>
 
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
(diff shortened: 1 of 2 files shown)


theComputeKid (approved):

Sqvid (approved):

spalicki:
@Sqvid You have approved this change and then re requested a review on yourself, do you want more time for this review?

theComputeKid:
@spalicki Could you clarify why we are modifying to `utils::make_unique` instead of using the std lib? Seems like a pattern, but still, wanted to clarify.

dzarukin:
> @spalicki Could you clarify why we are modifying to `utils::make_unique` instead of using the std lib? Seems like a pattern, but still, wanted to clarify.

std lib version is available only from C++14, while by default oneDNN gets compiled with C++11. That's the primary reason.

spalicki:
`std::make_unique` is not part of C++11, only C++14.

PR #2130, opened 2024-09-27
github: workflows: Enable Arm Hosted GitHub Runners

theComputeKid:
The oneDNN repo has now been enabled with Arm hosted GitHub runners. They have been activated for PRs opened against the repo.

This is an initial version of AArch64 CI for Linux, on c6g and c7g. We build for gcc-13 and clang-17, in debug and release, for openmp builds. For now, we enable smoke tests for most, but not all, configurations (in order to preserve runner resources). We have logic in the test scripts to skip tests on certain configs. As per suggestions from @vpirogov (and I agree), CI logic has been moved into cross-platform scripts.

The CI is dependent on MacOS pipelines passing. Total run time seems to be in line with x86:
- MacOS: 45 minutes
- Linux: 20-30 minutes

Overall time for a run: 1:15 (reasonable).

A few notes:

1. I have renamed the GitHub jobs generically so that they do not need to be updated when we upgrade compilers (e.g. gcc vs gcc-13). This means we can set them as required checks once then and forget.
2. I would like to rethink the CI scripts after initial work on CI is finished, to clean them up. Ideally, the scripts should allow developers to reproduce builds when debugging.
3. Threadpool builds still need to be added.
4. Full extensive CI testing needs to be added.
5. We should consider whether we should test multiple compiler versions in the future.
6. I believe we should be able to cache ACL builds in the future. We should explore this to save resources (approx 30% of compute time), instead of building every time. This can allow us to use smaller runners, of which more are available concurrently.

Unfortunately for oneDNN repo admins (@mgouicem, @vpirogov), you will need to tediously add all combinations of checks as "required", e.g.:

1. CI AArch64 / c7g, clang, OMP, Debug 
2. CI AArch64 / c6g, clang, OMP, Release
etc etc

Apologies for the inconvenience. However, this is a one-time effort as we don't need to update them later as those tags will rarely change. You will also need to remove the previous MacOS checks, e.g.: `macOS (gcc-14, Debug)` and replace with the current version-less MacOS checks, e.g.: `macOS (gcc, Debug)`.

For more info: 
1. https://github.com/apps/arm-hosted-gha-runners
2. https://gitlab.arm.com/tooling/gha-runner-docs/

### DIFF
--- a/.github/automation/build_aarch64.sh
+++ b/.github/automation/build_aarch64.sh
@@ -1,0 +1,51 @@
+#! /bin/bash
+
+# *******************************************************************************
+# Copyright 2024 Arm Limited and affiliates.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+
+# Build oneDNN for aarch64.
+
+set -o errexit -o pipefail -o noclobber
+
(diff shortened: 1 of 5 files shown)


theComputeKid:
@vpirogov and @mgouicem : change is ready

mgouicem (approved):
THanks !

PR #2134, opened 2024-09-30
cpu: aarch64: Fix C_graph_fusions_cpu test

Radu2k:
Remove asserts in brgemm_matmul_utils.cpp to make
test_benchdnn_modeC_graph_fusions_cpu fail or pass. The datatype support should be checked in
check_datatype. This has been renamed as well from check_isa_with_datatype to check_datatype.

# Description

By enabling all graph API tests on AArch64 as a follow-up of this change (https://github.com/oneapi-src/oneDNN/pull/2099#issuecomment-2379967176) unveiled that `ctest -R "test_benchdnn_modeC_graph_fusions_cpu"` failed due to the assets on lines 323, 324 and 325 in `src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp`.  

The test segfaulted because `pick_blocked_B_layout` is called before the init, when the object is constructed. These datatypes are checked in the init and return status unimplemented. 

This behaviour should be changed to calling `pick_blocked_B_layout` in configure after init has passed.

# Checklist
## General

### Bug fixes

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -129,9 +129,8 @@ bool post_ops_ok(brgemm_matmul_conf_t &bgmmc, const primitive_attr_t &attr,
                     true /*sum_requires_same_params*/, bcast_set));
 }
 
-status_t check_isa_with_datatype(
-        const brgemm_matmul_conf_utils_t &bm_conf_utils) {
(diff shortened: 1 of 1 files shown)


theComputeKid (approved):

mgouicem (approved):

mgouicem:
COuld you please run clang-format on your patch? it seems lightweight scans are flagging violations in src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp.

PR #2149, opened 2024-10-04
cpu: aarch64: hot fix for segfault in cached Winograd

taoye9:
# Description

This is a hot fix patch for aarch64 wino grad conv. The problem happens when same wino conv is taken from primitive cache due to wrong memory management behavior.

# Checklist

## General


theComputeKid (changes requested):
Need to make it memory safe.

theComputeKid (approved):

PR #2151, opened 2024-10-04
cpu: aarch64: hot fix for segfault in cached winograd

taoye9:
# Description

Same PR as https://github.com/oneapi-src/oneDNN/pull/2149 but to backport to release v3.6.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/acl_winograd_convolution.cpp
+++ b/src/cpu/aarch64/acl_winograd_convolution.cpp
@@ -79,13 +79,14 @@ status_t acl_wino_convolution_fwd_t::pd_t::init(engine_t *engine) {
 }
 
 status_t acl_wino_convolution_fwd_t::init(engine_t *engine) {
-    auto acp = pd()->acp_;
-    acl_obj_->conv.configure(&acp.src_tensor_info, &acp.wei_tensor_info,
-            acp.with_bias ? &acp.bia_tensor_info : nullptr,
-            &acp.dst_tensor_info, acp.padstride_info, acp.act_info,
-            true); // to support 5x5, 7x7 filter shapes in addition to 3x3
+    // commented due to hot fix solution for stateless API which should be replaced soon.
+    //     auto acp = pd()->acp_;
+    //     acl_obj_->conv.configure(&acp.src_tensor_info, &acp.wei_tensor_info,
+    //             acp.with_bias ? &acp.bia_tensor_info : nullptr,
+    //             &acp.dst_tensor_info, acp.padstride_info, acp.act_info,
+    //             true); // to support 5x5, 7x7 filter shapes in addition to 3x3
 
-    acl_obj_->aux_mem_req = acl_obj_->conv.workspace();
+    //     acl_obj_->aux_mem_req = acl_obj_->conv.workspace();
     return status::success;
 }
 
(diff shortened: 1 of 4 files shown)


theComputeKid (approved):
Even at its worst, it is better than having a segfault and the wrong answer.

src/cpu/aarch64/acl_winograd_convolution.hpp
```diff
@@ -59,7 +59,7 @@ struct acl_wino_convolution_fwd_t : public primitive_t {
 private:
     status_t execute_forward(const exec_ctx_t &ctx) const;
     const pd_t *pd() const { return (const pd_t *)primitive_t::pd().get(); }
-    std::unique_ptr<acl_obj_t<Op>> acl_obj_;
+    mutable std::unique_ptr<acl_obj_t<Op>> acl_obj_;
```
mgouicem, line 62: why not declare acl_obj_ directly in execute forward?
IIUC, this patch would still break in cache of concurrent execution.

src/cpu/aarch64/acl_winograd_convolution.cpp
```diff
@@ -131,6 +131,15 @@ status_t acl_wino_convolution_fwd_t::pd_t::init_conf() {
...
 status_t acl_wino_convolution_fwd_t::execute_forward(
         const exec_ctx_t &ctx) const {
+    acl_obj_ = std::make_unique<acl_obj_t<Op>>();
+    auto acp = pd()->acp_;
+    acl_obj_->conv.configure(&acp.src_tensor_info, &acp.wei_tensor_info,
+            acp.with_bias ? &acp.bia_tensor_info : nullptr,
+            &acp.dst_tensor_info, acp.padstride_info, acp.act_info,
+            true); // to support 5x5, 7x7 filter shapes in addition to 3x3
+
+    acl_obj_->aux_mem_req = acl_obj_->conv.workspace();
```
mgouicem, line 141: with the constructor and configure calls, would this impl still be faster?
If not, why not disable the impl from dispatch list?

theComputeKid (changes requested):
Need to make it thread-safe.

src/cpu/aarch64/acl_winograd_convolution.hpp
mgouicem: Thanks. I am still confused by the use of the mutex. Why not make the object local to the execute function call? Each execute call would have his own object and this will remove any need for synchronization.

Furthermore, as currently implemented, I believe you might still have an issue in case of concurrent execution as only the initialization is mutex protected. For example if you have
```
t0: lock -> acl_obj init *, dtor previous -> unlock ->  acl_obj.get() ->            ...             -> using handle
t1:                                                     lock          -> acl_obj init, dtor previous *-> unlock -> acl_obj.get() -> ... 
```

In the above example, t0 will get some pointer to object upon acl_obj.get() call, but the underlying object will be destroyed by t1 acl_obj init.

taoye9: hi, @mgouicem, thanks for your help, i've changed the acl_obj to a local variable. 

FYI, I'm kind of confused about what kind of synchronisation should be posed on oneDNN primitives and how the primitives cache works? could you please share some docs/info?

mgouicem (approved):

src/cpu/aarch64/acl_winograd_convolution.hpp
mgouicem: Regarding synchronization, primitive.execute function should be const qualified. So because it is not supposed to change any state of the primitive object, there should be no sync necessary in execution.

We do have synchronizations happening upon primitive cache accesses (single writer, multiple readers). This should be transparent from primitive implementation.

Regarding scratchpad, when user managed, scratchpad is just another output from a primitive perspective (so user responsibility to sync when needed). When library managed, it is trickier as it is not thread-safe by default. You can find more info in the scratchpad devguide [page](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_scratchpad.html).

theComputeKid (approved):

src/cpu/aarch64/acl_winograd_convolution.hpp
taoye9: thanks mgouicem, that's really helpful!

PR #2156, opened 2024-10-09
rfcs: graph api: support swish operation

TaoLv:
The RFC proposes adding support for Swish operation in Graph API.

Rendered version: [link](https://github.com/TaoLv/mkl-dnn/blob/lvtao/rfcs/swish/rfcs/20241008-graph-api-swish/README.md) .

### DIFF
--- a/rfcs/20241008-graph-api-swish/README.md
+++ b/rfcs/20241008-graph-api-swish/README.md
@@ -27,7 +27,7 @@ networks. For examples:
 
 - PyTorch supports Swish via the SiLU operation [[#7]][7]. The operation does
   not support specifying `factor` in the formula.
-- OpenVINO support Swish via the Swish operation [[#8]][8]. Unlike PyTorch's
+- OpenVINO supports Swish via the Swish operation [[#8]][8]. Unlike PyTorch's
   SiLU operation, OpenVINO's Swish also accepts a scalar input `Beta` as the
   multiplication `factor` for Sigmoid.
 - For ONNX, a PR is working in progress to add Swish operation [[#9]][9].
@@ -37,6 +37,9 @@ networks. For examples:
 - cuDNN backend API supports Swish as a mode (`CUDNN_POINTWISE_SWISH_FWD`) of
   its Pointwise operation [[#11]][11] and accepts attribute
   `CUDNN_ATTR_POINTWISE_SWISH_BETA` as the multiplication `factor`.
+- Please note that, even PyTorch has SiLU operation, there are still many model
+  scripts choosing to implement swish with a composition of smaller operations
+  [[#12]][12].
 
 ## Proposals
 
@@ -76,6 +79,10 @@ swish.finalize();
 Pros:
 
 - There is no need to define and maintain a new operation in oneDNN Graph API.
+- Composition of smaller operations makes it possible and scalable to extend whe
+  the activation has more variants or flavors in the future.
+- The approach of composition of Multiply and Sigmoid is also adopted in models
+  as mentioned above.
 
 Cons:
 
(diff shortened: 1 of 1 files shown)


rfcs/20241008-graph-api-swish/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Option 2: Support Swish as a dedicate operation
+
+As aforementioned, main stream frameworks and libraries all support Swish as a
+dedicate operation. We think that it's reasonable to add a new Swish operation
+in oneDNN Graph API. The proposed operation schema is as follow:
+
+- Operation Kind: `Swish` (C++), `dnnl_graph_op_swish` (C).
+- Input/output: Single input, single output.
+- Attribute: `beta` (optional). `beta = 1.f` if not provided.
```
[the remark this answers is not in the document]
TaoLv: Thank you for the review, @mgouicem. I didn't think too much about the naming here. It just followed the naming in OpenVINO and cuDNN. I feel alpha is also good to me if we think it's better to align with eltwise primitive.

TaoLv:
@mgouicem Thanks for the review. I just incorporate the comments into the RFC and leave the conclusion section open for discussion.

TaoLv:
> I would actually advocate for using a composition of simpler ops (option 1) ...

Thanks. I should have mentioned that we already have this composition for swish in our code ([link](https://github.com/oneapi-src/oneDNN/blob/9db41d4197914e88ca62c3036a5b7c6411b7b023/src/graph/backend/dnnl/passes/transform.cpp#L1719)). It indeed worked for some requests from frameworks. But we also see some issues as I mentioned in the cons of option 1.

> As an example, just search swish in [HuggingFace github](https://github.com/search?q=org%3Ahuggingface%20swish&type=code).

Thanks for the link. I copied the link to the RFC also. :)
For these cases, if framework developers want to optimize it with oneDNN, they will have to detect the pattern, rewrite it with PyTorch's SiLU operation, and then call oneDNN to optimize the SiLU operation, if we don't consider further fusion into other operations like matmul or conv. So it still becomes if we want to use two or more operations to optimize SiLU or just one dedicated swish operation.

> There are often "equivalent" formulas that are used in the wild.

Yes, this is really a good point. Using composition of smaller operations will give us the flexibility to support more variants, without breaking or adding API. I also added this into the proposals.

rfcs/20241008-graph-api-swish/README.md
```diff
@@ -0,0 +1,185 @@
...
+- Compared to a dedicate Swish operation, this proposal requires more users code
+  (at least one more logical tensor and one more operation).
+- It also requires complex logic in the backend to detect `Sigmoid + Multiply`
+  and map to the existing Swish kernels in oneDNN. It requires the input of
+  Sigmoid and the second input of Multiply to be the same tensor.
+- Considering that SiLU is a built-in operation in PyTorch, mapping it to two
+  operations in oneDNN Graph is troublesome for some integrations.
+- Currently, oneDNN Graph Sigmoid operation does not support a multiplication
+  `factor`. We may need to extend either the proposed Swish graph or the Sigmoid
+  operation to support cases where `factor != 1.f`.
```
TaoLv, line 98: Since PyTorch's SiLU operation only supports factor = 1.f, it can be simply decomposed to Sigmoid + Multiply. When we dispatch these pattern to oneDNN swish kernel, we just set alpha = 1.f. It works perfectly.
But if we want to support factor != 1.f for other cases, the operation will be decomposed into Multiply + Sigmoid + Multiply with the first Multiply to do `factor * x`. To dispatch this pattern into oneDNN swish kernel, we have more things to consider:
- We need to check that the factor input of the first Multiply should be a scalar.
- The value of the scalar should be const among iterations.
- The value of the scalar should be known at compilation stage as it's required to create primitive descriptors. But for now, the compile() API only accepts the logical tensors of input tensors, not the values of input tensors.

mgouicem: THat is correct. For that case, you would need an extra attribute to Sigmoid with alpha value.

TaoLv: Thanks! Or, can we simply rely on binary + eltwise post-op + binary post-op for the cases where we don't know factor at compilation?

mgouicem: That is an option as well. THough I would expect for the specific case of switch, the alpha parameter is constant, and using switch eltwise would be faster.
The other thing is GPU support. alpha might not be a memory object on GPU.

ElaineBao (approved):

TaoLv:
@mgouicem  Could you please take another look on this? The conclusion part has been updated in the last commit (I will squash them once ready for merge). Thank you!

PR #2159, opened 2024-10-09
generic: fix several gtest issues

sgeor255:
# Description

This PR fixes several gtest issues on generic vendor:
- `NDR.LocalSize[0]==0` assertion was being thrown due to problems with 0-dim input/outputs not handled correctly
- Several unsupported cases were not skipped
- PRELU was missing a check for post-ops in `init()`
- InnerProduct & RNN tests were not skipped (IP & RNN are not implemented on generic vendor yet)
- There are some examples that are still failing due to unimplemented features which will be fixed in a separate PR
- ~~There are some convolution backward data & deconvolution tests that are still failing which will be fixed in a separate PR (as the issue doesn't seem to be caused by missing skips)~~

### DIFF
--- a/src/gpu/generic/sycl/ref_convolution.cpp
+++ b/src/gpu/generic/sycl/ref_convolution.cpp
@@ -74,6 +74,8 @@ status_t ref_convolution_fwd_t::init(impl::engine_t *engine) {
 }
 
 status_t ref_convolution_fwd_t::execute(const exec_ctx_t &ctx) const {
+    if (memory_desc_wrapper(pd()->dst_md()).size() == 0) return status::success;
+
     parallel_for(ctx, kernel_, [&](::sycl::handler &cgh) {
         convolution_kernel_fwd_t convolution_kernel(pd()->conf_, cgh, ctx);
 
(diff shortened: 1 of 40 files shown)


t4c1 (approved):

sgeor255:
The failing CI check is in a file not added by this PR.

src/gpu/generic/sycl/ref_convolution.cpp
```diff
@@ -134,6 +136,8 @@ status_t ref_convolution_bwd_data_t::init(impl::engine_t *engine) {
 }
 
 status_t ref_convolution_bwd_data_t::execute(const exec_ctx_t &ctx) const {
+    if (memory_desc_wrapper(pd()->dst_md()).size() == 0) return status::success;
```
t4c1, line 139:
```suggestion
    if (memory_desc_wrapper(pd()->diff_src_md()).size() == 0) return status::success;
```

src/gpu/generic/sycl/ref_softmax.cpp
```diff
@@ -52,6 +52,8 @@ status_t ref_sycl_softmax_fwd_t::init(impl::engine_t *engine) {
 }
 
 status_t ref_sycl_softmax_fwd_t::execute_forward(const exec_ctx_t &ctx) const {
+    if (pd()->has_zero_dim_memory()) return status::success;
```
mgouicem, line 55: Could you clarify what failure do these checks prevents?
I would expect this kind of check to happen during pd creation time.

src/gpu/generic/sycl/ref_convolution.cpp
sgeor255: Thanks, I am guessing this is the fix for the failing convolution tests? :)

t4c1: yes, that is correct.

sgeor255:
make test
disable device_cpu
enable device_gpu
enable thr_cuda
enable arch_rtx

mgouicem:
make test
disable device_cpu
enable device_gpu
enable thr_generic
enable arch_rtx

PR #2163, opened 2024-10-11
gpu: nvidia: Added missing bf16 threshold for softmax

ShanoToni:
# Description

Softmax benchdnn was missing a precision threshold for bf16, this PR adds it for the nvidia backend, as it caused some failing tests there.


tests/benchdnn/softmax/softmax.cpp
```diff
@@ -240,10 +240,12 @@ void setup_cmp(compare::compare_t &cmp, const prb_t *prb, data_kind_t kind,
     // minor accuracy drops observed compared to f32
     const float trh = trh_f32;
 #else
-    const bool is_strict_acc
-            = prb->attr.acc_mode == dnnl_accumulation_mode_strict;
+    const bool is_strict_acc = !is_nvidia_gpu()
```
mgouicem, line 243: why not add DNNL_SYCL_CUDA in the `#if` condition above ?

ShanoToni: Changed the way the threshold is added to use DNNL_SYCL_CUDA

mgouicem (approved):

PR #2177, opened 2024-10-21
doc: Add status icon for CI pipelines

theComputeKid:
It can be beneficial for users to know whether to expect the main branch to build as expected, especially for AArch64, as our current way of pushing directly to main can break AArch64 CI. Users can look to see the current CI status in the latest git commit, but for better visibility, we can add a badge. This also gives us the status of the last full CI run instead of the current CI run that the git commit status will give us (which may not be complete at the time of viewing, and can hide the current CI status). Also recall that AArch64 CI does not fully complete on forks and so they will never show the correct CI status on the git commits, and instead will need to rely on the badge.

We added status badges in the README here: https://github.com/oneapi-src/oneDNN/commit/4d82471cf15e721a69a0e5a63f5996ec0b9ddda4

This change adds another badge for CI. An example view of badge can be seen from my fork's README: https://github.com/theComputeKid/oneDNN/tree/badges

I don't have the privileges to get the x86 CI badges from Azure, so will wait for @vpirogov or @mgouicem, in case they want to add x86 badges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ oneAPI Deep Neural Network Library (oneDNN)
 
 [![OpenSSF Best Practices](https://www.bestpractices.dev/projects/8762/badge)](https://www.bestpractices.dev/projects/8762)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/oneapi-src/oneDNN/badge)](https://securityscorecards.dev/viewer/?uri=github.com/oneapi-src/oneDNN)
+![AArch64 Build Status](https://github.com/oneapi-src/oneDNN/actions/workflows/ci-aarch64.yml/badge.svg?branch=main)
 
 oneAPI Deep Neural Network Library (oneDNN) is an open-source cross-platform
 performance library of basic building blocks for deep learning applications.


ranukund (approved):
Looks good!

vpirogov:
The issue I have with CI badges is that they show current status for the main development branch, but once added to README.md the same badges will be displayed for release branches and in forks, which is misleading. OpenSSF badges are focused on project practices, which are not as transient as build health.

Also Azure badge is horrible :(

[Build Status](https://dev.azure.com/mkldnn/dnnl/_build/latest?definitionId=7&branchName=main)

theComputeKid:
Can we can label them to indicate that they are for main? I was thinking more from the perspective of showing off github aarch64 runners. Is there any way we can make it work? Otherwise, no issue, we can drop it.

vpirogov:
I'd suggest we find a place outside of the code tree to host CI status. We can stick these into Github wiki for example. Will that work?

vpirogov:
How does [this](https://github.com/oneapi-src/oneDNN/wiki/CI-Status) look?

theComputeKid:
@vpirogov it looks good, we can do that.

btw, the aarch64 pipelines will never be successful on forks because only oneDNN + PRs have access to aarch64 runners.

vpirogov:
CI status badges for main and rls-v3.6 branches are now available [in wiki](https://github.com/oneapi-src/oneDNN/wiki/CI-Status).

PR #2195, opened 2024-10-31
[risc-v] cmake: added rvv intrinsics compilation check

dmitry-gorokhov:
# Description

OneDNN build with https://github.com/riscv-collab/riscv-gnu-toolchain fails with the following error:
image

The [GNU toolchain for RISC-V](https://github.com/riscv-collab/riscv-gnu-toolchain) now supports RVV by default - ratified RVV1.0. All intrinsics have prefix __riscv64. Some compilers (clang, llvm) can compile a code contained intrinsics without this prefix (have overloaded functions). But GCC from riscv-gnu-toolchain doesn't have them - so cannot compile max pooling primitive from oneDNN.

The fix adds cmake configuration time check that corresponding intrinsics are really supported by the compiler.

This change is upstreamed from [oneDNN fork repo](https://github.com/openvinotoolkit/oneDNN) maintained by [OpenVINO organization](https://github.com/openvinotoolkit).

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -438,7 +438,10 @@ if (DNNL_TARGET_ARCH STREQUAL "RV64")
     # Check if the RVV Intrinsics can be compiled with the current toolchain and flags
     include(CheckCXXSourceCompiles)
     check_cxx_source_compiles("#include <riscv_vector.h>
-                               int main() { return 0; };"
+                               int main() { 
+                                size_t size = 64;
+                                return vsetvl_e32m2(size); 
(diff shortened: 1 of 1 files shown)


vpirogov (approved):

mgouicem:
The workaround makes sense, thanks @dmitry-gorokhov.

Tagging @pazamelin who originally contributed this implementation. Would it make sense to just rely on the __riscv64 prefixed versions of the intrinsics to work with all compilers?

mgouicem (approved):

PR #2202, opened 2024-11-07
gpu: nvidia: skip unsupported gtests

sgeor255:
# Description

This PR adds skips in gtests for unimplemented primitives on nvidia.

### DIFF
--- a/tests/gtests/sycl/api/CMakeLists.txt
+++ b/tests/gtests/sycl/api/CMakeLists.txt
@@ -24,12 +24,12 @@ register_exe(${TEST_EXE} "${TEST_SOURCES}" "test" "dnnl_gtest")
 # so they need to be compiled with the correct device triple
 if(DNNL_WITH_SYCL)
     if(DNNL_SYCL_GENERIC)
-        CHECK_CXX_COMPILER_FLAG("-fsycl -fsycl-targets=nvptx64-nvidia-cuda" NVIDIA_TARGET_SUPPORTED)
+        CHECK_CXX_COMPILER_FLAG("-fsycl -fsycl-targets=nvptx64-nvidia-cuda,spir64" NVIDIA_TARGET_SUPPORTED)
     endif()
 
     # Enable linking SYCL kernels.
     if(DNNL_SYCL_CUDA OR (DNNL_SYCL_GENERIC AND NVIDIA_TARGET_SUPPORTED))
-        append(CMAKE_CXX_FLAGS "-fsycl-targets=nvptx64-nvidia-cuda")
+        append(CMAKE_CXX_FLAGS "-fsycl-targets=nvptx64-nvidia-cuda,spir64")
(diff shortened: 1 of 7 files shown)


tests/gtests/test_iface_attr_quantization.cpp
```diff
@@ -431,7 +431,12 @@ TEST_F(attr_quantization_test_t, TestLRN) {
     }
 }
 
+#define DEBUG_PRINT_LINE std::cout << "reached line " << __LINE__ << "\n";
```
ShanoToni, line 434: Is this needed?

ShanoToni (approved):

tests/gtests/test_iface_attr_quantization.cpp
sgeor255: It was left over from debugging, removed it.

tests/gtests/sycl/api/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ if(DNNL_WITH_SYCL)
     endif()
 
     # Enable linking SYCL kernels.
-    if(DNNL_SYCL_CUDA OR (DNNL_SYCL_GENERIC AND NVIDIA_TARGET_SUPPORTED))
+    if(DNNL_SYCL_GENERIC AND NVIDIA_TARGET_SUPPORTED)
```
dzarukin, line 31: Could you explain, please, why `DNNL_SYCL_CUDA` is removed?

sgeor255: The tests `Simple/sycl_memory_buffer_test.InteropReorderAndUserKernel/cpu` and `Simple/sycl_memory_buffer_test.EltwiseWithUserKernel/cpu` run with a CPU engine and try to run a SYCL kernel on CPU. They currently fail because the nvidia target triple is set. This results in a `UR_ERROR_INVALID_BINARY`. I am not sure if this is intended behaviour or the tests should be skipped, but removing the `DNNL_SYCL_CUDA` from the check here allows them to run successfully on CPU as it results in not setting the nvidia target triple.

dzarukin: What you are saying is that target shouldn't be set for CPU part of the test. You removed linker options for pure Nvidia target, but how this change would help for a generic case targeting Nvidia backend?

densamoilov: @sgeor255, I think the problem arises because due to overriding the default SYCL targets with `-fsycl-targets=nvptx64-nvidia-cuda` which throws off the default SYCL target, which is `spir64`. We could try to use both targets here as we expect the kernels to be compiled for CPU as well:
`-fsycl-targets=nvptx64-nvidia-cuda,spir64`

densamoilov (approved):

PR #2208, opened 2024-11-07
test: aarch64: Fix unimplemented error when --cpu-isa-hints=prefer_ymm

Ryo-not-rio:
When running with --cpu-isa-hints=prefer_ymm, tests fail on aarch64 due to an unimplmented error. This change will instead allow the code to run the reference implementations.

Reproducible:
```
./tests/benchdnn/benchdnn --conv --cpu-isa-hints=prefer_ymm --dt=bf16:bf16:bf16 --stag=axb --dtag=axb --attr-post-ops=sum+tanh:1:1 g1ic64ih1000oc64oh1000kh3ph128dh127
```

### DIFF
--- a/src/cpu/platform.cpp
+++ b/src/cpu/platform.cpp
@@ -82,6 +82,8 @@ status_t set_max_cpu_isa(dnnl_cpu_isa_t isa) {
 status_t set_cpu_isa_hints(dnnl_cpu_isa_hints_t isa_hints) {
 #if DNNL_X64
     return x64::set_cpu_isa_hints(isa_hints);
+#elif DNNL_AARCH64
+    return status::success;
 #else
     return status::unimplemented;
 #endif


tests/benchdnn/dnnl_common.cpp
```diff
@@ -272,7 +272,7 @@ void init_isa_settings() {
     if (hints.get() == isa_hints_t::no_hints)
         DNN_SAFE_V(dnnl_set_cpu_isa_hints(dnnl_cpu_isa_no_hints));
     else if (hints.get() == isa_hints_t::prefer_ymm)
-        DNN_SAFE_V(dnnl_set_cpu_isa_hints(dnnl_cpu_isa_prefer_ymm));
+        dnnl_set_cpu_isa_hints(dnnl_cpu_isa_prefer_ymm);
```
dzarukin, line 275: I feel that updating [this function](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/platform.cpp#L82) makes more sense. Aarch64 is fine to return `success` unconditionally.

mgouicem: +1 on Dmitri's solution as it is the simplest solution.

Down the line, it might also make sense to guard the platform specific isa/isa_hint enum values in the headers. Currently, DNNL_X64 is not exposed there but we could add it to dnnl_config.h (similar to what we do for experimental features).

theComputeKid:
Could you reformat the commit message to: `test: aarch64: Fix unimplemented error when --cpu-isa-hints=prefer_ymm`

The PR checks should have caught this but they are probably not working as intended.

tests/benchdnn/dnnl_common.cpp
Ryo-not-rio: Thank you for the review, I have update the code as suggested

dzarukin (approved):

PR #2209, opened 2024-11-08
cpu: aarch64: non-global stats mode for ACL lnorm

michalowski-arm:
* Change acl_layer_normalization to support non-global stats only
  * Remove mean and variance check for ACL lnorm from benchdnn


theComputeKid:
Could you please format the PR and the commit:
```
cpu: aarch64: non-global stats mode for ACL lnorm
```

theComputeKid:
We also need to remove skipped tests from GitHub actions from here:
https://github.com/oneapi-src/oneDNN/blob/74e7e4d3bb8c9f839fc5b44b7c8f8da0393161e3/.github/automation/test_aarch64.sh#L74

and in other places. Then if the CI passes, we are good.

PR #2212, opened 2024-11-08
src: cpu: aarch64: lowp_matmul_sq: Make weights constant

fadara01:
# Description

Setting the weights as constant allows us to avoid redundant pretranspose operations in Arm Compute Library (ACL) every time execute is called (they are now run once and cached). This delives big speedups especially for relatively small matmuls.
Note that this is a temp fix that needs to be handled carefully by primitive caches in frameworks, since the ACL object is now holding more state - i.e. we want to make sure that the cahce maps a layer with a specific set of weights to the oneDNN primitive storing those weights.
We're currently working on the proper fix for this which involves making lowp_gemm stateless and fixed-format in ACL and oneDNN.

Fixes # (github issue)

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes

## [RFC](https://github.com/oneapi-src/oneDNN/tree/rfcs) PR

- [ ] Does RFC document follow the [template](https://github.com/oneapi-src/oneDNN/blob/rfcs/rfcs/template.md#onednn-design-document-rfc)?

### DIFF
--- a/src/cpu/aarch64/matmul/acl_lowp_matmul_sq.hpp
+++ b/src/cpu/aarch64/matmul/acl_lowp_matmul_sq.hpp
@@ -1,0 +1,326 @@
+/*******************************************************************************
+* Copyright 2024 Arm Ltd. and affiliates
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef ACL_LOWP_MATMUL_SQ_HPP
+#define ACL_LOWP_MATMUL_SQ_HPP
+
+#include <random>
+
+#include "cpu/cpu_primitive.hpp"
+#include "cpu/matmul/cpu_matmul_pd.hpp"
+#include "cpu/matmul/matmul_utils.hpp"
+
+#include "arm_compute/core/utils/quantization/AsymmHelpers.h"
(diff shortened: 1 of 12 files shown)


src/cpu/aarch64/matmul/acl_lowp_matmul_sq.hpp
```diff
@@ -0,0 +1,326 @@
...
+                    arm_compute::TensorShape(K(), M()), 1,
+                    acl_utils::get_acl_data_t(src_d.data_type(), true),
+                    arm_compute::QuantizationInfo(1.0, 0, true));
+            almc_.src_tensor_info.set_are_values_constant(false);
+
+            almc_.wei_tensor_info = arm_compute::TensorInfo(
+                    arm_compute::TensorShape(N(), K()), 1,
+                    acl_utils::get_acl_data_t(wei_d.data_type(), true),
+                    arm_compute::QuantizationInfo(1.0, 0, true));
+            almc_.wei_tensor_info.set_are_values_constant(true);
```
mgouicem, line 159: Hi @fadara01 , 
What does constant mean here? Is it about constant values within the tensor or just that the shape is constant?
If it is about constant values in weights, it does not seem that primitive creation takes the weights, does it mean the primitive state changes upon first execution?

src/cpu/aarch64/matmul/acl_lowp_matmul_sq.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+        CHECK(r->configure(pd()->almc_));
+
+        mapper.add(this, std::move(r));
+
+        return status::success;
+    }
+
+    status_t execute(const exec_ctx_t &ctx) const {
+        std::lock_guard<std::mutex> _lock {this->mtx};
```
mgouicem, line 246: Is the mutex protection only about the quantization parameters that are reset, or is there something else due to constant weights?

mgouicem (commented):
One general note here, oneDNN primitive API has no concept of constant memory object, so such an optimization is unsafe (there is no guarantee from primitive implementation perspective that weights will not change between execute calls).

oneDNN Graph API has that concept though, and it might make sense to add it to primitive API if it helps. But I would hold off commiting this patch unless we extend API with constant memory objects.

openmodelzooci:

PR #2214, opened 2024-11-14
generic: sycl: Inner Product FWD

AD2605:
# Description

Implements the forward pass of the inner product for the generic sycl backend with support for different post-ops and data types

### DIFF
--- a/src/gpu/generic/sycl/README.md
+++ b/src/gpu/generic/sycl/README.md
@@ -93,6 +93,14 @@ The implementation supports both forward and backward directions.
 `pow`, `relu`, `round`, `soft_relu`, `sqrt`, `square`,`swish` and `tanh`
 * Supported formats: `NCDHW`, `NDHWC`, `NCHW`, `NHWC`, `NCW`, `NWC`, `NC`, `N`
 * Supported data types: `f32`, `bf16`, `f16`, `s32`, `s8`, `u8`
+  
+## Inner Product
+
+The implementation supports the forward direction only.
+
+* Supported formats: All plain formats are supported.
+* Supported data types: All possible data combinations listed in the oneDNN specification are supported.
+* Supported post-ops: All the post operations as mentioned in the specification are supported.
 
 ## Layer Normalization
 
--- a/src/gpu/generic/sycl/ref_inner_product.hpp
+++ b/src/gpu/generic/sycl/ref_inner_product.hpp
@@ -1,0 +1,154 @@
+/*******************************************************************************
+* Copyright 2023-2024 Intel Corporation
+* Copyright 2024-2025 Codeplay Software Limited
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef GPU_GENERIC_SYCL_REF_INNER_PRODUCT_HPP
+#define GPU_GENERIC_SYCL_REF_INNER_PRODUCT_HPP
+
+#include "gpu/generic/sycl/ref_matmul.hpp"
+#include "gpu/generic/sycl/sycl_gpu_primitive.hpp"
+#include "gpu/generic/sycl/sycl_post_ops.hpp"
+#include "gpu/generic/sycl/sycl_primitive_conf.hpp"
+#include "gpu/generic/sycl/sycl_utils.hpp"
+#include "gpu/gpu_inner_product_pd.hpp"
+#include "gpu/gpu_primitive.hpp"
+
+namespace dnnl::impl::gpu::generic::sycl {
+struct ref_inner_product_fwd_t : public gpu::generic::sycl::primitive_t {
+    using gpu::generic::sycl::primitive_t::primitive_t;
+
+    struct pd_t : public gpu_inner_product_fwd_pd_t {
+        using gpu_inner_product_fwd_pd_t::gpu_inner_product_fwd_pd_t;
+        using sm = primitive_attr_t::skip_mask_t;
+
+        DECLARE_COMMON_PD_T("dpcpp:ref:any", ref_inner_product_fwd_t);
+
+        status_t init(impl::engine_t *engine) {
+            auto src_dt = arg_md(DNNL_ARG_SRC)->data_type;
+            auto weights_dt = arg_md(DNNL_ARG_WEIGHTS)->data_type;
+            auto dst_dt = arg_md(DNNL_ARG_DST)->data_type;
+            auto bias_dt = with_bias() ? arg_md(DNNL_ARG_BIAS)->data_type
+                                       : data_type::undef;
+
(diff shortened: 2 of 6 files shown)


src/gpu/nvidia/cudnn_matmul_executor.hpp
```diff
@@ -392,16 +392,16 @@ struct cudnn_matmul_lt_exec_t final : public cudnn_matmul_lt_base_exec_t {
                             memory_tracking::names::key_matmul_dst_in_acc_dt)
                     : xpu::sycl::interop_memory_arg_t<
                             ::sycl::access::mode::read_write>();
-            auto arg_block_a_scratch = params->source_size_ != 0
```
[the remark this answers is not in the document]
AD2605: Yes.

>  If so, should we consider adding more tests to catch them?

From what I can tell, this file should be covering all the possible combinations, and I do see the case for which this was failing (i.e A = S8, ab format, B = S8, ba format, and C = S32, ab format), but maybe others with a better understanding could chime in, maybe @ShanoToni ?

src/gpu/nvidia/cudnn_matmul_lt_impl.hpp
```diff
@@ -717,7 +717,7 @@ struct cudnn_matmul_lt_impl_t {
             }
             if (!params->w_blocked_) {
                 transform_matrix(lt_handle, params, a_layout, a,
-                        blocked_a_layout, block_a_scratch, !params->trans_a_,
+                        blocked_a_layout, block_a_scratch, params->trans_a_,
```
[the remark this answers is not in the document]
AD2605: Yes,  this was one of the segmentation faults in the s8:s8:s32 case. 
and from what I can see, after making this change, it did not break any additional tests when compared against main.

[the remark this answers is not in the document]
Rbiessy: Ok, I was wondering if we should add test cases to catch this bug using matmul only but I suppose the inner product tests can catch it now.

sgeor255:
@AD2605  you should probably update https://github.com/oneapi-src/oneDNN/blob/main/src/gpu/generic/sycl/README.md to mention inner product is supported :)

AD2605:
@sgeor255 , updated the documentation

AD2605:
@oneapi-src/onednn-arch @oneapi-src/onednn-doc gentle nudge for reviewing the PR

src/gpu/generic/sycl/README.md
```diff
@@ -93,6 +93,14 @@ The implementation supports both forward and backward directions.
...
 * Supported formats: `NCDHW`, `NDHWC`, `NCHW`, `NHWC`, `NCW`, `NWC`, `NC`, `N`
 * Supported data types: `f32`, `bf16`, `f16`, `s32`, `s8`, `u8`
+  
+## Inner Product
+
+The implementation supports the forward direction only.
+
+* Supported formats: All plain formats are supported.
+* Supported data types: All possible data combinations as listed in the specification are supported.
+* Supported post-ops: All the post-ops as mentioned in the specification are supported.
```
ranukund, line 103:
```suggestion
* Supported post-ops: All the post operations as mentioned in the specification are supported.
```

ranukund (approved):
One minor comment added. Please incorporate as you see fit.

src/gpu/generic/sycl/README.md
AD2605: I have a slight preference towards the phrase `post-ops`, as the oneDNN specification too uses `post-ops` rather than post operations

ranukund: > I have a slight preference towards the phrase `post-ops`, as the oneDNN specification too uses `post-ops` rather than post operations

The general writing guidelines advise spelling out and expand the contracted terms. If this is a specialized term, suggest including it in code blocks for example, `post-ops` or italicize the term _post-ops_. In addition, we can call out that we are referring to the oneDNN specification for context. 

Supported _post-ops_: All the _post-ops_ mentioned in the oneDNN specification are supported.

src/gpu/generic/sycl/README.md
```diff
@@ -93,6 +93,14 @@ The implementation supports both forward and backward directions.
 `pow`, `relu`, `round`, `soft_relu`, `sqrt`, `square`,`swish` and `tanh`
 * Supported formats: `NCDHW`, `NDHWC`, `NCHW`, `NHWC`, `NCW`, `NWC`, `NC`, `N`
 * Supported data types: `f32`, `bf16`, `f16`, `s32`, `s8`, `u8`
+  
+## Inner Product
+
+The implementation supports the forward direction only.
+
+* Supported formats: All plain formats are supported.
+* Supported data types: All possible data combinations as listed in the specification are supported.
```
ranukund, line 102:
```suggestion
* Supported data types: All possible data combinations listed in the oneDNN specification are supported.
```

src/gpu/nvidia/cudnn_matmul_executor.hpp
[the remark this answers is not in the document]
ShanoToni: This is an oversight on the original PR, caused by differences in matmuls (oneDNN is col major, cuDNN is row major so we do BxA instead of AxB) the `arg_block_a_scratch` should be mapped to weights to be correct, so in the case where we need scratchpad memory only for weights and not source we would get correctness errors.

This is difficult to add more tests as cublasLT would only run when other implementations do not support it (cublas supports most already), additionally I am not sure what would the process for expanding the tests being run.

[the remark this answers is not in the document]
Rbiessy: Thanks for the explanation. I suppose it's fine then.

Rbiessy (approved):

src/gpu/generic/sycl/ref_inner_product.hpp
```diff
@@ -0,0 +1,154 @@
...
+            const bool ok = (set_default_params() == status::success)
+                    && is_fwd()
+                    && check_if_dtypes_valid(
+                            src_dt, dst_dt, bias_dt, weights_dt)
+                    && sycl_post_ops_t::post_ops_ok(attr())
+                    && (attr_.set_default_formats(dst_md()) == status::success)
+                    && attr()->has_default_values(sm::scales_runtime
+                            | sm::zero_points_runtime | sm::post_ops
+                            | sm::dropout | sm::scales_runtime_data_type
+                            | sm::zero_points_runtime_data_type)
```
mgouicem, line 55: why not defer attribute and datatype checking to init_matmul?

src/gpu/generic/sycl/ref_inner_product.hpp
```diff
@@ -0,0 +1,154 @@
...
+                    && memory_desc_wrapper(src_md()).is_plain()
+                    && memory_desc_wrapper(dst_md())
+                               .is_plain(); // Blocked memory formats are not supported
```
mgouicem, line 58: I think is_plain is not enough to flatten the dimensions.
Here you need to check:
- that strides are dense (see is_dense).
- that the order of flattened dimensions for src and weights match.

sgeor255: That's a good point, thanks. I had to create a new PR (https://github.com/oneapi-src/oneDNN/pull/2248) to address the comments as I don't have permissions to push to this one. I think we can close this one.

src/gpu/generic/sycl/ref_inner_product.hpp
sgeor255: Done in https://github.com/oneapi-src/oneDNN/pull/2248, thanks.

PR #2215, opened 2024-11-14
cpu: aarch64: non-global stats mode for ACL lnorm

michalowski-arm:
* Change acl_layer_normalization to support non-global stats only
  * Remove mean and variance check for ACL lnorm from benchdnn

### DIFF
--- a/src/cpu/aarch64/acl_layer_normalization.hpp
+++ b/src/cpu/aarch64/acl_layer_normalization.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2023 Arm Ltd. and affiliates
+* Copyright 2023-2024 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/cpu/aarch64/acl_layer_normalization.hpp
```diff
@@ -81,9 +81,8 @@ struct acl_layer_normalization_fwd_t : public primitive_t {
             // dir and flags
             ACL_CHECK_SUPPORT(
                     !is_fwd(), "ACL lnorm supports forward propagation only");
-            ACL_CHECK_SUPPORT(is_training() && !use_global_stats(),
-                    "ACL only supports forward training with lnorm if stats "
-                    "are provided (use global stats)");
+            ACL_CHECK_SUPPORT(use_global_stats(),
+                    "ACL does not support global stats with lnorm");
```
dzarukin, line 85: Does it mean that that ACL would compute mean and variance internally but wouldn't return them to the user?

michalowski-arm: That's correct

dzarukin: Any chance ACL will update their functionality to start returning mean and var to be aligned with oneDNN?

dzarukin: As an additional note, it seems that supporting only prop_kind::inference (and dropping prop_kind::training from supported) is a cleaner way to indicate that mean and variance are not going to be returned to the user.
@mgouicem, would you agree with this statement?

theComputeKid: @dzarukin that is a more involved change, I don't think it will happen any time soon.

dzarukin: > @dzarukin that is a more involved change, I don't think it will happen any time soon, but it is in our backlog.

Sounds good, then please proceed with supporting inference kind only, that would be more clear from user perspective and validation points.

theComputeKid (approved):

michalowski-arm:
CC @mgouicem

mgouicem (approved):

PR #2227, opened 2024-11-26
ngen: generate basic DWARF debug info sections in zebin output

bkammerd:
# Description

## Background
My team at Intel and I are creating an AI profiling tool that can generate flame graphs of AI workload performance based on an execution unit stall sampling feature on recent Intel GPUs ([link to blog post](https://brendangregg.com/blog/2024-10-29/ai-flame-graphs.html)). A crucial part of this tool is its ability to relate GPU kernels back to the source code defining them. We found that some workloads use oneDNN kernels that are generated at runtime through the nGEN library and that these kernels lack any sort of debug information in their ZEBin binaries. While this makes sense (there _is no kernel source_ to refer to in debug information), it limits the potential of existing tools like the one we're developing as well as future AI debugging/observability efforts.

## The Change
This PR includes a minimal amount of DWARF 5 debug information in the already generated ZEBin for a given kernel. The debug info simply defines a single compile unit and a single function with it (the kernel). The source code location of the kernel is defined as the source line that generates it within nGEN. The primary purpose of this is so that developers can view the output of a tool like ours and relate performance of a GPU kernel to the actual origin of the kernel. The following are some screenshots of our tool before (left) and after this patch (right) for `benchdnn`:

Screenshot_20241126_130154

# Checklist

## General

### DIFF
--- a/src/gpu/intel/jit/ngen/ngen_elf.hpp
+++ b/src/gpu/intel/jit/ngen/ngen_elf.hpp
@@ -126,6 +126,37 @@ private:
             MachineIntelGT = 205,
             ZebinExec = 0xFF12
         };
+
+        enum DWARF_UT : uint8_t {
+            COMPILE = 0x01,
+        };
+        enum DWARF_TAG : uint8_t {
+            COMPILATION_UNIT = 0x11,
+            SUBPROGRAM       = 0x2e,
+        };
+        enum DWARF_AT : uint8_t {
+            NAME        = 0x03,
+            STMT_LIST   = 0x10,
+            DECL_COLUMN = 0x39,
+            DECL_FILE   = 0x3a,
+            DECL_LINE   = 0x3b,
+        };
+        enum DWARF_FORM : uint8_t {
+            DATA2        = 0x05,
+            DATA4        = 0x06,
+            DATA8        = 0x07,
+            STRING       = 0x08,
+            DATA1        = 0x0b,
+            STRP         = 0x0e,
+            LINEPTR      = 0x17,
+            FLAG_PRESENT = 0x19,
+            LINE_STRP    = 0x1f,
+        };
+        enum DWARF_LNCT : uint8_t {
+            PATH            = 0x1,
+            DIRECTORY_INDEX = 0x2,
+        };
+
         union TargetMetadata {
             uint32_t all;
             struct {
@@ -173,7 +204,7 @@ private:
             uint32_t info = 0;
             uint64_t align = 0x10;
             uint64_t entrySize = 0;
-        } sectionHeaders[5];
+        } sectionHeaders[10];
         struct Note {
             uint32_t nameSize = 8;
             uint32_t descSize = 4;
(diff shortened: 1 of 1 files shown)


petercad:
Hi @bkammerd -- thanks for the PR! I like the overall idea of adding minimal debug information, and would be glad to help with that.

My main concern is that I don't know if it's been standardized anywhere how debug information should be embedded in zebin. I don't want us to get in a situation where oneDNN produces zebin that will be rejected by future zebin consumers (runtime, debuggers, etc.). The zebin spec [mentions](https://github.com/intel/intel-graphics-compiler/blob/ed93602c20e7221ac469cdf213e22b32b1804557/IGC/ZEBinWriter/zebin/spec/elf.md) debug information in passing but doesn't specify formats, versioning, etc.

I'd also like to enable debug information only when needed/requested by the user. For your tooling, do you rely on the user compiling e.g. SYCL device kernels with `-g` or are you using some other means to force IGC to compile with debug information? I'd like to understand how a user should let us know they need debug info.

bkammerd:
Thanks for the reply, @petercad. Yes you're correct that we've typically been relying on the kernel compiler stack (`-g`) to introduce debug info for kernels, which has worked fine for us so far. nGEN is a unique case we've run into. I chose to implement DWARF in the zebin format simply because that's what we've seen in other kernels compiled with debug information. That said, I'm open to other ideas that might make more sense w.r.t standardization.

petercad:
@bkammerd All right then, so actually you'll have issues with any oneDNN kernel, not just nGEN kernels. Our non-nGEN kernels are written in OpenCL, so SYCL debug flags don't affect them either. For that, we'd need to add `-g` to the OCL build options, but you still wouldn't get the file information you need for your tool, since the kernel source is a runtime-prepared string, not a file.

It seems to me you'll want some other side-channel by which oneDNN could convey to you location information with each kernel launch. As a side note, the ability to annotate kernel launches with information for tools to analyze would be useful for other reasons (e.g. information on tensor dimensions, data types, etc., could be attached to a kernel launch). Also, compiling in debug mode will significantly reduce performance for compiled kernels, and I'm thinking you won't want that (huge) performance hit for a performance debug tool.

Let me know if this seems reasonable; if so, we can think about next-level details.

bkammerd:
@petercad Thanks for the feedback. We're hesitant to dive into a new bespoke mechanism to get this information when it already works reliably for many of the libraries and frameworks we've tested. It seems to me like we'd want to adhere to the precedent of using embedded debug information to link executable code to source just as we do with CPU code. I see two major roadblocks: 1. Debug information is incorrectly married to other advanced debugging features and code generation options and 2. many kernels don't have a traditional "file on disc" source origin as you point out.

The first issue is the more difficult and will likely take collaboration from other projects either way. I think ideally for us, we would find a way to segregate DWARF generation from everything else that the `-g` flag currently does in GPU compiler stacks to more closely match the behavior of CPU compilers. Then perhaps we could agree on a standard way for a user to tell kernel producers to emit debug info with their kernels. This could be as simple as an environment variable. Thoughts?

With this patch I've dealt with 2 simply by pointing to the code responsible for generating the kernel. This may or may not be sufficient information depending on the use case. It seems like a reasonable choice so far for us, as we simply want to provide engineers with a full path through the stacks that ultimately launch a kernel. Improvements here would certainly be welcome however.

rjoursler:
@bkammerd , what kind of information do we need collected for this tool to perform useful analysis? I am curious how this is supposed to interact with the details for what oneDNN executes.

For example, the oneDNN library generates many different specialized kernels, with the same name, on demand  based on input data types, problem dimensions, post-operation fusions, etc. Should these different specializations be registered as separate kernels in this tool?

bkammerd:
@rjoursler, since this tool allows the user to visualize the performance of each individual kernel, the most important goal is that the kernels themselves are easily identifiable by the user. Typically, the kernel name plus the call stack leading up to its launch would be sufficient context. However, as you point out, oneDNN tends to generate different kernels with the same name, making this slightly less useful. I think that embedding some of the kernel parameters into the name could be a simple and useful change for this tool in its current state, but providing more comprehensive information about the kernel in the packaged debug information could be more impactful for analysis tools going forward.

petercad:
> We're hesitant to dive into a new bespoke mechanism to get this information when it already works reliably for many of the libraries and frameworks we've tested. It seems to me like we'd want to adhere to the precedent of using embedded debug information to link executable code to source just as we do with CPU code

Understood. But you will need a new mechanism of some kind, because oneDNN (along with other libraries/frameworks using OpenCL source compilation -- oneMKL, OpenVINO) has no way of telling OpenCL to embed this kind of debug information in built kernels. For our nGEN kernels, we can indeed embed it directly since we create the program binary from scratch. And the issue isn't really OpenCL or even GPU-specific, but applies to JIT'ed code in general. On CPU, as I recall, oneDNN has a sideband for providing this kind of information via ITT hooks.

bkammerd:
@petercad, I hear you. I think that we would personally prefer a solution that looks more like adding options to the relevant APIs such as OCL and friends that would allow this information to be relayed to which ever layer eventually compiles the kernel to a zebin so that it can be added as DWARF. Then we'd just need to address edge cases like I've done in this PR. You are correct though that JIT kernels in general make this harder. Where we have been successful, even with just-in-time compiled kernels is when they ship as SPIR-V, which can carry along the debug info along with it IIUC.

src/gpu/intel/jit/ngen/ngen_elf.hpp
```diff
@@ -183,26 +214,144 @@ class ELFCodeGenerator : public BinaryCodeGenerator<hw>
...
+                        uint32_t fileName0 = 0;
+                        uint8_t dirIndex0 = 0;
+                        uint32_t fileName1 = 0;
+                        uint8_t dirIndex1 = 0;
+                } header;
+        } debugLine;
+
+        struct {
+                const char fileName[strlen(__FILE__) + 1] = __FILE__;
+                const char dirName[strlen("oneDNN:") + 1] = "oneDNN:";
```
rjoursler, line 319: Indentation within this file should use 4 spaces.

src/gpu/intel/jit/ngen/ngen_elf.hpp
```diff
@@ -358,7 +531,7 @@ std::vector<uint8_t> ELFCodeGenerator<hw>::getBinary(const std::vector<uint8_t>
 
     binary.resize(paddedSzELF + paddedSzMetadata + paddedSzKernel);
 
-    (void) new(binary.data()) ZebinELF(paddedSzKernelName, metadata.size(), kernel.size());
+    (void) new(binary.data()) ZebinELF(paddedSzKernelName, metadata.size(), kernel.size(), __LINE__);
```
rjoursler, line 534: The debug information should come from the implementation. In particular, it will need passed into [make_kernel](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/2da2a65552814a455b0c2735ae24a94d39c9c6de/src/gpu/intel/jit/jit_generator.hpp#L188) and forwarded down to this function call from locations like: [gen_pooling](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/2da2a65552814a455b0c2735ae24a94d39c9c6de/src/gpu/intel/jit/pooling/gen_pooling.cpp#L150-L151), [gen_conv](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/2da2a65552814a455b0c2735ae24a94d39c9c6de/src/gpu/intel/jit/conv/gen_convolution.cpp#L206-L246), [gen_reorder](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/2da2a65552814a455b0c2735ae24a94d39c9c6de/src/gpu/intel/jit/reorder/gen_reorder.cpp#L208-L209), and also into the [gen_gemm_kernels calls to getBinary](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/2da2a65552814a455b0c2735ae24a94d39c9c6de/src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp#L895-L914)

src/gpu/intel/jit/ngen/ngen_elf.hpp
bkammerd: Fixed, thanks.

src/gpu/intel/jit/ngen/ngen_elf.hpp
rjoursler: I am suggesting the latter option, that the "location responsible for building the kernel" is set further up the stack. The core issue with the current method is that very different implementations (such as GEMM and pooling) will point to the same source code location. By pointing to the implementation generator, the debug info will be much more relevant.

> So are you suggesting that we add more comprehensive debug information from higher up the stack? If so, what would you include?

I think this is a good idea, for example, I am thinking we should add an asm->source code line mapping so that debugging is simpler within our team. This likely requires invasive changes, so I expect I will be handling it. Is there is any additional information that would be useful to collect from your perspective?

bkammerd: Okay that's clear, thanks. More information is always better than less. Our tool in its current form doesn't do anything that would require more debug info than what I've added here, but I can certainly imagine enhancements that indicate different properties of the kernel in the flame graph visualization. For these we might need to package some more parameters with the debug info such as the data type.

I also agree that mapping instructions back to source lines would be very useful. Our tool currently shows stall counts for individual kernel instructions, so it would be nice to add functionality that let's you zoom in on the kernels source and see a visualization of EU stalls per line.

rjoursler:
Closing as this was wrapped into (and extended by) #2297.

PR #2235, opened 2024-11-29
fix, build, docs: aarch64: mutex lock if the ACL kernel is not stateless

Sqvid:
# Description
Some of the underlying kernels used by ACL still require some state and are not safe to be called in parallel with different execution contexts. Eventually when all kernels are truly stateless, this guard can be removed. This feature depends on ACL >= 24.11.1 so the build system and docs have been updated to reflect that.

# Checklist

## General

### DIFF
--- a/.github/automation/build_acl.sh
+++ b/.github/automation/build_acl.sh
@@ -28,7 +28,7 @@ source ${SCRIPT_DIR}/common_aarch64.sh
 
 ACL_CONFIG=${ACL_CONFIG:-"Release"}
 ACL_ROOT_DIR=${ACL_ROOT_DIR:-"${PWD}/ComputeLibrary"}
-ACL_VERSION=${ACL_VERSION:-v24.09}
+ACL_VERSION=${ACL_VERSION:-v24.11.1}
 ACL_ARCH=${ACL_ARCH:-"armv8.2-a"}
 ACL_REPO="https://github.com/ARM-software/ComputeLibrary.git"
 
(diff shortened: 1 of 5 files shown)


Sqvid:
CI fails as expected for now because we are waiting for the new ACL release to become publicly available. Should I bump the ACL CI version here too?

Sqvid:
> Should I bump the ACL CI version here too?

Done.

mgouicem (approved):

mgouicem (commented):
It seems 24.11.1 is not out yet (see [log](https://github.com/oneapi-src/oneDNN/actions/runs/12084838859/job/33794260285?pr=2235) and [ACL release tags](https://github.com/ARM-software/ComputeLibrary/releases)). We should hold off merging this PR until it is publicly available so that we can run a full CI on it

Sqvid:
> It seems 24.11.1 is not out yet (see [log](https://github.com/oneapi-src/oneDNN/actions/runs/12084838859/job/33794260285?pr=2235) and [ACL release tags](https://github.com/ARM-software/ComputeLibrary/releases)). We should hold off merging this PR until it is publicly available so that we can run a full CI on it. Do you know when the public release is expected?

Yes indeed. We are expecting the next ACL release in the next 24-48 hours and then I will retrigger the aarch64 pipelines.
In the meantime if the doc team could review the changes that would be much apprecieated

ranukund (approved):
Looks good, thanks!

Radu2k:
All the tests pass. This patch shall be good to go in. 
Thanks.

PR #2239, opened 2024-12-02
supporting the amax operator

jinz2014:
I am not sure when you will add the operator.  Currently, the CPU implementation is missing in the PR, 

src/cpu/x64/jit_uni_reduction_kernel.cpp
```
        case reduction_amax:
            compute_op_ = [&](const Xbyak::Xmm &acc, const Xbyak::Xmm &to_acc) {
                // TODO abs(to_acc)
                uni_vmaxps(acc, acc, to_acc);
            };
            break;
```

I hope your expertise in oneDNN will fully support the operator with the PR.

Thanks.


src/gpu/intel/jit/ngen/ngen_pseudo.hpp
```diff
@@ -38,6 +38,18 @@ void min(const InstructionModifier &mod, const RegData &dst, const RegData &src0
...
+template <typename DT = void>
+void amax_(const InstructionModifier &mod, const RegData &dst, const RegData &src0, const RegData &src1) {
+    sel<DT>(mod | ge | f0[0], dst, src0, abs(src1));
+}
+#ifndef NGEN_WINDOWS_COMPAT
+template <typename DT = void>
+void amax(const InstructionModifier &mod, const RegData &dst, const RegData &src0, const RegData &src1) {
+    sel<DT>(mod | ge | f0[0], dst, src0, abs(src1));
+}
+#endif
```
petercad, line 51: I don't think there's any need to define an `amax` pseudo-instruction within nGEN, since it's not a widely-used operation and nGEN already provides the building blocks for it. Instead, you can keep the implementation completely within jit_reduction_injector.

src/gpu/intel/jit/jit_reduction_injector.cpp
```diff
@@ -41,6 +41,11 @@ void jit_reduction_injector_f32<hw>::sum_fwd(
     eadd(h, simd, acc, acc, val);
 }
 template <gpu_gen_t hw>
+void jit_reduction_injector_f32<hw>::amax_fwd(
+        int simd, const ngen::GRF &acc, const ngen::GRF &val) {
+    h.amax_(simd, acc, acc, val);
```
petercad, line 46: Instead of defining an amax pseudo-instruction, just write here what you'd like to do:
```suggestion
    h.max_(simd, acc, acc, abs(val));
```

petercad (changes requested):

src/gpu/intel/jit/jit_reduction_injector.cpp
jinz2014: I didn't know that it is fine to pass a variable of type ngen::GRF to abs(). Then, this will reuses the h.max_ function. Thanks.

Simonsays095 (changes requested):
I see changes to `src/gpu/intel/ocl/reduction` files, but not to `.cl` and/or `.h` files, which are used on the device side via opencl kernel compilation. Have you tested any reduction cases with an intel GPU? I suspect they will fail, as the opencl implementations will fall back to a no-op on reduction.

In this case, I think you'll need to modify `src/gpu/intel/ocl/reduction/ocl_reduction.h` to include the `REDUCTION_AMAX` algorithm (you've already defined this macro on the cpp/host side).

src/gpu/intel/jit/jit_reduction_injector.cpp
Simonsays095: Just a note: the jit_reduction_injector is currently only used in one dev-mode-only (i.e. `-DDNNL_DEV_MODE=on`) reduction implementation. If you want to test that this works, you'll need to compile in dev mode and see `src/gpu/intel/jit/jit_reduction.cpp` for the limited set of allowed cases for this implementation. Otherwise, the library currently uses opencl-based kernels exclusively.

src/gpu/intel/jit/ngen/ngen.hpp
```diff
@@ -1387,6 +1387,7 @@ template <typename... Targs> void ignoredep(Targs&&... args) { NGEN_NAMESPACE::B
 template <typename... Targs> void subdep(Targs&&... args) { NGEN_NAMESPACE::BinaryCodeGenerator<hw>::subdep(std::forward<Targs>(args)...); } \
 template <typename... Targs> void wrdep(Targs&&... args) { NGEN_NAMESPACE::BinaryCodeGenerator<hw>::wrdep(std::forward<Targs>(args)...); } \
 template <typename DT = void, typename... Targs> void min_(Targs&&... args) { NGEN_NAMESPACE::BinaryCodeGenerator<hw>::template min_<DT>(std::forward<Targs>(args)...); } \
+template <typename DT = void, typename... Targs> void amax_(Targs&&... args) { NGEN_NAMESPACE::BinaryCodeGenerator<hw>::template amax_<DT>(std::forward<Targs>(args)...); } \
```
petercad, line 1390: Thanks for the updates! Can you revert the remaining nGEN changes as well?

include/oneapi/dnnl/dnnl_graph.hpp
```diff
@@ -819,6 +819,7 @@ class op : public op_handle {
         Reciprocal = dnnl_graph_op_reciprocal,
         ReduceL1 = dnnl_graph_op_reduce_l1,
         ReduceL2 = dnnl_graph_op_reduce_l2,
+        ReduceAMax = dnnl_graph_op_reduce_amax,
```
Simonsays095, line 822: This addition causes a failure in `tests/gtests/graph/api/test_cpp_api_op.cpp:CreateAllOps`. Please add this enum value at the appropriate point in this file.

Simonsays095 (changes requested):
Please format the code and commit messages according to the [contribution guidelines](https://github.com/oneapi-src/oneDNN/blob/main/CONTRIBUTING.md) and the [coding standards](https://github.com/oneapi-src/oneDNN/blob/main/CODING_STANDARDS.md). In particular:
- Rename the first commit to include a scope.
- Clang-format each affected file

src/gpu/intel/jit/ngen/ngen.hpp
jinz2014: Yes. Thanks.

include/oneapi/dnnl/dnnl_graph.hpp
jinz2014: Did you encounter the following error ?
340: [ RUN      ] APIOp.CreateAllOps
340: [       OK ] APIOp.CreateAllOps (0 ms)
340: [ RUN      ] APIOp.CreateWithInputList
340: [       OK ] APIOp.CreateWithInputList (0 ms)
340: [ RUN      ] APIOp.CreateWithDefaultName
340: [       OK ] APIOp.CreateWithDefaultName (0 ms)
340: [ RUN      ] APIOp.SetInput
340: [       OK ] APIOp.SetInput (0 ms)
340: [ RUN      ] APIOp.SetOutput
340: [       OK ] APIOp.SetOutput (0 ms)
340: [ RUN      ] APIOp.SetAttr
340: [       OK ] APIOp.SetAttr (0 ms)
340: [ RUN      ] APIOp.ShallowCopy
340: [       OK ] APIOp.ShallowCopy (0 ms)
340: [ RUN      ] APIOp.SoftPlusAttr
340: unknown file: Failure
340: C++ exception with description "could not add op to the graph" thrown in the test body.

Simonsays095: No, this is the issue I saw:
>/Users/runner/work/oneDNN/oneDNN/oneDNN/tests/gtests/graph/api/test_cpp_api_op.cpp:125: Failure
Expected equality of these values:
  i
    Which is: 53
  static_cast<size_t>(kind)
    Which is: 54
[  FAILED  ] APIOp.CreateAllOps (0 ms)
[ RUN      ] APIOp.CreateWithInputList
[       OK ] APIOp.CreateWithInputList (0 ms)
...

CreateAllOps is failing because your added enum value shifts the value of the enums that come after it (53 -> 54). Your error looks like something else, but I'm not sure what.

jinz2014: Thank you for explaining the cause.

Simonsays095:
There are a few remaining errors in the graph API:
> The following tests FAILED:
	139 - test_graph_unit_interface_op_schema_cpu (SEGFAULT)
	174 - test_graph_unit_dnnl_pass_cpu (SEGFAULT)
	178 - test_graph_unit_dnnl_reduce_cpu (SEGFAULT)

@TaoLv could you help with some feedback on the changes to the graph API?

include/oneapi/dnnl/dnnl_graph.hpp
rongzha1, line 822: Thanks for your PR for Graph API. As you may know, not all graph OPs are directly mapped to primitive. Such as amax, it can be integrated as Graph API OPs. [Abs](https://oneapi-src.github.io/oneDNN/dev_guide_op_abs.html), and [ReduceMax](https://oneapi-src.github.io/oneDNN/dev_guide_op_reducemax.html)  amax(x) = max(abs(x)). then in Graph API backend, it can be fused to amax then mapped to primitive. May I know if there're any requirement or background for Graph API to add this new OP?

jinz2014: I am not sure if there is any requirement; I tried to add amax because cublaslt and hipblaslt already support the operation  

Thanks

vpirogov:
Closing as stale.

PR #2241, opened 2024-12-05
build: cmake: sdl: check for endbr64

spalicki:
# Description

Testing compiler for flags does not work well on old CMake and old GCC, so instead we can do simple compilation testing for ENDBR64 to unblock PP integration.

# Checklist


spalicki:
make test

mgouicem (approved):
LGTM.
For commit message, please use: `revert: scope: ...` or `scope: revert hash number`.

PR #2242, opened 2024-12-06
rfcs: graph api: support implicit causal mask in SDPA

ElaineBao:
# Description

The RFC proposes adding support for implicit causal mask in SDPA in Graph API.

Rendered version: [link](https://github.com/ElaineBao/oneDNN/blob/graph-api-causal-mask-support/rfcs/20241205-graph-api-causal-mask/README.md) .


rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,417 @@
+# Support Implicit Causal Mask in SDPA in Graph API
+
+## Background
+
+The causal mask is an essential component in Scaled Dot-Product Attention (SDPA), particularly for autoregressive language models like GPT. Its primary purpose is to prevent information leakage from future tokens during the training and inference processes of language models. To achieve this, causal mask is typically implemented as an upper triangular matrix with zeros on and below the main diagonal, and negative infinity above it [1]. This configuration ensures that each token can only attend to itself and the tokens that come before it in the sequence.
```
TaoLv, line 5: Thanks for the RFC.

nit: better to wrap the lines at 80 characters. Same for the below sections.

ElaineBao: Thanks for the reminder, fixed.

rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,485 @@
...
+![causal mask op](images/causal_mask_op.png)
+
+#### Mask Op Support
+
+`Mask` operation masks the input tensor based on attribute settings.
+
+
+| Attribute Name | Description | Value Type | Supported Values | Required or Optional |
+| :------------- | :---------- | :--------- | :--------------- | :------------------- |
+| [mask_option] (@ref dnnl::graph::op::attr::mask_option) | What kind of mask to apply | string     | causal_mask (default), causal_mask_bottom_right | Required             |
```
TaoLv, line 415: nit: it seems `mask_type` can be a better attribute name.

I assume that the purpose is to extend the value of this attribute to support more mask types, eg. padding mask, etc. Can you explain how this op will be defined for them? Do them require more inputs/outputs rather than a single attribute value change?

TaoLv: BTW, have you considered to let the new op just "generate mask" and keep the Select op which is already supported?

ElaineBao: For more mask types, this op needs to be extended, e.g.:
1. for padding mask, besides adding a new `mask_type`, actual sequence length of KV and Q are needed.
2. for sliding window mask, an attribute of window size is needed.
3. for alibi mask, an attribute of alibi slope is needed.

Regarding the new op of "generate mask" (consider it to be named as GenMask), compared with the Mask op, GenMask needs an input of shape, the value of the input can only be known at execution time, while in the Mask op, the value of the shape can be got from the logical tensor of `src`, which is at compilation stage. We want to have the shape information earlier for some potential optimization opportunities, so Mask op is proposed here.

rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,485 @@
...
+| :---- | :------------ | :------------------- |
+| 0     | `dst`         | Required             |
+
+##### Supported data types
+
+`GenInex` operation supports the following data type combinations.
+
+| src  | dst  |
+| :--- | :--- |
+| f32  | f32  |
```
rongzha1, line 307: maybe we can add s32/s64 for output?  pytorch index operations only accept s32/s64. such as https://pytorch.org/docs/stable/generated/torch.Tensor.index_add_.html

ElaineBao: good catch, will add :D

TaoLv: But in fact, oneDNN does not support s64 as memory/tensor data type.

ElaineBao: I also checked the document of CUDNN_POINTWISE_GEN_INDEX, although there's no explicit information about the output data type of this operation, as CUDNN supports INT32 and INT64  data type, we can infer that the op is likely to generate INT32 or INT64, as these are common for representing indices.

Given that, there will be two options for the output data type of GenIndex:

1. Keep f32 data type: if users need integer value, they can convert.
2. Use int32/int64 data type: oneDNN needs to add a new tensor data type, which seems to be overly complex if only consider the specific use case of implicit causal mask.

rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,485 @@
...
+#### Constant Op Support
+
+`Constant` operation takes a scalar as an attribute, and converts it to a tensor,
+which will then be passed to *Add* and *Sub* operations in the case of bottom-right
+causal mask to provide the information of s_q and s_kv. These values are needed during
+compilation stage to generate optimized kernels.
+
+| Attribute Name | Description | Value Type | Supported Values | Required or Optional |
+| :------------- | :---------- | :--------- | :--------------- | :------------------- |
+| [scalar] (@ref dnnl::graph::op::attr::scalar) | Specifies the constant value. | f32, bf16, f16 | arbitrary valid floating point value | Required             |
```
rongzha1, line 356: not sure if bf16 or f16 can work for bottom-right, especially for bf16, only 7 mantissa bits. it can't be used to represent the exact integer number due to the rounding error. maybe we need to add some pattern check when used for mask pattern

dzarukin (commented):
Hi @ElaineBao, thank you for the well-written document. In terms of options, I don't think we want Option 0, it doesn't sound as a rocket science API to leverage some good optimizations that make sense, it would be the matter of implementation effort, actually.
Choosing between Option 1 and Option 2, I'd really ask what PyTorch maintainers want more, because, after all, they are the only party utilizing Graph API.

rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,485 @@
...
+In this case, the aten operator utilizes different SDPA backends, each handling
+the is_causal boolean flag differently, for example:
+
+- cuDNN attention backend uses the cuDNN frontend API with this flag set.
+- Flash attention backend's CPU implementation masks unnecessary GEMM output with
+negative infinity [10].
+
+If integrating oneDNN Graph into the aten operation for this scenario, the
+integration currently needs to generate a causal mask inside the aten operator, and
+pass it to the library as an explicit mask.
```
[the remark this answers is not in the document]
ElaineBao: Hi @dzarukin , thank you for the review.

> It feels that generating explicit mask on a framework side would kill the benefit of memory savings which is the major treasure hunt here...

Yes, exactly. But there are two user scenarios as mentioned in [PyTorch](https://github.com/ElaineBao/oneDNN/blob/graph-api-causal-mask-support/rfcs/20241205-graph-api-causal-mask/README.md#pytorch) part. 
1. If user manually implements SDPA with small operations, then the mask is already there, integrating oneDNN Graph doesn't need to generate a mask additionally.
2. If user directly uses the fused operation - SDPA, there are also two cases:
    a. using it with an explicit mask: then similar to the scenario 1, no need to generate a mask additionally.
    b. using the fused operation with an implicit mask: this is the case that framework has to generate a mask for oneDNN Graph.

In summary, 1. and 2.a. are the cases that user chooses to give up the memory savings by themselves' design consideration, while 2.b. is the case that user has to give up the memory savings because of the current support status of oneDNN Graph. This RFC mainly focuses on solving 2.b.


> Does the request this RFC covers intends to support aten operation?

Yes, since in most user scenarios, the aten operation will be used.

rfcs/20241205-graph-api-causal-mask/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+#### Mask Op Support
+
+`Mask` operation masks the input tensor based on attribute settings.
+
+
+| Attribute Name | Description | Value Type | Supported Values | Required or Optional |
+| :------------- | :---------- | :--------- | :--------------- | :------------------- |
+| [mask_type] (@ref dnnl::graph::op::attr::mask_type) | What type of mask to apply | string     | causal_mask (default), causal_mask_bottom_right | Required             |
+
```
majing921201, line 416:
```suggestion
| [mask_type] (@ref dnnl::graph::op::attr::mask_type) | What type of mask to apply | string | none (default) causal_mask  causal_mask_bottom_right | Required |
```
causal_mask is not a must, so add "none" here

mgouicem (commented):
Thanks for the RFC!
I would side with Dima here. Option 0 seems like a no-go.
Choosing between options 1 and 2 would ultimately be decided by PyTorch and openVINO as they would be the primary users.


Maybe a couple of more considerations:
- time to solution: not clear but option 1 would require to support each individual new op as standalone for when they are not fused, whereas mask is just a single op to add. So if from FWK enabling is fine with either this might be faster to enable
- complexity: usually it is simpler from user perspective if there is a single way to do a specific thing. If we expose mask operation now and the ops of option 1 later, that would create multiple ways to achieve the same result.

dzarukin (approved):

TaoLv (approved):

TaoLv:
The top-left flavor implicit causal mask has been implemented in the library. The bottom-right flavor will be proposed in another RFC.

Issue #120, opened 2017-09-05
Need separate MKLDNN API for Conv2DBackpropBias


nhasabni:
convolution_backward_weights API emits gradients for both filter and bias, but Tensorflow operator for backprop bias and backprop filter are different. The operators get different inputs, and we cannot use convolution_backward_weights to compute backprop bias in Tensorflow.

vpirogov:
It's a design decision to merge weights and bias operations in convolution. In particular on the latest platforms it will be important to have these merged from performance perspective. We are also moving in the directions where we have more merged primitives and it's important to start implementing mechanisms that allow take advantage of these in Tensorflow.

Issue #119, opened 2017-09-05
we find XMM8-16 registers will be modified by MKL-DNN after mkldnn_excute()


a3213105:
only on win64 system. And the Intel manual says XMM 8-16 should be protected by callee

rsdubtso:
I think we have a fix for in c23f8c495a74da5c22c18f15f35bd88dc131c4c3. Can you please provide a testcase?

a3213105:
thanks @rsdubtso, i see the commit date is  13 Jul, i use the mkl lib from the parallel studio. maybe this commit is not merge into the lib which i used. i will try the latest release v0.10 first :).
if i have any question， i will ask you， thanks a lot

rsdubtso:
The mkl that comes with the studio is a separate product and is not API compatible. And it may have this bug depending on which version you use.

vpirogov:
The issue does not affect Intel MKL-DNN, so I will close it on Github. In Intel MKL 2018 at least LRN implementation does not restore xmm8-16 on Windows, I created an internal defect to track this.

PR #2243, opened 2024-12-06
gpu: generic: implement SYCL reduction

sgeor255:
# Checklist

## General

### DIFF
--- a/src/gpu/generic/sycl/reduction_kernels.hpp
+++ b/src/gpu/generic/sycl/reduction_kernels.hpp
@@ -1,0 +1,660 @@
+/*******************************************************************************
+* Copyright 2023-2024 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef GPU_GENERIC_SYCL_REDUCTION_KERNELS_HPP
+#define GPU_GENERIC_SYCL_REDUCTION_KERNELS_HPP
+
+#include "common/c_types_map.hpp"
+#include "common/dnnl_thread.hpp"
+#include "common/primitive_exec_types.hpp"
+#include "common/utils.hpp"
+#include "gpu/generic/sycl/sycl_io_helper.hpp"
+#include "gpu/generic/sycl/sycl_math_utils.hpp"
+#include "gpu/generic/sycl/sycl_primitive_conf.hpp"
+#include "xpu/sycl/memory_storage_base.hpp"
+#include "xpu/sycl/types.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace gpu {
+namespace generic {
+namespace sycl {
+
+struct init_kernel_t {
+    init_kernel_t(xpu::sycl::out_memory_arg_t &out_arg, alg_kind_t alg)
+        : out_arg_(out_arg), alg_(alg) {}
+
+    void operator()(::sycl::item<1> item) const {
+        auto *out = reinterpret_cast<float *>(out_arg_.get_pointer());
+        const int idx = item.get_linear_id();
+        float val = 0;
+        if (alg_ == alg_kind::reduction_min)
+            val = std::numeric_limits<float>::max();
+        else if (alg_ == alg_kind::reduction_max)
+            val = std::numeric_limits<float>::lowest();
+        else if (alg_ == alg_kind::reduction_mul)
+            val = 1;
+        else
+            val = 0;
+
+        out[idx] = val;
+    }
+
+private:
+    xpu::sycl::out_memory_arg_t out_arg_;
+    alg_kind_t alg_;
+};
+
+struct atomic_finalize_kernel_t {
+    atomic_finalize_kernel_t(::sycl::handler &cgh, const exec_ctx_t &ctx,
+            data_type_t dt, xpu::sycl::out_memory_arg_t &out_arg,
+            alg_kind_t alg, float p, float eps, sycl_post_ops_t &post_ops,
+            xpu::sycl::md_t dst_md, int reduce_size)
+        : dt_(dt)
+        , out_arg_(out_arg)
+        , alg_(alg)
+        , p_(p)
+        , eps_(eps)
+        , post_ops_(post_ops)
+        , po_args_(cgh, ctx, post_ops)
+        , dst_md_(dst_md)
+        , reduce_size_(reduce_size) {}
+
+    void operator()(::sycl::item<1> item) const {
+        void *out_ptr = out_arg_.get_pointer();
+        auto idx = item.get_linear_id();
+        auto val = load_float_value(dt_, out_ptr, idx);
+        if (alg_ == alg_kind::reduction_norm_lp_max) {
+            val = ::sycl::rootn(::sycl::max(val, eps_), p_);
+        } else if (alg_ == alg_kind::reduction_norm_lp_sum) {
+            val = ::sycl::rootn(val + eps_, p_);
+        } else if (alg_ == alg_kind::reduction_norm_lp_power_p_max) {
+            val = ::sycl::max(val, eps_);
+        } else if (alg_ == alg_kind::reduction_norm_lp_power_p_sum) {
+            val = val + eps_;
+        } else if (alg_ == alg_kind::reduction_mean) {
+            val = val / reduce_size_;
+        }
+
+        auto prev_val = val;
+        auto l_offset = idx;
+        dims_t pos;
+        for (int i = 0; i < dst_md_.ndims(); i++) {
+            const int d = dst_md_.ndims() - 1 - i;
+            const dim_t cur_dim = dst_md_.dims()[d];
+            pos[d] = l_offset % cur_dim;
+            l_offset = l_offset / cur_dim;
+        }
+        val = post_ops_.apply(val, prev_val, po_args_, pos);
+
+        store_float_value(dt_, val, out_ptr, idx);
+    }
+
+private:
+    data_type_t dt_;
+    xpu::sycl::out_memory_arg_t out_arg_;
+    alg_kind_t alg_;
+    float p_, eps_;
+    sycl_post_ops_t post_ops_;
+    post_op_input_args po_args_;
+    xpu::sycl::md_t dst_md_;
+    int reduce_size_;
+};
+
+struct Reducer {
+    alg_kind_t alg_;
+    float p_, eps_;
+    bool needs_prepare_;
+    bool iter_needs_finalize_;
+
+    Reducer(alg_kind_t alg, float p, float eps, bool needs_prepare,
+            bool iter_needs_finalize)
+        : alg_(alg)
+        , p_(p)
+        , eps_(eps)
+        , needs_prepare_(needs_prepare)
+        , iter_needs_finalize_(iter_needs_finalize) {}
+
+    inline float identity() const {
+        if (alg_ == alg_kind::reduction_min) {
+            return std::numeric_limits<float>::max();
+        } else if (alg_ == alg_kind::reduction_max) {
+            return std::numeric_limits<float>::lowest();
+        } else if (alg_ == alg_kind::reduction_mul) {
+            return 1.f;
+        }
+
+        return 0.f;
+    }
+
+    inline bool needs_finalize() const {
+        return (alg_ == alg_kind::reduction_mean
+                || alg_ == alg_kind::reduction_norm_lp_max
+                || alg_ == alg_kind::reduction_norm_lp_sum
+                || alg_ == alg_kind::reduction_norm_lp_power_p_max
+                || alg_ == alg_kind::reduction_norm_lp_power_p_sum);
+    }
+
+    inline float subgroup_reduce(
+            ::sycl::sub_group &subgroup, float sg_input) const {
+        if (alg_ == alg_kind::reduction_sum || alg_ == alg_kind::reduction_mean
+                || alg_ == alg_kind::reduction_norm_lp_max
+                || alg_ == alg_kind::reduction_norm_lp_sum
+                || alg_ == alg_kind::reduction_norm_lp_power_p_max
+                || alg_ == alg_kind::reduction_norm_lp_power_p_sum) {
+            return ::sycl::reduce_over_group(
+                    subgroup, sg_input, ::sycl::plus<float> {});
+        } else if (alg_ == alg_kind::reduction_min) {
+            return ::sycl::reduce_over_group(
+                    subgroup, sg_input, ::sycl::minimum<float> {});
+        } else if (alg_ == alg_kind::reduction_max) {
+            return ::sycl::reduce_over_group(
+                    subgroup, sg_input, ::sycl::maximum<float> {});
+        } else if (alg_ == alg_kind::reduction_mul) {
+            return ::sycl::reduce_over_group(
+                    subgroup, sg_input, ::sycl::multiplies<float> {});
+        }
+
+        return ::sycl::nan(0U);
+    }
+
+    inline float reduce(float lhs, float rhs) const {
+        if (alg_ == alg_kind::reduction_sum || alg_ == alg_kind::reduction_mean
+                || alg_ == alg_kind::reduction_norm_lp_max
+                || alg_ == alg_kind::reduction_norm_lp_sum
+                || alg_ == alg_kind::reduction_norm_lp_power_p_max
+                || alg_ == alg_kind::reduction_norm_lp_power_p_sum) {
+            return lhs + rhs;
+        } else if (alg_ == alg_kind::reduction_min) {
+            return ::sycl::min(lhs, rhs);
+        } else if (alg_ == alg_kind::reduction_max) {
+            return ::sycl::max(lhs, rhs);
+        } else if (alg_ == alg_kind::reduction_mul) {
+            return lhs * rhs;
+        }
+
+        return ::sycl::nan(0U);
+    }
+
+    template <::sycl::memory_order Order, ::sycl::memory_scope Scope,
+            ::sycl::access::address_space Space>
+    void atomic_op(data_type_t dt, void *ref, int idx, float val, int size) {
+        auto atomic_out = ::sycl::atomic_ref<float, Order, Scope, Space>(
+                reinterpret_cast<float *>(ref)[idx]);
+        if (alg_ == alg_kind::reduction_sum || alg_ == alg_kind::reduction_mean
+                || alg_ == alg_kind::reduction_norm_lp_max
+                || alg_ == alg_kind::reduction_norm_lp_sum
+                || alg_ == alg_kind::reduction_norm_lp_power_p_max
+                || alg_ == alg_kind::reduction_norm_lp_power_p_sum) {
+            atomic_out.fetch_add(val);
+        } else if (alg_ == alg_kind::reduction_min) {
+            atomic_out.fetch_min(val);
+        } else if (alg_ == alg_kind::reduction_max) {
+            atomic_out.fetch_max(val);
+        }
+    }
+
+    inline void prepare(float &val) {
+        if (needs_prepare_
+                && (alg_ == alg_kind::reduction_norm_lp_max
+                        || alg_ == alg_kind::reduction_norm_lp_sum
+                        || alg_ == alg_kind::reduction_norm_lp_power_p_max
+                        || alg_ == alg_kind::reduction_norm_lp_power_p_sum)) {
+            val = ::sycl::pow(::sycl::fabs(val), p_);
+        }
+    }
+
+    inline void finalize(float &val, int size) {
+        if (alg_ == alg_kind::reduction_mean) {
+            val /= size;
+        } else if (alg_ == alg_kind::reduction_norm_lp_max
+                && iter_needs_finalize_) {
+            val = ::sycl::rootn(::sycl::max(val, eps_), p_);
+        } else if (alg_ == alg_kind::reduction_norm_lp_sum
+                && iter_needs_finalize_) {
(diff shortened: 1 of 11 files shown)


src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,479 @@
...
+                range_t {conf.batch_size, global_row, global_col});
+        needs_atomic_reduction_ = needs_atomic_reduction_
+                || (global_row > std::min(
+                            tile_row, conf.num_sg_reductions * max_sg_size));
+
+        if (needs_atomic_reduction_ && conf.alg == dnnl_reduction_mul) {
+            return status::unimplemented;
+        }
+        if (needs_atomic_reduction_ && attr()->post_ops_.find(dnnl_sum) != -1) {
+            return status::unimplemented;
```
Rbiessy, line 297: It seems there are a few cases where we would return `unimplemented`. Could we fallback to a simple implementation instead?

src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,479 @@
...
+            if (red_iter != num_reductions_ - 1) { conf.dst_dt = dnnl_f32; }
+        }
+
+        conf.batch_groups = batch_groups;
+        confs_.push_back(conf);
+        dims[axes[red_iter]] = 1;
+    }
+
+    if (needs_reorder_) { CHECK(init_reorder(engine)); }
+    if (needs_atomic_reduction_) { CHECK(init_out_scratchpad()); }
```
Rbiessy, line 320: Can we check that the SYCL device supports atomic when `needs_atomic_reduction_` is `true`? I think we just need to check the query `info::device::atomic_memory_scope_capabilities` contains `memory_scope::work_group`.

src/gpu/generic/sycl/ref_reduction.hpp
```diff
@@ -0,0 +1,135 @@
...
+            using sm = primitive_attr_t::skip_mask_t;
+
+            memory_desc_wrapper src_wrap(src_md());
+            memory_desc_wrapper dst_wrap(dst_md());
+
+            bool ok = set_default_params() == status::success
+                    && attr()->has_default_values(sm::post_ops)
+                    && sycl_post_ops_t::post_ops_ok(attr())
+                    && attr_.set_default_formats(dst_md()) == status::success
+                    && src_wrap.is_plain() && dst_wrap.is_plain()
```
mgouicem, line 64: `is_plain` only checks that the memory is not blocked. 
However, below you seem to rely on the property that stride = accumulate(dims), which is not guaranteed by `is_plain`. Calling `is_dense` would guarantee that though.

mgouicem (commented):
Overall, it seems to be a pretty optimized implementation for a reference one :).
I would suggest to at least have a baseline implementation with fewer restrictions (and maybe less efficient).

Few other comments:
- atomics based reduction can happen only under non deterministic mode (see [doc](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_deterministic.html))
- sycl has a [reducer extension ](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reduction)(mandatory per spec). Any reason to not rely on that?

src/gpu/generic/sycl/sycl_primitive_conf.hpp
```diff
@@ -415,6 +415,39 @@ struct sycl_pooling_bwd_conf_t : public sycl_pooling_base_conf_t {
...
+    bool transpose;
+    bool bank_offset;
+    bool is_first_iter;
+    bool is_last_iter;
+
+    sycl_post_ops_t post_ops;
+
+    const int num_sg_reductions = 4;
+    static constexpr int local_row_wg = 256;
+    static constexpr int local_col_wg = 8;
```
mgouicem, line 448: Any reason for these values (maybe adding a few comments to clarify)?

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,658 @@
...
+        }
+    }
+};
+
+struct LocalMemTile {
+    using T = float;
+    using Index = int;
+
+    static constexpr int Dim = 3;
+    static constexpr bool CheckBounds = true;
```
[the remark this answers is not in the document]
sgeor255: This is true, added the comment.

src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,479 @@
...
+    exec_ctx_t reorder_ctx(ctx, std::move(reorder_args));
+
+    return reorder_p_->execute(reorder_ctx);
+}
+
+} // namespace sycl
+} // namespace generic
+} // namespace gpu
+} // namespace impl
+} // namespace dnnl
```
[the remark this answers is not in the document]
sgeor255: Done.

src/gpu/generic/sycl/ref_reduction.cpp
sgeor255: I have added a fallback simple implementation in #2249 .

src/gpu/generic/sycl/ref_reduction.cpp
sgeor255: Good point, done.

sgeor255:
@mgouicem 

> I would suggest to at least have a baseline implementation with fewer restrictions (and maybe less efficient).

Added a simple implementation in #2249 .

> * atomics based reduction can happen only under non deterministic mode (see [doc](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_deterministic.html))

I had missed this, thanks. Added a check.

> * sycl has a [reducer extension ](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reduction)(mandatory per spec). Any reason to not rely on that?

No particular reason. This implementation is based on an older implementation we had worked on before this extension was added.

> * test enabling: it seems gtest is disabled for GPU. It might be time to reenable it.

Enabled the tests. Also updated the documentation.

src/gpu/generic/sycl/sycl_primitive_conf.hpp
sgeor255: These were derived from testing and are values that should work in most cases. I added a comment to actually tune them in the future when we do an optimisation pass.

src/gpu/generic/sycl/ref_reduction.hpp
sgeor255: Added `is_dense()` checks, thanks. I also kept the `is_plain()` checks as their main purpose was to ensure unimplemented is returned for blocked formats (which doesn't seem to be the case for `is_dense()`).

Rbiessy (approved):

src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+        if (!IMPLICATION(needs_atomic_reduction_, supports_atomics)) {
+            return status::unimplemented;
+        }
+        // Atomics are only supported when deterministic == false
+        if (!IMPLICATION(needs_atomic_reduction_, !attr()->deterministic_)) {
+            return status::unimplemented;
+        }
+        if (needs_atomic_reduction_ && conf.alg == dnnl_reduction_mul) {
+            return status::unimplemented;
```
mgouicem, line 311: (minor) for these kind of checks, there is a `VDISPATCH` macro that allows to print diagnostics with ONEDNN_VERBOSE mode (see an example [here](https://github.com/oneapi-src/oneDNN/blob/0bff5bd0f56dcf5e68303198996e15a065ffb16d/src/gpu/generic/sycl/ref_convolution.hpp#L91)).

mgouicem (approved):

src/gpu/generic/sycl/ref_reduction.cpp
sgeor255: Done, thanks.

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,659 @@
...
+
+namespace dnnl {
+namespace impl {
+namespace gpu {
+namespace generic {
+namespace sycl {
+
+struct init_kernel_t {
+    xpu::sycl::out_memory_arg_t out_arg_;
+    dnnl_alg_kind_t alg_;
```
dzarukin, line 38: Please replace with `alg_kind_t`. Please update all algorithms with `alg_kind::name`. They are meant to be used from an `alg_kind` namespace, not the public API C type.

src/gpu/generic/sycl/ref_reduction.cpp
Rbiessy: I pushed the change for him as Svet is on holiday

src/gpu/generic/sycl/reduction_kernels.hpp
Rbiessy: This is done. I have applied similar changes to avoid using `dnnl_data_type_t` and `dnnl_dim_t`.

vpirogov (approved):

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,659 @@
...
+#include "xpu/sycl/types.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace gpu {
+namespace generic {
+namespace sycl {
+
+struct init_kernel_t {
+    xpu::sycl::out_memory_arg_t out_arg_;
```
dzarukin, line 37: Move members to private since there's a constructor?

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,659 @@
...
+
+struct atomic_finalize_kernel_t {
+    data_type_t dt_;
+    xpu::sycl::out_memory_arg_t out_arg_;
+    alg_kind_t alg_;
+    float p_, eps_;
+    sycl_post_ops_t post_ops_;
+    post_op_input_args po_args_;
+    xpu::sycl::md_t dst_md_;
+    int reduce_size_;
```
dzarukin, line 68: Same?

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,659 @@
...
+    static constexpr bool CheckBounds = true;
+    static constexpr int RowDim = Dim == 3 ? 1 : 0;
+    static constexpr int ColDim = Dim == 3 ? 2 : 1;
+    int row_tile_;
+    int col_tile_;
+    int row_lim_;
+    int col_lim_;
+    int local_id_;
+    int local_row_id_;
+    int local_col_id_;
```
dzarukin, line 252: `Index` type?

tests/gtests/test_reduction.cpp
```diff
@@ -61,11 +59,22 @@ class reduction_test_t
...
+        SKIP_IF_GENERIC(!(generic_supported_format_tag(p.src_format)
+                                && generic_supported_format_tag(p.dst_format)),
+                "Unsupported format tag");
 
         catch_expected_failures(
                 [&]() { Test(); }, p.expect_to_fail, p.expected_status);
     }
 
+    bool generic_supported_format_tag(memory::format_tag tag) {
+        return impl::utils::one_of(tag, dnnl_a, dnnl_ab, dnnl_abc, dnnl_abcd,
```
dzarukin, line 71:
```suggestion
        return impl::utils::one_of(tag, format_tag::a, ...
```

src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,499 @@
...
+        VDISPATCH_REDUCTION(
+                IMPLICATION(needs_atomic_reduction_, supports_atomics),
+                "Implementation needs to perform atomic reduction, but atomics "
+                "are not supported by current device");
+        VDISPATCH_REDUCTION(
+                IMPLICATION(needs_atomic_reduction_, !attr()->deterministic_),
+                "Atomic reduction is only supported in non-deterministic mode");
+        VDISPATCH_REDUCTION(IMPLICATION(needs_atomic_reduction_,
+                                    conf.alg != alg_kind::reduction_mul),
+                "Algorithm Mul is not supported with atomic reduction");
```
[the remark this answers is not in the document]
AD2605: IIUC, you are accumulating and storing it in f32, so you could implement an atomic multiplication using `compare_exchange_strong` and a do-while loop, which would enable any atomic operation we need.

src/gpu/generic/sycl/ref_reduction.cpp
```diff
@@ -0,0 +1,499 @@
...
+        VDISPATCH_REDUCTION(IMPLICATION(needs_atomic_reduction_,
+                                    attr()->post_ops_.find(dnnl_sum) == -1),
+                "Sum postop is not supported with atomic reduction");
```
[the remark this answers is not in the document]
sgeor255: The sum post-op requires the previous values stored in the output but with atomic reduction we do some intermediate computations and store the result in the output. This case is also handled by the fallback kernel.

src/gpu/generic/sycl/ref_reduction.cpp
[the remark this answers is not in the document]
sgeor255: As we discussed offline this case is handled by the fallback kernel and not that important to add currently.

src/gpu/generic/sycl/reduction_kernels.hpp
sgeor255: Done, thanks.

src/gpu/generic/sycl/reduction_kernels.hpp
sgeor255: Done, thanks.

tests/gtests/test_reduction.cpp
sgeor255: Done, thanks.

sgeor255:
The CI failures seem unrelated to this PR (copyright checks failed for files in `src/cpu/` not added in this PR).

src/gpu/generic/sycl/ref_reduction.cpp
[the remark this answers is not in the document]
AD2605: > The sum post-op requires the previous values stored in the output but with atomic reduction we do some intermediate computations and store the result in the output.

this needn't block the PR, but why cannot we add it to the final value of the reduction at the end, irrespective of the reduction type (atomic or not), to the output, instead of relying on a separate kernel ?

src/gpu/generic/sycl/ref_reduction.cpp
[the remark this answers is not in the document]
AD2605: I am okay with it either way. It was more about the reducing the number of kernels rather than performance since it's a reference implementation anyway

AD2605 (approved):

src/gpu/generic/sycl/ref_reduction.cpp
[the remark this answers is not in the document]
ShanoToni: The main problem with atomic reduction and the post ops sum comes from the fact that in the atomic reduction case we are reducing a dimension which does not fit inside the workgroup. In this case multiple workgroups are used and we lose any kind of synchronization the only guarantee we have here is that the atomic operation will increment the value each work item in the different workgroups will accumulate without a race condition, we do not know which workgroup's work item would work in the data first. 

The only way i can see this being done is selecting only one workgroup's work items (for example wg 0) to apply the post-op sum with an atomic_add, which would require a custom way for reductions to apply the sum post op.

ShanoToni (approved):

src/gpu/generic/sycl/reduction_kernels.hpp
```diff
@@ -0,0 +1,661 @@
...
+                auto val = load_float_value(dt, input, local_id);
+                reducer.finalize(val, finalize_param);
+                store_float_value(dt, val, input, local_id);
+            }
+        }
+    }
+
+    template <::sycl::memory_order Order, ::sycl::memory_scope Scope,
+            ::sycl::access::address_space
+                    Space /* , typename InMem, typename OutMem */>
```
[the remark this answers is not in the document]
sgeor255: Done, thanks.

vpirogov:
@sgeor255, this change breaks gtests on Intel GPUs:
```
119/225 Test #119: test_reduction_gpu .......................................***Exception: Numerical  0.35 sec
Note: Google Test filter = *:-*_CPU*
[==========] Running 58 tests from 16 test suites.
[----------] Global test environment set-up.
[----------] 5 tests from TestReductionEF/reduction_test_f32
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/0
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/0 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/1
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/1 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/2
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/2 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/3
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/3 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/4
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/4 (0 ms)
[----------] 5 tests from TestReductionEF/reduction_test_f32 (0 ms total)

[----------] 1 test from TestReductionZero/reduction_test_f32
[ RUN      ] TestReductionZero/reduction_test_f32.TestsReduction/0

        Start 120: test_reduction_buffer_gpu
120/225 Test #120: test_reduction_buffer_gpu ................................***Exception: Numerical  0.35 sec
Note: Google Test filter = *:-*_CPU*
[==========] Running 58 tests from 16 test suites.
[----------] Global test environment set-up.
[----------] 5 tests from TestReductionEF/reduction_test_f32
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/0
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/0 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/1
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/1 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/2
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/2 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/3
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/3 (0 ms)
[ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/4
[       OK ] TestReductionEF/reduction_test_f32.TestsReduction/4 (0 ms)
[----------] 5 tests from TestReductionEF/reduction_test_f32 (0 ms total)

[----------] 1 test from TestReductionZero/reduction_test_f32
[ RUN      ] TestReductionZero/reduction_test_f32.TestsReduction/0

```

sgeor255:
> @sgeor255, this change breaks gtests on Intel GPUs:
> 
> ```
> 119/225 Test #119: test_reduction_gpu .......................................***Exception: Numerical  0.35 sec
> Note: Google Test filter = *:-*_CPU*
> [==========] Running 58 tests from 16 test suites.
> [----------] Global test environment set-up.
> [----------] 5 tests from TestReductionEF/reduction_test_f32
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/0
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/0 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/1
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/1 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/2
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/2 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/3
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/3 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/4
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/4 (0 ms)
> [----------] 5 tests from TestReductionEF/reduction_test_f32 (0 ms total)
> 
> [----------] 1 test from TestReductionZero/reduction_test_f32
> [ RUN      ] TestReductionZero/reduction_test_f32.TestsReduction/0
> 
>         Start 120: test_reduction_buffer_gpu
> 120/225 Test #120: test_reduction_buffer_gpu ................................***Exception: Numerical  0.35 sec
> Note: Google Test filter = *:-*_CPU*
> [==========] Running 58 tests from 16 test suites.
> [----------] Global test environment set-up.
> [----------] 5 tests from TestReductionEF/reduction_test_f32
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/0
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/0 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/1
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/1 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/2
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/2 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/3
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/3 (0 ms)
> [ RUN      ] TestReductionEF/reduction_test_f32.TestsReduction/4
> [       OK ] TestReductionEF/reduction_test_f32.TestsReduction/4 (0 ms)
> [----------] 5 tests from TestReductionEF/reduction_test_f32 (0 ms total)
> 
> [----------] 1 test from TestReductionZero/reduction_test_f32
> [ RUN      ] TestReductionZero/reduction_test_f32.TestsReduction/0
> ```

Thanks for letting me know @vpirogov , I will take a look.

vpirogov:
It's a bit surprising that OpenCL runtime cases are broken too. This indicates the issue is coming from GPU common code, not from the implementation itself.

sgeor255:
> It's a bit surprising that OpenCL runtime cases are broken too. This indicates the issue is coming from GPU common code, not from the implementation itself.

@vpirogov https://github.com/oneapi-src/oneDNN/pull/2453 should fix this.

Issue #1232, opened 2022-01-15
Ability to generate JIT code with frame pointers to enable Linux perf stack unwinding


aalexand:
It seems that the JIT code generated by MKL DNN today does not set up frame pointers which means performance tools like Linux perf cannot unwind the stack by default.

It's understood that utilizing a frame pointer register can have negative performance impact due to one less register available for general purposes, but this may be an acceptable trade-off for some users.

It's also understood that the JIT code would not be symbolized unless further integration with the profiler is in place, but merely being able to unwind the code is useful since the caller functions would be symbolized and can provide enough context on what part of the call tree this JIT code is invoked by and hence what it does.

Would it be possible to have an opt-in mode where the generated JIT code would set up frame pointers to enable Linux perf stack unwinding in the default frame pointer mode?

dzarukin:
Hi @aalexand, thank you for the question. Did you have a chance to look through [this article](https://oneapi-src.github.io/oneDNN/dev_guide_profilers.html)? If following steps and using certain variables oneDNN supports, would it cover your need? Thank you.

aalexand:
@dzarukin Is there a value for ONEDNN_JIT_PROFILE that enables frame pointer generation while NOT enabling perf-pid or JIT dump file generation?

From the doc it is not clear that there is.

dzarukin:
> Is there a value for ONEDNN_JIT_PROFILE that enables frame pointer generation while NOT enabling perf-pid or JIT dump file generation?

There's no such value and, likely, support you are asking about.

Could you explain what's the problem are you trying to solve and why with exact approach you are asking about?

All of jit-ted implementation follow the same pattern - there's an `execute(...)` call, inside which there's a call to parallel section (`parallel(...)`, `parallel_nd(...)` or some of their modifications). Each thread in such section creates kernel arguments and use function call operator `operator()` of kernel which is defined in jit_generator.hpp.

I don't see what kind of stack you want to inspect with perf (rather than gdb) and what's the purpose. Sorry.

P.S. If you would like to see the stack chain I've just described, you may try to compile with `-fno-omit-frame-pointer` and see if it provides the necessary level of stack trace (though I have a feeling one may need to deal with threading first somehow). Thank you.

aalexand:
[...]

PR #2244, opened 2024-12-07
build: cmake: enabling frame pointers

spalicki:
Adding flags that build oneDNN with frame pointers, which can be used to walk CPU stack frames by profilers. First step for #1232

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -120,6 +120,16 @@ endif()
 if(MSVC)
     set(USERCONFIG_PLATFORM "x64")
     append_if(DNNL_WERROR CMAKE_CCXX_FLAGS "/WX")
+
+    # Generating frame pointers for easier performance profiling
+    if(DNNL_TARGET_ARCH STREQUAL "X64")
+        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+            append(CMAKE_CCXX_FLAGS "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer")
+        else()
+            append(CMAKE_CCXX_FLAGS "/Oy-")
+        endif()
+    endif()
+
     if(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
         append(CMAKE_CCXX_FLAGS "/MP")
         # increase number of sections in obj file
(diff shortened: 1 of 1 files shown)


spalicki:
make test

spalicki:
@theComputeKid @Radu2k If you prefer, I can limit this feature to x64, but it should be beneficial perf profiling on all architectures.

Radu2k (approved):
@spalicki That looks good to me. Thanks!

cmake/platform.cmake
```diff
@@ -233,6 +241,9 @@ elseif(UNIX OR MINGW)
         append(CMAKE_CCXX_FLAGS "-Wsign-compare")
     endif()
 
+    # Generating frame pointers for easier performance profiling
+    append(CMAKE_CCXX_FLAGS "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer")
```
mgouicem, line 245: Did you run any performance measurements? I don't think we should see any issue (most of the performance is in jitted kernels), but it might be worth checking to be on the safe side.

theComputeKid: The perf might also be arch dependant in this case.

theComputeKid: This is for Windows and Linux, I am curious to know the perf effect on Linux on AArch64. We might need to do a quick sanity check.

spalicki: There is no visible performance difference in execution time on JIT version, but weirdly enough kernel creation is consistently faster with frame pointer enabled version by ax 5%. In REF implementation both kernel creation and execution time fluctuates by ax 5% when using full socket per instance but on smaller 4-core per instance limited runs the performance is identical.

Sqvid: @spalicki I'll run similar benchmarking for aarch64 (I assume your results are x64-only?) and get back to you shortly. Would appreciate it if you could hold off from merging in the meantime.

Sqvid: @spalicki @mgouicem Thanks for your patience. I've tested this now and it looks like it does cause a noticeable performance drop for me on Neoverse (R) V1.  I used the matmul tests as a benchmark. There seem to be a slowdown against all tests but as an example:

```
./base/tests/benchdnn/benchdnn --mode=P --matmul --perf-template="%impl%,%name%,%prb%,%-ctime%,%-time%" --dt=bf16:bf16:bf16 --attr-post-ops=linear:2:1 2048x256:256x128_n"NCF:1*1"  
Output template: %impl%,%name%,%prb%,%-ctime%,%-time%
gemm:jit:bf16,NCF:1*1,--mode=P --matmul --dt=bf16:bf16:bf16 --attr-post-ops=linear:2:1 2048x256:256x128_nNCF:1*1,0.0136719,2.42505
total perf: min(ms):2.42505 avg(ms):2.43697
total: 3.02s; fill: 0.01s (0%);

./pr2244/tests/benchdnn/benchdnn --mode=P --matmul --perf-template="%impl%,%name%,%prb%,%-ctime%,%-time%" --dt=bf16:bf16:bf16 --attr-post-ops=linear:2:1 2048x256:256x128_n"NCF:1*1"
Output template: %impl%,%name%,%prb%,%-ctime%,%-time%
gemm:jit:bf16,NCF:1*1,--mode=P --matmul --dt=bf16:bf16:bf16 --attr-post-ops=linear:2:1 2048x256:256x128_nNCF:1*1,0.0146484,3.62427
total perf: min(ms):3.62427 avg(ms):3.63837
total: 3.02s; fill: 0.01s (0%);
```

Let's skip this on aarch64 for now. Thanks.

spalicki:
make test

tczeszun (approved):

spalicki:
make Lightweight_scans

PR #2245, opened 2024-12-09
[SYCL] Use L0 queries for L0 devices

mgouicem:
This PR replaces OCL queries with matching L0 queries for L0 devices on Intel devices.
Few notable things:
- Updated L0 headers to access some recent queries
- Added headers for Intel specific L0 extensions. These headers come from [compute-runtime](https://github.com/intel/compute-runtime/tree/master/level_zero/include). ~~We asked about forward compatibility status of these symbols and are waiting for clarification (hence why this PR is in draft)~~. Driver is backward compatible for a given major version of the level_zero driver, so we should be on the same level of compatibility as non-Intel specific functions.
- switched ngen to dynamically load L0


mgouicem:
make test 
disable device_cpu
enable device_gpu
enable thr_sycl
enable thr_ocl

src/gpu/intel/jit/ngen/ngen_level_zero.hpp
```diff
@@ -21,13 +21,100 @@
...
+
+ze_result_t func_zeModuleCreate(ze_context_handle_t hContext,
+        ze_device_handle_t hDevice, const ze_module_desc_t *desc,
+        ze_module_handle_t *phModule,
+        ze_module_build_log_handle_t *phBuildLog) {
+    static auto f = find_ze_symbol<decltype(&zeModuleCreate)>("zeModuleCreate");
+
+    if (!f) return ZE_RESULT_ERROR_UNINITIALIZED;
+    return f(hContext, hDevice, desc, phModule, phBuildLog);
+}
```
petercad, line 85: Could you reduce the boilerplate with some generic code + macros?

```
#define ZE_INDIRECT_API(f) \
    template <typename... Args> ze_result_t call_##f(Args&&... args) { \
        static auto f = find_ze_symbol<decltype(%f)>(#f);
        if (!f) return ZE_RESULT_ERROR_UNINITIALIZED;
        return f(std::forward<Args>(args)...);
    }

ZE_INDIRECT_API(zeModuleCreate)
ZE_INDIRECT_API(zeModuleDestroy)
/* ... */
```

src/gpu/intel/jit/ngen/ngen_level_zero.hpp
```diff
@@ -21,13 +21,100 @@
...
+void *find_ze_symbol(const char *symbol) {
+#if defined(__linux__)
+    void *handle = dlopen("libze_loader.so.1", RTLD_NOW | RTLD_LOCAL);
+#elif defined(_WIN32)
+    // Use LOAD_LIBRARY_SEARCH_SYSTEM32 flag to avoid DLL hijacking issue.
+    HMODULE handle = LoadLibraryExA(
+            "ze_loader.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+#endif
+    if (!handle) {
+        assert(!"not expected");
```
petercad, line 51: nGEN uses named exception classes (like `level_zero_error` below) instead of assertions to handle these kinds of situations, so it would be nice to match that in this code.

src/gpu/intel/jit/ngen/ngen_level_zero.hpp
mgouicem: Done, thanks!

src/gpu/intel/jit/ngen/ngen_level_zero.hpp
mgouicem: Done. Thanks!

src/gpu/intel/jit/ngen/ngen_level_zero.hpp
```diff
@@ -36,6 +44,54 @@ class level_zero_error : public std::runtime_error {
...
+    // Use LOAD_LIBRARY_SEARCH_SYSTEM32 flag to avoid DLL hijacking issue.
+    HMODULE handle = LoadLibraryExA(
+            "ze_loader.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+#endif
+    if (!handle) throw level_zero_error{ZE_RESULT_ERROR_UNINITIALIZED};
+
+    using zeInit_decl_t = ze_result_t (*)(ze_init_flags_t flags);
+    const ze_init_flags_t default_ze_flags = 0;
+#if defined(__linux__)
+    static const ze_result_t ze_result = reinterpret_cast<zeInit_decl_t>(
```
petercad, line 63: You can actually omit the calls to `zeInit` as the caller must invoke L0 prior to entering nGEN (to locate the device, create a context, etc.).

petercad (approved):

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+    ze_device_properties_t deviceProps = {ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES};
+    ze_device_ip_version_ext_t devicePropsIP
+            = {ZE_STRUCTURE_TYPE_DEVICE_IP_VERSION_EXT};
+    deviceProps.pNext = &devicePropsIP;
+    func_zeDeviceGetProperties(device, &deviceProps);
+    ip_version = devicePropsIP.ipVersion;
+    return status::success;
+}
+
+status_t get_lz_device_enabled_systolic_intel(
```
echeresh, line 300: Nit:
```suggestion
status_t get_l0_device_enabled_systolic_intel(
```

Looks like we don't use this abbreviation, it's either `l0`, `level0`, `level zero` (or `ze` like in API) and their capitalization variations.

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
     return lhs_ze_handle == rhs_ze_handle;
 }
 
+status_t get_device_ip(ze_device_handle_t device, uint32_t &ip_version) {
+    ze_device_properties_t deviceProps = {ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES};
+    ze_device_ip_version_ext_t devicePropsIP
+            = {ZE_STRUCTURE_TYPE_DEVICE_IP_VERSION_EXT};
+    deviceProps.pNext = &devicePropsIP;
+    func_zeDeviceGetProperties(device, &deviceProps);
```
echeresh, line 295: Missing status check?

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+status_t get_lz_device_enabled_systolic_intel(
+        ze_device_handle_t device, bool &mayiusesystolic) {
+    ze_device_module_properties_t deviceModProps
+            = {ZE_STRUCTURE_TYPE_DEVICE_MODULE_PROPERTIES};
+    // Note: supported by Intel Driver 24.05 and onwards
+    ze_intel_device_module_dp_exp_properties_t deviceModPropsExt
+            = {ZE_STRUCTURE_INTEL_DEVICE_MODULE_DP_EXP_PROPERTIES};
+    deviceModProps.pNext = &deviceModPropsExt;
+
+    func_zeDeviceGetModuleProperties(device, &deviceModProps);
```
echeresh, line 309: Status check?

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+    ze_device_ip_version_ext_t devicePropsIP
+            = {ZE_STRUCTURE_TYPE_DEVICE_IP_VERSION_EXT};
+    deviceProps.pNext = &devicePropsIP;
+    func_zeDeviceGetProperties(device, &deviceProps);
+    ip_version = devicePropsIP.ipVersion;
+    return status::success;
+}
+
+status_t get_lz_device_enabled_systolic_intel(
+        ze_device_handle_t device, bool &mayiusesystolic) {
```
echeresh, line 301: Nit, for consistency with the call args:
```suggestion
        ze_device_handle_t device, bool &mayiuse_systolic) {
```

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+
+status_t get_lz_device_enabled_native_float_atomics(
+        ze_device_handle_t device, uint64_t native_extensions) {
+    using namespace gpu::intel::compute;
+
+    ze_device_properties_t deviceProps = {ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES};
+    ze_float_atomic_ext_properties_t fltAtom
+            = {ZE_STRUCTURE_TYPE_FLOAT_ATOMIC_EXT_PROPERTIES};
+    deviceProps.pNext = &fltAtom;
+    func_zeDeviceGetProperties(device, &deviceProps);
```
echeresh, line 323: Missing status check, also please check other calls.

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+        native_extensions |= (uint64_t)native_ext_t::fp64_atomic_load_store;
+    if ((fltAtom.fp64Flags & atomic_add) == atomic_add)
+        native_extensions |= (uint64_t)native_ext_t::fp64_atomic_add;
+    if ((fltAtom.fp64Flags & atomic_add) == atomic_min_max)
+        native_extensions |= (uint64_t)native_ext_t::fp64_atomic_min_max;
+
+    return status::success;
+}
+
+status_t get_lz_device_eu_count(ze_device_handle_t device, int &eucount) {
```
echeresh, line 359: Nit: styling
```suggestion
status_t get_lz_device_eu_count(ze_device_handle_t device, int &eu_count) {
```

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,115 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+    // TODO: xelpg has no f64 support. check that the query properly handle that
+    ret = get_lz_device_enabled_native_float_atomics(device, native_extensions);
+    MAYBE_UNUSED(ret);
+
+    auto status
+            = jit::gpu_supports_binary_format(&mayiuse_ngen_kernels, engine);
+    if (status != status::success) mayiuse_ngen_kernels = false;
+
+    ip_version = 0;
+    get_device_ip(device, ip_version);
```
echeresh, line 396: Missing status check? Also it would nice to make `init_gpu_hw_info()` to return `status_t` to report errors back when needed.

echeresh (approved):

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: added status check and status propagation. Thanks.

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: Good point, added proper propagation of status_t to caller.

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,113 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+status_t get_l0_device_enabled_systolic_intel(
+        ze_device_handle_t device, bool &mayiuse_systolic) {
+    ze_device_module_properties_t deviceModProps
+            = {ZE_STRUCTURE_TYPE_DEVICE_MODULE_PROPERTIES};
+    // Note: supported by Intel Driver 24.05 and onwards
+    ze_intel_device_module_dp_exp_properties_t deviceModPropsExt
+            = {ZE_STRUCTURE_INTEL_DEVICE_MODULE_DP_EXP_PROPERTIES};
+    deviceModProps.pNext = &deviceModPropsExt;
+
+    CHECK(func_zeDeviceGetModuleProperties(device, &deviceModProps));
```
petercad, line 309: Do we have a driver support matrix? Users may have pretty old drivers (esp. on Linux), so I'm wondering if we need a fallback path here.

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,113 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+            = {ZE_STRUCTURE_TYPE_FLOAT_ATOMIC_EXT_PROPERTIES};
+    deviceProps.pNext = &fltAtom;
+    CHECK(func_zeDeviceGetProperties(device, &deviceProps));
+
+    ze_device_fp_atomic_ext_flags_t atomic_load_store
+            = ZE_DEVICE_FP_ATOMIC_EXT_FLAG_GLOBAL_LOAD_STORE
+            | ZE_DEVICE_FP_ATOMIC_EXT_FLAG_LOCAL_LOAD_STORE;
+    ze_device_fp_atomic_ext_flags_t atomic_add
+            = ZE_DEVICE_FP_ATOMIC_EXT_FLAG_GLOBAL_ADD
+            | ZE_DEVICE_FP_ATOMIC_EXT_FLAG_LOCAL_ADD;
```
petercad, line 330: No GPUs currently have floating point atomic adds in local memory, so we just need to know about global memory FP atomic support:
```suggestion
            = ZE_DEVICE_FP_ATOMIC_EXT_FLAG_GLOBAL_ADD;
```
I think a similar change makes sense for load/store and min/max.

mgouicem: I aligned with the `native_extension` definition that was used so far from OCL queries [here](https://github.com/oneapi-src/oneDNN/blob/316287183e2503a9db2d975dd04359728aabceb2/src/gpu/intel/ocl/ocl_utils.cpp#L394) (flag is set only if both local and global are available).

I am fine changing the semantic of `native_extensions` (it seems broken anyway as you mention). But that might have wider implications.

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: For backward compatibility, we validate only last driver available at time of release (see [README](https://github.com/oneapi-src/oneDNN/blob/main/README.md#validated-configurations)). In this particular case, this query appeared on Oct 31st 2023 ([99abb40a](https://github.com/intel/compute-runtime/commit/99abb40a45f51c8ea9779179e771c04a8be2777a))

For forward compatibility, I am working on getting clarification (and this PR will remain draft in the meantime).

src/gpu/intel/sycl/l0/utils.cpp
petercad: Good point. Then I think it makes sense to keep L0 and OCL aligned and change both in a future PR.

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: So I clarified the backward compatibility policy with Level Zero driver team. The extensions are backward compatible for a given major version of level zero API. So this is similar to regular (non-extensions) level zero APIs we currently use, and we should be fine with future driver updates.

However, this extension was introduced with [23.48.27912.11](https://github.com/intel/compute-runtime/releases/tag/23.48.27912.11) release, so this would become a requirement to our users. Do you think this could be an issue for some users?

petercad: > However, this extension was introduced with [23.48.27912.11](https://github.com/intel/compute-runtime/releases/tag/23.48.27912.11) release, so this would become a requirement to our users. Do you think this could be an issue for some users?

I don't have any hard data on the range of L0 driver versions users have in the wild, but it seems risky to drop support for drivers more than 1 year old, especially as we have been gracefully supporting much older drivers prior to this PR. At least, let's not fail engine/primitive initialization if this call fails. We could conservatively report "no systolic support" for these older drivers. What do you think?

src/gpu/intel/sycl/l0/utils.cpp
```diff
@@ -272,6 +287,114 @@ bool compare_ze_devices(const ::sycl::device &lhs, const ::sycl::device &rhs) {
...
+    LevelZeroCodeGenerator<HW::Unknown>::detectHWInfo(
+            context, device, hw, product);
+
+    gpu_arch = jit::convert_ngen_arch_to_dnnl(hw);
+    gpu_product_family = static_cast<int>(product.family);
+    stepping_id = product.stepping;
+
+    mayiuse_systolic = false;
+    CHECK(get_l0_device_enabled_systolic_intel(device, mayiuse_systolic));
+    // TODO: xelpg has no f64 support. check that the query properly handle that
```
dzarukin, line 386: Did you check this? :)

.github/automation/commit-msg-check.py
```diff
@@ -36,7 +37,7 @@ def __scopeCheck(msg: str):
     scopesArray = firstLine.split(":")[:-1]
 
     for scopes in scopesArray:
-        numWords = len(scopes.split())
+        numWords = len(re.split(',| ', scopes))
```
dzarukin, line 40: What does this suppose to do?

dzarukin (approved):

.github/automation/commit-msg-check.py
vpirogov: Let's drop this commit. The issue with checker is addressed in #2286.

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: Yes it was checked, I will drop the comment. :)

src/gpu/intel/sycl/l0/utils.cpp
mgouicem: + @vpirogov. Do we need/have a policy wrt to number of drivers oneDNN supports?

.github/automation/commit-msg-check.py
mgouicem: The new commit_msg_checker makes spaces between scopes mandatory. Is this something we want to enforce?

vpirogov: Looking at commits history space after `,` and `:` in scope is the common practice. All deviations are a part of 3% mentioned in #2286. So enforcing space just for the sake of consistency feels right.

petercad (approved):

mgouicem:
make test
disable device_cpu
enable device_gpu
enable thr_sycl
enable thr_ocl

densamoilov (approved):

src/gpu/intel/sycl/l0/utils.cpp
vpirogov: oneAPI policy is supporting the latest LTS and the latest rolling driver releases. I don't have concerns with dropping support for drivers >1 year old. Though I agree with @petercad's comment that we should fail gracefully if the driver is not supported. We could either assume that query returned negative or throw an error indicating that driver is not supported.

mgouicem: > Though I agree with @petercad's comment that we should fail gracefully if the driver is not supported. 

I guess it depends how we define gracefully. In the current state, the engine fails to create and an error message is printed upon the failure to query a hw property.
The alternative suggested by Peter is to not fail engine creation, but just proceed with disabled systolic support (the error message would still be printed).

vpirogov: Thanks for clarifications. I don't have preference between these two options.

vpirogov: As this is specific to SYCL configuration the change will not affect OpenVINO. Pytorch is relying on LTS driver, which does have queries support. With that we are ok to go with the changes as is.

vpirogov (approved):

PR #2246, opened 2024-12-11
Introduce GPU kernel metadata support

echeresh:
This PR introduces optional metadata support for GPU kernels. Metadata is stored in the binary form and may be used in an implementation-dependent manner.

The original use case originates from shapeless convolution:

- Shapeless kernel is reusable and it's created based on a so-called "descriptor"
- The descriptor is the only way to communicate kernel parameters to the kernel 
- In the opposite direction (from kernel creation back to the creation call site), there is no simple way to communicate anything back:
    - This is where kernel metadata is used, kernel creation goes through multiple steps where information about required kernel arguments is recorded. This information is then accessed by the caller to prepare kernel arguments accordingly (in the same order as registered originally)
- Examples of kernel arguments that we'd like to generate ad-hoc (instead of supplying them independently of kernel creation):
    - "Magic" values for integer division. Kernel grid indices contain fused dimension indices and we need to prepare "magic" values on the host and provide them to the kernel. Sometimes we need magic values of derivative sizes (like `div_up(ic, 32)`)
    - Some problem sizes don't need "magic" values: e.g. the outer kernel grid dimension is computed via inner dimension sizes
    - Some problem sizes/parameters are unused, e.g. for kernels supporting only 1D or 2D shapes

As a counter-point: we could replicate the kernel-side logic to figure out what parameters are needed (including magic values, etc) but it's non trivial (we need to repeat some logic to understand what divisions are to be used, what parameters are hard-coded, etc) and is generally more error-prone. The metadata-based approach is simpler in this regard: the kernel explicitly requests whatever it needs and arguments are prepared according to this.

### DIFF
--- a/src/common/cache_blob.hpp
+++ b/src/common/cache_blob.hpp
@@ -32,7 +32,7 @@ struct cache_blob_impl_t {
         : pos_(0), data_(data), size_(size) {}
 
     status_t add_binary(const uint8_t *binary, size_t binary_size) {
-        if (!binary || binary_size == 0) { return status::invalid_arguments; }
+        if (!binary && binary_size != 0) { return status::invalid_arguments; }
         if (pos_ + sizeof(binary_size) + binary_size > size_) {
             return status::invalid_arguments;
         }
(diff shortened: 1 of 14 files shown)


echeresh:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe2-lpg
enable arch_xe-hpg
enable arch_xe-hpc

PR #2249, opened 2024-12-11
gpu: generic: add simple SYCL reduction implementation

sgeor255:
# Description

This PR adds a fallback simple reduction implementations for cases not covered by #2243.

# Checklist

## General

### DIFF
--- a/src/gpu/generic/sycl/simple_reduction.cpp
+++ b/src/gpu/generic/sycl/simple_reduction.cpp
@@ -1,0 +1,57 @@
+#include "simple_reduction.hpp"
+
+#include "gpu/generic/sycl/engine.hpp"
+#include "gpu/generic/sycl/simple_reduction_kernels.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace gpu {
+namespace generic {
+namespace sycl {
+
+status_t simple_reduction_t::pd_t::init_conf() {
+    conf_.alg = desc()->alg_kind;
+    conf_.src_md = xpu::sycl::md_t(src_md());
+    conf_.dst_md = xpu::sycl::md_t(dst_md());
+    conf_.p = desc()->p;
+    conf_.eps = desc()->eps;
+
+    auto src_wrap = memory_desc_wrapper(src_md());
+    auto dst_wrap = memory_desc_wrapper(dst_md());
+    dst_nelems_ = dst_wrap.nelems();
+
+    const auto ndims = dst_wrap.ndims();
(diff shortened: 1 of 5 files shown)


src/gpu/generic/sycl/sycl_primitive_conf.hpp
```diff
@@ -415,6 +415,17 @@ struct sycl_pooling_bwd_conf_t : public sycl_pooling_base_conf_t {
...
 
+struct sycl_simple_reduction_conf_t {
+    dnnl_alg_kind_t alg = dnnl_alg_kind_undef;
+    xpu::sycl::md_t src_md;
+    xpu::sycl::md_t dst_md;
+    float p;
+    float eps;
+    sycl_post_ops_t post_ops;
+    dim_t reduce_dims[xpu::sycl::md_t::max_dims];
+    int reduce_size = 1;
```
ShanoToni, line 426: Any reason only this is initialised and not the floats?

src/gpu/generic/sycl/simple_reduction.cpp
```diff
@@ -0,0 +1,57 @@
...
+    CHECK(create_kernel(engine, kid, &kernel_));
+
+    return status::success;
+}
+
+status_t simple_reduction_t::execute(const exec_ctx_t &ctx) const {
+    auto dst_wrap = memory_desc_wrapper(pd()->dst_md());
+    return parallel_for(ctx, kernel_, [&](::sycl::handler &cgh) {
+        reduction_kernel_fwd_t reduction_kernel(pd()->conf_, cgh, ctx);
+        cgh.parallel_for(::sycl::range<1>(dst_wrap.nelems()), reduction_kernel);
```
ShanoToni, line 49: Instead of using `dst_wrap.nelems()` and creating a wrapper, this can be a member of the primitive set in the `init()`?

ShanoToni (approved):
A couple of small notes/ideas.

src/gpu/generic/sycl/sycl_primitive_conf.hpp
sgeor255: If you are referring to `p` and `eps`, those are set in `init_conf` to the values provided by the primitive desc, so they don't need to be explicitly initialised. `reduce_size` is used to accumulate the number of elements that are being reduced so it's convenient to initialise to 1 here so that it can directly be used in the loop.

src/gpu/generic/sycl/simple_reduction.cpp
sgeor255: Done

mgouicem (approved):

PR #2250, opened 2024-12-11
generic: sycl: Introduce spec constants for sycl matmul

ShanoToni:
# Description

Currently the sycl implementation of matmul has a kernel argument size error when run on intel hardware, the size of the kernel arguments could not be trivially reduced, requiring a more complex solution. This PR introduces a fix for this by introducing spec constants for the source, weights and destination memory descriptors. This fix should be temporary while the best solution would be to leverage SYCL RTC to handle the argument size error, when available.

Additionally, due to the kernels being compiled for the spec constants we cannot have the values of source, weights and destination md change. This would mean that the sycl implementation cannot support runtime dimensions currently.

# Checklist

## General

### DIFF
--- a/src/gpu/generic/sycl/matmul_kernels.hpp
+++ b/src/gpu/generic/sycl/matmul_kernels.hpp
@@ -18,6 +18,7 @@
 #define GPU_GENERIC_SYCL_MATMUL_KERNELS_HPP
 
 #include "common/primitive_exec_types.hpp"
+#include "gpu/generic/sycl/specialization_constants.hpp"
 #include "gpu/generic/sycl/sycl_io_helper.hpp"
 #include "gpu/generic/sycl/sycl_math_utils.hpp"
 #include "gpu/generic/sycl/sycl_post_ops.hpp"
@@ -358,7 +359,7 @@ struct matmul_kernel_fwd_t {
     matmul_kernel_fwd_t(const sycl_matmul_conf_t &conf, ::sycl::handler &cgh,
             const exec_ctx_t &ctx)
         : conf_(conf)
-        , data_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_SRC_0))
+        , src_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_SRC_0))
         , weights_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_WEIGHTS))
         , bias_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_BIAS))
         , dst_(CTX_INOUT_SYCL_KERNEL_MEMORY(DNNL_ARG_DST))
@@ -409,16 +410,23 @@ struct matmul_kernel_fwd_t {
                   CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_ATTR_DROPOUT_PROBABILITY))
         , po_args_(cgh, ctx, conf_.post_ops) {}
 
-    void operator()(::sycl::nd_item<1> item) const {
+    void operator()(::sycl::nd_item<1> item, ::sycl::kernel_handler kh) const {
         using data_block_t = register_block<register_block_M, register_block_K>;
         using weights_block_t
                 = register_block<register_block_K, register_block_N>;
         using dst_block_t = register_block<register_block_M, register_block_N>;
 
-        memory_tensor_t data_mem(data_, conf_.data_md);
-        memory_tensor_t weights_mem(weights_, conf_.weights_md);
+        // Get the value of the spec constant;
+        const auto &md_t_spec_const_pod_val = kh.get_specialization_constant<
+                detail::matmul::md_t_spec_const_id>();
+        const auto &src_md = md_t_spec_const_pod_val.data_md_t;
+        const auto &weights_md = md_t_spec_const_pod_val.weights_md_t;
+        const auto &dst_md = md_t_spec_const_pod_val.dst_md_t;
+
+        memory_tensor_t data_mem(src_, src_md);
+        memory_tensor_t weights_mem(weights_, weights_md);
         memory_tensor_t bias_mem(bias_, conf_.bias_md);
-        memory_tensor_t dst_mem(dst_, conf_.dst_md);
+        memory_tensor_t dst_mem(dst_, dst_md);
         memory_plain_t data_scale_mem(data_scale_, data_scales_dt_);
         memory_plain_t weights_scale_mem(weights_scale_, weights_scales_dt_);
         memory_plain_t dst_scale_mem(dst_scale_, dst_scales_dt_);
@@ -513,7 +521,11 @@ struct matmul_kernel_fwd_t {
             off_dst[matmul_dim_2] *= conf_.transpose_dst ? register_block_M
                                                          : register_block_N;
             int m = off_dst[conf_.transpose_dst ? matmul_dim_2 : matmul_dim_1];
-            int n = off_dst[conf_.transpose_dst ? matmul_dim_1 : matmul_dim_2];
+            // TODO: the following code is changed due to a correctness bug
+            // specific for PVC, needs further investigation and a better fix
+            // or explanation.
+            int n = off_dst[matmul_dim_2];
+            if (conf_.transpose_dst) { n = off_dst[matmul_dim_1]; }
 
             dims_t off_src, off_weights, off_bias;
             for (int i = max_supported_ndims - 1; i >= 0; i--) {
@@ -650,7 +662,7 @@ struct matmul_kernel_fwd_t {
 private:
     sycl_matmul_conf_t conf_;
 
-    xpu::sycl::in_memory_arg_t data_;
+    xpu::sycl::in_memory_arg_t src_;
     xpu::sycl::in_memory_arg_t weights_;
     xpu::sycl::in_memory_arg_t bias_;
     xpu::sycl::inout_memory_arg_t dst_;
--- a/src/gpu/generic/sycl/ref_matmul.cpp
+++ b/src/gpu/generic/sycl/ref_matmul.cpp
@@ -15,7 +15,14 @@
 *******************************************************************************/
 
 #include "gpu/generic/sycl/ref_matmul.hpp"
+#include "common/c_types_map.hpp"
 #include "gpu/generic/sycl/matmul_kernels.hpp"
+#include "gpu/generic/sycl/specialization_constants.hpp"
+#include "xpu/sycl/types.hpp"
+
+#define VCHECK_MATMUL(cond, msg, ...) \
+    VCONDCHECK(primitive, create, check, matmul, (cond), \
+            status::unimplemented, msg, ##__VA_ARGS__);
 
 namespace dnnl {
 namespace impl {
@@ -23,7 +30,7 @@ namespace gpu {
 namespace generic {
 namespace sycl {
 
-void ref_matmul_t::pd_t::init_conf() {
+status_t ref_matmul_t::pd_t::init_conf() {
     conf_ = sycl_matmul_conf_t();
 
     conf_.do_scale_data
@@ -49,19 +56,61 @@ void ref_matmul_t::pd_t::init_conf() {
     memory_desc_wrapper weights_d = weights_md();
     memory_desc_wrapper dst_d = dst_md();
     memory_desc_wrapper bias_d = weights_md(1);
-    for (const auto &mdw : {src_d, weights_d, dst_d, bias_d}) {
-        if (mdw.has_runtime_dims()) {
-            any_runtime_params_ = true;
-            return;
-        }
-    }
-    init_rt_conf(conf_, src_d, weights_d, dst_d, bias_d);
+    VCHECK_MATMUL(!utils::one_of(true, src_d.has_runtime_dims(),
+                          weights_d.has_runtime_dims(),
+                          dst_d.has_runtime_dims(), bias_d.has_runtime_dims()),
+            VERBOSE_RUNTIMEDIM_UNSUPPORTED);
+
+    return init_rt_conf(conf_, data_md_t, dst_md_t, weights_md_t, src_d,
+            weights_d, dst_d, bias_d);
 }
 
-void ref_matmul_t::pd_t::init_rt_conf(sycl_matmul_conf_t &conf,
+status_t ref_matmul_t::pd_t::init_rt_conf(sycl_matmul_conf_t &conf,
+        xpu::sycl::md_t_spec_const &data_md_t_,
+        xpu::sycl::md_t_spec_const &dst_md_t_,
+        xpu::sycl::md_t_spec_const &weights_md_t_,
         const memory_desc_wrapper src_d, const memory_desc_wrapper weights_d,
         const memory_desc_wrapper dst_d,
         const memory_desc_wrapper bias_d) const {
+
+    // Lambda because this function will not be used anywhere else
+    auto init_md_t_sc_from_md = [=](xpu::sycl::md_t_spec_const &md_t_sc,
+                                        const memory_desc_t *md) -> status_t {
+        constexpr int max_dims = 6;
+        using dim32_t = int32_t;
+
+        memory_desc_wrapper mdw(md);
+
+        VCHECK_MATMUL(mdw.format_kind() == format_kind::blocked,
+                VERBOSE_UNSUPPORTED_FORMAT_KIND);
+        VCHECK_MATMUL(
+                mdw.ndims() <= max_dims, VERBOSE_BAD_NDIMS, mdw, mdw.ndims());
+
+        const auto &blk = mdw.blocking_desc();
+
+        md_t_sc.data_type_ = mdw.data_type();
+#define CHECK_AND_ASSIGN(lhs, rhs) \
+    VCHECK_MATMUL((rhs) <= INT32_MAX, VERBOSE_BAD_PARAM, rhs); \
+    (lhs) = static_cast<dim32_t>(rhs)
+
+        CHECK_AND_ASSIGN(md_t_sc.ndims_, mdw.ndims());
+        CHECK_AND_ASSIGN(md_t_sc.offset0_, mdw.offset0());
+        CHECK_AND_ASSIGN(md_t_sc.inner_nblks_, blk.inner_nblks);
+
+        for (int d = 0; d < mdw.ndims(); d++) {
+            CHECK_AND_ASSIGN(md_t_sc.dims_[d], mdw.dims()[d]);
+            CHECK_AND_ASSIGN(md_t_sc.padded_dims_[d], mdw.padded_dims()[d]);
+            CHECK_AND_ASSIGN(
+                    md_t_sc.padded_offsets_[d], mdw.padded_offsets()[d]);
+            CHECK_AND_ASSIGN(md_t_sc.strides_[d], blk.strides[d]);
+            CHECK_AND_ASSIGN(md_t_sc.inner_blks_[d], blk.inner_blks[d]);
+            CHECK_AND_ASSIGN(md_t_sc.inner_idxs_[d], blk.inner_idxs[d]);
+        }
+#undef CHECK_AND_ASSIGN
+
+        return status::success;
+    };
+
     int matmul_dim_1 = ndims() - 2;
     int matmul_dim_2 = ndims() - 1;
 
@@ -73,7 +122,7 @@ void ref_matmul_t::pd_t::init_rt_conf(sycl_matmul_conf_t &conf,
                 data_md_copy.dims[matmul_dim_2]);
         conf.transpose_data = true;
     }
-    conf.data_md = xpu::sycl::md_t(&data_md_copy);
+    init_md_t_sc_from_md(data_md_t_, &data_md_copy);
 
     memory_desc_t weights_md_copy = *weights_d.md_;
     auto &weights_strides = weights_md_copy.format_desc.blocking.strides;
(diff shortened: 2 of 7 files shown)


tests/benchdnn/dnnl_common.hpp
```diff
@@ -310,7 +310,13 @@ int check_dnnl_status(dnnl_status_t status, const prb_t *prb, res_t *res) {
         case dnnl_unimplemented: {
             // Unconditionally set all Nvidia backend unimplemented cases as
             // not supported.
-            if (is_nvidia_gpu() || is_amd_gpu()) {
+            if (is_nvidia_gpu()
+                    || is_amd_gpu()
+#ifdef DNNL_SYCL_GENERIC
+                    // skip unimplemented configs for sycl impl
+                    || is_gpu()
+#endif
```
dzarukin, line 318: IIRC, the expectation was generic kernels cover oneDNN features completely, thus, all unimplemented cases should be properly reported by benchdnn and addressed by developers. Has this been changed?

@mgouicem, @vpirogov

src/xpu/sycl/types.hpp
```diff
@@ -146,6 +177,23 @@ struct md_t {
 #undef CHECK_AND_ASSIGN
     }
 
+    // This constructor is to be called inside the kernel
```
dzarukin, line 180: How do you control this copy constructor will be called only from the inside the kernel and not from other places?

src/xpu/sycl/types.hpp
```diff
@@ -146,6 +177,23 @@ struct md_t {
...
+    md_t(md_t_spec_const &mimicker) {
```
dzarukin, line 181:
```suggestion
    md_t(const md_t_spec_const &mimicker) : data_type_(...), ... {
        for () {
            ...
        }
    }
```

src/xpu/sycl/types.hpp
```diff
@@ -93,6 +93,37 @@ using in_memory_arg_t = memory_arg_t<::sycl::access::mode::read>;
...
+    dims32_t dims_;
+    dims32_t padded_dims_;
+    dims32_t padded_offsets_;
+    dim32_t offset0_;
+
+    dims32_t strides_;
+    dim32_t inner_nblks_;
+    dims32_t inner_blks_;
+    dims32_t inner_idxs_;
+};
```
dzarukin, line 119: // This struct cannot have a non trivial constructor, or any non trivial types, because...
// why exactly?
`static_assert(std::is_trivial_v<md_t_spec_const>, "md_t_spec_const must be trivial");`

tests/benchdnn/dnnl_common.hpp
Rbiessy: This is indeed the long term goal. As the generic backend is still a work in progress and new features are being added to oneDNN it is useful to be able to skip some configurations that are not supported yet.

src/xpu/sycl/types.hpp
ShanoToni: From the sycl specification [HERE](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_declaring_a_specialization_constant) we need the spec constant to be `device_copyable` [HERE](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable).

I can clarify this in the comment if you think that makes sense, or remove it all together?

Rbiessy: Looking into this a bit more DPC++ defines `SYCL_DEVICE_COPYABLE=1` so we may be able to avoid the need of `md_t_spec_const` if we specialize `is_device_copyable_v<md_t>` to return `true`.

tests/benchdnn/dnnl_common.hpp
ShanoToni: Removed it a bit prematurely, its added back in until a conclusion is reached

dzarukin: I don't see any unimplemented hits in current validation. It makes me think if there are any new unimplemented cases, they must be local to these changes. I can provide more guidance if there's a pointer on or a snapshot of what's going on.

src/xpu/sycl/types.hpp
dzarukin: All the knowledge not available from reading the code better be put in a comment as a reference with exact spot to look or as a summary.

tests/benchdnn/dnnl_common.hpp
ShanoToni: Without this change running the oneDNN compiled with generic vendor (sycl ref matmul used) causes benchdnn to fail with test unimplemented error when testing with runtime dims, spec constants require kernel to be compiled for the specific spec constants (specific sizes for src/dst/weights) so current approach does now support runtime dims.
We are wondering if this can should be reported as skipped by benchdnn, as it is expected.

src/gpu/generic/sycl/ref_matmul.cpp
```diff
@@ -52,16 +55,56 @@ void ref_matmul_t::pd_t::init_conf() {
     for (const auto &mdw : {src_d, weights_d, dst_d, bias_d}) {
         if (mdw.has_runtime_dims()) {
             any_runtime_params_ = true;
-            return;
+            return status::unimplemented;
```
mgouicem, line 58: (minor) please  don't hesitate to use [verbose macros](https://github.com/oneapi-src/oneDNN/blob/0bff5bd0f56dcf5e68303198996e15a065ffb16d/src/common/verbose.hpp#L87-L159) for this kind of checks.

src/gpu/generic/sycl/ref_matmul.cpp
```diff
@@ -52,16 +55,56 @@ void ref_matmul_t::pd_t::init_conf() {
...
+    // Lambda because this function will not be used anywhere else
+    auto init_md_t_sc_from_md = [=](xpu::sycl::md_t_spec_const &md_t_sc,
+                                        const memory_desc_t *md) -> void {
+        constexpr int max_dims = 6;
+        using dim32_t = int32_t;
+
+        memory_desc_wrapper mdw(md);
+
+        assert(mdw.format_kind() == format_kind::blocked);
+        assert(mdw.ndims() <= max_dims);
```
mgouicem, line 83: given that this function can fail, I would advocate to return a status instead of using asserts, and propagate by it in init_conf.

src/gpu/generic/sycl/ref_matmul.cpp
```diff
@@ -137,30 +180,38 @@ void ref_matmul_t::pd_t::init_rt_conf(sycl_matmul_conf_t &conf,
...
-        const auto weights_d
-                = ctx.memory_mdw(DNNL_ARG_WEIGHTS, pd()->weights_md());
-        const auto dst_d = ctx.memory_mdw(DNNL_ARG_DST, pd()->dst_md());
-        const auto bias_d = ctx.memory_mdw(DNNL_ARG_BIAS, pd()->weights_md(1));
-        pd()->init_rt_conf(conf, src_d, weights_d, dst_d, bias_d);
-    }
-
     parallel_for(ctx, kernel_, [&](::sycl::handler &cgh) {
-        matmul_kernel_fwd_t matmul_kernel(conf, cgh, ctx);
+        matmul_kernel_fwd_t matmul_kernel(pd()->conf_, cgh, ctx);
```
mgouicem, line 209: As we discussed, using specialization constants will trigger recompilation during execute function.
TBH, I am a bit bothered by the precedent this would set, as multiple users rely on the property that after primitive creation, there should be no recurrent jitting overheads. Are we sure there is no other venue to avoid the argument size limitation?

tests/benchdnn/dnnl_common.hpp
dzarukin: This sounds to me like a matmul-specific skip.
I propose to introduce `is_generic_gpu()` method identically to `is_nvidia_gpu()`, and put a check [here](https://github.com/oneapi-src/oneDNN/blob/main/tests/benchdnn/matmul/matmul.cpp#L556) (before the if() {} section).

The change in this PR will silence ALL drivers with ANY unimplemented hit which is not the desired behavior in a long term.
The practice to mark unimplemented as skipped is to figure out the minimal scope all of them fit and skip only for it. In your case is sounds like any runtime dimensions involved. You may relate to [this statement](https://github.com/oneapi-src/oneDNN/blob/main/tests/benchdnn/matmul/matmul.cpp#L585-L587) when crafting it

src/gpu/generic/sycl/matmul_kernels.hpp
```diff
@@ -409,16 +410,23 @@ struct matmul_kernel_fwd_t {
                   CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_ATTR_DROPOUT_PROBABILITY))
         , po_args_(cgh, ctx, conf_.post_ops) {}
 
-    void operator()(::sycl::nd_item<1> item) const {
+    void operator()(::sycl::nd_item<1> item, ::sycl::kernel_handler kh) const {
```
dzarukin, line 413:
```suggestion
    void operator()(::sycl::nd_item<1> item, const ::sycl::kernel_handler &kh) const {
```
?

src/gpu/generic/sycl/matmul_kernels.hpp
```diff
@@ -409,16 +410,23 @@ struct matmul_kernel_fwd_t {
...
         using dst_block_t = register_block<register_block_M, register_block_N>;
 
-        memory_tensor_t data_mem(data_, conf_.data_md);
-        memory_tensor_t weights_mem(weights_, conf_.weights_md);
+        // Get the value of the spec constant;
+        auto md_t_spec_const_pod_val = kh.get_specialization_constant<
+                detail::matmul::md_t_spec_const_id>();
+        auto data_md = md_t_spec_const_pod_val.data_md_t;
+        auto weights_md = md_t_spec_const_pod_val.weights_md_t;
+        auto dst_md = md_t_spec_const_pod_val.dst_md_t;
```
dzarukin, line 424: `const auto &` to remove unnecessary copies?

src/gpu/generic/sycl/matmul_kernels.hpp
```diff
@@ -409,16 +410,23 @@ struct matmul_kernel_fwd_t {
...
+
+        memory_tensor_t data_mem(data_, data_md);
```
dzarukin, line 426: Nit: for matmul it's called `src`, not `data`.

src/gpu/generic/sycl/ref_matmul.cpp
```diff
@@ -52,16 +55,56 @@ void ref_matmul_t::pd_t::init_conf() {
     for (const auto &mdw : {src_d, weights_d, dst_d, bias_d}) {
         if (mdw.has_runtime_dims()) {
             any_runtime_params_ = true;
-            return;
+            return status::unimplemented;
         }
     }
-    init_rt_conf(conf_, src_d, weights_d, dst_d, bias_d);
+    init_rt_conf(conf_, data_md_t, dst_md_t, weights_md_t, src_d,
+            weights_d, dst_d, bias_d);
```
dzarukin, line 62:
```suggestion
    CHECK(init_rt_conf(conf_, data_md_t, dst_md_t, weights_md_t, src_d,
            weights_d, dst_d, bias_d));
```

src/gpu/generic/sycl/ref_matmul.hpp
```diff
@@ -64,21 +66,30 @@ struct ref_matmul_t : public gpu::generic::sycl::primitive_t {
                     && md_dims_in_range(weights_md());
             if (!ok) return status::unimplemented;
 
-            init_conf();
-            return status::success;
+            auto init_status = init_conf();
+
+            return init_status;
```
dzarukin, line 71:
```suggestion
            return init_conf();
```

src/xpu/sycl/types.hpp
```diff
@@ -146,6 +177,21 @@ struct md_t {
 #undef CHECK_AND_ASSIGN
     }
 
+    md_t(const md_t_spec_const &mimicker)
```
dzarukin, line 180: Nit: 
```suggestion
    md_t(const md_t_spec_const &other)
```

src/gpu/generic/sycl/ref_matmul.cpp
Rbiessy: I don't follow, we discussed that the spec constant would only trigger recompilation if we supported runtime dimensions. We discussed it does not make sense to support runtime dimensions so in the current patch the kernel compilation should only happen in the `init` function.
From what I gathered there are no great solutions in the short term:

- Reducing the amount of arguments given to the kernel. We already use a struct as an argument to only give the arguments that may be used in the SYCL kernels.
- Reducing the number of post-op supported: we already support "only" 5 post-ops. I wouldn't mind reducing this further but this may disable more tests. My understanding was that we didn't want to reduce this further.

The long term solution is to use SYCL-RTC but this will be a large task. The spec constants are useful to be able to run more tests with the SYCL backend until SYCL-RTC is implemented.

mgouicem: > I don't follow, we discussed that the spec constant would only trigger recompilation if we supported runtime dimensions. We discussed it does not make sense to support runtime dimensions

I guess I missed that part. Given that this is a reference implementation, I would expect that at some point it would support runtime dimensions. Solving the current argument size limitation issue with specialization_constant would defer runtime dimension support until SYCL-RTC is enabled. Is that something we are fine with?

> Reducing the amount of arguments given to the kernel. We already use a struct as an argument to only give the arguments that may be used in the SYCL kernels.

IIUC, sycl_matmul_conf_t is used to pass argument shapes to the kernel right?
If so, it seems to use [sycl_post_ops_t](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_primitive_conf.hpp#L124) which is a [union type](https://github.com/oneapi-src/oneDNN/blob/main/src/gpu/generic/sycl/sycl_post_ops.hpp#L252), for which the biggest members are [prelu_post_op](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_post_ops.hpp#L256) with ~100 Bytes and [binary_post_op](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_post_ops.hpp#L170) with ~160 Bytes. So 5 post_ops gets us to ~800B, do I get that right? 
(which is about 1/4 of argument size limit).

In any case, here are a few suggestions:
- for prelu_post_op, using dim_t[xpu::sycl::md_t::max_dims] instead of dims_t for the [strides array](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_post_ops.hpp#L219) would half that size.
- for binary_post_op, we could use a couple of arrays for dims and strides instead of using md_t? If if we want to stick with sycl::md_t, we have about 50B related to blocking (inner_block, [here](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/xpu/sycl/types.hpp#L222-L224)). Do we actually use blocked layouts in generic sycl kernels? If not that would shape about 250B from post_ops, and ~500B from [matmul_conf](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_primitive_conf.hpp#L114).
- separate post-ops that require extra md from the others, so that we can reduce the number of post-ops that require passing md to kernel (e.g. binary/prelu), while not restricting others (e.g. eltwise).

> Reducing the number of post-op supported: we already support "only" 5 post-ops. I wouldn't mind reducing this further but this may disable more tests. My understanding was that we didn't want to reduce this further.

Correct 5 is already low. However it seems we are using a union for all postops. Splitting eltwise from binary/prelu postop could help limit only the number of binary/prelu post-ops (e.g. to 2 binary/prelu post-ops), while not restricting the number of eltwise post-ops. That could be an option as well.

Rbiessy: Thanks for the detailed suggestion. From previous investigations it did not seem possible to reduce enough the arguments size but we could investigate again. I'm not confident we can remove enough members from the binary post op.
Just one clarification, the argument size limit is 2kB so the 5 post-ops would represents about 40% of the total size. We need to shave off a total of 282B or at least 57B per post-op. I haven't looked into this myself but this seems difficult just for the binary post op which is apparently the biggest issue here.

```quote
Solving the current argument size limitation issue with specialization_constant would defer runtime dimension support until SYCL-RTC is enabled. Is that something we are fine with?
```

This sounds fine to me. From what I understand runtime dimensions are not that important if they are just meant for optimization purposes but the SYCL backend is not optimized yet. At this point it is easier to use spec-constant. This is just a workaround for us to be able to run tests and have a clearer picture of what operations are working as expected.

src/gpu/generic/sycl/ref_matmul.cpp
ShanoToni: Great suggestion, made the change and prefer it this way as well, thank you

src/gpu/generic/sycl/ref_matmul.cpp
ShanoToni: Asserts changed to verbose checks, propagating status_t now as well.

src/gpu/generic/sycl/matmul_kernels.hpp
ShanoToni: The kernel_handler cannot be const because the `get_specialization_constant()` is not marked caused and would throw an error.

src/gpu/generic/sycl/matmul_kernels.hpp
ShanoToni: Unneeded copies removed.

src/gpu/generic/sycl/matmul_kernels.hpp
ShanoToni: `data` renamed to `src`

src/gpu/generic/sycl/ref_matmul.cpp
ShanoToni: Check was not needed as currently function used to assert, now changed to return status and added check

src/gpu/generic/sycl/ref_matmul.hpp
ShanoToni: Suggested change made.

src/xpu/sycl/types.hpp
ShanoToni: `other` sounds better, change made.

src/gpu/generic/sycl/ref_matmul.cpp
AD2605: Hi @mgouicem , Thank you for your detailed suggestions. The motivation for going ahead with spec-constants was to keep the changes as minimal as possible, which would be fairly simple to revert and do not involve considerable refactoring / changes for one specific kernel which lead to code duplication, or changing the "core" of oneDNN, i.e. the de-facto way of doing things in oneDNN. The motivation of avoiding this is because we expect all of these changes to be reverted once SYCL-RTC is in place.

> In sycl::md_t, we have about 50B related to blocking (inner_block, [here](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/xpu/sycl/types.hpp#L222-L224)). Do we actually use blocked layouts in generic sycl kernels? If not that would shape about 250B from post_ops, and ~500B from [matmul_conf](https://github.com/oneapi-src/oneDNN/blob/36c227a9dcf2c66e2a61ad7d063f53a56acb456f/src/gpu/generic/sycl/sycl_primitive_conf.hpp#L114).

 I believe that the i8 matmuls which typically go to cublasLT, which expects data in a blocked formats, falls back to the generic kernel for the `binary_post_op`. Hence the support of blocking over there.

> separate post-ops that require extra md from the others, so that we can reduce the number of post-ops that require passing md to kernel (e.g. binary/prelu), while not restricting others (e.g. eltwise).

Since the post_ops_t struct is used by other reference primitives as well, this would mandate a change in them as well, and if something similar were to be developed just for the matmul kernel, it would lead to a lot of code duplication, hence it may not be viable. 

> for binary_post_op, we could use a couple of arrays for dims and strides instead of using md_t?

The kernel uses the `off_v` and the `off_v_masked` methods to calculate certain offsets, and those functions take care of padding if any etc etc. So that logic would need to be copied into another function if we were to isolate these two arrays into a struct of their own

If specialization constants is not preferable solution to this problem, we were also thinking of moving the kernel arguments to global memory. This would not trigger a re-compilation of the kernel when runtime-dimensions are used, and would have minimal effect on the kernel.

mgouicem: > I believe that the i8 matmuls which typically go to cublasLT, which expects data in a blocked formats, falls back to the generic kernel for the binary_post_op. Hence the support of blocking over there.

cublasLT blocking does not depend on blocking_desc structure (which contain inner_blk), but on dedicated [cublasLT_blocking_desc](https://github.com/oneapi-src/oneDNN/blob/611f2fecc87d89cc618496c5149acd062ba8c68f/src/common/memory_desc.hpp#L164), which is essentially an enum and does not require inner_blk array.

> Since the post_ops_t struct is used by other reference primitives as well, this would mandate a change in them as well.

Not sure I see the required changes to primitives themselves. Here is my thinking: the post_ops_t structure is initialized using the primitive_attr directly (see [here](https://github.com/oneapi-src/oneDNN/blob/611f2fecc87d89cc618496c5149acd062ba8c68f/src/gpu/generic/sycl/sycl_post_ops.hpp#L295)), so I don't expect a change of internal API here, but just a change in implementation of sycl_post_ops_t (e.g. instead of containing an array of union [here](https://github.com/oneapi-src/oneDNN/blob/611f2fecc87d89cc618496c5149acd062ba8c68f/src/gpu/generic/sycl/sycl_post_ops.hpp#L348), it would contain one array for lightweight eltwise_po, one small array for bigger binary_po/prelu, and maybe an extra array for indices to interleave these two).

openmodelzooci:

vpirogov:
Closing as stale.

PR #2252, opened 2024-12-12
graph, example, benchdnn: add int4 gated mlp

TaoLv:
1. Define Gated MLP patterns with int4 compressed weights.
2. Add the example and benchdnn test file accordingly.

### DIFF
--- a/examples/graph/gated_mlp_int4.cpp
+++ b/examples/graph/gated_mlp_int4.cpp
@@ -1,0 +1,357 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cassert>
+#include <chrono>
+#include <iomanip>
+#include <iostream>
+#include <memory>
+#include <random>
+#include <string>
+#include <vector>
+
+#include "oneapi/dnnl/dnnl.hpp"
+#include "oneapi/dnnl/dnnl_graph.hpp"
+
+#include "graph_example_utils.hpp"
+
+using namespace dnnl;
+using tag = memory::format_tag;
+
+using namespace dnnl::graph;
+using data_type = logical_tensor::data_type;
+using layout_type = logical_tensor::layout_type;
+using dim = logical_tensor::dim;
+using dims = logical_tensor::dims;
+
+struct mlp_dims_t {
+    dim mb;
+    dim ic;
+    dim oc;
+    dim gr; // group size for int4 group quantization
+};
+
+static const int min_runs = 4;
+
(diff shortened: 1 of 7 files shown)


src/graph/backend/dnnl/patterns/mlp.cpp
```diff
@@ -113,6 +113,99 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, gated_mlp_v1)
...
+*/
+DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, quantized_gated_mlp)
+        .set_priority(22.1f)
+        .set_kind(partition_kind_t::matmul_post_ops)
+        .set_attr<FCreatePattern>("FCreatePattern",
+                [](const std::shared_ptr<pb_graph_t> &pgraph) -> void {
+                    pm::pb_op_t *deq_up = pgraph->append_op(
+                            graph::op_kind::DynamicDequantize);
+                    pm::pb_op_t *deq_gate = pgraph->append_op(
+                            graph::op_kind::DynamicDequantize);
```
mgouicem, line 137: Should those be restricted to that same `in_edge`?

TaoLv: Hi Mourad, the dequantize ops here are for weights which are different for the two matmul ops (up and gate). But I guess you meant the src should be same for the two matmul ops. Currently, the pattern and the primitive based implementation support different src inputs for the two matmul ops. Do you think we'd better add a restriction for that?

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/patterns/mlp.cpp
mgouicem: Thanks Tao. Yes I meant the src. IIRC, the gated MLP pattern always have shared src, correct? 
I guess the question is hence the actual pattern definition: if optimizations can leverage that src is shared by both matmul (e.g. by keeping reordered buffer in SLM on GPU), it would be safer to define the pattern as such IMHO (and reflect that in tests as well).

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/patterns/mlp.cpp
TaoLv: Thanks. The restriction is added for the patterns now.

ElaineBao (approved):

examples/graph/gated_mlp_int4.cpp
```diff
@@ -0,0 +1,357 @@
+/*******************************************************************************
+* Copyright 2024 Intel Corporation
```
dzarukin, line 2:
```suggestion
* Copyright 2025 Intel Corporation
```

tests/benchdnn/inputs/graph/complex_fusion/harness_mlp_all
```diff
@@ -2,3 +2,6 @@
 
 # WA: use smaller problem to shape to pass correctness check for f32 on pvc.
 --reset --in-shapes=0:1x128+1:128x256+4:128x256+13:256x128 --case=complex_fusion/mlp/gated-mlp-f32.json
+
+# f16-int4 case
+--reset --case=complex_fusion/mlp/gated-mlp-int4.json
```
dzarukin, line 7: You may use `--batch=complex_fusion/harness_mlp_ci` in here and collect modified shapes down below the list.

dzarukin (approved):

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

PR #2257, opened 2024-12-12
xe: jit: use F pipe for zero initialization

echeresh:
PR fixes MFDNN-12763.

The original regression is introduced by https://github.com/oneapi-src/oneDNN/commit/ee32525d4f.

### DIFF
--- a/src/gpu/intel/jit/codegen/codegen.cpp
+++ b/src/gpu/intel/jit/codegen/codegen.cpp
@@ -522,7 +522,7 @@ private:
     void fill_buf(const ngen_operand_t &buf_op, int size,
             const ngen_operand_t &pattern = {}) const {
         auto &rd = buf_op.reg_buf_data();
-        type_t type = type_t::u32();
+        type_t type = (pattern.is_invalid() ? type_t::f32() : type_t::u32());
         int grf_size = ngen::GRF::bytes(hw);
         int step = 2 * grf_size;
         for (int i = 0; i < size; i += step) {


echeresh:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe2-lpg
enable arch_xe-hpg
enable arch_xe-hpc

echeresh:
make test perf-gpu
set primitive=reorder sum concat binary conv deconv pool

src/gpu/intel/jit/codegen/codegen.cpp
```diff
@@ -522,7 +522,7 @@ class ir_to_ngen_t : public ir_visitor_t {
     void fill_buf(const ngen_operand_t &buf_op, int size,
             const ngen_operand_t &pattern = {}) const {
         auto &rd = buf_op.reg_buf_data();
-        type_t type = type_t::u32();
+        type_t type = (pattern.is_invalid() ? type_t::f32() : type_t::u32());
```
rjoursler, line 525: Would it be reasonable to use different types based on the type of the buffer at the IR level?

rjoursler (approved):

src/gpu/intel/jit/codegen/codegen.cpp
echeresh: IR buffers don't have a type so it's not available here. But here we don't really want that. For zero-initialization it's better to use `F` pipe as prologue has a lot of arithmetic operations involving `I` and `L` pipes. So even though it's int8 convolution (and the buffer is used as a dword buffer) using `F` pipe should be more beneficial as we offload some work to the less utilized `F` pipe resulting in less stalls.

kealan-barbieri (approved):

echeresh:
make test perf-gpu
set primitive=reorder sum concat binary conv deconv pool

PR #2259, opened 2024-12-12
Restricted AArch64 CI to relevant changes

vpirogov:
There's no need to run AArch64 CI on changes that are unlikely to affect this platform.

### DIFF
--- a/.github/workflows/ci-aarch64.yml
+++ b/.github/workflows/ci-aarch64.yml
@@ -60,7 +60,7 @@ jobs:
       matrix:
         config: [
           { name: MacOS, label: macos-14, threading: SEQ, toolset: clang, build: Release, testset: SMOKE },
-          { name: c6g, label: ah-ubuntu_22_04-c6g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI }, 
+          { name: c6g, label: ah-ubuntu_22_04-c6g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI },
           { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: clang, build: Debug, testset: SMOKE },
           { name: c7g, label: ah-ubuntu_22_04-c7g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI }
         ]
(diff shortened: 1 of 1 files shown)


vpirogov:
The issue with approach taken in this PR is that Github still considers AArch64 CI checks required and prevents PR from being promoted.

vpirogov:
@theComputeKid, any ideas here?

theComputeKid:
One solution (but not ideal) is that we add a step in the AArch64 pipeline to read the label and then make a decision based on that to quit successfully early.

Drawbacks:
1. We will still spin up an instance for a few seconds
2. It will still book a slot in the Q
3. It will label the AArch64 as passed even though it was skipped

Advantages:
1. We will only spin up an instance for a few seconds instead of the whole run.


I'm working on a PR to reduce the configs AArch64, should have that out shortly.

mgouicem:
Stackoverflow suggests to create an action with same name but with opposite condition that would always run successfully (see [here](https://stackoverflow.com/questions/69348532/github-actions-required-status-check-doesnt-run-due-to-files-in-paths-not-chan))

vpirogov:
> Stackoverflow suggests to create an action with same name but with opposite condition that would always run successfully (see [here](https://stackoverflow.com/questions/69348532/github-actions-required-status-check-doesnt-run-due-to-files-in-paths-not-chan))

This approach is very limited as 'opposite condition' cannot always be described using trigger path syntax. Conditions suggested in the answer with `path: xxx` and `path-ignore: xxx` are mutually exclusive when exactly one path mask is used. This is discussed further down the thread.

vpirogov:
I would suggest that at this point we make sure that AArch64 CI runs only for relevant changes and figure out how to make it required later. Github developers [are aware](https://github.com/orgs/community/discussions/44490) of our pain.

We are taking the same approach for x64 CI: #2281. I expect codeowners (i.e. people allowed to merge PRs) would use CI as a tool to prevent breaking changes.

theComputeKid:
Does this mean that the CI will not run if not needed, but then also the "required check" will not be satisfied and an override will be needed?

theComputeKid:
I was playing around with a script that would, in the first step:

1. Detect if CI is needed
2. Export an environment variable that propagates that information

Then each further step would skip successfully if that env variable is present. It is over-engineering for the problem but it would get the job done gracefully.

vpirogov:
> Does this mean that the CI will not run if not needed, but then also the "required check" will not be satisfied and an override will be needed?

Yes, required check will not work with this setup. I believe from developer quality of life perspective it's more important to get this part completed than figuring out how to enable required checks. For example currently broken AArch64 CI would block all PRs from being merged to the repo.

theComputeKid (approved):
Whoever merges now needs to check whether a failure in CI is a genuine failure, or caused by skipping CI.

vpirogov:
> Whoever merges now needs to check whether a failure in CI is a genuine failure, or caused by skipping CI.

Whoever merges a PR needs to check whether it introduces failures in CI. Skipping CI will not flag any issues.

PR #2263, opened 2024-12-13
github: workflows: Streamline AArch64 CI

theComputeKid:
- Reduction in runners to increase throughput
- Shift to CMake for ACL
- Test CI test-set instead of smoke to improve coverage
- Linux runners now do not depend on MacOS

### DIFF
--- a/.github/workflows/ci-aarch64.yml
+++ b/.github/workflows/ci-aarch64.yml
@@ -34,40 +34,60 @@ concurrency:
 permissions: read-all
 
 jobs:
-  macos:
-    runs-on: macos-14
+  build-and-test:
     strategy:
       matrix:
         config: [
-          { toolset: clang, build: Release },
-          { toolset: gcc, build: Release }
+          { name: MacOS, label: macos-14, threading: SEQ, toolset: clang, build: Release, testset: SMOKE },
+          { name: c6g, label: ah-ubuntu_22_04-c6g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI }, 
+          { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: clang, build: Debug, testset: SMOKE },
+          { name: c7g, label: ah-ubuntu_22_04-c7g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI }
         ]
 
-    name: MacOS, ${{ matrix.config.toolset }}, ${{ matrix.config.build }}
+    name: ${{ matrix.config.name }}, ${{ matrix.config.toolset }}, ${{ matrix.config.threading }}, ${{ matrix.config.build }}
+    runs-on: ${{ matrix.config.label }}
     steps:
+
       - name: Checkout oneDNN
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           path: oneDNN
 
-      - name: Install Scons
-        uses: threeal/pipx-install-action@b0bf0add7d5aefda03a3d4e47d651df807889e10 # v1.0.0
+      - name: Get latest CMake and Ninja
+        uses: lukka/get-cmake@5979409e62bdf841487c5fb3c053149de97a86d3 # v3.31.2
         with:
-          packages: scons
+          cmakeVersion: 3.31.0
+          ninjaVersion: 1.12.0
+
+      - if: ${{ (startsWith(matrix.config.label,'ah-ubuntu') && (matrix.config.threading == 'OMP')) }}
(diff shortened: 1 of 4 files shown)


theComputeKid:
cc: @vpirogov @mgouicem

vpirogov:
This PR changes names of checks marked as 'required'. This will result in PRs including this commit to be blocked by these required checks. Updating the names of required checks will result in PRs not including this commit (for example ones targeting release branches) to be blocked. 

TLDR: Changing required check names is bad and should be planned carefully.

theComputeKid:
In order to reduce load from the CI, or any modifications that we need in the future, we would need to change the required checks. Otherwise, we cannot iterate improvements to CI. Also, I have to adjust to the increasing load the runners will now face.

theComputeKid:
Wait, I have an idea to solve this problem: we can create a final job that runs when all CI for AArch64 have passed, and that final job immediately passes, and we make that final job a required check. That way we can change the intermediary steps without worrying.

Thoughts @vpirogov ?

vpirogov:
Yep, this would work.

vpirogov:
> In order to reduce load from the CI, or any modifications that we need in the future, we would need to change the required checks. Otherwise, we cannot iterate improvements to CI. Also, I have to adjust to the increasing load the runners will now face.

I understand that the changes are necessary in this case. I'm just highlighting consequences of this change that need to be managed.

vpirogov (approved):

dzarukin:
Hey! I've just noticed that `ctest` would use both output options (last two):
`ctest -j3 --no-tests=error --verbose --output-on-failure ...`
This leads to a situation when the multi-million number of strings are dumped and it's impossible to find a failed test as browser can't load all the data to use a search line. Could you, please, remove `--verbose` out of here? Or if you have any other better solutions to finding an issue much faster, feel free to propose it. Thanks.

[Reference](https://github.com/oneapi-src/oneDNN/actions/runs/12324636883/job/34403322093?pr=2270) for a log where it's impossible to find an issue.

theComputeKid:
@dzarukin are you looking at the latest version of the code? I took out verbose after leaving it in for one of the drafts, for the reasons you mentioned. See here: https://github.com/oneapi-src/oneDNN/actions/runs/12353711166/job/34473404457?pr=2263

And here: https://github.com/oneapi-src/oneDNN/blob/e498aeabce71011e6975f40fa97be73d9a9fcf97/.github/automation/test_aarch64.sh#L69

dzarukin:
> @dzarukin are you looking at the latest version of the code? I took out verbose after leaving it in for one of the drafts, for the reasons you mentioned. See here: https://github.com/oneapi-src/oneDNN/actions/runs/12353711166/job/34473404457?pr=2263
> 
> And here:
> 
> https://github.com/oneapi-src/oneDNN/blob/e498aeabce71011e6975f40fa97be73d9a9fcf97/.github/automation/test_aarch64.sh#L69

Ah, nice, thanks. I'd still keep `--output-on-failure` so that it's visible what failed without reproducing tests to see the actual problem.

theComputeKid:
@dzarukin your wish is my command
https://github.com/oneapi-src/oneDNN/blob/674767abfce5c564df67db66a5fdf21c49c1f8fd/.github/automation/test_aarch64.sh#L69

mgouicem (approved):

.github/workflows/ci-aarch64.yml
```diff
@@ -201,21 +124,37 @@ jobs:
...
         env:
-          BUILD_TOOLSET: ${{ matrix.toolset }}
-          CMAKE_BUILD_TYPE: ${{ matrix.config }}
+          BUILD_TOOLSET: ${{ matrix.config.toolset }}
+          CMAKE_BUILD_TYPE: ${{ matrix.config.build }}
           DYLD_LIBRARY_PATH: ${{ github.workspace }}/ComputeLibrary/build
-          ONEDNN_THREADING: ${{ matrix.threading }}     
+          ONEDNN_THREADING: ${{ matrix.config.threading }}     
+
+  AArch64:
```
vpirogov, line 155: I updated required checks in the repo to use common subset between this and previous versions. Change to 'AArch64' will be done at later date to avoid blocking PRs that do not include changes from this one.

vpirogov:
It would still be awesome to figure out how to skip AArch64 tests for unrelated changes (think `src/gpu` and `src/x64` folders).

.github/workflows/ci-aarch64.yml
theComputeKid: Sounds good.

theComputeKid:
@vpirogov I'm going to look into that (plus play around with your PR) next.

PR #2265, opened 2024-12-13
xe: jit: gemm: remove unsupported GPU architecture strategies

rjoursler:
Removes Gen9 startegies as Gen9 GPUs are no longer supported.
Removes Gen11 strategies as Gen11 GPU's only appears on unsupported 10th generation processors
Removes XeHP strategies as that architecture was never launched.

Partially reverts 6b13988ede3564b97fad173341a8e899ce75b1cd as many of the modified strategies are on architectures that are not under active development. The fact that these strategies were changed is likely an issue with the upstream repo. In particular, one of the XeLP modifications triggered an out of registers issue on the following workload:

 
    benchdnn --matmul --engine=gpu --dt=bf16:s4:bf16
    --attr-scales=wei:per_ocic:bf16:32x1 --attr-zero-points=wei:common:1:u8
    --attr-fpmath=bf16:true 3x96x96:3x96x64


petercad (approved):

atkassen (approved):

rjoursler:
make test
disable device_cpu
enable device_gpu

PR #2267, opened 2024-12-13
src: gpu: intel: jit: conv: add reorder-based precomputed zero points

hidefromkgb:
# Description

This is a novel approach to calculating zero points ahead of time, appending the precalc results to the convolutions weights buffers when they are either uploaded to the GPU or reordered on the GPU to fit the layout requested by the conv primitive.

Perf results on YOLO-v8n:
[ATSM/DG2]
FP16 — 3.3633792 ms/iter
INT8+ZP — 2.6763831 ms/iter
INT8 no ZP — 2.4270376 ms/iter
[BMG]
FP16 — 1.46378 ms/iter
INT8+ZP — 1.42166 ms/iter
INT8 no ZP — 1.2961 ms/iter

Addresses MFDNN-12548.

## General


src/gpu/gpu_reorder_pd.hpp
```diff
@@ -62,9 +69,99 @@ struct gpu_reorder_pd_t : public reorder_pd_t {
...
     }
+
+    status_t maybe_build_zp_precompute_conv_pd(impl::engine_t *dst_engine) {
+        memory_desc_wrapper dst_mdw(dst_md());
+        auto &extra = dst_mdw.extra();
+        auto needs_conv
+                = memory_extra_flags::compensation_gpu_conv_asymmetric_src;
+        auto is_dst_gpu = (dst_engine->kind() == engine_kind::gpu);
+        do_zp_precomp_conv_ = is_dst_gpu && (extra.flags & needs_conv);
+        if (do_zp_precomp_conv_) {
```
dzarukin, line 92: To save some indentation.
```suggestion
        if (!do_zp_precomp_conv_) return status::success;
```

src/gpu/gpu_reorder_pd.hpp
```diff
@@ -62,9 +69,99 @@ struct gpu_reorder_pd_t : public reorder_pd_t {
...
+                    && (dst_md()->extra.flags == none);
+        const uint64_t c = compensation_gpu_conv_asymmetric_src;
+        const uint64_t b = compensation_gpu_conv_asymmetric_src_bwd;
+        const uint64_t s = compensation_gpu_conv_asymmetric_src_swap;
+        return utils::one_of(src_md()->extra.flags, c, c | b, c | b | s, none)
+                && utils::one_of(
+                        dst_md()->extra.flags, c, c | b, c | b | s, none);
     }
+
+    status_t maybe_build_zp_precompute_conv_pd(impl::engine_t *dst_engine) {
```
dzarukin, line 85: Nit: it's usually "create", not "build". relevant for all underlying calls.

src/gpu/gpu_reorder_pd.hpp
```diff
@@ -62,9 +69,99 @@ struct gpu_reorder_pd_t : public reorder_pd_t {
...
+    }
+
+public:
+    status_t maybe_build_zp_precompute_conv(
+            std::shared_ptr<impl::primitive_t> &zp_precomp_conv,
+            impl::engine_t *engine, gpu::primitive_t *primitive) const {
+        return (!do_zp_precomp_conv_)
+                ? status::success
+                : primitive->create_nested_primitive(
+                        zp_precomp_conv, zp_precomp_conv_pd_, engine);
```
dzarukin, line 126:
```suggestion
        if (!do_zp_precomp_conv_) return status::success;
        return primitive->create_nested_primitive(
                        zp_precomp_conv, zp_precomp_conv_pd_, engine);
```

src/gpu/gpu_utils.hpp
```diff
@@ -21,9 +21,9 @@
 #include <map>
 #include <vector>
 
-#include "oneapi/dnnl/dnnl.h"
-
+#include "common/convolution_pd.hpp"
```
dzarukin, line 24: Doesn't feel right that utils now depend on convolution. It should probably leave in gpu_reorder.hpp or maybe even in a new standalone hpp + cpp pair, preferably a cpp file owning this dependency alone.

dzarukin (commented):
What if we don't compute this partial summation on the library side but instead provide an API to request compensations and accept them as a per-channel array?

hidefromkgb:
> What if we don't compute this partial summation on the library side but instead provide an API to request compensations and accept them as a per-channel array?

An important remark: this is not just a per-channel array but a full-fledged hidden buffer with DHW spatials.
Still, it's definitely possible, in theory, to expose ZP precompute as a part of oneDNN API — but maybe not right now.

In fact, this would be more appropriate from the architectural standpoint to expose rather than hide, as the current approach is all about tackling a graph-level problem with primitive-level tools.
Furthermore it'd lift the scalar-ZP-only requirement — a serious limitation the proposed PR has.

On the downside, though, the users would then become responsible for allocating network nodes to handle the ahead-of-time ZP precalculation.

@karturov @dzarukin What do you think of this?

src/gpu/gpu_reorder_pd.hpp
hidefromkgb: Done.

src/gpu/gpu_reorder_pd.hpp
hidefromkgb: Done.

src/gpu/gpu_reorder_pd.hpp
hidefromkgb: Done.

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

src/common/memory_desc_wrapper.hpp
```diff
@@ -149,30 +149,28 @@ struct memory_desc_wrapper : public c_compatible {
...
         using namespace memory_extra_flags;
         if (flag_select & compensation_conv_s8s8) return sizeof(int32_t);
-        if ((flag_select & rnn_u8s8_compensation)
-                && !types::extra_flag_rnn_s8s8_compensation_is_set(flag_select))
-            return sizeof(float);
+        if (flag_select & rnn_u8s8_compensation) return sizeof(float);
         if (flag_select & compensation_conv_asymmetric_src)
             return sizeof(int32_t);
+        if (flag_select & compensation_gpu_conv_asymmetric_src)
+            return sizeof(int32_t);
```
echeresh, line 156: Do we need to handle `compensation_gpu_conv_asymmetric_src_bwd` here as well?
As a related comment: since `memory_extra_desc_t ` already has `*dhw` members which are only used by the GPU reorder implementation - we could add another field to indicate whether it's src/src_bwd/src_swap. Exposing this via two extra flags may be too much detail for the top level.

src/common/primitive_hashing.cpp
```diff
@@ -206,6 +204,14 @@ size_t get_md_hash(const memory_desc_t &md) {
                 & dnnl_memory_extra_flag_compensation_conv_asymmetric_src) {
             seed = hash_combine(seed, md.extra.asymm_compensation_mask);
         }
+
+        if (md.extra.flags
+                & dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src) {
```
echeresh, line 209: Similar question, should the other two values be handled here?

src/gpu/intel/jit/reorder/gen_reorder.cpp
```diff
@@ -158,7 +159,8 @@ status_t gen_reorder_t::pd_t::init(impl::engine_t *engine,
     cfg->set_zp_cfg(zp_cfg);
     VDISPATCH_REORDER_SC(
             init_kernel_info(), "kernel initialization unsuccessful");
-
+    VDISPATCH_REORDER_SC(
+            maybe_create_zp_precompute_conv_pd(dst_engine), VERBOSE_BAD_FLAGS);
```
echeresh, line 163: I think "bad flags" here is confusing - a lot of things can go wrong when creating convolution. I would use a custom message, like "failed to create nested convolution for precomputed zero-points"

There is one more similar message below.

echeresh:
> In fact, this would be more appropriate from the architectural standpoint to expose rather than hide, as the current approach is all about tackling a graph-level problem with primitive-level tools.
Furthermore it'd lift the scalar-ZP-only requirement — a serious limitation the proposed PR has.

> On the downside, though, the users would then become responsible for allocating network nodes to handle the ahead-of-time ZP precalculation.

@hidefromkgb What do you mean exactly by "expose"? Exposing the structure of the `memory_extra_desc_t` and its contents? If so, what is the potential benefit?

I assume we would still provide some API (as e.g. reorder) to compute a compensation buffer in this form. It can't be easily/efficiently pre-computed by the user as it's not really trivial - so we have to expose it from the library in some form, say now it's done under reorder. Then I would say - it's not really important for the users to have any details on the buffer contents as the flow is 1) allocate an extra buffer 2) call reorder (or whatever) primitive 3) pass this buffer to convolution. There is no need to expose more details.

> Furthermore it'd lift the scalar-ZP-only requirement — a serious limitation the proposed PR has.

This can be implemented under the current design, right? The user can pass vector zero-points as an additional reorder argument - then the resulting compensation buffer would be computed accordingly. This should require only minor API changes at worst - something like an extra attribute for the convolution for the user to pass to opt in for this feature.

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

src/common/memory_desc.cpp
```diff
@@ -471,8 +471,6 @@ status_t memory_desc_permute_axes(memory_desc_t &out_memory_desc,
     VCHECK_MEMORY(
             !memory_desc_wrapper(in_memory_desc).has_runtime_dims_or_strides(),
             invalid_arguments, VERBOSE_UNSUPPORTED_MEM_STRIDE);
-    VCHECK_MEMORY(in_memory_desc.extra.flags == 0, invalid_arguments,
-            VERBOSE_UNSUPPORTED_MD_FLAG, "extra");
```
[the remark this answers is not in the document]
hidefromkgb: Done.

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

src/gpu/intel/jit/reorder/gen_reorder.cpp
hidefromkgb: Done.

src/common/memory_desc_wrapper.hpp
hidefromkgb: We don't, as `compensation_gpu_conv_asymmetric_src_bwd` and `compensation_gpu_conv_asymmetric_src_swap` imply `compensation_gpu_conv_asymmetric_src` and should never be used without it.
As for exposure, I think it's not feasible to make these flags their own bool fields, since that doesn't limit their scope compared to them being flags.

src/common/primitive_hashing.cpp
hidefromkgb: No, as both imply `compensation_gpu_conv_asymmetric_src`.

src/gpu/gpu_utils.hpp
hidefromkgb: Done.

src/common/memory_desc.hpp
```diff
@@ -56,21 +56,24 @@ const rnn_packed_memory_format_t ldio_p = rnn_packed_memory_format_t::ldio_p;
...
+    dnnl_memory_extra_flag_rnn_u8s8_compensation = 4u,
     dnnl_memory_extra_flag_gpu_rnn_u8s8_compensation
     = dnnl_memory_extra_flag_rnn_u8s8_compensation,
-    dnnl_memory_extra_flag_compensation_conv_asymmetric_src = 0x8U,
-    dnnl_memory_extra_flag_rnn_s8s8_compensation = 0x16U,
+    dnnl_memory_extra_flag_compensation_conv_asymmetric_src = 8u,
+    dnnl_memory_extra_flag_rnn_s8s8_compensation = 16u,
+    dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src = 32u,
+    dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src_bwd = 64u,
+    dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src_swap = 128u,
```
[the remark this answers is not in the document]
hidefromkgb: Done.

echeresh (approved):

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

src/common/memory_desc.hpp
```diff
@@ -87,8 +96,25 @@ const memory_extra_flags_t rnn_s8s8_compensation
...
+// need for a separate flag on GPU so that reorders could not mess these two up
+const memory_extra_flags_t compensation_gpu_conv_asymmetric_src
+        = dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src;
+const memory_extra_flags_t compensation_gpu_conv_asymmetric_src_bwd
+        = dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src_bwd;
+const memory_extra_flags_t compensation_gpu_conv_asymmetric_src_swap
+        = dnnl_memory_extra_flag_compensation_gpu_conv_asymmetric_src_swap;
 } // namespace memory_extra_flags
 
+inline bool check_md_extra_flags(uint64_t flags) {
```
dzarukin, line 109: This name implies it checks all the flags while in the reality it checks only compensation for gpu. Please make a more narrow name to underline that fact.

src/common/memory_desc.cpp
```diff
@@ -461,6 +461,7 @@ status_t memory_desc_reshape(memory_desc_t &out_memory_desc,
 
 status_t memory_desc_permute_axes(memory_desc_t &out_memory_desc,
         const memory_desc_t &in_memory_desc, const int *perm) {
+    using namespace memory_extra_flags;
```
dzarukin, line 464: Might not be needed any longer.

src/gpu/gpu_reorder_pd.hpp
```diff
@@ -17,7 +17,14 @@
...
 #define GPU_GPU_REORDER_PD_HPP
 
+#include "common/c_types_map.hpp"
+#include "common/memory_desc.hpp"
 #include "common/reorder_pd.hpp"
+#include "common/stream.hpp"
+#include "gpu/gpu_engine.hpp"
+#include "gpu/gpu_primitive.hpp"
+#include "gpu/gpu_stream.hpp"
+#include "gpu/gpu_zero_points_conv.hpp"
```
dzarukin, line 27: Would it be possible to move all those function bodies to a .cpp and all these header along the way?

src/gpu/gpu_zero_points_conv.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#ifndef GPU_GPU_ZERO_POINTS_CONV_HPP
+#define GPU_GPU_ZERO_POINTS_CONV_HPP
+
+#include <cassert>
+#include <vector>
+
+#include "common/convolution_pd.hpp"
+#include "common/primitive_attr.hpp"
+#include "common/primitive_desc_iterator.hpp"
```
dzarukin, line 25: Ditto re moving bodies to .cpp

densamoilov (approved):

src/common/memory_desc.cpp
hidefromkgb: Done.

src/common/memory_desc.hpp
hidefromkgb: Renamed to `check_md_extra_flags_compensation_gpu`.

src/gpu/gpu_reorder_pd.hpp
hidefromkgb: Done.

src/gpu/gpu_zero_points_conv.hpp
hidefromkgb: Done.

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

hidefromkgb:
make test
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_rnn

hidefromkgb:
make test perf-gpu
set primitive=conv

PR #2270, opened 2024-12-13
src: attr: quantization refactor (part 1)

dzarukin:
Every year around Christmas time something happens to quantization:
* 2022 - a move to runtime happened, lots of obsolete code left behind
* 2023 - advanced quantization with groups appeared, even more code that could use some love left behind.
* 2023.5 - extension of zero-points for SRC argument happened, zero-points has become a warehouse of variable to access directly.
* 2024 - time for refactor now!

The whole point of this refactor is to move quantization attributes to C++ way of doing things - provide clear and simple interfaces to operate with objects and close members.
This part 1 covers scales which were not that bad in terms of interfaces but could be better with argument members which this part covers.

Interface for both scales and its new underlying object, quant_entry_t, provide getters for a mask, data_type and groups (no need to worry about ndims any longer!), as well as default values checks.
Initialization is still done through `set`, `reset` was replaced with `set`, 

> Any operations with masks should happen only after verifying that specific arg scale is not default! It's forced now through an invalid mask value which can't be used as is.

Among legacy use-cases here are the main changes across sources:
common:
* some primitive attribute checkers got more checks added.
cpu:
* many places updated `mask != 0` which was the default value for common and non-initialized scales to `mask > 0` as now the default mask is negative.
  - A check for equality is still valid while it's highly recommended to avoid unequal comparison (unless you really know what you are doing).
gpu:
* found some bugs in `gemm_with_post_ops` implementations.
* changed logic in several generic/cudnn kernels to match the new behavior.
tests:
* updated gtests to comply with updated primitive checks.

Part 2 will cover zero-points.

Disclaimer: the change is somewhat fundamental, the bug leaks are highly possible even if all tests are passing. Feel free to report any if I missed something.

### DIFF
--- a/src/common/matmul_pd.hpp
+++ b/src/common/matmul_pd.hpp
@@ -205,6 +205,17 @@ struct matmul_pd_t : public primitive_desc_t {
                                 utils::one_of(
                                         1, sc.group_dims_[0], sc.group_dims_[1])
                                         && wei_k_group_ok && wei_n_group_ok);
+
+                // Mask over K dim is allowed for decompression feature only.
+                const bool is_decompression_or_dynquant
+                        = utils::one_of(weights_md(0)->data_type, data_type::s8,
+                                  data_type::u8, data_type::s4, data_type::u4)
+                        && IMPLICATION(
+                                !types::is_integral_dt(src_md()->data_type),
+                                attr()->fpmath_.apply_to_int_);
+                ok = ok
+                        && IMPLICATION((mask & wei_qmask_K()),
+                                is_decompression_or_dynquant);
             } else if (arg == DNNL_ARG_SRC) {
                 ok = ok
                         && utils::one_of(mask, 0, src_qmask_K(),
--- a/src/common/matmul_pd.hpp
+++ b/src/common/matmul_pd.hpp
@@ -182,29 +182,26 @@ struct matmul_pd_t : public primitive_desc_t {
 
     virtual bool attr_scales_ok(const std::vector<int> &supported_args
             = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST}) const {
-        if (attr()->scales_.has_default_values()) return true;
+        const auto &scales = attr()->scales_;
+        if (scales.has_default_values()) return true;
 
-        bool ok = attr()->scales_.has_default_values(supported_args);
+        bool ok = scales.has_default_values(supported_args);
         for (int arg : supported_args) {
-            const auto &sc = attr()->scales_.get(arg);
-            const auto &mask = sc.mask_;
-            if (sc.has_default_values()) { continue; }
+            if (scales.has_default_values(arg)) { continue; }
 
+            const auto &mask = scales.get_mask(arg);
             if (arg == DNNL_ARG_WEIGHTS) {
-                const bool wei_k_group_ok
-                        = IMPLICATION(sc.ndims_ == 2 && sc.group_dims_[0] > 1,
-                                K() % sc.group_dims_[0] == 0);
-                const bool wei_n_group_ok
-                        = IMPLICATION(sc.ndims_ == 2 && sc.group_dims_[1] > 1,
-                                N() % sc.group_dims_[1] == 0);
+                const auto &g0 = scales.get_group(arg, 0);
+                const auto &g1 = scales.get_group(arg, 1);
+                const bool wei_k_group_ok = IMPLICATION(g0 > 1, K() % g1 == 0);
+                const bool wei_n_group_ok = IMPLICATION(g1 > 1, N() % g0 == 0);
 
                 // Any group is allowed to be greater than 1 but only one at a
                 // time, not both.
-                ok = ok && utils::one_of(sc.ndims_, 0, 2)
-                        && IMPLICATION(sc.ndims_ == 2,
-                                utils::one_of(
-                                        1, sc.group_dims_[0], sc.group_dims_[1])
-                                        && wei_k_group_ok && wei_n_group_ok);
+                ok = ok
+                        && IMPLICATION(!scales.get(arg).has_default_groups(),
+                                utils::one_of(1, g0, g1) && wei_k_group_ok
+                                        && wei_n_group_ok);
 
                 // Mask over K dim is allowed for decompression feature only.
                 const bool is_decompression_or_dynquant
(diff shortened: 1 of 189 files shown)


dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

tests/gtests/test_iface_attr_quantization.cpp
```diff
@@ -84,7 +84,7 @@ TEST_F(attr_quantization_test_t, TestBNorm) {
                 eng, prop_kind::forward_inference, md, md, 0.1f, flags));
         CHECK_UNIMPL(batch_normalization_forward::primitive_desc(eng,
                 prop_kind::forward_inference, md, md, 0.1f, flags,
-                gen_attr_with_scales()));
+                gen_attr_with_scales(/* with_wei = */ false)));
```
mgouicem, line 87: Does it mean that scales.has_default_values is not properly called in implementations?

src/common/primitive_attr_quant.hpp
```diff
@@ -28,175 +28,185 @@
...
-        auto predicate = [](const runtime_scales_t &s) {
-            return s.has_default_groups();
+    // This interface checks the content of all scales, and allows to ignore
+    // certain arguments.
+    bool has_default_data_type(
+            const std::vector<int> &supported_args = {}) const {
+        auto predicate = [](const quant_entry_t &s) {
+            // Note: `data_type::undef` represents `default_quant_entry`.
+            return utils::one_of(
+                    s.get_data_type(), default_data_type, data_type::undef);
```
mgouicem, line 160: (nit) maybe add `has_default_data_type` to quant_entry_t ?

src/cpu/ref_convolution_int8.cpp
```diff
@@ -35,7 +35,7 @@ namespace {
 void dequantize(float &d, dim_t g, dim_t C, dim_t c, const float *wei_scales,
         bool with_groups, int wei_mask, const float *src_scales) {
     // scale_idx_mult = 1 for per_channel scales and 0, otherwise
-    const int wei_scale_idx_mult = wei_mask != 0;
+    const int wei_scale_idx_mult = wei_mask > 0;
```
mgouicem, line 38: would it make sense to normalize mask with `quant_entry_t::get_mask()` method instead?
This would keep the use of -1 as default value an implementation details, and should be more safe to avoid some checks issues (e.g. comparison to 0 or usage of binary oring/xoring).

src/cpu/scale_utils.cpp
```diff
@@ -32,21 +32,19 @@ constexpr size_t scales_simd_w = 16;
...
-            = attr_scales.get(DNNL_ARG_WEIGHTS).ndims_;
+
     if ((with_src_scales && with_wei_scales) || force_scales_book
-            || (wei_scales_dt != data_type::f32 && with_wei_scales)
-            || (wei_scale_groups_ndims > 0 && with_wei_scales)) {
-        const int wei_mask = attr_scales.get(DNNL_ARG_WEIGHTS).mask_;
+            || !attr_scales.has_default_data_type(DNNL_ARG_WEIGHTS)
+            || !attr_scales.get(DNNL_ARG_WEIGHTS).has_default_groups()) {
+        const int wei_mask = attr_scales.get_mask(DNNL_ARG_WEIGHTS);
         const size_t precomputed_scales_size = wei_mask == 0
```
mgouicem, line 48: This comparison to 0 looks unsafe here, since default mask is -1 now (it should be fine now because with_wei_scales checks for has_default_values, but that seems slippery).

mgouicem (approved):
Looks good overall, thanks for the refactor.
Just one thing about the mask: moving from a default of 0 to -1 seems slippery.

Now the dev will have to always check if the value is non default before   doing a comparison to 0, or doing some binary operation (e.g. or/xor).
I think it might be worth to normalize the mask upon get_mask so that default value becomes an implementation detail.

src/gpu/nvidia/cudnn_matmul_lt.hpp
```diff
@@ -215,10 +215,9 @@ struct cudnn_matmul_lt_t : public gpu::primitive_t {
         }
 
         status_t create_scale_binary_pd(impl::engine_t *engine, int ARG) {
-            auto scale_md = dnnl_memory_desc();
-            scale_md.ndims = attr()->scales_.get(ARG).ndims_;
-            scale_md.data_type = attr()->scales_.get(ARG).data_type_;
-            scale_md.format_kind = dnnl_blocked;
+            memory_desc_t scale_md;
```
sgeor255, line 218: Is `scale_md.ndims = ...` missing on purpose here?

sgeor255 (approved):

sgeor255 (approved):

src/gpu/generic/sycl/ref_layer_normalizations.cpp
```diff
@@ -44,12 +44,8 @@ status_t ref_layer_normalization_fwd_t::pd_t::init_conf() {
     conf_.src_def = attr()->scales_.get(DNNL_ARG_SRC).has_default_values();
     conf_.dst_def = attr()->scales_.get(DNNL_ARG_DST).has_default_values();
 
-    conf_.scales_src_dt = conf_.src_def
-            ? data_type_t::dnnl_f32
-            : attr()->scales_.get(DNNL_ARG_SRC).data_type_;
-    conf_.scales_dst_dt = conf_.dst_def
-            ? data_type_t::dnnl_f32
-            : attr()->scales_.get(DNNL_ARG_DST).data_type_;
+    conf_.scales_src_dt = attr()->scales_.get_data_type(DNNL_ARG_SRC);
```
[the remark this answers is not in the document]
dzarukin: It shouldn't be. `get_data_type` will either return the actual data type is the scale was set, or `undef` if it was not. The check for `def` must be done at the point where scales are applied (whether they actually should or should not).

src/gpu/nvidia/cudnn_inner_product.hpp
```diff
@@ -39,15 +39,17 @@ struct cudnn_inner_product_fwd_t : public gpu::primitive_t {
...
-        bool attr_scales_ok() const {
-            const auto &scales = attr()->scales_;
-            const auto &supported_args
-                    = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST};
-            if (!scales.has_default_values(supported_args)) return false;
-            // cuDNN does not support scaling per dimension.
-            for (auto arg : supported_args)
-                if (scales.get(arg).mask_ != 0) return false;
-            return true;
+        bool attr_scales_ok(const std::vector<int> &supported_args
```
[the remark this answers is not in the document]
dzarukin: The idea is to re-use existing `attr_scales_ok()` methods whenever possible. Since this PR is not targeting NVidia backend refactor, I used the best version that suits the need of this PR.

src/gpu/nvidia/cudnn_matmul_lt.hpp
dzarukin: `ndims` in scales corresponds to the groups feature which is not supported for NVidia backend. Previous code was not correct in terms of that argument usage. Moreover, the memory descriptor would be re-created regardless of the value (which was always 0, btw) based on src memory desc, so the current version simplifies the former one.

src/common/matmul_pd.hpp
```diff
@@ -184,6 +184,18 @@ struct matmul_pd_t : public primitive_desc_t {
                                 utils::one_of(
                                         1, sc.group_dims_[0], sc.group_dims_[1])
                                         && wei_k_group_ok && wei_n_group_ok);
+
+                // Mask over K dim is allowed for decompression or dynamic
+                // quantization features only.
```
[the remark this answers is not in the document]
dzarukin: Removed "dynamic quantization" mention. It supposed to mean "weights decompression with int8 activations".

src/common/primitive_attr_quant.hpp
dzarukin: Discussed offline, `undef` is not the only default data type as there's an interface with just mask which means `f32` is a default data type which is reflected on a scales level.

Added a comment that the call can be removed in the future once each implementation supports data type for scales.

src/cpu/ref_convolution_int8.cpp
dzarukin: Discussed offline. Agreed on keeping mask as `INT_MIN` and get_mask() returning it to force developers check the scales state before querying the mask and using it as is. `INT_MIN` will prevent from collapsing into positive output for bit operations.

If this brings more issues in the future rather it solves, Mourad suggested to return 12 lowest bits of current mask from `get_mask()` which would simply return `0` was `INT_MIN` default value and restore the behavior that always existed.

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test
enable test_scope=NIGHTLY

echeresh (approved):

src/common/primitive_attr_quant.hpp
```diff
@@ -28,173 +28,205 @@
...
-    data_type_t data_type_ = data_type::f32;
+    void serialize(serialization_stream_t &sstream) const;
+
+    std::string get_verbose() const;
+
+private:
+    // Note: INT_MIN is used on purpose to avoid potential issues when
+    // `(mask & bit)` expression will return `true`. `INT_MIN` is represented
+    // as `10...0` in bits and will avoid such situations.
+    int mask_ = INT_MIN;
```
densamoilov, line 113: Is there any reason why we prefer signed int over unsigned int for bitmask?

densamoilov (approved):

PR #2280, opened 2024-12-17
rfcs: graph api: Constant Block Weight Mechanism Graph API design

xiang1guo:
The RFC proposes a design for handling constant block weight mechanism in graph API.

Rendered version: [link](https://github.com/oneapi-src/oneDNN/tree/xiang/rfcs/constant-weight-cache-design/rfcs/20241216-constant-block-weight-api-design)


rfcs/20241216-constant-block-weight-api-design/README.md
```diff
@@ -0,0 +1,243 @@
...
+optimal performance without extra handling of reorder or understanding the
+details of block layout. But it also introduces less flexibility for different 
+graph API usage and integration in framework side. 
+
+For example, for the operation direct optimization with graph API, the same
+compiled partition cache may correspond to different constant weights, so the
+graph library cannot directly store the constant weight cache in the compiled
+partition cache. Instead, it needs to distinguish between these different 
+constant weight tensors. However, distinguishing constant weights requires 
+additional information from the user side. This raises new considerations for
```
dzarukin, line 25: The problematic case is not quite clear to me: is the case that different data is supplied with a single compiled partition that it's caching is not possible? When re-usage of compiled partition happens exactly? What's the real-world scenario for it?

xiang1guo: Hi, Dima, thanks for the review! 
>is the case that different data is supplied with a single compiled partition that it's caching is not possible? When re-usage of compiled partition happens exactly? 

Consider a whole graph/model contains 2 subgraphs, the first subgraph compiles and caches the compiled partition with constant weight cache, and the second subgraph compilation will cache hit (it's possible) the first compiled partition cache and reuse the constant tensor cache in compiled partition cache(with constant cache feature enabled). However, the second subgraph has a different weight input(the constant weight cache cannot be reused in the first cp cache), thus leading to an accuracy issue.

>What's the real-world scenario for it?

Currently we are integrating graph API in PyTorch for SDPA op direct optimization, the workflow is that each SDPA's compilation, and execution with graph API happens in every iteration of the PyTorch SDPA operation call. Thus there may be some compilation cache hit and reuse but with different input. 
But we are safe now since SDPA doesn't contain any constant tensor cache(constant weight), so compiled partition cache reuse won't introduce any issues. However, we are planning for some other ops that may need constant tensor cache, like the MLP optimization on the OpenVINO side.

dzarukin: > the second subgraph compilation will cache hit (it's possible) the first compiled partition cache

Hasn't compiled partition have an ID which should be checked by the CP-cache? Will underneath LTs' IDs be ignored in case as this potential hit as well?

xiang1guo: >>the second subgraph compilation will cache hit (it's possible) the first compiled partition cache

>Hasn't compiled partition have an ID which should be checked by the CP-cache? Will underneath LTs' IDs be ignored in case as this potential hit as well?

We do have ID(both logical tensor and op ID) hashed into the key of CP-cache, it works very well on the legacy integration on IPEX and ITEX, since the whole model will be mapped to the graph API with unique ID. 
But for the current OP direct optimization, the integration exists under a single operation, they may use the same ID to create op and logical tensor. So the cp cache will hit. Actually, for such new integration solution, compilation happens on every iteration, more cp cache hit helps to eliminate the compilation time. That's why we want to address the issue without hurting cache hit.

openmodelzooci:

PR #2276, opened 2024-12-17
xe, benchdnn, x64: Enable large buffer GEMM and Convolution testing for GPU

rjoursler:
Fixes issues encountered when testing large buffer GEMM and convolution operations on the GPU.

PR #2272, opened 2024-12-14
cpu: x64: brgemm, matmul: add f32:f16 configuration support on AVX2 and AVX512_CORE (fixes MFDNN-11992)

dzarukin:
MFDNN-11992

The change adds f32:f16:f32 support on AVX512_CORE and AVX2 through a up-conversion path.
Extended a brgemm kernel (thanks @dmitry-gorokhov) and brgemm matmul copy routines to support the conversion.

PR #2286, opened 2024-12-18
github: automation: fixed rules for commit scope check

vpirogov:
This PR updates commit checker to avoid false positives in #2279 ([job link](https://github.com/oneapi-src/oneDNN/actions/runs/12393533080/job/34594975133)), #2276, and #2272.

As `:` can be used in commit message itself there's no good way to reliably identify cases with multi-level scope.

### DIFF
--- a/.github/automation/commit-msg-check.py
+++ b/.github/automation/commit-msg-check.py
@@ -31,22 +31,22 @@ def __scopeCheck(msg: str):
     status = "Message scope: "
 
     if not re.match('^[a-z0-9]+(, [a-z0-9]+)*: ', msg):
-        print(f"{status} FAILED: Commit message must follow the format <scope>:[ <scope>:] <short description>")
+        print(f"{status} FAILED: Commit message must follow the format "
+               "<scope>:[ <scope>:] <short description>")
         return False
 
     print(f"{status} OK")
     return True
 
-# Ensuring a character limit for the first line.
+# Ensure  a character limit for the first line.
 def __numCharacterCheck(msg: str):
     status = "Message length:"
-    summary = msg.partition("\n")[0]
-    msgSummaryLen = len(summary)
-    if msgSummaryLen <= 72:
+    if len(msg) <= 72:
         print(f"{status} OK")
         return True
     else:
-        print(f"{status} FAILED: Commit message summary must not exceed 72 characters.")
+        print(f"{status} FAILED: Commit message summary must not "
+               "exceed 72 characters.")
         return False
 
 def main():
@@ -58,7 +58,8 @@ def main():
     head: str = args.head
 
     commit_range = base + ".." + head
-    messages = subprocess.run(["git", "rev-list", "--format=oneline", commit_range], capture_output=True, text=True).stdout
+    messages = subprocess.run(["git", "rev-list", "--format=oneline",
+        commit_range], capture_output=True, text=True).stdout
 
     is_ok = True
     for i in messages.splitlines():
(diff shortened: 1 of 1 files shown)


.github/automation/commit-msg-check.py
```diff
@@ -20,29 +20,17 @@
...
-    # The last element of the split is the title, which we don't care about.
-    scopesArray = firstLine.split(":")[:-1]
-
-    for scopes in scopesArray:
-        numWords = len(scopes.split())
-        numCommas = scopes.count(",")
-
-        if numWords != numCommas + 1:
-            print(f"{status} FAILED: Same-level scopes must be comma-separated. Bad token: '{scopes}'")
-            return False
```
rjoursler, line 44: I don't see any issue with keeping this check for the top level scope. At minimum, the comment on line 25 should be updated if this is removed.

.github/automation/commit-msg-check.py
```diff
@@ -20,29 +20,17 @@
 
 import argparse
 import subprocess
+import re
 
 # * Ensuring the scopes end in colon and same level scopes are comma delimited.
 # TODO: Limit scopes to an acceptable list of tags.
```
rjoursler, line 26: Please add to this comment that the current implementation can only validate the top level scope due to ambiguous parsing of the `:` token between the scope title and the description.

.github/automation/commit-msg-check.py
```diff
@@ -79,7 +67,7 @@ def main():
       is_ok = is_ok and result
 
     if is_ok:
```
theComputeKid, line 69: Casually switching to snake_case halfway through the file that started with camelCase..

.github/automation/commit-msg-check.py
vpirogov: Right. Technically coding guidelines do not allow comma for top level scope, though I see that we do that from time to time in the commit history.

theComputeKid:
This means that e.g. `Removing: from doc` would be a valid commit title, since there is no validation of the initial tag.

.github/automation/commit-msg-check.py
vpirogov: Looks like argparse package set the tone for `main()`...

.github/automation/commit-msg-check.py
```diff
@@ -20,33 +20,23 @@
 
 import argparse
 import subprocess
+import re
 
-# * Ensuring the scopes end in colon and same level scopes are comma delimited.
+# Ensuring the scope ends in colon and same level scopes are comma delimited.
```
rjoursler, line 25:
```Suggestion
# Ensure the scope ends in a colon and that same level scopes are comma delimited.
```

rjoursler (approved):

vpirogov:
> This means that e.g. `Removing: from doc` would be a valid commit title, since there is no validation of the initial tag.

No, this one will fails because of capital 'R' :). Anything that follows the format of `<scope>: <description>` or `<scope1>, <scope2>: <description>` will fly. Testing on commits since oneDNN v3.5 release indicates that nothing really ugly passes and less than 3% of commits fail.

theComputeKid (approved):

.github/automation/commit-msg-check.py
```diff
@@ -67,7 +58,8 @@ def main():
     head: str = args.head
 
     commit_range = base + ".." + head
-    messages = subprocess.run(["git", "rev-list", "--format=oneline", commit_range], capture_output=True, text=True).stdout
+    messages = subprocess.run(["git", "rev-list", "--format=oneline",
+               commit_range], capture_output=True, text=True).stdout
```
mgouicem, line 62: (nit) according to [peps8](https://peps.python.org/pep-0008/#indentation), arguments should 
- either start on the next line with 4 spaces offset, 
- or start on same line but upon line break be aligned with the opening bracket

mgouicem (approved):

.github/automation/commit-msg-check.py
vpirogov: we need clang-format but for Python...

PR #2291, opened 2024-12-19
graph: backend: dnnl: support permute for scale and zps inputs

wzt1997:
## Description

We recently supported the compressed SDPA patterns which incorporate scales and zero points as inputs. Also, we aim to support the problem with a Key shape of ( N, H, S, D ) with `transpose_b`=true for the QK MatMul. For such case with transposed MatMul, the graph library adds `permute` ops prior to the pattern input, including the scale and zp tensors. To be more specific, for the following woq MatMul with tranpose_b=true:
image

After the graph compilation, we will have the graph as follows. The permute ops will not touch the physical memories but only change the mds.
image

However, we faced a issue that DNNL backend always regards scale and zp tensors as `abx` format, leading to failing results, hence we need to set the tag explicitly as `abx` and execute extra reorders. After the change, the graph after compilation is like:
image

### DIFF
--- a/src/graph/backend/dnnl/subgraph.cpp
+++ b/src/graph/backend/dnnl/subgraph.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2022-2024 Intel Corporation
+ * Copyright 2022-2025 Intel Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -442,6 +442,26 @@ void subgraph_rewriter_t::insert_op_before(const op_ptr &inserted_op,
     auto new_val = std::make_shared<value_t>(*inserted_op, 0, new_lt, true);
     auto in_dtype = in_val->get_logical_tensor().data_type;
     new_val->set_data_type(in_dtype);
+
+    if (inserted_op->get_kind() == op_kind::dnnl_permute
+            && (base_op->get_kind() == op_kind::dnnl_mul_scales
+                    || base_op->get_kind() == op_kind::dnnl_sub_zps)) {
+        // Only abx tag is respected for scale and zps inputs, should set
+        // strides explicitly and execute reorder.
+
+        dnnl::memory::desc in_md
+                = make_dnnl_memory_desc(in_val->get_logical_tensor());
+        const auto &perm = inserted_op->get_attr<std::vector<int64_t>>(
+                op_attr::permutation);
+        std::vector<int> int_perm(perm.size(), -1);
+        for (size_t i = 0; i < perm.size(); i++) {
+            int_perm[i] = static_cast<int>(perm[i]);
+        }
+        dnnl::memory::desc out_md = in_md.permute_axes(int_perm);
+        const auto &dims = out_md.get_dims();
+        // set the strides with abx tag.
+        new_val->set_strides(get_dense_strides(dims));
+    }
 
     if (k == std::numeric_limits<size_t>::max()) {
         k = inserted_op->num_outputs();
--- a/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


tests/benchdnn/inputs/graph/complex_fusion/mha/sdpa-compressed-k-int8-gs32.json
```diff
@@ -32,13 +32,13 @@
             "value": [
               1,
               1,
-              32,
-              1
+              1,
+              32
```
TaoLv, line 36: What's the purpose of changing the json file? My understanding is that all of these shape/attribute changes can be done through benchdnn command line.

src/graph/backend/dnnl/subgraph.cpp
```diff
@@ -443,6 +443,32 @@ void subgraph_rewriter_t::insert_op_before(const op_ptr &inserted_op,
...
+        }
+        dnnl::memory::desc out_md = in_md.permute_axes(int_perm);
+        const auto &ndims = out_md.get_ndims();
+        const auto &dims = out_md.get_dims();
+
+        std::vector<dim_t> out_strides(ndims, 1);
+        // calculate the strides with abx for the output logical tensor.
+        for (int idx = ndims - 2; idx >= 0; --idx) {
+            out_strides[idx] = out_strides[idx + 1] * dims[idx + 1];
+        }
```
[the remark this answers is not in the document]
wzt1997: Updated. Thanks for the suggestion.

tests/benchdnn/inputs/graph/complex_fusion/mha/sdpa-compressed-k-int8-gs32.json
wzt1997: Thanks for reminding. I've moved the test case to the batch file with cml rewriting.

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/subgraph.cpp
```diff
@@ -443,6 +443,26 @@ void subgraph_rewriter_t::insert_op_before(const op_ptr &inserted_op,
     auto in_dtype = in_val->get_logical_tensor().data_type;
     new_val->set_data_type(in_dtype);
 
+    if (inserted_op->get_kind() == op_kind::dnnl_permute
+            && (base_op->get_kind() == op_kind::dnnl_mul_scales
+                    || base_op->get_kind() == op_kind::dnnl_sub_zps)) {
```
rongzha1, line 448: do we need to check dnnl_add_zps ?

wzt1997: No. This check only affects dynamic dequant.

src/graph/backend/dnnl/subgraph.cpp
```diff
@@ -443,6 +443,26 @@ void subgraph_rewriter_t::insert_op_before(const op_ptr &inserted_op,
...
+        // as DNNL backend only respect abx tag for scale and zps, should set
+        // strides explicitly and execute reorder.
```
xiang1guo, line 450: Nit: seems not the `DNNL backend` (maybe primitive kernel ??) respect abx tag. More clear clarification for this special case helps for later revisitor. Thanks~

xiang1guo (approved):

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

gyhintel (approved):

wzt1997:
During the validation of dynamic per-channel quantization cases, especially for those with zps and transposed matmul, some issues were discovered. So I pushed 2 new commits for enhancing the handling for dynamic per-channel quantization cases in the PR, please review again:
1. Apart from per-group quantization, DNNL backend always accepts `mask`=0 for zero points, hence improved the op def check for dynamic quantization ops to allow zp tensor of which nelems=1.
2. Added corresponding compressed SDPA case with per-channel quantization, and the transposed test case.

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

gyhintel (approved):

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

PR #2295, opened 2024-12-19
tests: benchdnn: add experimental support for sycl graph

spalicki:
Adds ability to run benchdnn tests using graph execution by adding `--execution-mode=[direct(default)|graph]` option. It is currently limited to SyCL GPU engine executing on L0 backend.

### DIFF
--- a/cmake/testing.cmake
+++ b/cmake/testing.cmake
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2020-2024 Intel Corporation
+# Copyright 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ set(DNNL_TEST_SET_COVERAGE "0")
 set(DNNL_TEST_SET_COVERAGE_STR "")
 set(DNNL_TEST_SET_HAS_NO_CORR "0")
 set(DNNL_TEST_SET_HAS_ADD_BITWISE "0")
+set(DNNL_TEST_SET_HAS_GRAPH_EXE "0")
 
 function(check_consistency entry)
     if(NOT DNNL_TEST_SET_COVERAGE EQUAL 0)
@@ -57,6 +58,8 @@ foreach(entry ${DNNL_TEST_SET})
         set(DNNL_TEST_SET_HAS_NO_CORR "1")
     elseif(entry STREQUAL "ADD_BITWISE")
         set(DNNL_TEST_SET_HAS_ADD_BITWISE "1")
+    elseif(entry STREQUAL "GRAPH_EXE")
+        set(DNNL_TEST_SET_HAS_GRAPH_EXE "1")
     elseif(entry STREQUAL "CI_NO_CORR") # Left here for compatibility till v4.0
         set(DNNL_TEST_SET_COVERAGE ${DNNL_TEST_SET_CI})
         set(DNNL_TEST_SET_COVERAGE_STR "CI")
@@ -68,7 +71,7 @@ foreach(entry ${DNNL_TEST_SET})
         message(FATAL_ERROR
                 "The DNNL_TEST_SET entry ${entry} is not recognized. "
                 "Supported values are:"
-                "NIGHTLY, CI, SMOKE, NO_CORR, ADD_BITWISE.")
+                "NIGHTLY, CI, SMOKE, NO_CORR, ADD_BITWISE, GRAPH_EXE.")
     endif()
 endforeach()
 
@@ -78,4 +81,7 @@ if(DNNL_TEST_SET_HAS_NO_CORR EQUAL 1)
 endif()
 if(DNNL_TEST_SET_HAS_ADD_BITWISE EQUAL 1)
     message(STATUS "Enabled testing modifier: Add bitwise validation")
+endif()
+if(DNNL_TEST_SET_HAS_GRAPH_EXE EQUAL 1)
+    message(STATUS "Enabled testing modifier: Use graph execution")
 endif()
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -366,8 +366,40 @@ int execute_and_wait(perf_function_t &exec_func, const dnnl_engine_t &engine,
 
     execute_unmap_args(args, dnnl_args);
 
-    auto status = exec_func(stream, dnnl_args);
-    DNN_SAFE(dnnl_stream_wait(stream), CRIT);
+    dnnl_status_t status = dnnl_runtime_error;
+    bool run_regular_exec = true;
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_DPCPP
+    while (execution_mode == execution_mode_t::graph && is_gpu(engine)) {
+        void *queue_ptr;
+        DNN_SAFE(dnnl_sycl_interop_stream_get_queue(stream, &queue_ptr), CRIT);
+        sycl::queue queue = *static_cast<sycl::queue *>(queue_ptr);
+        const bool can_run_sycl_graph = queue.get_device().get_backend()
+                == sycl::backend::ext_oneapi_level_zero;
+        if (!can_run_sycl_graph) break;
+
+        BENCHDNN_PRINT(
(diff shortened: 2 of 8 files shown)


tests/benchdnn/common.hpp
```diff
@@ -93,6 +93,7 @@ enum { CRIT = 1, WARN = 2 };
 extern int verbose;
 extern bool canonical;
 extern bool mem_check;
+extern bool use_sycl_graph;
```
dzarukin, line 96: Please update benchdnn doc page knobs_common to have this new option in there.

cmake/testing.cmake
```diff
@@ -79,3 +82,6 @@ endif()
 if(DNNL_TEST_SET_HAS_ADD_BITWISE EQUAL 1)
     message(STATUS "Enabled testing modifier: Add bitwise validation")
 endif()
+if(DNNL_TEST_SET_HAS_SYCL_GRAPH EQUAL 1)
+    message(STATUS "Enabled testing modifier: Use experimental SyCL graph execution")
```
Rbiessy, line 86: Please rename the few occurrences of "SyCL" to "SYCL"

tests/benchdnn/doc/knobs_common.md
```diff
@@ -146,6 +146,12 @@ Additional information is printed to the stdout depending on a level `N`. `N` is
 a non-negative integer value. The default value is `0`. Refer to
 [verbose](knobs_verbose.md) for details.
 
+### --use-sycl-graph
+`--use-sycl-graph=BOOL`, Instructs the driver to execute using the experimental
```
ranukund, line 150:
```suggestion
`--use-sycl-graph=BOOL` instructs the driver to execute using the experimental
```

tests/benchdnn/doc/knobs_common.md
```diff
@@ -146,6 +146,12 @@ Additional information is printed to the stdout depending on a level `N`. `N` is
...
+SyCL Graph backend when `BOOL` is set to `true`. When `BOOL`is `false`
```
ranukund, line 151:
```suggestion
SYCL Graph backend when `BOOL` is set to `true`. When `BOOL` is set to `false`
```

ranukund (approved):
A few edits suggested, please incorporate as you see fit!

tests/benchdnn/dnnl_common.cpp
```diff
@@ -366,8 +366,40 @@ int execute_and_wait(perf_function_t &exec_func, const dnnl_engine_t &engine,
 
     execute_unmap_args(args, dnnl_args);
 
-    auto status = exec_func(stream, dnnl_args);
-    DNN_SAFE(dnnl_stream_wait(stream), CRIT);
+    dnnl_status_t status = dnnl_runtime_error;
+    bool run_regular_exec = true;
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_DPCPP
+    while (use_sycl_graph && is_gpu(engine)) {
```
ShanoToni, line 372: Why is `while` with a `break`s preferred here instead of just `if`s or a lambda?

tests/benchdnn/dnnl_common.cpp
```diff
@@ -366,8 +366,40 @@ int execute_and_wait(perf_function_t &exec_func, const dnnl_engine_t &engine,
...
+        sycl::ext::oneapi::experimental::command_graph graph {
+                queue.get_context(), queue.get_device()};
+
+        graph.begin_recording(queue);
+        status = exec_func(stream, dnnl_args);
+        graph.end_recording(queue);
+        DNN_SAFE(dnnl_stream_wait(stream), CRIT);
+
+        auto exec = graph.finalize();
+        queue.ext_oneapi_graph(exec).wait();
```
Rbiessy, line 391: Can you give a bit more context on the reasons behind this PR in general? I see it is adding a way to test the SYCL Graph extension with benchdnn. Is the intention to test for correctness and/or measure the impact on performance? Wouldn't it be useful to also be able to re-use the graph `n` times?
Also the PR says it is an experimental support. Is there a plan for the future of SYCL-Graph in benchdnn?

petercad:
General suggestions: for use in CI, it would be nice to be able to apply `--use-sycl-graph` on a per-layer basis. I'd also suggest a more generic name (perhaps something like `--gpu-execution-mode=graph`), because other GPU runtimes also support similar functionality.

tests/benchdnn/dnnl_common.cpp
vpirogov: The main goal is enabling functional validation of oneDNN primitives in SYCL Graph record/replay mode. While use of SYCL Graph does not require changes in oneDNN API it puts some limitations on what the implementation can do internally.

'Experimental' refers to [support status](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_graph.asciidoc#5-status) of SYCL Graph extension support in the compiler.

tests/benchdnn/dnnl_common.cpp
dzarukin: If the functionality can't be supported, the fallback to a regular execution must happen. Without `while` it's gonna be a more tangled connection between two or three `if`s. Replacing this code with lambda would be identical to this `while` as we need objects from checking part and executing part, can't break it down into two without passing those objects around which doesn't help with readability. So to day, it's just a style preference.

dzarukin:
> General suggestions: for use in CI, it would be nice to be able to apply `--use-sycl-graph` on a per-layer basis. I'd also suggest a more generic name (perhaps something like `--gpu-execution-mode=graph`), because other GPU runtimes also support similar functionality.

By "per layer" do you mean to run like this:
```
--matmul --gpu-execution-mode=,graph 32x32:32x64
```
and two problems will be executed, one with sycl-graph, another - not?

spalicki:
> General suggestions: for use in CI, it would be nice to be able to apply `--use-sycl-graph` on a per-layer basis. I'd also suggest a more generic name (perhaps something like `--gpu-execution-mode=graph`), because other GPU runtimes also support similar functionality.

The function is limited to GPU today due to L0 dependency, but in the end, it should be extended to CPU as well. Maybe then `--execution-mode=[direct(default)|graph]`?

petercad:
> By "per layer" do you mean to run like this:
>    `--matmul --gpu-execution-mode=,graph 32x32:32x64`
> and two problems will be executed, one with sycl-graph, another - not?

@dzarukin My main thought was that it would be nice to add a few targeted matmul test cases for SYCL graph (since there are some special SYCL graph code paths), but we don't need to test the whole CI test suite with SYCL graph. By making a per-layer knob we can add some graph test cases to the regular CI test suites instead of creating new test suites for SYCL graph.

> The function is limited to GPU today due to L0 dependency, but in the end, it should be extended to CPU as well. Maybe then --execution-mode=[direct(default)|graph]?

That sounds good to me.

tests/benchdnn/dnnl_common.cpp
Rbiessy: Ok, thanks. Would it be safer to re-use the graph a number of times to make sure that the limitations you mentioned are respected? Is this planned for a separate PR maybe?
I think it looks odd to create a graph without re-using it.

spalicki:
> My main thought was that it would be nice to add a few targeted matmul test cases for SYCL graph (since there are some special SYCL graph code paths), but we don't need to test the whole CI test suite with SYCL graph.
@petercad The knob I added can be used for whole test suite, batch file, but also per-layer. My initial thought was to simply test the CI suite on SYCL Graph during our weekly testing (just to make sure it works), but I can create a limited scope batch files for some selected primitives.

tests/benchdnn/common.hpp
```diff
@@ -107,6 +99,28 @@ extern std::string driver_name;
...
+#define LIST_OF_EXECUTION_MODES(X) \
+    X(direct) \
+    X(graph)
+
+enum class execution_mode_t {
+#define EXECUTION_MODES_ENUM(name, ...) name,
+    LIST_OF_EXECUTION_MODES(EXECUTION_MODES_ENUM)
+#undef EXECUTION_MODES_ENUM
+};
+extern execution_mode_t execution_mode;
```
dzarukin, line 119: I can see why you defined the enum this way but amount of macros is freaking me out...
Could you, please, make it simple through regular `const char *` as it's done for others? I don't think this will ever get any extension...

dzarukin (approved):

PR #2297, opened 2024-12-19
xe: jit: enable kernel debug information

rjoursler:
Downstreams new nGEN code. In particular downstreams support for DWARF debug information. The basic debug information was originally implemented in #2227, and was further extended to include a line table. The line table is only enabled in DNNL_DEV_MODE to avoid overhead in release builds (potentially this condition should be removed to improve the information collected by profiling tools) and under C++20 builds (due to a dependency on `std::source_location`.) This line table works with `gdb-onapi` as can be seen with the following picture:

image

Sample usage (requires using SYCL gpu runtime for gdb to work):
```
$ cat env/gdb-oneapi_env.sh
source ~/intel/gdb-oneapi/setvars.sh --force
export ZET_ENABLE_PROGRAM_DEBUGGING=1

for f in /sys/class/drm/card*/prelim_enable_eu_debug; do
    val=$(cat $f)
    if [ $val == 0 ]; then
        sudo sh -c "echo 1 > $f"
    fi
done
$ source env/gdb-oneapi_env.sh

:: initializing oneAPI environment ...
   -bash: BASH_VERSION = 5.1.16(1)-release
   args: Using "$@" for setvars.sh arguments:
:: debugger -- latest
:: oneAPI environment initialized ::
$ gdb-oneapi --args ~/dnnl/build/tests/benchdnn/benchdnn --matmul --engine=gpu 2048x13:13x512
...
(gdb) break _entry
Function "_entry" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (_entry) pending.
(gdb) r
...
Thread 2.1414 hit Breakpoint 1, with SIMD lanes [0-15], gemm_kernel ()
    at ~/dnnl/src/gpu/intel/jit/gemm/generator/pieces/common.cxx:42
42          or_(1, cr0, cr0, cr0Enable);
(gdb)
```

### DIFF
--- a/src/gpu/intel/jit/ngen/ngen_config.hpp
+++ b/src/gpu/intel/jit/ngen/ngen_config.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
--- a/src/gpu/intel/jit/ngen/ngen_config.hpp
+++ b/src/gpu/intel/jit/ngen/ngen_config.hpp
@@ -31,4 +31,13 @@ using half = dnnl::impl::float16_t;
 #define NGEN_BFLOAT16_TYPE
 #define NGEN_HALF_TYPE
 
+#if (!defined(NDEBUG) || defined(DNNL_DEV_MODE)) && (__cplusplus >= 202002L || _MSVC_LANG >= 202002L)
+#if __has_include(<version>)
+#include <version>
+#if __cpp_lib_source_location >= 201907L
+#define NGEN_ENABLE_SOURCE_LOCATION true
+#endif
+#endif
+#endif
+
 #endif /* header guard */
--- a/src/gpu/intel/jit/binary_format.cpp
+++ b/src/gpu/intel/jit/binary_format.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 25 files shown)


src/gpu/intel/jit/ngen/ngen_config.hpp
```diff
@@ -30,5 +30,7 @@ using half = dnnl::impl::float16_t;
 
 #define NGEN_BFLOAT16_TYPE
 #define NGEN_HALF_TYPE
+#define NGEN_ENABLE_SOURCE_LOCATION true
```
petercad, line 33: I think we should only enable source_location on debug builds; similarly for use of `__FILE__` below. oneDNN users may not want to leak internal directory paths into released applications/libraries (especially closed-source products).
```suggestion
#define NGEN_ENABLE_SOURCE_LOCATION (!defined(NDEBUG))
```

rjoursler: The `__FILE__` macro is already used in many locations throughout oneDNN, for example, it is used in [common/verbose.hpp](https://github.com/oneapi-src/oneDNN/blob/main/src/common/verbose.hpp), [cpu/x64/jit_generator.hpp](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_generator.hpp). If this is a concern, it should be handled in a separate task addressing the library wide problem.

rjoursler:
make test
disable device_cpu
enable device_gpu

src/gpu/intel/jit/ngen/ngen_config.hpp
petercad: @rjoursler -- fair enough, though I think that's the only place (well, that and the aarch64 equivalent) where directory paths would be visible in the binary outside of debug mode. Though LTO could remove them...

@vpirogov What would you like to do here?

rjoursler:
make test
disable device_cpu
enable device_gpu

petercad (approved):

echeresh (approved):

rjoursler:
make test
disable device_cpu
enable device_gpu

rjoursler:
make test
disable device_cpu
enable device_gpu

PR #2298, opened 2024-12-19
common: workaround for GCC 8 format-truncation warning

spalicki:
Some SyCL and OCL builds with GCC host compiler fail due to false-positive format-truncation warning in first `snprintf`, this workaround prevents that.

### DIFF
--- a/src/common/utils.hpp
+++ b/src/common/utils.hpp
@@ -538,7 +538,14 @@ const char *format_cvt_impl(T &&t) {
 
 template <typename... Args>
 std::string format_impl(const char *fmt, Args... args) {
+    // volatile here is a workaround for GCC 8 format-truncation warning e.g.:
+    // ‘%d’ directive output truncated writing 1 byte into a region of size 0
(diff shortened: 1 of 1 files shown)


src/common/utils.hpp
```diff
@@ -538,7 +538,8 @@ const char *format_cvt_impl(T &&t) {
 
 template <typename... Args>
 std::string format_impl(const char *fmt, Args... args) {
-    size_t sz = snprintf(nullptr, 0, fmt, args...);
+    // volatile here is a workaround for GCC format-truncation warning
```
mgouicem, line 541: In the comment could you specify the compiler version that this fixes as well as the warning itself?

mgouicem (approved):
This is interesting, thanks. Does that warning happen only for gcc8, and not gcc9+ ?

spalicki:
> This is interesting, thanks. Does that warning happen only for gcc8, and not gcc9+ ?

Yes, all GCC 8 versions.

dzarukin (approved):

PR #2300, opened 2024-12-20
src: handle DNNL_VERBOSE=OFF support through an early exit (fixes MFDNN-12862)

dzarukin:
Fixes MFDNN-12862

Disabled verbose shouldn't print data even if requested. The change is complied with this requirement with a marginal price of the library size for disabled verbose build, but easier way to support this mode.

### DIFF
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -168,7 +168,7 @@ uint32_t get_verbose(verbose_t::flag_kind verbosity_kind,
         component_t::flag_kind filter_kind) noexcept {
 #if defined(DISABLE_VERBOSE)
     return verbose_t::none;
-#else
+#endif
     // we print all verbose by default
     static int flags = component_t::all;
     // record filter parsing result to instruct verbose printing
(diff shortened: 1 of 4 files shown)


dzarukin:
make test

src/common/verbose.cpp
```diff
@@ -308,29 +307,8 @@ bool get_verbose_timestamp() {
...
-    return std::string();
-}
-
-std::string rt_dims2fmt_str(primitive_kind_t prim_kind,
-        const memory_desc_t *src_md, const memory_desc_t *wei_md,
-        const memory_desc_t *dst_md) {
-    return std::string();
-}
-
-void verbose_printf_impl(const char *raw_fmt_str, verbose_t::flag_kind kind) {}
```
mgouicem, line 330: Could you clarify why these were not enough? At a glance, it seems that under DISABLE_VERBOSE, the behavior after the patch should be similar (return empty string or empty function).

dzarukin: The number of functions was growing, they are used across different places in verbose-reporting functionality. I decided that instead of handling each function under `if (disabled)` section (including future added) separately, it's just easier to force some of them to return empty strings under that build flag and call it a day. Should be no issues with future functions.

avmanerikar (approved):

TaoLv (approved):

echeresh (approved):

PR #2304, opened 2024-12-20
Fix sdpa descriptor hashing function

umar456:
# Description

The sdpa descriptor has hadn't been updated with the new scales and zero points parameters so the primitive caching mechanism was reusing primitives with invalid configurations. This PR updates the sdpa primitive hash functions to take those parameters into account.

### DIFF
--- a/src/common/primitive_hashing.cpp
+++ b/src/common/primitive_hashing.cpp
@@ -211,6 +211,43 @@ size_t get_md_hash(const memory_desc_t &md) {
     return seed;
 }
 
+// Generate a hash for runtime scales
+size_t get_runtime_scale_hash(const runtime_scales_t &scales) {
+    size_t seed = 0;
+    seed = hash_combine(seed, scales.mask_);
+    // scales: groups
+    const int ndims = scales.ndims_;
+    seed = hash_combine(seed, ndims);
+    if (ndims > 0) seed = get_array_hash(seed, scales.group_dims_, ndims);
+    // scales: data type
+    seed = hash_combine(seed, static_cast<size_t>(scales.data_type_));
+    return seed;
+}
+
+// Generate a hash for zero points
+size_t get_zero_points_hash(const zero_points_t &zps) {
+    size_t seed = 0;
+    for (int arg : {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST})
+        if (!zps.has_default_values(arg)) {
+            // zero_points: arg
+            seed = hash_combine(seed, arg);
+
+            int mask = 0;
+            zps.get(arg, &mask);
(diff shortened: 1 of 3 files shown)


src/common/primitive_hashing.cpp
```diff
@@ -757,6 +768,13 @@ size_t get_desc_hash(const sdpa_desc_t &desc) {
     // Scale type
     seed = hash_combine(seed, static_cast<size_t>(desc.scale_dt));
     seed = hash_combine(seed, desc.invert_scale);
+    seed = hash_combine(seed, desc.kv_head_number);
```
densamoilov, line 771: Do we also need to update this? https://github.com/oneapi-src/oneDNN/blob/f6f966988b9b0d0a3e1dcbabd4f6f7672bfb3390/src/common/serialization.cpp#L610

densamoilov: And also the comparison function: https://github.com/oneapi-src/oneDNN/blob/f6f966988b9b0d0a3e1dcbabd4f6f7672bfb3390/src/common/type_helpers.hpp#L972

umar456: Yes, Updating it now.

umar456:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_graph

src/common/primitive_hashing.cpp
```diff
@@ -757,6 +768,13 @@ size_t get_desc_hash(const sdpa_desc_t &desc) {
...
+    seed = hash_combine(seed, get_runtime_scale_hash(desc.kq_scales));
+    seed = hash_combine(
+            seed, get_zero_points_hash(desc.kq_zero_points, DNNL_ARG_WEIGHTS));
+    seed = hash_combine(seed, get_runtime_scale_hash(desc.vs_scales));
+    seed = hash_combine(
+            seed, get_zero_points_hash(desc.vs_zero_points, DNNL_ARG_WEIGHTS));
```
dzarukin, line 777: If dynamic quantization makes to the micro-kernel and DNNL_ARG_SRC would hit its usage, it would be not obvious to update this place. Suggest to create one more function that gets a hash for whole zero-points object. Applies to rest places.

dzarukin: Nit: please follow the order of members in `sdpa_desc_t`. It's easier to make sure nothing is left behind. APplies to rest places as well.

umar456:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_graph

dzarukin (approved):

densamoilov (approved):

PR #2305, opened 2024-12-20
fixups for CI workflows

vpirogov:
Few fixups for automation:
* Fixed syntax in labels.yml:
```
An unknown config option was under component:build: any-glob-to-any-file
An unknown config option was under component:examples: any-glob-to-any-file
```
* Added `component:build` label for `CMakeLists.txt`
* Changed `src/cpu/common/**` to `src/common/**` in AArch64 CI trigger
* Excluded `src/xpu` from x64 CI trigger


vpirogov:
Labeler workflow on this PR uses `labels.yml` from main (because of `pull_request_target` trigger), so warnings are still there.

theComputeKid (approved):
AArch64 changes look good. Didn't know that you needed those changes for labels.yml, I've been using what you originally had for ages for my own work.

PR #2309, opened 2024-12-23
Backport: cpu: x64: fixes matmul segfault 

xuxinzen:


tprimak:
v3.7-pc is already out

vpirogov:
There's no need to backport PRs to `-pc` branches unless explicitly requested. If the bug in question impacts oneDNN v3.6 it make sense to backport the fix to `rls-v3.6` branch.

PR #2314, opened 2024-12-24
benchdnn: graph: check partition number by default

TaoLv:
Currently on main branch, `--expected-n-partitions` is 0 by default (if not provided in benchdnn command line), which means to skip checking the partition number returned from an input graph.
This PR changes the default value of `--expected-n-partitions` to 1 which means the input graph should be fused into one partition.
Most of the cases we have in CI/Nightly should fused into one partition. With this check enabled by default, we can detect that if a PR breaks a fusion into multiple partitions. Before this change, those multiple partitions will silently pass the correctness check of benchdnn.
Apparently, we still have a few cases which cannot fuse into one partition due to:
- the cases previously were added for graph compiler backend.
- the cases can fuse on cpu but not on gpu, or vice versa.

Partition number check is skipped for these cases explicitly by setting `--expected-n-partitions=0` in the command lines. We can track and remove these settings when the fusions are implemented in the future.

### DIFF
--- a/tests/benchdnn/graph/graph.cpp
+++ b/tests/benchdnn/graph/graph.cpp
@@ -382,7 +382,7 @@ std::string case_to_str(const std::string &json_file,
         s << " ";
     }
 
-    if (expected_n_partitions != 0) {
+    if (expected_n_partitions != 1) {
         s << "--expected-n-partitions=" << std::to_string(expected_n_partitions)
           << " ";
     }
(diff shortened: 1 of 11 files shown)


TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/graph.cpp
```diff
@@ -557,6 +544,19 @@ int doit(const prb_t *prb, res_t *res) {
         return FAIL;
     }
 
+    if (prb->expected_n_partition != 0) {
+        // If the expected partition num is specified by user with command line
```
rongzha1, line 548: a little confusion in comments, **_1_** is default but not specified by cml

TaoLv: Thanks. I will remove this comment.

ElaineBao (approved):

xiang1guo (approved):

wzt1997 (approved):

tests/benchdnn/graph/graph.cpp
dzarukin: Maybe fuse two conditions:
```
if (prb->expected_n_partition > 0 && partitions.size() != prb->expected_n_partition) {
    ...
}
```

dzarukin (approved):

vpirogov (approved):

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

PR #2316, opened 2024-12-26
graph: backend: dnnl: enhance fp8 matmul patterns

wzt1997:
### Description

Enhance support for fp8 MatMul patterns in oneDNN Graph.

- Support more fp8 MatMul patterns.
- Add more test cases in benchdnn graph.

### DIFF
--- a/tests/benchdnn/inputs/graph/pattern/f8/f8_bf16_matmul_add_fusion.json
+++ b/tests/benchdnn/inputs/graph/pattern/f8/f8_bf16_matmul_add_fusion.json
@@ -1,0 +1,374 @@
+{
+  "version": "3.7.0",
+  "engine_kind": "cpu",
+  "fpmath_mode": "strict",
+  "fpmath_mode_apply_to_int": "false",
+  "input_ports": [
+    0, 
+    3, 
+    7, 
+    9
+  ],
+  "output_ports": [
+    10
+  ],
+  "graph": [
+    {
+      "id": 0,
+      "name": "DEQUANTIZE_1",
+      "kind": "Dequantize",
+      "attrs": {
+        "axis": {
+          "type": "s64",
+          "value": 0
+        },
+        "qtype": {
+          "type": "string",
+          "value": "per_tensor"
+        },
+        "scales": {
+          "type": "f32[]",
+          "value": [
+            1
+          ]
+        }
+      },
+      "inputs": [
+        {
+          "id": 0,
+          "dtype": "f8_e5m2",
+          "shape": [
+            16, 
+            1024
+          ],
+          "stride": [
+            1024, 
+            1
+          ],
+          "layout_type": "strided",
+          "property_type": "undef"
+        }
+      ],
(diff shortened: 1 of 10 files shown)


tests/benchdnn/inputs/graph/pattern/f8/f8_bf16_matmul_add_fusion.json
```diff
@@ -0,0 +1,365 @@
+{
+    "version": "3.0.0",
```
rongzha1, line 2: please update the json with the latest version

wzt1997: Thanks for reminding. Have fixed.

dzarukin (approved):

tests/benchdnn/inputs/graph/pattern/f8/f8_bf16_matmul_add_fusion.json
TaoLv: @wzt1997  changing the version number only is not enough. Some new fields are still missing in this json file: fpmath_mode_apply_to_int, input_ports, output_ports, etc.  It's confusing that a json file with version = 3.7.0 but without these fields.

wzt1997: I partly agree with you. For `input_ports` and `output_ports`, I agree we should align them in the test cases( although it seems they are ignored and do not affect the testing currently ). 
But, for other fields such as `fpmath_mode_apply_to_int`, according to the semantic, it only affects integral inputs, hence it does not have to appear here.

TaoLv: We don't have other reliable ways to generate graph json files besides dumping them from the library. If this one is dumped with v3.7 (as suggested by the version number), it should contain these fields, no matter if they take effect or not.

wzt1997: OK. I've updated the test case. Also it's found that there are many other test case with version=3.7.0 do not match the requirement, such as the single dynamic dequantize and gated-mlp. If necessary, we may need to enhance them with dumped json files.

TaoLv: The gated-mlp case was added before `fpmath_mode_apply_to_int` was introduced into the library. I would not insist on upgrading all legacy cases to the latest version of the library. Instead, I suggest when a new case is being added, it should comply with the latest version of the library.

rongzha1 (approved):

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

wzt1997:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

PR #2323, opened 2024-12-28
Fix for issue with generated CMake config file

vpirogov:
This PR is a follow up to #2225 including two changes:
* Restored `target_link_libraries_install` function to strip paths from libraries that end up in CMake config.
* Removes CMake policy settings [redundant for version 3.13](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-13).

Fixes MFDNN-12931.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,66 +16,6 @@
 
 cmake_minimum_required(VERSION 3.13)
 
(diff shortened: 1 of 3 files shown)


densamoilov (approved):

dzarukin (approved):

vpirogov:
make test
disable run_tests

PR #2326, opened 2024-12-30
build: switched from FindPythonInterp to FindPython2

vpirogov:
FindPythonInterp was deprecated in CMake 3.12 and removed in CMake 3.27.

With this PR oneDNN build system does not have any policy warnings with CMake 3.31, the latest version available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2016-2019 Intel Corporation
+# Copyright 2016-2024 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


dzarukin (approved):

spalicki (approved):

CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ if(POLICY CMP0082)
     cmake_policy(SET CMP0082 NEW)
 endif()
 
+# The FindPythonInterp and FindPythonLibs modules are removed.
```
spalicki, line 24: Can we add version of CMake this policy was added in to the comment? e.g.
`# 3.27: The FindPythonInterp and FindPythonLibs modules are removed.`

PR #2331, opened 2025-01-02
benchdnn: graph: extend --dt to support specifying tensor id

TaoLv:
We introduced `--dt` in 461388e801 to support floating-point data type rewrite for the whole graph.

In this PR, the knob is further extended to support data type rewrite for specified tensor IDs, in a format of `--dt=ID0:DT0+ID1:DT1...`. It provides more flexibility to test a json file with different data types, especially for the various zps/scales data types for quantizations which were excluded previously in whole graph data type rewrite. Due to the flexibility of this knob and the restriction of op definition in the library, users need to be cautious when rewriting the data type of a specific tensor in a graph as it may lead to graph construction failures in the library during testing (eg. incompatible input and output tensor data types).

 An example as below:
`benchdnn --graph --dt=0:s8+2:s8+6:s8+8:s8 --case=complex_fusion/mha/sdpa-compressed-kv-int4-gs32.json`

It can be used to test int8 compressed KV graph by rewriting rewrite the int4 compressed KV graph (0 is the key tensor, 2 is the zps pf key tensor, 6 is the value tensor, and 8 is the zps of values tensor).


tests/benchdnn/graph/bench_graph.cpp
```diff
@@ -31,16 +31,18 @@ void check_correctness(const settings_t &s) {
     for_(const auto &i_expected_n_partition : s.expected_n_partition_vec)
     for_(const auto &i_fpmath_mode : s.fpmath_mode_vec)
     for_(const auto &i_dt : s.dt)
+    for_(const auto &i_dts : s.dts)
```
rongzha1, line 34: what about rename them as i_id_dts and s.id_dts?  we already have ` i_dt ` , i_id_dts is more self-explain than i_dts.

tests/benchdnn/graph/parser.cpp
```diff
@@ -117,6 +117,34 @@ bool parse_op_attrs(std::vector<std::map<size_t, std::string>> &op_attrs_vec,
...
+        parse_key_value(dts_tmp, dts_str, option_name);
+        dts_map.clear();
+        // convert size_t:string to size_t:dnnl_data_type_t
+        for (const auto &e : dts_tmp) {
+            std::map<size_t, dnnl_data_type_t> tmp;
+            for (const auto &v : e) {
+                tmp[v.first] = str2dt(v.second.c_str());
+            }
+            dts_map.push_back(tmp);
+        }
```
dzarukin, line 142: It's probably cheaper to replace strings inplace in `dst_tmp[i]` and then do (pseudo-code):
```
dts_map.clear();
dts_map.resize(dts_tmp.size());
for (size()) {
    dst_map[i] = std::move(dts_tmp[i]);`
}
```
`push_back` will likely re-trigger many maps re-creation due to its capacity policy.

tests/benchdnn/graph/graph.hpp
```diff
@@ -83,7 +85,8 @@ std::string case_to_str(const std::string &json_file,
         const std::map<size_t, std::string> &op_attrs,
         const graph_fpmath_mode_t &fpmath_mode,
         const size_t expected_n_partitions, const int64_t mb,
-        const dnnl_data_type_t dt);
+        const dnnl_data_type_t dt,
```
dzarukin, line 88: Feels like this `dt` can be incorporated into `dts`. `--dt=f16` should be considered as just a convenient shortcut to `--dt=ID0:f16+ID1:f16+...`. Benchdnn can always report the latter form (new one) instead of the short.

tests/benchdnn/graph/graph.hpp
```diff
@@ -53,6 +53,8 @@ struct settings_t : public base_settings_t {
     std::vector<size_t> expected_n_partition_vec {1};
     std::vector<graph_fpmath_mode_t> fpmath_mode_vec {graph_fpmath_mode_t {}};
     std::vector<dnnl_data_type_t> dt {dnnl_data_type_undef};
+    std::vector<std::map<size_t, dnnl_data_type_t>> dts {
+            {{(size_t)-1, dnnl_data_type_undef}}};
```
dzarukin, line 57:
```suggestion
    std::vector<std::map<size_t, dnnl_data_type_t>> dts {
            {{SIZE_MAX, dnnl_data_type_undef}}};
```

tests/benchdnn/graph/parser.cpp
TaoLv: Thanks. I changed the code to reserve capacity for `dts_map` and replace `push_back` with assign.

tests/benchdnn/graph/graph.hpp
TaoLv: Thank you for the suggestion. I feel hesitate to change that as:
- Expanding `--dt=f16` to a list of `ID:DT` pairs requires to deserialize the json file firstly to get all the tensor IDs contained in the file. We may need to adjust the current benchmark flow to achieve that.
- For the repro string, it seems to me that replacing `--dt=f16` with `--dt=ID0:f16+ID1:f16+...` is confusing (and also lengthy) if `--dt=f16` is what the user specified in the command line.

tests/benchdnn/graph/graph.hpp
TaoLv: Thanks. It's fixed now.

tests/benchdnn/graph/flex_rewrite.cpp
```diff
@@ -1148,6 +1149,45 @@ void flex_rewrite::dt_rewrite(deserialized_graph &dgraph) {
...
 
+void flex_rewrite::dts_rewrite(deserialized_graph &dgraph) {
+    // check the IDs and data types in dts.
+    for (const auto &v : dts_) {
+        if (v.second == dnnl_data_type_undef) return;
+
+        bool found_id = false;
+        for (auto &aop : dgraph.ops_) {
+            for (auto &lt : aop.in_lts_) {
+                if (lt.id_ == v.first) found_id = true;
```
wzt1997, line 1160: Can we fuse the two for loops in the function, such as updating the data type if `lt.id_ == v.first`, and returning  `FAIL` if the id is not found?

wzt1997: OK. Thanks for the check.

wzt1997 (approved):

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
    field of all logical tensors in the input JSON file from `f32` to `f16`. If
    `--dt` is not specified or specified as `undef`, the original data types
    contained in the input JSON file will be used for testing.
+  - `--dt=ID:DT[+ID:DT]` -- Another format to specify the data types in the
+    input JSON file. `ID` specifies the input or output tensor of an operation
+    in the JSON file. `DT` is the target data type. Users can specify the data
```
ranukund, line 66:
```suggestion
    in the JSON file. `DT` is the target data type. To specify the data
```

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
...
+    types of multiple tensors using `+` to concatenate the `ID` and `DT` pairs.
+    An error will be raised if an `ID` is not contained in the JSON file. oneDNN
```
ranukund, line 68:
```suggestion
    An error will occur if `ID` is not contained in the JSON file. oneDNN
```

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
...
+    operations have restrictions for input and output tensor data types.
+    Changing the data type of a tesnor may lead to graph construction failures,
+    eg. fail to `add_op()` into graph.
```
ranukund, line 71:
```suggestion
    for example, failure to perform the `add_op()` operation in the graph.
```

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
    field of all logical tensors in the input JSON file from `f32` to `f16`. If
    `--dt` is not specified or specified as `undef`, the original data types
    contained in the input JSON file will be used for testing.
+  - `--dt=ID:DT[+ID:DT]` -- Another format to specify the data types in the
+    input JSON file. `ID` specifies the input or output tensor of an operation
+    in the JSON file. `DT` is the target data type. Users can specify the data
+    types of multiple tensors using `+` to concatenate the `ID` and `DT` pairs.
+    An error will be raised if an `ID` is not contained in the JSON file. oneDNN
+    operations have restrictions for input and output tensor data types.
+    Changing the data type of a tesnor may lead to graph construction failures,
```
ranukund, line 70:
```suggestion
    Changing the data type of a tensor may lead to graph construction failures,
```

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
    field of all logical tensors in the input JSON file from `f32` to `f16`. If
    `--dt` is not specified or specified as `undef`, the original data types
    contained in the input JSON file will be used for testing.
+  - `--dt=ID:DT[+ID:DT]` -- Another format to specify the data types in the
+    input JSON file. `ID` specifies the input or output tensor of an operation
+    in the JSON file. `DT` is the target data type. Users can specify the data
+    types of multiple tensors using `+` to concatenate the `ID` and `DT` pairs.
```
ranukund, line 67:
```suggestion
    types of multiple tensors, use `+` to concatenate the `ID` and `DT` pairs.
```

tests/benchdnn/doc/driver_graph.md
```diff
@@ -61,6 +61,14 @@ where *graph-knobs* are:
...
+    An error will be raised if an `ID` is not contained in the JSON file. oneDNN
+    operations have restrictions for input and output tensor data types.
+    Changing the data type of a tesnor may lead to graph construction failures,
+    eg. fail to `add_op()` into graph.
 
 and *graph-case* is a JSON file which is dumped by a library or created from scratch.
```
ranukund, line 73: Is this a continuation of the previous bullet?

ranukund (commented):
I've left a few comments, please incorporate as you see fit, thanks!

tests/benchdnn/doc/driver_graph.md
TaoLv: From the content, i think it should be a continuation of "where *graph-knobs* are:" at line 9.

tests/benchdnn/graph/bench_graph.cpp
TaoLv: Renamed.

TaoLv:
> I've left a few comments, please incorporate as you see fit, thanks!

Thank you @ranukund . Suggestions are incorporated now.

tests/benchdnn/doc/driver_graph.md
ranukund: I think it will make sense if we have graph knobs and graph case as two separate bullets and have nested formatting for all the other attributes.

For example,

Where

- [graph_knobs] can have the following attributes:
-  `--mb=INT` -- Override minibatch size specified in the JSON file default
    case. When set to `0`, use batch size as defined by the
    individual test..........
- [graph_case] is a JSON file......

TaoLv: Thanks for the suggestion. I converted them into a list in the latest commit.

ranukund (approved):
Looks good, thank you!

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/parser.cpp
```diff
@@ -117,6 +117,35 @@ bool parse_op_attrs(std::vector<std::map<size_t, std::string>> &op_attrs_vec,
...
+        std::vector<std::map<size_t, dnnl_data_type_t>> &dt_maps,
+        const char *str, const std::string &option_name) {
+    std::string dts_str;
+    if (!parse_string(dts_str, str, option_name)) return false;
+
+    if (dts_str.find(":") == std::string::npos) {
+        // format like --dt=f32,bf16,f16
+        const std::vector<dnnl_data_type_t> def_dt = {dnnl_data_type_undef};
+        parser::parse_dt(dts, def_dt, str, option_name);
+    } else {
```
dzarukin, line 130: Based on this logic it looks like the following will work:
```
--dt=f16 --dt=0:f32+... --case=some_case
```
`if` branch will update `dt`, then `else` branch will update `dt_map`, then `dt_rewrite` will update the whole graph, and then `dt_map_rewrite` will update specific IDs.
Is it an expected flow?
If yes, should it be properly documented?
If no, then it probably should be restricted here by additional checking whether the other object was updated or not.

tests/benchdnn/graph/parser.cpp
```diff
@@ -117,6 +117,35 @@ bool parse_op_attrs(std::vector<std::map<size_t, std::string>> &op_attrs_vec,
...
 }
 
+bool parse_dt(std::vector<dnnl_data_type_t> &dts,
+        std::vector<std::map<size_t, dnnl_data_type_t>> &dt_maps,
+        const char *str, const std::string &option_name) {
+    std::string dts_str;
+    if (!parse_string(dts_str, str, option_name)) return false;
+
+    if (dts_str.find(":") == std::string::npos) {
+        // format like --dt=f32,bf16,f16
```
dzarukin, line 127:
```suggestion
        // `dts` object: format like --dt=f32,bf16,f16
```

tests/benchdnn/graph/parser.cpp
```diff
@@ -117,6 +117,35 @@ bool parse_op_attrs(std::vector<std::map<size_t, std::string>> &op_attrs_vec,
...
+        const std::vector<dnnl_data_type_t> def_dt = {dnnl_data_type_undef};
+        parser::parse_dt(dts, def_dt, str, option_name);
+    } else {
+        // format like --dt=0:f32+1:f32,0:f16+1:f16
```
dzarukin, line 131:
```suggestion
        // `dt_maps` object: format like --dt=0:f32+1:f32,0:f16+1:f16
```

tests/benchdnn/graph/flex_rewrite.hpp
```diff
@@ -47,6 +48,7 @@ struct flex_rewrite {
     graph_fpmath_mode_t fpmath_mode_;
     int64_t mb_;
     dnnl_data_type_t dt_;
+    std::map<size_t, dnnl_data_type_t> dt_map_;
```
dzarukin, line 51:
```suggestion
    dnnl_data_type_t dt_; // Updates whole graph with a single dt value.
    std::map<size_t, dnnl_data_type_t> dt_map_; // Updates specific LT with selected dt values.
```

tests/benchdnn/graph/parser.cpp
TaoLv: Good catch! I did not expect the two --dt flavors would be specified in a single command line. I just added a check in the `bench()` function to error out.

tests/benchdnn/graph/bench_graph.cpp
```diff
@@ -65,13 +67,23 @@ int bench(int argc, char **argv) {
...
                 || parse_mb(s.mb, def.mb, argv[0]) || parse_reset(s, argv[0]);
+        // Graph driver does not support two styles of --dt in a single command line.
+        const bool has_dt = s.dt.size() != 1 || s.dt[0] != dnnl_data_type_undef;
+        const bool has_dt_map
+                = s.dt_map.size() != 1 || (s.dt_map[0].count(SIZE_MAX) == 0);
+        if (has_dt && has_dt_map) {
+            BENCHDNN_PRINT(0, "%s\n",
+                    "Error: --dt is specified twice with different styles.");
+            SAFE_V(FAIL);
+        }
```
dzarukin, line 85: I would move it into `parse_dt` since it has both objects and can check their state to figure out if both options were passed. Afterall, it's a parser obligation to make sure the input is valid (when it can).

TaoLv: Sure, it's moved to the parser function now.

dzarukin (approved):

TaoLv:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

PR #2333, opened 2025-01-02
benchdnn: binary: update filling (maybe fixed MFDNN-12857)

dzarukin:
MFDNN-12857

Make src1 input integer only to avoid potential cancellation in sdpa patterns.

### DIFF
--- a/tests/benchdnn/binary/binary.cpp
+++ b/tests/benchdnn/binary/binary.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
 *******************************************************************************/
 
 #include <algorithm>
+#include <random>
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


dzarukin:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_binary
enable benchdnn_graph

densamoilov (approved):

tests/benchdnn/binary/binary.cpp
```diff
@@ -54,17 +55,35 @@ int fill_mem(
...
-    const int f_min = (dt == dnnl_u8 || input_idx == 2) ? 0 : -range / 2;
-
-    benchdnn_parallel_nd(nelems, [&](int64_t i) {
-        const int64_t gen = (12 * i + 5 * input_idx + 16) % (range + 1);
-        const float scale = 1.25f;
-        float value = (f_min + gen) * scale;
-        // Remove zeroes in src1 to avoid division by zero
-        if (input_idx == 1 && value == 0.0f) value = 1.0f;
-        mem_fp.set_elem(i, round_to_nearest_representable(dt, value));
+    int min_val = MAX2(-8, static_cast<int>(lowest_dt(mem_dt.dt())));
```
mgouicem, line 58: It seems:
- older code path would generate random data between [0; 16] for unsigned types whereas the patch generates data between [0; 8], is that expected?
- not specific to this patch: why use a fixed interval of 16 values instead of mem_dt range?

dzarukin: > why use a fixed interval of 16 values instead of mem_dt range?

This patch targets the sdpa pattern when mask gets added to the matmul output. Adding to f16 anything big leads to rounding mismatch and to error amplification. f16 is really bad data type as its exponent range is very low. As we didn't explore much of f16 sdpa before, it's just "kind of new" (depends on shapes, actually). bf16 was fine due to a wider range. If it was purely for binary primitive, mem_dt would likely work.

> is that expected?

I realized over the time that it doesn't really matter, but the bigger the range, the more potential instabilities it can bring. I don't think we lose anything by decreasing it from 0-16 to 0-8.

mgouicem (approved):

PR #2337, opened 2025-01-03
src: cpu: x64: jit_brdgmm_dw_conw: make brdgmm support fusing sum op

amakarev:
Brdgmm kernel could not be selected when using the **--attr-post-ops=sum** flag. jit_dw was used instead.
This PR makes it possible to use brdgmm kernel with **sum post-op**

The changes look trivial, In fact, there was already support, but brdgmm could not be dispatched because the function(has_default_values) signature was updated a long time ago but the function call was not updated. As a result, data_type_undef was always used inside, which prevented the brdgmm kernel from being selected

### DIFF
--- a/src/cpu/x64/jit_brdgmm_dw_conv.cpp
+++ b/src/cpu/x64/jit_brdgmm_dw_conv.cpp
@@ -152,8 +152,8 @@ status_t brdgmm_dw_convolution_fwd_t::pd_t::init(engine_t *engine) {
     VDISPATCH_CONV(!has_zero_dim_memory(), VERBOSE_EMPTY_TENSOR, "");
     VDISPATCH_CONV(
             (isa != isa_undef) && mayiuse(isa), "undefined or unsupported isa");
-    VDISPATCH_CONV(
-            attr()->has_default_values(skip_mask), VERBOSE_UNSUPPORTED_ATTR);
+    VDISPATCH_CONV(attr()->has_default_values(skip_mask, dst_type),
+            VERBOSE_UNSUPPORTED_ATTR);
 
     auto &jcp = jcp_;
 


inteldimitrius (approved):
LGTM

ankalinin (approved):

dzarukin:
`dst_dt` argument is responsible to support s8 sum post-op while dst data type is u8, or vice versa. It doesn't have anything about enabling the sum post-op per se. `undef` data type passed requires sum post-op to check if it uses exactly the same data type as destination tensor, not allowing divergence from the first sentence.

It would be nice to see the following:
* There's actually some coverage for brdgmm + sum post+op in ci (or some nightly) file.
* The support flipped from jit to brgemm for a regular sum post-op case (same data type) and for a sum_dt != dst_dt (updating description with verbose before and after should be fine).

amakarev:
> `dst_dt` argument is responsible to support s8 sum post-op while dst data type is u8, or vice versa. It doesn't have anything about enabling the sum post-op per se. `undef` data type passed requires sum post-op to check if it uses exactly the same data type as destination tensor, not allowing divergence from the first sentence.
> 
> It would be nice to see the following:
> 
> * There's actually some coverage for brdgmm + sum post+op in ci (or some nightly) file.
> * The support flipped from jit to brgemm for a regular sum post-op case (same data type) and for a sum_dt != dst_dt (updating description with verbose before and after should be fine).

Got it, thanks for the clarification

dzarukin:
Thanks for providing the logs, I think I see what's going on now.
If user doesn't specify the data type for sum_dt, everything dispatches fine:
```
onednn_verbose,v1,primitive,create:cache_miss,cpu,convolution,brdgmm_dw:avx512_core,forward_inference,src:f32::blocked:acdb::f0 wei:f32:a:blocked:debcA16a::f0 bia:undef::undef::: dst:f32::blocked:acdb::f0,attr-scratchpad:user attr-post-ops:sum,alg:convolution_direct,g32mb2_ic32oc32_ih7oh5kh3sh1dh0ph0_iw7ow5kw3sw1dw0pw0,0.364014
```
but once it's specified, even to the actual `dst_dt`, it stops dispatching properly because of the check you are modifying:
```
onednn_verbose,v1,primitive,create:dispatch,convolution,cpu,convolution,brdgmm_dw:undef,forward_inference,src:f32::blocked:acdb::f0 wei:f32:a:any:any::f0 bia:undef::undef::: dst:f32::blocked:acdb::f0,attr-scratchpad:user attr-post-ops:sum:1:0:f32,alg:convolution_direct,g32mb2_ic32oc32_ih7oh5kh3sh1dh0ph0_iw7ow5kw3sw1dw0pw0,unsupported attribute,src/cpu/x64/jit_brdgmm_dw_conv.cpp:156
```

It looks quite unexpected to me, but it's probably not your war, so I'm fine to pass this change.

Hope you verified it's actually faster with brdgmm because the verbose from you doesn't show it is (though it's a first run, but still).

@mgouicem, we have a debate to talk through.

PR #2338, opened 2025-01-03
Enabled clang-tidy checks in header files

vpirogov:
Our current setup does not run any clang-tidy check on header files. One wrinkle is that we don't want it to touch any header files from third party components embedded in the source code, but current version of clang-tidy does not offer a way to exclude these. There are couple options:
* Uses `ExcludeHeaderFilterRegex` option available in clang-tidy >=19.0
* Move all third party components to separate folder (like `/third-party`) (in this PR)

With these changes CI will continue to scan source code only to avoid noise. Scan for headers will be available with `ONEDNN_USE_CLANG_TIDY=CHECK_ALL` option or manual `clan-tidy` invocation. With that we have 1390 [unique warnings](https://github.com/user-attachments/files/18367422/warnings.log) for Intel CPU/GPU build.

Related to MFDNN-12943.


src/gpu/intel/ocl/mdapi/.clang-format
```diff
@@ -0,0 +1,18 @@
+#===============================================================================
+# Copyright 2019-2020 Intel Corporation
```
asimonov1, line 2: Why 2020?

vpirogov: It's a copy of `.clang-format` from another location. Copyrights will be updated later once we figure out what to do with the new hits :)

theComputeKid:
We could temporarily add another non blocking check that checks for warnings in the headers, and allow the current one to remain while we fix them.

theComputeKid:
Also, 3p should ideally be in a separate folder anyways.

.clang-tidy
```diff
@@ -1,5 +1,12 @@
+# Enables checks in all project headers. Without this option all hits in
+# headers are ignored.
 HeaderFilterRegex: '.*'
 
+# This option requires clang-tidy >= v19.0. When enabled for older version
+# it renders whole file invalid and causes clang-tidy to use default
+# set of checks.
+ExcludeHeaderFilterRegex: '/src/(common/ittnotify|common/spdlog|cpu/aarch64/xbyak_aarch64|cpu/x64/xbyak|gpu/intel/jit/ngen|gpu/intel/sycl/l0/level_zero)/'
```
rjoursler, line 8: Do we have any standard support matrix for clang-tidy (especially with the poor fallback behavior)? In general, I don't think there is much reason to maintain backward compatibility for static-analyzers beyond the latest broadly distributed release. In my experience,  getting to use the latest improvements is a net benefit to code quality. I just wonder what counts as broadly distributed. For example with Ubuntu, this package is available in 24.10, but  is only in a [prerelease state](https://launchpad.net/ubuntu/noble/+package/clang-tidy-19) for the 24.10 LTS release. Does that seem sufficient for a broadly used distribution like Ubuntu?

If the v19.0 distribution is too recent, we should be able to rely on the equivalent command line argument via cmake to get a similar effect.

vpirogov: > If the v19.0 distribution is too recent, we should be able to rely on the equivalent command line argument via cmake to get a similar effect.

`--exclude-header-filter` was introduced in v19.0 as well

rjoursler: > --exclude-header-filter was introduced in v19.0 as well

True, I was meaning that we could check the clang-tidy version within CMake and then add or remove that command line option accordingly.

rjoursler: I think separating out 3rd party components is likely a good idea anyway. From a GPU perspective, I think there might be a couple of hijinks around the thirdparty nGEN and gemmstone components though as there is some unnecessary coupling within those projects (although extending MFDNN-12666 to target all third party projects could generally help enable upstream third-party development).

I am also just fine with requiring clang-tidy-19, rather than make these checks dependent on that refactoring. It might be be useful to see if there are any newer checks implemented in clang-tidy which would further motivate this change.

vpirogov: How does this look?

I did not touch `src/gpu/intel/jit/gemm` yet, as I'm not exactly clear on code structure there.

rjoursler: That looks great. Thanks.

> I did not touch src/gpu/intel/jit/gemm yet, as I'm not exactly clear on code structure there.

I think this patch should accomplish that: [0001-build-relocation-src-gpu-intel-jit-gemm-gemmstone-fi.patch](https://github.com/user-attachments/files/18349697/0001-build-relocation-src-gpu-intel-jit-gemm-gemmstone-fi.patch), although there is some cross-pollination that would be nice to remove in the future:
```Bash
$ git grep "gpu/" ~/dnnl/third_party/gemmstone/
third_party/gemmstone/include/emulation.hpp:#include "gpu/intel/jit/emulation.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/gpu_post_ops.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/jit/jit_generator.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/jit/jit_post_op_injector.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/serialization.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/microkernels/entrance_agent.hpp"
third_party/gemmstone/include/generator.hpp:#include "gpu/intel/microkernels/package.hpp"
third_party/gemmstone/include/microkernel_provider.hpp:#include "gpu/intel/microkernels/package.hpp"
third_party/gemmstone/include/problem.hpp:#include "gpu/intel/gpu_post_ops.hpp"
```

vpirogov: That's a bit too much of entanglement to consider it a third party component...

vpirogov: Some stuff in `src/gpu/intel/microkernels`  apparently also comes from gemmstone. I'll just disable clang-tidy checks on these paths for now.

vpirogov:
make test

vpirogov:
make test Lightweight_scans

vpirogov:
make test Lightweight_scans

mgouicem (approved):

theComputeKid:
To turn this on for AArch64, do I need to define `DNNL_USE_CLANG_TIDY` as `CHECK_ALL`? I want to add this to CI.

theComputeKid (approved):

vpirogov:
> To turn this on for AArch64, do I need to define `DNNL_USE_CLANG_TIDY` as `CHECK_ALL`? I want to add this to CI and see how bad it is..

That's correct. You'll see warnings from common parts of the code though in this case, so maybe make sure that 'CHECK' is clean first for AArch64.

densamoilov (approved):

vpirogov:
make test

PR #2339, opened 2025-01-03
xe: enable source code information for OpenCL kernels

rjoursler:
As the difficult case for handling nGEN was already handled in #2297, it seems only fitting to enable debug information in the simpler case where we use OpenCL C kernels. This PR limits the scope of support for this to Linux due to the nuances required for handling temporary files. It should be fairly easy to extend with Windows support when it becomes useful.

### DIFF
--- a/cmake/gen_gpu_kernel_list.cmake
+++ b/cmake/gen_gpu_kernel_list.cmake
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2020-2021 Intel Corporation
+# Copyright 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -46,6 +46,11 @@ endfunction()
 function(gen_gpu_kernel_list ker_list_templ ker_list_src ker_sources headers)
     set(_sources "${SOURCES}")
 
+    set(MINIFY "ON")
+    if(DNNL_DEV_MODE OR CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(MINIFY "OFF")
+    endif()
+
     set(KER_LIST_EXTERN)
     set(KER_LIST_ENTRIES)
     set(KER_HEADERS_EXTERN)
@@ -62,6 +67,7 @@ function(gen_gpu_kernel_list ker_list_templ ker_list_src ker_sources headers)
             COMMAND ${CMAKE_COMMAND}
                 -DCL_FILE="${header_path}"
                 -DGEN_FILE="${gen_file}"
+                -DMINIFY="${MINIFY}"
                 -P ${PROJECT_SOURCE_DIR}/cmake/gen_gpu_kernel.cmake
             DEPENDS ${header_path}
         )
@@ -81,6 +87,7 @@ function(gen_gpu_kernel_list ker_list_templ ker_list_src ker_sources headers)
             COMMAND ${CMAKE_COMMAND}
                 -DCL_FILE="${ker_path}"
                 -DGEN_FILE="${gen_file}"
+                -DMINIFY="${MINIFY}"
                 -P ${PROJECT_SOURCE_DIR}/cmake/gen_gpu_kernel.cmake
             DEPENDS ${ker_path}
         )
--- a/src/gpu/intel/compute/kernel.hpp
+++ b/src/gpu/intel/compute/kernel.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


rjoursler:
make test
disable device_cpu
enable device_gpu

petercad (approved):

src/gpu/intel/compute/kernel.hpp
```diff
@@ -34,6 +34,43 @@ namespace gpu {
...
+#if defined(__linux__) && (defined(DNNL_DEV_MODE) || !defined(NDEBUG))
+struct program_src_t {
+    program_src_t() = default;
+    program_src_t(const std::string &src_str) {
+        // Only enable if gdb-oneapi debugging is active
+        if (getenv_int("ZET_ENABLE_PROGRAM_DEBUGGING", 0) == 0) return;
+
+        const int name_size = 29;
+        char name[name_size] = "/tmp/dnnl_ocl_jit_src.XXXXXX";
+        int fd = mkstemp(name);
```
mgouicem, line 46: @avmanerikar @vpirogov 

Is creating temporary files this way compatible with our SDL?

src/gpu/intel/ocl/ocl_gpu_engine.cpp
```diff
@@ -249,6 +250,9 @@ status_t ocl_gpu_engine_t::build_program_from_source(
     std::string pp_code_str = pp_code.str();
     const char *pp_code_str_ptr = pp_code_str.c_str();
 
+    src = {pp_code_str};
+    if (src) { options += " -g -s " + std::string(src.name()); }
```
mgouicem, line 254: It seems to be the only place where program_src_t is needed.
Why introduce program_src_t, and not do everything in this function directly?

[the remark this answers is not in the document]
rjoursler: The `-s` specifies the source location to use for debug information. For more details, see https://www.intel.com/content/www/us/en/docs/distribution-for-gdb/tutorial-debugging-dpcpp-linux/2023-0/debugging-an-opencl-application.html.

rjoursler: The goal here is that this file is temporary and is deleted after it is no longer necessary. In particular, this means that this file is necessary so long the resulting `compute::kernels` is instantiated, so that this file can be referenced by a debugger.  The `program_src_t` structure is largely responsible for handling this lifetime management. We will also need a similar file if we are to introduce debug information for our [ir based](https://github.com/oneapi-src/oneDNN/tree/main/src/gpu/intel/jit/ir) kernels as currently all nGEN debug information just maps to the code which [lowers the IR to nGEN](https://github.com/oneapi-src/oneDNN/blob/main/src/gpu/intel/jit/codegen/codegen.cpp).

cmake/gen_gpu_kernel_list.cmake
```diff
@@ -46,6 +46,11 @@ endfunction()
 function(gen_gpu_kernel_list ker_list_templ ker_list_src ker_sources headers)
     set(_sources "${SOURCES}")
 
+    set(MINIFY "ON")
+    if(DNNL_DEV_MODE OR CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(MINIFY "OFF")
```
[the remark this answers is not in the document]
rjoursler: In testing, I do not see any measurable impact on compilation time.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
mgouicem: ok I see. Indeed we need to pass something around, thanks.

src/gpu/intel/compute/kernel.hpp
avmanerikar: SDL requires that temporary files and directories must be created with minimal permissions. While using `mkstemp()` will take care of this via exclusive access, setting a predictable filename `"/tmp/dnnl_ocl_jit_src.XXXXXX"` still leaves it open to symlink attacks. 
The recommendation would be to set the file permissions explicitly or use [`fopen()`](https://github.com/oneapi-src/oneDNN/blob/c39b83bc20f1f0afbba3cdc86c9925d85214adb3/src/common/utils.cpp#L151) which performs the necessary checks.

rjoursler: Thanks, I decided against using `dnnl::impl::fopen` as that introduces a race condition between checking if a temporary filename is valid and creating the file which requires additional logic to handle. To address your suggestions, I added a check that `/tmp` is not a symlink to guard against symlink attacks. In addition, per https://man7.org/linux/man-pages/man3/mkstemp.3.html

> The file is opened with the [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) O_EXCL flag, guaranteeing that the caller is the process that creates the file. 

the file created by mkstemp is the file created by this process, and thus not a symlink. In regards to permissions, from the same source above:

>  The file is created with permissions 0600, that is, read plus write for owner only

As this file needs it contents written to and subsequently read by the debugger, these are the minimum required. 

Finally, I modified file deletion to follow the pattern `unlink(name); close(fd)`, ensuring that `name` cannot map to a different file descriptor upon file deletion. 

Please let me know if I missed anything.

avmanerikar: The fix looks good. Thanks for addressing this!

avmanerikar (approved):

rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

PR #2341, opened 2025-01-06
cpu: x64: enable matmul-based IP for forward inference

densamoilov:
This PR introduces a matmul-based inner product implementation for inference that will be used instead of the current brgemm-based one. The latter will be disabled for inference.

This PR is targeting oneDNN v3.8 and will be merged after the oneDNN v3.7 code freeze date.

The implementation allows using blocked weights layouts directly or via the special tag `any`.
The Inner Product weights must meet **ONE** of the following requirements to enable using the blocked layouts:
 * Weights don't have spatial.
 * Weights have unit spatial.
 * Weights have non-unit spatial but the number of input channels is a multiple of K block.

If none of the above requirements are met then a plain layout will be used.

The new matmul-based implementation is able to leverage blocked weights layouts for all cases used in the performance testing and therefore provides the best possible performance.
Below is a comparison of the matmul-based vs brgemm-based inner product. > 100% means matmul-based is better.

image _________________ image

### DIFF
--- a/src/cpu/x64/jit_brgemm_inner_product.hpp
+++ b/src/cpu/x64/jit_brgemm_inner_product.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2024 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -67,7 +67,9 @@ struct brgemm_inner_product_fwd_t : public primitive_t {
             // better readability
             if (!mayiuse(isa)) return status::unimplemented;
 
-            VDISPATCH_INNER_PRODUCT(is_fwd(), VERBOSE_BAD_PROPKIND);
+            VDISPATCH_INNER_PRODUCT(
+                    get_prop_kind() == prop_kind::forward_training,
+                    VERBOSE_BAD_PROPKIND);
             VDISPATCH_INNER_PRODUCT(
                     expect_data_types(src_dt, wei_dt, data_type::undef, dst_dt,
                             data_type::undef),
--- a/src/cpu/x64/matmul_inner_product.cpp
+++ b/src/cpu/x64/matmul_inner_product.cpp
@@ -1,0 +1,276 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <map>
+
+#include "cpu/x64/matmul_inner_product.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace x64 {
+
+status_t create_matmul_pd(std::shared_ptr<primitive_desc_t> &matmul_pd,
+        engine_t *engine, const memory_desc_t *src_md,
+        const memory_desc_t *wei_md, const memory_desc_t *dst_md,
+        const memory_desc_t *bia_md, const primitive_attr_t *attr) {
+    auto matmul_desc = matmul_desc_t();
+
+    CHECK(matmul_desc_init(&matmul_desc, src_md, wei_md, bia_md, dst_md));
+
+    primitive_desc_iterator_t it(
+            engine, (op_desc_t *)&matmul_desc, attr, nullptr);
+
+    matmul_pd = *(++it);
+    if (!matmul_pd) return status::unimplemented;
+
+    return status::success;
+}
+
+status_t init_matmul_md(memory_desc_t &mm_md, const memory_desc_t &ip_md,
+        format_tag_t tag, bool swap_dims) {
+    auto p_dims = ip_md.dims;
+    auto p_dim1 = utils::array_product(p_dims + 1, ip_md.ndims - 1);
+
+    if (swap_dims) {
+        dims_t dims_2d = {p_dim1, p_dims[0]};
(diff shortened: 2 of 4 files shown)


densamoilov:
make test
enable device_cpu
disable device_gpu

src/cpu/x64/jit_brgemm_inner_product.hpp
```diff
@@ -67,7 +67,9 @@ struct brgemm_inner_product_fwd_t : public primitive_t {
             // better readability
             if (!mayiuse(isa)) return status::unimplemented;
 
-            VDISPATCH_INNER_PRODUCT(is_fwd(), VERBOSE_BAD_PROPKIND);
+            VDISPATCH_INNER_PRODUCT(
+                    get_prop_kind() == prop_kind::forward_training,
+                    VERBOSE_BAD_PROPKIND);
```
mgouicem, line 72: Could you clarify why jit_brgemm_ip gets restricted to forward training?

IIUC, jit_brgemm_ip is after the new matmul_ip in the dispatch list, so there is a chance that if a forward inference case is not handled by matmul_ip (per the documented restrictions), it will also be skipped by jit_brgemm_ip and will go to a lower performance impl.

densamoilov: > so there is a chance that if a forward inference case is not handled by matmul_ip (per the documented restrictions)

If a blocked layout cannot be used for weights then it'll fall back to a plain layout so all cases that can be handled by brgemm-ip can also be handled by matmul-based ip.

The goal is to remove brgemm-ip completely so we can't use it as a complement.

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -0,0 +1,271 @@
...
+                src_md_.data_type, src_tag));
+    else if (!memory_desc_matches_tag(*src_md(), src_tag))
+        return status::unimplemented;
+
+    const auto dst_tag = ab;
+    if (dst_md()->format_kind == format_kind::any)
+        CHECK(memory_desc_init_by_tag(dst_md_, dst_md_.ndims, dst_md_.dims,
+                dst_md_.data_type, dst_tag));
+    else if (!memory_desc_matches_tag(*dst_md(), dst_tag))
+        return status::unimplemented;
```
dzarukin, line 178: You probably want all those unimplemented status returned to be wrapped in verbose call to save time figuring out which one failed.

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -0,0 +1,271 @@
...
+                VERBOSE_UNSUPPORTED_ATTR);
+    else if (wei_mask != 0)
+        return status::unimplemented;
+
+    CHECK(create_matmul_pd(matmul_pd_, engine, &mm_src_md, &mm_wei_md,
+            &mm_dst_md, with_bias() ? weights_md(1) : nullptr, &matmul_attr));
+
+    // Fallback to a generic GEMM-based Inner Product is preferred rather
+    // than using a reference or GEMM-based MatMul implementations here.
+    if (std::string(matmul_pd_->name()).find("ref") != std::string::npos)
```
dzarukin, line 197: Minor: maybe use the inversion to include brg_matmul only to be more precautious?

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -0,0 +1,271 @@
...
+        return status::unimplemented;
+    if (std::string(matmul_pd_->name()).find("gemm") != std::string::npos)
+        return status::unimplemented;
+
+    if (weights_md()->format_kind == format_kind::any) {
```
dzarukin, line 202:
```suggestion
    // Initializing IP weights format based on Matmul weights format.
    if (weights_md()->format_kind == format_kind::any) {
```

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -0,0 +1,271 @@
...
+    using namespace memory_tracking::names;
+
+    exec_args_t matmul_args = ctx.args();
+    exec_ctx_t matmul_ctx(ctx, std::move(matmul_args));
+
+    nested_scratchpad_t ns(ctx, key_nested, matmul_);
+    matmul_ctx.set_scratchpad_grantor(ns.grantor());
+
+    return matmul_->execute(matmul_ctx);
+    return status::success;
```
dzarukin, line 265: Remove?

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+status_t init_matmul_md(memory_desc_t &mm_md, const memory_desc_t &ip_md,
+        format_tag_t tag, bool swap_dims) {
+    auto p_dims = ip_md.dims;
+    auto p_dim1 = utils::array_product(p_dims + 1, ip_md.ndims - 1);
+
+    if (swap_dims) {
+        dims_t dims_2d = {p_dim1, p_dims[0]};
```
dzarukin, line 63: ?
```suggestion
        dims_t dims_2d = (swap_dims ? {p_dim1, p_dims[0]} : {p_dims[0], p_dim1});
```

dzarukin (approved):

src/cpu/x64/matmul_inner_product.cpp
densamoilov: This syntax is not legal in the language. The closest thing we could get is using the ternary operator for each dimension but the code would be hard to read.
`dims_t dims_2d = {swap_dims ? p_dim1 : p_dims[0], swap_dims ? p_dims[0] : p_dim1};`

ankalinin (approved):

src/cpu/x64/matmul_inner_product.cpp
densamoilov: Fixed

src/cpu/x64/matmul_inner_product.cpp
densamoilov: Added

src/cpu/x64/matmul_inner_product.cpp
densamoilov: Replaced the condition

src/cpu/x64/matmul_inner_product.cpp
densamoilov: Done, thanks for pointing this out!

densamoilov:
make test
enable device_cpu
disable device_gpu

PR #2345, opened 2025-01-07
xe: conv_v2: enable Stream-K kernels

echeresh:
Jira: MFDNN-11721

PR updates performance modeling and benchmarking logic to handle Stream-K, and also updates the registry to use new Stream-K kernels.

ResNet-50 performance data on PVC is below. A few comments:

- Excluded cases: first convolution and strided backward by data
    - 1st convolution is very slow now, needs special tiling approach and maybe a reorder (to be covered in the next month)
    - Strided backward by data is not supported, needs dynamic filter indexing (also to be covered soon)
- Large batch performance is relatively good though this is mostly a baseline, and more optimization work is underway

In the table below ratios > 1 mean that "v2-new" is faster.

| Propagation | Batch size | Data Type | # of layers | Ratio v2-new to v2-old | Ratio v2-new to v1 |
|-------------|------------|-----------|-------------|--------------------------------------------|--------------------------------------------|
| FWD_I | 1 | s8 | 22 | 1.22 | 0.88 |
| FWD_I | 1 | bf16 | 22 | 0.87 | 0.80 |
| FWD_I | 1 | f32 | 22 | 1.01 | 0.78 |
| FWD_I | 32 | s8 | 22 | 0.93 | 0.86 |
| FWD_I | 32 | bf16 | 22 | 1.02 | 0.78 |
| FWD_I | 32 | f32 | 22 | 1.01 | 0.90 |
| BWD_D | 32 | bf16 | 16 | 0.96 | 1.29 |
| BWD_D | 32 | f32 | 16 | 0.99 | 0.97 |
| BWD_W | 32 | bf16 | 22 | 3.08 | 0.73 |
| BWD_W | 32 | f32 | 22 | 2.32 | 0.93 |
| FWD_I | 128 | s8 | 22 | 0.97 | 0.95 |
| FWD_I | 128 | bf16 | 22 | 1.00 | 0.80 |
| FWD_I | 128 | f32 | 22 | 1.03 | 0.92 |
| BWD_D | 128 | bf16 | 16 | 0.99 | 0.92 |
| BWD_D | 128 | f32 | 16 | 1.00 | 0.97 |
| BWD_W | 128 | bf16 | 22 | 6.24 | 0.74 |
| BWD_W | 128 | f32 | 22 | 2.74 | 0.87 |


echeresh:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe2-lpg
enable arch_xe-hpg
enable arch_xe-hpc

echeresh:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe2-lpg
enable arch_xe-hpg
enable arch_xe-hpc

echeresh:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe2-lpg
enable arch_xe-hpg
enable arch_xe-hpc

src/gpu/intel/jit/v2/conv/planner/bench.cpp
```diff
@@ -614,6 +614,7 @@ bench_data_t bench(const bench_manager_t &bench_mger,
 
 bool try_create(
         const bench_manager_t &bench_mger, const kernel_desc_t &kernel_desc) {
+    clear_primitive_cache();
```
rjoursler, line 633: Would it be reasonable to just set the primitive cache capacity to 0?

rjoursler:
Thanks for the data comparing to `v1`. What is the before and after performance improvement from this optimization?

src/gpu/intel/serialization.hpp
```diff
@@ -239,6 +239,8 @@ struct deserializer_t {
         }
     }
 
+    bool empty() const { return idx >= s.get_data().size(); }
```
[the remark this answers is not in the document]
echeresh: I think this is not the same, "empty" for deserializer means we reached the end, and the original data may be non-empty.

src/gpu/intel/jit/utils/utils.hpp
```diff
@@ -1344,26 +1344,34 @@ void stringify_to_cpp_file(const std::string &file_name,
...
-std::string serialize_to_hex(const T &t) {
+inline std::string data_to_hex(const std::vector<uint8_t> &data) {
     std::ostringstream oss;
-    serialized_data_t s;
-    s.append(t);
-    for (uint8_t d : s.get_data()) {
+    for (auto v : data) {
         oss << std::uppercase << std::hex << std::setw(2) << std::setfill('0')
-            << (int)d;
+            << (int)v;
```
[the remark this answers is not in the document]
echeresh: Thanks, changed.

src/gpu/intel/jit/v2/conv/planner/bench.cpp
echeresh: We still want to enable kernel cache to reduce primitive creation time in general. In a typical benchmark run we create 100+ primitives with different shapes but all of them use the same kernel underneath - kernel/primitive cache helps with that.

A caveat here is that shapes may repeat, and we want to test the same shape with a different kernel. In such places we need to clear the cache to avoid undesired primitive cache hits. However we still need primitive cache enabled for other cases.

echeresh:
> Thanks for the data comparing to `v1`. What is the before and after performance improvement from this optimization?

Sure, added another column to compare with the current `v2`.

echeresh:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe-hpg
enable arch_xe-hpc
enable arch_xe2-lpg
enable arch_xe2-hpg-bmg

src/gpu/intel/jit/v2/conv/planner/bench.cpp
rjoursler: This is not working as you intend. The external [primtive_cache interace](https://github.com/oneapi-src/oneDNN/blob/6bc11a06251a230b8fd2c3ab8b355c37e881045e/src/common/primitive_cache.cpp#L150-L157) (called from [here](https://github.com/oneapi-src/oneDNN/blob/6bc11a06251a230b8fd2c3ab8b355c37e881045e/src/gpu/intel/jit/v2/conv/planner/bench.cpp#L518))  will also clear the kernel cache because the kernel_cache is not intended to be externally visible. You will need to add a new interface or environment variable to get your desired behavior.

echeresh: Got you now. So for your suggestion we need to introduce an interface to control kernel/primitive caches separately and use it instead (now these abstractions are hidden even internally).

I implemented this in the last commit, please take a look.

Also tagging @densamoilov to review primitive cache changes.

mgouicem (approved):
Common part looks fine, thanks.
Would it make sense to update benchdnn with the new profiling kind?

echeresh:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_xe-hpg
enable arch_xe-hpc
enable arch_xe2-lpg
enable arch_xe2-hpg-bmg

echeresh:
> Common part looks fine, thanks. Would it make sense to update benchdnn with the new profiling kind?

The new profiling kind is not really new, it is to access individual kernel times which are always recorded.

But in regular benchmarking via bennchdnn we usually need the elapsed/wall time (end of last kernel minus start of the first kernel). It is more representative. For more in-depth performance analysis I guess using profiling tools (like OpenCL intercept layer) is the right option.

PR #2349, opened 2025-01-07
graph: backend: dnnl: support select with binary primitive

Jiexin-Zheng:
# Description
1. `cond` input is defined for dnnl binary op
2. For now, as primitive doesn't support broadcast for `cond` input, we use binary select primitive for non-broadcast case only, the lowering logic is: always lower select to binary primitive and then decide which impl path to use in pass `decompose_select_to_multiple_binary_ops` and decompose it to multiple binary ops if necessary.
3. It's unsupported on GPU as binary primitive doesn't support either.

# Performance
relative perf:

platform: Intel(R) Xeon(R) Platinum 8490H

case | speedup
-- | --
./tests/benchdnn/benchdnn --graph --mode=P --reset --in-shapes=1:1x12x128x128+2:1x12x128x128 --case=complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json | 98.24%
./tests/benchdnn/benchdnn --graph --mode=P --reset --dt=bf16 --in-shapes=1:1x12x128x128+2:1x12x128x128 --case=complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json | 170.12%
./tests/benchdnn/benchdnn --graph --mode=P --reset --in-shapes=1:1x12x128x128+2:1x12x128x128 --case=complex_fusion/mha/MHA-distill_bert-inf-int8-bs1.json | 140.41%

### DIFF
--- a/src/graph/backend/dnnl/dnnl_op_def.hpp
+++ b/src/graph/backend/dnnl/dnnl_op_def.hpp
@@ -701,6 +701,7 @@ DNNL_GRAPH_OP_SCHEMA(dnnl_binary, 1,
                 .set_num_outputs(2)
                 .set_input(0, "a")
                 .set_input(1, "b")
+                .set_input(2, "cond")
                 .set_output(0, "output")
                 .set_output(1, "scratchpad")
                 // Attributes inherited from front binary ops (Add, Multiply,
--- a/src/graph/backend/dnnl/dnnl_shape_infer.cpp
+++ b/src/graph/backend/dnnl/dnnl_shape_infer.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -15,9 +15,9 @@
 *******************************************************************************/
 
 #include <algorithm>
-#include <unordered_set>
-
 #include "graph/interface/shape_infer.hpp"
+#include "oneapi/dnnl/dnnl.hpp"
+#include <unordered_set>
 
 #include "graph/backend/dnnl/dnnl_shape_infer.hpp"
 #include "graph/backend/dnnl/internal_attrs.hpp"
@@ -484,17 +484,65 @@ status_t infer_dnnl_pool_bwd_output_shape(op_t *n,
     return status::success;
 }
 
+status_t infer_binary_select_output_shape(op_t *n,
+        std::vector<logical_tensor_t *> &inputs,
+        std::vector<logical_tensor_t *> &outputs) {
+    auto in0 = logical_tensor_wrapper_t(inputs[0]);
+    auto in1 = logical_tensor_wrapper_t(inputs[1]);
+    auto in2 = logical_tensor_wrapper_t(inputs[2]);
+
+    const bool shapes_should_match = n->has_attr(op_attr::auto_broadcast)
+            ? "none" == n->get_attr<std::string>(op_attr::auto_broadcast)
+            : false;
+
+    dims input0_dims = in0.vdims();
+    dims input1_dims = in1.vdims();
+    dims input2_dims = in2.vdims();
+    dims inferred_out_shape;
+
+    if (shapes_should_match) { // no broadcast
+        VCHECK_INVALID_SHAPE(
+                (input0_dims == input1_dims && input1_dims == input2_dims),
+                "%s, all input dims should match each other if there is no "
+                "broadcast. input0 dims: %s, input1 dims: %s, input2 dims: %s ",
+                op_t::kind2str(n->get_kind()).c_str(),
+                dims2str(input0_dims).c_str(), dims2str(input1_dims).c_str(),
+                dims2str(input2_dims).c_str());
+        inferred_out_shape = std::move(input0_dims);
+    } else { // can broadcast
+        status_t ret1 = broadcast(input0_dims, input1_dims, inferred_out_shape);
+        VCHECK_INVALID_SHAPE((ret1 == status::success),
+                "%s, failed to implement numpy broadcasting",
+                op_t::kind2str(n->get_kind()).c_str());
+    }
+
+    auto out0 = logical_tensor_wrapper_t(outputs[0]);
+    // check if given or partial set shape aligns with inferred shape
+    if (!out0.is_shape_unknown() || out0.ndims() != -1) {
+        VCHECK_INVALID_SHAPE(validate(inferred_out_shape, out0.vdims()),
+                "%s, inferred out shape and output shape are not compatible",
+                op_t::kind2str(n->get_kind()).c_str());
+        if (!out0.is_shape_unknown()) return status::success;
+    }
+
+    set_shape_and_strides(*outputs[0], inferred_out_shape);
+    return status::success;
+}
+
 status_t infer_dnnl_binary_output_shape(op_t *n,
         std::vector<logical_tensor_t *> &inputs,
         std::vector<logical_tensor_t *> &outputs) {
     const bool is_bias_add = n->has_attr(op_attr::is_bias_add)
             && n->get_attr<bool>(op_attr::is_bias_add);
-
-    auto ret = is_bias_add
-            ? infer_bias_add_output_shape(n, inputs, outputs)
-            : infer_elemwise_arithmetic_output_shape(n, inputs, outputs);
-
-    return ret;
+    const algorithm algo = static_cast<dnnl::algorithm>(
+            n->get_attr<int64_t>(op_attr::alg_kind));
+    if (algo == algorithm::binary_select) {
+        return infer_binary_select_output_shape(n, inputs, outputs);
+    } else if (is_bias_add) {
+        return infer_bias_add_output_shape(n, inputs, outputs);
+    } else {
+        return infer_elemwise_arithmetic_output_shape(n, inputs, outputs);
+    }
 }
 
 } // namespace dnnl_impl
--- a/src/graph/backend/dnnl/kernels/sdp_decomp.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -60,6 +60,8 @@ status_t sdp_decomp_kernel_t<quantized, dt>::compile_impl(
     pass_pipeline_t pipeline = pass_pipeline_t(vis);
     pass_pipeline_t select_pipeline = pass_pipeline_t(vis);
     BACKEND_DNNL_ADD_PASS(pipeline, lower_down);
+    // Decompose select to binary ops if necessary
+    BACKEND_DNNL_ADD_PASS(pipeline, decompose_select_to_binary_ops);
     BACKEND_DNNL_ADD_PASS(pipeline, fuse_reshape_for_gqa);
     // Fusion and canonicalization passes begin
     if (quantized) {
--- a/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -56,6 +56,23 @@ bool sdp_decomp_config_t::initial_check(const std::shared_ptr<subgraph_t> &sg,
         auto scale_sz = ltw(inputs[graph_inport[2]]).nelems();
         VCHECK_SDP_DECOMP(scale_sz == 1, false,
                 "Only supports single scale value, but got %lld", scale_sz);
+    }
+
+    // Check select cond and src0 shape
+    if (graph_inport[5] != -1 && graph_inport[6] != -1) {
+        const auto select_cond_dims = ltw(inputs[graph_inport[5]]).vdims();
+        const auto select_src0_dims = ltw(inputs[graph_inport[6]]).vdims();
+        VCHECK_SDP_DECOMP(select_cond_dims.size() == select_src0_dims.size(),
+                false,
+                "Select cond and src0 dims should be same, but got %zu and %zu",
+                select_cond_dims.size(), select_src0_dims.size());
+        for (size_t i = 0; i < select_cond_dims.size(); i++) {
+
+            VCHECK_SDP_DECOMP(select_cond_dims[i] == select_src0_dims[i], false,
+                    "Select cond and src0 dims should be same, but got %lld "
+                    "and %lld",
+                    select_cond_dims[i], select_src0_dims[i]);
+        }
     }
 
 #if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
--- a/src/graph/backend/dnnl/passes/lower.cpp
+++ b/src/graph/backend/dnnl/passes/lower.cpp
@@ -664,114 +664,33 @@ static status_t select_handler(
     auto cond = in_vals[0];
     auto src0 = in_vals[1];
     auto src1 = in_vals[2];
-    cond->set_data_type(dnnl::impl::data_type::u8);
+    // For the binary select operation, the conditional input tensor can
+    // only be of `s8` data type.
+    cond->set_data_type(dnnl::impl::data_type::s8);
 
-    //TODO: This reorder can be removed once eltwise_clip support int8 input
-    op_ptr type_cast = std::make_shared<op_t>(op_kind::dnnl_reorder);
-    type_cast->set_attr<bool>(op_attr::change_layout, false);
-
-    op_ptr clip = std::make_shared<op_t>(op_kind::dnnl_eltwise);
-    clip->set_attr<int64_t>(op_attr::alg_kind,
-            static_cast<int64_t>(dnnl::algorithm::eltwise_clip));
-    clip->set_attr<float>(op_attr::alpha, 0.f);
-    clip->set_attr<float>(op_attr::beta, 1.f);
-
-    // After reorder and clip. The cond value is 0 or 1.
-    // Then output = src0.*cond+src1.*(cond*-1 + 1)
-    op_ptr mul1 = std::make_shared<op_t>(op_kind::dnnl_binary);
-    mul1->set_attr<int64_t>(op_attr::alg_kind,
-            static_cast<int64_t>(dnnl::algorithm::binary_mul));
-    mul1->merge_attributes(op->get_attributes());
-
-    op_ptr mul2 = std::make_shared<op_t>(op_kind::dnnl_binary);
-    mul2->set_attr<int64_t>(op_attr::alg_kind,
-            static_cast<int64_t>(dnnl::algorithm::binary_mul));
-    mul2->merge_attributes(op->get_attributes());
-
-    op_ptr linear = std::make_shared<op_t>(op_kind::dnnl_eltwise);
-    linear->set_attr<int64_t>(op_attr::alg_kind,
-            static_cast<int64_t>(dnnl::algorithm::eltwise_linear));
-    const float alpha_value = -1.0f, beta_value = 1.0f;
-    linear->set_attr<float>(op_attr::alpha, alpha_value);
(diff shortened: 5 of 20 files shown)


tests/gtests/graph/unit/utils.hpp
```diff
@@ -171,16 +171,12 @@ static inline void verify_op_schema(const dnnl::impl::graph::op_kind_t op_kind_,
     const op_schema_t *op_schema_
             = op_schema_registry_t::get_op_schema(op_kind_);
     EXPECT_TRUE(nullptr != op_schema_);
-
```
TaoLv, line 174: please revert.

tests/gtests/graph/unit/utils.hpp
```diff
@@ -1165,6 +1161,169 @@ inline void construct_select_float_MHA(dnnl::impl::graph::graph_t *agraph,
     agraph->add_op(&reorder_output);
 }
 
+inline void construct_select_float_MHA_without_broadcast(
+        dnnl::impl::graph::graph_t *agraph,
+        impl::data_type_t dtype = impl::data_type::f32, int batch_size = 1,
+        int seq_len = 128, int num_head = 12, int head_dim = 768,
+        bool transpose = false, bool distil = true, bool gpt = false) {
```
TaoLv, line 1168: If this is needed, it can be incorporated into the function above, probably with a flag in the parameter:
```suggestion
        bool transpose = false, bool distil = true, bool gpt = false, bool bcast_select = true) {
```

src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -651,122 +651,195 @@ static status_t dynamic_dequant_handler(
...
+
+        insert_empty_scratchpad(new_op);
+        rewriter.to_insert(new_op);
+        rewriter.to_remove(op);
+    } else {
+        auto in_vals = op->get_input_values();
+        auto out_vals = op->get_output_values();
+        assertm(in_vals.size() == 3 && out_vals.size() == 1,
+                "select should have three inputs and a output");
+        auto cond = in_vals[0];
```
rongzha1, line 727: this decompose code is almost same as [decompose_select_to_multiple_binary_ops() ](https://github.com/oneapi-src/oneDNN/pull/2349/files#diff-a5f0988a1a4cd1050cc529cadde6563a84adbf70d0d9370cfe8ffec60c08af61R2302) could you refactor them by extract method to avoid code duplication

tests/gtests/graph/unit/backend/dnnl/test_large_partition.cpp
```diff
@@ -1304,6 +1744,65 @@ TEST(test_large_partition_execute, Int8Bf16DistilBertMha_CPU) {
     strm->wait();
 }
 
+TEST(test_large_partition_execute, Int8Bf16DistilBertMhaWithoutBroadcast_CPU) {
```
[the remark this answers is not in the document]
Jiexin-Zheng: These unit tests are added to test the mha patterns without broadcasted `cond` and `select_other_input`, do you think we should use benchdnn graph instead?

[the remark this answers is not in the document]
TaoLv: If the test purpose can be achieved by benchdnn, then yes, benchdnn is the first choice.  If the test purpose is something should not happen or should fail, then I guess gtests are still needed.

src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -651,122 +651,195 @@ static status_t dynamic_dequant_handler(
...
-
-    // add new ops and delete select op
-    rewriter.to_insert(type_cast);
-    rewriter.to_insert(clip);
-    rewriter.to_insert(mul1);
-    rewriter.to_insert(linear);
-    rewriter.to_insert(mul2);
-    rewriter.to_insert(add);
-    rewriter.to_remove(op);
+    // For now, as primitive doesn't support broadcast for cond input, we use
```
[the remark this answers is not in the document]
Jiexin-Zheng: I have modified the lowering logic as: always lower select to binary primitive and then decide which impl path to use in pass `decompose_select_to_multiple_binary_ops` and decompose it to multiple binary ops if necessary, so the function is much shorter now.

src/graph/backend/dnnl/passes/lower.cpp
Jiexin-Zheng: I have wrapped it as an individual function, thanks for the suggestion~

tests/gtests/graph/unit/backend/dnnl/test_large_partition.cpp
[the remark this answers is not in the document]
Jiexin-Zheng: I have move them into the benchdnn graph batch files.

Jiexin-Zheng:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

Jiexin-Zheng:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/transform.hpp
```diff
@@ -119,6 +119,15 @@ status_t fuse_to_dnnl_sum(std::shared_ptr<subgraph_t> &sg);
 // make the input shape meet the requirement of oneDNN binary primitive
 status_t binary_canonicalization(std::shared_ptr<subgraph_t> &sg);
 
+// For now, we support two impl paths for select op: one is to use binary
+// primitive with select alorithm, the other is to use multiple binary ops(we
+// call it "leagcy impl" here). However, during the lowering pass, we directly
```
dzarukin, line 124:
```suggestion
// call it "legacy impl" here). However, during the lowering pass, we directly
```

dzarukin (approved):

src/graph/backend/dnnl/kernels/sdp_decomp.cpp
```diff
@@ -391,4 +392,4 @@ template struct sdp_decomp_kernel_t<true, dnnl::memory::data_type::f32>;
 } // namespace dnnl_impl
 } // namespace graph
 } // namespace impl
-} // namespace dnnl
+} // namespace dnnl
```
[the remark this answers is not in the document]
TaoLv: It still needs a newline at the end of file.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -58,6 +58,26 @@ bool sdp_decomp_config_t::initial_check(const std::shared_ptr<subgraph_t> &sg,
                 "Only supports single scale value, but got %lld", scale_sz);
     }
 
+    // Check select cond and src0 shape
+    if (graph_inport[5] != -1 && graph_inport[6] != -1) {
+        auto select_cond_dims = ltw(inputs[graph_inport[5]]).vdims();
+        auto select_src0_dims = ltw(inputs[graph_inport[6]]).vdims();
```
TaoLv, line 64: Please make them const.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -58,6 +58,26 @@ bool sdp_decomp_config_t::initial_check(const std::shared_ptr<subgraph_t> &sg,
...
+                false,
+                "Select cond and src0 dims should be same, but got %zu and %zu",
+                select_cond_dims.size(), select_src0_dims.size());
+        for (size_t i = 0; i < select_cond_dims.size(); i++) {
+
+            VCHECK_SDP_DECOMP(select_cond_dims[i] == select_src0_dims[i], false,
+                    "Select cond and src0 dims should be same, but got %lld "
+                    "and "
+                    "%lld",
+                    select_cond_dims[i], select_src0_dims[i]);
```
TaoLv, line 76: The code format does not look correct.

src/graph/backend/dnnl/passes/transform.cpp
```diff
@@ -2266,7 +2266,8 @@ status_t binary_canonicalization(std::shared_ptr<subgraph_t> &sg) {
         int32_t src1_ndims = src1_lt.ndims;
         int32_t target_ndims = std::max(src0_ndims, src1_ndims);
         std::vector<int32_t> in_ndims {src0_ndims, src1_ndims};
-        for (size_t i = 0; i < cur_op->num_inputs(); ++i) {
+        std::vector<size_t> input_indices = {0, 1};
+        for (auto i : input_indices) {
```
TaoLv, line 2270: Is this correct?  Previously `num_inputs()` is 2 - 32 per the schema definition. Now the code only handles the first two?

src/graph/backend/dnnl/passes/transform.cpp
```diff
@@ -2297,6 +2298,149 @@ status_t binary_canonicalization(std::shared_ptr<subgraph_t> &sg) {
...
+                    && op->get_attr<bool>(op_attr::is_select)))
+            continue;
+
+        // For the binary select primitive, broadcast semantics are not
+        // supported for the third conditional input tensor. For this case, the
+        // shape of the conditional input tensor must match that of the source 0
+        // tensor.
+        // The binary select primitive is unsupported on GPU.
+        const bool require_broadcast
+                = check_broadcast_requirement_for_inputs_pair(op, 0, 2);
```
gyhintel, line 2316: (1,2) and (0,1)  need to be checked?

src/graph/backend/dnnl/passes/transform.cpp
```diff
@@ -2297,6 +2298,149 @@ status_t binary_canonicalization(std::shared_ptr<subgraph_t> &sg) {
     return infer_shape(sg);
 }
 
+status_t decompose_select_to_multiple_binary_ops(
```
TaoLv, line 2301: nit: I feel `ops` already implies `multiple`.

```suggestion
status_t decompose_select_to_binary_ops(
```

src/graph/backend/dnnl/passes/utils.cpp
```diff
@@ -645,6 +646,32 @@ bool inverse_mul_scales(std::shared_ptr<op_t> &scale_op) {
...
+                      .ndims();
+    const size_t input2_ndims
+            = logical_tensor_wrapper_t(in_vals[index2]->get_logical_tensor())
+                      .ndims();
+
+    if (input1_ndims != input2_ndims) { return true; }
+
+    for (size_t i = 0; i < input1_ndims; i++) {
+        if (input1_dims[i] != input2_dims[i]) { return true; }
+    }
```
TaoLv, line 671: Since you already checked the length of the two vectors, I think you can simplify it with:

```suggestion
    if (input1_dims != input2_dims) return true;
```

src/graph/interface/shape_infer.cpp
```diff
@@ -32,9 +32,6 @@ namespace dnnl {
 namespace impl {
 namespace graph {
 
-// utils function
-namespace {
-
```
TaoLv, line 37: why remove it?

TaoLv (commented):
Do you have any performance data to share?

src/graph/backend/dnnl/dnnl_shape_infer.cpp
```diff
@@ -484,17 +484,65 @@ status_t infer_dnnl_pool_bwd_output_shape(op_t *n,
...
-
-    auto ret = is_bias_add
-            ? infer_bias_add_output_shape(n, inputs, outputs)
-            : infer_elemwise_arithmetic_output_shape(n, inputs, outputs);
-
-    return ret;
+    const bool is_select = n->has_attr(op_attr::is_select)
+            && n->get_attr<bool>(op_attr::is_select);
+    if (is_select) {
+        return infer_binary_select_output_shape(n, inputs, outputs);
```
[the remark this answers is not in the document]
Jiexin-Zheng: No, they are different as the frontend and backend select op have different inputs sequence and broadcast rules, you check take a look at code in the else branch.

src/graph/backend/dnnl/passes/transform.cpp
Jiexin-Zheng: No, broadcast for src0 and src1 of select is still supported, and primitive only requires cond and src0 have same dims.

src/graph/interface/shape_infer.cpp
Jiexin-Zheng: 1. We have many utils functions in `shape_infer.cpp`, they are defined in `shape_infer.cpp` and declared in `shape_infer.hpp`, and this temporary namespace only has function `dims2str`(it's declared and defined in this namespace and used in `shape_infer.cpp` only)
2. `dims2str` can be used and has been used for this PR in other files, it should be in the same namespace of other utils functions, that is `dnnl::impl::graph`

src/graph/backend/dnnl/passes/transform.cpp
Jiexin-Zheng: This pass is applied before postop fusion pass, so input number is always 2 before. For this PR, although binary select has three inputs, since `cond` dims has been promised to be the same that of `src0` by pass `decompose_select_to_binary_ops`, we only need to unsqueeze `src0` and `src1`.

Jiexin-Zheng:
> Do you have any performance data to share?

Sure, I have attached it to the PR description.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
Jiexin-Zheng: I have formatted it.

src/graph/backend/dnnl/passes/transform.hpp
Jiexin-Zheng: Fixed, thanks for the review~

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
Jiexin-Zheng: Fixed.

src/graph/backend/dnnl/passes/transform.cpp
Jiexin-Zheng: Fixed.

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Fixed.

src/graph/backend/dnnl/kernels/sdp_decomp.cpp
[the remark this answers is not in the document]
Jiexin-Zheng: Fixed.

gyhintel (approved):

src/graph/backend/dnnl/dnnl_shape_infer.cpp
```diff
@@ -484,17 +484,65 @@ status_t infer_dnnl_pool_bwd_output_shape(op_t *n,
...
+        VCHECK_INVALID_SHAPE(
+                (input0_dims == input1_dims && input1_dims == input2_dims),
+                "%s, all input dims should match each other if there is no "
+                "broadcast. input0 dims: %s, input1 dims: %s, input2 dims: %s ",
+                op_t::kind2str(n->get_kind()).c_str(),
+                dims2str(input0_dims).c_str(), dims2str(input1_dims).c_str(),
+                dims2str(input2_dims).c_str());
+        inferred_out_shape = std::move(input0_dims);
+    } else { // can broadcast
+        status_t ret1 = broadcast(input0_dims, input1_dims, inferred_out_shape);
```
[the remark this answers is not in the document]
Jiexin-Zheng: Another difference is that the binary select op should not have the `one_way_broadcast` check, and it's a must to change the order, because the backend select op has the inputs sequence as `(src0, src1, cond)` while the frontend select has the inputs sequence as `(src0, src1, cond)`

src/graph/backend/dnnl/dnnl_op_def.hpp
```diff
@@ -713,6 +714,7 @@ DNNL_GRAPH_OP_SCHEMA(dnnl_binary, 1,
                         {"NXC", "NCX"})
                 // New added attributes
                 .set_attr(op_attr::is_bias_add, false, attribute_kind::b, false)
+                .set_attr(op_attr::is_select, false, attribute_kind::b, false)
```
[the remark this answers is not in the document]
Jiexin-Zheng: Hi, yixin, thanks for the suggestion, I have remove the `is_select` attr.

Jiexin-Zheng:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/transform.cpp
ElaineBao: if `cond` dims has been promised to be the same of `src0`, then it should fall into the condition of `if (in_ndims[i] == target_ndims) { continue; }`, so no unsqueeze inserted. If this is the case, no need to limit the input_indices?

src/graph/backend/dnnl/passes/transform.cpp
ElaineBao: It seems `check_broadcast_requirement_for_inputs_pair` is only used here, I'm not sure if it worth a separate function or not, but it's not a hard requirement, feel free to resolve

src/graph/interface/shape_infer.cpp
ElaineBao: where is this function used?

src/graph/backend/dnnl/passes/transform.cpp
Jiexin-Zheng: The `in_ndims` only has two elements, the access for the third element is not legal.

src/graph/backend/dnnl/passes/transform.cpp
Jiexin-Zheng: Thanks~

src/graph/backend/dnnl/passes/transform.cpp
ElaineBao: ok, then it seems the original code is designed for 2 elements

src/graph/interface/shape_infer.cpp
Jiexin-Zheng: It's used in the shape inference process of binary select op: https://github.com/oneapi-src/oneDNN/blob/325bca9713c0f132a5306023010ff574c07b57d2/src/graph/backend/dnnl/dnnl_shape_infer.cpp#L509

ElaineBao (approved):

Jiexin-Zheng:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/transform.cpp
TaoLv: > This pass is applied before postop fusion pass, so input number is always 2 before. For this PR, although binary select has three inputs, since `cond` dims has been promised to be the same that of `src0` by pass `decompose_select_to_binary_ops`, we only need to unsqueeze `src0` and `src1`.

This explanation looks suspicious as the code has quite a few assumption to work properly. You may need to at least add comment for that.
BTW: I feel `for (size_t i : {0, 1}) { .... }` should work without defining `input_indices`.

src/graph/backend/dnnl/passes/utils.cpp
```diff
@@ -646,6 +647,30 @@ bool inverse_mul_scales(std::shared_ptr<op_t> &scale_op) {
     return true;
 }
 
+bool check_broadcast_requirement_for_inputs_pair(
+        const std::shared_ptr<op_t> &op, size_t index1, size_t index2) {
```
TaoLv, line 651: I remember i had a suggestion for this function signature.
```suggestion
bool need_broadcast_for_inputs(
        const std::shared_ptr<op_t> &op, size_t in1, size_t in2) {
```

src/graph/backend/dnnl/passes/utils.cpp
```diff
@@ -646,6 +647,30 @@ bool inverse_mul_scales(std::shared_ptr<op_t> &scale_op) {
...
+    const size_t input1_ndims
+            = logical_tensor_wrapper_t(in_vals[index1]->get_logical_tensor())
+                      .ndims();
+    const size_t input2_ndims
+            = logical_tensor_wrapper_t(in_vals[index2]->get_logical_tensor())
+                      .ndims();
+
+    if (input1_ndims != input2_ndims) { return true; }
+
+    if (input1_dims != input2_dims) { return true; }
```
TaoLv, line 670: The `operation==` of vector also checks the length of the two vector? If so, you may not need the check `ndims` above.

TaoLv (commented):
please separate benchdnn inputs changes into a standalone commit.

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Fixed

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Fixed.

PR #2350, opened 2025-01-07
Regression ci

Ryo-not-rio:
Adding a regression test ci to compare a PR against the main branch


theComputeKid:
Changing this to draft so we don't annoy the rest of the world by tagging the reviewers. You should be able to continue working on it as usual pushing to the draft PRs and the github actions will still run.

.github/workflows/regression-test.yml
```diff
@@ -0,0 +1,157 @@
...
+      matrix:
+        config: [
+          { name: c7g, label: ah-ubuntu_22_04-c7g_4x-50, threading: OMP, toolset: gcc, build: Release, testset: CI }
+        ]
+
+    name: ${{ matrix.config.name }}, ${{ matrix.config.toolset }}, ${{ matrix.config.threading }}, ${{ matrix.config.build }}
+    runs-on: ${{ matrix.config.label }}
+    steps:
+
+      - name: Checkout oneDNN
```
jondea, line 52: I worry about replicating all this oneDNN building logic. Can we not have a common build oneDNN pipeline and use the artifacts from it to test or benchmark?

PR #2352, opened 2025-01-07
Add support for non-power of 2 head size with quantization

umar456:
# Description

This pull request adds the ability to support quantization with head size that is not a power of two. This was a limitation on the vs ugemm and was required by the phi3-mini-4k-instruct model.

The shape that was failing was  Q [1x32xSEQ_LENx96] KV [1x32xSEQ_LENx96] which now passes with this change.

Additionally this PR addresses a failure to build when the vs work group was configured with an 16x16 work group tile.

### DIFF
--- a/src/gpu/intel/ocl/micro_sdpa.cpp
+++ b/src/gpu/intel/ocl/micro_sdpa.cpp
@@ -325,8 +325,9 @@ status_t micro_sdpa_t::pd_t::init_microkernels(impl::engine_t *engine) {
         problem_vs.aOffset = ABOffset::Calc;
     }
     if (with_value_scales() || with_value_zp()) {
-        problem_vs.aqGroupM
-                = (vs_common_scales || vs_common_zp) ? 1 : value_group_size();
+        problem_vs.aqGroupM = (vs_common_scales || vs_common_zp)
+                ? 1
+                : utils::rnd_up_pow2(value_group_size());
         problem_vs.aqGroupK = 1;
     }
     opts_vs.scaleA = with_value_scales() && !vs_common_scales;
--- a/src/gpu/intel/ocl/micro_sdpa.hpp
+++ b/src/gpu/intel/ocl/micro_sdpa.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -325,8 +325,9 @@ status_t micro_sdpa_t::pd_t::init_microkernels(impl::engine_t *engine) {
         problem_vs.aOffset = ABOffset::Calc;
     }
     if (with_value_scales() || with_value_zp()) {
-        problem_vs.aqGroupM
-                = (vs_common_scales || vs_common_zp) ? 1 : value_group_size();
+        problem_vs.aqGroupM = (vs_common_scales || vs_common_zp)
+                ? 1
+                : utils::rnd_up_pow2(value_group_size());
```
petercad, line 330: This is only safe if there is only one group in the `m` dimension (if group size >= head size).

umar456: Ah good point. I will add a check for this.

umar456: Added the check here: https://github.com/oneapi-src/oneDNN/pull/2352/files#diff-bba4b0f9326dd28e41aa9188d531360420f3040755ba76beb260df3441365641R144

syurkevi (approved):

src/gpu/intel/ocl/micro_sdpa.hpp
```diff
@@ -140,6 +141,16 @@ struct micro_sdpa_t : public gpu_primitive_t {
                         value_group_size());
             }
 
+            if (desc()->vs_scales.has_default_values()
+                    || desc()->vs_zero_points.has_default_values()) {
+                int vgs = value_group_size();
+                VDISPATCH_SDPA(
+                        math::is_pow2<int>(vgs) || vgs == val_md()->dims[3],
+                        "the value group size(%d) for must be a power of 2 or "
+                        "must equal to the number of values(%d).",
```
petercad, line 150:
```suggestion
                        "the value group size (%d) must be a power of 2 or "
                        "equal to the number of values (%d).",
```

petercad (approved):

umar456:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_graph

umar456:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_graph

dzarukin (approved):

PR #2356, opened 2025-01-08
ci: Add initial regression test workflow

Ryo-not-rio:
Adding a regression test CI to test against the main branch


dzarukin:
Hi @Ryo-not-rio, could you please describe what you are trying to achieve and why did you choose specifically this way of resolution?

Ryo-not-rio:
@dzarukin I've updated the PR message with the information

theComputeKid:
Will this require further changes once #2388 goes in?

.github/workflows/ci-aarch64.yml
```diff
@@ -174,6 +179,72 @@ jobs:
           CTEST_PARALLEL_LEVEL: 6
           DYLD_LIBRARY_PATH: ${{ github.workspace }}/ComputeLibrary/build
           ONEDNN_THREADING: ${{ matrix.config.threading }}
+
+      ## Regression test steps ##
```
theComputeKid, line 183: We only want to do this for the release builds.

theComputeKid (changes requested):

dzarukin:
> @dzarukin I've updated the PR message with the information

Hi @Ryo-not-rio, thank you for adding more information. I feel there are better ways to do what you are trying to achieve. As you could know this repo supports github actions with custom syntax. It seems that supporting that flow for aarch64 using dedicated private infrastructure underneath for performance validation purpose would be more beneficial and will give you more flexible control over desired content and outcome.
I just don't see much value in the proposed solution for most users of the library as it tries to solve pretty narrow scope on a very limited set of platforms (Aarch64 Linux only with shell scripts?) and either can be done in a dedicated private repo, or through the way I've described above.

Ryo-not-rio:
@dzarukin I understand that there aarch64 may be a less frequent use-case, however we would like a check in the CI which would prevent the merging of any PR which accidentally causes a regression. This includes patches for non-aarch64 machines which inadvertently causes a regression and I think the best way to do that is to add to the existing CI. 

We already have unit testing specifically for aarch64 and the regression tests do not add a massive overhead so I am struggling to understand the  problem with extending them. The scripts we have added are also not architecture specific so it can be reused to create regression tests for other architectures as well. 

If you are worried about the execution time of the CI, these tests only adds about 5 minutes to the existing ci-aarch64 workflow: https://github.com/oneapi-src/oneDNN/actions/runs/12888105593/job/35932228418?pr=2356

dzarukin:
> @dzarukin I understand that aarch64 may be a less frequent use-case, however we would like a check in the CI which would prevent the merging of any PR which accidentally causes a regression. This includes patches for non-aarch64 machines which inadvertently causes a regression and I think the best way to do that is to add to the existing CI.
> 
> We already have unit testing specifically for aarch64 and the regression tests do not add a massive overhead so I am struggling to understand the problem with extending them. The scripts we have added are also not architecture specific so it can be reused to create regression tests for other architectures as well.
> 
> If you are worried about the execution time of the CI, these tests only adds about 5 minutes to the existing ci-aarch64 workflow: https://github.com/oneapi-src/oneDNN/actions/runs/12888105593/job/35932228418?pr=2356

I have certain doubts about the methods used to solve the problem that can't be generally solved.
What's going to happen from administrative point of view if:
* A sporadic PR triggers a problem due to machine instability which can be caused by many reason (extra job running, frequency changing, software updates, etc). Would that trigger block the PR promotion? If yes, how the developer supposed to address this gap?

Other related questions:
* The coverage seems to be unified and hard-coded (at least I failed to find a way where it can be customized). Who decides on what to cover and its importance?
* Once the coverage eats all of the machine time bandwidth, what's going to happen?

As I mentioned earlier, every team working on oneDNN solves performance validation on their own way because performance targets and priority workloads are different for each dedicated hardware, and the way to measure what's important is also inconsistent between architectures.

So I kindly suggest to re-think the current approach and build something locally to run on a regular basis than trying to implement something on a common ground.

.github/workflows/ci-aarch64.yml
```diff
@@ -174,6 +183,55 @@ jobs:
...
```
vpirogov, line 187: Please change the term to 'performance test' or 'performance regression test' to avoid confusion with functional regression tests.

tests/regression/bench_regression.sh
```diff
@@ -0,0 +1,30 @@
+#! /bin/bash
```
vpirogov, line 1: This script belongs in `.github/automation`, not in `tests`.

tests/regression/benchdnn_comparison.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/python3
```
vpirogov, line 1: This script belongs in `.github/automation`, not in `tests`.

tests/regression/inputs/conv
```diff
@@ -0,0 +1,22 @@
+# *******************************************************************************
```
vpirogov, line 1: Consider using existing input files from `tests/benchdnn/inputs`, say `[conv/shapes_resnet_50](https://github.com/oneapi-src/oneDNN/blob/main/tests/benchdnn/inputs/conv/shapes_resnet_50)` and `[matmul/shapes_bert_large](https://github.com/oneapi-src/oneDNN/blob/main/tests/benchdnn/inputs/matmul/shapes_bert_large)`. 

If you really need specific set for AArch64 add something like `inputs/conv/perf_aarch64`.

vpirogov:
> I understand that aarch64 may be a less frequent use-case, however we would like a check in the CI which would prevent the merging of any PR which accidentally causes a regression.

Based on our experience avoiding regressions requires extensive coverage and infrastructure that can run relevant tests on demand. Running few testing points on every PR is unlikely to catch regressions. This considerations is not a blocker for this PR though, we can try and see how this system works.

.github/workflows/ci-aarch64.yml
```diff
@@ -174,6 +183,55 @@ jobs:
...
+
+      - shell: bash
+        if: ${{ matrix.config.build == 'Release' }}
+        run: |
+          OMP_NUM_THREADS=4 bash ${{ github.workspace }}/oneDNN/tests/regression/bench_regression.sh ${{ github.workspace }}/oneDNN_main/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN/build/tests/benchdnn/benchdnn main.txt new.txt
+          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/tests/regression/bench_regression.sh ${{ github.workspace }}/oneDNN_main/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN/build/tests/benchdnn/benchdnn main.txt new.txt
+        env:
+          DYLD_LIBRARY_PATH: ${{ github.workspace }}/ComputeLibrary/build
+
+      - name: Compare regression test results
```
vpirogov, line 225: This seems to compare performance of code in the PR to the `HEAD` of main branch. Note, that PR branch is almost always behind `HEAD`, so this would result in false positives.

tests/regression/inputs/conv
Ryo-not-rio: For the precommit tests, we really want a very few specific tests so would it not make more sense to have these tests easily accessible and in the same place within `.github/automation`?

.github/workflows/ci-aarch64.yml
Ryo-not-rio: Thanks for noticing this, I've changed this to test against the base branch of the PR

Ryo-not-rio:
>Based on our experience avoiding regressions requires extensive coverage and infrastructure that can run relevant tests on demand. Running few testing points on every PR is unlikely to catch regressions

Yes, I agree. We intend on doing a more thorough performance test with our nightlies and for the precommit we just want to make sure the critical paths do not regress, similar to a smoke test

Ryo-not-rio:
waiting on https://github.com/oneapi-src/oneDNN/pull/2599 before we can merge it

.github/workflows/ci-aarch64.yml
```diff
@@ -20,18 +20,6 @@ name: "CI AArch64"
 #* To avoid duplicate jobs running when both push and PR is satisfied, we use this:
 #* https://github.com/orgs/community/discussions/26940#discussioncomment-5686753
 on:
```
theComputeKid, line 22: What's going here?

theComputeKid (changes requested):

.github/workflows/ci-aarch64.yml
```diff
@@ -108,6 +95,15 @@ jobs:
         with:
           version: ${{ fromJson(steps.get-versions.outputs.output).dependencies.clang }}
 
+      - name: setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Install scipy
```
theComputeKid, line 103: Do we really need to pull in scipy?

.github/workflows/ci-aarch64.yml
```diff
@@ -149,15 +145,64 @@ jobs:
...
+      #   working-directory: ${{ github.workspace }}/oneDNN/build
+      #   env:
+      #     BUILD_TOOLSET: ${{ matrix.config.toolset }}
+      #     CMAKE_BUILD_TYPE: ${{ matrix.config.build }}
+      #     CTEST_PARALLEL_LEVEL: 6
+      #     DYLD_LIBRARY_PATH: ${{ github.workspace }}/ComputeLibrary/build
+      #     ONEDNN_THREADING: ${{ matrix.config.threading }}
+
+      ## Performance test steps ##
+      - name: Checkout oneDNN base
```
theComputeKid, line 159: Is there a way to separate the perf logic from the usual CI logic, while keeping the same workspace?

theComputeKid (changes requested):

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -0,0 +1,81 @@
...
+
+    with open(file2) as f:
+        r2 = f.readlines()
+
+    # Trim non-formatted lines and split the prolem from time
+    r1 = [x.split(",") for x in r1 if x[0:8] == "--mode=P"]
+    r2 = [x.split(",") for x in r2 if x[0:8] == "--mode=P"]
+
+    if len(r1) != len(r2):
+        raise Exception("The number of benchdnn runs do not match")
```
Sqvid, line 41: Also check for the case that r1 == r2 == 0. Something probably went wrong in that case.

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Convert to dict and trim \n
```
Sqvid, line 43: This comment seems misleading. You are just intialising new defaultdicts right? Where is the stripping of the newline.

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -0,0 +1,81 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+
+import sys
+import os
+from collections import defaultdict
+from scipy.stats import ttest_ind
```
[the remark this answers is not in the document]
Ryo-not-rio: I think it only takes 8s to download scipy, I think it's not too big of an issue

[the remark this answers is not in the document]
Sqvid: Oki that is fine then. Thanks.

.github/automation/performance/bench_performance.sh
```diff
@@ -0,0 +1,30 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+for i in {1..5}
+do
+    $1 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul >> $3
+    $2 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul >> $4
+    $1 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv >> $3
+    $2 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv >> $4
+done
```
vpirogov, line 30: (nit) Here and in other new files: missing EoL in the last line

.github/workflows/ci-aarch64.yml
Ryo-not-rio: I think it's okay to have scipy in this case since it only takes 8s to download

tests/regression/inputs/conv
vpirogov: If you are referring to benchdnn input files then the rationale for keeping them in inputs directory is to make manually reproducing issues with benchdnn a bit more straightforward. Compare:
```
benchdnn --matmul --batch=inputs/matmul/perf_aarch64
benchdnn --matmul --batch=<REPO_ROOT>/.github/automation/perf_aarch64
```

Also helps with reuse and maintenance of these input files. If you strongly feel that these should sit in automation I'm fine with that as well.

vpirogov:
@Ryo-not-rio, do you know why AArch64 CI did not trigger on this PR?

vpirogov:
Ok, now AArch64 CI did trigger, but all performance testing part was skipped.

.github/workflows/ci-aarch64.yml
Ryo-not-rio: Creating a new action doesn't work well with building the main branch and accessing the pre-built PR-branch, creating a new workflow will force it to start up a new instance so depends on if we prioritize neatness of the yaml file or the speed of the CI

vpirogov:
Consider adding a commit with intentional regression to test that the check actually catches it.

tests/regression/inputs/conv
Ryo-not-rio: In the case where the precommit performance tests fail, the failed tests would be shown to the user so there wouldn't be a need to run the full batches. In the case the user wants to run all the performance tests at once, they can run `bench_performance.sh` so I would prefer to have these files at an easily accessible place since we're likely to update them

Ryo-not-rio:
@vpirogov Thanks for your comments, I have confirmed the pipeline works with a regression and this PR is ready for review.

cc: @theComputeKid

vpirogov (approved):

theComputeKid (approved):

PR #2358, opened 2025-01-08
xe: jit: gemm: robust flag register handling in k loop check

petercad:
Fixes potential issues in grouped quantization cases where the k loop check could happen too early, clashing with flag register usage by scale/zp loads in case virtual flag registers are in use. The existing mechanism for timing the k loop check is a bit fragile; this patch makes it more robust by disallowing usage of the k loop check flag once the check has been made.

Xe2 example failing case prior to this patch:
```
--matmul --engine=gpu --dt=f16:s4:f16 --wtag=acb --attr-scales=wei:per_ocic:f16:128x1 --attr-zero-points=wei:per_ocic:u4:128x1 --attr-fpmath=f16:true --skip-impl=ref 3x96x512:3x512x64
```


petercad:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

petercad:
Precommit/lightweight scans failures are unrelated to this PR.

rjoursler (approved):

dyoussif (approved):

Issue #1652, opened 2023-05-18
About default dilation in convolution


MOLOjl:
https://github.com/oneapi-src/oneDNN/blob/5dc0698b11f6f02ddd5566ab5d5b0a27622d5199/src/common/convolution.cpp#L122
Why the default dilation is 0 but 1?

vpirogov:
I tried to dig up history, but the reason we use 0 seems to be lost.

MOLOjl:
All right, thanks anyway.

Issue #438, opened 2019-03-27
failed to create dilated convolution 


nwpu2fjl:
I tried to create dilated convolution on my MacBook Pro (Retina, 15-inch, Mid 2015).
![image](https://user-images.githubusercontent.com/12624852/55055969-ae8cdc80-509f-11e9-9f27-48f3e22091e5.png)
But I got an error!
![image](https://user-images.githubusercontent.com/12624852/55055983-bcdaf880-509f-11e9-9af5-009e729194d4.png)
What should I use dilated convolution?

nwpu2fjl:
Sorry, It's my fault! The error is caused by invalid arguments.
![image](https://user-images.githubusercontent.com/12624852/55057735-fa8e5000-50a4-11e9-905e-1c1964c348ba.png)
It must be that our understanding of the dilated concept is different. In my opinion, the dilated value of no-dilated convolution should be 1, but factly 0 in mkldnn.
If I set the destination dimension(- 1 in my previous concept) by the below formulate, it is right.
dst_dim = (input_dim + 2 * pad - (dilated * (weight_dim - 1) + 1)) / stride + 1.
Am I right?

igorsafo:
Hi @nwpu2fjl  MKL-DNN uses the following formula: dst = (src - ((dilate + 1)*(weights - 1) + 1))/ stride + 1 (https://github.com/intel/mkl-dnn/blob/master/src/common/convolution.cpp#L117), so you can just pass modified dilation to MKL-DNN to match your formula: mkldnn_dilate = dilate - 1

nwpu2fjl:
Thanks for your reply. Right, I have modified my dilation(mkldnn_dilate = my dilation - 1), and got correct result!

PR #2359, opened 2025-01-09
doc: conv: add note for the dilation computation formula

### LINKED ISSUE #2355: Correctness Issue: The Formula Used in OneDNN and PyTorch for MaxPool2d is Inconsistent
The example provided in the OneDNN examples (https://github.com/oneapi-src/oneDNN/blob/v3.6.2/examples/primitives/pooling.cpp) uses a formula that is inconsistent with the formula in PyTorch (https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html).

The formula in OneDNN:
```
const memory::dim OH = (IH - ((KH - 1) * DH + KH) + PH_L + PH_R) / SH + 1;  
const memory::dim OW = (IW - ((KW - 1) * DW + KW) + PW_L + PW_R) / SW + 1;
```
The formula in PyTorch:
```
const memory::dim OH = std::floor((IH - DH * (KH - 1) - 1 + PH_L + PH_R) / SH + 1);  
const memory::dim OW = std::floor((IW - DW * (KW - 1) - 1 + PH_L + PH_R) / SW + 1);
```
微信截图_20250108150729
In my test case, I used the following parameters:
```
const memory::dim N = 3; // batch size  
const memory::dim IC = 3; // input channels  
const memory::dim IH = 112; // input tensor height  
const memory::dim IW = 112; // input tensor width  
const memory::dim KH = 3; // kernel height  
const memory::dim KW = 3; // kernel width  
const memory::dim PH_L = 1; // height padding: left  
const memory::dim PH_R = 1; // height padding: right  
const memory::dim PW_L = 1; // width padding: left  
const memory::dim PW_R = 1; // width padding: right  
const memory::dim SH = 2; // height-wise stride  
const memory::dim SW = 2; // width-wise stride  
const memory::dim DH = 1; // height-wise dilation  
const memory::dim DW = 1; // width-wise dilation
```

Using the OneDNN formula, I get OW = 55 and OH = 55.
Using the PyTorch formula, I get OW = 56 and OH = 56.

The resulting shapes are inconsistent!!!!

If I use the PyTorch formula to calculate OW and OH, and then generate the descriptor based on the OW and OH calculated from PyTorch, the program (https://github.com/oneapi-src/oneDNN/blob/v3.6.2/examples/primitives/pooling.cpp) will result in a compilation error!!!

I am even confused.

shu1chen:
# Description

The formula for dilated convolution and dilated pooling in oneDNN differs from that used in PyTorch and some other applications. This difference has led to confusion and several related questions in oneDNN, including #2355, #1652, and #438. This PR clarifies the differences in the documentation and updates the code example comments accordingly.

Fixes #2355

### DIFF
--- a/doc/primitives/convolution.md
+++ b/doc/primitives/convolution.md
@@ -100,6 +100,12 @@ Here:
 - \f$OW = \left\lfloor{\frac{IW - DKW + PW_L + PW_R}{SW}}
         \right\rfloor + 1,\f$ where \f$DKW = 1 + (KW - 1) \cdot (DW + 1)\f$.
 
+@note In oneDNN, convolution without dilation is defined by setting the dilation
+parameters to `0`. This differs from PyTorch and TensorFlow, where a non-dilated
+case corresponds to a dilation value of `1`. As a result, the PyTorch and
+TensorFlow dilation parameters need to be adjusted by subtracting `1` (for example,
+\f$DH_onednn = DH_torch - 1\f$, and \f$DW_onednn = DW_torch - 1\f$).
+
 #### Deconvolution (Transposed Convolution)
 
 Deconvolutions (also called fractionally strided convolutions or transposed
--- a/examples/primitives/pooling.cpp
+++ b/examples/primitives/pooling.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2022 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


doc/primitives/convolution.md
```diff
@@ -100,6 +100,12 @@ Here:
 - \f$OW = \left\lfloor{\frac{IW - DKW + PW_L + PW_R}{SW}}
         \right\rfloor + 1,\f$ where \f$DKW = 1 + (KW - 1) \cdot (DW + 1)\f$.
 
+@note Some applications may use an alternative formula for dimensions computation:
```
vpirogov, line 103: I would suggest explaining the behavior with plain language first. Something along these lines:

> In oneDNN convolution without dilation is defined by setting the dilation parameters to `0`. This is different from PyTorch and TensorFlow where non-dilated case corresponds to dilation value of `1`.

You can still keep the formulas if you feel that these add value.

+@mgouicem
+@ranukund

shu1chen: Thank you for the suggestion. I have updated it accordingly. I think these formulas make it clear how to apply the alternative formulas with oneDNN, so I keep them.

doc/primitives/convolution.md
```diff
@@ -100,6 +100,16 @@ Here:
...
+@note In oneDNN, convolution without dilation is defined by setting the dilation 
+parameters to `0`. This differs from PyTorch and TensorFlow, where a non-dilated 
+case corresponds to a dilation value of `1`.
+
+@note For convolution with dilation, PyTorch and TensorFlow use alternative 
+formulas to calculate dimensions: \f$DKH_alt = 1 + (KH - 1) \cdot DH_alt\f$, and
+\f$DKW_alt = 1 + (KW - 1) \cdot DW_alt\f$. To apply the formulas with oneDNN, 
+the dilation parameters need to be adjusted as follows: 
+\f$DH_onednn = DH_alt - 1\f$, and \f$DW_onednn = DW_alt - 1\f$.
+
```
mgouicem, line 112: I think this second note is redundant with the one above and both can be simplified in the same note. Something like:

```
@note In oneDNN, convolution without dilation is defined by setting the dilation 
parameters to `0`. This differs from PyTorch and TensorFlow, where a non-dilated 
case corresponds to a dilation value of `1`. As a result, the PyTorch and Tensorflow
dilation parameters need to be adjusted by subtracting `1` (for example 
\f$DW_onednn = DW_torch - 1\f$, \f$DH_onednn = DH_torch - 1\f$, ...).

shu1chen: This looks much better. Thanks!

vpirogov (approved):

mgouicem (approved):

PR #2361, opened 2025-01-09
doc: clarify conversions can be impacted by double-rounding

mgouicem:
Specify in the documentation that there could be double rounding going on when converting f32 to very low accuracy datatypes (fp4/fp8).


doc/programming_model/data_types.md
```diff
@@ -99,7 +99,12 @@ No downconversions are allowed by default, but can be enabled using
 the floating-point math controls described in @ref
 dev_guide_attributes_fpmath_mode.
 
-
+The \f$convert_{dst\_dt}\f$ conversion is not guaranteed to be correctly
```
ranukund, line 102: Will this be the result of the conversion is not guaranteed to be correctly rounded to the nearest value?

doc/programming_model/data_types.md
```diff
@@ -99,7 +99,12 @@ No downconversions are allowed by default, but can be enabled using
...
+rounded. In particular, some hardware platforms have no direct
+conversion instructions from f32 to low-precision datatypes (e.g. fp8
```
ranukund, line 104:
```suggestion
conversion instructions to convert f32 data types to low-precision data types such as fp8
```

doc/programming_model/data_types.md
```diff
@@ -99,7 +99,12 @@ No downconversions are allowed by default, but can be enabled using
...
+or fp4), and will perform conversion through an intermediate datatype
```
ranukund, line 105:
```suggestion
or fp4, and will perform conversion through an intermediate data type
```

doc/programming_model/data_types.md
```diff
@@ -99,7 +99,12 @@ No downconversions are allowed by default, but can be enabled using
...
+(e.g. f16/bf16), which may cause [double
```
ranukund, line 106:
```suggestion
(for example, f16 or bf16), which may result in [double
```

ranukund (commented):
A few comments suggested, please incorporate as you see fit, thanks!

doc/programming_model/data_types.md
```diff
@@ -99,7 +99,12 @@ No downconversions are allowed by default, but can be enabled using
...
+rounding](https://en.wikipedia.org/wiki/Rounding#Double_rounding).
 
 ### Rounding mode and denormal handling
```
vpirogov, line 109: Does it make sense to place this note to [rounding mode](https://github.com/oneapi-src/oneDNN/blob/main/doc/programming_model/data_types.md#rounding-mode-and-denormal-handling) section instead?

vpirogov (approved):

doc/programming_model/data_types.md
mgouicem: That is an option, but given that it is not about rounding mode but rounding accuracy, I thought it might be better to put with the general formula.

vpirogov: Got it, thanks.

doc/programming_model/data_types.md
mgouicem: fixed, thanks.

doc/programming_model/data_types.md
mgouicem: fixed thanks

doc/programming_model/data_types.md
mgouicem: fixed, thanks

doc/programming_model/data_types.md
mgouicem: I clarified that it would be faithfully rounded, and what that means (one of the two closest representable values is returned, but not necessarily the closest).

ranukund (approved):
Thank you!

PR #2363, opened 2025-01-09
xe: jit: gemm: remove mixing of oneDNN and gemmstone code

rjoursler:
As a side-effect, also enables targeting an out-of-tree gemmstone by setting ONEDNN_GEMMSTONE_DIR.


rjoursler:
make test
disable device_cpu
disable device_gpu

petercad:
@rjoursler Is the idea to then move gemmstone to the third-party directory? The directories are getting buried pretty deep :)

petercad:
> Do you have a preferred location? I had considered moving this `gpu/intel/jit/gemmstone` or even `gpu/intel/gemmstone`, but wasn't sure on developers opinions.

`gpu/intel/jit/gemmstone` is fine by me, but of course it doesn't matter much if we're going to move gemmstone to third_party "really soon." FYI, OV has to modify a bit of code on their side every time we move gemmstone around, so it's probably best to delay this PR until rls-v3.7 is branched.

vpirogov:
> That is the basic idea, or more accurately, to get that process started. According to [this comment](https://github.com/oneapi-src/oneDNN/pull/2338#discussion_r1907851297), there is likely a bit too much entanglement between the projects to move into the proposed `thirdparty` location.

My comment was about gemmstone being dependent on oneDNN internals. With these changes it looks cleanly separated so moving it to thirdparty makes perfect sense.

PR #2364, opened 2025-01-09
Arch: coverity fixes (fixed MFDNN-12954)

dzarukin:
MFDNN-12954

Engine is peculiar, Coverity understands it's global object and that it has some singletons and that their initialization order od destruction is uncertain and may lead to issues.

Rest cases are less interesting.

### DIFF
--- a/examples/tutorials/matmul/cpu_matmul_quantization.cpp
+++ b/examples/tutorials/matmul/cpu_matmul_quantization.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2022 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -203,7 +203,10 @@ int compare_vectors(const std::vector<float> &v1,
 
 } // namespace
 
-engine eng(engine::kind::cpu, 0); // We create a global engine for simplicity
+const engine &eng() {
+    static const engine eng(engine::kind::cpu, 0);
+    return eng;
+}
 
 // Quantize float data into X_int_m oneDNN memory using the q10n parameters
 //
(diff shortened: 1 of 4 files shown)


dzarukin:
make test

tests/gtests/dnnl_test_common.hpp
```diff
@@ -437,7 +437,13 @@ static void fill_data(const memory::dim nelems, const memory &mem, data_t mean,
 
 inline void fill_data(memory::data_type dt, const memory &mem, float mean,
         float deviation, double sparsity = 1.) {
-    size_t nelems = mem.get_desc().get_size() / memory::data_type_size(dt);
+    const auto dt_size = memory::data_type_size(dt);
+    if (dt_size <= 0) {
+        assert(!"unexpected data type");
```
densamoilov, line 442: Is it considered an error or dt can be undef?

examples/tutorials/matmul/cpu_matmul_quantization.cpp
```diff
@@ -203,7 +203,10 @@ int compare_vectors(const std::vector<float> &v1,
 
 } // namespace
 
-engine eng(engine::kind::cpu, 0); // We create a global engine for simplicity
```
densamoilov, line 206: This engine is used only within a single translation unit isn't it?

tests/gtests/dnnl_test_common.hpp
dzarukin: `undef` dt is considered an error, as it's not clear what is the correct behavior in that case.

examples/tutorials/matmul/cpu_matmul_quantization.cpp
dzarukin: I believe so.

densamoilov: Can you elaborate on the problem with it being global?

tests/gtests/test_sum.cpp
```diff
@@ -249,10 +249,13 @@ class sum_test_t : public ::testing::TestWithParam<sum_test_params> {
                 int mant_digits
                         = dnnl::impl::nstl::numeric_limits<src_data_t>::digits;
                 int want_mant_digits = 3;
+                int digits_shift = mant_digits - want_mant_digits;
+                uint_type max_val = (uint_type)-1;
+                // Move left to keep mask value in uint_type range, move left
+                // to flush all digits but `want_mant_digits`.
```
mgouicem, line 255: Is the comment even needed? ;)
```suggestion
                // Move right to clear trailing mantissa bits, move left
                // to keep `want_mant_digits` leading mantissa bits.
                // Procceeding this way avoids overflow on mask 
```

mgouicem (approved):

avmanerikar (approved):

examples/tutorials/matmul/cpu_matmul_quantization.cpp
dzarukin: Discussed offline, the order of destruction for internal singletons and related to them objects might expose issues.

tests/gtests/test_sum.cpp
dzarukin: It was shifting left only before and it didn't come to me where the problem might be. So, this is kind of documenting shift to the right and why it's necessary here.

PR #2370, opened 2025-01-10
x64: conv: handle dilation int overflow (fixes MFDNN-12967)

tczeszun:
Disabling dilation values (`> INT_MAX`) to avoid `int` overflow.
In future we may change `jcp` fields from `int` to `size_t`, but there are a lot of dependencies in a many places of the code, so it's better to disable it for now.
[MFDNN-12967](https://jira.devtools.intel.com/browse/MFDNN-12967)

### DIFF
--- a/src/cpu/x64/jit_avx512_common_conv_kernel.cpp
+++ b/src/cpu/x64/jit_avx512_common_conv_kernel.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2016-2024 Intel Corporation
+* Copyright 2016-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@
 #include "common/type_helpers.hpp"
 #include "common/utils.hpp"
 
+#include "cpu/cpu_convolution_pd.hpp"
 #include "cpu/platform.hpp"
 #include "cpu/x64/cpu_barrier.hpp"
 #include "cpu/x64/injectors/injector_utils.hpp"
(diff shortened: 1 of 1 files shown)


inteldimitrius (approved):
LGTM

src/cpu/x64/jit_avx512_common_conv_kernel.cpp
```diff
@@ -822,6 +822,13 @@ status_t jit_avx512_common_conv_fwd_kernel::init_conf(jit_conv_conf_t &jcp,
     jcp.stride_h = (ndims == 3) ? 1 : cd.strides[ndims - 4];
     jcp.stride_w = cd.strides[ndims - 3];
 
+    // Big int (> INT_MAX) values are unsupported and jcp fields may overflow
+    // TODO: change data type of jcp fields to size_t
+    if ((ndims == 5 && cd.dilates[0] > INT_MAX)
+            || (ndims >= 4 && cd.dilates[ndims - 4] > INT_MAX)
+            || (cd.dilates[ndims - 3] > INT_MAX))
+        return status::unimplemented;
```
dzarukin, line 830: Please use verbose messaging for unimplemented status.

src/cpu/x64/jit_avx512_common_conv_kernel.cpp
```diff
@@ -822,6 +822,13 @@ status_t jit_avx512_common_conv_fwd_kernel::init_conf(jit_conv_conf_t &jcp,
     jcp.stride_h = (ndims == 3) ? 1 : cd.strides[ndims - 4];
     jcp.stride_w = cd.strides[ndims - 3];
 
+    // Big int (> INT_MAX) values are unsupported and jcp fields may overflow
+    // TODO: change data type of jcp fields to size_t
+    if ((ndims == 5 && cd.dilates[0] > INT_MAX)
```
dzarukin, line 827: For consistent style.
```suggestion
    if ((ndims == 5 && cd.dilates[ndims - 5] > INT_MAX)
```

ankalinin (approved):

src/cpu/x64/jit_avx512_common_conv_kernel.cpp
tczeszun: done, thanks

src/cpu/x64/jit_avx512_common_conv_kernel.cpp
tczeszun: done, thanks

dzarukin (approved):

PR #2374, opened 2025-01-11
Revert "Enable bfloat16 for micro sdpa kernel"

h-sadia:
Reverts oneapi-src/oneDNN#2344


vpirogov:
Let's leave it as it is for now. Reverting and recommitting is not worth it.

PR #2378, opened 2025-01-11
Fixed issue with installed multiple versions of the OpenCL which lead…

redradist:
# Description
This PR fix issue with multiple version OpenCL installed on machne as well as brings new versions of OpenCL 3_0

Related PR https://github.com/oneapi-src/oneDNN/pull/2001, issue was more deeply investigated and PR was reworked

For details explanation of an issue with `${OPENCL_INCLUDE_DIR}/CL/cl.h` https://stackoverflow.com/a/79347124/5157245

### DIFF
--- a/cmake/FindOpenCL.cmake
+++ b/cmake/FindOpenCL.cmake
@@ -47,18 +47,18 @@ function(_FIND_OPENCL_VERSION)
   set(CMAKE_REQUIRED_QUIET ${OpenCL_FIND_QUIETLY})
 
   CMAKE_PUSH_CHECK_STATE()
-  foreach(VERSION "2_2" "2_1" "2_0" "1_2" "1_1" "1_0")
+  foreach(VERSION "3_0" "2_2" "2_1" "2_0" "1_2" "1_1" "1_0")
     set(CMAKE_REQUIRED_INCLUDES "${OpenCL_INCLUDE_DIR}")
 
     if(APPLE)
       CHECK_SYMBOL_EXISTS(
         CL_VERSION_${VERSION}
-        "${OpenCL_INCLUDE_DIR}/Headers/cl.h"
+        "Headers/cl.h"
         OPENCL_VERSION_${VERSION})
     else()
       CHECK_SYMBOL_EXISTS(
         CL_VERSION_${VERSION}
-        "${OpenCL_INCLUDE_DIR}/CL/cl.h"
+        "CL/cl.h"
         OPENCL_VERSION_${VERSION})
     endif()
 


cmake/FindOpenCL.cmake
```diff
@@ -47,18 +47,18 @@ function(_FIND_OPENCL_VERSION)
...
   CMAKE_PUSH_CHECK_STATE()
-  foreach(VERSION "2_2" "2_1" "2_0" "1_2" "1_1" "1_0")
+  foreach(VERSION "3_0" "2_2" "2_1" "2_0" "1_2" "1_1" "1_0")
     set(CMAKE_REQUIRED_INCLUDES "${OpenCL_INCLUDE_DIR}")
 
     if(APPLE)
       CHECK_SYMBOL_EXISTS(
         CL_VERSION_${VERSION}
-        "${OpenCL_INCLUDE_DIR}/Headers/cl.h"
+        "Headers/cl.h"
```
densamoilov, line 56: I read the stackoverflow post you shared and it seems that this change merely tricks CMake to make it pick up the header files you want it to pick up because removing the full path to the header files adds ambiguity that resolves in your favor.

Have you tried to unset `OpenCL_INCLUDE_DIR` prior to calling to `find_package(OpenCL)`?

redradist: As you can see, `OpenCL_INCLUDE_DIR` is already used in this `.cmake file`, line 51 `set(CMAKE_REQUIRED_INCLUDES "${OpenCL_INCLUDE_DIR}")`, but it does not help in case of multiple versions installed.
But in my case I need multiple versions, because NVIDIA do not support latest OpenCL and it ships with old OpenCL header file, that is why in some rare (but very important cases, like to use NVIDIA with OpenCL and oneDNN) it could improve experience ...
I've tried to do multiple thing, including adding defines before and after, but non of it helped, only this little trick helped

redradist: @densamoilov Is there any other issues with this PR ??
It should work anyway because `set(CMAKE_REQUIRED_INCLUDES "${OpenCL_INCLUDE_DIR}")` specified before `CHECK_SYMBOL_EXISTS`

redradist:
Guys is there any other comments to this PR ??

vpirogov:
> Guys is there any other comments to this PR ??

I don't understand what this change does based on [stackoverflow discussion](https://stackoverflow.com/a/79347124/5157245). Specifically this part:

> CHECK_SYMBOL_EXISTS try to compile test program with the following include statment: #include <${OPENCL_INCLUDE_DIR}/CL/cl.h>, - where it was in my case: #include </usr/include/CL/cl.h>, after that compiler struggle between 2 type of header files for OpenCL in /usr/include/CL/ and /usr/local/cuda/include/CL/

Also note, that the code you are trying to change comes from [CMake upstream](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindOpenCL.cmake) verbatim.

If you want to pursue this further consider one of the following options:
* Open a [merge request in CMake upstream](https://gitlab.kitware.com/cmake/cmake/-/merge_requests) to get input from the CMake developers.
* Check if updating FindOpenCL.cmake from upstream solves your problem. If it does we'll happily accept the PR with update.
* Provide a oneDNN build command and list of packages that demonstrate the issue so that we can reproduce the issue on our end and investigate.

redradist:
> > Guys is there any other comments to this PR ??
> 
> I don't understand what this change does based on [stackoverflow discussion](https://stackoverflow.com/a/79347124/5157245). Specifically this part:
> 
> > CHECK_SYMBOL_EXISTS try to compile test program with the following include statment: #include <${OPENCL_INCLUDE_DIR}/CL/cl.h>, - where it was in my case: #include </usr/include/CL/cl.h>, after that compiler struggle between 2 type of header files for OpenCL in /usr/include/CL/ and /usr/local/cuda/include/CL/
> 
> Also note, that the code you are trying to change comes from [CMake upstream](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindOpenCL.cmake) verbatim.
> 
> If you want to pursue this further consider one of the following options:
> 
>     * Open a [merge request in CMake upstream](https://gitlab.kitware.com/cmake/cmake/-/merge_requests) to get input from the CMake developers.
> 
>     * Check if updating FindOpenCL.cmake from upstream solves your problem. If it does we'll happily accept the PR with update.
> 
>     * Provide a oneDNN build command and list of packages that demonstrate the issue so that we can reproduce the issue on our end and investigate.

@vpirogov Done, PR is merged:

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/10366

For more details see comments in PR of CMake

Here it is merged in master branch Here it is merged in master branch https://gitlab.kitware.com/cmake/cmake/-/merge_requests/10366

vpirogov (approved):

vpirogov:
> Could someone with permissions to restart this job ?

Disregard this one.

vpirogov:
make test

vpirogov:
make test

densamoilov (approved):

PR #2388, opened 2025-01-13
Build acl cache sequentially

Ryo-not-rio:
When using github actions caches with the matrix strategy, the `cache-hit` variable is shared amongst all the strategies, potentially causing a race condition when run in parallel. This work creates just the caches sequentially, avoiding this problem

### DIFF
--- a/.github/automation/build_acl.sh
+++ b/.github/automation/build_acl.sh
@@ -1,7 +1,7 @@
 #! /bin/bash
 
 # *******************************************************************************
-# Copyright 2020-2024 Arm Limited and affiliates.
+# Copyright 2020-2025 Arm Limited and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -26,7 +26,7 @@ SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 # Defines MP, CC, CXX and OS.
 source ${SCRIPT_DIR}/common_aarch64.sh
 
-CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-"Release"}
+ACL_BUILD_TYPE=${ACL_BUILD_TYPE:-"Release"}
 ACL_ROOT_DIR=${ACL_ROOT_DIR:-"${PWD}/ComputeLibrary"}
 ACL_REPO="https://github.com/ARM-software/ComputeLibrary.git"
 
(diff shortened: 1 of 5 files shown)


.github/automation/build_acl.sh
```diff
@@ -36,24 +36,45 @@ elif [[ "$ACL_THREADING" == "SEQ" ]]; then
...
 elif [[ "$ACL_ACTION" == "build" ]]; then
     set -x
-    cmake --build $ACL_ROOT_DIR/build 
+    cd $ACL_ROOT_DIR
+    set -x
+    scons $MP Werror=0 debug=$ACL_DEBUG neon=1 opencl=0 embed_kernels=0 \
+        os=$ACL_OS arch=armv8.2-a build=native multi_isa=$ACL_MULTI_ISA_SUPPORT \
+        fixed_format_kernels=1 cppthreads=0 openmp=$ACL_OPENMP examples=0 \
+        validation_tests=0
+    set +x
```
theComputeKid, line 77: extra `set +x`

.github/automation/build_acl.sh
```diff
@@ -36,24 +36,45 @@ elif [[ "$ACL_THREADING" == "SEQ" ]]; then
...
-	-DARM_COMPUTE_BUILD_EXAMPLES=1 \
-	-DARM_COMPUTE_BUILD_TESTING=1 \
-    -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE
-    set +x
 elif [[ "$ACL_ACTION" == "build" ]]; then
     set -x
-    cmake --build $ACL_ROOT_DIR/build 
+    cd $ACL_ROOT_DIR
+    set -x
+    scons $MP Werror=0 debug=$ACL_DEBUG neon=1 opencl=0 embed_kernels=0 \
```
theComputeKid, line 73: Did CMake not work at all? It is surprising that it was building fine before.

.github/ci-aarch64.json
theComputeKid, line 1: Is it better to have this file in `automation`?

.github/workflows/aarch64-acl.yml
```diff
@@ -0,0 +1,131 @@
...
+permissions: read-all
+
+jobs:
+  # Cache is built sequentially to avoid cache-hit race conditions
+  build-cache:
+    strategy:
+      matrix:
+        config: [
+          { name: MacOS, label: macos-14, threading: SEQ, toolset: clang, build: Release, testset: SMOKE },
+          { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: clang, build: Debug, testset: SMOKE },
```
theComputeKid, line 51: We shouldn't need the testset, the testset is specifically for oneDNN, not ACL.

.github/workflows/aarch64-acl.yml
```diff
@@ -0,0 +1,131 @@
...
+    steps:
+
+      - name: Checkout oneDNN
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: oneDNN
+
+      - name: Install Scons MacOS
+        if: ${{ matrix.config.name == 'MacOS' }}
+        uses: threeal/pipx-install-action@b0bf0add7d5aefda03a3d4e47d651df807889e10 # v1.0.0
```
theComputeKid, line 66: It might be better to install scons via brew to avoid dependency on a 3p action.
`brew install scons`

.github/workflows/aarch64-acl.yml
theComputeKid, line 51: We can use c6g (small) to build both linux configs, as we don't need a particularly powerful machine to build ACL (plus we don't run any tests).

.github/workflows/aarch64-acl.yml
```diff
@@ -0,0 +1,131 @@
...
+        with:
+          packages: scons
+
+      - name: Install scons linux
```
theComputeKid, line 70: If we keep the casing consistent:
`Install scons (MacOS)`
`Install scons (Linux)`

theComputeKid (changes requested):

.github/automation/build_acl.sh
Ryo-not-rio: No, wasn't working for gcc or macOS, I imagine you hadn't changed ACL versions after switching to cmake so we were always reusing the cached build

theComputeKid:
Looks good. Just need to squash commits.

.github/workflows/ci-aarch64.yml
```diff
@@ -55,7 +55,7 @@ concurrency:
 permissions: read-all
 
 jobs:
-  build-and-test:
+  run-tests:
```
theComputeKid, line 58: we are still building oneDNN

.github/workflows/aarch64-acl.yml
```diff
@@ -0,0 +1,129 @@
...
+on:
+  push:
+    branches: [ main, 'rls-*' ]
+    paths:
+      - '.github/automation/ci-aarch64.json'
+      - '.github/workflows/aarch64-acl.yml'
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+       - '.github/automation/ci-aarch64.json'
```
theComputeKid, line 31: Will this work correctly when we update ACL in a PR and the ci-aarch64.yml runs at the same time as the aarch64-acl.yml? Should ci-aarch64.yml not depend on aarch64-acl.yml finishing first?

Ryo-not-rio: I've changed this workflow to a reusable workflow so this problem should be solved

theComputeKid (approved):
It is unfortunate that we have to build ACL sequentially so that we can cache it properly. This is a limitation of GitHub's API. But thanks for spotting it and fixing it.

theComputeKid (approved):

Radu2k (approved):
LGTM. 👍

PR #2394, opened 2025-01-13
common: align verbose level interface

rjoursler:
Align dnnl_set_verbose(int level) and DNNL_VERBOSE=level handling of verbose_t::warn.

### DIFF
--- a/src/common/verbose.hpp
+++ b/src/common/verbose.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2018-2024 Intel Corporation
+* Copyright 2018-2025 Intel Corporation
 * Copyright 2023 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/common/verbose.hpp
```diff
@@ -174,7 +174,11 @@ struct verbose_t {
         warn = 1 << 9,
         // the upper 8 bits are reserved for devinfo levels
         debuginfo = 1 << 24,
+
         //
+        level1 = error | exec_profile | warn,
+        level2 = error | exec_profile | warn | create_profile,
+
```
avmanerikar, line 181: (Random spot): It will be useful to cover the new levels for logging as well:
https://github.com/oneapi-src/oneDNN/blob/71a6c0603200fda39fe749901f348494e200a8a5/src/common/verbose.hpp#L255

rjoursler: I added the mappings and assigned them to `log_manager_t::info`, as that seems to be the highest level of information they correspond to.

avmanerikar: Looks good. Thanks for the update.

avmanerikar (approved):

rjoursler:
make test

dzarukin (approved):

spalicki (approved):

PR #2403, opened 2025-01-14
src: cpu: aarch64: move jit_sve conv above acl conv in convolution_list

michalowski-arm:
In benchdnn testing, jit_sve conv performs about x1.3 better than acl conv on average and jit_sve_1x1 conv performs x1.3-x3 better on average depending on the number of threads. In testing with CNNs, we got a 10-15% speed-up by using jit_sve and jit_sve_1x1 conv. So, they should be the preferred implementations whenever possible.

### DIFF
--- a/src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
+++ b/src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
@@ -1,6 +1,7 @@
 /*******************************************************************************
 * Copyright 2021-2023 Intel Corporation
 * Copyright 2021-2024 FUJITSU LIMITED
+* Copyright 2024-2025 Arm Ltd. and affiliates
 *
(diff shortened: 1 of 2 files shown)


theComputeKid (approved):

theComputeKid:
@vpirogov @mgouicem By moving jit above ACL, we also solve the random failures in AArch64 CI. Could you please look at this ASAP? Thanks.

dzarukin:
> @vpirogov @mgouicem By moving jit above ACL, we also solve the random failures in AArch64 CI. Could you please look at this ASAP? Thanks.

Just for the record: I wouldn't call "solving" a solution when making one code running over the other. Rather "sweeping under the rug". Once the code gets executed, sporadic issues will pop up back, but with less context.

dzarukin (approved):

theComputeKid:
@dzarukin I agree, and that is why the reported issue is still open and not linked to this PR.

PR #2410, opened 2025-01-15
benchdnn: graph: better mask filling for SDPA pattern

dzarukin:
There are several reports (MFDNN-12857, MFDNN-12933) about unstable correctness results of SDPA patterns, especially when micro-kernel implementation is picked up. Test cases were reduced in shapes to make existing implementation passing.

The core of the issue can be described as this:
* Mask in SDPA pattern is the mechanism to control connections between tokens. Several masks options are [available](https://medium.com/@swarms/understanding-masking-in-pytorch-for-attention-mechanisms-e725059fd49f) in PyTorch, thanks @TaoLv for the link.
  - Padding mask which has `[mb, 1, 1, seq_len]` shape. In this case if the framework padded several elements over `seq_len`, or `N`, dimension, those several points will be filled with `-infinity` to remove the connection between real and imaginary tokens.
  - Look-ahead mask (or causal mask) which has `[mb, 1, seq_len, seq_len]` shape. In this case each token wants to connect to itself and past tokens, but not to the future ones, thus, upper-top triangle of the mask tensor is filled with `-infinity`.
* Mask filling in benchdnn was based on binary op and with latest extension (which didn't help and was not necessary) fills values in [-8; 8] integer range.
* Micro-kernel implementation instead of doing `QK_output * scale + mask` does `QK_output + mask / scale` with scaling at SoftMax stage. Division of mask over scale is not precise (as no floating-point values are expected there), which is the main point of data discrepancy between the benchdnn reference path and the library path.

This PR aligns mask filling with real expectations:
* Padding masks will be ignored and filled with zeros in case when they are represented with "Add" op, or Condition will be set with zeros/ones, depending on what the QK_output's input number in Select is. The reason behind ignoring it is that the library simply can't make an assumption about padding as it's not doing this padding. It's a framework level responsibility, benchdnn will assume the user requested `seq_len` passed to the case.
* Causal mask will be aligned the way it should look like. When passed explicitly and having last two non-unit dimensions, it will be assumed the mask asked is look-ahead, and properly filled memory will be passed inside the library. When passed implicitly though a special subgraph sequence, the buffer will still be generated for proper reference pass to have results matched, however the library won't take it.

The PR also has a side fix for MFDNN-12994 that simply disables the run if `--mode-modifier=M` was specified as currently the driver doesn't support this capability.


dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -136,6 +136,120 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                auto *aop_in_lt = &aop.in_lts_[i];
+                auto *parent_op = &dg_->get_op_by_out_lt(aop_in_lt->id_);
+                if (!parent_op->empty()) {
+                    if (aop_in_lt->get_data_type()
+                            != logical_tensor::data_type::boolean) {
+                        // Filtering out condition for implicit causal mask.
+                        //
+                        // This is the qk_data, need to know its offset to
+                        // properly fill condition for padding mask.
+                        qk_data_offset = i;
```
ElaineBao, line 175: IIUC, The logic here wants to make sure that an implicit causal mask should not encounter this condition.
However, It seems it can still run into this condition:
1. The src_0 (qk_data) of the select operation in implicit causal mask has a parent_op (which passes the first condition of `!parent_op->empty()`).
2. The data type of src_0 is not boolean (which passes the second condition of `aop_in_lt->get_data_type() != logical_tensor::data_type::boolean`).

So it seems possible for the implicit causal mask to encounter `qk_data_offset = i`, which will make the data filling later incorrect.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -136,6 +136,120 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                    filling_type = filling_type_t::minus_infinity;
+                } else {
+                    // Padding mask.
+                    assert(qk_data_offset == 1 || qk_data_offset == 2);
+                    // Fill condition depending on qk values tensor to use only
+                    // its values, which is equivalent of not using a mask.
+                    if (qk_data_offset == 1) {
+                        filling_type = filling_type_t::one;
+                    } else if (qk_data_offset == 2) {
+                        filling_type = filling_type_t::zero;
```
ElaineBao, line 237: Could you please elaborate why the filling type is one or zero regarding qk_data_offset = 1 / 2?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -196,8 +324,10 @@ int partition_data_displacer_t::displace_input_data(
 
     // do the reverse job
     auto *parent_op = &dg_->get_op_by_out_lt(tensor.id_);
+    bool backward_path_launched = false;
     while (filling_type == filling_type_t::quantization && !parent_op->empty()
             && op_ids_set_.find(parent_op->id_) != op_ids_set_.end()) {
+        backward_path_launched = true;
```
ElaineBao, line 330: It seems this flag is only used for printing, is it intended?

dzarukin: Yes, that message is only meaningful for quantization filling. The other possibility is to introduce same message next to each break. I decided to stick to this one.

tests/benchdnn/graph/input_displacer.cpp
dzarukin: Select takes the value from src_0 or src_1 depending on the value in cond. The desired value is always the one coming from qk_matmul. Depending on its position in the op, condition should use `1` to pick those values from src_0, or `0` to pick it from src_1.

ElaineBao: Sorry, I thought it was filling the mask instead of the condition. If it's filling the condition, then make sense 👍

tests/benchdnn/graph/input_displacer.cpp
dzarukin: Agreed, thanks for testing it for me. I've introduced an additional variable to track whether the condition tensor had a parent or not, it should resolve it.

ElaineBao (approved):

TaoLv (approved):

wzt1997 (approved):

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -398,4 +534,34 @@ int partition_data_displacer_t::gen_pow2_filling(dnn_mem_t &mem,
...
+    const int64_t N = dims[ndims - 1];
+
+    benchdnn_parallel_nd(batch, M, N, [&](int64_t b, int64_t m, int64_t n) {
+        int64_t idx = b * M * N + m * N + n;
+        // Note: we use a different seed for each chunk to avoid
+        // repeating patterns. We could use discard(idx_start) too but
+        // it has a complexity in O(idx_start). We also add 1 to avoid
+        // seeding with 0.
+        std::minstd_rand int_seed(idx + 1);
+        int_seed.discard(1);
```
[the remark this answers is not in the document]
dzarukin: Good catch, an artifact of copy-paste, removed.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -187,7 +305,21 @@ int partition_data_displacer_t::displace_input_data(
...
+    } else if (filling_type == filling_type_t::zero) {
+        static const std::vector<float> user_set {0.f};
+        fill_cfg_t fill_cfg(user_set, "Explicit_padding_mask");
+        SAFE(gen_fixed_set_filling(mem_replace, mem.md_, fill_cfg, res), WARN);
+    } else if (filling_type == filling_type_t::one) {
+        static const std::vector<float> user_set {1.f};
+        fill_cfg_t fill_cfg(user_set, "Explicit_padding_mask");
+        SAFE(gen_fixed_set_filling(mem_replace, mem.md_, fill_cfg, res), WARN);
     } else {
         assert(!"unexepcted filling type");
```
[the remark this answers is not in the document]
dzarukin: Fixed, thanks.

avmanerikar (approved):

PR #2417, opened 2025-01-15
benchdnn: softmax: update setup_cmp

dzarukin:
Way of checking softmax type of comparison did take into account acc_mode (by mistake).
This PR adds an argument to skip acc_mode (as it shouldn't be considered as an attribute) and updates softmax comparison to become p2p back again for logsoftmax.


dzarukin:
make test
disable benchdnn_all
enable benchdnn_softmax

spalicki (approved):

avmanerikar (approved):

PR #2420, opened 2025-01-15
x64: updates for brgemm kernel and avx2 brgemm convolutions

ankalinin:
This request contains a few general brgemm code changes and some performance updates for avx2:

- brgemm: introduce brgemm_desc_finalize function  :  this function must be called after all the parameters of the brgemm descriptor are set.  As result brgemm blocking is called only once having all needed information
- brgemm conv: update to reduce the number of generated kernels: small update for brgemm non-amx convolutions 
- brgemm: update non-amx blocking: more careful distribution of available vmm registers. It's especially important for avx2.  It also determine which order of data loading in microkernel to choose - n_bcast_1_load or 1_bcast_n_load .
- brgemm kernel: update gemm microkernel: just update the kernel code
- brgemm kernel: reduce number of vpad variants: it reduces the size of generated kernel
- brgemm conv: knob for brgemm microkernel decomposition: we may want to tune microkernel decomposition
- brgemm conv: introduce loop order gcndhw: for some shapes such loop order makes sense in terms of performance
- brgemm conv: update cache constants and ur estimation: update platform values used in convolution blocking selection
- brgemm conv: heuristic for big convolution on avx2

Performance testing by openvino on MTL. Here are a ratio brgemm to jit implementation for rls-v3.5 and for this brgemm update over v3.5:

image

### DIFF
--- a/src/cpu/x64/brgemm/brgemm.cpp
+++ b/src/cpu/x64/brgemm/brgemm.cpp
@@ -261,13 +261,6 @@ status_t brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
                 brg->dt_b == u8, is_superset(brg->isa_impl, avx512_core_amx)))
         return status::unimplemented;
 
-    CHECK(brgemm_blocking(brg));
-
-    // avx2_vnni_2 kernel with xf16 data type requires blocked weights.
-    if (brg->isa_impl == avx2_vnni_2 && brg->is_xf16()
-            && brg->LDB % brg->ld_block > 0)
-        return status::unimplemented;
-
     return status::success;
 }
 
@@ -290,8 +283,6 @@ status_t brdgmm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
     if (utils::everyone_is(
                 false, brg->is_int8, brg->is_bf16, brg->is_f32, brg->is_f16))
         return status::unimplemented;
-
-    CHECK(brdgmm_blocking(brg));
 
     return status::success;
 }
@@ -373,9 +364,6 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     if (brg->is_int8 && brg->dt_d == bf16)
         brg->is_bf16_emu
                 = !(mayiuse(avx512_core_bf16) || brg->isa_impl == avx2_vnni_2);
-
-    // Rerun blocking heuristic due to reduced zmm register count
-    if (brg->is_bf16_emu && brg->is_dgmm) CHECK(brdgmm_blocking(brg));
 
     if (!brg->attr()) return status::success;
 
@@ -484,13 +472,6 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     CHECK(init_zp_type(brg->zp_type_b, DNNL_ARG_WEIGHTS));
     CHECK(init_zp_type(brg->zp_type_c, DNNL_ARG_DST));
 
-    // Post-ops may use vector registers so brgemm/brdgmm blocking may need to
-    // be updated
-    if (brg->is_dgmm)
-        CHECK(brdgmm_blocking(brg));
-    else
-        CHECK(brgemm_blocking(brg));
-
     return status::success;
 }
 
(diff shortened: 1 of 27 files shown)


ankalinin:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl

Radu2k (approved):
LGTM, also no regressions in terms of performance. Thanks!

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -172,10 +169,17 @@ struct brg_blocking_t : public jit_brgemm_conv_conf_t {
...
     }
 
-    static int estimate_ur(int oc_block) {
-        const auto est_ur = (oc_block == 64)
-                ? 6
-                : ((oc_block == 48) ? 9 : ((oc_block == 32) ? 14 : 28));
+    static int estimate_ur(cpu_isa_t isa, int oc_block) {
+        auto est_ur = 1;
+        if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2)) {
+            est_ur = (oc_block == 32)
```
tczeszun, line 175: nit: `switch` would be more readable to me

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -1378,79 +1383,73 @@ status_t brgemm_convolution_fwd_t<isa>::execute(const exec_ctx_t &ctx) const {
...
-                    assert(!"Unknown exec type");
-                last_btc.n = n;
-                last_btc.g = g;
-                last_btc.icc = icc;
-                last_btc.odb = odb;
-                last_btc.ohb = ohb;
-                last_btc.owb = owb;
+        int n {0}, g {0}, ocb {0}, odb {0}, ohb {0}, owb {0}, icc {0};
+        for (icc = 0; icc < _pd->ic_chunks; icc++) {
+            BRGEMM_CONV_ITERATOR_INIT;
```
tczeszun, line 1389: shouldn't we introduce gcndhw to 1x1 as well?

src/cpu/x64/jit_brgemm_conv_utils.cpp
ankalinin: agree. updated. thanks

src/cpu/x64/jit_brgemm_conv.cpp
ankalinin: gcndhw  introduced for convolutions with big weights. For 1x1 convolutions weights are usually much smaller. So, I didn't performance experiments for 1x1 convolutions.
I think it would be relatively easy to add gcndhw when needed

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -252,6 +252,7 @@ struct brgemm_desc_t {
     bool with_scales = false;
     bool skip_zp_b_compensation = false;
     bool skip_scales = false;
+    bool n_bcast_1_load {false};
```
dzarukin, line 255: For consistency.
```suggestion
    bool n_bcast_1_load = false;
```

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -583,6 +539,42 @@ status_t brgemm_desc_set_attr(
...
+        // brgemm_kernel implementation. TODO: remove this restriction
+        const int min_bd_block
+                = brg->bdb_tail > 0 ? brg->bdb_tail : brg->bd_block;
+        if ((max_vpad > min_bd_block)) return status::unimplemented;
+    }
+
+    // avx2_vnni_2 kernel with xf16 data type requires blocked weights.
+    if (brg->isa_impl == avx2_vnni_2 && brg->is_xf16()
+            && brg->LDB % brg->ld_block > 0)
+        return status::unimplemented;
```
dzarukin, line 564: If `ld_block` is determined at `brgemm_blocking`, this check should move there. It seems a bit off we get some
non-blocked weights and then just return unimplemented.
If it's determined earlier at the stack, it should move to that earlier place then.

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -583,6 +539,42 @@ status_t brgemm_desc_set_attr(
...
+
+    brg->LDC2_N
+            = (brg->brgattr.LDC2_N != 0) ? brg->brgattr.LDC2_N : brg->ld_block;
```
dzarukin, line 567: Based on the description of the member, it seems this is a part of blocking. Thus, can this be done inside the blocking routine? It might be confusing to update blocking in different places across the brgemm creation stack.

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -583,6 +539,42 @@ status_t brgemm_desc_set_attr(
...
+
+    brg->is_blocked = (brg->LDA2 != brg->LDA || brg->LDB2 != brg->LDB
+            || brg->LDC2_M != brg->LDC || brg->LDC2_N != brg->ld_block);
```
dzarukin, line 570: Ditto.

dzarukin (approved):

src/cpu/x64/brgemm/brgemm_types.hpp
ankalinin: done

src/cpu/x64/brgemm/brgemm.cpp
ankalinin: moved to `brgemm_blocking`

src/cpu/x64/brgemm/brgemm.cpp
ankalinin: This piece of code updated. `LDC2_N` and `is_blocking` calculation are moved to `brgemm_blocking`

src/cpu/x64/brgemm/brgemm.cpp
ankalinin: Yeah. 
`LDC2_N` and `is_blocking` calculation are moved to `brgemm_blocking`. Ditto removed

ankalinin:
I removed commit "brgemm conv: update loop nest - move loop by ic_chunks to the top: we usually don't use blocking by input channels. If we do it makes more sense to have a corresponding block loop at the top level" from this PR as it not directly related to purpose of avx2 convolution optimization and should be updated and included to other PR

densamoilov (approved):

Radu2k (approved):

PR #2425, opened 2025-01-16
xe: softmax: workaround for bug in wg_reduce

syurkevi:
# Description

The current reusable softmax implementation relies on work group collective functions. If the total size of the workgroup isn't a multiple of the subgroup size the OpenCL runtime will perform the reduction with the inactive portion of the subgroup producing incorrect results in the wg_max and wg_add reductions. 
Since reusable softmax reduces the <1024 sizes it is creating incorrect reference results for SDPA testing that targets similar data sizes.

# Checklist

## General

### DIFF
--- a/src/gpu/intel/ocl/reusable_softmax.cpp
+++ b/src/gpu/intel/ocl/reusable_softmax.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


syurkevi:
make test
disable device_cpu
enable device_gpu
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_softmax
enable benchdnn_graph

umar456 (approved):
This fixes many of the SDPA tests that were failing with odd shaped tensors. I wonder why this was being missed by the benchdnn tests.

pv-pterab-s (approved):
lgtm

src/gpu/intel/ocl/reusable_softmax.cpp
```diff
@@ -180,6 +180,20 @@ status_t reusable_softmax_fwd_t::pd_t::init_dispatch_workgroup_per_reduction(
     conf.gws_params = dispatch.get_compile_params();
     rt_conf.gws_params = dispatch.get_runtime_params();
 
+    auto dispatch_lws = dispatch.get_runtime_params().nd_range.local_range();
+    auto dispatch_gws = dispatch.get_runtime_params().nd_range.global_range();
+
+    auto *device_info = compute_engine->device_info();
+    const size_t multiple_of_sg_lws
+            = utils::rnd_up(dispatch_lws[0], device_info->max_subgroup_size());
```
rjoursler, line 188: A better workaround would be to just set the subgroup size in the dispatch config so that no fixup is required. It seems like this is also a bug in the OpenCL runtime, could you provide a reproducer to them as well?

PR #2430, opened 2025-01-16
brgconv: fix long pd time for problems with dilation (fixed MFDNN-12916)

dzarukin:
MFDNN-12916

The reason for slowdown is executing a really long cycle where a full brgemm descriptor as going to be created. In the reality, only a short set of M values must be check.
The reason why this loop is called is dilated convolution can fail the condition for exec_vpad type (left a TODO comment there). This type would only call verification for two M values and that's it.

The solution was to adopt the approach for exec_base type from the actual initialization step.

Measured on sdpdal774583
Timing before (cpdtime is a primitive descriptor creation time):
```
run: --mode=P --conv --dir=FWD_I --dt=u8:s8:u8 --stag=acb --dtag=acb --attr-scales=wei:per_oc --attr-post-ops=relu:0.1+linear:350.855:35 mb1ic32iw129024oc32ow129024kw3pw5dw4
Output template: %impl%,%+cpdtime%
brg_conv_fwd:avx2_vnni_2,37.1177
```

Timing after:
```
run: --mode=P --conv --dir=FWD_I --dt=u8:s8:u8 --stag=acb --dtag=acb --attr-scales=wei:per_oc --attr-post-ops=relu:0.1+linear:350.855:35 mb1ic32iw129024oc32ow129024kw3pw5dw4
Output template: %impl%,%+cpdtime%
brg_conv_fwd:avx2_vnni_2,1.89819
```

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv.cpp
+++ b/src/cpu/x64/jit_brgemm_conv.cpp
@@ -604,10 +604,9 @@ status_t brgemm_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
             for (int kw = kw_s; kw < kw_f; kw++) {
                 brgemm_convolution_utils::get_ow_range(
                         jcp_, ow, kw, ow_s, ow_f);
-                if (ow_f - ow_s <= 0) continue;
-
-                auto M = ow_f - ow_s;
+                const auto M = ow_f - ow_s;
                 if (M <= 0) continue;
+
                 for (const auto &key_value_pair : batchsizes) {
                     const int kd_b = key_value_pair.first[0];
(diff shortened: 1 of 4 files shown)


dzarukin:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_conv
set test_set=NIGHTLY

densamoilov (approved):

ankalinin (approved):

dzarukin:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_conv
set test_set=NIGHTLY

PR #1999, opened 2024-07-18
rfcs: add proposal for collecting and analyzing synthetic data

rjoursler:
[Document Link](https://github.com/oneapi-src/oneDNN/blob/rjoursle/synthetic_rfc/rfcs/20240718-synthetic-benchmark/README.md)
This RFC proposes new tooling to enable a data pipeline on synthetic data analyze and validate oneDNN performance.

PR #2434, opened 2025-01-17
test: benchdnn: inputs: matmul: add random ci test

rjoursler:
The test_matmul_ci test suite coverage has allowed a lot of bug escapes to nightly testing for GPU development. This test suite is intended to supplement that testing with some randomly generated problems. Sample results on PVC which shows a functionality error rate of around 3.6%, with around 25% of the problems being skipped due to falling back to reference.

```
// XeHPC
11:FAILED (errors:2318607 total:2683350) __REPRO: --matmul --engine=gpu --dt=f32:u8:f16 --wtag=abc --dtag=abc --attr-fpmath=strict:true --skip-impl=ref 3x2670x95:3x95x335
92:FAILED (errors:90070 total:180360) __REPRO: --matmul --engine=gpu --dt=f32:u8:f32 --stag=cab --wtag=cab --attr-fpmath=strict:true --skip-impl=ref 5x216x2430:5x2430x167
160:FAILED (errors:446 total:1115) __REPRO: --matmul --engine=gpu --dt=f32:u8:f32 --stag=cab --wtag=cab --dtag=bac --attr-fpmath=strict:true --skip-impl=ref 1x223x370:1x370x5
200:FAILED (errors:48320 total:58968) __REPRO: --matmul --engine=gpu --dt=f32:u8:f32 --stag=acb --wtag=bac --dtag=bac --attr-fpmath=f16:true --skip-impl=ref 3x84x4350:3x4350x234
225:FAILED (errors:708235 total:1643040) __REPRO: --matmul --engine=gpu --dt=f32:u8:f16 --stag=acb --wtag=bac --dtag=bac --attr-fpmath=strict:true --skip-impl=ref 2x280x30:2x30x2934
227:FAILED (errors:365998 total:688536) __REPRO: --matmul --engine=gpu --dt=f32:s8:f16 --stag=acb --wtag=bac --attr-fpmath=strict:true --skip-impl=ref 6x524x31:6x31x219
248:FAILED (errors:30076089 total:36065696) __REPRO: --matmul --engine=gpu --dt=f32:u8:bf16 --stag=cab --attr-fpmath=f16:true --skip-impl=ref 4x4468x18:4x18x2018
287:FAILED (errors:39535 total:39900) __REPRO: --matmul --engine=gpu --dt=f32:u8:f32 --stag=acb --dtag=abc --attr-fpmath=tf32:true --skip-impl=ref 5x2x381:5x381x3990
294:FAILED (errors:115353565 total:161063100) __REPRO: --matmul --engine=gpu --dt=f32:s8:bf16 --stag=abc --wtag=bac --dtag=abc --attr-fpmath=tf32:true --skip-impl=ref 4x5445x73:4x73x7395
311:FAILED (errors:10949500 total:25139667) __REPRO: --matmul --engine=gpu --dt=f32:s8:bf16 --dtag=abc --attr-fpmath=strict:true --skip-impl=ref 7x1679x1:7x1x2139
(... 26 more cases ...)
============================
tests:1000 passed:702 skipped:252 mistrusted:10 unimplemented:0 invalid_arguments:0 failed:36 listed:0
total: 72.70s; fill: 17.62s (24%); compute_ref: 1.10s (2%); compare: 4.28s (6%);
```

For reference, the current test_matmul_ci test takes about 16x as long as this one and performs around 10x more test cases
```
// XeHPC test_matmul_ci
tests:8913 passed:7631 skipped:1268 mistrusted:14 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total: 1142.52s; fill: 84.72s (7%); compute_ref: 910.19s (80%); compare: 28.03s (2%);
```

Running the same workload on CPU finds 1 functional error although around 80% of the workloads is skipped due to dispatching to reference or being an unsupported implementation.
```
//Saphire Rapids
520:FAILED (errors:360 total:360) __REPRO: --matmul --dt=f16:f16:f32 --stag=cab --wtag=cab --skip-impl=ref 3x1x4908:3x4908x120
============================
tests:1000 passed:205 skipped:794 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:1 listed:0
total: 10.34s; fill: 5.33s (52%); compute_ref: 1.71s (17%); compare: 1.15s (11%);
```
Due to all the skipped problems, execution time is minimal, taking only 10 seconds to complete.

The python script is based on experimental performance analysis done as part of exploring #1999, most of which is on the development branch `rjoursle/synthdnn`. 

# Details

## Why Random Sampling Works:

Say we are testing a function that takes 10 boolean inputs, then brute-force validation would take `2^10` tests. In practice, not all boolean inputs are going to be dependent, so we will be able to get complete test coverage with significantly fewer tests. For example, if each input is completely independent, 2 tests would suffice. In practice, we expect a more complex dependency structure. For demonstration, say this structure is `{{2*i ,2*i+1, 2*i+2}: i in [0, 5)}`. In this case, we can cover the whole testing space in around 8 tests. The issue here is that we need significant implementation knowledge to craft and maintain this test set. If we switch to random sampling though, we can (likely) cover the testing space in 24 tests (as each dependent set likely achieved all 8 configurations) significantly less than the original `2^10` cases. In addition, this scales well as generalizing the demo case to  `n` inputs results in (likely) testing coverage in  `8 * log(n)` tests (see coupon collector problem the general idea behind how the math works out). By its nature, the random sampling will be robust to the input dependency structure as well, making this technique practical for validation.

## Test Set Stability
One of the key issues with random testing is reproducibility. While the generating script is deterministic, as we are using a psuedo-random generator, we can expect the generated output to change significantly as features are added to it. While we can introduce measures to maintain test stability, I think this is misguided. The main method for introducing stability is to append test cases to a batch file as new features are implemented. This introduces a few problems, first it changes the covering complexity discussed above from `O(k*log(n)` to `O(k*n)` and, in addition, it introduces a maintenance burden as developers have to add the new cases.  Because of this, the proposition here is that we simply regenerate the test suite when new tests are necessary. To address stability, we should instead limit how often new test suites are generated. As this generator becomes feature complete, I expect these updates will be naturally limiting. Because of this, I propose that we have periodic updates as new features are added to the generator.

## Coverage
The intention is that this test suite covers all configurations explicitly listed in the [developer guide](https://oneapi-src.github.io/oneDNN/dev_guide_matmul.html) to ensure this suite is applicable for all hardware platforms. As not all listed configurations are well supported on all platforms, testing is limited to optimized implementations. The test coverage is incomplete and will be extended in future PRs. The current coverage includes all data types combinations supported by source, weights, and destination,  support for bias is not yet implemented. All layout combinations of `any` and buffers with contiguous `m`, `n`, or `k` dimensions are tested as well, with the additional restriction that dst has contiguous `n`. Additionally, except for `fpmath-mode`, no attributes or post-ops are supported.


vpirogov:
Maybe a python script + fixed seed instead of 10k lines of autogenerated tests?

rjoursler:
> Maybe a python script + fixed seed instead of 10k lines of autogenerated test cases?

Sure!  Just to be clear as to the intention, should these be generated at build time by CMAKE or at runtime by benchdnn?

vpirogov:
> Sure! Just to be clear as to the intention, should these be generated at build time or at runtime?

I think @dzarukin had some thought on that.

rjoursler:
make test
enable test_device_cpu
enable test_device_gpu

rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

tests/synthdnn/synthdnn.py
```diff
@@ -0,0 +1,97 @@
...
+    matmul_parser.add_argument("-t", "--types", default="all",
+                               help="dt:dt:dt(optional fpmath-mode), comma separated list of type configurations or \"all\" for every supported type")
+def matmul_main(args):
+    batchFile = open(args.batch_file, "w+t") if args.batch_file != None else None
+    if args.benchdnn != None and batchFile == None:
+        batchFile = NamedTemporaryFile("w+t")
+
+    if batchFile:
+        log(f"generating batch file: {batchFile.name}")
+        region = matmul_sampler.Region(args.region)
```
Simonsays095, line 62: The `Region` class is outside of `matmul_sampler`, and needs to be imported/used separately.

tests/synthdnn/synthdnn.py
```diff
@@ -0,0 +1,97 @@
...
+    matmul_parser.add_argument("--engine", default="cpu", help="engine used for benchdnn execution")
+    matmul_parser.add_argument("--collect", default="corr", help="benchdnn collection type, can be one of [corr, perf]")
+    matmul_parser.add_argument("-n", "--name", default="", help="sample name")
+
+    # Sampler Arguments
+    matmul_parser.add_argument("-l", "--layouts", default="all",
+                               help="stag:wtag:dtag, comma separated list of layouts or \"all\" for every supported layout")
+    matmul_parser.add_argument("-m", "--iter-mode", default="zip", help ="iteration mode, must be one of zip or product")
+    matmul_parser.add_argument("-r", "--region", default="(1,1,1,1):(8,8192,8192,8192):(1,1,1,1)",
+                               help="(m_min,k_min,n_min)-(m_max,k_max,n_max)/(m_align,k_align, n_align)")
```
Simonsays095, line 51: The help message needs to be updated to use the `:`-delimited string format.

tests/synthdnn/matmul/sampler.py
```diff
@@ -0,0 +1,148 @@
...
+                        self.region.alignment[i],
+                        )
+                dims_tuple = tuple(dims)
+                if dims_tuple not in self.seen:
+                    self.seen.add(dims_tuple)
+                    return Dims(dims[:-3], dims[-3], dims[-2], dims[-1])
+
+            raise RuntimeError(
+                "Can not effectively sample problems larger than: "
+                + str(min_size)
```
Simonsays095, line 147: This variable doesn't exist. Should probably be something like:
```suggestion
                f"Can not sample >{len(self.seen)} problems in region {self.region}"
```

tests/synthdnn/matmul/primitive.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __iter__(self):
+        return iter(self.values)
+
+    @staticmethod
+    def supported():
+        support_matrix =  [
+            [["f64"], ["f64"], ["f64"], [""]],
+            [["f32"], ["f32"], ["f32"], ["", "(tf32)","(f16)", "(bf16)"]],
+            [["f16"], ["f16"], ["f16", "u8", "s8"], [""]],
```
Simonsays095, line 115: This row is entirely duplicates of entries in the following row.

tests/synthdnn/matmul/primitive.py
```diff
@@ -0,0 +1,147 @@
...
+            [["f16"], ["f16", "u8", "s8","u4","s4"], ["f16", "u8", "s8"], ["", "(f16:true)"]],
+            [["f16"], ["f16", "u8", "s8"], ["f32"], [""]],
+            [["bf16"], ["bf16"], ["bf16", "u8", "s8"], [""]],
```
Simonsays095, line 118: Same with this row.

tests/synthdnn/matmul/primitive.py
```diff
@@ -0,0 +1,147 @@
...
+            [["f32", "bf16"], ["bf16", "u8", "s8","u4","s4"], ["bf16", "u8", "s8"], ["", "(bf16:true)"]],
+            [["bf16"], ["bf16", "u8", "s8"], ["f32"], [""]],
+            [["f32"], ["u8", "s8", "u4", "s4"], ["f32"], [""]],
+            [["u8","s8"], ["u8", "s8"], ["f32", "bf16", "f16", "s32", "u8", "s8"], [""]],
```
Simonsays095, line 122: The [documentation's support matrix](https://oneapi-src.github.io/oneDNN/dev_guide_matmul.html#data-types) differs from these lines quite a bit. Is there a reason not to effectively copy the intended support into this table?

tests/synthdnn/synthdnn.py
```diff
@@ -0,0 +1,97 @@
...
+        elif(args.collect == "perf"):
+            benchdnn_args = f"--engine={args.engine} --matmul --mode=F --cold-cache=all --perf-template=sample,{args.name},%prb%,%0Gflops%,%0Gbw% --memory-kind=usm_device --attr-scratchpad=user"
+            if args.name.find(',') != -1:
+                error(f"sample name {args.name} contains invalid character: ,")
+        else:
+            error(f"unknown collection method {args.collect}")
+
+        cmd = f"{args.benchdnn} {benchdnn_args} --batch={batchFile.name}"
+        log(f"executing: {cmd}")
+        ret = os.system(cmd)
```
Simonsays095, line 87: In general, I think `subprocess.call` is better than `os.system`, and would eventually be required for report generation (to capture the output).

tests/synthdnn/matmul/primitive.py
Simonsays095: I was looking at bf16 mainly:
image

dzarukin: Hardware support for u8 weights appeared with AMX only, so it's not advertised.

tests/synthdnn/synthdnn.py
rjoursler: I intend to leave this as is until `subprocess.call`  becomes necessary as I am attempting to minimize the script dependencies. By using subprocess, we increase the number of dependencies as the `os` modules is already used elsewhere.

tests/synthdnn/matmul/primitive.py
rjoursler: Thanks for the information @dzarukin, I will go ahead and drop that combination.

@Simonsays095, thanks for pointing that out, I believe I have closed that gap. The testing set still isn't quite aligned as I am attempting to match the documentation with what I have seen in practice.  In particular, I have not seen any mixed bf16/f16 workloads, so I am dropping those cases. In addition, the f16 and bf16 data types are used in similar scenarios, so I aligned their coverage. We can change this if there are disagreements with the approach.

tests/synthdnn/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+Report Generation: Not yet implemented, intended usage:
+```
+python3 synthdnn.py <primtive> --engine=gpu --collect perf <benchdnn_file> > data.log
+python3 synthdnn.py report --report-kind <kind>  data.log
+
+```
+
+See `synthdnn.py -h` for additional details.
```
dzarukin, line 28: I recommend to move all Python stuff to scripts/ folder where verbose_converter lives.

tests/synthdnn/README.md
```diff
@@ -0,0 +1,28 @@
+# Synthdnn
+
+Synthdnn is a suite of scripts for collecting and analyzing oneDNN performance
+across a randomly generated data. The general architecture is intended to follow
+a data pipeline composed of synthetic problem generation, data collection, and
+data analysis. The `synthdnn.py` script provides a command line interface to
+these tools. Sample Usage:
+
+
+Data Generation:
```
dzarukin, line 10: Suggest to differentiate and rename:
`Data` that refers to benchdnn problems or cases to be renamed into `Problems` or `Cases`.
`Data` that refers to performance numbers to be renamed into `Performance Data`.
Otherwise, the context of different data might be mixed up.

tests/synthdnn/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+python3 synthdnn.py <primitive> [sampling controls] -b <batch_file>
+```
```
dzarukin, line 13: Please specify a fixed line that is used to generate `test_matmul_random_ci`.

I also think that `test` is over-used across the validation. Let's make it `harness_matmul_generated_ci` and include that into `test_matmul_ci`.

tests/synthdnn/README.md
```diff
@@ -0,0 +1,28 @@
...
+python3 synthdnn.py <primitive> [sampling controls] --engine=<engine> --collect <data_kind> <benchdnn_file>
+```
+
+Report Generation: Not yet implemented, intended usage:
+```
+python3 synthdnn.py <primtive> --engine=gpu --collect perf <benchdnn_file> > data.log
+python3 synthdnn.py report --report-kind <kind>  data.log
+
+```
```
dzarukin, line 26: I don't think this section should make into production so far.

tests/benchdnn/CMakeLists.txt
```diff
@@ -165,6 +168,17 @@ foreach(driver ${all_drivers})
...
+        set(SYNTHDNN "${CMAKE_CURRENT_SOURCE_DIR}/../synthdnn/synthdnn.py")
+        set(RANDOM_CI "${CMAKE_CURRENT_BINARY_DIR}/inputs/${driver}/test_${driver}_random_ci")
+        execute_process(COMMAND ${PYTHON3_EXECUTABLE} ${SYNTHDNN} ${driver} -s400 -b ${RANDOM_CI} RESULT_VARIABLE RET OUTPUT_VARIABLE OUT ERROR_VARIABLE OUT)
+        if(${RET})
+            message(Warn "Generation of ${RANDOM_CI} exited with error code ${RET}")
+        else()
+            set(test_files_ci "${test_files_ci};test_${driver}_random_ci")
+        endif()
+    endif()
+
```
dzarukin, line 181: I proposed to keep the process of (re-)generating problems manual, so this change shouldn't be needed.

tests/benchdnn/inputs/matmul/test_matmul_random_ci
```diff
@@ -0,0 +1,400 @@
+--reset --allow-enum-tags-only=0 --stag=bac --wtag=cba --dtag=cab --dt=f8_e4m3:f8_e4m3:bf16  1x362x16:1x16x12
```
dzarukin, line 1: `--allow-enum-tags-only=0` is fine for converter but not for a generator. It shouldn't run beyond the library capabilities.

tests/benchdnn/inputs/matmul/test_matmul_random_ci
```diff
@@ -0,0 +1,400 @@
...
+--reset --allow-enum-tags-only=0 --stag=cba --wtag=acb --dtag=any --dt=bf16:u8:f32 --attr-fpmath=bf16:true 6x953x602:6x602x54
+--reset --allow-enum-tags-only=0 --stag=abc --wtag=cba --dtag=cab --dt=f32:s8:bf16  3x3118x29:3x29x2
```
dzarukin, line 3: If there's a half of `dtag` not having `c` in the last dimensions, all those times we would be testing the reference implementation, as non-dense N support doesn't make much sense out there. It's fine to have 3-5% covered, but not beyond that. Can the script have weighted distribution for those?

Same reasoning behind weights, it's highly unlikely users would pick non-performant cases. Thus, 70% of the time weights format must be `any`, and forced plain layouts to one that are actually used out there for rest.

src tag should be mostly tied to `c` last as well.

dzarukin: `--dt=f32:s8:bf16` doesn't have much sense without `--attr-fpmath=f32:true`

tests/synthdnn/README.md
rjoursler: I modified the script to log the generation command in the generated file. I don't intend to include the command here as that seems less flexible and maintainable  (for example, if we start using multiple generated test suites).

tests/benchdnn/CMakeLists.txt
rjoursler: Agreed.

tests/benchdnn/inputs/matmul/test_matmul_random_ci
rjoursler: I modified the layouts to follow the following restrictions in our documentation:

> The MatMul primitive is generally optimized for the case in which memory objects use plain memory formats. 
Additionally, the  and must have at least one of the axes m or k and n or k contiguous (i.e., stride=1) respectively.

>  The memory format of the destination tensor should always be plain with n axis contiguous.

From what I have seen, all 3d layouts corresponding to these restrictions are used on GPU. Additionally, GPU also supports `k` axis contiguous for `dst` as well as it simplifies internal reuse although I when ahead and dropped that.  As an aside, I don't really see a reason to foist the `dst` restriction on oneDNN users, it is fairly straightforward to just normalize to the current restriction using the identity `(AB)^T = B^TA^T`. By requiring contiguous `n`, we are just pushing this transform up to the frameworks.

> Can the script have weighted distribution for those?

I think this would be overly complicated to implement. It would be simpler to add separately generated test suites which are guided by bug escapes encountered in practice.

> Thus, 70% of the time weights format must be any

> src tag should be mostly tied to c last as well.

These seem very device/framework/workload specific and seem like they should be in more targeted test suites if they are found to be necessary.

tests/synthdnn/synthdnn.py
```diff
@@ -0,0 +1,97 @@
...
+def error(output):
+    print("synthdnn: error: " + output)
+    exit(1)
+
+
+def setup_matmul_subparser(subparsers):
+    matmul_parser = subparsers.add_parser('matmul', help='call with -h for information')
+    matmul_parser.add_argument("--subprogram_main", default=matmul_main, help=argparse.SUPPRESS)
+
+    matmul_parser.add_argument("-b", "--batch-file", default=None, help="batch file used for storing the sample")
```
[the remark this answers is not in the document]
rjoursler: Good idea. Thanks for the suggestion.

rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

scripts/synthdnn/matmul/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+# This file marks this directory as a package and is needed to allow relative
+# imports. See https://docs.python.org/3/tutorial/modules.html#packages.
```
atkassen, line 18: nit: this file is unnecessary as you aren't using relative imports.

atkassen:
nit: ideally the stuff in scripts/ is formatted with `black`.

rjoursler:
make test
enable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul

scripts/synthdnn/README.md
```diff
@@ -0,0 +1,27 @@
+# Synthdnn
```
densamoilov, line 1: (random spot)
If I understand the idea correctly, the proposal is to include a non-static test set in the coverage to reduce the chances of bugs escaping.

- What do you suggest should be the frequency of updating the test set?
- How should we handle unimplemented cases? For example, the reference matmul for CPU does not support configurations with mixed float and integer data types. In my opinion, it doesn't make much sense to enable those configurations in the reference implementations just to cover synthetic test cases.
- I guess the auto-generated test set could also be used to find gaps in our static coverage to extend it accordingly. Do you see any problems with that?
- How many test cases do you think is reasonable to generate?

I think this could be a good addition to our validation process, as it can be used to identify gaps though I'm not sure how efficient it will be.

densamoilov: Thanks Roy for the detailed response. Overall, I'm OK with the idea, I think we can try it out and see how it goes.
@dzarukin @vpirogov, @mgouicem, I think this requires a nod from the architecture team to proceed.

rjoursler:
make test
enable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul

rjoursler:
make test
enable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul

scripts/synthdnn/README.md
dzarukin: I think we converged on the idea of randomized coverage with files existing permanently and having their content  updated on a regular but scheduled basis.

We agreed on the purpose of such approach which allows to have more extensive coverage and this coverage can be controlled in terms of testing run time.

The implementation of current solution can be improved in a more scalable way. Currently, I don't see an easy path to extend existing solution to any other benchdnn driver or with more knobs matmul supports. I suppose, to move forward till the finish line where every driver is covered in a full scope the core design principles must be established.

I propose to re-use existing IR object from verbose_converter, which is already complete for verbose conversion and has all necessary objects and abstractions. Besides it can solve the problem of benchdnn line generation through the converter leaving only necessity to generate values for the IR and verify their sanity with library limitations.
It doesn't have to happen in this PR but I see it as essential part of the future solution.

dzarukin (approved):

echeresh (approved):

rjoursler:
make test

PR #2436, opened 2025-01-17
cpu: x64: ukernel: transform: add incorrect ldb check

### LINKED ISSUE #2414: Memory corruption caused by transformation ukernel
# Summary
Transformation ukernel for 4x4 BF16 block writes 80 bytes instead of expected 32 bytes

# Version
Main branch, commit 281d20d643bdee595d2ce12db3985f3ebb8cb43e

# Environment
SPR CPU
Ubuntu 22.04

# Steps to reproduce
Use ukernels for 4x4 BF16 input tensors on a machine with AMX-BF16 support.

# Observed behavior
The transformation kernel writes 80 bytes to the output memory, while the data is only 32 bytes long. The generated code has the following encoding instructions:
```
   0x7ffff141e1dc:      vmovdqu16 (%rax),%zmm17{%k7}{z}
   0x7ffff141e1e2:      vmovdqu16 0x8(%rax),%zmm2{%k7}{z}
   0x7ffff141e1ec:      vinsertf64x4 $0x1,%ymm2,%zmm17,%zmm17
   0x7ffff141e1f3:      vpermw %zmm17,%zmm1,%zmm17
   0x7ffff141e1f9:      vmovups %zmm17,(%rbx)
```
Here, k7=0xf. The kernel correctly reads two rows, each with four elements, performs the permutation, and then writes 64 bytes instead of 16 bytes. Thus, it writes 48 extra zero bytes.

# Expected behavior
Transformation kernel doesn't write extra bytes to the output memory.

dzarukin:
This fixes #2414

### DIFF
--- a/doc/ukernel/operations/transform.md
+++ b/doc/ukernel/operations/transform.md
@@ -2,16 +2,26 @@ Data transformation {#dev_guide_ukernel_transform}
 =======================================
 
 >
-> [API Reference](@ref dnnl_api_ukernel_brgemm)
+> [API Reference](@ref dnnl::ukernel::transform)
 >
 
 ## General
 
-The transform ukernel allows users to convert data from one format to the other,
-similar to what reorder primitive provides functionally.
+The [BRGeMM ukernel](@ref dev_guide_ukernel_brgemm) might require the B tensor
+in a specific memory layout depending on target data types and the machine
+architecture. Check the requirement by calling the
+[get_B_pack_type()](@ref dnnl::ukernel::brgemm::get_B_pack_type) function. If it
+returns the [pack32](@ref dnnl::ukernel::pack_type::pack32) type, it implies
+that packing is required, otherwise, packing is not required.
 
-The only output data format supported by this routine is packed format, which is
-required by B matrices in [BRGeMM ukernel](@ref dev_guide_ukernel_brgemm).
+The transform ukernel allows the conversion of data from the original layout,
+which is described as either
+[non-transposed](@ref dnnl::ukernel::pack_type::no_trans) or
+[transposed](@ref dnnl::ukernel::pack_type::trans) to the layout requested by
+the BRGeMM ukernel.
+
+The only supported output packing type is `pack32`.
+
 This is an out-of-place operation.
 
 ## Data Types
@@ -34,7 +44,9 @@ No attribute is supported for transform ukernel.
 
 ## Implementation limitations
 
-- Destination leading dimension only supported values are: 16, 32, 48, or 64.
+- Destination leading dimension, or `out_ld`, must be one of the following
+  values: `16`, `32`, `48`, or `64`. This is the implementation limitation,
+  there are no efficient kernels supported for other leading dimension values.
 
 ## Examples
 
--- a/src/cpu/x64/brgemm/capi/brgemm_api.cpp
+++ b/src/cpu/x64/brgemm/capi/brgemm_api.cpp
@@ -685,6 +685,8 @@ status_t dnnl_transform_create(transform_t **transform, dim_t K, dim_t N,
         pack_type_t in_pack_type, dim_t in_ld, dim_t out_ld, data_type_t in_dt,
         data_type_t out_dt) {
     if (transform == nullptr) return status::invalid_arguments;
+    VCHECK_BRGEMM(utils::one_of(out_ld, 16, 32, 48, 64),
+            "Transform routine supports only \'out_ld\' of 16, 32, 48, or 64.");
 
     *transform
             = new transform_t(K, N, in_pack_type, in_ld, out_ld, in_dt, out_dt);


mgouicem (approved):

ienkovich:
Adding this restriction to the transform documentation would be very useful for users.

dzarukin:
> Adding this restriction to the transform documentation would be very useful for users.

Fair point. I'll update doc page as well.

xuxinzen (approved):

dzarukin:
> Adding this restriction to the transform documentation would be very useful for users.

I just figured that the limitation is documented: https://oneapi-src.github.io/oneDNN/dev_guide_ukernel_transform.html#implementation-limitations

Though the limitation is not attached to a specifically named argument, and I'll give a touch up for better wording.

doc/ukernel/operations/transform.md
```diff
@@ -2,16 +2,26 @@ Data transformation {#dev_guide_ukernel_transform}
...
 
 ## General
 
-The transform ukernel allows users to convert data from one format to the other,
-similar to what reorder primitive provides functionally.
+The [BRGeMM ukernel](@ref dev_guide_ukernel_brgemm) might require the B tensor
+in a specific memory layout depending on target data types and the machine
+architecture. Users should check the requirement by calling the
+[get_B_pack_type()](@ref dnnl::ukernel::brgemm::get_B_pack_type) function. If it
+returned the [pack32](@ref dnnl::ukernel::pack_type::pack32) type, the packing
```
densamoilov, line 14:
```suggestion
[get_B_pack_type()](@ref dnnl::ukernel::brgemm::get_B_pack_type) function. If it
returns the [pack32](@ref dnnl::ukernel::pack_type::pack32) type, the packing
```

doc/ukernel/operations/transform.md
```diff
@@ -2,16 +2,26 @@ Data transformation {#dev_guide_ukernel_transform}
...
+is required, otherwise, it's not.
+
+The transform ukernel allows users to convert data from the original layout
+which are described as either
```
densamoilov, line 18:
```suggestion
which is described as either
```

doc/ukernel/operations/transform.md
```diff
@@ -34,7 +44,9 @@ No attribute is supported for transform ukernel.
 
 ## Implementation limitations
 
-- Destination leading dimension only supported values are: 16, 32, 48, or 64.
+- Destination leading dimension, or `out_ld`, only supported values are: `16`,
+  `32`, `48`, or `64`. This is the implementation limitation, there's no
```
densamoilov, line 48:
```suggestion
  `32`, `48`, or `64`. This is the implementation limitation, there are no
```

doc/ukernel/operations/transform.md
```diff
@@ -34,7 +44,9 @@ No attribute is supported for transform ukernel.
 
 ## Implementation limitations
 
-- Destination leading dimension only supported values are: 16, 32, 48, or 64.
+- Destination leading dimension, or `out_ld`, only supported values are: `16`,
```
densamoilov, line 47:
```suggestion
- Destination leading dimension, or `out_ld`, must be one of the following values: `16`,
```

doc/ukernel/operations/transform.md
```diff
@@ -34,7 +44,9 @@ No attribute is supported for transform ukernel.
...
+  `32`, `48`, or `64`. This is the implementation limitation, there's no
+  effecient kernels supported for other leading dimension values.
```
densamoilov, line 49:
```suggestion
  efficient kernels supported for other leading dimension values.
```

densamoilov: Could you please clarify if there are no efficient kernels or kernels at all for other leading dimensions?

densamoilov (approved):

doc/ukernel/operations/transform.md
dzarukin: Thanks for the suggestions, incorporated.

From the end-user perspective it should be no difference whether there's no kernels or they are not efficient at all. In theory we can plug simple_reorder in, but it would be no point in doing that. It's just a comment explaining where the limitation is coming from.

dzarukin:
@ranukund, could you, please, give it a look? Thank you.

doc/ukernel/operations/transform.md
```diff
@@ -2,16 +2,26 @@ Data transformation {#dev_guide_ukernel_transform}
...
-similar to what reorder primitive provides functionally.
+The [BRGeMM ukernel](@ref dev_guide_ukernel_brgemm) might require the B tensor
+in a specific memory layout depending on target data types and the machine
+architecture. Users should check the requirement by calling the
+[get_B_pack_type()](@ref dnnl::ukernel::brgemm::get_B_pack_type) function. If it
+returns the [pack32](@ref dnnl::ukernel::pack_type::pack32) type, the packing
+is required, otherwise, it's not.
+
+The transform ukernel allows users to convert data from the original layout
+which is described as either
```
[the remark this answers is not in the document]
dzarukin: Thanks for the edits, incorporated.

PR #2453, opened 2025-01-20
gpu: reduction: fix failing gtests

sgeor255:
# Description

#2243 caused some failures in the reduction gtests on intel GPU. This PR reverts a [skip which was incorrectly removed](https://github.com/oneapi-src/oneDNN/pull/2243/files#diff-38eb9abaf39baf897c0c419e65db5b41e8f52d2ecb9aee896d1a99c8ba1fa5b5L52-L53) which skips the tests on non-cpu engines. The `ifdef` for INTEL vendor is added as the tests should be enabled on generic backend.

# Checklist

## General

### DIFF
--- a/tests/gtests/test_reduction.cpp
+++ b/tests/gtests/test_reduction.cpp
@@ -49,6 +49,10 @@ protected:
 
         SKIP_IF(unsupported_data_type(src_dt),
                 "Engine does not support this data type.");
+#if DNNL_GPU_VENDOR == DNNL_VENDOR_INTEL
+        SKIP_IF(get_test_engine().get_kind() != engine::kind::cpu,
+                "Engine does not support this primitive.");
+#endif
         SKIP_IF_CUDA(p.aalgorithm != algorithm::reduction_max
                         && p.aalgorithm != algorithm::reduction_min
                         && p.aalgorithm != algorithm::reduction_sum


Rbiessy (approved):

tests/gtests/test_reduction.cpp
```diff
@@ -49,6 +49,10 @@ class reduction_test_t
 
         SKIP_IF(unsupported_data_type(src_dt),
                 "Engine does not support this data type.");
+#if DNNL_GPU_VENDOR == DNNL_VENDOR_INTEL
+        SKIP_IF(get_test_engine().get_kind() != engine::kind::cpu,
+                "Engine does not support this primitive.");
+#endif
```
mgouicem, line 55: The Intel GPU backend should have a bunch of implementation to pick from, for both OCL and SYCL runtimes.
So even though your previous PR made [those bugs](https://github.com/oneapi-src/oneDNN/pull/2243#issuecomment-2598761745) visible, I don't think rehiding these issues is the proper fix here.

sgeor255: My bad, I assumed those tests had been disabled on purpose and I re-added the check. From a quick look, both `intel::ocl::atomic_reduction_t` & `intel::ocl::combined_reduction_t` implementations (which are the top 2 in the implementation list) can't handle 0-dimensional inputs correctly. `intel::ocl::ref_reduction_t` seems to work correctly.

Simonsays095: For the record, I've created #2469 which fixes these failures by handling the zero-dim cases "correctly".

PR #2457, opened 2025-01-21
graph: backend: dnnl: fix check for sdpa

ElaineBao:
# Description

There are two fixes:

- [x] fix select check for sdpa primitive: sdpa prmitive doesn't support select after scale(optional) and mask(optional), filter them out.
- [x] add check for sdp decomp kernel: when cpu runtime = threadpool,  sdpa with implicit causal mask will first run into sdp decomp kernel, which is not supported, filter them out.

### DIFF
--- a/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
@@ -207,9 +207,11 @@ status_t sdp_primitive_config_t::initial_check(
                 post_op = get_post_op(post_op);
             }
             // mask
-            if (post_op && post_op->get_kind() == graph::op_kind::Add) {
-                // Mask exists, update post_op and traverse to next op
-                post_op = get_post_op(post_op);
+            if (post_op) {
+                if (post_op->get_kind() == graph::op_kind::Add) {
+                    // Mask exists, update post_op and traverse to next op
+                    post_op = get_post_op(post_op);
+                }
                 // Not support select after scale(optional) and mask(optional)
                 // Distill-Bert:[mm1] --> [scale]* --> [mask]* --> [select] --> ...
                 VCHECK_SDP_PRIMITIVE(post_op
--- a/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
@@ -457,6 +457,8 @@ impl::status_t sdp_decomp_config_t::record_input_offset(
                     graph::op_kind::SoftMax};
     for (const auto &cur_op : sg->get_ops()) {
         const auto &op_kind = cur_op->get_kind();
+        VCHECK_SDP_DECOMP(op_kind != graph::op_kind::GenIndex,
+                status::unimplemented, "Not support implicit causal mask");
         VCHECK_SDP_DECOMP(op_kind != graph::op_kind::DynamicDequantize,
                 status::unimplemented,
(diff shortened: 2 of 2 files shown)


ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -206,18 +206,17 @@ status_t sdp_primitive_config_t::initial_check(
                 scale = post_op;
                 post_op = get_post_op(post_op);
             }
+            // Not support select after scale(optional) and mask(optional)
+            // Distill-Bert:[mm1] --> [scale]* --> [mask]* --> [select] --> ...
```
gyhintel, line 210: Should be [mm1] --> [scale]* --> [select] -->  ... ?

ElaineBao: should be after mask, let me fix, thanks

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -207,9 +207,11 @@ status_t sdp_primitive_config_t::initial_check(
                 post_op = get_post_op(post_op);
             }
             // mask
-            if (post_op && post_op->get_kind() == graph::op_kind::Add) {
-                // Mask exists, update post_op and traverse to next op
-                post_op = get_post_op(post_op);
+            if (post_op) {
```
gyhintel, line 210: This condition is always true. Do we need it?

ElaineBao: no, for implicit causal mask, it's not true

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -207,9 +207,11 @@ status_t sdp_primitive_config_t::initial_check(
...
+                if (post_op->get_kind() == graph::op_kind::Add) {
+                    // Mask exists, update post_op and traverse to next op
+                    post_op = get_post_op(post_op);
+                }
                 // Not support select after scale(optional) and mask(optional)
                 // Distill-Bert:[mm1] --> [scale]* --> [mask]* --> [select] --> ...
                 VCHECK_SDP_PRIMITIVE(post_op
```
xiang1guo, line 217: Nit: Seems that we don't need to check `post_op`  in `VCHECK_SDP_PRIMITIVE`. The check here equals to the following condition, but post_op is not nullptr, remove it seems to be more clear:
```
if(!post_op || post_op->get_kind() == graph::op_kind::Select))
```

ElaineBao: post_op could be nullptr for implicit causal mask

gyhintel (approved):

xiang1guo (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

avmanerikar (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2460, opened 2025-01-21
Enable brgemm-based RNN for int8 on avx2

densamoilov:
The gemm api based RNN implementation for int8 on avx2 was accidently disabled. Given that we want to stop using the gemm api it makes sense to enable brgemm as a replacement rather than enabling gemm api back.

This change will be backported to rls-v3.7.


densamoilov:
make test
disable test_device_gpu
enable test_device_cpu

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -4936,7 +4936,15 @@ status_t create_brgemm_matmul_copy_b(
...
             else {
-                assert(one_of(conf->isa, avx2_vnni, avx2_vnni_2));
+                // TODO: jit_avx2_vnni_brgemm_matmul_copy_b_int8_t can handle
+                // avx2 if no compensation is required. Consider enabling it
+                // for avx2 and renaming the kernel (drop "vnni" part).
+                const bool is_comp_required = conf->s8s8_compensation_required
+                        || conf->has_zero_point_a;
+                MAYBE_UNUSED(is_comp_required);
+                assert(one_of(conf->isa, avx2_vnni, avx2_vnni_2, avx2)
+                        && IMPLICATION(conf->isa == avx2, !is_comp_required));
```
mgouicem, line 4946: given that this function returns a status, why not do a proper check instead of an assert?

src/cpu/rnn/rnn_reorders.hpp
```diff
@@ -858,7 +859,10 @@ struct rnn_brgemm_weights_reorder_s8_t : public primitive_t {
         const auto &blocked_d = dst_d;
         const auto &pdims = blocked_d.padded_dims();
 
-        const int o_block = pd()->otag_ == ldgOI64o4i ? 64 : 32;
+        const auto otag = pd()->otag_;
+        const int o_block = otag == ldgOI64o4i               ? 64
+                : utils::one_of(otag, ldgOI32o4i, ldOI32o4i) ? 32
+                                                             : 16;
```
mgouicem, line 865: (nit) why not read inner_blk directly instead?

mgouicem (approved):

tprimak (approved):

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
densamoilov: I thought about that but what threw me off was that we don't return `unimplemented` status from here so I wasn't sure about the policy regarding ukernels and I hoped that you or @dzarukin could clarify that.

dzarukin: It seems it was just a style of doing things - enjoying the assert instead of a status.
Switching to verbose-guided messages with proper exits would be more preferrable, but deserves a standalone change.

densamoilov: Ok, let's do it later.

src/cpu/rnn/rnn_reorders.hpp
densamoilov: Makes sense, changed it.

PR #2475, opened 2025-01-22
xe: conv_v2: add Xe2/Xe3 support

echeresh:
Jira: https://jira.devtools.intel.com/browse/MFDNN-13005

Reusing XeHPC kernels for Xe2/Xe3.

### DIFF
--- a/src/gpu/intel/jit/v2/conv/kernel_desc.cpp
+++ b/src/gpu/intel/jit/v2/conv/kernel_desc.cpp
@@ -339,7 +339,7 @@ void kernel_desc_t::set_defaults() {
     }
 }
 
-bool fit_tag(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
+bool is_compatible(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
         const problem_t &prb, bool exact) {
     auto &desc_tag = kernel_desc.layout_tag(abc);
     auto &prb_tag = prb.layout_tag(abc);
@@ -359,11 +359,26 @@ bool fit_tag(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
     return true;
 }
 
-bool fit_impl(const kernel_desc_t &desc, const problem_t &prb, bool exact) {
+bool is_compatible(const hw_desc_t &hw_desc, const hw_t &hw, bool exact) {
+    if (!exact && hw != hw_desc.hw) {
+        switch (hw_desc.hw) {
+            case ngen::HW::XeHPC:
+                return utils::one_of(
+                        hw.to_ngen(), ngen::HW::Xe2, ngen::HW::Xe3);
+            default: break;
+        }
+    }
+    return hw_desc.hw == hw.to_ngen();
+}
+
+bool is_compatible(
+        const kernel_desc_t &desc, const problem_t &prb, bool exact) {
+    gpu_check(is_compatible(desc.hw_desc, prb.hw(), exact))
+            << "HW does not match";
     gpu_check(prb.prop() == desc.prop) << "Propagation kind does not match";
-    gpu_check(fit_tag(tensor_kind_t::a, desc, prb, exact));
-    gpu_check(fit_tag(tensor_kind_t::b, desc, prb, exact));
-    gpu_check(fit_tag(tensor_kind_t::c, desc, prb, exact));
+    gpu_check(is_compatible(tensor_kind_t::a, desc, prb, exact));
+    gpu_check(is_compatible(tensor_kind_t::b, desc, prb, exact));
+    gpu_check(is_compatible(tensor_kind_t::c, desc, prb, exact));
     gpu_check(prb.is_depthwise() == desc.is_dw)
             << "Mixing depthwise/non-depthwise descriptor and problem";
     if (desc.use_stream_k) {
@@ -401,6 +416,7 @@ void fit_tag_to(
 }
 
 void fit_to_impl(kernel_desc_t &desc, const problem_t &prb) {
+    desc.hw_desc = hw_desc_t(prb.hw().to_ngen());
     fit_tag_to(tensor_kind_t::a, desc, prb);
     fit_tag_to(tensor_kind_t::b, desc, prb);
     fit_tag_to(tensor_kind_t::c, desc, prb);
@@ -415,7 +431,7 @@ void fit_to_impl(kernel_desc_t &desc, const problem_t &prb) {
 }
 
 bool kernel_desc_t::can_fit(const problem_t &prb) const {
-    return fit_impl(*this, prb, /*exact=*/false);
+    return is_compatible(*this, prb, /*exact=*/false);
 }
 
 void kernel_desc_t::fit_to(const problem_t &prb) {
@@ -450,7 +466,7 @@ status_t kernel_desc_t::set_post_ops(const post_ops_t &attr_post_ops,
 }
 
 bool kernel_desc_t::matches(const problem_t &prb) const {
-    return fit_impl(*this, prb, /*exact=*/true);
+    return is_compatible(*this, prb, /*exact=*/true);
 }
 
 std::string kernel_desc_t::cmd_str() const {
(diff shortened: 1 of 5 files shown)


echeresh:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_sum
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -301,11 +301,26 @@ bool fit_tag(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
     return true;
 }
 
-bool fit_impl(const kernel_desc_t &desc, const problem_t &prb, bool exact) {
+bool is_compatible(const hw_desc_t &hw_desc, const hw_t &hw, bool exact) {
+    if (!exact && hw != hw_desc.hw) {
+        switch (hw_desc.hw) {
+            case ngen::HW::XeHPC:
+                return utils::one_of(
+                        hw.to_ngen(), ngen::HW::Xe2, ngen::HW::Xe3);
```
rjoursler, line 309: Could we make compatibility checks against feature requirements (i,e, dpas, 2d send, etc) rather than hardware generations? It generally seems like performant strategies behave well on multiple architectures. By forcing strategies to certain architecture we induces extra tuning work as strategies from platform X is not available for platform Y. What if instead, we can gate strategies by whether there is a performance model associated with the platform, so that a simple remodeling step can be used to take advantage of work on different platforms?

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -301,11 +301,26 @@ bool fit_tag(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
...
+            << "HW does not match";
     ir_check(prb.prop() == desc.prop) << "Propagation kind does not match";
-    ir_check(fit_tag(tensor_kind_t::a, desc, prb, exact));
-    ir_check(fit_tag(tensor_kind_t::b, desc, prb, exact));
-    ir_check(fit_tag(tensor_kind_t::c, desc, prb, exact));
+    ir_check(is_compatible(tensor_kind_t::a, desc, prb, exact));
+    ir_check(is_compatible(tensor_kind_t::b, desc, prb, exact));
+    ir_check(is_compatible(tensor_kind_t::c, desc, prb, exact));
     ir_check(prb.is_depthwise() == desc.is_dw)
             << "Mixing depthwise/non-depthwise descriptor and problem";
```
rjoursler, line 325: Random spot: 2d messages have different alignment requirements on Xe2, but I don't see any check for it. Where is that check happening?

rjoursler (approved):

kealan-barbieri (approved):

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
echeresh: Good point, thanks. I'll open another PR to refactor block 2D requirements to remove them from the kernel descriptor. This way we can tailor a kernel descriptor to the target GPU platform on the fly.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
echeresh: > Could we make compatibility checks against feature requirements (i,e, dpas, 2d send, etc) rather than hardware generations?

I think this step is essentially part of `kernel_desc_t::is_supported()`. Any incompatible/invalid kernels are filtered out during selection.

Gating/reusing strategies is a more challenging thing. I don't like the idea to always require refitting as it's an extra burden:

- There are different variations of hardware like integrated/discrete or with different EU counts or different memory configuration. In theory all such platforms may have performance models tuned per-configuration but it's not feasible in practice so instead we group hardware by some characteristics. Grouping Xe2/Xe3 together can be a viable option in this paradigm if Xe2/Xe3 performance numbers correlate well.
- During early enabling (or preSi enabling) we usually don't have hardware with "full" specs so any performance model is not reliable and subject to change significantly. Moreover with preSi enabling we may not have a way to get the needed amount of benchmark data

At the same mixing performance models built for different platforms does not sound reasonable. Even though we may have good correlation between platforms, they are not calibrated to be comparable. So as a starting point we can follow these rules:

1. If platform A doesn't have any kernels with A-specific performance model than we use the closest platform B to evaluate performance of A-compatible kernels.
    - This is a low-effort approach to support platforms during early enabling or platforms that are very similar to other platforms
2. Once we have at least a single A-specific modeled kernel we look for A-based performance modeled kernels only

Transition between 1 and 2 may be problematic but this approach seems to be a compromise.

rjoursler: > Gating/reusing strategies is a more challenging thing. I don't like the idea to always require refitting as it's an extra burden:

I agree, I am not trying to imply that we shouldn't reuse old models. If anything, I suspect we should allow more reuse (under the assumption that old models roughly correlate, but are less accurate) by having some kind of model priority list and an inter-model comparison function, something like
```
if(model_A_estimate(strategy1) * 1.10 < model_B_estimate(strategy2)) {
    return strategy2;
} else {
    return strategy1;
}
```
The idea here is to expose strategy issues during development. By using model B more aggressively, we should get some testing of new strategies on other platforms which should prevent surprises when we eventually perform a refit on platform A.

echeresh: To some extent it's still refitting, but in a simpler way, via scaling factors. Theoretical-based factors (e.g. like `B_EU_count / A_EU_count`) may be not reliable as different HW characteristics contribute to the final performance like EU count, cache bandwidth/latency, etc.

But I agree, this approach may work, we need to check how good/accurate it may be in practice.

rjoursler: > But I agree, this approach may work, we need to check how good/accurate it may be in practice.

And to be clear, I only intend this as an interim measure for testing purposes without generating model churn.  The intention would be to refit before oneDNN releases so that these scales are unused (unless this magically is super effective). If this proposal is inaccurate, other testing methods (or even no testing) could be reasonable as well.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
echeresh: Addressed in https://github.com/oneapi-src/oneDNN/pull/2545

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

PR #2479, opened 2025-01-22
gpu: nvidia: ip: respect acc_mode for sum post-op

sgeor255:
# Description

Currently the inner product error threshold in benchdnn is set to 0. In some cases for large shapes on nvidia backend there are some precision issues (e.g. the cases reported in MFDNN-12610). This PR adjusts the error threshold so that such cases are not reported as failures.

Fixes MFDNN-12610.

### DIFF
--- a/src/gpu/nvidia/README.md
+++ b/src/gpu/nvidia/README.md
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
 The inner product primitives is an implementation of matrix multiplication plus
 bias activation. There are two implementation of inner product in cuDNN backend.
 
+With `sum` post-op, the accumulation mode attribute affects the behavior as
+follows:
+- `relaxed`: Uses GEMM’s beta parameter for a fused and optimized sum post-op
+  but may reduce output precision for large `f16` inputs.
+- `strict` (default): Converts GEMM output to `f32`, performs `sum` post-op as a
+  separate operation, then converts it back to the original type. This attribute
+  provides better output precision but reduced performance.
+
 #### Using GEMM
 
 The default backend for inner product is the gemm backend using `cublasGemmEx`
--- a/src/gpu/nvidia/cudnn_gemm_inner_product_impl.hpp
+++ b/src/gpu/nvidia/cudnn_gemm_inner_product_impl.hpp
@@ -77,8 +77,8 @@ struct cudnn_gemm_inner_product_fwd_impl_t
     bool need_reorder_;
 
     virtual status_t init(impl::engine_t *, inner_product_pd_t *pd,
-            bool with_relu, bool with_eltwise, bool with_sum,
-            bool need_reorder) override {
+            bool with_relu, bool with_eltwise, bool with_sum, bool need_reorder,
+            bool use_f32_sum) override {
         need_reorder_ = need_reorder;
         // GEMM is column major, here the data is row major.
         // By switching the weight and source we convert the row major to
@@ -121,8 +121,10 @@ struct cudnn_gemm_inner_product_fwd_impl_t
         use_acc_dst_ = ((pd->dst_md()->data_type == data_type::s8)
                 || (with_bias_
                         && pd->weights_md(1)->data_type
-                                != pd->dst_md()->data_type));
+                                != pd->dst_md()->data_type)
+                || use_f32_sum);
         with_sum_ = with_sum;
+        with_f32_sum_ = use_f32_sum;
         // scaling factor to add the previous destination value to the current
         // computation. This is equivalent of
         sum_scale_ = sum_scale(pd);
@@ -154,12 +156,23 @@ struct cudnn_gemm_inner_product_fwd_impl_t
 
         if (with_bias_) {
             CHECK(convert_data_type(pd->weights_md(1), &data_types_[io::bia]));
+
             // format is always nchw
             set_bias_dims(CUDNN_TENSOR_NCHW, ndims_, pd->OC());
 
             CHECK(create_and_set_tensor_descriptor(&tensor_descs_[io::bia],
                     data_types_[io::bia], ndims_, dims_[io::bia],
                     strides_[io::bia]));
+
+            if (with_f32_sum_) {
+                pd->scratchpad_registry().registrar().book(
+                        memory_tracking::names::key_iprod_bias_bf16_convert_wsp,
+                        memory_desc_wrapper(pd->weights_md(1)).nelems(),
+                        types::data_type_size(data_type::f32));
+                CHECK(create_and_set_tensor_descriptor(&bias_f32_desc_,
+                        CUDNN_DATA_FLOAT, ndims_, dims_[io::bia],
+                        strides_[io::bia]));
+            }
         }
         if (use_acc_dst_) {
             pd->scratchpad_registry().registrar().book(
@@ -178,10 +191,10 @@ struct cudnn_gemm_inner_product_fwd_impl_t
 
     void execute(cudnnHandle_t cudnn_handle, cublasHandle_t cublas_handle,
             const std::vector<void *> &args) const override {
-        assert(args.size() == 9);
+        assert(args.size() == 10);
         auto x = args[0], w = args[1], b = args[2], y = args[3],
              workspace = args[4], src_scale = args[6], wei_scale = args[7],
-             dst_scale = args[8];
+             dst_scale = args[8], bias_f32 = args[9];
         auto w_arg = w;
         if (need_reorder_) {
             void *transformed_w = args[5];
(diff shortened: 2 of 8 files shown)


tests/benchdnn/ip/ip.cpp
```diff
@@ -278,7 +278,16 @@ void skip_invalid_prb(const prb_t *prb, res_t *res) {}
...
+    // The nvidia implementation has precision issues in some cases
+    // for large problems with post-op sum
+    if (is_nvidia_gpu()
+            && prb->attr.post_ops.find(attr_t::post_ops_t::kind_t::SUM) != -1
+            && prb->dst_dt() == dnnl_f16) {
+        const float trh = epsilon_dt(prb->dt[2]);
+        cmp.set_threshold(trh);
+    } else {
+        cmp.set_threshold(0.f);
+    }
```
mgouicem, line 290: Do you know why this difference ? Is sum post-op applied over f32 intermediate value or over f16 values for NV backend?

tests/benchdnn/ip/ip.cpp
```diff
@@ -278,7 +278,16 @@ void skip_invalid_prb(const prb_t *prb, res_t *res) {}
 
 void setup_cmp(compare::compare_t &cmp, const prb_t *prb, data_kind_t kind,
         const args_t &ref_args) {
-    cmp.set_threshold(0.f);
+    // The nvidia implementation has precision issues in some cases
```
Rbiessy, line 281: I wouldn't call it precision issues, wasn't our conclusion that this is differences in rounding modes that we can't change in cuDNN?

AD2605 (approved):

tests/benchdnn/ip/ip.cpp
dzarukin: I'd say this change can fly in only in case sum post-op is done through a native cuDNN fusion (single call) with f16 accumulation internally, otherwise, the issue is likely inside the implementation that doesn't convert the output to f32 and accumulate pieces in f32.

dzarukin: @sgeor255, thanks for looking into implementation details, that's a good start.
I suggest to conduct a little experiment - make `cublasGemmEx` sum-less and return f32 and append sum post-op (with data copied to a dedicated buffer) with f32 as well (through the `cudnnAddTensor` call) to simulate the proper behavior.
If this experiment goes fine, it would prove that there's something wrong indeed with cuda libraries, and I'll re-iterate on that check again. Probably the same one should exist (or appear) for matmul as well.

dzarukin: When changing data addresses the issue it always means rounding/accumulation mechanics stands on its way. Smaller ranges usually lead to situations when final numbers remain exact and conversion to f16/f32 and back don't change the number and the check passes.

When exp number if x.5, in the reality, it can be x.5002, which would be rounded towards `x + 1`, while the library might have `x.4998` which would be rounded towards `x`. That library result can be a product of different effects (accumulation in lower data types, inexact division/multiplication, intermediate roundings, etc.) which can't be examined with cuda.

sgeor255: @dzarukin thanks for the suggestion, I tested doing the sum post-op separately with `cudnnAddTensor` and the test is passing for all sum scale values I tested.

dzarukin: @sgeor255 thanks for checking it!

Then it likely means non-zero `beta` is not aligned with proper f32 conversions.
I would suggest to keep a "split" implementation version as a gold one and use `beta`-fusion if `acc-mode` is specified to be `relaxed`. `acc-mode` isn't used much inside the library yet, but it can become an essential requirement for cuda/cuDNN/cuBLAS functionalities given observed behavior.

tests/benchdnn/ip/ip.cpp
sgeor255: I updated the comment in the latest commit.

Rbiessy: I don't see what the difference is? I'd say something like "different precision guarantees"

sgeor255: Looks like I updated the check, not the comment. Updated the comment as suggested now, thanks.

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
 The inner product primitives is an implementation of matrix multiplication plus
 bias activation. There are two implementation of inner product in cuDNN backend.
 
+With `sum` post-op, the accumulation mode attribute affects behaviour as
```
Rbiessy, line 218:
```suggestion
With `sum` post-op, the accumulation mode attribute affects behavior as
```
American English is preferred

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
...
+follows:
+- `relaxed`: Uses GEMM’s beta parameter for a fused, optimised sum post-op but
```
Rbiessy, line 220:
```suggestion
- `relaxed`: Uses GEMM’s beta parameter for a fused, optimized sum post-op but
```
American English

Rbiessy (approved):

src/gpu/nvidia/README.md
sgeor255: Done

Rbiessy (approved):

sgeor255:
I think this PR still needs a review from @oneapi-src/onednn-doc

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
 The inner product primitives is an implementation of matrix multiplication plus
 bias activation. There are two implementation of inner product in cuDNN backend.
 
+With `sum` post-op, the accumulation mode attribute affects behavior as
```
ranukund, line 218:
```suggestion
With `sum` post-op, the accumulation mode attribute affects the behavior as
```

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
...
+follows:
+- `relaxed`: Uses GEMM’s beta parameter for a fused, optimized sum post-op but
```
ranukund, line 220:
```suggestion
- `relaxed`: Uses GEMM’s beta parameter for a fused and optimized `sum` post-op but
```

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
...
+  may reduce output precision for large `f16` inputs.
+- `strict` (default): Converts GEMM output to `f32`, performs sum as a separate
```
ranukund, line 222:
```suggestion
- `strict` (default): Converts GEMM output to `f32`, performs `sum` post-op as a separate
```
...Perform sum post-op?

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
...
+  operation, then converts it back to the original type. This is more precise
```
ranukund, line 223:
```suggestion
  operation, then converts it back to the original type. This attribute provides better output precision
```

src/gpu/nvidia/README.md
```diff
@@ -215,6 +215,14 @@ limitations when using Nvidia backend for eltwise primitive:
...
+  but less performant.
```
ranukund, line 224:
```suggestion
  but reduced performance.
```

ranukund (commented):
Edits suggested, please incorporate as you see fit! Thanks!

sgeor255:
Thanks @ranukund , I added the suggested changes.

ranukund (approved):
Thank you!

src/gpu/nvidia/cudnn_gemm_inner_product_impl.hpp
```diff
@@ -271,6 +296,10 @@ struct cudnn_gemm_inner_product_fwd_impl_t
 
         return status::success;
     }
+
+    ~cudnn_gemm_inner_product_fwd_impl_t() {
+        if (with_f32_sum_) { cudnnDestroyTensorDescriptor(bias_f32_desc_); }
```
dzarukin, line 301: I can't find where `tensor_descs_` gets destroyed but given there are other additional tensors created but only a new one gets destroyed here is suspicious. My naive understanding is either this destroy call should move to the place where it's already done (I failed to identify it), or other tensor should be destroyed here as well. Could you check it for me, please, which option is better?

dzarukin (commented):
Thanks for the addressing the feedback. One relatively minor comment and it's good to go.

sgeor255:
It looks like the precommit check failures are in files unrelated to this PR.

dzarukin (approved):

PR #2480, opened 2025-01-22
ci: add oneDNN cache

Ryo-not-rio:
context:
We need to build oneDNN main for our regression tests, this change adds a pipeline to create a cache for oneDNN main so we do not have to rebuild the same patchset every time.


vpirogov:
Closing as stale.

PR #2490, opened 2025-01-23
cpu: x64: disable small shapes for brgemm matmul reorder

xuxinzen:
Fixes MFDNN-13029

Dispatch small shapes to jit reorder for performance purpose.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2024 Intel Corporation
+* Copyright 2022-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test
enable benchdnn_nightly
disable test_device_gpu

src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
```diff
@@ -393,38 +393,56 @@ status_t brgemm_matmul_copy_reorder_t::execute_body(
...
-                                    * kernel_conf.tr_a_dt_sz
-                            + dst_d.offset0() * kernel_conf.tr_a_dt_sz);
-                    (*a_kernel_)(&ker_exec_ctx);
-                });
-
+        const auto work_amount = kernel_conf.batch * k_chunks * m_chunks;
+        const auto cache_size
+                = kernel_conf.a_dt_sz * kernel_conf.M_blk * kernel_conf.K_blk;
+        // // The threshold is empirical value
+        const int nthr = work_amount * cache_size
```
densamoilov, line 403: Why do we base the heuristic on the L2 size here? Does it really depend on it?

xuxinzen:
make test
enable benchdnn_nightly
disable test_device_gpu

src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
xuxinzen: For my understanding, the answer is yes. The basic idea in this PR is to find a heuristic value that fits the cache into L2 and then we can use single thread to improve performance. A machine with different L1/L2 cache size could be perform different.
This changes did not work on user's side, performance is better but still has gap with jit reorder. So I decided to dispatch such shape to jit reorder for now.

densamoilov (approved):

tprimak (approved):

PR #2501, opened 2025-01-24
Arch coverity fixes

mgouicem:


spalicki (approved):

third_party/ngen/ngen_level_zero.hpp
```diff
@@ -63,7 +63,14 @@ void *find_ze_symbol(const char *symbol) {
     void *f = reinterpret_cast<void *>(GetProcAddress(handle, symbol));
 #endif
 
-    if (!f) throw level_zero_error{ZE_RESULT_ERROR_UNINITIALIZED};
+    if (!f) {
+#if defined (__linux__)
+	dlclose(handle);
+#elif defined(_WIN32)
+	FreeLibrary(handle);
+#endif
```
dzarukin, line 71: I failed to find any place where `dlclose` is called in a proper situation.
It sounds like hadnle should be wrapped into a `unique_ptr` with a custom deleter which would `dlclose` the handle upon success or failure.

avmanerikar (approved):

atkassen (approved):

third_party/ngen/ngen_level_zero.hpp
mgouicem: Currently we just let the library open and never call dlclose. If we use a smart pointer, we would have to keep it alive while we want the library to stay open (otherwise, we incur the risk for the library to be effectively loaded/unloaded multiple times over the lifetime of the program).
I guess we have three options:
- close level_zero library only when we fail to find symbols (this patch),
- never close level_zero, even in the case where we fail to find a symbol (same as today).
- or explicitly load/unload the library through shared pointer and tie its lifetime to oneDNN objects.

dzarukin: Thanks for the explanation.

petercad: The reason for the dlopen "leak" is that nGEN doesn't have any stateful objects that can encapsulate the scope where it needs to call upon L0. Actually, ze_loader must be loaded prior to nGEN in any case (the caller is responsible for creating L0 device, etc. objects before it can call nGEN).

Anyway, if we truly wanted to avoid reference count leaks on ze_loader, we would need to dlopen/dlclose around every call to L0.

mgouicem: let's drop that commit, as it brings little value anyways.

PR #2506, opened 2025-01-24
gpu:amd: Update bnorm lambda capture

ShanoToni:
# Description

When compiling oneDNN with C++20, the current bnorm operator causes errors due to the lambda capturing `this`.
In C++20 implicit capture of this is an error, while in C++17 explicit capture will flag as a warning.

PR adds use of the already existing compiler workaround for the lambda capture WA_THIS_COPY_CAPTURE

# Checklist

## General


Rbiessy (approved):
Thanks LGTM. Correct me if I'm wrong @ShanoToni, the PR is just adding 2 `WA_THIS_COPY_CAPTURE` as suggested and the rest of the changes are just due to formatting.

PR #2509, opened 2025-01-24
x64: utils: jit_io_helper: fix xf16 store from Xmm (fixes MFDNN-12635)

tczeszun:
This is a short fix for [MFDNN-12635](https://jira.devtools.intel.com/browse/MFDNN-12635).

Function `jit_io_helper_t<Vmm>::store_bf16` doesn't support `Xmm`, that's why it crashed when using `bf16` with blocked format with channels blocked by 4.
I changed a way it works to use `jit_io_helper_t<Vmm>::store_byte_by_byte` function instead. 
Though, I have other idea that maybe using `store_bf16` could work if we introduced another `Opmask` to filter out the lower half of `Xmm`. It would require some changes to `jit_io_helper_t` class, so I prefer to not submit it now as a bugfix, but after some time (as it requires some analysis), that's why I left a `TODO` comment.

Also I'm adding reorder regression input to benchdnn, as it was likely not covered on CPU side.

### DIFF
--- a/src/cpu/x64/utils/jit_io_helper.cpp
+++ b/src/cpu/x64/utils/jit_io_helper.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -277,12 +277,13 @@ void jit_io_helper_t<Vmm>::prepare_i8_data_to_store(const Vmm &i8_vmm) {
 
 template <typename Vmm>
 void jit_io_helper_t<Vmm>::prepare_xf16_data_to_store(const Vmm &vmm) {
-    assert(!is_superset(isa_, avx512_core));
     const auto &cvt_lower_vmm =
             typename vreg_traits<Vmm>::Vmm_lower_t(vmm.getIdx());
 
     if (data_type_ == data_type::bf16)
-        host_->vcvtneps2bf16(cvt_lower_vmm, vmm, Xbyak::VexEncoding);
+        host_->vcvtneps2bf16(cvt_lower_vmm, vmm,
+                mayiuse(avx512_core) ? Xbyak::EvexEncoding
+                                     : Xbyak::VexEncoding);
     else
         host_->uni_vcvtps2phx(cvt_lower_vmm, vmm);
 }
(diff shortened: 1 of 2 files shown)


tczeszun:
make test
enable benchdnn_nightly
disable test_device_gpu

src/cpu/x64/utils/jit_io_helper.cpp
```diff
@@ -277,12 +277,13 @@ void jit_io_helper_t<Vmm>::prepare_i8_data_to_store(const Vmm &i8_vmm) {
...
 void jit_io_helper_t<Vmm>::prepare_xf16_data_to_store(const Vmm &vmm) {
-    assert(!is_superset(isa_, avx512_core));
     const auto &cvt_lower_vmm =
             typename vreg_traits<Vmm>::Vmm_lower_t(vmm.getIdx());
 
     if (data_type_ == data_type::bf16)
-        host_->vcvtneps2bf16(cvt_lower_vmm, vmm, Xbyak::VexEncoding);
+        host_->vcvtneps2bf16(cvt_lower_vmm, vmm,
+                mayiuse(avx512_core) ? Xbyak::EvexEncoding
+                                     : Xbyak::VexEncoding);
```
dzarukin, line 286: How come an instruction that supports only Evex encoding has an option for Vex encoding?

tests/benchdnn/inputs/reorder/harness_reorder_regression
```diff
@@ -11,3 +11,10 @@
 --stag=abdc --dtag=abcd
 --attr-zero-points=src0:common:1
 1x32x128x33
+
+# Test bf16 with aBcde4b format
+--reset
+--skip-impl=ref,simple # ! test jit version only
```
dzarukin, line 17:
```suggestion
--skip-impl=simple #skip non-jit version
```

dzarukin (approved):

src/cpu/x64/utils/jit_io_helper.cpp
tczeszun: Good question, but according to xbyak it has parameter for encoding:
https://github.com/oneapi-src/oneDNN/blob/main/third_party/xbyak/xbyak_mnemonic.h#L1260
And I see it's not the only place where Vex encoding is used, for example pooling:
https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_uni_pool_kernel.cpp#L931

dzarukin: I think we should limit that instruction to Evex only, regardless if xbyak provides an opportunity for encoding argument.

tczeszun: I'll do separate PR for this, as it occurs in many files.

xuxinzen (approved):

tczeszun:
make test
enable benchdnn_nightly
disable test_device_gpu

PR #2512, opened 2025-01-24
gpu: quantization refactoring prerequisites

dzarukin:
This PR updates gemm_with_post_ops implementation with a new way of loading scales as they are going to be affected after refactor. The main issue is that `data_type::undef` becomes a valid data type for scales, meaning, they were not specified, which would trigger a bunch of issues in the existing implementation.

This PR guards from that situation.

Thanks to @rjoursler for a patch to deal with undefined data types.

### DIFF
--- a/src/gpu/intel/ocl/gemm/gemm_with_post_ops.cl
+++ b/src/gpu/intel/ocl/gemm/gemm_with_post_ops.cl
@@ -62,7 +62,11 @@ __kernel void gemm_post_ops(__global SRC_DATA_T *src,
         APPLY_POST_OPS_SERIAL(accumulator, POST_OP_DATA_T, sum_src,
                 POST_OP_DATA_T, d0, 1, d1, 1, d2, 1, d3, 1, 0, 1, 0, 1);
 
-        if (C_SCALES) accumulator /= DST_SCALES_TO_REF(c_scales[0]);
+        if (C_SCALES) {
+            POST_OP_DATA_T c_scale = 1;
+            load(&c_scale, c_scales);
+            accumulator /= c_scale;
+        }
         if (DST_ZERO_POINT) accumulator += dst_zp[0];
     }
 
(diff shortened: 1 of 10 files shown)


dzarukin:
make test
disable test_device_cpu

src/gpu/intel/ocl/gemm/gemm_with_post_ops.cl
```diff
@@ -62,7 +62,11 @@ __kernel void gemm_post_ops(__global SRC_DATA_T *src,
         APPLY_POST_OPS_SERIAL(accumulator, POST_OP_DATA_T, sum_src,
                 POST_OP_DATA_T, d0, 1, d1, 1, d2, 1, d3, 1, 0, 1, 0, 1);
 
-        if (C_SCALES) accumulator /= DST_SCALES_TO_REF(c_scales[0]);
+        float c_scale = 1;
+        if (C_SCALES) {
+            load(&c_scale, c_scales);
```
rjoursler, line 67: To ensure there are no type conversion issues when `c_scale` is used with `accumulator`:

```Suggestion
        if (C_SCALES) {
            POST_OP_DATA_T c_scale = 1;
            load(&c_scale, c_scales);
```

src/gpu/intel/ocl/ocl_math_utils.h
```diff
@@ -81,6 +81,10 @@ int rnd_down(int a, unsigned int b) {
 #define MATH_UTILS_DECLARE_BF16 1
 #endif
 
+#if DST_SCALES_DT_E8M0
```
rjoursler, line 84: Is there a reason we do not need to be concerned with the other buffers data types similar to the above declarations?

rjoursler (approved):

echeresh (approved):

src/gpu/intel/ocl/gemm/gemm_with_post_ops.cl
dzarukin: Thanks, updated.

src/gpu/intel/ocl/ocl_math_utils.h
dzarukin: e8m0 is designed for scales only. It seems that so far it is supported for dst only and there's no src or weights counterparts (I found only commented cases in benchdnn matmul_fp4 input list).
To be extended as support grows, I guess...

PR #2516, opened 2025-01-27
Improve diagnostics for unimplemented RNN cases

densamoilov:
Implements MFDNN-13023.

This PR improves diagnostics for unimplemented cases when domain specific tags are used. The idea is to normalize input tags prior to validating them to avoid skipping unimplemented cases that are supposed to implemented. This will help us to catch accidently disabled implementations or reduced coverage (e.g. MFDNN-12818).

For example:
```bash
# If unimplemented then the status is SKIPPED (not an error) - without this PR
 --rnn --cfg=u8u8u8u8 --tag=tnc:any:tnc --alg=VANILLA_GRU --activation=UNDEF l1t1mb3sic16slc16dhc16dic16n"uniform"

# If unimplemented then the status is UNIMPLEMENTED (error) - with this PR
 --rnn --cfg=u8u8u8u8 --tag=tnc:any:tnc --alg=VANILLA_GRU --activation=UNDEF l1t1mb3sic16slc16dhc16dic16n"uniform"
```

### DIFF
--- a/tests/benchdnn/rnn/rnn.cpp
+++ b/tests/benchdnn/rnn/rnn.cpp
@@ -565,8 +565,8 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
     dnnl_dims_t dst_layer_dims = {prb.n_iter, prb.mb, prb.dlc(PRIMITIVE)};
 
     dnnl_dims_t src_layer_dims = {prb.n_iter, prb.mb, prb.slc};
-    auto src_layer_d = dnn_mem_t::init_md(
-            3, src_layer_dims, prb.cfg[SRC_LAYER].dt, prb.tag[0]);
+    auto src_layer_d = dnn_mem_t::init_md(prb.ndims(SRC_LAYER), src_layer_dims,
+            prb.cfg[SRC_LAYER].dt, prb.tag[0]);
     if (prb.tag[0] != tag::any) {
         dims_t src_layer_strides(query_md_ndims(src_layer_d));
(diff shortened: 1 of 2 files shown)


densamoilov:
make test

tests/benchdnn/rnn/rnn.cpp
```diff
@@ -825,13 +836,24 @@ void skip_unimplemented_prb(const prb_t *prb_, res_t *res) {
...
+
+        if (is_cpu()) {
+            const auto src_tag
+                    = normalize_tag(prb.tag[0], prb.ndims(SRC_LAYER));
+            const auto wei_tag
+                    = normalize_tag(prb.tag[1], prb.ndims(WEIGHTS_LAYER));
+            const auto dst_tag
+                    = normalize_tag(prb.tag[2], prb.ndims(DST_LAYER));
+
+            const bool tags_not_ok = src_tag != "abc" || wei_tag != tag::any
```
dzarukin, line 848: Nit: I'd invert the condition for it to be positive (`tags_ok`), because it's usually easier to append the valid configuration through a dedicated variable than to fit it into negation condition.

dzarukin (approved):

inteldimitrius (approved):

tests/benchdnn/rnn/rnn.cpp
densamoilov: The reason why is used `not` is because there are already `is_f16_not_ok` and `is_acl_f16_not_ok` above so I just followed the style. I don't have a strong preference and can drop `not` if you insist.

PR #2523, opened 2025-01-27
x64: conv: addressing coverity issues

inteldimitrius:
This PR fixes coverity issues with `Severity == Major` in Convolution primitive, in particular:
- [x]  `6980313` Improper use of negative value in file: `/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp:1365`. 
`brg_idx` could be equal `-1`, which causes issues in `call_brgemm(..., brg_idx, ...)`. 
Solution: add  an assertion in `call_brgemm()`.
- [x]  `6980310` Improper use of negative value in file: `/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp:1610`. 
The same as in `6980313`, but for tail computations.

PS. Other coverity hits in convolution will be marked as false positive.

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp
@@ -1258,6 +1258,10 @@ void brgemm_convolution_bwd_strided_t<isa>::ker_base(
     const auto call_brgemm = [&](int iw, int brg_idx, int oc_block_s,
                                      int n_oc_blocks, size_t comp_ker_offs,
                                      bool do_postops, bool do_only_comp) {
+        if (brg_idx < 0) {
+            assert(!"Requested brgemm kernel was not created.");
+            return;
+        }
         const auto kh_ee = kh_e;
         int k_sum = 0;
         int32_t *src_zp = jcp.src_zero_point
@@ -1513,6 +1517,10 @@ void brgemm_convolution_bwd_strided_t<isa>::ker_trans(
          is_first_call_postops_state_changed = false;
     const auto call_brgemm = [&](int brg_idx, int oc_block_s, int n_oc_blocks,
                                      size_t comp_ker_offs, bool do_postops) {
+        if (brg_idx < 0) {
+            assert(!"Requested brgemm kernel was not created.");
+            return;
+        }
         const auto kh_ee = kh_e;
         const auto kw_e = kw_f;
         const auto pbuf_base = inp_buffer;


src/cpu/x64/jit_avx512_core_amx_conv_utils.hpp
```diff
@@ -283,6 +283,7 @@ inline void execute_backward_convolution_body(const exec_ctx_t &ctx,
                 p.dst_scale = &dst_scales[0];
                 p.acc_s32 = wsp + ithr * jcp.wsp_buffer_size;
                 p.last_h = (ih + ih_step <= ih_e);
+                assert(iwb >= 0);
```
dzarukin, line 286: Assert can't fix the issue for the user, if the issue is there.
What is the scenario when `iwb` can become negative?

inteldimitrius:
make test

src/cpu/x64/jit_avx512_core_amx_convolution.cpp
```diff
@@ -387,7 +387,7 @@ jit_avx512_core_amx_convolution_fwd_t::execute_forward_reduced_lowering(
                     limit_idx++;
                 p.ohb = limit_idx;
                 p.last_h = (oh + oh_step <= oh_e);
-                const int zp_owb = nstl::min(jcp.l_pad_blk, owb)
+                const size_t zp_owb = nstl::min(jcp.l_pad_blk, owb)
```
dzarukin, line 390: If `nstl::min(jcp.l_pad_blk, owb)` will produce `int`, this change won't resolve an overflow.

src/cpu/x64/jit_avx512_core_amx_convolution.cpp
```diff
@@ -781,7 +781,7 @@ status_t jit_avx512_core_amx_convolution_fwd_t::execute_forward(
                 assert(limit_idx < 6);
                 p.ohb = limit_idx;
                 p.last_h = (oh + oh_step <= oh_e);
-                const int zp_owb = nstl::min(jcp.l_pad_blk, owb)
+                const size_t zp_owb = nstl::min(jcp.l_pad_blk, owb)
```
dzarukin, line 784: Ditto.

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -834,8 +834,11 @@ inline int brgemm_convolution_fwd_t<isa>::get_comp_offset(const int g,
     const auto comp_idx
             = get_comp_ker_idx(kd_b, kd_e, kh_b, kh_e, kw_b, kw_e, cur_comp_oh);
     assert(IMPLICATION(jcp.req_cal_comp_pad, comp_idx >= 0));
-    return jcp.req_cal_comp_pad ? g * comp_ocb_sz + ocb * comp_ker_sz
-                    + comp_idx * comp_kw_sz + ow * comp_ow_sz
+    const int req_cal_comp_offset = g * static_cast<int>(comp_ocb_sz)
+            + ocb * static_cast<int>(comp_ker_sz)
+            + comp_idx * static_cast<int>(comp_kw_sz)
+            + ow * static_cast<int>(comp_ow_sz);
```
dzarukin, line 840: What does this suppose to address?

src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp
```diff
@@ -1513,6 +1515,7 @@ void brgemm_convolution_bwd_strided_t<isa>::ker_trans(
          is_first_call_postops_state_changed = false;
     const auto call_brgemm = [&](int brg_idx, int oc_block_s, int n_oc_blocks,
                                      size_t comp_ker_offs, bool do_postops) {
+        assert(brg_idx >= 0);
```
dzarukin, line 1518: Again, `assert` doesn't fix anything.
This lambda should start returning status. If `brg_idx` is negative, it should report `status::runtime_error`.
And it should start checking the status of `call_brgemm_kernel` if it has one.

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -834,8 +834,11 @@ inline int brgemm_convolution_fwd_t<isa>::get_comp_offset(const int g,
...
+    return jcp.req_cal_comp_pad ? req_cal_comp_offset
```
densamoilov, line 841: I'm a little confused by the description  that you cast `int64_t` to `int` here to avoid int overflow? Are we sure that `req_cal_comp_offset` cannot exceed INT32_MAX?

src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp
inteldimitrius: Updated. Now it's aligned with the same lambda in brgemm: https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_brgemm_conv.cpp#L2048

src/cpu/x64/jit_brgemm_conv.cpp
inteldimitrius: Sorry, this issue look false positive. I removed from PR.

inteldimitrius:
Hi @dzarukin @densamoilov, sorry I did nonsense. The most of issues are false positives, I removed it from PR.

inteldimitrius:
make test

dzarukin (approved):

PR #2528, opened 2025-01-27
cpu: fixes some coverity issues(overflow, dereference null pointer issues)

xuxinzen:
Partially fixes MFDNN-12964

Fixes potential overflow/underflow, and dereference null pointer issue

### DIFF
--- a/src/cpu/reorder/simple_sparse_reorder.hpp
+++ b/src/cpu/reorder/simple_sparse_reorder.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2023-2024 Intel Corporation
+* Copyright 2023-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -88,7 +88,8 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
 
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
-        const auto nelems = output_d.nelems(true);
+        if (output_d.nelems(true)) return 0;
+        const auto nelems = static_cast<size_t>(output_d.nelems(true));
         const auto tmp_output_sz = nelems * output_d.data_type_size();
         const auto nnz_per_blocks_sz
                 = nelems / output_d.blk_size() * sizeof(dim_t);
@@ -218,6 +219,8 @@ struct simple_sparse_reorder_t : public primitive_t {
             CHECK(reorder_primitive_desc_create(
                     reorder_pd_, engine, src_md(), &converted_dst_md, attr()));
 
+            const memory_desc_wrapper dst_d(dst_md());
+            if (dst_d.nelems(true) < 0) return status::unimplemented;
             const size_t scratchpad_sz_ = simple_sparse_reorder_impl<
                     SIMPLE_SPARSE_REORDER_TEMPL_CALL>::
                     get_scratchpad_size(src_md(), dst_md());
--- a/src/cpu/x64/lrn/jit_avx512_common_lrn_bwd_base.cpp
+++ b/src/cpu/x64/lrn/jit_avx512_common_lrn_bwd_base.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2024 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


xuxinzen:
make test
disable device_gpu

xuxinzen:
make test
disable device_gpu

src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -88,8 +88,8 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
 
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
-        const auto nelems = output_d.nelems(true);
-        const auto tmp_output_sz = nelems * output_d.data_type_size();
+        const size_t nelems = static_cast<size_t>(output_d.nelems(true));
```
tczeszun, line 91: I think it would compile but it doesn't make sense to me, casting negative value to unsigned will give us some weird number. I think it's better to mark the issue as intentional, or completely change `get_scratchpad_size` return type to be able to return negatives as error codes.

src/cpu/x64/jit_uni_batch_normalization.cpp
```diff
@@ -2241,7 +2241,7 @@ struct driver_t : public c_compatible {
 
             if (C_blks_thr == 0 || N_thr == 0) continue;
 
-            size_t coff_base = global_C_blk_s * simd_w;
+            size_t coff_base = static_cast<size_t>(global_C_blk_s) * simd_w;
```
tczeszun, line 2244: same problem, function is wrongly designed to return void while `global_C_blk_s` requires exception handling
I think it's better to mark it as intentional and leave comment that it's not expected that `global_C_blk_s` ever becomes negative

tczeszun: Alternatively you could silently return when `global_C_blk_s < 0` which will stop kernel from execution instead of getting segfaults. Not sure if silent returns won't make coverity issue again.

src/cpu/x64/lrn/jit_avx512_common_lrn_bwd_base.cpp
```diff
@@ -76,6 +76,12 @@ void jit_avx512_common_lrn_kernel_bwd_t<f16>::store_data(
 template <>
 void jit_avx512_common_lrn_kernel_bwd_t<bf16>::store_data(
         bool nt, const Address addr, Zmm zr) {
+    const bool is_bf16_supported
```
tczeszun, line 79: maybe I missed something, but I don't see this issue in coverity

src/cpu/x64/lrn/jit_avx512_common_lrn_fwd_base.cpp
```diff
@@ -117,6 +117,13 @@ void jit_avx512_common_lrn_kernel_fwd_t<f16>::store_data(
 template <>
 void jit_avx512_common_lrn_kernel_fwd_t<bf16>::store_data(
         const Address addr, Zmm zr, Ymm yr) {
+    const bool is_bf16_supported = mayiuse(avx512_core_bf16)
```
tczeszun, line 120: also don't see this in coverity

src/cpu/x64/lrn/jit_avx512_common_lrn_bwd_base.cpp
xuxinzen: The CID is 6925547. `compute_loop()` dereferences null `this->bf16_emu_` which in `store_data()`.

src/cpu/x64/lrn/jit_avx512_common_lrn_fwd_base.cpp
xuxinzen: The CID is 3674949, the cause is the same as above.

src/cpu/x64/jit_uni_batch_normalization.cpp
xuxinzen: I prefer to mark it as intentional or false positive. When `global_C_blk_s < 0`, it means `C_blk_s == -1 && C_blk_e == -1 `(from [thread_balance()](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_uni_batch_normalization.cpp#L2170)), and then `C_blks_thr == 0`. It won't reach the code here.

src/cpu/reorder/simple_sparse_reorder.hpp
xuxinzen: Thanks, updated.

src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -218,6 +219,7 @@ struct simple_sparse_reorder_t : public primitive_t {
             CHECK(reorder_primitive_desc_create(
                     reorder_pd_, engine, src_md(), &converted_dst_md, attr()));
 
+            if (output_d.nelems(true) < 0) return status::unimplemented;
```
tczeszun, line 222: I think it won't compile, since I don't see `output_d` definition here

tczeszun (approved):

xuxinzen:
make test
disable device_gpu

src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -218,6 +219,8 @@ struct simple_sparse_reorder_t : public primitive_t {
             CHECK(reorder_primitive_desc_create(
                     reorder_pd_, engine, src_md(), &converted_dst_md, attr()));
 
+            const memory_desc_wrapper dst_d(dst_md());
+            if (dst_d.nelems(true) < 0) return status::unimplemented;
```
tczeszun, line 223: nit: maybe you could replace it with verbose check

PR #2529, opened 2025-01-27
cpu: matmul: correct ldc for gemm matmul

xuxinzen:
Fixes MFDNN-13097

### DIFF
--- a/src/cpu/matmul/gemm_x8s8s32x_matmul.cpp
+++ b/src/cpu/matmul/gemm_x8s8s32x_matmul.cpp
@@ -245,7 +245,9 @@ status_t gemm_x8s8s32x_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
     const char transB = helper.transB();
     const dim_t lda = helper.lda();
     const dim_t ldb = helper.ldb();
-    const dim_t ldc = helper.ldc();
+    const dim_t ldc = dst_d.ndims() == 2 && dst_d.count_non_unit_dims(1)
+            ? N
+            : helper.ldc();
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test
disable device_gpu
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_matmul

src/cpu/matmul/gemm_x8s8s32x_matmul.cpp
```diff
@@ -245,7 +245,9 @@ status_t gemm_x8s8s32x_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
     const char transB = helper.transB();
     const dim_t lda = helper.lda();
     const dim_t ldb = helper.ldb();
-    const dim_t ldc = helper.ldc();
+    const dim_t ldc = dst_d.ndims() == 2 && dst_d.count_non_unit_dims(1)
+            ? N
+            : helper.ldc();
```
ankalinin, line 250: Just curious. Why not move this logic into a helper.ldc()?

ankalinin (approved):

src/cpu/matmul/gemm_x8s8s32x_matmul.cpp
xuxinzen: Good question! The root problem for this issue is in `transA()`, `transB()` and `transC()`. These functions may not return correct value when have unit dimension. I tried to fixed them before, but they involved too many files and I still cannot figure out the best solution. So I decided to promote a temporary solution here

densamoilov (approved):

PR #2531, opened 2025-01-28
api: ukernel: make get_B_pack_type static for better usability

dzarukin:
Certain user applications can't create brgemm objects at creation stage because of lacking certain information, like strides.
But current API mandates a brgemm object to query an information about packing the tensor B.
Thus, such users get into a trap when they are forced to move ukernel creation to their execution stage, but the decision about tensor B and its packing must have been done at creation stage which makes API unusable.

To address this kind of situation, this PR makes the function that is responsible for packing requirement `static`, or independent from the object, which would let query that information based on data types (which are available) and make specific choices ahead of time.

### DIFF
--- a/examples/ukernels/cpu_brgemm.cpp
+++ b/examples/ukernels/cpu_brgemm.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


dzarukin:
make test
disable test_device_gpu
disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_matmul

mgouicem (approved):

examples/ukernels/cpu_brgemm.cpp
```diff
@@ -57,17 +57,27 @@ void brgemm_example() {
...
-    const memory::dim ldc = N; // Leading dimension for accumulator.
-    const memory::dim ldd = N; // Leading dimension for an actual output.
-    const memory::dim batch_size = n_calls - 1;
-
     memory::data_type a_dt = dt::u8;
     memory::data_type b_dt = dt::s8;
     memory::data_type c_dt = dt::s32; // Accumulator data type.
     memory::data_type d_dt = dt::f32; // Output data type.
 
+    // Query the packing requirement from the ukernel. It's enough to query
```
tczeszun, line 65: (random place)
Regarding description, so strides aren't known before execution? I thought it only happens with runtime dims.

tczeszun (approved):

examples/ukernels/cpu_brgemm.cpp
dzarukin: It's not about execution, it's about creation.
User's application doesn't have strides available at brgemm creation time. Yet they had to create a brgemm object to query the information about packing, and if yes, that would limit B strides options for the kernel.

PR #2533, opened 2025-01-28
cpu: x64: matmul: fix strides for weights in ba when K == 1

densamoilov:
It seems that when we introduced support for non-trivial strides in 78979db92f5b899f142ac2651b1787df7f968944 we also introduced a bug for the case when B tensor is in `ba` format and `K == 1`.

When K == 1 we always pick "ab" format for B (see set_or_check_B_tag) regardles of whether the actual tag was "ab" or  "ba".
Since the implementation assumes the "ab" format is used we cannot use bgmmc.B_strides[1] directly as the strides could be specified for "ba" therefore we need to use bgmmc.N instead.

Fixes MFDNN-13030.


densamoilov:
make test

ankalinin (approved):

inteldimitrius (approved):

PR #2535, opened 2025-01-28
benchdnn, ngen, xe: fix memory use miss-estimation triggering out-of-memory issues

rjoursler:
Partial fix for MFDNN-13045
In addition, memory estimation of iGPU memory within benchnn is currently incorrect, as the device memory needs to be added to the system memory. As identifying when this extra memory is a tricky problem, this PR just requires the device memory requirements fit within the available CPU memory.

In addition, there is a memory miss estimation coming from blocked memory formats used in the reference primitive. In the worst case, the padding in the blocked formats could increase memory utilization of a buffer by 16x. This issue is expected to be addressed in a future PR by @dzarukin.

### DIFF
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -1138,7 +1138,10 @@ static int check_total_size(
     }
 
     size_t total_size_cpu = check_mem_size_args.total_size_cpu;
-    if (is_cpu()) total_size_cpu += check_mem_size_args.total_size_device;
+
+    // Add device size as a simple method to account for integrated devices and
+    // mapping/unmapping memory
+    total_size_cpu += check_mem_size_args.total_size_device;
     bool fits_cpu_ram = total_size_cpu <= benchdnn_cpu_limit;
 
     if (!fits_cpu_ram) {


src/gpu/intel/compute/device_info.cpp
```diff
@@ -67,7 +67,11 @@ uint64_t get_future_extensions(
 
 bool device_info_t::is_integrated() const {
     auto family = static_cast<ngen::ProductFamily>(gpu_product_family_);
-    return ngen::getPlatformType(family) == ngen::PlatformType::Integrated;
+    auto type = ngen::getPlatformType(family);
+    if(type == ngen::PlatformType::Unknown) {
+        type = ngen::getPlatformType(device_id_);
+    }
+    return type == ngen::PlatformType::Integrated;
```
Simonsays095, line 74:
```suggestion
    gpu_assert(type != ngen::PlatformType::Unknown) << "Unknown GPU platform type";
    return type == ngen::PlatformType::Integrated;
```

Simonsays095 (approved):

third_party/ngen/ngen_core.hpp
```diff
@@ -300,6 +300,153 @@ static inline constexpr14 PlatformType getPlatformType(ProductFamily family) {
     return PlatformType::Unknown;
 }
 
+static inline constexpr14 PlatformType getPlatformType(uint32_t deviceID) {
+    // A list of device ids can be found in:
+    // https://github.com/intel/compute-runtime/blob/master/shared/source/dll/devices/devices_base.inl
+    // only NAMEDDEVICE ids are included.
+    switch(deviceID) {
```
echeresh, line 307: Looks like PTL is missing:
https://github.com/intel/compute-runtime/blob/d7ca6568fc70582bd88fbda239efdef57ec03763/shared/source/dll/devices/devices_base.inl#L10

I think we need `DEVICE` ids as well.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1126,7 +1185,8 @@ static int check_total_size(
     }
 
     size_t total_size_cpu = check_mem_size_args.total_size_cpu;
-    if (is_cpu()) total_size_cpu += check_mem_size_args.total_size_device;
+    if (is_unified_test_memory())
+        total_size_cpu += check_mem_size_args.total_size_device;
```
echeresh, line 1189: I would make it flipped, something like: `if(!is_discrete_gpu()) add-extra-size`
If some ID is missing/unknown - we better be on the safer side and take a pessimistic estimate. This would save time on debugging.
If the concern is that in this case we don't catch some missing device IDs - then I would incorporate a device IDs check similar to this one:
https://github.com/oneapi-src/oneDNN/blob/00ba39488366411d5bb9736aa576d34d3b9485f4/src/gpu/intel/jit/binary_format.cpp#L244

tests/benchdnn/dnnl_common.cpp
```diff
@@ -30,6 +30,14 @@
 #include "oneapi/dnnl/dnnl_sycl.hpp"
 #endif
 
+#if DNNL_GPU_VENDOR == DNNL_VENDOR_INTEL
+#define NGEN_NO_OP_NAMES
+#include "ngen/ngen_opencl.hpp"
```
echeresh, line 35: As an option this query could be implemented in the library, e.g. see `dnnl_impl_gpu_mayiuse_ngen_kernels`. This way we could avoid benchdnn dependency on nGEN.

@dzarukin @mgouicem Please weigh in.

third_party/ngen/ngen_core.hpp
```diff
@@ -300,6 +300,153 @@ static inline constexpr14 PlatformType getPlatformType(ProductFamily family) {
...
+        case 0xA7AA:
+        case 0xA7AB:
+        case 0xA7AC:
+        case 0xA7AD:
+            return PlatformType::Integrated;
+
+        default: return PlatformType::Unknown;
+    }
+}
+
```
petercad, line 449: Let's not rely on PCI device IDs as new ones are frequently added/unembargoed, so the library ends up being constantly out-of-date. Instead, can you add new `ProductFamily` entries as needed and extend the existing `ProductFamily` detection? This is based on GMDID, described [here](https://github.com/intel-innersource/drivers.gpu.specification.platforms/blob/generated_cs/gen/templates/doc/listing.md).

tests/benchdnn/dnnl_common.cpp
dzarukin: I suggested to make a generic check for summed device + cpu memory against cpu memory regardless of the GPU type. This would allow to avoid much code with low risk of important cases being skipped.

rjoursler:
make test
set test_scope=NIGHTLY
disable benchnn_all
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

tests/benchdnn/dnnl_common.cpp
rjoursler: Agreed. I investigated the interface with differentiating between iGPU and dGPU as that is something we have encountered for GPU work in the past. Given the complexity required in solving this problem, I agree this simpler solution is better for now.

echeresh (approved):

rjoursler:
make test
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

PR #2558, opened 2025-01-29
cpu: x64: brgemm: update conditions for zp with big ow

xuxinzen:
Partially fixed MFDNN-12557

When the shapes has big ow but small ic/oc, it may require small calculation work in pre-calculated compensation. So it's better to calculate the pre-calculate compensation instead of calculating the compensation in brgemm kernel. 

Summarized performance data on SRF and CLX:
image
image

Raw data:
[brg_zp_avx2_vnni_2.xlsx](https://github.com/user-attachments/files/18596637/brg_zp_avx2_vnni_2.xlsx)
[brg_zp_avx512_core_vnni.xlsx](https://github.com/user-attachments/files/18596639/brg_zp_avx512_core_vnni.xlsx)

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_utils.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
 *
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test
disable device_gpu

ankalinin (approved):

densamoilov (approved):

PR #2562, opened 2025-01-30
cpu: x64: disable MM IP for unsupported ISAs

densamoilov:
We don't support Inner Product for training on AVX2, the only exception is an f32 data type configuration.

### DIFF
--- a/src/cpu/x64/matmul_inner_product.hpp
+++ b/src/cpu/x64/matmul_inner_product.hpp
@@ -132,6 +132,13 @@ struct matmul_inner_product_bwd_data_t : public primitive_t {
             const auto diff_dst_dt = invariant_dst_md()->data_type;
             const auto wei_dt = invariant_wei_md()->data_type;
 
+            const bool is_f32
+                    = utils::everyone_is(f32, diff_src_dt, wei_dt, diff_dst_dt);
+
+            VDISPATCH_INNER_PRODUCT(mayiuse(avx2), VERBOSE_UNSUPPORTED_ISA);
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul_inner_product.hpp
```diff
@@ -132,6 +132,10 @@ struct matmul_inner_product_bwd_data_t : public primitive_t {
             const auto diff_dst_dt = invariant_dst_md()->data_type;
             const auto wei_dt = invariant_wei_md()->data_type;
 
+            const bool is_f32
+                    = utils::everyone_is(f32, diff_src_dt, wei_dt, diff_dst_dt);
+            if (!mayiuse(avx512_core) && !(mayiuse(avx2) && is_f32))
+                return status::unimplemented;
```
dzarukin, line 138:
```suggestion
            VDISPATCH_INNER_PRODUCT(mayiuse(avx2), VERBOSE_UNSUPPORTED_ISA);
            const bool is_f32
                    = utils::everyone_is(f32, diff_src_dt, wei_dt, diff_dst_dt);
            VDISPATCH_INNER_PRODUCT(IMPLICATION(!is_f32, mayiuse(avx512)), VERBOSE_UNSUPPORTED_DT_CFG);
```

dzarukin (approved):

avmanerikar (approved):

inteldimitrius (approved):

src/cpu/x64/matmul_inner_product.hpp
densamoilov: Thanks! Keep forgetting about the macros :\

PR #2564, opened 2025-01-30
gpu: nvidia: matmul: fix issues with scaling

sgeor255:
# Description

Currently tests that use the cublasLt implementation of matmul with src, weight or dst scaling fail. This PR fixes several issues:
- For dst scaling, the binary operation was created with the incorrect datatype for the scales tensor. Additionally, the memory descriptor for the scales tensor was not initialised correctly and was triggering an assertion due to inner_idxs not being initialised to sensible values. Additionally, a `shared_ptr` containing the binary_pd was freed incorrectly at the end of the `create_scaling_format_desc` function, resulting in a segmentation fault.
- src & weight scaling currently cannot be supported with this implementation. The current approach was to use the binary primitive to scale the src/weights before performing matmul. However, since the cublasLt implementation only works with integer types, the output of the src/weight scaling binary was being converted to an integer type. The scaling factors are floating point numbers, therefore the scaled src/weights were being truncated and then passed to the matmul, outputting incorrect results. I tested using the alpha parameter of the gemm parameter to perform this scaling, however since the compute type is set to an integer type, the scaling is performed in integer type too, also resulting in incorrect output. This PR therefore removes the code related to src/weight scaling as it wasn't correct and adds a return in the init of the primitive for cases that have src/wei scales. 

# Checklist

## General

### DIFF
--- a/src/common/memory_tracking.hpp
+++ b/src/common/memory_tracking.hpp
@@ -266,8 +266,6 @@ enum {
     key_matmul_wei_trans,
     key_matmul_dst_trans,
     key_matmul_dst_cast_acc,
-    key_matmul_lt_src_scale,
-    key_matmul_lt_wei_scale,
     key_matmul_sparse_tmp_ptr,
     key_pool_dst_bf16cvt,
     key_pool_dst_plain2blocked_cvt,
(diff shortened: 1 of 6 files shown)


dzarukin:
Hi @sgeor255, I did some clean up in #2270 and merged it. Could you, please, rebase and check if all changes are still needed?

mgouicem (approved):

Rbiessy (approved):

ShanoToni (approved):

PR #2571, opened 2025-01-31
Fix LSTM INT8 for AVX2

densamoilov:
Incorrect format was used for projection weights on avx2. This PR fixes it.
There is no tracker for the issue.


mgouicem (commented):
How was it uncovered? do we need to add tests?

Also, you are adding support for that layout in reorder ~~but no implementation seems to match or require that tag. Could you clarify how that plays with existing impls?~~ Sorry I had missed the rnn_utils.cpp hunk, my bad.

densamoilov:
make test

mgouicem (approved):
Ok after offline discussion, it seems the issue was caught by nightly, so no new tests needed.
Thanks.

ankalinin (approved):

PR #2576, opened 2025-01-31
Fix a condition for when A can be treated as plain (non-transposed)

densamoilov:
Fixes MFDNN-13124.

When M == 1 MatMul always considers A to be non-transposed even if A md was created using "ba" tag so the condition for `treat_transposed_A_as_plain` was incorrect.


vpirogov:
PR linters fail because of leading space in the commit message.

dzarukin (approved):

spalicki (approved):

ankalinin (approved):

PR #2581, opened 2025-02-03
amd: fix build errors (fixes MFDNN-13026 and fixes MFDNN-13141)

dzarukin:
MFDNN-13026
MFDNN-13141

### DIFF
--- a/.github/automation/commit-msg-check.py
+++ b/.github/automation/commit-msg-check.py
@@ -38,16 +38,22 @@ def __scopeCheck(msg: str):
     print(f"{status} OK")
     return True
 
-# Ensure  a character limit for the first line.
+# Ensure a character limit for the first line.
 def __numCharacterCheck(msg: str):
     status = "Message length:"
     if len(msg) <= 72:
         print(f"{status} OK")
         return True
     else:
-        print(f"{status} FAILED: Commit message summary must not "
-               "exceed 72 characters.")
-        return False
+        # Fixup commits usually include the full name of the commit they are
+        # fixing, which adds 6 more symbols to the message. Let them in.
+        if re.match('^fixup: ', msg):
+            print(f"{status} Fixup message, OK")
+            return True
+        else:
+            print(f"{status} FAILED: Commit message summary must not "
+                   "exceed 72 characters.")
+            return False
 
 def main():
     parser = argparse.ArgumentParser()
--- a/src/gpu/amd/miopen_binary.cpp
+++ b/src/gpu/amd/miopen_binary.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2024 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 * Copyright 2020 Codeplay Software Limited
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,6 +14,8 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 *******************************************************************************/
+
+#include "common/compiler_workarounds.hpp"
 
 #include "gpu/amd/miopen_binary.hpp"
 #include "gpu/amd/stream.hpp"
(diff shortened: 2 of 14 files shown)


.github/automation/commit-msg-check.py
```diff
@@ -38,16 +38,22 @@ def __scopeCheck(msg: str):
...
-        return False
+        # Fixup commits usually include the full name of the commit they are
+        # fixing, which adds 6 more symbols to the message. Let them in.
+        if re.match('^fixup: ', msg):
+            print(f"{status} Fixup message, OK")
+            return True
+        else:
+            print(f"{status} FAILED: Commit message summary must not "
+                   "exceed 72 characters.")
+            return False
```
Rbiessy, line 56: Is it intentional to update this file as part of an amd PR? It should probably be separated.

src/gpu/amd/miopen_binary.cpp
```diff
@@ -40,7 +40,7 @@ status_t miopen_binary_t::execute(const exec_ctx_t &ctx) const {
         auto arg_scale1
                 = CTX_IN_SYCL_MEMORY(DNNL_ARG_ATTR_SCALES | DNNL_ARG_SRC_1);
 
-        compat::host_task(cgh, [=](const compat::interop_handle &ih) {
+        compat::host_task(cgh, [=, this](const compat::interop_handle &ih) {
```
Rbiessy, line 43: Shouldn't this use `WA_THIS_COPY_CAPTURE`, just like https://github.com/oneapi-src/oneDNN/pull/2506 to avoid the original issue mentioned in MFDNN-13026?

dzarukin: Copy-pasted from existing Nvidia backend code base.
Code enhancements would be appreciated for sure.

.github/automation/commit-msg-check.py
dzarukin: It is, because, otherwise, the linter doesn't let it pass due to a long message.

src/gpu/amd/miopen_binary.cpp
vpirogov: If this WA is required for build to succeed this is not a code enhancement :)

dzarukin: It's working for the other backend, why wouldn't it for this?
If it wouldn't for some reason I'll submit another PR.

vpirogov:
make test
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_cdna2
enable compiler_icx-oss

vpirogov:
make test
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_cdna2
enable compiler_icx-oss
enable build_mode_cpp20

.github/automation/commit-msg-check.py
vpirogov: > Is it intentional to update this file as part of an amd PR? It should probably be separated.

Having this change as a separate commit in this PR looks fine.

dzarukin:
make test
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_cdna2
enable compiler_icx-oss
enable build_mode_cpp20

src/gpu/amd/miopen_binary.cpp
Rbiessy: Thanks for fixing this. So it seems for the NVidia backend we can use `[=, this]` because the warnings were disabled in [CMake](https://github.com/oneapi-src/oneDNN/blob/main/cmake/SYCL.cmake#L81) with `-Wno-c++20-extensions`. On AMD the warnings are emitted and it fails due to `-Werror`.
I'm pretty sure we can't use capture by reference to capture buffer's accessors in SYCL. It could work for USM but the paths for USM and buffers are the same.
I think we could improve the situation by removing `-Wno-c++20-extensions` and using `WA_THIS_COPY_CAPTURE` when needed.

Rbiessy (approved):

dzarukin:
make test
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_cdna2
enable compiler_icx-oss
enable build_mode_cpp20

ShanoToni (approved):

PR #2583, opened 2025-02-03
gpu: ocl: benchdnn: enable RNG memory fill in --mode=F

yehudaorel:
# Description

Systems with data compression enabled by the driver by default wont generate meaningful performance data with `benchdnn ... --mode=F `

- Reuse [ocl_philox.h](https://github.com/oneapi-src/oneDNN/blob/main/src/gpu/intel/ocl/ocl_philox.h) kernel to directly fill gpu mem with random vals
- cold-cache support

Fixes [MFDNN-12589](https://jira.devtools.intel.com/browse/MFDNN-12589)

# Checklist

## General

### DIFF
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -628,12 +628,8 @@ int dnn_mem_t::memset_rng(size_t size) const {
                 cl_platform_id platform;
                 clGetDeviceInfo(ocl_device, CL_DEVICE_PLATFORM,
                         sizeof(cl_platform_id), &platform, nullptr);
-                auto clSetKernelArgMemPointerINTEL
-                        = (clSetKernelArgMemPointerINTEL_fn)
-                                clGetExtensionFunctionAddressForPlatform(
-                                        platform,
-                                        "clSetKernelArgMemPointerINTEL");
-                clSetKernelArgMemPointerINTEL(ocl_kernel, 0, mem_handle);
+                DNN_SAFE_V(dnnl::impl::xpu::ocl::usm::set_kernel_arg(
+                        engine_, ocl_kernel, 0, mem_handle));
                 clSetKernelArg(ocl_kernel, 1, sizeof(uint64_t), &buffSize);
                 clSetKernelArg(ocl_kernel, 2, sizeof(uint64_t), &blockSize);
                 clSetKernelArg(ocl_kernel, 3, sizeof(uint64_t), &DEFAULT_SEED);
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -20,6 +20,7 @@
 #include <memory>
 #include <numeric>
 #include <string>
+#include <fstream>
 
 #include "oneapi/dnnl/dnnl.h"
 
@@ -30,7 +31,7 @@
 #if DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
 #include "oneapi/dnnl/dnnl_ocl.hpp"
 #include "src/xpu/ocl/usm_utils.hpp"
-#include "utils/ocl_philox.h"
+
 #define OCL_TEST(x) \
     do { \
         cl_int s = (x); \
@@ -558,7 +559,7 @@ void dnn_mem_t::memset(int value, size_t size) const {
     SAFE_V(FAIL);
 }
 
-dnnl_status_t dnn_mem_t::memset_rng(size_t size) const {
+int dnn_mem_t::memset_rng(size_t size) const {
     bool is_opencl = is_opencl_engine(engine_);
     bool is_sycl = is_sycl_engine(engine_);
     auto mem = m_padded_ ? m_padded_ : m_;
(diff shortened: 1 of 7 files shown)


tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,72 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+                clEnqueueNDRangeKernel(ocl_queue, ocl_kernel, 1, nullptr, &gws, nullptr, 0, nullptr, nullptr);
+                DNN_SAFE_V(dnnl_stream_wait(stream));
+                return dnnl_status_t::dnnl_success;
+            }
+        }
+#endif
+    } else if (is_sycl) {
+#ifdef DNNL_WITH_SYCL
+        // TODO: add sycl support
+         return dnnl_status_t::dnnl_unimplemented;
```
rjoursler, line 619: There should be some fixed size window where data compression can happen, and I believe it is on the order of 256 bytes, so it is not particularly big. What if we (ab)use the binary primitive to broadcast random data into the buffer by doing `binary_max(uncompressed_src, uncompressed_src)` with broadcast semantics? That should be a bit simpler as then we won't have to maintain all this runtime specific code. although it will introduce memory size alignment requirements.

Simonsays095: I don't follow - how does `binary_max` generate random data?

echeresh:
@yehudaorel Did you compare it with the current data filling approach? How much is the additional overhead on RNG-based data filling?

tests/benchdnn/dnnl_memory.cpp
rjoursler: Binary max doesn't generate the data, instead we setup 1 incompressible source buffer at startup. Essentially, we do something like
```C++
out_data get_mem_rng(size) {

    static dnnl_mem isrc = []() {
        uint8_t rand_data[256];
        fill_rng(rand_data);
        dnnl_mem isrc;
        reorder(isrc, rand_data)
        return isrc;
    }

    dnnl_mem out_data(utils::rnd_up(size, 256))
    out_data = binary_max(isrc, isrc);
    return out_data;
}
```

yehudaorel:
> @yehudaorel Did you compare it with the current data filling approach? How much is the additional overhead on RNG-based data filling?

Based on my testing, this will introduce ~5% overhead, which in realistic terms will only be a couple seconds difference when running larger tests.

timed 1perf: [A570M]

- **base(main) with cold-cache**
  - real    4m19.663s
  - user    2m32.354s
  - sys     0m56.664s
  
- **base(main) with no-cold-cache:**
  - real    3m18.686s
  - user    2m15.223s
  - sys     0m27.828s
 
- **PRNG with cold-cache:**
  - real    4m11.391s
  - user    2m1.066s
  - sys     0m58.836s
  
- **PRNG with no-cold-cache:**
  - real    4m3.358s
  - user    2m17.656s
  - sys     0m30.760s

```
perf,--mode=F --matmul --engine=gpu --memory-kind=buffer --stag=ab --wtag=ba --dtag=ab --attr-fpmath=f16:true 1x1024:1024x1024,0.160625,26.1634,13.0562
total perf: min(ms):0.160625 avg(ms):0.164043
total: 0.07s; fill: 0.00s (0%);

=== Device Timing Summary ===

                Total Execution Time (ns):             74905580
Total Device Time for CL GPU backend (ns):             12576009

== CL GPU Backend ==

                                            Kernel,       Calls,           Time (ns),    Time (%),        Average (ns),            Min (ns),            Max (ns)
         gemm_kernel[SIMD8 {512; 8; 4} {16; 8; 4}],          73,            11977260,       95.24,              164072,              160625,              167395
philox_fill_kernel[SIMD16 {65536; 1; 1} {0; 0; 0}],           1,              559479,        4.45,              559479,              559479,              559479
  philox_fill_kernel[SIMD16 {256; 1; 1} {0; 0; 0}],           2,               39270,        0.31,               19635,               19583,               19687


```

tests/benchdnn/dnnl_memory.cpp
echeresh: @rjoursler Do you have details on how compression is done? For example, if it can detect a repeating pattern at the page granularity (4 kilobytes) then any pattern of a power of two size would result in some subset of pages having the same data which exposes compression opportunities for the hardware.

echeresh: Discussed with Roy offline - the current hardware seems to work with 256-byte blocks (and future hardware may have possibly larger blocks). So using a small block of 256 bytes with random data as a pattern should be enough. 

But I'm not sure about this trick with broadcasting via binary primitive as it looks like `dst` must match either `src0` or `src1`:

https://github.com/oneapi-src/oneDNN/blob/a6a303a0805687f965453bbf7b66eb34cb6ebe6f/src/common/binary.cpp#L130-L131

rjoursler: Thanks for finding that @echeresh, it sounds like a separate kernel would would still be required.

karturov:
@yehudaorel , 5% overhead is acceptable.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,75 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+        DNN_SAFE_V(dnnl_ocl_interop_engine_get_context(engine_, &ocl_ctx));
+        DNN_SAFE_V(dnnl_ocl_interop_get_device(engine_, &ocl_device));
+        philox_program = clCreateProgramWithSource(
+                ocl_ctx, 1, &philox_rng_source, nullptr, &err);
+        OCL_TEST(clBuildProgram(
+                philox_program, 1, &ocl_device, nullptr, nullptr, nullptr));
+        ocl_kernel = clCreateKernel(philox_program, "philox_fill_kernel", &err);
+
+        //TODO: generate random seed to avoid repeating values
+        uint64_t seed = 12345;
```
dzarukin, line 587: tests/benchdnn/dnn_types.hpp:223 says that using seed as `-1` brings better entropy. Probably better to use the same approach here.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,75 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+        const uint64_t buffSize = (uint64_t)size;
```
dzarukin, line 588: I'd put an assert that `size` is less than UINT64_MAX to make sure this cast is valid.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,75 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+                auto buf = static_cast<cl_mem>(mem_handle);
+                clSetKernelArg(ocl_kernel, 0, sizeof(cl_mem), &buf);
+                clSetKernelArg(ocl_kernel, 1, sizeof(uint64_t), &buffSize);
+                clSetKernelArg(ocl_kernel, 2, sizeof(uint64_t), &blockSize);
+                clSetKernelArg(ocl_kernel, 3, sizeof(uint64_t), &seed);
+                clEnqueueNDRangeKernel(ocl_queue, ocl_kernel, 1, nullptr, &gws,
+                        nullptr, 0, nullptr, nullptr);
+
+                DNN_SAFE_V(dnnl_stream_wait(stream));
+                return dnnl_status_t::dnnl_success;
```
dzarukin, line 602:
```suggestion
                return dnnl_success;
```

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,75 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+            }
+            case memory_kind_ext_t::usm:
+            case memory_kind_ext_t::usm_device:
+            case memory_kind_ext_t::usm_shared: {
+                DNN_SAFE_V(dnnl::impl::xpu::ocl::usm::set_kernel_arg(
```
dzarukin, line 607: Looking underneath the call, it seems that it can be simply replaced by clSetKernelArgMemPointerINTEL since the code below still operates with OpenCL objects directly.
Exposing this symbol wouldn't be required.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +558,75 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+                        engine_, ocl_kernel, 0, mem_handle));
+                clSetKernelArg(ocl_kernel, 1, sizeof(uint64_t), &buffSize);
+                clSetKernelArg(ocl_kernel, 2, sizeof(uint64_t), &blockSize);
+                clSetKernelArg(ocl_kernel, 3, sizeof(uint64_t), &seed);
+
+                clEnqueueNDRangeKernel(ocl_queue, ocl_kernel, 1, nullptr, &gws,
+                        nullptr, 0, nullptr, nullptr);
+                DNN_SAFE_V(dnnl_stream_wait(stream));
+                return dnnl_status_t::dnnl_success;
```
dzarukin, line 616: Ditto.

karturov:
make test perf-gpu
disable arch_gpu_xe-hpc
disable arch_gpu_xe-lp
disable arch_gpu_xe2-lpg

karturov:
make test perf-gpu
disable arch_gpu_xe-hpc
disable arch_gpu_xe-lp
disable arch_gpu_xe2-lpg

petercad:
One side effect of RNG fill is that it will increase power usage, leading to significant performance variability for power-limited sizes. Has anyone checked if 1perf handles this situation properly (e.g. no false regressions)?

@echeresh I remember that when you first introduced `--mode=po` you mentioned the low power usage of constant/zero fill as an asset in performance tracking.

echeresh:
> One side effect of random fill is that it will increase power usage, leading to significant performance variability for power-limited sizes. Has anyone checked if 1perf handles this situation properly (e.g. no false regressions)?
> 
> @echeresh I remember that when you first introduced `--mode=po` you mentioned the low power usage of constant/zero fill as an asset in performance tracking.

Good point. Yes, this is a known issue/behavior. @yehudaorel I'd suggest to do a couple of performance runs with the same commit and look at the run-to-run variability. We need to confirm that it's stable with the random data filling.

I'll list two performance effects and implications for our performance testing. Unfortunately, it seems there is no approach that is reliable and fast enough for our performance testing (but still, I hope we can manage it with our performance testing infrastructure).

- Zero (or constant) initialized data vs random data requires less power and causes less frequency throttling
    - Implications: performance with random data is 1) generally lower and 2) less stable - there is a higher chance of throttling (which depends on the test cases and hardware)
    - A good thing: a lot of cases in our performance testing are not compute-bound so it seems even if frequency drops, it often bounces back without compromising the numbers
    - A bad thing: the most stable way to measure performance is to run a case long enough to ensure we run at a sustained frequency. I assume it's hundreds of milliseconds per case vs ~10ms per case as of now
- L3 cache write-back overhead. If a kernel writes to global memory, it may write to L3 cache only leaving L3 -> global memory transfer for future queue synchronization (which triggers cache flush) or future kernel submissions (if they evict that L3 data to free up it for their data)
    - Implications: first kernel runs are generally faster - they have "fresh" L3 and don't have to evict previous data

Both effects result in a significant gap between minimum/average times. When moving to random data filling, stability should suffer more due to the first effect.

As for the practical side, we have some "mitigations" in our performance testing so it's not that bad:

- We use the minimum time for performance reports. It's more stable and suffers less from throttling and L3 cache impact
- When a regression is detected, we do several re-runs to confirm the regression is there
- We have quite a few non compute-bound cases which limits the impact of the 1st effect

petercad:
> * Zero (or constant) initialized data vs random data requires less power and causes less frequency throttling
>   
>   * Implications: performance with random data is 1) generally lower and 2) less stable - there is a higher chance of throttling (which depends on the test cases and hardware)

Another side-effect (actually, the one I'm most worried about) is that frequency throttling causes the performance of one power-bound layer to affect the performance of a following layer in batch testing. Some possible results:

* Reordering layers in batch testing may affect the performance reported for each layer.
* An optimization that leads to greater compute efficiency in one power-bound layer may reduce frequency for the following layers, causing (false) apparent regressions for those layers.

echeresh:
> Another side-effect (actually, the one I'm most worried about) is that frequency throttling causes the performance of one power-bound layer to affect the performance of following layers in batch testing. Some possible results:
> 
> * Reordering layers in batch testing may affect the performance reported for each layer.
> * An optimization that leads to greater compute efficiency in one power-bound layer may reduce frequency for the following layers, causing (false) apparent regressions for those layers.

I agree, in isolation throttling is less of a problem as the first few runs are not usually affected (unless the sizes are huge).

I guess we need to step into some related issues to see what helps and what doesn't. For example, we can add a short delay after compute-bound shapes in benchdnn to let the frequency recover, or, if nothing helps, switch to longer runs.

petercad:
> I agree, in isolation throttling is less of a problem as the first few runs are not usually affected (unless the sizes are huge).
> 
> I guess we need to step into some related issues to see what helps and what doesn't. For example, we can add a short delay after compute-bound shapes in benchdnn to let the frequency recover, or, if nothing helps, switch to longer runs.

Can we detect memory-bound vs. compute-bound cases automatically (for select primitives at least)? So by default fill mode would be constant for compute-bound, random for memory-bound, with a knob to override the default.

yehudaorel:
> Can we detect memory-bound vs. compute-bound cases automatically (for select primitives at least)? So by default fill mode would be constant for compute-bound, random for memory-bound, with a knob to override the default.

I believe this could be done, although for compute-bound with constant filling cases, we might end up with original issue of inflated performance data again due to data compression.

tests/benchdnn/utils/ocl_philox.h
```diff
@@ -0,0 +1,73 @@
...
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key0.s67);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key0.s89);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key0.sAB);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key0.sCD);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key0.sEF);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key1.s01);
+    ctr = PHILOX_4UINT_ROUND(ctr_mul, ctr, key1.s23);
+
+    return ctr[~idx & 3L];
+}
```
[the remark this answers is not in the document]
atkassen: @dzarukin do you mean to make an ocl_philox.cl as part of benchdnn which `#include`s the library ocl_philox.h?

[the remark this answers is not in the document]
dzarukin: The last time I checked, benchdnn needed only a portion of what resides in philox implementation inside the library.
I suggested to split what we have in the library into two parts, the kernel itself (into .cl) and auxiliary functions/definitions (into .h). Include the kernel in benchdnn and call it a day.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -548,6 +559,98 @@ void dnn_mem_t::memset(int value, size_t size) const {
...
+            std::ifstream file("../../../src/gpu/intel/ocl/ocl_philox.cl");
+            if (!file.is_open()) {
+                std::cerr << "Failed to open ocl_philox.cl" << std::endl;
+                return dnnl_invalid_arguments;
+            }
+            std::stringstream buffer;
+            buffer << file.rdbuf();
+            philox_kernel_source = buffer.str();
+        }
+        const char *philox_rng_source = philox_kernel_source.c_str();
```
atkassen, line 593: @yehudaorel can you simply inline the kernel here as a string literal? This should prevent any fragility w.r.t. directory structure. Notably, using relative paths will not work with performance testing.

PR #2587, opened 2025-02-04
Fix memory storage allocation for SYCL CPU

densamoilov:
SYCL CPU engine is inherited from classic CPU engine that overrides an interface for creating a memory storage therefore SYCL CPU engine must also override it.

This bug was introduced during the recent architecture refactoring (8187bb585b2e242fba778266cb8d2c396fe669d1) and caused SYCL CPU engine to allocate classic CPU memory storage. ~~It means that we haven't been testing usm/buffer for SYCL CPU since the bug was introduced. Also, it means that set_data_handle()/get_data_handle() and their counterparts for SYCL buffers might not work correctly (could manifest as weird issues in validation)~~.

**Update:** Only the `memory_t` objects created inside the library are affected.

I also, added additional checks to reduce the chance of introducing such a bug in the future.

### DIFF
--- a/src/cpu/cpu_engine.cpp
+++ b/src/cpu/cpu_engine.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2016-2024 Intel Corporation
+* Copyright 2016-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


densamoilov:
make test

mgouicem (approved):
LGTM.
Shouldn't that be covered by gtests?

Rbiessy (approved):

densamoilov:
@mgouicem,
> Shouldn't that be covered by gtests?

I double checked it and it seems that only the `memory_t` objects created inside the library are affected meaning that public API always goes to SYCL interop API that creates correct `memory_t` objects. So from the gtest's perspective everything is good.

Unit tests would be the best way to test such things but we don't have them yet. I think the checks that I added should be enough.

PR #2589, opened 2025-02-04
Enable matmul-based RNN implementation back for SYCL CPU

densamoilov:
Fixes MFDNN-13148 and MFDNN-12386.

Regular, raw CPU pointers can only be used with memory_t objects created for the classic CPU engine. Use service engine to create such objects.


densamoilov:
make test

ankalinin (approved):

PR #2590, opened 2025-02-04
gtest: pool: remove test case with ih=0

sgeor255:
# Checklist

## General

### DIFF
--- a/tests/gtests/test_pooling_forward.cpp
+++ b/tests/gtests/test_pooling_forward.cpp
@@ -91,11 +91,6 @@ bool generic_check_format_tags(memory::format_tag format) {
             memory::format_tag::any);
 }
 
-bool generic_has_input_zero_dim(pool_test_params_t &p) {
-    auto &pd = p.test_pd;
-    return pd.mb == 0 || pd.id == 0 || pd.ih == 0 || pd.iw == 0;
-}
-
 template <typename data_t>
 void check_pool_fwd(const pool_test_params_t &p, const memory &src,
         const memory &dst, const memory &ws) {
@@ -244,9 +239,6 @@ protected:
                 "Unsupported format tag");
         SKIP_IF_GENERIC(!generic_check_format_tags(p.dst_format),
                 "Unsupported format tag");
-        // XXX: Enable when 0-dim input is supported in generic implementation
-        SKIP_IF_GENERIC(generic_has_input_zero_dim(p),
-                "Input dims == 0 are not supported");
 
         catch_expected_failures(
                 [&]() { Test(); }, p.expect_to_fail, p.expected_status);
@@ -785,12 +777,7 @@ INSTANTIATE_TEST_SUITE_P(TestPoolingForwardZeroDim, pooling_test_float,
                         algorithm::pooling_max, memory::format_tag::nhwc,
                         memory::format_tag::nhwc,
                         EXPAND_SIZES_2D(
-                                0, 4, 4, 4, 4, 4, 3, 3, 0, 0, 1, 1, 1, 1)},
-                pool_test_params_float {prop_kind::forward_training,
-                        algorithm::pooling_max, memory::format_tag::nchw,
-                        memory::format_tag::nchw,
-                        EXPAND_SIZES_2D(
-                                2, 4, 0, 4, 4, 4, 3, 3, 1, 1, 1, 1, 1, 1)}));
+                                0, 4, 4, 4, 4, 4, 3, 3, 0, 0, 1, 1, 1, 1)}));
 
 INSTANTIATE_TEST_SUITE_P(TestPoolingForwardEF, pooling_test_float,
         ::testing::Values(


ShanoToni (approved):

Rbiessy (approved):

vpirogov:
ping @oneapi-src/onednn-arch

dzarukin:
@sgeor255 Could you, please, post a benchdnn problem that this fix targets?

src/gpu/generic/sycl/ref_pooling.cpp
```diff
@@ -65,18 +65,55 @@ status_t ref_pooling_fwd_t::pd_t::init_conf() {
...
+
+        switch (dst_mdw.data_type()) {
+            CASE(f16)
+            CASE(bf16)
+            CASE(f32)
+            CASE(s8)
+            CASE(u8)
+            default:
+                assert(!"unhandled datatype");
+                return status::unimplemented;
```
dzarukin, line 107: This error should be reported at `pd_t::init()` level. Here the only thing possible is runtime_error which shouldn't be reachable by the user.

sgeor255: There is already a check for the supported datatypes in the `pd::init()` function. I mainly added this default case here in case we ever add more datatypes to pooling and forget to update this switch. 
Changed it to `runtime_error`, thanks for the suggestion.

dzarukin:
> > @sgeor255 Could you, please, post a benchdnn problem that this fix targets?
> 
> @dzarukin I found this issue through gtests, `TestPoolingForwardZeroDim/pooling_test_float.TestsPooling/2` tests this case:
> 
> ```
> onednn_verbose,v1,primitive,create:cache_miss,gpu,pooling,dpcpp:ref:any,forward_training,src:f32::blocked:abcd::f0 dst:f32::blocked:abcd::f0 ws:u8::blocked:abcd::f0,,a
> lg:pooling_max,mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1,36.1221
> ```
> 
> I am not able to run it with benchdnn as it seems it doesn't allow `ih` to be set to 0:
> 
> ```
> ./benchdnn --pool --engine=gpu -v1 mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1
> ERROR: setting `ih` was not specified or set to 0. Full descriptor input: `mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1`.
> Error: Function 'bench' at (/home/svet/oneDNN-nvidia/tests/benchdnn/pool/bench_pool.cpp:93) returned '1'
> ```

Seems like the case is totally broken (hope it's not me adding it)...
I don't think the shape is legit with `ih` equal to 0 because there's nothing to fuse with the kernel. Primitive_desc_create function should have checked for it...
If there are no cases with channels equal to 0, let's remake this one into it. If there's let's just drop this one.

sgeor255:
> > > @sgeor255 Could you, please, post a benchdnn problem that this fix targets?
> > 
> > 
> > @dzarukin I found this issue through gtests, `TestPoolingForwardZeroDim/pooling_test_float.TestsPooling/2` tests this case:
> > ```
> > onednn_verbose,v1,primitive,create:cache_miss,gpu,pooling,dpcpp:ref:any,forward_training,src:f32::blocked:abcd::f0 dst:f32::blocked:abcd::f0 ws:u8::blocked:abcd::f0,,a
> > lg:pooling_max,mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1,36.1221
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am not able to run it with benchdnn as it seems it doesn't allow `ih` to be set to 0:
> > ```
> > ./benchdnn --pool --engine=gpu -v1 mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1
> > ERROR: setting `ih` was not specified or set to 0. Full descriptor input: `mb2ic4_ih0oh4kh3sh1dh1ph1_iw4ow4kw3sw1dw1pw1`.
> > Error: Function 'bench' at (/home/svet/oneDNN-nvidia/tests/benchdnn/pool/bench_pool.cpp:93) returned '1'
> > ```
> 
> Seems like the case is totally broken (hope it's not me adding it)... I don't think the shape is legit with `ih` equal to 0 because there's nothing to fuse with the kernel. Primitive_desc_create function should have checked for it... If there are no cases with channels equal to 0, let's remake this one into it. If there's let's just drop this one.

There is a case with ic=0 (`TestPoolingForwardZeroDim/pooling_test_float.TestsPooling/0`):

```
onednn_verbose,v1,primitive,create:cache_miss,gpu,pooling,dpcpp:ref:any,forward_training,src:f32::blocked:abcd::f0 dst:f32::blocked:abcd::f0 ws:u8::blocked:abcd::f0,,a
lg:pooling_max,mb2ic0_ih4oh4kh3sh1dh0ph1_iw4ow4kw3sw1dw0pw1,36.717
```

I removed the case with ih=0 as suggested, thanks.

tests/gtests/test_pooling_forward.cpp
```diff
@@ -785,12 +777,7 @@ INSTANTIATE_TEST_SUITE_P(TestPoolingForwardZeroDim, pooling_test_float,
                         algorithm::pooling_max, memory::format_tag::nhwc,
                         memory::format_tag::nhwc,
                         EXPAND_SIZES_2D(
-                                0, 4, 4, 4, 4, 4, 3, 3, 0, 0, 1, 1, 1, 1)},
-                pool_test_params_float {prop_kind::forward_training,
-                        algorithm::pooling_max, memory::format_tag::nchw,
-                        memory::format_tag::nchw,
-                        EXPAND_SIZES_2D(
-                                2, 4, 0, 4, 4, 4, 3, 3, 1, 1, 1, 1, 1, 1)}));
+                                0, 4, 4, 4, 4, 4, 3, 3, 0, 0, 1, 1, 1, 1)}));
```
dzarukin, line 780: With the test updated, I feel most changes from the library can be dropped. Could you confirm, please?

sgeor255: @dzarukin I think even though we removed the unsupported test and benchdnn doesn't allow cases with ih==0, it is still possible to run such a case if using the oneDNN API programmatically. Additionally, the cuda implementation has this behaviour implemented (https://github.com/oneapi-src/oneDNN/blob/main/src/gpu/nvidia/cudnn_pooling.cpp#L43-L45). Should I remove the library changes in this PR and add a check in the `init()` for input dimensions != 0 (except C dimension)? Or is this a check that should be added in the primitive desc create so it applies to all pooling implementations? Or does it make sense to keep the changes in this PR?

dzarukin: Zero spatial has no sense for pooling operation, it shouldn't be used by users and must be prohibited on the library API side (so it's not an impl level, it's a higher level). I'll take care of it. Please remove all related to zero-spatial changes as they are not supposed to be needed. If there are any changes needed for zero-batch or zero-channel cases, those should remain. Thanks.

sgeor255: Removed the library changes as they are not required in this case, thanks.

dzarukin (approved):

Issue #2572, opened 2025-01-31
test_graph_unit_dnnl_convolution_cpu fails on AArch64 CI after recent quantization refactor


michalowski-arm:
# Summary
After a recent quantization refactor (https://github.com/oneapi-src/oneDNN/pull/2270) we see a failure in Nightly AArch64 pipeline (https://github.com/oneapi-src/oneDNN/actions/runs/13066726551/job/36460347340). Upon bisecting, the first bad commit appears to be af1410c2.

# Version
ACL version v24.11.1

# Environment
OneDNN GitHub CI for AArch64, c7g AWS instance

# Steps to reproduce
`ctest -R "test_graph_unit_dnnl_convolution_cpu" --rerun-failed --output-on-failure`

# Observed behavior

> [ RUN      ] test_convolution_execute.ConvBiasAddEltwise
> /opt/actions-runner/_work/oneDNN/oneDNN/oneDNN/tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp:2557: Failure
> Expected equality of these values:
>   dst[i]
>     Which is: -0.072591752
>   param.ref_dst[I]
>     Which is: -0.072591871
> [  FAILED  ] test_convolution_execute.ConvBiasAddEltwise (23 ms)

# Expected behavior
Test OK

Ryo-not-rio:
@dzarukin this seems to be affecting our nightly and precommit pipelines as well: https://github.com/oneapi-src/oneDNN/actions/runs/13128702961

vpirogov:
+@TaoLv who was looking into this.

theComputeKid:
Please remember to revert https://github.com/oneapi-src/oneDNN/pull/2599 in the same patch as the fix for this issue to show the issue is fixed. Thanks.

PR #2591, opened 2025-02-04
github: workflows: Add GitHub hosted Cobalt-100 runners

theComputeKid:
GitHub-hosted Cobalt-100 runners are now [available](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/).

It comes with the caveat of the restrictions of the free GitHub-hosted runners (size, limited runners), however, let's put this in and observe the impact it has on CI. But the advantage is that this architecture is different from the Arm-Hosted Graviton-based runners.

### DIFF
--- a/.github/automation/common_aarch64.sh
+++ b/.github/automation/common_aarch64.sh
@@ -24,6 +24,13 @@ set -o errexit -o pipefail -o noclobber
 
 export OS=$(uname)
 
+# Num threads on system.
+if [[ "$OS" == "Darwin" ]]; then
+    export MP="-j$(sysctl -n hw.ncpu)"
+elif [[ "$OS" == "Linux" ]]; then
+    export MP="-j$(nproc)"
+fi
+
 if [[ "$BUILD_TOOLSET" == "gcc" ]]; then
     export CC=gcc-${GCC_VERSION}
     export CXX=g++-${GCC_VERSION}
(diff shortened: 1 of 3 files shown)


theComputeKid:
@vpirogov This PR is ready but do we wait until https://github.com/oneapi-src/oneDNN/issues/2572 is fixed? That might take a while.

The CI job I added passes. So if we wanted to, we could merge this while we wait. I'll leave it to you to decide.

vpirogov (approved):

Ryo-not-rio (approved):

vpirogov:
> @vpirogov This PR is ready but do we wait until #2572 is fixed? That might take a while.

No need to wait for #2572 of course.

vpirogov:
It may make sense to suppress `test_graph_unit_dnnl_convolution` test as part of this PR to make CI green again though.

theComputeKid:
@vpirogov I'll create a new PR for that failure so that the person who fixes the test can revert the suppression easily, without needing to revert this PR as a whole.

vpirogov:
> @vpirogov I'll create a new PR for that failure so that the person who fixes the test can revert the suppression easily, without needing to revert this PR as a whole.

Thanks! Note, that reverts are done at a commit granularity, not a PR granularity. So as long as the change is in separate commit it can be reverted individually.

PR #2592, opened 2025-02-04
xe, benchdnn: fix large buffer concat and add nighly gpu tests 

rjoursler:
This fixes some potential index overflows in `gen9_concat`. The other concat implementations do not seem to suffer from the same integer overflow shortcomings. In addition, this adds a large buffer test suite. The large buffer tests suite encounters a few failures which mostly appear to be issues with the reorder primitive fp8 conversions and f64 conversions.  These are unrelated to large buffer support so I do not intend to fix them before merging.

This PR also contains a small bonus commit fixing the `ref_prelu` implementation.


rjoursler:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_concat

rjoursler:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_concat
enable benchdnn_prelu

dzarukin (approved):

src/gpu/intel/ocl/reusable_simple_concat.cpp
```diff
@@ -87,6 +88,8 @@ static status_t init_conf_common(impl::engine_t *engine, const concat_pd_t *pd,
     dim_t final_padding = 0;
     for (int i = 0; i < pd->n_inputs(); ++i) {
         if (pd->src_md(i)->padded_dims[concat_dim] == 0) continue;
+        max_bytes = std::max(max_bytes,
+                into<dim_t>(memory_desc_wrapper(pd->src_md(i)).size()));
```
atkassen, line 92: Was there a case that required this logic? Ignoring any striding, the destination tensor is going to be the largest tensor involved in a concat handled by this implementation just by the nature of the operation.

rjoursler: You are correct, I don't think this is technically necessary due to the current implementation requirement that src and dst have the same types.  This is largely for future proofing (and potentially edge case handling), as these restrictions can be easy to miss in testing.

atkassen (approved):

PR #2593, opened 2025-02-04
cpu: x64: matmul: correct batch layout support for int4 wei

xuxinzen:
Fixes MFDNN-13095

When `is_wei_batch_layout_trivial == true`, [B_batch_offset ](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/main/src/cpu/x64/matmul/brgemm_matmul.cpp#L1495) may not be divided by 2 and then cause correctness issue.


xuxinzen:
make test

ankalinin (approved):

densamoilov (approved):

PR #2598, opened 2025-02-04
Fix crash in sparse reorder

densamoilov:
Fixes MFDNN-13062.

Incorrect condition `if (output_d.nelems(true)) return 0;` was introduced in b20dcfcd409 resulting in returning an incorrect scratchpad size an consequently the crash.
This PR removes the condition and improves verbose diagnostics. 

The piece of code below is removed because `nelems()` can return a value that is < 0 only if md contains runtime dimensions, and that check was moved to `is_applicable()`.
```cpp
const memory_desc_wrapper dst_d(dst_md());
if (dst_d.nelems(true) < 0) return status::unimplemented
```

### DIFF
--- a/src/cpu/reorder/simple_sparse_reorder.hpp
+++ b/src/cpu/reorder/simple_sparse_reorder.hpp
@@ -77,13 +77,25 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
                 && (is_format_tag(fmt_o)
                         && (fmt_o == format_tag::any))>::type> {
 
-    static bool is_applicable(const memory_desc_wrapper &input_d,
+    static status_t is_applicable(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d, const primitive_attr_t *attr) {
-        // This reorder expects a non-plain format for destination.
-        return input_d.is_blocking_desc() && output_d.is_sparse_desc()
-                && output_d.sparse_desc().encoding == sparse_encoding::packed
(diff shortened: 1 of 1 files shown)


densamoilov:
make test

dzarukin (approved):

src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -77,18 +77,33 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
...
+
+        VDISPATCH_REORDER_IC(
+                output_d.sparse_desc().encoding == sparse_encoding::packed,
+                VERBOSE_UNSUPPORTED_SPARSE_CFG);
+
+        VDISPATCH_REORDER_IC(output_d.blocking_desc().inner_nblks > 0,
+                VERBOSE_UNSUPPORTED_SPARSE_CFG);
+
+        VDISPATCH_REORDER_IC(
+                output_d.blk_size() % 64 == 0, VERBOSE_UNSUPPORTED_SPARSE_CFG);
```
avmanerikar, line 100: (nit): 
Perhaps a more descriptive message can be printed using VERBOSE_BAD_PARAM or VERBOSE_UNSUPPORTED_FEATURE?

avmanerikar (approved):

ankalinin (approved):

src/cpu/reorder/simple_sparse_reorder.hpp
densamoilov: I made them a little more descriptive, thanks!

PR #2611, opened 2025-02-06
Handle large strides in matmul properly

densamoilov:
Fixes MFDNN-13075.

This PR adds checks for large (> INT32_MAX) strides to brgemm-based matmul primitive so that such cases can be properly handled by returning an "unimplemented".
Adding support for large strides will require extending brgemm and copy routines which doesn't seem to be a good investment at this point because the shapes are not used in real world workloads for CPU.


densamoilov:
make test

xuxinzen (approved):

ankalinin (approved):

amakarev (approved):

PR #2614, opened 2025-02-06
gpu: generic: skip unimplemented cases

sgeor255:
# Description

Currently, when running benchdnn tests on the generic SYCL backend, unimplemented cases are reported as failures. This makes it challenging to identify actual failures. It also clutters the CI logs with false-positive failure reports. Furthermore, the generic implementations are still under development, and there are several known limitations (e.g., lack of blocked format support) that should not be flagged as errors. For instance, the reorder issue reported in MFDNN-13088 currently appears as a failing test, but it is actually due to unsupported features like blocked formats and zero-points. This PR changes the status of such tests to SKIPPED, as is done with the NVIDIA and AMD backends, to prevent them from being reported as failures.

# Checklist

## General


Rbiessy (approved):
Thanks. I believe this would help us prioritize issues that are bugs instead of missing features.

AD2605 (approved):

PR #2615, opened 2025-02-06
github: workflows: Move CI into dedicated AArch64 folders

theComputeKid:
@vpirogov [suggests](https://github.com/oneapi-src/oneDNN/pull/2599#issuecomment-2637228454) to separate AArch64 CI into dedicated folders.

Requires a bit of testing, will update PR when ready. Can't check nightly is OK until this PR is merged to main, but we can test precommit.


theComputeKid:
Discovered that creating subfolders inside workflows is not going to work.. Disappointing.

theComputeKid:
This gift keeps on giving - we have "build*" in the gitignore, so I can't add build.sh to an AArch64 folder.. So now I have to add a redundant aarch64-build.sh in the aarch64 folder..

vpirogov:
> Discovered that creating subfolders inside workflows is not going to work.. Disappointing. Well, I'll keep "aarch64" in the name of all AArch64 workflows so we can regex them. And I'll continue with moving automation scripts into an AArch64 subfolder.

Looks like a lot of people are upset with this quirk: https://github.com/orgs/community/discussions/15935

vpirogov:
> This gift keeps on giving - we have "build*" in the gitignore, so I can't add build.sh to an AArch64 folder.. So now I have to add a redundant aarch64-build.sh in the aarch64 folder..

This does not sound right. There are build.sh/build.bat files sitting in .github/automation.

theComputeKid:
@vpirogov : I think the "build*" edit to gitignore was made after the scripts were already in. But no worries, I changed it to "/build*" and now it will only match build* folders in root, which is what we want.

theComputeKid:
@vpirogov : https://github.com/oneapi-src/oneDNN/commit/41c2804ba3d5e386ab9101fb9887312e0cb59eec
Here we go.

vpirogov:
> @vpirogov : I think the "build*" edit to gitignore was made after the scripts were already in. But no worries, I changed it to "/build*" and now it will only match build* folders in root, which is what we want.

My git (2.40.1) still does not bark at new files with 'build' in the name. But your fix makes sense based on [.gitignore documentation](https://git-scm.com/docs/gitignore).

theComputeKid:
@vpirogov : ready.

vpirogov:
Do you want to update [x64 CI trigger](https://github.com/oneapi-src/oneDNN/blob/main/.github/automation/.azure-pipeline.yml) as part of this PR as well?

vpirogov (approved):

theComputeKid:
@vpirogov : I tried many (10) times and still couldn't get Azure to skip, so I gave up. Could you take a look?

PR #2616, opened 2025-02-06
ci: linters: fix clang-tidy check

hmaciak:
This PR is to fix clang-tidy checks.

### DIFF
--- a/.github/workflows/pr-linter.yml
+++ b/.github/workflows/pr-linter.yml
@@ -82,7 +82,7 @@ jobs:
 
   pr-clang-tidy:
     name: Clang-Tidy
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-linter.yml
```diff
@@ -98,6 +98,8 @@ jobs:
           ONEDNN_ACTION: configure
 
       - name: Build oneDNN
-        run: .github/automation/build_linters.sh
+        run: |
+            .github/automation/build_linters.sh | tee build.log
+            grep -q "warning" build.log && exit 1 || true
```
vpirogov, line 103: Consider this:
```
grep -i "warning:" build.log | sort -u && exit 1 || true
```

Clang-tidy may generate duplicate warnings during build, so having a summary of unique warnings in the end is useful.

Also having a colon added to pattern would reduce false matches.

PR #2617, opened 2025-02-06
[backport rls-v3.7] Fix bug with conversion of mask from bf16 to float

umar456:
# Description

This PR fixes an issue with SDPA with bf16 data types with explicit masks. This issue was caused because of a copy operation from the native type(bf16) to the internal type of the kernel(float). With float16 there are implicit conversation in OpenCL but the internal data type uses ushort to represent bf16 so the conversion was incorrectly representing those values.

### DIFF
--- a/src/gpu/intel/ocl/micro_sdpa.cl
+++ b/src/gpu/intel/ocl/micro_sdpa.cl
@@ -114,10 +114,12 @@ DECLARE_2D_TILE(mask_tile_type, MSK_DATA_T, SUBGROUP_SIZE, mask_br, mask_bc,
 DECLARE_2D_TILE_BLOCK_OPS(mask_tile_type, MSK_DATA_T, SUBGROUP_SIZE, mask_br,
         mask_bc, mask_nbr, mask_nbc)
 #endif
-#endif
-
 DECLARE_2D_TILE(mask_tile_type_float, float, SUBGROUP_SIZE, mask_br, mask_bc,
         mask_nbr, mask_nbc)
+DECLARE_2D_TILE_COPY_REBLOCK(mask_tile_type, SUBGROUP_SIZE, mask_br, mask_bc,
+        mask_nbr, mask_nbc, mask_tile_type_float, SUBGROUP_SIZE, mask_br,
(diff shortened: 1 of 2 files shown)


umar456:
make test
disable benchnn_all
enable benchdnn_graph
enable test_device_gpu

h-sadia (approved):

syurkevi (approved):

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -132,12 +134,12 @@ DECLARE_2D_TILE_BLOCK2D_OPS(a_tile_type_dst, DST_DATA_T, SUBGROUP_SIZE,
...
         ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
         ugemm_vs_c_type_nblock1, a_tile_type_dst, SUBGROUP_SIZE,
-        ugemm_vs_sg_tile_m, 1, 1, ugemm_vs_sg_tile_n)
+        ugemm_vs_sg_tile_m, 1, 1, ugemm_vs_sg_tile_n, CONVERT_DATA_T)
 #else
 DECLARE_2D_TILE_COPY_REBLOCK(a_tile_type, SUBGROUP_SIZE, ugemm_vs_c_type_block0,
         ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
         ugemm_vs_c_type_nblock1, a_tile_type_dst, SUBGROUP_SIZE,
-        ugemm_vs_sg_tile_m, 8, 1, ugemm_vs_sg_tile_n / 8)
+        ugemm_vs_sg_tile_m, 8, 1, ugemm_vs_sg_tile_n / 8, CONVERT_DATA_T)
```
h-sadia, line 142: why are sending the the CONVERT_DATA_T macro along with copy functions? Are we expecting it to to cater for mixed data types?

umar456: This converts data from the internal type(f32) to the destination data type which can be half or bf16. This is the same behavior as before but I moved the conversion function into the macro argument so that we can specify how we want to perform the conversion.

PR #2620, opened 2025-02-06
xe: jit: gemm: move catalog generation to its own compilation unit

rjoursler:
GCC can take upwards of 10 minutes to compile the gemm catalog. This patch moves kcatalog generation to its own compilation unit to limit recompilations.


atkassen (approved):

rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

PR #2622, opened 2025-02-06
cpu: x64: brgemm: fixed issues with uncheked return value

amakarev:
Fix for unchecked return value of brgemm_init_tiles function.
Found by coverity scan

### DIFF
--- a/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
@@ -2672,7 +2672,8 @@ void jit_brgemm_amx_uker_base_t::generate() {
     Label permute_index_table;
     if (brg.is_input_convert() || brg.amx_wary_k_tail()) {
         // save tiles description for later use
-        brgemm_init_tiles(brg, (char *)(&palette_));
+        auto status = brgemm_init_tiles(brg, (char *)(&palette_));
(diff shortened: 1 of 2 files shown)


src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -2672,7 +2672,8 @@ void jit_brgemm_amx_uker_base_t::generate() {
     Label permute_index_table;
     if (brg.is_input_convert() || brg.amx_wary_k_tail()) {
         // save tiles description for later use
-        brgemm_init_tiles(brg, (char *)(&palette_));
+        auto status = brgemm_init_tiles(brg, (char *)(&palette_));
+        if (status != status::success) assert(!"tiles configuration failed");
```
dzarukin, line 2676: Lets get one more time over this: `assert` is not considered as a fix for coverity. Moreover, this results in a broken compilation in release mode.

Secondly, `brgemm_init_tiles` is a tricky function. The only reason one can seek for a successful status specifically here is because it involves Tmm check in the kernel definition (and an explicit check in a change below). Once this guard is gone, this code will be wrong as `unimplemented` is a valid status for non-AMX platforms.

The least this code should have is a `return` statement.

amakarev: so,I close the pull request and mark it as intentional. Thanks for clarification

PR #2626, opened 2025-02-07
cpu: Fixes some coverity issues (copy instead of move, const auto instead of const auto&).

davideberius:
cpu: Fixes some coverity issues (copy instead of move, const auto instead of const auto&).

### DIFF
--- a/src/cpu/ref_fused_convolution.hpp
+++ b/src/cpu/ref_fused_convolution.hpp
@@ -196,7 +196,7 @@ struct ref_fused_convolution_fwd_t : public primitive_t {
                 arg_cache.append_inout_arg(
                         DNNL_ARG_FROM, sp_begin, from_md, true);
                 arg_cache.append_inout_arg(DNNL_ARG_TO, sp_end, to_md, false);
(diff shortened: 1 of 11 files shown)


src/cpu/ref_fused_convolution.hpp
```diff
@@ -196,7 +196,7 @@ struct ref_fused_convolution_fwd_t : public primitive_t {
                 arg_cache.append_inout_arg(
                         DNNL_ARG_FROM, sp_begin, from_md, true);
                 arg_cache.append_inout_arg(DNNL_ARG_TO, sp_end, to_md, false);
-                args_.push_back(arg_cache);
+                args_.push_back(std::move(arg_cache));
```
xuxinzen, line 200: I did not see related coverity issue in the latest Coverity report. Could you point me out the CID or check if these changes are needed?

davideberius: There are a variety of CIDs represented here, but they are in the minor category.  This particular one is 6925574.

densamoilov (approved):

PR #2619, opened 2025-02-06
cpu: fixes coverity issues

xuxinzen:
Partially fixes MFDNN-12964

1. Check `g0`, `g1` before division
2. `nelems` can be negative, cast it into `size_t` may cause potential issue.

PR #2630, opened 2025-02-07
Backport: cpu: fixes coverity issues

xuxinzen:
Backport of #2619

### DIFF
--- a/src/cpu/reorder/simple_sparse_reorder.hpp
+++ b/src/cpu/reorder/simple_sparse_reorder.hpp
@@ -100,7 +100,7 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
 
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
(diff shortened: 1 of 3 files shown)


xuxinzen:
make test

avmanerikar (approved):

src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -100,7 +100,7 @@ struct simple_sparse_reorder_impl<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
 
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
-        const auto nelems = static_cast<size_t>(output_d.nelems(true));
```
densamoilov, line 103: `nelems` can only return < 0 when one of the dimensions is runtime because this is how wildcard defined: https://github.com/oneapi-src/oneDNN/blob/c07fe05d9c63d5a316773d055888844218931702/include/oneapi/dnnl/dnnl_types.h#L2279

This reorder doesn't support runtime dimensions so we return unimplemented before we can reach this place so it is safe to keep `size_t`.

densamoilov (approved):

PR #2647, opened 2025-02-10
src:cpu:aarch64: Enable JIT fp32 -> fp16 reorder

renato-arantes:
# Description

This PR adds support for fp16 JIT reorder for data type f32->f16. Partially solves [issue](https://github.com/oneapi-src/oneDNN/issues/2185)

## General

## Performance improvements

Below are small test case logs that demonstrate performance numbers before and after.

`./tests/benchdnn/benchdnn --reorder --sdt=f32 --ddt=f16 --mode=p 1x4096x4096`

**(new) jit:uni**: `total perf: min(ms):0.125977 avg(ms):0.136528`
**(old) simple:any**: `total perf: min(ms):10.1409 avg(ms):10.168`

### DIFF
--- a/src/cpu/aarch64/jit_uni_reorder.cpp
+++ b/src/cpu/aarch64/jit_uni_reorder.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2018-2023 Intel Corporation
 * Copyright 2020-2024 FUJITSU LIMITED
-* Copyright 2022-2024 Arm Ltd. and affiliates
+* Copyright 2022-2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Ryo-not-rio (approved):
LGTM

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -170,7 +170,8 @@ struct jit_uni_reorder_kernel_f32_t : public kernel_t, public jit_generator {
                 || (p.itype == bf16 && p.otype == f32 && mayiuse_bf16()
                         && p.beta == 0.f);
 
-        bool f16_ok = (p.itype != f16 && p.otype != f16)
+        bool is_f16 = (p.itype == f16 || p.otype == f16);
+        bool f16_ok = (p.itype == f32 && p.otype == f16 && p.beta == 0.f)
                 || (p.itype == f16 && p.otype == f32 && p.beta == 0.f);
```
mgouicem, line 175: (nit) would it make sense to align logic for bf16 and f16 for readability?
- bf16_ok includes the `IMPLICATION(is_bf16, ...)` logic through `|| (p.itype != bf16 && p.otype != bf16)`
-  but f16_ok does not include it and handles it separately with is_f16.

mgouicem (approved):

src/cpu/aarch64/jit_uni_reorder.cpp
Sqvid: @mgouicem This is a good point. I am going to merge this change as-is for now but we have more patches for this file on the way and the logic will be cleaned up there. Thanks.

PR #2577, opened 2025-01-31
ngen: remove ONEAPI_PVCSendWARWA environment variable control

petercad:
`getenv` usage in #2472 is producing build errors on Windows. Removing the environment variable control for now; will reintroduce later.

PR #2650, opened 2025-02-10
ngen: reinstate ONEAPI_PVC_SEND_WAR_WA environment variable

petercad:
Followup to #2577, reinstating the environment variable for Linux only (since it is not needed on Windows) and updating to the final name.

### DIFF
--- a/third_party/ngen/ngen_auto_swsb.hpp
+++ b/third_party/ngen/ngen_auto_swsb.hpp
@@ -1356,6 +1356,27 @@ inline BasicBlockList getBasicBlocks(HW hw, const Program &program)
         }
     }
 
+#ifndef NGEN_DISABLE_GETENV
+#ifndef _WIN32
+    // Check ONEAPI_PVC_SEND_WAR_WA environment variable.
+    static bool checkedEnv = false;
+    static bool haveEnv = false;
+    static bool envEnablePWW = true;
+
+    if (!checkedEnv) {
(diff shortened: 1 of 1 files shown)


petercad:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_conv

third_party/ngen/ngen_auto_swsb.hpp
```diff
@@ -1356,6 +1356,27 @@ inline BasicBlockList getBasicBlocks(HW hw, const Program &program)
         }
     }
 
+#ifndef NGEN_DISABLE_GETENV
+#ifndef _WIN32
+    // Check ONEAPI_PVC_SEND_WAR_WA environment variable.
+    static bool checkedEnv = false;
+    static bool haveEnv = false;
+    static bool envEnablePWW = true;
```
rjoursler, line 1364: Can we just initialize `envEnablePWW` with a lambda to avoid all the call once logic?

petercad: There are two flags to initialize (haveEnv, envEnablePWW), so we'd only be saving the `checkedEnv` flag. A lambda initialization is certainly possible, but I'm not sure it would improve readability or save lines.

kealan-barbieri (approved):

rjoursler (approved):

PR #2648, opened 2025-02-10
xe: address clang-tidy complaints

atkassen:
Addresses [MFDNN-13007](https://jira.devtools.intel.com/browse/MFDNN-13007).

PR #2654, opened 2025-02-10
common: address some clang-tidy complaints

atkassen:
Partially addresses [13015](https://jira.devtools.intel.com/browse/MFDNN-13015) with some pre-work for #2648, the most common clang-tidy hits, and some other low-hanging fruit.

### DIFF
--- a/src/common/nstl.hpp
+++ b/src/common/nstl.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2016-2024 Intel Corporation
+* Copyright 2016-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -348,9 +348,9 @@ private:
     std::vector<T> _impl;
 
 public:
-    typedef typename std::vector<T>::iterator iterator;
-    typedef typename std::vector<T>::const_iterator const_iterator;
-    typedef typename std::vector<T>::size_type size_type;
+    using iterator = typename std::vector<T>::iterator;
+    using const_iterator = typename std::vector<T>::const_iterator;
+    using size_type = typename std::vector<T>::size_type;
     vector() {}
     vector(size_type n) : _impl(n) {}
     vector(size_type n, const T &value) : _impl(n, value) {}
@@ -382,9 +382,9 @@ private:
     std::map<Key, T> _impl;
 
 public:
-    typedef typename std::map<Key, T>::iterator iterator;
-    typedef typename std::map<Key, T>::const_iterator const_iterator;
-    typedef typename std::map<Key, T>::size_type size_type;
+    using iterator = typename std::map<Key, T>::iterator;
+    using const_iterator = typename std::map<Key, T>::const_iterator;
+    using size_type = typename std::map<Key, T>::size_type;
     map() {}
     ~map() {}
     size_type size() const { return _impl.size(); }
--- a/src/common/nstl.hpp
+++ b/src/common/nstl.hpp
@@ -286,7 +286,7 @@ struct numeric_limits<int4_t> {
 };
 
 template <typename T>
-struct is_integral {
+struct is_integral { // NOLINT(readability-identifier-naming)
     static constexpr bool value = false;
 };
 template <>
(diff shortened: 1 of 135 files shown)


atkassen:
make test

src/common/utils.hpp
```diff
@@ -52,17 +52,17 @@ namespace impl {
 
 #define DNNL_SHORT_CIRCUIT_SELF_ASSIGN(other) \
     do { \
-        if (this == &other) return *this; \
+        if (this == &SILENCE_PAREN_WARNING(other)) return *this; \
```
dzarukin, line 55: How does it help to silence the warning?

src/common/z_magic.hpp
```diff
@@ -36,7 +36,8 @@
 #define PRAGMA_MACRO(x) PRAGMA_MACRo(x)
 #endif
 
-#define UNUSED(x) ((void)x)
+#define SILENCE_PAREN_WARNING(x) x
```
dzarukin, line 39: PARENT? Or PAREN(THESIS)?

echeresh (approved):

dzarukin (approved):
Thanks for addressing those numerous hits!

src/graph/backend/dnnl/kernels/kernel_base.hpp
```diff
@@ -119,10 +119,6 @@ using FCreateKernel = std::function<kernel_ptr(void)>;
 #define DEF_KERNEL_METHOD_STR(name) \
     std::string str() const override { return #name; }
 
-#define DNNL_DISALLOW_COPY_AND_ASSIGN(T) \
-    T(const T &) = delete; \
-    T &operator=(const T &) = delete;
-
```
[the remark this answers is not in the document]
atkassen: > I guess we will need to include the whole common/utils.hpp
It already is:
`graph/backend/dnnl/kernels/kernel_base.hpp` > `graph/interface/logical_tensor.hpp` > `common/type_helpers.hpp` > `common/utils.hpp`

If I put it back, the build fails.

src/common/utils.hpp
atkassen: clang-tidy doesn't complain about _some_ un-parenthesized arguments.
```c
#define macro(x) x
```
seems to be one of them. In any case, I have removed the macro since this and most of the other instances can be fixed by just adding parentheses. The exception in this PR is the change to `DNNL_DISALLOW_COPY_AND_ASSIGN`.

src/common/z_magic.hpp
atkassen: Removed the macro altogether.

atkassen:
make test

src/common/nstl.hpp
```diff
@@ -377,16 +377,16 @@ class vector : public c_compatible {
 };
 
 template <typename Key, typename T>
-class map : public c_compatible {
+class map : public c_compatible { // NOLINT(readability-identifier-naming)
```
densamoilov, line 380: (nit) I think the map and vector classes from this file are not used anywhere in the library so I think we can safely remove them.

src/common/primitive.hpp
```diff
@@ -153,22 +153,24 @@ struct nested_scratchpad_t {
 #define ARG_TYPE(t) \
```
densamoilov, line 153: Is this dead code now?

densamoilov (approved):

src/common/nstl.hpp
atkassen: It is still used [here](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_brgemm_conv.cpp#L1459-L1460), [here](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_brgemm_conv_comp_pad_kernel.cpp#L284-L285), [here](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_brgemm_conv_comp_pad_kernel.cpp#L336-L337), and [here](https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/x64/jit_brgemm_conv_comp_pad_kernel.cpp#L591-L592). If it is ok to replace with `std::vector` in those places, I will do it.

src/common/primitive.hpp
atkassen: Yes, I have combined the two macros.

atkassen:
@oneapi-src/onednn-cpu-aarch64, please review. Thanks!

Sqvid (approved):
LGTM. Reviewed `src/cpu/aarch64/**`

atkassen:
make test

PR #2658, opened 2025-02-11
benchdnn: improve memory estimation

dzarukin:
This PR brings a bunch of general improvements:
* Add "run ref" status line so that it's easier to distinguish between GPU and CPU problems when validating GPU cases with CPU reference.
* Use mprotect for --mode=R, previously it was used for --mode=C only. This brings a consistent behavior to running the test case with the library side.
* The major one - improves memory estimation for prim_ref objects. It usually happens that CPU ref primitive requires weights in blocked format and padding is possible in this case. In real extreme cases, it could lead up to 64x more memory for weights that it was estimated for reference because original estimation would consider ref memories in f32 abx format. With this PR, the logic would collect real sizes of prim_ref, combine them with the library primitive and perform a check. If it doesn't fit, current prim_ref wouldn't be used (it's allowed up to 4 prim_ref versions). If none of them fits but the library one fits with pure ref, that one will be executed.

### DIFF
--- a/tests/benchdnn/utils/task.hpp
+++ b/tests/benchdnn/utils/task.hpp
@@ -38,8 +38,11 @@ struct task_t {
         , perf_template_(perf_template)
         , idx_(idx) {}
 
-    int create() {
-        BENCHDNN_PRINT(1, "create: %s\n", prb_.str());
+    int create(bool in_parallel) {
+        // Report creation status for problems only in sequential mode as in
+        // parallel it's still not clear which one failed.
+        if (!in_parallel) BENCHDNN_PRINT(1, "create: %s\n", prb_.str());
+
         if (skip_start(&res_, idx_)) return OK;
         if (bench_mode == bench_mode_t::list) return res_.state = LISTED, OK;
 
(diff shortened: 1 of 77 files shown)


dzarukin:
make test
set test_scope=NIGHTLY

echeresh (approved):

tests/benchdnn/utils/task.hpp
```diff
@@ -43,8 +43,11 @@ struct task_t {
         , perf_template_(perf_template)
         , idx_(idx) {}
 
-    int create() {
-        BENCHDNN_PRINT(1, "create: %s\n", prb_.str());
+    int create(bool in_parallel) {
+        // Report creation status for problems only in sequanetial mode as in
+        // parallel it's still not clear which one failed.
```
rjoursler, line 48: Sequanetial is mispelled.

rjoursler (approved):

tests/benchdnn/utils/task.hpp
dzarukin: Thank you for spotting. Someday I manage vim to check spelling in comments...

dzarukin:
make test
set test_scope=NIGHTLY

densamoilov (approved):

dzarukin:
make test
set test_scope=NIGHTLY

dzarukin:
make test
set test_scope=NIGHTLY

dzarukin:
make test
set test_scope=NIGHTLY

TaoLv (approved):

PR #2660, opened 2025-02-11
graph, benchdnn: several minor fixes for SDPA with mask

ElaineBao:
# Description

This PR includes 3 minor fixes for SDPA with explicit or implicit mask:
[x] enhance check of axis in implicit causal mask to include negative axis (-1 & -2).
[x] support 1D / 2D mask for explicit causal mask (MFDNN-13164).
[x] add a internal attribute for internal op dnnl_gen_index for constant propagation (SET_ATTR_IS_CONSTANT)

### DIFF
--- a/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
@@ -139,7 +139,26 @@ status_t sdp_primitive_config_t::locate_io(std::shared_ptr<subgraph_t> &sg,
 
     if (add) {
         auto m0 = add->get_input_value(0), m1 = add->get_input_value(1);
-        attn_mask_ = in_tensor_list(m1.get(), inputs) ? m1 : m0;
+        if (in_tensor_list(m1.get(), inputs)) {
+            attn_mask_ = m1;
+        } else if (in_tensor_list(m0.get(), inputs)) {
+            attn_mask_ = m0;
+        } else if (m1->has_producer()
+                && m1->get_producer().get_kind() == op_kind::dnnl_unsqueeze
+                && in_tensor_list(
+                        m1->get_producer().get_input_value(0).get(), inputs)) {
+            // consider the case when mask is not 4D,
+            // unsqueeze op is inserted to broadcast the mask
+            attn_mask_ = m1;
+        } else if (m0->has_producer()
+                && m0->get_producer().get_kind() == op_kind::dnnl_unsqueeze
+                && in_tensor_list(
+                        m0->get_producer().get_input_value(0).get(), inputs)) {
+            attn_mask_ = m0;
+        } else {
+            VCHECK_SDP_PRIMITIVE(
+                    false, status::unimplemented, "explicit mask is not found");
+        }
     }
 
     return status::success;
--- a/tests/benchdnn/graph/custom_driver.cpp
+++ b/tests/benchdnn/graph/custom_driver.cpp
@@ -69,8 +69,8 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
 
 int execute(const prb_t *prb, const args_t &args, res_t *res) {
     dnn_mem_t &dst = const_cast<dnn_mem_t &>(args.find(DNNL_ARG_DST));
-    size_t axis = prb->axis;
     auto ndims = dst.ndims();
+    const size_t axis = prb->axis < 0 ? (prb->axis + ndims) : prb->axis;
     auto dims = dst.dims();
     auto strides = dst.strides();
 
--- a/tests/benchdnn/inputs/graph/op/harness_bf16_all
+++ b/tests/benchdnn/inputs/graph/op/harness_bf16_all
@@ -148,6 +148,7 @@
 --reset --dt=bf16 --case=op/f32/static_transpose.json
 --reset --dt=bf16 --case=op/f32/genindex.json
 --reset --dt=bf16 --op-attrs=0:axis:2 --case=op/f32/genindex.json
+--reset --dt=bf16 --op-attrs=0:axis:-1 --case=op/f32/genindex.json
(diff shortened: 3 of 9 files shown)


TaoLv (commented):
Could you please also make a PR for rls-v3.7? Let's see if we have any chance to fix them also before the release.

tests/benchdnn/inputs/graph/op/harness_bf16_all
```diff
@@ -148,6 +148,7 @@
 --reset --dt=bf16 --case=op/f32/static_transpose.json
 --reset --dt=bf16 --case=op/f32/genindex.json
 --reset --dt=bf16 --op-attrs=0:axis:2 --case=op/f32/genindex.json
+--reset --dt=f16 --op-attrs=0:axis:-1 --case=op/f32/genindex.json
```
dzarukin, line 151:
```suggestion
--reset --dt=bf16 --op-attrs=0:axis:-1 --case=op/f32/genindex.json
```

dzarukin (approved):

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -139,7 +139,28 @@ status_t sdp_primitive_config_t::locate_io(std::shared_ptr<subgraph_t> &sg,
 
     if (add) {
         auto m0 = add->get_input_value(0), m1 = add->get_input_value(1);
-        attn_mask_ = in_tensor_list(m1.get(), inputs) ? m1 : m0;
+        if (in_tensor_list(m1.get(), inputs)) {
+            attn_mask_ = m1;
+        } else if (in_tensor_list(m0.get(), inputs)) {
+            attn_mask_ = m0;
+        } else if (m1->has_producer()
+                && m1->get_producer().get_kind() == op_kind::dnnl_unsqueeze
```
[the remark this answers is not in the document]
ElaineBao: the unsqueeze comes from [binary_canonicalization pass](https://github.com/oneapi-src/oneDNN/blob/yixin/sdpa_minor_fix/src/graph/backend/dnnl/kernels/sdp_primitive.cpp#L83), when the input shapes are not the same for two inputs, unsqueeze will be inserted. this pass is needed for layout propagation.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

gyhintel (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -139,7 +139,28 @@ status_t sdp_primitive_config_t::locate_io(std::shared_ptr<subgraph_t> &sg,
...
+            attn_mask_ = m1;
+        } else {
+            VCHECK_SDP_PRIMITIVE(
+                    (m0->has_producer()
+                            && m0->get_producer().get_kind()
+                                    == op_kind::dnnl_unsqueeze
+                            && in_tensor_list(
+                                    m0->get_producer().get_input_value(0).get(),
+                                    inputs)),
+                    status::unimplemented, "explicit mask is not found");
```
TaoLv, line 161: can you please explain why we have the verbose check in this branch and what the conditions are?

tests/benchdnn/graph/custom_driver.cpp
```diff
@@ -69,8 +69,9 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
 
 int execute(const prb_t *prb, const args_t &args, res_t *res) {
     dnn_mem_t &dst = const_cast<dnn_mem_t &>(args.find(DNNL_ARG_DST));
-    size_t axis = prb->axis;
     auto ndims = dst.ndims();
+    size_t axis = prb->axis;
+    if (prb->axis < 0) { axis = prb->axis + ndims; }
```
TaoLv, line 74:
```suggestion
    const size_t axis = prb->axis < 0 ? (prb->axis + ndims) : prb->axis;
```

BTW, since it is passed by the user through benchdnn command line, do we need to verify `axis` in [0, ndims)?

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
ElaineBao: because if unsqueeze is not in add's post-src (checked in previous `else if`), the only possibility is to have it in add's first src, so this verbose check should always be true, except some unexpected pattern happens

tests/benchdnn/graph/custom_driver.cpp
ElaineBao: axis can be [-ndims, ndims), so I think negative values of axis is acceptable. And whether we need to verify the user input to be in the range of [-ndims, ndims), my understanding is that user should follow [our document](https://oneapi-src.github.io/oneDNN/dev_guide_op_genindex.html) to set the value of this attribute

ElaineBao: OK, let me add a check for axis :)
But I feel it's more useful to have such check at library side instead of benchdnn side. If we want to add a check at library side, we can only check at compilation stage (since ndims may not exist when get_partition).

TaoLv: Thanks and yes, we need to have the similar check on library side at some places. But here i feel it's actually custom driver for reference.

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
ElaineBao: fixed, thanks!

tests/benchdnn/graph/custom_driver.cpp
ElaineBao: fixed, thanks!

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2662, opened 2025-02-11
gpu: amd: add support for rocm6.1

sgeor255:
# Description

This PR enables compilation of oneDNN with rocm6.1 by updating the cmake configuration to find rocblas and miopen. Additionally, the `int8x4` datatype was removed in miopen version 3, so this PR adds a check for the miopen version and disables support for the `int8x4` datatype.

### DIFF
--- a/src/gpu/amd/README.md
+++ b/src/gpu/amd/README.md
@@ -3,14 +3,15 @@
 ## General information
 
 Support for AMD backend is implemented via SYCL HIP backend. The feature is
-disabled by default. Users must enable it at build time with a CMake option
-`DNNL_GPU_VENDOR=AMD`. The AMD GPUs can be used via oneDNN engine abstraction.
-The engine should be created using `dnnl::engine::kind::gpu` engine kind or the
-user can provide a `sycl::device` objects that corresponds to AMD GPUs.
+disabled by default and is currently experimental. You must enable it at the
+build time with a CMake option `DNNL_GPU_VENDOR=AMD`. The AMD GPUs can be used
+via oneDNN engine abstraction. The engine should be created using
+`dnnl::engine::kind::gpu` engine kind or you can provide `sycl::device` objects
+that correspond to AMD GPUs.
 
 ## Pre-requisites
 * [oneAPI DPC++ Compiler with support for HIP AMD](https://github.com/intel/llvm/blob/sycl/sycl/doc/GetStartedGuide.md#build-dpc-toolchain-with-support-for-hip-amd), version [2022-12](https://github.com/intel/llvm/releases/tag/2022-12)
-* [AMD ROCm](https://github.com/RadeonOpenCompute/ROCm), version 5.3 or newer
+* [AMD ROCm](https://github.com/RadeonOpenCompute/ROCm), version 5.3 or newer. The latest supported version currently is 6.1.
 * [MIOpen](https://github.com/ROCmSoftwarePlatform/MIOpen), version 2.18 or newer (optional if AMD ROCm includes the required version of MIOpen)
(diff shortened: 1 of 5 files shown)


src/gpu/amd/sycl_hip_utils.hpp
```diff
@@ -131,14 +131,19 @@ inline status_t convert_data_type(const memory_desc_t *mem_desc,
...
+            if (vectorized
+                    && mem_desc->format_desc.blocking.inner_blks[0] == 4) {
+#if MIOPEN_HAS_INT8X4
+                *miopen_data_type = miopenDataType_t::miopenInt8x4;
+#else
+                return status::unimplemented;
+#endif
+            } else {
+                *miopen_data_type = miopenDataType_t::miopenInt8;
+            }
```
dzarukin, line 144: Would `miopenDataType_t::miopenInt8` work for `vectorized && mem_desc->format_desc.blocking.inner_blks[0] == 4` condition?
If yes, maybe change the scope of the macro to avoid `unimplemented` when the condition is hit...

sgeor255: I had a look at convolution (as the miopen documentation mentions int8x4 is used primarily for convolution forward pass) and with the current implementation I believe it won't work as an miopen blocked format would need to be used for those cases, which is currently not supported in the oneDNN miopen implementations. I think it's okay to return unimplemented for now and when blocked format support is added this could be enabled, but let me know what you think @dzarukin .

dzarukin: I'm fine with the solution, just tried to iterate if we could save people (if there are any) from upgrading the ROCm version and getting surprise unimplemented statuses in cases which worked before.

dzarukin (approved):

vpirogov (approved):

src/gpu/amd/README.md
```diff
@@ -3,14 +3,15 @@
 ## General information
 
 Support for AMD backend is implemented via SYCL HIP backend. The feature is
-disabled by default. Users must enable it at build time with a CMake option
-`DNNL_GPU_VENDOR=AMD`. The AMD GPUs can be used via oneDNN engine abstraction.
-The engine should be created using `dnnl::engine::kind::gpu` engine kind or the
-user can provide a `sycl::device` objects that corresponds to AMD GPUs.
+disabled by default and is currently experimental. Users must enable it at build
```
ranukund, line 6:
```suggestion
disabled by default and is currently experimental. You must enable it at the build
```

src/gpu/amd/README.md
```diff
@@ -3,14 +3,15 @@
...
+time with a CMake option `DNNL_GPU_VENDOR=AMD`. The AMD GPUs can be used via
+oneDNN engine abstraction. The engine should be created using
+`dnnl::engine::kind::gpu` engine kind or the user can provide a `sycl::device`
```
ranukund, line 9:
```suggestion
`dnnl::engine::kind::gpu` engine kind or you can provide `sycl::device`
```

src/gpu/amd/README.md
```diff
@@ -3,14 +3,15 @@
...
+objects that corresponds to AMD GPUs.
```
ranukund, line 10:
```suggestion
objects that correspond to AMD GPUs.
```

ranukund (approved):
Minor edits suggested, thanks

src/gpu/amd/README.md
sgeor255: Done, thanks.

src/gpu/amd/README.md
sgeor255: Done, thanks.

Rbiessy (approved):

PR #2666, opened 2025-02-11
gtests: fix test_iface_sparse fails for mem_debug (fixes MFDNN-13144)

avmanerikar:
# Description

Fixes MFDNN-13144.

With `-DDNNL_ENABLE_MEM_DEBUG=ON`, [test_iface_sparse](https://github.com/oneapi-src/oneDNN/blob/main/tests/gtests/test_iface_sparse.cpp) crashes due to a mismatch between failed `malloc` counts during memory object instantiation. The PR adds a fix for the expected failure.

### DIFF
--- a/tests/gtests/test_iface_sparse.cpp
+++ b/tests/gtests/test_iface_sparse.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2024 Intel Corporation
+* Copyright 2022-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -196,7 +196,7 @@ TEST(iface_sparse_test_t, TestSparseMDSize) {
     ASSERT_EQ(md.get_size(2), 0u);
 }
 
-TEST(iface_sparse_test_t, TestSparseMemoryCreation) {
+HANDLE_EXCEPTIONS_FOR_TEST(iface_sparse_test_t, TestSparseMemoryCreation) {
     engine eng = get_test_engine();
 
     const bool is_unimplemented = (eng.get_kind() == engine::kind::gpu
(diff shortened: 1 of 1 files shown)


avmanerikar:
make test
enable build_experimental
enable build_experimental_sparse
disable benchdnn_all
enable benchdnn_matmul

tests/gtests/test_iface_sparse.cpp
```diff
@@ -212,7 +217,7 @@ TEST(iface_sparse_test_t, TestSparseMemoryCreation) {
     memory mem;
 
     // Default memory constructor.
-    EXPECT_NO_THROW(mem = memory(md, eng));
+    EXPECT_OUT_OF_MEM(mem = memory(md, eng));
```
dzarukin, line 220: We have `HANDLE_EXCEPTIONS_FOR_TEST` to replace `TEST` instead of introducing a dedicated lambda for each memory object creation. Just remove the test from validating for failed memory and that's it.

avmanerikar: Thanks for the suggestion. This has been fixed in the latest update.

avmanerikar:
make test
disable benchdnn_all
enable benchdnn_matmul
enable build_experimental
enable build_experimental_sparse

dzarukin (approved):

tests/gtests/test_iface_sparse.cpp
```diff
@@ -292,20 +297,23 @@ TEST(iface_sparse_test_t, TestSparseMemorySetGetDataHandles) {
...
-    // Creating a memory object without underlying buffers.
-    for (int i = 0; i < nhandles; i++) {
-        EXPECT_NO_THROW(mem.set_data_handle(DNNL_MEMORY_NONE, i));
-    }
+    mem = memory(md, eng);
 
-    for (int i = 0; i < nhandles; i++) {
-        void *h = mem.get_data_handle(i);
-        ASSERT_EQ(h, nullptr);
+    {
```
densamoilov, line 302: nit: putting the piece of code in a separate block doesn't give us anything here because we don't create any object within it

densamoilov (approved):

tests/gtests/test_iface_sparse.cpp
avmanerikar: Thanks for the comment. I have added the block to take care of initialization errors for the memory object `mem` as it is called to get and set the data handles.

avmanerikar:
make test Lightweight_scans

PR #2678, opened 2025-02-12
x64: pool: Accuracy issue in benchnn pooling test

jstachowintel:
It is similar use case as https://github.com/oneapi-src/oneDNN/pull/2627. The problem is with huge input sizes, so I decided disable `src_size <= INT_MAX` for jit:avx512_core implementation.

### DIFF
--- a/src/gpu/intel/jit/gemm/generator/pieces/k_loop.cxx
+++ b/src/gpu/intel/jit/gemm/generator/pieces/k_loop.cxx
@@ -871,35 +871,70 @@ void BLASKernelGenerator<hw>::kLoop(KLoop type, const GEMMProblem &problem, GEMM
     if (remaskA && remaskB && loadBFirst)
         ls.swapLast2();
 
-    // A/B quantization parameter repacking.
+    // A/B quantization parameter repacking and remasking.
     auto reqRepackAq = every(kaq_load);
     auto reqRepackBq = every(kbq_load);
     auto reqRepackAqLate = every(kaq_loadLate);
     auto reqRepackBqLate = every(kbq_loadLate);
 
-    bool remaskAs = as2D && (minOPCount > 1) && (problem.aqGroupK == 1);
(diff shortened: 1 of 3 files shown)


src/cpu/x64/jit_uni_pooling.hpp
```diff
@@ -75,6 +75,13 @@ struct jit_uni_pooling_fwd_t : public primitive_t {
             CHECK(jit_uni_pool_kernel<isa>::init_conf(
                     jpp_, scratchpad, attr_, this));
 
+            dim_t src_size = static_cast<dim_t>(jpp_.mb) * jpp_.c * jpp_.id
```
rjoursler, line 78: Why do we need to type cast? Is this check just going to fail if we happen to get problem dimensions exceeding `INT_MAX`?

jstachowintel: If the product of our input sizes exceeds `INT_MAX`, an overflow will occur, and we want to skip this implementation. We need to use casting to get the correct, not overflowed number that can be compared with `INT_MAX`.

rjoursler: This issue is that `jpp_.mb` may have already "overflowed". As I understand this code,  all the problem dimensions are set [here](https://github.com/oneapi-src/oneDNN/blob/7437899ec01fb323d99e8e50626a046ca7a70d83/src/cpu/x64/jit_uni_pool_kernel.cpp#L155-L164).   The `src_d` memory descriptor returns a `dim_t` which is truncated to an for storing in `int` for storing in `jpp`. If we set one of those dimensions to something like `jpp.mb` `UINT_MAX + 1`, then `src_size` will be zero and the check will pass.  It seem better to just directly check the memory descirptor sizes [similar to here](https://github.com/oneapi-src/oneDNN/pull/2670) or to change the dimensions types to `dim_t` so that no cast is necessary.

PR #2679, opened 2025-02-12
x64: conv: fix huge shape limitation calculation

tczeszun:
small update to #2663

### DIFF
--- a/src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
+++ b/src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
@@ -3912,12 +3912,14 @@ status_t jit_avx512_core_amx_bwd_data_kernel_t::init_conf(jit_conv_conf_t &jcp,
             = jcp.with_bias ? types::data_type_size(bias_d.data_type()) : 0;
     jcp.typesize_acc = sizeof(int32_t);
 
-    dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw * jcp.ic
+    const dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw
+            * (jcp.is_nspc ? jcp.ic : rnd_up(jcp.ic, jcp.ic_block))
             * jcp.typesize_in;
     VDISPATCH_CONV_IC(src_size <= INT_MAX, VERBOSE_UNSUPPORTED_FEATURE,
             "src size > INT_MAX is not supported");
 
-    dim_t dst_size = (dim_t)jcp.mb * jcp.od * jcp.oh * jcp.ow * jcp.oc
+    const dim_t dst_size = (dim_t)jcp.mb * jcp.od * jcp.oh * jcp.ow
+            * (jcp.is_nspc ? jcp.oc : rnd_up(jcp.oc, jcp.oc_block))
             * jcp.typesize_out;
     VDISPATCH_CONV_IC(dst_size <= INT_MAX, VERBOSE_UNSUPPORTED_FEATURE,
             "dst size > INT_MAX is not supported");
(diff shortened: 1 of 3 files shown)


xuxinzen (approved):

densamoilov (approved):

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
```diff
@@ -3912,12 +3912,14 @@ status_t jit_avx512_core_amx_bwd_data_kernel_t::init_conf(jit_conv_conf_t &jcp,
...
 
-    dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw * jcp.ic
+    const dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw
+            * (jcp.is_nspc ? jcp.ic : rnd_up(jcp.ic, jcp.ic_block))
             * jcp.typesize_in;
     VDISPATCH_CONV_IC(src_size <= INT_MAX, VERBOSE_UNSUPPORTED_FEATURE,
             "src size > INT_MAX is not supported");
 
-    dim_t dst_size = (dim_t)jcp.mb * jcp.od * jcp.oh * jcp.ow * jcp.oc
+    const dim_t dst_size = (dim_t)jcp.mb * jcp.od * jcp.oh * jcp.ow
```
rjoursler, line 3921: The fact that there is a cast to `dim_t` seems to signal that there are other issues here. Are there guards to ensure that the dimension variables already fit in `int`? In general, it seems like it would be better to change the dimensions data types to `dim_t` so no cast is necessary or to just directly query the memory descriptor similar to what is done in #2276.

tczeszun: @rjoursler I was testing it with `-ftrapv` enabled, so it detected potential int overflow places during benchdnn run and I've fixed these places already in my previous commits regarding similar issues.

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
```diff
@@ -3912,12 +3912,14 @@ status_t jit_avx512_core_amx_bwd_data_kernel_t::init_conf(jit_conv_conf_t &jcp,
             = jcp.with_bias ? types::data_type_size(bias_d.data_type()) : 0;
     jcp.typesize_acc = sizeof(int32_t);
 
-    dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw * jcp.ic
+    const dim_t src_size = (dim_t)jcp.mb * jcp.id * jcp.ih * jcp.iw
```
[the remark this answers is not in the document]
tczeszun: @densamoilov thanks, I added new commit that fixes this. I may squash if needed.

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
rjoursler: >  with -ftrapv enabled, so it detected potential int overflow places during benchdnn run

That is not a sufficient test, it will miss overflow due to [implicit conversions](https://godbolt.org/z/YjY5eYPdY), which is exactly what is occurring [here](https://github.com/oneapi-src/oneDNN/blob/28685a6053a7f1b70c38d2c4f2c5514cf47e5858/src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp#L2287) as we are implicitly converting `dim_t` to `int`.

tczeszun: @rjoursler I see, thanks for clarifying. I'll keep this in mind to spend some time on this and I'll open new PR with md query. And I'll push this PR as is, as it does not break anything.

rjoursler (approved):

PR #2684, opened 2025-02-13
x64: conv: fix int overflows (fixes MFDNN-13152)

tczeszun:
fixes [MFDNN-13152](https://jira.devtools.intel.com/browse/MFDNN-13152)

### DIFF
--- a/src/cpu/x64/jit_avx512_core_bf16_convolution.cpp
+++ b/src/cpu/x64/jit_avx512_core_bf16_convolution.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


densamoilov (approved):

xuxinzen (approved):

src/cpu/x64/jit_avx512_core_bf16_convolution.cpp
```diff
@@ -1501,8 +1501,9 @@ void jit_avx512_core_bf16_convolution_bwd_weights_t ::compute_diff_weights(
             jcp.src_tag, format_tag::nwc, format_tag::nhwc, format_tag::ndhwc);
     const bool is_ddst_layout_nxc = utils::one_of(
             jcp.dst_tag, format_tag::nwc, format_tag::nhwc, format_tag::ndhwc);
-    const int wei_size = jcp.ngroups * jcp.nb_oc * jcp.oc_block * jcp.nb_ic
-            * jcp.ic_block * jcp.kh * jcp.kw * jcp.kd;
+    const int wei_size = static_cast<size_t>(jcp.ngroups) * jcp.nb_oc
```
dzarukin, line 1504: Casting to `size_t` to put the value into `int` might be incorrect. Is it intentional? Same one below.

tczeszun: changed type to `size_t`, test passed after this change (`-ftrapv` did not catch anything)

PR #2686, opened 2025-02-13
aarch64: matmul: addition of JIT int8 kernel

Shreyas-fuj:
# Description

This PR introduces the quantised(int8) matmul kernel in OneDNN which gives good speed-ups for quantised model inference on Graviton 3 CPUs.

Kernel supports:

1. Shapes – 2d,3d,4d

2. Zeropoint
- Source – common(s32)
- Weight – common(s32)
- Destination – common(s32)
3. Scales
- Source – common(f32)
- Weight – common(f32),per_oc(f32)
- Destination - common(f32)
4. Bias – mask - 1xN(f32)

5. Dynamic quantisation of Source matrix from f32->int8 based on common policy for scales and zp(This feature is available when the flag `ONEDNN_AARCH64_MATMUL_SRC_QUANT` is enabled at runtime by : 
`export ONEDNN_AARCH64_MATMUL_SRC_QUANT=ON`

# Checklist

## General

- [y] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
```
make test

97% tests passed, 6 tests failed out of 219

Total Test time (real) = 4218.29 sec

The following tests FAILED:
	162 - test_graph_unit_dnnl_convolution_cpu (Failed)
	168 - test_graph_unit_dnnl_large_partition_cpu (Failed)
	191 - test_benchdnn_modeC_binary_ci_cpu (Failed)
	192 - test_benchdnn_modeC_binary_different_dt_ci_cpu (Failed)
	200 - test_benchdnn_modeC_graph_ci_cpu (Failed)
	214 - test_benchdnn_modeC_self_ci_cpu (Failed)
Errors while running CTest
Output from these tests are in: /home/shreyas/G/shr-fuj/oneDNN_open_source/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
make: *** [Makefile:71: test] Error 8
```
- [y] Have you formatted the code using clang-format?

### DIFF
--- a/include/oneapi/dnnl/dnnl_types.h
+++ b/include/oneapi/dnnl/dnnl_types.h
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2016-2025 Intel Corporation
-* Copyright 2024 FUJITSU LIMITED
+* Copyright 2024-2025 FUJITSU LIMITED
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -1047,6 +1047,9 @@ typedef enum {
     dnnl_aCBdef8b8c,
     dnnl_abdEC16e4c,
     dnnl_abDC16d4c,
+    dnnl_BA24b8a,
+    dnnl_aCB24c8b,
+    dnnl_abDC24d8c,
 
     /// Just a sentinel, not real memory format tag. Must be changed after new
     /// format tag is added.
--- a/src/cpu/aarch64/matmul/jit_int8_kernel_types.hpp
+++ b/src/cpu/aarch64/matmul/jit_int8_kernel_types.hpp
@@ -1,0 +1,94 @@
+/*******************************************************************************
+* Copyright 2025 FUJITSU LIMITED
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef CPU_AARCH64_JIT_INT8_KERNEL_TYPES_HPP
+#define CPU_AARCH64_JIT_INT8_KERNEL_TYPES_HPP
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+namespace matmul {
+
+typedef enum {
+    none = 0,
+    per_tensor = 1,
+    per_m = 2,
+    per_n = 3,
+    per_k = 4,
+} jit_int8_broadcast_t;
+
+struct dyn_vals_t {
+    int f = 0;
+    dim_t M = 0;
+    dim_t K = 0;
+    dim_t N = 0;
+    dim_t B = 0;
+    int is_s8 = 0, is_u8 = 0;
+    int mtail, ktail, ntail, m_blk, k_blk, n_blk;
+    int get_min_max = 0, reorder_a = 0, reorder_b = 0, cal_src = 0;
+    int is_mtail = 0, is_ktail = 0;
+};
+
+struct dyn_params_t {
+    const float *dyn_src;
+    const int8_t *src;
+    int8_t *dst;
+    float *max, *min;
+    int *nk, *nm, *nn;
+    int *tl, *mtl, *ntl;
+};
+
+struct brg_int8_t {
+    int M, K, N;
+    const int m_blk = 8, n_blk = 4, k_blk = 8;
+    const int ld_block = 6, rd_block = 4, bd_block = 8;
+    int na, nb;
+    int m_tail, n_tail, k_tail;
+    int is_m_tail, is_k_tail, is_n_tail, is_zp_cal;
+    int dst_dt_sz;
+    bool is_s8;
+    bool is_bias;
+    bool with_scales;
+    bool with_dst_scales;
+    bool is_oc_scales;
+    jit_int8_broadcast_t zp_type_a = jit_int8_broadcast_t::none;
+    jit_int8_broadcast_t zp_type_b = jit_int8_broadcast_t::none;
+    jit_int8_broadcast_t zp_type_c = jit_int8_broadcast_t::none;
+    bool is_zp_b_int8 = false;
+    bool b_reo = true;
+    data_type_t zp_b_dt;
+    dim_t B;
+};
+
+struct call_params_t {
+    const uint8_t *src, *wei;
+    float *dst;
+    const float *bias, *scales, *dst_scales;
+    dim_t M, K, N;
+    char *buf_B_ptr_;
+    int *na, *nb;
+    int32_t *src_zero_point, *wei_zero_point, *dst_zero_point;
+    const int8_t *wei_zero_point_buf;
+    float *zp_a_ptr, *zp_b_ptr;
+};
+
+} // namespace matmul
+} // namespace aarch64
+} // namespace cpu
+} // namespace impl
+} // namespace dnnl
+#endif
--- a/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
+++ b/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
@@ -1,0 +1,1478 @@
+/*******************************************************************************
+* Copyright 2025 FUJITSU LIMITED
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cassert>
+
+#include <math.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <algorithm>
+#include <iostream>
+#include <iterator>
+#include <ranges>
+#include <vector>
+
+#include "common/c_types_map.hpp"
+#include "common/dnnl_thread.hpp"
+#include "common/math_utils.hpp"
+#include "common/memory_tracking.hpp"
+#include "common/nstl.hpp"
+#include "common/tag_traits.hpp"
+#include "common/type_helpers.hpp"
+#include "common/utils.hpp"
+#include "cpu/aarch64/jit_generator.hpp"
+#include "cpu/cpu_primitive.hpp"
+#include "cpu/matmul/matmul_utils.hpp"
+#include "cpu/scale_utils.hpp"
+
+#include "cpu/platform.hpp"
+#include "cpu/primitive_attr_postops.hpp"
+
+#include "cpu/aarch64/matmul/jit_int8_kernel_types.hpp"
+#include "cpu/aarch64/matmul/jit_int8_matmul.hpp"
+#include "cpu/aarch64/matmul/jit_int8_matmul_utils.hpp"
+
+#define GET_OFF(field) (uint32_t) offsetof(call_params_t, field)
+
+#define LDR_IMM(reg, addr, off) \
+    { \
+        const uint64_t IMM12_MASK = ~uint64_t(0xfff); \
+        if ((off & IMM12_MASK) == 0) { \
+            ldr(reg, ptr(addr, off)); \
+        } else { \
+            add_imm(X_DEFAULT_ADDR, addr, off, X_TMP_0); \
+            ldr(reg, ptr(X_DEFAULT_ADDR)); \
+        } \
+    }
+
+#define VCHECK_BG(f, msg, ...) \
+    VCHECK(primitive, create, dispatch, brgemm_matmul, f, msg, ##__VA_ARGS__);
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+namespace matmul {
+
+using namespace Xbyak_aarch64;
+using namespace dnnl::impl::cpu::matmul;
+using namespace dnnl::impl::format_tag;
+using namespace dnnl::impl::memory_tracking::names;
+using namespace dnnl::impl::utils;
+
+using namespace nstl;
+
+using namespace data_type;
+
+struct jit_int8_matmul_kernel_t : public jit_generator {
+
+    DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_int8_matmul_kernel_t)
+
+    XReg reg_param = abi_param1;
+    XReg reg_a = x3;
+    XReg reg_b = x4;
+    XReg reg_c = x5;
+    XReg reg_aux_a = x6;
+    XReg reg_aux_b = x7;
+    XReg reg_aux_c = x8;
+    XReg reg_aux_a1 = x9;
+    XReg reg_zp_aux_b_buf = x10;
+    XReg reg_aux_c1 = x11;
+    XReg reg_ld_loop = x12;
+    XReg reg_rd_loop = x13;
+    XReg reg_bd_loop = x14;
+    XReg reg_tmp = x15;
+    XReg reg_tmp_1 = x16;
+    XReg reg_bias = x17;
+    XReg reg_zp_a = x18;
+
+    XReg reg_scales = x20;
+    XReg reg_aux_scales = x24; //used X_TMP_1
+    XReg reg_na = x25; //used X_TMP_2
+    XReg reg_zp_b = x26; //used X_TMP_3
+    XReg reg_zp_aux_b = x27; //used X_TMP_4
+    PReg prd_ld = p1;
+    PReg prd_st = p2;
+    PReg prd_b = p3;
+    PReg prd_8 = p4;
+    PReg prd_zp_b_tl = p5;
+    XReg reg_zp_val_c = x2;
+
+    XReg reg_zp_val_a = reg_scales;
+    XReg reg_zp_val_b = reg_aux_scales;
+
+    call_params_t inp;
+
+    void operator()(const call_params_t *p) {
+        return jit_generator::operator()(p);
+    }
+
+    ZReg loadb(int ld) { return ZReg(ld + 1); }
+    ZReg acc(int bd, int ld) {
+        return ZReg(bd * brg_.ld_block + ld + brg_.ld_block + 1);
+    }
+    void zero_regs() {
+        for (int a = 0; a < brg_.bd_block / 2; a++)
+            for (int b = 0; b < brg_.ld_block; b++)
+                eor(acc(a, b).d, acc(a, b).d, acc(a, b).d);
+    }
+    void store_regs(int bdb, int ldb, int tail) {
+        for (int a = 0; a < bdb; a++) {
+            for (int b = 0; b < ldb; b++) {
+                if (brg_.is_s8)
+                    scvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+                else
+                    ucvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+            }
+        }
+
+        for (int a = 0; a < bdb; a++) {
+            for (int b = 0; b < ldb; b += 2) {
+                if (b + 1 < ldb) {
+                    uzp1(z31.d, acc(a, b).d, acc(a, b + 1).d);
+                    uzp2(acc(a, b + 1).d, acc(a, b).d, acc(a, b + 1).d);
+                    mov(acc(a, b).d, z31.d);
+                } else {
+                    uzp1(z31.d, acc(a, b).d, acc(a, b).d);
+                    uzp2(acc(a, b + 1).d, acc(a, b).d, acc(a, b).d);
+                    mov(acc(a, b).d, z31.d);
+                }
+            }
+        }
+
+        if (brg_.zp_type_a != jit_int8_broadcast_t::none) {
+            for (int b = 0; b < ldb; b += 2) {
+                PReg p = (brg_.is_n_tail && b >= ldb - 2) ? prd_b : P_ALL_ONE;
+                ld1w(z31.s, p, ptr(reg_zp_a, b / 2, MUL_VL));
+                for (int a = 0; a < bdb; a++) {
+                    fsub(acc(a, b).s, acc(a, b).s, z31.s);
+                    fsub(acc(a, b + 1).s, acc(a, b + 1).s, z31.s);
+                }
+            }
+        }
+
+        if (brg_.zp_type_b != jit_int8_broadcast_t::none) {
+            int ao = 0;
+            if (brg_.is_zp_b_int8) {
+                mov(reg_tmp_1, reg_zp_aux_b_buf);
+                int ilp = (brg_.is_n_tail) ? n_blks : 3;
+                for (int i = 0; i < ilp; i++) {
+                    PReg p = (brg_.is_n_tail && i == ilp - 1) ? prd_zp_b_tl
+                                                              : prd_8;
+                    ld1b(ZRegB(i + 1), p, ptr(reg_tmp_1));
+                    if (brg_.zp_b_dt == u8) {
+                        uunpklo(ZRegH(i + 1), ZRegB(i + 1));
+                        uunpklo(ZRegS(i + 1), ZRegH(i + 1));
+                        ucvtf(ZRegS(i + 1), P_ALL_ONE, ZRegS(i + 1));
+                    } else {
+                        sunpklo(ZRegH(i + 1), ZRegB(i + 1));
+                        sunpklo(ZRegS(i + 1), ZRegH(i + 1));
+                        scvtf(ZRegS(i + 1), P_ALL_ONE, ZRegS(i + 1));
+                    }
+                    add_imm(reg_tmp_1, reg_tmp_1, 8, X_TMP_0);
+                }
+            }
+            for (int a = 0; a < bdb; a++) {
+                ld1rw(z31.s, P_ALL_ONE, ptr(reg_zp_aux_b, ao * 4));
+                ld1rw(z0.s, P_ALL_ONE, ptr(reg_zp_aux_b, (ao + 1) * 4));
+                for (int b = 0; b < ldb; b += 2) {
+                    if (brg_.is_zp_b_int8) {
+                        fmul(z4.s, z31.s, ZRegS(b / 2 + 1));
+                        fmul(z5.s, z0.s, ZRegS(b / 2 + 1));
+                        fsub(acc(a, b).s, acc(a, b).s, z4.s);
+                        fsub(acc(a, b + 1).s, acc(a, b + 1).s, z5.s);
+                    } else {
+                        fsub(acc(a, b).s, acc(a, b).s, z31.s);
+                        fsub(acc(a, b + 1).s, acc(a, b + 1).s, z0.s);
+                    }
+                }
+                ao += 2;
+            }
+        }
+
+        if (brg_.with_scales) {
+            for (int b = 0; b < ldb; b += 2) {
+                PReg p = (brg_.is_n_tail && b >= ldb - 2) ? prd_b : P_ALL_ONE;
+                if (brg_.is_oc_scales) {
+                    ld1w(z31.s, p, ptr(reg_scales, b / 2, MUL_VL));
+                } else {
+                    ld1w(z31.s, p, ptr(reg_scales));
+                }
+
+                for (int a = 0; a < bdb; a++) {
+                    fmul(acc(a, b).s, acc(a, b).s, z31.s);
+                    fmul(acc(a, b + 1).s, acc(a, b + 1).s, z31.s);
+                }
+            }
+        }
+
+        if (brg_.is_bias) {
+            for (int b = 0; b < ldb; b += 2) {
+                PReg p = (brg_.is_n_tail && b >= ldb - 2) ? prd_b : P_ALL_ONE;
+                ld1w(z31.s, p, ptr(reg_bias, b / 2, MUL_VL));
+                for (int a = 0; a < bdb; a++) {
+                    fadd(acc(a, b).s, acc(a, b).s, z31.s);
+                    fadd(acc(a, b + 1).s, acc(a, b + 1).s, z31.s);
+                }
+            }
+        }
+
+        if (brg_.with_dst_scales) {
+            ld1rw(z31.s, P_ALL_ONE, ptr(reg_aux_scales));
+            for (int b = 0; b < ldb; b += 2) {
+                for (int a = 0; a < bdb; a++) {
+                    fmul(acc(a, b).s, acc(a, b).s, z31.s);
+                    fmul(acc(a, b + 1).s, acc(a, b + 1).s, z31.s);
+                }
+            }
+        }
+
+        if (brg_.zp_type_c != jit_int8_broadcast_t::none) {
+            LDR_IMM(reg_zp_val_c, reg_param, GET_OFF(dst_zero_point));
+            ldr(W_TMP_0, ptr(reg_zp_val_c));
+            dup(z0.s, W_TMP_0);
+            scvtf(z0.s, P_ALL_ONE, z0.s);
+            for (int b = 0; b < ldb; b += 2) {
+                for (int a = 0; a < bdb; a++) {
+                    fadd(acc(a, b).s, acc(a, b).s, z0.s);
+                    fadd(acc(a, b + 1).s, acc(a, b + 1).s, z0.s);
+                }
+            }
+        }
+
+        mov(reg_tmp, reg_aux_c);
+        add_imm(reg_tmp_1, reg_aux_c, brg_.N * brg_.dst_dt_sz, X_TMP_0);
+        for (int a = 0; a < bdb; a++) {
+            for (int b = 0; b < ldb; b += 2) {
+                PReg p = (brg_.is_n_tail && b >= ldb - 2) ? prd_st : P_ALL_ONE;
+                int vl = b / 2;
+                st1w(acc(a, b).s, p, ptr(reg_tmp, vl, MUL_VL));
+                if (a >= bdb - 1 && brg_.is_m_tail) {
+                    if (brg_.m_tail % 2 == 0)
+                        st1w(acc(a, b + 1).s, p, ptr(reg_tmp_1, vl, MUL_VL));
+                } else {
+                    st1w(acc(a, b + 1).s, p, ptr(reg_tmp_1, vl, MUL_VL));
+                }
+            }
+            add_imm(reg_tmp, reg_tmp, 2 * brg_.N * brg_.dst_dt_sz, X_TMP_0);
+            add_imm(reg_tmp_1, reg_tmp_1, 2 * brg_.N * brg_.dst_dt_sz, X_TMP_0);
+        }
+    }
+
+    void microkernel(int rdb, int bdb, int ldb, int tail) {
+        int a_off = 0, rd, ld, bd;
+        mov(reg_tmp, reg_aux_b);
+        for (rd = 0; rd < rdb; rd++) {
+            int ao = 0;
+
+            for (ld = 0; ld < ldb; ld++) {
+                PReg p = (brg_.is_n_tail && ld == ldb - 1) ? prd_ld : P_ALL_ONE;
+                ld1b(loadb(ld).b, p, ptr(reg_tmp, ld, MUL_VL));
+            }
+            for (bd = 0; bd < bdb; bd++) {
+                add_imm(X_DEFAULT_ADDR, reg_aux_a, a_off + ao, X_TMP_0);
+                ld1rqb(z0.b, P_ALL_ONE, ptr(X_DEFAULT_ADDR));
+                ao += brg_.m_blk * 2;
+
+                for (ld = 0; ld < ldb; ld++) {
+                    if (brg_.is_s8)
+                        smmla(acc(bd, ld).s, z0.b, loadb(ld).b);
+                    else
+                        ummla(acc(bd, ld).s, z0.b, loadb(ld).b);
+                }
+            }
+            a_off += brg_.m_blk * brg_.k_blk;
+            add_imm(reg_tmp, reg_tmp, brg_.k_blk * brg_.n_blk * brg_.ld_block,
+                    X_TMP_0);
+        }
+    }
+
+    void loop_k(int bdb, int ldb, int tail) {
+        zero_regs();
+        mov(reg_aux_a, reg_aux_a1);
+        mov(reg_aux_b, reg_b);
+        if (k_full_blks > 0) {
+            mov(reg_rd_loop, k_full_blks);
+            Label l0;
+            L(l0);
+            microkernel(brg_.rd_block, bdb, ldb, tail);
+            add_imm(reg_aux_a, reg_aux_a,
+                    brg_.m_blk * brg_.k_blk * brg_.rd_block, X_TMP_0);
+            add_imm(reg_aux_b, reg_aux_b,
+                    brg_.k_blk * brg_.n_blk * brg_.ld_block * brg_.rd_block,
+                    X_TMP_0);
+            sub(reg_rd_loop, reg_rd_loop, 1);
+            cmp(reg_rd_loop, 0);
+            b(GT, l0);
+        }
+        if (k_tail_blk > 0) {
+            microkernel(k_tail_blk, bdb, ldb, tail);
+            add_imm(reg_aux_a, reg_aux_a, brg_.m_blk * brg_.k_blk * k_tail_blk,
+                    X_TMP_0);
+            add_imm(reg_aux_b, reg_aux_b,
+                    brg_.k_blk * brg_.n_blk * brg_.ld_block * k_tail_blk,
+                    X_TMP_0);
+        }
+        if (k_residual_blk > 0) { microkernel(1, bdb, ldb, tail); }
+        store_regs(bdb, ldb, tail);
+    }
+
+    void loop_k_zp(int bdb, int ldb, int is_a, int is_b) {
+        eor(z3.d, z3.d, z3.d);
+        eor(z4.d, z4.d, z4.d);
+        for (int i = 0; i < 6; i++)
+            eor(acc(2, i).d, acc(2, i).d, acc(2, i).d);
+        mov(reg_aux_a, reg_aux_a1);
+        mov(reg_aux_b, reg_b);
+        if (k_full_blks > 0) {
+            mov(reg_rd_loop, k_full_blks);
+            Label l0;
+            L(l0);
+            zp_comp(brg_.rd_block, bdb, ldb, is_a, is_b);
+            add_imm(reg_aux_a, reg_aux_a,
+                    brg_.m_blk * brg_.k_blk * brg_.rd_block, X_TMP_0);
+            add_imm(reg_aux_b, reg_aux_b,
+                    brg_.k_blk * brg_.n_blk * brg_.ld_block * brg_.rd_block,
+                    X_TMP_0);
+            sub(reg_rd_loop, reg_rd_loop, 1);
+            cmp(reg_rd_loop, 0);
+            b(GT, l0);
+        }
+        if (k_tail_blk > 0) {
+            zp_comp(k_tail_blk, bdb, ldb, is_a, is_b);
+            add_imm(reg_aux_a, reg_aux_a, brg_.m_blk * brg_.k_blk * k_tail_blk,
+                    X_TMP_0);
+            add_imm(reg_aux_b, reg_aux_b,
+                    brg_.k_blk * brg_.n_blk * brg_.ld_block * k_tail_blk,
+                    X_TMP_0);
+        }
+        if (k_residual_blk > 0) { zp_comp(1, bdb, ldb, is_a, is_b); }
+
+        if (brg_.zp_type_b != jit_int8_broadcast_t::none && is_b == 1) {
+            uzp1(z3.d, z3.d, z4.d);
+            scvtf(z3.s, P_ALL_ONE, z3.s);
+            if (!brg_.is_zp_b_int8) {
+                ldr(W_TMP_0, ptr(reg_zp_val_b));
+                dup(z0.s, W_TMP_0);
+                scvtf(z0.s, P_ALL_ONE, z0.s);
+                fmul(z3.s, P_ALL_ONE, z0.s);
+            } else {
+                if (brg_.zp_type_a != jit_int8_broadcast_t::none) {
+                    ldr(W_TMP_0, ptr(reg_zp_val_a));
+                    dup(z0.s, W_TMP_0);
+                    mov_imm(W_TMP_0, brg_.K);
+                    dup(z1.s, W_TMP_0);
+                    scvtf(z0.s, P_ALL_ONE, z0.s);
+                    scvtf(z1.s, P_ALL_ONE, z1.s);
+                    fmul(z0.s, z1.s, z0.s);
+                    fsub(z3.s, z3.s, z0.s);
+                }
+            }
+            st1w(z3.s, P_ALL_ONE, ptr(reg_zp_b));
+        }
+
+        if ((brg_.zp_type_a != jit_int8_broadcast_t::none) && is_a == 1) {
+            ldr(W_TMP_0, ptr(reg_zp_val_a));
+            dup(z2.s, W_TMP_0);
+            scvtf(z2.s, P_ALL_ONE, z2.s);
+            uzp1(acc(2, 0).d, acc(2, 0).d, acc(2, 1).d);
+            uzp1(acc(2, 2).d, acc(2, 2).d, acc(2, 3).d);
+            uzp1(acc(2, 4).d, acc(2, 4).d, acc(2, 5).d);
+
+            scvtf(acc(2, 0).s, P_ALL_ONE, acc(2, 0).s);
+            scvtf(acc(2, 2).s, P_ALL_ONE, acc(2, 2).s);
+            scvtf(acc(2, 4).s, P_ALL_ONE, acc(2, 4).s);
+            if (brg_.zp_type_b != jit_int8_broadcast_t::none
+                    && !brg_.is_zp_b_int8) {
+                ldr(W_TMP_0, ptr(reg_zp_val_b));
+                dup(z0.s, W_TMP_0);
+                mov_imm(W_TMP_0, brg_.K);
+                dup(z1.s, W_TMP_0);
+                scvtf(z0.s, P_ALL_ONE, z0.s);
+                scvtf(z1.s, P_ALL_ONE, z1.s);
+                fmul(z0.s, z1.s, z0.s);
+                fsub(acc(2, 0).s, acc(2, 0).s, z0.s);
+                fsub(acc(2, 2).s, acc(2, 2).s, z0.s);
+                fsub(acc(2, 4).s, acc(2, 4).s, z0.s);
+            }
+            fmul(acc(2, 0).s, P_ALL_ONE, z2.s);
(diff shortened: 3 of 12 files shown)


cmake/options.cmake
```diff
@@ -1,23 +1,23 @@
-#===============================================================================
-# Copyright 2018-2025 Intel Corporation
+#== == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == == =
```
jondea, line 1: Could you revert these whitespace changes please? It could be that your IDE auto-formatted them without you realizing.

cmake/options.cmake
```diff
@@ -26,14 +26,18 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
...
-# ========
+#== == == ==
+#Features
+#== == == ==
 
 option(DNNL_VERBOSE
     "allows oneDNN be verbose whenever ONEDNN_VERBOSE
     environment variable set to 1" ON) # enabled by default
 
+option(DNNL_AARCH64_MATMUL_SRC_QUANT
```
jondea, line 37: Could you explain the rationale behind using a compile flag for this? Usually features like this would be enabled as runtime options, e.g. https://oneapi-src.github.io/oneDNN/v2.5/group_dnnl_api_mathmode.html

jondea (commented):
Thanks for the contribution. I haven't looked in detail at the JIT code itself, I first want to understand the API and motivation. Also, do you have any performance numbers and related benchdnn arguments?

cmake/options.cmake
Shreyas-fuj: Sorry I did'nt notice this. Done.

cmake/options.cmake
Shreyas-fuj: Currently there are minor expected accuracy issues, with this when we run all tests at benchdnn level. As a result the benchdnn tests fail which might break some existing pipeline. Therefore we have given this as an optional feature. The small accuracy difference is expected as this implementation uses sing scale and zp for the entire matrix as of now.

cmake/options.cmake
Shreyas-fuj: > Could you revert these whitespace changes please? It could be that your IDE auto-formatted them without you realizing.

Sure, I will share the performance results shortly.

Shreyas-fuj:
> Thanks for the contribution. I haven't looked in detail at the JIT code itself, I first want to understand the API and motivation. Also, do you have any performance numbers and related benchdnn arguments?

Please find the performance numbers(f32 vs int8) below taken on a 32core Graviton 3E machine at benchdnn level:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/SHANKA~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/SHANKA~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


Shapes | f32 | int8 | speedup
-- | -- | -- | --
4096x4096:4096x128256 | 1826.83 | 259.224 | 7.047302719
2048x4096:4096x128256 | 883.757 | 128.232 | 6.891860066
580x4096:4096x128256 | 251.723 | 34.3804 | 7.321700736
4096x4096:4096x14336 | 204.306 | 31.1897 | 6.550431713
4096x14336:14336x4096 | 199.682 | 29.3344 | 6.807093378
290x4096:4096x128256 | 129.629 | 16.0703 | 8.066370883
2048x14336:14336x4096 | 98.4181 | 14.115 | 6.972589444
2048x4096:4096x14336 | 99.4149 | 14.6779 | 6.773101057
145x4096:4096x128256 | 70.0352 | 8.38655 | 8.350895183
4096x4096:4096x4096 | 58.6478 | 8.47308 | 6.921662489



</body>

</html>
These shapes were taken from some of the widely used LLM models.
The f32 results were obtained using the command:
`./benchdnn --matmul --mode=p MxK:KxN`
for int8:
`./benchdnn --matmul --dt=s8:s8:f32 --mode=p MxK:KxN`

Other arguments like scale and zp can be give to int8 kernel as example shown below:
`./benchdnn --matmul --dt=s8:s8:f32 --attr-scales=src:common:3+wei:common:4 --attr-zero-points=src:common:3+wei:common:2 85x324:324x243`

cmake/options.cmake
vpirogov: This functionality is unlikely to be discovered and used by anyone if it's hidden behind a build option. So we need to consider the desired end state here. If current implementation is temporary it makes sense to complete it so that it passes tests before adding to the library.

src/cpu/matmul/cpu_matmul_list.cpp
```diff
@@ -79,6 +80,7 @@ constexpr impl_list_item_t impl_list[] = REG_MATMUL_P({
         CPU_INSTANCE_AARCH64_ACL(acl_lowp_matmul_t)
         CPU_INSTANCE_AARCH64_ACL(acl_matmul_t)
         CPU_INSTANCE_AARCH64(brgemm_matmul_t<sve_256>)
+        CPU_INSTANCE_AARCH64(jit_int8_matmul_t)
```
vpirogov, line 83: Could you please elaborate on why we need new direct JIT implementation instead of extending BRGEMM? BRGEMM-based approach is generally preferred over direct JIT from maintenance perspective.

Shreyas-fuj: The memory layouts required for the smmla instruction is quite different and cannot be easily integrated in the current brgemm microkernel. If sdot was used instead, the same approach could have been used.(smmla is a faster instruction in arm for int8 matrix multiplication.) Therefore there would be lot of changes to brgemm kernel which would increase its size and complication, so we have maintained this seperately for fast int8 matmuls.

cmake/options.cmake
Shreyas-fuj: The functionality is complete but there will definitely be small accuracy issues since f32 is converted to u8 and then back to f32. The benchdnn tests currently dont accommodate this accuracy threshold.

Shreyas-fuj: I have now changed it to a runtime feature which can be enabled by  `export ONEDNN_AARCH64_MATMUL_SRC_QUANT=ON` .

Shreyas-fuj:
```
./benchdnn --matmul --dt=f32:s8:f32 32x32:32x32
[   0][DST][0:0] exp_f32:        -486 exp:        -486 got:    -485.795 diff:0.204712 rdiff:0.000421218
[   1][DST][0:1] exp_f32:         206 exp:         206 got:     209.134 diff: 3.13385 rdiff:0.0152129
[   2][DST][0:2] exp_f32:         804 exp:         804 got:     803.276 diff:0.724426 rdiff:0.000901028
[   3][DST][0:3] exp_f32:        -388 exp:        -388 got:    -392.063 diff: 4.06299 rdiff:0.0104716
[   4][DST][0:4] exp_f32:        -179 exp:        -179 got:    -176.378 diff: 2.62206 rdiff:0.0146484
[   5][DST][0:5] exp_f32:         363 exp:         363 got:      350.74 diff: 12.2599 rdiff:0.0337737
[   6][DST][0:6] exp_f32:         327 exp:         327 got:     322.016 diff: 4.98425 rdiff:0.0152424
[   7][DST][0:7] exp_f32:        -341 exp:        -341 got:    -337.134 diff: 3.86615 rdiff:0.0113377
[   8][DST][0:8] exp_f32:        -323 exp:        -323 got:     -317.48 diff: 5.51968 rdiff:0.0170888
[   9][DST][0:9] exp_f32:        -554 exp:        -554 got:    -546.772 diff: 7.22833 rdiff:0.0130475
[COMPARE_STATS][DST]: trh=0 err_max_diff: 14.8188 err_max_rdiff: 7.05512 all_max_diff: 14.8188 all_max_rdiff: 7.05512
[PRIM_REF][INFO]: L2_size:524288 bytes; per_core_L3_size:1048576 bytes; nthr:32; impl_name:gemm:jit:f32
[PRIM_REF][REPRO]: --matmul 32x32:32x32
0:FAILED (errors:1024 total:1024) __REPRO: --matmul --dt=f32:s8:f32 32x32:32x32
```

1. This is an example of accuracy issue. If we consider this, this much error would be introduced when we apply common scales and zp to the entire matrix. 
2. I agree the top level layer will take care of this like pytorch, but if we see OpenVino it expects OneDNN to handle the src quantisation. S0 we have just given this as a feature in case the frame work requires. Removing it all together would not be a good idea. In future PRs we can work on accuracy by introducing different quantisation methods.
3. Regarding benchdnn tests, they are not run with this feature ON.(because this feature is used only when top level framework requires it. Its better to have it this way rather than using reference implementation which is very slow).

dzarukin: f32:u8 mode is not supported by the library without additional attributes. It's not restricted yet but it will be.

There's no API to allow f32 activation down-conversion to s8/u8 and perform integer computations.

Given that benchdnn fills data with integers and powers of 2, I would say the implementation does something wrong in here regardless whether it does f32 computations of s32.

Shreyas-fuj: @dzarukin , @jondea So can you suggest the method ahead for this? Should we completely remove support for f32:u8 ?

dzarukin: It depends what kind of a problem you are trying to solve. If you are solving decompression, that's solvable through `fpmath=DT:true`, where DT is `strict` or `f16/bf16`. If you are trying to address dynamic quantization, that involves more that just math on a primitive side, it does require quantization-on-the-fly, which is not a straightforward thing to do.

src/cpu/matmul/cpu_matmul_list.cpp
vpirogov: Enabling `smmla` in BRGEMM may be worth the effort in the long run as it's an abstraction for small matrix multiplication. With BRGEMM enabled you won't have to write individual JIT kernels for any function that uses this instruction.

cmake/options.cmake
Radu2k: > f32:u8 mode is not supported by the library without additional attributes. It's not restricted yet but it will be.
Why will f32:u8 be restricted? What additional attributes are needed and why are they needed? 

> There's no API to allow f32 activation down-conversion to s8/u8 and perform integer computations.
What do you mean by this?

src/cpu/matmul/cpu_matmul_list.cpp
Shreyas-fuj: smmla (Signed matrix-multiply-accumulate) is not only used for small matrices, it works equally well for large matrices. In fact its better when matrices are large like those in LLMs.

vpirogov: > smmla (Signed matrix-multiply-accumulate) is not only used for small matrices, it works equally well for large matrices.

This is no different from any other instruction accelerating matrix matrix multiplication.

cmake/options.cmake
vpirogov: @Radu2k, 

Let me provide broader context for this discussion first. oneDNN has strictly requirement to numerical behavior and tests that enforce it. It's critical to make sure all implementations produce consistent and predictable results. In default scenario oneDNN primitives are not allowed to reduce precision of any inputs and must use `f32` or `s32` as intermediate accumulation data type. This behavior can be modified with two flags:
* [floating-point math mode](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_fpmath_mode.html) can be used to allow downconversions to more narrow floating point data type. Note, that math mode does not allow downconversions to integer data types.
* [accumulation mode](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_accumulation_mode.html) can be used to allow accumulation to lower precision data types. 

As you can see the implementation we are discussing is not conformant with these requirements hence the failed tests.

Now to specific questions:
> Why will f32:u8 be restricted? What additional attributes are needed and why are they needed?

oneDNN numerical behavior for `--dt=f32:s8:f32` requires that u8 weights are dequantized to f32 using [quantization attributes](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_quantization.html). The fact that API allows to create matmul with `s8` weights and no quantization attributes is bug.

> There's no API to allow `f32` activation down-conversion to `s8`/`u8` and perform integer computations.

Default numerical behavior does not allow downconversions. [Floating-point math mode](https://oneapi-src.github.io/oneDNN/dev_guide_attributes_fpmath_mode.html) only supports downconversion to floating point data types. The main reason why downconversion to integer data types is not supported is because this process requires quantization parameters and has many different ways to do that. Typically it's up to framework to define how to do that based on model level accuracy considerations. For these cases OpenVINO, for example, calls oneDNN with `s8:s8:f16` data type combination.

cmake/options.cmake
```diff
@@ -424,4 +424,4 @@ option(DNNL_AARCH64_USE_ACL "Enables use of AArch64 optimised functions
     from Arm Compute Library.
     This is only supported on AArch64 builds and assumes there is a
     functioning Compute Library build available at the location specified by the
-    environment variable ACL_ROOT_DIR." OFF)
+    environment variable ACL_ROOT_DIR." OFF)
```
vpirogov, line 427: Pls revert this change.

cmake/options.cmake
vpirogov: > @dzarukin , @jondea So can you suggest the method ahead for this? Should we completely remove support for f32:u8 ?

I would suggest to drop dynamic quantization case as it's not consistent with library numerical behavior and not covered by tests.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        return status::unimplemented;
+
+    bool is_s8_wei = utils::everyone_is(s8, wei_type);
+    bool is_u8 = utils::everyone_is(u8, src_type, wei_type);
+    bool is_s8 = utils::everyone_is(s8, src_type, wei_type);
+    bool is_src_f32 = (utils::everyone_is(u8, wei_type)
+                              || utils::everyone_is(s8, wei_type))
+            && utils::everyone_is(f32, src_type);
+
+    static std::string val = getenv_string_user("AARCH64_MATMUL_SRC_QUANT");
```
vpirogov, line 593: This method allows users to alter behavior of application relying on oneDNN without control from the application developer.

cmake/options.cmake
Shreyas-fuj: @vpirogov, understood. Thanks for the clarification. I will remove it from this PR.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
Shreyas-fuj: Sure, I will remove this from the PR.

Shreyas-fuj: @vpirogov  Is there any plan to support this type of quantization(when requested through attributes), if so I can remove just the functionality which enables this now and keep the quantization logic code intact. Or else should I remove the entire implementation?

cmake/options.cmake
Shreyas-fuj: Sure

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
vpirogov: > @vpirogov Is there any plan to support this type of quantization(when requested through attributes)

We are not currently working on enabling dynamic quantization as this approach does not bring performance benefits over performing quantization on application side and calling `s8`/`u8` primitive on oneDNN side. Though if you want to explore this approach I'd suggest partnering with OpenVINO to confirm viability of the approach and create [an RFC](https://github.com/oneapi-src/oneDNN/tree/rfcs) with proposal.

> I can remove just the functionality which enables this now and keep the quantization logic code intact. Or else should I remove the entire implementation?

There's no reason to merge dead code to production branches, so please remove the implementation. If you want to keep this code in case it's needed in the future please open a separate draft PR.

cmake/options.cmake
Radu2k: Thanks @vpirogov for the detailed explanation! 

It makes sense to let frameworks handle the dynamic quantization and provide quantized inputs, since they can better manage accuracy tradeoffs at the model level.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
Shreyas-fuj: Sure, I will make the necessary changes. Thanks.

cmake/options.cmake
Shreyas-fuj: I am not sure why this is showing up. I have restored it but again when I push this comes up. Any suggestions for this? Even in the git diff I cant identify what is the change.

cmake/options.cmake
Shreyas-fuj: @vpirogov @jondea I have removed the src dynamic quantization from this PR, please have a look.
@Radu2k @dzarukin Please share if you have any other review comments from your side. Thanks!

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -715,6 +715,9 @@ typedef enum {
     dnnl_aBC16b32c,
     dnnl_AB16a16b,
     dnnl_AB16a32b,
+    dnnl_BA24b8a,
+    dnnl_aCB24c8b,
+    dnnl_abDC24d8c,
```
dzarukin, line 720: New tags must be put at the end of the list next to the `tag_last` to avoid ABI issues.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+struct jit_int8_matmul_t : public primitive_t {
+    struct pd_t : public dnnl::impl::cpu::matmul::cpu_matmul_pd_t {
+        using ::dnnl::impl::cpu::matmul::cpu_matmul_pd_t::cpu_matmul_pd_t;
+
+        DECLARE_COMMON_PD_T("jit:int8", jit_int8_matmul_t);
+
+        status_t init(engine_t *engine);
+
+        bool formats_ok() const {
```
dzarukin, line 160: Is this call unused?

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,216 @@
...
+            int nt = b.N % (b.n_blk * b.ld_block);
+            int kt = b.K % (b.k_blk * 4);
+            if ((m == 1 && mt == 0) || (k == 1 && kt == 0)
+                    || (n == 1 && nt == 0) || (k == 0 && kt == 1))
+                return -1;
+            return k + n * 2 + m * 2 * 2 + z * 2 * 2 * 2;
+        }
+
+    private:
+        brg_int8_t brg;
```
dzarukin, line 194: Trailing underscores for class members?

dzarukin:
The code is very raw, I have high level of confidence it would fail on lots of benchdnn cases.

Shreyas-fuj:
> The code is very raw, I have high level of confidence it would fail on lots of benchdnn cases.

HI @dzarukin, I will look into these. But there are no failures on benchdnn test cases. It was verified before raising the PR.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+            }
+            if (src_d.dims()[0] != weights_d.dims()[0]
+                    || src_d.dims()[1] != weights_d.dims()[1])
+                return status::unimplemented;
+            break;
+        }
+        default: return status::unimplemented;
+    }
+
+    VDISPATCH_MATMUL(set_default_formats(), VERBOSE_UNSUPPORTED_TAG);
```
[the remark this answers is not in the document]
Shreyas-fuj: This line was added because, bias cases were failing without this even though we have initialized memory before.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+
+    bool no_post_ops = attr()->post_ops_.has_default_values();
+    const bool ok = (is_s8 || is_u8) && utils::everyone_is(f32, dst_type)
+            && no_post_ops && (get_sve_length() == 32);
+    if (!ok) return status::unimplemented;
+
+    switch (dims) {
+        case 2: {
+            VCHECK_BG(memory_desc_init_by_tag(src_md_, format_tag::ab),
+                    VERBOSE_UNSUPPORTED_TAG);
```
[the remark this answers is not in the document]
Shreyas-fuj: Then the jit:uni reorder would be called to convert it to tag BA24b8a. We have an internal more efficient reorder designed for ab->BA24b8a which would be better than jit:uni, but if the tag is ba,
ba->BA24b8a is handled well by jit:uni itself.

include/oneapi/dnnl/dnnl_types.h
Shreyas-fuj: The location I had added the tags failed some checks on the PR. The checks suggested these places for me to add these tags, once I did the checks passed. Should I change it anyways?

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+                int mtail
+                        = (m_ed == num_a_blocks) ? ((b.m_tail > 0) ? 1 : 0) : 0;
+                int m_blk_adr = (batch
+                                        * (num_a_blocks * b.m_blk
+                                                * div_up(K, b.k_blk) * b.k_blk))
+                        + m_st * b.m_blk * div_up(K, b.k_blk) * b.k_blk;
+                int zp_ptr_b_adr
+                        = (batch * (num_a_blocks * b.m_blk)) + m_st * b.m_blk;
+
+                int idx = pd()->get_idx(1, 0, ktail, 0, b);
```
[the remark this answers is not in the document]
Shreyas-fuj: This is in the execute function. If here something fails we cannot return an error because primitive would already be selected and returning error at this stage would fail the model execution from top level frameworks. So here we have to give values cautiously. (Same is done in x86 src/cpu/aarch64/matmul/brgemm_matmul.cpp).
If idx==-1 we do not create the kernel itself. This is done in the init function. Please have a look.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+        for (int m = 0; m < 2; m++)
+            for (int n = 0; n < 2; n++)
+                for (int k = 0; k < 2; k++) {
+                    int idx = pd()->get_idx(z, m, k, n, b);
+                    if (idx == -1) continue;
+                    b.is_m_tail = m;
+                    b.is_k_tail = k;
+                    b.is_n_tail = n;
+                    b.is_zp_cal = z;
+                    int8_kernels_[idx]
```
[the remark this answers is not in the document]
Shreyas-fuj: the get_idx function can never return value greater than 15.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+                        * brg.k_blk,
+                sizeof(char));
+    book_precomputed_scales(scratchpad, attr()->scales_, N());
+
+    return status::success;
+}
+
+status_t jit_int8_matmul_t::init(engine_t *engine) {
+
+    auto &b1 = pd()->get_b();
```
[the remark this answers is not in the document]
Shreyas-fuj: Here we need to modify some parameters of b1 according to the kernel required, so we have created b1 and copied the values of brg and then sent for kernel creation.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+                = !attr()->scales_.get(DNNL_ARG_SRC).has_default_values();
+        auto is_dst_scl
+                = !attr()->scales_.get(DNNL_ARG_DST).has_default_values();
+        auto is_wei_scl
+                = !attr()->scales_.get(DNNL_ARG_WEIGHTS).has_default_values();
+        auto dst_scl_msk = attr()->scales_.get(DNNL_ARG_DST).get_mask();
+        auto wei_scl_msk = attr()->scales_.get(DNNL_ARG_WEIGHTS).get_mask();
+        auto src_scl_msk = attr()->scales_.get(DNNL_ARG_SRC).get_mask();
+
+        if ((is_src_scl && src_scl_msk != 0)
```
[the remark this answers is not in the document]
Shreyas-fuj: YEs, will make this change.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+
+        if (!zero_points.has_default_data_type(DNNL_ARG_WEIGHTS) && wt_int8) {
+            brg->is_zp_b_int8 = true;
+            if (zero_points.get_data_type(DNNL_ARG_WEIGHTS) == u8)
+                brg->zp_b_dt = u8;
+            else
+                brg->zp_b_dt = s8;
+        }
+
+        if (!zero_points.common(DNNL_ARG_SRC)
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+            if (N() == DNNL_RUNTIME_DIM_VAL) ok = false;
+        }
+        return ok;
+    };
+
+    auto check_bias = [&]() -> bool {
+        const auto bia_dt = weights_md(1)->data_type;
+
+        const bool is_bia_dt_correct = IMPLICATION(true, one_of(bia_dt, f32))
+                && IMPLICATION(false, one_of(bia_dt, f32, src_type));
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change

dzarukin:
> > The code is very raw, I have high level of confidence it would fail on lots of benchdnn cases.
> 
> HI @dzarukin, I will look into these. But there are no failures on benchdnn test cases. It was verified before raising the PR.

Sorry, that was a quite premature statement. Let me put it this way:
* Unless I'm really missing something, I would expect to see issues around memory formats because of the way that's used to initialize memory descriptors. Especially transposed tags for src and weights. That's the primary concern I have.
* It looks like the PR is not rebased on top of the latest main, and that was a second source of my conclusion because with existing change the rebased version wouldn't work.
* There will be clang-tidy issues raised. I marked some of those, but it may find additional ones on top, would be good to address that before promotion.
* Styling is minor but still good to follow.
* Verbose support is also minor but really helpful to save time debugging the reason.
* Personal observation: it seems the existing approach follows a data-centric classes style which has its advantage but the more it grows the harder to keep it accurate. Changing a value for a class to specify which reorder, A or B, to initialize might shot at somebody at some point. Better be careful with such things, because the deeper the whole goes, the harder to debug it.

Shreyas-fuj:
> src_d.has_runtime_dims_or_strides()

Thanks for the clarification @dzarukin .
Yes, I will rebase the code to the main and make changes and test and push again.

Shreyas-fuj:
> > > The code is very raw, I have high level of confidence it would fail on lots of benchdnn cases.
> > 
> > 
> > HI @dzarukin, I will look into these. But there are no failures on benchdnn test cases. It was verified before raising the PR.
> 
> Sorry, that was a quite premature statement. Let me put it this way:
> 
> * Unless I'm really missing something, I would expect to see issues around memory formats because of the way that's used to initialize memory descriptors. Especially transposed tags for src and weights. That's the primary concern I have.
> * It looks like the PR is not rebased on top of the latest main, and that was a second source of my conclusion because with existing change the rebased version wouldn't work due to how quantization is used/checked.
> * There will be clang-tidy issues raised. I marked some of those, but it may find additional ones on top, would be good to address that before promotion.
> * Styling is minor but still good to follow.
> * Verbose support is also minor but really helpful to save time debugging the reason.
> * Personal observation: it seems the existing approach follows a data-centric classes style which has its advantage but the more it grows the harder to keep it accurate. Changing a value for a class to specify which reorder, A or B, to initialize might shot at somebody at some point. Better be careful with such things, because the deeper the whole goes, the harder to debug it.

Thanks for the clarification @dzarukin .
Yes, I will rebase the code to the main and make changes and test and push again.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
Shreyas-fuj: Will make this change

include/oneapi/dnnl/dnnl_types.h
dzarukin: [This](https://github.com/oneapi-src/oneDNN/blob/main/include/oneapi/dnnl/dnnl_types.h#L1051) is the right spot.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
dzarukin: I can see a `continue` on a `-1` returned value, I can't see where it's guaranteed that the index in this function won't hit the empty spot in the array where the kernel was not generated. I don't see it's checked here or anywhere else, thus, the question.

It's also not obvious to me or static analysis tool why this function would not return `-1` for given values.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
dzarukin: Can't insist on making it better but if it receives an extension in the future, expect surprises.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
dzarukin: I can see reading from `b1`, can't see writing to `b1`. Could you point me to the line where it gets updated?

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+        return true;
+    };
+
+    VDISPATCH_MATMUL(init_zp_type(&brg), VERBOSE_UNSUPPORTED_ZP_CFG);
+    VDISPATCH_MATMUL(check_bias(), VERBOSE_UNSUPPORTED_BIAS_CFG);
+    VDISPATCH_MATMUL(check_attr_scales(), VERBOSE_UNSUPPORTED_SCALES_CFG);
+
+    bool no_post_ops = attr()->post_ops_.has_default_values();
+    const bool ok = (is_s8 || is_u8) && utils::everyone_is(f32, dst_type)
+            && no_post_ops && (get_sve_length() == 32);
```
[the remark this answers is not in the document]
Shreyas-fuj: Yes, will make this change.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1347 @@
...
+    const auto wei_type = weights_md(0)->data_type;
+    const auto dst_type = dst_md(0)->data_type;
+
+    const memory_desc_wrapper src_d(src_md_);
+    const memory_desc_wrapper weights_d(weights_md_);
+    const memory_desc_wrapper dst_d(dst_md_);
+
+    if (src_d.has_runtime_dims_or_strides()
+            || weights_d.has_runtime_dims_or_strides())
+        return status::unimplemented;
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,216 @@
...
+    XReg reg_zp_val_b = reg_aux_scales;
+
+    call_params_t inp;
+
+    void operator()(const call_params_t *p) {
+        return jit_generator::operator()(p);
+    }
+
+    jit_int8_matmul_kernel_t(
+            const dnnl::impl::cpu::aarch64::matmul::brg_int8_t &k)
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
Shreyas-fuj: Will add this.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
dzarukin: It seems that you are covering weights. But even for them, if the user specified a fixed format, you can't change its format in here, because user will pass the data in the format passed into the execute function and not in the format you instructed them to. It's a property of tag::any, or format_kind::any, but I can't see handling it properly here.

FYI: benchdnn doesn't check if user's requested format coincides with the one the library initialized internal memory descriptor with, but that check would raise the flag as it would hold.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
dzarukin: Maybe initialize bias explicitly then to avoid an unnecessary call?

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,216 @@
...
+            const bool is_src
+                    = memory_desc_wrapper(src_md()).matches_one_of_tag(
+                            format_tag::ab, format_tag::abc, format_tag::abcd);
+            return is_dst && is_wei && is_src;
+        }
+        const brg_int8_t &get_b() const { return brg; }
+
+        const dyn_vals_t &get_d() const { return dyn; }
+
+        int get_idx(int z, int m, int k, int n, brg_int8_t b) const {
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
Shreyas-fuj: Understood. Added a check here.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
Shreyas-fuj: I have added the check for >15.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
Shreyas-fuj: Thanks for pointing out. I have added the formats_ok() check to handle this.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
Shreyas-fuj: Done.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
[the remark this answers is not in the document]
Shreyas-fuj: Sorry for the confusion. The variable b should not be a constant but b1 should. I have made this change.

include/oneapi/dnnl/dnnl_types.h
Shreyas-fuj: Done, thanks.

Shreyas-fuj:
@dzarukin Thanks for the extensive review. Your valuable comments and suggestions are highly appreciated. I have made all the suggested changes, request you to please have a look.

Shreyas-fuj:
@vpirogov @dzarukin @Radu2k @jondea , please let me know if there are any other changes required for approval. Thanks.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -0,0 +1,1401 @@
...
+            = (is_s8 || is_u8) && utils::everyone_is(f32, dst_type);
+
+    VDISPATCH_MATMUL(problem_dt_correct, VERBOSE_UNSUPPORTED_DT);
+    VDISPATCH_MATMUL(no_post_ops, VERBOSE_UNSUPPORTED_ATTR);
+    VDISPATCH_MATMUL(formats_ok(), VERBOSE_UNSUPPORTED_TAG);
+    VDISPATCH_MATMUL(get_sve_length() == 32, VERBOSE_UNSUPPORTED_ISA);
+
+    switch (dims) {
+        case 2: {
+            VCHECK_BG(memory_desc_init_by_tag(src_md_, format_tag::ab),
```
dzarukin, line 683: Minor: coupled with `formats_ok` this works now, but you may make it more bullet proof by using the following logic:
```
if (format_any) {
    initialize_with_ab
} else {
    // Plain copy of initialized user's md should work.
    src_md_ = src_desc();
}
```
This would help to avoid future modifications and debugging once you'd like to add transposed formats for this implementation.

dzarukin (commented):
The last big piece I have is header dependencies for the list file, once it's addressed, it's good to go from my side. Thank you.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,218 @@
...
+    brg_int8_t_ brg;
+
+    int ldb;
+    int bdb;
+    int rdb;
+    int k_full_blks;
+    int k_tail_blk;
+    int k_residual_blk;
+    int n_blks;
+};
```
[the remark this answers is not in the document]
Shreyas-fuj: HI @dzarukin, if we do that then it would be difficult for other kernels(like convolution) to use the jit micro kernel defined in the cpp file by including the hpp file. We have done this so that by including the hpp file, other kernels can use the mm microkernel. Any suggestions for this?

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,218 @@
...
+        DECLARE_COMMON_PD_T("jit:int8", jit_int8_matmul_t);
+
+        status_t init(engine_t *engine);
+
+        bool formats_ok() const {
+
+            const memory_desc_wrapper src_d(src_md_);
+            const memory_desc_wrapper weights_d(weights_md_);
+            const memory_desc_wrapper dst_d(dst_md_);
+            const bool is_dst = dst_d.matches_one_of_tag(format_tag::ab,
```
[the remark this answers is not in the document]
Shreyas-fuj: Sure, will do this.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
Shreyas-fuj: Sure, will do this.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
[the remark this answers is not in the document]
dzarukin: @vpirogov already made that suggestion - to follow the brgemm path and add support there - an universal kernel with API to configure and control with a potential to be exposed through external ukernel API. I share that vision and I don't think that including current struct in other primitives helps in the long-term for maintainability and scalability.

I'm also very skeptical that one can simply re-enable matmul kernel for convolutions. I can see the other way around with some performance penalties, but not the way forward.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
Shreyas-fuj: But here the condition for now should be :
if(tag=ab) initialise with ab
else initialise with BA24b8a

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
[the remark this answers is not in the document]
Shreyas-fuj: @dzarukin , we are also looking at convolution, can we make this change in future if we are sure that the convolution kernel can not use this mm kernel?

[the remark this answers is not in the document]
dzarukin: It seems there's slight misunderstanding - I don't ask to make a kernel property of matmul, I'm asking to remove an implementation header (`jit_int8_matmul.hpp`) dependency on jit machinery which would save from potential compilation culprits due to extra dependencies and also save compilation time.

This is the proposal:
```
/// These types form a jit_int8_kernel_types.hpp new header file
/// It's fine to include into jit_int8_matmul.hpp as these are POD mostly.
dyn_vals_t
dyn_params_t 
jit_int8_broadcast_t
brg_int8_t
```

```
/// This struct and its dependencies forms a jit_int8_kernel.hpp
/// Function bodies moves to cpp counterpart.
jit_int8_matmul_kernel_t
```

`jit_int8_matmul.hpp` adds a forward declaration for `jit_int8_matmul_kernel_t`.
`jit_int8_matmul.cpp` adds the `jit_int8_kernel.hpp` header.

Everything compiles, nicely extends to convolution, no dependencies.
Future Convolution implementation's cpp file will include kernel header same as matmul.

Do you see any issues with having this done?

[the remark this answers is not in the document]
Shreyas-fuj: > This will bring a ton of dependencies into the list file which is not good. For x64 we keep those in .cpp and make a forward declaration in a header file. I suggest you to follow and drop jit_utils header dependency here.

But if utils dependency is required in hpp because of these declarations :
```
 std::unique_ptr<jit_int8_matmul_utils_kernel_t> reo_ker_a_;
 std::unique_ptr<jit_int8_matmul_utils_kernel_t> reo_ker_b_;
```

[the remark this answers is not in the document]
dzarukin: `struct jit_int8_matmul_utils_kernel_t;` above `jit_int8_matmul_t` will do the trick. It is referred as forward declaration above.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,218 @@
...
+            int kt = b.K % (b.k_blk * 4);
+            if ((m == 1 && mt == 0) || (k == 1 && kt == 0)
+                    || (n == 1 && nt == 0) || (k == 0 && kt == 1))
+                return -1;
+            return k + n * 2 + m * 2 * 2 + z * 2 * 2 * 2;
+        }
+
+    private:
+        brg_int8_t_ brg;
+        dyn_vals_t_ dyn;
```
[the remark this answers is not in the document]
Shreyas-fuj: Yes, will make this change thanks.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
Shreyas-fuj: Understood. Will add this.

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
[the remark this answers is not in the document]
Shreyas-fuj: @vpirogov , understood. Thanks for the detailed explanation. Will make this change and update.

Shreyas-fuj:
@dzarukin , I have made the suggested changes, request you to have a look. Thanks!

dzarukin (approved):

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,111 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef CPU_AARCH64_jit_int8_matmul_HPP
+#define CPU_AARCH64_jit_int8_matmul_HPP
```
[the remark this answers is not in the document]
Shreyas-fuj: Will make this change, thanks.

Shreyas-fuj:
@dzarukin, thanks for the review!
@jondea, any comments from ARM side for approval? Thanks.

include/oneapi/dnnl/dnnl_types.h
Shreyas-fuj: @jondea, I have corrected this.

Shreyas-fuj:
@jondea, requesting for approval from ARM side if there are no more review comments. Thanks.

Shreyas-fuj:
I have just rebased and force pushed the commit, but the clang-tidy is failing with the error:
```
Getting Git version info
Temporarily overriding HOME='/home/runner/work/_temp/b589a90e-5114-4043-bb40-2090bc2bf684' before making global git config changes
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /home/runner/work/oneDNN/oneDNN
Deleting the contents of '/home/runner/work/oneDNN/oneDNN'
Initializing the repository
Disabling automatic garbage collection
Setting up auth
Fetching the repository
Determining the checkout info
  /usr/bin/git branch --list --remote origin/aarch64_matmul_int8
  /usr/bin/git tag --list aarch64_matmul_int8
  Error: A branch or tag with the name 'aarch64_matmul_int8' could not be found
```
Any idea why this is happening?

dzarukin:
> I have just rebased and force pushed the commit, but the clang-tidy is failing with the error:
> 
> ```
> Getting Git version info
> Temporarily overriding HOME='/home/runner/work/_temp/b589a90e-5114-4043-bb40-2090bc2bf684' before making global git config changes
> Adding repository directory to the temporary git global config as a safe directory
> /usr/bin/git config --global --add safe.directory /home/runner/work/oneDNN/oneDNN
> Deleting the contents of '/home/runner/work/oneDNN/oneDNN'
> Initializing the repository
> Disabling automatic garbage collection
> Setting up auth
> Fetching the repository
> Determining the checkout info
>   /usr/bin/git branch --list --remote origin/aarch64_matmul_int8
>   /usr/bin/git tag --list aarch64_matmul_int8
>   Error: A branch or tag with the name 'aarch64_matmul_int8' could not be found
> ```
> 
> Any idea why this is happening?

Can the fork be a reason? I'm not totally sure if the linter is designed to work with forks.

Shreyas-fuj:
> > I have just rebased and force pushed the commit, but the clang-tidy is failing with the error:
> > ```
> > Getting Git version info
> > Temporarily overriding HOME='/home/runner/work/_temp/b589a90e-5114-4043-bb40-2090bc2bf684' before making global git config changes
> > Adding repository directory to the temporary git global config as a safe directory
> > /usr/bin/git config --global --add safe.directory /home/runner/work/oneDNN/oneDNN
> > Deleting the contents of '/home/runner/work/oneDNN/oneDNN'
> > Initializing the repository
> > Disabling automatic garbage collection
> > Setting up auth
> > Fetching the repository
> > Determining the checkout info
> >   /usr/bin/git branch --list --remote origin/aarch64_matmul_int8
> >   /usr/bin/git tag --list aarch64_matmul_int8
> >   Error: A branch or tag with the name 'aarch64_matmul_int8' could not be found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any idea why this is happening?
> 
> Can the fork be a reason? I'm not totally sure if the linter is designed to work with forks.

Even I think there is a linter issue, I checked the tracking info on my local repo, aarch64_matmul_int8 is tracking the correct branch in upstream.

vpirogov:
> Any idea why this is happening?

Linter is currently broken for forks. Will be fixed by #2859. It should not block the PR though.

atkassen:
> Linter is currently broken for forks. Will be fixed by #2859. It should not block the PR though.

I've just merged the fix, try rebasing again.

Shreyas-fuj:
> > Linter is currently broken for forks. Will be fixed by #2859. It should not block the PR though.
> 
> I've just merged the fix, try rebasing again.

Thanks! I have rebased, it seems to be working fine now.

jondea (approved):
You've addressed all my comments, thank you! I do have one (non-blocking) question though: is this vector length agnostic? Not that it needs to be, but I can't see any reference to anything vector length specific

Shreyas-fuj:
> You've addressed all my comments, thank you! I do have one (non-blocking) question though: is this vector length agnostic? Not that it needs to be, but I can't see any reference to anything vector length specific

@jondea , thanks for the approval, right now its written only for GR3 cpus, but can be easily extended to other vector lengths in future by adjusting the block sizes.

Shreyas-fuj:
Hi @vpirogov, is there anything to be done from our side which is blocking the merge of this PR?

src/cpu/aarch64/matmul/jit_int8_kernel_types.hpp
```diff
@@ -0,0 +1,94 @@
...
+    const int8_t *wei_zero_point_buf;
+    float *zp_a_ptr, *zp_b_ptr;
+};
+
+} // namespace matmul
+} // namespace aarch64
+} // namespace cpu
+} // namespace impl
+} // namespace dnnl
+#endif
```
vpirogov, line 94: (nit) EOL

```suggestion
#endif

```

src/cpu/aarch64/matmul/jit_int8_matmul_utils.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    dyn_vals_t dyn_;
+};
+
+} // namespace matmul
+} // namespace aarch64
+} // namespace cpu
+} // namespace impl
+} // namespace dnnl
+#endif
```
vpirogov, line 86: (nit) EoL

```suggestion
#endif

```

src/cpu/aarch64/matmul/jit_int8_matmul.hpp
```diff
@@ -0,0 +1,111 @@
...
+    std::unique_ptr<jit_int8_matmul_utils_kernel_t> reo_ker_a_;
+    std::unique_ptr<jit_int8_matmul_utils_kernel_t> reo_ker_b_;
+};
+
+} // namespace matmul
+} // namespace aarch64
+} // namespace cpu
+} // namespace impl
+} // namespace dnnl
+#endif
```
vpirogov, line 111: (nit) EoL

```suggestion
#endif

```

vpirogov:
@Shreyas-fuj, There're a couple comments from @dzarukin that should be addressed before promotion:
* https://github.com/uxlfoundation/oneDNN/pull/2686#discussion_r1983749114
* https://github.com/uxlfoundation/oneDNN/pull/2686#discussion_r1983758248

Shreyas-fuj:
> @Shreyas-fuj, There're a couple comments from @dzarukin that should be addressed before promotion:
> 
> * [aarch64: matmul: addition of JIT int8 kernel #2686 (comment)](https://github.com/uxlfoundation/oneDNN/pull/2686#discussion_r1983749114)
> * [aarch64: matmul: addition of JIT int8 kernel #2686 (comment)](https://github.com/uxlfoundation/oneDNN/pull/2686#discussion_r1983758248)

Hi @vpirogov , the second comment was not insisted. I had missed the first comment, thanks.

src/cpu/aarch64/matmul/jit_int8_matmul_utils.hpp
```diff
@@ -0,0 +1,86 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef CPU_AARCH64_jit_int8_matmul_utils_HPP
+#define CPU_AARCH64_jit_int8_matmul_utils_HPP
+
+#include "common/primitive.hpp"
```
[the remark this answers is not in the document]
Shreyas-fuj: Hi @dzarukin , sorry I had missed this. I am not sure what this means can you explain?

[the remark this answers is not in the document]
dzarukin: It means `common/primitive.hpp` header shouldn't be needed for this file.

Shreyas-fuj:
@vpirogov , I have made the EOL changes suggested by you. Thanks.

Shreyas-fuj:
Hi @vpirogov, I have addressed all the comments. Thanks!

PR #2690, opened 2025-02-13
Update clang-format version

vpirogov:
We are stuck at clang-format-11, which is quite old, because of behavior differences with newer version. The intent of this PR is to update version in CI and address formatting differences that arise.

### DIFF
--- a/.github/automation/clang-format.sh
+++ b/.github/automation/clang-format.sh
@@ -16,7 +16,7 @@
 # limitations under the License.
 #===============================================================================
 
-CLANG_FORMAT=clang-format-11
+CLANG_FORMAT=clang-format-17
 
 echo "Checking ${CLANG_FORMAT}"
 if ! ${CLANG_FORMAT} --version; then
--- a/.github/automation/clang-format.sh
+++ b/.github/automation/clang-format.sh
@@ -33,7 +33,10 @@ echo $(git status) | grep "nothing to commit" > /dev/null
 
 if [ $? -eq 1 ]; then
     echo "Clang-format check FAILED! Found not formatted files!"
+    echo "Impacted files: "
     echo "$(git status)"
+    echo "Changes:"
+    echo "$(git diff)"
     RETURN_CODE=3
 else
     echo "Clang-format check PASSED! Not formatted files not found..."
--- a/.github/workflows/pr-linter.yml
+++ b/.github/workflows/pr-linter.yml
@@ -33,7 +33,7 @@ concurrency:
 jobs:
   pr-formatting:
     name: Formatting
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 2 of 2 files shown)


theComputeKid (approved):

rjoursler:
> I'm lost on why clang-tidy-18 wants to do this:

To me, this just looks like a bug. From the documentation for [AlignAfterOpenBracket](https://releases.llvm.org/18.1.6/tools/clang/docs/ClangFormatStyleOptions.html#alignafteropenbracket), we use DontAlign which states:

> BAS_DontAlign (in configuration: DontAlign) Don’t align, instead use ContinuationIndentWidth, e.g.:

The ContinuationIndentWidth we set is 8 and Clang-18 is instead indenting by 10.

It looks like clang-tidy-17 works as expected though.

vpirogov:
clang-format-14 has `AllowShortFunctionsOnASingleLine: Inline` broken. Here's sample diff:
```
diff --git a/src/common/engine_impl.hpp b/src/common/engine_impl.hpp
index e42b8fa27f..ad6e26cb74 100644
--- a/src/common/engine_impl.hpp
+++ b/src/common/engine_impl.hpp
@@ -56,11 +56,17 @@ public:
     }

 #ifdef ONEDNN_BUILD_GRAPH
-    void *get_allocator() const { return (void *)(&allocator_); };
-    void set_allocator(graph::allocator_t *alloc) { allocator_ = *alloc; }
+    void *get_allocator() const {
+        return (void *)(&allocator_);
+    };
+    void set_allocator(graph::allocator_t *alloc) {
+        allocator_ = *alloc;
+    }
 #endif

-    virtual status_t init() { return status::success; }
+    virtual status_t init() {
+    }

     virtual status_t create_memory_storage(memory_storage_t **storage,
             engine_t *engine, unsigned flags, size_t size, void *handle) const {

```

vpirogov:
I've experimented with several versions so far and every one has some quirks. In depth analysis of some changes reported below indicates either bugs or behavior changes of existing functionality. I was not able to find knobs to reduce produced diff.

Here's the data to illustrate the issue we are facing:
| version | # files changed | diff |
| :- | -: | :- |
| 14.0.6 | 159 | [clang-format-14.diff.txt](https://github.com/user-attachments/files/18805262/clang-format-14.diff.txt) |
| 17.0.6 | 140 | [clang-format-17.diff.txt](https://github.com/user-attachments/files/18805273/clang-format-17.diff.txt) |
| 18.1.2 | 426 | [clang-format-18.diff.txt](https://github.com/user-attachments/files/18805277/clang-format-18.diff.txt) |

vpirogov:
clang-format-17 hates pure virtual function declarations for some reason:
```
diff --git a/src/common/engine.hpp b/src/common/engine.hpp
index 65cdce5d5b..abe84715e3 100644
--- a/src/common/engine.hpp
+++ b/src/common/engine.hpp
@@ -113,23 +113,27 @@ struct dnnl_engine : public dnnl::impl::c_compatible {
      * a NULL-terminated list */
     virtual const dnnl::impl::impl_list_item_t *get_reorder_implementation_list(
             const dnnl::impl::memory_desc_t *src_md,
-            const dnnl::impl::memory_desc_t *dst_md) const = 0;
+            const dnnl::impl::memory_desc_t *dst_md) const
+            = 0;
```

This specific change can be reverted with `BreakBeforeBinaryOperators: NonAssignment`, but it leads to massive changes elsewhere.

vpirogov:
None of the newer versions look particularly good. Closing for now.

PR #2692, opened 2025-02-13
src: add debuginfo logging for CHECK and CHECK_BOOL

rjoursler:
Adds debug info logging in the CHECK and CHECK_BOOL macros. In addition, to make the decision for logging level reasonable, this PR standardizes the debug info logging levels across the library to be aligned with the levels in spdlog.


src/common/utils.hpp
```diff
@@ -74,13 +75,19 @@ static_assert(sizeof(void *) == 8, "oneDNN supports 64-bit architectures only");
 #define CHECK(f) \
     do { \
         dnnl::impl::status_t _status_ = f; \
-        if (_status_ != dnnl::impl::status::success) return _status_; \
+        if (_status_ != dnnl::impl::status::success) { \
+            VDEBUGINFO(debug, common, utils, "CHECK() failed"); \
```
atkassen, line 79: Maybe
```suggestion
            VDEBUGINFO(debug, common, utils, #f " = %s", dnnl_status2str(f)); \
```
is a more useful message for debugging.

atkassen (approved):

src/common/verbose.hpp
```diff
@@ -181,6 +183,15 @@ struct verbose_t {
         all = (uint32_t)-1,
     };
 
+    enum debug_flag_kind : uint32_t {
+        debug_critical = 1,
+        debug_error = 2,
+        debug_warn = 3,
+        debug_info = 4,
+        debug_debug = 5,
```
echeresh, line 191: Just an observation: having `debuginfo`/`debug_info`/`debug_debug` referring to different things gets confusing.

Maybe we should drop `debuginfo` as it's semantically loaded due to clash with logging levels. It could be named something like "developer verbose" (`dev_verbose` in API and `ONEDNN_VERBOSE=dev=...` in the environment variable) - to align with `ONEDNN_DEV_MODE`.

echeresh (approved):

src/common/verbose.hpp
```diff
@@ -181,6 +183,15 @@ struct verbose_t {
         all = (uint32_t)-1,
     };
 
+    enum debug_flag_kind : uint32_t {
```
avmanerikar, line 186: (Random Spot): How does this align with the logging levels in `spdlog`? From the [mapping](https://github.com/oneapi-src/oneDNN/blob/e80b57a5865b5cce2d607f34b7e6529e97f6df9e/src/common/verbose.hpp#L287) in verbose for logging levels, all debug verbose information is printed at the `spdlog::debug` level. Are these expected to add another level like:
```
[ ... 18:16:10.201] [oneDNN] [DEBUG] [TRACE] onednn_verbose,primitive,exec, ...
[ ... 18:16:10.201] [oneDNN] [DEBUG] [INFO] onednn_verbose,primitive,exec, ...
```

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2018-2023 Intel Corporation
+* Copyright 2018-2025 Intel Corporation
```
theComputeKid, line 2: 2018-2023,2025

src/common/verbose.hpp
mgouicem: +1
BTW, what is debug_debug and how does it differ from debug_info?

Also, would it make sense to keep some unused values in between (e.g. if you want to expand on trace with trace_accuracy, trace_perf, ...)?

rjoursler: Agreed, I noticed this naming issue when developing this, but I wanted to get the initial concept reviewed before investing more time renaming, just in case there were disagreements with the methodology.

 > Maybe we should drop debuginfo as it's semantically loaded due to clash with logging levels.
 
I decided against renaming debuginfo. As can be seen by `debug_debug`, the setup of `debug_flags` was already problematic. Instead, I changed `debug_flags` to an enum class `debug_level` which allows dropping the prefix resolving the name clashes for `debug_debug` and `debug_info`. 

> BTW, what is debug_debug and how does it differ from debug_info?

I mainly included this to keep alignment with spdlogs levels as this provides a standardized set of levels. To me, the key difference is that info provides a summary of concrete information where as debug provides "meta" information. To give an example in the  currently implementation, the info level would correspond to the strategy used for jitting a binary and the debug level would correspond to information targeting why that strategy was chosen.

> Also, would it make sense to keep some unused values in between (e.g. if you want to expand on trace with trace_accuracy, trace_perf, ...)?

We could, but I don't see much utility in this. The debuginfo levels mainly target developers, so API stability provides minimal benefit. In addition, 5 levels already appears sufficient in practice, adding more levels just makes it difficult to determine the right mapping.  Instead, the most useful change I forsee is to add more filtering so we can specify something like gpu trace information. Could we add unused values in the future if we encounter issues in practice?

tczeszun (approved):

PR #2699, opened 2025-02-14
Fixups for recent changes.

dzarukin:
get_B_pack_type is not working for int8 on AVX and older.
Windows has slightly higher precision issue with logsoftmax than Linux.
All failures popped-up in Nightly.

### DIFF
--- a/examples/ukernels/cpu_brgemm.cpp
+++ b/examples/ukernels/cpu_brgemm.cpp
@@ -66,8 +66,17 @@ void brgemm_example() {
     // packing requirements once for multiple objects.
     // Based on this information, specific `ldb` value can be used, since
     // transform has a limited set of values supported.
-    const bool need_pack
-            = brgemm::get_B_pack_type(a_dt, b_dt) == pack_type::pack32;
+    bool need_pack = false;
+    try {
+        need_pack = brgemm::get_B_pack_type(a_dt, b_dt) == pack_type::pack32;
+    } catch (error &e) {
+        if (e.status == dnnl_unimplemented)
+            throw example_allows_unimplemented {
+                    "Kernel is not supported on this platform.\n"};
+
+        // on any other error just re-throw
+        throw;
+    }
 
     const memory::dim lda = K;
     // `ldb` for `need_pack = true` must be one of 16, 32, 48, or 64. This
--- a/src/cpu/x64/brgemm/capi/brgemm_api.cpp
+++ b/src/cpu/x64/brgemm/capi/brgemm_api.cpp
@@ -168,7 +168,7 @@ status_t brgemm_t::get_B_pack_type(
     brgemm_utils::set_isa_impl(&brg);
     if (brg.isa_impl == cpu_isa_t::isa_undef) {
         VCHECK_BRGEMM_STATUS(
-                status::invalid_arguments, false, "get_B_pack_type failed");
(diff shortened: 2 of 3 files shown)


avmanerikar (approved):

examples/ukernels/cpu_brgemm.cpp
```diff
@@ -66,8 +66,17 @@ void brgemm_example() {
...
+    bool need_pack = false;
+    try {
+        need_pack = brgemm::get_B_pack_type(a_dt, b_dt) == pack_type::pack32;
+    } catch (error &e) {
+        if (e.status == dnnl_unimplemented)
+            throw example_allows_unimplemented {
+                    "Kernel is not supported on this platform.\n"};
+
+        // on any other error just re-throw
+        throw;
```
densamoilov, line 78:
```suggestion
        std::rethrow_exception(std::current_exception());
```

examples/ukernels/cpu_brgemm.cpp
```diff
@@ -66,8 +66,17 @@ void brgemm_example() {
     // packing requirements once for multiple objects.
     // Based on this information, specific `ldb` value can be used, since
     // transform has a limited set of values supported.
-    const bool need_pack
-            = brgemm::get_B_pack_type(a_dt, b_dt) == pack_type::pack32;
+    bool need_pack = false;
+    try {
+        need_pack = brgemm::get_B_pack_type(a_dt, b_dt) == pack_type::pack32;
```
densamoilov, line 71: Are we sure that we want to handle this scenario with exceptions? Another option would be returning `undef` pack type that would indicate that no type is defined for the given configuration.

dzarukin: Relying on `undef` likely means an API extension with:
* `undef` meaning something went wrong, re-consider using something other than ukernels;
* `no_pack` meaning packing is not needed;
* other valid values depending on the platform.
, which is probably fine, but not in this PR.

Besides, if going this way, something must be done to ukernel object as well. f32 case won't require packing, will return `no_pack` (with new API) but the kernel is not supported on AVX system, and it won't be a ctor but `finalize()` call.

densamoilov: > ndef meaning something went wrong, re-consider using something other than ukernels

For the situations when something went wrong (e.g. invalid arguments) we throw exceptions.

> no_pack meaning packing is not needed;

I couldn't find `no_pack` type, is it something that's going to be introduced? I suggested to rely on `pack_type` based on my understanding that `pack_type` is essentially a layout meaning that anything other than `undef` is OK that is, the operation is defined/supported for the configuration. 

One more thing, how do we expect users who don't want to use exceptions (`--fno-exceptions`) to handle such situations? For primitive descriptors that cannot be created they can use `allow_empty` option but what tools will they have to handle a failed `get_B_pack_type` call?

densamoilov (approved):

examples/ukernels/cpu_brgemm.cpp
dzarukin: Let me come up with an exception free proposal in a different PR. You raised a good concern.

examples/ukernels/cpu_brgemm.cpp
dzarukin: I'll keep this as is since they are going to be removed.

PR #2709, opened 2025-02-17
gpu: generic: fix checks for scales datatype

sgeor255:
# Description

This PR fixes several checks for the datatype of scales which were incorrectly returning `unimplemented`. 

# Checklist

## General

### DIFF
--- a/src/gpu/generic/sycl/ref_binary.hpp
+++ b/src/gpu/generic/sycl/ref_binary.hpp
@@ -87,6 +87,8 @@ struct ref_binary_t : public gpu::generic::sycl::primitive_t {
             const auto &scales = attr()->scales_;
             bool dt_ok = true;
             for (auto arg : supported_args) {
+                if (scales.has_default_values(arg)) continue;
+
                 dt_ok = dt_ok && is_supported_type(scales.get_data_type(arg));
             }
             return dt_ok && attr_scales_ok(supported_args);
--- a/src/gpu/generic/sycl/ref_sum.hpp
+++ b/src/gpu/generic/sycl/ref_sum.hpp
@@ -64,9 +64,9 @@ struct ref_sum_t : public gpu::generic::sycl::primitive_t {
                                 && src_d.ndims() <= xpu::sycl::md_t::max_dims,
                         VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "src");
 
-                VDISPATCH_SUM(attr()->scales_.has_default_values()
-                                || is_supported_type(
-                                        scales.get_data_type(DNNL_ARG_SRC + i)),
+                VDISPATCH_SUM(IMPLICATION(!attr()->scales_.has_default_values(),
+                                      is_supported_type(scales.get_data_type(
+                                              DNNL_ARG_SRC + i))),
                         VERBOSE_UNSUPPORTED_ATTR);
             }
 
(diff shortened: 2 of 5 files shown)


ShanoToni (approved):

src/gpu/generic/sycl/ref_sum.hpp
```diff
@@ -64,9 +64,9 @@ struct ref_sum_t : public gpu::generic::sycl::primitive_t {
                                 && src_d.ndims() <= xpu::sycl::md_t::max_dims,
                         VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "src");
 
-                VDISPATCH_SUM(attr()->scales_.has_default_values()
-                                || is_supported_type(
-                                        scales.get_data_type(DNNL_ARG_SRC + i)),
+                VDISPATCH_SUM(IMPLICATION(!attr()->scales_.has_default_values(),
+                                      is_supported_type(scales.get_data_type(
+                                              DNNL_ARG_SRC + i))),
```
mgouicem, line 69: (nit) Isn't that the same since `IMPLICATION(a, b)` is equivalent to `!a || b` ?

src/gpu/generic/sycl/ref_binary.hpp
```diff
@@ -87,6 +87,8 @@ struct ref_binary_t : public gpu::generic::sycl::primitive_t {
             const auto &scales = attr()->scales_;
             bool dt_ok = true;
             for (auto arg : supported_args) {
+                if (scales.has_default_values(arg)) continue;
+
```
mgouicem, line 91: It might also be worth checking that unsupported args for scales are set to default with 

scales.has_default_values(supported_args); (see [here](https://github.com/oneapi-src/oneDNN/blob/fe85475acb78db87cf2798c93353411760673474/src/common/primitive_attr_quant.hpp#L156))

src/gpu/generic/sycl/ref_sum.hpp
sgeor255: I can't remember why I added this change, I think it was mainly for readability, but yes, you are right that it is the same check. I could revert it if preferable?

mgouicem: This was just to make sure I was not missing something thanks.

src/gpu/generic/sycl/ref_binary.hpp
mgouicem: There are two versions of has_default_values:
- one that take arg_kind, and checks that the scale for that arg_kind is default
- one that takes an std::vector of supported args, and checks that unsupported args are default.

IMHO, it would be safer here to rely on the second flavor and call `scales.has_default_values(supported_args)` outside of the loop.

sgeor255: @mgouicem does this make sense? Or would you still like me to make a change here?

mgouicem: @sgeor255, we need 1 things to be checked:
- that all scales set by user are supported by the implementation. Calling `scales.has_default_values(supported_args)` outside of the loop checks that.
- that all scales supported by the implementation are properly set. The check you currently have inside the loop does that (you are checking if scales are not default before accessing data_type).

Here we would need both and it seems the first one is missing.

sgeor255: I see, thanks, updated it as suggested.

dzarukin (approved):

PR #2711, opened 2025-02-17
x64: pool: Accuracy issue in benchnn pooling test

jstachowintel:
Opening one more time this PR. The previous one was [here](https://github.com/oneapi-src/oneDNN/pull/2678).

It is similar use case as https://github.com/oneapi-src/oneDNN/pull/2627. The problem is with huge input sizes, so I decided disable src_size <= INT_MAX for jit:avx512_core implementation.

### DIFF
--- a/src/cpu/x64/jit_uni_pool_kernel.cpp
+++ b/src/cpu/x64/jit_uni_pool_kernel.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2017-2024 Intel Corporation
+* Copyright 2017-2025 Intel Corporation
 * Copyright 2018 YANDEX LLC
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jstachowintel:
@rjoursler 
Unfortunately, I closed the previous PR, so I opened another one.
Practically I agree that a user can pass an overflowed value for any input attribute, so I added changed has_large_buffers, where I think that we don't need to check wei, dst, and bia memory descs in the pooling case. Changing variable types in [struct](https://github.com/oneapi-src/oneDNN/blob/7437899ec01fb323d99e8e50626a046ca7a70d83/src/cpu/x64/jit_primitive_conf.hpp#L488) on dim_t would be inaccurate because we would never use these extra 4 bytes.

rjoursler:
Thanks for making the change. One note about your comment:

> Changing variable types in [struct](https://github.com/oneapi-src/oneDNN/blob/7437899ec01fb323d99e8e50626a046ca7a70d83/src/cpu/x64/jit_primitive_conf.hpp#L488) on dim_t would be inaccurate because we would never use these extra 4 bytes.

I expect the overall cost of these 4 bytes would be effectively immaterial on real world workloads. The fact that they are unused is largely irrelevant, especially considering the fact that we were just discussing a scenario where they would be used. On the other hand, keeping code type correct provides many benefits as it prevents bugs due to overflows/underflows and it allows code to be better optimized as type conversions prevent many optimization passes performed by compilers.

Overall, I do think directly using the memory descriptor is better in this case. There are a lot of edge cases around calculating the memory size such as sub-byte sized types, non-trivial strides, or even if there is a non-trivial offset0. Just relying on the memory descriptor to handle those cases seems much more maintainable than reimplementing the subset of that logic is applicable to any particular implementation.

rjoursler (approved):

src/cpu/x64/jit_uni_pooling.hpp
inteldimitrius, line 1: The file remains untouched, however the copyright header changed.

inteldimitrius (approved):
LGTM

PR #2717, opened 2025-02-18
ci: linters: path conditioning

hmaciak:
This PR is to provide triggering part of linter jobs on changes in code base only.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Build oneDNN
         run: |
-          .github/automation/build_linters.sh 2>&1 | tee build.log
+          .github/automation/x64/build_linters.sh 2>&1 | tee build.log
           grep -i "warning:" build.log | sort -u
           grep -q -i "warning:" build.log && exit 1 || true
         env:
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -6,8 +6,21 @@ on:
     paths:
       - ".github/automation/x64/**"
       - ".github/workflows/clang-tidy.yml"
+      - "cmake/**"
+      - "examples/**"
+      - "include/**"
+      - "src/common/**"
+      - "src/cpu/*"
+      - "src/cpu/gemm/**"
+      - "src/cpu/matmul/**"
+      - "src/cpu/reorder/**"
+      - "src/cpu/rnn/**"
       - "src/cpu/x64/**"
-      - "src/gpu/intel/**"
+      - "src/gpu/*"
+      - "src/gpu/intel/**""
+      - "src/gpu/graph/**"
+      - "tests/**"
+      - "CMakeLists.txt"
 
 ## Declare default permissions as read only.
 permissions: read-all
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -55,4 +55,3 @@ jobs:
           grep -q -i "warning:" build.log && exit 1 || true
         env:
           ONEDNN_ACTION: build
-          
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -17,7 +17,7 @@ on:
       - "src/cpu/rnn/**"
       - "src/cpu/x64/**"
       - "src/gpu/*"
-      - "src/gpu/intel/**""
+      - "src/gpu/intel/**"
       - "src/gpu/graph/**"
       - "tests/**"
       - "CMakeLists.txt"
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -9,7 +9,7 @@ on:
       - "src/cpu/x64/**"
       - "src/gpu/intel/**"
 
-# Declare default permissions as read only.
+## Declare default permissions as read only.
 permissions: read-all
 
 # Kill stale checks
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -1,0 +1,44 @@
+name: "Clang-Tidy"
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+    paths:
+      - ".github/automation/x64/**"
+      - ".github/workflows/clang-tidy.yml"
+      - "src/cpu/x64/**"
+      - "src/gpu/intel/**"
+
+# Declare default permissions as read only.
+permissions: read-all
+
+# Kill stale checks
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  pr-clang-tidy:
+    name: Clang-Tidy
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-linter.yml
```diff
@@ -31,16 +31,39 @@ concurrency:
...
+    outputs:
+      changed: ${{ steps.changes.outputs.changed }}
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
       - name: Check commit messages
         run: python3 ./.github/automation/commit-msg-check.py "${{ github.event.pull_request.head.sha }}" "${{ github.event.pull_request.base.sha }}"
+      - name: Check codebase changes
```
vpirogov, line 46: Any reason to use this approach instead of [PR trigger rules](https://github.com/oneapi-src/oneDNN/blob/main/.github/workflows/ci-aarch64.yml#L35)?

.github/workflows/pr-linter.yml
```diff
@@ -31,16 +31,39 @@ concurrency:
...
+          if git diff --name-only origin/${{ github.base_ref }}..origin/${{ github.head_ref }}  | grep -q -E "^.github/|^examples/|^include/|^src/|^tests/"; then
+            echo "Found changes in codebase"
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "No changes in codebase"
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  pr-clang-format:
+    name: Formatting
```
vpirogov, line 58: This check must run unconditionally on all PRs so that we can mark it as 'Required'. If a required check does not run for some reason PR promotion gets blocked.

.github/workflows/pr-linter.yml
```diff
@@ -83,6 +108,8 @@ jobs:
   pr-clang-tidy:
     name: Clang-Tidy
     runs-on: ubuntu-latest
+    needs: pr-changes
```
vpirogov, line 111: This check is x64 specific currently, so it needs to run only on cpu-x64 and gpu-intel related changes.

.github/workflows/pr-linter.yml
hmaciak: As far as I know PR trigger rules are applied to all jobs in workflow, and it's not possible to split it to run some of them every time and some of them only on codebase changes.

hmaciak: If you want we can move to this solution https://github.com/dorny/paths-filter , but I'm assuming it's the same case as clang download from KyleMayes repo.

hmaciak:
https://github.com/oneapi-src/oneDNN/actions/runs/13411787185/job/37463465392?pr=2717
https://github.com/oneapi-src/oneDNN/actions/runs/13411867383/job/37463726800?pr=2717
https://github.com/oneapi-src/oneDNN/actions/runs/13411943062/job/37463963635?pr=2717

.github/workflows/pr-linter.yml
vpirogov: > As far as I know PR trigger rules are applied to all jobs in workflow, and it's not possible to split it to run some of them every time and some of them only on codebase changes.

It's reasonable to split this workflow into two: one that runs on every PR (commit messages, clang-format, format tags) and one that runs on x64 changes (clang-tidy).

hmaciak: Done.

theComputeKid:
Thanks. This does make sense. I have a proposal for cutting down x64 time even more: Clang-Tidy doesn't need a build to run, could we:

1. Run configure in CMake to generate compile_commands.json.
2. Run run-clang-tidy.py (or whatever the name of that script is) over that compile_commands.json.

I hypothesise we could save a lot of time. It might not work if we are generating files on the fly.

vpirogov:
> I have a proposal for cutting down x64 time even more: Clang-Tidy doesn't need a build to run, could we:

We discussed this approach with @hmaciak already [here](https://github.com/oneapi-src/oneDNN/pull/2697#issuecomment-2660207461). It would require handling edge cases carefully though, so we'll do this as a separate improvement.

.github/workflows/pr-linter.yml
```diff
@@ -27,20 +27,29 @@ permissions: read-all
...
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  pr-formatting:
-    name: Formatting
-    runs-on: ubuntu-22.04
+  pr-commits:
+    name: Check commits
```
vpirogov, line 35:
```suggestion
    name: Commit messages
```

.github/workflows/pr-linter.yml
```diff
@@ -27,20 +27,29 @@ permissions: read-all
...
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
       - name: Check commit messages
         run: python3 ./.github/automation/commit-msg-check.py "${{ github.event.pull_request.head.sha }}" "${{ github.event.pull_request.base.sha }}"
+
+  pr-clang-format:
+    name: Formatting
```
vpirogov, line 46:
```suggestion
    name: Clang-Format
```

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,45 @@
+name: "Clang-Tidy"
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+    paths:
+      - ".github/automation/x64/**"
+      - ".github/workflows/clang-tidy.yml"
+      - "src/cpu/x64/**"
+      - "src/gpu/intel/**"
```
vpirogov, line 10: We need a bit wider set of sources here :
```suggestion
      - cmake/**
      - examples/**
      - include/**
      - src/common/**
      - src/cpu/*
      - src/cpu/gemm/**
      - src/cpu/matmul/**
      - src/cpu/reorder/**
      - src/cpu/rnn/**
      - src/cpu/x64/**
      - tests/**
      - CMakeLists.txt
```

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,45 @@
...
+          ONEDNN_ACTION: configure
+
+      - name: Build oneDNN
+        run: |
+          .github/automation/x64/build_linters.sh 2>&1 | tee build.log
+          grep -i "warning:" build.log | sort -u
+          grep -q -i "warning:" build.log && exit 1 || true
+        env:
+          ONEDNN_ACTION: build
+
```
vpirogov, line 45:
```suggestion

```

.github/workflows/pr-linter.yml
vpirogov: To avoid messing up required check we can add something along these lines:
```
  status:
    needs: "Commit messages"
    needs: "clang-format"
    needs: "Format tags consistency"
    runs-on: ubuntu-latest
    name: "Formatting"
    steps:
      - name: Print success
        run: echo Success
```

vpirogov (approved):

theComputeKid (approved):

PR #2725, opened 2025-02-20
common: verbose: update logic for printing verbose headers (fixes MFDNN-13224)

avmanerikar:
# Description

Fixes MFDNN-13224.

Querying for verbose mode using `dnnl::impl::get_verbose()` has the unintentional side effect of printing verbose headers upon a match even when no logging is implied. The PR adds an option to disable printing of verbose headers for such cases (An example being the verbose checks made to determine if OpenCL C warnings should be printed or not when building the program from source).

### DIFF
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -85,7 +85,15 @@ static constexpr char verbose_version[] = "v1";
 
 static setting_t<uint32_t> verbose {0};
 
-void print_header(const filter_status_t &filter_status) noexcept {
+// Component filters help manage verbose output by parsing and printing for
+// matching components. The filter status is tracked from verbose initializaton,
+// allowing queries for the component type during verbose printing.
+filter_status_t &filter_status() {
+    static filter_status_t filter_status;
+    return filter_status;
+}
+
+void print_header() noexcept {
     static std::atomic_flag version_printed = ATOMIC_FLAG_INIT;
     if (!version_printed.test_and_set()) {
         verbose_printf("info,oneDNN v%d.%d.%d (commit %s)\n",
@@ -150,16 +158,16 @@ void print_header(const filter_status_t &filter_status) noexcept {
                 "mode,implementation,backend,exec_time\n",
                 get_verbose_timestamp() ? "timestamp," : "");
 #endif
-        if (filter_status.status == filter_status_t::flags::valid)
+        if (filter_status().status == filter_status_t::flags::valid)
             verbose_printf(
                     "common,info,filter format is enabled, hit components: "
                     "%s\n",
-                    filter_status.components.c_str());
-        else if (filter_status.status == filter_status_t::flags::invalid)
+                    filter_status().components.c_str());
+        else if (filter_status().status == filter_status_t::flags::invalid)
             verbose_printf(
                     "common,error,filter format is ill-formed and is not "
                     "applied, error: %s\n",
-                    filter_status.err_msg.c_str());
+                    filter_status().err_msg.c_str());
     }
 }
 
@@ -171,8 +179,6 @@ uint32_t get_verbose(verbose_t::flag_kind verbosity_kind,
 #endif
     // we print all verbose by default
     static int flags = component_t::all;
-    // record filter parsing result to instruct verbose printing
-    static filter_status_t filter_status;
 
     if (!verbose.initialized()) {
         // Assumes that all threads see the same environment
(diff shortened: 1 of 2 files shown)


avmanerikar:
make test
disable test_device_cpu

src/gpu/intel/ocl/ocl_gpu_engine.cpp
```diff
@@ -153,26 +153,29 @@ status_t create_ocl_kernel_from_cache_blob(const ocl_gpu_engine_t *ocl_engine,
...
-    if (!get_verbose(verbose_t::error) && !get_verbose(verbose_t::warn))
-        return err_;
+#if defined(DISABLE_VERBOSE)
+    return err_;
+#endif
+
+    bool is_err = err_ == CL_SUCCESS;
+    bool is_vwarn_enabled = get_verbose(verbose_t::warn);
+
+    if (is_err || !is_vwarn_enabled) return err_;
```
atkassen, line 163:
```suggestion
    bool is_err = err_ != CL_SUCCESS;
    bool is_vwarn_enabled = get_verbose(verbose_t::warn);

    if (!is_err || !is_vwarn_enabled) return err_;
```
I think the use of `is_err` on L169 is correct after this change.

avmanerikar: You are right. Thanks for the catch!

avmanerikar:
make test
disable test_device_cpu

src/gpu/intel/ocl/ocl_gpu_engine.cpp
```diff
@@ -153,26 +153,29 @@ status_t create_ocl_kernel_from_cache_blob(const ocl_gpu_engine_t *ocl_engine,
 cl_int maybe_print_debug_info(
         cl_int err_, cl_program program, cl_device_id dev) {
     // Return error code if verbose is not enabled.
-    if (!get_verbose(verbose_t::error) && !get_verbose(verbose_t::warn))
-        return err_;
+#if defined(DISABLE_VERBOSE)
+    return err_;
+#endif
```
rjoursler, line 158: I don't follow why this  `#ifdef` is necessary.  If VERBOSE is disabled than `get_verbose` should always return `verbose_t::none` which has a value of `0`, so the original if statement should become `if(!0  && !0)`,  causing the early exit to always be taken.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
```diff
@@ -153,26 +153,29 @@ status_t create_ocl_kernel_from_cache_blob(const ocl_gpu_engine_t *ocl_engine,
...
+
+    bool is_err = err_ != CL_SUCCESS;
+    bool is_vwarn_enabled = get_verbose(verbose_t::warn);
+
+    if (!is_err || !is_vwarn_enabled) return err_;
```
rjoursler, line 163: This condition is wrong, the goal here is to log the build information when there is a build error and verbose error logging is enabled or when warnings are requested. We do not want a different condition as the call to `clGetProgramBuildInfo` adds overhead, so the call should only be made when may actually log data.

```Suggestion
     if(!(is_err && get_verbose(verbose_t::error)) || is_vwarn_enabled)) return err_;
```

avmanerikar: Verbose error logging is enabled by default, hence querying for `get_verbose(verbose_t::error)` will trigger the verbose headers to be printed without actually printing anything.  

https://github.com/oneapi-src/oneDNN/blob/46605647d69221411a69c61b33f11bf79d85ca35/src/common/verbose.cpp#L259

> ... the goal here is to log the build information when there is a build error and verbose error logging is enabled or when warnings are requested ...

This condition is then fulfilled and also rechecked on L174 before printing the messages.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
avmanerikar: Check this [comment](https://github.com/oneapi-src/oneDNN/pull/2725#discussion_r1964534634) for the explanation.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
rjoursler: > hence querying for get_verbose(verbose_t::error) will trigger the verbose headers to be printed without actually printing anything

This doesn't seem right. The issue is that querying for verbose support queries should not imply there will be any logging. We will hit the same issue when `verbose_t::warn` is enabled, as that will print a header but there may be no build logs associated with the OpenCL build. In general, the crux of this issue is that we need to change the control flow here based on whether verbose is enabled, as there is no point in executing the OpenCL queries if the results will just be disregarded, but at the same time, we cannot know what the result of the queries are until they are run.

avmanerikar: Thanks for the discussion. The issue stems from the fact that the verbose headers are printed unconditionally when querying for the verbose mode. I have updated the verbose calls accordingly to disable this action.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
avmanerikar: With the latest changes, this update is no longer required..

src/common/verbose.cpp
```diff
@@ -287,7 +287,7 @@ uint32_t get_verbose(verbose_t::flag_kind verbosity_kind,
     int result = verbose.get() & verbosity_kind;
     if (verbosity_kind == verbose_t::debuginfo)
         result = verbose_t::get_debuginfo(verbose.get());
-    if (result) print_header(filter_status);
+    if (result && add_headers) print_header(filter_status);
```
rjoursler, line 290: Why not just delay `print_header` until the first `verbose_printf_impl` call? In general, creating a log while querying for support is a problematic by design as it enables misuse of the interface.

src/gpu/intel/ocl/ocl_gpu_engine.cpp
```diff
@@ -153,26 +153,27 @@ status_t create_ocl_kernel_from_cache_blob(const ocl_gpu_engine_t *ocl_engine,
...
 
     size_t log_length = 0;
     auto err = clGetProgramBuildInfo(
             program, dev, CL_PROGRAM_BUILD_LOG, 0, nullptr, &log_length);
     gpu_assert(err == CL_SUCCESS);
 
-    bool is_err = get_verbose(verbose_t::error) && err_ != status::success;
-    bool is_warn = get_verbose(verbose_t::warn);
-    if (log_length > 1 && (is_err || is_warn)) {
+    bool is_err_status = is_err_mode && err_ != CL_SUCCESS;
```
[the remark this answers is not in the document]
avmanerikar: Fixed in the latest update. Thanks!

src/common/verbose.cpp
avmanerikar: The `print_header` call itself is dependent on `verbose_printf_impl` to print the log headers so checking for the first printed message therein will be problematic. The typical use case for `get_verbose()` is to check for the right verbose level right before printing the message like in [`VINFO`](https://github.com/oneapi-src/oneDNN/blob/efb7e7c4583612a9a2a42873379656be52046ba9/src/common/verbose.hpp#L88), hence, checking and adding the headers here is required. One solution would be to introduce a separate function `query_verbose` set aside for only querying the current verbose level. But for now, the option to disable the headers prevents the unintentional logging of headers.

rjoursler: > The print_header call itself is dependent on verbose_printf_impl to print the log headers so checking for the first printed message therein will be problematic.

I don't see the issue here, `print_header` is guarded by [a latch](https://github.com/oneapi-src/oneDNN/blob/f73bb6cc7925c842d1f0b2e8171f47194032b03e/src/common/verbose.cpp#L89-L90) using an atomic flag. The fact that `print_header` depends on `verbose_printf` does result in small amount of co-recursion, but after the latch is set, everything should just work as expected.

avmanerikar: That's true. But one would also need to check for the component filter along with the verbose mode before printing the headers. While that can be done by calling `get_verbose` within `verbose_print_impl`, it also means having to specify an additional `component_t::flag_kind` parameter to `verbose_printf` and the upstream verbose macros.

rjoursler: > it also means having to specify an additional component_t::flag_kind parameter to verbose_printf

I don't quite follow this either. In `get_verbose`, header printing is handled by the line:
```
if (result) print_header(filter_status);
```
If we move print_header into verbose_printf, we will change the printing behavior because this function returns
```
return filter_result ? result : 0;
```
To me, this change in behavior looks like we are fixing a bug as currently, due to filtering, we can hit the same scenario by printing the header and then never logging anything. 

The other important piece of that line is `filter_status`. As `filter_status` is global variable that is initialized at the same time as verbose. All that should be required is to move the scope of this variable so it is accessible in `verbose_printf`? I don't see any runtime information that needs forwarded, is there something else that I am missing?

avmanerikar: Thanks for the discussion and the comments. The header printing logic has been updated accordingly.

rjoursler (approved):

src/common/verbose.cpp
```diff
@@ -85,7 +85,12 @@ static constexpr char verbose_version[] = "v1";
 
 static setting_t<uint32_t> verbose {0};
 
-void print_header(const filter_status_t &filter_status) noexcept {
+// Component filters help manage verbose output by parsing and printing for
+// matching components. The filter status is tracked from verbose initializaton,
+// allowing queries for the component type during verbose printing.
+static filter_status_t filter_status;
```
dzarukin, line 91: Having an object wrapped in a function in a global space is a better practice as it won't require initialization if not requested.

```suggestion
filter_status_t &filter_status() {
    static filter_status_t filter_status;
    return filter_status;
}
```

dzarukin (approved):

src/common/verbose.cpp
avmanerikar: Thanks! Fixed in the latest commits.

avmanerikar:
make test

PR #2727, opened 2025-02-21
graph: backend: enable genindex ref implementation for gpu

gyhintel:
# Description
The GenIndex op is used to construct SDPA with an implicit causal mask. Since no primitive supports this op, we implement a reference kernel with OCL in Graph.

### DIFF
--- a/src/graph/backend/dnnl/kernels/gen_index.cpp
+++ b/src/graph/backend/dnnl/kernels/gen_index.cpp
@@ -25,6 +25,11 @@
 #include "graph/backend/dnnl/passes/utils.hpp"
 
 #include "graph/backend/dnnl/op_executable.hpp"
+
+#define VCHECK_GENINDEX(cond, status, msg, ...) \
+    VCONDCHECK(graph, create, check, genindex_t, (cond), status, msg, \
+            ##__VA_ARGS__);
+
 namespace dnnl {
 namespace impl {
 namespace graph {
@@ -40,6 +45,16 @@ status_t genindex_t::compile_impl(const dnnl_partition_impl_t *part,
     subgraph_ = std::make_shared<subgraph_t>(part->get_ops(), p_engine_,
             part->get_fpmath_mode(), part->get_use_blocked_layout(), true);
     BACKEND_DNNL_CHECK(set_given_inputs_outputs(subgraph_, inputs, outputs));
+
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE
+    if (p_engine_.get_kind() == engine::kind::gpu) {
+        int ndims = inputs[0].ndims;
+        VCHECK_GENINDEX(ndims <= MAX_NDIMS, status::invalid_arguments,
+                "only tensors of 6 or fewer dimensions are supported for "
+                "genindex GPU, but got %dD",
+                ndims);
+    }
+#endif
 
     subgraph_visualizer_t vis(part->id(), [this](const value_t *val) {
         return this->memory_planner_.get_memory_info(val);
@@ -84,7 +99,7 @@ status_t genindex_t::compile_impl(const dnnl_partition_impl_t *part,
 
 void genindex_t::prepare_args_set(const execution_args_set_t *res,
         const std::vector<tensor_t> &inputs,
-        const std::vector<tensor_t> &outputs, const scratchpad_t &scratchpad) {
+        const std::vector<tensor_t> &outputs) {
     // update the data of partition in/outputs args
     for (const auto &mem_idx : res->get_mems_use_external_inputs()) {
         mem_idx.first.set_data_handle(inputs[mem_idx.second].get_data_handle());
@@ -92,13 +107,6 @@ void genindex_t::prepare_args_set(const execution_args_set_t *res,
     for (const auto &mem_idx : res->get_mems_use_external_outputs()) {
         mem_idx.first.set_data_handle(
                 outputs[mem_idx.second].get_data_handle());
-    }
-
-    grantor_t var_grantor = memory_planner_.internal_temporary_grantor(
-            scratchpad.get_buffer());
-
-    for (auto &mem_offkey : res->get_mems_use_internal_temporary()) {
-        mem_offkey.first.set_data_handle(var_grantor.get(mem_offkey.second));
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/gpu/intel/ocl/graph/genindex.cl
```diff
@@ -0,0 +1,65 @@
...
+
+__kernel void gen_index(__global float *src,  __global int *dst) {
+        int id = get_global_id(0);
+        int result, offset = 0;
+        int idx;
+
+        idx = id % D0;
+        id = id / D0;
+        offset += idx * S0;
+#if AXIS == 0
```
dzarukin, line 25: Would be nice to make `axis` a runtime parameter and adjust the kernel to avoid multiple potential compilations.

src/graph/backend/dnnl/kernels/gen_index.cpp
```diff
@@ -135,7 +135,34 @@ status_t genindex_t::sycl_execute_impl(const stream_t *g_stream,
...
+            returned_event = subgraph_->execs_[i]->execute_sycl(
+                    p_stream, res->get_exec_args()[i], deps);
+            deps = {returned_event};
+        }
+
+        scratchpad.set_deps(returned_event);
+        if (sycl_event) *sycl_event = returned_event;
+
+        return status::success;
+    }
```
dzarukin, line 165: Would it be reasonable to move all the infrastructure to call the kernel next to the kernel the same way it's done for primitives? The kernel itself is not very useful by just laying there...

src/gpu/intel/ocl/graph/genindex.cl
gyhintel: Okay, I have changed the `axis` to a runtime input parameter. Please help take a look at it again. thanks!

src/graph/backend/dnnl/kernels/gen_index.cpp
gyhintel: Yes, It should be more reasonable to iterate through all kernels one by one like the primitives during the compilation stage and execute the compiled kernel during the execution stage.  But there is only one ref kernel for genindex currently so we directly select the chosen one to compile. Can you further share your idea about how to organize the logic? Thanks!

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/gen_index.cpp
dzarukin: What I meant is it would be nicer to have a complete implementation in ocl/graph directory rather than a stripped kernel. Each implementation ends up in a list, but this one would not, that would be the only difference, and you'll grab it mostly same way as you do with other implementations.

gyhintel: Putting the ocl kernel in the gpu/ocl/graph directory is to reuse the gpu infra (CMake code scan, program generation, kernel build, and the conversion of ocl kernel to sycl kernel). 
Now the kernel build and execute call is in the graph backend directory. Now I try to move this logic to ocl/graph directory. I implement a POC in this commit [f147ed7](https://github.com/oneapi-src/oneDNN/commit/f147ed7f67bc47fd38df8fe93d7ac1f5dc7dd17f). How do you think about it? 
Now the kernel creation check has been completed in graph backend. If there are multiple implementations for genindex on GPU. It may be good to put it in the graph backend directory.

gyhintel:
@echeresh @rjoursler Can you help take a look at this PR. Thanks!

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -2490,34 +2502,113 @@ struct genindex_executable_t : public op_executable_t {
...
+            kernel_ = kernels[0];
+        }
+#endif
     }
 
     void execute(const stream &stream,
             const std::unordered_map<int, memory> &args) const override;
 
+#if (DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE) \
+        && (DNNL_GPU_VENDOR == DNNL_VENDOR_INTEL)
```
TaoLv, line 2531: I feel we need to move this check into the function. That means we still need the `execute_sycl` function when the library is built with SYCL runtime, the function raises error when the GPU vendor is not Intel.

TaoLv (approved):

TaoLv:
Anything we need to do with benchdnn to test it?

src/graph/backend/dnnl/op_executable.hpp
gyhintel: Changed. Please take a look at it again. Thanks!

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

gyhintel:
> Anything we need to do with benchdnn to test it?

I started a weekly job to test it on other GPU vendors.

TaoLv:
> > Anything we need to do with benchdnn to test it?
> 
> I started a weekly job to test it on other GPU vendors.

I'm not asking for other GPU vendors. Just wanted to confirm that we have tests which will dispatch the reference kernel added here.

src/gpu/intel/ocl/graph/genindex.cl
```diff
@@ -0,0 +1,53 @@
+/*******************************************************************************
```
echeresh, line 1: Nit: rename `genindex.cl` -> `gen_index.cl` to align with naming of other files.

src/gpu/intel/ocl/graph/genindex.cl
```diff
@@ -0,0 +1,53 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+__kernel void gen_index(__global int *dst, int axis) {
+    int id = get_global_id(0);
+    int result, offset = 0;
+    int idx;
```
rjoursler, line 20: Use `long` or `off_t` to avoid integer overflows on large buffers.

src/gpu/intel/ocl/graph/genindex.cl
gyhintel: Changed. Please take a look again. Thanks!

src/gpu/intel/ocl/graph/genindex.cl
gyhintel: Changed. Please take a look again. Thanks!

gyhintel:
> > > Anything we need to do with benchdnn to test it?
> > 
> > 
> > I started a weekly job to test it on other GPU vendors.
> 
> I'm not asking for other GPU vendors. Just wanted to confirm that we have tests which will dispatch the reference kernel added here.

Now only single genindex op(genindex.json) should be dispatched to reference kernel. The implicit causal mask pattern(sdpa-plain-implicit-causal-mask-fp32-bs1.json) will be dispatched to the micro kernel by default. If we force it to run into ref path, the GPU binary primitive still doesn't support GreaterEqual.

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/gpu/intel/ocl/graph/genindex.cl
rjoursler: The `id` and `offset` variables also look like they need their type changed as well.

gyhintel: Changed. Thanks!

rjoursler (approved):

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2740, opened 2025-02-24
ci: aarch64: Initial nightly performance test

Ryo-not-rio:
Adding initial nightly regression tests. The list of tests is subject to change depending on how crucial each model is and how long it takes to run these tests


Ryo-not-rio:
Closing PR as this needs to be pushed to the oneDNN repo on a separate branch to test the nightly functionality

PR #2741, opened 2025-02-24
benchdnn: inputs: graph: sdpa: add tests for d_qk != d_v

TaoLv:
MFDNN-13232

Add test cases where the head size of QK is different from the head size of V.
The decomposition kernel on CPU is fixed to support the cases.


ElaineBao (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2745, opened 2025-02-24
Minor improvements for arg_usage and post-ops verbose

dzarukin:
PR includes:
* Better handling of dangling symbols on the benchdnn input line.
* `arg_usage` function improvements not to fallback on known arguments but decide on the spot if it's needed or not.
* Verbose diagnostics improvements for post-ops injector and underlying calls.

Last two items are asked by @EgorDuplensky from OV team.

### DIFF
--- a/src/common/convolution_pd.hpp
+++ b/src/common/convolution_pd.hpp
@@ -264,7 +264,8 @@ struct convolution_fwd_pd_t : public convolution_pd_t {
         if (utils::one_of(arg, DNNL_ARG_SRC, DNNL_ARG_WEIGHTS))
             return arg_usage_t::input;
 
-        if (arg == DNNL_ARG_BIAS && with_bias()) return arg_usage_t::input;
+        if (arg == DNNL_ARG_BIAS)
+            return with_bias() ? arg_usage_t::input : arg_usage_t::unused;
 
         if (arg == DNNL_ARG_DST) return arg_usage_t::output;
 
@@ -426,8 +427,8 @@ struct convolution_bwd_weights_pd_t : public convolution_pd_t {
 
         if (arg == DNNL_ARG_DIFF_WEIGHTS) return arg_usage_t::output;
 
-        if (arg == DNNL_ARG_DIFF_BIAS && with_bias())
-            return arg_usage_t::output;
+        if (arg == DNNL_ARG_DIFF_BIAS)
+            return with_bias() ? arg_usage_t::output : arg_usage_t::unused;
 
         return primitive_desc_t::arg_usage(arg);
     }
--- a/src/cpu/x64/injectors/jit_uni_binary_injector.cpp
+++ b/src/cpu/x64/injectors/jit_uni_binary_injector.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2024 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -29,6 +29,9 @@ namespace impl {
 namespace cpu {
 namespace x64 {
 namespace binary_injector {
+
+#define VCHECK_BIN_INJ_BOOL(cond, msg) \
+    VCONDCHECK(primitive, create, check, binary_injector, cond, false, msg);
 
 bcast_set_t get_all_strategies_supported_by_injector() {
     return bcast_set_t {broadcasting_strategy_t::scalar,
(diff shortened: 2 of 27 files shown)


dzarukin:
make test

dzarukin:
make test

mgouicem (approved):

src/cpu/x64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -48,11 +51,14 @@ bool is_alg_supported(alg_kind_t alg) {
 }
 
 bool is_supported(cpu_isa_t isa, alg_kind_t alg, data_type_t dt) {
-    if (dt != data_type::f32) return false;
-
-    return is_isa_supported(isa) && is_alg_supported(alg);
+    VCHECK_ELT_INJ_BOOL(dt == data_type::f32, VERBOSE_UNSUPPORTED_DT);
+    VCHECK_ELT_INJ_BOOL(is_isa_supported(isa), VERBOSE_UNSUPPORTED_ISA);
+    VCHECK_ELT_INJ_BOOL(is_alg_supported(alg), "Unsupported algorithm");
```
avmanerikar, line 56:
```suggestion
    VCHECK_ELT_INJ_BOOL(is_alg_supported(alg), VERBOSE_BAD_ALGORITHM);
```
(nit)

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -90,17 +93,22 @@ bool is_bcast_supported(const dnnl::impl::memory_desc_t &src1_desc,
...
+    return true;
 }
 
 bool is_supported(cpu_isa_t isa, const dnnl::impl::memory_desc_t &src1_desc,
         const memory_desc_wrapper &dst_d,
         const bcast_set_t &supported_strategy_set) {
-    return is_data_supported(isa, src1_desc.data_type)
-            && is_bcast_supported(src1_desc, dst_d, supported_strategy_set);
+    VCHECK_BIN_INJ_BOOL(is_data_supported(isa, src1_desc.data_type),
+            VERBOSE_UNSUPPORTED_DT_CFG);
```
avmanerikar, line 110:
```suggestion
    VCHECK_BIN_INJ_BOOL(is_data_supported(isa, src1_desc.data_type),
            VERBOSE_ISA_DT_MISMATCH);
```
(nit)

avmanerikar (approved):

src/cpu/x64/injectors/jit_uni_eltwise_injector.cpp
dzarukin: It's definitely not a BAD algorithm, the algorithm is valid, just not supported.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
dzarukin: Updated this one, thanks.

src/common/convolution_pd.hpp
```diff
@@ -264,7 +264,8 @@ struct convolution_fwd_pd_t : public convolution_pd_t {
         if (utils::one_of(arg, DNNL_ARG_SRC, DNNL_ARG_WEIGHTS))
             return arg_usage_t::input;
 
-        if (arg == DNNL_ARG_BIAS && with_bias()) return arg_usage_t::input;
+        if (arg == DNNL_ARG_BIAS)
+            return with_bias() ? arg_usage_t::input : arg_usage_t::unused;
```
densamoilov, line 268: Can you please explain why doing this kind of check for all primitive specific arguments is better than doing it in a single place inside `primitive_desc_t::arg_usage` function?

dzarukin: There's no single place, that's the thing. So far it will go till the end of the common function and hopefully reaches the unused return, which could be the signal of error instead saying that the argument wasn't handled properly.

Handling the arg at the spot is just cleaner and wouldn't let collisions happen with the common function.

densamoilov: I'm not sure I understand your point. The handling of the arguments was designed intentionally this way: the primitive-specific arguments are supposed to be checked in their respective primitive-specific functions first. If we reach the end of the primitive-specific function without hitting "return", we assume it's unused, and we handle it in the common function. If a check in the primitive-specific function is missed, we will get a nullptr during execution, which will cause a crash. How does this change modify the behavior?"

> Handling the arg at the spot is just cleaner and wouldn't let collisions happen with the common function.
Collisions are not expected by design, so I don't think it should be a cause for concern.

IMO, checking each primitive-specific argument is redundant because the whole mechanism was designed like a 'switch' statement, where 'unused' is the default case. Checking whether it's the default or not in each 'case' seems redundant to me. Though I don't mind if we merge it since the work has already been done and won't require much maintenance.

densamoilov (approved):

PR #2746, opened 2025-02-24
src: attr: quantization refactor (part 3)

dzarukin:
This is the finishing part of quantization refactor that beautifies skip_mask values for scales and zero-points and introduces a single class for both of them with ability to extend one over the other.


dzarukin:
make test
enable arch_gpu_ampere
enable compiler_icx-oss

mgouicem (approved):
(nit) would it makes sense to also remove usage of DNNL_ARG_ATTR_OUTPUT_SCALES from primitive_exec_types.cpp and miopen_{convolution,inner_product}.cpp as part of this PR?

sgeor255 (approved):

xuxinzen (approved):

echeresh (approved):

dzarukin:
@oneapi-src/onednn-cpu-aarch64 @theComputeKid could you, please, take a look? The last piece unless new bugs revealed.

jondea (approved):

PR #2757, opened 2025-02-26
benchdnn: inputs: rnn: fix shapes leading to unstable results

densamoilov:
When t > 1 the results can become numerically unstable causing some cases to fail. The refactoring of the input files in 10d64ff23e9ebdafa8277db54c3413e3549f3073 introduced some problematic shapes that should be either run with --skip-nonlinear=true or excluded/adjusted. This commit adjusts the shapes as it looks like a more reasonable solution.

### DIFF
--- a/tests/benchdnn/inputs/rnn/shapes_small
+++ b/tests/benchdnn/inputs/rnn/shapes_small
@@ -1,8 +1,8 @@
 # small shapes
 
-l8t12mb12_sic16_n"uniform"
+l8t3mb12_sic16_n"uniform"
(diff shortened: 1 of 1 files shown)


densamoilov:
make test
disable device_gpu
disable benchdnn_all
enable benchdnn_rnn
enable benchdnn_lstm
enable benchdnn_augru
enable benchdnn_gru

dzarukin (approved):

rjoursler (approved):

mgouicem (approved):

PR #2758, opened 2025-02-26
ci: clang-tidy: speed-up clang-tidy

hmaciak:
This PR is to speed up clang-tidy builds by building only changed files.


hmaciak:
Checks are based on compile_command.json file content which should contain all c and cpp files used in full build. 
Also it should contain all necessary dirs which are related to header files. In this configuration dirs like src/cpu/aarch64 or src/cpu/sycl are not prestent in json file, so it should prevent failures regarding checking "wrong" files.

hmaciak:
Example check (source check failed in purpose) https://github.com/oneapi-src/oneDNN/actions/runs/13632125476/job/38102022530?pr=2758

hmaciak:
ping @vpirogov

.github/automation/x64/build_linters.sh
```diff
@@ -17,11 +17,12 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
           -DDNNL_EXPERIMENTAL_PROFILING=ON \
           -DDNNL_EXPERIMENTAL_UKERNEL=ON \
           -DONEDNN_EXPERIMENTAL_LOGGING=ON \
-          -DDNNL_USE_CLANG_TIDY=CHECK \
+          -DDNNL_USE_CLANG_TIDY=CHECK_ALL \
```
vpirogov, line 20: I don't think we need `DNNL_USE_CLANG_TIDY` at all for this check.

.github/automation/x64/build_linters.sh
```diff
@@ -17,11 +17,12 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
...
           -DDNNL_CPU_RUNTIME=OMP \
           -DDNNL_GPU_RUNTIME=OCL \
           -DDNNL_WERROR=ON \
-          -DDNNL_BUILD_FOR_CI=ON
+          -DDNNL_BUILD_FOR_CI=ON \
```
vpirogov, line 24: This is not needed either.

.github/workflows/clang-tidy.yml
```diff
@@ -48,10 +51,28 @@ jobs:
...
-          .github/automation/x64/build_linters.sh 2>&1 | tee build.log
-          grep -i "warning:" build.log | sort -u
-          grep -q -i "warning:" build.log && exit 1 || true
-        env:
-          ONEDNN_ACTION: build
+          touch source-check.log
+          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.(c|cpp)$'); 
+          do
+            if grep -q "$file" "build/compile_commands.json"; then
+              clang-tidy -p build --header-filter='' $file 2>&1 | tee -a source-check.log
```
vpirogov, line 60: Why do we need `--header-filter` here?

.github/workflows/clang-tidy.yml
```diff
@@ -48,10 +51,28 @@ jobs:
...
+          grep -q -i "warning:" source-check.log && exit 1 || true
+
+      - name: Check header files
+        if: always()
+        run: |
+          touch headers-check.log
+          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.(h|hpp)$'); 
+          do
+            header_dir=$(dirname "$file") 
+            if [[ "$header_dir" =~ ^include ]] || grep -q "$header_dir" "build/compile_commands.json"; then
```
vpirogov, line 73: Could you please explain this logic?

.github/workflows/clang-tidy.yml
vpirogov, line 60: Do you see why something like this will not do the trick?

```suggestion
          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.(c|h|cpp|hpp)$'); 
          do
            if grep -q "$file" "build/compile_commands.json"; then
              clang-tidy -p build $file 2>&1 | tee -a source-check.log
```

.github/workflows/clang-tidy.yml
hmaciak: The reason is compile_commands.json does not contain header files. It only contains c/cpp files which are used in full build. If we want to ensure somehow that we check the "right" header files we need to check full path without file name. For example:

headers:
1. src/cpu/file.hpp  -> path is src/cpu -> in compile_commands.json for sure some cpp is in this directory, so we should check this header file.
2. src/cpu/x64/file.hpp  -> path is src/cpu/x64 -> in compile_commands.json for sure some cpp is in this directory, so we should check this header file.
3. src/cpu/aarch64/file.hpp  -> path is src/cpu/aarch64 -> in compile_commands.json for sure there is no cpp in this directory, so we do not check this header file.
4. The only exception are files in include, where for example include/oneapi dir is not present in compile_commands.json.

c/cpp:
1. src/cpu/file.cpp  -> in compile_commands.json this cpp file exists, so we should check this file.
2. src/cpu/x64/file.cpp -> in compile_commands.json this cpp file exists, so we should check this file.
3. src/cpu/aarch64/file.cpp -> in compile_commands.json for sure there is no such cpp file, so we do not check it.


This is the main difference between checking c/cpp files and headers. As you can see headers are more tricky because we don't have the list of exact files.  Also it seems to provide more visibility when we check c/cpp files only and headers separately, but this is only my opinion :)

.github/workflows/clang-tidy.yml
hmaciak: Same comment as above https://github.com/oneapi-src/oneDNN/pull/2758#discussion_r1982419383
If we want to check all in one step, we have to use full paths like in check header step. This may decrease certainty about checked c/cpp files, but it's also possible that it won't decrease it :)

.github/workflows/clang-tidy.yml
vpirogov: Got it, thanks. [Documentation for `clang-tidy`](https://clang.llvm.org/docs/HowToSetupToolingForLLVM.html) indicates that it is expected to run on C/C++ files only. Header files included by sources are analyzed as well if they match the mask defined by `HeaderFilterRegex` in .clang-tidy or '--header-filter' option. So it looks like we should not be invoking `clang-tidy` on header files at all. 

With that we have a question of how to make sure that the changed headers are analyzed. In ideal world this should be done by finding a source that includes particular header and running clang-tidy on it, though this may not be practical. For starters we can assume that header is not touched in isolation and see how it goes.

.github/workflows/clang-tidy.yml
vpirogov: Yep, looks like we'll need some cleanup in headers before we can enable these by default.

.github/workflows/clang-tidy.yml
```diff
@@ -48,10 +51,28 @@ jobs:
...
-      - name: Build oneDNN
+      - name: Check source files
         run: |
-          .github/automation/x64/build_linters.sh 2>&1 | tee build.log
-          grep -i "warning:" build.log | sort -u
-          grep -q -i "warning:" build.log && exit 1 || true
-        env:
-          ONEDNN_ACTION: build
+          touch source-check.log
+          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.(c|cpp)$');
```
vpirogov, line 57: I tested this one and it looks like there's no value in running clang-tidy on `.c` files.

.github/workflows/clang-tidy.yml
vpirogov: I think the options for this part are:
1. Remove it altogether.
2. Use the same logic as for source check, but without the headers filter. In that case this check should only print diagnostics, but not fail the job.

Once we are done with addressing hits in headers #2 will become the default behavior.

.github/workflows/clang-tidy.yml
```diff
@@ -48,10 +51,28 @@ jobs:
...
-          grep -i "warning:" build.log | sort -u
-          grep -q -i "warning:" build.log && exit 1 || true
-        env:
-          ONEDNN_ACTION: build
+          touch source-check.log
+          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.(c|cpp)$'); 
+          do
+            if grep -q "$file" "build/compile_commands.json"; then
+              clang-tidy -p build $file 2>&1 | tee -a source-check.log
+            fi
```
vpirogov, line 61: A little diagnostic printing may be helpful to understand what's going on.

```suggestion
            if grep -q "$file" "build/compile_commands.json"; then
              echo Analyzing $file
              clang-tidy -p build $file 2>&1 | tee -a source-check.log
            else
              echo Skipped $file as it's not used in x64 OpenMP/OpenCL configuration.
            fi
```

.github/workflows/clang-tidy.yml
vpirogov: Upon experimenting I'm leaning towards option (2).

rjoursler: >  For starters we can assume that header is not touched in isolation and see how it goes.

This doesn't seem like a good assumption to me, in my experience this happens relatively frequently, especially for small changes. If we excluded header checks here, we will just hit issues when we later run a full clang-tidy suite.

> With that we have a question of how to make sure that the changed headers are analyzed. In ideal world this should be done by finding a source that includes particular header and running clang-tidy on it

Ideally, this would be handled by the build system, but least last I was looking into it, there are not tools for this. Although there are a few wrappers scripts floating around that try to cache clang-tidy calls. (see https://gitlab.kitware.com/cmake/cmake/-/issues/21092 for some discussions).

> indicates that it is expected to run on C/C++ files only

What if we just generate some psuedo-cpp targets via CMake that we can run clang-tidy against for this check? The main thing that clang-tidy needs to know is how to compile the file, which for headers, implies knowing what preprocessor defines are set. Since most of our `#include` are solely a function of the defines set by the compiler, this will get us 99% of the way to being correct without worrying about dependencies. One other advantage here is it can avoid duplicated logging as some of our core headers could be checked many times.

.github/automation/x64/build_linters.sh
```diff
@@ -17,11 +17,10 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
           -DDNNL_EXPERIMENTAL_PROFILING=ON \
           -DDNNL_EXPERIMENTAL_UKERNEL=ON \
           -DONEDNN_EXPERIMENTAL_LOGGING=ON \
-          -DDNNL_USE_CLANG_TIDY=CHECK \
           -DDNNL_CPU_RUNTIME=OMP \
           -DDNNL_GPU_RUNTIME=OCL \
```
rjoursler, line 21: Side Question: Would it make sense to modify this to target the SYCL runtime? Since Intel's GPU implementations always require OpenCL support handle OpenCL kernels, this switch should result in a  strictly better test.

vpirogov: SYCL runtime would not compile with `clang`, we'll need `icx` at least to generate commands and provide SYCL headers. Just switching the runtime gets you this:

```
$ clang-tidy -p build examples/cnn_inference_f32.cpp
1383 warnings and 21 errors generated.
Error while processing /nfs/pdx/home/vpirogov/work/dnnl/examples/cnn_inference_f32.cpp.
error: too many errors emitted, stopping now [clang-diagnostic-error]
error: unknown argument: '-fno-sycl-id-queries-fit-in-int' [clang-diagnostic-error]
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:32:2: error: "Unsupported compiler" [clang-diagnostic-error]
#error "Unsupported compiler"
 ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:80:15: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
        const sycl::device &adevice, const sycl::context &acontext) {
              ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:80:44: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
        const sycl::device &adevice, const sycl::context &acontext) {
                                           ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:94:8: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
inline sycl::context get_context(const engine &aengine) {
       ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:99:10: error: variable has incomplete type 'void' [clang-diagnostic-error]
    auto ctx = *static_cast<sycl::context *>(ctx_ptr);
         ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:99:29: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    auto ctx = *static_cast<sycl::context *>(ctx_ptr);
                            ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:108:8: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
inline sycl::device get_device(const engine &aengine) {
       ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:113:10: error: variable has incomplete type 'void' [clang-diagnostic-error]
    auto dev = *static_cast<sycl::device *>(dev_ptr);
         ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:113:29: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    auto dev = *static_cast<sycl::device *>(dev_ptr);
                            ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:124:50: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
inline stream make_stream(const engine &aengine, sycl::queue &aqueue) {
                                                 ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:137:8: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
inline sycl::queue get_queue(const stream &astream) {
       ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:142:10: error: variable has incomplete type 'void' [clang-diagnostic-error]
    auto queue = *static_cast<sycl::queue *>(queue_ptr);
         ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:142:31: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    auto queue = *static_cast<sycl::queue *>(queue_ptr);
                              ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:157:1: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
sycl::buffer<T, ndims> get_buffer(const memory &amemory) {
^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:173:29: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    if (!handle_ptr) return sycl::buffer<T, ndims>(sycl::range<1>(1));
                            ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:173:42: error: 'T' does not refer to a value [clang-diagnostic-error]
    if (!handle_ptr) return sycl::buffer<T, ndims>(sycl::range<1>(1));
                                         ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:156:20: note: declared here
template <typename T, int ndims = 1>
                   ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:173:52: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    if (!handle_ptr) return sycl::buffer<T, ndims>(sycl::range<1>(1));
                                                   ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:175:10: error: cannot form a reference to 'void' [clang-diagnostic-error]
    auto &buf_u8 = *static_cast<sycl::buffer<uint8_t, 1> *>(handle_ptr);
         ^
/nfs/pdx/home/vpirogov/work/dnnl/include/oneapi/dnnl/dnnl_sycl.hpp:175:33: error: use of undeclared identifier 'sycl' [clang-diagnostic-error]
    auto &buf_u8 = *static_cast<sycl::buffer<uint8_t, 1> *>(handle_ptr);
                                ^
Suppressed 1495 warnings (1379 in non-user code, 112 NOLINT, 4 with check filters).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
Found compiler error(s).
```

.github/workflows/clang-tidy.yml
vpirogov: > What if we just generate some psuedo-cpp targets which just include the target header via CMake that we can run clang-tidy against for this check?

That's an interesting idea to explore. For headers clang-tidy need a bit more then defines alone:
* Include paths to find dependencies
* Some compile options, like `-std` that affect behavior
Effectively this problem boils down to finding a C++ file in `compile_commands.json` that include particular header and running clang-tidy on it.

I would suggest to not hold this PR though, as it offers material improvement over what we have now.

vpirogov:
Few minor improvements looking at the results on [a test PR](https://github.com/oneapi-src/oneDNN/pull/2839):
1. clang-tidy version matters for the results, let's add a print with version like we do for [clang-format](https://github.com/oneapi-src/oneDNN/blob/main/.github/automation/clang-format.sh#L21).
2. Check should fail on errors as well, now it passes.
3. An empty line before 'Analyzing' would improve readability.

vpirogov (approved):

.github/automation/x64/build_linters.sh
atkassen: Could we use the clang-tidy packaged with the DPC++ compiler?

<details><summary>clang-tidy output</summary>

```
.../compiler/latest/bin/compiler/clang-tidy -p build examples/cnn_inference_f32.cpp
3079 warnings generated.
examples/example_utils.hpp:82:20: warning: function 'validate_engine_kind' defined in a header file; function definitions in header files can lead to ODR violations [misc-definitions-in-headers]
   82 | dnnl::engine::kind validate_engine_kind(dnnl::engine::kind akind) {
      |                    ^
examples/example_utils.hpp:82:20: note: make as 'inline'
   82 | dnnl::engine::kind validate_engine_kind(dnnl::engine::kind akind) {
      |                    ^
      | inline
examples/example_utils.hpp:97:8: warning: invalid case style for struct 'example_allows_unimplemented' [readability-identifier-naming]
   97 | struct example_allows_unimplemented : public std::exception {
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |        example_allows_unimplemented_t
   98 |     example_allows_unimplemented(const char *message) noexcept
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     example_allows_unimplemented_t
examples/example_utils.hpp:110:31: warning: the parameter 'example' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
  110 |         std::function<void()> example) {
      |                               ^
      |         const                &
include/oneapi/dnnl/dnnl.hpp:4204:5: warning: return type 'const post_ops' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type]
 4204 |     const post_ops get_post_ops() const {
      |     ^~~~~
include/oneapi/dnnl/dnnl.hpp:4223:38: warning: the const qualified parameter 'ops' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param]
 4223 |     void set_post_ops(const post_ops ops) {
      |                                      ^
      |                                     &
include/oneapi/dnnl/dnnl_common.hpp:131:67: warning: use nullptr [modernize-use-nullptr]
  131 |     std::shared_ptr<typename std::remove_pointer<T>::type> data_ {0};
      |                                                                   ^
      |                                                                   nullptr
include/oneapi/dnnl/dnnl_common.hpp:380:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  380 |     inline enum_name operator&(enum_name lhs, enum_name rhs) { \
      |            ^
      |            (        )
include/oneapi/dnnl/dnnl_common.hpp:385:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  385 |     inline enum_name operator^(enum_name lhs, enum_name rhs) { \
      |            ^
      |            (        )
include/oneapi/dnnl/dnnl_common.hpp:390:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  390 |     inline enum_name &operator|=(enum_name &lhs, enum_name rhs) { \
      |            ^
      |            (        )
include/oneapi/dnnl/dnnl_common.hpp:390:34: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  390 |     inline enum_name &operator|=(enum_name &lhs, enum_name rhs) { \
      |                                  ^
      |                                  (        )
include/oneapi/dnnl/dnnl_common.hpp:396:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  396 |     inline enum_name &operator&=(enum_name &lhs, enum_name rhs) { \
      |            ^
      |            (        )
include/oneapi/dnnl/dnnl_common.hpp:396:34: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  396 |     inline enum_name &operator&=(enum_name &lhs, enum_name rhs) { \
      |                                  ^
      |                                  (        )
include/oneapi/dnnl/dnnl_common.hpp:402:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  402 |     inline enum_name &operator^=(enum_name &lhs, enum_name rhs) { \
      |            ^
      |            (        )
include/oneapi/dnnl/dnnl_common.hpp:402:34: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  402 |     inline enum_name &operator^=(enum_name &lhs, enum_name rhs) { \
      |                                  ^
      |                                  (        )
include/oneapi/dnnl/dnnl_common.hpp:408:12: warning: macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
  408 |     inline enum_name operator~(enum_name rhs) { \
      |            ^
      |            (        )
Suppressed 3149 warnings (3043 in non-user code, 106 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

</details>

hmaciak:
ping @theComputeKid

Sqvid:
Hi @hmaciak thank you for these changes. @theComputeKid is away from the office for the next few weeks so he may or may not be able to provide a timely review. I'd be happy to approve the PR, though I have left a comment above that you could hopefully comment on. Thanks.

Sqvid (approved):

.github/workflows/clang-tidy.yml
```diff
@@ -48,10 +51,36 @@ jobs:
...
-          .github/automation/x64/build_linters.sh 2>&1 | tee build.log
-          grep -i "warning:" build.log | sort -u
-          grep -q -i "warning:" build.log && exit 1 || true
-        env:
-          ONEDNN_ACTION: build
+          echo -e "Checking Clang-Tidy $(clang-tidy --version)\n"
+          touch source-check.log
+          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.cpp$'); 
+          do
+            if grep -q "$file" "build/compile_commands.json"; then
```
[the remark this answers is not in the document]
Sqvid: Ah it looks like the workflow is scanning the x64 paths anyway. So I don't think my comment is relevant.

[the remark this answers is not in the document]
hmaciak: Yep, this config is for x64 specific changes :)

PR #2768, opened 2025-02-27
build: nvidia: fix install dependencies

sgeor255:
# Description

Currently when oneDNN built and installed with `GPU_VENDOR=NVIDIA`, cuDNN and cuBLAS are not correctly propagated to the exported `DNNL::dnnl` target, resulting in linker errors when used in consumer applications. This PR fixes this issue.


Rbiessy (approved):

src/CMakeLists.txt
```diff
@@ -297,6 +297,14 @@ if(DNNL_BLAS_VENDOR STREQUAL "ACCELERATE")
     DESTINATION ${LIB_CONFIG_INSTALL_DIR})
 endif()
 
+if((DNNL_GPU_RUNTIME STREQUAL "DPCPP" OR DNNL_GPU_RUNTIME STREQUAL "SYCL") AND DNNL_GPU_VENDOR STREQUAL "NVIDIA")
+    install(FILES
+        "../cmake/FindcuBLAS.cmake"
+        "../cmake/FindcublasLt.cmake"
+        "../cmake/FindcuDNN.cmake"
+        DESTINATION ${LIB_CONFIG_INSTALL_DIR})
+endif()
```
mgouicem, line 306: Do I get it right that you are adding those cmake files to the install target?
Also why in `src/CMakeLists.txt` and not in `cmake/SYCL.cmake` ?

src/CMakeLists.txt
```diff
@@ -297,6 +297,14 @@ if(DNNL_BLAS_VENDOR STREQUAL "ACCELERATE")
     DESTINATION ${LIB_CONFIG_INSTALL_DIR})
 endif()
 
+if((DNNL_GPU_RUNTIME STREQUAL "DPCPP" OR DNNL_GPU_RUNTIME STREQUAL "SYCL") AND DNNL_GPU_VENDOR STREQUAL "NVIDIA")
```
mgouicem, line 300: (nit)
```suggestion
if(DNNL_SYCL_CUDA)
```

mgouicem (commented):
Thanks for the PR. Do you know why we don't have a similar issue with other packages (e.g. HIP, MIOPEN, rocBLAS, ACL, TBB, ...)?

@densamoilov @spalicki

sgeor255:
> Thanks for the PR. Do you know why we don't have a similar issue with other packages (e.g. HIP, MIOPEN, rocBLAS, ACL, TBB, ...)?
> 
> @densamoilov @spalicki

Thanks for the review @mgouicem . I am not sure about the AMD-related dependencies as I haven't tested, but I think it's likely that the same issue is present and a similar change will be needed.

 TBB it is handled similarly to my changes here: https://github.com/oneapi-src/oneDNN/blob/main/cmake/config.cmake.in#L24 (in the same file I added the changes for finding cuDNN and cuBLAS). Below TBB, there's similar code for openCL and BLAS.

src/CMakeLists.txt
sgeor255: > Do I get it right that you are adding those cmake files to the install target?

Yes, with this change those files will get installed as they are used by the `find_package(...)` calls in the generated `dnnl-config.cmake` file in `install_dir/lib/cmake`.

> Also why in src/CMakeLists.txt and not in cmake/SYCL.cmake ?

I saw that a few other files are installed here and thought it would be good to keep all similar calls in one place. I am happy to move them to `SYCL.cmake` if that's preferable?

src/CMakeLists.txt
sgeor255: Done, thanks!

mgouicem (approved):

PR #2770, opened 2025-02-27
cpu: x64: conv: update estimate_brgemm_ur for better blocking

xuxinzen:
Fixes MFDNN-12557

Missing initialized some variables in `estimate_brgemm_ur() `(like zp_type_a, max_vpad) may cause some shapes fall into `exec_base` instead of `exec_vpad`.

The perf data was collect on ADL with 6 core per sockets.
image
[brg_zp_avx2_vnni.xlsx](https://github.com/user-attachments/files/19017009/brg_zp_avx2_vnni.xlsx)


xuxinzen:
make test

inteldimitrius (approved):
LGTM

tprimak (approved):

PR #2779, opened 2025-02-28
tests: benchdnn: update skip impl prb on deconv for gpu

xuxinzen:
Fixes MFDNN-13276

### DIFF
--- a/tests/benchdnn/deconv/deconv.cpp
+++ b/tests/benchdnn/deconv/deconv.cpp
@@ -345,11 +345,14 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
             prb->attr, res, dnnl_deconvolution, prb->get_dt(SRC));
     skip_unimplemented_prelu_po(prb->attr, res, dnnl_deconvolution);
 
-    // GPU supports only post ops and all but x8s8bf16 cfg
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test

tests/benchdnn/deconv/deconv.cpp
```diff
@@ -345,11 +345,14 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
             prb->attr, res, dnnl_deconvolution, prb->get_dt(SRC));
     skip_unimplemented_prelu_po(prb->attr, res, dnnl_deconvolution);
 
-    // GPU supports only post ops and all but x8s8bf16 cfg
+    // GPU supports only post ops and all but x8s8bf16 and f32xf16f32 cfg
```
dzarukin, line 348: The generic support for f32-xf16 deconv has just landed. Where unimplemented cases are coming from?

xuxinzen: f32xf16 has just landed on CPU. But according to GPU team, f32xf16 is not supported and the detail of the failure cases can be found in MFDNN-13276.

dzarukin: GPU... my bad, thanks for clarifications.

dzarukin (approved):

echeresh (approved):

PR #2790, opened 2025-03-01
[GPU] Shapeless conv: add scales support

echeresh:
Jira: [MFDNN-13245](https://jira.devtools.intel.com/browse/MFDNN-13245)

### DIFF
--- a/src/common/primitive_attr_quant.hpp
+++ b/src/common/primitive_attr_quant.hpp
@@ -104,6 +104,8 @@ struct quant_entry_t : public c_compatible {
 
     void serialize(serialization_stream_t &sstream) const;
 
+    static quant_entry_t deserialize(deserializer_t &d);
+
     std::string get_verbose() const;
 
 private:
@@ -253,6 +255,8 @@ struct scales_t : public quant_entries_t {
         return quant_entries_t::has_default_data_type(arg);
     }
 
+    static scales_t deserialize(deserializer_t &d);
+
 private:
     static constexpr data_type_t default_data_type_ = data_type::f32;
 
@@ -299,6 +303,8 @@ struct zero_points_t : public quant_entries_t {
     bool has_default_data_type(int arg) const {
         return quant_entries_t::has_default_data_type(arg);
     }
+
+    static zero_points_t deserialize(deserializer_t &d);
 
 private:
     static constexpr data_type_t default_data_type_ = data_type::s32;
--- a/src/gpu/intel/jit/v2/conv/builder.cpp
+++ b/src/gpu/intel/jit/v2/conv/builder.cpp
@@ -293,10 +293,18 @@ public:
                         = buf_mgr.get("bias_reduced", plan.bias_layout.size());
         }
 
+        arg_helper_t arg_helper(desc);
+        for (int arg : {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST}) {
+            if (desc.scales.has_default_values(arg)) continue;
+            auto name = arg_helper.scales_name(arg);
+            auto &e = entries_[name];
+            e.mem_buf = var_mgr.get_arg(name);
+        }
+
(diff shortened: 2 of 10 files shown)


src/common/primitive_attr_quant.hpp
```diff
@@ -80,6 +80,8 @@ struct quant_entry_t : public c_compatible {
 
     int get_mask() const { return mask_; }
     data_type_t get_data_type() const { return data_type_; }
+    int get_group_ndims() const { return group_ndims_; }
+    const dims_t &get_group_dims() const { return group_dims_; }
```
dzarukin, line 84: Discussed offline. The proposal is to introduce a `deserialize` method to quant_entry, scales and zero-points instead of exposing internals.

echeresh: Updated to use common API.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/common/primitive_attr_quant.hpp
```diff
@@ -253,6 +255,8 @@ struct scales_t : public quant_entries_t {
         return quant_entries_t::has_default_data_type(arg);
     }
 
+    static scales_t deserialize(deserializer_t &d);
```
dzarukin, line 258: I think you can move it to quant_entries_t and have a single interface.

dzarukin (approved):

src/common/primitive_attr_quant.hpp
echeresh: `deserialize()` API  is type-dependent so it can't be inherited from the base class in this form:

https://github.com/uxlfoundation/oneDNN/blob/70a851368209585354b6e40d273aac30ec3b8202/src/common/serialization.hpp#L214-L216

So we have to have explicit static interfaces in both classes. However I added helper `deserialize_entries()` which has the common logic so nu much duplication.

In theory we can switch to two-step deserialization with member function (which can be inherited):

```c++
T t;
t.deserialize();
```

This approach has some minor cons though: the default constructor must be always present and public in `T` and we have less control over the object state preceding deserialization (what if it was "normally" constructed? do we need to reset it to the default state before starting deserialization?).

kealan-barbieri (approved):

src/gpu/intel/jit/v2/conv/builder.cpp
```diff
@@ -642,11 +651,41 @@ class epilogue_tile_builder_t : public ir_builder_t {
...
+        auto to_rhs_mask = [](int arg, int mask) {
+            if (mask == 0) return 0;
+            switch (arg) {
+                case DNNL_ARG_WEIGHTS:
+                    gpu_assert(mask == ((1 << 0) | (1 << 1)))
+                            << "Only per-output channels scales are supported, "
+                               "mask: "
+                            << mask;
+                    return (1 << 1);
+                case DNNL_ARG_DST: return mask;
```
kealan-barbieri, line 663: no limits on dst scale? Are groups future work?

echeresh: Yes, I think so. Group scales are not supported by convolution (checked in `src/common/convolution.cpp`) and the only supported masks are "common" and "per-channel" (also checked, both supported).

This is where it's checked in common:

https://github.com/uxlfoundation/oneDNN/blob/2b248c42a72a9cff77e01e9d837a34e676f5a1ca/src/common/convolution.cpp#L176-L184

PR #2794, opened 2025-03-01
[GPU] [conv:v2] add strided tensor support

dyoussif:
addresses MFDNN-12455
Assumes innermost dimension has stride 1.
1) Add placeholder stride vars for each tensor.
2) use stride vars in layout_t instead of calcuating based off dims.
3) Infer stride_var reqs for mod and gt/lt reqs :
examples
`ic % 16 == 0  => stride_iw % 16 ==0, stride_ih % 16 ==0, stride_mb % 16 == 0`
`ic*iw < 256 => stride_iw*iw<256  stride_ih < 256 `

This approach was chosen as opposed to limiting spatial dimensions to be dense because it is more general
and reqs would still need to be inferred/generated for non-spatial dimension stride vars.

### DIFF
--- a/src/gpu/intel/jit/conv/problem.cpp
+++ b/src/gpu/intel/jit/conv/problem.cpp
@@ -47,6 +47,18 @@ const std::vector<pvar_t> &conv_dims() {
     return _conv_dims;
 }
 
+pvar_t prb_stride(const pvar_t &dim, tensor_kind_t tensor_kind) {
+
+    auto dims = conv_layout_dims(tensor_kind, true);
+    for (auto &d : dims) {
+        if (d == dim) {
+            auto str = to_string(tensor_kind) + "_";
+            return pvar_t(str + dim.str() + "_stride");
+        }
+    }
+    return pvar_t();
+}
+
 const std::vector<pvar_t> &conv_index_dims(prop_kind_t prop) {
     auto get_dims = [&](prop_kind_t prop) {
         std::vector<pvar_t> ret;
--- a/src/gpu/intel/jit/v2/conv/bridge.hpp
+++ b/src/gpu/intel/jit/v2/conv/bridge.hpp
@@ -58,6 +58,69 @@ inline pvar_tile_t to_shape(const convolution_pd_t *pd) {
     shape[pvars::pd] = pd->padFront();
     shape[pvars::ph] = pd->padT();
     shape[pvars::pw] = pd->padL();
+    memory_desc_wrapper mdw_src(pd->invariant_src_md());
+    memory_desc_wrapper mdw_wei(pd->invariant_wei_md());
+    memory_desc_wrapper mdw_dst(pd->invariant_dst_md());
+
+    bool src_strided = (mdw_src.is_plain() && !mdw_src.is_dense());
+    bool wei_strided = (mdw_wei.is_plain() && !mdw_wei.is_dense());
+    bool dst_strided = (mdw_dst.is_plain() && !mdw_dst.is_dense());
+    auto get_stride = [&](const memory_desc_t *md, int index,
+                              int non_spatial_ndims = 2) -> dim_t {
+        if (pd->ndims() > index) {
+            index = index >= non_spatial_ndims
+                    ? pd->ndims() - (index + 1 - non_spatial_ndims)
+                    : index;
+            if (md->dims[index] == 1) return 0;
+            return md->format_desc.blocking.strides[index];
+        }
+        return 0;
+    };
+    if (src_strided) {
+        shape[prb_stride(pvars::mb, tensor_kind_t::src)]
+                = get_stride(pd->invariant_src_md(), 0);
+        shape[prb_stride(pvars::ic, tensor_kind_t::src)]
+                = get_stride(pd->invariant_src_md(), 1);
+        shape[prb_stride(pvars::g, tensor_kind_t::src)]
+                = shape[prb_stride(pvars::ic, tensor_kind_t::src)]
+                * shape[pvars::ic];
+        shape[prb_stride(pvars::id, tensor_kind_t::src)]
+                = get_stride(pd->invariant_src_md(), 4);
+        shape[prb_stride(pvars::ih, tensor_kind_t::src)]
+                = get_stride(pd->invariant_src_md(), 3);
+        shape[prb_stride(pvars::iw, tensor_kind_t::src)]
+                = get_stride(pd->invariant_src_md(), 2);
+    }
+    if (wei_strided) {
+        shape[prb_stride(pvars::g, tensor_kind_t::wei)] = pd->with_groups()
+                ? get_stride(pd->invariant_wei_md(), 0, 3)
+                : 0;
+        shape[prb_stride(pvars::oc, tensor_kind_t::wei)]
+                = get_stride(pd->invariant_wei_md(), 1, 3);
+        shape[prb_stride(pvars::ic, tensor_kind_t::wei)]
+                = get_stride(pd->invariant_wei_md(), 2, 3);
+        shape[prb_stride(pvars::kd, tensor_kind_t::wei)]
+                = get_stride(pd->invariant_wei_md(), 5, 3);
+        shape[prb_stride(pvars::kh, tensor_kind_t::wei)]
+                = get_stride(pd->invariant_wei_md(), 4, 3);
+        shape[prb_stride(pvars::kw, tensor_kind_t::wei)]
+                = get_stride(pd->invariant_wei_md(), 3, 3);
+    }
+    if (dst_strided) {
+        shape[prb_stride(pvars::mb, tensor_kind_t::dst)]
+                = get_stride(pd->invariant_dst_md(), 0);
+        shape[prb_stride(pvars::oc, tensor_kind_t::dst)]
+                = get_stride(pd->invariant_dst_md(), 1);
+        shape[prb_stride(pvars::g, tensor_kind_t::dst)]
+                = shape[prb_stride(pvars::oc, tensor_kind_t::dst)]
+                * shape[pvars::oc];
+        shape[prb_stride(pvars::od, tensor_kind_t::dst)]
+                = get_stride(pd->invariant_dst_md(), 4);
+        shape[prb_stride(pvars::oh, tensor_kind_t::dst)]
+                = get_stride(pd->invariant_dst_md(), 3);
+        shape[prb_stride(pvars::ow, tensor_kind_t::dst)]
+                = get_stride(pd->invariant_dst_md(), 2);
+    }
     return shape;
 }
 
--- a/src/gpu/intel/jit/v2/conv/kernel_desc.cpp
+++ b/src/gpu/intel/jit/v2/conv/kernel_desc.cpp
@@ -249,7 +249,11 @@ bool is_grf_usage_ok(const kernel_desc_t &desc) {
 }
 
 prb_reqs_t kernel_desc_t::reqs() const {
-    return generate_2d_reqs(*this);
+    auto reqs = generate_2d_reqs(*this);
+    set_stride_reqs(tensor_kind_t::src, src_tag, reqs);
+    set_stride_reqs(tensor_kind_t::wei, wei_tag, reqs);
+    set_stride_reqs(tensor_kind_t::dst, dst_tag, reqs);
+    return reqs;
 }
 
 bool kernel_desc_t::is_supported(const hw_t &hw, const problem_t *prb) const {
@@ -332,6 +336,15 @@ void kernel_desc_t::set_missing() {
     }
 }
 
+void kernel_desc_t::set_stride_reqs(const tensor_kind_t kind,
+        const layout_tag_t &desc_tag, prb_reqs_t &reqs) const {
+    if (!desc_tag.is_strided()) return;
+    auto tag = append_groups(kind, desc_tag, is_dw);
+    auto &entries = tag.raw_tag().entries();
+    auto dim = tag.desc().prb_dim(entries.rbegin()->index());
+    reqs.add(prb_stride(dim, kind).var() == expr_t(1));
+}
+
 bool is_compatible(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
         const problem_t &prb, bool exact) {
     auto &desc_tag = kernel_desc.layout_tag(abc);
@@ -347,6 +360,10 @@ bool is_compatible(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
     if (!type_ok && is_out && kernel_desc.use_stream_k) type_ok = true;
     gpu_check(type_ok) << to_string(abc) << " tag " << prb_tag
                        << " does not match kernel descriptor tag " << desc_tag;
+    if (prb_tag.is_strided())
+        gpu_check(prb_tag.matches(desc_tag, prb.shape(), /*check_type=*/false))
+                << to_string(abc) << " tag " << prb_tag
+                << " does not match kernel descriptor tag " << desc_tag;
     return true;
 }
 
@@ -403,9 +420,10 @@ void fit_tag_to(
     auto &prb_tag = prb.layout_tag(abc);
     bool is_out_stream_k
             = (abc == tensor_kind_t::c) && kernel_desc.use_stream_k;
-    if (desc_tag.type() != prb_tag.type() && !is_out_stream_k) {
-        desc_tag = layout_tag_t(
-                desc_tag.desc(), prb_tag.type(), desc_tag.raw_tag());
+    if ((desc_tag.type() != prb_tag.type() && !is_out_stream_k)
+            || prb_tag.is_strided()) {
+        desc_tag = layout_tag_t(desc_tag.desc(), prb_tag.type(),
+                desc_tag.raw_tag(), prb_tag.is_strided());
     }
 }
 
@@ -425,7 +443,11 @@ void fit_to_impl(kernel_desc_t &desc, const problem_t &prb) {
 }
 
 bool kernel_desc_t::can_fit(const problem_t &prb) const {
-    return is_compatible(*this, prb, /*exact=*/false);
+    auto desc = *this;
+    desc.src_tag.set_strided(prb.src_tag().is_strided());
+    desc.wei_tag.set_strided(prb.wei_tag().is_strided());
+    desc.dst_tag.set_strided(prb.dst_tag().is_strided());
+    return is_compatible(desc, prb, /*exact=*/false);
 }
 
 void kernel_desc_t::fit_to(const problem_t &prb) {
@@ -758,6 +780,19 @@ void kernel_desc_t::init_kernel_iface(kernel_iface_t &kernel_iface) const {
         kernel_iface.register_arg(var);
         if (d == pvars::sw)
             kernel_iface.register_arg("sw_magic", type_t::u64());
+        if (!is_conv_index(d)) continue;
+        for (auto &t_kind :
+                {tensor_kind_t::src, tensor_kind_t::wei, tensor_kind_t::dst}) {
+            auto &tag = layout_tag(t_kind);
+            if (!tag.is_strided()) continue;
+            auto inner_dim = tag.raw_tag().entries().back();
+            if (prb_stride(d, t_kind).is_undef()
+                    || tag.desc().prb_dim(inner_dim.index()) == d)
+                continue;
+            auto stride_var
+                    = var_t::make(type_t::s32(), prb_stride(d, t_kind).str());
+            kernel_iface.register_arg(stride_var);
+        }
     }
     if (use_stream_k) {
         kernel_iface.register_arg("sk_iters_per_tile_main", type_t::s32());
@@ -1021,7 +1056,9 @@ status_t kernel_desc_t::init_primitive_plan(primitive_init_plan_t &plan,
                                  : jit::layout_t(md, /*do_normalize=*/false));
         bool is_out_stream_k = use_stream_k && t.is_output;
         bool zero_out = is_out_stream_k;
-        if (compute_layout != user_layout) {
+        if (compute_layout != user_layout
+                && !compute_layout.normalize().is_strictly_equal(
+                        user_layout.normalize(), true, false)) {
             user_name += "_user";
             scratchpad_key++;
             pd->scratchpad_registry().registrar().book(
@@ -1034,7 +1071,8 @@ status_t kernel_desc_t::init_primitive_plan(primitive_init_plan_t &plan,
         plan.add_user_buffer(user_name, user_layout, t.is_input, t.is_output,
                 t.arg_key, zero_out);
         if (user_name == t.name) {
-            gpu_assert(user_layout == compute_layout)
+            gpu_assert(compute_layout.normalize().is_strictly_equal(
+                    user_layout.normalize(), true, false))
                     << "Incompatible user/kernel layouts. User: "
                     << user_layout.str()
                     << ", kernel: " << compute_layout.str();
--- a/src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
+++ b/src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
@@ -145,6 +145,24 @@ block_2d_params_t to_block_2d_params(const prop_kind_t &prop,
     return params;
 }
 
+void generate_2d_stride_var_reqs(const layout_tag_t &tag, prb_reqs_t &reqs,
+        const tensor_kind_t &tensor_kind, const block_2d_params_t &params,
+        const int base_align) {
+    using ir_utils::safe_div;
+    std::vector<pvar_t> pvars;
+    for (int i = tag.raw_tag().nentries() - 2; i >= 0; i--) {
+        auto &e = tag.raw_tag().entries()[i];
+        gpu_assert(!e.is_blocked);
+        auto dim = tag.desc().prb_dim(e.index());
+        if (utils::one_of(dim, params.w_dim, params.h_dim)) continue;
+        reqs.add_no_simplify(
+                prb_stride(dim, tensor_kind).var() % base_align == 0);
+    }
+    auto &e = tag.raw_tag().entries()[tag.raw_tag().nentries() - 1];
+    auto dim = tag.desc().prb_dim(e.index());
+    reqs.add_no_simplify(dim.var() % base_align == 0);
+}
+
 void generate_2d_reqs(const kernel_desc_t &desc, tensor_kind_t tensor_kind,
         prb_reqs_t &reqs) {
     using ir_utils::safe_div;
@@ -160,8 +178,14 @@ void generate_2d_reqs(const kernel_desc_t &desc, tensor_kind_t tensor_kind,
         auto &e = tag.raw_tag().entries()[i];
         gpu_assert(!e.is_blocked);
         auto dim = tag.desc().prb_dim(e.index());
-        strides[dim] = stride;
-        stride *= dim;
+        if (tag.is_strided() && i < tag.raw_tag().nentries() - 1) {
+            stride_t s_stride(1);
+            s_stride *= prb_stride(dim, tensor_kind);
+            strides[dim] = s_stride;
+        } else {
+            strides[dim] = stride;
+            stride *= dim;
+        }
     }
     int type_size = tag.type().size();
     auto params = to_block_2d_params(desc.prop, tensor_kind, type_size,
@@ -186,7 +210,10 @@ void generate_2d_reqs(const kernel_desc_t &desc, tensor_kind_t tensor_kind,
     reqs.add_no_simplify(
             P.mod(safe_div(block_2d_pitch_alignment(desc.hw_desc), type_size))
             == 0);
-    reqs.add_no_simplify(params.base_stride.mod(base_align) == 0);
+    if (tag.is_strided())
+        generate_2d_stride_var_reqs(tag, reqs, tensor_kind, params, base_align);
+    else
+        reqs.add_no_simplify(params.base_stride.mod(base_align) == 0);
     if ((is_fwd || is_bwd_w) && params.h_dim == pvars::iw) {
         reqs.add_no_simplify((params.y_stride == 1) | (pvars::pw.var() == 0));
         reqs.add_no_simplify((params.y_stride == 1) | (pvars::kw.var() == 1));
--- a/src/gpu/intel/jit/v2/conv/plan_registry.cpp
+++ b/src/gpu/intel/jit/v2/conv/plan_registry.cpp
@@ -66,8 +66,9 @@ kernel_desc_t plan_registry_t::find_best(
         auto sk_desc = to_stream_k(e.desc);
         sk_desc.spec.mode = spec_mode;
         sk_desc.spec.specialize(prb);
+        if (sk_desc.is_empty()) continue;
         gpu_trace() << "Trying kernel desc: " << sk_desc.cmd_str();
-        if (sk_desc.is_empty() || !sk_desc.can_fit(prb)) continue;
+        if (!sk_desc.can_fit(prb)) continue;
         time = e.model_set.time(prb, sk_desc);
         if (time < min_time) {
             min_time = time;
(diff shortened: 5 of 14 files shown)


dyoussif:
make test
disable device_cpu
disable run_scans
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv

src/gpu/intel/jit/v2/conv/bridge.hpp
```diff
@@ -58,6 +58,71 @@ inline pvar_tile_t to_shape(const convolution_pd_t *pd) {
...
+    memory_desc_wrapper mdw_src(pd->invariant_src_md());
+    memory_desc_wrapper mdw_wei(pd->invariant_wei_md());
+    memory_desc_wrapper mdw_dst(pd->invariant_dst_md());
+
+    bool src_strided = (mdw_src.is_plain() && !mdw_src.is_dense());
+    bool wei_strided = (mdw_wei.is_plain() && !mdw_wei.is_dense());
+    bool dst_strided = (mdw_dst.is_plain() && !mdw_dst.is_dense());
+    if (src_strided) {
+        shape[pvars::src_mb_stride]
+                = pd->invariant_src_md()->format_desc.blocking.strides[0];
```
rjoursler, line 70: The value in the blocking descriptor is undefined if the dimension size is 1. Please provide a reasonable default value in this case.

dyoussif: Thanks for pointing this out.  Dimensions with size 1 should now have a default stride of 1.

rjoursler: Could setting the stride to `1` cause unintended conflicts with alignment checks? If a dimension size, the dimension should be irrelevant for alignment checks, but checks for `stride % value` will fail.

dyoussif: It looks like setting stride to `0` works fine with current set of generated 2d reqs.

dyoussif:
> @dyoussif I see some assertions while testing PVC:
> 
> ```
> $ ./build/tests/benchdnn/benchdnn -v5 --engine=gpu --conv --impl=v2 --dir=FWD_I --stag=axb --dtag=axb mb1128ic32ih16oc32oh16kh1
> create: --conv --engine=gpu --dir=FWD_I --stag=axb --dtag=axb --impl=v2 mb1128ic32ih16oc32oh16kh1ph0
> Assertion parts.size() == 2 failed at ../src/gpu/intel/jit/v2/ir/tensor.hpp:334
> 
> Aborted (core dumped)
> ```
> 
> Would you please take a look? I think this is coming from mismatch in parse/stringify.

It should be fixed now.

echeresh:
> Infer stride_var reqs for mod and gt/lt reqs :
examples
ic % 16 == 0  => stride_iw % 16 ==0, stride_ih % 16 ==0, stride_mb % 16 == 0
ic*iw < 256 => stride_iw*iw<256  stride_ih < 256 

@dyoussif I'm not sure the modulus implication this always holds. In general the outer stride doesn't depend on the inner dimensions (we only need to ensure no overlapping). Please take a look at this case on PVC:

```
./build/tests/benchdnn/benchdnn -v5 --engine=gpu --conv --impl=v2 --dir=FWD_I --dt=s8 --strides=1280x1x80::1280x1x80 mb128ic64iw16oc64ow16kw3
```

src/gpu/intel/jit/conv/problem.cpp
```diff
@@ -45,6 +45,62 @@ const std::vector<pvar_t> &conv_dims() {
     return _conv_dims;
 }
 
+pvar_t prb_stride(pvar_t dim, tensor_kind_t tensor_kind) {
+    auto ret = pvars::src_mb_stride;
+    if (tensor_kind == tensor_kind_t::src) {
+
+        if (dim == pvars::mb)
+            ret = pvars::src_mb_stride;
```
echeresh, line 53: For simplicity we can handle it generically `return pvar_t("src_" + dim.str() + "_stride");`

src/gpu/intel/jit/conv/problem.cpp
```diff
@@ -45,6 +45,62 @@ const std::vector<pvar_t> &conv_dims() {
     return _conv_dims;
 }
 
+pvar_t prb_stride(pvar_t dim, tensor_kind_t tensor_kind) {
```
echeresh, line 48:
```suggestion
pvar_t prb_stride(const pvar_t &dim, tensor_kind_t tensor_kind) {
```

src/gpu/intel/jit/ir/problem.cpp
```diff
@@ -123,6 +123,39 @@ pvar_t b("b");
...
+pvar_t ih_stride("ih_stride");
+pvar_t iw_stride("iw_stride");
+pvar_t kd_stride("kd_stride");
+pvar_t kh_stride("kh_stride");
+pvar_t kw_stride("kw_stride");
+pvar_t mb_stride("mb_stride");
+pvar_t oc_stride("oc_stride");
+pvar_t od_stride("od_stride");
+pvar_t oh_stride("oh_stride");
+pvar_t ow_stride("ow_stride");
```
echeresh, line 158: Looks like these are unused

src/gpu/intel/jit/v2/conv/bridge.hpp
```diff
@@ -58,6 +58,52 @@ inline pvar_tile_t to_shape(const convolution_pd_t *pd) {
...
     shape[pvars::ph] = pd->padT();
     shape[pvars::pw] = pd->padL();
+    memory_desc_wrapper mdw_src(pd->invariant_src_md());
+    memory_desc_wrapper mdw_wei(pd->invariant_wei_md());
+    memory_desc_wrapper mdw_dst(pd->invariant_dst_md());
+
+    bool src_strided = (mdw_src.is_plain() && !mdw_src.is_dense());
+    bool wei_strided = (mdw_wei.is_plain() && !mdw_wei.is_dense());
+    bool dst_strided = (mdw_dst.is_plain() && !mdw_dst.is_dense());
+    auto set_stride = [&](const memory_desc_t *md, int index,
```
echeresh, line 68: Nit:
```suggestion
    auto get_stride = [&](const memory_desc_t *md, int index,
```

src/gpu/intel/jit/v2/conv/bridge.hpp
```diff
@@ -58,6 +58,52 @@ inline pvar_tile_t to_shape(const convolution_pd_t *pd) {
...
+            index = index > non_spatial_ndims
+                    ? pd->ndims() - (index - non_spatial_ndims)
+                    : index - 1;
+            if (md->dims[index] == 1) return 0;
+            return md->format_desc.blocking.strides[index];
+        }
+        return 0;
+    };
+    if (src_strided) {
+        shape[pvars::src_mb_stride] = set_stride(pd->invariant_src_md(), 1);
```
echeresh, line 80: Nit: can it start from `0`? Usually indexing is zero-based...

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -327,6 +331,23 @@ void kernel_desc_t::set_defaults() {
     }
 }
 
+void kernel_desc_t::set_stride_reqs(tensor_kind_t kind,
+        const layout_tag_t &desc_tag, prb_reqs_t &reqs) const {
+    if (desc_tag.is_strided()) {
```
echeresh, line 336: Nit: to reduce indentation:
```suggestion
    if (!desc_tag.is_strided()) return;
```

src/gpu/intel/jit/v2/conv/plan_registry_data.cpp
```diff
@@ -25,117 +25,117 @@ namespace conv {
...
-        "hw=xehpc prop=fwd src=ABx32a4b:s8 wei=aCBx32b4c:s8 dst=axb:s8 fma=dpas simd=16 regs=256 iter=ic4kw8mb32oc64 spec=oc@64 ext=out_b2,out_b4 model=0132A59E441460B03E0050713F337B923F3400653E02F3686046CCC6363F",
-        "hw=xehpc prop=fwd src=ABx32a2b:f16 wei=aCBx32b2c:f16 dst=axb:f16 fma=dpas simd=16 regs=256 iter=ic2kw8mb32oc64 spec=oc@64 ext=out_b1,out_b4 model=013AF19F443000793E00306E3F9A01903F1680903E02AF4066478BEFC33F",
-        "hw=xehpc prop=fwd src=axb:s8 wei=aCBx16b4c:s8 dst=axb:s8 fma=dpas simd=16 regs=256 iter=ic4kw8oc32ow16 tg=oc2ow4 spec=oc@64 ext=out_b2,out_b4 model=01E33E004505E0223F0280553F68728B3FFB3F6D3F0200489F46DFE60C41",
-        "hw=xehpc prop=fwd src=axb:f16 wei=aCBx16b2c:f16 dst=axb:f16 fma=dpas simd=16 regs=256 iter=ic2kw8oc32ow16 tg=oc2ow4 spec=oc@64 ext=out_b1,out_b4 model=0158E2DE440400303F0180783F99E3873F04C0293F02E6FE60463C8CE840",
+        "hw=xehpc prop=fwd src=axb:s8:0 wei=axcb:s8:0 dst=axb:s8:0 fma=dpas simd=16 regs=256 iter=ic64oc64ow32 2d=1 ext=out_b2,out_b4,stream_k model=015DBA59441308D13E08C8673FEB49D441F6FFA74102487A344669B2AD3F",
```
echeresh, line 28: Can we avoid these changes and treat `"tag:type"` by default as non-strided?

src/gpu/intel/jit/v2/ir/reqs.cpp
```diff
@@ -670,8 +680,61 @@ class req_impl_t {
     req_rhs_t rhs_;
 };
 
+void prb_reqs_t::add_stride_reqs(
+        const pvar_t stride, std::vector<pvar_t> vars) {
```
echeresh, line 684:
```suggestion
        const pvar_t &stride, const std::vector<pvar_t> &vars) {
```

src/gpu/intel/jit/v2/ir/reqs.cpp
```diff
@@ -670,8 +680,61 @@ class req_impl_t {
...
+    if (vars.size() == 0) return;
+    std::vector<req_impl_t> new_reqs;
+    for (auto &r : reqs_) {
```
echeresh, line 687: Can you please explain what is the purpose of this function? Looks like it handles the stride and inner dims together - but they are independent in general...

src/gpu/intel/jit/v2/ir/tensor.hpp
```diff
@@ -284,16 +284,22 @@ class layout_tag_t {
     layout_tag_t() = default;
 
     layout_tag_t(const layout_desc_t &desc, const type_t &type,
-            const layout_raw_tag_t &raw_tag)
-        : desc_(desc), type_(type), raw_tag_(raw_tag) {}
-    layout_tag_t(const type_t &type, const std::string &str_tag)
-        : layout_tag_t({}, type, layout_raw_tag_t(str_tag)) {}
+            const layout_raw_tag_t &raw_tag, const bool is_strided = false)
```
echeresh, line 287:
```suggestion
            const layout_raw_tag_t &raw_tag, bool is_strided = false)
```

src/gpu/intel/jit/v2/ir/tensor.hpp
```diff
@@ -284,16 +284,22 @@ class layout_tag_t {
...
+        , type_(type)
+        , raw_tag_(raw_tag)
+        , is_strided_(is_strided) {}
+    layout_tag_t(const type_t &type, const std::string &str_tag,
+            const bool is_strided = false)
+        : layout_tag_t({}, type, layout_raw_tag_t(str_tag), is_strided) {}
     layout_tag_t(const layout_desc_t &desc, const type_t &type,
-            const std::string &str_tag)
-        : layout_tag_t(desc, type, layout_raw_tag_t(str_tag)) {}
+            const std::string &str_tag, const bool is_strided = false)
```
echeresh, line 296:
```suggestion
            const std::string &str_tag, bool is_strided = false)
```

src/gpu/intel/jit/v2/ir/tensor.hpp
```diff
@@ -284,16 +284,22 @@ class layout_tag_t {
...
+        : layout_tag_t(desc, type, layout_raw_tag_t(str_tag), is_strided) {}
 
     bool is_empty() const { return raw_tag_.is_empty(); }
     bool is_any() const { return raw_tag_.is_any(); }
+    bool is_strided() const { return is_strided_; }
+    void set_strided(const bool strided) { is_strided_ = strided; }
```
echeresh, line 302:
```suggestion
    void set_strided(bool strided) { is_strided_ = strided; }
```

src/gpu/intel/jit/v2/ir/reqs.cpp
```diff
@@ -73,10 +73,20 @@ class linear_cmp_simplifier_t : public ir_mutator_t {
...
     if (is_a_mod_b_eq_0(_e, a, b)) {
-        a = simplify_expr(a);
+        auto e = _e;
+        auto vars = find_objects<const_var_t>(_e);
+        for (auto var : vars) {
+            auto pvar = pvar_t::from_var(var.as<const_var_t>());
+            if (pvar.is_undef() || !reqs) continue;
+            auto factor = reqs->max_factor(pvar);
+            if (factor <= 1) continue;
+            e = substitute(e, var, expr_t(reqs->max_factor(pvar)));
```
echeresh, line 86: Can you please explain the logic here? At first glance, it doesn't look right, e.g. if we have reqs with:

```
ic % 4 == 0
```

and add another one:

```
ic % 64 == 0
```

then it's translated to `4 % 64 == 0` which is `false` and which is not correct (instead they should be combined). Is my understanding correct?

echeresh: > ic % 64  == 0 this implies  ic % 4 ==0

Right. So I think this code is now broken - the second condition is simplified to `false` (and causes assertion) while it shouldn't. The correct result would be to drop `ic % 4 == 0` requirement and keep `ic % 64 == 0` only.

```
    prb_reqs_t reqs;
    reqs.add(pvars::ic.var() % 4 == 0);
    reqs.add(pvars::ic.var() % 64 == 0);

```

dyoussif: Discussed offline.  Made changes to simplify req generation in send plan instead.

src/gpu/intel/jit/v2/ir/reqs.cpp
dyoussif: Discussed offline.  Kernel desc reqs generated directly in kernel_desc_2d_reqs.cpp.

src/gpu/intel/jit/v2/conv/plan_registry_data.cpp
dyoussif: done.

src/gpu/intel/jit/v2/conv/bridge.hpp
```diff
@@ -58,6 +58,52 @@ inline pvar_tile_t to_shape(const convolution_pd_t *pd) {
...
+            index = index >= non_spatial_ndims
+                    ? pd->ndims() - (index + 1 - non_spatial_ndims)
+                    : index;
+            if (md->dims[index] == 1) return 0;
+            return md->format_desc.blocking.strides[index];
+        }
+        return 0;
+    };
+    if (src_strided) {
+        shape[pvars::src_mb_stride] = get_stride(pd->invariant_src_md(), 0);
```
echeresh, line 80: It seems these global variables are used only in this function - suggest to remove them and instead generate stride vars on-the-fly via a local helper function, e.g. `shape[src_stride(pvars::ic)] = ...`.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -332,6 +336,21 @@ void kernel_desc_t::set_missing() {
...
+    if (!desc_tag.is_strided()) return;
+    std::vector<pvar_t> dims;
+    auto tag = append_groups(kind, desc_tag, is_dw);
+    auto entries = tag.raw_tag().entries();
+    for (auto it = entries.rbegin(); it != entries.rend(); it++) {
+        auto dim = tag.desc().prb_dim(it->index());
+        if (prb_stride(dim, kind).is_undef()) continue;
+        if (it == entries.rbegin())
+            reqs.add(prb_stride(dim, kind).var() == expr_t(1));
+        dims.push_back(dim);
```
echeresh, line 350: `dims` are not used now?

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -332,6 +336,21 @@ void kernel_desc_t::set_missing() {
...
+        const layout_tag_t &desc_tag, prb_reqs_t &reqs) const {
+    if (!desc_tag.is_strided()) return;
+    std::vector<pvar_t> dims;
+    auto tag = append_groups(kind, desc_tag, is_dw);
+    auto entries = tag.raw_tag().entries();
+    for (auto it = entries.rbegin(); it != entries.rend(); it++) {
+        auto dim = tag.desc().prb_dim(it->index());
+        if (prb_stride(dim, kind).is_undef()) continue;
+        if (it == entries.rbegin())
+            reqs.add(prb_stride(dim, kind).var() == expr_t(1));
```
echeresh, line 349: Since it's only for `rbegin()` - we don't need to iterate through all entries.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -758,6 +786,23 @@ void kernel_desc_t::init_kernel_iface(kernel_iface_t &kernel_iface) const {
...
         if (d == pvars::sw)
             kernel_iface.register_arg("sw_magic", type_t::u64());
+        if (!is_conv_index(d)) continue;
+        for (auto &t_kind :
+                {tensor_kind_t::src, tensor_kind_t::wei, tensor_kind_t::dst}) {
+            auto tag = src_tag;
+            if (t_kind == tensor_kind_t::wei)
+                tag = wei_tag;
+            else if (t_kind == tensor_kind_t::dst)
+                tag = dst_tag;
```
echeresh, line 796: Can we use `auto &tag = layout_tag(t_kind)`?

src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
```diff
@@ -145,6 +145,49 @@ block_2d_params_t to_block_2d_params(const prop_kind_t &prop,
     return params;
 }
 
+void generate_2d_stride_var_reqs(layout_tag_t &tag, stride_t &P,
+        prb_reqs_t &reqs, tensor_kind_t &tensor_kind, const kernel_desc_t &desc,
```
echeresh, line 149:
```suggestion
        prb_reqs_t &reqs, const tensor_kind_t &tensor_kind, const kernel_desc_t &desc,
```

Please double check where we can add `const`

src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
```diff
@@ -170,8 +213,15 @@ void generate_2d_reqs(const kernel_desc_t &desc, tensor_kind_t tensor_kind,
     auto W = params.w_dim.var();
     auto H = params.h_dim.var();
     auto P = strides.at(params.h_dim);
```
echeresh, line 215: I think the loop above that computes `strides` should be adjusted. Now `strides` are initialize assuming the dense tensor - this is not correct for stride layouts.

Once this is fixed, it seems all pitch-related requirements can be directly done on `P` value, without extra processing.

dyoussif:
make test
disable device_cpu
disable run_scans
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
dyoussif: done.

src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
dyoussif: adjusted `strides` to use stride vars when applicable.

src/gpu/intel/jit/v2/conv/bridge.hpp
dyoussif: done.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -345,8 +359,10 @@ bool is_compatible(tensor_kind_t abc, const kernel_desc_t &kernel_desc,
             && kernel_desc.ext.has(extensions_t::out_size(prb_type.size())))
         type_ok = true;
     if (!type_ok && is_out && kernel_desc.use_stream_k) type_ok = true;
-    gpu_check(type_ok) << to_string(abc) << " tag " << prb_tag
-                       << " does not match kernel descriptor tag " << desc_tag;
+    gpu_check(type_ok
+            && prb_tag.matches(desc_tag, prb.shape(), /*check_type=*/false))
```
echeresh, line 363: This needs more careful handling. There are some kernels targeting 1st convolution that assume mismatch between tags. Mismatched tags are handled via reorder that is generated when mismatch is found. For example, look at this case on PVC/BMG:

```
--engine=gpu --mode=F --conv --impl=v2 --dt=bf16 mb128ic3ih224oc64oh112kh7sh2ph3
```

It should show worse performance after this change.

Why do we need to have a strict check here? Can you please give more details?

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -404,8 +420,11 @@ void fit_tag_to(
     bool is_out_stream_k
             = (abc == tensor_kind_t::c) && kernel_desc.use_stream_k;
     if (desc_tag.type() != prb_tag.type() && !is_out_stream_k) {
-        desc_tag = layout_tag_t(
-                desc_tag.desc(), prb_tag.type(), desc_tag.raw_tag());
+        desc_tag = layout_tag_t(desc_tag.desc(), prb_tag.type(),
+                desc_tag.raw_tag(), prb_tag.is_strided());
+    } else if (prb_tag.is_strided()) {
+        desc_tag = layout_tag_t(desc_tag.desc(), prb_tag.type(),
+                desc_tag.raw_tag(), prb_tag.is_strided());
```
echeresh, line 427: Looks like both branches do the same...

src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
```diff
@@ -145,6 +145,19 @@ block_2d_params_t to_block_2d_params(const prop_kind_t &prop,
...
 
+void generate_2d_stride_var_reqs(const layout_tag_t &tag, prb_reqs_t &reqs,
+        const tensor_kind_t &tensor_kind, const int base_align) {
+    using ir_utils::safe_div;
+    std::vector<pvar_t> pvars;
+    for (int i = tag.raw_tag().nentries() - 2; i >= 0; i--) {
+        auto &e = tag.raw_tag().entries()[i];
+        gpu_assert(!e.is_blocked);
+        auto dim = tag.desc().prb_dim(e.index());
+        reqs.add_no_simplify(
```
echeresh, line 156: The current version skips W/H dimensions: https://github.com/uxlfoundation/oneDNN/blob/72e39e2d0d3831e01de170d69211f7913b6b2fad/src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp#L140

I suggest to do the same here - for consistency.

src/gpu/intel/jit/v2/conv/plan_registry.cpp
```diff
@@ -66,7 +66,8 @@ kernel_desc_t plan_registry_t::find_best(
         auto sk_desc = to_stream_k(e.desc);
```
echeresh, line 66:
```suggestion
        auto sk_desc = to_stream_k(e.desc);
        if (sk_desc.is_empty()) continue;
```

src/gpu/intel/jit/v2/ir/send.hpp
```diff
@@ -515,6 +515,36 @@ struct send_2d_desc_t {
         }
     }
 
+    static std::vector<expr_t> split_by_add(const expr_t &e) {
```
echeresh, line 518: There is `op_split()` in `src/gpu/intel/jit/ir/linear_expr.hpp`. Will it work here?

src/gpu/intel/jit/v2/ir/tensor.hpp
```diff
@@ -285,17 +285,23 @@ class layout_tag_t {
     layout_tag_t() = default;
 
     layout_tag_t(const layout_desc_t &desc, const type_t &type,
-            const layout_raw_tag_t &raw_tag)
-        : desc_(desc), type_(type), raw_tag_(raw_tag) {}
-    layout_tag_t(const type_t &type, const std::string &str_tag)
-        : layout_tag_t({}, type, layout_raw_tag_t(str_tag)) {}
+            const layout_raw_tag_t &raw_tag, bool is_strided = false)
```
echeresh, line 288: Can we treat all non-blocked tags/layouts as strided? Do you see any issues with that?

If no performance impact - it would be nice to not split between strided/dense layouts as the dense layout is a special case of the strided layout.

If this requires significant changes in this PR, let's keep as is and refactor later.

src/gpu/intel/jit/v2/ir/tensor.hpp
```diff
@@ -320,21 +326,25 @@ class layout_tag_t {
         jit::stringify(out, raw_tag_);
         out << ":";
         jit::stringify(out, type_);
+        out << ":";
+        jit::stringify(out, is_strided_);
```
echeresh, line 330: Looks like there are still warnings when kernel entries are loaded during stringify/parse check. You can check them with `ONEDNN_VERBOSE=debuginfo=100`. Please take a look.

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
dyoussif: We don't want strategies with blocked tags to be used with strided. Refined the check so it is only used for strided tensors.

src/gpu/intel/jit/v2/conv/kernel_desc_2d_reqs.cpp
dyoussif: added check.

src/gpu/intel/jit/v2/ir/send.hpp
dyoussif: Expressions come in the form `(a+b)*c`. We need to factor out c and reapply it to each summand.

src/gpu/intel/jit/v2/ir/tensor.hpp
dyoussif: 1x1 conv does reshaping, which doesn't work with strided convs. To avoid issues it is simpler to treat strided separately.

src/gpu/intel/jit/v2/ir/tensor.hpp
dyoussif: Warning is coming from adding `is_strided `to each tag dynamically, which doesn't match the original strategy, but is expected. Not sure how to resolve this.

src/gpu/intel/jit/v2/ir/tensor.hpp
echeresh: Reshape/normalization is called from a place where we have the primitive descriptor - we can have extra checks there to understand if reshape is applicable or not. PD has strided layouts -> skip normalization.

src/gpu/intel/jit/v2/ir/tensor.hpp
echeresh: If I understand correctly, that warning is because of mismatch between the stringified version in `src/gpu/intel/jit/v2/conv/plan_registry_data.cpp` (no `:0` at the end of tags) and the current version of `stringify()` (`:0` at the end of tags).

I think the best option is to not add `:0` in tag stringification - assume `0` by default. This requires to update parse/stringify to only change strided tags (i.e. add `:1` but not `:0`).

src/gpu/intel/jit/v2/ir/send.hpp
echeresh: I see, thanks. Can you please check if this would work here?

https://github.com/uxlfoundation/oneDNN/blob/48e7e7379e0b07fd541012ceed837f6bbde21ebc/src/gpu/intel/jit/ir/linear_expr.cpp#L211-L212

It's expand followed by split - should work like you described.

src/gpu/intel/jit/v2/ir/tensor.hpp
dyoussif: i see. changed.

src/gpu/intel/jit/v2/ir/send.hpp
dyoussif: changed to use linear_normalize_expander.

src/gpu/intel/jit/v2/ir/tensor.hpp
dyoussif: It still seems that we need to generate two different kernels: one non-strided with 1x1 reshape and the other with strided. Would prefer not to consolidate everything to use strided  in this PR.

src/gpu/intel/jit/conv/problem.cpp
```diff
@@ -47,6 +47,25 @@ const std::vector<pvar_t> &conv_dims() {
...
+    auto dims = conv_layout_dims(tensor_kind, true);
+    for (auto &d : dims) {
+        if (d == dim) {
+            auto str = "";
+            if (tensor_kind == tensor_kind_t::src)
+                str = "src_";
+            else if (tensor_kind == tensor_kind_t::wei)
+                str = "wei_";
+            else if (tensor_kind == tensor_kind_t::dst)
+                str = "dst_";
```
echeresh, line 61:
```suggestion
            auto str = to_string(tensor_kind) + "_";
```

src/gpu/intel/jit/v2/ir/tensor.hpp
echeresh: Sounds good, let's keep as is and address later if needed.

It's not about the number of kernels but rather about complexity: if we have strided vs non-strided branches in the code while they can be implemented in a generic way then adding this separation may be not the best thing. Reusability should also work though: 1) dense with reshaped 1x1 and strided 1D can share the same kernel and 2) 3x3 dense and any strided cases can share the same kernel as well.

echeresh (approved):

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
```diff
@@ -425,7 +443,11 @@ void fit_to_impl(kernel_desc_t &desc, const problem_t &prb) {
 }
 
 bool kernel_desc_t::can_fit(const problem_t &prb) const {
-    return is_compatible(*this, prb, /*exact=*/false);
+    auto desc = *this;
+    desc.src_tag.set_strided(prb.src_tag().is_strided());
```
kealan-barbieri, line 447: nit: is `desc` doing anything here or can you just do `src_tag.set_strided(prb.src_tag() ....` ?

kealan-barbieri (approved):

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
dyoussif: `can_fit` is const, so we are working on a copy of the kernel_desc in this function.

dyoussif:
make test
disable device_cpu
disable run_scans
disable benchdnn_all
enable benchdnn_nightly
enable benchdnn_conv

src/gpu/intel/jit/v2/conv/kernel_desc.cpp
kealan-barbieri: got it, thanks.

PR #2801, opened 2025-03-03
xe: microkernels: disambiguate vISA variables from separate invocations

petercad:
Fixes a variable aliasing bug in the microkernel service library when multiple microkernels are used.

The vISA shim that transfers control to the microkernel has several series of variables (`COPY*`, `CLOBBER*`) that are tied to physical GRFs. They are necessarily global scope vISA variables, which meant that variables from different microkernel invocations could alias and as a result vISA could copy input/output arguments to the wrong registers or the calling kernel's variables could be trashed. It's quite surprising this wasn't found earlier. Likely it's the cause of some failing SDPA configs -- FYI @umar456.

This PR annotates the variable names with their physical locations so that any aliasing that does occur will not be a problem.

### DIFF
--- a/src/gpu/intel/microkernels/shim.cpp
+++ b/src/gpu/intel/microkernels/shim.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable arch_gpu_xe-lp
disable arch_gpu_xe-lpg
disable benchdnn_all
enable benchdnn_graph

umar456 (approved):

syurkevi (approved):

PR #2802, opened 2025-03-03
common, xe: consolidate serialization API

echeresh:
PR consolidates serialization API in `src/gpu/intel` and `src/common`.

List of changes:

- Moved `serialized_t`/`serialized_data_t`/`deserializer_t` functionality from `src/gpu/intel` to `src/common`
    - Merged `serialized_data_t`, `serialized_t` and `serialization_stream_t` functionality into `serialization_stream_t`
- Renamed `src/common/{serialization_stream.hpp => serialization.hpp}` as it now includes broad serialization API (not just stream)
- Renamed `src/common/{serialization.hpp => primitive_serialization.hpp}` 

I'm not 100% sure on the names so please share your suggestions if any.


src/common/primitive_serialization.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "common/serialization.hpp"
+#include "common/type_helpers.hpp"
+#include "oneapi/dnnl/dnnl.h"
+
+namespace dnnl {
+namespace impl {
+namespace primitive_serialization {
+
+void serialize_post_ops(
+        serialization_stream_t &sstream, const post_ops_t &post_ops);
```
rjoursler, line 31: This code seems excessively verbose and the style is unaligned with how the GPU serialization tends to work. What if we just use a templated function instead?

```
namespace  primitive {

template <typename T> void serialize(serialization_stream_t sstream, const T &t);

}
```
And then use `extern template` in this file. After that, usage is `primitive::serialize()` which seems more readable. Debatably, we could also drop the namespace guard, as it doesn't seem like it serves much of a purpose as implementing multiple serializers for these core types seems like it would be an antipattern.

echeresh: I agree, switching to overloading + dropping the namespace look reasonable to me.

+ @densamoilov who implemented this originally.

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc

src/common/primitive_serialization.hpp
```diff
@@ -0,0 +1,76 @@
...
+*******************************************************************************/
+
+#ifndef COMMON_PRIMITIVE_SERIALIZATION_HPP
+#define COMMON_PRIMITIVE_SERIALIZATION_HPP
+
+#include "common/c_types_map.hpp"
+#include "common/primitive_attr.hpp"
+#include "common/serialization.hpp"
+#include "common/type_helpers.hpp"
+#include "oneapi/dnnl/dnnl.h"
```
dzarukin, line 24: Minor: probably can skip this one in favor of c_types_map.

src/common/serialization.hpp
```diff
@@ -17,59 +17,269 @@
...
+            typename T::value_type t = {};
+            pop(t);
+            v.emplace_back(t);
+        }
+    }
+
+    bool empty() const { return idx >= s.get_data().size(); }
+
+    size_t idx;
+    const serialization_stream_t &s;
```
dzarukin, line 266: Make them `private`?

src/common/serialization.hpp
```diff
@@ -17,59 +17,269 @@
...
+                std::is_same<decltype(&U::serialize),
+                        void (U::*)(serialization_stream_t &) const>::value,
+                bool>);
+        template <typename U>
+        static no_t test(...);
+
+        static const bool value = (sizeof(test<T>(0)) == sizeof(yes_t));
+    };
+
+    // Append helper function for structures with the member function
```
dzarukin, line 92: QQ: is `append` supposed to be public? Or should a ctor be the only user-facing?

dzarukin (approved):

src/common/primitive_serialization.hpp
echeresh: Removed, thanks.

src/common/serialization.hpp
echeresh: Yes, it's supposed to be public as it's used as `sstream.append(...)` in many places.
I put some suggestions to unify API in the description (e.g. switch to uniform `serialize(sstream, obj)` usage) but I'd left it for the future.

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc

src/common/primitive_serialization.hpp
echeresh: Dropped the namespace and switched to overloads (we don't need template functions yet so overloads work fine).

rjoursler (approved):

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc

dzarukin (approved):

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc

PR #2773, opened 2025-02-27
xe: enable building with out-of-tree nGEN

rjoursler:
This is intended to enable better co-development between nGEN and oneDNN.

To build with an out-tree nGEN on this branch, developers just need to set `-DONEDNN_NGEN_DIR=<ngen_dir>` in cmake and remove ngen_config.hpp from the target so that oneDNN's configuration is used .

PR #2809, opened 2025-03-04
fix a build issue when DNNL_DEV_MODE is OFF.

sungeunk:
# Description
fix a build issue when DNNL_DEV_MODE is OFF.

### DIFF
--- a/third_party/ngen/ngen_asm.hpp
+++ b/third_party/ngen/ngen_asm.hpp
@@ -1,6 +1,6 @@
 #ifndef NGEN_ASM_HPP
 #define NGEN_ASM_HPP
-#if NGEN_ASM
+#ifdef NGEN_ASM
 
 #include "ngen_config.hpp"
 


rjoursler (approved):

third_party/ngen/ngen_asm.hpp
```diff
@@ -1,6 +1,6 @@
 #ifndef NGEN_ASM_HPP
 #define NGEN_ASM_HPP
-#if NGEN_ASM
+#ifdef NGEN_ASM
```
rjoursler, line 4: Coincidentally, this change also appeared #2773, so there is not need to port this to main.

petercad (approved):

third_party/ngen/ngen_asm.hpp
sungeunk: @rjoursler  Could you cherry-pick #2773 to rls-v3.8-pc branch? I'm integrating rls-v3.8-pc branch to public OpenVINO.

rjoursler: I don't think backporting #2773 makes sense in this case. That change is relatively invasive, so keeping the simple fix provided here is much better. I can go ahead and take over getting this merged in #2824 so that we can get this merged soon.

rjoursler:
Closing as #2824 merged these changes.

PR #2821, opened 2025-03-05
benchdnn: inputs: graph: fix test cases related to int8/f8 add

TaoLv:
Mainly about int8/f8 Conv/matmul + Add fusions. Remove the dequantize op on the src1 of Add.

The original cases were executed with graph dump. Two graphs for two partitions will be dumped (1 for the main fusion, and 1 for the dequantize op). The graph of the main fusion pattern is used here to replace the original case. 

Fixes MFDNN-12928

### DIFF
--- a/tests/benchdnn/inputs/graph/pattern/f8/f8_conv_add_add_fusion.json
+++ b/tests/benchdnn/inputs/graph/pattern/f8/f8_conv_add_add_fusion.json
@@ -1,12 +1,13 @@
 {
-  "version": "3.6.0",
+  "version": "3.8.0",
   "engine_kind": "cpu",
   "fpmath_mode": "strict",
+  "fpmath_mode_apply_to_int": "false",
   "input_ports": [
     0, 
     1, 
     2, 
-    3
+    10305
   ],
   "output_ports": [
     14512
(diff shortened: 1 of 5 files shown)


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/inputs/graph/pattern/f8/f8_conv_add_add_fusion.json
```diff
@@ -134,128 +135,6 @@
         }
       ]
     }, 
-    {
```
ElaineBao, line 137: remove `--expected-n-partitions=0` at below position? https://github.com/uxlfoundation/oneDNN/blob/1ae1905345983d74c302ac12f8d9f9ada7022ab5/tests/benchdnn/inputs/graph/pattern/harness_f8_all#L2

TaoLv: It's because fp8 fusions are not supported on old platforms before SPR and several partitions will be returned. We have to skip the partition number check for this case as it will run on all platforms.

ElaineBao (approved):

mgouicem (approved):

tests/benchdnn/inputs/graph/pattern/f8/f8_conv_add_add_fusion.json
dzarukin: fp8 data-typed-cases should be reported as skipped for systems below AVX512_CORE_FP16 due to lack of emulation support.

TaoLv: Currently the logic in graph driver is that, even there are multiple partitions returned, if each of them is implemented, we still test the partitions one by one. In this case, as fp8 conv is not supported before SPR, several partitions (fp8 quant/dequant, f32 conv, etc.) are returned. fp8 quant/dequant are actually supported through reorders on the platforms, so we don't skip the case.

PR #2824, opened 2025-03-05
ngen: fix a build issue when DNNL_DEV_MODE is OFF

rjoursler:
Taking over #2809 so that we can get it merged ASAP due to the release schedule.

The only effective change here is to fix the commit message formatting.


rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

skazakov1 (approved):

PR #2825, opened 2025-03-05
Bunch of functional fixes (fixes MFDNN-13193, fixes MFDNN-13282, fixes MFDNN-13236)

dzarukin:
* Remove mask retrieval for the lowest precompute_scales calls as it's already determined based on the number of scales needed for weights.
* Restored the memory usage for scratchpad size after the refactor for a single implementation.
* Adjusted softmax threshold a little more to fit Windows threadpool output values.
* Added support for mode-modifier=M for sparse functionality. Metadata can't be generated with random values - it leads to an access a random memory location which is prohibited. Fill metadata as for correctness mode, rest buffer are kept with random values.

### DIFF
--- a/src/cpu/x64/jit_avx512_core_x8s8s32x_conv_kernel.cpp
+++ b/src/cpu/x64/jit_avx512_core_x8s8s32x_conv_kernel.cpp
@@ -1771,9 +1771,11 @@ status_t jit_avx512_core_x8s8s32x_fwd_kernel::init_conf(jit_conv_conf_t &jcp,
 void jit_avx512_core_x8s8s32x_fwd_kernel::init_scratchpad(
         memory_tracking::registrar_t &scratchpad, const jit_conv_conf_t &jcp,
         const primitive_attr_t &attr) {
-    const int wei_mask = attr.scales_.get_mask(DNNL_ARG_WEIGHTS);
-    const dim_t scales_count = wei_mask == 0 ? 1 : jcp.oc * jcp.ngroups;
-    dim_t count = wei_mask == 0 ? (dim_t)16 : scales_count;
+    dim_t count = 16;
+    if (!attr.scales_.has_default_values(DNNL_ARG_WEIGHTS)) {
+        const int wei_mask = attr.scales_.get_mask(DNNL_ARG_WEIGHTS);
+        if (wei_mask > 0) count = jcp.oc * jcp.ngroups;
+    }
     scratchpad.book<float>(key_conv_adjusted_scales, count);
 }
 
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -205,6 +205,14 @@ size_t dnn_mem_t::size() const {
     return dnnl_memory_desc_get_size(md_);
 }
 
+bool dnn_mem_t::is_sparse_md() const {
+#ifdef DNNL_EXPERIMENTAL_SPARSE
+    return query_md_sparse_encoding(md_) != dnnl_sparse_encoding_undef;
+#else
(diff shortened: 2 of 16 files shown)


src/cpu/x64/jit_avx512_core_x8s8s32x_conv_kernel.cpp
```diff
@@ -1765,9 +1765,12 @@ status_t jit_avx512_core_x8s8s32x_fwd_kernel::init_conf(jit_conv_conf_t &jcp,
...
         memory_tracking::registrar_t &scratchpad, const jit_conv_conf_t &jcp,
         const primitive_attr_t &attr) {
-    const int wei_mask = attr.scales_.get_mask(DNNL_ARG_WEIGHTS);
-    const dim_t scales_count = wei_mask == 0 ? 1 : jcp.oc * jcp.ngroups;
-    dim_t count = wei_mask == 0 ? (dim_t)16 : scales_count;
+    dim_t count = 16;
+    if (!attr.scales_.has_default_values(DNNL_ARG_WEIGHTS)) {
+        const int wei_mask = attr.scales_.get_mask(DNNL_ARG_WEIGHTS);
+        const dim_t scales_count = wei_mask == 0 ? 1 : jcp.oc * jcp.ngroups;
+        if (wei_mask > 0) count = scales_count;
```
mgouicem, line 1772: (nit)
```suggestion
        if (wei_mask > 0) count = wei_mask == 0 ? 1 : jcp.oc * jcp.ngroups;
```

mgouicem (approved):

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -900,7 +913,16 @@ int dnn_mem_t::initialize(
     SAFE(initialize_memory_create(handle_info), CRIT);
 
     if (handle_info.is_allocate()) {
-        if (!has_bench_mode_modifier(mode_modifier_t::no_ref_memory)) map();
+        // Indirectly accessed memory objects can't have random data in them
+        // because it leads to undefined jump to a random memory location that
+        // the test object doesn't own.
+        // For such objects every driver will guarantee they get filled. To get
```
densamoilov, line 919:
```suggestion
        // For such objects every driver will guarantee they get filled. To get them
```

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -900,7 +913,16 @@ int dnn_mem_t::initialize(
     SAFE(initialize_memory_create(handle_info), CRIT);
 
     if (handle_info.is_allocate()) {
-        if (!has_bench_mode_modifier(mode_modifier_t::no_ref_memory)) map();
+        // Indirectly accessed memory objects can't have random data in them
```
densamoilov, line 916: I know what the problem is but I fail to comprehend the comment :) 
Does it say that the metadata buffers cannot be filled with random values because it will result in a crash during decoding the tensor?

densamoilov (approved):

src/cpu/x64/jit_avx512_core_x8s8s32x_conv_kernel.cpp
dzarukin: Updated, thanks.

tests/benchdnn/dnnl_memory.cpp
dzarukin: Thanks for pointing to this. I've updated the comment, please check it out.

dzarukin:
make test

dzarukin:
make test

avmanerikar (approved):

vpirogov (approved):

PR #2829, opened 2025-03-06
graph: backend: graph_compiler: src: fix brgemm coredump issue

rongzha1:
maintain GC code.

change brgemm to new process , final before use.

Validate by IPEX GC backend. cc @chunyuan-w


TaoLv (approved):

TaoLv:
@rongzha1  how is this tested?

rongzha1:
> @rongzha1 how is this tested?

since GC is disabled in CI, for api and unit tests, I tested it in local SPR by CTest.  It passed all tests except known issue.(MHA numeric issue and wrong partition numbers which is set based on dnnl backend)
For model level tests, I asked IPEX to test in v3.7 adoption test. Coredump was fixed and no regression was reported.

chunyuan-w (approved):

PR #2831, opened 2025-03-06
[rls-v3.8-pc] xe: enable building with out-of-tree nGEN

rjoursler:
Backport of #2773. OpenVino hit build issues due to the specification on nGEN related configuration defines in their CMAKE file. Because this PR consolidateds nGEN defines into `src/gpu/intel/jit/ngen/ngen_config.hpp` and resolved all the resulting warnings that occurred, it also fixes the issues with setting `#define` in their CMAKE.


rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

atkassen (approved):

kealan-barbieri (approved):

spalicki (approved):

PR #2834, opened 2025-03-06
common: disabling #pragma omp simd on windows debug builds

spalicki:
`#pragma omp simd` on Windows debug builds in some corner cases creates incorrect code that leads to wrong results. We are disabling it using `/Qopenmp-simd-` compilation flag, but it seems to not be enough.

### DIFF
--- a/src/common/dnnl_thread.hpp
+++ b/src/common/dnnl_thread.hpp
@@ -184,13 +184,18 @@ inline int dnnl_get_current_num_threads() {
 #define OMP_GET_NUM_THREADS() 1
 #endif
 
-// MSVC still supports omp 2.0 only
-#if defined(_MSC_VER) && !defined(__clang__) && !defined(__INTEL_COMPILER)
+// Disabling OMP SIMD feature in the following scenarios:
+// * For MSVC as it only supports OpenMP 2.0
+// * In debug mode on Windows to avoid incorrect code generation
+//   by Intel(R) oneAPI DPC++/C++ Compiler
+#if defined(_MSC_VER) \
+        && ((!defined(__clang__) && !defined(__INTEL_COMPILER)) \
+                || defined(_DEBUG))
 #define collapse(x)
 #define PRAGMA_OMP_SIMD(...)
 #else
 #define PRAGMA_OMP_SIMD(...) PRAGMA_MACRO(CHAIN2(omp, simd __VA_ARGS__))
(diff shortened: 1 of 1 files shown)


src/common/dnnl_thread.hpp
```diff
@@ -184,8 +184,12 @@ inline int dnnl_get_current_num_threads() {
...
 #endif
 
-// MSVC still supports omp 2.0 only
-#if defined(_MSC_VER) && !defined(__clang__) && !defined(__INTEL_COMPILER)
+// Disabling OMP SIMD feature on MSVC (still supports omp 2.0 only) and other
+// Windows debug builds completely due to incorrect results (_DEBUG is MSVC
+// specific).
+#if defined(_DEBUG) \
+        || (defined(_MSC_VER) && !defined(__clang__) \
+                && !defined(__INTEL_COMPILER))
```
dzarukin, line 192: I'd go a safer way:
```suggestion
#if defined(_MSC_VER) && ((!defined(__clang__) && !defined(__INTEL_COMPILER)) || defined(_DEBUG))
```

And pairing undef must be matched as well.

src/common/dnnl_thread.hpp
```diff
@@ -184,8 +184,12 @@ inline int dnnl_get_current_num_threads() {
 #define OMP_GET_NUM_THREADS() 1
 #endif
 
-// MSVC still supports omp 2.0 only
-#if defined(_MSC_VER) && !defined(__clang__) && !defined(__INTEL_COMPILER)
+// Disabling OMP SIMD feature on MSVC (still supports omp 2.0 only) and other
+// Windows debug builds completely due to incorrect results (_DEBUG is MSVC
+// specific).
```
vpirogov, line 189:
```suggestion
// Disabling OMP SIMD feature in the following scenarios:
// * For MSVC as it only supports OpenMP 2.0
// * In debug mode on Windows to avoid incorrect code generation by Intel C++/DPC++ Compiler
```

vpirogov (approved):

src/common/dnnl_thread.hpp
spalicki: Done.

src/common/dnnl_thread.hpp
spalicki: Done.

mgouicem (approved):

Issue #2303, opened 2024-12-20
test_benchdnn_modeC_conv_ci_cpu fails on AArch64 CI for c7g instance


renato-arantes:
# Summary

The test `test_benchdnn_modeC_conv_ci_cpu` with `fpmath` mode enabled for `bf16` fails on AArch64 GitHub CI at an AWS `c7g`.

# Version

ACL_VERSION: v24.11.1

# Environment

OneDNN GitHub CI for AArch64 on a `c7g` AWS instance.

# Steps to reproduce

`benchdnn --conv --dir=FWD_D --attr-fpmath=bf16`

# Observed behaviour

Test fail:

```
2024-12-19T17:06:18.6979025Z run: --conv --dir=FWD_D --attr-fpmath=bf16 ic17ih8oc17oh4kh1sh2ph0n"conv_basic_2d:1x1_stride_tail"
2024-12-19T17:06:18.6979225Z [   4][DST][0:0:1:0] exp_f32:         -25 exp:         -25 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6979403Z [  11][DST][0:0:2:3] exp_f32:         -19 exp:         -19 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6979583Z [  14][DST][0:0:3:2] exp_f32:         -29 exp:         -29 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6979751Z [  15][DST][0:0:3:3] exp_f32:         -16 exp:         -16 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6979922Z [  20][DST][0:1:1:0] exp_f32:          24 exp:          24 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6980092Z [  27][DST][0:1:2:3] exp_f32:         -35 exp:         -35 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6980363Z [  30][DST][0:1:3:2] exp_f32:         -30 exp:         -30 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6980534Z [  31][DST][0:1:3:3] exp_f32:         -26 exp:         -26 got:         nan diff:     nan rdiff:     nan
2024-12-19T17:06:18.6980713Z [  36][DST][0:2:1:0] exp_f32:         -17 exp:         -17 got:         nan diff:     nan rdiff:     nan
[...]

PR #2838, opened 2025-03-07
cpu: aarch64: dispatch fpmath_mode::bf16 conv to ACL

michalowski-arm:
jit_sve_conv is above acl_conv in the convolution list since it performs better for fp32, however for fpmath_mode::bf16 jit_conv does not support lower precision calculation so we skip it for acl_conv if oneDNN was compiled with ACL

When tested with ResNet50 and `ONEDNN_DEFAULT_FPMATH_MODE=BF16` we see:

- ~1.34x speedup for 8 threads
- ~1.41x speedup for 16 threads
- ~1.17x speedup for 32 threads

These are in line with results when ACL conv was above jit conv for `FPMATH_MODE::BF16`

For a minimal benchdnn reproducer use:
```
OMP_NUM_THREADS=16 ONEDNN_DEFAULT_FPMATH_MODE=BF16 ./build/tests/benchdnn/benchdnn -v5 --mode=p --conv mb1_ic256oc256_ih28oh14kh3sh2dh0ph1_iw28ow14kw3sw2dw0pw1 mb1_ic1024oc2048_ih14oh7kh1sh2dh0ph0_iw14ow7kw1sw2dw0pw0
```

### DIFF
--- a/src/cpu/aarch64/jit_sve_1x1_convolution.hpp
+++ b/src/cpu/aarch64/jit_sve_1x1_convolution.hpp
@@ -1,6 +1,7 @@
 /*******************************************************************************
 * Copyright 2021-2023 Intel Corporation
 * Copyright 2021-2024 FUJITSU LIMITED
+* Copyright 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -54,6 +55,13 @@ struct jit_sve_1x1_convolution_fwd_t : public primitive_t {
 
         status_t init(engine_t *engine) {
             using namespace utils;
+#if (DNNL_AARCH64_USE_ACL)
+            if (get_fpmath_mode() == fpmath_mode::bf16) {
+                // prefer ACL to jit for fpmath_mode::bf16 if available
+                // since it supports lower precision calculation
+                return status::unimplemented;
+            }
+#endif
 
             bool ok = true && is_fwd()
                     && set_default_alg_kind(alg_kind::convolution_direct)
--- a/src/cpu/aarch64/jit_sve_convolution.hpp
+++ b/src/cpu/aarch64/jit_sve_convolution.hpp
@@ -1,6 +1,7 @@
 /*******************************************************************************
 * Copyright 2020-2021 Intel Corporation
 * Copyright 2020-2024 FUJITSU LIMITED
+* Copyright 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


Sqvid:
Thanks for the patch! Could you put up the performance data for this change please?

cc: @Ryo-not-rio

src/cpu/aarch64/jit_sve_1x1_convolution.hpp
```diff
@@ -54,6 +55,13 @@ struct jit_sve_1x1_convolution_fwd_t : public primitive_t {
 
         status_t init(engine_t *engine) {
             using namespace utils;
+#if (DNNL_AARCH64_USE_ACL)
```
fadara01, line 58: would it not be cleaner if we had ACL higher than jit in conv list cpu, and return `status::unimplemented` from ACL conv if the conv is fp32?
This would allow us to avoid the `#if (DNNL_AARCH64_USE_ACL)` here

fadara01:
This addresses fast-math regressions caused by: https://github.com/uxlfoundation/oneDNN/pull/2403

src/cpu/aarch64/jit_sve_1x1_convolution.hpp
michalowski-arm: Even with jit higher in the list we see some calls go to indirect_gemm:acl, for example these are results for Resnet50 running in fp32 mode:
```
impl               ncalls  time(ms)  overall%  agg_ncalls(avg)  agg_time(ms)  agg_overall%
jit_1x1:sve_256    36396   6015.02   47.46     36396.00         6015.02       47.46
indirect_gemm:acl  17187   5484.52   43.28     26791.50         11499.55      90.74
jit:blk            53583   1173.38   9.26      35722.00         12672.93      100.00
```

If we return `status::unimplemented` for acl, these calls would go to either jit_brgconv (if they are supported), which is generally slightly worse than acl, or worse it would go to ref.

fadara01: Ahh, I see, so there might be fp32 convs that jit does not support, but ACL does.
Then, yeah I think this change makes sense.

fadara01 (approved):
LGTM

michalowski-arm:
> Thanks for the patch! Could you put up the performance data for this change please?

Updated the PR message with some performance results

Sqvid:
Holding off on this for now as it re-exposes this issue https://github.com/uxlfoundation/oneDNN/issues/2303.

Ryo-not-rio (approved):

theComputeKid:
The bug exposed already exists in the previous release so we can go ahead with it.

PR #2839, opened 2025-03-07
Test PR for differential clang-tidy check

vpirogov:


src/common/batch_normalization_pd.hpp
```diff
@@ -148,6 +148,7 @@ struct batch_normalization_pd_t : public primitive_desc_t {
     }
 };
 
+// NOLINTBEGIN(google-default-arguments)
```
dzarukin, line 151: Could you share, please, what is this one about?

dzarukin (approved):

src/common/batch_normalization_pd.hpp
vpirogov: This one was not intended for review, just to test clang-tidy check developer experience...

clang-tidy complains about use of default arguments on override methods in primitive descriptors like this one:
```
    const memory_desc_t *arg_md(
            int arg, bool user_input = false) const override {
```

The rationale is that different default values on base class and derivative classes result in counter-intuitive behavior in some cases. Best example I found is here: https://stackoverflow.com/questions/3533589/can-virtual-functions-have-default-parameters

dzarukin: I see. Well, will need to change then. Thanks for explaining.

vpirogov: I chose to suppress it because we are heavily relying on these default in the implementations. The only scenario when issue will arise if someone choses to change these default in derivative classes, which may be covered by a comment instead.

PR #2842, opened 2025-03-07
clang-tidy fixes in common headers

vpirogov:
This PR addresses clang-tidy warnings detected with `ONEDNN_USE_CLANG_TIDY=CHECK_ALL` in API, examples, and common.

### DIFF
--- a/include/oneapi/dnnl/dnnl_ukernel.hpp
+++ b/include/oneapi/dnnl/dnnl_ukernel.hpp
@@ -19,6 +19,7 @@
 
 #ifndef ONEAPI_DNNL_DNNL_UKERNEL_HPP
 #define ONEAPI_DNNL_DNNL_UKERNEL_HPP
+// NOLINTBEGIN(readability-identifier-naming)
 
 #include "oneapi/dnnl/dnnl.hpp"
 #include "oneapi/dnnl/dnnl_ukernel.h"
@@ -467,4 +468,5 @@ struct transform : public handle<dnnl_transform_t> {
 
 /// @} dnnl_api
 
+// NOLINTEND(readability-identifier-naming)
 #endif /* ONEAPI_DNNL_DNNL_UKERNEL_HPP */
--- a/src/common/gemm_types.hpp
+++ b/src/common/gemm_types.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2024 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -57,10 +57,10 @@ struct gemm_desc_t : public op_desc_t {
         return utils::make_unique<gemm_desc_t>(*this);
     }
 
-    memory_desc_t a_desc {};
-    memory_desc_t b_desc {};
-    memory_desc_t c_desc {};
-    memory_desc_t bias_desc {};
+    memory_desc_t a_desc;
+    memory_desc_t b_desc;
+    memory_desc_t c_desc;
+    memory_desc_t bias_desc;
     // Type for accumulating A*B.
     dnnl_data_type_t acc_type {};
     // Sum across k dimension in either A or B tensor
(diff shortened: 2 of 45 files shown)


densamoilov (approved):

vpirogov:
Apparently `using` [is not an equivalent](https://stackoverflow.com/questions/48613758/using-vs-typedef-is-there-a-subtle-lesser-known-difference) of `typedef` in GCC...

vpirogov:
make test

TaoLv (approved):

include/oneapi/dnnl/dnnl_ukernel.hpp
```diff
@@ -19,6 +19,7 @@
 
 #ifndef ONEAPI_DNNL_DNNL_UKERNEL_HPP
 #define ONEAPI_DNNL_DNNL_UKERNEL_HPP
+// NOLINTBEGIN(readability-identifier-naming)
```
mgouicem, line 22: Is it complaining about the uppercase A, B, C, D or is there something else?

src/common/gemm_types.hpp
```diff
@@ -57,16 +57,16 @@ struct gemm_desc_t : public op_desc_t {
...
+    memory_desc_t bias_desc;
     // Type for accumulating A*B.
-    dnnl_data_type_t acc_type {};
+    dnnl_data_type_t acc_type;
     // Sum across k dimension in either A or B tensor
     // and output to sum_ab tensor.
-    sum_ab_t sum_ab {};
-    dnnl_data_type_t sum_ab_type {};
+    sum_ab_t sum_ab;
+    dnnl_data_type_t sum_ab_type;
```
mgouicem, line 69: I believe coverity requires explicit zero-initialization. I guess we will have to disable the check of either clang-tidy or coverity to not end up in an infinite editing loop.

+ @dzarukin

src/common/utils.hpp
```diff
@@ -213,6 +214,7 @@ template <typename... Args>
 constexpr bool any_null(Args... ptrs) {
     return one_of(nullptr, ptrs...);
 }
+// NOLINTEND(performance-unnecessary-value-param)
```
mgouicem, line 217: The matching NOLINTBEGIN seems to be missing.

src/common/gemm_types.hpp
vpirogov: We had these warnings only in one file, so my hope that coverity does not make a fuss about this one.

atkassen: If clang-tidy is complaining about these initializers, it is wrong. Removing the initializer leaves these members uninitialized, and @mgouicem is right that coverity will complain.

`memory_desc_t` OTOH doesn't need an explicit `{}` initializer because it will call the default constructor anyway.

When I run with clang-tidy 18, I don't see any complaints about `acc_type`, `sum_ab`, or `sum_ab_type`.

vpirogov: Thanks, @atkassen.

src/common/ittnotify.hpp
```diff
@@ -24,7 +24,9 @@ namespace dnnl {
 namespace impl {
 namespace itt {
 
-typedef enum {
+// GCC treats using and typedef diffently for enums and structs
```
atkassen, line 27:
```suggestion
// GCC treats using and typedef differently for enums and structs
```

src/common/memory_tracking.hpp
```diff
@@ -428,8 +428,8 @@ struct registry_t {
     public:
         common_iterator_t(const void *base_ptr_,
                 const std::unordered_map<key_t, entry_t> &map,
-                bool is_begin = true) {
-            base_ptr = base_ptr_;
+                bool is_begin = true)
+            : base_ptr(base_ptr_) {
             if (is_begin) {
                 iter = map.cbegin();
             } else {
```
atkassen, line 435:
```suggestion
            : base_ptr(base_ptr_)
            , iter(is_begin ? map.cbegin() : map.cend()) {}
```

atkassen (approved):

vpirogov:
make test

PR #2845, opened 2025-03-10
example: Add an example of Pointnet inference implementation

s-Nick:
# Description

*Converted to DRAFT to address comments, it will be closed and moved to oneAPI Samples as suggested once comments are solved* 

This PR adds a useful example of how to implement a full model using oneDNN. It implements inference of Pointnet model using the ModelNet10 dataset. With the example there is also a python script that, using a pre-trained model, converts data to the point cloud to use as input for the inference example.

The introduction of this example it is necessary to help us moving existing [portDNN](https://github.com/codeplaysoftware/portDNN) users to oneDNN, showing that everything they are used to achieve with portDNN it's possible with oneDNN. It would also allow Codeplay Software to properly archive portDNN.

# Checklist

## General

### DIFF
--- a/examples/network/README.md
+++ b/examples/network/README.md
@@ -1,0 +1,35 @@
+# PointNet Convolutional Neural Network Sample for 3D Pointcloud Classification
+
+[PointNet][pointnet-paper] is a convolutional neural network architecture for applications concerning 3D recognition such as object classification and part segmentation. These sample codes implement a variant of PointNet for 3D object classification, for inference only with ModelNet10, showing a larger example of using oneDNN. Some rough instructions for how it might be used are provided.
+
+## Obtaining the model weights and classes and preparing an input pointcloud
+
+A preprocessing script is provided which unpacks the weights from a pretrained pytorch model. The script also prepares an input pointcloud for testing inference. The pointcloud is made from 3D scans taken from the [ModelNet10][modelnet] dataset. The script requires an installation of [PyTorch][pytorch]. First download the pretrained PointNet weights and move the pth file into the same directory of the model.
+
+```bash
+wget http://3dvision.princeton.edu/projects/2014/3DShapeNets/ModelNet10.zip
+unzip ModelNet10.zip
+rm ModelNet10.zip
+python3 prepareData.py ModelNet10/ pointnet_model.pth
+```
+
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`
+
+## Testing on a pointcloud
+
+The oneDNN samples are built in the default CMake configuration. The sample
+is built by the target `network-pointnet-cpp`. The samples must first
+be passed the directory where the binary weights files are stored and the second
+argument should be the preprocessed pointcloud that should be classified. The expected
+output is of a classification index and a series of times in nanoseconds that corresond
+to the total time to run the network on an input, not including data transfer time.
+
+```bash
+network-pointnet-cpp ModelNet10/directory/extracted_data ModelNet10/directory/input_cloud/itemName_cloud.bin
+
+```
+
+[pointnet-paper]: https://arxiv.org/pdf/1612.00593.pdf
+[pytorch]: https://pytorch.org/
+[modelnet]: https://modelnet.cs.princeton.edu/
+
--- a/examples/network/README.md
+++ b/examples/network/README.md
@@ -1,10 +1,12 @@
 # PointNet Convolutional Neural Network Sample for 3D Pointcloud Classification
 
-[PointNet][pointnet-paper] is a convolutional neural network architecture for applications concerning 3D recognition such as object classification and part segmentation. These sample codes implement a variant of PointNet for 3D object classification, for inference only with ModelNet10, showing a larger example of using oneDNN. Some rough instructions for how it might be used are provided.
+[PointNet][pointnet-paper] is a convolutional neural network architecture for applications concerning 3D recognition such as object classification and part segmentation. These sample codes implement a variant of PointNet for 3D object classification, for inference only with ModelNet10, providing a comprehensive example of using oneDNN. You can see the following initial instructions on using the samples.
 
-## Obtaining the model weights and classes and preparing an input pointcloud
+## Obtain the Model Weights and Classes and Preparing an Input pointcloud
 
-A preprocessing script is provided which unpacks the weights from a pretrained pytorch model. The script also prepares an input pointcloud for testing inference. The pointcloud is made from 3D scans taken from the [ModelNet10][modelnet] dataset. The script requires an installation of [PyTorch][pytorch]. First download the pretrained PointNet weights and move the pth file into the same directory of the model.
+A preprocessing script is provided which unpacks the weights from a pre-trained pytorch model. The script also prepares an input pointcloud for testing inference. The pointcloud is made from 3D scans taken from the [ModelNet10][modelnet] dataset. The script requires an installation of [PyTorch][pytorch].
+
+First, download the pre-trained PointNet weights and then move the pth file into the same directory containing the model.
 
 ```bash
 wget http://3dvision.princeton.edu/projects/2014/3DShapeNets/ModelNet10.zip
@@ -13,16 +15,17 @@ rm ModelNet10.zip
 python3 prepareData.py ModelNet10/ pointnet_model.pth
 ```
 
-The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`.
 
-## Testing on a pointcloud
+## Test on a pointcloud
 
 The oneDNN samples are built in the default CMake configuration. The sample
-is built by the target `network-pointnet-cpp`. The samples must first
-be passed the directory where the binary weights files are stored and the second
-argument should be the preprocessed pointcloud that should be classified. The expected
-output is of a classification index and a series of times in nanoseconds that corresond
-to the total time to run the network on an input, not including data transfer time.
+is built by the target `network-pointnet-cpp`.
+The oneDNN samples are built using the default CMake configuration.
+To run the sample, provide as first argument the path to the directory containing the binary weight files and as second argument the path to the preprocessed point cloud file to be classified.
+
+The expected output includes a classification index and (if enabled) timing measurements in nanoseconds, representing the total time taken to run the network on the input, excluding data transfer time.
+
 
 ```bash
 network-pointnet-cpp ModelNet10/directory/extracted_data ModelNet10/directory/input_cloud/itemName_cloud.bin
--- a/examples/network/pointnet.cpp
+++ b/examples/network/pointnet.cpp
@@ -351,7 +351,7 @@ struct FCLayer : public Layer<T> {
     void execute(dnnl::memory &in_mem) override {
 
         matmul_pd = dnnl::matmul::primitive_desc(
-                this->engine_, src_md, weights_md, this->out_desc_);
+                this->engine_, src_md, weights_md, bias_md, this->out_desc_);
 
         // Create the primitive.
         auto matmul_prim = dnnl::matmul(matmul_pd);
@@ -476,10 +476,10 @@ private:
     dnnl::binary::primitive_desc sum_pd;
 };
 template <typename T>
-struct SoftmaxLayer : public Layer<T> {
-    SoftmaxLayer(dnnl::engine &engine, dnnl::stream &stream, const int batch,
+struct LogSoftMaxLayer : public Layer<T> {
+    LogSoftMaxLayer(dnnl::engine &engine, dnnl::stream &stream, const int batch,
             const int channels, const int rows, const int cols,
-            dnnl::algorithm algo = dnnl::algorithm::softmax_accurate,
+            dnnl::algorithm algo = dnnl::algorithm::softmax_log,
             dnnl::memory::format_tag format = dnnl::memory::format_tag::nhwc,
             dnnl::memory::data_type data_type = dnnl::memory::data_type::f32)
         : Layer<T>(engine, stream) {
@@ -509,48 +509,12 @@ struct SoftmaxLayer : public Layer<T> {
         // Primitive execution.
         softmax_prim.execute(this->stream_, softmax_args);
     }
-    ~SoftmaxLayer() = default;
+    ~LogSoftMaxLayer() = default;
 
 private:
     dnnl::memory::desc src_md;
     dnnl::memory src_mem;
     dnnl::softmax_forward::primitive_desc softmax_pd;
-};
-
-template <typename T>
-struct LogLayer : public Layer<T> {
-    LogLayer(dnnl::engine &engine, dnnl::stream &stream, const int batch,
-            const int channels, const int rows, const int cols,
-            dnnl::memory::format_tag format = dnnl::memory::format_tag::nhwc,
-            dnnl::memory::data_type data_type = dnnl::memory::data_type::f32)
-        : Layer<T>(engine, stream) {
-
-        src_md = dnnl::memory::desc(
-                {batch, channels, rows, cols}, data_type, format);
-        this->out_desc_ = dnnl::memory::desc(
-                {batch, channels, rows, cols}, data_type, format);
-        this->out_mem_ = dnnl::memory(this->out_desc_, this->engine_);
-
-        eltwise_log_pd = dnnl::eltwise_forward::primitive_desc(this->engine_,
-                dnnl::prop_kind::forward_training, dnnl::algorithm::eltwise_log,
-                src_md, this->out_desc_);
-    }
-
-    void execute(dnnl::memory &in_mem) override {
-        auto eltwise_log = dnnl::eltwise_forward(eltwise_log_pd);
-
-        std::unordered_map<int, dnnl::memory> eltwise_args;
-        eltwise_args.insert({DNNL_ARG_SRC, in_mem});
-        eltwise_args.insert({DNNL_ARG_DST, this->out_mem_});
-
-        // Primitive execution: element-wise (ReLU).
-        eltwise_log.execute(this->stream_, eltwise_args);
-    }
-    ~LogLayer() = default;
-
-private:
-    dnnl::memory::desc src_md;
-    dnnl::eltwise_forward::primitive_desc eltwise_log_pd;
 };
 
 template <typename T>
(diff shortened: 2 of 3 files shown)


examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        weights_mem = dnnl::memory(weights_md, this->engine_);
+        bias_mem = dnnl::memory(bias_md, this->engine_);
+        dst_mem = dnnl::memory(this->out_desc_, this->engine_);
+
+        write_to_dnnl_memory(weights.data(), weights_mem);
+        write_to_dnnl_memory(bias.data(), bias_mem);
+
+        this->out_ptr_ = sycl::malloc_device<T>(batch * out_channels, q);
+
+        matmul_pd = dnnl::matmul::primitive_desc(
```
AD2605, line 401: Fully Connected (also known as the [linear layer](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html)) would technically map to the inner product primitive, and not a matmul

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+                this->engine_, src_md, weights_md, this->out_desc_);
```
AD2605, line 402: I might be misunderstanding / misremembering, but I see you create a `bias_md` above, and do not pass it to the `primitive_desc` constructor. Do you mean to use the bias ?

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        dst_dims = {1, batch, out_channels};
+
+        src_md = dnnl::memory::desc(
+                src_dims, data_type, dnnl::memory::format_tag::abc);
+        this->out_desc_ = dnnl::memory::desc(
+                dst_dims, data_type, dnnl::memory::format_tag::abc);
+        weights_md = dnnl::memory::desc(
+                weights_dims, data_type, dnnl::memory::format_tag::abc);
+        bias_md = dnnl::memory::desc(
+                bias_dims, data_type, dnnl::memory::format_tag::abc);
```
AD2605, line 386: PS: The bias in a fully connected layer (aka linear layer in pytorch) is one dimensional, one value per output channel

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        std::unordered_map<int, dnnl::memory> conv_args;
+        conv_args.insert({DNNL_ARG_SRC, conv_src_mem});
+        conv_args.insert({DNNL_ARG_WEIGHTS, conv_weights_mem});
+        conv_args.insert({DNNL_ARG_BIAS, conv_bias_mem});
+        conv_args.insert({DNNL_ARG_DST, conv_dst_mem});
+
+        conv_prim.execute(this->stream_, conv_args);
+
+        // Wait for the computation to finalize.
+        this->stream_.wait();
```
sgeor255, line 177: I think this wait and all other waits in layers' execute functions shouldn't be needed. You may need to remove the `read_from_dnnl_memory` below as they call `queue.memcpy()` and wait directly, rather than returning the event that you can use for synchronisation. I think you could just pass the output memory/pointer to the next layer directly though without needing to copy it, which should fix this issue.

examples/network/prepareData.py
```diff
@@ -0,0 +1,105 @@
...
+            read_pc(file_names[i], data, i)
+
+        with open("%s_cloud.bin" % furniture_item, "wb") as f:
+            data.tofile(f)
+
+        # load pretrained model and write weights to binary files for oneDNN to
+        # parse
+        if torch.cuda.is_available():
+            pretrained_model = torch.load(
+                pointnet_path, map_location=torch.device("cuda"))
```
AD2605, line 80: This seems unnecessary. as we are only loading to save the weights to disk,  hence no need to copy the weighs to the GPU first

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        prev = predicted[i];
+    }
+    if (count > mode_count) {
+        mode = prev;
+        mode_count = count;
+    }
+    std::cout << "classed as " << mode << " (i.e., " << object_map[mode] << ")"
+              << std::endl;
+
+#if CHECK_PERF
```
AD2605, line 1048: nit: ideally this should be a command line argument so that there's no need to recompile when trying to measure perf

examples/network/pointnet.cpp
s-Nick: @sgeor255  Thank you for pointing me in the right direction. I removed the usage of raw memory pointers in favor of `dnnl::memory` and everything works smoothly without sync as you said. You can check the changes in [9d2cc7b](https://github.com/oneapi-src/oneDNN/pull/2845/commits/9d2cc7beeff98f8c8579f11247805fe68d4ee230).

examples/network/pointnet.cpp
s-Nick: Thank you for the review @AD2605, due to many changes I forgot to add back bias to it, I address it in [3faeea6](https://github.com/oneapi-src/oneDNN/pull/2845/commits/3faeea64bb456ab57a8089c30f61bb0fbf1bd751)

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        dnnl::memory::dims dst_dims = {in_n, filt_f, oh, ow};
+        dnnl::memory::dims bias_dims = {filt_f};
+        dnnl::memory::dims strides_dims = {1, 1};
+        dnnl::memory::dims padding_dims_l = {0, 0};
+        dnnl::memory::dims padding_dims_r = {0, 0};
+        dnnl::memory::dims dilates = {0, 0};
+
+        const auto sycl_queue = dnnl::sycl_interop::get_queue(this->stream_);
+        this->out_ptr_
+                = sycl::malloc_device<T>(in_n * filt_f * oh * ow, sycl_queue);
```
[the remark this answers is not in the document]
s-Nick: Thank you for the comment, I changed the approach following [this comment](https://github.com/oneapi-src/oneDNN/pull/2845#discussion_r1987068476).
When I was porting it from portDNN, I switch to `raw pointers` because that is how it is implemented there and I am not super familiar with oneDNN. Now everything uses `dnnl::memory` as you pointed out and it is much better. The changes are in [9d2cc7b](https://github.com/oneapi-src/oneDNN/pull/2845/commits/9d2cc7beeff98f8c8579f11247805fe68d4ee230)

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+    if (!file.is_open()) {
+        throw std::runtime_error("Failed to open file " + name);
+    }
+    std::vector<char> output {std::istreambuf_iterator<char> {file}, {}};
+    return output;
+}
+
+// read image data from disk
+template <typename DType, typename DeviceMem, typename Backend>
+DeviceMem read_image_data(
```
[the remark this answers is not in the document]
s-Nick: I addressed this one and other minor changes in  [8a1f3e5](https://github.com/oneapi-src/oneDNN/pull/2845/commits/8a1f3e55d127cb9155e2e7d9ead51cb7a8b5bd44)

examples/network/pointnet.cpp
s-Nick: I followed portDNN implementation and kept `matmul`. I am now aware that inner product would be better, but I don't think this is an important change to make. Would you be fine if I keep the matmul implementation?

examples/network/pointnet.cpp
AD2605: No issues either way, but matmul also supports a broadcastable bias, so a 1 dimensional bias would have worked here as well.

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`
+
+## Testing on a pointcloud
+
+The oneDNN samples are built in the default CMake configuration. The sample
+is built by the target `network-pointnet-cpp`. The samples must first
+be passed the directory where the binary weights files are stored and the second
+argument should be the preprocessed pointcloud that should be classified. The expected
```
vpirogov, line 23: oneDNN examples are used in various CI environments and across different architectures. Dependency on Pytorch would be a bit of a hassle. I would suggest to put this one into [oneAPI Samples](https://github.com/oneapi-src/oneAPI-samples/tree/master/Libraries/oneDNN) instead of the main repo.

+@onednnsupporttriage

s-Nick: Thank you for suggesting a better place for our sample. Once all comments are solved, I'll open another PR to the appropriate repo.

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
+# PointNet Convolutional Neural Network Sample for 3D Pointcloud Classification
+
+[PointNet][pointnet-paper] is a convolutional neural network architecture for applications concerning 3D recognition such as object classification and part segmentation. These sample codes implement a variant of PointNet for 3D object classification, for inference only with ModelNet10, showing a larger example of using oneDNN. Some rough instructions for how it might be used are provided.
```
ranukund, line 3:
```suggestion
[PointNet][pointnet-paper] is a convolutional neural network architecture for applications concerning 3D recognition such as object classification and part segmentation. These sample codes implement a variant of PointNet for 3D object classification, for inference only with ModelNet10, providing a comprehensive example of using oneDNN. You can see the following initial instructions on using the samples.
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Obtaining the model weights and classes and preparing an input pointcloud
```
ranukund, line 5:
```suggestion
## Obtain the Model Weights and Classes and Prepare an Input pointcloud
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+A preprocessing script is provided which unpacks the weights from a pretrained pytorch model. The script also prepares an input pointcloud for testing inference. The pointcloud is made from 3D scans taken from the [ModelNet10][modelnet] dataset. The script requires an installation of [PyTorch][pytorch]. First download the pretrained PointNet weights and move the pth file into the same directory of the model.
```
ranukund, line 7:
```suggestion
A preprocessing script is provided which unpacks the weights from a pre-trained PyTorch model. The script also prepares an input pointcloud for testing inference. The pointcloud is made from 3D scans taken from the [ModelNet10][modelnet] dataset. The script requires an installation of [PyTorch][pytorch]. 

First, download the pre-trained PointNet weights and then move the pth file into the same directory containing the model.
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```bash
+wget http://3dvision.princeton.edu/projects/2014/3DShapeNets/ModelNet10.zip
+unzip ModelNet10.zip
+rm ModelNet10.zip
+python3 prepareData.py ModelNet10/ pointnet_model.pth
+```
+
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`
```
ranukund, line 16:
```suggestion
The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`.
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Testing on a pointcloud
```
ranukund, line 18:
```suggestion
## Test on a pointcloud
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The oneDNN samples are built in the default CMake configuration. The sample
+is built by the target `network-pointnet-cpp`. The samples must first
+be passed the directory where the binary weights files are stored and the second
+argument should be the preprocessed pointcloud that should be classified. The expected
+output is of a classification index and a series of times in nanoseconds that corresond
```
ranukund, line 24:
```suggestion
output is a classification index and a series of times in nanoseconds that correspond
```

examples/network/README.md
```diff
@@ -0,0 +1,35 @@
...
+rm ModelNet10.zip
+python3 prepareData.py ModelNet10/ pointnet_model.pth
+```
+
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`
+
+## Testing on a pointcloud
+
+The oneDNN samples are built in the default CMake configuration. The sample
+is built by the target `network-pointnet-cpp`. The samples must first
```
ranukund, line 21: Is the directory containing the binary weight files passed as the first argument when running a oneDNN sample? Suggesting rewrite for consideration to improve readability, please modify as you see fit

To test a sample, the directory where the binary weights files are stored must be passed as the first argument. The second argument should be the preprocessed pointcloud that should be classified.

ranukund (commented):
A few edits suggested, please incorporate as you see fit.

examples/network/prepareData.py
```diff
@@ -0,0 +1,105 @@
...
+            break
+        x, y, z = tuple([np.float32(s) for s in line.strip().split(" ")])
+        file_contents.append([x, y, z])
+    return file_contents
+
+
+def normalize(points):
+    norm_pointcloud = points - np.mean(points, axis=0)
+    norm_pointcloud /= np.max(np.linalg.norm(norm_pointcloud, axis=1))
+    return norm_pointcloud
```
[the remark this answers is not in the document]
s-Nick: This normalization seems to be common when using point cloud.
The idea is to centering all the points and then scaling to a unit sphere. There are more advanced techniques, but I think this simple one is good enough.

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+            "bn2.running_mean.bin", "bn2.running_var.bin", 32, 512, 256);
+
+    add_fc_layer(feature_transform_block, eng, stream,
+            data_dir + "fc3.weight.bin", data_dir + "fc3.bias.bin", 32, 256,
+            10);
+
+    add_softmax_layer(feature_transform_block, eng, stream, 32, 10, 1, 1);
+
+    add_log_layer(feature_transform_block, eng, stream, 32, 10, 1, 1);
+
```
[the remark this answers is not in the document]
s-Nick: Addressed in [3faeea6](https://github.com/uxlfoundation/oneDNN/pull/2845/commits/3faeea64bb456ab57a8089c30f61bb0fbf1bd751)

examples/network/README.md
s-Nick: Thank you for your feedback @ranukund, I update everything in [dedc77d](https://github.com/uxlfoundation/oneDNN/pull/2845/commits/dedc77d0958c75bc99893538da780e7ef13085eb)

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,1066 @@
...
+template <typename T>
+inline void add_conv_bias_layer(Network<T> &net, dnnl::engine &handle,
+        dnnl::stream &stream, std::string const &filter_file,
+        std::string const &bias_file, const int in_n, const int in_c,
+        const int in_h, const int in_w, const int filt_f, const int filt_c,
+        const int filt_h, const int filt_w) {
+    net.add_layer(std::make_unique<ConvBiasLayer<T>>(handle, stream,
+            filter_file, bias_file, in_n, in_c, in_h, in_w, filt_f, filt_c,
+            filt_h, filt_w));
+}
```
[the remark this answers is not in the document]
s-Nick: No particularly other utility, just wrap to `unique_pointer`

examples/network/README.md
chaoyu3: oneAPI oneDNN sample code(https://github.com/oneapi-src/oneAPI-samples/tree/master/Libraries/oneDNN) looks a better place for this sample,  @onednnsupporttriage can help on this code check.

examples/network/README.md
```diff
@@ -0,0 +1,38 @@
...
+rm ModelNet10.zip
+python3 prepareData.py ModelNet10/ pointnet_model.pth
+```
+
+The weights will be saved to `data/` and the input pointcloud will be saved as `itemName_cloud.bin`.
+
+## Test on a pointcloud
+
+The oneDNN samples are built in the default CMake configuration. The sample
+is built by the target `network-pointnet-cpp`.
```
shu1chen, line 23: Since this example will be moved to [oneAPI Samples](https://github.com/oneapi-src/oneAPI-samples/tree/master/Libraries/oneDNN) and will not be built with the oneDNN source code, could you please add the bash command line to build this example and link libdnnl.so?

examples/network/pointnet.cpp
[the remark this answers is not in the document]
AD2605: Then I suggest we remove these thin wrappers completely, or if you insist on having these, you can  condense almost 100 lines of these wrappers into the following

```cpp
template<class LayerType, T, typename... Args>
inline void make_unique_wrapper(Network<T>& network, const Args&... args) {
    net.add_layer(std::make_unique<LayerType>(args...));
}
```

ranukund (approved):
Minor edits suggested, thanks!

kminemur:
Thank you for the great commit introducing this new sample!

However, I encountered some issues while trying to run the sample on Ubuntu24.04 LTS environment. The current steps may be completed, particularly regarding the target model version and the corresponding PyTorch version. Additionally, there are some setup steps that need clarification:

**Python Environment:**
It’s unclear which numpy and torch versions are compatible with the point cloud dataset and the pre-trained model provided.

Necessary packages include:
```
pip install numpy==xxx
pip install torch==yyy
```

**Extracted Data Manipulation:**
Running prepareData.py failed due to redundant files being extracted when unzipping ModelNet10.zip.

Files like .DS_Store and README.txt in the ModelNet10 directory need to be removed for the script to work properly.

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,877 @@
...
+
+        dnnl::memory::dims src_dst_dims = {batch, channels, rows, cols};
+        src_md = dnnl::memory::desc(src_dst_dims, data_type, format);
+        this->out_desc_ = dnnl::memory::desc(src_dst_dims, data_type, format);
+
+        this->out_mem_ = dnnl::memory(this->out_desc_, this->engine_);
+        constexpr int axis = 1;
+
+        softmax_pd = dnnl::softmax_forward::primitive_desc(this->engine_,
+                dnnl::prop_kind::forward_training, algo, src_md,
```
AD2605, line 495: Minor: Should be `dnnl::prop_kind::forward_inference`

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,877 @@
...
+                dnnl::prop_kind::forward_inference,
+                dnnl::algorithm::pooling_max, src_md, this->out_desc_,
+                strides_dims, kernel_dims, dilation_dims, padding_dims_l,
+                padding_dims_r);
+    }
+
+    void execute(dnnl::memory &in_mem) override {
+        auto pooling_prim = dnnl::pooling_forward(pooling_pd);
+
+        // Primitive arguments. Set up in-place execution by assigning src as DST.
```
AD2605, line 302: Seems to be an unrelated comment, pooling does not support in-place operations

examples/network/pointnet.cpp
```diff
@@ -0,0 +1,877 @@
...
+        write_to_dnnl_memory(bias.data(), bias_mem);
+    }
+
+    void execute(dnnl::memory &in_mem) override {
+
+        matmul_pd = dnnl::matmul::primitive_desc(
+                this->engine_, src_md, weights_md, bias_md, this->out_desc_);
+
+        // Create the primitive.
+        auto matmul_prim = dnnl::matmul(matmul_pd);
```
AD2605, line 357: The creation of primitive descriptor and the primitive should be inside the class constructor, otherwise to each execute call, you will have the creation of primitive descriptor and the primitive, which are expensive processes. The primitive descriptor creation iterates over the available implementations and chooses the first implementation which returns `status::success`, and is also responsible for allocating any scratchpad memory typically. The primitive creation builds the kernel, thus these needs to be moved to the class constructor. Thus in the execute, all you need to do is set the arguments and invoke layer

Rbiessy:
Closing as the example has been pushed to https://github.com/SYCLOPS-Project-EU/pointnet-onednn-sample instead.

PR #2849, opened 2025-03-10
cpu: aarch64: modify acl_pooling for stateless functions

zhili03:
# Description

Make pooling use stateless pooling op from ACL.

# Checklist

## General

## Performance improvements

```
./tests/benchdnn/benchdnn --pool --mode=P --dt=f16:f16 --tag=axb --alg=avg_np ic64ih56oh56kh3ph1

stateful:
OMP_NUM_THREADS=4
total perf: min(ms):0.301025 avg(ms):0.318544
OMP_NUM_THREADS=16
total perf: min(ms):0.121338 avg(ms):0.131665

stateless:
OMP_NUM_THREADS=4
total perf: min(ms):0.304443 avg(ms):0.320418
OMP_NUM_THREADS=16
total perf: min(ms):0.123047 avg(ms):0.133939
```

### DIFF
--- a/src/common/memory_tracking.hpp
+++ b/src/common/memory_tracking.hpp
@@ -283,6 +283,7 @@ enum {
     key_pool_src_f32_accum,
     key_pool_src_plain2blocked_cvt,
     key_pool_reduction,
+    key_pool_wsp_buffer,
     key_precomputed_scales,
     key_prelu_reduction,
     key_reducer_space,
--- a/src/cpu/aarch64/acl_pooling.cpp
+++ b/src/cpu/aarch64/acl_pooling.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2023 Arm Ltd. and affiliates
+* Copyright 2022-2023, 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


zhili03:
Hi @Sqvid , thanks for it, will fix the format issue and commit messages in my next commit

Ryo-not-rio:
Hi Zhibo, could you add performance numbers and the corresponding benchdnn command? Thank you!

zhili03:
Hi Ryo, sure, will do in the next commit as well, thanks!

zhili03:
Hi @Sqvid @Ryo-not-rio some CI pipelines are failing because the stateless pooling function haven't been merge to ACL yet, does that mean this PR cannot be merged at least until next ACL's release? Thanks

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,35 +21,72 @@ namespace impl {
...
+    // Configure pooling operation when workspace tensor is not used, mem allocation happens
+    else {
+        pooling_op_->configure(
+                &asp.src_info, &asp.dst_info, asp.pool_info, nullptr);
+    }
+
+    return status::success;
+}
+
 status_t acl_pooling_fwd_t::execute_forward(const exec_ctx_t &ctx) const {
```
Sqvid, line 251: Is this thread-safe or does it need to be locked?

.github/automation/aarch64/ci.json
```diff
@@ -1,7 +1,7 @@
 {
     "dependencies": {
-        "acl": "v25.02",
+        "acl": "v52.1.0",
```
Sqvid: This change is already in main. You probably need to rebase.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,35 +21,72 @@ namespace impl {
 namespace cpu {
 namespace aarch64 {
 
+status_t acl_pooling_fwd_t::init(engine_t *engine) {
+    auto asp = pd()->asp_;
+
+    auto op = std::make_unique<arm_compute::experimental::op::CpuPool2d>();
+
+    pooling_op_ = std::move(op);
```
Sqvid: This should be initialised in the constructer via the initializer list syntax

src/cpu/aarch64/acl_pooling.hpp
```diff
@@ -17,72 +17,38 @@
...
     struct pd_t : public cpu_pooling_fwd_pd_t {
         using cpu_pooling_fwd_pd_t::cpu_pooling_fwd_pd_t;
-
-        DECLARE_COMMON_PD_T("acl", acl_pooling_fwd_t);
+        DECLARE_COMMON_PD_T("acl", acl_pooling_fwd_t, USE_GLOBAL_SCRATCHPAD);
 
         status_t init(engine_t *engine) {
-            bool ok = set_default_params() == status::success
-                    && is_fwd() // ACL supports forward propagation only
+            auto scratchpad = scratchpad_registry().registrar();
```
Sqvid: In general it would be good to move all function definitions to the *.cpp file and keep only declarations in the *.hpp

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,35 +21,72 @@ namespace impl {
...
+
+    // Configure pooling operation when workspace tensor is used, mem allocation happens
+    if (asp.use_ws) {
```
Sqvid: if you do

```cpp
auto& ws_info = asp.use_ws ? &asp.ws_info : nullptr;
```

Then you can reduce the clutter with  a single call to `configure()`

Sqvid (changes requested):
Thank you for the patch! I have left a few comments.

.github/automation/aarch64/ci.json
zhili03: OK! Will rebase then

src/cpu/aarch64/acl_pooling.hpp
zhili03: OK, will do, thanks!

src/cpu/aarch64/acl_pooling.cpp
zhili03: no, the wrapper `CpuPool2d` from ACL is not thread safe. But I wonder is a lock necessary for the configure function here as it seems that concurrent access would not happen here?

Sqvid: Sorry the comment was intended for `execute_forward()`

zhili03: No worries! then I need to investigate if this is thread safe as `scratchpad` is used here.

.github/automation/aarch64/ci.json
Sqvid: I think I was misremembering with this. [It's just that there is another open PR](https://github.com/uxlfoundation/oneDNN/pull/2992/files#diff-52ae88455683350ca19a278cf88cfa2974e821987001a23a4193733e93b1199b) with the exact same change :laughing:

zhili03: No problem! I'll just keep this.

src/cpu/aarch64/acl_pooling.hpp
```diff
@@ -17,72 +17,38 @@
...
-private:
-    std::unique_ptr<acl_pooling_obj_t> acl_pooling_obj_;
-}; // acl_pooling_resource_t
-
 struct acl_pooling_fwd_t : public primitive_t {
     struct pd_t : public cpu_pooling_fwd_pd_t {
         using cpu_pooling_fwd_pd_t::cpu_pooling_fwd_pd_t;
-
-        DECLARE_COMMON_PD_T("acl", acl_pooling_fwd_t);
+        DECLARE_COMMON_PD_T("acl", acl_pooling_fwd_t, USE_GLOBAL_SCRATCHPAD);
```
almayne, line 44: I don't think we need to USE_GLOBAL_SCRATCHPAD here, as the scratchpad mode will be set to user, not library.

Sqvid: @almayne I think @zhili03 is right to use `USE_GLOBAL_SCRATCHPAD` here. It is true that upstream frameworks will probably use `scratchpad_mode::user` but this is a sensible default for us within oneDNN I think.

almayne: We probably want to keep that mutex then, if we can't guarantee how the scratchpad will be managed?

Sqvid: Yes I suspect we need the mutex in `execute_forward` which I raised here https://github.com/uxlfoundation/oneDNN/pull/2849#discussion_r2142142989

zhili03: I got some different findings - it might be fine even without a mutex lock for the scratchpad? [primitive-attributes-scratchpad](https://www.intel.com/content/www/us/en/docs/onednn/developer-guide-reference/2025-0/primitive-attributes-scratchpad.html) and [scratchpad-mode](https://www.intel.com/content/www/us/en/docs/onednn/developer-guide-reference/2025-0/enum-dnnl-scratchpad-mode.html#DOXID-GROUP-DNNL-API-ATTRIBUTES-1GGAC24D40CEEA0256C7D6CC3A383A0FA07FAD521F765A49C72507257A2620612EE96) mention that when `ONEDNN_ENABLE_CONCURRENT_EXEC` set to OFF by default, the scratchpad will have limited thread safety - the library won't let access of a primitive's `execute_forward()` from different threads at the same time. I'd like to know in oneDNN, if a user is allowed to run the same `primitive object` concurrently from two threads? If that is the case, then it would be better to lock it.

Sqvid: As far as I understand it is permitted within the programming model to execute the same primitive object in multiple threads.

vpirogov:
@zhili03, @Sqvid, is this PR still relevant?

Sqvid:
@vpirogov Yes, still in review

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,265 @@ namespace impl {
...
+    if (asp_.use_ws) {
+        scratchpad.book(memory_tracking::names::key_pool_ind_plain2blocked_cvt,
+                dst_d.nelems(), sizeof(uint32_t));
+    }
+    return status::success;
+}
+
+acl_pooling_fwd_t::acl_pooling_fwd_t(const pd_t *apd)
+    : primitive_t(apd)
+    , pooling_op_ {
```
Sqvid: nit: style: use `pooling_op_(make_unique...)` instead of `pooling_op_ { make_unique... }` for consistency.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,265 @@ namespace impl {
...
 
-    // import_memory() and free() methods do not allocate/free any additional
-    // memory, only acquire/release pointers.
-    acl_obj.src_tensor.allocator()->import_memory(const_cast<void *>(src_base));
-    acl_obj.dst_tensor.allocator()->import_memory(dst_base);
-    if (acl_obj.use_ws) acl_obj.ws_tensor.allocator()->import_memory(ws_base);
+    arm_compute::Tensor ws_tensor;
 
-    acl_obj.pool.run();
+    if (asp.use_ws) {
```
Sqvid: Can this `if` clause be combined with the other `if (asp.use_ws)` one below?

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,265 @@ namespace impl {
...
+                cutoff = 25000;
+        }
+    }
+    return problem_size > cutoff * thread_count;
+}
+
+status_t acl_pooling_fwd_t::pd_t::init_scratchpad(
+        memory_tracking::registrar_t &scratchpad) {
+    const memory_desc_wrapper dst_d(&dst_md_);
+    scratchpad.book(memory_tracking::names::key_pool_reduction, dst_d.nelems(),
```
Sqvid: Are these `memory_tracking` key names representing their use? I think historically we have added our own when needed.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,265 @@ namespace impl {
...
+acl_pooling_fwd_t::acl_pooling_fwd_t(const pd_t *apd)
+    : primitive_t(apd)
+    , pooling_op_ {
+              std::make_unique<arm_compute::experimental::op::CpuPool2d>()} {}
+
+status_t acl_pooling_fwd_t::execute(const exec_ctx_t &ctx) const {
+    return execute_forward(ctx);
+}
+
+const acl_pooling_fwd_t::pd_t *acl_pooling_fwd_t::pd() const {
```
Sqvid: Thanks for moving the definitions to the `.cpp` file. However, I should have mentioned this particular function might be better left in the `.hpp` file [as discussed here](https://github.com/uxlfoundation/oneDNN/pull/2992#discussion_r2153657576).

Sqvid (changes requested):

src/cpu/aarch64/acl_pooling.cpp
zhili03: Upon checking [memory_tracking,hpp](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/memory_tracking.hpp), there are 
    `key_pool_dst_bf16cvt`,
    `key_pool_dst_plain2blocked_cvt`,
    `key_pool_ind_plain2blocked_cvt`,
    `key_pool_src_bf16cvt`,
    `key_pool_src_f32_accum`,
    `key_pool_src_plain2blocked_cvt`,
    `key_pool_reduction`
I think the `key_pool_reduction` seems to be suitable? Correct me if I was wrong as I do not know exactly the meanings of other names (I can only guess they specify the properties for src/dst/ind tensors of different data types). Thanks

zhili03:
For the most recent [CI failure](https://github.com/uxlfoundation/oneDNN/actions/runs/17325118469/job/49187346077?pr=2849), that failed test case passed on my local develop machine,
`./tests/benchdnn/benchdnn --pool --tag=axb --alg=avg_np ic5ih15oh14kh3ph1n"pool_ci_2d:only_pad_l"
0:PASSED (11 ms) __REPRO: --pool --tag=axb --alg=avg_np ic5ih15oh14kh3ph1npool_ci_2d:only_pad_l
total: 0.01s; create_pd: 0.00s (3%); create_prim: 0.00s (1%); fill: 0.01s (45%); execute: 0.00s (4%); compute_ref: 0.00s (4%); compare: 0.00s (7%);`. Could reviewers @Sqvid @Ryo-not-rio @almayne help look into this? Thanks

Sqvid:
> For the most recent [CI failure](https://github.com/uxlfoundation/oneDNN/actions/runs/17325118469/job/49187346077?pr=2849), that failed test case passed on my local develop machine

I believe the one you posted was the last successful test. Since you hit an `assert()` the failing `REPRO` line was not hit.

Here is a failing case:
```
./build/tests/benchdnn/benchdnn --pool --dt=f16:f16 --tag=axb --alg=avg_np ic64ih56oh56kh3ph1
```
The failing `assert()` is here:
https://github.com/uxlfoundation/oneDNN/blob/2945509677ef27f475d9cd8d26059af8317161dc/src/common/memory_tracking.hpp#L399

You will need to compile with asserts enabled. i.e, with `CMAKE_BUILD_TYPE={Debug|RelWithAssert}`.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,282 @@ namespace impl {
...
+        if (req.size > 0) {
+            size_t align = req.alignment ? req.alignment : 64;
+            if (align & (align - 1)) align = utils::rnd_up_pow2(align);
+            scratchpad.book(memory_tracking::names::key_pool_int, req.size, 1,
+                    align, align);
+        }
+    }
+
+    const memory_desc_wrapper dst_d(&dst_md_);
+    scratchpad.book(memory_tracking::names::key_pool_int, dst_d.nelems(),
```
dzarukin: Using two identical keys is leading to an error (likely what's been caught).

Either you are booking the biggest size among two if scratchpad buffers don't intersect when used (better memory utilization), or allocate into two separate keys.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,282 @@ namespace impl {
 namespace cpu {
 namespace aarch64 {
 
+using pooling_key_t = decltype(memory_tracking::names::key_pool_reduction);
```
dzarukin: Unused.

src/cpu/aarch64/acl_pooling.cpp
zhili03: deleted, as the second booking is never used.

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -21,33 +21,276 @@ namespace impl {
...
+status_t acl_pooling_fwd_t::pd_t::init_scratchpad(
+        memory_tracking::registrar_t &scratchpad,
+        const arm_compute::experimental::MemoryRequirements &aux_mem_req) {
+
+    if (!aux_mem_req.empty()) {
+        const auto &req = aux_mem_req[0];
+        if (req.size > 0) {
+            size_t align = req.alignment ? req.alignment : 64;
+            if (align & (align - 1)) align = utils::rnd_up_pow2(align);
+            scratchpad.book(memory_tracking::names::key_pool_int, req.size, 1,
```
Sqvid: Include `src/common/memory_tracking.hpp`

src/common/memory_tracking.hpp
```diff
@@ -279,6 +279,7 @@ enum {
     key_pool_dst_bf16cvt,
     key_pool_dst_plain2blocked_cvt,
     key_pool_ind_plain2blocked_cvt,
+    key_pool_int,
```
Sqvid: `key_pool_int` is a little confusing as a name.\

Since it is used internally as a working space I would suggest calling it
`key_pool_wsp_buffer` since a [similarly named key exists already](https://github.com/uxlfoundation/oneDNN/blob/ddcb9bebe093078a407dd3ba8d543717419d384e/src/common/memory_tracking.hpp#L181).

Make sure you keep the enum list alphabetically sorted after the change as well.

src/cpu/aarch64/acl_pooling.cpp
zhili03: Hi, `memory_tracking.hpp` is included from `acl_utils.hpp` where `acl_pooling.hpp` includes, I wonder shall we include `memory_tracking.hpp` here again? thanks

Sqvid: It doesn't matter too much. I would prefer to include what we use directly but I won't block the PR for this.

zhili03: Great! I'll add `memory_tracking.hpp` here for better consistency.

zhili03:
> LGTM, thanks!

Thanks for your help :)

Radu2k (approved):

Sqvid:
@dzarukin Would appreciate a review from @uxlfoundation/onednn-arch, thanks!

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -15,39 +15,283 @@
...
+
+    // Compute Library supports forward propagation only
+    bool ok = set_default_params() == status::success && is_fwd()
+            && utils::everyone_is(src_md()->data_type, dst_md()->data_type)
+            && utils::one_of(
+                    src_md()->data_type, data_type::f32, data_type::f16)
+            && attr()->has_default_values()
+            && attr_.set_default_formats(dst_md(0)) == status::success
+            && !is_dilated() && !has_zero_dim_memory();
+    if (!ok) return status::unimplemented;
```
dzarukin: Minor: might want to add verbose reporting here.

dzarukin (approved):

src/cpu/aarch64/acl_pooling.cpp
zhili03: thanks! will fix soon

src/cpu/aarch64/acl_pooling.cpp
```diff
@@ -15,39 +15,284 @@
...
+    // Compute Library supports forward propagation only
+    bool ok = set_default_params() == status::success && is_fwd()
+            && utils::everyone_is(src_md()->data_type, dst_md()->data_type)
+            && utils::one_of(
+                    src_md()->data_type, data_type::f32, data_type::f16)
+            && attr()->has_default_values()
+            && attr_.set_default_formats(dst_md(0)) == status::success
+            && !is_dilated() && !has_zero_dim_memory();
+
+    ACL_CHECK_SUPPORT(!ok, "ACL supports forward propagation only");
```
Sqvid: This error message will be confusing since `ok` might be `false` even for forward propagation (e.g. if `dt == s8`). Just print something more generic like "unsupported primitive options".

PR #2851, opened 2025-03-10
common: gemm: fix illegal ld for size-1 dimensions

petercad:
oneDNN strides are undefined for size-1 dimensions, so gemm_desc_t::get_ld returns ld = 1 rather than relying on the undefined stride. However, this is only a legal ld for 1x1 matrices. This can cause various problems for downstream GEMM implementations that are expecting a legal ld.


petercad:
make test

dzarukin (approved):

kealan-barbieri (approved):

src/common/gemm_types.hpp
```diff
@@ -125,9 +125,11 @@ struct gemm_desc_t : public op_desc_t {
                 || md.dims[md.ndims - 2] == 1 || strides[md.ndims - 2] == 1);
         switch (get_trans(md)) {
             case transpose::trans:
-                return md.dims[md.ndims - 1] > 1 ? strides[md.ndims - 1] : 1;
+                return md.dims[md.ndims - 1] > 1 ? strides[md.ndims - 1]
+                                                 : md.dims[md.ndims - 2];
             default:
-                return md.dims[md.ndims - 2] > 1 ? strides[md.ndims - 2] : 1;
+                return md.dims[md.ndims - 2] > 1 ? strides[md.ndims - 2]
+                                                 : md.dims[md.ndims - 1];
```
rjoursler, line 132: Out of curiosity, how is this causing failures?

Also, would it make sense to round this up to a well aligned value, or is that irrelevant for strategy selection?

petercad: Well, GEMM (the original BLAS routine) requires leading dimension >= size in stride-1 dimension.

So any time GEMM is implemented with an external BLAS library (optional on some platforms when building oneDNN), an error may be reported due to incorrect GEMM arguments.

There are also assumptions within the Xe GEMM implementation of oneDNN that the LD is valid: e.g., when
- determining the memory footprint of a matrix
- using block 2D messages (which have the same restriction)

In terms of rounding -- in my mind, this is implementation specific so shouldn't be done here. Currently gen_gemm_t does its own rounding to avoid unnecessary alignment restrictions.

rjoursler (approved):

PR #2862, opened 2025-03-12
ukernel: add a common shim layer for better maintainability and split x64 APIs and abstractions into several files

dzarukin:
The common shim layer should allow to modify existing or introduce new APIs without updating all backends but just the one desired.
With current model every change to API forces a developer to update all backends, otherwise, a compilation or linking failure will pop up.

Splitting x64 abstractions follows primitive abstraction coding practice.


dzarukin:
make test

src/cpu/x64/ukernel/transform.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+#ifdef DNNL_EXPERIMENTAL_UKERNEL
+
+using namespace dnnl::impl;
+using namespace dnnl::impl::cpu::x64;
+using namespace dnnl::impl::cpu::ukernel;
+
+#define VCHECK_BRGEMM(cond, msg, ...) \
+    VCONDCHECK(ukernel, create, check, brgemm, (cond), \
+            status::invalid_arguments, msg, ##__VA_ARGS__)
```
Radu2k, line 29: I think we should move this in the corresponding BRGeMM file and import it here as this is a BRGEMM check.

Radu2k (approved):
LGTM. There is one recommendation that I would suggest.

src/cpu/x64/ukernel/transform.cpp
dzarukin: This is a transform one, I'll rename it, thanks for catching.

src/cpu/ukernel/brgemm.cpp
```diff
@@ -0,0 +1,157 @@
...
+using namespace dnnl::impl::cpu::ukernel;
+
+status_t dnnl_brgemm_create(brgemm_t **brgemm, dim_t M, dim_t N, dim_t K,
+        dim_t batch_size, dim_t lda, dim_t ldb, dim_t ldc, data_type_t a_dt,
+        data_type_t b_dt, data_type_t c_dt) {
+#if DNNL_X64
+    return x64::ukernel::dnnl_brgemm_create(
+            brgemm, M, N, K, batch_size, lda, ldb, ldc, a_dt, b_dt, c_dt);
+#endif
+    return status::success;
```
mgouicem, line 38:
```suggestion
    return status::unimplemented;
```

Why return success and not unimplemented?
Also, we have to think if we will ever add reference path or not to cover backend specific limitations.

mgouicem (commented):
LGTM overall.
Just a couple of points:
- why return success by default for unimplemented features (instead of unimplemented)?
- for arch that don't support this API extension, maybe we should either provide reference fallback, or we should disable in CMakeList.txt.

dzarukin:
> LGTM overall. Just a couple of points:
> 
> * why return success by default for unimplemented features (instead of unimplemented)?
> * for arch that don't support this API extension, we might be better off disabling the feature in [CMakeList.txt](https://github.com/uxlfoundation/oneDNN/blob/2fa9425167a8c5057833051d8b486e9ccd78b314/src/CMakeLists.txt#L84). (or maybe in the future provide a reference fallback).

Addressed both of those. Please take a look, thanks for raising the points.

mgouicem (approved):

ankalinin (approved):

PR #2863, opened 2025-03-12
graph: interface: several fixes and code improves

TaoLv:
1. JSON file serialization: removed the trailing whitespaces which may cause unintended edits when a JSON file is opened and closed with an editor (eg. vscode) configured with removing trailing whitespaces. Also added a newline at the end of the JSON file.
2. Moved the graph::serialize() method to cpp file so changing it or the JSON utility will not cause the whole graph component recompile (due to graph.hpp is almost included everywhere).
3. Backend interface: removed two lock guards which seem unnecessary.
4. fpmath_t: make the constructor explicit so we don't confuse it with fpmath_mode_t. At some places, we passed fpmath_mode_t to functions which requires fpmath_t as an argument.

### DIFF
--- a/src/graph/interface/backend.hpp
+++ b/src/graph/interface/backend.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2020-2024 Intel Corporation
+ * Copyright 2020-2025 Intel Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

ElaineBao (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2866, opened 2025-03-12
xe: ocl: prevent double->float literal conversion in post ops

rjoursler:
When cl_khr_f64 is unsupported, the compiler warns about double literals.

### DIFF
--- a/src/gpu/intel/ocl/ocl_eltwise.h
+++ b/src/gpu/intel/ocl/ocl_eltwise.h
@@ -35,6 +35,12 @@
 #endif
 #endif
 
+#if DT_F64 == 1
+#define POST_OP_LITERAL(x) x
+#else
+#define POST_OP_LITERAL(x) x##f
+#endif
+
 POST_OP_DATA_T relu_fwd(POST_OP_DATA_T s, float alpha) {
     return s > 0 ? s : s * alpha;
 }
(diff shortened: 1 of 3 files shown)


rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

src/gpu/intel/ocl/ocl_eltwise.h
```diff
@@ -217,19 +223,19 @@ POST_OP_DATA_T pow_bwd(
 
 POST_OP_DATA_T gelu_erf_fwd(POST_OP_DATA_T s) {
     const POST_OP_DATA_T sqrt_2_over_2
-            = AS_POST_OP_DATA_T(0.707106769084930419921875);
```
atkassen, line 220: Other than the numbers that would be truncated, why not just keep everything in explicit `float` or `int` precision and let type promotion/conversion do its thing?

rjoursler: There are a few of reasons:

1. This can prevent some hidden precision loss. For example, consider the expression `2.0f * alpha`. By our API, `alpha` is a float, so `2.0f * alpha` will will have type float. If we are accumulating in double, this typing is incorrect. As such we need to perform a conversion on `2.0f`, and as such, we don't end up save anything over  directly setting the literal to the right type.

2. Circumvents complexity in determining which literals are exactly representable in float. While a lot of the cases are fairly obvious (0, 1, etc), it requires knowledge of the IEEE floating point specification to verify correctness.

3. To maintain consistency throughout this file.  In other files where we do not have "complex" constants, just relying on type promotion is fine with me.

kealan-barbieri (approved):

atkassen (approved):

PR #2875, opened 2025-03-13
cpu: pooling: fix crashes of large tensor processing

asimonov1:
This PR fixes crashes of simple_nchw and simple_nhwc in some cases that were found in [MFDNN-13286](https://jira.devtools.intel.com/browse/MFDNN-13286).
It also reduces memory allocation in simple_nchw in case of back propagation for non-f32 data types: simple_nchw requested scratchpad memory for every available thread even if the number of work items was smaller (e.g. it requested more than 4TB for mb1ic1iw4294967311ow858993461kw7sw5pw0 according to logs in MFDNN-13286; the size depends on the number of available threads). This helps in some cases (particularly, in the cases from MFDNN-13286).

### DIFF
--- a/src/cpu/nchw_pooling.hpp
+++ b/src/cpu/nchw_pooling.hpp
@@ -174,8 +174,9 @@ struct nchw_pooling_bwd_t : public primitive_t {
             return status::success;
         }
 
-        dim_t channel_block_size_;
+        dim_t channel_block_size_ {1};
         int nthr_; // To not exceed the limit in execute used for set up.
+        int nbuf_ {0};
 
     private:
         void init_scratchpad() {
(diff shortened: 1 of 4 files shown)


asimonov1:
make test
disable benchdnn_all
disable test_device_gpu
disable build_gpu_runtime_ocl
enable benchdnn_pool

src/cpu/nchw_pooling.hpp
```diff
@@ -192,19 +191,26 @@ struct nchw_pooling_bwd_t : public primitive_t {
...
-            dim_t src_sz_ = ID() * IH() * IW();
-            dim_t C_per_thr = nstl::min(MB() * IC() / nthr_, IC());
-            const dim_t max_block_size
-                    = platform::get_per_core_cache_size(1) / 2;
-            dim_t data_size_per_ch = (dst_sz_ + src_sz_) * 6; // f32 + bf16
-            channel_block_size_ = nstl::max(
-                    nstl::min(C_per_thr, max_block_size / data_size_per_ch),
-                    (dim_t)1);
+        void calculate_nthr_and_channel_block_size() {
+            nthr_ = dnnl_get_max_threads();
```
dzarukin, line 195: `nthr_` are supposed to be always equal to the upper level the user sets. The reason for that is the following:

Parallel function creates a pool of specified number of threads, and when the number of threads changes (let's consider convolution before pooling which used all threads), the underlying OMP/TBB implementation will re-create the pool object, which is global, at runtime, and this re-creation is expensive, and then the next op might use again the full number of threads and trigger the underlying pool object re-creation again, so paying double the price.

Allowed scenarios are limited to either passing `nthr=1` instead of `nthr_` from `pd` to the parallel call, which causes a sequential run and avoid underlying parallel section creation all together, or inside the parallel call make tail threads/workers drop the work with a condition.

This way it has way less overhead on spawning extra threads than on the singleton re-creation. This will require careful alignment between parallelizing logic with blocking/balancing logic.

asimonov1: Reimplemented. A value of `nthr_` is restored.

If I understand correctly how `parallel_nd_ext` is implemented (including the called function `balance211`), the current implementation should work.

asimonov1:
make test
disable benchdnn_all
disable test_device_gpu
disable build_gpu_runtime_ocl
enable benchdnn_pool

src/cpu/nchw_pooling.hpp
```diff
@@ -185,26 +186,35 @@ struct nchw_pooling_bwd_t : public primitive_t {
                 size_t src_sz_ = ID() * IH() * IW();
                 auto scratchpad = scratchpad_registry().registrar();
 
+                nbuf_ = nstl::min(static_cast<dim_t>(nthr_),
+                        MB() * utils::div_up(IC(), channel_block_size_));
```
dzarukin, line 190: I'd probably leave a comment to make sure that parallel sections are in compliance with the future changes. The assert added there won't save release mode and will likely explode somewhere.

src/cpu/nchw_pooling.hpp
```diff
@@ -176,6 +176,7 @@ struct nchw_pooling_bwd_t : public primitive_t {
 
         dim_t channel_block_size_;
         int nthr_; // To not exceed the limit in execute used for set up.
+        int nbuf_;
```
dzarukin, line 179: Probably fine to zero initialize in the beginning to avoid else branch. Same with `channel_block_size_`.

dzarukin (approved):

src/cpu/nchw_pooling.hpp
asimonov1: Done

src/cpu/nchw_pooling.hpp
asimonov1: Done

ankalinin (approved):

PR #2876, opened 2025-03-13
gpu: intel: ocl: add s32 support for binary

yehudaorel:
# Description

Currently the binary primitive doesn't support s32 dtype on GPU, which will affect the reference implementation of SDPA with implicit causal mask

The following change enables support for s32 data types for gpu binary primitive.

Fixes [MFDNN-12984](https://jira.devtools.intel.com/browse/MFDNN-12984)

# Checklist

## General

### DIFF
--- a/tests/benchdnn/inputs/binary/test_binary_gpu
+++ b/tests/benchdnn/inputs/binary/test_binary_gpu
@@ -36,6 +36,10 @@
 --sdt=f32:f32
 --batch=option_set_minimal
 
+--ddt=s32
+--sdt=s32:s32
+--batch=option_set_minimal
+
 --ddt=s8
 --sdt=s8:s8
 --attr-scales=,src:common:0.25+src1:common:0.5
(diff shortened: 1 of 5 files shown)


karturov (commented):
Documentation needs to be adjusted as well here: https://uxlfoundation.github.io/oneDNN/dev_guide_binary.html#implementation-limitations

atkassen:
Does simple binary also need to be adjusted to allow s32?

dzarukin (approved):

tests/benchdnn/inputs/binary/test_binary_gpu
```diff
@@ -87,6 +91,10 @@
 --sdt=f32:f32
 --batch=option_set_all
 
+--ddt=s32
+--sdt=s32:s32
+--batch=option_set_all
```
ElaineBao, line 96: Do you need to remove the unimplemented statement here? 

https://github.com/yehudaorel/oneDNN/blob/gpu_binary_s32/tests/benchdnn/binary/binary.cpp#L150-L156

karturov (approved):

echeresh (approved):

PR #2877, opened 2025-03-13
ref: sycl: enable zp support in reorder and skip unsupported tests

AD2605:
# Description

This PR enables zero point support to the SYCL backend's reorder, and adds skips in benchdnn to skip unsupported cases for the SYCL backend

## General

### DIFF
--- a/src/gpu/generic/sycl/ref_reorder.cpp
+++ b/src/gpu/generic/sycl/ref_reorder.cpp
@@ -27,15 +27,21 @@ namespace sycl {
 status_t ref_reorder_t::pd_t::init_conf() {
     conf_ = sycl_reorder_conf_t();
 
-    conf_.src_md = xpu::sycl::md_t(src_md(0));
-    conf_.dst_md = xpu::sycl::md_t(dst_md());
+    conf_.src_md = xpu::sycl::md_t(arg_md(DNNL_ARG_FROM));
+    conf_.dst_md = xpu::sycl::md_t(arg_md(DNNL_ARG_TO));
 
-    conf_.wk_size = memory_desc_wrapper(src_md(0)).nelems();
+    auto dst_nelems = memory_desc_wrapper(arg_md(DNNL_ARG_TO)).nelems(true);
+    conf_.num_elements = dst_nelems;
 
-    conf_.do_scale_src = !attr()->scales_.has_default_values(DNNL_ARG_SRC_0);
-    conf_.scale_src_mask = attr()->scales_.get_mask(DNNL_ARG_SRC_0);
-    conf_.do_scale_dst = !attr()->scales_.has_default_values(DNNL_ARG_DST);
-    conf_.scale_dst_mask = attr()->scales_.get_mask(DNNL_ARG_DST);
+    conf_.apply_src_scale = !attr()->scales_.has_default_values(DNNL_ARG_FROM);
+    conf_.src_scale_mask = attr()->scales_.get_mask(DNNL_ARG_FROM);
+    conf_.apply_dst_scale = !attr()->scales_.has_default_values(DNNL_ARG_TO);
+    conf_.dst_scale_mask = attr()->scales_.get_mask(DNNL_ARG_TO);
+    conf_.apply_src_zp
+            = !attr()->zero_points_.has_default_values(DNNL_ARG_FROM);
+    conf_.src_zp_mask = attr()->zero_points_.get_mask(DNNL_ARG_FROM);
+    conf_.apply_dst_zp = !attr()->zero_points_.has_default_values(DNNL_ARG_TO);
+    conf_.dst_zp_mask = attr()->zero_points_.get_mask(DNNL_ARG_TO);
     conf_.post_ops = sycl_post_ops_t(attr(), dst_md());
 
     return status::success;
@@ -51,7 +57,8 @@ status_t ref_reorder_t::execute(const exec_ctx_t &ctx) const {
     parallel_for(ctx, kernel_, [&](::sycl::handler &cgh) {
         reorder_kernel_t reorder_kernel(pd()->conf_, cgh, ctx);
 
-        cgh.parallel_for(get_range(ctx, pd()->conf_.wk_size), reorder_kernel);
+        cgh.parallel_for(
+                get_range(ctx, pd()->conf_.num_elements), reorder_kernel);
     });
 
     return status::success;
--- a/src/gpu/generic/sycl/ref_reorder.hpp
+++ b/src/gpu/generic/sycl/ref_reorder.hpp
@@ -54,11 +54,11 @@ struct ref_reorder_t : public gpu::generic::sycl::primitive_t {
             VDISPATCH_REORDER(!dst_d.has_runtime_dims_or_strides(),
                     VERBOSE_RUNTIMEDIM_UNSUPPORTED);
             VDISPATCH_REORDER(
-                    check_data_types(src_d, dst_d), VERBOSE_UNSUPPORTED_DT_CFG);
-            VDISPATCH_REORDER(
                     check_formats(src_d, dst_d), VERBOSE_UNSUPPORTED_TAG);
             VDISPATCH_REORDER(
-                    attr()->has_default_values(sm::scales | sm::post_ops),
+                    check_data_types(src_d, dst_d), VERBOSE_UNSUPPORTED_DT_CFG);
+            VDISPATCH_REORDER(attr()->has_default_values(sm::post_ops
+                                      | sm::scales | sm::zero_points),
                     VERBOSE_UNSUPPORTED_ATTR);
             VDISPATCH_REORDER(IMPLICATION(!attr()->scales_.has_default_values(),
                                       scales_ok()),
@@ -108,6 +108,7 @@ struct ref_reorder_t : public gpu::generic::sycl::primitive_t {
                 const auto dt = scales.get_data_type(arg);
                 if (!is_supported_type(dt)) { return false; }
             }
+
             return true;
         }
     };
--- a/src/gpu/generic/sycl/reorder_kernels.hpp
+++ b/src/gpu/generic/sycl/reorder_kernels.hpp
@@ -37,106 +37,116 @@ struct reorder_kernel_t {
     reorder_kernel_t(const sycl_reorder_conf_t &conf, ::sycl::handler &cgh,
             const exec_ctx_t &ctx)
         : conf_(conf)
-        , src_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_SRC_0))
-        , dst_(CTX_INOUT_SYCL_KERNEL_MEMORY(DNNL_ARG_DST))
+        , src_(CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_FROM))
+        , dst_(CTX_INOUT_SYCL_KERNEL_MEMORY(DNNL_ARG_TO))
         , src_scale_(CTX_IN_SYCL_KERNEL_MEMORY(
-                  DNNL_ARG_ATTR_SCALES | DNNL_ARG_SRC_0))
-        , dst_scale_(CTX_IN_SYCL_KERNEL_MEMORY(
-                  DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST))
-        , scales_src_dt_(conf_.do_scale_src
-                          ? ctx.memory_mdw(
-                                       DNNL_ARG_ATTR_SCALES | DNNL_ARG_SRC_0)
-                                    .data_type()
-                          : data_type_t::dnnl_f32)
-        , scales_dst_dt_(conf_.do_scale_dst
-                          ? ctx.memory_mdw(DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST)
-                                    .data_type()
-                          : data_type_t::dnnl_f32) {}
+                  DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM))
+        , dst_scale_(
+                  CTX_IN_SYCL_KERNEL_MEMORY(DNNL_ARG_ATTR_SCALES | DNNL_ARG_TO))
+        , src_zero_points(CTX_IN_SYCL_KERNEL_MEMORY(
+                  DNNL_ARG_ATTR_ZERO_POINTS | DNNL_ARG_FROM))
+        , dst_zero_points(CTX_IN_SYCL_KERNEL_MEMORY(
+                  DNNL_ARG_ATTR_ZERO_POINTS | DNNL_ARG_TO))
+        , scales_src_dt_(ctx.memory_mdw(DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM)
+                                 .data_type())
+        , scales_dst_dt_(ctx.memory_mdw(DNNL_ARG_ATTR_SCALES | DNNL_ARG_TO)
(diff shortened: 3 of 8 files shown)


AD2605:
make test
disable benchdnn_all
disable build_vendor_intel
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_ampere
enable benchdnn_reorder

src/gpu/generic/sycl/ref_reorder.cpp
```diff
@@ -27,15 +27,25 @@ namespace sycl {
 status_t ref_reorder_t::pd_t::init_conf() {
     conf_ = sycl_reorder_conf_t();
 
-    conf_.src_md = xpu::sycl::md_t(src_md(0));
-    conf_.dst_md = xpu::sycl::md_t(dst_md());
+    conf_.src_md = xpu::sycl::md_t(arg_md(DNNL_ARG_FROM));
+    conf_.dst_md = xpu::sycl::md_t(arg_md(DNNL_ARG_TO));
 
-    conf_.wk_size = memory_desc_wrapper(src_md(0)).nelems();
+    //get padded number of elements from source and destination
```
dzarukin, line 33: I can see destination, but I can't see source. Is it a typo?

src/gpu/generic/sycl/ref_reorder.cpp
```diff
@@ -27,15 +27,25 @@ namespace sycl {
...
+    auto dst_nelems = memory_desc_wrapper(arg_md(DNNL_ARG_TO)).nelems(true);
 
-    conf_.do_scale_src = !attr()->scales_.has_default_values(DNNL_ARG_SRC_0);
-    conf_.scale_src_mask = attr()->scales_.get_mask(DNNL_ARG_SRC_0);
-    conf_.do_scale_dst = !attr()->scales_.has_default_values(DNNL_ARG_DST);
-    conf_.scale_dst_mask = attr()->scales_.get_mask(DNNL_ARG_DST);
+    // To cover cases when src has more padding than destination, in that case
+    // simply setting the range as dst_nelems does not cover all the source elements
```
dzarukin, line 37: I'm not sure I'm following the comment. Could you re-phrase, please?

src/gpu/generic/sycl/reorder_kernels.hpp
```diff
@@ -37,106 +37,129 @@ struct reorder_kernel_t {
...
+        int ndims = conf_.src_md.ndims();
+
+        dims_t dst_padded_index;
+        float src_scale = conf_.do_scale_src ? src_scales_tensor.load(0) : 1;
+        float dst_scale = conf_.do_scale_dst ? dst_scales_tensor.load(0) : 1;
+        float src_zp = conf_.apply_src_zp ? src_zp_tensor.load(0) : 0;
+        float dst_zp = conf_.apply_dst_zp ? dst_zp_tensor.load(0) : 0;
+
+        for (std::size_t flattened_index = item.get_global_id(0);
+                flattened_index < conf_.num_elements;
```
dzarukin, line 91: Is `item` always nicely divided with this number? No need to handle tails somehow?

tests/benchdnn/reorder/reorder.cpp
```diff
@@ -354,6 +354,26 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
             res->reason = skip_reason::case_not_supported;
             return;
         }
+
+        // Skip blocked format tags and 4 bit formats for Nvidia/AMD/Generic SYCL backends
+        const auto &engine = get_test_engine();
+
+        if (is_generic_gpu(engine) || is_nvidia_gpu(engine)
+                || is_amd_gpu(engine)) {
```
dzarukin, line 362: NVidia and AMD backends are skipped by default.
```suggestion
        if (is_generic_gpu()) {
```

tests/benchdnn/reorder/reorder.cpp
```diff
@@ -354,6 +354,26 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
+            const bool is_4bit_format = is_nibble_sized_type(prb->sdt)
+                    || is_nibble_sized_type(prb->ddt);
+
+            // sycl reorder implementation does not support grouped zero points / scales
+            bool zero_point_has_groups = !prb->attr.scales.is_def();
+            bool scales_has_groups = !prb->attr.zero_points.is_def();
+
+            if (is_blocked_format(prb->stag) || is_blocked_format(prb->dtag)
```
dzarukin, line 370: Please convert `is_blocked_format` into lambda since that's the only piece of code relying on the format. Though I'm sharing the passion they can't support blocked layouts. There are functions that return physical offset by logical one and it should do the trick.

tests/benchdnn/reorder/reorder.cpp
```diff
@@ -354,6 +354,26 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
             return;
         }
+
+        // Skip blocked format tags and 4 bit formats for Nvidia/AMD/Generic SYCL backends
+        const auto &engine = get_test_engine();
+
+        if (is_generic_gpu(engine) || is_nvidia_gpu(engine)
+                || is_amd_gpu(engine)) {
+            const bool is_4bit_format = is_nibble_sized_type(prb->sdt)
+                    || is_nibble_sized_type(prb->ddt);
```
dzarukin, line 364: Please rename into `is_subbyte_type` and move to utils/numerics.{c,h}pp files to unify with other numeric functions.

tests/benchdnn/reorder/reorder.cpp
AD2605: a separate declaration is intentional here, as we will be adding similar skips for the other primitives as well. 

> Though I'm sharing the passion they can't support blocked layouts. There are functions that return physical offset by logical one and it should do the trick.

I agree, but I believe the decision to not support blocked formats is driven by the fact that blocked formats are used for performance.  Since the generic SYCL backend does not aim to be performant, it need not support it.

src/gpu/generic/sycl/ref_reorder.cpp
AD2605: I'll remove this comment, these are reminiscent of  supporting blocked format in reorder, and does not hold it's meaning any more

src/gpu/generic/sycl/ref_reorder.cpp
AD2605: This comment is to be removed, thanks for pointing it out

src/gpu/generic/sycl/reorder_kernels.hpp
AD2605: This is essentially a grid strided loop, the total `global_range` may be lesser than the number of elements, but it strides with a `global_range` index,  covering all the values. Thus it automatically handles all sizes, as each workitem is only responsible for one element at a time, and does not read a vector

src/gpu/generic/sycl/ref_reorder.hpp
```diff
@@ -98,14 +98,23 @@ struct ref_reorder_t : public gpu::generic::sycl::primitive_t {
         }
 
         bool scales_ok() const {
-            const std::vector<int> supported_args
-                    = {DNNL_ARG_SRC, DNNL_ARG_DST};
+            if (!attr()->scales_.has_default_values(
+                        DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM)) {
```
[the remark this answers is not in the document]
AD2605: I maybe wrong, but the API seems to suggest using `DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM`,
https://uxlfoundation.github.io/oneDNN/dev_guide_reorder.html

src/gpu/generic/sycl/ref_reorder.hpp
```diff
@@ -98,14 +98,23 @@ struct ref_reorder_t : public gpu::generic::sycl::primitive_t {
...
+                // Check if scales are applied and then check the data type
+                // otherwise either data type can come up as data_type::undef
+                // returning false unecessarily
+                const auto src_scale_dt = attr()->scales_.get_data_type(
+                        DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM);
+                if (!is_supported_type(src_scale_dt)) { return false; }
```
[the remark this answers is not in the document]
AD2605: The verbose is handled above,  this function call is wrapped in the `VDISPTACH_REORDER` macro,
see- https://github.com/uxlfoundation/oneDNN/blob/1ae1905345983d74c302ac12f8d9f9ada7022ab5/src/gpu/generic/sycl/ref_reorder.hpp#L63C1-L64C51

This change just fixes a bug in that function

tests/benchdnn/reorder/reorder.cpp
dzarukin: > we will be adding similar skips for the other primitives as well

Could you, please, share more details on what primitives and what kind of checks you are anticipating? Blocked layout are supposed to be useful for weights only and I'm almost sure they are utilized through `any` only, and not through direct blocked tag.

src/gpu/generic/sycl/ref_reorder.hpp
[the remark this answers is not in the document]
dzarukin: That's for execution arguments, but not for scales/zero-points. Those know what they are working with.

src/gpu/generic/sycl/ref_reorder.hpp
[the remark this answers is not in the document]
dzarukin: It's generally better to lower it from upper level to lower to the exact place of failing instead of figuring out which condition triggered it, especially, if it's not one.

tests/benchdnn/reorder/reorder.cpp
dzarukin: I get this part. Reorder working with explicit blocked tags is normal as some functionality must move plain data to blocked. I'm missing which **other** primitives will get same kind of check and why, my question was targeting that clarification.

AD2605: >  I'm missing which other primitives will get same kind of check and why, my question was targeting that clarification.

I was expecting other primitives to test with explicit blocked formats in the `test_prim` scripts, especially convolutions and matmuls, but it seems like there are none, except reorder. For now, I can make this a lambda, and add it back in case I come across a test which explicitly mentions blocked formats

tests/benchdnn/reorder/reorder.cpp
AD2605: Done

tests/benchdnn/reorder/reorder.cpp
AD2605: Made it into a lambda

src/gpu/generic/sycl/reorder_kernels.hpp
```diff
@@ -37,106 +37,129 @@ struct reorder_kernel_t {
...
+        , src_zero_points(CTX_IN_SYCL_KERNEL_MEMORY(
+                  DNNL_ARG_ATTR_ZERO_POINTS | DNNL_ARG_SRC))
+        , dst_zero_points(CTX_IN_SYCL_KERNEL_MEMORY(
+                  DNNL_ARG_ATTR_ZERO_POINTS | DNNL_ARG_DST))
         , scales_src_dt_(conf_.do_scale_src
-                          ? ctx.memory_mdw(
-                                       DNNL_ARG_ATTR_SCALES | DNNL_ARG_SRC_0)
+                          ? ctx.memory_mdw(DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM)
                                     .data_type()
                           : data_type_t::dnnl_f32)
```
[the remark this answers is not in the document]
AD2605: removed the conditional

AD2605:
make test
disable build_vendor_intel
disable build_vendor_intel
disable test_device_cpu
enable build_vendor_generic
enable arch_gpu_ampere
enable compiler_icx-oss

ShanoToni (approved):

src/gpu/generic/sycl/ref_reorder.hpp
```diff
@@ -98,14 +98,22 @@ struct ref_reorder_t : public gpu::generic::sycl::primitive_t {
...
 
         bool scales_ok() const {
-            const std::vector<int> supported_args
-                    = {DNNL_ARG_SRC, DNNL_ARG_DST};
+            if (!attr()->scales_.has_default_values(DNNL_ARG_FROM)) {
+                // Check if scales are applied and then check the data type
+                // otherwise either data type can come up as data_type::undef
+                // returning false unecessarily
+                const auto src_scale_dt = attr()->scales_.get_data_type(
+                        DNNL_ARG_ATTR_SCALES | DNNL_ARG_FROM);
```
dzarukin, line 106: I still insist current line return `data_type::undef` unconditionally and passes it into `is_supported_type`. Depending on the handling of `undef`, it either always `false` or `true`. Which one is it?

```suggestion
                        DNNL_ARG_FROM);
```

src/gpu/generic/sycl/sycl_io_helper.hpp
```diff
@@ -238,8 +238,8 @@ struct memory_tensor_t {
         store_float_vec<width>(md_.data_type(), val, mem_.get_pointer(), idx);
     }
 
-    inline float load_md(dims_t offsets) const {
-        return load(md_.off_v(offsets));
+    inline float load_md(dims_t offsets, bool padded = false) const {
```
dzarukin, line 241: Minor: I'd align the name with the one used in `off_v`. Feel free to resolve.

src/gpu/generic/sycl/sycl_io_helper.hpp
```diff
@@ -250,8 +250,19 @@ struct memory_tensor_t {
         return load(md_.off_v(offsets_masked));
     }
 
-    inline void store_md(float val, dims_t offsets) {
-        store(val, md_.off_v(offsets));
+    inline void store_md(float val, dims_t offsets, bool padded = false) {
+        store(val, md_.off_v(offsets, padded));
+    }
+
+    // flattened_index must be lesser than nelems
```
dzarukin, line 257: Nit: maybe `less`?

src/gpu/generic/sycl/sycl_primitive_conf.hpp
```diff
@@ -198,16 +198,19 @@ struct sycl_shuffle_conf_t {
...
     xpu::sycl::md_t src_md;
     xpu::sycl::md_t dst_md;
-    xpu::sycl::md_t scales;
 
     bool do_scale_src;
     int scale_src_mask;
     bool do_scale_dst;
     int scale_dst_mask;
+    bool apply_src_zp;
+    int src_zp_mask;
```
dzarukin, line 207: I don't see a good reason to have completely different names for things that basically do pretty similar thing:
`do_scale_src` versus `apply_src_zp` and `scale_src_mask` versus `src_zp_mask`. Please consider aligning.

AD2605: Fair, Changed all of  `apply_*` and `src_/dst_*`, making all of that consistent

src/gpu/generic/sycl/sycl_io_helper.hpp
AD2605: Changed the variable name from `offsets` to `pos`

AD2605:
@uxlfoundation/onednn-arch gentle ping

dzarukin (approved):

PR #2885, opened 2025-03-14
rfcs: graph api: support bottom-right implicit causal mask in SDPA

ElaineBao:
# Description

The RFC proposes adding support for bottom-right implicit causal mask in oneDNN Graph.

Rendered version: [link](https://github.com/uxlfoundation/oneDNN/blob/yixin/rfc/graph-api-bottom-right-causal-mask/rfcs/20250314-graph-api-bottom-right-causal-mask/README.md)


rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+A key component of SDPA is the causal mask, which restricts the attention mechanism
+appropriately. Currently, oneDNN Graph supports two types of causal masks:
+
+1. Explicit causal mask: Users explicitly construct a causal mask matrix using
+   operations such as `Add` and `Select`.
+
+   ![oneDNN Graph SDPA mask with Add](images/onednn_graph_sdpa_mask1.png) ![oneDNN Graph SDPA mask with Select](images/onednn_graph_sdpa_mask2.png)
+
+2. Implicit causal mask: oneDNN Graph supports top-left implicit casual mask
```
rongzha1, line 23:
```suggestion
2. Implicit causal mask: oneDNN Graph supports top-left implicit causal mask
```

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,294 @@
...
+Scaled Dot-Product Attention (SDPA), introduced in [1], is the core operation in
+Transformer blocks and serves as the backbone for numerous language and generative
+models (e.g., BERT, Stable Diffusion, GPT, DeepSeek).
+
+oneDNN supports SDPA and its optimization through Graph API by defining an SDPA
+graph, partitioning the graph, and optimizing the underlying kernels. The
+floating-point SDPA — supporting f32, bf16, or f16 — is defined as illustrated below:
+
+![oneDNN Graph SDPA pattern](images/onednn_graph_sdpa.png)
+
```
rongzha1, line 14: Does brown module mean optional ? If there are two colors of modules in the flowchart, better to add annotations explaining what each color represents. Same comments for the other pictures.

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,294 @@
...
+  (an ABI-breaking change that must wait for the oneDNN v4.0 release).
+
+### Option 3: Pass s_kv and s_q by a New Shape Operation
+
+Another option to avoid allocating dedicated GPU tensors is to introduce a
+new Shape operation to extract s_kv and s_q from previous tensor's shape.
+The `Shape` operation generates a 1D shape tensor along specified axes of an
+input tensor. With this operation, s_kv and s_q can be extracted from the last
+two axes of scaled output tensor's shape.
+
```
rongzha1, line 235: could you add attribute to `shape` in this picture to show how it works?

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,294 @@
...
+![causal mask subgraph bottom_right with Shape op](images/causal_mask_subgraph_bottom_right_with_shape.png)
+
+#### Shape Operation Specification
+
+##### Attributes
+
+| Attribute Name | Description | Value Type | Supported Values | Required or Optional |
+| :------------- | :---------- | :--------- | :--------------- | :------------------- |
+| [axes] (@ref dnnl::graph::op::attr::axes) | Specifies the axes for generating shape values. | s64 | A s64 list containing the element in the range of [-N, N-1], negative value means counting from the last axis | Required |
```
rongzha1, line 244: Does the output order of shape relate to the order of axes？ does it support shape.size() larger than original input ndims? does it support scalar shape?

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
ElaineBao: Updated the picture, thanks.

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,313 @@
...
+
+- No need to create GPU tensors for s_kv and s_q scalars,
+  and other scalars like attn_scale, -inf.
+- Consistent subgraph design with the top-left implicit causal mask.
+- Align with cuDNN Graph's usage.
+
+#### Cons
+
+- Requires extending the Graph API to support scalar logical tensors
+  (an ABI-breaking change that must wait for the oneDNN v4.0 release).
```
mgouicem, line 246: Is it really ABI breaking? If we really want to support host side scalar at execution time, maybe a naive proposition could be:
- extend logical_tensor properties with host_scalar
- during execution, the compiled graph would take a tensor object with cpu engine.

Could the above be an option?

mgouicem (commented):
Maybe another would be to add an offset attribute to GenIndex?

Regarding host-side logical_tensor, I guess we should differentiate two cases:
- host-side scalars that are constant for the lifetime of the graph. For those, attributes should do the trick, as those scalars are often part of the op/sub-graph definition.
- host-side scalars that are actual execution parameter and can change from execution to execution of the same subgraph. For those cases, couldn't we support that through logical_tensor properties?

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
ElaineBao: This could be an option, but it requires mixed engines in one compiled_partition (i.e. Q/K/V tensors are passed with gpu engine, scalar is passed with cpu engine). We had some preliminary discussion on the meeting that we don't want to support this, but maybe we need to revisit this option?

ElaineBao:
> Maybe another would be to add an offset attribute to GenIndex?
> 
> Regarding host-side logical_tensor, I guess we should differentiate two cases:
> 
> * host-side scalars that are constant for the lifetime of the graph. For those, attributes should do the trick, as those scalars are often part of the op/sub-graph definition.
> * host-side scalars that are actual execution parameter and can change from execution to execution of the same subgraph. For those cases, couldn't we support that through logical_tensor properties?

Thank for clarifying the two cases. For the sequence length of q and kv, they are not the first case, the sequence length could change from run to run, They are the second case.

For logical_tensor properties, as I have replied [here](https://github.com/uxlfoundation/oneDNN/pull/2885#discussion_r1996446617), it's ok to extend the properties, but the difficult part is whether we want to support mixed engines of tensors at execution time.

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
mgouicem: Could you clarify the arguments against mixed engines for arguments?
API-wise, this is something we already support with reorder scenario, and here it would not limited to logical tensors with that host_scalar property.

ElaineBao: Summarize some offline discussion here:

Yixin: 
From Graph API perspective, when users call graph creation API, only one engine kind is passed in. 
When users call graph compilation API, only one engine kind is passed in. 
It seems we don't have a place to pass both cpu and gpu engine before execution. 
And during execution, if tensors with difference engine kind are passed in, firstly it has some disalignment with graph creation and compilation, secondly the library eventually needs to create gpu tensor, which means this doesn't help with the performance.

Mourad:
> For the 2nd question, when we create graph, users pass only one engine kind. when we compile graph, users pass only one engine kind. It seems we don't have a place to pass both cpu and gpu engine before execution. 

That is fine I think. at that point, only logical_tensors are passed, and the new logical_tensor property can be taken into account:

> And during execution, if tensors with difference engine kind are passed in, firstly it has some disalignment with graph creation and compilation, 

I don't think it is disalignment since a) logical tensors have no engine and b) logical_tensor property would express host-side scalars.

> secondly the library eventually needs to create gpu tensor, which means this doesn't help with the performance.

not sure I understand this part. here the logical_tensor would be created as host-side scalar, the actual tensor would be created with cpu engine, and inside library, we would pass that host-side scalar directly as kernel parameter, not copy it explicitely to device memory.

Yixin:
at execution time, if we don't have this constraint, then all the tensors can be passed with cpu engine, so we need to check if it's a scalar or not, if it's a scalar, we can pass it as kernel parameter, if it's not, then the tensor needs to be reordered.. Is it a lot of effort?

ElaineBao: Thanks, @mgouicem! I update the RFC to add a new option based on your suggestion. 👍

A few more points I’d like to discuss with you:
1. Once the `host_scalar` property is added to a logical tensor, it will be available for both CPU and GPU engines. What does this property mean for CPU, especially for SYCL CPU?
2. Users can pass a logical tensor with the `host_scalar` property to any partition, but not all partitions will support it. Initially, only the SDPA partition may handle this property. Other partitions should check the property of each logical tensor and return a compilation failure if a scalar is passed.
3. Since GPU computation is asynchronous during execution, I'm not sure if users can retain the scalar until execution completes. We may need to store the scalar when users initiate execution.
4. If a logical tensor has `host_scalar` property, it cannot have `constant` property anymore, which means the scalar cannot be defined as a constant scalar.

mgouicem: Thanks for updating the RFC!
1. The meaning of host_scalar is the same for CPU or GPU device. Note that in SYCL, host is different than CPU device and they don't necessarily share the same memory space. 
2. correct, we would have to document when that is supported or not.
3. The user should not have to retain the scalar. It will be copied as part of the kernel submission.
4. that is correct. Though I thought the premise of this discussion was that the scalar was not constant? :)

ElaineBao: > that is correct. Though I thought the premise of this discussion was that the scalar was not constant? :)

For sequence length case, yes, but I guess for -inf, it's always a constant...

LuFinch (approved):
LGTM

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
```diff
@@ -0,0 +1,382 @@
...
+    } layout;
+} dnnl_graph_logical_tensor_t;
+```
+
+Currently supported dnnl_graph_tensor_property_t values:
+- `dnnl_graph_tensor_property_undef`: Undefined tensor property.
+- `dnnl_graph_tensor_property_variable`: The tensor may change during computation
+  or between iterations.
+- `dnnl_graph_tensor_property_constant`: The tensor remains unchanged during computation and across
+  iterations, allowing optimization and caching (e.g., constant weight tensors in inference scenarios).
```
TaoLv, line 280: I think this part can be safely removed. But I suggest to add more information and usage about the newly added `host_scalar` property here. I think we had quite a few discussions about setting the property under different runtime combinations. Better to have them documented here.

TaoLv (approved):

rfcs/20250314-graph-api-bottom-right-causal-mask/README.md
ElaineBao: Thanks for the suggestion, updated.

dzarukin (approved):

PR #2893, opened 2025-03-14
rfcs: ukernel API updates to support exception-free mode and rt lds

dzarukin:
[Link to a rendered document](https://github.com/uxlfoundation/oneDNN/tree/dzarukin/rfcs/ukernel_api_updates/rfcs/20250312-ukernel-api-updates)

### DIFF
--- a/rfcs/20250312-ukernel-api-updates/README.md
+++ b/rfcs/20250312-ukernel-api-updates/README.md
@@ -1,0 +1,289 @@
+# Ukernel API Updates
+
+## Motivation
+
+ukernel API got some traction in the community. During the integration some pain
+points emerged, and this document will discuss some of them and propose.
+
+## Proposal
+
+### Exception-free Mode
+
+The implementation of ukernel API has its limitations, mostly related to
+hardware features available. Some underlying C API calls may return unsuccessful
+status, but C++ API calls will not propagate that and just throw an exception.
+
+There's no other way to recover from the failure but to catch exceptions which
+doesn't suit every user's environment. To overcome that limitation, the
+following is proposed:
+
+* `enum class pack_type` gets an additional special value `no_pack`. With that,
+  `undef` value starts indicating that the kernel is unusable on the current
+  system and a fallback to a different frontend is required. For example,
+  querying `pack_type` on SSE41 systems will start returning this value.
+  `no_pack` means that brgemm is supported but packing for a given tensor, A or
+  B, is not required, and plain memory format for requested tensor configured
+  through the brgemm object constructor promises to be functionally correct.
+  This replaces the former `no_trans` value that was used to denote the same
+  message. The difference is `no_trans` still has physical representation for
+  the tensor and may confuse the user in case their tensor is originally in
+  `trans` `pack_type`, or transposed memory format. The proposed value doesn't
+  have that physical meaning and serves for differentiating from the unsupported
+  state and is uniform across multiple backends which may have different packing
+  requirements.
+* `brgemm` object provides several APIs to configure it and fills internal
+  object state with user choices. Then a `finalize()` call must be used to
+  initiate the backend object preparation. During that preparation, underlying
+  calls may return unsuccessful statuses and the C++ `finalize()` will throw an
+  exception. Instead, the proposal is to start returning a `bool` value with
+  `true` meaning success and `false` meaning no success instead of current
+  `void`. In the latter case the fallback to a different frontend must be done
+  as the brgemm object indicates it doesn't support given settings on a given
+  system.
+
+These are correspondent changes in the API and usage:
+```cpp
+/// dnnl_ukernel_types.h
+
+/// Packing specification
+typedef enum {
+    /// Undefined pack type. (A guard value.) <-- No longer true.
+    /// When returned, ukernel API can't be used.
+    dnnl_pack_type_undef = 0,
+    /// NEW TYPE.
+    /// A value indicating the brgemm object can proceed without packing data
+    /// for a requested tensor.
+    dnnl_pack_type_no_pack,
+    /// Plain, not transposed layout. Similar to format_tag::ab.
+    dnnl_pack_type_no_trans,
+    /// Plain, transposed layout. Similar to format_tag::ba.
+    dnnl_pack_type_trans,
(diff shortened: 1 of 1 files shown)


dzarukin:
@Radu2k @mgouicem

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,289 @@
...
+doesn't suit every user's environment. To overcome that limitation, the
+following is proposed:
+
+* `enum class pack_type` gets an additional special value `no_pack`. With that,
+  `undef` value starts indicating that the kernel is unusable on the current
+  system and a fallback to a different frontend is required. For example,
+  querying `pack_type` on SSE41 systems will start returning this value.
+  `no_pack` means that brgemm is supported but packing for a given tensor, A or
+  B, is not required, and plain memory format for requested tensor configured
+  through the brgemm object constructor promises to be functionally correct.
```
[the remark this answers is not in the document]
dzarukin: Fair point, I removed a new type proposal since the static call doesn't let query values from a configured brgemm object anyway. Thanks.

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,262 @@
...
+    /// Executes a BRGeMM ukernel object.
+    ///
+    /// @param A Base pointer to a tensor A.
+    /// @param B Base pointer to a tensor B.
+    /// @param A_B_offsets Vector of pairs of tensors A and B offsets for
+    ///     each batch. The number of batches must coincide with the
+    ///     `batch_size` value passed at object construction stage.
+    /// @param C Pointer to a tensor C (accumulation buffer).
+    /// @param scratchpad Pointer to a scratchpad buffer.
+    /// @param actual_lds NEW ARGUMENT. Vector of actual leading dimensions for
```
mgouicem, line 149: (nit) why not just `lds`?

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,262 @@
...
+    ///     tensors A, B, and C. Must be specified if brgemm object was created
+    ///     with `DNNL_RUNTIME_DIM_VAL` instead of regular values for `lda`,
+    ///     `ldb`, or `ldc`. When `get_B_pack_type` returns values other than
+    ///     `no_trans`, `actual_lds[1]` for `ldb` must be used in transform
+    ///     routine. If any of given values were not specified as runtime values
+    ///     at kernel creation, they will be used instead.
+    void execute(const void *A, const void *B,
+            const std::vector<std::pair<memory::dim, memory::dim>> &A_B_offsets,
+            void *C, void *scratchpad,
+            const std::vector<memory::dim> &actual_lds = {}) const {
```
mgouicem, line 159: Given that:
- there could be potential overheads due to heap allocation in std::vector
- and that we need only `lda`, `ldb`, `ldc`, `ldd`.

Would it make sense to actually take `lda`, `ldb`, `ldc`, `ldd` separately ? Or use `const std::array<memory::dim, 4> &lds` ?

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,262 @@
...
+            printf("API can't be used\n");
+            fallback_call();
+        }
+    }
+
+    ...
+
+    // Note that for `need_pack == true` the `ldb` value is fixed.
+    dnnl::ukernel::brgemm brg(M, N, K, batch, DNNL_RUNTIME_DIM_VAL,
+            need_pack ? transform::out_ld : DNNL_RUNTIME_DIM_VAL,
```
mgouicem, line 236: what is `transform::out_ld`?

mgouicem (commented):
LGTM, thanks.
Just one open question on the usage of std::vector for a fixed size array.

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+```cpp
+    ...
+
+    using namespace dnnl::ukernel;
+
+    const auto pack_type = brgemm::get_B_pack_type(a_dt, b_dt);
+    const bool need_pack = pack_type != pack_type::no_trans;
+
+    if (need_pack) {
```
jondea, line 219: You haven't checked for `pack_type::undef`, although I'm guessing this will be handled by `transform` returning a null pointer. If that is the case, it might make sense to make this a documented and expected feature when implementing transforms.

rfcs/20250312-ukernel-api-updates/README.md
jondea: Or, at the risk of being overly verbose `execute_lds` to make it clear how they differ and when they are necessary.

rfcs/20250312-ukernel-api-updates/README.md
jondea: It would also avoid potential out of bounds accesses and make their internal use easier.

rfcs/20250312-ukernel-api-updates/README.md
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    // Configurating `brg` object...
+
+    // This is possible with proposed API change.
+    const bool ok = brg.finalize();
+    if (!ok) {
+        printf("API can't be used\n");
+        fallback_call();
+    }
```
jondea, line 111: This is a lot of checks, which I have no doubt will in time be checked in subtly different ways by each consumer. In your example later, you don't explicitly check the `pack_type`, and instead check whether transform returns a null pointer. This makes me wonder whether we need to explicitly overload the meaning of  `dnnl::ukernel::pack_type::undef` from the `get_B_pack_type` functions at all.

In short, telling the user to always check whether the transform is a null pointer rolls up all the error cases into one, and gives us more future freedom to change the API.

jondea:
Broadly speaking this looks good to me. My comments are all minor optional suggestions.

I am a little sceptical about the utility of runtime dimensions generally though. Given that brgemm is a very lightweight API, it does feel like it would be more natural for the caller to handle changing dimensions by regenerating the brgemm object.

PR #2894, opened 2025-03-17
graph: fix the intermediate data types in SDPA patterns

TaoLv:
Address MFDNN-13091

The purpose is to align the intermediate data types in the pattern with those in ATen SDPA and GPU ukernel SDPA.

1. matmul/softmax/binary ops are extended to support mixed data types for inputs and outputs.
2. The two SDPA examples are changed to use f32 intermediate data type, rather than bf16 or f16.
3. The backend is changed to dispatch to GPU ukernel SDPA only when f32 intermediate data type is required.
4. Add test cases for f16/bf16 SDPA with f32 intermediate data type.

With these, now a typical f16 SDPA which can be dispatched to GPU ukernel SDPA looks like below:

image

### DIFF
--- a/doc/graph/fusion_patterns/sdpa.md
+++ b/doc/graph/fusion_patterns/sdpa.md
@@ -92,8 +92,11 @@ optional.
 
 oneDNN supports the floating-point SDPA pattern with data types f32, bf16, and
 f16. You can specify the data type via the input and output logical tensors'
-data type fields for each operation. oneDNN does not support mixing different
-floating data types in a floating-point SDPA pattern.
+data type fields for each operation.
+
+oneDNN supports bf16 or f16 SDPA with f32 intermediate type, which means the
+Q/K/V tensors have bf16 or f16 data type while the output of the first MatMul,
+Scale, Mask, and the input of SoftMax are in f32 data type.
 
 oneDNN supports the quantized SDPA pattern with int8-f32 mixed precision,
 int8-bf16 mixed precision, and int8-f16 mixed precision data types.
@@ -128,9 +131,9 @@ platforms follow the general description in @ref dev_guide_data_types.
 4. GPU
    - Optimized implementation is available for 4D Q/K/V tensors with shape
      defined as (N, H, S, D).
-   - Optimized implementation is available for floating-point SDPA with `f16`
-     data type and `D <= 256` on Intel Graphics Products with Intel(R) Xe Matrix
-     Extensions (Intel(R) XMX) support.
+   - Optimized implementation is available for `f16` or `bf16` SDPA with `f32`
+     intermediate data type and `D <= 256` on Intel Graphics Products with
+     Intel(R) Xe Matrix Extensions (Intel(R) XMX) support.
 
 ## Example
 
--- a/doc/graph/programming_model/low_precision.md
+++ b/doc/graph/programming_model/low_precision.md
@@ -52,7 +52,6 @@ Graph operations support bf16 and f16 data types.
 
 A TypeCast operation performing down conversion should be inserted clearly to
 indicate the use of low numeric precision. oneDNN Graph implementation fully
-honors the API-specified numeric precision and only performs the computation
-using the API-specified or higher numeric precision.
+honors the API-specified numeric precision.
 
 @img{bf16_programming.jpg,Figure 2: Overview of bf16 programming model.,80%,}
--- a/src/graph/interface/op_def.hpp
+++ b/src/graph/interface/op_def.hpp
@@ -1032,12 +1032,15 @@ DNNL_GRAPH_OP_SCHEMA(SoftMax, 1,
         op_schema_t()
                 .set_num_inputs(1)
                 .set_num_outputs(1)
-                .set_input(0, "src", "T")
-                .set_output(0, "dst", "T")
+                .set_input(0, "src", "T1")
+                .set_output(0, "dst", "T2")
                 .set_attr(op_attr::axis, false, attribute_kind::i, (int64_t)1)
                 .set_type_constraints(
-                        "T", {data_type::f32, data_type::bf16, data_type::f16})
-                .set_shape_inference_function(infer_identity_output_shape))
+                        "T1", {data_type::f32, data_type::bf16, data_type::f16})
+                .set_type_constraints(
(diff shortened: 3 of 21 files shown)


tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
```diff
@@ -15,6 +15,10 @@
 --reset --dt=f32,bf16,f16 --case=complex_fusion/mha/sdpa-plain-wo-mask-f16.json
 --reset --dt=f32,bf16,f16 --case=complex_fusion/mha/sdpa-plain-scale-by-mul-f16.json
 --reset --dt=f32,bf16,f16 --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
+# f16 inputs + f32 intermediates + f16 outputs
+--reset --case=complex_fusion/mha/sdpa-plain-simplified-f16-f32.json
```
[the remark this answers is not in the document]
TaoLv: Yes, we can. I provide a separate file here as I think this one will be more frequently used, tested, and modified. It will be easier for us to manipulate it or share it with others with this new file.
The legacy pure f16 case `sdpa-plain-simplified-f16.json` is also kept for testing.

doc/graph/programming_model/low_precision.md
```diff
@@ -52,7 +52,6 @@ Graph operations support bf16 and f16 data types.
 
 A TypeCast operation performing down conversion should be inserted clearly to
 indicate the use of low numeric precision. oneDNN Graph implementation fully
-honors the API-specified numeric precision and only performs the computation
-using the API-specified or higher numeric precision.
+honors the API-specified numeric precision.
```
mgouicem, line 55: Just to make sure we are aligned. This still allows to use f32 values to store f16/bf16 data, as long as we respect roundings to f16/bf16 accuracy, right?

src/graph/interface/op_def.hpp
```diff
@@ -1127,13 +1128,17 @@ DNNL_GRAPH_OP_SCHEMA(Subtract, 1,
...
+                .set_output(0, "dst", "T3")
                 .set_attr(op_attr::auto_broadcast, false, attribute_kind::s,
                         "numpy", {"none", "numpy"})
                 .set_type_constraints(
-                        "T", {data_type::f32, data_type::bf16, data_type::f16})
+                        "T1", {data_type::f32, data_type::bf16, data_type::f16})
+                .set_type_constraints(
+                        "T2", {data_type::f32, data_type::bf16, data_type::f16})
+                .set_type_constraints(
+                        "T3", {data_type::f32, data_type::bf16, data_type::f16})
```
mgouicem, line 1141: This requires some documentation about type promotion as users might wonder what happens for example with f16 <- f16 + bf16.

dzarukin (commented):
LGTM. Move approve responsibility to @mgouicem. :)

doc/graph/programming_model/low_precision.md
TaoLv: Yes, in my understanding, it's still allowed for backend implementations. From this perspective, it seems I need to keep the original statement. My intention here was to align the implementations. As the original statement sounds like different backends (eg. DNNL & GC, CPU & GPU) can have different numerical behaviors.

src/graph/interface/op_def.hpp
TaoLv: In fact, we don't allow f16 + bf16. It's mentioned in the "supported data types" section in the op document. When src0 and src1 have different data types, one of them should be f32 and the other one (f16 or bf16) will be promoted to f32 for calculation.

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -128,9 +128,9 @@ platforms follow the general description in @ref dev_guide_data_types.
 4. GPU
    - Optimized implementation is available for 4D Q/K/V tensors with shape
      defined as (N, H, S, D).
-   - Optimized implementation is available for floating-point SDPA with `f16`
-     data type and `D <= 256` on Intel Graphics Products with Intel(R) Xe Matrix
-     Extensions (Intel(R) XMX) support.
+   - Optimized implementation is available for `f16` or `bf16` SDPA with `f32`
+     intermediate data type and `D <= 256` on Intel Graphics Products with
```
ElaineBao, line 132: What does intermediate data type mean when users construct graph? Do we need to describe it in text or picture?

TaoLv: I'm not sure how to make it better. Please feel free to make a suggestion here. Thanks!

ElaineBao: From the picture in the description, the intermediate data type means the data type starting from the output of first MatMul, to the input of Softmax. Maybe we can add an intermediate dtype mark on the picture.
image

And then describe the relationship between input and intermediate dtypes.

wzt1997 (approved):

src/graph/interface/op_def.hpp
```diff
@@ -54,13 +54,17 @@ DNNL_GRAPH_OP_SCHEMA(Add, 1,
...
```
rongzha1, line 67: Since not all combinations of data types are supported, do we need to add a dt check function like `check_matmul_dtype()` ?

src/graph/interface/op_def.hpp
```diff
@@ -1121,13 +1128,17 @@ DNNL_GRAPH_OP_SCHEMA(Subtract, 1,
...
```
rongzha1, line 1141: Same comments as `Add`, do we need to add a dt check function like `check_matmul_dtype()` ?

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao (approved):

xiang1guo (approved):

mgouicem (approved):

ranukund (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2900, opened 2025-03-17
common: postops: add post-ops support for binary select op

avmanerikar:
# Description

The PR adds updates post-ops implementations so that they support binary select operation as a post-op:

$dst[i] = cond[i]\quad?\quad src_0[i] \quad:\quad src_1[i]$

The updates account for the third conditional input that the select op requires for computation.

Implements MFDNN-12883.

This also introduces a new benchdnn format for providing arguments to multiple `src` tensors for a single post-op:
```
--attr-post-ops=select:DT[.S1_MASK_INPUT[.S1_TAG]][:S2_MASK_INPUT[.S2_TAG]]
```
#### Testcase:
```
./benchdnn --matmul --attr-post-ops=select:f32.per_tensor.ab:per_tensor.ab 5x10:10x5
```
(Currently, there is no broadcasting support for the conditional `src2` tensor. The data type, mask policy and tag for the `src2` tensor are fixed as `s8`, `common` and `any` respectively).

### DIFF
--- a/doc/programming_model/attributes_post_ops.md
+++ b/doc/programming_model/attributes_post_ops.md
@@ -281,6 +281,29 @@ Currently the following scenarios are optimized:
   post-operation which format may be queried from attributes using
   `dnnl::post_ops::get_params_binary(...)` function call.
 
+For the binary select operation, an additional conditional tensor is required 
+to execute the operation which is implemented using:
+
+~~~cpp
+void dnnl::post_ops::append_binary(
+        algorithm alg, // binary algorithm to apply
+        const memory::desc &src1 // memory descriptor for a second memory operand
+        const memory::desc &src2 // memory descriptor for a third memory operand
+        );
+~~~
+
+The `alg`, `src1`, and `src2` parameters are the same as in 
+@ref dev_guide_binary for the select operation.
+
+The binary post-op thus becomes:
+
+\f[
+    \dst[:] = \operatorname{binary}(\operatorname{Op}(...), Source\_1[:], Source\_2[:])
+\f]
+
+There is no broadcasting support for the conditional tensor. The select op is only 
+supported for CPU implementations.
+
 @anchor dev_guide_attributes_post_ops_prelu
 ### Prelu Post-op
 
--- a/include/oneapi/dnnl/dnnl.h
+++ b/include/oneapi/dnnl/dnnl.h
@@ -718,7 +718,7 @@ dnnl_status_t DNNL_API dnnl_post_ops_get_params_dw(
 
 /// Appends a binary post-op.
 ///
-/// The kind of this post operation is #dnnl_binary.
+/// This post operation is categorized as #dnnl_binary.
 ///
 /// In the simplest case when the binary is the only post operation, the
 /// computations would be:
@@ -736,6 +736,32 @@ dnnl_status_t DNNL_API dnnl_post_ops_get_params_dw(
 dnnl_status_t DNNL_API dnnl_post_ops_append_binary(dnnl_post_ops_t post_ops,
         dnnl_alg_kind_t alg_kind, const_dnnl_memory_desc_t src1_desc);
 
+/// Appends a binary post-op with ternary operators.
+///
+/// This post operation is categorized as #dnnl_binary.
+///
+/// In the simplest case when the binary is the only post operation, the
+/// computations will be:
+///
+///     dst[:] <- binary_op (dst[:], another_input1[:], another_input2[:])
+///
+/// where binary_op is configured with the given parameters. binary_op supports
+/// broadcast semantics only for the second operand and not for the third
+/// operand.
+///
+/// @param post_ops Post-ops.
+/// @param alg_kind Binary algorithm for the post-op.
+/// @param src1_desc Memory descriptor of a second operand.
+/// @param src2_desc Memory descriptor of a third operand. If the specificed
+/// algorithm is not one that requires a ternary input, src2_desc will be
+/// ignored.
+
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+dnnl_status_t DNNL_API dnnl_post_ops_append_binary_v2(dnnl_post_ops_t post_ops,
+        dnnl_alg_kind_t alg_kind, const_dnnl_memory_desc_t src1_desc,
+        const_dnnl_memory_desc_t src2_desc);
+
 /// Returns the parameters of a binary post-op.
 ///
 /// @param post_ops Post-ops.
@@ -749,6 +775,25 @@ dnnl_status_t DNNL_API dnnl_post_ops_append_binary(dnnl_post_ops_t post_ops,
 dnnl_status_t DNNL_API dnnl_post_ops_get_params_binary(
         const_dnnl_post_ops_t post_ops, int index, dnnl_alg_kind_t *alg_kind,
         const_dnnl_memory_desc_t *src1_desc);
+
+/// Returns the parameters of a binary post-op with ternary operators.
+///
+/// @param post_ops Post-ops.
+/// @param index Index of the binary post-op.
+/// @param alg_kind Output binary algorithm kind.
+/// @param src1_desc Output memory descriptor of a second operand.
+/// @param src2_desc Output memory descriptor of a third operand. If the
+/// specified algorithm is not one that requires a ternary input, src2_desc
+/// will be ignored.
+
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+/// @returns #dnnl_invalid_arguments if @p index does not refer to a binary
+///     post-op.
+dnnl_status_t DNNL_API dnnl_post_ops_get_params_binary_v2(
+        const_dnnl_post_ops_t post_ops, int index, dnnl_alg_kind_t *alg_kind,
+        const_dnnl_memory_desc_t *src1_desc,
+        const_dnnl_memory_desc_t *src2_desc);
 
 /// Appends a prelu forward post-op.
 ///
--- a/include/oneapi/dnnl/dnnl.hpp
+++ b/include/oneapi/dnnl/dnnl.hpp
@@ -3741,10 +3741,10 @@ struct post_ops : public handle<dnnl_post_ops_t> {
 
     /// Appends a binary post-op.
     ///
-    /// The kind of this post operation is #dnnl_binary.
+    /// This post operation is categorized as #dnnl_binary.
     ///
     /// In the simplest case when the binary is the only post operation, the
-    /// computations would be:
+    /// computations will be:
     ///
     ///     dst[:] <- binary_op (dst[:], another_input[:])
     ///
@@ -3757,6 +3757,32 @@ struct post_ops : public handle<dnnl_post_ops_t> {
         error::wrap_c_api(dnnl_post_ops_append_binary(get(),
                                   convert_to_c(aalgorithm), src1_desc.get()),
                 "could not append a binary post-op");
+    }
+
+    /// Appends a binary post-op with ternary operators.
+    ///
+    /// This post operation is categorized as #dnnl_binary.
+    ///
+    /// In the simplest case when this is the only post operation, the
+    /// computations will be:
+    ///
+    ///     dst[:] <- binary_op (dst[:], another_input1[:], another_input2[:])
+    ///
+    /// where binary_op is configured with the given parameters. binary_op
+    /// supports broadcast semantics only for the second operand and not for the
+    /// third operand.
+    ///
+    /// @param aalgorithm Binary algorithm for the post-op.
+    /// @param src1_desc Memory descriptor of the second operand.
+    /// @param src2_desc Memory descriptor of the third operand. If the specified
+    /// algorithm is not one that requires a ternary input, src2_desc will be
+    /// ignored.
+    void append_binary(algorithm aalgorithm, const memory::desc &src1_desc,
+            const memory::desc &src2_desc) {
+        error::wrap_c_api(
+                dnnl_post_ops_append_binary_v2(get(), convert_to_c(aalgorithm),
+                        src1_desc.get(), src2_desc.get()),
+                "could not append a binary post-op with ternary operators");
     }
 
     /// Returns the parameters of a binary post-op.
@@ -3776,6 +3802,33 @@ struct post_ops : public handle<dnnl_post_ops_t> {
         error::wrap_c_api(dnnl_memory_desc_clone(&cloned_md, cdesc),
                 "could not clone a memory descriptor");
         src1_desc = memory::desc(cloned_md);
+    }
+
+    /// Returns the parameters of a binary post-op with ternary operators.
+    ///
+    /// @param index Index of the binary post-op.
+    /// @param aalgorithm Output binary algorithm kind.
+    /// @param src1_desc Output memory descriptor of the second operand.
+    /// @param src2_desc Output memory descriptor of the third operand.
+    void get_params_binary(int index, algorithm &aalgorithm,
+            memory::desc &src1_desc, memory::desc &src2_desc) const {
+        dnnl_alg_kind_t c_alg;
+        const_dnnl_memory_desc_t cdesc1, cdesc2;
+        error::wrap_c_api(dnnl_post_ops_get_params_binary_v2(
+                                  get(), index, &c_alg, &cdesc1, &cdesc2),
+                "could not get parameters of a binary post-op with ternary "
+                "operators");
+        aalgorithm = static_cast<dnnl::algorithm>(c_alg);
+        dnnl_memory_desc_t cloned_md1 = nullptr;
+        dnnl_memory_desc_t cloned_md2 = nullptr;
+
+        error::wrap_c_api(dnnl_memory_desc_clone(&cloned_md1, cdesc1),
+                "could not clone a memory descriptor");
+        src1_desc = memory::desc(cloned_md1);
+
+        error::wrap_c_api(dnnl_memory_desc_clone(&cloned_md2, cdesc2),
+                "could not clone a memory descriptor");
+        src2_desc = memory::desc(cloned_md2);
     }
 
     /// Appends a prelu forward post-op.
--- a/src/common/primitive_attr.cpp
+++ b/src/common/primitive_attr.cpp
@@ -206,51 +206,77 @@ status_t post_ops_t::append_dw(data_type_t wei_dt, data_type_t bias_dt,
     return success;
 }
 
-status_t post_ops_t::validate_binary(
-        alg_kind_t alg, const memory_desc_t *user_src1_desc) const {
+status_t post_ops_t::validate_binary(alg_kind_t alg,
+        const memory_desc_t *user_src1_desc,
+        const memory_desc_t *user_src2_desc) const {
 
     if (len() == post_ops_limit) return out_of_memory;
     using namespace alg_kind;
     bool alg_ok = one_of(alg, binary_add, binary_mul, binary_max, binary_min,
             binary_div, binary_sub, binary_ge, binary_gt, binary_le, binary_lt,
-            binary_eq, binary_ne);
-    if (!alg_ok) return invalid_arguments;
-    if (!memory_desc_sanity_check(*user_src1_desc)) return invalid_arguments;
+            binary_eq, binary_ne, binary_select);
+    bool is_ternary_op = (alg == binary_select);
+
+    VCHECK_ATTR(alg_ok, VERBOSE_BAD_ALGORITHM);
+    VCHECK_ATTR(memory_desc_sanity_check(*user_src1_desc),
+            VERBOSE_MEM_DESC_CHECK_FAIL);
 
     // Additional check to restrict run-time dimension usage until supported.
     for (int d = 0; d < user_src1_desc->ndims; ++d) {
-        if (user_src1_desc->dims[d] == DNNL_RUNTIME_DIM_VAL)
-            return invalid_arguments;
+        VCHECK_ATTR(user_src1_desc->dims[d] != DNNL_RUNTIME_DIM_VAL,
+                VERBOSE_RUNTIMEDIM_UNSUPPORTED);
+    }
+
+    // Additional checks if the algorithm involves ternary inputs
+    if (is_ternary_op) {
+        VCHECK_ATTR(memory_desc_sanity_check(*user_src2_desc),
+                VERBOSE_MEM_DESC_CHECK_FAIL);
+        for (int d = 0; d < user_src2_desc->ndims; ++d) {
+            VCHECK_ATTR(user_src2_desc->dims[d] != DNNL_RUNTIME_DIM_VAL,
+                    VERBOSE_RUNTIMEDIM_UNSUPPORTED);
+        }
     }
 
     return success;
 }
 
-status_t post_ops_t::append_binary(
-        alg_kind_t alg, const memory_desc_t *user_src1_desc) {
-    auto status = validate_binary(alg, user_src1_desc);
-    if (status != success) return status;
+status_t post_ops_t::append_binary(alg_kind_t alg,
+        const memory_desc_t *user_src1_desc,
+        const memory_desc_t *user_src2_desc) {
+    CHECK(validate_binary(alg, user_src1_desc, user_src2_desc));
 
     entry_.emplace_back();
     auto &e = entry_.back();
     e.kind = primitive_kind::binary;
     e.binary.alg = alg;
+
     e.binary.user_src1_desc = *user_src1_desc;
     e.binary.src1_desc = *user_src1_desc;
+
+    if (alg == alg_kind::binary_select) {
+        e.binary.user_src2_desc = *user_src2_desc;
+        e.binary.src2_desc = *user_src2_desc;
+    }
     return success;
 }
 
-status_t post_ops_t::prepend_binary(
-        alg_kind_t alg, const memory_desc_t *user_src1_desc) {
-    auto status = validate_binary(alg, user_src1_desc);
-    if (status != success) return status;
+status_t post_ops_t::prepend_binary(alg_kind_t alg,
+        const memory_desc_t *user_src1_desc,
+        const memory_desc_t *user_src2_desc) {
+    CHECK(validate_binary(alg, user_src1_desc, user_src2_desc));
 
     entry_.emplace(entry_.begin());
     auto &e = entry_[0];
     e.kind = primitive_kind::binary;
     e.binary.alg = alg;
+
     e.binary.user_src1_desc = *user_src1_desc;
     e.binary.src1_desc = *user_src1_desc;
+
+    if (alg == alg_kind::binary_select) {
+        e.binary.user_src2_desc = *user_src2_desc;
+        e.binary.src2_desc = *user_src2_desc;
+    }
     return success;
 }
 
@@ -270,17 +296,32 @@ status_t post_ops_t::set_default_formats(const memory_desc_t *dst_md) {
 
         auto &src1_md = entry_[idx].binary.src1_desc;
         const memory_desc_wrapper src1_mdw(src1_md);
-        if (!src1_mdw.format_any()) continue;
 
         const memory_desc_wrapper dst_mdw(dst_md);
-        assert(!dst_mdw.format_any());
 
         // 1D tensors should be plain abx.
-        if (src1_mdw.count_non_unit_dims(1))
-            CHECK(memory_desc_init_by_strides(src1_md, nullptr));
-        else
-            CHECK(memory_desc_init_by_blocking_desc(
-                    src1_md, dst_mdw.blocking_desc()));
+        if (src1_mdw.format_any()) {
+            assert(!dst_mdw.format_any());
+
+            if (src1_mdw.count_non_unit_dims(1))
+                CHECK(memory_desc_init_by_strides(src1_md, nullptr));
+            else
+                CHECK(memory_desc_init_by_blocking_desc(
+                        src1_md, dst_mdw.blocking_desc()));
+        }
+
+        auto &src2_md = entry_[idx].binary.src2_desc;
+        const memory_desc_wrapper src2_mdw(src2_md);
+
+        if (entry_[idx].is_binary_with_ternary_op() && src2_mdw.format_any()) {
+            assert(!dst_mdw.format_any());
+
+            if (src1_mdw.count_non_unit_dims(1))
+                CHECK(memory_desc_init_by_strides(src2_md, nullptr));
+            else
+                CHECK(memory_desc_init_by_blocking_desc(
+                        src2_md, dst_mdw.blocking_desc()));
+        }
     }
 
     return status::success;
@@ -333,12 +374,22 @@ status_t post_ops_t::entry_t::validate_binary(
     if (!is_binary()) return status::success;
 
     VCHECK_ATTR(dst_md->ndims == binary.user_src1_desc.ndims,
-            VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "dst", "bin_po",
+            VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "dst", "bin_po src1",
             dst_md->ndims, binary.user_src1_desc.ndims);
 
     VCHECK_ATTR(IMPLICATION(engine_kind == dnnl_cpu,
                         binary.user_src1_desc.data_type != data_type::f64),
             VERBOSE_INVALID_DATATYPE, "bin_po src1");
+
+    if (is_binary_with_ternary_op()) {
+        VCHECK_ATTR(dst_md->ndims == binary.user_src2_desc.ndims,
+                VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "dst", "bin_po src2",
+                dst_md->ndims, binary.user_src2_desc.ndims);
+
+        VCHECK_ATTR(IMPLICATION(engine_kind == dnnl_cpu,
+                            binary.user_src2_desc.data_type != data_type::f64),
+                VERBOSE_INVALID_DATATYPE, "bin_po src2");
+    }
 
     return status::success;
 }
@@ -711,6 +762,14 @@ status_t dnnl_post_ops_append_binary(post_ops_t *post_ops, alg_kind_t alg_kind,
     return post_ops->append_binary(alg_kind, user_src1_desc);
 }
 
+status_t dnnl_post_ops_append_binary_v2(post_ops_t *post_ops,
+        alg_kind_t alg_kind, const memory_desc_t *user_src1_desc,
+        const memory_desc_t *user_src2_desc) {
+    if (post_ops == nullptr) return invalid_arguments;
+
+    return post_ops->append_binary(alg_kind, user_src1_desc, user_src2_desc);
+}
+
 status_t dnnl_post_ops_get_params_binary(const post_ops_t *post_ops, int index,
         alg_kind_t *alg_kind, const memory_desc_t **user_src1_desc) {
     CHECK(simple_get_params_check(post_ops, index, primitive_kind::binary));
@@ -718,6 +777,19 @@ status_t dnnl_post_ops_get_params_binary(const post_ops_t *post_ops, int index,
     const auto &b = post_ops->entry_[index].binary;
     if (alg_kind) *alg_kind = b.alg;
     if (user_src1_desc) *user_src1_desc = &b.user_src1_desc;
+
+    return success;
+}
+
+status_t dnnl_post_ops_get_params_binary_v2(const post_ops_t *post_ops,
+        int index, alg_kind_t *alg_kind, const memory_desc_t **user_src1_desc,
+        const memory_desc_t **user_src2_desc) {
+    CHECK(simple_get_params_check(post_ops, index, primitive_kind::binary));
+
+    const auto &b = post_ops->entry_[index].binary;
+    if (alg_kind) *alg_kind = b.alg;
+    if (user_src1_desc) *user_src1_desc = &b.user_src1_desc;
+    if (user_src2_desc) *user_src2_desc = &b.user_src2_desc;
 
     return success;
 }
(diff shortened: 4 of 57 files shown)


src/common/primitive_desc.hpp
```diff
@@ -201,12 +208,18 @@ struct primitive_desc_t : public c_compatible {
...
+                            (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx)
+                                    | DNNL_ARG_SRC_2)))
                     continue;
 
-                return &po.entry_[idx].binary.src1_desc;
+                return (arg
+                               == (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx)
+                                       | DNNL_ARG_SRC_1))
+                        ? &po.entry_[idx].binary.src1_desc
+                        : &po.entry_[idx].binary.src2_desc;
```
dzarukin, line 222: Maybe using a `switch (arg)` would be a little cleaner with more than a single arg handled.

src/cpu/binary_injector_utils.cpp
```diff
@@ -79,6 +84,11 @@ memory_desc_t get_src1_desc(
     }
 }
 
+memory_desc_t get_src2_desc(const post_ops_t::entry_t &post_op) {
+    assert(post_op.is_binary_with_ternary_op());
+    return post_op.binary.src2_desc;
```
dzarukin, line 89:
```suggestion
    if (post_op.is_binary_with_ternary_op()) {
        return post_op.binary.src2_desc;
    } else {
        return memory_desc_t();
    }
```

src/cpu/binary_injector_utils.hpp
```diff
@@ -52,6 +52,8 @@ std::vector<broadcasting_strategy_t> extract_bcast_strategies(
 memory_desc_t get_src1_desc(
         const post_ops_t::entry_t &post_op, const memory_desc_wrapper &dst_d);
 
+memory_desc_t get_src2_desc(const post_ops_t::entry_t &post_op);
```
dzarukin, line 55: Please keep the identical signature even though dst_d wouldn't be used inside.

src/cpu/primitive_attr_postops.cpp
```diff
@@ -235,7 +235,7 @@ dim_t get_prelu_weights_off(const memory_desc_t &weights_md,
             weights_md, l_offset, dst_dims, dst_ndims, weights_mask);
 }
 
-dim_t get_binary_src1_off(const memory_desc_t &src1_md, const dim_t l_offset,
+dim_t get_binary_srcn_off(const memory_desc_t &src1_md, const dim_t l_offset,
```
dzarukin, line 238:
```suggestion
// Note: src_md is either src1_md or src2_md
dim_t get_binary_src_off(const memory_desc_t &src_md, const dim_t l_offset,
```

src/cpu/primitive_attr_postops.cpp
```diff
@@ -304,13 +304,27 @@ void ref_post_ops_t::execute(float &res, const args_t &args) const {
...
                         (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_1));
+                const auto src2_binary_po = CTX_IN_MEM(const void *,
+                        (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_2));
+
+                bool src2_val = false;
+                if (e.is_binary_with_ternary_op()) {
+                    const auto &src2_desc = e.binary.src2_desc;
+                    const auto src2_off = get_binary_srcn_off(src2_desc,
+                            args.l_offset, dst_d.dims(), dst_d.ndims());
+                    src2_val = static_cast<bool>(io::load_int_value(
```
dzarukin, line 320: I'd add an assert for alg == select here to catch any potential issues with s8 being treated other than bool value.

tests/benchdnn/dnn_types.cpp
```diff
@@ -590,6 +594,12 @@ std::vector<std::pair<int, int>> attr_t::post_ops_t::get_po_masks(
                     : policy2mask(
                             DNNL_ARG_SRC_1, e.binary.policy, prim_kind, ndims);
             arg = DNNL_ARG_SRC_1;
+
+            if (e.is_binary_kind_with_ternary_op()) {
+                mask2 = e.binary.mask;
```
dzarukin, line 599: I don't see it working if user specified the policy instead of a mask number.

tests/benchdnn/utils/parser.cpp
```diff
@@ -206,6 +206,9 @@ attr_t::post_ops_t parse_attr_post_ops_func(const std::string &s) {
         } else if (e.is_binary_kind()) {
             const auto dt_str = get_substr(subs, subs_pos, ':');
             e.binary.src1_dt = str2dt(dt_str.c_str());
+
+            if (e.is_binary_kind_with_ternary_op()) e.binary.src2_dt = dnnl_s8;
```
dzarukin, line 210: No new inputs for benchdnn? Since we target sdpa fusion, I'd check that select post-op can be fused into matmul successfully.

After some thinking, it seems that the only change for post-op input is an algorithm, mask and format will be re-used and the data type is fixed. What about a broadcast case the Graph team asked for?

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1800,15 +1800,25 @@ int init_ref_memory_args_default_case(int exec_arg, dnn_mem_t &mem,
...
+                        /* int = */ true, alg, "def_binary_post_op");
+                const auto it = fill_cfg_map.find(
+                        is_src2_mem ? DNNL_ARG_SRC_2 : DNNL_ARG_SRC_1);
+                const bool has_external_cfg = it != fill_cfg_map.end();
+                const fill_cfg_t &binary_fill_cfg
+                        = has_external_cfg ? (*it).second : def_binary_cfg;
+                TIME_FILL(SAFE(
+                        fill_random_real(mem, ref_mem, res, binary_fill_cfg),
+                        WARN));
+            }
```
dzarukin, line 1821: I'm not sure I'm following how it's working... I'd expect an `} else if (exec_arg & DNNL_ARG_SRC_2) {` to appear instead of mixing stuff together... and a separate memory object to be filled.

tests/benchdnn/dnn_types.cpp
```diff
@@ -1168,9 +1196,16 @@ dnnl_primitive_attr_t create_dnnl_attr(
...
                 const auto &src1_md = attr_args.get_md(
                         (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_1));
+                const auto &src2_md = attr_args.get_md(
+                        (DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_2));
                 assert(query_md_ndims(src1_md) != 0);
-                DNN_SAFE_V(dnnl_post_ops_append_binary(
-                        ops, e.binary.alg, src1_md));
+
+                if (e.is_binary_kind_with_ternary_op())
+                    assert(query_md_ndims(src2_md) != 0);
```
dzarukin, line 1204: In release mode this code should break regular post-ops as C call would be under if, unless I'm really missing something.

```suggestion
                if (e.is_binary_kind_with_ternary_op()) {
                    assert(query_md_ndims(src2_md) != 0);
                }
```

tests/benchdnn/dnn_types.cpp
```diff
@@ -1065,6 +1078,21 @@ int attr_args_t::prepare_post_ops_mds(const attr_t &attr, int ndims,
             mds.emplace((DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx)
                                 | po_rhs_tensor_entry.arg_attr_mask),
                     std::move(rhs_tensor_desc));
+
+            if (e.is_binary_kind_with_ternary_op()) {
+
+                const post_ops_rhs_tensor_entry_t post_op_rhs_src2_tensor_entry
+                        = {e.binary.src2_dt, 0, e.binary.tag, DNNL_ARG_SRC_2};
```
dzarukin, line 1085: Would be great to have an explanation why `mask = 0`...

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -3859,11 +3859,37 @@ struct post_ops : public handle<dnnl_post_ops_t> {
     /// @param aalgorithm Binary algorithm for the post-op.
     /// @param src1_desc Memory descriptor of a second operand.
     void append_binary(algorithm aalgorithm, const memory::desc &src1_desc) {
-        error::wrap_c_api(dnnl_post_ops_append_binary(get(),
-                                  convert_to_c(aalgorithm), src1_desc.get()),
+        error::wrap_c_api(
+                dnnl_post_ops_append_binary_v2(get(), convert_to_c(aalgorithm),
+                        src1_desc.get(), nullptr),
```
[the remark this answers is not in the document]
avmanerikar: Updated along with the other suggestions. Thanks!

src/common/primitive_desc.hpp
avmanerikar: Using `switch(arg)` would require precomputing and unrolling for the values that depend in the runtime `idx` variable. This has been updated with a cleaner `if-else` implementation instead. Thanks!

src/cpu/primitive_attr_postops.cpp
avmanerikar: The `e.is_binary_with_ternary_op()` call accounts for the algorithm check but I have added the assert anyways in case there are more ternary op algorithms to be added in the future.

tests/benchdnn/dnn_types.cpp
avmanerikar: Ideally, there should be no mask requirement for the src2 tensor since broadcasting is disabled for the tensor and it has the same dimensions as src0. This has been updated in the code - thanks for the catch!

tests/benchdnn/utils/parser.cpp
avmanerikar: Since broadcasting support is not enabled for `select op`, the broadcast example from the Graph team may not be of use for this case. Other than `matmul`, are there any additional cases of interest - perhaps, fusion with `binary` or `conv` can also be helpful?

tests/benchdnn/dnnl_common.cpp
avmanerikar: I have added the explanation for the working in the comments. The filling for both the binary `src1/2` tensors is the same except for the minimum fill value. For the `src2` tensor, the condition `(is_binary_src1_arg || attr.post_ops.entry[bin_po_idx].is_binary_kind_with_ternary_op())` ensures that the values are filled only when the binary algorithm needs a third input and ignored otherwise.

tests/benchdnn/dnn_types.cpp
avmanerikar: The code has been re-written so that it is no longer required.

src/gpu/gpu_reorder_pd.hpp
```diff
@@ -62,6 +62,10 @@ struct gpu_reorder_pd_t : public reorder_pd_t {
 
     bool post_ops_ok() const {
         const auto &post_ops = attr()->post_ops_;
+        for (auto idx = 0; idx < post_ops.len(); ++idx) {
+            if (post_ops.entry_[idx].is_binary_with_ternary_op()) return false;
+        }
+
```
atkassen, line 68: This shouldn't be necessary as reorder only supports at most a single sum post-op (as evidenced by the condition below).

avmanerikar: Fixed. Thanks!

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -2487,6 +2519,81 @@ void jit_uni_binary_injector_t<isa, Vmm>::inject_binary(
...
+        const tail_lode_mode_t tail_load_mode) const {
+
+    const auto alg = post_op.binary.alg;
+    const auto rhs_arg_data_type
+            = get_src1_desc(post_op, rhs_arg_static_params_.dst_d).data_type;
+
+    if (alg == alg_kind::binary_select) {
+        // The select op is implemented by adding the results of
+        // the src0 tensor ANDed with the select input and the
+        // src1 tensor ANDed with the complimented sleect input.
```
atkassen, line 2535: nit: complimented sleect -> complemented select

include/oneapi/dnnl/dnnl.h
```diff
@@ -583,6 +583,16 @@ int DNNL_API dnnl_post_ops_len(const_dnnl_post_ops_t post_ops);
...
         const_dnnl_post_ops_t post_ops, int index);
 
+/// Returns the algorithm kind of a post-op entry.
+///
+/// @param post_ops Post-ops.
+/// @param index Post-op entry index.
+/// @returns The algorithm kind of the post-op with the specified index.
+/// @returns #dnnl_alg_kind_undef if there is no post-op at the specified
+///     index or if the post-op kind does not have algorithm kind as a member.
+dnnl_alg_kind_t DNNL_API dnnl_post_ops_get_alg_kind(
```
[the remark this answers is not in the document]
avmanerikar: This provides a cleaner option to querying the algorithm kind for forward declaration. A use case is here: https://github.com/uxlfoundation/oneDNN/blob/44304301623ddcc509dfc91f1b47874c264c5cab/tests/benchdnn/dnnl_common.hpp#L901

include/oneapi/dnnl/dnnl.h
```diff
@@ -750,6 +785,25 @@ dnnl_status_t DNNL_API dnnl_post_ops_get_params_binary(
...
+/// @param alg_kind Output binary algorithm kind.
+/// @param src1_desc Output memory descriptor of a second operand.
+/// @param src2_desc Output memory descriptor of a third operand. If the
+/// specificed algorithm is not one that requires a ternary input, src2_desc
+/// will be ignored.
+
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+/// @returns #dnnl_invalid_arguments if @p index does not refer to a binary
+///     post-op.
```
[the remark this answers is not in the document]
avmanerikar: Doxygen can process multiple `@return` statements into a single paragraph. Verifying from the source: https://www.doxygen.nl/manual/commands.html#cmdreturn

src/common/primitive_attr.hpp
```diff
@@ -431,10 +437,16 @@ struct dnnl_post_ops : public dnnl::impl::c_compatible {
             dnnl::impl::dim_t padding_l_size);
     dnnl::impl::status_t append_binary(dnnl::impl::alg_kind_t alg,
             const dnnl::impl::memory_desc_t *user_src1_desc);
```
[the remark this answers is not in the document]
avmanerikar: Updated. Thanks!

include/oneapi/dnnl/dnnl.h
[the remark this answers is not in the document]
dzarukin: That's the new API to maintain. "Cleaner" is not a strong enough motivation to introduce it. @mgouicem - what would be your thoughts on this?

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_cpu

ShanoToni (approved):

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -575,10 +610,14 @@ Xbyak::Address jit_uni_binary_injector_t<isa, Vmm>::prepare_rhs_arg_addr(
     switch (rhs_broadcasting_strategy) {
         case broadcasting_strategy_t::scalar: return host_->ptr_b[rhs_addr_reg];
         case broadcasting_strategy_t::no_broadcast: {
+            // the address is not cached for the ternary input to avoid
+            // overwriting the address for the binary src input as they are
+            // loaded consecutively
```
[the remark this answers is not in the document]
avmanerikar: The trouble with using separate registers for the third (`src2`) tensor is that additional registers are not always available for allocation when `rhs_arg_static_params` are defined for the primitive (for example, the binary primitive). Also, the addresses for the `src1` and `src2` tensors are loaded consecutively for each unrolled loop so reusing the cache registers is not an option.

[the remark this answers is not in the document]
tczeszun: I see, so as I understood it only would work if we could change the order of post-ops execution or add separate registers for src2 address and cache.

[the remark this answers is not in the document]
avmanerikar: Specifically we will need two separate cache registers for the `src1` and `src2` addresses. One can work with a single address register as it gets refreshed for every `vmm_idx` and can be temporarily pushed to the stack but the cached addresses will need to be available for all `vmm_idx`'s. I may be wrong but changing the order of post-ops execution will not affect the availability of caching registers in this case.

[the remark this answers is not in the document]
avmanerikar: Wouldn't that still require assigning additional resources for post-ops during primitive execution? The availability will vary from primitive to primitive.

src/common/primitive_attr.cpp
```diff
@@ -718,6 +781,20 @@ status_t dnnl_post_ops_get_params_binary(const post_ops_t *post_ops, int index,
     return success;
 }
 
+status_t dnnl_post_ops_get_params_binary_v2(const post_ops_t *post_ops,
+        int index, alg_kind_t *alg_kind, const memory_desc_t **user_src1_desc,
+        const memory_desc_t **user_src2_desc) {
+    if (!simple_get_params_check(post_ops, index, primitive_kind::binary))
+        return invalid_arguments;
```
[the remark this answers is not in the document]
avmanerikar: There are verbose checks within `simple_get_params_check()` that print the messages so an additional message is not required here.

src/cpu/x64/injectors/jit_uni_postops_injector.cpp
```diff
@@ -393,12 +394,22 @@ bool post_ops_ok(const post_ops_ok_args_t &post_ops_ok_args) {
...
-                    assert(dst_d && "dst_d is null");
-                    return binary_injector::is_supported(isa,
-                            binary_injector::get_src1_desc(entry, *dst_d),
-                            *dst_d, enabled_bcast_strategy);
-                default: assert(!"Unhandled post_op type");
+                    if (entry.is_like_binary()) {
+                        assert(dst_d != nullptr && "dst_d is null");
+                        bool ok = binary_injector::is_supported(isa,
+                                binary_injector::get_src1_desc(entry, *dst_d),
+                                *dst_d, enabled_bcast_strategy);
```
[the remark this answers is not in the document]
avmanerikar: Done. Thanks!

tests/benchdnn/softmax/softmax.cpp
```diff
@@ -315,8 +316,13 @@ fill_cfg_t binary_po_fill_cfg(
...
         const auto alg = attr.post_ops.entry[bin_po_idx].kind;
-        cfg = fill_cfg_t(mem.dt(), 0.f, 16.f, /* int = */ true, alg,
-                "softmax_binary_post_op");
+        const bool is_src1_arg = !(exec_arg
+                ^ (DNNL_ARG_ATTR_MULTIPLE_POST_OP(bin_po_idx)
+                        | DNNL_ARG_SRC_1));
+
+        if (is_src1_arg)
+            cfg = fill_cfg_t(mem.dt(), 0.f, 16.f, /* int = */ true, alg,
+                    "softmax_binary_post_op");
```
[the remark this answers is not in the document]
avmanerikar: The call to `binary_po_fill_cfg()` configures filling parameters for the `src1` tensor for the binary post-op and not the `src2` tensor. This is an additional check that ensures the function is dealing with `src1` tensor only. The flag `is_post_ops_arg` does not distinguish between the two tensors for the binary postop.

include/oneapi/dnnl/dnnl.h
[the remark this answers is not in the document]
avmanerikar: The code is rewritten so that it no longer requires the additional function. Thanks!

tests/benchdnn/softmax/softmax.cpp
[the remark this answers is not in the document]
dzarukin: Could you point to the spot which guarantees src2 won't reach this point? Because if it does it won't likely work as expected anyway, with this extra precaution measure or not.

tests/benchdnn/dnn_types.cpp
```diff
@@ -598,6 +598,10 @@ std::vector<std::pair<int, int>> attr_t::post_ops_t::get_po_masks(
 
         assert(mask >= 0);
         v_masks.emplace_back(DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | arg, mask);
+
+        if (e.is_binary_kind_with_ternary_op())
+            v_masks.emplace_back(
+                    DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_2, -1);
```
dzarukin, line 604: Would be nice to explain `-1` in a comment.

tests/benchdnn/dnn_types.cpp
```diff
@@ -1065,6 +1069,15 @@ int attr_args_t::prepare_post_ops_mds(const attr_t &attr, int ndims,
             mds.emplace((DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx)
                                 | po_rhs_tensor_entry.arg_attr_mask),
                     std::move(rhs_tensor_desc));
+
+            if (e.is_binary_kind_with_ternary_op()) {
+                auto rhs_select_tensor_desc = dnn_mem_t::init_md(
+                        ndims, dims, e.binary.src2_dt, dtag);
```
dzarukin, line 1075: I'm not sure I understand the logic behind taking the destination tag. It either should be unconditional `tag::any` so that we check at least the code path inside the library that configures the tensor, or should follow src1 tag logic with `tag::any` being a default.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1794,22 +1794,48 @@ int init_ref_memory_args_default_case(int exec_arg, dnn_mem_t &mem,
     const bool is_dropout_p = (exec_arg == DNNL_ARG_ATTR_DROPOUT_PROBABILITY);
     const bool is_dropout_seed = (exec_arg == DNNL_ARG_ATTR_DROPOUT_SEED);
     const bool is_rounding_seed = (exec_arg == DNNL_ARG_ATTR_ROUNDING_SEED);
+    const bool is_binary_arg = (exec_arg & DNNL_ARG_SRC_1);
 
     if (is_post_ops_arg) {
-        if (exec_arg & DNNL_ARG_SRC_1) {
+        if (is_binary_arg) {
```
dzarukin, line 1800: Unless there's a bug, I don't see why `DNNL_ARG_SRC_2` would go into this `if`... Could you check it for me, please?

tests/benchdnn/dnnl_common.hpp
```diff
@@ -894,9 +894,17 @@ void init_memory_args(dnn_mem_map_t &mem_map, const prb_t *prb,
...
-        const auto &po_md = query_md(const_pd, po_arg);
-        mem_map.emplace(po_arg, dnn_mem_t(po_md, test_engine));
+        int po_arg1 = DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_1;
+        const auto &po_md1 = query_md(const_pd, po_arg1);
+        mem_map.emplace(po_arg1, dnn_mem_t(po_md1, test_engine));
+
+        dnnl_alg_kind_t alg;
+        const auto status = dnnl_post_ops_get_params_binary_v2(
+                const_po, idx, &alg, nullptr, nullptr);
+        if ((status != dnnl_success) || (alg != dnnl_binary_select)) continue;
```
dzarukin, line 904:
```suggestion
        const auto alg = query_bin_po_alg(const_po, idx);
        if (alg != dnnl_binary_select) continue;
```
`query_bin_po_alg` would be a part of dnnl_query.{c,h}pp files.

avmanerikar: Done. Thanks!

tests/benchdnn/dnn_types.cpp
avmanerikar: For the select operation, the `cond` tensor tag will need to be similar to  `src0` tensor so the destination tag is used to initialize the `src2` descriptor.

tests/benchdnn/dnnl_common.cpp
avmanerikar: It is not a bug but a consequence of how the macros are defined. With:
```
#define DNNL_ARG_ATTR_MULTIPLE_POST_OP_BASE 32768
#define DNNL_ARG_SRC_1 2
#define DNNL_ARG_SRC_2 3
```
`(exec_arg & DNNL_ARG_SRC_1) ` is set for both cases when exec_arg holds `DNNL_ARG_SRC_1` or `exec_arg` holds `DNNL_ARG_SRC_2`. That is why the code checks the args with the more precise masks:
```
const bool is_src1_arg = !(exec_arg ^ (DNNL_ARG_ATTR_MULTIPLE_POST_OP(bin_po_idx)
                            | DNNL_ARG_SRC_1));
const bool is_src2_arg = !(exec_arg ^ (DNNL_ARG_ATTR_MULTIPLE_POST_OP(bin_po_idx)
                            | DNNL_ARG_SRC_2));
```

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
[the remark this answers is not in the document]
tczeszun: It will for sure, but we could always keep the non-cached version for cases where it's not worth using.

tests/benchdnn/dnn_types.cpp
dzarukin: The fact that it's needed the same doesn't imply where it comes from.
User sets it up regardless of the requirement - either they let the library deduce it and query it (tag::any), or they manually specify it (tag::XXX). Both options are a property of post-ops parser, and the benchdnn internals.

tczeszun (approved):

atkassen (approved):

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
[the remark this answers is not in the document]
tczeszun: resolving, as we came to conclusion to add TODO for caching and other bcast strategies

vpirogov (approved):

tests/benchdnn/dnnl_common.cpp
avmanerikar: Done. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_cpu

avmanerikar:
make test

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

avmanerikar:
make test

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -144,8 +144,9 @@ bool any_binary_postop_rhs_non_scalar_broadcast(
                             get_src1_desc(entry, dst_d), dst_d,
                             get_all_strategies_supported_by_injector());
                     return !utils::one_of(bcast_type,
-                            broadcasting_strategy_t::scalar,
-                            broadcasting_strategy_t::unsupported);
+                                   broadcasting_strategy_t::scalar,
+                                   broadcasting_strategy_t::unsupported)
+                            || entry.is_binary_with_ternary_op();
```
dzarukin, line 149: Shouldn't this code be changed to check bcast_type of src2_desc and `&&`-ed to the original condition?

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -517,42 +519,82 @@ void jit_uni_binary_injector_t<isa, Vmm>::compute_vector_range(
...
+        const Vmm dst_vmm(vmm_idx);
+
+        // For binary ops with ternary inputs, a temporary vmm will be needed
+        // for the additional input.
+        if (needs_ternary_input) {
+            const auto rhs2_arg_data_type
+                    = get_src2_desc(post_op, dst_d).data_type;
+            rhs2_arg_addr = prepare_rhs_arg_addr(vmm_idx, rhs_arg_idx + 1,
+                    post_op, rhs_arg_params,
+                    broadcasting_strategy_t::no_broadcast, true, true);
```
dzarukin, line 547: To pass `no_broadcast`, there must be a check added to `validate_binary_with_dst_consistency`.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -2495,6 +2554,79 @@ void jit_uni_binary_injector_t<isa, Vmm>::inject_binary(
...
+                load_rhs(rhs_arg_data_type, dst, rhs_addr, tail_load_mode,
+                        with_tail);
+            if (types::is_integral_dt(rhs_arg_data_type)) cvt_to_f32(dst);
+
+            host_->vpand(tmp_vmm, dst, tmp_vmm);
+            pop_vmm(host_, dst);
+            host_->vpaddd(dst, dst, tmp_vmm);
+        }
+    } else
+        assert(!"unsupported algorithm");
```
dzarukin, line 2627:
```suggestion
    } else {
        assert(!"unsupported algorithm");
    }
```

tests/benchdnn/utils/dnnl_query.hpp
```diff
@@ -48,6 +48,8 @@ int query_n_outputs(const_dnnl_primitive_desc_t pd);
 bool query_post_ops_has_kind(dnnl_primitive_t prim, dnnl_primitive_kind_t kind);
 bool query_post_ops_has_kind(
         const_dnnl_post_ops_t post_ops, dnnl_primitive_kind_t kind);
+bool query_po_alg_kind(
+        const_dnnl_post_ops_t post_ops, int idx, dnnl_alg_kind_t alg);
```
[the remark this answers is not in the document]
avmanerikar: Updated in the latest changes along with the rest of suggestions. Thanks!

src/cpu/x64/injectors/jit_uni_postops_injector.cpp
```diff
@@ -393,12 +394,23 @@ bool post_ops_ok(const post_ops_ok_args_t &post_ops_ok_args) {
...
```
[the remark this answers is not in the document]
avmanerikar: `binary_injector::is_supported()` already has verbose checks in place for `src1_desc` so this is redundant.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
avmanerikar: There is no broadcasting type assigned to `src2`. Since `src2` has the same dimensions as `src0`, the broadcasting type is always `broadcasting_strategy_t::no_broadcast` which is non-scalar.

avmanerikar:
make test

avmanerikar:
make test

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

src/common/primitive_attr.cpp
```diff
@@ -333,12 +374,18 @@ status_t post_ops_t::entry_t::validate_binary(
     if (!is_binary()) return status::success;
 
     VCHECK_ATTR(dst_md->ndims == binary.user_src1_desc.ndims,
-            VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "dst", "bin_po",
+            VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "dst", "bin_po src1",
             dst_md->ndims, binary.user_src1_desc.ndims);
 
-    VCHECK_ATTR(IMPLICATION(engine_kind == dnnl_cpu,
-                        binary.user_src1_desc.data_type != data_type::f64),
-            VERBOSE_INVALID_DATATYPE, "bin_po src1");
```
dzarukin, line 341: This should stay. Likely a product of bad rebase.

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -249,6 +249,20 @@ supported, positioned after `MASK_INPUT`, to specify physical memory format.
...
+arguments for both the tensors:
+
+```
+--attr-post-ops=select:DT[.S1_MASK_INPUT[.S1_TAG]][:S2_MASK_INPUT[.S2_TAG]]
+```
+
+The arguments for each tensor are provided in the form `DT.MASK_INPUT.TAG`, 
+with the arguments for the third tensor positioned after that of the second
+and are separated by `:`. The arguments are optional for the third tensor and
+the data type value for the third tensor is fixed at `s8`.
```
dzarukin, line 265: @ranukund, could you, please, check out this page?

tests/benchdnn/dnn_types.cpp
```diff
@@ -747,18 +752,46 @@ std::ostream &operator<<(std::ostream &s, const attr_t::post_ops_t &post_ops) {
...
+                                == mask_input_t::policy);
+                        s << e.binary.src2_policy;
+                    }
+                }
+                if (e.binary.src2_tag != tag::any) {
+                    if (!delim_added)
+                        s << ":";
+                    else
+                        s << src_delim;
+                    s << e.binary.src2_tag;
```
dzarukin, line 792: Nit: 
```suggestion
                    s << (delim_added ? src_delim : ":") << e.binary.src2_tag;
```

tests/benchdnn/inputs/matmul/test_matmul_ci
```diff
@@ -201,6 +201,15 @@
 --attr-post-ops=mul:f32,relu,sum,prelu,prelu:per_oc
 2x10x3x20:2x10x20x4n"postops+runtime_dims_4d"
 
+# Post-ops with binary select op
+--reset
+--dt=f32,s8:s8:s32,bf16
+--stag=ab --wtag=ab --dtag=ab
+--attr-post-ops=select:s8,select:f32,\
+                select:f32.per_tensor,select:f32.per_oc,\
+                select:f32.per_tensor+mul:s8
```
dzarukin, line 210: Three options should be enough.
```suggestion
--attr-post-ops=select:f32,\
                select:f32.per_oc,\
                select:s8.per_tensor+add:f32
```

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_cpu

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -3732,6 +3732,32 @@ struct post_ops : public handle<dnnl_post_ops_t> {
                 "could not append a binary post-op");
     }
 
+    /// Appends a binary post-op with ternary operators.
+    ///
+    /// The kind of this post operation is #dnnl_binary.
```
ranukund, line 3737:
```suggestion
    /// This post operation is categorized as #dnnl_binary.
```

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -3732,6 +3732,32 @@ struct post_ops : public handle<dnnl_post_ops_t> {
...
+    ///
+    /// In the simplest case when this is the only post operation, the
+    /// computations would be:
```
ranukund, line 3740:
```suggestion
    /// computations will be:
```

doc/programming_model/attributes_post_ops.md
```diff
@@ -281,6 +281,29 @@ Currently the following scenarios are optimized:
...
+
+~~~cpp
+void dnnl::post_ops::append_binary(
+        algorithm alg, // binary algorithm to apply
+        const memory::desc &src1 // memory descriptor for a second memory operand
+        const memory::desc &src2 // memory descriptor for a third memory operand
+        );
+~~~
+
+The `alg`, `src1` and `src2` parameters are the same as in
```
ranukund, line 295:
```suggestion
The `alg`, `src1`, and `src2` parameters are the same as in 
```

ranukund (approved):
Minor comments suggested, please incorporate as you see fit, thanks!

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

tests/benchdnn/utils/parser.cpp
```diff
@@ -220,17 +220,117 @@ attr_t::post_ops_t parse_attr_post_ops_func(const std::string &s) {
...
+                }
+
+                if (s2_pos != std::string::npos) {
+                    const auto unknown_str
+                            = get_substr(subs, subs_pos, src_delim);
+                    BENCHDNN_PRINT(0, "%s \'%s\' %s\n",
+                            "Warning: Additional unrecognized arguments",
+                            unknown_str.c_str(), "are specified.");
+                }
+            };
```
dzarukin, line 315: What if the logic gets like this:
```
for (const auto &s : {subs, subs2}) {
    process_the_string(...)
}
```
`process_the_string` (pick a better name) will fill attr.binary objects. Then do post-processing and if some entries were updated that are not supposed, provide a warning. Right now I returns it unconditionally whether the restricted input was specified or not.

dzarukin (commented):
One last comment to address. I'm fine with the rest.

tests/benchdnn/utils/parser.cpp
avmanerikar: Updated in the latest changes. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_gpu

dzarukin (approved):

PR #2901, opened 2025-03-17
benchdnn: fix out-of-memory encountered on iGPU systems

rjoursler:
Adds a new benchdnn_combined_limit for the combined device and CPU memory limit. A new combined capacity factor was introduced, as the capacity factor appears unnecessarily strict from what has been observed in practice. We likely could increase the current capacity factor, but that is avoided in this patch to limit the scope of this change.

Fixes [MFDNN-13353](https://jira.devtools.intel.com/browse/MFDNN-13353).


rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

dyoussif (approved):

PR #2907, opened 2025-03-18
common: make the primitive cache mutex a member instead of global

mgouicem:


src/common/cache_utils.hpp
```diff
@@ -114,10 +114,7 @@ struct cache_t {
     virtual value_t get_or_add(const key_t &key, const value_t &value) = 0;
     virtual void remove_if_invalidated(const key_t &key) = 0;
     virtual void update_entry(const key_t &key, const object_t &p) = 0;
-    static utils::rw_mutex_t &rw_mutex() {
-        static utils::rw_mutex_t mutex;
-        return mutex;
-    }
+    mutable utils::rw_mutex_t rw_mutex_;
```
rjoursler, line 117: I believe you can just do the following and drop removing all the getter calls.

```Suggestion
utils::rw_mutex_t& rw_mutex() const {
     return rw_mutex_;
}
mutable utils::rw_mutex_t rw_mutex_;
```

rjoursler (approved):
Out of curiosity, what is the motivation for making this change?

vpirogov:
@mgouicem, does this one still make sense?

mgouicem:
Nope (it should have been a draft PR), closing.

PR #2908, opened 2025-03-18
build: removed -fcf-protection build option for old GCC

vpirogov:
Restores ability to build oneDNN with GCC 7.x.

Backport of #2946.


cmake/SDL.cmake
```diff
@@ -30,9 +30,10 @@ macro(sdl_unix_common_ccxx_flags var)
     append(${var} "-fPIC -Wformat -Wformat-security")
 endmacro()
 
-macro(sdl_gnu_common_ccxx_flags var)
+macro(sdl_gnu_common_ccxx_flags var gnu_version)
```
spalicki, line 33: This macro is also used in cmake\host_compiler.cmake, which might not compile correctly now. Host compiler is also limited to GCC>8 in cmake/host_compiler_id.cmake, but that is a separate issue.

vpirogov: > This macro is also used in cmake\host_compiler.cmake, which might not compile correctly now. 

Good catch.

> Host compiler is also limited to GCC>8 in cmake/host_compiler_id.cmake, but that is a separate issue.

Host compiler is SYCL-specific concept. It's not a concern in this case.

vpirogov:
make test
disable test_device_cpu
disable test_device_gpu

cmake/host_compiler.cmake
```diff
@@ -78,7 +78,7 @@ if(DPCPP_HOST_COMPILER_KIND MATCHES "^(GNU|CLANG)$")
 
     if(DPCPP_HOST_COMPILER_KIND STREQUAL "GNU")
         platform_gnu_nowarn_ccxx_flags(DPCPP_CXX_NOWARN_FLAGS ${DPCPP_HOST_COMPILER_MAJOR_VER}.${DPCPP_HOST_COMPILER_MINOR_VER})
-        sdl_gnu_common_ccxx_flags(DPCPP_HOST_COMPILER_OPTS)
+        sdl_gnu_common_ccxx_flags(DPCPP_HOST_COMPILER_OPTS CMAKE_CXX_COMPILER_VERSION)
```
spalicki, line 81: `DPCPP_HOST_COMPILER_VER`, `CMAKE_CXX_COMPILER_VERSION` will pass main/SYCL compiler (ICX/clang) version.

vpirogov: Shouldn't code between the meetings :(

spalicki (approved):

PR #2914, opened 2025-03-18
xe: ocl: prevent double -> float literal conversion [rnn, softmax, (de)conv]

yehudaorel:
# Description

On older arch, if cl_khr_f64 is unsupported, the compiler throws warning about double literals.

Additionally fixes warning in conv:
`incompatible pointer types initializing '__private float *__private' with an expression of type '__private float2 (*)[1][6]' `

### DIFF
--- a/src/gpu/intel/ocl/ocl_post_ops.h
+++ b/src/gpu/intel/ocl/ocl_post_ops.h
@@ -188,7 +188,7 @@ float fwd_Xnary(bool is_binary, unsigned algorithm, float x, float y,
 #define APPLY_ALL_PO_STAGES(accumulator, acc_elem_dt, ...) \
     { \
         const int nelems = sizeof(accumulator) / sizeof(acc_elem_dt); \
-        acc_elem_dt *acc = &accumulator; \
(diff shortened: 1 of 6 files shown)


atkassen (approved):

src/gpu/intel/ocl/ocl_post_ops.h
```diff
@@ -188,7 +188,7 @@ float fwd_Xnary(bool is_binary, unsigned algorithm, float x, float y,
 #define APPLY_ALL_PO_STAGES(accumulator, acc_elem_dt, ...) \
     { \
         const int nelems = sizeof(accumulator) / sizeof(acc_elem_dt); \
-        acc_elem_dt *acc = &accumulator; \
+        acc_elem_dt *acc = (acc_elem_dt *)&accumulator; \
```
rjoursler, line 191: Lets avoid performing a pointer cast here as it can suppress incorrect pointer conversions. Instead, lets fix the caller of this macro similar to the following:

```
APPLY_POST_OPS_SERIAL(&C[0][0], DATA_T, ...
```

PR #2918, opened 2025-03-19
[GPU] Drop generated gemm cases causing crashes

kealan-barbieri:
# Description

Commenting out generated cases observed to crash testing, disabling all other testcases from running. 

These cases should be reenabled pending https://jira.devtools.intel.com/browse/MFDNN-13381. 

# Checklist

## General


dzarukin (approved):
This way we will soon disable all generated cases :D

rjoursler (approved):

rjoursler:
@dzarukin Would it be reasonable to make benchdnn robust against crashes during the testing of a workload. It would likely require running most testing on child process so that unrecoverable errors can be handled by the parent relaunching the process, but I am not sure how difficult it would be to integrate into the current code base..

PR #2920, opened 2025-03-19
gpu:intel: doc workaround for ocl compiler bug in sdpa ukernels

pv-pterab-s:
Comment giving instruction on maneuvering around major OpenCL compiler bug when working with SDPA ukernel. Debugging is very difficult without this hint.


umar456:
I don't see the value in this comment because we are using the results of the microkernel.

pv-pterab-s:
> I don't see the value in this comment because we are using the results of the microkernel.

This is directive from Konstantin. Whenever working with this kernel, unused variables will eventually pop up even in the most unanticipated ways. Had I known this workaround from Peter, tons of time debugging would have been saved. I suggested a wiki page or note, but he wants it in the code instead.

petercad (approved):

PR #2922, opened 2025-03-19
benchdnn: matmul: resolve the regression introduced by grouped scales and zero-points support

dzarukin:
MFDNN-13114

Couple changes to improve the situation reported in the tracker above:
* Avoid unnecessary get_idx calls for each K while it should be done once per a group of K.
* Use `ba` format for weights in ref kernel to avoid bad leading dimensions and have memory dense in K-loop.

Seems to give a significant boost compared to what it was before.

### DIFF
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -68,6 +68,18 @@ dnn_mem_t::dnn_mem_t(const_dnnl_memory_desc_t md, dnnl_data_type_t dt,
     }
 }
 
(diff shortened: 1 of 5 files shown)


dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul

dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul

mgouicem (approved):

spalicki (approved):

PR #2926, opened 2025-03-20
ukernel: x64: initialize amx palette once in finalize

mgouicem:
This PR moves palette generation from brgemm::set_hw_context (which can be called many times by user) to brgemm::finalize (which is called once ahead of time). 

fixes MFDNN-13158


mgouicem:
make test
enable benchdnn_brgemm

dzarukin (approved):

inteldimitrius (approved):

PR #1521, opened 2022-12-15
cpu: risc-v: pooling: Add NCHW pooling with RVV intrinsics

pazamelin:
# Description

Hello, OneDNN team!

First of all we would like to thank you for all the amazing work you are doing.
OneDNN is a key element in many deep learning frameworks enabling highly optimized training and inference on various platforms.

As you probably know, RISC-V committee has some time ago ratified vector extension v1.0. There is no CPU with RVV 1.0 hardware implementation available on the market yet, but it is expected they may appear within one year. Hovewer, there are publicly available tools which already enable development for this new architecture. These tools include gcc and clang compilers capable to generate RVV 1.0 code and QEMU emulator that allows functional testing of such code.

Currently, OneDNN can be successfully compiled for RISC-V ISA, but only generic reference implementations are available in that scenario. We think having oneDNN optimized for RISC-V would be beneficial both for the library itself and the emerging RISC-V ecosystem. So, we propose to begin the process of adding primitive implementations that will utilize RVV extension. 
This PR adds an implementation of Pooling primitive optimized with RVV intrinsics. For now, only NCHW layouts with alg_kind::pooling_max and f32 are supported, but we are eager to improve it in the near future. Please bear in mind that this is an initial version which will be subject to further optimizations.
Also, we understand that implementing the kernel with JIT assembler could be a better option than using intrinsics.
Hovewer, there seems to be no functional JIT generator for RISC-V available yet. Will that be a blocker?

By default, this implementation is not compiled, since RVV extension may not be available on a particular RISC-V CPU or QEMU instance. If RVV is available, compilation can be enabled with the `DNNL_RISCV_USE_RVV` CMake option. 

We are looking forward to hear your opinion on this.
Thanks

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?

We cross-compile oneDNN for RISC-V using gcc with flags "-march=rv64gcv -mabi=lp64d",
then execute tests in QEMU running linux kernel [patched](https://patchwork.kernel.org/project/linux-riscv/cover/20220921194629.1480202-1-stillson@rivosinc.com/) for vector ISA support.
In that case, all gtests except these two pass:
   - [test_convolution_eltwise_forward_f32.log](https://github.com/oneapi-src/oneDNN/files/10235410/test_convolution_eltwise_forward_f32.log) - 8 tests failed 
[...]

PR #2929, opened 2025-03-21
[cpu] riscv64: update intrinsics

zhangfeiv0:
# Description

Currently, oneDNN can be compiled normally on RISC-V architecture platforms, but when using newer compilers and enabling vector extensions, it does not compile RVV-related code (specifically, the code introduced in #1521). Below is the configuration output information regarding RVV when using clang-17:
```shell
export CC=clang-17 CXX=clang++-17
cmake .. -DDNNL_CPU_RUNTIME=SEQ -DCMAKE_CXX_FLAGS=-march=rv64gcv -DCMAKE_C_FLAGS=-march=rv64gcv
```
```plaintext

-- Performing Test CAN_COMPILE_RVV_INTRINSICS
-- Performing Test CAN_COMPILE_RVV_INTRINSICS - Failed
-- Can compile RVV Intrinsics: FALSE
-- DNNL_RISCV_USE_RVV_INTRINSICS: FALSE

```
This is because the RVV intrinsics in #1521 are based on the old version specifications, and the new versions all have the __riscv64 prefix, as detailed in [rvv-intrinsic-doc](https://github.com/riscv-non-isa/rvv-intrinsic-doc). Currently, most open-source software projects (such as OpenCV, OpenBLAS, and SLEEF) have already adopted the new RVV intrinsics.

Based on #1521, the following modifications were made:

1. Updated intrinsics in rvv_nchw_pooling.cpp to the latest version;
2. Added version checks for intrinsics in CMake.

These changes enable Clang 17+ or GCC 14+ to properly compile RVV-related code, while older compiler versions will fall back to generic implementations.

# Checklist

## General

Tested on both QEMU and [Banana Pi BPI-F3](https://docs.banana-pi.org/en/BPI-F3/BananaPi_BPI-F3) (vector length is 256 bits). Due to the long testing time required for `matmul`, it has been omitted here, as this PR only modifies the names of intrinsics. Below are the test results:

[result.log](https://github.com/user-attachments/files/19380139/result.log)

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -474,8 +474,7 @@ if (DNNL_TARGET_ARCH STREQUAL "RV64")
 
                                #include <riscv_vector.h>
                                int main() {
-                                size_t size = 64;
-                                return __riscv_vsetvl_e32m2(size);
+                                return 0;
                                };"
                                CAN_COMPILE_RVV_INTRINSICS
     )
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -464,10 +464,18 @@ endif()
 if (DNNL_TARGET_ARCH STREQUAL "RV64")
     # Check if the RVV Intrinsics can be compiled with the current toolchain and flags
     include(CheckCXXSourceCompiles)
-    check_cxx_source_compiles("#include <riscv_vector.h>
+    check_cxx_source_compiles("#if !defined(__riscv) || !defined(__riscv_v)
+                               #error \"RISC-V or vector extension(RVV) is not supported by the compiler\"
+                               #endif
+
+                               #if defined(__riscv_v_intrinsic) && __riscv_v_intrinsic < 12000
+                               #error \"Wrong intrinsics version, v0.12 or higher is required for gcc or clang\"
+                               #endif
+
+                               #include <riscv_vector.h>
                                int main() {
                                 size_t size = 64;
-                                return vsetvl_e32m2(size);
+                                return __riscv_vsetvl_e32m2(size);
                                };"
                                CAN_COMPILE_RVV_INTRINSICS
     )
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -469,7 +469,7 @@ if (DNNL_TARGET_ARCH STREQUAL "RV64")
                                #endif
 
                                #if defined(__riscv_v_intrinsic) && __riscv_v_intrinsic < 12000
-                               #error \"Wrong intrinsics version, v0.12 or higher is required for gcc or clang\"
+                               #error \"RISC-V intrinsics v0.12 or higher is required\"
                                #endif
 
                                #include <riscv_vector.h>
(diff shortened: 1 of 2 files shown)


cmake/platform.cmake
```diff
@@ -464,10 +464,18 @@ endif()
...
+
+                               #if defined(__riscv_v_intrinsic) && __riscv_v_intrinsic < 12000
+                               #error \"Wrong intrinsics version, v0.12 or higher is required for gcc or clang\"
+                               #endif
+
+                               #include <riscv_vector.h>
                                int main() {
                                 size_t size = 64;
-                                return vsetvl_e32m2(size);
+                                return __riscv_vsetvl_e32m2(size);
```
mgouicem, line 478: Why not just return 0, given that __riscv_v_intrinsic is already check to be 0.12 or above?

mgouicem (commented):
LGTM thanks. Do you know what becomes the minimal compiler version for GCC/clang after this change?

Tagging @pazamelin @aaronfranke @dmitry-gorokhov for awareness.

zhangfeiv0:
The following is an excerpt from [rvv-intrinsic-doc](https://github.com/riscv-non-isa/rvv-intrinsic-doc). After the introduction of this PR, the minimum GCC version required will be GCC 14, and the minimum Clang version will be Clang 17.
image

Furthermore, below is the configuration output when compiling OneDNN with different versions of the compiler.
<details> <summary>use gcc 13 configuration output</summary>
iShot_2025-03-21_20 09 55
</details>

<details> <summary>use gcc 14 configuration output</summary>
iShot_2025-03-21_20 12 07
</details>

<details> <summary>use clang 16 configuration output</summary>
iShot_2025-03-21_20 30 24
</details>

<details> <summary>use clang 17 configuration output</summary>
iShot_2025-03-21_20 31 33
</details>

cmake/platform.cmake
zhangfeiv0: Done.

cmake/platform.cmake
```diff
@@ -464,10 +464,17 @@ endif()
 if (DNNL_TARGET_ARCH STREQUAL "RV64")
     # Check if the RVV Intrinsics can be compiled with the current toolchain and flags
     include(CheckCXXSourceCompiles)
-    check_cxx_source_compiles("#include <riscv_vector.h>
+    check_cxx_source_compiles("#if !defined(__riscv) || !defined(__riscv_v)
+                               #error \"RISC-V or vector extension(RVV) is not supported by the compiler\"
+                               #endif
+
+                               #if defined(__riscv_v_intrinsic) && __riscv_v_intrinsic < 12000
+                               #error \"Wrong intrinsics version, v0.12 or higher is required for gcc or clang\"
```
vpirogov, line 472:
```suggestion
                               #error \"RISC-V intrinsics v0.12 or higher is required\"
```

zhangfeiv0: Done.

vpirogov (approved):

zhangfeiv0:
Hi, @mgouicem 
If everything is okay, can we proceed with the merge?

mgouicem (approved):

PR #2946, opened 2025-03-25
build: removed -fcf-protection build option for old GCC

vpirogov:
While we do not officially support GCC 7.x anymore, there's no reason to break builds because of unsupported compiler options.


vpirogov:
make test
disable test_device_cpu
disable test_device_gpu

spalicki (approved):

mgouicem (approved):

PR #2264, opened 2024-12-13
cpu: aarch64: thread_local ThreadpoolScheduler setting

davsva01:
# Description

Makes setting of ThreadpoolScheduler in ACL thread_local to enable use of multiple (thread_local) Schedulers in ACL.

# Checklist

## General

- [X] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [X] Have you formatted the code using clang-format?

PR #2958, opened 2025-03-26
cpu: aarch64: default num_threads to max for acl_threadpool

Sqvid:
# Description
Default num_threads to max for acl_threadpool. This brings it in line with the other schedulers. This fixes an out-of-bounds crash in upstream frameworks where ops are configured to be single threaded and then are run in parallel.

Also included some cleanup:

- Use `std::mutex` instead of `arm_compute::Mutex` (was a `typedef`)
- Use `std::lock_guard` instead of `arm_compute::lock_guard` (was a minimal wrapper)
- Add missing `thread_local` to the `acl_tp_benchmark_scheduler` (probably accidentally omitted from https://github.com/uxlfoundation/oneDNN/pull/2264)

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/acl_thread.cpp
+++ b/src/cpu/aarch64/acl_thread.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2024 Arm Ltd. and affiliates
+* Copyright 2022-2025 Arm Ltd. and affiliates
 *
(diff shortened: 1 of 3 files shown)


Sqvid:
cc: @theComputeKid @davsva01 @jondea

davsva01 (approved):
Looks good to me!

cfRod (approved):

PR #2964, opened 2025-03-26
ngen: fix missing field intializer warning

rjoursler:
Address a GCC warning caught when using GCC as a DPC++ host compiler.

Fixes [MFDNN-13429](https://jira.devtools.intel.com/browse/MFDNN-13429l)

### DIFF
--- a/cmake/SDL.cmake
+++ b/cmake/SDL.cmake
@@ -44,13 +44,13 @@ endmacro()
 # only. To prevent warnings on users' side who use the library and turn
 # this warning on, let's use it too. Applicable for the library sources
 # and interfaces only (tests currently rely on that fact heavily)
-macro(sdl_gnu_src_ccxx_flags var)
+macro(sdl_unix_src_ccxx_flags var)
     append(${var} "-Wmissing-field-initializers")
 endmacro()
 
-macro(sdl_gnu_example_ccxx_flags var)
+macro(sdl_unix_example_ccxx_flags var)
     # At this point the flags for src and examples are the same
-    sdl_gnu_src_ccxx_flags(${var})
+    sdl_unix_src_ccxx_flags(${var})
 endmacro()
 
 set(ONEDNN_SDL_COMPILER_FLAGS)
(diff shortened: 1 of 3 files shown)


echeresh (approved):

cmake/platform.cmake
```diff
@@ -245,6 +245,9 @@ elseif(UNIX OR MINGW)
 
         # Align with GCC -Wall
         append(CMAKE_CCXX_FLAGS "-Wsign-compare")
+
+        # Align with GCC host compiler options
+        append(DPCPP_SRC_CXX_FLAGS "-Wmissing-field-initializers")
```
spalicki, line 250: This is added in SDL.cmake to both host (for DPCPP) and main compiler via [sdl_gnu_src_ccxx_flags ](https://github.com/uxlfoundation/oneDNN/blob/5f6951bbf11ed1df06c3ad7f8a9a28d4bc6cf542/cmake/SDL.cmake#L47) macro. If it is not being applied, it should be fixed there.

rjoursler: I am not quite following what action I should take based on your feedback here.  This macro is only applied for dpc++ [builds with the GCC host compiler](https://github.com/uxlfoundation/oneDNN/blob/5f6951bbf11ed1df06c3ad7f8a9a28d4bc6cf542/cmake/host_compiler.cmake#L82). This change always enables this warning for dpc++ so that we do not need to compile in that very specific configuration to trigger warnings. I chose against using `sdl_gnu_src_ccxx_flags` directly here as not all GCC warnings are available in dpc++, to prevent new warnings added in SDL.cmake triggering issues on non GCC builds.

spalicki: I assume you specifically want to block some configuration from using this flag, since with your current design only icx main and gcc host compiler will have it and e.g. clang host compiler will not. But once we add it to main compiler of all DPCPP builds all the warnings will pop up anyway, so at this point why not add it to all configurations?

In that case let's not add the same flag in 2 places especially since `sdl_gnu_src_ccxx_flags` is used for only this one flag, so let's just change the macro name to `sdl_unix_src_ccxx_flags`, and move its usage in both [host_compiler.cmake](https://github.com/uxlfoundation/oneDNN/blob/5f6951bbf11ed1df06c3ad7f8a9a28d4bc6cf542/cmake/host_compiler.cmake#L82) and [SDL.cmake](https://github.com/uxlfoundation/oneDNN/blob/5f6951bbf11ed1df06c3ad7f8a9a28d4bc6cf542/cmake/SDL.cmake#L67C9-L67C31) out of GNU if block just below `sdl_unix_common_ccxx_flags` (in both cases using only SRC variable). Unless that creates massive number of warnings in other builds.

rjoursler: Thanks for clarifying.

rjoursler:
make test
disable test_device_cpu
enable test_device_gpu

spalicki (approved):

PR #2971, opened 2025-03-28
enable graph API build on nv gpu[Do not review]

Jiexin-Zheng:
TBD...

### DIFF
--- a/src/graph/backend/dnnl/op_executable.cpp
+++ b/src/graph/backend/dnnl/op_executable.cpp
@@ -1643,6 +1643,23 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
 
     // 1. sqrt_variance = sqrt(variance + epsilon)
 
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    // binary + sqrt post-op fusion is unsupported on NVIDIA GPU
+    memory::dims epsilon_dims(variance.get_ndims(), 1);
+    desc.epsilon_desc_ = memory::desc(
+            epsilon_dims, memory::data_type::f32, memory::format_tag::a);
+
+    primitive_attr add_attr;
+    desc.add_pd_ = dnnl::binary::primitive_desc(p_engine, algorithm::binary_add,
+            variance, desc.epsilon_desc_, variance, add_attr);
+
+    primitive_attr sqrt_attr;
+    desc.sqrt_pd_ = dnnl::eltwise_forward::primitive_desc(p_engine,
+            prop_kind::forward, algorithm::eltwise_sqrt, variance, variance,
+            0.0f, 0.0f, sqrt_attr);
+#else
+
     // temp = variance + epsilon
     memory::dims epsilon_dims(variance.get_ndims(), 1);
     desc.epsilon_desc_ = memory::desc(
@@ -1657,6 +1674,7 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
     desc.add_pd_ = dnnl::binary::primitive_desc(p_engine, algorithm::binary_add,
             variance, desc.epsilon_desc_, variance, add_attr);
 
+#endif
     // 2. updated_weight = weights * scale / sqrt_variance
 
     // expand 1D scale and variance to same ndims with weights
@@ -1714,7 +1732,12 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
     memory::dims scratchpad_dims = variance.get_dims();
     // sqrt_variance, zero_bias and others (like epsilon),
     // or no need to alloc bias
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    size_t factor = bias.is_zero() ? 4 : 3;
+#else
     size_t factor = bias.is_zero() ? 3 : 2;
+#endif
     scratchpad_dims[0] *= factor;
     desc.scratchpad_desc_ = memory::desc(
             scratchpad_dims, variance.get_data_type(), memory::format_tag::a);
--- a/src/graph/backend/dnnl/op_executable.hpp
+++ b/src/graph/backend/dnnl/op_executable.hpp
@@ -495,8 +495,8 @@ struct conv_fwd_executable_t : public op_executable_t {
                     auto prim = dnnl::reorder(psrc_mem, to_mem);
                     auto e = dnnl::sycl_interop::execute(prim, stream,
                             {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                    {DNNL_ARG_TO,
-                                            const_cast<memory &>(to_mem)}},
+                                    { DNNL_ARG_TO,
+                                        const_cast<memory &>(to_mem) }},
                             sycl_deps);
                     sycl_deps = {e};
                     if (stream.get_engine().get_kind() == engine::kind::cpu)
@@ -505,8 +505,8 @@ struct conv_fwd_executable_t : public op_executable_t {
                     auto prim = dnnl::reorder(psrc_mem, dst_mem);
                     auto e = dnnl::sycl_interop::execute(prim, stream,
                             {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                    {DNNL_ARG_TO,
-                                            const_cast<memory &>(dst_mem)}},
+                                    { DNNL_ARG_TO,
+                                        const_cast<memory &>(dst_mem) }},
                             sycl_deps);
                     sycl_deps = {e};
                 }
@@ -561,16 +561,16 @@ struct conv_fwd_executable_t : public op_executable_t {
                     auto prim = dnnl::reorder(psrc_mem, to_mem);
                     auto e = dnnl::ocl_interop::execute(prim, stream,
                             {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                    {DNNL_ARG_TO,
-                                            const_cast<memory &>(to_mem)}},
+                                    { DNNL_ARG_TO,
+                                        const_cast<memory &>(to_mem) }},
                             ocl_deps);
                     ocl_deps = {e};
                 } else {
                     auto prim = dnnl::reorder(psrc_mem, dst_mem);
                     auto e = dnnl::ocl_interop::execute(prim, stream,
                             {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                    {DNNL_ARG_TO,
-                                            const_cast<memory &>(dst_mem)}},
+                                    { DNNL_ARG_TO,
+                                        const_cast<memory &>(dst_mem) }},
                             ocl_deps);
                     ocl_deps = {e};
                 }
@@ -626,7 +626,8 @@ struct deconv_fwd_executable_t : public op_executable_t {
                 auto prim = dnnl::reorder(psrc_mem, dst_mem);
                 auto e = dnnl::sycl_interop::execute(prim, stream,
                         {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                {DNNL_ARG_TO, const_cast<memory &>(dst_mem)}},
+                                { DNNL_ARG_TO,
+                                    const_cast<memory &>(dst_mem) }},
                         sycl_deps);
                 sycl_deps = {e};
             }
@@ -649,7 +650,8 @@ struct deconv_fwd_executable_t : public op_executable_t {
                 auto prim = dnnl::reorder(psrc_mem, dst_mem);
                 auto e = dnnl::ocl_interop::execute(prim, stream,
                         {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                {DNNL_ARG_TO, const_cast<memory &>(dst_mem)}},
+                                { DNNL_ARG_TO,
+                                    const_cast<memory &>(dst_mem) }},
                         ocl_deps);
                 ocl_deps = {e};
             }
@@ -827,7 +829,8 @@ struct matmul_executable_t : public op_executable_t {
                 auto prim = dnnl::reorder(psrc_mem, dst_mem);
                 auto e = dnnl::sycl_interop::execute(prim, stream,
                         {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                {DNNL_ARG_TO, const_cast<memory &>(dst_mem)}},
+                                { DNNL_ARG_TO,
+                                    const_cast<memory &>(dst_mem) }},
                         sycl_deps);
                 sycl_deps = {e};
             }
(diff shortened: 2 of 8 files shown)


src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1643,6 +1643,23 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
 
     // 1. sqrt_variance = sqrt(variance + epsilon)
 
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    // binary + sqrt post-op fusion is unsupported on NVIDIA GPU
+    memory::dims epsilon_dims(variance.get_ndims(), 1);
+    desc.epsilon_desc_ = memory::desc(
+            epsilon_dims, memory::data_type::f32, memory::format_tag::a);
```
TaoLv, line 1651: this piece of code seems the same as in the else branch. Can we move it above the runtime and vendor check?

src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1714,7 +1732,12 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
     memory::dims scratchpad_dims = variance.get_dims();
     // sqrt_variance, zero_bias and others (like epsilon),
     // or no need to alloc bias
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    size_t factor = bias.is_zero() ? 4 : 3;
```
TaoLv, line 1737: leave a comment please.

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -495,8 +495,8 @@ struct conv_fwd_executable_t : public op_executable_t {
                     auto prim = dnnl::reorder(psrc_mem, to_mem);
                     auto e = dnnl::sycl_interop::execute(prim, stream,
                             {{DNNL_ARG_FROM, const_cast<memory &>(psrc_mem)},
-                                    {DNNL_ARG_TO,
-                                            const_cast<memory &>(to_mem)}},
+                                    { DNNL_ARG_TO,
+                                        const_cast<memory &>(to_mem) }},
```
TaoLv, line 499: revert the changes please.

src/graph/backend/dnnl/op_executable.cpp
Jiexin-Zheng: Solved

src/graph/backend/dnnl/op_executable.hpp
Jiexin-Zheng: done

src/graph/backend/dnnl/op_executable.cpp
Jiexin-Zheng: done

Jiexin-Zheng:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -1478,6 +1491,44 @@ struct bn_folding_t : public op_executable_t {
...
+        if (scratchpad.get_engine().get_kind() == engine::kind::gpu) {
+
+            buf_start += epsilon_mem.get_desc().get_size();
+
+            // variance + epsilon
+            memory variance_epsilon = make_dnnl_memory(desc_.epsilon_desc_,
+                    scratchpad.get_engine(), (void *)buf_start);
+
+            // 1. sqrt_variance = sqrt(variance + epsilon)
+            engine cpu_eng(engine::kind::cpu, 0);
```
rongzha1, line 1505: could you explain why to use CPU engine in GPU? When CPU_RUNTIME == NONE, can reorder still work correctly ?

Jiexin-Zheng:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/utils.cpp
```diff
@@ -409,6 +409,21 @@ bool post_binary_fusible(
...
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    if (static_cast<dnnl::algorithm>(
+                elt_op->get_attr<int64_t>(op_attr::alg_kind))
+                    == dnnl::algorithm::eltwise_sqrt
+            && ekind == dnnl_gpu) {
+        return false;
+    }
+#endif
+    return true;
+}
```
xiang1guo, line 425: Seems `base_op` is not used? May need to check `base_op_kind` here and rename the function name to indicate that?

tests/gtests/graph/unit/backend/dnnl/test_binary_op.cpp
```diff
@@ -317,6 +317,59 @@ TEST(test_binary_op_execute, MinEltwise) {
     }
 }
 
+TEST(test_binary_op_execute, BinarySqrt) {
+    graph::engine_t *eng = get_engine();
```
xiang1guo, line 321: What's the point of adding this test case? Seems no special check for NV GPU? I noticed that we already have binary+sqrt post ops check in [here](https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/graph/unit/backend/dnnl/test_binary_op.cpp#L587).

src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1648,6 +1648,32 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
     desc.epsilon_desc_ = memory::desc(
             epsilon_dims, memory::data_type::f32, memory::format_tag::a);
 
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    // binary + sqrt post-op fusion is unsupported on NVIDIA GPU
+    if (p_engine.get_kind() == dnnl::engine::kind::gpu) {
```
TaoLv, line 1654: I feel that the gppu vendor check should be moved into this if-branch.

src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1714,7 +1741,19 @@ bn_folding_t::desc_t bn_folding_t::create_desc(std::shared_ptr<op_t> &op,
     memory::dims scratchpad_dims = variance.get_dims();
     // sqrt_variance, zero_bias and others (like epsilon),
     // or no need to alloc bias
+    // binary + sqrt post-op fusion is unsupported on NVIDIA GPU, so we need
+    // one more scratchpad for sqrt
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    size_t factor = 0;
+    if (p_engine.get_kind() == dnnl::engine::kind::gpu) {
```
TaoLv, line 1749: Similar here.

src/graph/backend/dnnl/passes/utils.cpp
TaoLv: If it's not NV GPU, there function will be preprocessed as below and unused variables will be raised by compilers.

```cpp
bool post_eltwise_fusible(
        const op_t *base_op, const op_t *elt_op, graph::engine_kind_t ekind) {
    return true;
}
```

tests/gtests/graph/unit/CMakeLists.txt
```diff
@@ -93,3 +93,9 @@ foreach(i RANGE ${count})
         endif()
     endif()
 endforeach()
+
+if(DNNL_WITH_SYCL)
+    if(DNNL_SYCL_CUDA OR (DNNL_SYCL_GENERIC AND NVIDIA_TARGET_SUPPORTED))
+        append(CMAKE_CXX_FLAGS "-fsycl-targets=nvptx64-nvidia-cuda,spir64")
+    endif()
+endif()
```
TaoLv, line 101: Please add a newline here.

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Yes, you are right, I should check the base op kind, but I think it's not necessary to rename the function name as this function targets on all the checks, other check similar to this can be put inside this func.

tests/gtests/graph/unit/backend/dnnl/test_binary_op.cpp
Jiexin-Zheng: I didn't add special check for NV GPU as this case should work for all the platforms, and the case you referred has three post-ops, from my perspective, that case is different from this one.

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Does this work for you?
```cpp   
// Suppress unused variable
    (void)base_op;
    (void)elt_op;
    (void)ekind;
    return true;
```
Or do you have better suggestion?

TaoLv: There is a macro `UNUSED` in the library serving the purpose.

tests/gtests/graph/unit/CMakeLists.txt
Jiexin-Zheng: Done

src/graph/backend/dnnl/op_executable.cpp
Jiexin-Zheng: Done

src/graph/backend/dnnl/op_executable.cpp
Jiexin-Zheng: Done

src/graph/backend/dnnl/passes/utils.cpp
Jiexin-Zheng: Done.

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -1481,21 +1494,53 @@ struct bn_folding_t : public op_executable_t {
...
                 scratchpad.get_engine(), (void *)buf_start);
 
-        // 1. sqrt_variance = sqrt(variance + epsilon)
-        if (variance.get_engine().get_kind() == engine::kind::cpu) {
-            float *ptr = (float *)epsilon_mem.get_data_handle();
-            *ptr = desc_.epsilon_;
-        } else {
+        if (scratchpad.get_engine().get_kind() == engine::kind::gpu) {
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
```
TaoLv, line 1499: Why do we need to handle this here? I support for NVIDIA GPU, execution will run into the `execute_sycl()` function below?

src/graph/backend/dnnl/op_executable.hpp
Jiexin-Zheng: As mentioned in https://github.com/uxlfoundation/oneDNN/pull/2971#discussion_r2025248390, this PR will not change the logic here now, so this question is not related to this PR anymore, but we can still discuss about this offline. Anyway, thanks for the suggestion~

Jiexin-Zheng:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -2722,7 +2787,10 @@ struct sdpa_executable_t : public op_executable_t {
             const std::vector<::sycl::event> &deps) const override {
 
 #if DNNL_GPU_VENDOR != DNNL_VENDOR_INTEL
-        return status::unimplemented;
+        assertm(false,
+                "genindex opexcutable is only implemented for intel vendor "
+                "under OCL runtime ");
+        throw std::runtime_error("Unimplement");
```
xiang1guo, line 2793: Not `GenIndex` and not `OCL` runtime

xiang1guo (approved):

Jiexin-Zheng:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

gyhintel (approved):

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -2787,7 +2787,10 @@ struct sdpa_executable_t : public op_executable_t {
             const std::vector<::sycl::event> &deps) const override {
 
 #if DNNL_GPU_VENDOR != DNNL_VENDOR_INTEL
-        return status::unimplemented;
+        assertm(false,
+                "sdpa opexcutable is only implemented for intel vendor under "
+                "SYCL runtime ");
+        throw std::runtime_error("Unimplement");
```
TaoLv, line 2793: why the change?

Jiexin-Zheng: The original statement `return status::unimplemented;` will return a `status` class, while function `::sycl::event execute_sycl()` requires a `sycl::event` class, which causes build issue, so I just adopt an exception throw following the way at https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/op_executable.hpp#L2575

rongzha1 (approved):

src/graph/backend/dnnl/op_executable.hpp
TaoLv: We need to clean these eventually. @gyhintel

gyhintel: For this question, 
1. If we don't have the implementation for other vendors. I think we should return `unimplemented status` in the compilation stage.
2. If we plan to implement it with other vendors. We need to consider giving a sycl generic implementation.

TaoLv: I agree. We need to either do the check at the higher level in advance to avoiding reaching here, or revise this function to return a status code. For the latter one, we can do it in another PR as I see it will touch many places.

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -2742,18 +2810,19 @@ struct sdpa_executable_t : public op_executable_t {
...
 
-        if (!deps.empty()) sycl_stream->sycl_ctx().set_deps(deps);
+        if (!deps.empty()) sycl_stream_impl->sycl_ctx().set_deps(deps);
 
         sdpa_prim_->execute(ctx);
 
-        auto return_event = sycl_stream->get_output_event();
-        sycl_stream->after_exec_hook();
+        ::sycl::event return_event = sycl_stream_impl->get_output_event();
+        strm_t->after_exec_hook();
```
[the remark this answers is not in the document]
Jiexin-Zheng: `dnnl::impl::utils::downcast<sycl::stream_t *>` is only available for intel vendor, we need a more general implementation for other platforms, that is why I used `dnnl::impl::xpu::sycl::stream_impl_t`, or we will encounter build issue, this following the way at https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/op_executable.hpp#L2539

src/graph/backend/dnnl/op_executable.hpp
Jiexin-Zheng: Will create another PR to fix this after all the gtest and benchdnn failures are solved, since we can have a better design for it at that time.

TaoLv: Last question for this PR - do you see any issue if we remove the vendor check here? I assumed that if it's NV GPU, it should not reach this `sdpa_executable_t` at all.

Jiexin-Zheng: Did you mean remove the whole macro and assertion here? If so, I didn't see any issue for now, we can remove them if necessary.

TaoLv: Yes, please.

Jiexin-Zheng: Done.

PR #2979, opened 2025-03-29
graph: interface: op: softmax to support inf_as_zero mode

TaoLv:
This is a draft API implementation for RFC https://github.com/uxlfoundation/oneDNN/pull/2932 .

~CPU/GPU kernels to support safe softmax are not included yet.~

CPU/GPU reference path are implemented in this PR.

Fixes MFDNN-13295.

### DIFF
--- a/doc/graph/operations/Softmax.md
+++ b/doc/graph/operations/Softmax.md
@@ -14,6 +14,13 @@ where \f$ C \f$ is a size of tensor along axis dimension.
 | Attribute Name                           | Description                                               | Value Type | Supported Values                     | Required or Optional |
 |:-----------------------------------------|:----------------------------------------------------------|:-----------|:-------------------------------------|:---------------------|
 | [axis](@ref dnnl::graph::op::attr::axis) | Represents the axis from which the SoftMax is calculated. | s64        | Arbitrary s64 value (`1` in default) | Optional             |
+| [mode](@ref dnnl::graph::op::attr::mode) | Specifies the computation mode of SoftMax                 | string     | `none` (default), `inf_as_zero`      | Optional             |
+
+When the operation attribute `mode` is not set or set to `none`, the operation
+performs the normal SoftMax calculation. In this case, the operation will
+generate `NaN` if all the input elements are `-infinity` along the `axis`
+dimension. To prevent this, you can set the attribute to `inf_as_zero` so that
+the operation generates zeros for `-infinity` inputs. 
 
 ## Execution arguments
 
--- a/examples/graph/sdpa.cpp
+++ b/examples/graph/sdpa.cpp
@@ -319,6 +319,7 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     auto probs = logical_tensor(id++, dt, score_sz, layout_type::strided);
     auto softmax = op(id++, op::kind::SoftMax, "softmax");
     softmax.set_attr<int64_t>(op::attr::axis, -1);
+    softmax.set_attr<std::string>(op::attr::mode, "inf_as_zero");
     softmax.add_inputs({masked_score});
     softmax.add_outputs({probs});
 
--- a/scripts/generate_dnnl_debug.py
+++ b/scripts/generate_dnnl_debug.py
@@ -35,8 +35,9 @@ def template(body, banner):
 %s""" % (
         banner,
         os.path.basename(__file__),
-        body
+        body,
     )
+
 
 def header(body):
     return (
@@ -232,8 +233,11 @@ def func_to_str(enum, values):
     func += func_to_str_decl(enum) + " {\n"
     for v in values:
         func += '%sif (v == %s) return "%s";\n' % (indent, v, sanitize_value(v))
-    if (enum == "dnnl_primitive_kind_t"):
-        func += '%sif (v == dnnl::impl::primitive_kind::sdpa) return "sdpa";\n' % indent
+    if enum == "dnnl_primitive_kind_t":
+        func += (
+            '%sif (v == dnnl::impl::primitive_kind::sdpa) return "sdpa";\n'
+            % indent
+        )
     func += '%sassert(!"unknown %s");\n' % (indent, abbrev)
     func += '%sreturn "unknown %s";\n}\n' % (indent, abbrev)
     return func
@@ -301,12 +305,18 @@ def generate(ifile, banners):
         enum = v_enum.attrib["name"]
         if maybe_skip(enum):
             continue
-        values = [v_value.attrib["name"] for v_value in v_enum.findall("EnumValue")]
+        values = [
+            v_value.attrib["name"] for v_value in v_enum.findall("EnumValue")
+        ]
 
         h_body += func_to_str_decl(enum, is_header=True) + ";\n"
         s_body += func_to_str(enum, values) + "\n"
 
-        if enum in ["dnnl_format_tag_t", "dnnl_data_type_t", "dnnl_sparse_encoding_t"]:
+        if enum in [
+            "dnnl_format_tag_t",
+            "dnnl_data_type_t",
+            "dnnl_sparse_encoding_t",
+        ]:
             h_benchdnn_body += (
                 str_to_func_decl(enum, is_header=True, is_dnnl=False) + ";\n"
             )
@@ -336,6 +346,7 @@ $ castxml --castxml-cc-gnu-c clang --castxml-output=1 \\
     )
     sys.exit(1)
 
+
 for arg in sys.argv:
     if "-help" in arg:
         usage()
@@ -354,8 +365,8 @@ file_paths = (
 banners = []
 for file_path in file_paths:
     with open(file_path, "r") as f:
-        m = re.match(r'^/\*+\n(\*.*\n)+\*+/\n', f.read())
-        banners.append('' if m == None else m.group(0))
+        m = re.match(r"^/\*+\n(\*.*\n)+\*+/\n", f.read())
+        banners.append("" if m == None else m.group(0))
 
 for file_path, file_body in zip(file_paths, generate(ifile, banners)):
     with open(file_path, "w") as f:
--- a/scripts/generate_dnnl_debug.py
+++ b/scripts/generate_dnnl_debug.py
@@ -238,6 +238,11 @@ def func_to_str(enum, values):
             '%sif (v == dnnl::impl::primitive_kind::sdpa) return "sdpa";\n'
             % indent
         )
+    if enum == "dnnl_alg_kind_t":
+        func += (
+            '%sif (v == dnnl::impl::alg_kind::softmax_accurate_inf_as_zero) return "softmax_accurate_inf_as_zero";\n'
+            % indent
+        )
     func += '%sassert(!"unknown %s");\n' % (indent, abbrev)
     func += '%sreturn "unknown %s";\n}\n' % (indent, abbrev)
     return func
--- a/src/graph/backend/dnnl/dnnl_op_def.hpp
+++ b/src/graph/backend/dnnl/dnnl_op_def.hpp
@@ -1002,6 +1002,8 @@ DNNL_GRAPH_OP_SCHEMA(dnnl_softmax, 1,
                 .set_output(1, "scratchpad")
                 // Attributes inherited from SoftMax
                 .set_attr(op_attr::axis, false, attribute_kind::i, (int64_t)1)
+                .set_attr(op_attr::mode, false, attribute_kind::s, "none",
+                        {"none", "inf_as_zero"})
                 // New added attributes
                 .SET_ATTR_IS_CONSTANT // used for constant prop and cache
                 .set_attr(op_attr::fusion_info_key, false, attribute_kind::i,
--- a/src/graph/backend/dnnl/dnnl_op_def.hpp
+++ b/src/graph/backend/dnnl/dnnl_op_def.hpp
@@ -1177,6 +1177,7 @@ DNNL_GRAPH_OP_SCHEMA(dnnl_sdpa, 1,
                 // mask_type attribute indicates existence of explicit mask,
                 // top-left implicit causal mask or bottm-right implicit causal mask
                 .set_attr(op_attr::mask_type, true, attribute_kind::i)
+                .set_attr(op_attr::mode, true, attribute_kind::s)
                 .set_shape_inference_function(infer_dnnl_sdpa_output_shape)
                 .SET_LAYOUT_PROPAGATOR(layout_propagator_for_sdpa)
                 .SET_EXECUTABLE_CREATOR(executable_creator<sdpa_executable_t>)
--- a/src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
@@ -181,9 +181,14 @@ status_t mqa_decomp_config_t::construct_params(std::shared_ptr<subgraph_t> &sg,
     dnnl::primitive_attr sub_softmax_attr
             = make_primitive_attr(original_softmax, mgr);
     sub_softmax_dst_md = memory::desc(sub_mm1_dst_dims, dt_src_user, tag::abc);
+    const auto mode = mqa_op[2]->get_attr<std::string>(op_attr::mode);
+    const dnnl::algorithm algo = mode == "inf_as_zero"
+            ? static_cast<dnnl::algorithm>(
+                    dnnl::impl::alg_kind::softmax_accurate_inf_as_zero)
+            : dnnl::algorithm::softmax_accurate;
     auto sub_softmax_pd = softmax_forward::primitive_desc(p_engine,
-            prop_kind::forward_inference, algorithm::softmax_accurate,
-            sub_mm1_dst_md, sub_softmax_dst_md, sub_mm1_dst_md.get_ndims() - 1,
+            prop_kind::forward_inference, algo, sub_mm1_dst_md,
+            sub_softmax_dst_md, sub_mm1_dst_md.get_ndims() - 1,
             sub_softmax_attr);
     sub_softmax_prim = softmax_forward(sub_softmax_pd);
 
--- a/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
@@ -230,9 +230,14 @@ impl::status_t sdp_decomp_config_t::construct_params(
     // create softmax primitive attr
     dnnl::primitive_attr sub_softmax_attr = make_primitive_attr(sdp_op[2], mgr);
     sub_softmax_dst_md = memory::desc(sub_mm1_dst_dims, dt_src_user, tag::abcd);
+    const auto mode = sdp_op[2]->get_attr<std::string>(op_attr::mode);
+    const dnnl::algorithm algo = mode == "inf_as_zero"
+            ? static_cast<dnnl::algorithm>(
+                    dnnl::impl::alg_kind::softmax_accurate_inf_as_zero)
+            : dnnl::algorithm::softmax_accurate;
     auto sub_softmax_pd = softmax_forward::primitive_desc(p_engine,
-            prop_kind::forward_inference, algorithm::softmax_accurate,
-            sub_mm1_dst_md, sub_softmax_dst_md, sub_mm1_dst_md.get_ndims() - 1,
+            prop_kind::forward_inference, algo, sub_mm1_dst_md,
+            sub_softmax_dst_md, sub_mm1_dst_md.get_ndims() - 1,
             sub_softmax_attr);
     sub_softmax_prim = softmax_forward(sub_softmax_pd);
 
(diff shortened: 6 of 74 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

examples/graph/sdpa.cpp
```diff
@@ -319,6 +319,7 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     auto probs = logical_tensor(id++, dt, score_sz, layout_type::strided);
     auto softmax = op(id++, op::kind::SoftMax, "softmax");
     softmax.set_attr<int64_t>(op::attr::axis, -1);
+    softmax.set_attr<std::string>(op::attr::mode, "inf_as_zero");
```
Jiexin-Zheng, line 322: Is this mode always required for sdpa? If so, from my perspective, it's better to add some comments here.

TaoLv: It's not required for softmax or sdpa. I think PyTorch integration requires it to match PyTorch SDPA's behavior. I suppose users will check op document to understand the meaning of the attributes.

Jiexin-Zheng (approved):

src/graph/backend/dnnl/dnnl_op_def.hpp
```diff
@@ -988,6 +988,8 @@ DNNL_GRAPH_OP_SCHEMA(dnnl_softmax, 1,
                 .set_output(1, "scratchpad")
                 // Attributes inherited from SoftMax
                 .set_attr(op_attr::axis, false, attribute_kind::i, (int64_t)1)
+                .set_attr(op_attr::mode, false, attribute_kind::s, "none",
+                        {"none", "inf_as_zero"})
```
ElaineBao, line 992: Do you need to add this mode attr to internal dnnl_sdpa op also?

TaoLv: Yes, i guess so. I will add it along with GPU ukernel enabling. We will need a way to pass the flag along to the kernel in backend.

doc/graph/operations/Softmax.md
```diff
@@ -14,6 +14,13 @@ where \f$ C \f$ is a size of tensor along axis dimension.
 | Attribute Name                           | Description                                               | Value Type | Supported Values                     | Required or Optional |
 |:-----------------------------------------|:----------------------------------------------------------|:-----------|:-------------------------------------|:---------------------|
 | [axis](@ref dnnl::graph::op::attr::axis) | Represents the axis from which the SoftMax is calculated. | s64        | Arbitrary s64 value (`1` in default) | Optional             |
+| [mode](@ref dnnl::graph::op::attr::mode) | Spcifies the computation mode of SoftMax                  | string     | `none` (default), `inf_as_zero`      | Optional             |
+
+When the operation attribute `mode` is not set or set to `none`, the operation
+performs the normal SoftMax calculation. In this case, the operation will
+generate `NaN`s if all the input elements are `-infinity`s along the `axis`
+dimension. To prevent from that, the user can set the attribute to `inf_as_zero`, the
+operation will generate zeros for `-infinity` inputs.
```
ranukund, line 23:
```suggestion
operation generates zeros for `-infinity` inputs. 
```

ranukund (approved):
Edits suggested, please incorporate as you see fit, thanks!

ElaineBao (approved):

src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1377,15 +1377,21 @@ softmax_executable_t::desc_t softmax_executable_t::create_desc(
...
 
-    const dnnl::algorithm algo
+    impl::alg_kind_t algo
             = op->get_kind() == dnnl_impl::op_kind::dnnl_logsoftmax
-            ? dnnl::algorithm::softmax_log
-            : dnnl::algorithm::softmax_accurate;
+            ? impl::alg_kind::softmax_log
+            : impl::alg_kind::softmax_accurate;
+
+    auto mode = op->get_attr<std::string>(op_attr::mode);
```
ElaineBao, line 1385: for dnnl_logsoftmax, there's no attribute of mode?

TaoLv: Thank you. I refactored the logic a bit here. Now we only get the attribute from dnnl_softmax op.

src/graph/backend/dnnl/op_executable.cpp
```diff
@@ -1377,15 +1377,23 @@ softmax_executable_t::desc_t softmax_executable_t::create_desc(
...
+    }
 
-    dnnl::softmax_forward::primitive_desc pd;
-    pd = dnnl::softmax_forward::primitive_desc(p_engine,
-            prop_kind::forward_inference, algo, src, dst,
-            static_cast<int>(axis), prm_attr);
+    dnnl_primitive_desc_t softmax_pd = nullptr;
+    dnnl_softmax_forward_primitive_desc_create(&softmax_pd, p_engine.get(),
+            impl::prop_kind::forward_inference, algo, src.get(), dst.get(),
+            static_cast<int>(axis), prm_attr.get());
```
ElaineBao, line 1395: Is it possible to use c++ api to create primitive_desc?

doc/graph/operations/Softmax.md
```diff
@@ -14,6 +14,13 @@ where \f$ C \f$ is a size of tensor along axis dimension.
 | Attribute Name                           | Description                                               | Value Type | Supported Values                     | Required or Optional |
 |:-----------------------------------------|:----------------------------------------------------------|:-----------|:-------------------------------------|:---------------------|
 | [axis](@ref dnnl::graph::op::attr::axis) | Represents the axis from which the SoftMax is calculated. | s64        | Arbitrary s64 value (`1` in default) | Optional             |
+| [mode](@ref dnnl::graph::op::attr::mode) | Spcifies the computation mode of SoftMax                  | string     | `none` (default), `inf_as_zero`      | Optional             |
```
xiang1guo, line 17:
```suggestion
| [mode](@ref dnnl::graph::op::attr::mode) | Specifies the computation mode of SoftMax                  | string     | `none` (default), `inf_as_zero`      | Optional             |
```

src/graph/backend/dnnl/op_executable.cpp
TaoLv: No, it's not possible. The new `softmax_accurate_inf_safe` algorithm is not part of API. So I switched to the internal interface and types to create primitive descriptor and then wrap it into a C++ object

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -204,10 +204,16 @@ impl::status_t sdp_decomp_config_t::construct_params(
...
-            sub_softmax_attr);
+    const auto mode = sdp_op[2]->get_attr<std::string>(op_attr::mode);
+    const impl::alg_kind_t algo = mode == "inf_as_zero"
+            ? impl::alg_kind::softmax_accurate_inf_as_zero
+            : impl::alg_kind::softmax_accurate;
+    dnnl_primitive_desc_t softmax_pd = nullptr;
+    dnnl_softmax_forward_primitive_desc_create(&softmax_pd, p_engine.get(),
+            impl::prop_kind::forward_inference, algo, sub_mm1_dst_md.get(),
+            sub_softmax_dst_md.get(), sub_mm1_dst_md.get_ndims() - 1,
+            sub_softmax_attr.get());
```
xiang1guo, line 215: Do we need to check the returned status of this func?

src/graph/backend/dnnl/op_executable.cpp
dzarukin: It is, I just forgot to update C++ counterpart (feel free to add it into the corresponding commit).

wzt1997:
Some updates of benchdnn graph have been committed to support better validation for softmax:
1. Create reference softmax primitive with `inf_as_zero` algorithm.
2. Change data filling for explicit casual mask that always fill -`INF` for the very last row.
3. Tolerate `nan` outputs for softmax without `inf_as_zero` attribute.
4. Skip `nan` values in the normalization check.

tests/benchdnn/utils/compare.cpp
```diff
@@ -198,10 +198,9 @@ int compare_t::compare_norm(const dnn_mem_t &exp_mem, const dnn_mem_t &got_mem,
         static thread_local diff_norm_t diff_norm_ithr;
         driver_check_func_args_t args(exp_mem, got_f32, i, dt, trh_);
 
-        if ((std::isnan(args.exp_f32) && is_integral_dt(dt))
-                || std::isinf(args.exp)) {
-            // Don't include integer max values or inf values into norm as they
-            // make it irrelevant for validation.
+        if ((std::isnan(args.exp_f32)) || std::isinf(args.exp)) {
```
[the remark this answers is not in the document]
wzt1997: It's to tolerate `nan` results for the validation of softmax without `inf_as_zero`. The correctness check might fall back to norm check, and as the results are partially `nan`, we should not take them into consideration. 
According to @xiang1guo, The reason why we have `is_integral_dt(dt)` here, is probably that we have only `nan` mixed with normal results for integral cases before, and it's acceptable to change the condition if we see more cases.

doc/graph/operations/Softmax.md
TaoLv: Fixed it. Nice catch.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
TaoLv: Now it's changed back to C++ API for primitive descriptor creation.

ElaineBao (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -550,6 +550,7 @@ int partition_data_displacer_t::gen_causal_mask_filling(
     benchdnn_parallel_nd(batch, M, N, [&](int64_t b, int64_t m, int64_t n) {
         int64_t idx = b * M * N + m * N + n;
         float val = m >= n ? 0.f : -INFINITY;
+        if (m == M - 1) val = -INFINITY;
```
dzarukin, line 553:
```suggestion
        // The line below masks out the whole prompt to verify the softmax output returns zeroes, not NaNs, as expected by PyTorch.
        if (m == M - 1) val = -INFINITY;
```

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -343,7 +344,13 @@ int ref_partition_t::check_partition_correctness(
                 // into NaNs.
                 || (op_driver == dnnl_driver_t::reorder
                         && op.get().in_lts_.front().get_data_type()
-                                == logical_tensor::data_type::f8_e4m3);
+                                == logical_tensor::data_type::f8_e4m3)
+                // Only safe softmax can tolerant input where the entire line
```
dzarukin, line 348:
```suggestion
                // Only softmax_inf_as_zero can tolerant input where the entire line
```

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -322,6 +322,7 @@ int ref_partition_t::check_partition_correctness(
         size_t op_id = op.get().id_;
         const auto op_kind = op.get().kind_;
         const auto ref_prim = ref_prims_.at(op_id);
+        std::string mode {};
```
dzarukin, line 325: std::string has default ctor.
```suggestion
        std::string mode;
```

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -368,6 +375,9 @@ int ref_partition_t::check_partition_correctness(
         // reset the state
         res->state = EXECUTED;
 
+        // TODO(zhitao): need to check whether the operation that produces the
+        // output args is the children of the operations that affect
+        // output_has_nans.
```
dzarukin, line 380: Why do you think is this needed? The flag is passed till the very end, if the parent expects nans, children might have them, too.

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -1943,7 +1943,15 @@ bool get_softmax_sdt_and_ddt(const deserialized_op_t &base_op_ref,
 bool get_softmax_alg(
         const deserialized_op_t &base_op_ref, ::softmax::alg_t &alg) {
     const auto &op_kind = base_op_ref.kind_;
-    if (op_kind == "SoftMax" || op_kind == "SoftMaxBackward") {
+    if (op_kind == "SoftMax") {
+        alg = ::softmax::alg_t::SOFTMAX;
+
+        std::string mode {};
```
dzarukin, line 1949: Ditto.

tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
```diff
@@ -136,3 +135,7 @@
 # d_qk != d_v
 --reset --dt=f32,bf16,f16 --in-shapes=8:1x16x384x32,8:1x16x384x64,8:1x16x384x128 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json
 --reset --dt=f32,bf16,f16 --in-shapes=3:1x16x384x32,3:1x16x384x64,3:1x16x384x128 --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
+
+# safe softmax
+--reset --dt=f32,bf16,f16 --op-attrs=3:mode:inf_as_zero --case=complex_fusion/mha/sdpa-plain-simplified-f16.json
```
dzarukin, line 140: We probably need to make this mode a default in all sdpa-related graphs as this is the end goal. I'm fine to have that change separately.

tests/benchdnn/utils/compare.cpp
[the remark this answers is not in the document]
dzarukin: But do we actually have those cases? The only filling change to catch it targets sdpa and it's supposed to be safe from this effect.

tests/benchdnn/graph/ref_partition.cpp
wzt1997: My concern is that, take the following pattern for example:
```
/*
                |
         _____MatMul_______
        /                  \
        |                  |
       SQRT              ReLU
        |                  |
*/
```
We should allow `nan`s from the branch of `Sqrt` as designed now, but for the other branch, the driver should not tolerant that.  As designed now,  `output_has_nans` will be set true once we have seen some special operations, while it can lead to missing some bug on other branches.
As we have not seen such cases, I just left a comment here and may fix it in the future if necessary.

tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
wzt1997: Got it. I think we can do it separately to make this PR simple for now.

tests/benchdnn/utils/compare.cpp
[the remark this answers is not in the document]
wzt1997: Yes, we have as we displace the last row with -`INF` now. For instance, some bf16 sdpa cases might fail in p2p check due to precision issue and relies on the norm check instead. If we don't tolerate `nan`s in the norm check, it will fail either. For example:
```
./tests/benchdnn/benchdnn --graph --dt=bf16 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json
[5934][0:0:92:46] exp_f32:    -4.79688 exp:     -4.8125 got:    -4.78125 diff: 0.03125 rdiff:0.00649351
[35079][0:1:164:7] exp_f32: 1.44724e-11 exp: 1.44951e-11 got: 1.45519e-11 diff:5.68434e-14 rdiff:0.00392157
[35081][0:1:164:9] exp_f32: 2.93658e-11 exp: 2.93312e-11 got: 2.91038e-11 diff:2.27374e-13 rdiff:0.00775194
[44610][0:1:313:2] exp_f32:   -0.162598 exp:   -0.163086 got:   -0.162109 diff:0.000976562 rdiff:0.00598802
[124072][0:5:18:40] exp_f32:-2.37022e-06 exp:-2.36928e-06 got:-2.26498e-06 diff:1.04308e-07 rdiff:0.0440252
[138171][0:5:238:59] exp_f32:     1.01953 exp:     1.02344 got:     1.01562 diff:0.0078125 rdiff:0.00763359
[143748][0:5:326:4] exp_f32:     5.01562 exp:           5 got:     5.03125 diff: 0.03125 rdiff: 0.00625
[172369][0:7:5:17] exp_f32:    -2.39844 exp:    -2.40625 got:    -2.39062 diff:0.015625 rdiff:0.00649351
[178025][0:7:93:41] exp_f32:     2.97656 exp:     2.96875 got:     2.98438 diff:0.015625 rdiff:0.00526316
[186338][0:7:223:34] exp_f32:    0.133301 exp:    0.133789 got:    0.132812 diff:0.000976562 rdiff:0.00729927
[COMPARE_STATS]: trh=0 err_max_diff: 0.03125 err_max_rdiff:0.164179 all_max_diff: 0.03125 all_max_rdiff: 0.00625
[COMPARE_STATS] Norm check is allowed; error_to_total_ratio: 19/393216; allowed_ratio: 384/393216;
[L0] = 0.164179
[L1] exp:     nan got:     nan diff:     nan rel_diff:     nan
[L2] exp:    -nan got:    -nan diff:    -nan rel_diff:    -nan
[L8] exp:       8 got:       8 diff: 0.03125 rel_diff:0.00390625
[COMPARE_STATS]: trh=0.0001 (compare against [L2] rel_diff)
Error: Function 'doit' at (/home/sdp/zhitaowa/oneDNN-src/tests/benchdnn/graph/graph.cpp:743) returned '1'
0:FAILED (errors:1 total:1) (241 ms) __REPRO: --graph --dt=bf16 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json
```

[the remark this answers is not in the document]
dzarukin: Wait, all sdpa cases must be moved to inf_as_zero softmax to match the behavior, it's not expected for it to work without the attribute, that's the only reason we are adding that last row with infs. This shouldn't be a driver.

tests/benchdnn/graph/ref_partition.cpp
dzarukin: Probably would be good to add this picture to the comment, I didn't figure it out just from the text. Thanks for the explanation.

wzt1997: OK. I have attached the graph in the TODO.

tests/benchdnn/utils/compare.cpp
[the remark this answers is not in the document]
wzt1997: I got your idea, here are some changes and the motivations:
1. All test cases are by default with `inf_as_zero` softmax to align with pytorch behavior.
2. As some  `inf_as_zero` softmax cases will fail with norm check( may be due to the whole line of `0`s ), we still need to ignore `nan`s during the norm check.
3. For softmax without `inf_as_zero`, benchdnn graph driver does not tolerant the `nan`s. If anyone requires such computation, we can see if we need to support the validation for it.

src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
```diff
@@ -180,8 +180,12 @@ status_t mqa_decomp_config_t::construct_params(std::shared_ptr<subgraph_t> &sg,
     dnnl::primitive_attr sub_softmax_attr
             = make_primitive_attr(original_softmax, mgr);
     sub_softmax_dst_md = memory::desc(sub_mm1_dst_dims, dt_src_user, tag::abc);
+    // const auto mode = mqa_op[2]->get_attr<std::string>(op_attr::mode);
+    // const dnnl::algorithm algo = mode == "inf_as_zero"
+    //         ? dnnl::algorithm::softmax_accurate_inf_as_zero
+    //         : dnnl::algorithm::softmax_accurate;
```
atkassen, line 186: Unneeded?

atkassen (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

densamoilov (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

dzarukin:
make test
disable benchdnn_all
enable benchdnn_graph

scripts/generate_dnnl_debug.py
```diff
@@ -354,8 +370,8 @@ def usage():
 banners = []
 for file_path in file_paths:
     with open(file_path, "r") as f:
-        m = re.match(r'^/\*+\n(\*.*\n)+\*+/\n', f.read())
-        banners.append('' if m == None else m.group(0))
+        m = re.match(r"^/\*+\n(\*.*\n)+\*+/\n", f.read())
+        banners.append("" if m == None else m.group(0))
```
atkassen, line 374: nit:
```suggestion
        banners.append("" if m is None else m.group(0))
```
or
```suggestion
        banners.append("" if m else m.group(0))
```

dzarukin: Thanks for the suggestions. I'll keep those for the next PR to main only, this was just black formatting.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_graph

PR #2980, opened 2025-03-31
benchdnn: graph: support input displace for scalar

wzt1997:
## General

To support bottom-right casual mask( [RFC ](https://github.com/uxlfoundation/oneDNN/blob/yixin/rfc/graph-api-bottom-right-causal-mask/rfcs/20250314-graph-api-bottom-right-causal-mask/README.md)), Graph API will introduce scalar logical tensor. In current SDPA scenarios, the scalar values are relevant with sequence lengths, hence it requires input displacement with shape information.

The work in the PR includes: 
1. Support scalar displacement for `Add`->`Sub`->`GE` operation chain.
2. Resolve the TODO( [link ](https://github.com/uxlfoundation/oneDNN/blob/633978a82ec14ccc01a9cccaf173e89c25cbbc89/tests/benchdnn/graph/input_displacer.hpp#L37)) in input displacer by moving `fill_cfg` into the displacement arguments.

### DIFF
--- a/tests/benchdnn/graph/input_displacer.cpp
+++ b/tests/benchdnn/graph/input_displacer.cpp
@@ -73,12 +73,24 @@ partition_data_displacer_t::partition_data_displacer_t(
                     lt = &parent_op->in_lts_[0]) {
                 parent_op = &dg_->get_op_by_out_lt(lt->id_);
                 if (parent_op->empty()) {
-                    if (aop.kind_ == "Divide" || aop.kind_ == "Multiply") {
+                    if (aop.kind_ == "Divide") {
+                        // Division has values > 1.f to reduce final values.
+                        static const std::vector<float> user_set {
+                                2.f, 4.f, 8.f};
                         // There's a special case for Divide, when second (user)
                         // input should be displaced with power-of-2 values.
-                        quantize_displace_.emplace(lt->id_,
-                                std::make_tuple(
-                                        aop, i, *lt, filling_type_t::pow2));
+                        displace_args_.emplace(lt->id_,
+                                displace_args_t {aop, i, *lt,
+                                        filling_type_t::fixed_setting,
+                                        {user_set, "Div displacer"}});
+                    } else if (aop.kind_ == "Multiply") {
+                        // Multiplication has values <= 1.f to reduce final values.
+                        static const std::vector<float> user_set {
+                                0.25f, 0.5f, 1.f};
+                        displace_args_.emplace(lt->id_,
+                                displace_args_t {aop, i, *lt,
+                                        filling_type_t::fixed_setting,
+                                        {user_set, "Mul displacer"}});
                     }
                     break;
                 }
@@ -105,13 +117,12 @@ partition_data_displacer_t::partition_data_displacer_t(
                                         == f8_main_op_kind.end()))
                             break;
 
-                        quantize_displace_.emplace(parent_op_in_lt.id_,
-                                std::make_tuple(aop, i, parent_op_in_lt,
-                                        filling_type_t::quantization));
+                        displace_args_.emplace(parent_op_in_lt.id_,
+                                displace_args_t {aop, i, parent_op_in_lt,
+                                        filling_type_t::quantization});
                         break;
                     }
-                }
-                if (parent_op->kind_ == "StaticReshape") {
+                } else if (parent_op->kind_ == "StaticReshape") {
                     // StaticReshape is accepted when the pattern is
                     // "StaticReshape + Matmul" and it doesn't have any
                     // predecessors in the partition
@@ -122,9 +133,9 @@ partition_data_displacer_t::partition_data_displacer_t(
                             || op_ids_set_.find(prev_parent_op.id_)
                                     == op_ids_set_.end()) {
                         if (aop.kind_ == "MatMul") {
-                            quantize_displace_.emplace(parent_op_in_lt.id_,
-                                    std::make_tuple(aop, i, parent_op_in_lt,
-                                            filling_type_t::quantization));
+                            displace_args_.emplace(parent_op_in_lt.id_,
+                                    displace_args_t {aop, i, parent_op_in_lt,
+                                            filling_type_t::quantization});
                         }
                         break;
                     }
@@ -203,6 +214,8 @@ partition_data_displacer_t::partition_data_displacer_t(
             if (!causal_mask_lt) break;
 
             filling_type_t filling_type = filling_type_t::undef;
+            std::string cfg_name;
+            float user_set_value = 0.f;
             if (aop.kind_ == "Add") {
                 const auto ndims = causal_mask_lt->shape_.size();
                 if (ndims < 2) {
@@ -219,7 +232,8 @@ partition_data_displacer_t::partition_data_displacer_t(
                     // are computed. To avoid numerical instabilities, a zero
                     // mask can be applied without compromising validation
                     // capabilities.
-                    filling_type = filling_type_t::zero;
+                    filling_type = filling_type_t::fixed_setting;
+                    cfg_name = "Explicit_padding_mask";
                 } else {
                     // This is a look-ahead (or causal) mask case, when future
                     // tokens (row < col) are set to infinity to remove all
@@ -229,17 +243,21 @@ partition_data_displacer_t::partition_data_displacer_t(
             } else if (aop.kind_ == "Select") {
                 if (select_cond_has_parent) {
                     // Implicit causal mask case.
-                    filling_type = filling_type_t::minus_infinity;
+                    filling_type = filling_type_t::fixed_setting;
+                    user_set_value = -INFINITY;
+                    cfg_name = "Implicit_causal_mask";
                 } else {
                     // Padding mask.
                     assert(qk_data_offset == 1 || qk_data_offset == 2);
                     // Fill condition depending on qk values tensor to use only
                     // its values, which is equivalent of not using a mask.
+                    filling_type = filling_type_t::fixed_setting;
                     if (qk_data_offset == 1) {
-                        filling_type = filling_type_t::one;
+                        user_set_value = 1.f;
                     } else if (qk_data_offset == 2) {
-                        filling_type = filling_type_t::zero;
+                        user_set_value = 0.f;
                     }
+                    cfg_name = "Explicit_padding_mask";
                 }
             }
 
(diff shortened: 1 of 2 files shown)


tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +246,91 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                                displace_args_.emplace(op->in_lts_[i].id_,
+                                        displace_args_t {*op, i, op->in_lts_[i],
+                                                filling_type_t::fixed_setting,
+                                                {user_set, cfg_name}});
+                            }
+                        }
+                    };
+
+            set_scalar_displace_args(&aop, SEQ_LEN_Q);
+            set_scalar_displace_args(child_op, SEQ_LEN_KV);
```
dzarukin, line 333: Please use constexpr int constants in here instead macro in a header file for better readability.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -264,16 +343,16 @@ int partition_data_displacer_t::displace_input_data(
...
-
-    const auto &main_op = ::std::get<0>(displace);
-    auto main_op_offset = ::std::get<1>(displace);
-    auto tensor = ::std::get<2>(displace);
-    const auto filling_type = ::std::get<3>(displace);
+    const displace_args_t &d_args = displace_args_.at(lt_id);
+    const auto &main_op = d_args.main_op_;
+    auto main_op_offset = d_args.main_op_offset_;
+    auto tensor = d_args.tensor_;
+    auto &fill_cfg = d_args.fill_cfg_;
```
dzarukin, line 354: Two members above seem to fit `const` category as well.
```suggestion
    const auto &fill_cfg = d_args.fill_cfg_;
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -203,6 +217,8 @@ partition_data_displacer_t::partition_data_displacer_t(
             if (!causal_mask_lt) break;
 
             filling_type_t filling_type = filling_type_t::undef;
+            std::string cfg_name {};
```
dzarukin, line 220:
```suggestion
            std::string cfg_name;
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +246,91 @@ partition_data_displacer_t::partition_data_displacer_t(
             } else if (aop.kind_ == "Select") {
                 if (select_cond_has_parent) {
                     // Implicit causal mask case.
-                    filling_type = filling_type_t::minus_infinity;
+                    filling_type = filling_type_t::fixed_setting;
+                    user_set_value = -1e30;
```
dzarukin, line 250: Why change to a specific value?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +246,91 @@ partition_data_displacer_t::partition_data_displacer_t(
...
             if (filling_type == filling_type_t::undef) {
                 BENCHDNN_PRINT(
                         7, "%s\n", "[DISPLACE]: Filling type was not set");
                 break;
+            } else if (filling_type == filling_type_t::fixed_setting) {
+                const std::vector<float> user_set {user_set_value};
+                displace_args_.emplace(causal_mask_lt->id_,
+                        displace_args_t {aop, offset, *causal_mask_lt,
+                                filling_type, {user_set, cfg_name}});
+            } else {
```
dzarukin, line 276: I'd do `else if` to avoid random errors.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +246,91 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                                const std::vector<float> user_set {
+                                        user_set_value};
+
+                                displace_args_.emplace(op->in_lts_[i].id_,
+                                        displace_args_t {*op, i, op->in_lts_[i],
+                                                filling_type_t::fixed_setting,
+                                                {user_set, cfg_name}});
+                            }
+                        }
+                    };
```
dzarukin, line 330: Is host-scalar only identified by the pattern and not by the property of Op?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +246,91 @@ partition_data_displacer_t::partition_data_displacer_t(
...
             }
 
-            quantize_displace_.emplace(causal_mask_lt->id_,
-                    std::make_tuple(
-                            aop, offset, *causal_mask_lt, filling_type));
+            break;
+        }
+
+        // Fill scalar for bottom-right implicit casual mask
+        if (aop.kind_ == "Add") {
```
ElaineBao, line 287: How does this condition co-work with the same check of Add's op kind in L222: https://github.com/uxlfoundation/oneDNN/blob/aa6722e9137ea1c4b4b62926ce4929dd6a34aa02/tests/benchdnn/graph/input_displacer.cpp#L222C15-L222C38

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -74,11 +74,26 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                        // Multiplication has values <= 1.f to reduce final values.
+                        static const std::vector<float> pow2_mul_vals {
+                                0.25f, 0.5f, 1.f};
+                        // Guard set.
+                        static const std::vector<float> dummy {};
+                        const bool is_div = aop.kind_ == "Divide";
+                        const bool is_mul = aop.kind_ == "Multiply";
+                        const auto &user_set = is_div
+                                ? pow2_div_vals
+                                : (is_mul ? pow2_mul_vals : dummy);
```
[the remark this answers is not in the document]
wzt1997: Good suggestion. I split them to make it more clear.

tests/benchdnn/graph/input_displacer.cpp
wzt1997: Thanks for reminding. The change is removed.

tests/benchdnn/graph/input_displacer.cpp
wzt1997: As we have a check in [L165](https://github.com/uxlfoundation/oneDNN/blob/aa6722e9137ea1c4b4b62926ce4929dd6a34aa02/tests/benchdnn/graph/input_displacer.cpp#L165) that handles the op chain `Add`/`Select`->`Softmax`. This part handles the input displacement for:
1. Explicit casual mask with Add/Softmax.
2. `-INF` filling for Softmax with implicit casual mask.
( NOTE: for explicit casual mask, the other input of Select( besides MM1 ouput ) will not be handled specially as MM1  output will always be chosen. )

While for bottom-right implicit casual mask, it contains an op chain `Add`->`Sub`->`GE`->`Select`, so they will not interfere with the above handling.

tests/benchdnn/graph/input_displacer.cpp
wzt1997: Changed. Thanks.

tests/benchdnn/graph/input_displacer.cpp
wzt1997: For current request, it can be identified in both ways. For checking with porperty, it can only be supported after Yixin's update for bottom-right implicit casual mask is merged.

ElaineBao: I think checking pattern here can cover two scenarios, first is s_kv and s_q are passed as host_scalar, second is s_kv and s_q are passed as device memory. @wzt1997 can you confirm the implementation applies to two scenarios?

wzt1997: Verified that checking through pattern can support both scenarios: with/without `host_scalar` property.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +243,92 @@ partition_data_displacer_t::partition_data_displacer_t(
...
-                    std::make_tuple(
-                            aop, offset, *causal_mask_lt, filling_type));
+            // The following subtract and greaterEqual must also be a part of
+            // the partition.
+            if (op_ids_set_.find(child_op->id_) == op_ids_set_.end()
+                    || op_ids_set_.find(next_child_op->id_)
+                            == op_ids_set_.end())
+                break;
+
+            const auto set_scalar_displace_args =
```
ElaineBao, line 303: nit: as this pattern doesn't necessarily have to be scalars, please change the variable name here.

wzt1997:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao (approved):

wzt1997:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +243,92 @@ partition_data_displacer_t::partition_data_displacer_t(
...
+                    cfg_name = "Explicit_padding_mask";
                 }
             }
 
             if (filling_type == filling_type_t::undef) {
                 BENCHDNN_PRINT(
                         7, "%s\n", "[DISPLACE]: Filling type was not set");
                 break;
+            } else if (filling_type == filling_type_t::fixed_setting) {
+                const std::vector<float> user_set {user_set_value};
```
dzarukin, line 269: It would be better to fill a vector above instead of a single value and pass it directly to the `emplace`.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -229,29 +243,92 @@ partition_data_displacer_t::partition_data_displacer_t(
...
-                            aop, offset, *causal_mask_lt, filling_type));
+            // The following subtract and greaterEqual must also be a part of
+            // the partition.
+            if (op_ids_set_.find(child_op->id_) == op_ids_set_.end()
+                    || op_ids_set_.find(next_child_op->id_)
+                            == op_ids_set_.end())
+                break;
+
+            const auto set_seq_len_displace_args =
+                    [&, this](const deserialized_op_t *op, int which_seq_len) {
```
dzarukin, line 304:
```suggestion
                    [&](const deserialized_op_t *op, int which_seq_len) {
```

dzarukin (approved):

wzt1997:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #2982, opened 2025-03-31
cpu: aarch64: Add BF16 support for eltwise JIT via reordering to FP32

aditew01:
Currently, PyTorch’s mapping to eltwise algorithms for `BF16` introduces multiple reorder operations:
reorder(BF16 → FP32) → Eltwise → reorder(FP32 → BF16). 
Relevant iDeep PR: https://github.com/intel/ideep/pull/350

This PR proposes moving the BF16-to-FP32 conversion directly into the JIT-generated code. By doing so, we eliminate the need for explicit reorder operations, resulting in performance improvements for these ops.

Benchdnn perf for Gelu ~ similar times for fp32 and bf16 equivalent. 

`OMP_NUM_THREADS=1 ./benchdnn --eltwise   --alg=gelu_erf --alpha=0 --beta=0 --mode=p --dt=bf16  1024x4096` 

Ref implementation: `total perf: min(ms):74.1855 avg(ms):74.2793` 
Current implementation:  `total perf: min(ms):6.7478 avg(ms):6.78948` **~12x speedup** 
FP32 run: `total perf: min(ms):6.71167 avg(ms): 6.7328` 

Speedup when benchmarked with PyTorch : 
Screenshot 2025-03-31 at 07 53 35

Tests: `./test/benchdnn/benchdnn --eltwise --batch=inputs/eltwise/test_eltwise_bfloat16` 

cc: @jondea @theComputeKid  @Sqvid

### DIFF
--- a/src/cpu/aarch64/jit_uni_eltwise.cpp
+++ b/src/cpu/aarch64/jit_uni_eltwise.cpp
@@ -119,14 +119,13 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
             // - unpack BF16 to FP32 by zero-extending
             // - compute eltwise alg in FP32
             // - down convert back to BF16 using bfcvt, and pack result
-            mov(ZReg(tmp0).s, P_ALL_ONE, ZReg(vmm_src).s);
-            lsl(ZReg(vmm_src.getIdx()).s, ZReg(vmm_src).s, 16);
-            and_(ZReg(tmp0).s, 0xFFFF0000); // 0xFFFF0000
+            mov(tmp0.s, P_ALL_ONE, vmm_src.s);
+            lsl(vmm_src.s, vmm_src.s, 16);
+            and_(tmp0.s, 0xFFFF0000);
             eltwise_injector_->compute_vector_range(
                     {vmm_src.getIdx(), tmp0.getIdx()});
-            bfcvt(ZReg(vmm_src.getIdx()).h, P_ALL_ONE,
-                    ZReg(vmm_src.getIdx()).s);
-            bfcvtnt(ZReg(vmm_src.getIdx()).h, P_ALL_ONE, ZReg(tmp0.getIdx()).s);
+            bfcvt(vmm_src.h, P_ALL_ONE, vmm_src.s);
+            bfcvtnt(vmm_src.h, P_ALL_ONE, tmp0.s);
         } else {
             eltwise_injector_->compute_vector(vmm_src.getIdx());
         }
--- a/src/cpu/aarch64/jit_uni_eltwise.cpp
+++ b/src/cpu/aarch64/jit_uni_eltwise.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2017-2022 Intel Corporation
 * Copyright 2021-2023 FUJITSU LIMITED
-* Copyright 2022 Arm Ltd. and affiliates
+* Copyright 2022, 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jondea:
Thanks for the contribution! Considering PyTorch is the right big picture view, but out of interest, what's the rough speedup for just oneDNN for BF16 eltwise? Are we going from ref to jit?

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -113,7 +113,26 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
...
         // perspective and will complicate the compute logic significantly.
         ldr(vmm_src, ptr(reg_src));
-        eltwise_injector_->compute_vector(vmm_src.getIdx());
+
+        if (is_bf16()) {
+            // Convert BF16 input to FP32, apply eltwise op, then convert back to BF16:
+            // - unpack BF16 to FP32 by zero-extending (zip with zero)
+            // - compute eltwise alg in FP32
+            // - down convert back to BF16 using bfcvt, and pack result
+            zip1(ZReg(tmp0.getIdx()).h, ZReg(vmm_zero.getIdx()).h,
```
jondea, line 122: I've had a think about this and I don't think we need zip/uzp in this case. We can just operate on odd/evens.

For unpacking,
- for odds, use https://dougallj.github.io/asil/doc/lsl_z_p_zi_32.html
- for the evens, use the src register but zero out the odds

Then https://dougallj.github.io/asil/doc/bfcvtnt_z_p_z_32.html should work.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -136,13 +155,30 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
...
+            // Convert BF16 input to FP32, apply eltwise op, then convert back to BF16
+            shll(VReg4S(v_bf16[0].getIdx()), VReg4H(v_bf16[0].getIdx()), 16);
+            eltwise_injector_->compute_vector(v_bf16.getIdx());
+            bfcvt(ZReg(v_bf16[0].getIdx()).h, P_ALL_ONE,
+                    ZReg(v_bf16[0].getIdx()).s);
+        } else {
+            eltwise_injector_->compute_vector(xmm_src.getIdx());
+        }
+
         if (!is_fwd) {
```
jondea, line 173: Have you tested bwd? If not, we can always just return unimplemented for this case

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -136,13 +155,30 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
...
-        ld1(xmm_src[0], ptr(reg_src));
-        eltwise_injector_->compute_vector(xmm_src.getIdx());
+        if (is_bf16())
+            ld1(v_bf16[0], ptr(reg_src));
+        else
+            ld1(xmm_src[0], ptr(reg_src));
+
+        if (is_bf16()) {
+            // Convert BF16 input to FP32, apply eltwise op, then convert back to BF16
+            shll(VReg4S(v_bf16[0].getIdx()), VReg4H(v_bf16[0].getIdx()), 16);
```
jondea, line 165: Can we avoid this getIdx and casting stuff by using a VReg directly and reinterpret with `.s` `.h` etc?

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -320,6 +360,7 @@ status_t jit_uni_eltwise_bwd_t<isa, d_type>::execute(
 
 template struct jit_uni_eltwise_fwd_t<sve_512, data_type::f32>;
 template struct jit_uni_eltwise_fwd_t<sve_256, data_type::f32>;
+template struct jit_uni_eltwise_fwd_t<sve_256, data_type::bf16>;
```
jondea, line 363: We should definitely test and enable for 128 and 512 if we are able to.

jondea (changes requested):
Thank you for your hard work on this. I know this part of the code base is unfamiliar to both of us so there has been a lot of learning, but we are definitely making progress!

src/cpu/aarch64/jit_uni_eltwise.cpp
aditew01: > LSL Zdn.S, Pg/M, Zdn.S, #const  - https://developer.arm.com/documentation/ddi0602/2025-03/SVE-Instructions/LSL--immediate--predicated---Logical-shift-left-by-immediate--predicated--?lang=en  

what I see from this instruction is not present in https://github.com/fujitsu/xbyak_aarch64/blob/main/src/xbyak_aarch64_mnemonic.h (maybe I missed it) .
we might use this -> `LSL Zdn.S, Zdn.S, #const ` instead  ?
 
>use the src register but zero out the odds 
(I'm using tmp1 and not the src for this) I'm using zip2 for this , can you please suggest if there's something more specific. 

if I get it correct: this is what you're suggesting: 

1. tmp0 -> left shift vmm_src , s.t., the odd position holds remaining values for compute. -> this needs zeroing as well right ? 
2. vmm_src -> zero out odds and compute the even vals. (use zip ?)
3. bfcvtnt -> tmp0 to vmm_src

src/cpu/aarch64/jit_uni_eltwise.cpp
aditew01: for BWD cases, currently it falls back to the the `ref` implementation for `dt==bf16`

aditew01:
> Thanks for the contribution! Considering PyTorch is the right big picture view, but out of interest, what's the rough speedup for just oneDNN for BF16 eltwise? Are we going from ref to jit?

Yes, we're mapping from ref -> jit. We're seeing ~10x speedup for larger shapes on 1 thread. I'll update the description with the numbers.

src/cpu/aarch64/jit_uni_eltwise.cpp
aditew01: done!

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -113,7 +113,24 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
...
         // perspective and will complicate the compute logic significantly.
         ldr(vmm_src, ptr(reg_src));
-        eltwise_injector_->compute_vector(vmm_src.getIdx());
+
+        if (is_bf16()) {
+            // Convert BF16 input to FP32, apply eltwise op, then convert back to BF16:
+            // - unpack BF16 to FP32 by zero-extending
+            // - compute eltwise alg in FP32
+            // - down convert back to BF16 using bfcvt, and pack result
+            mov(ZReg(tmp0).s, P_ALL_ONE, ZReg(vmm_src).s);
```
jondea, line 122: This looks great! Although do we need all the `ZReg` casts? `tmp0` and `vmm_src` already are `ZReg`s right?

aditew01: done!

jondea:
@rpushkarr you may be interested in this PR

jondea (approved):

spalicki (approved):

PR #2985, opened 2025-03-31
Update project version (3.8.0 to 3.9.0) in CMakeLists

vgvozdeva:
# Description

Update project version (3.8.0 to 3.9.0) in CMakeLists on main branch. PR will be merged right after creation of v3.8 release branch.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -18,4 +18,4 @@ abstract: >-
   oneDNN has experimental support for the following architectures: NVIDIA GPU,
   AMD GPU, OpenPOWER Power ISA (PPC64), IBMz (s390x), and RISC-V.
 license: Apache-2.0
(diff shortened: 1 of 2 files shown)


mgouicem (approved):

vpirogov (approved):

TaoLv:
Do we need to change the version number if the citation file when we bump the version of the library?
I see v3.7 in https://github.com/uxlfoundation/oneDNN/blob/main/CITATION.cff

vpirogov:
> Do we need to change the version number if the citation file when we bump the version of the library? I see v3.7 in https://github.com/uxlfoundation/oneDNN/blob/main/CITATION.cff

Good catch, Tao. Looks like I missed CITATION.cff last time...

PR #2992, opened 2025-04-01
cpu: aarch64: Update ACL reorder API

Ryo-not-rio:
This patch updates the ACL reorder to use the new API which  includes a flag for transposing or not transposing during the reorder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ On a CPU based on Arm AArch64 architecture, oneDNN CPU engine can be built with
 machine learning applications and provides AArch64 optimized implementations
 of core functions. This functionality currently requires that ACL is downloaded
 and built separately. See [Build from Source] section of the Developer Guide for
-details. oneDNN only supports Compute Library versions 24.11.1 or later.
+details. oneDNN only supports Compute Library versions 25.03.1 or later.
 
 [Arm Compute Library (ACL)]: https://github.com/arm-software/ComputeLibrary
 
--- a/cmake/ACL.cmake
+++ b/cmake/ACL.cmake
@@ -1,5 +1,5 @@
 # ******************************************************************************
-# Copyright 2020-2024 Arm Limited and affiliates.
+# Copyright 2020-2025 Arm Limited and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -31,7 +31,7 @@ endif()
 
 find_package(ACL REQUIRED)
 
-set(ACL_MINIMUM_VERSION "24.11.1")
+set(ACL_MINIMUM_VERSION "25.03.1")
 
 if(ACL_FOUND)
     file(GLOB_RECURSE ACL_VERSION_FILE ${ACL_INCLUDE_DIR}/*/arm_compute_version.embed)
--- a/src/cpu/aarch64/acl_reorder.hpp
+++ b/src/cpu/aarch64/acl_reorder.hpp
@@ -22,6 +22,28 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
+    int do_transpose(const dnnl::impl::memory_desc_t* src_md,const dnnl::impl::memory_desc_t* dst_md) {
+        auto src_md_ = dnnl::impl::md2fmt_tag_str(src_md);
+        auto dst_md_ = dnnl::impl::md2fmt_tag_str(dst_md);
+
+        bool equal = true;
+        for (int i = 0; i < src_md->ndims; i++) {
+            equal &= tolower(src_md_[i]) == tolower(dst_md_[i]);
+        }
+        if (equal) return 0;
+
+        equal = true;
+        for (int i = 0; i < src_md->ndims - 1; i++) {
+            equal &= tolower(src_md_[i]) == tolower(dst_md_[i + 1]);
+        }
+        equal &= tolower(src_md_[src_md->ndims - 1]) == tolower(dst_md_[0]);
+        if (equal) return 1;
+
+        return -1;
+    }
+}
+
 namespace dnnl {
 namespace impl {
 namespace cpu {
@@ -40,6 +62,7 @@ struct acl_reorder_conf_t {
     arm_compute::TensorInfo dst_info;
     arm_compute::WeightFormat src_wf;
     arm_compute::WeightFormat dst_wf;
+    bool                   transpose;
 };
 
 struct acl_reorder_resource_t : public resource_t {
@@ -58,7 +81,8 @@ struct acl_reorder_resource_t : public resource_t {
             &acl_obj_->src_tensor,
             &acl_obj_->dst_tensor,
             app.src_wf,
-            app.dst_wf
+            app.dst_wf,
+            app.transpose
             );
         // clang-format on
 
@@ -121,24 +145,17 @@ struct acl_reorder_fwd_t : public primitive_t {
             ACL_CHECK_SUPPORT(format_tag::undef == src_tag,
                     "Only ab, ba or cdba source formats supported");
 
-            auto dst_tag = memory_desc_matches_one_of_tag(*dst_md,
-                    format_tag::BA8b4a, format_tag::BA4b4a, format_tag::Ab4a,
-                    format_tag::Ab8a, format_tag::Acdb8a, format_tag::Acdb4a);
-            ACL_CHECK_SUPPORT(format_tag::undef == dst_tag,
-                    "Only Ab4a/Ab8a, BA8b4a/BA4b4a and Acdb8a/Acdb4a "
-                    "destination formats supported");
+            int transpose = do_transpose(src_md, dst_md);
+            ACL_CHECK_SUPPORT(transpose == -1, "Reorder not supported by ACL");
+            _pd->app_.transpose = (bool) transpose;
 
-            if (dst_tag == format_tag::BA4b4a || dst_tag == format_tag::Acdb4a
-                    || dst_tag == format_tag::Ab4a) {
-                _pd->app_.dst_wf = arm_compute::WeightFormat::OHWIo4;
-            } else if (mayiuse(sve_256)
-                    && (dst_tag == format_tag::BA8b4a
-                            || dst_tag == format_tag::Acdb8a
-                            || dst_tag == format_tag::Ab8a)) {
-                _pd->app_.dst_wf = arm_compute::WeightFormat::OHWIo8;
-            } else {
-                return status::unimplemented;
-            }
+            ACL_CHECK_SUPPORT(dst_md->format_desc.blocking.inner_nblks > 2, "destination format not supported");
+            
+            _pd->app_.dst_wf = arm_compute::WeightFormat::OHWI;
+            if (dst_md->format_desc.blocking.inner_nblks >= 1)
+                _pd->app_.dst_wf = (arm_compute::WeightFormat) ((long int) _pd->app_.dst_wf + 0x000100 * (dst_md->format_desc.blocking.inner_blks[0] - 1));
+            if (dst_md->format_desc.blocking.inner_nblks >= 2)
+                _pd->app_.dst_wf = (arm_compute::WeightFormat) ((long int) _pd->app_.dst_wf + 0x100000 * (dst_md->format_desc.blocking.inner_blks[1] - 1));
 
             arm_compute::TensorShape acl_tensor_shape_in;
             arm_compute::TensorShape acl_tensor_shape_out;
@@ -146,18 +163,14 @@ struct acl_reorder_fwd_t : public primitive_t {
             // Switch for 2 or 4 dim tensors
             switch (src_md->ndims) {
                 case 2: {
-                    if (src_tag == format_tag::ab
-                            && dst_md->data_type == data_type::bf16
-                            && utils::one_of(dst_tag, format_tag::BA8b4a,
-                                    format_tag::BA4b4a)) { // bf16
+                    if ((src_tag == format_tag::ab && (transpose == 1)) ||
+                    (src_tag == format_tag::ba && (transpose == 0))) {
                         acl_tensor_shape_in = arm_compute::TensorShape(
                                 src_md->dims[0], src_md->dims[1]);
                         acl_tensor_shape_out = arm_compute::TensorShape(
                                 dst_md->padded_dims[0], dst_md->padded_dims[1]);
-                    } else if (src_tag == format_tag::ba
-                            && dst_md->data_type == data_type::f32
-                            && !utils::one_of(dst_tag, format_tag::BA8b4a,
-                                    format_tag::BA4b4a)) { // f32
+                    } else if ((src_tag == format_tag::ba && (transpose == 1)) || 
+                                (src_tag == format_tag::ab && (transpose == 0))) {
                         acl_tensor_shape_in = arm_compute::TensorShape(
                                 src_md->dims[1], src_md->dims[0]);
                         acl_tensor_shape_out = arm_compute::TensorShape(
@@ -198,12 +211,11 @@ struct acl_reorder_fwd_t : public primitive_t {
             const arm_compute::DataType dst_acl_data_t
                     = acl_utils::get_acl_data_t(dst_md->data_type);
             _pd->app_.dst_info = arm_compute::TensorInfo(
-                    acl_tensor_shape_out, 1, dst_acl_data_t, acl_layout);
+                    acl_tensor_shape_out, 1, dst_acl_data_t, arm_compute::DataLayout::NCHW);
 
             ACL_CHECK_VALID(arm_compute::NEReorderLayer::validate(
                     &_pd->app_.src_info, &_pd->app_.dst_info, _pd->app_.src_wf,
-                    _pd->app_.dst_wf));
-
+                    _pd->app_.dst_wf, _pd->app_.transpose));
             // Init scratch memory, not used so 0 in this implementation
             _pd->init_scratchpad_md();
 
--- a/src/common/verbose.hpp
+++ b/src/common/verbose.hpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2018-2025 Intel Corporation
-* Copyright 2023 Arm Ltd. and affiliates
+* Copyright 2023, 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -382,6 +382,7 @@ std::string md2fmt_str(
         const char *name, const memory_desc_t *md, format_kind_t user_format);
 std::string md2dim_str(
         const memory_desc_t *md, dims_type_t dims_type = dims_type_t::dims);
+std::string md2fmt_tag_str(const memory_desc_t *md);
 std::string arg2str(int arg);
 // Returns a verbose string of dimensions or descriptor from src, wei, and/or
 // dst memory descs. Can be called externally to provide info about actual


src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -121,43 +145,32 @@ struct acl_reorder_fwd_t : public primitive_t {
...
-            } else {
-                return status::unimplemented;
-            }
+            int transpose = do_transpose(src_md, dst_md);
+            ACL_CHECK_SUPPORT(transpose == -1, "Reorder not supported by ACL");
+            _pd->app_.transpose = (bool) transpose;
+
+            ACL_CHECK_SUPPORT(dst_md->format_desc.blocking.inner_nblks > 2, "destination format not supported");
+            
+            _pd->app_.dst_wf = arm_compute::WeightFormat::OHWI;
```
jondea, line 154: You might want to make a temp references to make some of this logic look cleaner. E.g.
```
auto& dst_wf = _pd->app_.dst_wf;
auto& dst_format = dst_md->format_desc.blocking;
```

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -121,43 +145,32 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+            if (dst_md->format_desc.blocking.inner_nblks >= 1)
+                _pd->app_.dst_wf = (arm_compute::WeightFormat) ((long int) _pd->app_.dst_wf + 0x000100 * (dst_md->format_desc.blocking.inner_blks[0] - 1));
+            if (dst_md->format_desc.blocking.inner_nblks >= 2)
+                _pd->app_.dst_wf = (arm_compute::WeightFormat) ((long int) _pd->app_.dst_wf + 0x100000 * (dst_md->format_desc.blocking.inner_blks[1] - 1));
```
jondea, line 158: These hex numbers need some explaining, maybe a comment and/or use the enums?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -22,6 +22,28 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
+    int do_transpose(const dnnl::impl::memory_desc_t* src_md,const dnnl::impl::memory_desc_t* dst_md) {
+        auto src_md_ = dnnl::impl::md2fmt_tag_str(src_md);
```
jondea, line 27: This is very clever, but if I'm honest it does make me a bit nervous to use format strings like this.

jondea (changes requested):

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -22,6 +22,41 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
+    int do_transpose(const dnnl::impl::memory_desc_t* src_md,const dnnl::impl::memory_desc_t* dst_md) {
+        if (src_md->ndims != 4 && src_md->ndims != 2) return -1;
```
jondea, line 27: I'd put this logic inline. We only use the function once, and you introduce a new API {-1,0,1} which is not immediately obvious. Inline, it would be `transpose = true/false` or `return unimplemented` where the meaning would be immediately clear.

If anything, I'd spin out the helper function which looks for the densest dimension. They could be the same function for src and dst.

jondea (changes requested):

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done!

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done!

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -22,6 +22,16 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
+int find_dense_idx(const dnnl_memory_desc *md, int dense_blk) {
```
jondea: A quick comment here would be great, especially explaining what `dense_blk` means. Naively, I wouldn't expect you would need any extra args because you just look for the minimum stride and that's the densest.

.github/automation/aarch64/ci.json
```diff
@@ -1,6 +1,6 @@
 {
     "dependencies": {
-        "acl": "v52.0.1",
+        "acl": "v52.1.0",
```
jondea, line 3: Do any other references to the ACL version need changing? Or is this a bugfix?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -121,23 +132,73 @@ struct acl_reorder_fwd_t : public primitive_t {
...
-                _pd->app_.dst_wf = arm_compute::WeightFormat::OHWIo4;
-            } else if (mayiuse(sve_256)
-                    && (dst_tag == format_tag::BA8b4a
-                            || dst_tag == format_tag::Acdb8a
-                            || dst_tag == format_tag::Ab8a)) {
-                _pd->app_.dst_wf = arm_compute::WeightFormat::OHWIo8;
+            auto &transpose = _pd->app_.transpose;
+            auto &dst_blocking = dst_md->format_desc.blocking;
+
+            ACL_CHECK_SUPPORT(src_md->ndims != dst_md->ndims,
```
jondea: We have the general `VDISPATCH` macro for this now

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -121,23 +132,73 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+                                    *dst_md, dnnl::impl::format_tag::Acdb4a)
+                                || memory_desc_matches_tag(*dst_md,
+                                        dnnl::impl::format_tag::Acdb8a))) {
+                    transpose = true;
+                } else {
+                    return status::unimplemented;
+                };
             } else {
-                return status::unimplemented;
+                uint32_t blk_mult = 1;
```
jondea: Apologies if I have already asked this, but can't we fold this blk logic into the helper function?

jondea (changes requested):
Generally this looks great, I really like the way you have simplified the logic while expanding the coverage.

Is this new interface tested with existing `benchdnn` input files?

.github/automation/aarch64/ci.json
Ryo-not-rio: This is the only place we need to change this as it's a minor version change

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done.

Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: totally! done

.github/automation/aarch64/ci.json
jondea: I may have misunderstood, but minor versions can add features which may be used by oneDNN, and therefore the minimum version may need to be bumped. I'm guessing that's not the case here though?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -22,6 +22,27 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
+
+/*
+* Find the index of the dense block.
```
jondea: Isn't it the index of the dense _dimension_ after blocking?

.github/automation/aarch64/ci.json
Ryo-not-rio: ah yes, you're totally right, we were operating under the impression that minor versions can't introduce new features but that's definitely wrong. We need this change to go in before we can have a required minimum minor version: https://github.com/uxlfoundation/oneDNN/pull/3399/files

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

.github/automation/aarch64/ci.json
Ryo-not-rio: done!

cmake/ACL.cmake
```diff
@@ -33,7 +33,7 @@ find_package(ACL REQUIRED)
 
 # Required. The minimum compatible major-version as per Semantic Versioning.
 set(ACL_MIN_MAJOR_VERSION "52")
-set(ACL_MIN_MINOR_VERSION "0")
+set(ACL_MIN_MINOR_VERSION "1")
```
Sqvid, line 36: I think you need to update README.md as well

Sqvid (changes requested):

cmake/ACL.cmake
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -22,6 +22,27 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/reorder/cpu_reorder_pd.hpp"
 
+namespace {
```
Sqvid: Could we move all function definitions to the *.cpp and keep only declarations in the *.hpp. Also an anonymous namespace in a header file won't have much affect since it will be visible in any file that includes it.

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
 
             auto src_tag = memory_desc_matches_one_of_tag(
                     *src_md, format_tag::ab, format_tag::ba, format_tag::cdba);
-            ACL_CHECK_SUPPORT(format_tag::undef == src_tag,
+            VDISPATCH_REORDER_IC(format_tag::undef != src_tag,
                     "Only ab, ba or cdba source formats supported");
 
+            // We need this check as ACL validation is currently broken.
```
Sqvid: Can we preface this comment with `TODO:` so that it is `grep`-able? Also why v53? will it be a backwards-incompatible fix?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+                int dst_dense_idx = find_dense_idx(dst_md);
+
+                transpose = src_dense_idx != dst_dense_idx;
+            }
+
+            auto &dst_wf = _pd->app_.dst_wf;
+
+            dst_wf = arm_compute::WeightFormat::OHWI;
+            // Offsets to calculate the enum for ComputeLibrary weight formats
+            // defined in arm_compute/core/CoreTypes.h
```
Sqvid: Can't we just include `arm_compute/core/CoreTypes.h` and then `static_cast` to get these values? That seems less brittle.

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -197,13 +260,12 @@ struct acl_reorder_fwd_t : public primitive_t {
 
             const arm_compute::DataType dst_acl_data_t
                     = acl_utils::get_acl_data_t(dst_md->data_type);
-            _pd->app_.dst_info = arm_compute::TensorInfo(
-                    acl_tensor_shape_out, 1, dst_acl_data_t, acl_layout);
+            _pd->app_.dst_info = arm_compute::TensorInfo(acl_tensor_shape_out,
+                    1, dst_acl_data_t, arm_compute::DataLayout::NCHW);
```
Sqvid: What is the point of this change? It looks like `acl_layout = NCHW` a few lines up anyway.

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+            const auto block_by_offset = 0x100000;
+            for (int i = 0; i < dst_blocking.inner_nblks; i++) {
+                auto idx = dst_blocking.inner_idxs[i];
+                auto blk = dst_blocking.inner_blks[i];
+                if (idx == 0) {
+                    auto offset = dst_blocking.inner_nblks == 1
+                            ? interleave_offset
+                            : block_by_offset;
+                    dst_wf = (arm_compute::WeightFormat)(
+                            (long int)dst_wf + offset * (blk - 1));
```
Sqvid: Change these to `static_cast<long int>`

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+                } else if (idx == 1) {
+                    auto offset = dst_blocking.inner_nblks == 1
+                            ? block_by_offset
+                            : interleave_offset;
+                    // Set block_by
+                    dst_wf = (arm_compute::WeightFormat)(
+                            (long int)dst_wf + offset * (blk - 1));
```
Sqvid: ditto

Sqvid (changes requested):

src/cpu/aarch64/acl_reorder.hpp
Sqvid: Nevermind this, I misunderstood the comment.

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

Sqvid (approved):
LGTM, thanks

vpirogov:
Ping @uxlfoundation/onednn-arch for review.

README.md
```diff
@@ -179,7 +179,7 @@ On a CPU based on Arm AArch64 architecture, oneDNN CPU engine can be built with
 machine learning applications and provides AArch64 optimized implementations
 of core functions. This functionality currently requires that ACL is downloaded
 and built separately. See [Build from Source] section of the Developer Guide for
-details. oneDNN only supports Compute Library version 52.
+details. The minimimum supported version of ACL is 52.1.0.
```
dzarukin:
```suggestion
details. The minimum supported version of ACL is 52.1.0.
```

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,233 @@
 
 #include "cpu/aarch64/acl_reorder.hpp"
 
+namespace {
+/*
+* Find the index of the dense dimension.
+* The stride of the dense block will be multiplied by
+* the blocking of all prior blocks.
+*/
+int find_dense_idx(const dnnl_memory_desc *md) {
```
dzarukin:
```suggestion
int find_innermost_dense_idx(const memory_desc *md) {
```

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -218,28 +77,15 @@ struct acl_reorder_fwd_t : public primitive_t {
...
-        return execute_forward(ctx);
-    }
+    status_t execute(const exec_ctx_t &ctx) const override;
 
 private:
     // To guard the const execute_forward, the mutex must be 'mutable'
     mutable std::mutex mtx;
     status_t execute_forward(const exec_ctx_t &ctx) const;
-    const pd_t *pd() const { return (const pd_t *)primitive_t::pd().get(); }
+    const pd_t *pd() const;
```
dzarukin: This may increase the access time to each `pd()` call if not properly inlined by a compiler...

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+            VDISPATCH_REORDER_IC((dst_md->ndims == 2 || dst_md->ndims == 4),
+                    "ACL only supports 2D and 4D reorders");
+            // Check if a transpose is needed during the reorder
+            if (src_md->ndims == 4) {
+                if (memory_desc_matches_tag(
+                            *src_md, dnnl::impl::format_tag::cdba)
+                        && (memory_desc_matches_tag(
+                                    *dst_md, dnnl::impl::format_tag::Acdb4a)
+                                || memory_desc_matches_tag(*dst_md,
+                                        dnnl::impl::format_tag::Acdb8a))) {
```
dzarukin: Maybe `memory_desc_matches_one_of_tag` to combine both tags?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
-                _pd->app_.dst_wf = arm_compute::WeightFormat::OHWIo8;
+            auto &transpose = _pd->app_.transpose;
+            auto &dst_blocking = dst_md->format_desc.blocking;
+
+            VDISPATCH_REORDER_IC(src_md->ndims == dst_md->ndims,
+                    "Number of dimensions in src and dst do not match");
+            VDISPATCH_REORDER_IC((dst_md->ndims == 2 || dst_md->ndims == 4),
+                    "ACL only supports 2D and 4D reorders");
+            // Check if a transpose is needed during the reorder
+            if (src_md->ndims == 4) {
```
dzarukin: I'm not sure I'm following the logic: if ndims is 4, then there's a limitation for applicable tags but for other ndims just dense index alignment? Why is that? Any additional comment explaining the reasoning?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
-                return status::unimplemented;
+                int src_dense_idx = find_dense_idx(src_md);
+                int dst_dense_idx = find_dense_idx(dst_md);
+
+                transpose = src_dense_idx != dst_dense_idx;
+            }
+
+            auto &dst_wf = _pd->app_.dst_wf;
+
+            dst_wf = arm_compute::WeightFormat::OHWI;
```
dzarukin: Introduce a default initializer for `_pd->app_.dst_wf` instead?

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+            // Offsets to calculate the enum for ComputeLibrary weight formats
+            // defined in arm_compute/core/CoreTypes.h
+            const auto interleave_offset = 0x000100;
+            const auto block_by_offset = 0x100000;
```
dzarukin: If named constants are available for this, I'd definitely suggest to rely on those instead.

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+            if (src_md->ndims == 4) {
+                if (memory_desc_matches_tag(
+                            *src_md, dnnl::impl::format_tag::cdba)
+                        && (memory_desc_matches_tag(
+                                    *dst_md, dnnl::impl::format_tag::Acdb4a)
+                                || memory_desc_matches_tag(*dst_md,
+                                        dnnl::impl::format_tag::Acdb8a))) {
+                    transpose = true;
+                } else {
+                    return status::unimplemented;
```
dzarukin: Given that `VDISPATCH_REORDER_IC` was introduced, it's a good practice to rely on it at other spots as well.

src/cpu/aarch64/acl_reorder.hpp
```diff
@@ -118,26 +140,71 @@ struct acl_reorder_fwd_t : public primitive_t {
...
+                            (long int)dst_wf + offset * (blk - 1));
+                } else if (idx == 1) {
+                    auto offset = dst_blocking.inner_nblks == 1
+                            ? block_by_offset
+                            : interleave_offset;
+                    // Set block_by
+                    dst_wf = (arm_compute::WeightFormat)(
+                            (long int)dst_wf + offset * (blk - 1));
+                } else {
+                    return status::unimplemented;
```
dzarukin: Ditto.

src/cpu/aarch64/acl_reorder.hpp
Sqvid: Might as well annotate this with `inline` as well when we add it back then, to signal that intent.

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: We only support the two listed reorders for 4 dimensions in ACL so it's simpler to list the supported reorders rather than write another function to figure out if a transpose is necessary or not for 4d reorders. I think we can add the logic to check for transpose for 4d if we ever do enable more 4d reorders in ACL

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: agreed but we don't have named constants for this sadly

README.md
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.hpp
Ryo-not-rio: done

Sqvid:
@jondea Are you happy with the patch in it's current form?

jondea (approved):

PR #2993, opened 2025-04-01
src: gpu: intel: jit: gemm: add optional mode matching, restrict fpmath

hidefromkgb:
This PR resolves [MFDNN-13425](https://jira.devtools.intel.com/browse/MFDNN-13425).


kealan-barbieri (approved):

hidefromkgb:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

hidefromkgb:
make test perf-gpu
set primitive=matmul ip

src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
```diff
@@ -527,55 +527,70 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
...
+            fpmath_f16, /*optional=*/true, [](Type dt) -> const char * {
+                if (dt == Type::f32) { return "[SH]"; }
+                if (dt.isInt8() || dt.isInt4()) return "[OH]";
+                if (dt.isF8()) return "H";
+                return nullptr;
+            });
+
+    add_mode_matches(!(fpmath_f16 || fpmath_bf16), /*optional=*/false,
+            [](Type dt) -> const char * {
+                if (dt.isInt4()) return "[FO]";
```
petercad, line 591: I'm not sure this part is 100% correct. We have two kinds of x8:x4:f16 kernels in the catalog: `FOS` and `[FO]OS`. We can always use the `[FO]OS` ones (upconverting int4->int8 prior to dpas). However, if there are no wei zp then we can pass the int4 wei data directly to dpas -- `FOS` -- and that can be faster. Here, then, if I'm reading the patch correctly, you want optional to be determined by the presence of wei zp.

src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
```diff
@@ -527,55 +527,70 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
...
+                match_params.push_back(base);
+                match_params.back().selector.precisions[0] = def[0];
+            }
             match_params.back().selector.precisions[1] = match(problem_.Tb);
         }
         if (match(problem_.Ta) && match(problem_.Tb)) {
-            match_params.push_back(base);
+            if (optional) match_params.push_back(base);
             match_params.back().selector.precisions[0] = match(problem_.Ta);
             match_params.back().selector.precisions[1] = match(problem_.Tb);
```
rjoursler, line 563: The logic here is really weird. The point of the call to `push_back` is to add a new value match_params, and then the subsequent calls set the value to the desired result. With this change, we are just mutating the same match structure up to 3 different times. and selecting the last mutation. I have trouble seeing how this is the desired logical effect. Could you explain how this is intended to work?

kealan-barbieri: at least in https://github.com/uxlfoundation/oneDNN/pull/2916 this is intended for cases where `fpmath_t::apply_to_int_`  is `true`, in which case integer inputs must be upconverted. In such a case where both inputs are integer type we have only one valid `match_param` where both inputs are substituted with some compound tag, eg. `[OB]` . We could avoid permuting it three times by putting this case first and using `else if` though.

src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
```diff
@@ -527,55 +527,70 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
...
+        return (type == dnnl_u4) || (type == dnnl_s4) || (type == dnnl_u8)
+                || (type == dnnl_s8);
+    };
+    bool full_int = is_int_type(a_type) && is_int_type(b_type);
+
     bool fpmath_tf32 = mode & mode_tf32;
-    bool fpmath_bf16 = mode & mode_bf16x1;
-    bool fpmath_f16 = mode & mode_f16x1;
+    bool fpmath_bf16 = (mode & mode_bf16x1) && !full_int;
+    bool fpmath_f16 = (mode & mode_f16x1) && !full_int;
```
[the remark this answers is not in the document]
kealan-barbieri: I think part of this PR is preserving the existing (incorrect) behavior where we still perform integer computation if the type signature is say "u8:s4:f16" with `fpmath=f16:true`, effectively ignoring fpmath mode to follow the "dynamic quantization" pattern.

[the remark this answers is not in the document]
hidefromkgb: > I think part of this PR is preserving the existing (incorrect) behavior where we still perform integer computation if the type signature is say "u8:s4:f16" with `fpmath=f16:true`, effectively ignoring fpmath mode to follow the "dynamic quantization" pattern.

This is exactly right, which is why I'm calling this a workaround and not even merging it into main.

src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
rjoursler: Thanks for the explanation. Lets just directly modify `base` or `match_params[0]` outside this lambda in this case.  Using `add_mode_matches` just injects unnecessary complexity as we are actually adding any mode matches  when optional is set.

kealan-barbieri: split `mod` and `add` functionality in https://github.com/uxlfoundation/oneDNN/pull/2916 to align with this pattern.

hidefromkgb:
Superseded by #3076.

PR #2994, opened 2025-04-01
ngen: workaround for SYCL + GCC 12.3 compiler bug

petercad:
Addresses MFDNN-13382. When building under SYCL with gcc 12.3 as the host compiler, accuracy failures are seen in certain fp8 kernels. The culprit is a compiler bug that looks similar to a gcc 13 bug noticed earlier. A similar workaround addresses the failure in this situation.


petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

kealan-barbieri (approved):

atkassen (approved):

PR #2998, opened 2025-04-01
benchdnn: speed up convolution reference by better prim_ref

dzarukin:
1) CPU list lacks x8:s8:f16 combination which is frequent for GPU. Enabling it adds some cases to go into CPU.
2) Benchdnn's prim ref used original sum_dt if it was specified which could limit CPU capabilities which doesn't have f16 natively supported.
3) Benchdnn's prim ref used `tag::abx` for dst which would end up in format mismatching in most cases when there's more that 1 non-unit dimension. Switching to `tag::any` enabled more cases.


dzarukin:
make test

echeresh (approved):

avmanerikar (approved):

dzarukin:
make test

PR #3000, opened 2025-04-02
fix for big shapes pr

ankalinin:
This is a fix for [MFDNN-13409](https://jira.devtools.intel.com/browse/MFDNN-13409)
- remove unnecessary checks for size in convolutions: we have unified check by using `has_large_size` after #2827 
- update `has_large_size` function: exclude minibatch from size calculation because we process data image by image in all convolutions

Pre-commit shows no new fails excepting those that have already been in CI.


ankalinin:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl

tprimak (approved):

dzarukin (approved):
Would be great to verify that no new GPU cases (with CPU reference) regress.

spalicki (approved):

PR #3002, opened 2025-04-02
api: make md queries return const object

mgouicem:
This is a draft PR to test change to md queries semantic:
- C++ API avoids the cloning
- md queries that can now return const memory::desc.

### DIFF
--- a/include/oneapi/dnnl/dnnl.hpp
+++ b/include/oneapi/dnnl/dnnl.hpp
@@ -4642,7 +4642,7 @@ struct primitive_desc_base : public handle<dnnl_primitive_desc_t> {
     /// @returns The requested memory descriptor.
     /// @returns A zero memory descriptor if the primitive does not have a
     ///     parameter of the specified kind or index.
-    memory::desc query_md(query what, int idx = 0) const {
+    const memory::desc query_md(query what, int idx = 0) const {
         std::vector<query> valid_q {query::src_md, query::diff_src_md,
                 query::weights_md, query::diff_weights_md, query::dst_md,
                 query::diff_dst_md, query::workspace_md, query::scratchpad_md,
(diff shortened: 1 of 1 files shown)


include/oneapi/dnnl/dnnl.hpp
```diff
@@ -4654,21 +4654,18 @@ struct primitive_desc_base : public handle<dnnl_primitive_desc_t> {
...
                 get(), dnnl::convert_to_c(what), idx);
-        if (!cdesc) return memory::desc();
 
-        dnnl_memory_desc_t cloned_md = nullptr;
-        error::wrap_c_api(dnnl_memory_desc_clone(&cloned_md, cdesc),
-                "could not clone a memory descriptor");
+        auto nc_cdesc = const_cast<dnnl_memory_desc_t>(cdesc);
 
-        return memory::desc(cloned_md);
+        return nc_cdesc ? memory::desc(nc_cdesc) : memory::desc();
```
densamoilov, line 4660: I believe changing the API semantics in this way may be problematic. It implicitly binds the lifetime of the returned object to that of the primitive descriptor, which significantly complicates the API's usability. The code snippet below illustrates one way this can lead to issues.
```cpp
auto src_md = convolution_forward::primitive_desc(engine, ...).src_desc();
// src_md is no longer valid.
```
If we're aiming to reduce overhead, I'd recommend making the option to skip cloning opt-in, similar to how the `allow_empty` feature works.

densamoilov: Or, as an alternative, count references.

densamoilov: One more thing. `memory::desc` class is based on the `handle` class that releases the handle upon destruction so it seems that it will try to release `nc_cdesc`.

openmodelzooci:

PR #3005, opened 2025-04-02
xe: jit: gemm: fix zero pool memory leak under SYCL when not recording

petercad:
Closes MFDNN-13250. A memory leak was inadvertently introduced when introducing initial SYCL graph recording support for zero pools in GEMM. The memory leak is expected when under recording mode (temporary measure -- will be replaced with async malloc/free when graph recording support available), but not in regular execution.

### DIFF
--- a/src/gpu/intel/jit/gemm/gen_gemm.cpp
+++ b/src/gpu/intel/jit/gemm/gen_gemm.cpp
@@ -24,6 +24,10 @@
 #include "gpu/intel/jit/gemm/gemm_walk_orders.hpp"
 #include "gpu/intel/jit/gemm/include/driver_info.hpp"
 
+#ifdef DNNL_WITH_SYCL
+#include "gpu/intel/sycl/stream.hpp"
+#endif
+
 namespace dnnl {
 namespace impl {
 namespace gpu {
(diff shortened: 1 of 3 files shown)


petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

atkassen (approved):

src/gpu/intel/jit/gemm/gen_gemm.cpp
```diff
@@ -231,11 +235,17 @@ status_t gen_gemm_t::execute(const gemm_exec_ctx_t &ctx) const {
     auto zero_pool = zero_pool_;
 
 #ifdef DNNL_WITH_SYCL
-    if (!zero_pool) {
```
kealan-barbieri, line 234: are we now doing a lookup for `zero_pool` even if its already initialized?

petercad: Yes -- the motivation is that we have two different flows (regular mode and graph recording) and we don't know which one we'll get until execute time.

For regular execution, the zero pool is looked up once as part of the primitive initialization, and reused for all executions.

In graph recording mode, we can't use the global zero pool (synchronization and memory lifetime issues) so we call lookup_zero_pool again to grab a per-graph zero pool.

kealan-barbieri: thank you for clarifying.

petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

PR #3007, opened 2025-04-02
cpu: x64: matmul: correct blocked_B layout initialization

xuxinzen:
Fixes MFDNN-13348

`blocked_16n_B_layout_tag` is not initialized correctly and then the result of `bgmmc.blocked_B` could be incorrect. This could cause segfault in the later computation.


xuxinzen:
make test

ankalinin (approved):

tczeszun (approved):

PR #3014, opened 2025-04-03
gpu: intel: sycl: l0: remove dependency to OCL for atomics query

mgouicem:
It turns out the atomics support is to be queried in LZ from device module property, and not from device property. This removes the call to OCL to query atomics support and uses the proper query in LZ.

Fixes MFDNN-13201


mgouicem:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

rjoursler (approved):

Issue #2165, opened 2024-10-14
 Extend support for JIT Backward Convolution Operators with ARM  SVE 128bit 


snadampal:
# Summary

On aarch64 platform, Convolution backward operators are supported via jitted SVE kernels. Today the support exists only for SVE 512 and SVE 256bit width, but not for SVE 128bit processors like AWS Graviton4. The request is to extend the existing SVE jitted kernels to support 128bit width.

# Problem statement

resnet50 model training requires backward convolution operators, and these currently executed with reference 'c' kernels. Extending the following oneDNN operators for SVE 128bit accelerates these operators with SIMD and improved the performance by several orders.
Here are the details on the existing  oneDNN jitted kernels for backward pass operators:
~~~

https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/cpu_convolution_list.cpp#L262
 
{{backward_data, f32, f32, f32}, REG_BWD_D_PK({
…..
            CPU_INSTANCE_AARCH64(jit_uni_dw_convolution_bwd_data_t<sve_512,data_type::f32>)
            CPU_INSTANCE_AARCH64(jit_sve_1x1_convolution_bwd_data_t<f32,f32,f32,sve_512>)
            CPU_INSTANCE_AARCH64(jit_sve_convolution_bwd_data_t<f32,f32,f32,sve_512>)
            CPU_INSTANCE_AARCH64(jit_uni_dw_convolution_bwd_data_t<sve_256,data_type::f32>)
            CPU_INSTANCE_AARCH64(jit_sve_1x1_convolution_bwd_data_t<f32,f32,f32,sve_256>)
            CPU_INSTANCE_AARCH64(jit_sve_convolution_bwd_data_t<f32,f32,f32,sve_256>)
….
}
 
{{backward_weights, f32, f32, f32}, REG_BWD_PK({
……
            CPU_INSTANCE_AARCH64(jit_uni_dw_convolution_bwd_weights_t<sve_512,data_type::f32>)
            CPU_INSTANCE_AARCH64(jit_sve_1x1_convolution_bwd_weights_t<f32,f32,f32,sve_512>)
            CPU_INSTANCE_AARCH64(jit_sve_convolution_bwd_weights_t<f32,f32,f32,sve_512>)
            CPU_INSTANCE_AARCH64(jit_uni_dw_convolution_bwd_weights_t<sve_256,data_type::f32>)
[...]

PR #3022, opened 2025-04-04
cpu: aarch64: enable jit conv for 128

jondea:
# Description

Draft: needs some optimization tweaks, also deconv tests fail due to `check_zero_padding`

Naively enable 128 by copying the equivalent invocations to 512 and 256. Note that `is_1stconv` is hard coded to `true` for `sve_128`, which misses out on some performance.

Some more optimization is necessary, but this speeds up some cases, specifically backward. In some cases this was slower than Arm Compute Library (ACL), so unlike the 512 and 256 counterparts, it has been set below the ACL implementations in the CPU list.

# Checklist

## General

## Performance improvements


jondea:
@kasturedeeksha is this a reasonable approach? (It's a draft, so I know there may be some code quality issues)

jondea:
Should help #2165

openmodelzooci:

jondea:
This has been replaced by several other PRs which have broadly acheived the same thing.

PR #3028, opened 2025-04-04
cpu: x64: brdgmm: enable sum with different data type

xuxinzen:
Brdgmm kernel already has the support of sum w/ data type.


xuxinzen:
make test

ankalinin (approved):

dzarukin (approved):

PR #2366, opened 2025-01-09
[GPU] Improve reference kernel coverage.

dyoussif:
Addresses https://jira.devtools.intel.com/browse/MFDNN-12444. 
Broaden ocl:ref coverage and add some smoke tests for key compute bound primitives.

PR #3049, opened 2025-04-08
tests: benchdnn: remove fpmath=f16 for some cases

dyoussif:
These test cases  are not supported by cpu and return unimplemented. Revert back to original cases.

Fixup for #2366.


dyoussif:
make test

dzarukin (approved):

xuxinzen (approved):

PR #3055, opened 2025-04-09
[WIP] common: verbose: asynchronous verbose mode for execution time tracking

avmanerikar:
WORK IN PROGRESS (PROTOTYPE) - DO NOT REVIEW

Description
This PR proposes a prototype for introducing an asynchronous verbose mode to accurately track kernel execution times for GPU engines. For the verbose mode, retrieving the kernel timing causes significant overhead as it requires the GPU kernel execution to be synchronized and also because it is tracked on the host.
The asynchronous mode uses event-based tracking to measure the GPU kernel execution to avoid the synchronization issue.
The prototype is created for a SYCL GPU API with event callbacks that provide the verbose information for GPU kernels as they get executed.

Addresses MFDNN-12088.

# Checklist
- [ ] Was the RFC approved?


src/gpu/intel/ocl/stream.hpp
```diff
@@ -102,6 +102,53 @@ struct stream_t : public compute::compute_stream_t {
...
+            double *start_ms, double *duration_ms) override {
+        cl_event exec_evt = get_output_event();
+        if (!exec_evt) return status::runtime_error;
+
+        clRetainEvent(exec_evt);
+        async_tracked_event_ = exec_evt;
+
+        auto *td = new TimingData(start_ms, duration_ms);
+
+        cl_int err = clSetEventCallback(
```
mgouicem, line 134: Not all stream are created with profiling events. Please check for `stream::is_profiling_enabled()` ([this function](https://github.com/uxlfoundation/oneDNN/blob/bd9f25f6e3a8e15a8d97d63eb27d7a13d088021a/src/common/stream.hpp#L74)).

As a fallback, you can use a callback function that would just call `get_ms()` on host.

src/gpu/intel/ocl/stream.hpp
```diff
@@ -102,6 +102,53 @@ struct stream_t : public compute::compute_stream_t {
...
+                exec_evt, CL_COMPLETE, async_tracker_callback, td);
+
+        return (err == CL_SUCCESS) ? status::success : status::runtime_error;
+    }
+
+    status_t get_async_exec_times() {
+        if (!async_tracked_event_) return status::runtime_error;
+
+        cl_int err = clWaitForEvents(1, async_tracked_event_);
```
mgouicem, line 143: This seems like the wrong event to use here, get_output_event() should be used after the call to stream::enqueue_primitive, not before (otherwise you will not get execution time from primitive execution).

src/common/primitive_iface.cpp
```diff
@@ -97,11 +97,23 @@ status_t primitive_execute(
 
     if (get_verbose(verbose_t::exec_profile,
                 prim_kind2_comp_kind(primitive_iface->pd()->impl()->kind()))) {
+
+#ifdef EXPERIMENTAL_ASYNC_VERBOSE
+        // tracks the execution times for the primitive asynchronously by
+        // directly querying the event times for the stream.
+        // gets rid of the synchronization overheads on the host device
+        double start_ms, duration_ms;
+        CHECK(stream->register_async_tracker(&start_ms, &duration_ms));
```
mgouicem, line 106: start_ms and duration_ms are on the stack, I am surprised the callback function does not segfault since those are out of scope when the callback function is running.

src/gpu/intel/ocl/stream.hpp
avmanerikar: Thanks for the suggestion! This is now added in the latest update with the other suggested changes.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,91 @@ status_t stream_t::barrier() {
...
+        // stream.wait() for capturing execution time.
+        wait();
+        *duration_ms = get_msec() - *start_ms;
+        delete async_timing_data_;
+        async_timing_data_ = nullptr;
+
+        return status::success;
+
+    } else {
+        cl_int err = clWaitForEvents(1, &async_tracked_event_);
```
echeresh, line 177: Isn't this call synchronous? We enqueue a kernel, record an event and execution blocks here, until the kernel finishes. Am I missing something?

avmanerikar: Yes. This was a fallback for failure cases where the verbose info is then printed with the default `stream.wait()` calls. The implementation has been updated to avoid repetition.

avmanerikar:
make test
disable test_device_cpu
enable test_device_gpu
enable build_gpu_runtime_ocl
enable build_gpu_runtime_sycl

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,89 @@ status_t stream_t::barrier() {
...
+
+    // Check if there is any event in the queue - for default
+    // implementation, there is no kernel on queue to record for.
+    if (!event) return status::success;
+
+    if (!stream_timing_data_) return status::success;
+
+    cl_event exec_evt = event;
+    cl_int retain_status = clRetainEvent(exec_evt);
+    if (retain_status != CL_SUCCESS) return status::runtime_error;
```
echeresh:
```suggestion
    OCL_CHECK(clRetainEvent(exec_evt));
```

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,89 @@ status_t stream_t::barrier() {
...
+    async_tracked_event_ = exec_evt;
+
+    // local container for storing timing data for the event
+    auto event_timing_data_ = new async_timing_data_t;
```
echeresh: Looks like this is never deleted.

src/gpu/intel/ocl/kernel.cpp
```diff
@@ -219,12 +219,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
...
     } else {
-        bool save_event = save_events_ || stream.is_profiling_enabled();
         cl_int err = clEnqueueNDRangeKernel(queue, *kernel, ndims, nullptr,
                 range.global_range().data(),
                 range.local_range() ? range.local_range().data() : nullptr, 0,
-                nullptr, save_event ? &event.unwrap() : nullptr);
+                nullptr, &event.unwrap());
         OCL_CHECK(err);
+
+        ocl_stream->register_async_tracker(event);
```
echeresh: Missing check of the return status?

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,89 @@ status_t stream_t::barrier() {
...
+}
+
+status_t stream_t::check_async_exec_times(
+        std::string &vinfo, double *start_ms) {
+
+    if (!stream_timing_data_->timing_stat) {
+        CHECK(wait());
+        double duration_ms = get_msec() - stream_timing_data_->start_ms;
+
+        VPROF(static_cast<int>(stream_timing_data_->start_ms), primitive, exec,
```
echeresh: I think we should use double as before:
```suggestion
                VPROF(stream_timing_data_->start_ms, primitive, exec,
```

src/gpu/intel/ocl/kernel.cpp
```diff
@@ -219,12 +219,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
         OCL_CHECK(err);
         xpu::ocl::event_t::from(out_dep).events = {event};
     } else {
-        bool save_event = save_events_ || stream.is_profiling_enabled();
         cl_int err = clEnqueueNDRangeKernel(queue, *kernel, ndims, nullptr,
                 range.global_range().data(),
                 range.local_range() ? range.local_range().data() : nullptr, 0,
-                nullptr, save_event ? &event.unwrap() : nullptr);
+                nullptr, &event.unwrap());
```
echeresh: This makes `clEnqueueNDRangeKernel()` to always return an `event` which may have performance impact.

I think we don't really need an event without profiling enabled - as we can't query it anyway so the previous code, without changes, based on `save_event` should probably work here.

src/gpu/intel/ocl/stream.hpp
```diff
@@ -116,8 +121,14 @@ struct stream_t : public compute::compute_stream_t {
             cl_context ctx, cl_device_id dev, cl_int *err) const;
 
     std::unique_ptr<mdapi_helper_t> mdapi_helper_;
+
+    cl_event async_tracked_event_ = nullptr;
+    async_timing_data_t *stream_timing_data_ = nullptr;
```
echeresh: Please use smart pointers to make sure it's properly destroyed in case of an early return (e.g. because of some error).

```suggestion
    std::unique_ptr<async_timing_data_t> stream_timing_data_;
```

Edit: I see that you used `shared_ptr` in SYCL, let's align OpenCL with SYCL. It also seems `unique_ptr` would be enough.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,89 @@ status_t stream_t::barrier() {
...
+    delete stream_timing_data_;
+    stream_timing_data_ = nullptr;
+
+    return status::success;
+}
+
+void CL_CALLBACK async_tracker_callback(
+        cl_event event, cl_int event_command_status, void *user_data) {
+
+    if (event_command_status != CL_COMPLETE) return;
```
echeresh: When is this case possible? Do we still need to release the event in this case?

src/gpu/intel/sycl/stream.cpp
```diff
@@ -174,6 +174,67 @@ status_t stream_t::resume_recording() {
...
+                if (td->profiler_enabled) {
+                    auto start_ms = event.get_profiling_info<
+                            ::sycl::info::event_profiling::command_start>();
+                    auto end_ms = event.get_profiling_info<
+                            ::sycl::info::event_profiling::command_end>();
+                    td->start_ms = start_ms * 1e-6;
+                    td->end_ms = end_ms * 1e-6;
+                    td->duration_ms = (end_ms - start_ms) * 1e-6;
+                } else {
+                    td->end_ms = get_msec();
```
echeresh, line 220: I suggest to move this branch and handling out of the host task - we can check and exit earlier, without submitting a task and adding extra overhead.

src/gpu/intel/ocl/stream.cpp
avmanerikar: Fixed along with the suggestions. Thanks!

src/gpu/intel/ocl/stream.cpp
avmanerikar: `event_timing_data_` is intentionally added for two reasons:
-  it locally stores the event-wise timing information as `stream_timing_data_` will not be preserved  for the OpenCL callback. 
-  it accounts for multiple events queueed on the stream.

src/gpu/intel/ocl/stream.cpp
avmanerikar: Thanks for the catch. The check is now adjusted so that the event is released for all cases.

src/gpu/intel/sycl/stream.cpp
avmanerikar: The branch only changes the method for obtaining the timing data based on whether profiling capabilites are enabled, so it needs to be included in the `host_task`.

avmanerikar:
make test
disable test_device_cpu
enable test_device_gpu
enable build_gpu_runtime_ocl
enable build_gpu_runtime_sycl

src/gpu/intel/ocl/stream.cpp
echeresh: Sorry, my comment wasn't clear enough. I see `new` but no `delete` - looks like a memory leak.

avmanerikar: Got it. It was an artifact of code rewrite - the fix has been added in the OpenCL callback.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,81 @@ status_t stream_t::barrier() {
...
+    stream_timing_data_->start_ms = get_msec();
+    stream_timing_data_->profiler_enabled = is_profiling_enabled();
+    stream_timing_data_->vinfo = vinfo;
+    return status::success;
+}
+
+status_t stream_t::register_async_tracker(cl_event event) {
+
+    // Checks if there is any event in the queue - for default
+    // implementation, there is no kernel on queue to record for.
```
echeresh: Did you mean "there is no event" here? Not sure I understand why "no kernel".

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,81 @@ status_t stream_t::barrier() {
     return impl()->barrier();
 }
 
+status_t stream_t::init_async_tracking(std::string &vinfo, double *start_ms) {
+    stream_timing_data_->start_ms = get_msec();
```
echeresh: After taking a second look - it seems we need some changes to ensure it's thread-safe. It's not safe to have `stream_timing_data_` stored inside the stream object as there may be multiple concurrent submissions to the same stream (which is a valid scenario).

There is an option to use a mutex specifically with `ONEDNN_VERBOSE` mode but I would try to explore less intrusive options first - to avoid deviations with verbose mode and just to follow the common design.

Please take a look at `xpu::{sycl,ocl}::context_t`:

https://github.com/uxlfoundation/oneDNN/blob/608fc3ae0b6ee5d53d2a60dc97263f322a020838/src/xpu/ocl/stream_impl.hpp#L98

It's used to track events, and TLS helps to avoid issues with concurrent execution. Can we move the "timing data" structure to the context?

src/gpu/intel/ocl/stream.cpp
```diff
@@ -136,6 +136,81 @@ status_t stream_t::barrier() {
...
+
+    cl_event exec_evt = event;
+    OCL_CHECK(clRetainEvent(exec_evt));
+    async_tracked_event_ = exec_evt;
+
+    // local container for storing timing data for the event
+    auto *event_timing_data_ = new async_timing_data_t(*stream_timing_data_);
+
+    cl_int err = clSetEventCallback(
+            exec_evt, CL_COMPLETE, async_tracker_callback, event_timing_data_);
```
echeresh, line 181: After another look - I think we need to be more careful here for multi-kernel primitives.

For example, you can test this case:

```
./build/tests/benchdnn/benchdnn -v5 --conv --mode=F --engine=gpu --dt=bf16 --dir=BWD_W mb128ic512ih32oc512oh32kh3
```

It includes three kernels, and the benchdnn timing approach is to take the end time stamp from the last kernel and the starting time stamp from the very first kernel. `ONEDNN_VERBOSE` approach should be aligned with benchdnn.

For example we have these three kernels:

```
zero_out
gen_conv
conv_reorder
```

`ONEDNN_VERBOSE` profiling-based time should be calculated as `conv_reorder_end - zero_out_start`.

I believe for this we need more sophisticated logic - we need to bookkeep the events and register the callback only after `enqueue_primitive` (and not inside `parallel_for`).

src/gpu/intel/sycl/stream.cpp
echeresh: I guess you can check whether the profiling capabilities are enabled before the host task. But anyway, after thinking more, my main concern was the overhead, and since this is specifically for verbose when overhead is expected, this shouldn't be very important. Feel free to resolve.

src/gpu/intel/sycl/stream.cpp
```diff
@@ -174,6 +174,67 @@ status_t stream_t::resume_recording() {
...
+
+    stream_timing_data_->timing_stat = true;
+    return status::success;
+}
+
+status_t stream_t::check_async_exec_times(
+        std::string &vinfo, double *start_ms) {
+
+    if (!stream_timing_data_ || stream_timing_data_->timing_stat) {
+        stream_timing_data_.reset();
```
echeresh: Please double check the synchronization and potential lifetime issues. I assume `host_task()` is asynchronous so there is a chance this `reset()` happens before `host_task` accesses `stream_timing_data_` data. We need to ensure this never happens - e.g. move this `reset()` inside `host_task`.

src/gpu/intel/ocl/stream.cpp
avmanerikar: Fixed. Thanks!

src/gpu/intel/ocl/stream.cpp
avmanerikar: Thanks for the suggestion. This has been updated in the latest commits - the timing data is now defined and maintained as a part of `xpu::context_t`.

src/gpu/intel/ocl/stream.cpp
avmanerikar: As mentioned in the [RFC](https://github.com/uxlfoundation/oneDNN/tree/amanerik/rfcs/async-verbose-mode/rfcs/20250527-async-verbose-mode), this is indeed a complicated scenario that the current logic for the asynchronous mode does not account for.  For now, I have added a verbose warning indicating the fallback to the blocking verbose when there are multiple kernels involved.

src/gpu/intel/sycl/stream.cpp
avmanerikar: That's correct. Upon closer inspection, the `reset()` seems unnecessary as the `timing_data` is refreshed with every initialization. It is therefore removed in the latest update.

avmanerikar:
make test
disable test_device_cpu
enable test_device_gpu
enable build_gpu_runtime_ocl
enable build_gpu_runtime_sycl

src/gpu/intel/ocl/stream.cpp
echeresh: I see, thanks. We will need to quickly follow up on that after this PR as having inconsistent timing approaches between different cases is not good for performance analysis or debugging. Also it's needed for proper #2973 resolution.

Related question: for proper multi-kernel support and out-of-order queue support, do you see a path to extend the callback solution to support these scenarios as well? or do we have to move from callbacks to event polling as Peter suggested in [MFDNN-12088](https://jira.devtools.intel.com/browse/MFDNN-12088)?

For the fallback, do you see a simple way to implement it? For SYCL, `host_task()` is submitted in-order but we need to know inside it if it's a single- or multi-kernel execution. In the latter case, we have to skip prints inside the callback but we don't know if there are more kernels to submit at the time of the first kernel submission.

For OpenCL, I can think of the sequence below in this exact order. Let's say the primitive includes two kernels, A and B:

- Kernel A is submitted
- Callback for kernel A event is set
- Kernel A is completed, callback is signaled
    - Note, at this point we don't know that kernel B will be submitted so the callback will print the execution time incorrectly
- Kernel B is submitted
- ...

Here, the callback is signaled before kernel B is submitted. Is this scenario possible in practice? If yes, can we handle it correctly?

avmanerikar: Currently, the asynchronous mode uses two `async_timing_data_` objects to keep track of profiling information: (i) `event_timing_data_` to record individual event timings and (ii) `stream_timing_data_` to record stream-wise timings.
For multi-kernel and out-of-order queue support, I propose the following extension - it is based on Peter's idea from MFDNN-12088, but leverages the timing data structure for aggregation rather a pending event list:

1) For each stream, we maintain a list (`primitive_timing_data_`) of `async_timing_data_` objects (one per primitive) instead of a single `stream_timing_data_` instance. For the listed instances, The `vinfo` field serves as the primitive marker, associating each timing data object with its respective primitive.
2) As each event callback is triggered, the corresponding `primitive_timing_data_` object is updated with the new end time and duration. This ensures that each primitive's timing aggregates all its relevant kernel events, even in out-of-order execution.
3) During stream destruction, timing data for each primitive is printed in the order the primitives were queued. Optionally, event polling can also be used to check for primitives whose events are complete, but it should be avoided as it does not account for the aforementioned scenario with late event submissions.

There remain two considerations for the proposed extension:
- The profiling info is printed in an aggregated manner for all the primitives at the end of the stream. This can be cumbersome when there are many primitives executed on the stream.
- the `primitive_timing_data_` objects must be shared safely across all callbacks which requires atomic operations or proper synchronization to ensure thread safety in presence of concurrent callbacks.

The [RFC](https://github.com/uxlfoundation/oneDNN/pull/3393) has also been updated to include this proposal.

echeresh: @avmanerikar Thanks!

I have one comment. With the updated approach we have to introduce a separate "aggregation" step - which should combine data from all primitive kernels, etc. Taking this into account, to me there is no benefit in using callbacks as now the callback responsibility is just to record the kernel time stamps. But this can be directly queried from recorded events on the host. At the same time, callbacks add complexity, e.g. SYCL/OpenCL need different approaches, future L0 support probably needs different handling (is there support for L0 callbacks?).

On the other hand, with direct host-side event querying, the approach is relatively uniform: step 1: we record events (regardless of the underlying runtime), step 2: we poll the recorded events to see if they are completed or not and aggregate their times to print verbose.

avmanerikar: Using a host-side event querying approach is possible, but it comes with three main limitations compared to the current callback-based method:

- Obtaining timing data from recorded events requires that stream profiling capabilities are enabled at creation time. By default, this is not the case and the corresponding flags must be enabled by the user (`CL_QUEUE_PROFILING_ENABLE` in OpenCL and `property::queue::enable_profiling` in SYCL).
- The current implementation has a fallback in place which uses `get_msec()` to record the timings when device-side polling is disabled. A host-side polling approach can also implement such a fallback but the accuracy will be limited by the polling frequency. One would need to either poll continuously (at the cost of CPU cycles) or accept a lower timing resolution.
- Shifting the timing capture to the host will also reintroduce the synchronization overhead which was previously taken care of by the asynchronous callbacks.

echeresh: > Obtaining timing data from recorded events requires that stream profiling capabilities are enabled at creation time. By default, this is not the case and the corresponding flags must be enabled by the user (CL_QUEUE_PROFILING_ENABLE in OpenCL and property::queue::enable_profiling in SYCL).

I agree. But this is something we know beforehand, right? The stream can be checked on entry to primitive execution and we can (and have to) operate in two modes: profiling-based timing and host-side timing (via `get_msec()`). So in the latter case we can keep the existing code under a condition:

```c++
if (!stream_with_profiling) {
    double ms = get_msec();
    execute();
    wait();
    double duration = get_msec() - ms;
    // print verbose line
} else {
    execute(); // event are recorded inside
    maybe_print_verbose(); // from completed events - or do this on stream.wait()/stream destruction
}
```

Regarding the overhead comment: this is true but I think you suggested to print verbose on stream destruction (or this can also be done during stream synchronization). This approach may or may not be user-friendly - but it can be a starting point before further adjustments.

With callbacks we will need synchronization (via locking, etc) for multi-kernel primitives or OOO queues to ensure the required events are all completed: the last kernel event callback must ensure all previous events are handled before trying to aggregate them. This extra handling will add more complexity to the callback solution.

petercad: > Obtaining timing data from recorded events requires that stream profiling capabilities are enabled at creation time. By default, this is not the case and the corresponding flags must be enabled by the user (CL_QUEUE_PROFILING_ENABLE in OpenCL and property::queue::enable_profiling in SYCL).

IMO, it is not worth worrying about device timing unless we have stream profiling enabled. Without it, we don't get accurate times, and we might as well use existing verbose timing.

Seconding that we should avoid event callbacks, this does not have native support in HW and is not very efficient. I think it's implemented by polling event completion in a separate thread.

vpirogov: > IMO, it is not worth worrying about device timing unless we have stream profiling enabled. 

We still want to avoid synchronizations in that case though to avoid breaking SYCL Graph.

echeresh: @vpirogov It won't be trivial - without synchronization and without profiling events the host-side measurements can be inaccurate. Can we disable time measurement in the SYCL Graph (when profiling events are not available)? It should be the simplest approach.

avmanerikar: > I agree. But this is something we know beforehand, right? The stream can be checked on entry to primitive execution and we can (and have to) operate in two modes: profiling-based timing and host-side timing (via `get_msec()`). So in the latter case we can keep the existing code under a condition:

@echeresh , that's right. But there is also the issue of how profiling is enabled. For example, for OpenCL API, profiling must be enabled by the user on the command-queue which is then passed into the oneDNN stream:
```
    cl_command_queue ocl_queue {};
    cl_queue_properties props[] = {CL_QUEUE_PROPERTIES, CL_QUEUE_PROFILING_ENABLE, 0};
    ocl_queue = clCreateCommandQueueWithProperties(ocl_interop::get_context(engine),
        ocl_interop::get_device(engine), props, ...);
```
A similar action is required for SYCL API via SYCL interop. A runtime-agnostic stream with `stream::flags::profiling` can be created but that requires DNNL_EXPERIMENTAL_PROFILING to be enabled during runtime. For the asynchronous mode to be enabled by default, profiling will also have to be enabled by default.

> With callbacks we will need synchronization (via locking, etc) for multi-kernel primitives or OOO queues to ensure the required events are all completed: the last kernel event callback must ensure all previous events are handled before trying to aggregate them. This extra handling will add more complexity to the callback solution.

If the verbose messages are to be printed at the time of stream destruction, synchronization is not necessary since all the events will be completed. Since the timing data is stored in the `primitive_timing_data_` list for each primitive, the callbacks will update the respective primitive end times as they complete. This works for both multi-kernel and OOO queue scenarios.
 
> IMO, it is not worth worrying about device timing unless we have stream profiling enabled. Without it, we don't get accurate times, and we might as well use existing verbose timing.
> 
> Seconding that we should avoid event callbacks, this does not have native support in HW and is not very efficient. I think it's implemented by polling event completion in a separate thread.

@petercad, I agree. But the idea for introducing asynchronous verbose mode was to triage two issues: (i) to make the verbose profiling mode non-blocking (ii) to get accurate execution timings by reducing host-to-device synchronization overheads. 
With the host-side event polling approach, there will be higher synchronization overheads unless, as you mentioned, the profiling capabilities are enabled. If there is a workaround that enables profiling by default, then this implementation can be possible.

echeresh: > For the asynchronous mode to be enabled by default, profiling will also have to be enabled by default.

The asynchronous mode can be automatic - only for streams with enabled profiling. For other cases we use the current approach with synchronization + host-side timing. One exception: the SYCL Graph - we skip verbose timing for this scenario (unless stream is with profiling).

> If the verbose messages are to be printed at the time of stream destruction, synchronization is not necessary since all the events will be completed. 

Agree, I think my comment was about potential advantages of the callback solution - which doesn't seem to have any. Synchronization I mentioned - is only when we start thinking about intermediate printing with callbacks. Again, in practice it all comes with more complexity.


> The current implementation has a fallback in place which uses get_msec() to record the timings when device-side polling is disabled. 

@avmanerikar I want to mention that this approach is not reliable - and I suggest to not use it to avoid misleading. In simple words, if we don't have profiling events and don't have synchronization then using callbacks can't provide reliable times for kernel execution. Below is an example. Say we have three single-kernel primitives, the user submits all three back-to-back:

```
submit primitive/kernel A
submit primitive/kernel B
submit primitive/kernel C
```

Without profiling events, `start_ms` for `C` is measured at submission time. Submission is relatively lightweight so `A`, `B` and `C` submission timestamps are close to each other. However, if kernels `A` and `B` are long then the `C` callback  would also include times for `A` and `B` and report the whole duration as time of `C` execution - this is misleading.

vpirogov: > It won't be trivial - without synchronization and without profiling events the host-side measurements can be inaccurate. Can we disable time measurement in the SYCL Graph (when profiling events are not available)?

No free lunch... Thanks, @echeresh.

avmanerikar: @echeresh Thanks for the explanation! There is one more consideration remaining before switching to an event-polling based mechanism:

Currently, the stream profiling capabilities in oneDNN must be enabled by setting the appropriate flags on the underlying command queue and then passing that queue to the stream. Since the asynchronous mode (for both proposed methods) will require stream profiling to function accurately, its usage would be limited to scenarios where users create profiling-enabled streams via OpenCL/SYCL interop.

In this case, it would be preferable to have a more convenient way to enable profiling at stream creation - say, via a runtime variable like `ENABLE_ASYNC_VERBOSE` - which would ensure the underlying OpenCL/SYCL queues have the correct profiling properties enabled (unless there is a design limitation preventing this).

On a related note, the stream profiler created by enabling `ONEDNN_EXPERIMENTAL_PROFILING` already captures timing information in much the same way as the proposed event-polling mechanism. If we had a convenient way to define profiling-enabled streams, could we leverage the existing experimental profiler to implement asynchronous verbose logging directly?

@mgouicem, please feel free to correct me if I'm missing something.

echeresh: @avmanerikar Yeah, these are good questions.

Now, we are in a situation when GPU profiling can be used for two purposes:

1. `ONEDNN_EXPERIMENTAL_PROFILING` feature, in particular `dnnl_query_profiling_data()` API
2. For `ONEDNN_VERBOSE` time tracking purposes

Strictly speaking, these two scenarios are independent. On the other hand, if we need to track events in both cases, then we have an option to extend the existing profiler implementation to also be used for verbose-related tracking because in both cases we need to track events and aggregate information from them.

@avmanerikar As you mentioned, one issue with the current state of things is that profiling capabilities are only queried when the user builds with `ONEDNN_EXPERIMENTAL_PROFILING`:

https://github.com/uxlfoundation/oneDNN/blob/f4016359548f16fbf97e4354da863384b51fddd0/src/xpu/ocl/stream_impl.hpp#L88-L91

There are various options and considerations here, but what I have in mind is the following:

- Introduce a special flag in `stream_t` to mean "async verbose requested"
    - This flag is set when a) the queue (either user queue or created with `dnnl_stream_profiling`) is created with profiling capabilities and b) `ONEDNN_VERBOSE` requires time tracking
- Introduce a special `verbose_profiler` in `src/gpu/gpu_stream.hpp` (actually we can rely on whether it's `nullptr` or not for "async verbose requested" flag
    - Here, we need to also extend the profiler to map a primitive `stamp` to a verbose line
- We will record events for the verbose async tracking as well
- Now, at `stream.wait()` or stream destruction we do synchronization and call `stream_profiler_t::get_info()` and print verbose lines and clear the entries afterwards
    - To prevent large memory consumption, we may also want to do periodic event flushing to check whether they are completed, print their verbose lines and release them

I noticed that the existing stream profiler is not in the thread-local stream context and instead uses recursive locking to ensure thread safety. But I think this may be fine for the initial version.

@avmanerikar @mgouicem @petercad What do you think about this path? Any other suggestions?

avmanerikar: @echeresh, thanks for enumerating the considerations. 
> I noticed that the existing stream profiler is not in the thread-local stream context and instead uses recursive locking to ensure thread safety. But I think this may be fine for the initial version.

Following from the previous[ comment](https://github.com/uxlfoundation/oneDNN/pull/3055#discussion_r2279773864) and after an offline discussion with @mgouicem, extending the stream profiler does serve as a good option for asynchronous verbose tracking without the introduction of new timing data structures. Pending any further comments, I plan to switch to this method for implementation.

> * Now, at `stream.wait()` or stream destruction we do synchronization and call `stream_profiler_t::get_info()` and print verbose lines and clear the entries afterwards

This can also be implemented by utilizing the knowledge of the output event to print the verbose info:
https://github.com/uxlfoundation/oneDNN/blob/50577007bc1a6df25183be97f2f70ec5045ea8bf/src/xpu/ocl/stream_impl.cpp#L266

>   * To prevent large memory consumption, we may also want to do periodic event flushing to check whether they are completed, print their verbose lines and release them

With periodic event flushing, there is still the consideration about multiple kernels and out-of-order queues as it will not account for the scenario described in the previous [discussion](https://github.com/uxlfoundation/oneDNN/pull/3055#discussion_r2252630038). But this can be scope for future coverage.

echeresh: @avmanerikar Thanks for the summary. Then looks like we agreed on the following steps.

> With periodic event flushing, there is still the consideration about multiple kernels and out-of-order queues as it will not account for the scenario described in the previous https://github.com/uxlfoundation/oneDNN/pull/3055#discussion_r2252630038. But this can be scope for future coverage.

Event flushing can be opportunistic - for that we need to check the first event in the list, extract the related primitive execution stamp and then query all events belonging to this stamp. If all are complete then print a verbose line, drop these events and move to the next available event. Whenever we find a pending event - just break, allow further execution and re-check during the next event flushing.

mgouicem: It seems we are discussing a few distinct things and would like to separate these for clarity (and hopefully settle on some of those to move forward).

**Granularity at which we collect performance data** `parallel_for` does not seem to be the right place to do so (as discussed above). `stream_t::enqueue_primitive` or `primitive_execute` seem to be good spots as they could trigger start/end of collection for a whole primitive, and we also get out_event to track completion. Do we all agree on that?

**What abstraction to collect performance data**. Here I 100% agree with Eugene proposal to use stream_profiler for both VERBOSE and ONEDNN_EXPERIMENTAL_PROFILING APIs. It seems we agree on that (right?).
Whether it should be extended when stream has no profiling events is up for discussion. IMHO, it could be with something trivial if we don't care about accuracy (e.g. take host timestamp upon first event record of stream_profiler, and timestamp upon out_event completion in callback for example). But printing nothing or 0 is another option too. IMHO Main point here should be to get primitive_desc info, and not disturb GPU execution with extra sync.

**When to print data**. To trigger printing, we could either do callbacks or event polling.
@echeresh @petercad Could you help with input for this one?

Event polling:
- pros: 
  - some level of control for host overhead
- cons: 
  - Increased events lifetime (impact memory consumption, and could impact runtime overheads too as runtime impl cannot reclaim them).
  - If polling on `stream::wait` / `~stream`, could lead to memory leaks (e.g. some applications might never call stream::wait and rely on runtime object events or runtime wait function).
  - Polling at fixed frequency requires spawning extra thread (so a CPU threading runtime)
  - if no profiling_event, taking host timestamp on event completion might not be feasible if polling rate is low (which might be ok).
  - deferred printing. if stopping on wait or low polling rate, this might lead to lose of data when debugging (e.g. when segfaults occur).

Callback/host_task: 
- pros:
  - collected event lifetime is bounded to primitive execution lifetime
  - no extra thread to spawn by oneDNN: asynchronicity and sync with out event completion handled by runtime
  - can add timestamp collection when profiling events are not available.
  - Verbose line is printed as soon as primitive ended (it could limit loss of data in some cases).
- cons:
  - runtime overhead?
  - impact GPU execution that is being profiled?

If the above is accurate I would favor callback. @echeresh @petercad Am I missing anything in the pros/cons above (especially impact of callbacks/host_tasks on GPU execution)?

echeresh: @mgouicem Thanks for the write-up. Please see my comments below.

> Whether it should be extended when stream has no profiling events is up for discussion. IMHO, it could be with something trivial if we don't care about accuracy (e.g. take host timestamp upon first event record of stream_profiler, and timestamp upon out_event completion in callback for example). But printing nothing or 0 is another option too.

Well, this is a good question. I have a slight preference to keep the current behavior with synchronization. Yes, it's not accurate but at least it's more accurate for longer kernels and has well-defined and easy-to-explain behavior. For example, relying on host timestamps based on event records is less reliable and hence less useful. Printing time as `0` and printing upon last kernel submission is probably OK too but I'm not sure what we gain by changing the current default with synchronization and host side timing.

> Polling at fixed frequency requires spawning extra thread (so a CPU threading runtime)

In the initial version I would keep this in the main thread. Yes, this will add overhead but it shouldn't be that big - at least should be much less than with implicit synchronization. My reasoning is that even enabled queue profiling may degrade performance by ~5% so our users can't expect true E2E numbers with verbose. Some overhead is always be there.

> deferred printing. if stopping on wait or low polling rate, this might lead to lose of data when debugging (e.g. when segfaults occur).

Yeah, this was my concern too. Even with callbacks and/or OOO queues we may have weirdly ordered or delayed output. However for that to happen, the user should explicitly use profiling capabilities together with `ONEDNN_VERBOSE` - so in most cases we can assume their intention in such cases is to get event-based timing.

Debugging may be a concern, I agree.

As an observation: even now, as the execution line is printed after execution - we can't reliably identify the case at fault when segfault is inside execution. Maybe we can introduce "exec_entry" lines for that purpose, e.g. enabled with `ONEDNN_VERBOSE=all`. Or just some additional output even in the default case for the purpose of improved debugging.

Regarding callbacks my main concern is that this is going to be a more complex solution:

- I'm not sure we have guarantees on callback ordering in case of multiple kernels. Probably they are called in-order in practice but I guess it's not specified anywhere. We may need to have synchronization inside callbacks and logic to identify what callback is the last one to aggregate all events
- Callbacks are probably called in a separate thread - will we need synchronization to ensure output from CPU and GPU primitives is not intermixed?
- L0 doesn't have native API for callbacks and a related request was rejected: https://jira.devtools.intel.com/browse/GSD-6333. What will we do for proper L0 runtime support? Assuming L0 support is in our plans (?)

As for the pros you mentioned:

> - collected event lifetime is bounded to primitive execution lifetime
> - no extra thread to spawn by oneDNN: asynchronicity and sync with out event completion handled by runtime
> - can add timestamp collection when profiling events are not available.
> - Verbose line is printed as soon as primitive ended (it could limit loss of data in some cases).

1 - can be mitigated by explicit flushing done periodically
2 - can be done in the main thread
3 - I wouldn't rely on that as times may be far off (as commented above)
4 - agree here

Regarding the event polling and switching to asynchronous verbose I have two concerns:

- Explicit event polling may add significant overhead
    - However, it's unclear how much is the overhead. For example, if we end up with ~5% E2E impact and getting accurate device times, it's a good deal
- Delayed/mixed verbose output, e.g. executed as GPU primitive -> CPU primitive but will be printed in reverse
    - Well, with events and asynchronous execution we have to accept this. Callbacks can mitigate that though (as well as more frequent flushing) but this is just the implication of asynchronous execution

Taking all this into account, I'm leaning towards the event polling solution. IMO, it can help us achieve the end goal with less effort and have acceptable behavior.

mgouicem: Thanks @echeresh 

So can I assume we are aligned on:
- reusing stream_profiler abstraction for collecting performance data
- gathering time to print in verbose at primitive level (enqueue_primitive) instead of kernel submission (parallel_for).

So it leaves open 2 questions:
- what to print: host-time, device time from profiling events, or a mix of both
- which mechanism to trigger prints: callback/polling.

I'll try to answer your point on those two questions:

**What to print**

>  I have a slight preference to keep the current behavior with synchronization. Yes, it's not accurate but at least it's more accurate for longer kernels and has well-defined and easy-to-explain behavior.

I think the main point here is not about performance measurement accuracy. oneDNN users seem to ask we drop synchronizations, otherwise oneDNN verbose interferes too much with program execution as a whole. I agree there will be some overhead anyway, but it should be better / less intrusive than explicit synchronization for each primitive execution.

Regarding well-defined / easy-to-explain behavior, I agree that printing different data based on the nature of the stream can be awkward. This seems to stem from the ask to print profiling event timings (vs host-time we have today). Here are a few options if we want to simplify:
- only print host-time through ONEDNN_VERBOSE, if user wants kernel time, they could rely on onetrace or other profiling tool (essentially, we drop the requirement to print profiling event data).
- by default print host-time, and introduce ONEDNN_VERBOSE_MODIFIER env variable so that user can ask for data from profiling events (and we would print 0 when those are not available). That would also allow user to specify how multi-kernel primitive should aggregate their timings, similar to the experimental API we currently have.

> For example, relying on host timestamps based on event records is less reliable and hence less useful.

I was just suggesting to take only 2 timestamps on host, like we do today: one at the start of enqueue_primitive (or first `stream_profiler::record` call in enqueue_primitive, if we want to hide all profiling in stream_profiler), and one at the end (in call back function). I am not suggesting to take host timestamps for every `stream_profiler::record` call. 


> As an observation: even now, as the execution line is printed after execution - we can't reliably identify the case at fault when segfault is inside execution. Maybe we can introduce "exec_entry" lines for that purpose, e.g. enabled with ONEDNN_VERBOSE=all. Or just some additional output even in the default case for the purpose of improved debugging.

That is orthogonal to async verbose but a good point nonetheless. I opened [MFDNN-14095](https://jira.devtools.intel.com/browse/MFDNN-14095) to track.

**which mechanism to trigger prints**

Now wrt callback function:
> I'm not sure we have guarantees on callback ordering in case of multiple kernels.

I don't think so but do we need that property?
- if you are referring to multiple kernels within the same primitive, it does not matter: we need to attach a callback to the output event of the enqueue_primitive call only (which is the event we return from primitive execution), not to all events.
- if you are referring to multiple kernels across multiple primitives, there is no guarantee in their order of completion either for out-of-order or multiple in-order streams, so why guarantee that for verbose prints?

> Callbacks are probably called in a separate thread - will we need synchronization to ensure output from CPU and GPU primitives is not intermixed?

Synchronization indeed is necessary and typically handled at the logging level (printf and spdlog are thread-safe). Note that it is not specific to callback, we already have to handle case where multiple threads issue calls to oneDNN.

> What will we do for proper L0 runtime support? Assuming L0 support is in our plans (?)

There are plans to support callback on host in level-zero, see NEO-14577.

Now regarding the concern about complexity, this PR actually relies on host_task enqueuing and callback. It should not be too intrusive as you can see. The main change required IMHO is to add that callback/host_task in `enqueue_primitive`/`primitive_execute` to the primitive output event instead of `parallel_for`.

Regarding event polling: 
> 1 - can be mitigated by explicit flushing done periodically
> 2 - can be done in the main thread

Those two are contradictory. If using only the main thread, we cannot poll at fixed interval of time, we can do it only upon oneDNN API calls. And because that depends on application, it essentially means we have no control over the lifetime of these events.

Also, using callbacks allows to remove need from introducing a state: we can just allocate that on heap and callback execution will free the recorded events.
With polling, we will need to store those recorded events in the stream (and it would be harder to make stream stateless down the line), or globally and make it thread-safe.

If we poll at each call to oneDNN, it also has higher overhead than host_task/callback since those are typically executed in a separate thread.

echeresh: @mgouicem Thanks for the comments. I'm fine with the callback solution if you see it being a better option. My comments in this thread were largely about the correctness of the callback solution for various scenarios. As long as we have a solid implementation in mind - callback vs event polling is not that important, IMO.

@petercad @rjoursler If you see any issues with the callback solution for device time tracking, please share.

> oneDNN users seem to ask we drop synchronizations, otherwise oneDNN verbose interferes too much with program execution as a whole.

I clicked through the "Issue Links" for [MFDNN-12088](https://jira.devtools.intel.com/browse/MFDNN-12088), and my impression is that it not the main concern. AFAIU, the main concern is that verbose numbers do not match device measured times via benchdnn. I think @petercad commented similarly as well (if I understood it right) that synchronization by itself is not that big of a problem when we can't use device timing.

> I was just suggesting to take only 2 timestamps on host, like we do today: one at the start of enqueue_primitive (or first stream_profiler::record call in enqueue_primitive, if we want to hide all profiling in stream_profiler), and one at the end (in call back function). I am not suggesting to take host timestamps for every stream_profiler::record call.

If I understand you right, this may be very inaccurate in case of back-to-back submissions, see my comment [here](
https://github.com/uxlfoundation/oneDNN/pull/3055#discussion_r2274529555).

> Note that it is not specific to callback, we already have to handle case where multiple threads issue calls to oneDNN.
> There are plans to support callback on host in level-zero, see NEO-14577.

Yep, you are right, missed that, multithreading is not a concern. Thanks for the Jira, that is I was asking about so we are good here - looks like it's in progress already.

>  The main change required IMHO is to add that callback/host_task in enqueue_primitive/primitive_execute to the primitive output event instead of parallel_for.
> Also, using callbacks allows to remove need from introducing a state: we can just allocate that on heap and callback execution will free the recorded events.

I think this should work. A few questions/comments:

- With this approach we don't need to rely on the existing profiler, correct?
- What if we have a multi-kernel OOO scenario, and the last kernel event callback is called while some of the preceding events are not complete yet?
- Even for in-order scenarios this assumes proper ordering (also multi-kernel scenario). Say, for kernels `A` -> `B`, if `B` event is complete, there should be a guarantee that `A` event is also ready at that point. I'm probably overthinking but just wanted to mention it here.
- The callback, being called in a separate thread, shouldn't access any oneDNN related state/objects as they may not be alive at that point.

petercad: Agreed, the main problem to solve is that verbose times currently don't reflect actual device execution time, and it leads to no end of confusion to our users, who expect to be able to analyze their workload's performance based on verbose timing.

So I think host overhead is not a primary concern (as long as it doesn't affect device timing significantly), though I agree we should minimize it as much as possible.

My major concern with the callbacks is that I don't know how efficiently they are implemented, and it will cause complications for L0 if we are using an API that is only now coming into existence. On the first concern -- I guess we'll just have to try it; on the second -- I guess we will be asking users to have very recent L0 drivers anyway for e.g. mutable command lists.

avmanerikar:
@echeresh, @mgouicem, @petercad. Thanks for the comments and the discussion.
I am closing this PR and have created a new implementation in #4187 based on the following recommendations:
- a `stream_profiler_t` based approach to track device measured times for primitive execution.
- calculation of aggregate execution times at primitive level instead of the kernel level.
- non-blocked printing of profiling information through asynchronous callbacks triggered for the output event obtained via `get_output_event()`.

PR #3085, opened 2025-04-13
cpu: x64: fix using new scale mask interface

densamoilov:
Fixes MFDNN-13473 (large buffer case).

The issue is not related to the large shapes directly. Using the large shapes results in falling back to this implementation that had a bug manifesting as a segfault. We recently changed the interface of scales mask but overlooked this place that required some adjustments.

### DIFF
--- a/src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_conv_kernel.cpp
+++ b/src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_conv_kernel.cpp
@@ -1229,8 +1229,12 @@ void jit_avx512_core_x8s8s32x_1x1_conv_kernel::init_scratchpad(
     using namespace dnnl::impl::memory_tracking::names;
 
     const int wei_mask = attr.scales_.get_mask(DNNL_ARG_WEIGHTS);
+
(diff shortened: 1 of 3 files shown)


ankalinin (approved):

dzarukin (approved):
Seems like we really have some gaps in CPU coverage as they don't get caught during and post-refactor validation.

I've updated some other places like that (non-1x1 avx512, and both avx2, seems like I missed this one) with a slightly different version of the code. If you feel up to align it, that would be great, but it's up to you.

densamoilov:
> Seems like we really have some gaps in CPU coverage as they don't get caught during and post-refactor validation.

All cases to go to the brgemm implementation. The only reason we caught this bug was because we skipped the brgemm implementation for extra large shapes. I'm not sure if there are some cases that are not supported by the brgemm implementation but supported by this one.

> I've updated some other places like that (non-1x1 avx512, and both avx2, seems like I missed this one) with a slightly different version of the code. If you feel up to align it, that would be great, but it's up to you.

I refactored the code a bit to make it consistent with the way you handled such cases in other implementations. Thanks for pointing this out.

PR #3089, opened 2025-04-14
cpu: aarch64: Add BF16 support for eltwise JIT via reordering to FP32

aditew01:
Backport to [#2982 ](https://github.com/uxlfoundation/oneDNN/pull/2982) 

It fixes a major performance gap when using BF16 precision on aarch64. It should directly translate to improvement in performance for a majority of models in PyTorch.

### DIFF
--- a/src/cpu/aarch64/jit_uni_eltwise.cpp
+++ b/src/cpu/aarch64/jit_uni_eltwise.cpp
@@ -119,14 +119,13 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel {
             // - unpack BF16 to FP32 by zero-extending
             // - compute eltwise alg in FP32
             // - down convert back to BF16 using bfcvt, and pack result
-            mov(ZReg(tmp0).s, P_ALL_ONE, ZReg(vmm_src).s);
-            lsl(ZReg(vmm_src.getIdx()).s, ZReg(vmm_src).s, 16);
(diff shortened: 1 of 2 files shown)


theComputeKid:
@Sqvid can we get this in v3.8?

aditew01:
cc: @spalicki , do you think we can backport this change to the release?

vpirogov:
@theComputeKid, do you feel these changes were sufficiently tested to make sure we won't break Pytorch?

vpirogov:
> I think this is a feature so it should not be included in a backport.

As oneDNN v3.8 is not released yet, adding a feature is a possibility. We just need to manage the risk, that's why I asked about validation.

aditew01:
@vpirogov @theComputeKid 
The correctness tests in `benchdnn` pass for all algorithms in BF16, which I take as a green signal.
A supporting datapoint is that we have performed integration testing as part of the ARM/Tool-Solutions merge.
Additionally, I’ve tested it with a few models with integration in PyTorch and it lgtm!  
Please let me know your thoughts / if you'd like some specific data points. 

PS: integration with PyTorch requires an iDeep change as marked here: https://github.com/intel/ideep/pull/354

theComputeKid:
Alright then. I can see that the tests are extensive, if Vadim is OK with it, I am.

theComputeKid (approved):

theComputeKid:
You need an approval from arch.

vpirogov (approved):

PR #3101, opened 2025-04-15
benchdnn: graph: fix false positives by reducing the matmul range

dzarukin:
And bumping L2 norm trh as there are still cases not addressed by reduced filling.

Should cover all cases MFDNN-13052


dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_graph

wzt1997 (approved):

avmanerikar (approved):

PR #3001, opened 2025-04-02
graph: backend: dnnl: dispatch select to binary primitive

gyhintel:
# Description

Dispatch the select implementation to the binary primitive. This PR is based on [#2989](https://github.com/uxlfoundation/oneDNN/pull/2989)

PR #3102, opened 2025-04-16
[backport][rls-v3.8]graph: backend: dnnl: dispatch select to binary primitive

gyhintel:
Backport #3001 to rls-v3.8


rongzha1 (approved):

ElaineBao (approved):

src/common/binary.cpp
```diff
@@ -137,9 +137,6 @@ status_t binary_md_check(const engine_t *engine, alg_kind_t alg_kind,
         if (src2_md != nullptr) {
             VCHECK_BINARY(utils::one_of(src2_md->dims[d], 1, dims[d]),
                     VERBOSE_BAD_DIM, "src2", d);
-            VCHECK_BINARY(IMPLICATION(src0_md->dims[d] != dims[d],
-                                  src2_md->dims[d] == src0_md->dims[d]),
-                    VERBOSE_INCONSISTENT_DIM, "src0", d, "src2", d);
```
dzarukin, line 142: I'd cherry-pick the PR associated with this change as is.

gyhintel: I cherry-pick this commit from https://github.com/uxlfoundation/oneDNN/commit/f1c4c60aeff5e3b4203421e4873cec2968aa3f8f in the main branch. What else do you think I should do?

dzarukin: Alright, I confused this change with [2d41f6f986](https://github.com/uxlfoundation/oneDNN/commit/2d41f6f98656c4982574617abb1ef3c37cb6215d).

Why do you need this condition to be removed as well? What's the use case?

gyhintel: According to the [select op](https://uxlfoundation.github.io/oneDNN/dev_guide_op_select.html) definition, src0 and src2 can be broadcast. Now the [distilbert](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/inputs/graph/complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json) case is the use case.

dzarukin: My understanding of ternary operator is that if condition has some non-unit dim, same value for that dim should be present either in src0 or in src1, otherwise, it doesn't make any sense as operation.
This particular condition might be a little bit more restrictive than it should, but the condition for consistency must present.

avmanerikar: > My understanding of ternary operator is that if condition has some non-unit dim, same value for that dim should be present either in src0 or in src1, otherwise, it doesn't make any sense as operation. This particular condition might be a little bit more restrictive than it should, but the condition for consistency must present.
The ternary operator will still work as expected with the second check removed. The  check in [L127-L141](https://github.com/uxlfoundation/oneDNN/blob/538d7d5390c78624773a1902d62426155269b520/src/common/binary.cpp#L127-L141) only ensure that the non-unit src2 dimensions default to src0 which is not strictly required.

dzarukin: With your guarantees, I'll resolve the discussion.

dzarukin (approved):

PR #3107, opened 2025-04-16
cpu: x64: ip: fix integer overflow in offsets

amakarev:
Previously, a large offset exceeding INT_MAX would trigger an assertion failure, leading to a segmentation fault. This change modifies the behavior to call a fallback function that correctly handles offsets exceeding the standard range

### DIFF
--- a/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
@@ -186,6 +186,8 @@ private:
     const reg64_t reg_converted_stride = rsi;
     const reg64_t reg_zp_comp_pad_a = rsi;
 
+    const reg64_t reg_long_offt = r11;
+
     constexpr static int abi_param1_offs_ = 0;
     constexpr static int reg_zp_comp_a_offs_ = 8;
     constexpr static int reg_zp_comp_b_offs_ = 16;
@@ -1004,7 +1006,8 @@ void jit_brgemm_amx_uker_base_t::apply_post_ops_to_range(
 
                 auto zmm = accm(bd);
                 const auto d_offset = D_offset(bi, bdb, bd, ldb_pos);
-                auto addr = EVEX_compress_addr(reg_D, d_offset);
+                auto addr = EVEX_compress_addr_safe(
+                        reg_D, d_offset, reg_long_offt);
 
                 cvt2ps(brg.sum_dt, zmm_prev_dst, addr, true, false, k_mask);
                 if (p_sum_zp_reg_set) vsubps(zmm_prev_dst, zmm_sum_zp);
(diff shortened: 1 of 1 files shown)


src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1481,7 +1484,7 @@ void jit_brgemm_amx_uker_base_t::store_vector(
     const auto d_offset = D_offset(bi, bdb, inp_bd, ldb_pos);
 
     auto ptr_C = EVEX_compress_addr(reg_C, c_offset);
-    auto ptr_D = EVEX_compress_addr(reg_D, d_offset);
+    auto ptr_D = EVEX_compress_addr_safe(reg_D, d_offset, reg_long_offt);
```
densamoilov, line 1487: Is the problem only with strides for tensor D or other tensors are also affected?

amakarev: i will check

amakarev: I think there might be a problem with other tensors too. added changes

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -186,6 +186,8 @@ struct jit_brgemm_amx_uker_base_t : public jit_base_brgemm_kernel_t {
     const reg64_t reg_converted_stride = rsi;
     const reg64_t reg_zp_comp_pad_a = rsi;
 
+    const reg64_t reg_long_offt = r11;
```
dzarukin, line 189:
```suggestion
    const reg64_t reg_long_offt = reg_A;
```

And a comment why it's safe to use `reg_A` at the point it's used would be nice.

densamoilov: Honestly, I think creating an alias using another alias is a bad practice. Leaving a comment on why it's ok to share r11 should be sufficient and not confusing.

densamoilov (approved):

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
dzarukin: Just followed the consistent practice, but I don't mind. Wish we had an allocator for that...

amakarev: @dzarukin i assume that this is Caller-saved register according to System V ABI.

dzarukin: If you were replying to allocator piece, that's not what I meant. An allocator supposed to solve a manual register search problem and find a free one at runtime without developer's attention.

ankalinin (approved):

PR #3110, opened 2025-04-16
xe: jit: gemm: improve debugability

rjoursler:
Enables source location forwarding for some simple wrapper functionality. This enables better debugging by enabling developers to focus on the higher level details when using a debugger.


rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

src/gpu/intel/jit/gemm/generator/pieces/gemm_setup.cxx
```diff
@@ -59,11 +59,11 @@ void BLASKernelGenerator<hw>::gemmCheck32(const GEMMProblem &problem, GEMMStrate
...
     auto flag = state.raVFlag.alloc();
 
-    auto mulHigh = [&](Subregister dst, Subregister src0, Subregister src1) {
+    auto mulHigh = [&](Subregister dst, Subregister src0, Subregister src1, SourceLocation loc = {}) {
         if (emulate)
-            emul32High(1, dst.ud(), src0, src1);
+            emul32High(1, dst.ud(), src0, src1, loc);
         else
-            mul(1, dst, src0, src1);
+            mul(1, dst, src0, src1, loc);
```
petercad, line 66: I wonder if we should be thinking about a more general mechanism for "scoping" location information in this way, instead of passing around lots of SourceLocation objects. Something like this:

```c++
auto mulHigh = [&](..., SourceLocationScoper scoper = {this}) {
// rest of code as before
};
```

And then `SourceLocationScoper` is defined something like:

```c++
template <HW hw>
class SourceLocationScoper {
    SourceLocation loc;
    BinaryCodeGenerator *generator = nullptr;
public:
    explicit SourceLocationScoper(ngen::BinaryCodeGenerator<hw> *g)
                : loc{std::source_location::current()}, generator(g) {
        g->enterLocationScope(loc);
    }
    ~SourceLocationScoper() {
        generator->exitLocationScope();
    }
}
```
The new generator methods `{enter,exit}LocationScope` would set/clear a location override (new variable in the generator class). You'd have a counter as well to properly support nested scopes (only the outermost is honored).

rjoursler: I don't quite see what the benefit is here.  Is this just supposed to be a performance optimization? For release builds, Source Location will be an empty object, so I expect it would mostly be optimized away anyway. If the goal is to just avoid forwarding `loc` in the source code, I don't see much benefit as forwarding the same location for a "large" operation seems misguided anyway. 

In general, I agree we could use some improvement here. The main idea that I have considered is to encode nops for each scope so that we can traverse them in the debugger.

petercad: It's for better readability/easier programmability, rather than performance. The idea is that you only need to change one line to combine location information:

```c++
auto mulHigh = [&](..., SourceLocationScoper scoper = {this}) {
      emul32High(...);    // don't have to pass loc here
      mul(...);           // don't have to pass loc here
};
```
instead of modifying every single instruction:
```c++
auto mulHigh = [&](..., SourceLocationScoper loc = {}) {
      emul32High(..., loc);    // have to pass loc here
      mul(..., loc);           // have to pass loc here
};
```
When there are a lot of nested instructions, it's easy to miss a loc and just a lot of code changes to make. Since this pattern is appearing in lots of places, it'd be nice to simplify.

PR #3114, opened 2025-04-17
graph: backend: dnnl: fix int8 sdpa with softmax

gyhintel:
# Description

- Support int8 output softmax with scale and zp in large partition and decompose kernel
- For int8 output softmax
    - Only with scale, we would treat the scale as the dst scale
    - With scale and zp, or only zp, we would treat the scale and zp as post-binary.

### DIFF
--- a/src/graph/backend/dnnl/kernels/sdp_decomp.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp.cpp
@@ -77,11 +77,14 @@ status_t sdp_decomp_kernel_t<quantized, dt>::compile_impl(
     BACKEND_DNNL_ADD_PASS(pipeline, fuse_post_ops);
     BACKEND_DNNL_ADD_PASS(pipeline, insert_permute_for_matmul);
     if (quantized) {
+        BACKEND_DNNL_ADD_PASS(pipeline, remove_quant_data_with_no_effect);
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_scales);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_scales);
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_zero_points);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_zero_points);
-        BACKEND_DNNL_ADD_PASS(pipeline, remove_quant_data_with_no_effect);
+        // fuse those new post-binaries converted from add_zps and mul_scales
+        BACKEND_DNNL_ADD_PASS(pipeline, replace_quant_data_with_binary_post_op);
+        BACKEND_DNNL_ADD_PASS(pipeline, fuse_post_ops);
     }
     pipeline.reset_visualize_arg(true, false);
     BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_transpose_to_predecessor);
--- a/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
@@ -522,11 +522,20 @@ impl::status_t sdp_decomp_config_t::record_input_offset(
 impl::status_t sdp_decomp_config_t::record_sdp_ops(
         std::shared_ptr<subgraph_t> &sg, bool is_quantize) {
     const auto get_wei_pre_op = [](const op_ptr &op) -> op_ptr {
-        const auto out_val = op->get_input_value(1);
-        if (out_val->has_producer()) {
-            auto &producer = out_val->get_producer();
-            if (producer.get_kind() != op_kind::dnnl_reorder) return nullptr;
-            return producer.shared_from_this();
+        auto in_val = op->get_input_value(1);
+        if (in_val->has_producer()) {
+            auto *producer = &in_val->get_producer();
+            if (producer->get_kind() == op_kind::dnnl_permute) {
+                in_val = producer->get_input_value(0);
+                if (in_val->has_producer())
+                    producer = &in_val->get_producer();
+                else
+                    return nullptr;
+            }
+            if (producer == nullptr
+                    || producer->get_kind() != op_kind::dnnl_reorder)
+                return nullptr;
+            return producer->shared_from_this();
         } else
             return nullptr;
     };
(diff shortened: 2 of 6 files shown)


src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -229,6 +229,25 @@ impl::status_t sdp_decomp_config_t::construct_params(
     // softmax
     // create softmax primitive attr
     dnnl::primitive_attr sub_softmax_attr = make_primitive_attr(sdp_op[2], mgr);
+
+    dnnl_pops = {};
+    ori_dnnl_pops = sub_softmax_attr.get_post_ops();
+    for (int i = 0; i < ori_dnnl_pops.get()->len(); i++) {
+        auto alg = static_cast<algorithm>(
```
rongzha1, line 236: always add const unless you expect it will be changed later:  alg, post_shape, post_stride etc.
```suggestion
        const auto alg = static_cast<algorithm>(
```

src/graph/backend/dnnl/kernels/sdp_decomp.cpp
```diff
@@ -77,11 +77,14 @@ status_t sdp_decomp_kernel_t<quantized, dt>::compile_impl(
     BACKEND_DNNL_ADD_PASS(pipeline, fuse_post_ops);
     BACKEND_DNNL_ADD_PASS(pipeline, insert_permute_for_matmul);
     if (quantized) {
+        BACKEND_DNNL_ADD_PASS(pipeline, remove_quant_data_with_no_effect);
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_scales);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_scales);
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_zero_points);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_zero_points);
-        BACKEND_DNNL_ADD_PASS(pipeline, remove_quant_data_with_no_effect);
+        // fuse those new post-binaries converted from add_zps and mul_scales
```
rongzha1, line 85: are these new passed only works for softmax? what's the diff between last `if (quantized)` (line 65) and this one (line 79) ? could you add some comments?

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -229,6 +229,25 @@ impl::status_t sdp_decomp_config_t::construct_params(
...
+                ori_dnnl_pops.get()->entry_[i].binary.alg);
+        const dnnl::impl::memory_desc_t &ori_desc
+                = ori_dnnl_pops.get()->entry_[i].binary.user_src1_desc;
+        auto post_shape = ori_desc.dims;
+        auto post_stride = ori_desc.format_desc.blocking.strides;
+        auto post_dt = static_cast<memory::data_type>(ori_desc.data_type);
+        dims post_stride_dims = dims(post_stride, post_stride + ori_desc.ndims);
+        auto new_sub_md = memory::desc({1, 1, post_shape[2], post_shape[3]},
+                post_dt, post_stride_dims);
```
ElaineBao, line 245: Why we cannot directly use post_shape here?

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -519,9 +561,12 @@ impl::status_t sdp_decomp_config_t::record_sdp_ops(
...
         if (out_val->has_producer()) {
-            auto &producer = out_val->get_producer();
-            if (producer.get_kind() != op_kind::dnnl_reorder) return nullptr;
-            return producer.shared_from_this();
+            auto *producer = &out_val->get_producer();
+            if (producer->get_kind() == op_kind::dnnl_permute)
+                producer = &producer->get_input_value(0)->get_producer();
+            if (producer && producer->get_kind() != op_kind::dnnl_reorder)
+                return nullptr;
+            return producer->shared_from_this();
```
ElaineBao, line 569: Do we have the case that producer is nullptr at L567? Then `return producer->shared_from_this()` cannot work

src/graph/backend/dnnl/passes/transform.cpp
```diff
@@ -1207,6 +1207,16 @@ status_t fuse_dst_scales(std::shared_ptr<subgraph_t> &sg) {
         if (consumers.size() != 1) continue;
         auto &next_op = consumers[0].get_op();
         if (next_op.get_kind() != op_kind::dnnl_mul_scales) continue;
+        if (impl::utils::one_of(cur_op->get_kind(), op_kind::dnnl_softmax,
+                    op_kind::dnnl_layernorm, op_kind::dnnl_groupnorm)) {
```
ElaineBao, line 1211: Can you please add a comment here, why these three ops, why it cannot be fused

src/graph/backend/dnnl/kernels/sdp_decomp.cpp
gyhintel: Yes, these two passes only work for softmax. These 2 `if(quantized)` are the same as other kernel files (eg, matmul.cpp).  It is to handle quantize op at different stages. Line 65 is to handle the src and wei dequantize. Line 79 is to handle the src and wei dequantize.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
gyhintel: Good catch. For u8s8, this producer may be nullptr. Fixed, pls take another look at it.

src/graph/backend/dnnl/passes/transform.cpp
gyhintel: Added, thanks!

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
gyhintel: Added, thanks.

ElaineBao (approved):

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3120, opened 2025-04-17
[GPU] Improve convolution CI coverage

echeresh:
This is a stopgap solution for [MFDNN-12689](https://jira.devtools.intel.com/browse/MFDNN-12689). The long-term solution is to extend random testing to cover all primitives, see more details in https://github.com/uxlfoundation/oneDNN/pull/2434.

Adjustments made:

- Updated `shapes_basic_gpu` to have more variation in shapes
    - Extended 12 -> 20 shapes, randomly generated, more variability (strides, non-1x1, padding, etc)
- Dropped/combined some higher-level cases, examples:
    - Combined some post-ops (e.g. testing two post-ops doesn't have much value if we already tested three post-ops)
    - Dropped `--dtag` - there is no much value in testing mixed combinations (like `--stag=any --dtag=axb`). Usually we have both `axb` or both `any`
    - Dropped `--mb` knob - better to have more variability in batch sizes through the shapes
    - Combined some attribute testing - e.g. if we validate both scales/post-ops together then no much value in testing them separate

The commented out shapes pending [MFDNN-13553](https://jira.devtools.intel.com/browse/MFDNN-13553) resolution.


tests/benchdnn/inputs/conv/test_conv_gpu_ci
```diff
@@ -1,83 +1,68 @@
...
+--attr-scales=
+--attr-post-ops=
+--attr-zero-points=
 
-# Wino
---reset
---mb=2,32
---stag=any,axb
+# Winograd
 --alg=wino
```
rjoursler, line 44: Given that the Winograd implementation is not being actively developed, it is likely reasonable to just drop it from CI testing and rely on nightly for its validation.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
```diff
@@ -1,83 +1,68 @@
...
 --dt=f32,bf16
---attr-post-ops=
---batch=shapes_basic_gpu
+--batch=shapes_ci_gpu
 
-## Fpmath-mode
+## FP math mode
 --dir=FWD_B,BWD_D,BWD_WB
 --dt=f32
 --attr-fpmath=tf32,bf16
```
rjoursler, line 21: `f16` math mode?

tests/benchdnn/inputs/conv/test_conv_gpu_ci
```diff
@@ -1,83 +1,68 @@
...
 --dt=f32,bf16,f16
---attr-post-ops=,relu,sum:0.25+relu:0.5
---batch=shapes_basic_gpu
+--attr-post-ops=,sum:0.25+relu:0.5
+--batch=shapes_ci_gpu
+--attr-post-ops=
+
 ## Backward
 --dir=BWD_D,BWD_WB
 --dt=f32,bf16
```
rjoursler, line 15: We support f16 in training now, maybe that should be covered?

tests/benchdnn/inputs/conv/test_conv_gpu_ci
```diff
@@ -1,83 +1,68 @@
 --reset
 
---mb=2,16
 --stag=any,axb
---dtag=any,axb
```
rjoursler, line 5: When do we check `--dtag=axb` now? It looks like we are always testing it with the default value of `any`.

> Dropped --dtag - there is no much value in testing mixed combinations (like --stag=any --dtag=axb). Usually we have both axb or both any

While generally true, mixed layouts is fairly important use case to avoid reorders for interoperation with frameworks. It seems like we should give it some testing at a minimum.

dzarukin (approved):

tests/benchdnn/inputs/conv/test_conv_gpu_ci
```diff
@@ -1,83 +1,68 @@
...
---dt=u8:s8:s32,s8:s8:f16
---stag=any,axb
---dtag=any,axb
---attr-post-ops=,sum:0.5:3+add:f32:per_oc+add:f32:per_tensor
---attr-zero-points=
---batch=shapes_basic_gpu
---attr-post-ops=
+--dir=FWD_I
+--dt=u8:s8:s32
+--attr-scales=,src:common:0.25,wei:per_oc,src:common:0.25+wei:per_oc+dst:common:2
```
rjoursler, line 35: These 4 scale combinations seems like they could be improved if we scrambled the policies a bit more.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
echeresh: `--stag=axb --dtag=any` is \*typically\* resolved to `--stag=axb --dtag=axb` by the implementation (however strictly speaking not always as there are some nuances with 1st convolution). But for clarity maybe better to have it set explicitly like:

```
--stag=axb
--dtag=axb
--batch=option_set_gpu_ci

--stag=any
--dtag=any
--batch=option_set_gpu_ci
```

> While generally true, mixed layouts is fairly important use case to avoid reorders for interoperation with frameworks. It seems like we should give it some testing at a minimum.

Good point. This is usually first/last layers - when the framework needs NHWC/NCHW specifically while keeping output/input as `any`.  I'll add limited coverage similar to the scheme above.

echeresh: Updated in the last version.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
echeresh: Added.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
echeresh: Reduced to two combinations, thanks.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
echeresh: Removed.

tests/benchdnn/inputs/conv/test_conv_gpu_ci
echeresh: Thanks, added.

echeresh:
make test
disable test_device_cpu

PR #3123, opened 2025-04-18
benchdnn: improve performance related to memory operations (fixes MFDNN-13451)

dzarukin:
MFDNN-13451

Thanks to @echeresh, who pivoted all the optimizations from the tracker and provided a PoC for them.

This PR implements 4 out of 6 of specified in the tracker. 2 not included is check_zero_pad on inputs and simplified RNG-related filling changes.

On a list attached to the tracker the speed up looks like this:
Before:
```
total: 76.37s; create_pd: 0.04s (0%); create_prim: 7.55s (10%); fill: 31.24s (41%); execute: 2.17s (3%); compute_ref: 9.06s (12%); compare: 23.82s (31%);
```
After:
```
total: 54.19s; create_pd: 0.04s (0%); create_prim: 7.19s (13%); fill: 19.07s (35%); execute: 2.13s (4%); compute_ref: 8.65s (16%); compare: 15.23s (28%);
```

### DIFF
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -49,60 +49,61 @@ extern "C" dnnl_status_t dnnl_memory_desc_set_data_type(
         dnnl_memory_desc_t memory_desc, dnnl_data_type_t data_type);
 
 dnn_mem_t::dnn_mem_t(const_dnnl_memory_desc_t md, dnnl_engine_t engine,
-        const handle_info_t &handle_info) {
+        bool prefill, const handle_info_t &handle_info) {
     if (query_md_ndims(md) > 0) {
         auto status = dnnl_memory_desc_clone(&md_, md);
         (void)status;
         assert(status == dnnl_success);
-        active_ = (initialize(engine, handle_info) == OK);
+        active_ = (initialize(engine, prefill, handle_info) == OK);
     }
 }
 
 dnn_mem_t::dnn_mem_t(const_dnnl_memory_desc_t md, dnnl_data_type_t dt,
-        const std::string &tag, dnnl_engine_t engine) {
+        const std::string &tag, dnnl_engine_t engine, bool prefill) {
     const int ndims = query_md_ndims(md);
     if (ndims > 0) {
         auto md_wrapper = dnn_mem_t::init_md(ndims, query_md_dims(md), dt, tag);
         md_ = md_wrapper.release();
-        active_ = (initialize(engine) == OK);
+        active_ = (initialize(engine, prefill) == OK);
     }
 }
 
 dnn_mem_t::dnn_mem_t(const_dnnl_memory_desc_t md, dnnl_data_type_t dt,
-        const dnnl_dims_t strides, dnnl_engine_t engine) {
+        const dnnl_dims_t strides, dnnl_engine_t engine, bool prefill) {
     const int ndims = query_md_ndims(md);
     if (ndims > 0) {
         auto status = dnnl_memory_desc_create_with_strides(
                 &md_, ndims, query_md_dims(md), dt, strides);
         (void)status;
         assert(status == dnnl_success);
-        active_ = (initialize(engine) == OK);
+        active_ = (initialize(engine, prefill) == OK);
     }
 }
 
 dnn_mem_t::dnn_mem_t(int ndims, const dnnl_dims_t dims, dnnl_data_type_t dt,
-        const std::string &tag, dnnl_engine_t engine) {
+        const std::string &tag, dnnl_engine_t engine, bool prefill) {
     if (ndims > 0) {
         auto md_wrapper = dnn_mem_t::init_md(ndims, dims, dt, tag);
         md_ = md_wrapper.release();
-        active_ = (initialize(engine) == OK);
+        active_ = (initialize(engine, prefill) == OK);
     }
 }
 
 dnn_mem_t::dnn_mem_t(int ndims, const dnnl_dims_t dims, dnnl_data_type_t dt,
-        const dnnl_dims_t strides, dnnl_engine_t engine) {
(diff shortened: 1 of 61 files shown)


dzarukin:
make test

rjoursler (approved):

TaoLv (approved):

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -229,6 +231,10 @@ size_t dnn_mem_t::sizeof_dt() const {
     return dnnl_data_type_size(dt());
 }
 
+float dnn_mem_t::get_f32_elem(int64_t idx) const {
```
echeresh, line 234: Better to move both `{get,set}_f32_elem()` to the header for better inlining. For example I see this case gets ~2x speed-up on LNL after moving them to the header:

```
./benchdnn -v5 --conv --engine=gpu --fast-ref=0 --dir=FWD_D mb32ic128ih208oc64oh208kh1ph0
# total: 10.94s; create_pd: 0.00s (0%); create_prim: 0.03s (0%); fill: 0.44s (4%); execute: 0.19s (2%); compute_ref: 9.49s (87%); compare: 0.33s (3%);

# When defined in the header -> inlined:
# total: 5.87s; create_pd: 0.00s (0%); create_prim: 0.02s (0%); fill: 0.32s (5%); execute: 0.20s (3%); compute_ref: 4.43s (76%); compare: 0.36s (6%);
```

echeresh (approved):

tests/benchdnn/dnnl_memory.cpp
rjoursler: @echeresh, do you get the same results if we just [enable link time optimization](https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html)? If so, maybe we should just enable that so we don't need to worry too much about code location in regards to performance.

dzarukin: LTO won't help much with that, it's rather C++17 that makes GCC do better magic with headers. At least it optimized much disc space from XByak the last time I checked it, must do something about performance, too.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -49,60 +49,61 @@ extern "C" dnnl_status_t dnnl_memory_desc_set_data_type(
         dnnl_memory_desc_t memory_desc, dnnl_data_type_t data_type);
 
 dnn_mem_t::dnn_mem_t(const_dnnl_memory_desc_t md, dnnl_engine_t engine,
-        const handle_info_t &handle_info) {
+        bool prefill_memory, const handle_info_t &handle_info) {
```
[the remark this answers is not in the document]
dzarukin: Sure thing, applied, thanks.

tests/benchdnn/dnnl_memory.hpp
```diff
@@ -212,7 +222,18 @@ struct dnn_mem_t {
     int initialize_memory_create_opencl(const handle_info_t &handle_info);
     int initialize_memory_create(const handle_info_t &handle_info);
 
-    int initialize(dnnl_engine_t engine,
+    // `Prefill_memory` is a flag that controls whether the underlying memory
```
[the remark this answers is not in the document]
dzarukin: Applied along with the previous change.

tests/benchdnn/dnnl_memory.cpp
dzarukin: > Better to move both `{get,set}_f32_elem()` to the header for better inlining. For example I see this case gets ~2x speed-up on LNL after moving them to the header:
> 
> ```
> ./benchdnn -v5 --conv --engine=gpu --fast-ref=0 --dir=FWD_D mb32ic128ih208oc64oh208kh1ph0
> # total: 10.94s; create_pd: 0.00s (0%); create_prim: 0.03s (0%); fill: 0.44s (4%); execute: 0.19s (2%); compute_ref: 9.49s (87%); compare: 0.33s (3%);
> 
> # When defined in the header -> inlined:
> # total: 5.87s; create_pd: 0.00s (0%); create_prim: 0.02s (0%); fill: 0.32s (5%); execute: 0.20s (3%); compute_ref: 4.43s (76%); compare: 0.36s (6%);
> ```

Fine with me, put into a header, thanks for checking.

dzarukin:
make test

tests/benchdnn/dnnl_memory.cpp
echeresh: @rjoursler LTO shows some improvement though it's worse - but I haven't looked further. One thing I noticed - it resulted in quite slower compilation times when enabled for the whole library.

I feel like we may step on some issues with LTO like slower compilation/linking times, potential issues with support from different compilers, etc - so it may be not enough ROI for this particular case.

tests/benchdnn/dnnl_memory.hpp
```diff
@@ -225,7 +225,18 @@ struct dnn_mem_t {
...
+    // to catch issues with no/bad access.
+    //
+    // Some memory objects created will be filled right away with different
+    // values, e.g., reference f32 memories. In that case, read/write access
+    // with a special value is a waste of memory bandwidth so developer decides
+    // when to enable/disable it.
+    //
+    // The flag is propagated to the toppest dnn_mem_t constructors. In case
+    // when in doubt, always use `true` to stay on the safe side of things.
+    int initialize(dnnl_engine_t engine, bool prefill,
```
avmanerikar, line 239: It could be useful to have `prefill` default to `true` as a precaution.

dzarukin: I purposely didn't make that argument with a default value so that developers would check and think what value to put in. After all, the default value wouldn't save from misusing the `false` value.

avmanerikar (approved):

dzarukin:
make test

PR #3129, opened 2025-04-19
examples: graph: add int8 sdpa example

TaoLv:
Draft an example for int8 SDPA.

MFDNN-13484.


TaoLv:
make test
disable benchdnn_all

PR #3136, opened 2025-04-22
rfcs: add gemm external zero point precompute

hidefromkgb:
This is a proposal for user-supplied precomputed zero points, primarily intended for boosting GEMM performance in LLMs.

@dzarukin, @mgouicem, @vpirogov, please review.


rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+by the K dimension to multiply the resulting reductions to the ZPs extracted from B; see formulas below.
+
+Calculating partial K reductions of A within the GEMM kernel would never be efficient; there is, however,
+a way to perform them painlessly while downcasting A. Everything that needs adding to GEMM to use them is
+an argument identifier to take that reduction buffer from the user.
+
+The general case, trivial but impractical:
+
+```math
+C_{m,n}=\sum_{k=0}^{K-1}{A_{m,k}(B_{k,n}-Z_B(k))}=\sum_{k=0}^{K-1}{A_{m,k}B_{k,n}-\sum_{k=0}^{K-1}A_{m,k}Z_B(k)
```
mgouicem, line 22:
```suggestion
C_{m,n}=\sum_{k=0}^{K-1}{A_{m,k}(B_{k,n}-Z_B(k))}=\sum_{k=0}^{K-1}{A_{m,k}B_{k,n}}-\sum_{k=0}^{K-1}A_{m,k}Z_B(k)
```

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+a novel approach has been proposed which is to dynamically downcast both A and B matrices from a floating
+point type to INT8 — but to avoid losing too much precision the downcasting algorithm should also produce
+zero points.
+
+Of particular interest is the variation where the B matrix, or "weights", which is usually constant, gets
+downcast ahead of time. The problem is that zero points on B mean that A (the "sources") is to be reduced
+by the K dimension to multiply the resulting reductions to the ZPs extracted from B; see formulas below.
+
+Calculating partial K reductions of A within the GEMM kernel would never be efficient; there is, however,
+a way to perform them painlessly while downcasting A. Everything that needs adding to GEMM to use them is
```
mgouicem, line 16: Note that oneDNN does not provide a reduction post-op, so the user will have to fuse downcast+reduction to the previous kernel themselves. Is that ok for the target usage?

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+For modern LLM networks, which usually are a bunch of relatively small GEMMs, performance is priority #1.
+To optimize LLMs for performance (i.e. use DPAS and DP4A, the hardware matmul instructions of Intel GPUs)
+a novel approach has been proposed which is to dynamically downcast both A and B matrices from a floating
+point type to INT8 — but to avoid losing too much precision the downcasting algorithm should also produce
+zero points.
+
+Of particular interest is the variation where the B matrix, or "weights", which is usually constant, gets
+downcast ahead of time. The problem is that zero points on B mean that A (the "sources") is to be reduced
+by the K dimension to multiply the resulting reductions to the ZPs extracted from B; see formulas below.
```
mgouicem, line 13: Could you share examples/data where it is more beneficial to do extra conversion+reduction in hot path dynamically vs using f16/bf16 directly?
Is it for platforms that don't have dpas and only dp4a?

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Calculating partial K reductions of A within the GEMM kernel would never be efficient; there is, however,
```
mgouicem, line 15: I guess it depends: if user has to call an extra reduction kernel before the GEMM, could it actually be more efficient to do the reduction during input packing for DPAS instruction?

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+```diff
+ /// Dropout RNG seed value passed via a buffer.
+ #define DNNL_ARG_ATTR_DROPOUT_SEED 511
+ 
+-/// Output scaling factors provided at execution time.
+-/// Deprecated value.
+-#define DNNL_ARG_ATTR_OUTPUT_SCALES 513
++/// The argument marked with this attribute is user-precomputed.
++/// Used exclusively for zero points at the time of writing.
++#define DNNL_ARG_ATTR_USER_PRECOMP 512
```
mgouicem, line 47: (nit) given that you are not proposing to reuse the OUTPUT_SCALES value, I guess the removal of OUTPUT_SCALES is orthogonal to your proposal.

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+point type to INT8 — but to avoid losing too much precision the downcasting algorithm should also produce
+zero points.
+
+Of particular interest is the variation where the B matrix, or "weights", which is usually constant, gets
+downcast ahead of time. The problem is that zero points on B mean that A (the "sources") is to be reduced
+by the K dimension to multiply the resulting reductions to the ZPs extracted from B; see formulas below.
+
+Calculating partial K reductions of A within the GEMM kernel would never be efficient; there is, however,
+a way to perform them painlessly while downcasting A. Everything that needs adding to GEMM to use them is
+an argument identifier to take that reduction buffer from the user.
```
mgouicem, line 17: IIUC, B is constant in the case of interest, and user should already have zero-points of B known when computing the reduction of A blocks.
If that is the case, why can't they pre-compute the whole compensation term(`-zp(B)*block_sum(A)`), and pass that instead?

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+ /// number of source arguments.
+```
+
+```diff
+ 
+ /// Zero points provided at execution time.
+ #define DNNL_ARG_ATTR_ZERO_POINTS 8192
++/// Convenience constant for execution-time user-precomputed zero points.
++#define DNNL_ARG_ATTR_USER_PRECOMP_ZERO_POINTS \
++    (DNNL_ARG_ATTR_USER_PRECOMP | DNNL_ARG_ATTR_ZERO_POINTS)
```
mgouicem, line 59: For the case at hand, it seems the user can precomp the whole compensation term, and pass that instead.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: Yes, as the OpenVINO team who proposed the op already have this primitive implemented.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: AFAICT on both Xe2LPG and Xe2HPG this is exactly the case: on several LLMs OpenVINO focuses on, reduction + INT8 with no ZP takes less per-network time than raw FP16, even though not by much, ~15%.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: Since we typically need many partial sums per K, one sum per group, it has potential to get nasty very fast due to register pressure — and that's even without touching the reduction ops themselves.
To reduce efficiently, the kernel has to read A in patterns wildly different than those used in GEMM.

hidefromkgb: @petercad, please check whether my take is correct.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: If we leave `DNNL_ARG_ATTR_OUTPUT_SCALES` be, it's going to equal `DNNL_ARG_ATTR_USER_PRECOMP | DNNL_ARG_SRC`, which is the most common ZP usecase.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: Well, they certainly can, but what if the same A is used in multiple GEMMs with different Bs?
The proposed scheme, though, isn't getting in the way if the user decides to precompute everything and just pass a dummy scalar B ZP equal to 1.

rfcs/20250416-gemm-external-zp-precompute/README.md
hidefromkgb: Yes, this is definitely possible, but in this case the user can just pass a non-grouped precomputed buffer for A and a scalar identity ZP on B.

rfcs/20250416-gemm-external-zp-precompute/README.md
petercad: Andrey is right. The problem is that DPAS needs its src2 data in k-major format, so these partial k-reductions are horizontal (cross-lane) reductions, which is very slow on current hardware, even if we managed to share the work cooperatively across threads.

rfcs/20250416-gemm-external-zp-precompute/README.md
echeresh: @petercad Can you please add more details? I think I'm missing something. From the formula `R` is already `M x G` and multiplying it by zero-points + reduction should give a vector of size `M` so it's not a matrix.

image

hidefromkgb: @echeresh, what you see here is not a matrix product, but a single element of C (column #m, row #n) being computed.

echeresh: @hidefromkgb Hmm, okay. But I see `m` index being used in the `R`-related term while there is no `n`. I assumed this is a broadcast by `n`, isn't it?

rfcs/20250416-gemm-external-zp-precompute/README.md
```diff
@@ -0,0 +1,87 @@
...
+ 
+ /// Arguments for fused depthwise convolution.
+ /// See @ref dev_guide_attributes_post_ops_depthwise_fusion
+```
+
+API-wise there is nothing more that should be added, as these 2 arguments are capable of doing everything
+related to accepting any precomputed buffers from the user. All that's needed from the user is to bitwise
+OR either of these constants with the buffer identifier, and pass the result to zero point related oneDNN
+primitive attribute functions (e.g. `dnnl_primitive_attr_set_zero_points`) on PD initialization, and then
```
echeresh, line 68: Can you please add an example of how `set_zero_points()` would look like? What should be the argument value? For example now, with regular zero-points, it's just `DNNL_ARG_WEIGHTS`.

hidefromkgb: I expect the precomputed A's to have an ID of `DNNL_ARG_ATTR_USER_PRECOMP | DNNL_ARG_SRC`.

rfcs/20250416-gemm-external-zp-precompute/README.md
dzarukin: To be precise, it's not a "primitive" but a custom kernel that performs dynamic quantization of the tensor A which needs to compute max for scales, I presume, and as it does so, will also collect a partial sums values into a dedicated tensor.

rfcs/20250416-gemm-external-zp-precompute/README.md
dzarukin: Please expand the whole call to `set_zero_points()` because it would definitely have a different semantics.

As a side point - it may likely require a dedicated API to avoid confusion.

rfcs/20250416-gemm-external-zp-precompute/README.md
mgouicem: @petercad I guess the question is about what are the shapes we are caring about:
- precomputing the full compensation term into bias is M x N elements (and user does not need to pass B zero-points)
- taking col_sum for A is M x (K / G), and zero-points for B is N x (K/G) elements. So a total of (K/G) x (M + N) elements.

So for larger shapes, indeed, the bias term would be bigger. But in that case, I expect we would be compute bound, and the size of the bias should be small anyway compared to A and B matrices. For example if we take something like llama2 shapes like M=N=4096, and K=11008, taking the compensation into bias term directly should be negligible, no?

petercad: @mgouicem That's true, however for larger shapes you still have the compute overhead -- the amount of precomputation is on the same order as GEMM itself, $\Theta(m\cdot n\cdot k)$.

rfcs/20250416-gemm-external-zp-precompute/README.md
petercad: The quantization kernels though are almost certainly memory-bound so they should be able to absorb the extra compute from the horizontal reduction. Also the reduction computation is only performed once (quadratic time) instead of repeatedly inside GEMM (cubic time).

isanghao:
Looks good to me. Just in case, please note that A is also grouped and group_size(A) may be different from group_size(B).

vpirogov:
We'll need to address few things before this RFC gets approved:
* Demonstrate performance benefits of the proposed approach in a PoC with OpenVINO.
* Address considerations shared in https://github.com/uxlfoundation/oneDNN/pull/3222#issuecomment-3008185206
In the end the goal of this process is to make sure we have API that will be stable.

Before this happens we have several options to get this feature into OpenVINO's hands.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```math
+C_{m,n}=\sum_{g=0}^{G-1}\sum_{k={K\over{G}}g}^{{K\over{G}}(g+1)-1}{A_{m,k}(B_{k,n}-Z_B(g,n))}=\sum_{k=0}^{K-1}{A_{m,k}B_{k,n}}-\overbrace{\sum_{g=0}^{G-1}Z_B(g,n)\underbrace{\sum_{k={K\over{G}}g}^{{K\over{G}}(g+1)-1}A_{m,k}}_{R_{m,g}}}^{\zeta},\;K\mathrm{mod}{G}=0
+```
+
+, where R is the matrix of partial K reductions of A, i.e. the reduction buffer
+introduced above.
+
+*N.B.:* In the formula above, the `ζ` expression in the simplest case where
+`G = 1`, is a single product of 2 values: `Z<sub>B</sub>(0,n)` and
```
mgouicem:
```suggestion
`G = 1`, is a single product of 2 values: $Z_B(0,n)$ and
```

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+`R<sub>m,0</sub>`. That way for each `C<sub>m,n</sub>` the pairs of values are
```
mgouicem:
```suggestion
$R_{m,0}$. That way for each $C_{m,n}$ the pairs of values are
```

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+different, so precomputing all of them would mean converting the 2 vectors into
+an M by N matrix, which can be appended as bias, avoiding the extra
+manipulations written below. But this approach results in the same performance
+as doing zero-points math conventionally which is far below the target.
+
```
mgouicem, line 44: Could you spell out in RFC the requirement that the same precomputed reductions could be reused across multiple Matmul calls (e.g. same activation tensor is passed to different Matmul with different weights and zero-point groups).
That would help justify why different mask is needed and cannot be just inferred from zero-point shapes.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+The following limitations to be applied:
+* Can't be used without weights zero-points specified.
+* Must have full A matrix mask, e.g., standard M by K times K by N MatMul should
+  have the mask of 3 for this feature, meaning broadcast is not supported.
+
+This design keeps in mind the potential extension for zero-points compensation
+functionality on CPU which is limited to weights format kind "any" and can't be
+applied for any other format, while this API can help to get that case
+supported.
+
```
mgouicem, line 80: Could you add the performance data in the description to the rfc for future reference?
Or at least the main points (improves perf for int8 by XX, for int4 by XXX, ...).

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+The alternative design approach is to extend the primitive descriptor signature
+with an additional flag. It would deduce the grouping argument, the most
+critical one, based on other attributes, which doesn't provide flexibility of
+choice to the user, and at the same time mandates the check at execute time if
```
mgouicem: I would say the biggest argument against this option is that inferring pre-computed reductions from zero-points contradicts one of the requirements from openVINO.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+deduced value was actually passed by the user through the memory descriptor with
+the memory object for precomputed reductions, which may result in an error
+in case of mismatch. Such behavior is undesired. Additionally, it costs more to
+extend the signature and doesn't give a flexibility to attach it to, for example
+convolutions.
```
mgouicem, line 90: If you want to keep that option for future reference, maybe put as separate section?
Not sure if relevant though, since it does not meet requirement.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
```diff
@@ -0,0 +1,85 @@
...
+This design keeps in mind the potential extension for zero-points compensation
+functionality on CPU which is limited to weights format kind "any" and can't be
+applied for any other format, while this API can help to get that case
+supported.
+
+The alternative design approach is to extend the primitive descriptor signature
+with an additional flag. It would deduce the grouping argument, the most
+critical one, based on other attributes, which doesn't provide flexibility of
+choice to the user, and at the same time mandates the check at execute time if
+deduced value was actually passed by the user through the memory descriptor with
```
mgouicem: (nit as this option does not meet requirements) This statement of checking for mistmatch is actually true for all arguments to execute function (passed descriptor could be different than the one used for creation). There is nothing specific to this option here.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
dzarukin: Sounds good, added a sentence, thanks.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
dzarukin: Added the paragraph for performance.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
dzarukin: I put this argument in there, thanks.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
dzarukin: Put in a different paragraph and provided better justification, thanks.

rfcs/20250819-attributes-api-precomputed-reductions/README.md
dzarukin: True, I was not quite precise in words what I wanted to say, updated that sentence, thanks.

mgouicem (approved):

PR #3139, opened 2025-04-22
cpu: aarch64: Nuke winograd convolution

theComputeKid:
Trying to find out if we really need winograd. Need to run performance tests on it to know for sure.


vpirogov:
Note, that Winograd convolution algorithm in oneDNN requires opt-in with `algorithm::convolution_auto` or `algorithm::convolution_winograd`. Pytorch seem to always use `algorithm::convolution_direct`.

Sqvid:
It would be ideal if we could document the motivation, and justification for doing this in the commit message so there's a record of it in-tree.

murste01:
I've done PyTorch e2e benchmarking with & without this patch using the Torchbench benchmark suite. We don't see any notable use of Winograd, or change in performance with the introduction of this patch. I'm happy with this change from a performance POV.

Radu2k (approved):

Ryo-not-rio (approved):

theComputeKid:
@vpirogov could you approve from arch/doc?

dzarukin:
> I've done PyTorch e2e benchmarking with & without this patch using the Torchbench benchmark suite. We don't see any notable use of Winograd, or change in performance with the introduction of this patch. I'm happy with this change from a performance POV.

You may want to double check it with OpenVINO team, they were relying on it and reported several regressions when it was removed/corrupted.

Sqvid:
>You may want to double check it with OpenVINO team, they were relying on it and reported several regressions when it was removed/corrupted.

cc: @alvoron

PR #3068, opened 2025-04-10
[wip] POC for Root Mean Square (RMS) layer normalization support

mzhukova:
TBD

PR #3147, opened 2025-04-23
rfcs: extend layer norm to support root mean square normalization

mzhukova:
Proposal to extend current Layer Normalization primitive to support RMS normalization via new flag.
Note "open questions" section for providing your opinion on better naming.
[Read here](https://github.com/uxlfoundation/oneDNN/blob/mzhukova/rfcs/20251004-rms-norm/rfcs/20251004-rms-norm/README.md)

Draft of implementation: https://github.com/uxlfoundation/oneDNN/pull/3068
JIRA: https://jira.devtools.intel.com/browse/MFDNN-13287

// Note: Current proposal is aligned with keras RMS normalization, but not Layer Normalization with rms_scaling due to https://github.com/keras-team/keras/issues/21234.

### DIFF
--- a/rfcs/20251004-rms-norm/README.md
+++ b/rfcs/20251004-rms-norm/README.md
@@ -110,9 +110,10 @@ explciitly indicate what is happening with the variance.
 
 #### Compatibility with Existing Flags
 
-The combination of `dnnl_use_global_stats` and `dnnl_rms_norm` would be supported,
-but users must only provide pre-computed RMS norm statistics
-(instead of both mean and variance as required for Layer normalization).
+The combination of `dnnl_use_global_stats` and `dnnl_rms_norm` will be supported.
+However, users must provide pre-computed RMS norm statistics
+(instead of both mean and variance as required for Layer Normalization)
+using the `DNNL_ARG_VARIANCE` index.
 
 Since RMSNorm is a simplified version of LayerNorm, it works with `dnnl_use_scale`
 and `dnnl_use_shift` flags in the same manner.
--- a/rfcs/20251004-rms-norm/README.md
+++ b/rfcs/20251004-rms-norm/README.md
@@ -51,6 +51,9 @@ using frameworks like TensorFlow and PyTorch.
 Additionally, consider [Keras RMS Normalization Layer](https://github.com/keras-team/keras/blob/v3.9.2/keras/src/layers/normalization/rms_normalization.py)
 and [PyTorch RMSNorm](https://pytorch.org/docs/stable/generated/torch.nn.RMSNorm.html) APIs.
 
+Note that this proposal currently does not cover the [Keras Layer Normalization](https://keras.io/api/layers/normalization_layers/layer_normalization/)
+with the `rms_scaling=True` option, see [keras#21234 Issue](https://github.com/keras-team/keras/issues/21234).
+
 ## Proposal
 
 The proposal is to extend the LayerNorm primitive to support RMSNorm
--- a/rfcs/20251004-rms-norm/README.md
+++ b/rfcs/20251004-rms-norm/README.md
@@ -1,0 +1,177 @@
+# Proposal to Extend Layer Normalization to Support Root Mean Square Normalization
+
+## Introduction
+
+Root Mean Square Normalization (RMSNorm) is a normalization technique similar
+to Layer Normalization (LayerNorm) but computationally more efficient.
+
+In oneDNN, LayerNorm computes the mean and variance using the following formulas:
+```math
+\mu(t, n) = \frac{1}{C} \sum_{c} src(t, n, c)
+```
+```math
+\sigma^2(t, n) = \frac{1}{C} \sum_{c} (src(t, n, c) - \mu(t, n))^2
+```
+
+It then normalizes the input using these statistics:
+```math
+dst(t, n, c) = \frac{src(t, n, c) - \mu(t, n)}{\sqrt{\sigma^2(t, n) + \epsilon}}
+  * \gamma(c) + \beta(c)
+```
+Here `gamma` and `beta` are optional scale and shift parameters, and `epsilon`
+is a small constant to prevent division by zero.
+
+The purpose of computing the mean and variance is to re-center and re-scale the
+input data, which helps improve stability and convergence.
+
+RMSNorm, on the other hand, skips the re-centering step and normalizes using only
+the root mean square statistic (`RMS`):
+```math
+RMS(t, n) = \sqrt{\frac{1}{C} \sum_{c} src(t, n, c)^2 + \epsilon}
+```
+The `epsilon` is added to avoid further dividing by zero.
+Note that it is not always the part of the definition for `RMS` statistics.
+
+This results in a simpler normalization formula:
+```math
+dst(t, n, c) = \frac{src(t, n, c)}{RMS(t, n)} * \gamma(c) + \beta(c)
+```
+
+TLDR; RMSNorm could be considered a simplified version of
+LayerNorm where the mean is assumed to be zero. As a result, the variance is
+equal to `RMS^2`, and moreover if the input data is such that it has a mean of zero,
+RMSNorm and LayerNorm produce the same results.
+
+For more details, including comparisons with LayerNorm and examples of benefits,
+refer to the article by [Biao Zhang and Rico Sennrich](https://arxiv.org/abs/1910.07467).
+This article provides more details regarding the method and showcases some experiments
+using frameworks like TensorFlow and PyTorch.
+
+Additionally, consider TensorFlow
+[LayerNorm with RMS parameter](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LayerNormalization)
+and PyTorch [RMSNorm](https://pytorch.org/docs/stable/generated/torch.nn.RMSNorm.html) APIs.
+
+## Proposal
+
+The proposal is to extend the LayerNorm primitive to support RMSNorm
+as this is a subset of existing functionality computations-wise.
+As for other Normalization primitives, the proposal is to not extend them at this point.
+
+For the work-in-progress POC, refer to [PR 3068](https://github.com/uxlfoundation/oneDNN/pull/3068).
+For more details about the proposal, see the following sections.
+
+Note, that previously, if users wanted to implement RMSNorm, they had to rely on the `dnnl_use_global_stats` flag,
+manually set the mean to zero, and compute the `RMS^2` statistics themselves.
+This approach is not convenient and adds extra complexity (computations, memory overhead) for the users.
+
+### Extension to the Normalization Flags
+
+This support of RMSNorm be achieved by adding a new flag to the existing `dnnl_normalization_flags_t`
+enum and `normalization_flags` enum in the C++ API:
+```cpp
+/// Flags for normalization primitives.
+enum class normalization_flags : unsigned {
+  <old flags...>
+
+  /// Use Root Mean Square (RMS) Layer Normalization.
+  /// In forward propagation, this means that the mean is set to zero, and RMS
+  /// is used instead of variance.
+  /// The RMS norm is provided as an output during forward propagation for
+  /// training.
+  /// In backward propagation, the library computes the derivative with respect
+  /// to the RMS norm, assuming that the mean is zero.
+  use_rms_norm = dnnl_use_rms_norm,
+};
+```
+
+```c
+/// Flags for normalization primitives.
+typedef enum {
+  <old flags...>
+
+  /// Use RMS normalization
+  ///
+  /// If specified:
+  ///  - on forward propagation use RMS normalization instead of true mean and
+  ///    variance
+  ///  - on backward propagation assume mean as zero
+  dnnl_use_rms_norm = 0x20U,
+} dnnl_normalization_flags_t;
+```
+
+The new flag `dnnl_use_rms_norm` will be used to indicate that the RMS
+normalization is used instead of the standard Layer Normalization.
+
+**Naming Option 2**:
+Alternative namings for the flag could signify that the mean is **assumed** to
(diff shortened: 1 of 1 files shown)


mzhukova:
also inviting @pengzhao-intel and @gaurides to take a look

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+Since RMSNorm is a simplified version of LayerNorm, it works with `dnnl_use_scale`
+and `dnnl_use_shift` flags in the same manner.
+
+Additionally, like LayerNorm, it will not support `dnnl_fuse_norm_relu` or `dnnl_fuse_norm_add_relu`.
+
+### Changes to Layer Normalization Primitive Inputs and Outputs
+
+Since the main purpose of RMSNorm is to avoid the overhead of mean computations,
+it seems reasonable to omit the mean as input/output and use `RMS^2` in place of the variance, which means:
```
dzarukin, line 125: Technically, it's `RMS^2 - eps`. So only sum squared and divided by N is provided. Maybe put a remark everywhere you meant this instead if you want to keep a simple signature.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+An additional flag for LayerNorm to enable RMSNorm will be added to the `--flags` option.
+
+```
+`--flags=[|G|C|H|R]` -- layer normalization flags, default `none`; where
+            multiple simultaneous flags are supported.
+            `G` is dnnl_use_global_stats;
+            `C` is dnnl_use_scale;
+            `H` is dnnl_use_shift;
+            `R` is dnnl_use_rms_norm;
```
dzarukin, line 147: R stands for `Relu` in bnorm, we better avoid collision since it might be extended to it.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+            Refer to [layer normalization primitive](https://uxlfoundation.github.io/oneDNN/dev_guide_layer_normalization.html)
+            for details.
```
dzarukin, line 149: This can be safely dropped from RFC. The official doc won't be updated until the support is in.

rfcs/20251004-rms-norm/README.md
mzhukova: reworded in the latest commit to avoid confusion w.r.t eps

rfcs/20251004-rms-norm/README.md
mzhukova: sure, double checked flag and seems like `M` could be a candidate (as in ~second letter), update now

rfcs/20251004-rms-norm/README.md
mzhukova: this was just copy-pasted from benchdnn docs on the flag, but I've removed the particular reference to avoid confusion

mzhukova:
confirmed with @pengzhao-intel and @gaurides, that this options is preferred --
> (Preferred) Omit mean in the output:
Pros: Cleaner design and avoids unnecessary memory requirement.
Cons: Requires users to adjust their code to handle the absence of the mean for RMSNorm comparing to LayerNorm.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+Alternative namings for the flag could signify that the mean is **assumed** to
+be zero, for instance `dnnl_{use, force}_zero_mean`. But this seems misleading
+as the mean is not explicitly set to zero, but simply ignored. Also it doesn't
+explciitly indicate what is happening with the variance.
+
+#### Compatibility with Existing Flags
+
+The `dnnl_use_global_stats` flag technically could be applied to RMSNorm, but
+it is not particularly meaningful since RMSNorm never requires the mean computations.
+Therefore the proposal is to make it incompatible with the `dnnl_use_global_stats` flag.
```
mgouicem, line 116: I don't know if we can do that, as there should still be the case where variance is precomputed and rovided by user. I guess this calls for splitting `dnnl_use_global_stats`. User would then have:
- `dnnl_use_mean | dnnl_use_variance` to cover existing `dnnl_use_global_stats` for LayerNorm
- `dnnl_no_mean | dnnl_use_variance` to cover RMS norm with precomputed variance.
- `dnnl_no_mean` to cover RMS norm with embedded variance computation (with mean = 0).

mgouicem (commented):
Regarding open questions:
- I would vote for `dnnl_use_zero_mean` or `dnnl_no_mean`, as this is closer to how we split `dnnl_use_scale` and `dnnl_use_shift`.
- I would vote for omitting mean, but it might be worth checking with PyTorch team what the expectation is.

rfcs/20251004-rms-norm/README.md
dzarukin: > dnnl_no_mean | dnnl_use_variance to cover RMS norm with precomputed variance.

If variance is pre-computed, user can simply pass mean as zeros if they want to emulate RMS. It sounds like a really minor memory optimization with an API change which can be avoided.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+The purpose of computing the mean and variance is to re-center and re-scale the
+input data, which helps improve stability and convergence.
+
+RMSNorm, on the other hand, skips the re-centering step and normalizes using only
+the root mean square statistic (`RMS`):
+```math
+RMS(t, n) = \sqrt{\frac{1}{C} \sum_{c} src(t, n, c)^2 + \epsilon}
+```
+The `epsilon` is added to avoid further dividing by zero.
```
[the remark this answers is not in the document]
mzhukova: I could change to that wording no problem, but I guess my point here was that since `RMS()` is in the denominator, dividing by zero could be an issue.

rfcs/20251004-rms-norm/README.md
mgouicem: Well from user perspective, it means:
- either allocate an extra buffer on device and enqueue zeroing of this memory
- or pre-allocate a zero memory buffer and manage it's lifetime.

Maybe @pengzhao-intel can chime in.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,177 @@
...
+equal to `RMS^2`, and moreover if the input data is such that it has a mean of zero,
+RMSNorm and LayerNorm produce the same results.
+
+For more details, including comparisons with LayerNorm and examples of benefits,
+refer to the article by [Biao Zhang and Rico Sennrich](https://arxiv.org/abs/1910.07467).
+This article provides more details regarding the method and showcases some experiments
+using frameworks like TensorFlow and PyTorch.
+
+Additionally, consider TensorFlow
+[LayerNorm with RMS parameter](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LayerNormalization)
```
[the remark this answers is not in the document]
mzhukova: Thanks for details @gaurides, I'll update the section accordingly. From a brief look into what you shared, the rest of the RFC / proposal would stay the same.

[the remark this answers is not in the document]
gaurides: Hi @mzhukova, the rms_scaling flag in LayerNormalization api in TF/keras is different than RMSNormalization. It makes me wonder that it will be confusing to have "dnnl_use_rms_norm" flag to indicate RMSNormalization op.
How to distinguish these two scale LN + rms_scaling vs RMSNormalization ?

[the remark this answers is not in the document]
mzhukova: hi @gaurides, a small update -- I think there is an issue with Layer Normalization + rms scaling (either doc or implementation), so I've created an issue https://github.com/keras-team/keras/issues/21234 to clear the confusion.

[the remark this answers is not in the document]
gaurides: thanks @mzhukova for checking. 
I too observed the same that the mean is indeed used in the variance calculation in case of LN + rms_scaling. I think when they say center and scale are ignored, i believe it refers to the inputs to [LN api] (https://keras.io/api/layers/normalization_layers/layer_normalization/)
Yes, their documentation can be clear to specify the differences.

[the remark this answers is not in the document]
mzhukova: > I think when they say center and scale are ignored, i believe it refers to the inputs to [LN api](https://keras.io/api/layers/normalization_layers/layer_normalization/)

I'm more concerned with this part _"and the inputs are scaled by gamma and the **inverse square root of the square of all inputs**. This is an approximate and faster approach that **avoids ever computing the mean of the input**"_, which essentially defines RMS Norm. But in any case, let's wait for clarification from their side.

rfcs/20251004-rms-norm/README.md
mzhukova: > `dnnl_no_mean | dnnl_use_variance` to cover RMS norm with precomputed variance.

@mgouicem Could we instead add support for `dnnl_use_rms_norm | dnnl_use_global_stats` and require a single input (== RMS norm) in the future if needed?

> `dnnl_no_mean | dnnl_use_variance` to cover RMS norm with precomputed variance

This feels a bit confusing to me because it might suggest that we're skipping the mean (so no re-centering), but scaling would still use the true variance instead of RMS norm. It seems to me that it might be clearer to explicitly say that RMS normalization is performed with user-provided statistics via `dnnl_use_rms_norm | dnnl_use_global_stats`, but it is the same argument that I have in the pros/cons list for the naming options.

mgouicem: > I find having dnnl_use_variance to be a bit confusing in the context of RMS normalization, as we do not require variance, but rather the RMS norm, right..?

Agree, strictly speaking, if we set mean to 0 it is no more variance. I mentioned variance nomenclature just because it allows to just piggy-back on existing APIs. I am fine with clearly separating those. We would need a new `DNNL_ARG_RMS_NORM` argument kind as well if we want that separation.

> I believe {dnnl_use_rms_norm, dnnl_use_zero_mean} | dnnl_use_global_stats does indicate that the user needs to provide pre-computed statistics only for the RMS norm.

Sure that works. Though I would recommend not to use `dnnl_use_*` nomenclature, as it currently tags things that user need to provide, but here it is more used as an algorithm kind (maybe `dnnl_rms`?).

mzhukova: @mgouicem I've updated to use `dnnl_rms_norm`, please let me know if that would work :) Also, would you prefer this over `dnnl_{use_zero, no}_mean`?
I don't want to complicate API, I think re-using `DNNL_ARG_VARIANCE` should be fine, but let me know if you think otherwise.

rfcs/20251004-rms-norm/README.md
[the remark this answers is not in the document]
mzhukova: @gaurides I think we should proceed with the current proposal. keras now has a separate layer for RMS normalization, which this RFC covers. I don't believe we need to support the option of no re-centering + re-scaling using true variance vs RMS statistics (as seen in keras with Layer Normalization + `rms_scaling=True`), since it removes the benefit of reduced computations.

I'll include the newly created keras issue in the RFC description for future reference, but I suggest we move forward with RFC and implementation. Let me know what you think!

mzhukova:
hi @dzarukin, @mgouicem and @gaurides, thank you for your feedback!
I believe I've addressed all the current comments, so could you please re-review/close opened threads and let me know if anything else is needed here or approve.

rfcs/20251004-rms-norm/README.md
mgouicem: Thanks @mzhukova , in the current state, `dnnl_rms_norm` is prefered, thanks.

rfcs/20251004-rms-norm/README.md
```diff
@@ -0,0 +1,161 @@
...
+Alternative namings for the flag could signify that the mean is **assumed** to
+be zero, for instance `dnnl_{use_zero, no}_mean`. But this seems misleading
+as the mean is not explicitly set to zero, but simply ignored. Also it doesn't
+explciitly indicate what is happening with the variance.
+
+#### Compatibility with Existing Flags
+
+The combination of `dnnl_use_global_stats` and `dnnl_rms_norm` would be supported,
+but users must only provide pre-computed RMS norm statistics
+(instead of both mean and variance as required for Layer normalization).
```
mgouicem, line 115: Could you precise that user need to pass pre-computed RMS norm with DNNL_ARG_VARIANCE.

mgouicem (approved):

rfcs/20251004-rms-norm/README.md
mzhukova: done in [92e9f5d](https://github.com/uxlfoundation/oneDNN/pull/3147/commits/92e9f5d29441fb429903c45d479499c2b2ce1cbf)

rfcs/20251004-rms-norm/README.md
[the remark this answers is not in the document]
gaurides: @mzhukova sounds good. At this point, I am not aware of a model using LayerNorm + rms_scaling. Please document that this case is not supported. We can revisit that portion if/when a model needs it.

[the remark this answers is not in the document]
gaurides: @mzhukova sounds good. At this point, I am not aware of a model using LayerNorm + rms_scaling. Please document that this case is not supported. We can revisit that portion if/when a model needs it.

gaurides (approved):
LGTM

rfcs/20251004-rms-norm/README.md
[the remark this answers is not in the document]
mzhukova: added in [b739951](https://github.com/uxlfoundation/oneDNN/pull/3147/commits/b7399514984c153029d7da6cba015dc2126cc1b5)

dzarukin (approved):

PR #3151, opened 2025-04-23
benchdnn: utils: cfg: consolidate adjustment logic under one umbrella

dzarukin:
* Update safe_n_acc to return fixed number of elements for very low precision dt.
* Move all logic spread across configs under a base config.
* Utilize a single call across cfgs to re-use the filling logic.

A better version of https://github.com/uxlfoundation/oneDNN/pull/3101 (the commit was reverted on main) which was exposed to the issue after matmul driver coverage was updated between the validation and promotion. Besides, the issue was caught on Nightly coverage which was not validated.

And a fix for cold-cache variable overflow.

### DIFF
--- a/tests/benchdnn/utils/cfg.hpp
+++ b/tests/benchdnn/utils/cfg.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2023 Intel Corporation
+* Copyright 2022-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


dzarukin:
make test
set test_scope=NIGHTLY

dzarukin:
make test

tests/benchdnn/utils/cfg.hpp
```diff
@@ -195,6 +225,60 @@ struct base_cfg_t {
     void set_range_max(data_kind_t dk, int new_value) {
         cfg_entry_.at(dk).set_range_max(new_value);
     }
+
+    void get_verbose(
```
avmanerikar, line 229: (nit): perhaps rename `get_verbose` to be more descriptive, e.g., `verbose_print_fill_cfg`?

avmanerikar (approved):

tests/benchdnn/utils/cfg.hpp
dzarukin: Updated, thanks.

PR #3166, opened 2025-04-25
graph: utils: pm: support commutative select

ShanSimu:
# Description

The original setting for "commutative" was limited to operators with only two inputs. However, the select operator has three inputs, and only two of them are commutative.
A new member variable, `std::set<size_t> commutative_inputs`, has been added to `op_schema` to store the indices of commutative inputs. The function `set_commutative_inputs(std::set<size_t> inputs = {0, 1})` uses default parameters to maintain support for the original commutative operators.
In `nested_matcher.cpp`, the functions `has_commutative_inputs`, `match_commutative_inputs`, and `match_node_inputs` have been updated to include checks for whether a port belongs to the commutative inputs.

This PR is for fix MFDNN-13383. Another method is also implemented [here](https://github.com/uxlfoundation/oneDNN/tree/shaojiec/select_input_extend), which is achieved by swapping the mapping method of ports inside and outside the subgraph.


src/graph/interface/op_schema.cpp
```diff
@@ -102,15 +102,24 @@ op_schema_t &op_schema_t::set_output(
     return *this;
 }
 
-op_schema_t &op_schema_t::set_commutative_inputs() {
-    assertm(num_inputs_.size() == 1 && *(num_inputs_.begin()) == 2,
-            "commutative inputs can only be enabled for ops with two inputs");
+op_schema_t &op_schema_t::set_commutative_inputs(std::set<size_t> inputs) {
```
ElaineBao, line 105:
```suggestion
op_schema_t &op_schema_t::set_commutative_inputs(const std::set<size_t> &inputs) {
```

src/graph/interface/op_schema.cpp
```diff
@@ -102,15 +102,24 @@ op_schema_t &op_schema_t::set_output(
...
+                "each input value must be less than input numbers");
+    }
     commutative_inputs_enabled_ = true;
+    commutative_inputs = inputs;
     return *this;
 }
 
-bool op_schema_t::get_commutative_inputs() const {
-    return commutative_inputs_enabled_;
+bool op_schema_t::get_commutative_inputs(std::set<size_t> inputs) const {
```
ElaineBao, line 118:
```suggestion
bool op_schema_t::get_commutative_inputs(const std::set<size_t> &inputs) const {
```

src/graph/utils/pm/nested_matcher.cpp
```diff
@@ -70,10 +70,10 @@ namespace pm {
 
 namespace {
 // check if an op's inputs are commutative
-bool has_commutative_inputs(op_t *op) {
+bool has_commutative_inputs(op_t *op, std::set<size_t> inputs) {
```
ElaineBao, line 73:
```suggestion
bool has_commutative_inputs(op_t *op, const std::set<size_t> &inputs) {
```

src/graph/utils/pm/nested_matcher.cpp
```diff
@@ -252,7 +254,8 @@ bool match_node_inputs(const binding_t &b, match_context_t *ctx,
     if (node_inputs_matcher.get_node()->get_inputs().size()
             == VARIADIC_INPUT_NUM) {
         matching_status = node_inputs_matcher.match_variadic_inputs();
-    } else if (!has_commutative_inputs(node_inputs_matcher.get_op())) {
+    } else if (!has_commutative_inputs(node_inputs_matcher.get_op(),
+                       {b.bind_op_port, b.bind_port})) {
```
ElaineBao, line 258: I'm wondering, for select, whose first input is non-commutative, and second & third inputs are commutative, how can we make sure all the inputs are correctly visited and matched?

src/graph/utils/pm/nested_matcher.cpp
```diff
@@ -200,12 +200,14 @@ bool node_inputs_matcher_t::match_commutative_inputs() {
 
     for (size_t node_input_offset = 0; node_input_offset < node_inputs_.size();
             ++node_input_offset) {
+        if (!has_commutative_inputs(get_op(), {node_input_offset})) continue;
```
ElaineBao, line 203: maybe it's better to change the name of this function to is_commutative_input(), as it's verifying the single input port now.

src/graph/utils/pm/nested_matcher.cpp
ShanSimu: Take https://github.com/intel-sandbox/onednn-graph-perf-analysis/blob/master/json/mha/bf16/mha-pt-starcoder-inf-bf16-bs1.json as an example. When `bind_node` is `select`, `bind_port` is 2 and `bind_op_port` is 1 which conform to `commutative inputs = {1, 2}`.

ElaineBao: I understand that. I mean how is the 0-th input matched? As it's non-commutative.

src/graph/interface/op_schema.cpp
ShanSimu: Updated

src/graph/utils/pm/nested_matcher.cpp
ShanSimu: Updated

src/graph/utils/pm/nested_matcher.cpp
ShanSimu: Updated

ElaineBao:
please format the commit message with correct component, graph: interface: ..., graph: utils: pm: ...

src/graph/utils/pm/nested_matcher.cpp
ShanSimu: Do you mean the first or the second input will swap with the first one? I doubt whether that's a valid graph, since the data type of the 0-th input is boolean after all. As for other cases, like when all three input data types are the same but only two are commutative, I don't think we have such a requirement.

src/graph/interface/op_schema.hpp
```diff
@@ -301,6 +302,7 @@ class op_schema_t {
     shape_infer_fn tensor_inference_function_ = nullptr;
     std::vector<op_def_constraint_fn> op_def_constraint_functions_;
     bool commutative_inputs_enabled_ = false;
+    std::set<size_t> commutative_inputs;
```
wzt1997, line 305: better use `commutative_inputs_`

ShanSimu: Updated

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
```diff
@@ -1108,6 +1108,46 @@ TEST(test_utils_pattern_matcher, OptionalWithLargerPort) {
     ASSERT_EQ(fusion_ops.size(), 3U);
 }
 
+TEST(test_utils_pattern_matcher, CommutativeSelect) {
+    auto graphp = std::make_shared<pb_graph_t>();
+    // Pattern that captures
+    // MatMul  Add
+    //   \     /
+    //    Select
+    auto pmatmul = graphp->append_op(MatMul);
```
ElaineBao, line 1117: Please add another input for select's 0-th input (e.g. GreaterEqual), we want to know if the three inputs (both commutative and non-commutative) can all be matched

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
```diff
@@ -1108,6 +1108,46 @@ TEST(test_utils_pattern_matcher, OptionalWithLargerPort) {
...
+    select.add_input(lt_vec[2]);
+    select.add_output(lt_vec[7]);
+
+    ASSERT_EQ(agraph.add_op(&matmul), status::success);
+    ASSERT_EQ(agraph.add_op(&add), status::success);
+    ASSERT_EQ(agraph.add_op(&select), status::success);
+    agraph.finalize();
+
+    std::vector<op_t *> fusion_ops;
+    EXPECT_TRUE(match_pattern(agraph.get_ops()[0].get(), graphp, fusion_ops));
```
ElaineBao, line 1147: And here we need to match from the select op, to verify our assumption.

ShanSimu: I don't understand "match from the select op". Is there anything to be modified?

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
ShanSimu: 0-th input for select has been added.

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
ShanSimu: It seems impossible to declare `select` first and `edges` at last because edge related functions of `pb_graph_t` are protected.

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
ShanSimu: The logic of `match_commutative_inputs()` has been modified: when `node_input_offset` and `op_input_offset` are not equal, it will check whether both belong to `commutative_inputs_`. If they do not, `match_input_by_offset()` will not be called.

src/graph/utils/pm/nested_matcher.cpp
```diff
@@ -252,7 +255,7 @@ bool match_node_inputs(const binding_t &b, match_context_t *ctx,
     if (node_inputs_matcher.get_node()->get_inputs().size()
             == VARIADIC_INPUT_NUM) {
         matching_status = node_inputs_matcher.match_variadic_inputs();
-    } else if (!has_commutative_inputs(node_inputs_matcher.get_op())) {
+    } else if (!is_commutative_input(node_inputs_matcher.get_op())) {
```
ElaineBao, line 258: what will happen if this function is called without a set of input ports?

ShanSimu: It will return directly `commutative_inputs_enabled_`

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
```diff
@@ -1108,6 +1108,53 @@ TEST(test_utils_pattern_matcher, OptionalWithLargerPort) {
     ASSERT_EQ(fusion_ops.size(), 3U);
 }
 
+TEST(test_utils_pattern_matcher, CommutativeSelect) {
```
ElaineBao, line 1111: The commit message shoud be gtest: graph: ...

src/graph/interface/op_schema.hpp
```diff
@@ -167,11 +167,12 @@ class op_schema_t {
...
 
-    /*! @brief Enable commutative inputs */
-    op_schema_t &set_commutative_inputs();
+    /*! @brief Enable commutative inputs for specific inputs*/
+    op_schema_t &set_commutative_inputs(
+            const std::set<size_t> &inputs = {0, 1});
 
     /*! @brief Get whether the commutative inputs option is enabled or not */
-    bool get_commutative_inputs() const;
+    bool get_commutative_inputs(const std::set<size_t> &inputs = {}) const;
```
ElaineBao, line 175: Add a brief description for this function, check if the inputs set are all commutative inputs, if inputs is an empty set, check if the op has commutative inputs.

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
ShanSimu: Updated. Just typo...

src/graph/interface/op_schema.hpp
ShanSimu: Updated

ElaineBao (approved):

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/interface/op_schema.cpp
```diff
@@ -102,15 +102,23 @@ op_schema_t &op_schema_t::set_output(
...
-            "commutative inputs can only be enabled for ops with two inputs");
+op_schema_t &op_schema_t::set_commutative_inputs(
+        const std::set<size_t> &inputs) {
+    assertm(num_inputs_.size() == 1 && *(num_inputs_.begin()) > 1,
+            "The number of commutative inputs is at least two");
+    assertm(std::all_of(inputs.begin(), inputs.end(),
+                    [&](size_t idx) { return idx < *(num_inputs_.begin()); }),
+            "each input value must be less than input numbers");
     commutative_inputs_enabled_ = true;
+    commutative_inputs_ = inputs;
```
TaoLv, line 113: Why do we need both `commutative_inputs_enabled_` and `commutative_inputs_`?

TaoLv: with this, we plan to support only 1 pair of commutative inputs, correct?

src/graph/interface/op_schema.cpp
```diff
@@ -102,15 +102,23 @@ op_schema_t &op_schema_t::set_output(
...
     return *this;
 }
 
-bool op_schema_t::get_commutative_inputs() const {
-    return commutative_inputs_enabled_;
+bool op_schema_t::get_commutative_inputs(const std::set<size_t> &inputs) const {
+
+    return commutative_inputs_enabled_
+            && std::includes(commutative_inputs_.begin(),
+                    commutative_inputs_.end(), inputs.begin(), inputs.end());
```
TaoLv, line 121: Not sure how this function is used. But from it's implementation, i think the function name can be improved:

```
bool op_schema_t::is_commutative_inputs(size_t input1, size_t input2) {
      // make sure both {input1, input2} and {input2, input1} are checked.
      const bool ret = .....;
      return ret;
}
```

src/graph/interface/op_schema.hpp
```diff
@@ -167,11 +167,13 @@ class op_schema_t {
     op_schema_t &set_output(size_t out_offset, std::string &&out_name,
             std::string &&dtype_string = "any");
 
-    /*! @brief Enable commutative inputs */
-    op_schema_t &set_commutative_inputs();
+    /*! @brief Enable commutative inputs for specific inputs*/
+    op_schema_t &set_commutative_inputs(
+            const std::set<size_t> &inputs = {0, 1});
```
TaoLv, line 172: How many places is this function called? If possible, let's remove the default argument and add it explicitly at where it's needed.

src/graph/interface/op_schema.hpp
TaoLv: The default argument here is confusing.

src/graph/interface/op_schema.cpp
ShanSimu: > Why do we need both `commutative_inputs_enabled_` and `commutative_inputs_`?

The commutative_inputs_enabled_ flag can indeed be removed—checking whether commutative_inputs_ is empty is sufficient to determine if commutative behavior is enabled.

> with this, we plan to support only 1 pair of commutative inputs, correct?

We support a set of commutative inputs, not just a single pair.

src/graph/interface/op_schema.cpp
ShanSimu: In fact, the inputs may exceed two, and while we currently don't have such an op, it might be beneficial to retain this extensibility for future use.

src/graph/interface/op_schema.cpp
TaoLv: what does `a set of commutative inputs` mean? More than two inputs can be commutative? Do you have a test case for that?
I was thinking about an extreme case that input 0 and 1 are commutative and input 2 and 3 are commutative.

ShanSimu: Yes, `a set of commutative inputs` means more than two inputs can be commutative. We currently don't have such an op or test case.
`input 0 and 1 are commutative and input 2 and 3 are commutative` is another corner case I hadn't considered, and currently there are no corresponding test cases for it.

src/graph/interface/op_schema.cpp
ShanSimu: In actual usage, one way this function is called is with no parameters, solely to check whether the current op supports commutativity. If we change the parameters to `size_t input1, size_t input2`, it would make this zero-argument usage less straightforward or even impossible without overloading or adding a separate method.

src/graph/interface/op_schema.hpp
ShanSimu: There are only four usages relying on the default argument. Removing it would indeed make the function interface clearer and more explicit.

src/graph/interface/op_schema.cpp
TaoLv: My point is not the arguments of the function. It's more about the function name. You can still use `std::set` to pass the inputs.

- is_commutative_inputs(inputs): check if the given inputs are commutative, return true or false accordingly.
- get_commutative_inputs(): return the set of commutative inputs.

src/graph/interface/op_schema.cpp
TaoLv: In this case, I would suggest to stick on what's supported and tested for now. That's only 1 pair of commutative inputs is supported. Update the code and comment to reflect this design and restriction.
For the cases with more than 2 commutative inputs, or two pairs of commutative inputs, we can add support once the request pops up.

ShanSimu: `commutative_inputs_enabled_` has been removed and replaced with `!commutative_inputs_.empty()`.

`commutative_inputs_` has been changed to `std::array<size_t, 2>`, strictly limiting it to two inputs. This change should be documented in the comments to clarify the restriction.

src/graph/interface/op_schema.cpp
ShanSimu: The function name has been changed to `is_commutative_inputs`

src/graph/interface/op_schema.hpp
ShanSimu: Updated

src/graph/interface/op_schema.hpp
ShanSimu: The default argument here is to maintain compatibility with zero-argument usage(in [nested_matcher.cpp#L73](https://github.com/uxlfoundation/oneDNN/blob/0fc69752f83811c008847ee1ee73076a70244dad/src/graph/utils/pm/nested_matcher.cpp#L73)), where we simply check if the op has commutative inputs. It might make sense to split this into two separate functions:

- is_commutative_inputs(inputs)
- is_commutative_op()

ShanSimu: Updated

src/graph/interface/op_schema.cpp
ShanSimu: Since the zero-argument usage is separated, the arguments have been changed to `const size_t input0, const size_t input1`

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/interface/op_schema.cpp
```diff
@@ -102,15 +102,30 @@ op_schema_t &op_schema_t::set_output(
...
+    return is_commutative_op()
+            && ((commutative_inputs_[0] == input0
+                        && commutative_inputs_[1] == input1)
+                    || (commutative_inputs_[0] == input1
+                            && commutative_inputs_[1] == input0));
+}
+
+bool op_schema_t::is_commutative_op() const {
+    return commutative_inputs_[0] != SIZE_MAX
+            && commutative_inputs_[1] != SIZE_MAX;
```
TaoLv, line 128: Is it possible to just make it empty to represent that there are no commutative ops?

ShanSimu: `std::array<T, N>` is a fixed-size container, so `empty()` always returns false. I also tried using `std::pair`, but it likewise always contains two values—and it doesn't even have an `empty()` function.

TaoLv: Thanks for the explain. I would suggest to not use `SIZE_MAX` as magical numbers in this case.
If we don't consider `std::optional` which requires C++17, two options I can imagine:
1. Still use std::set as your previous solution, but check the size of the set when needed. It should be either empty or 2 elements.
2. Use a vector/set/pair/etc of std::pair. Check empty if it's not set. Once set, access the elements in the pair.

TaoLv: Casting @uxlfoundation/onednn-graph  to see if other suggestions here.

TaoLv (approved):
Please rebase and test the PR.

ShanSimu:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3168, opened 2025-04-25
doc: update environment.yml file

mzhukova:
- Added doxygen and graphviz into dependencies as currently both are required to build documentation (will re-evaluate the need for dot/graphviz later).
- Disallowed default channel.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,5 +6,7 @@ dependencies:
   - python=3.9.5
   - pip=21.1.2
   - doxyrest=2.1.2
+  - doxygen=1.8.14
+  - graphviz=2.40.1
   - sphinx=4.0.2
   - sphinx-book-theme=0.0.41
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -1,7 +1,7 @@
 name: onednn-doc
 channels:
   - conda-forge
-  - defaults
+  - nodefaults
 dependencies:
   - python=3.9.5
   - pip=21.1.2


mzhukova:
make docs

mzhukova:
make test
enable run_documentation

mzhukova:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

doc/environment.yml
```diff
@@ -1,10 +1,11 @@
 name: onednn-doc
 channels:
   - conda-forge
-  - defaults
+  - nodefaults
 dependencies:
   - python=3.9.5
   - pip=21.1.2
   - doxyrest=2.1.2
+  - doxygen=1.8.14
```
spalicki, line 9: Is dot/graphviz also installed with doxygen as a dependency?

mzhukova: yep, dot is a part of doxygen installation

vpirogov (approved):

densamoilov (approved):

doc/environment.yml
spalicki: It seems that it is not a direct dependency. To have dot come from conda we must add `graphviz` package as well. I do not know which version works well with 1.8.14 Doxygen.

mzhukova: I will double check on that and extend if needed

mzhukova: updated now

spalicki (approved):

PR #2742, opened 2025-02-24
{include, benchdnn}: graph: fix clang-tidy warnings in headers

TaoLv:
MFDNN-13013: fix clang-tidy warning in headers.

Corresponding violations are explained by the commit messages.

PR #3186, opened 2025-04-29
Fixup clang-tidy warnings when using clang-14

Simonsays095:
Building with clang-tidy results in a few warnings related to functions declared in headers. This PR addresses these (and two other warnings) in the library.

- `misc-definitions-in-headers`  fixed in CPU IP, CPU brgemm (aarch64 and x64), and GPU SDPA. See https://github.com/llvm/llvm-project/issues/61471 for why this error is showing up, despite being enclosed in anonymous namespaces.
- `clang-diagnostic-ambiguous-reversed-operator` fixed in CPU brgemm (this is a C++20 warning)
- `cppcoreguidelines-prefer-member-initializer` fixed in graph, which was suppressing the warning incorrectly.


tests/benchdnn/graph/graph.hpp
```diff
@@ -42,8 +42,8 @@ struct settings_t : public base_settings_t {
 
     // ctor to save certain fields from resetting
     settings_t(const char *perf_template) : settings_t() {
-        this->perf_template
-                = perf_template; // NOLINT(cppcoreguidelines-prefer-member-initializer)
+        // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer)
+        this->perf_template = perf_template;
     }
```
atkassen, line 47: Couldn't this be
```suggestion
    settings_t(const char *perf_template) : perf_template(perf_template) {}
```
and we can get rid of the suppression altogether?

vpirogov:
@Simonsays095, full clang-tidy build in CI [is clean](http://intel-ci.intel.com/f024bb1d-6619-f123-a10a-a4bf010d0e2d). Is this particular clang-tidy version or specific build options?

Simonsays095:
> @Simonsays095, full clang-tidy build in CI [is clean](http://intel-ci.intel.com/f024bb1d-6619-f123-a10a-a4bf010d0e2d). Is this particular clang-tidy version or specific build options?

This is using clang-14, and according to https://reviews.llvm.org/D147379, this warning was suppressed at some point (I'm not sure, but my guess is somewhere in clang-15) in favor of the `google-build-namespaces` warning. It looks like the CI you link is using clang-19, where this has been transferred to the other warning flag.

tests/benchdnn/graph/graph.hpp
Simonsays095: It's possible that this will have some unintended effect on value initialization, but I'll give it a shot and run graph tests to see.

Simonsays095:
make test
enable build_mode_cpp20
disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_ip
enable benchdnn_graph

tests/benchdnn/graph/graph.hpp
TaoLv: Please see the discussion here: https://github.com/uxlfoundation/oneDNN/pull/2742#discussion_r1968133318

Simonsays095: @TaoLv What do we gain from delegating to a default constructor? I _think_ that because all member variables have default values, they should pick those up (identically to what the default constructor would do) before overriding from `perf_template` from the member initializer list.

atkassen: All of the members of `settings_t` has a default initializer/constructor, so there is no need to explicitly call `settings_t()` here -- default initialization happens anyway. But delegating to another constructor forbids you from adding the `perf_template` initializer, as noted in the linked discussion.

tests/benchdnn/graph/graph.hpp
```diff
@@ -41,10 +41,8 @@ struct settings_t : public base_settings_t {
     settings_t() = default;
 
     // ctor to save certain fields from resetting
-    settings_t(const char *perf_template) : settings_t() {
-        this->perf_template
-                = perf_template; // NOLINT(cppcoreguidelines-prefer-member-initializer)
-    }
+    settings_t(const char *perf_template) : perf_template(perf_template) {}
```
dzarukin, line 44: This breaks the way it works. The ctor should create a brand new object and THEN update the perf_template. This will just assign template and won't do reset.

dzarukin (commented):
Why checking clang-14 at all?

tests/benchdnn/graph/graph.hpp
Simonsays095: This constructor calls the base class constructor (`base_settings_t`), followed by setting all `settings_t` variables to their defaults except `perf_template`, which gets set from the argument instead of the default. This should be equivalent to the default constructor followed by a `perf_template` override.

Simonsays095:
make test
enable build_mode_cpp20
disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_ip
enable benchdnn_graph

tests/benchdnn/graph/graph.hpp
atkassen: [Manual CI](https://intel-ci.intel.com/f025faa3-f3a2-f154-a9a2-a4bf010d0e2d) seems fine.

dzarukin: Ci wouldn't trigger a problem since it's perf-template in question.
But I checked and it's fine for graph, because they don't rely on `base_settings_t` for perf_template. And even if perf-template would be supported (while it's not), seems to work fine.
Maybe I got some phantom pains which are not there any more. But for rest classes to get rid of the NOLINT, it will require base_settings to introduce a new ctor and use that one. Haven't checked if that works as expected.

Simonsays095:
make test
enable build_mode_cpp20
disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_ip
enable benchdnn_graph

Simonsays095:
[Actual CI](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f0278084-e321-f13d-9f7b-a4bf010d0e2d?objectId=job-f0278084-e321-f13d-9f7b-a4bf010d0e2d&filterName0=jobsPageSearch&filterDepth=1) (github links to an aborted run)

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -312,7 +312,7 @@ struct jit_brgemm_amx_uker_base_t : public jit_base_brgemm_kernel_t {
         bd_iteration_t *similar {nullptr};
         Label lstart;
 
-        bool operator==(const dim_iteration_t &_rhs) const override {
+        bool operator==(const bd_iteration_t &_rhs) const {
             // `downcast` will catch a type mismatch in debug mode.
             // Note: it supports only a pointer type so far.
             const bd_iteration_t &rhs
```
atkassen, line 318: If this is the intention, then the downcast below is unnecessary.

Simonsays095: Thanks, Andy. This is the intention (unless the CPU team points out an issue with this approach), so I've simplified the function as you suggest by removing the downcast.

atkassen:
> [...] according to https://reviews.llvm.org/D147379, this warning was suppressed at some point (I'm not sure, but my guess is somewhere in clang-15) in favor of the `google-build-namespaces` warning. It looks like the CI you link is using clang-19, where this has been transferred to the other warning flag.

Does it make sense to enable this flag, then? It seems helpful for avoiding ODR violations. There are a handful of other locations where clang-tidy will complain, however.

```
src/cpu/reorder/simple_reorder.hpp:118
src/cpu/reorder/simple_sparse_reorder.hpp:66
src/cpu/rnn/ref_rnn.hpp:42
src/cpu/x64/cpu_isa_traits.hpp:378
src/cpu/x64/jit_generator.hpp:64
src/gpu/gpu_batch_normalization_pd.hpp:31
src/gpu/gpu_inner_product_pd.hpp:31
src/graph/backend/dnnl/patterns/data_type_check_pass.hpp:35
src/graph/interface/partition_hashing.hpp:45
```
(maybe more)

src/cpu/cpu_inner_product_pd.hpp
```diff
@@ -76,7 +76,7 @@ inline bool dense_gemm_consitency_check(const memory_desc_wrapper &src_d,
             && src_d.is_dense(true) && dst_d.is_dense() && wei_d.is_dense(true);
 }
 
-void transpose_md(memory_desc_t &md) {
+inline void transpose_md(memory_desc_t &md) {
```
atkassen, line 79: What is different here that clang-tidy doesn't complain? It's an unnamed namespace in a header file with inline functions, just like src/cpu/aarch64/matmul/brgemm_matmul.hpp.

Simonsays095:
> > [...] according to https://reviews.llvm.org/D147379, this warning was suppressed at some point (I'm not sure, but my guess is somewhere in clang-15) in favor of the `google-build-namespaces` warning. It looks like the CI you link is using clang-19, where this has been transferred to the other warning flag.
> 
> Does it make sense to enable this flag, then? It seems helpful for avoiding ODR violations. There are a handful of other locations where clang-tidy will complain, however.
> 
> ```
> src/cpu/reorder/simple_reorder.hpp:118
> src/cpu/reorder/simple_sparse_reorder.hpp:66
> src/cpu/rnn/ref_rnn.hpp:42
> src/cpu/x64/cpu_isa_traits.hpp:378
> src/cpu/x64/jit_generator.hpp:64
> src/gpu/gpu_batch_normalization_pd.hpp:31
> src/gpu/gpu_inner_product_pd.hpp:31
> src/graph/backend/dnnl/patterns/data_type_check_pass.hpp:35
> src/graph/interface/partition_hashing.hpp:45
> ```
> 
> (maybe more)

I'm in favor of adding it, but I'd like feedback from someone on the architecture team. In either case, I think the bulk of this work should be done in a more targeted PR.

tests/benchdnn/graph/graph.hpp
atkassen: I've opened #3198 to refactor the settings ctors.

src/cpu/cpu_inner_product_pd.hpp
Simonsays095: It's not complaining about the inline functions (those would actually be fine, otherwise) - it's complaining about the variables declared in the unnamed namespace (which are no longer in an unnamed namespace). I moved the inline functions to the cpp just to keep their internal linkage, which I believe was the intent behind the unnamed namespace.

dzarukin (approved):

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -312,18 +312,13 @@ struct jit_brgemm_amx_uker_base_t : public jit_base_brgemm_kernel_t {
...
 
-        bool operator==(const dim_iteration_t &_rhs) const override {
-            // `downcast` will catch a type mismatch in debug mode.
-            // Note: it supports only a pointer type so far.
-            const bd_iteration_t &rhs
-                    = *utils::downcast<const bd_iteration_t *>(&_rhs);
-            bool res = dim_iteration_t::operator==(rhs)
-                    && A_shift == rhs.A_shift && C_shift == rhs.C_shift
-                    && D_shift == rhs.D_shift && bd_mask == rhs.bd_mask
+        bool operator==(const bd_iteration_t &rhs) const {
```
densamoilov, line 315: Since this class is not polymorphic can you please make `dim_iteration_t`'s destructor non virtual?

densamoilov (approved):

atkassen (approved):

TaoLv (approved):

Simonsays095:
@Sqvid Can you please review for the aarch64 team?

Sqvid:
@Simonsays095 Yes taking a look today, sorry for the delay!

Sqvid (approved):
Done. Thank you for the patch and sorry again for the wait.

PR #3192, opened 2025-04-30
build: enable -Wundef diagnostics (fixes MFDNN-12446)

dzarukin:
https://jira.devtools.intel.com/browse/MFDNN-12446

It's mostly to cover OpenVINO needs to rebase and update codes if there're issues in there.

Certain choices are documented in commit messages.


dzarukin:
make test

dzarukin:
Hi @theComputeKid, could you please review this PR? Moreover, please take a look on the failed job for Mac. I've worked around one place (acl_thread) in the PR, but two more spots popped up. I need to come up with a way to work around the problem. In the core it will be the same thing to mark correspondent headers as system. The question is what the team prefers: multiple proxy headers for each affected spot so it's easier to revert them later or a single proxy header that would be included across the affected spots to have less new files. Please let me know your preference, I'll implement it.

src/cpu/aarch64/acl_thread_proxy.hpp
```diff
@@ -0,0 +1,31 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef CPU_AARCH64_ACL_THREAD_PROXY_HPP
+#define CPU_AARCH64_ACL_THREAD_PROXY_HPP
+
+// The purpose of the proxy header file is exactly to enable system_header
```
Sqvid, line 20: I am not sure about this pattern in general. As you can see from the build logs there are multiple ACL headers that have `Wundef` hits. Surely we would end up with a large number of `*_proxy.hpp` headers?

src/cpu/aarch64/acl_thread_proxy.hpp
```diff
@@ -0,0 +1,31 @@
+/*******************************************************************************
+* Copyright 2025 Arm Ltd. and affiliates
```
Sqvid, line 2: This should probably be your organisation rather than Arm.

Sqvid (changes requested):
Thank you for the patch. I have left some comments regarding the proxy-header pattern. Please share your thoughts.

src/cpu/aarch64/acl_thread_proxy.hpp
Sqvid: I will let @jondea and @theComputeKid weigh in but neither proliferating a large number of `_proxy.hpp` files nor a single large `acl_proxy.hpp` seem ideal.

src/gpu/intel/jit/config/ngen_config.hpp
```diff
@@ -41,7 +41,8 @@ using half = dnnl::impl::float16_t;
 #define NGEN_HALF_TYPE
 
 #if (!defined(NDEBUG) || defined(DNNL_DEV_MODE)) \
-        && (__cplusplus >= 202002L || _MSVC_LANG >= 202002L)
+        && ((defined(__cplusplus) && __cplusplus >= 202002L) \
```
rjoursler, line 44: Do we really need to check if `__cplusplus` is defined? According to https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros, it is a predefined macro and so it will always be defined.

rjoursler (approved):

src/cpu/aarch64/acl_thread_proxy.hpp
Sqvid: @dzarukin [I have raised a patch](https://review.mlplatform.org/c/ml/ComputeLibrary/+/14388) on the ComputeLibrary side to address this and will hopefully have an update for you soon. If it gets merged perhaps we only enable this warning for non-AArch64 systems for now and then turn it back on when we bump ComputeLibrary. Let me know your thoughts.

src/cpu/binary_injector_utils.cpp
```diff
@@ -34,7 +34,7 @@ std::vector<const void *> prepare_binary_args(const post_ops_t &post_ops,
             post_ops_binary_rhs_arg_vec.emplace_back(CTX_IN_MEM(const void *,
                     DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) | DNNL_ARG_SRC_1));
         }
-#if DNNL_X64
+#if defined(DNNL_X64)
```
densamoilov, line 37: I'm not sure if it's correct. All arch specific macros are defined but only one of them must be defined as `1` and the others must be defined as `0`.

See https://github.com/uxlfoundation/oneDNN/blob/96be1ead1444c328e1c0351bec88c18370639a3e/src/cpu/platform.hpp#L34 for more details.

dzarukin: Let me double-check when did it report this hit and for which backend.
But it definitely was highlighted. It just means that it's not compiled with platform.hpp at least.

src/gpu/intel/jit/config/ngen_config.hpp
dzarukin: I think I can drop it unless there's a case when this header is going to be used with plain C compiler, in which case this macro is not defined.

src/cpu/aarch64/acl_thread_proxy.hpp
dzarukin: Let me explore if it's an easy thing to do. Dropping it for Aarch64 is an option for me.

src/cpu/binary_injector_utils.cpp
densamoilov: I believe there is inconsistency in how we identify the architecture. We have two ways to do that:
1. Through CMake: https://github.com/uxlfoundation/oneDNN/blob/96be1ead1444c328e1c0351bec88c18370639a3e/CMakeLists.txt#L64-L76 and here https://github.com/uxlfoundation/oneDNN/blob/96be1ead1444c328e1c0351bec88c18370639a3e/cmake/platform.cmake#L32
2. Through the platform.hpp: https://github.com/uxlfoundation/oneDNN/blob/96be1ead1444c328e1c0351bec88c18370639a3e/src/cpu/platform.hpp#L40-L53

So, it seems that when it's identified via CMake then there is only macro defined but when it's identified via platform.hpp then all macros are defined.

So I think we should make it consistent before starting to rely on the identification logic.

densamoilov: FYI: MFDNN-5545

src/cpu/aarch64/acl_thread_proxy.hpp
dzarukin: Looks like platform.cmake has necessary support so I dropped the flag for AARCH64 and proxy headers altogether.

src/gpu/intel/jit/config/ngen_config.hpp
dzarukin: Removed.

src/cpu/binary_injector_utils.cpp
dzarukin: So, I'm definitely not covering build and aligning macro definition change as a part of this PR.

I dropped the check for X64 from this file, will monitor if Aarch64 backend fails anywhere. And if yes, will strip support from the implementation level at runtime which must have been done in the first place.

densamoilov: My point is that the code `#if defined(DNNL_ANYARCH)` is not correct based on our own specification of the `DNNL_ANYARCH` macros because they are always defined if the `platform.hpp` header is included. The behavior of `#if defined(DNNL_ANYARCH)` code depends on whether `platform.hpp` was included directly or indirectly.

dzarukin: DNNL_AARCH64_USE_ACL doesn't represent an architecture, so I'm not sure what your comment refers to. It is the way it was initially designed. If the intention to make it acting like architecture, then it's a different kind of project not related to Wundef issues.

densamoilov: Never mind, I confused DNNL_AARCH64_USE_ACL with DNNL_AARCH64.

src/cpu/aarch64/acl_thread_proxy.hpp
Sqvid: @dzarukin The fix has been merged in ACL. We should be able to enable when we next bump the version.

Sqvid (approved):
LGTM, thanks

TaoLv (approved):

src/gpu/gpu_impl_list.hpp
```diff
@@ -74,7 +74,7 @@ namespace gpu {
 #if defined(DNNL_WITH_SYCL) \
         && ((DNNL_GPU_VENDOR == DNNL_VENDOR_GENERIC) \
                 || (DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA) \
-                || (DNNL_AMD_ENABLE_SYCL_KERNELS == 1))
+                || defined(DNNL_AMD_ENABLE_SYCL_KERNELS))
```
densamoilov, line 77: If we don't care about the value then let's remove it from the definition as well:
https://github.com/uxlfoundation/oneDNN/blob/48e7e7379e0b07fd541012ceed837f6bbde21ebc/cmake/options.cmake#L333

third_party/gtest/internal/gtest-port-arch.h
```diff
@@ -35,80 +35,210 @@
 #ifndef GOOGLETEST_INCLUDE_GTEST_INTERNAL_GTEST_PORT_ARCH_H_
 #define GOOGLETEST_INCLUDE_GTEST_INTERNAL_GTEST_PORT_ARCH_H_
 
+// Seems to be an internal undefined variable for us.
```
densamoilov, line 38: This change creates discrepancy between the upstream version and our copy complicating the process of updating gtest. To me it seems that a better way to handle this would be disabling the check for gtest component as we don't use it to build the library. Do you see any problems with this?

densamoilov (approved):

src/gpu/gpu_impl_list.hpp
dzarukin: Makes sense to me, removed, thanks.

third_party/gtest/internal/gtest-port-arch.h
dzarukin: Discussed offline. A trick with removing an option worked fine. gtest.h is still modified since it's included in oneDNN gtests. To avoid the change, a single entry point to gtest.h must be done and a proxy header technique, same as one used for TBB, should be applied there, but it will require whole gtest folder update, so I decided to skip on it this time.

PR #3193, opened 2025-04-30
xe: gemm: clone PVC FHS kernels to FBS

Simonsays095:
Addresses [MFDNN-13596](https://jira.devtools.intel.com/browse/MFDNN-13596). Replaces existing PVC FBS (`--dt=bf16:u4:f32`) kernels with equivalents from the more heavily optimized FHS (`--dt=f16:u4:f32`) set of kernels. These kernels are used in PVC, Xe2, and Xe3 gemm problems.

We have no bf16 layers in performance testing, so we don't see any performance changes in CI. For the following layer (tested manually), Windows PTL time reduced from 35 ms to 10 ms, in line with FHS kernels.
> --matmul --engine=gpu  --dt=f16:u4:f16 --stag=ab --wtag=ba --dtag=ab --attr-scratchpad=user --attr-fpmath=bf16:true --attr-scales=wei:per_ocic:bf16:64x1 --attr-zero-points=wei:per_ocic:s8:64x1 --mode=P 1009x4096:4096x14336


petercad:
@Simonsays095 Do `./ktool -g -A 12p8 FBS` and `./ktool -t -A 12p8 FBS` pass for the new kernels?

Simonsays095:
> @Simonsays095 Do `./ktool -g -A 12p8 FBS` and `./ktool -t -A 12p8 FBS` pass for the new kernels?

Thanks for the suggestion, Peter - this uncovered a few issues (at least some in gemmstone and one in nGEN) which I'm making tickets for. FHS has a few accuracy issues and a few kernels with errors under `ktool -t`. I've modified the last two FBS kernels to use a smaller tile size, and now both kernel generation and accuracy are on par with FHS.

petercad (approved):

PR #3212, opened 2025-05-05
ukernel: exception free mode

dzarukin:
PR implements the first part of the proposal from https://github.com/uxlfoundation/oneDNN/pull/2893 and also updates some cosmetics leftovers from older PRs.

### DIFF
--- a/include/oneapi/dnnl/dnnl_ukernel.hpp
+++ b/include/oneapi/dnnl/dnnl_ukernel.hpp
@@ -277,10 +277,8 @@ struct brgemm : public handle<dnnl_brgemm_t> {
         dnnl_pack_type_t c_pack_type;
         dnnl_status_t status = dnnl_brgemm_get_B_pack_type(&c_pack_type,
                 memory::convert_to_c(a_dt), memory::convert_to_c(b_dt));
-        if (status != dnnl_success)
-            error::wrap_c_api(status, "could not query B pack type");
-
-        return static_cast<pack_type>(c_pack_type);
+        return status == dnnl_success ? static_cast<pack_type>(c_pack_type)
+                                      : dnnl::ukernel::pack_type::undef;
     }
 
     /// Returns the size of a scratchpad memory needed for the BRGeMM ukernel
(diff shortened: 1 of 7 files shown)


dzarukin:
make test

densamoilov (approved):

include/oneapi/dnnl/dnnl_ukernel.hpp
```diff
@@ -277,9 +279,7 @@ struct brgemm : public handle<dnnl_brgemm_t> {
         dnnl_pack_type_t c_pack_type;
         dnnl_status_t status = dnnl_brgemm_get_B_pack_type(&c_pack_type,
                 memory::convert_to_c(a_dt), memory::convert_to_c(b_dt));
-        if (status != dnnl_success)
-            error::wrap_c_api(status, "could not query B pack type");
-
+        (void)status;
```
densamoilov, line 282: The `c_pack_type` variable is not initialized and if `dnnl_brgemm_get_B_pack_type` returns a status that is not `success ` then we will get an undefined behavior.

Shouldn't we return `undef` if `status != success`?

dzarukin: That's a good point. I've added initialization into a C API call. Now it should be good with any kind of status.

dzarukin:
make test
disable test_device_gpu

include/oneapi/dnnl/dnnl_ukernel.hpp
densamoilov: The standard practice is to keep "out" variables unchanged if the getter function call fails to complete. So this logic should be implemented at the C++ level because we are free to define semantics of the C++ API and we want `get_B_pack_type` to return `undef` in all cases except for successful ones.

This is how we generally handle exceptionless cases in our C++ API:
https://github.com/uxlfoundation/oneDNN/blob/37ba9f94319a44ace08d9e0ae0a27006f92a1fe7/include/oneapi/dnnl/dnnl.hpp#L3191-L3198

dzarukin: Given that C API users can rely on a status, the proposed solution is fine with me. Updated, thanks.

vpirogov (approved):

PR #3216, opened 2025-05-06
cpu: x64: brgemm: fix handling big offsets

densamoilov:
Fixes MFDNN-13215.

There is an integer overflow in the BRGEMM kernel that results in undefined behavior when calculating the offset for the prefetch instruction. On Windows, using the DPC++ compiler in Release mode, this leads to an offset that fails some of xbyak’s internal checks, resulting in a runtime error that was reported in the tracker. In contrast, in Debug mode, on Linux, or when using different compilers, the generated offset passes the Xbyak checks and is passed to the prefetch instruction, which means the instruction silently fails to behave as expected. After fixing the overflow, the runtime error became consistent, and the reported xbyak error was "offset too big."

So the fix is to handle the big offsets via a temporary register (the standard solution).

### DIFF
--- a/src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
@@ -159,6 +159,12 @@ private:
     const reg64_t reg_C = r15;
     const reg64_t reg_aux_C = r14;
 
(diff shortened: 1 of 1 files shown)


densamoilov:
make test

ankalinin (approved):

dzarukin (approved):

densamoilov:
make test

PR #3220, opened 2025-05-06
build: aarch64: Update minimum ACL version to 52.0.0

Sqvid:
# Description
ComputeLibrary-v52.0.0 is the first release to use semantic versioning. The CMake rules have been updated to add sanity checks based on this new paradigm.

# Checklist

## General

### DIFF
--- a/cmake/ACL.cmake
+++ b/cmake/ACL.cmake
@@ -1,5 +1,5 @@
 # ******************************************************************************
-# Copyright 2020-2024 Arm Limited and affiliates.
+# Copyright 2020-2025 Arm Limited and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


cmake/ACL.cmake
```diff
@@ -31,31 +31,62 @@ endif()
...
+                # known incompatible maximum version:
+                if("${ACL_MAX_MAJOR_VERSION}" STREQUAL "")
+                    message(WARNING
+                        "Build may fail. Using a newer ACL version than officially supported.\n"
+                        "Detected ACL version ${ACL_MAJOR_VERSION}, but "
+                        "supported version is ${ACL_MIN_MAJOR_VERSION}\n"
+                    )
+                else()
+                    if("${ACL_MAJOR_VERSION}" GREATER "${ACL_MAX_MAJOR_VERSION}")
+                    message(FATAL_ERROR
```
Ryo-not-rio, line 79: We might want to make this a warning too, otherwise it will make testing oneDNN with new versions of ComputeLibrary a headache

src/cpu/aarch64/matmul/acl_matmul.cpp
```diff
@@ -178,11 +178,9 @@ status_t acl_matmul_t::execute_forward(const exec_ctx_t &ctx) const {
...
     std::unique_lock<std::mutex> locker {mtx_, std::defer_lock};
 
-    // Some of the underlying kernels used by ACL still require some state and
-    // are not safe to be called in parallel with different execution contexts.
-    // Eventually when all kernels are truly stateless, this guard can be
-    // removed.
-    if (!acl_obj_->asm_gemm.has_stateless_impl()) { locker.lock(); }
+    // Non-fixed-format kernels in ACL hold shared state and are not safe to be
+    // called in parallel with different execution contexts.
+    if (!IsFixedFormat) { locker.lock(); }
```
Ryo-not-rio, line 183: Looks like a non-related diff got jumbled in

Ryo-not-rio (changes requested):
Just a random matmul change to be removed

cmake/ACL.cmake
vpirogov: By definition of semantic versioning ACL v53 will not be API compatible, so headache is almost guaranteed. With semantic versioning we need to track minimal supported version in terms of MAJOR.MINOR:
* If ACL version is >= minimal supported version we are good. The reason MINOR is important here because we may rely on API introduced in a minor release, so oneDNN will not work with older versions.
* If ACL version is unknown or the major version is larger than minimal supported throw a warning.

src/cpu/aarch64/matmul/acl_matmul.cpp
Sqvid: No the rules for which kernels need to be locked has changed since v25.02. So this is needed to keep things thread-safe.

cmake/ACL.cmake
Sqvid: @vpirogov I agree, the current flow is as follows:

Unreleased/undetermined version = WARNING. You are on your own.
Less than minimum = FATAL_ERROR. By definition incompatible.
Greater than minimum = WARNING. There is an API break but it may or may not break oneDNN. On your own.
Greater than known maximum = FATAL_ERROR. Known to be broken.

Anything else is SUCCESS.

Sqvid: @vpirogov If the flow above makes sense to you, could you review it on behalf of @uxlfoundation/onednn-maintain?

src/cpu/aarch64/matmul/acl_matmul.cpp
Sqvid: @Ryo-not-rio bump for re-review

cmake/ACL.cmake
vpirogov: Looks very comprehensive. Thanks, @Sqvid.

vpirogov (approved):

Ryo-not-rio (approved):

theComputeKid (approved):

PR #3231, opened 2025-05-08
[Benchdnn][GPU] Drop fwks_llm shapes larger than 20k

kealan-barbieri:
# Description

In spite of @dzarukin s heroic work these huge shapes still cause timeouts on integrated and older gen platforms. Drop LLM shapes with dims over 20k to hopefully avoid timeouts on any platform. 

Fixes # [MFDNN-13114](https://jira.devtools.intel.com/browse/MFDNN-13114)

# Checklist

## General


kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul

dzarukin (approved):
Some notes, your choice:
1. Maybe comment instead of removing. This way at least one could run them. Grouping at the end of the file would really help to do it with several magical buttons.
2. Consider transferring LLM file into a dedicated test_matmul_llm_gpu file to break it into pieces when running under ECMD. This should speed up things, not piling up LLMs time on top of large matmul time.

Simonsays095 (approved):

echeresh:
@kealan-barbieri I see some slow steps with XeLP/XeLPG in CI. I guess some LLM cases hit reference - hence it's slow. If this is the case, it would be nice to somehow avoid running LLM files on these platforms. What do you think?

kealan-barbieri:
@echeresh I think youre right all fwks inputs should definitely have `--skip-impl=ref`, I'll add it in. If its still slow Dmitry suggested splitting this into a separate top level test-set so it can run in parallel with other steps.

echeresh:
> @echeresh I think youre right all fwks inputs should definitely have `--skip-impl=ref`, I'll add it in. If its still slow Dmitry suggested splitting this into a separate top level test-set so it can run in parallel with other steps.

But there is a risk we miss unexpected reference, say on Xe2 with KPI cases which is not good. I think this needs a different approach. One thought I have is to:

- Rely on `--check-ref-impl=true`
- Introduce `skip_ref_impl_*()` to catch allowed reference dispatches. In other words, this is to override `--check-ref-impl` behavior for selected cases (e.g. XeLP + `int4`)
    - We already have `skip_unimplemented_*()` serving the same purpose for unimplemented cases that are expected

Maybe there a simpler option though. +@dzarukin for his thoughts.

dzarukin:
> Maybe there a simpler option though. +@dzarukin for his thoughts.

If the goal is to remove Xe-LP from consideration, easier to handle it on Infra side by explicit target removal. Then a dedicated test file would definitely a must.

There's no API to query GPU ISA so there's no *simple* or *direct* way to filter out undesired ISA programmable way.

echeresh:
> If the goal is to remove Xe-LP from consideration, easier to handle it on Infra side by explicit target removal. Then a dedicated test file would definitely a must.

Thanks, as a quick solution, handling it via a separate target should probably help.

But in general, it's not like LLM cases vs other cases. For example, I see some "reference" cases in `tests/benchdnn/inputs/matmul/option_set_fwks_key_gpu` (FP8 and cases with zero-points would dispatch to `ref` on XeLP). If going this path, we need to consolidate XeLP-non-optimized cases in one input file.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul

kealan-barbieri:
Separated into its own file with `--check-ref-impl=true`, filed https://jira.devtools.intel.com/browse/IPLQA-449 requesting the new set be disabled on XeLP.

PR #3233, opened 2025-05-09
rfcs: graph api: support SDPA training

ElaineBao:
## Description

The RFC proposes adding support for SDPA training in oneDNN Graph API.

Rendered version: [link](https://github.com/uxlfoundation/oneDNN/blob/yixin/rfc/sdpa-training/rfcs/20250428-graph-api-sdpa-training/README.md)


rfcs/20250428-graph-api-sdpa-training/README.md
```diff
@@ -0,0 +1,563 @@
...
+used to apply the mask and scale the output.
+
+![cuDNN dropout](images/cudnn_dropout.png)
+
+#### SDPA Backward
+
+With the precomputed SoftMax $Stats$, the SoftMax output $P$ can be reconstructed as:
+
+> $$
+> P = e^{S - stats} \in \mathbb{R}^{B \times H_{q} \times S_{q} \times S_{kv}}
```
dzarukin, line 202: What I'm missing is why to save `Stats` when one can simple save full softmax output and use it?

ElaineBao: The primary reason SDPA forward keeps Stats instead of the softmax output (P) is mainly to optimize memory usage:

- Softmax Output (P): Storing the full softmax output requires significant memory, especially for long sequences, as it scales with the size of the attention matrix (B × H × S_q × S_kv).
- Softmax Stats: Storing only the statistics (e.g., log_sumexp) requires much less memory because it reduces the storage to a smaller tensor (e.g., B × H × S_q × 1), which is independent of the sequence length S_kv.

By keeping the statistics instead of the full output, the memory footprint is significantly reduced, which is critical for large-scale models and long sequences.

Frameworks like cuDNN and PyTorch store softmax stats instead of the full softmax output for the same reasons. This approach aligns with industry best practices and ensures compatibility with other PyTorch SDP backends.

TaoLv: If memory footprint is the consideration, I'm wondering if it's also a request on CPU if passing softmax output from forward to backward can save a lot of computation.

rfcs/20250428-graph-api-sdpa-training/README.md
```diff
@@ -0,0 +1,563 @@
...
+
+> $$
+> S_{scaled} = f_{scale}(S) \in \mathbb{R}^{B \times H_{q} \times S_{q} \times S_{kv}}
+> $$
+> $$
+> S_{masked} = f_{mask}(S_{scaled}) \in \mathbb{R}^{B \times H_{q} \times S_{q} \times S_{kv}}
+> $$
+
+The following sections explore how to support SDPA training in the oneDNN Graph
+API. We will start from `f32` data type.
```
TaoLv, line 70: Just to clarify, do you plan to cover bf16, f16, or even fp8 training in this document?

rfcs/20250428-graph-api-sdpa-training/README.md
dzarukin: @ElaineBao, could you please put an explanation for that into the document since it's not an obvious optimization. BTW, this saves space only when the softmax input is also stashed for further computations, otherwise, it will require it for nothing. It's nice to confirm it's always the case whether optional pieces present in the pattern or not.

@TaoLv, I'd say CPU has less demand on memory footage. If the model doesn't fit the host, it'll never fit the device. From performance perspective, re-compute exponent over reading ready to use memory, it's definitely cheaper just to read.

dzarukin (commented):
I'm concerned about two things:
1. There's no exploration for backend specific things that are supported or not currently. It will give a better picture what the cost is going to be.
2. Validation topic is not covered at all. I'd really like to see the plan how it's going to be verified it's actually working.

rfcs/20250428-graph-api-sdpa-training/README.md
ElaineBao: Plan to start from f32, I add the clarification at the section of implementation effort. Thanks.

ElaineBao:
> I'm concerned about two things:
> 
> 1. There's no exploration for backend specific things that are supported or not currently. It will give a better picture what the cost is going to be.
> 2. Validation topic is not covered at all. I'd really like to see the plan how it's going to be verified it's actually working.

Thanks for the suggestions, add a section of implementation effort to cover these two topics.

rfcs/20250428-graph-api-sdpa-training/README.md
```diff
@@ -0,0 +1,616 @@
...
+
+#### Proposal 1.B: Add Dropout Operation to oneDNN Graph API
+
+The proposed definition for the Dropout operation is as follows:
+
+| **Inputs/Attributes/Outputs** | **Name**           | **Description**                                                               |
+|-------------------------------|--------------------|-------------------------------------------------------------------------------|
+| Input                         | `src`             | The tensor to which Dropout will be applied.                                   |
+| Input                         | `seed`            | Seed for the Philox random number generator used in Dropout.                   |
+| Input                         | `offset`          | Offset for the Philox random number generator used in Dropout.                 |
```
TaoLv, line 355: I'm not sure if Philox is only requested algorithm to do random number generation for Dropout. @mgouicem do you have any comments here?

ElaineBao: In Pytorch, cuda and xpu Dropout uses Philox, cpu uses MT19937.

RNG Type | MT19937 | Philox
-- | -- | --
Algorithm | Mersenne Twister | Counter-based block cipher-like RNG
Parallelism | Not parallelizable | Highly parallelizable
Performance | Optimized for single-threaded CPU | Optimized for multi-threaded/GPU
Period | (2^{19937} - 1) | Very large (depends on counter size)
Use Case | General-purpose RNG | HPC, GPU, and parallel environments

reference: 
- cuda: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Dropout.cu#L403
- xpu: 
https://github.com/intel/torch-xpu-ops/blob/main/src/ATen/native/xpu/sycl/Dropout.cpp#L399
- cpu: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/CPUGeneratorImpl.cpp#L94

TaoLv: If we define a new Dropout operation, we will expect it to be general and can be used to define different fusions on different platforms.

TaoLv: How about RNG operation? Does it have different implementations in frameworks?

ElaineBao: There's no dedicated RNG operation in PyTorch. In cuDNN, it uses Philox.

ElaineBao: update: PyrTorch has `random_` operation, which uses the same random number generator as dropout, meaning both philox and mt19937 are used.

rfcs/20250428-graph-api-sdpa-training/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+##### Cons
+
+- Requires adding a new Dropout operation to the oneDNN Graph API.
+- A reference implementation for Dropout is needed. Currently, the oneDNN
+  Primitive API has limited support for Dropout, implemented as a primitive
+  attribute for `MatMul` without Philox offset input. Supporting SDPA training
+  may require extending this implementation.
+
+#### Proposal 1.C: Add RNG (Random Number Generator) Operation to oneDNN Graph API
```
Valentine233, line 378: For CPU, will the API be different if we use another generator other than RNG?

ElaineBao: IIUC, you mean `will the API be different if you use another generator other than Philox`. 

I think there will be at least one more attribute for selecting different generator algorithms, e.g. Philox, MT19937. And if different generator requires different inputs (e.g. seed and offset for Philox, and maybe other inputs for MT19937), then other inputs may be added.

So the question will be what `another generator` will you use and what inputs are needed for it?

Valentine233: Yes, you are right that I refer to `Philox` here. As CPU has not the request yet, I leave this as an open. Just to confirm if other generators may impact the API.

LuFinch:
Hi, I have confirmed that Pytorch team currently doesn't commit any models that use `dropout!=0` during training.  Hence dropout can be low priority.

rfcs/20250428-graph-api-sdpa-training/README.md
mgouicem: Some comment here on RNG for dropout: the same random mask has to be used between forward and backward pass, the only ways I know of to achieve that would be:
- use any RNG on FWD, write the mask tensor to memory, and pass that mask tensor to bwd. This require more memory bandwidth and more memory capacity for training.
- generate the same mask tensor on both fwd and bwd. This saves bandwidth and memory capacity at the cost of extra compute. Note that here the exact same algorithm shall be used between fwd and bwd
  - using a CBPRNG (counter based PRNG), like philox, makes rng values depend only on seed and tensor value index. So implementation is free to traverse input tensor as it pleases.
  - using a classic PRNG (eg. MT), requires to fix the order of tensor traversal to generate the same bits for a given tensor value index. Here the first option (write/read mask tensor) would likely have lower impact on perf.

rfcs/20250428-graph-api-sdpa-training/README.md
```diff
@@ -0,0 +1,648 @@
...
+
+## Conclusions
+
+The decision is to implement Proposal 1.B (Add Dropout Operation to oneDNN Graph
+API) and Proposal 2.B (Extend Existing SoftMax Operation). This approach enables
+comprehensive support for SDPA training in the oneDNN Graph API, aligning with
+requirements from major frameworks and ensuring interoperability.
+
+A dedicated RFC will be opened to discuss the detailed definition and
+implementation of the Dropout operation.
```
ElaineBao, line 568: A dedicated RFC for dropout is opened: https://github.com/uxlfoundation/oneDNN/pull/3939

PR #3236, opened 2025-05-09
rfcs: host-side scalars support

mgouicem:
[Link to rendered document](https://github.com/uxlfoundation/oneDNN/blob/mgouicem/rfcs/host_scalar/rfcs/20250430-host-side-scalars/README.md)


rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,155 @@
...
+We have two paths:
+- encapsulate host-side scalars in memory objects: this would create
+  overhead for passing a scalar but would not require execution API
+  changes. There is a question of object lifetime to clarify here: for
+  host-side scalar, we have to guarentee that they are not accessed
+  asynchronously, otherwise users will incur extra overhead of
+  allocation and copy for a single scalar.
+- expose new execute function to allow passing scalars. For both C and
+  C++ APIs, it would require to take extra unordered_map for scalar
+  arguments.
```
dzarukin, line 124: It seems to me this part should be extended further with API examples.
If the proposal is to put scalars into a map, what the execution argument is gonna be for it? A new one, or re-used? If re-used, what kind of internal changes what be required to distinguish regular memory from a scalar?
How the `dnnl::memory` object creation with host_scalar would look like? New API, or re-use one? If re-used, is engine redundant there? Or the ctor with memory allocation, which kills the whole purpose of the feature?

If the execute function allows to split constants and memories, not quite clear how the implementation should identify them which comes where if there are more than one scalar.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,155 @@
...
+- no clear benefit for performance other than specific scalar case.
+
+## How should host-side scalar be passed to execute call?
+
+Currently, the oneDNN primitive API only accepts `dnnl::memory` objects
+(through an `unordered_map` in C++ interface, or an array of
+`arg_kind`/`memory` for C interface).
+
+We have two paths:
+- encapsulate host-side scalars in memory objects: this would create
```
TaoLv, line 116: Creating a memory object for a host scalar requires to create a host engine as well:
- when `CPU_RT=NONE`, I'm not sure if the user can create a CPU engine.
- when `CPU_RT=SYCL`, the address of the host scalar is not a SYCL memory.

Or, do you think we can use a GPU engine to create a memory object for a host scalar?

I think we discussed it when adding the host scalar support to graph API, but feel like to also mention it here.

petercad:
@mgouicem Did you consider host USM as an option? It neatly encapsulates the concept of device-accessible, host-resident (and accessible) memory.

Pros:
* No changes to primitive implementations (either kernels or host-side code)
* No changes to memory/memory descriptor objects

Cons:
* Asynchronous access from kernels places some burden on users
* (Usually slight) performance hit for the first GPU access to host USM memory due to the long latency of the PCI read.

So, it's not perfect, but at the same time might be the right balance between developer time and performance.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+Pros:
+- avoids user from recreating primitives when scalars are changed
+- simple enabling internally
+
+Cons:
+- ties the memory descriptor to a specific device (host) which we
+  avoided so far.
+
```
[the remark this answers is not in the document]
mgouicem: > It seems that option 2 (like option 3) requires oneDNN to be compiled with CPU support and requires the user to create a CPU engine -- is that correct? If so, it seems like a lot of overhead (space + time) for users that otherwise don't compile for CPU, like OV's Intel GPU plugin.

It requires CPU engine but not CPU support. Though exposing a dedicated dummy host_engine, or just "null" engine makes sense!

rfcs/20250430-host-side-scalars/README.md
mgouicem: Yes we would need an engine. We can either rely on existing CPU engine, or expose a new host_engine or "null" engine.

Exposing a host engine would be cleaner IMHO, and it would be limited to memory creation.

mzhukova: hey @TaoLv, I've added this into RFC (let me know if any more info is needed on this), is it OK to resolve the thread?

rfcs/20250430-host-side-scalars/README.md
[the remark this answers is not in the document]
mzhukova: hi @petercad, I've added this note to RFC, I hope it make sense, let me know if anything else is needed here or feel free to resolve this thread! :)

mzhukova:
> Did you consider host USM as an option? It neatly encapsulates the concept of device-accessible, host-resident (and accessible) memory.

@petercad I've added this as option 4, thanks for suggestion! we're currently awaiting some evaluation of whether this approach would work

rfcs/20250430-host-side-scalars/README.md
mzhukova: hi @dzarukin, if I understood correctly, the ask is only for 2nd option, which is currently not what is proposed, but ~possible alternative? would you still like to see that level of details? do you have any concerns with recommended option (which is having scalars as memory objects)?

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+## Motivation
+
+Currently, primitives assume that all inputs passed at execute time
+are on the same device as the one they execute on (with the exception
+of reorder).
+
+A new request is to pass scalar values that reside on host device
+to GPU kernels at execution time.
+A good example is the scale for SDPA operator in Pytorch [^1], which is
+a python variable on the host and not a torch.tensor object.
```
densamoilov: Is this motivation still relevant given that there is a way to pass the scalar to oneDNN? Was there a concern about associated overhead?

I just remembered that we received some complaints in the past about the CPU, where passing a single scalar requires creating the entire memory object, which incurs unnecessary overhead. So maybe it would make sense to look at the issue beyond the GPU.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+Cons:
+- Ties the memory descriptor to a specific device (host) which we
+  avoided so far.
+
+> Note: Options 2 and 3 formally require creating memory objects with a `CPU` engine.
+However, since this is a very specific use case that does not otherwise
+require full CPU support in oneDNN, it is suggested to introduce a new `host` or "null" engine.
+This would allow users to avoid building oneDNN with complete CPU support just for this scenario.
+
+### Option 3: explicitly take engine kind in memory desc
```
densamoilov: Formally, our specification [states](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onednn/source/data_model/memory/objects#descriptors) that memory descriptors are engine agnostic so unless we are fine with changing the current data model this option doesn't look viable.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
+# Host-side scalars Support in oneDNN Primitives
```
CaoZhongZ, line 1: How about we just call it as "scalar" instead of "host-side scalar". Remove it will ease the unnecessary pain in considering engine or other things. Is there a case of Device-side scalars?

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+A new request is to pass scalar values that reside on host device
+to GPU kernels at execution time.
+A good example is the scale for SDPA operator in Pytorch [^1], which is
+a python variable on the host and not a torch.tensor object.
+
+There are two challenges here:
+- How to specify to primitive descriptor creation that a given input is
+  a host scalar?
+    - Currently, we don't have any engine specified to input memory descriptors.
```
CaoZhongZ, line 15: If we abandon the concept of "host scalar" we don't have to worry about engine. I'm comfortable to associate a scalar value with GPU engine.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+- How to pass host-side scalars to the execute function?
+  - Currently, we only take dnnl::memory objects.
```
CaoZhongZ: ~~Yes, pass the pointer to the value with dnnl::memory. Submission process treat the contents as value and send to GPU the same way send pointers. JITed kernel use the value already pull in the registers.~~

CaoZhongZ (changes requested):

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+- Avoids user from recreating primitives when scalars are changed.
+- Simple enabling internally.
+- Allows passing scalar as a kernel argument, that should not introduce
+a performance penalty.
+
+Cons:
+- Ties the memory descriptor to a specific device (host) which we
+  avoided so far.
+
+> Note: Options 2 and 3 formally require creating memory objects with a `CPU` engine.
```
CaoZhongZ: Create it with GPU engine is fine? Why have we stick it with CPU?

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+(through an `unordered_map` in C++ interface, or an array of
+`arg_kind`/`memory` for C interface).
+
+We have two paths:
+1. (recommended) Encapsulate host-side scalars in memory objects: this would create
+overhead for passing a scalar but would not require execution API
+changes. There is a question of object lifetime to clarify here: for
+host-side scalar, we have to guarantee that they are not accessed
+asynchronously, otherwise users will incur extra overhead of
+allocation and copy for a single scalar.
```
CaoZhongZ: Lifetime should be fine as long as onednn use setArgument APIs to pass the value at submission time. Also the reason we shouldn't use USM solution.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+descriptors as it would likely provide little benefit over host-side
+scalar specific memory descriptor, but being more general, it would
+also come with more complications.
+
+> Note: Options 2 and 3 formally require creating memory objects with a `CPU` engine.
+However, since this is a very specific use case that does not otherwise
+require full CPU support in oneDNN, it is suggested to introduce a new `host` or "null" engine.
+This would allow users to avoid building oneDNN with complete CPU support just for this scenario.
+
+### Option 4 (2nd recommended): do nothing and suggest user to use USM `malloc_host`
```
CaoZhongZ: It's a nasty solution. USM incurs memory cache allocator behavior when we call frequently to onednn.

rfcs/20250430-host-side-scalars/README.md
CaoZhongZ: My bad, could just pass value with dnnl::memory. Call a set_value (similar to set_data_handle) API to modify the content each time we use the same memory object.

rfcs/20250430-host-side-scalars/README.md
mzhukova: The idea here was to support specific input which is happened to be a scalar value in reality and it is coming from the host (that's why the title), but is then being used in GPU kernel. So this is not really about ~host-side scalar support vs ~device-side scalar, it is about improving (decreasing overhead e.g. for Pytorch) for a specific scenario of input.
I'll try to think how to make it more clear in the title.

CaoZhongZ: Yes, the scalar value can be materialized on both host and GPU with the overhead "transparent" to users. In a sense the value is already on GPU as we want when threads start to execute. Assign GPU engine is somehow not conflict with its nature, that's the point.

rfcs/20250430-host-side-scalars/README.md
mgouicem: > Is this motivation still relevant given that there is a way to pass the scalar to oneDNN?

The issue here is more about where that scalar resides. If it was a torch.tensor, it would reside on the device and we would be good (aligned with current API). However, because it is on the host, it is not aligned with our current API for execution time arguments (those take only device memory, that live on the same engine as the primitive, except for reorder).

rfcs/20250430-host-side-scalars/README.md
mgouicem: yes correct, this option is to open discussion about reevaluating memory descriptors being engine agnostic.

rfcs/20250430-host-side-scalars/README.md
mgouicem: >  Assign GPU engine is somehow not conflict with its nature, that's the point.

It is, otherwise we would have no way to differentiate between:
- a scalar value on the host transfered to GPU as kernel argument
- a scalar value that resides on the device memory, which we can pass to GPU kernel only through pointer.

rfcs/20250430-host-side-scalars/README.md
mgouicem: Unfortunately, we need to differentiate between a scalar on host memory and a scalar on device memory at primitive creation time, since the generated kernel will be different in both cases (in first case, we take scalar directly as kernel argument, in second case, we take the address of the scalar as kernel argument).

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+The new kind of memory descriptor would be created through a dedicated function
+It would be very lightweights and be limited:
+- only scalar value, so no `ndims`, no `dims`, ...
+- only scalar datatype can be configurable by user.
+
+```C
+dnnl_status_t DNNL_API dnnl_memory_desc_create_host_scalar(
+        dnnl_memory_desc_t *memory_desc, dnnl_data_type_t data_type);
+```
```
TaoLv, line 69: Is there any implication or consideration about using this API and host-side scalar memory in a CPU only build and CPU primitives? It seems we don't have a `host-side scalar` concept in that case. Do we expect that the CPU primitives will work with both a normal scalar memory or a host-side scalar memory (created from this new API)?

mgouicem: I wonder if we should just call this `scalar` instead of `host_side_scalar` as it seems to tie it to a device and makes for confusion.

From CPU/GPU perspective it should be treated the same: a scalar is passed to execute function, and implementation has to manage that.

From execute argument creation, we could even ditch the engine from memory object creation with a dedicated function/constructor as suggested by @densamoilov [here](https://github.com/uxlfoundation/oneDNN/pull/3412#discussion_r2161990682).
That would make the API simpler and make scalar engine-independant (as it should).

rfcs/20250430-host-side-scalars/README.md
mzhukova: hey @dzarukin as for extending primitive execution APIs, we didn't explore this option further in details, as the recommendation is to get with mem descr / mem constructor changes. Let me know if you want the other option considered instead. 
As for the memory object creation, I've put more details / how the API would look like and you could also follow the link to the POC.

rfcs/20250430-host-side-scalars/README.md
mzhukova: Resolving this thread for now, as RFC got updated and this doesn't seem to be relevant anymore.. Please feel free to reopen / start a new one

rfcs/20250430-host-side-scalars/README.md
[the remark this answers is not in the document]
mzhukova: Resolving this thread for now, as RFC got updated and this doesn't seem to be relevant anymore.. Please feel free to reopen / start a new one

rfcs/20250430-host-side-scalars/README.md
mzhukova: hey @densamoilov, I've updated the RFC, let me know if this thread is still relevant (or please resolve otherwise)

rfcs/20250430-host-side-scalars/README.md
mzhukova: resolving as this seems answered

rfcs/20250430-host-side-scalars/README.md
mzhukova: I believe Mourad cleared the confusion, so resolving this thread

rfcs/20250430-host-side-scalars/README.md
mzhukova: (seems like a duplicate, so resolving)

rfcs/20250430-host-side-scalars/README.md
mzhukova: Put some updates wrt main proposed option and USM, I believe this should resolve the thread

rfcs/20250430-host-side-scalars/README.md
mzhukova: Moved out USM option from recommended (and keeping it just for the record), so resolving this thread

rfcs/20250430-host-side-scalars/README.md
mzhukova: Updated the recommended option and API to not have engine, POC link is available, so resolving this thread.
Feel free to reopen / start a different one if any clarification is needed still.

mzhukova:
I've re-requested review from everyone who participated, also tagging @mgouicem and @pv-pterab-s :)

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,205 @@
...
+// C++ API
+
+// Scalar value
+float value = 42.0f;
+
+// Create a memory object for the scalar (no engine needed)
+memory scalar_mem(memory::desc::host_scalar(memory::data_type::f32), &value);
+
+// Use as regular memory object in execute function...
+```
```
mgouicem, line 69: It seems C and C++ API are a bit different here: C++ API takes a desc to create scalar_mem object while the C API takes only datatype.

Any reason for that difference? Personally I am fine with both, maybe slight preference for a static function to create memory without desc and only dt, as it would have lower overhead (no desc to allocate for host-side scalars during execution).
But overloading constructor would be better aligned with existing APIs (e.g. sparse).

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,205 @@
...
+POC for API changes is available in the draft PR [^3],
+you could check it out for simple example, API tests, and some implementation
+details (although the implementation is not complete yet).
+
+Pros:
+- Users do not need to recreate primitives when scalar values change.
+- Allows passing scalars as kernel arguments, minimizing performance overhead.
+
+Cons:
+- Ties memory descriptor to a host.
```
mgouicem: this one is no more relevant since engine is no more passed.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,205 @@
...
+// Create a memory object for the scalar (no engine needed)
+memory scalar_mem(memory::desc::host_scalar(memory::data_type::f32), &value);
+
+// Use as regular memory object in execute function...
+```
+
+POC for API changes is available in the draft PR [^3],
+you could check it out for simple example, API tests, and some implementation
+details (although the implementation is not complete yet).
+
```
mgouicem, line 79: (repeating points from https://github.com/uxlfoundation/oneDNN/pull/3454#issuecomment-3031633111)

We need to clarify what happens with set/get_data_handle, as well as get_engine.

rfcs/20250430-host-side-scalars/README.md
mzhukova: Oh, I'm so sorry for the confusion, it was the typo (POC had the correct API/tests version).
Not updated, the APIs are aligned.

rfcs/20250430-host-side-scalars/README.md
mzhukova: now removed

rfcs/20250430-host-side-scalars/README.md
mzhukova: Sure, will do.

For `{get, set}_data_handle`, I think we should "disable" them and return nullptr/error. Since for scalars we copy from the user's value, `get_data_handle` wouldn't return the user's pointer, and for `set_data_handle`, it probably makes more sense to again copy the user's data instead of just setting the pointer, but this changes the meaning..
However, we still need some way to get and set the scalar memory, so users don't have to recreate the memory object every time the value changes, right?

For the engine, it would be just be nullptr.

What do you think?

mzhukova: Sorry I didn't see the newest comment before replying. It seems like we're aligned here.
Should I introduce new `{get,set}_host_scalar` in this RFC then?

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,204 @@
...
+    that host-side scalar value changes.
+- Potential extra jitting overhead if internal implementation considers that
+    parameter constant.
+
+This option is not recommended if the host-side scalar value is expected to
+frequently change.
+
+### Option 3: Take engine kind in memory descriptor
+
+Here, we would allow users to pass any memory on host or device. This would
```
densamoilov, line 207: I'm a bit puzzled by this option. Suppose memory descriptors become engine-aware, what impact would that have? The primitive descriptor would still be created for a particular engine and memory objects would still be created for a particular engine. If we allow passing memory objects created for different engines then how would making memory descriptors engine-aware help? Would it just be to validate that the engine kinds in memory descriptors and memory objects are consistent?

In any case, I believe that making memory descriptors engine-aware is a deal-breaker as they were designed to be engine-agnostic and contain no engine-specific information. This, in my opinion, is a major drawback of this option so I just want to make sure I understand this option correctly.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,204 @@
...
+For both C and C++ APIs, this option would require introducing a new `execute`
+function overload that accepts an additional container (such as `unordered_map`
+or similar) specifically for scalar arguments. This would allow users to pass
+host-side scalars directly as arguments, separate from the usual memory objects.
+
+Pros:
+- Provides a clear and explicit way to pass scalars without overloading the
+    semantics of memory objects.
+- Avoids tying memory descriptors to a specific device or engine.
+- Can be implemented in a backward-compatible manner by adding new overloads.
```
densamoilov, line 253: I think another important benefit would be getting rid of unnecessary overhead associated with memory descriptor and object creation.

rfcs/20250430-host-side-scalars/README.md
densamoilov: > Should I introduce new {get,set}_host_scalar in this RFC then?

Yes, please. I like the idea of "disabling" `set_data_handle()` and `get_data_handle()` and introducing scalar counterparts. On top of the reasons you've already shared, I have one more: we have an interop API that offers interop versions of these interfaces, so making them scalar-compatible (e.g. via overloading) could result in very confusing interfaces/logic, since those interop interfaces differ from the generic ones.

I also agree on the `get_engine()` proposal, returning `nullptr` sounds reasonable.

densamoilov (approved):
I agree that the recommended option (Expose a new memory descriptor kind) is what we should go with. The design is consistent with our programming model, solves the original problem and doesn't prevent us from introducing a new execute function that takes scalar arguments by value, if we decide to do so in the future.

We also seem to agree on what to do with the auxiliary interfaces such as `{set|get}_data_handle()` and `get_engine()`.

As for the `set_scalar` and `get_scalar` interfaces, we can make them templated or simply use `void*`. I have a slight preference for the templated option, but I'd be okay with the `void*` approach as well.

There are also `map`/`unmap` interfaces, which we can also "disable" as they don't make sense for the scalars as they are always "mapped".

Mourad, Maria, thank you for the proposal!

rfcs/20250430-host-side-scalars/README.md
densamoilov: Thank you, its looks good to me.

rfcs/20250430-host-side-scalars/README.md
mzhukova: incorporated into RFC, so resolving

rfcs/20250430-host-side-scalars/README.md
mzhukova: added this as well

mzhukova:
I would like to ping @uxlfoundation/onednn-arch again to hear your thoughts on latest addition https://github.com/uxlfoundation/oneDNN/pull/3236/commits/ee1300fadc148d37f96333d355cb1429f087a5b4, that is also available in POC https://github.com/uxlfoundation/oneDNN/pull/3506/commits/d1465c9ddb4aa20184b4e07abf7be1fd10b8cf9a

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -85,6 +85,56 @@ Cons:
...
+**Option 1 (recommended):** Introduce new "set" APIs for scales and zero points:
+```C
+void dnnl::primitive_attr::set_host_scale(int arg);
+void dnnl::primitive_attr::set_host_zero_point(int arg);
+```
+This allows us to explicitly specify that the scaling factor or zero point
+is a host-side scalar for primitive descriptor creation.
+Which allows us to prepare a correct kernel that passes the value as a kernel parameter,
+exit early if host-side scalars are not supported for the implementation,
+as well as provide more verbose error messages in case of any failure.
```
mgouicem: I agree with option 1 being favored over option 2. Main argument is that passing quantization from host without copy requires per-tensor scale/zp, and new API makes it explicit that other masks are not supported.

(nit 1) would this new scale/zp attr be incompatible with existing one? (I think it should be incompatible).
(nit 2) new setter might need associated getters.

Maybe option 3: what about using current API and introduce a new special mask value for host scale?

mzhukova: @mgouicem thanks for the input!

as for 1, could you elaborate on incompatibility? do you mean something along the lines ~we should have a check during execute that if `set_host_scale()` was used for primitive creation, then we should be getting host scalar memory object vs "regular" one?

as for 2, I didn't find any getter functions for current scales or zp, but I guess we need to provide a way to query whether pd was created using "host scales" or not, would it be enough to just introduce an API like `is_host_scale()/is_host_zp()` (or maybe) `is_host_attr_value()`?

as for the option 3, I have a slight preference towards option 1 since it is more explicit, but I could explore whether having e.g. `mask == -1` would cause any issues

mgouicem: 1. we would have two scales setter: `set_scales()` and `set_host_scale()`. We need to clarify how they interact with each other if user calls both. I was suggesting to make them mutually exclusive (so if both are called, the first call would succeed but later calls would fail).
2. yep, we had a grey area about what to return when scales are not set, but getter is called. maybe let's keep that getter business on hold until someone actually asks for it.
3. I would argue that reusing the current API is actually the more explicit option, since it deals with point 1. above (no need to clarify interaction between two symbols to set scales), and can be defined as accepting scalar only. For the value, we already have a special value (INT_MIN) to check if scales were set (see [here](https://github.com/uxlfoundation/oneDNN/blob/8a168123b3ac80f8504f5a108d976b7bd8db7849/src/common/primitive_attr_quant.hpp#L114)). I would advocate to not use -1 though, as this would cause bits related to dimensions to be set. Maybe something like (INT_MIN + (INT_MIN >> 1)).

densamoilov: >  was suggesting to make them mutually exclusive (so if both are called, the first call would succeed but later calls would fail

It may be a little too strict. I don't think calling one of the interfaces should make the attributes special. I don't see problems with overwriting the scales for a particular argument.

mzhukova: Added option about new mask into RFC, my slight preference is still for `set_host_{scale, zp}` as this could be extended same way for other attributes. Also, the mask is "naturally" 0 for host scalars scsales/zps too, so I wonder if having a different one might look a bit confusing.. I'm not super attached though.

I wonder if anyone else in @uxlfoundation/onednn-arch has any opinion.

Note: https://github.com/uxlfoundation/oneDNN/pull/3506 currently implements option with new API, but I'll change this if we decide the other way.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,278 @@
...
+// Create a memory object for the scalar (no engine needed)
+memory scalar_mem(memory::desc::host_scalar(memory::data_type::f32), value);
+
+// Use as regular memory object in execute function...
+```
+
+Additional considerations for the new memory object:
+- The `get_engine()` method will return `nullptr` for host-side scalar memory objects.
+- The `{set, get}_data_handle()` APIs will either return `nullptr` or an error code, since the scalar value is stored internally as a copy (and not user-provided memory) that changes the previous meaning/behavior of those functions.
+- To retrieve or update the scalar value stored in a memory object, a new API `{set, get}_host_scalar_value()` will be introduced.
```
dzarukin, line 74: I'm not sure a getter is particularly useful in here, it's always easier to set a needed value and that's it.
Also, if you allow to set the value, there should be a ctor that doesn't take the value, or a default one as NAN (or something) is provided.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,278 @@
...
+you could check it out for simple example, API tests, and some implementation
+details (although the implementation is not complete yet).
+
+Pros:
+- Users do not need to recreate primitives when scalar values change.
+- Allows passing scalars as kernel arguments, minimizing performance overhead.
+
+Cons:
+- Requires explicit support in each non-host implementation to handle host-side scalars
+    correctly.
```
dzarukin, line 86: I might be wrong, but this is the reality regardless of the option to support host-side scalars from API point of view.

rfcs/20250430-host-side-scalars/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+**Option 2**: Introduce new mask value for existing APIs specifically for
+host-side scalars:
+```C
+#define DNNL_MASK_HOST_SCALAR (INT_MIN + (INT_MIN >> 1))
+```
+This would allow us to specify that a scaling factor or zero point is a
+host-side scalar without introducing new APIs.
+The user would pass this mask value to the existing `set_scales(...)`
+or `set_zero_points(...)` APIs.
```
dzarukin: If the intention of the option to identify host scalar through the mask (or just through the argument), I'd suggest to pass a new bool flag instead:
```
    set_scales_mask(int arg, int mask, bool is_host = false);
```

This way it's at least explicit. But my vote would still be for Option 1.

I also agree with Denis that no special attention to the scales call shouldn't happen - we allow to reset scales if the call has already been done before the current one. In case when set_host_scales() happen on top of regular one, it just resets what was set and specify that it's going to be a host scalar, and vice verse. So, Con 2 and 3 shouldn't be relevant.

rfcs/20250430-host-side-scalars/README.md
mzhukova: > I'm not sure a getter is particularly useful in here, it's always easier to set a needed value and that's it.

One application that I could think of -- it is useful for when you lost track of what was happening before and then primitive failed (or returned unimplemented), and you just want to query the scalar value so to fallback to some reference implementation.

> Also, if you allow to set the value, there should be a ctor that doesn't take the value, or a default one as NAN (or something) is provided.

Just setting the value is also useful if you want to "save time" on re-creating the whole memory object. I'm not sure what's the point now to have an empty storage w/a a value, but I guess that could be done if necessary. Do you see a useful case for that?

rfcs/20250430-host-side-scalars/README.md
mzhukova: resolving as outdated

rfcs/20250430-host-side-scalars/README.md
mzhukova: Per a couple of offline discussions, I've updated and RFC and moved recommendation to option 2b (Dima's suggestion in https://github.com/uxlfoundation/oneDNN/pull/3236#discussion_r2252669718), that is essentially a middle ground, where we would not introduce a new API (except for C `..._v2`) and would not change the meaning of the mask, that were two main concerns in this discussion. As the result, I'm resolving this thread. Thanks everyone for the input!

rfcs/20250430-host-side-scalars/README.md
mzhukova: This was poor wording, I've updated.

rfcs/20250430-host-side-scalars/README.md
mzhukova: Agree with the boolean flag as an option, I've added it into RFC as option 2b.

rfcs/20250430-host-side-scalars/README.md
mzhukova: I believe your understanding is correct, this option is only for the sake of having a complete review of anything possibly with the current API.

mzhukova:
Minor update: after a few discussions it was decided, to have both options 2b and 1 for attributes setting, with 2b being main entry point for all scales and zero-points setting, and 1 being a convenience API (so to not force user to set extra unnecessary parameters).

mzhukova (approved):

PR #3243, opened 2025-05-09
benchdnn: fix case timer for parallel creation scenario

dzarukin:
More precise timing for GPU on Linux now.

### DIFF
--- a/tests/benchdnn/common.cpp
+++ b/tests/benchdnn/common.cpp
@@ -144,9 +144,13 @@ void parse_result(res_t &res, const char *pstr) {
                 + " total:" + std::to_string(res.total) + ")";
     }
 
+    using bt = timer::timer_t;
+
     const auto &tct = res.timer_map.get_timer(timer::names::test_case_timer);
     // Round to integer for nicer input.
-    const int64_t tct_ms = static_cast<int64_t>(tct.ms());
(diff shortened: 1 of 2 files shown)


dzarukin:
make test
disable test_device_cpu

echeresh (approved):

fivemserverhostings:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemtebex:
Great implementation of this feature. I particularly like the approach to error handling.

tests/benchdnn/common.cpp
```diff
@@ -144,9 +144,14 @@ void parse_result(res_t &res, const char *pstr) {
                 + " total:" + std::to_string(res.total) + ")";
     }
 
-    const auto &tct = res.timer_map.get_timer(timer::names::test_case_timer);
+    using bt = timer::timer_t;
+    using namespace timer::names;
```
spalicki, line 148: nit: maybe let's not use it? It is not a long namespace, we are not "using" it in other files, and it is easier for new people to figure out where the parameters come from.

spalicki (approved):

tests/benchdnn/common.cpp
dzarukin: Sure, updated, thanks.

PR #3244, opened 2025-05-10
intel: ocl: rnn: modify k_limit to fix pvc regression

h-sadia:
# Description
These regressions occured as a result of changing the k_limit for fused kernels in RNN while LBR_GRU was added to the existing available algorithms.
Machine: PVC 141703
Base Commit: 34d53b4
Following graphs show that the performance was restored.
Regression graph:
image

Graph after proposed changes:
image


h-sadia:
make test perf-gpu
set primitive=rnn

atkassen (approved):

rjoursler (approved):

PR #3249, opened 2025-05-12
gpu: generic: Change generic sycl attr scales func

ShanoToni:
# Description

Adding latest changes from commit e210602  to the `attr_scales_ok` function in the generic sycl implementation.
Change is required as without it native backends and generic sycl do not compile.

### DIFF
--- a/src/gpu/generic/sycl/ref_convolution.hpp
+++ b/src/gpu/generic/sycl/ref_convolution.hpp
@@ -108,10 +108,10 @@ struct ref_convolution_fwd_t : public gpu::generic::sycl::primitive_t {
                     attr()->has_default_values(sm::scales | sm::zero_points
                             | sm::post_ops | sm::sum_dt),
                     VERBOSE_UNSUPPORTED_ATTR);
-            VDISPATCH_CONV(
-                    IMPLICATION(!attr()->scales_.has_default_values(),
-                            attr_scales_ok()
-                                    && check_convolution_scales_types(attr())),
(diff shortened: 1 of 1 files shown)


Rbiessy (approved):

src/gpu/generic/sycl/ref_convolution.hpp
```diff
@@ -110,7 +110,10 @@ struct ref_convolution_fwd_t : public gpu::generic::sycl::primitive_t {
                     VERBOSE_UNSUPPORTED_ATTR);
             VDISPATCH_CONV(
                     IMPLICATION(!attr()->scales_.has_default_values(),
-                            attr_scales_ok()
+                            attr_scales_ok({{DNNL_ARG_SRC, {0}},
+                                    {DNNL_ARG_WEIGHTS, {0}},
+                                    {DNNL_ARG_DST, {0}}})
+                                            == status::success
```
dzarukin, line 116: Sorry, my bad again. I asked if we can add the build for generic backend in PreCI to have less escapes like this.

It can be replaced as this:
```
CHECK(attr_scales_ok({{DNNL_ARG_SRC, {0}}, {DNNL_ARG_WEIGHTS, {0}}, {DNNL_ARG_DST, {0}}}));
```
Verbose checks will be inside the call, no need to have an extra one outside of it.

ShanoToni: Change made as suggested, thanks.

dzarukin (approved):

PR #3252, opened 2025-05-12
[GPU] Modify GEMM attr group to support reshape

kealan-barbieri:
# Description

Address performance regressions in batched gemm with grouped attrs. Previously disabled 3d->2d reshape to avoid correctness issues, in some cases we can instead reshape grouped dims to preserve correctness and performance. 

Added support for `per_tensor` grouped attrs with 3d->2d reshape to support OV use cases.  

Fixes # https://jira.devtools.intel.com/browse/MFDNN-13651

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

## Performance improvements

- [x] Have you submitted performance data that demonstrates performance improvements?

PR #3253, opened 2025-05-12
[GPU][rls-v3.9-pc] Modify GEMM attr group to support reshape

kealan-barbieri:
# Description

Backport of #3252 .


kealan-barbieri:
Reverted these changes in rls-v3.9-pc, so improvements are unnecessary.

PR #3255, opened 2025-05-12
xe: jit: gemm: fp8 compressed weights

petercad:
Closes MFDNN-13609. Adds support for f8 -> f16 compressed weights.


petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

dzarukin (approved):
LGTM. Casting @mgouicem for a second pair of eyes.

Simonsays095 (approved):
Do we have any tests to validate these new kernels/type combinations?

petercad:
> LGTM. Casting @mgouicem for a second pair of eyes.

Thanks for reviewing! Do you know if there was a reason `is_fpsubtype` doesn't have any support for f8/f4 types? Was this just an oversight during f8/f4 enabling?

baodii:
Hi, @petercad I can't pass `./benchdnn --matmul --reset --engine=gpu --skip-impl=ref --stag=ab --wtag=ba --dtag=ab --dt=f16:f8_e5m2:f16 --attr-fpmath=f16:true 8192x4096:4096x4096`. Could you check it for me? Since we would use this shape in our case.

petercad:
@baodii You should remove `--attr-fpmath=f16:true` -- it is only for integer->floating point decompression. But I will update the PR so it passes with `--attr-fpmath=f16:true` also.

CaoZhongZ:
Not very relevant question, if --fp-math=f16:true is dummy flag why would it affect performance a little bit?

```
perf,gpu,jit:gemm:any,,--mode=P --matmul --engine=gpu --dt=f16:f8_e5m2:f16 --stag=ab --wtag=ba --dtag=ab --attr-scales=wei:common:0.5:f16 --attr-scratchpad=user 4096x4096:4096x4096,137.439,73.9016,1.22325,112356,1.41439,97171.6
perf,gpu,jit:gemm:any,,--mode=P --matmul --engine=gpu --dt=f16:f8_e5m2:f16 --stag=ab --wtag=ba --dtag=ab --attr-scales=wei:common:0.5:f16 --attr-scratchpad=user --attr-fpmath=f16:true 4096x4096:4096x4096,137.439,57.2483,1.39038,98850.2,1.54536,88936.3
```

petercad:
> Not very relevant question, if --fp-math=f16:true is dummy flag why would it affect performance a little bit?

@CaoZhongZ -- That looks like measurement noise, since this layer is power-limited; the flag doesn't affect the generated kernel.

petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

petercad:
> Do we have any tests to validate these new kernels/type combinations?

Thanks for the reminder -- added test cases.

petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

dzarukin (approved):

src/gpu/intel/jit/gemm/selector/db/kernel.db
```diff
@@ -360,6 +362,31 @@ auto _CATALOG_ = kcatalog::toArray({
...
 {{'E', "gemm", {"O", "S", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {1, 1, 1}, "ixy"}, "sB4 sB4 aB wg 4x8 kc4 cab4 ks8 nse bo sr bk0 sm sn l4 pab", {8, (LoopType) 0, 128, {(LoopType) 128, (LoopType) 255, (LoopType) 255}, {524288, 262144, 16777216}, {524288, 262144, 16777216}, {32, 16, 8}, {4, 8, 1}, 1, (WGType) 1, 257, 32768, 0, {1, 4, 4}, {false, false, true}}, {'W', 1, {512}}},
+{{'E', "gemm", {"Q", "H", "S"}, {"A", "N", "N"}}, {-1, -1, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {1, 4, 1}, "xyz"}, "sB8 sB8 sB wg 2x1x16 akr kc8 fg 0.25 nse sr sb32 bk0 bm0 pab", {8, (LoopType) 0, 128, {(LoopType) 0, (LoopType) 1, (LoopType) 2}, {16777216, 8192, 16777216}, {8192, 8192, 16777216}, {32, 4, 8}, {2, 1, 16}, 1, (WGType) 0, 262917, 0, 1024, {32, 4, 4}, {false, false, false}}, {'W', 1, {128}}},
```
kealan-barbieri, line 365: Do we plan to continue to use explicit tags for cases with upconversion, eg. `"[QH]"` in this case?

rogerxfeng8 (approved):
Verified accuracy correct with fp8 quantized workload.  fp8 hits 1.37x fp16 TPS.

src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
```diff
@@ -581,17 +577,6 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
         return nullptr;
     });
 
-    if (fpmath_strict) {
-        if (problem_.Tb.isInt4() && !(fpmath_f16 || fpmath_bf16)) {
```
kealan-barbieri, line 585: similarly dropping this block disables `--matmul --engine=gpu --dt=f8_e4m3:u4:f8_e5m2 --stag=ab --wtag=ab --dtag=ab --attr-fpmath=strict:true 2x13:13x512_n"DLRM:0*1"`

petercad: This causes problems because it is trying to reuse HHS strategies for e.g. QHS, EHS, etc. and many of these cases will run out of registers.

petercad: @kealan-barbieri On the other hand, we can reuse e.g. `OHS` for `QHS` and `FHS` for `EHS`. In general, we should be able to reuse upconversion strategies as long as the data type is equivalent. So for the case you mention (could be either `[fH][qH]S` or `[fB][qB]S` but the first is faster on current HW), I think we should introduce some `[FH][OH]S` strategies and use them.

petercad:
make test perf-gpu
set primitive=matmul ip

PR #3259, opened 2025-05-13
Fix Wundef build host compiler issue (fixes MFDNN-13664)

dzarukin:
MFDNN-13664

For host compiler it's a workaround, doesn't add much value to what's already been enabled.

### DIFF
--- a/src/cpu/rnn/ref_rnn.hpp
+++ b/src/cpu/rnn/ref_rnn.hpp
@@ -59,9 +59,12 @@ void gates_reduction(const rnn_utils::rnn_conf_t &rnn,
         rnn_utils::cell_position_t cell_position, const gates_t *ws_gates_,
         acc_t *diff_bias_) {
 
-    // @todo block k on simd-width to enable vectorization in
+    // @TODO: block k on simd-width to enable vectorization in
     // parallel_nd path
-#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP && _OPENMP >= 201307 \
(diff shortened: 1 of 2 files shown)


spalicki (approved):

src/cpu/rnn/ref_rnn.hpp
```diff
@@ -59,9 +59,12 @@ void gates_reduction(const rnn_utils::rnn_conf_t &rnn,
         rnn_utils::cell_position_t cell_position, const gates_t *ws_gates_,
         acc_t *diff_bias_) {
 
-    // @todo block k on simd-width to enable vectorization in
+    // @TODO: block k on simd-width to enable vectorization in
     // parallel_nd path
-#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP && _OPENMP >= 201307 \
+    // Note: `defined(_OPENMP)` is needed for `icx`. Somehow `-fsycl` removes
```
densamoilov, line 64: DPC++ is a 2-pass compiler meaning that the the same source code is compiled 2 times - host and device compiler passes.
The host compiler defines `_OPENMP` while the device compiler does not. This is why we may see the warning.

SYCL RT provides a macro `__SYCL_DEVICE_ONLY__` to guard the code that is supposed to be compiled only for host or device.

densamoilov (approved):

src/cpu/rnn/ref_rnn.hpp
dzarukin: Updated the comment for better clarity, thanks.

PR #3263, opened 2025-05-14
benchdnn: inputs: graph: replace mul with sub in gated mlp test

xiang1guo:
Fix MFDNN-13430.

Rounding precision issue due to the uncertainty of elements' accumulation order inside the kernel calculation. 

Replace multiply with Subtract(Add is not enough) between 2 matmuls in the test cases. It helps to reduce the intermediate value after binary.


TaoLv (approved):

wzt1997 (approved):

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

PR #3264, opened 2025-05-14
gpu: generic: Add skip for unsupported zeropad dt

ShanoToni:
# Description

When testing on generic gpu, if `benchdnn` attempt to run zeropad with unsupported data type the test fails. 
PR introduces check for `f64` specifically as `skip_unimplemented_data_type` already checks for other datat ypes, but because zeropad does not support `f64` and the `skip_unimplemented_data_type` checks if the engine supports the data type not the primitive running the test with `f64` is reported as failure.

see MFDNN-13216

### DIFF
--- a/tests/benchdnn/zeropad/zeropad.cpp
+++ b/tests/benchdnn/zeropad/zeropad.cpp
@@ -116,6 +116,12 @@ static dnnl_status_t perf_func(
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_data_type({prb->dt}, FWD_D, res);
 
+    // check on sycl engine if dt is one of supported zeropad dt's
+    // if not supported skip
+    if (is_generic_gpu() && prb->dt == dnnl_f64) {
+        res->state = SKIPPED;
+        res->reason = skip_reason::case_not_supported;
+    }
     if (is_nvidia_gpu() || is_amd_gpu()) {
         res->state = SKIPPED;
         res->reason = skip_reason::case_not_supported;


tests/benchdnn/zeropad/zeropad.cpp
```diff
@@ -116,6 +116,12 @@ static dnnl_status_t perf_func(
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_data_type({prb->dt}, FWD_D, res);
 
+    // check on sycl engine if dt is one of supported zeropad dt's
+    // if not supported skip
+    if (is_generic_gpu() && prb->dt == dnnl_f64) {
+        res->state = SKIPPED;
+        res->reason = skip_reason::case_not_supported;
+    }
```
dzarukin, line 124: Please check `tests/benchdnn/dnnl_common.cpp:942` (function is_f64_supported) out. Looks like a better fit to me.

ShanoToni: The function `is_f64_supported` checks if the engine has f64 support, while the error arises from specifically zeropad not being implemented with f64 support. 
So during the `doit` of benchdnn zeropad reports failure from [here](https://github.com/uxlfoundation/oneDNN/blob/2398b29ba8fb07e0bf8ae1176073d57585aca541/src/common/memory_zero_pad.cpp#L278)

dzarukin: Just to double check - you say that f64 is generally supported for a generic gpu backend (like convolutions and matmuls) but there's no zeropad implementation and this is the reason it returns unimplemented? I'm asking because it's mentioned in the PR desc/any comment that f64 is a valid data type for generic gpu backend.

How does it work for Intel GPUs then? If there's a dedicated kernel for it, I'd say we better add a new f64 instance into the spot you mentioned and comment it as needed for generic gpu backend.

ShanoToni:
Changed the PR from checking and skipping `f64` zeropad in benchdnn to adding the data type to the list of `typed_zero_pad`

dzarukin (approved):

sgeor255 (approved):

PR #3269, opened 2025-05-14
benchdnn: mem_check: reduce the allowed combined memory limit

dzarukin:
Workaround for https://jira.devtools.intel.com/browse/MFDNN-13654

Reduce the number of allowed cases when it comes to CPU RAM limit.


dzarukin:
make test

rjoursler (approved):

dzarukin:
make test

avmanerikar (approved):

PR #3281, opened 2025-05-15
[GPU] Fixup GEMM strategy override

kealan-barbieri:
# Description

Use strategy values rather than driver_info which may be uninitialized. 

# Checklist

## General


Simonsays095:
In what cases are these member variables of `driver_info_` uninitialized? From what I understand, all `CommonDriverInfo` models use these variables (not 100% sure about this), and it gets initialized in the `entry_` and copied into `driver_info_` (either from the strategy or the entry) a few lines earlier in `update_driver_info`.

petercad (approved):

kealan-barbieri:
@Simonsays095 there are some cases where `kInterleave` is set but `kInterleaveChunk` is not so in `driverInfo` unroll for `K` dim actually gets set to 0 after its been initialized to the non-zero default value from the strategy, looking into whether theres a better fix within that initialization.

Simonsays095:
This is included as part of #3616.

PR #3289, opened 2025-05-19
cpu: aarch64: eltwise: Use predication & get rid of tail loop

Ryo-not-rio:
The jit_uni_eltwise code uses a SVE vectorized loop and a NEON tail loop. This can be unified by using SVE predication, making the code cleaner and shorter without introducing any regressions.

# Performance
No performance change


src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -165,52 +182,20 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
         cmp(reg_work_amount, simd_w());
         b(GE, vectorized_loop_start);
 
-        L(vectorized_loop_end);
-
         L(reminder_loop_start);
```
jondea, line 185: Could you write a few comments explaining how this remainder logic works now? Also, I don't think it makes sense to call this a remainder _loop_

Ryo-not-rio:
I've updated the PR with the benchdnn test cases used to check for regressions!

src/cpu/aarch64/jit_uni_eltwise.cpp
Ryo-not-rio: done!

jondea:
Excellent thanks, have you done correctness testing for the bwd case? It looks like the logic has non-trivially changed for bf16/fp16. If so, why was it not caught in previous testing/

Ryo-not-rio:
I didn't realize we don't support jit backwards for f16 and bf16, I've removed the unnecessary code to avoid confusion and have double checked the f32 backwards still works

jondea:
It could be that your changes actually fixed it, and we could have supported it, but it's definitely not necessary in this PR. If anything, I'd rather keep the changes isolated.

Radu2k (approved):

Issue #3266, opened 2025-05-14
get_fpmath_mode() not detecting --attr-fpmath=bf16


Ryo-not-rio:
# Summary
When running benchdnn tests, setting the flag `--attr-fpmath` does not seem to get detected when calling get_fpmath_mode(). This was discovered when running convolution tests where we expect the ACL implmentation to run for fast-math due to [this change](https://github.com/uxlfoundation/oneDNN/pull/2838/files).  This change works when the environment variable `ONEDNN_DEFAULT_FPMATH_MODE=BF16` is used so it is just a detection issue in the `get_fpmath_mode()` function.

# Version
v3.9.0

# Environment
aarch64 with ACL

# Steps to reproduce
```
ONEDNN_VERBOSE=2 OMP_NUM_THREADS=16 ./tests/benchdnn/benchdnn --conv --dir=FWD_D --attr-fpmath=bf16 mb1ic64ih200iw267oc256oh200ow267kh1kw1ph0pw0
```

# Observed behavior
Jit implmentation runs: 
```
onednn_verbose,v1,primitive,exec,cpu,convolution,jit_1x1:sve_256,forward_training,src:f32:a:blocked:aBcd8b::f0 wei:f32:a:blocked:ABcd8b8a::f0 bia:undef::undef::: dst:f32:a:blocked:aBcd8b::f0,attr-fpmath:bf16,alg:convolution_direct,mb1_ic64oc256_ih200oh200kh1sh1dh0ph0_iw267ow267kw1sw1dw0pw0,2.12402
```

# Expected behavior
ACL version runs:
```
onednn_verbose,v1,primitive,exec,cpu,convolution,gemm:acl,forward_training,src:f32:a:blocked:acdb::f0 wei:bf16:a:blocked:AcdB8a4b::f0 bia:undef::undef::: dst:f32:a:blocked:acdb::f0,attr-fpmath:bf16,alg:convolution_direct,mb1_ic64oc256_ih200oh200kh1sh1dh0ph0_iw267ow267kw1sw1dw0pw0,1.49683
```

dzarukin:
Hi @Ryo-not-rio, it seems the conclusions you've made are not quite there. According to the shared verbose line, fpmath mode is there:
```
--->> attr-fpmath:bf16 (marked with asterisk in a longer line)
[...]

PR #3290, opened 2025-05-19
cpu: aarch64: conv: fix fast math fallback

Ryo-not-rio:
When running fast math convolution, we should run ACL instead of JIT. This is a change introduced in https://github.com/uxlfoundation/oneDNN/pull/2838/files. However, this patch was not working when using --attr-fpmath as outlined in https://github.com/uxlfoundation/oneDNN/issues/3266. This is causing the nightly performance tests to fail.

performance:
~2.5x speedups for benchdnn conv cases using --attr-fpmath=bf16


Sqvid (approved):
Thank you for the fix! Approved.

Radu2k (approved):

PR #3292, opened 2025-05-19
cpu: x64: reorder: change couting mask product

jstachowintel:
[MFDNN-7333](https://jira.devtools.intel.com/browse/MFDNN-7333) [CPU] Jit reorder doesn't support compensation mask other than 1 or 3


src/cpu/x64/jit_uni_reorder.cpp
```diff
@@ -2693,17 +2693,18 @@ void jit_uni_reorder_t::reduce_compensation(char *out,
...
-    // Note: We do not need to explicitly zero-out compensation buffer, as the
-    // per_thread buffers are already zeroed out in the padded area.
-    const auto G = pd()->with_groups_ ? od.padded_dims()[0] : 1;
-    const auto N = od.padded_dims()[pd()->with_groups_ ? 1 : 0];
-    const auto GN = G * N;
+    auto masked_dim_prod = 1;
+    auto mask = pd()->prb_.compensation_mask;
+    for (int dim = 0; dim < pd()->prb_.ndims; dim++) {
+        if ((mask >> dim) & 1) { masked_dim_prod *= od.padded_dims()[dim]; }
+    }
```
jstachowintel, line 2700: Instead of using a fixed way to calculate the compensation buffer size, I now use a more flexible approach: I iterate through the compensation mask as a bitmask and calculate the product for every used dimension.

src/cpu/x64/jit_uni_reorder_utils.cpp
```diff
@@ -133,20 +133,6 @@ status_t cvt_mem_desc_to_layout_desc(const memory_desc_t &md_,
...
-    const int grp_bit = 1 << 1;
-    auto check_flag_and_mask = [&](int flag, int mask) {
-        return (dst_d.extra().flags & flag) && (mask & grp_bit);
-    };
-
-    return check_flag_and_mask(
-                   compensation_conv_s8s8, dst_d.extra().compensation_mask)
-            || check_flag_and_mask(compensation_conv_asymmetric_src,
-                    dst_d.extra().asymm_compensation_mask);
-}
```
jstachowintel, line 148: I removed this method in order to simplify the way supported masks are checked. Previously, the logic relied on an extra helper function (is_with_groups) to distinguish between different cases based on additional flags and masks in the memory descriptor. Now, the logic is more straightforward—we simply verify whether the mask belongs to a set of supported values (such as {0x1, 0x2, 0x3}).

src/cpu/x64/jit_uni_reorder_utils.cpp
```diff
@@ -349,12 +334,6 @@ status_t prb_init(prb_t &p, const memory_desc_t &imd, const memory_desc_t &omd,
                 ? om_d.extra().compensation_mask
                 : (p.req_asymmetric_comp ? om_d.extra().asymm_compensation_mask
                                          : tr::prb_t::invalid_comp_mask);
-
-        if (p.compensation_mask == tr::prb_t::asymmetric_comp_mask)
-            return unimplemented;
-
-        assert(p.compensation_mask == tr::prb_t::standard_comp_mask
-                || p.compensation_mask == tr::prb_t::comp_mask_with_groups);
```
jstachowintel, line 357: After introducing a fixed set of supported masks, these checks are no longer necessary. Additionally, we want to support the asymmetric compensation mask.

vpirogov:
Closing as stale.

PR #3293, opened 2025-05-19
xe, doc: add alignment checks

echeresh:
Related Jira: [MFDNN-13700](https://jira.devtools.intel.com/browse/MFDNN-13700).

For Intel GPUs, most non-reference kernels rely on block loads/stores which require:

- 16 bytes for block load/stores
    - I think this is Xe-LP/Xe-HP while LSC load/store in Xe-HPG+ needs only 8 bytes (@petercad please correct me)
- 64 bytes for block 2D load/store

So listing 64 byte as the minimal requirement and 128 byte as the preferred (see @petercad comment in Jira).

Checking only USM pointers - I think SYCL/OpenCL buffers and sub-buffer operations ensure that any buffer is aligned at `CL_DEVICE_MEM_BASE_ADDR_ALIGN` (or similar value in SYCL) which is at least 128 bytes.

USM pointer check is under the assumption that we have page granularity mapping between CPU/GPU addresses which looks like a reasonable assumption.


petercad:
> 16 bytes for block load/stores
> I think this is Xe-LP/Xe-HP while LSC load/store in Xe-HPG+ needs only 8 bytes (@petercad please correct me)

Pre-XeHPG requirements are 4b for load/16b for store. XeHPG and later is 4b for both load and store.

64b is the base address alignment requirement for block 2D, although for GEMM there are tricks to make it work down to 8b for XeHPC, 16b for Xe2/3.

128b is preferred alignment for latest GPUs.

petercad (approved):

echeresh:
make test
disable test_device_cpu

src/gpu/intel/sycl/interop_kernel.cpp
```diff
@@ -144,6 +144,7 @@ status_t interop_kernel_t::parallel_for(impl::stream_t &stream,
                             auto *m = utils::downcast<
                                     const xpu::sycl::usm_memory_storage_t *>(
                                     mem_storage);
+                            CHECK(check_alignment(m->usm_ptr()));
```
mgouicem, line 147: Wouldn't that prevent reference implementation from executing as well?

echeresh: Good catch, thanks. I updated to check the kernel name as well. Let's assume all kernels named with "ref_*" do not use block access and do not have any specific alignment requirements.

echeresh:
make test
disable test_device_cpu

echeresh:
make test
disable test_device_cpu

doc/programming_model/alignment.md
```diff
@@ -0,0 +1,26 @@
+Memory Alignment Requirements {#dev_guide_alignment}
```
vpirogov, line 1: (nit) Having whole section for memory alignment looks like an overkill. Maybe we could add this info to existing section like [Input Validation Notes](https://uxlfoundation.github.io/oneDNN/dev_guide_c_and_cpp_apis.html#input-validation-notes).

+@ranukund for additional insights.

echeresh: Thanks, moved this to the API page and put a second-level section. Please take a look.

echeresh:
make test
disable test_device_cpu

doc/programming_model/alignment.md
vpirogov: LGTM

echeresh:
make test
disable test_device_cpu

doc/programming_model/alignment.md
vpirogov: Agree that this is rather obscure detail that does not belong in library introduction.

ranukund (approved):
Thank you!

PR #3305, opened 2025-05-20
rfcs: asynchronous CPU runtime support

mgouicem:
[link to rendered](https://github.com/uxlfoundation/oneDNN/blob/7b95addc4d8daf594c7e0f97ba285395dd10e512/rfcs/20250520-async-threadpool-support/README.md)

@cfRod @TaoLv

### DIFF
--- a/rfcs/20250520-async-threadpool-support/README.md
+++ b/rfcs/20250520-async-threadpool-support/README.md
@@ -1,0 +1,168 @@
+# Proposal to support asynchronous threadpool runtime
+
+## Background
+
+XLA is switching to a new threapool runtime on CPU. In particular:
+- it has an asynchronous `parallel_for` implementation, meaning the
+  `parallel_for` function call returns while submitted tasks are still
+  running on the threadpool.
+- XLA will have a single threadpool for both primitive creation and
+  execution. This allows to overlap primitive creation time with
+  primitive execution. As a result, multiple primitives will be
+  submitted from a threadpool to that same threadpool, so parallelism
+  will always be nested.
+
+We have some properties that we can leverage on oneDNN side:
+- Primitive and memory object lifetimes will be maintained by thunk
+  runtime while execution happens: no need to reference count oneDNN
+  objects to extend their lifetime.
+- consecutive `parallel_for` calls happen in order (similar to sycl
+  in\_order queues). We can leverage that to implicitly synchronize
+  multiple parallel regions.
+
+## API changes
+
+### Waiting on threadpool tasks completion
+
+Currently the external threadpool interface that users need to
+implement already exposes flags, to inform oneDNN implementation of
+the threadpool implementation properties.  The only flag we expose as
+of now is the `ASYNCHRONOUS` flag, which when set, implies that the
+threadpool iface implementation has a non-blocking implementation of
+the parallel method.
+
+However, until now, when the threadpool iface implementation was
+ASYNCHRONOUS, we used to explicit synchronize inside oneDNN library,
+and we plan on removing that synchronization. With that change, we
+need to add a wait() method to the threadpool iface, so that we can
+properly handle waiting through threadpool in stream:wait() method.
+
+Additions to threadpool_iface.hpp:
+```c++
+struct threadpool_iface {
+     // Does nothing if SYNCHRONOUS, waits for all jobs for ASYNCHRONOUS 
+     virtual void wait() = 0;
+}
+```
+
+### Dependency tracking
+
+Technically, we need to expose events or some dependency tracking
+mecanism to express dependencies between different `parallel_for`
+calls. However, because thunk runtime has implicit synchronization in
+a threadpool, we can leverage that.
+
+Here there are two options:
+- (recommended) no API extension is needed. We need to document that the ASYNCHRONOUS flag implies this property
+- we expose a new `IN_ORDER` flag, and we would require it to be set when ASYNCHORNOUS is set.
+
+## Internal changes
+
+The main idea to support asynchronous execution in oneDNN
+implementation will be to tie the lifetime of temporary objects during
+excecution to the closure of the std::function object (often a lambda)
+that will be submited to the threapool. We then can transfert the
+ownership of this std::function to the threadpool implementation, that
+will maintain the std::function object alive while it is beeing
+executed.
+
+### Managing std::function lifetime
+
+Here the change can be local to src/common/dnnl_thread.hpp, where we need to:
+- when redirecting internal `parallel*` to threadpool implemenation
+  `parallel_for`, we need to stop capturing references to variables in
+  lambda, but copy those variables. In general, those are just `dim_t`
+  variables and should have no additional overhead.
+- we should move ownership of these function objects through std::move
+  semantic.
+
+Here this adds a requirement to the threadpool iface implemenation to
+properly handle the lifetime of the lambda closure as well.
+
+### Managing local variable lifetime
+
+oneDNN implemenations currently use quiete a few variables allocated
+on the stack of the main thread, and then pass those by reference to
+the lambda capture in `parallel_for`. However, when parallel_for is
+asynchronous, parallel tasks might access these variables while the
+main thread already exited the function scope (and hence freed those
+variables..
+
+Hence, oneDNN CPU implementation cannot pass stack allocated variables
+to a lambda capture by reference anymore. A few solutions are::
+  - to allocate those on the heap behind a shared_ptr, which would be
+    passed by copy to lambda capture.
+  - to directly declare those variables inside the lambda (this is for
+    example the preference way for `DEFINE_ARG_SCALES_BUFFER`). This
+    introduces duplication of initialization but when initilization is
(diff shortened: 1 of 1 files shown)


Sqvid:
cc: @theComputeKid @jondea @davsva01

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+However, until now, when the threadpool iface implementation was
+ASYNCHRONOUS, we used to explicit synchronize inside oneDNN library,
+and we plan on removing that synchronization. With that change, we
+need to add a wait() method to the threadpool iface, so that we can
+properly handle waiting through threadpool in stream:wait() method.
+
+Additions to threadpool_iface.hpp:
+```c++
+struct threadpool_iface {
```
cfRod, line 42: is it always safe to assume parallel_for() is non-blocking for async threadpools and blocking otherwise?

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+     // Does nothing if SYNCHRONOUS, waits for all jobs for ASYNCHRONOUS 
+     virtual void wait() = 0;
```
cfRod, line 44: Who is expected to call tp->wait()?

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+### No computation on main thread
+
+Main thread can no more be used to carry computation after a parallel
+region, as there is no guarentee of completion of the parallel
+tasks. In particular, running a parallel reduction in a parallel
+region, and then reduce the partial sums in the main thread is no more
+possible: that reduction of partial sums will have to be submitted to
+the threadpool with a `parallel_for` call with a single task.
+
+## Validation
```
cfRod, line 144: Will there be any tests on the framework side for this? i.e TensorFlow and XLA

rfcs/20250520-async-threadpool-support/README.md
cfRod: Will there be an impact on TensorFlow and PyTorch?

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+  execution. This allows to overlap primitive creation time with
+  primitive execution. As a result, multiple primitives will be
+  submitted from a threadpool to that same threadpool, so parallelism
+  will always be nested.
+
+We have some properties that we can leverage on oneDNN side:
+- Primitive and memory object lifetimes will be maintained by thunk
+  runtime while execution happens: no need to reference count oneDNN
+  objects to extend their lifetime.
+- consecutive `parallel_for` calls happen in order (similar to sycl
```
cfRod, line 19: In ACL, we usually chain calls to standalone post op calls after the main conv/ip call - for cases where we dont support explicit fusion https://github.com/uxlfoundation/oneDNN/blob/73ec3de65693929c1089ce225b37dff6bd71dffa/src/cpu/aarch64/acl_post_ops.cpp Wondering what impact it would have such a case?

rfcs/20250520-async-threadpool-support/README.md
mgouicem: we call it internally to implement stream->wait().
This is used to wait for work completion in our tests and examples.

rfcs/20250520-async-threadpool-support/README.md
mgouicem: Good question, the answer is it depends on the dependency between these calls.
- if internally, ACL takes the threadpool as input and does no computation on input/output data on the main thread (all actual computations go through the threadpool), then no problem with keeping these calls on the main thread.
- if internally, ACL does some computation on input/output in the main thread, then these should be submitted to the threadpool so that they are executed in-order with the previous call. I would suggest to make the submission inside ACL though, and not on oneDNN side. This allow to safely call ACL API from main thread, without ACL users guessing internal implementation details.

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+executed.
+
+### Managing std::function lifetime
+
+Here the change can be local to src/common/dnnl_thread.hpp, where we need to:
+- when redirecting internal `parallel*` to threadpool implemenation
+  `parallel_for`, we need to stop capturing references to variables in
+  lambda, but copy those variables. In general, those are just `dim_t`
+  variables and should have no additional overhead.
+- we should move ownership of these function objects through std::move
```
densamoilov, line 76: If there are two lambdas and the 1st lambda calls the 2nd lambda and the 1st lambda is submitted to `parallel_for` using `std::move` then how will the 2nd lambda be handled?

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+  - to directly declare those variables inside the lambda (this is for
+    example the preference way for `DEFINE_ARG_SCALES_BUFFER`). This
+    introduces duplication of initialization but when initilization is
+    cheap, it simplifies logic (and removes overhead of
+    synchornization
+  - to pass those stack variables by copy when they are small, or when
+    it is cheaper to copy than to initialize those, or to handle
+    parallel access to a shared ressource.
+
+### Managing execution context lifetime
```
densamoilov, line 104: This is a generic thought about handling all the lifetime requirements.

If my understanding is correct then I strongly believe that we will need to come up with a robust design that will enforce developers to follow the requirements and prohibit them from bypassing them intentionally or unintentionally. Otherwise, I'm afraid, we will have to spend a lot of time dealing with bug escapes and maintaining the code.

A few things that come to my mind about the possible design.
1. Introduce a "kernel argument" abstraction that could contain any allowed type. The abstraction could be used to pass the kernel arguments over to the kernel while handling all the requirements. Also, it can omit/add something based on the runtime type, if necessary.
2. Make it impossible to call `parallel_nd` explicitly. Instead, we could have something like an interface class that would have public interfaces taking all kernel arguments (that are currently passed to the lambda explicitly and implicitly) and validate their type. Or any other way to enforce that.

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+XLA is switching to a new threapool runtime on CPU. In particular:
+- it has an asynchronous `parallel_for` implementation, meaning the
+  `parallel_for` function call returns while submitted tasks are still
+  running on the threadpool.
+- XLA will have a single threadpool for both primitive creation and
+  execution. This allows to overlap primitive creation time with
+  primitive execution. As a result, multiple primitives will be
+  submitted from a threadpool to that same threadpool, so parallelism
+  will always be nested.
```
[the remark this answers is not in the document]
mgouicem: From oneDNN perspective, we used to explicitly disable nested parallelism and some implementation relied on that. For example, sdpa creates nested primitives, open a parallel region in the top level primitive, and assumes the nested primitives under that parallel region will run sequentially. There would no more be any such guarentee.

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+     virtual void wait() = 0;
+}
+```
+
+### Dependency tracking
+
+Technically, we need to expose events or some dependency tracking
+mecanism to express dependencies between different `parallel_for`
+calls. However, because thunk runtime has implicit synchronization in
+a threadpool, we can leverage that.
```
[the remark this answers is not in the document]
mgouicem: exactly.

rfcs/20250520-async-threadpool-support/README.md
```diff
@@ -0,0 +1,157 @@
...
+  `parallel_for` function call returns while submitted tasks are still
+  running on the threadpool.
+- XLA will have a single threadpool for both primitive creation and
+  execution. This allows to overlap primitive creation time with
+  primitive execution. As a result, multiple primitives will be
+  submitted from a threadpool to that same threadpool, so parallelism
+  will always be nested.
+
+We have some properties that we can leverage on oneDNN side:
+- Primitive and memory object lifetimes will be maintained by thunk
```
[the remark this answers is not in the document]
mgouicem: The user is responsible of primitive object lifetime.
Here XLA runtime will need to keep conv object alive for as long as con.execute is not complete.
If conv is destroyed before execute completes, the program will crash (as execute typically accesses pd() and primitive members during execution).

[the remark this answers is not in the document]
densamoilov: Ok, so we will require users to either call `wait()` or submit a "callback" to the threadpool that will do the cleanup to handle such situations?

rfcs/20250520-async-threadpool-support/README.md
densamoilov: The thing I was thinking about isn't a valid concern so we should be fine.

openmodelzooci:

vpirogov (approved):

PR #3307, opened 2025-05-21
Pmanczak/matmul goes into ref path with mask 12

pmanczak:
Enabling binary injector broadcast for per_spatial (mask 12) case. 
Requested in [MFDNN-8093](https://jira.devtools.intel.com/browse/MFDNN-8093).

### DIFF
--- a/tests/benchdnn/inputs/matmul/harness_matmul_regression_f32
+++ b/tests/benchdnn/inputs/matmul/harness_matmul_regression_f32
@@ -32,3 +32,6 @@
 # test dst strides
 --reset
 --strides=16x1::16x1 3x1:1x1_n"non_unit_dims_with_strides"
+
+--reset 
+--dt=f32 --attr-post-ops=add:f32:12 2x16x49x32:2x16x32x49_n"per_hw_binary_po"


tests/gtests/internals/test_bcast_strategy.cpp
```diff
@@ -332,6 +332,9 @@ INSTANTIATE_TEST_SUITE_P(SupportedStrategies, bcast_strategy_test_t,
                 std::make_tuple(memory::dims {2, 2, 1, 2},
                         memory::dims {2, 2, 1, 1},
                         impl::broadcasting_strategy_t::spatial),
+                std::make_tuple(memory::dims {2, 2, 2, 2},
+                        memory::dims {1, 1, 2, 2},
+                        impl::broadcasting_strategy_t::per_spatial),
```
dzarukin, line 337: Does "per_spatial" mask covers 1D and 3D spatial cases?

pmanczak: No, it is only 4D, as stated in the task. Should it be extended to include 1D and 3D?

dzarukin: "Per_spatial" implies **any** spatial. Either it should follow its name and be generic or be specialized instead to avoid confusion.

pmanczak: Since in 3D per_spatial would be the same as per_w, I decided to choose the second option and renamed it to something more specific - per_spatial_4d.

tczeszun (approved):

tests/gtests/internals/test_bcast_strategy.cpp
dzarukin: Let's keep the (bad) naming scheme and use per_hw instead implying 4D somehow, though I'm not the number of dimensions is actually checked.

PR #3309, opened 2025-05-21
ci: aarch64: add ctime comparison to benchdnn_comparison.py

gausah01:
This patch updates the nightly regression scripts to include ctime (creation time) comparison alongside execution time in benchdnn_comparison.py. It modifies bench_nightly_performance.sh and bench_pr_performance.sh to output ctime and updates the Python logic to detect regressions based on ctime.

# Description

This patch updates the nightly regression scripts to compare ctime (creation time) in addition to execution time for benchdnn runs.

## General

Previously, only execution time was considered when detecting regressions. This patch ensures creation time regressions are also caught in nightly CI.

- Updated bench_nightly_performance.sh and bench_pr_performance.sh to include %-ctime% in the perf template.
- Modified benchdnn_comparison.py to:
    - Parse and collect ctime values.
    - Compare median ctime between baseline and new runs.
    - Mark the test as failed if ctime regression exceeds 10%.

### DIFF
--- a/.github/automation/performance/bench_nightly_performance.sh
+++ b/.github/automation/performance/bench_nightly_performance.sh
@@ -26,14 +26,14 @@ PERF_TEMPLATE="--perf-template=%prb%,%-time%,%-ctime%"
 INPUTS_DIR="${SCRIPT_DIR}/inputs"
 
 TESTS=(
-        "$1 --matmul --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/matmul_nightly >> $3"
-        "$2 --matmul --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/matmul_nightly >> $4"
-        "$1 --conv --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/conv_nightly >> $3"
-        "$2 --conv --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/conv_nightly >> $4"
-        "$1 --eltwise --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/eltwise_nightly >> $3"
-        "$2 --eltwise --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/eltwise_nightly >> $4"
-        "$1 --reorder --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/reorder_nightly >> $3"
-        "$2 --reorder --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/reorder_nightly >> $4"
+        "$1 --matmul --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/matmul_nightly' >> $3"
+        "$2 --matmul --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/matmul_nightly' >> $4"
+        "$1 --conv --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/conv_nightly' >> $3"
+        "$2 --conv --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/conv_nightly' >> $4"
+        "$1 --eltwise --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/eltwise_nightly' >> $3"
+        "$2 --eltwise --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/eltwise_nightly' >> $4"
+        "$1 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $3"
+        "$2 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $4"
     )
 
 N=5
--- a/.github/automation/performance/bench_pr_performance.sh
+++ b/.github/automation/performance/bench_pr_performance.sh
@@ -25,14 +25,14 @@ PERF_TEMPLATE="--perf-template=%prb%,%-time%,%-ctime%"
 INPUTS_DIR="${SCRIPT_DIR}/inputs"
 
 TESTS=(
-        "$1 --matmul --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/matmul >> $3"
-        "$2 --matmul --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/matmul >> $4"
-        "$1 --conv --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/conv >> $3"
-        "$2 --conv --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/conv >> $4"
-        "$1 --eltwise --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/eltwise >> $3"
-        "$2 --eltwise --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/eltwise >> $4"
-        "$1 --reorder --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/reorder >> $3"
-        "$2 --reorder --mode=P $PERF_TEMPLATE --batch=$INPUTS_DIR/reorder >> $4"
+        "$1 --matmul --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/matmul' >> $3"
+        "$2 --matmul --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/matmul' >> $4"
+        "$1 --conv --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/conv' >> $3"
+        "$2 --conv --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/conv' >> $4"
+        "$1 --eltwise --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/eltwise' >> $3"
+        "$2 --eltwise --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/eltwise' >> $4"
+        "$1 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder' >> $3"
+        "$2 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder' >> $4"
     )
 
 N=5


.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        r1_samples[k].append(float(v[:-1]))
-    for k, v in r2:
-        r2_samples[k].append(float(v[:-1]))
+    r1_exec = defaultdict(list)
+    r1_ctime = defaultdict(list)
+    r2_exec = defaultdict(list)
+    r2_ctime = defaultdict(list)
+
+    for parts in r1:
+        if len(parts) == 3:
```
Ryo-not-rio, line 53: Do we need this check? The line is validated when we do `r1 = [x.split(",") for x in r1 if x[0:8] == "--mode=P"]`

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-    for k, v in r1:
-        r1_samples[k].append(float(v[:-1]))
-    for k, v in r2:
-        r2_samples[k].append(float(v[:-1]))
+    r1_exec = defaultdict(list)
+    r1_ctime = defaultdict(list)
+    r2_exec = defaultdict(list)
+    r2_ctime = defaultdict(list)
+
+    for parts in r1:
```
Ryo-not-rio, line 52: can be written as `for key, exec_time, ctime in r1`

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
             continue
 
-        r2_times = r2_samples[prb]
-
-        res = ttest_ind(r2_times, r1_times, alternative='greater')
-        r1_med = statistics.median(r1_times)
-        r2_med = statistics.median(r2_times)
-        times[prb] = (r1_med, r2_med)
-        times_str = f" {times[prb][0]} vs {times[prb][1]}"
+        res = ttest_ind(r2_exec[prb], r1_exec[prb], alternative='greater')
```
Ryo-not-rio, line 71: `r1_exec[prb]` and `r2_exec[prb]` should be made variables for cleanliness and to avoid accessing the dictionary every time

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        r1_med_exec = statistics.median(r1_exec[prb])
+        r2_med_exec = statistics.median(r2_exec[prb])
+        r1_med_ctime = statistics.median(r1_ctime[prb])
+        r2_med_ctime = statistics.median(r2_ctime[prb])
+        # times[prb] = (r1_med, r2_med)
+        # times_str = f" {times[prb][0]} vs {times[prb][1]}"
```
Ryo-not-rio, line 77: remove

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
         # both the median time and min time has not 
         # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or \
-                ((r2_med - r1_med) / r1_med < 0.1 and \
-                (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1)
-        if not passed:
-            failed_tests.append(prb + times_str)
-            passed = False
+        exec_regressed = res.pvalue <= 0.05 and ((r2_med_exec - r1_med_exec) / r1_med_exec >= 0.1)
+        ctime_regressed = ((r2_med_ctime - r1_med_ctime) / r1_med_ctime >= 0.1)
```
Ryo-not-rio, line 84: could you also add a ttest for the ctime?

Sqvid:
>Mark the test as failed if ctime regression exceeds 10%.

Maybe @Ryo-not-rio can comment on whether we should be failing for increased `ctime`? I guess it comes down to whether this is a metric we want to optimise for or just track for diagnostic purposes. (I probably lean towards the latter).

Ryo-not-rio (changes requested):
Thanks for updating the tests, I've left a few comments here and there. Also looks like `benchdnn_comparison.py` needs to go through a formatter

Ryo-not-rio:
>Maybe @Ryo-not-rio can comment on whether we should be failing for increased ctime? I guess it comes down to whether this is a metric we want to optimise for or just track for diagnostic purposes. (I probably lean towards the latter).

I think we can leave it as an error for now, since we want to catch unintentional regressions on ctime as well. If it so happens that we introduce an intentional regression, then we can disable ctime for that specific test later

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,39 +44,50 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        times_str = f" {times[prb][0]} vs {times[prb][1]}"
+        res = ttest_ind(r2_exec[prb], r1_exec[prb], alternative='greater')
+        r1_med_exec = statistics.median(r1_exec[prb])
+        r2_med_exec = statistics.median(r2_exec[prb])
+        r1_med_ctime = statistics.median(r1_ctime[prb])
+        r2_med_ctime = statistics.median(r2_ctime[prb])
+        # times[prb] = (r1_med, r2_med)
+        # times_str = f" {times[prb][0]} vs {times[prb][1]}"
 
         # pass the test if:
```
almayne, line 79: This comment could do with an update. You are no longer checking the min time and you've flipped the logic ( > or < has become <= and >=).
Also, you have a new and different test being applied for the ctime. I assume not checking the p value is intentional?

.github/automation/performance/benchdnn_comparison.py
gausah-arm: Thanks for the suggestion! I initially tried using a t-test for ctime, but since the number of samples is small and the variance is low, the t-test often returns a high p-value which as far as i understood (is not significant) and just ignores the regressions even when there’s a large regression. I have tested this by manually adding regressions and increase ctime by ~250% That’s why I went with a simpler threshold-based check for ctime instead — it helps reliably catch regressions like a 10% increase in creation time. Let me know if you'd still prefer the t-test added alongside it.

.github/automation/performance/benchdnn_comparison.py
gausah-arm: Done

.github/automation/performance/benchdnn_comparison.py
gausah-arm: Done

.github/automation/performance/benchdnn_comparison.py
gausah-arm: Done

.github/automation/performance/benchdnn_comparison.py
gausah-arm: Thanks for the suggestion — I’ve updated the comment to reflect the current logic. It now describes the regression conditions for both execution time and creation time accurately.

Sqvid:
also nitpick but the commit message scope should be "ci: aarch64: ...".
"perf: ..." would be for performance improvements

gausah-arm:
> also nitpick but the commit message scope should be "ci: aarch64: ...". "perf: ..." would be for performance improvements

Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,52 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        r1_med_ctime = statistics.median(ctime1)
+        r2_med_ctime = statistics.median(ctime2)
+
+        # A test fails if either:
+        # - Execution time shows statistically significant regression (t-test p ≤ 0.05 and ≥ 10% slowdown)
+        # - Creation time shows ≥ 10% slowdown in median value
+        exec_regressed = res.pvalue <= 0.05 and ((r2_med_exec - r1_med_exec) / r1_med_exec >= 0.1)
+        ctime_regressed = ((r2_med_ctime - r1_med_ctime) / r1_med_ctime >= 0.1)
+
+        if r1_med_exec == 0 or min(exec1) == 0:
```
Ryo-not-rio, line 82: this needs to be before any of the division calculations happen

gausah-arm: Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,54 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        # both the median time and min time has not
-        # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or (
-            (r2_med - r1_med) / r1_med < 0.1
-            and (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1
+        # A test fails if either:
+        # - Execution time shows statistically significant regression (t-test p ≤ 0.05 and ≥ 10% slowdown)
+        # - Creation time shows ≥ 10% slowdown in median value
+        exec_regressed = res.pvalue <= 0.05 and (
+            (r2_med_exec - r1_med_exec) / r1_med_exec >= 0.1
```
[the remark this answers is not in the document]
gausah-arm: I was calculating regression on the basis of median values for ctime and hence changed for exec_time. But now I have added the min values and updated the checks for both execution time and ctime.

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        r2_ctime[key].append(float(ctime))
 
     failed_tests = []
     times = {}
-    for prb, r1_times in r1_samples.items():
-        if prb not in r2_samples:
+    for prb in r1_exec:
+        if prb not in r2_exec:
             warnings.warn(f"{prb} exists in {file1} but not {file2}")
             continue
```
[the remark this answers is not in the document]
Sqvid: Yeah I am assuming that if this check fails something is likely wrong right?

.github/automation/performance/benchdnn_comparison.py
Ryo-not-rio: Not resolved yet, how did you test this? I tried with a ttest locally and it detected regressions as expected

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+    for key, exec_time, ctime in r1:
+        r1_exec[key].append(float(exec_time))
+        r1_ctime[key].append(float(ctime))
+
+    for key, exec_time, ctime in r2:
+        r2_exec[key].append(float(exec_time))
+        r2_ctime[key].append(float(ctime))
 
     failed_tests = []
     times = {}
```
Ryo-not-rio, line 61: remove, no longer used

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+            and (min(ctime2) - min(ctime1)) / min(ctime1) < 0.1
         )
-        if not passed:
-            failed_tests.append(prb + times_str)
-            passed = False
+
+        if exec_regressed or ctime_regressed:
+            failed_tests.append(
+                f"{prb} exec: {r1_med_exec:.4f} → {r2_med_exec:.4f}, "
+                f"ctime: {r1_med_ctime:.5f} → {r2_med_ctime:.5f}"
```
Ryo-not-rio, line 106: Would be great if we could print to three significant figures instead of rounding to decimal places

.github/automation/performance/benchdnn_comparison.py
gausah01: I have now added a ttest for ctime alongside the existing threshold-based checks. To verify this I manually added some regressions in new.txt by changing the ctime for 4-5 reorders which was increased by 20-25% compared to base.txt, ensuring both median and min increased by more than 10%. The ttest returned a p-values<=0.05 and the regression was detected by the script

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        r2_times = r2_samples[prb]
-
-        res = ttest_ind(r2_times, r1_times, alternative="greater")
-        r1_med = statistics.median(r1_times)
-        r2_med = statistics.median(r2_times)
-        times[prb] = (r1_med, r2_med)
-        times_str = f" {times[prb][0]} vs {times[prb][1]}"
-
-        if r1_med == 0 or min(r1_times) == 0:
+        if r1_med_ctime == 0 or min(ctime1) == 0:
```
Ryo-not-rio: Please make this one if statement with the exec time

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
 
-        # pass the test if:
-        # the t-test passes (i.e. pvalue > 0.05) OR
-        # both the median time and min time has not
-        # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or (
-            (r2_med - r1_med) / r1_med < 0.1
-            and (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1
+        # A test fails if either:
+        # - Execution time shows statistically significant regression (t-test p ≤ 0.05 and ≥ 10% slowdown in median and min value)
```
Ryo-not-rio: please run the code formatter. This line is too long

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,67 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        # - Creation time shows ≥ 10% slowdown in median and min value
```
Ryo-not-rio: This comment needs updating

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,70 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        # - Creation time shows a statistically significant regression:
+        #   p ≤ 0.05 and ≥ 10% slowdown in both median and min creation times
+        exec_regressed = res.pvalue <= 0.05 and (
+            (r2_med_exec - r1_med_exec) / r1_med_exec >= 0.1
+            and (min(exec2) - min(exec1)) / min(exec1) >= 0.1
         )
-        if not passed:
-            failed_tests.append(prb + times_str)
-            passed = False
+        ctime_regressed = (
```
Ryo-not-rio: nit: unnecessary bracket

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,70 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        # both the median time and min time has not
-        # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or (
-            (r2_med - r1_med) / r1_med < 0.1
-            and (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1
+        # A test fails if either:
+        # - Execution time shows a statistically significant regression:
+        #   p ≤ 0.05 and ≥ 10% slowdown in both median and min execution times
+        # - Creation time shows a statistically significant regression:
+        #   p ≤ 0.05 and ≥ 10% slowdown in both median and min creation times
```
Ryo-not-rio: nit: we can make this concise just by saying:
```
# A test fails if either execution time or creation time:
# - shows a statistically significant regression and
# - shows ≥ 10% slowdown in both median and min times
```

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,70 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        ctime1 = r1_ctime[prb]
+        ctime2 = r2_ctime[prb]
+        res = ttest_ind(exec2, exec1, alternative="greater")
+        ctime_test = ttest_ind(ctime2, ctime1, alternative="greater")
+        r1_med_exec = statistics.median(exec1)
+        r2_med_exec = statistics.median(exec2)
+        r1_med_ctime = statistics.median(ctime1)
+        r2_med_ctime = statistics.median(ctime2)
+
+        if (
```
Ryo-not-rio: nit:
```
if 0 in [r1_med_exec, min(exec1), r1_med_ctime, min(ctime1): 
```

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,62 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or (
-            (r2_med - r1_med) / r1_med < 0.1
-            and (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1
+        # A test fails if either execution time or creation time:
+        # - shows a statistically significant regression and
+        # - shows ≥ 10% slowdown in both median and min times
+        exec_regressed = res.pvalue <= 0.05 and (
+            (r2_med_exec - r1_med_exec) / r1_med_exec >= 0.1
+            and (min(exec2) - min(exec1)) / min(exec1) >= 0.1
```
Ryo-not-rio: this should be an `or`

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,62 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
         )
-        if not passed:
-            failed_tests.append(prb + times_str)
-            passed = False
+        ctime_regressed = ctime_test.pvalue <= 0.05 and (
+            (r2_med_ctime - r1_med_ctime) / r1_med_ctime >= 0.1
+            and (min(ctime2) - min(ctime1)) / min(ctime1) >= 0.1
```
Ryo-not-rio: this should be an `or`

gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/bench_nightly_performance.sh
```diff
@@ -24,14 +24,14 @@ IFS=$'\n' # Prevents shuffling from using spaces as delimiters
...
 TESTS=(
-        "$1 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul_nightly >> $3"
-        "$2 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul_nightly >> $4"
-        "$1 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv_nightly >> $3"
-        "$2 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv_nightly >> $4"
-        "$1 --eltwise --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/eltwise_nightly >> $3"
-        "$2 --eltwise --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/eltwise_nightly >> $4"
-        "$1 --reorder --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/reorder_nightly >> $3"
-        "$2 --reorder --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/reorder_nightly >> $4"
+        "$1 --matmul --mode=P --perf-template=%prb%,%-time%,%-ctime% --batch=${SCRIPT_DIR}/inputs/matmul_nightly >> $3"
```
Sqvid: Can we please save `perf-template` into a string. Typing it out every time seems unnecessarily error prone.

E.g if you accidentally typed `--perf-template=%prb%,%time%,%-ctime%` anywhere then the meaning would change.

.github/automation/performance/bench_pr_performance.sh
```diff
@@ -23,14 +23,14 @@ IFS=$'\n' # Prevents shuffling from using spaces as delimiters
...
-        "$1 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul >> $3"
-        "$2 --matmul --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/matmul >> $4"
-        "$1 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv >> $3"
-        "$2 --conv --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/conv >> $4"
-        "$1 --eltwise --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/eltwise >> $3"
-        "$2 --eltwise --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/eltwise >> $4"
-        "$1 --reorder --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/reorder >> $3"
-        "$2 --reorder --mode=P --perf-template=%prb%,%-time% --batch=${SCRIPT_DIR}/inputs/reorder >> $4"
+        "$1 --matmul --mode=P --perf-template=%prb%,%-time%,%-ctime% --batch=${SCRIPT_DIR}/inputs/matmul >> $3"
+        "$2 --matmul --mode=P --perf-template=%prb%,%-time%,%-ctime% --batch=${SCRIPT_DIR}/inputs/matmul >> $4"
```
Sqvid: Same as earlier

Sqvid (changes requested):

.github/automation/performance/bench_nightly_performance.sh
Sqvid: nit: I would probably also define

```sh
INPUTS_DIR="${SCRIPTS_DIR}/inputs"
```

and then use that in these locations since that seems to be the most common use of `SCRIPTS_DIR` anyway.

Sqvid (changes requested):

.github/automation/performance/bench_nightly_performance.sh
gausah01: Done

.github/automation/performance/bench_pr_performance.sh
gausah01: Done

.github/automation/performance/bench_nightly_performance.sh
gausah01: Done

Sqvid: You need single quotes for the value of `--batch` or it will break if `SCRIPT_DIR` has spaces in it (which is allowed).

E.g. `"$1 --matmul --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/matmul_nightly' >> $3"`

Sqvid: nitpick (meaning I will approve it even if you don't change this), but:

```sh
PERF_TEMPLATE=${PERF_TEMPLATE:-"%prb%,%-time%,%-ctime%"}

# and then modify so that:
... --perf-template='$PERF_TEMPLATE' ...
```

Then we can override PERF_TEMPLATE at runtime using an env variable

gausah01: Done

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -44,46 +44,62 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-        # pass the test if:
-        # the t-test passes (i.e. pvalue > 0.05) OR
-        # both the median time and min time has not
-        # slowed down by more than 10%
-        passed = res.pvalue > 0.05 or (
-            (r2_med - r1_med) / r1_med < 0.1
-            and (min(r2_times) - min(r1_times)) / min(r1_times) < 0.1
+        # A test fails if either execution time or creation time:
+        # - shows a statistically significant regression and
+        # - shows ≥ 10% slowdown in both median or min times
```
Ryo-not-rio, line 86: This comment needs to be updated to `shows ≥ 10% slowdown in either median or min times`

Ryo-not-rio: We'll update this in a different patch since it's such a small fix

PR #3310, opened 2025-05-21
cpu: aarch64: Revert "cpu: aarch64: Nuke winograd"

Ryo-not-rio:
This reverts commit ef1365bfea776be2cf685e1210a87c180f3f6cb3.

Keep Winograd as OpenVINO uses `convolution_auto`

### DIFF
--- a/doc/primitives/convolution.md
+++ b/doc/primitives/convolution.md
@@ -395,7 +395,8 @@ fall back to an explicit GEMM algorithm.
 @anchor dg_winograd_conv
 ### Winograd Convolution
 
-oneDNN supports the Winograd convolution algorithm on GPU systems.
+oneDNN supports the Winograd convolution algorithm on GPU and AArch64 CPU systems.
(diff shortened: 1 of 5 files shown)


Sqvid:
Let's keep this as a draft for the time being.

@alvoron any comments from your end on how hot the dnnl-winograd path is would be very helpful. I see it is theoretically an option since [algorithm is set to convolution_auto](https://github.com/openvinotoolkit/openvino/blob/f3f8d8efea602abd95c09c3bcf59bceaf34f526d/src/plugins/intel_cpu/src/nodes/executors/dnnl/dnnl_convolution_primitive.cpp#L236), but maybe you can provide some insight as to whether this is the primary path on AArch64. Thank you.

Sqvid:
This was a clean revert right? Or were any manual changes required?

Sqvid:
info: reverted PR is here https://github.com/uxlfoundation/oneDNN/pull/3139

PR #3317, opened 2025-05-22
synthdnn: generalize type interface

Simonsays095:
Improves the `-t` option for synthdnn's matmul driver, to use globbing-style wildcards matching in addition to index matches. This replaces the following syntax:
- `all` --> `*` to match any supported type
- `*` --> `%0` to match the first supplied type

And it comes with the following upgrades:
- `%N` matches the Nth type (currently only `%0` used as a shortcut in the support matrix)
- Any type can be given instead of all 3 types. This will match any supported configuration using that type
- Each type can match using globbing-style wildcards. So `[su]4` will match `s4` and `u4` types, and `f*` will match all floating-point types except `bf16`, for example.

These changes make it easier to target similar types in synthetic test sets. For example, `?f16` will match `bf16` and `f16` types, and `[su]4` will match int4 types, as mentioned earlier.

This PR does not affect problem set generation, only the syntax of the interface used to determine which problems are considered.


scripts/synthdnn/synthdnn.py
```diff
@@ -165,7 +165,7 @@ def setup_matmul_subparser(subparsers):
         "-t",
         "--types",
         default="all",
-        help='dt:dt:dt(optional fpmath-mode), comma separated list of type configurations or "all" for every supported type',
+        help='dt:dt:dt(optional fpmath-mode), comma separated list of type configurations. "%%N" will match the Nth given data type. Giving a single data type instead of 3 will match any supported configurations using that type.',
```
dzarukin, line 168: It seems to me a documentation page on the syntax should be updated/added as well. It would be good to explain new capabilities with examples of command line syntax and the outcome.

skazakov1 (approved):

scripts/synthdnn/synthdnn.py
Simonsays095: I updated the readme with a more verbose description of the options, as well as a few examples. I also removed a shortcut syntax we had (generation and collection in one command) to simplify the syntax. Let me know if you'd like any changes to the readme, please.

dzarukin: Thanks for taking care of it. Definitely looks better.

dzarukin (approved):

PR #3320, opened 2025-05-22
synthdnn: add run collection type

skazakov1:
The synthdnn is going to be used to generate verbose log for 1perf. "run" collection type makes the log simpler and shorter, without any kind of testing.

### DIFF
--- a/scripts/synthdnn/synthdnn.py
+++ b/scripts/synthdnn/synthdnn.py
@@ -73,7 +73,7 @@ def setup_collect_args(parser, req):
     parser.add_argument(
         "--collect",
         default="corr",
-        help="benchdnn collection type, can be one of [corr, perf]",
+        help="benchdnn collection type, can be one of [corr, perf, run]",
     )
     parser.add_argument("-n", "--name", default="", help="sample name")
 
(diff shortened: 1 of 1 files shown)


scripts/synthdnn/synthdnn.py
```diff
@@ -73,7 +73,7 @@ def setup_collect_args(parser, req):
     parser.add_argument(
         "--collect",
         default="corr",
-        help="benchdnn collection type, can be one of [corr, perf]",
+        help="benchdnn collection type, can be one of [corr, perf, run]",
```
dzarukin, line 76: Would be nice to see a doc page explaining the benefits of run.

scripts/synthdnn/synthdnn.py
```diff
@@ -113,6 +113,8 @@ def collect_main(args):
 
     if args.collect == "corr":
         benchdnn_args = f"--engine={args.engine} --matmul --mode-modifier=P {get_optional_args(args)}"
+    elif args.collect == "run":
+        benchdnn_args = f"--engine={args.engine} --matmul --mode=R --mode-modifier=P {get_optional_args(args)}"
     elif args.collect == "perf":
         benchdnn_args = f"--engine={args.engine} --matmul --mode=F --cold-cache=all --perf-template=sample,{args.name},%prb%,%0Gflops%,%0Gbw% --memory-kind=usm_device --attr-scratchpad=user {get_optional_args(args)}"
```
rjoursler, line 119: Collect is mostly a wrapper around benchdnn. It's core purpose is to set a standardized output data format (i.e. perf-template) for processing by later scripts. Because of this, it likely makes more sense just forward non-format related command line arguments to benchdnn, rather than maintain a second interface.

PR #3324, opened 2025-05-23
gpu: nvidia: Add check for deconv bias mixed dt

ShanoToni:
# Description
`cudnnConvolutionBackwardBias` used in deconv does not support mixed datatypes causing benchdnn to fail for:
```
--deconv --engine=gpu --dir=BWD_WB --dt=f16:f16:f16
``` 
with `CUDNN_STATUS_BAD_PARAM`
Adding a check to ensure datatypes match.


ShanoToni:
make test
disable build_vendor_intel
disable test_device_cpu
enable arch_gpu_ampere
enable compiler_icx-oss

Rbiessy:
Question out of curiosity, how come using `--dt=f16:f16:f16` ends up testing configurations with mixed datatypes?

ShanoToni:
> Question out of curiosity, how come using `--dt=f16:f16:f16` ends up testing configurations with mixed datatypes?
Fair question here `dt` represents src, weights and dst. bias dt would be specified with `--bia-dt={undef [default], f32, bf16, f16, ...}`
Since we do not pass it, it defaults to f32

PR #2931, opened 2025-03-21
graph: refine and support GQA pattern with ukernel

xiang1guo:
## Description
Follow up PR based on #2930. Part 2 for addressing MFDNN-12871 which is requested by PyTorch upstream. This PR refined the legacy GQA pattern and added a pass to handle new pattern. Now, the GQA pattern and graph compilation fusion processed pattern as follows.
GQA pattern that graph API support. 
Q[batch size, group num, group size, seq len, head size]
K[batch size, group num, 1, seq len, head size]
V[batch size, group num, 1, seq len, head size]
image

The processed GQA pattern after graph compilation.
image

## Works
- Added a pass to transform 5D GQA into 4D so that the ukernel can support it.
- Added some check and layout propagation to avoid unnecessary reorder.

## Validation
```
onednn_verbose,v1,graph,exec,gpu,100002,sdp,bmm1;scale_div;mask_add;softmax;bmm2,,in0_f16:1:strided:undef:32x2x8x384x64:393216s196608s24576s64s1 in1_f16:3:strided:undef:32x2x1x384x64:49152s24576s24576s64s1 in2_f16:8:strided:undef:1:1 in3_f16:11:strided:undef:32x1x1x1x384:384s384s384s384s1 in4_f16:19:strided:undef:32x2x1x384x64:49152s24576s24576s64s1 out0_f16:20:strided:undef:32x2x8x384x64:393216s196608s24576s64s1,fpm:strict,sdp_primitive_v1_kernel_t,dnnl_backend,1.71313
0:PASSED __REPRO: --graph --engine=gpu --dt=f16 --case=/home/gta/xiangguo/oneDNN/tests/benchdnn/inputs/graph/complex_fusion/mha/GQA-fp16-v2.json
tests:1 passed:1 skipped:0 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total: 7.64s; fill: 0.00s (0%); compute_ref: 0.00s (0%); compare: 0.00s (0%);
```

PyTorch upstream team also helped to verified the new design that the new solution works fine.

PR #2930, opened 2025-03-21
graph: introduce internal dnnl_sdpa op

xiang1guo:
## Description
- This is the first part to support GQA pattern refinement requested by PyTorch upstream. Address MFDNN-12871. The PR added a new internal dnnl_sdpa op for better fusion and alignment with sdpa primitive. With this, the internal compilation will transform a subgraph into sdpa op, see following pic.
- The refactor also aims to reduce graph compilation time by reducing layout propagation and memory planning time with a simplified internal sdpa op.

**Compiled graph before this PR**

image

**Compiled graph after this PR**

image

## Works
- Added a dnnl_sdpa op. Currently only support float SDPA.
- Added a new sdp_primitive_v1 kernel for simplicity. The final goal is to merge this kernel with sdp_primitive kernel.
- Move the sdpa primitive ukernel creation process into op_executable which is now aligned with other kernels

## Follow-up
There will be another PR to refine the GQA pattern based on this new internal dnnl_sdpa.

### Correctness check
There are total 66 case can be supported by GPU ukernel, those 42 float SDPA cases can run into sdp_primitive_v1_kernel_t now, the other 24 cases are quantization SDPA which all run into sdp_primitive_kernel_t .
Kernel | PR | main branch
-- | -- | --
sdp_primitive_v1_kernel_t | 42 | 0
sdp_primitive_kernel_t | 24 | 66
larger_partition_kernel_t | 76 | 76

### Performance test
- SDPA partition level performance has comparable performance with main branch.
- Graph compilation time of v1 kernel(exclude primitive creation time) got 2x speedup compared to legacy sdp_primitive_kernel.

PR #3334, opened 2025-05-27
graph: backend: dnnl: sdpa v1 kernel support quantize SDPA and backend refactor

xiang1guo:
### Background
This is a follow-up work based on #2930 and #2931. The PR mainly focuses on supporting quantized SDPA with internal dnnl_sdpa. It helps to reduce graph compilation time and also simplifies the backend optimization pass. 
### Works
- [x] Support compressed SDPA with internal dnnl_sdpa.
- [x] Support legacy GQA pattern with internal dnnl_sdpa.
- [x] Merge `sdp_primitive_kernel_t` and `sdp_primitive_v1_kernel_t`
### TODO
- Support CPU decompose kernel with internal dnnl_sdpa.
### Fused graph
- compressed SDPA dot graph:
**before fusion**
image
**after fusion**
image

- legacy GQA dot graph 
**before fusion**
image
**after fusion**
image


xiang1guo:
We have alternative solution in #3423

PR #3335, opened 2025-05-27
build: add call to find_package(OpenCL) in config.cmake.in

AD2605:
# Description

If building oneDNN, [this line](https://github.com/uxlfoundation/oneDNN/blob/743e4273d18bcb0e733723bf9289f14e8366687f/cmake/SYCL.cmake#L127) is hit during it's CMake configuration, it adds `OpenCL::OpenCL` as a target to be available in the `dnnl-targets.cmake` which is subsequently installed. 

Applications which use such a build of oneDNN then require a `find_package(OpenCL)` for the target to be exposed which otherwise leads to CMake errors. This dependency is something which is ideally handled by oneDNN, and this PR adds the support for the same.


Rbiessy (approved):

vpirogov:
make test
test_scope=NIGHTLY
disable test_device_cpu
disable test_device_gpu

AD2605:
It seems the clang-tidy check in the CI timed out

vpirogov (approved):

PR #3352, opened 2025-05-28
gpu: cuda: Fix matmul queue recording

EwanC:
When `sycl_ext_codeplay_enqueue_native_command` commands are used as part of sycl queue recording to a graph, then their usage should to be updated to put the underlying CUDA stream into recording mode to enable correct behavior. See https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_codeplay_enqueue_native_command.asciidoc#cuda-stream-record-native-graph-nodes

This patch updates the `sycl_ext_codeplay_enqueue_native_command()` callsites to that. Additionally, an `after_exec_hook()` implementation is added to the NVIDIA stream class to reset events after execution. Otherwise this results in an error when recording future oneDNN primitive, executions as they try to add a dependency on the previous events in this list, which is an error as the event represents a node in another graph.

With these fixes when building llama.cpp with oneDNN for the NVIDIA SYCL backend the `./bin/test-backend-ops -b SYCL0 -o MUL_MAT` tests now all complete successfully.

This patch is specific to `MUL_MAT` invocations of `compat::host_task`, rather than a generic `compat::host_task` solution because CUDA stream capture needs to be done after cuDNN and cuBLAS handles are created, and for the same native stream. Not every `compat::host_task` usage across the operations use the same patterns for these restrictions, so rather than making a large scope change that will need tested with many different `benchdnn` invocations, instead I've kept the scope of this PR to `MUL_MAT` as the critical operation that enables llama.cpp backend-ops testing with graphs, and can be verified with `benchdnn --matmul --execution-mode=graph`. Then follow-up PRs can be made for the other operations which can be verified with the respective benchdnn invocations.

With a DPC++ build after https://github.com/intel/llvm/pull/19091 using CUDA 12.9. I see the same results with `./benchdnn --execution-mode=graph  --engine=gpu --matmul --batch=inputs/matmul/test_matmul_ci` as without graph execution mode. For older DPC++ builds, or those with CUDA < 12.9, the operations using CUDA async malloc as part of the implementation don't pass. Note that I wasn't able to stress the `cudnn_matmul_lt_t` path, so i've not updated that `compat::host_task` usage as I didn't want to touch code I can't test the effects of, but I could do if maintainers desired.

### DIFF
--- a/src/gpu/nvidia/cudnn_matmul.cpp
+++ b/src/gpu/nvidia/cudnn_matmul.cpp
@@ -48,9 +48,12 @@ status_t cudnn_matmul_t::execute(const exec_ctx_t &ctx) const {
             matmul_impl_, pd()->params_, src_d, weights_d, dst_d, bias_d);
 
     if (pd()->params_->has_runtime_params_) {
-        auto &evts = cuda_stream->sycl_ctx().get_sycl_deps().events;
-        for (auto e : evts) {
-            e.wait();
+        if (cuda_stream->queue().ext_oneapi_get_state()
+                == sycl::ext::oneapi::experimental::queue_state::executing) {
+            auto &evts = cuda_stream->sycl_ctx().get_sycl_deps().events;
+            for (auto e : evts) {
+                e.wait();
+            }
         }
     }
     return status;
--- a/src/gpu/nvidia/cudnn_matmul_executor.hpp
+++ b/src/gpu/nvidia/cudnn_matmul_executor.hpp
@@ -79,6 +79,8 @@ protected:
                     auto cudnn_handle
                             = cuda_stream->get_cudnn_handle(native_stream);
 
+                    cuda_stream->begin_recording_if_graph(ih, native_stream);
+
                     void *reorder_scratch
                             = arg_bias_scratch.get_native_pointer(ih);
                     void *bias = arg_bias.get_native_pointer(ih);
@@ -93,6 +95,8 @@ protected:
                     matmul_impl_->execute(cublas_handle, cudnn_handle, params,
                             weights, src, dst, bias, reorder_scratch, src_scale,
                             wei_scale, dst_scale);
+
+                    cuda_stream->end_recording_if_graph(ih, native_stream);
 
                     if (params->has_runtime_params_) {
                         sync_device();
--- a/src/gpu/nvidia/stream.cpp
+++ b/src/gpu/nvidia/stream.cpp
@@ -42,6 +42,54 @@ cudnnHandle_t &stream_t::get_cudnn_handle(CUstream cuda_stream) {
     e->activate_stream_cudnn(cuda_stream);
     return *(e->get_cudnn_handle());
 }
+
+// If interop_handle is being submitted to a SYCL queue that is recording to a
+// graph, then put cuda stream into capture mode.
+void stream_t::begin_recording_if_graph(
+        const ::sycl::interop_handle &ih, CUstream cuda_stream) {
+#if SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND >= 2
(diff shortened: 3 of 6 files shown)


ShanoToni:
Some thoughts on the TODO
- To generalise the graph recording Id recommend [Here](https://github.com/uxlfoundation/oneDNN/blob/faf928e7c94973f1fa8e3b264228c766967fe169/src/gpu/nvidia/sycl_cuda_compat.hpp#L39) making the following change.
```cpp
template <typename HandlerT, typename FnT>
void host_task(HandlerT &cgh, const FnT &task) {
#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
    cgh.ext_codeplay_enqueue_native_command(task);
#else
    cgh.host_task(task);
#endif
}
```
to
```cpp
template <typename HandlerT, typename FnT>
void host_task(nvidia::stream_t *cuda_stream, HandlerT &cgh, const FnT &task) {
    auto native_stream = cuda_stream->get_underlying_stream();
    cuda_stream->begin_recording_if_graph(cgh, native_stream);
#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
    cgh.ext_codeplay_enqueue_native_command(task);
#else
    cgh.host_task(task);
#endif
    cuda_stream->end_recording_if_graph(cgh, native_stream);
}
```
If this change is possible and works, it will affect any operator which calls the `compat::host_task` and the only operator specific change would be passing the `cuda_stream` to the `host_task` function itself.

- For regression tests I am not sure how that would work, I would recommend looking into the `/test/gtest/sycl/api` I believe that would be the correct approach.

Rbiessy:
One more comment about the regression tests, you may want to try using `benchdnn`, it looks like it has a flag to use SYCL-Graph `--execution-mode=graph`. Maybe that would be enough to test everything you need. I'm not too familiar with it but I happen to see PRs related to it like: https://github.com/uxlfoundation/oneDNN/pull/3340

Rbiessy (commented):
LGTM!
Note the commits need to be squashed to pass the GitHubCI.
Were you able to confirm the benchdnn tests work fine with the graph execution mode?
There are some expected failures for Nvidia devices so I think you would need to compare the number of failures with and without your changes.
The benchdnn command should be `./tests/benchdnn/benchdnn --matmul --execution-mode=graph --engine=gpu --batch=tests/benchdnn/inputs/matmul/test_matmul_ci` and same for every operations listed by `benchdnn --help`. You'll need to drop `--execution-mode=graph` without your changes.

EwanC:
```quote
Were you able to confirm the benchdnn tests work fine with the graph execution mode?
There are some expected failures for Nvidia devices so I think you would need to compare the number of failures with and without your changes.
The benchdnn command should be ./tests/benchdnn/benchdnn --matmul --execution-mode=graph --engine=gpu --batch=tests/benchdnn/inputs/matmul/test_matmul_ci and same for every operations listed by benchdnn --help. You'll need to drop --execution-mode=graph without your changes.
```

The benchdnn tesing was a good shout, I tried it and definitely stressed the graph path. So I think is a better way to verifying things that adding large amounts of new graph specific tests to `test/gtest/sycl/api`.

By running it I actually found a new interesting fail where as part of a `sycl_ext_codeplay_enqueue_native_command` call oneDNN is trying to add a `cudaGraphAddMemAllocNode`, which doesn't work with our DPC++ Implementation. I want to look into possible DPC++ fixes, and also see if there's any other fails it's hiding, before taking this PR out of draft.

Rbiessy:
@uxlfoundation/onednn-arch I mentioned this PR in our last sync, please let us know if you're keen to merge it.

vpirogov:
Closing as stale.

PR #3354, opened 2025-05-28
cpu: x64: matmul: correct LDD when M == 1

xuxinzen:
Fixes MFDNN-13737


xuxinzen:
make test
set test_scope=NIGHTLY
disable test_device_gpu

ankalinin (approved):

PR #3363, opened 2025-05-29
cpu: aarch64: extend brgemm conv to sve_128

jondea:
Naively extend to `sve_128`, with ~2x speedups over `gemm:ref` in most cases. In practice, this PR will just speed up `BWD_D` because if oneDNN is built with Arm compute Library (ACL), the `FWD_D` numbers will be unchanged because ACL is higher in the impl list.

There are almost certainly more optimizations to be made, specifically the heuristics need tweaking, we skip any `brg_blocking_t` which have `oc_block < 16` because they always seem slower even when `est_eff` suggests otherwise.

Benchmarks when built without ACL (to demonstrate speedups over `gemm:ref` for `FWD_D`). Shapes are taken from PyTorch ResNet50 training example
```
                                        brg conv / gemm:ref (<1 means brg conv is faster)
version                                    build
threads                                       1     8     16    32    64
dir   desc
BWD_D mb128ic1024ih14oc256oh14kh1ph0         .37   .40   .47   .50   .50
      mb128ic1024ih14oc512oh14kh1ph0         .37   .41   .48   .52   .51
      mb128ic128ih28oc128oh28kh3ph1          .55   .60   .69   .67   .65
      mb128ic128ih28oc512oh28kh1ph0          .54   .59   .69   .69   .68
      mb128ic2048ih7oc512oh7kh1ph0           .39   .42   .50   .52   .52
      mb128ic256ih14oc1024oh14kh1ph0         .43   .47   .55   .54   .54
      mb128ic256ih14oc256oh14kh3ph1          .45   .49   .58   .58   .58
      mb128ic256ih56oc128oh56kh1ph0          .50   .54   .62   .60   .57
      mb128ic256ih56oc64oh56kh1ph0           .50   .54   .62   .60   .60
      mb128ic512ih28oc128oh28kh1ph0          .59   .64   .73   .70   .68
      mb128ic512ih28oc256oh28kh1ph0          .60   .66   .77   .74   .71
      mb128ic512ih7oc2048oh7kh1ph0           .49   .54   .63   .62   .61
      mb128ic512ih7oc512oh7kh3ph1            .50   .51   .60   .60   .61
      mb128ic64ih56oc256oh56kh1ph0           .35   .38   .44   .45   .47
      mb128ic64ih56oc64oh56kh3ph1            .51   .55   .63   .60   .57
FWD_D mb128ic1024ih14oc2048oh7kh1sh2ph0      .55   .61   .72   .75   .78
      mb128ic1024ih14oc256oh14kh1ph0         .45   .49   .58   .61   .60
      mb128ic1024ih14oc512oh14kh1ph0         .51   .56   .65   .67   .67
      mb128ic128ih28oc128oh28kh3ph1          .19   .21   .24   .27   .29
      mb128ic128ih28oc512oh28kh1ph0          .65   .70   .79   .77   .76
      mb128ic128ih56oc128oh28kh3sh2ph1       .17   .19   .22   .26   .28
      mb128ic2048ih7oc512oh7kh1ph0           .52   .58   .67   .68   .69
      mb128ic256ih14oc1024oh14kh1ph0         .51   .55   .65   .67   .67
      mb128ic256ih14oc256oh14kh3ph1          .41   .45   .55   .59   .60
      mb128ic256ih28oc256oh14kh3sh2ph1       .44   .49   .59   .64   .64
      mb128ic256ih56oc128oh56kh1ph0          .53   .58   .68   .67   .67
      mb128ic256ih56oc512oh28kh1sh2ph0       .24   .26   .31   .35   .36
      mb128ic256ih56oc64oh56kh1ph0           .34   .37   .43   .44   .46
      mb128ic3ih224oc64oh112kh7sh2ph3        .32   .34   .39   .42   .42
      mb128ic512ih14oc512oh7kh3sh2ph1        .55   .61   .73   .76   .78
      mb128ic512ih28oc1024oh14kh1sh2ph0      .53   .58   .68   .70   .70
      mb128ic512ih28oc128oh28kh1ph0          .55   .60   .70   .70   .69
      mb128ic512ih28oc256oh28kh1ph0          .55   .60   .71   .70   .70
      mb128ic512ih7oc2048oh7kh1ph0           .54   .59   .69   .68   .68
      mb128ic512ih7oc512oh7kh3ph1            .50   .47   .55   .60   .62
      mb128ic64ih56oc256oh56kh1ph0           .52   .56   .65   .64   .63
      mb128ic64ih56oc64oh56kh3ph1            .26   .28   .33   .36   .37

```

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit? (except known failures)

## Performance improvements


src/cpu/aarch64/matmul/brgemm_matmul_reorders.cpp
```diff
@@ -1,6 +1,8 @@
 /*******************************************************************************
 * Copyright 2022-2023 Intel Corporation
 * Copyright 2024 FUJITSU LIMITED
+* Copyright 2025 Arm Ltd. and affiliates
+*                                                                              ││    │                                                                         │  5 │*
```
aditew01, line 5: Looks like a Typo to me.

src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -550,7 +551,7 @@ struct jit_brgemm_matmul_copy_b_f32_t : public jit_brgemm_matmul_copy_b_t,
 void jit_brgemm_matmul_copy_b_f32_t::copy_16_8_x_n_block(
         int nrows, int ncolumns) {
 
-    int n_blk_step = is_sve_256 ? 8 : 16;
+    int n_blk_step = get_sve_length() / typesize_in_;
```
renato-arantes, line 554: Correct me if I'm wrong, but here you are calculating the channel dimension block size, which should be 4, 8 or 16 for SVE_128, 256 and 512. What if the `typesize_in_` is not f32?

src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -610,7 +611,7 @@ void jit_brgemm_matmul_copy_b_f32_t::compute_k_loop(int ncolumns) {
         L(K_end_label);
     };
 
-    int k_unroll = is_sve_256 ? 8 : 16;
+    int k_unroll = get_sve_length() / typesize_in_;
```
renato-arantes, line 614: Same as above.

src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
isharif168, line 554: is it possible if we can create a function for this utility and use that in place of calculating over here , so that if new type comes in future , it would be good to use, what you think about it ?

jondea: Do you mean something like
```
get_sve_length(data_type)
```

src/cpu/aarch64/matmul/brgemm_matmul_reorders.cpp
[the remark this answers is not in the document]
jondea: Good question.
It is my tmux column numbers.
It is careless copy pasting.
It is me using more git magic, and less copy pasting in the future...

src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
jondea: I have added a new function, `simd_elems`, which does this and I have used it where appropriate in this PR.

jondea: After looking more deeply into this, I've decided to walk this back a bit and revisit it later. `get_sve_length()` currently gets the `sveLen_` directly from `HwCap`. I think the correct behaviour would be to use `get_max_cpu_isa()` which allows the user to restrict the ISA. I don't want to implement a new function which is potentially wrong, and I think untangling this is outside of the scope of this PR.

I've kept `simd_elems` but only for the case where the caller explicitly provides the isa.

renato-arantes (approved):

aditew01 (approved):

src/cpu/aarch64/jit_brgemm_post_ops.hpp
```diff
@@ -197,19 +198,12 @@ struct jit_brgemm_kernel_diff_bias_t : public jit_generator {
...
-                simd_w_ = cpu_isa_traits<sve_256>::vlen / sizeof(float);
-                break;
-            default: {
-                assert(!"unsupported isa");
-                return;
-            }
-        }
+
+        if (!utils::one_of(brg_.isa_impl, sve_512, sve_256, sve_128))
+            assert(!"unsupported isa: only SVE 512, 256 and 128 supported");
```
isharif168: Do we really require this check here instead we can assert in simd_elems?

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -278,6 +279,18 @@ inline size_t data_type_vnni_simd_elems(data_type_t data_type) {
...
 
+// Maximum number of elements of a given type in a SIMD (SVE/Neon) vector for a
+// given ISA
+inline size_t simd_elems(data_type_t dt, cpu_isa_t cpu_isa) {
+    switch (cpu_isa) {
+        case sve_512: return data_type_vnni_simd_elems<sve_512>(dt);
+        case sve_256: return data_type_vnni_simd_elems<sve_256>(dt);
+        case sve_128:
+        case asimd: return data_type_vnni_simd_elems<sve_128>(dt);
+        default: return 0;
```
isharif168: should we assert here in default ? Your thoughts?

src/cpu/aarch64/jit_brgemm_post_ops.hpp
jondea: It's a good point, but I think it is up to the kernels to enforce which ISAs they support. E.g. here we don't support `asimd`, but `asimd` should return a non-zero number for `simd_elems`.

src/cpu/aarch64/cpu_isa_traits.hpp
jondea: `asserts` should generally be avoided in oneDNN. Where possible, errors should be dealt with using normal control flow (e.g. `status_t::unimplemented`). As such, I wouldn't want a generic helper function to assert. I was reluctant to use an assert in [src/cpu/aarch64/jit_brgemm_post_ops.hpp](https://github.com/uxlfoundation/oneDNN/pull/3363/files/3eed9939b91cdaf250956d75ba64767723ba7a4b#diff-3610d1c6563e544b9ab7903e162af2d273977e75d951453a5deede52041a4f2c) but at this point there is no ability to return a status. In this case I erred on the side of leaving the behaviour the same, but I wouldn't mind removing it in the future. I think that there's a lot of tech debt in this code, but I think it should be rationalized with a dedicated PR which has as few functional changes as possible.

Back to `simd_elems`, I think 0 is a reasonable value to return if a machine has no vector ISA (I actually don't know if this is possible, it certainly isn't for Armv8 because Neon is mandatory).

src/cpu/aarch64/jit_brgemm_post_ops.hpp
isharif168: Thats what I was skeptical about, that for `asimd` , this if condition will fail and it will call the assert to terminate, so handling in `simd_elems `is necessary (it will never reach there if isa is `asimd`), correct me if I am wrong please.

Ryo-not-rio (approved):
I've confirmed all the unit tests pass on a 128 SVE machine locally both with and without ACL so happy for this to go in

Sqvid (approved):
LGTM too

Sqvid:
Would appreciate a review from @uxlfoundation/onednn-arch, thanks.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1847,19 +1847,12 @@ void jit_brgemm_kernel_t::bdb_loop() {
...
-            break;
-        default: {
-            assert(!"unsupported isa");
-            return;
-        }
-    }
+
+    if (!one_of(brg.isa_impl, sve_512, sve_256, sve_128))
+        assert(!"unsupported isa: only SVE 512, 256 and 128 supported");
+
```
dzarukin:
```suggestion
    if (!one_of(brg.isa_impl, sve_512, sve_256, sve_128)) {
        assert(!"unsupported isa: only SVE 512, 256 and 128 supported");
    }
```

src/cpu/aarch64/cpu_isa_traits.hpp
dzarukin: I'd put an assert here, it's not expected to get `0` from this function anyway, but finding this spot with assert would happen in no time when the next arch will be added.

src/cpu/aarch64/jit_brgemm_post_ops.hpp
```diff
@@ -197,19 +198,12 @@ struct jit_brgemm_kernel_diff_bias_t : public jit_generator {
...
+
```
dzarukin:
```suggestion
        if (!utils::one_of(brg_.isa_impl, sve_512, sve_256, sve_128)) {
            assert(!"unsupported isa: only SVE 512, 256 and 128 supported");
        }
```

src/cpu/aarch64/jit_brgemm_post_ops.hpp
```diff
@@ -854,19 +848,11 @@ struct jit_brgemm_kernel_post_ops : public jit_generator {
...
-            case sve_256:
-                simd_w_ = cpu_isa_traits<sve_256>::vlen / sizeof(float);
-                break;
-            default: {
-                assert(!"unsupported isa");
-                return;
-            }
-        }
+        if (!utils::one_of(brg.isa_impl, sve_512, sve_256, sve_128))
+            assert(!"unsupported isa: only SVE 512, 256 and 128 supported");
```
dzarukin, line 855: Ditto.

dzarukin (approved):

src/cpu/aarch64/jit_brgemm_post_ops.hpp
Sqvid: When we eventually get around to using `>=clang-format-15` we should add the [InsertBraces rule](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces).

Sqvid (changes requested):
Temporarily blocking till the end-to-end benchmarking is complete.

michalowski-arm:
Just a heads up this is now giving incorrect results when running u8/s8 brgemm, eg.: `./tests/benchdnn/benchdnn --brgemm --dt=u8:s8:u8 13x192:192x32_n"int8:no_tail:21"` (from test_benchdnn_modeC_brgemm_ci_cpu), which gives expected results with sve_256 but not sve_128. On the other hand, bf16 brgemm from #3731 seems to work as expected, and with some extra minor changes the bf16 brgconv sve_128 also worked in all the cases I tested.

aditew01:
we've run in some issues with bf16 conv, with this change as is (after including the stubs in cpu_conv_list.cpp), following this PR : https://github.com/uxlfoundation/oneDNN/pull/3731 . We can raise followup a patch on top of this. 
cc: @michalowski-arm

jondea:
Good spot both, I will sort this one way or another and get it in.

aditew01:
This is not necessarily a blocker for this to go in. The BF16 conv needs to be enabled explicitly in the `cpu_conv_list`. We can fix the related issues and push the change.

jondea:
Thanks @michalowski-arm, int8 brgemm now passes, thankfully it was a very simple change

src/cpu/aarch64/jit_brgemm_post_ops.hpp
jondea: Done

src/cpu/aarch64/jit_brgemm_post_ops.hpp
jondea: Done

src/cpu/aarch64/cpu_isa_traits.hpp
jondea: All good points. I have added the assert

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Done

jondea:
@Sqvid I've done some e2e benchmarks and the impact is minimal, but sometimes negative. However, we have work in progress to mitigate it. On balance, it is worth getting this in so that we can build on top of it.

Sqvid:
@jondea Removed the block. You can merge when you're ready.

PR #3365, opened 2025-05-29
tests: gtest: add check for khr_get_default_context

spalicki:
Build issues fixes:
1. `ext_oneapi_get_default_context` is deprecated in newest Intel Compiler in favor of `khr_get_default_context`. I have added a check for new implementation based on presence of mandatory `SYCL_KHR_DEFAULT_CONTEXT` macro.
2. added `defined()` guard to `MIOPEN_HAS_INT8X4` macro

### DIFF
--- a/src/gpu/amd/sycl_hip_utils.hpp
+++ b/src/gpu/amd/sycl_hip_utils.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2020-2024 Intel Corporation
+* Copyright 2020-2025 Intel Corporation
 * Copyright 2020 Codeplay Software Limited
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


dzarukin (approved):

src/gpu/amd/sycl_hip_utils.hpp
```diff
@@ -134,7 +134,7 @@ inline status_t convert_data_type(const memory_desc_t *mem_desc,
         case data_type_t::dnnl_s8: {
             if (vectorized
                     && mem_desc->format_desc.blocking.inner_blks[0] == 4) {
-#if MIOPEN_HAS_INT8X4
+#if defined(MIOPEN_HAS_INT8X4)
```
Rbiessy, line 137: Can you explain why this is needed?

As far as I can tell `MIOPEN_HAS_INT8X4` is either set to `1` or not defined at all, see here: https://github.com/uxlfoundation/oneDNN/blob/4f1fd078770c369b9363a27e578ec88d91ccd0b7/cmake/FindMIOpen.cmake#L73

Also note there is another occuerence of this macro in https://github.com/uxlfoundation/oneDNN/blob/4f1fd078770c369b9363a27e578ec88d91ccd0b7/src/gpu/amd/miopen_gemm_inner_product_impl.hpp#L62

spalicki: We compile oneDNN with `-Wundef`, so if it is not defined it will create and warning/werror.

spalicki: Done.

Rbiessy (approved):

PR #3378, opened 2025-05-30
cpu: aarch64: eltwise: make jit eltwise vector length agnostic

Ryo-not-rio:
Rewrite jit eltwise to use runtime vector length detection and vector length agnostic. This cleans up the code and removes the need for the isa template parameter.

## Performance
Testing on neoverse-v1, there are no regressions on any of the tests listed in `test_eltwise_ci`


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
```diff
@@ -79,11 +80,12 @@ bool is_supported(cpu_isa_t isa, alg_kind_t alg);
 
 } // namespace eltwise_injector
 
-template <cpu_isa_t isa>
 struct jit_uni_eltwise_injector_f32 {
-    using TReg = typename cpu_isa_traits<isa>::TReg;
-    using TRegS = typename cpu_isa_traits<isa>::TRegS;
-
+    typedef Xbyak_aarch64::ZReg TReg;
```
jondea: I don't think this is the right approach. If we are specializing our class to SVE then we should explicitly use `ZReg` everywhere rather than aliasing `TReg`

src/cpu/aarch64/jit_uni_eltwise.hpp
```diff
@@ -36,13 +37,12 @@ namespace aarch64 {
...
 
-template <cpu_isa_t isa, impl::data_type_t d_type>
+template <impl::data_type_t d_type>
 struct jit_uni_eltwise_fwd_t : public primitive_t {
     struct pd_t : public cpu_eltwise_fwd_pd_t {
         using cpu_eltwise_fwd_pd_t::cpu_eltwise_fwd_pd_t;
 
-        DECLARE_COMMON_PD_T(
-                JIT_IMPL_NAME_HELPER("jit:", isa, ""), jit_uni_eltwise_fwd_t);
+        DECLARE_COMMON_PD_T("jit: sve", jit_uni_eltwise_fwd_t);
```
jondea: impl names don't typically have spaces

jondea (changes requested):
Thank you for taking a look at this, it's great to reduce the number of implementation by 3x, but I'm not convinced on the exact approach. Specifically, think about how we might extend the primitive and injectors to Neon. I think there's could be a lot of code reuse if we do it right. Almost all the `alg_compute_vector` functions would have similar logic for Neon (provided we have helper functions for providing p_all in SVE cases). I think we can use sve_128 as a stand in for > sve_128.

Ryo-not-rio:
@jondea I've changed it back to being a template, so we can reuse it for NEON as you suggested

src/cpu/aarch64/injectors/jit_uni_postops_injector.cpp
```diff
@@ -236,7 +236,7 @@ bool post_ops_ok(const post_ops_ok_args_t &post_ops_ok_args) {
                 case eltwise:
                     if (entry.is_eltwise()) {
                         const auto alg = entry.eltwise.alg;
-                        return eltwise_injector::is_supported(isa, alg);
+                        return eltwise_injector::is_supported(sve_128, alg);
```
jondea: I think we can leave this as-is because `eltwise_injector::is_supported` checks if it is a superset.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -166,7 +166,8 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
             st1w(vmm_src.s, pg_s / T_z, ptr(reg_dst));
         }
 
-        const auto shift = cpu_isa_traits<isa>::vlen;
+        const auto shift
+                = isa == sve_128 ? get_sve_length() : cpu_isa_traits<isa>::vlen;
```
jondea: Makes sense to use the `vlen()` function

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -365,16 +367,12 @@ status_t jit_uni_eltwise_bwd_t<isa, d_type>::execute(
     return status::success;
 }
 
-template struct jit_uni_eltwise_fwd_t<sve_512, data_type::f32>;
```
jondea, line 368: I love the smell of freshly deleted lines of code in the morning!

src/cpu/aarch64/jit_uni_softmax.cpp
```diff
@@ -61,8 +62,8 @@ struct jit_softmax_base_t : public jit_generator {
     const memory_desc_wrapper src_d_, dst_d_, diff_dst_d_;
 
     virtual void operator()(const call_params_t *p) = 0;
-    std::unique_ptr<jit_uni_eltwise_injector_f32<isa>> exp_injector_;
-    std::unique_ptr<jit_uni_eltwise_injector_f32<isa>> log_injector_;
+    std::unique_ptr<jit_uni_eltwise_injector_f32<sve_128>> exp_injector_;
```
jondea: This looks deeply unsettling because
1. We are mixing ISAs (i.e. sve_512 and sve_128)
2. This will silently fail if we extend this primitive to `asmid`

Can we do some `constexpr` magic to either
- Make `jit_uni_eltwise_injector_f32<sve_256>` compile to `jit_uni_eltwise_injector_f32<sve_128>`, maybe using?
- Introduce a constexpr helper `injector_helper<isa>::isa` which maps asimd -> asimd, sve_* -> sve_128?

Stuff like this is pushing me towards thinking we need a VLA `sve` enum value, although that's definitely out of the scope of this PR.

jondea (changes requested):
Sorry, another round. This is great work though, it's just very new territory so there's some kinks to work out.

src/cpu/aarch64/injectors/jit_uni_postops_injector.cpp
Ryo-not-rio: I've changed `is_supported` to only support sve_128, I think that makes it less confusing, I've changed this part of the code to use the new constexpr function

src/cpu/aarch64/jit_uni_eltwise.cpp
Ryo-not-rio: done

src/cpu/aarch64/jit_uni_softmax.cpp
Ryo-not-rio: Fixed!

vpirogov:
Ping @uxlfoundation/onednn-arch for review.

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -102,6 +102,11 @@ cpu_isa_t DNNL_API get_max_cpu_isa_mask(bool soft = false);
 status_t set_max_cpu_isa(dnnl_cpu_isa_t isa);
 dnnl_cpu_isa_t get_effective_cpu_isa();
 
+// If isa is a superset of sve_128, return sve_128, else return isa
+constexpr cpu_isa_t to_vla_sve(const cpu_isa_t isa) {
+    return (cpu_isa_t)(isa & sve_128);
```
dzarukin, line 107: I'm not so sure it returns `isa` if `isa` doesn't contain `sve_128`. I'd expect it returning `0` cast to `cpu_isa_t` which might not be expected.

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2018-2023 Intel Corporation
 * Copyright 2020-2024 FUJITSU LIMITED
-* Copyright 2023 Arm Ltd. and affiliates 
+* Copyright 2023, 2025 Arm Ltd. and affiliates
```
dzarukin, line 4: Maybe dash?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2398,8 +2398,7 @@ void jit_uni_eltwise_injector_f32<isa>::register_table_entries() {
     }
 }
 
-template struct jit_uni_eltwise_injector_f32<sve_512>;
-template struct jit_uni_eltwise_injector_f32<sve_256>;
+// We only need sve_128 as the injector is fully vector length agnostic
```
dzarukin, line 2401: If the point about vlen agnostic is true, maybe stripping a templated argument then?
At least, would be nice to consider and if considered, a TODO would be a good thing to have here.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -202,8 +202,9 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
 private:
     using TReg = typename cpu_isa_traits<isa>::TReg;
     using TRegS = typename cpu_isa_traits<isa>::TRegS;
-
-    int vlen() { return cpu_isa_traits<isa>::vlen; }
+    int vlen() {
+        return isa == sve_128 ? get_sve_length() : cpu_isa_traits<isa>::vlen;
```
dzarukin: Feels like if `isa` is `sve_128`, it should be handled by `cpu_isa_traits` as well and this change shouldn't be needed.

src/cpu/aarch64/jit_uni_eltwise.cpp
dzarukin: One could go further and strip data type as a templated argument, but it's up to you.

dzarukin (approved):

src/cpu/aarch64/cpu_isa_traits.hpp
Sqvid: Our convention is to use a dash only when all years in the range had a modification. In this case there was no change in 2024.

src/cpu/aarch64/jit_uni_eltwise.cpp
jondea: Yes, we have separate work planned for this

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Ryo-not-rio: We've kept it as a template as we're looking to reuse jit_uni_eltwise for a NEON implementation

src/cpu/aarch64/jit_uni_eltwise.cpp
Ryo-not-rio: kept for the NEON implementation

src/cpu/aarch64/cpu_isa_traits.hpp
Ryo-not-rio: I've tested all values of cpu_isa_t and it returns what I expect it to return

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: We are also exploring options to make this clearer, possibly by adding an `sve` enum value to signal VLA SVE. Or making the arch orthogonal to the SVE vector length. Either way, I think this is best done in an RFC, and we can come back to it

src/cpu/aarch64/jit_uni_eltwise.cpp
jondea: I agree with @dzarukin, this isn't really the right place to handle this, but I don't think it's something we should sort out in this PR. But, if we don't, then we need a decent comment here explaining that we are using `sve_128` as a proxy for VLA SVE and a TODO. I also think that it would be clearer to rearrange the logic to check for `asimd` rather than `== sve_128`, because looking at it makes me wonder what happens to `sve_256` and `sve_512`

jondea (changes requested):

src/cpu/aarch64/jit_uni_eltwise.cpp
Ryo-not-rio: I've swapped around the logic and added a comment with a TODO

PR #3380, opened 2025-05-30
doc: build: updates to build from source

ranukund:
# Description

Addressing the recommendations in the [DOC 13142](https://jira.devtools.intel.com/browse/DOC-13142)

### DIFF
--- a/doc/build/build.md
+++ b/doc/build/build.md
@@ -1,5 +1,6 @@
 Build from Source {#dev_guide_build}
 ====================================
+You can install the oneDNN library using the source distribution.
 
 ## Download the Source Code
 
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
 
 ~~~sh
 git clone https://github.com/uxlfoundation/oneDNN.git
+cd oneDNN
 ~~~
 
 ## Build the Library
 
-### Set up the environment for the compiler
+You can quickly get started with building the library. The various stages involved in building the library are as follows:
 
-Ensure that all [software dependencies](https://github.com/uxlfoundation/oneDNN#requirements-for-building-from-source)
-are in place and have at least the minimal supported version.
+1. Set up the environment for the compiler
 
-### Generate the build system
+   Ensure that all [software dependencies](https://github.com/uxlfoundation/oneDNN#requirements-for-building-from-source)
+   are in place and have at least the minimal supported version.
 
-The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). Use the following command to generate a build system:
-~~~sh
-cmake -B <path-to-build> [-S <path-to-source>] [<options>]
-~~~
-In most cases, both `-B` and `-S` options are skipped with the assumption that
-`<path-to-build>` is the current folder and `<path-to-source>` is the one
-higher in the tree:
-~~~sh
-cd <path-to-onednn-source>
-mkdir -p build ; cd build
-cmake .. [<options>]
-~~~
-The following are a few useful options defined by CMake:
+2. Generate the build system
 
-- `G` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
-  `Ninja`, `"Unix Makefiles"`).
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
+   ~~~
+   In most cases, both `-B` and `-S` options are skipped with the assumption that
+  `<path-to-build>` is the current folder and `<path-to-source>` is the one
+   higher in the tree:
+   ~~~sh
+   cd <path-to-onednn-source>
+   mkdir -p build ; cd build
+   cmake .. [<options>]
+   ~~~
+   The following are a few useful options defined by CMake:
 
-- `CMAKE_INSTALL_PREFIX` to control the library installation location.
+   - `G` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
+    `Ninja`, `"Unix Makefiles"`).
 
-- `CMAKE_BUILD_TYPE` to select between build type (`Release`, `Debug`,
-  `RelWithDebInfo`).
+   - `CMAKE_INSTALL_PREFIX` to control the library installation location.
 
-- `CMAKE_PREFIX_PATH` to specify directories to be searched for the
-  dependencies located at non-standard locations.
+   - `CMAKE_BUILD_TYPE` to select between build type (`Release`, `Debug`,
+    `RelWithDebInfo`).
 
-See @ref dev_guide_build_options for detailed description of build-time
-configuration options defined by oneDNN.
+   - `CMAKE_PREFIX_PATH` to specify directories to be searched for the
+    dependencies located at non-standard locations.
 
-### Build the library
+   See @ref dev_guide_build_options for detailed description of build-time
+   configuration options defined by oneDNN.
 
-CMake provides a unified method for building a project, independent of the
-generator or operating system used:
-~~~sh
-cmake --build <path-to-build> [<options>]
-~~~
-Full list of options can be found [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project).
+3. Build the library
 
-### Linux/macOS
+   CMake provides a unified method for building a project, independent of the
+   generator or operating system used:
+   ~~~sh
+   cmake --build <path-to-build> [<options>]
+   ~~~
+   Full list of options can be found [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project).
 
-#### GCC, Clang, or Intel oneAPI DPC++/C++ Compiler
+You can build the library on Linux, macOS, or Windows using the compiler of your choice.
 
-- Set up the environment for the compiler
+### Build on Linux and macOS
 
-~~~sh
-# Uncomment the following lines to build with GCC
-# export CC=gcc
-# export CXX=g++
+#### Use GCC, Clang, or Intel oneAPI DPC++/C++ Compiler
 
-# Uncomment the following lines to build with Clang
-# export CC=clang
-# export CXX=clang++
+1. Set up the environment for the compiler
+   ~~~sh
+   # Uncomment the following lines to build with GCC
+   # export CC=gcc
+   # export CXX=g++
 
-# Uncomment the following lines to build with Intel oneAPI DPC++/C++ Compiler (x64 only)
-# export CC=icx
-# export CXX=icpx
-~~~
+   # Uncomment the following lines to build with Clang
+   # export CC=clang
+   # export CXX=clang++
 
-- Generate the build system
+   # Uncomment the following lines to build with Intel oneAPI DPC++/C++ Compiler (x64 only)
+   # export CC=icx
+   # export CXX=icpx
+   ~~~
 
-~~~sh
-mkdir -p build ; cd build
-cmake ..
-~~~
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake ..
+   ~~~
 
-- Build the library
+3. Build the library
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
+   # compilation. Parallelization can be controlled with:
+   cmake --build . --parallel $(nproc)
+   ~~~
 
-~~~sh
-# Some generators, like Unix Makefiles, might default to single-threaded
-# compilation. Parallelization can be controlled with:
-cmake --build . --parallel $(nproc)
-~~~
+#### Use Intel oneAPI DPC++/C++ Compiler with SYCL runtime
 
-#### Intel oneAPI DPC++/C++ Compiler with SYCL runtime
+1. Set up the environment for the compiler
 
-- Set up the environment for the compiler
+   Intel oneAPI DPC++/C++ Compiler uses the `setvars.sh` script to set all the
+   required variables. The command below assumes you installed to the default
+   folder. If you customized the installation folder, `setvars.sh` (Linux/macOS)
+   is in your custom folder.
+   ~~~sh
+   source /opt/intel/oneapi/setvars.sh
 
-Intel oneAPI DPC++/C++ Compiler uses the `setvars.sh` script to set all
-required variables. The command below assumes you installed to the default
-folder. If you customized the installation folder, `setvars.sh` (Linux/macOS)
-is in your custom folder.
-~~~sh
-source /opt/intel/oneapi/setvars.sh
+   # Set Intel oneAPI DPC++/C++ Compiler as default C and C++ compilers
+   export CC=icx
+   export CXX=icpx
+   ~~~
 
-# Set Intel oneAPI DPC++/C++ Compiler as default C and C++ compilers
-export CC=icx
-export CXX=icpx
-~~~
-
-- Generate the build system
-
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DDNNL_CPU_RUNTIME=SYCL \
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_CPU_RUNTIME=SYCL \
          -DDNNL_GPU_RUNTIME=SYCL
-~~~
+   ~~~
 
-@note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
-use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
-contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
-environment variable of the same name to specify path to the OpenCL runtime if
-it is installed in a custom location.
+   @note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
 
-- Build the library
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
 
-~~~sh
-cmake --build . --parallel $(nproc)
-~~~
+#### Use GCC targeting AArch64 on x64 host
 
-#### GCC targeting AArch64 on x64 host
+1. Set up the environment for the compiler
+   ~~~sh
+   export CC=aarch64-linux-gnu-gcc
+   export CXX=aarch64-linux-gnu-g++
+   ~~~
 
-- Set up the environment for the compiler
-
-~~~sh
-export CC=aarch64-linux-gnu-gcc
-export CXX=aarch64-linux-gnu-g++
-~~~
-
-- Generate the build system
-
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DCMAKE_SYSTEM_NAME=Linux \
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DCMAKE_SYSTEM_NAME=Linux \
          -DCMAKE_SYSTEM_PROCESSOR=AARCH64 \
          -DCMAKE_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib
-~~~
+   ~~~
 
-- Build the library
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
 
-~~~sh
-cmake --build . --parallel $(nproc)
-~~~
+#### Use GCC with Arm Compute Library (ACL) on AArch64 host
 
-#### GCC with Arm Compute Library (ACL) on AArch64 host
+1. Set up the environment for the compiler
 
-- Set up the environment for the compiler
+   Download [Arm Compute Library](https://github.com/ARM-software/ComputeLibrary)
+   or build it from source and set `ACL_ROOT_DIR` to directory where it is
+   installed.
+   ~~~sh
+   export ACL_ROOT_DIR=<path/to/ComputeLibrary>
+   export CC=gcc
+   export CXX=g++
+   ~~~
 
-Download [Arm Compute Library](https://github.com/ARM-software/ComputeLibrary)
-or build it from source and set `ACL_ROOT_DIR` to directory where it is
-installed.
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_AARCH64_USE_ACL=ON
+   ~~~
 
-~~~sh
-export ACL_ROOT_DIR=<path/to/ComputeLibrary>
-export CC=gcc
-export CXX=g++
-~~~
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
 
-- Generate the build system
+### Build on Windows
 
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DDNNL_AARCH64_USE_ACL=ON
-~~~
+#### Use Microsoft Visual C++ Compiler
 
-- Build the library
+1. Set up the environment for the compiler
 
-~~~sh
-cmake --build . --parallel $(nproc)
-~~~
+   Microsoft Visual Studio uses the `VsDevCmd.bat` script to set all
+   required variables. The command below assumes you installed to the default
+   folder. If you customized the installation folder, `VsDevCmd.bat` is in your
+   custom folder.
+   ~~~bat
+   "C:\Program Files\Microsoft Visual Studio\2022\Professional\Common7\Tools\VsDevCmd.bat" -startdir=none -arch=x64 -host_arch=x64
+   ~~~
+   or open `x64 Native Tools Command Prompt` from start menu instead.
 
-### Windows
+2. Generate the build system
+   ~~~bat
+   mkdir build
+   cd build
+   cmake .. -G "Visual Studio 17 2022"
+   ~~~
 
-#### Microsoft Visual C++ Compiler
+3. Build the library
+   ~~~bat
+   cmake --build . --config=Release
+   ~~~
 
-- Set up the environment for the compiler
+   @note Currently, the oneDNN build system has limited support for multi-config
+   generators. Build configuration is based on the `CMAKE_BUILD_TYPE` option
+   (`Release` by default), and CMake must be rerun from scratch every time
+   the build type changes to apply the new build configuration. You can choose
+   a specific build type with the `--config` option (the solution file supports
+   both `Debug` and `Release` builds), but it must refer to the same build type
+   (`Release`, `Debug`, etc.) as selected with the `CMAKE_BUILD_TYPE` option.
 
-Microsoft Visual Studio uses the `VsDevCmd.bat` script to set all
-required variables. The command below assumes you installed to the default
-folder. If you customized the installation folder, `VsDevCmd.bat` is in your
-custom folder.
-~~~bat
-"C:\Program Files\Microsoft Visual Studio\2022\Professional\Common7\Tools\VsDevCmd.bat" -startdir=none -arch=x64 -host_arch=x64
-~~~
-or open `x64 Native Tools Command Prompt` from start menu instead.
+   @note You can also open `oneDNN.sln` to build the project from the
+   Microsoft Visual Studio IDE.
 
-- Generate the build system
+#### Use Intel oneAPI DPC++/C++ Compiler with SYCL Runtime
 
-~~~bat
-mkdir build
-cd build
-cmake .. -G "Visual Studio 17 2022"
-~~~
+1. Set up the environment for the compiler
 
-- Build the library
+   Intel oneAPI DPC++/C++ Compiler uses the `setvars.bat` script to set all
+   required variables. The command below assumes you installed to the default
+   folder. If you customized the installation folder, `setvars.bat` is in your
+   custom folder.
+   ~~~bat
+   "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
 
-~~~bat
-cmake --build . --config=Release
-~~~
+   :: Set Intel oneAPI DPC++/C++ Compiler as default C and C++ compilers
+   set CC=icx
+   set CXX=icx
+   ~~~
+   or open `Intel oneAPI Commmand Prompt` from start menu instead.
 
-@note Currently, the oneDNN build system has limited support for multi-config
- generators. Build configuration is based on the `CMAKE_BUILD_TYPE` option
- (`Release` by default), and CMake must be rerun from scratch every time
- the build type changes to apply the new build configuration. You can choose
- a specific build type with the `--config` option (the solution file supports
- both `Debug` and `Release` builds), but it must refer to the same build type
- (`Release`, `Debug`, etc.) as selected with the `CMAKE_BUILD_TYPE` option.
+2. Generate the build system
+   ~~~bat
+   mkdir build
+   cd build
 
-@note You can also open `oneDNN.sln` to build the project from the
-Microsoft Visual Studio IDE.
-
-#### Intel oneAPI DPC++/C++ Compiler with SYCL Runtime
-
-- Set up the environment for the compiler
-
-Intel oneAPI DPC++/C++ Compiler uses the `setvars.bat` script to set all
-required variables. The command below assumes you installed to the default
-folder. If you customized the installation folder, `setvars.bat` is in your
-custom folder.
-~~~bat
-"C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
-
-:: Set Intel oneAPI DPC++/C++ Compiler as default C and C++ compilers
-set CC=icx
-set CXX=icx
-~~~
-or open `Intel oneAPI Commmand Prompt` from start menu instead.
-
-- Generate the build system
-
-~~~bat
-mkdir build
-cd build
-
-cmake .. -G Ninja ^
+   cmake .. -G Ninja ^
          -DDNNL_CPU_RUNTIME=SYCL ^
          -DDNNL_GPU_RUNTIME=SYCL
-~~~
+   ~~~
 
-@warning Intel oneAPI DPC++/C++ Compiler on Windows requires CMake v3.23 or later.
+   @warning Intel oneAPI DPC++/C++ Compiler on Windows requires CMake v3.23 or later.
 
-@warning Intel oneAPI DPC++/C++ Compiler does not support CMake's Microsoft Visual
-Studio generator.
+   @warning Intel oneAPI DPC++/C++ Compiler does not support CMake's Microsoft Visual
+   Studio generator.
 
-@note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
-use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
-contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
-environment variable of the same name to specify path to the OpenCL runtime if
-it is installed in a custom location.
+   @note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
 
-- Build the library
-
-~~~bat
-cmake --build .
-~~~
+3. Build the library
+   ~~~bat
+   cmake --build .
+   ~~~
 
 ## Validate the Build
 
 After building the library, you can run a predefined test set using:
+
 ~~~sh
 ctest
 ~~~
@@ -294,26 +287,30 @@ When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI toolk
 Make sure the correct oneDNN library is present in
 `LD_LIBRARY_PATH` by setting it explicitly if needed.
 
-## Build documentation
+## (Optional) Build Documentation
 
-- Install the requirements
-~~~sh
-conda env create -f ../doc/environment.yml
-conda activate onednn-doc
-~~~
+1. Install the requirements
+   ~~~sh
+   conda env create -f ../doc/environment.yml
+   conda activate onednn-doc
+   ~~~
 
-- Build the documentation
-~~~sh
-cmake --build . --target doc
-~~~
+2. Build the documentation
+   ~~~sh
+   cmake --build . --target doc
+   ~~~
 
-## Install library
+## Install the Library
 
-Install the library, headers, and documentation
+Install the library, headers, and documentation.
+
+@note
+The install directory is specified by the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)
+CMake variable. When installing in the default directory, the following command
+needs to be run with administrative privileges using `sudo` on Linux/Mac or a
+command prompt run as administrator on Windows.
+
 ~~~sh
 cmake --build . --target install
 ~~~
-The install directory is specified by the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)
-cmake variable. When installing in the default directory, the above command
-needs to be run with administrative privileges using `sudo` on Linux/Mac or a
-command prompt run as administrator on Windows.
+


ranukund:
Hi @uxlfoundation/onednn-doc @uxlfoundation/onednn-arch please review when you have a moment, thanks!

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
 ~~~
 
 ## Build the Library
 
-### Set up the environment for the compiler
+You can quickly get started with building the library. The various stages involved in building the library are as follows:
 
-Ensure that all [software dependencies](https://github.com/uxlfoundation/oneDNN#requirements-for-building-from-source)
-are in place and have at least the minimal supported version.
+1. Set up the environment for the compiler
```
Marina-Zubova: We provide here the general instruction or high-level overview of the building process. However, step 1 'Set up the environment for the compiler' is not an overview, but a required step before proceeding with build from source. To make sure developer don't overlook this step, I suggest moving it to between 'Download the Source Code' and 'Build the library' to make it more visible. We can call it 'Prerequisites for Building from Source' or similar.

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
+   ~~~
+   In most cases, both `-B` and `-S` options are skipped with the assumption that
+  `<path-to-build>` is the current folder and `<path-to-source>` is the one
+   higher in the tree:
+   ~~~sh
+   cd <path-to-onednn-source>
```
Marina-Zubova: Fixing inconsistency issue
```suggestion
   cd <path-to-source>
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-~~~
-The following are a few useful options defined by CMake:
+2. Generate the build system
 
-- `G` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
-  `Ninja`, `"Unix Makefiles"`).
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
```
Marina-Zubova: This command following by the next code snippet looks redundant. Developer needs to run a cmake command only once, while here 2 cmake commands are provided. At the same time, the first command (line 29) provides a more extended version and has extra explanation. So, it's suggested to replace line 37 with the command on line 29
```suggestion
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~
+   In most cases, both `-B` and `-S` options are skipped with the assumption that
+  `<path-to-build>` is the current folder and `<path-to-source>` is the one
+   higher in the tree:
+   ~~~sh
+   cd <path-to-onednn-source>
+   mkdir -p build ; cd build
+   cmake .. [<options>]
```
Marina-Zubova, line 37: replacing the command
```suggestion
   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
 
-- `G` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
-  `Ninja`, `"Unix Makefiles"`).
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
+   ~~~
+   In most cases, both `-B` and `-S` options are skipped with the assumption that
+  `<path-to-build>` is the current folder and `<path-to-source>` is the one
```
Marina-Zubova: Provide examples to avoid confusion with directories
```suggestion
  `<path-to-build>` is the current folder (usually ~/oneDNN/build for local installation) and `<path-to-source>` is the one
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   higher in the tree:
```
Marina-Zubova:
```suggestion
   higher in the tree (usually ~/oneDNN):
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~sh
+   cd <path-to-onednn-source>
+   mkdir -p build ; cd build
+   cmake .. [<options>]
+   ~~~
+   The following are a few useful options defined by CMake:
```
Marina-Zubova, line 44: CMake options should be expanded with CPU/GPU runtime arguments, they should be clearly listed and explained.

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
 
-- `CMAKE_INSTALL_PREFIX` to control the library installation location.
+   - `G` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
```
Marina-Zubova: for more clarity on syntax
```suggestion
   - `-G <generator-name>` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+    `Ninja`, `"Unix Makefiles"`).
 
-- `CMAKE_BUILD_TYPE` to select between build type (`Release`, `Debug`,
-  `RelWithDebInfo`).
+   - `CMAKE_INSTALL_PREFIX` to control the library installation location.
```
Marina-Zubova: for more clarity on syntax
```suggestion
   - `-DCMAKE_INSTALL_PREFIX=<path>` to control the library installation location.
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
 
-- `CMAKE_PREFIX_PATH` to specify directories to be searched for the
-  dependencies located at non-standard locations.
+   - `CMAKE_BUILD_TYPE` to select between build type (`Release`, `Debug`,
```
Marina-Zubova:
```suggestion
   - `-DCMAKE_BUILD_TYPE=<build-type>` to select between build type (`Release`, `Debug`,
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+    `RelWithDebInfo`).
 
-See @ref dev_guide_build_options for detailed description of build-time
-configuration options defined by oneDNN.
+   - `CMAKE_PREFIX_PATH` to specify directories to be searched for the
```
Marina-Zubova:
```suggestion
   - `-DCMAKE_PREFIX_PATH=<path>` to specify directories to be searched for the
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+    dependencies located at non-standard locations.
```
Marina-Zubova, line 56: Suggestion to add runtime arguments
```suggestion
- `-DONEDNN_CPU_RUNTIME=<cpu-runtime> to specify CPU runtime (`OMP`, `TBB`, `SYCL` or other). OpenMP is the default runtime for CPU.
- `-DONEDNN_GPU_RUNTIME=<gpu-runtime> to specify GPU runtime (`OCL`, `SYCL`).
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
 
 ~~~sh
 git clone https://github.com/uxlfoundation/oneDNN.git
+cd oneDNN
 ~~~
 
 ## Build the Library
 
-### Set up the environment for the compiler
+You can quickly get started with building the library. The various stages involved in building the library are as follows:
```
Marina-Zubova: We need to clarify that this is a general build process description followed by specific build from source instructions for specific OS, compiler, etc. This way developers who need specific instructions will scroll down to find them under the general description.

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-### Linux/macOS
+   CMake provides a unified method for building a project, independent of the
+   generator or operating system used:
+   ~~~sh
+   cmake --build <path-to-build> [<options>]
+   ~~~
+   Full list of options can be found [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project).
 
-#### GCC, Clang, or Intel oneAPI DPC++/C++ Compiler
+You can build the library on Linux, macOS, or Windows using the compiler of your choice.
```
Marina-Zubova, line 73: Need to add - When the library is built, make sure to proceed with 'Validate the build'(relative link) and 'Install library'(relative link) sections.

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-# export CC=gcc
-# export CXX=g++
+#### Use GCC, Clang, or Intel oneAPI DPC++/C++ Compiler
 
-# Uncomment the following lines to build with Clang
-# export CC=clang
-# export CXX=clang++
+1. Set up the environment for the compiler
+   ~~~sh
+   # Uncomment the following lines to build with GCC
```
Marina-Zubova, line 81: Assuming we will have 3 tabs here for 3 compilers soon, so keeping this code snippet as it is for now, even if there are usability issues identified.

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake ..
+   ~~~
 
-- Build the library
+3. Build the library
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
+   # compilation. Parallelization can be controlled with:
```
Marina-Zubova: the comment in the code snippet looks weird, suggest moving it to under the command
```suggestion
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake ..
+   ~~~
 
-- Build the library
+3. Build the library
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
```
Marina-Zubova:
```suggestion
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-cmake ..
-~~~
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake ..
+   ~~~
 
-- Build the library
+3. Build the library
```
Marina-Zubova, line 100:
```suggestion
3. Build the library

To speed up the build process, multi-threaded compilation is recommended. You can control the number of parallel jobs using the `--parallel` option.

**For Linux:**
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
+   # compilation. Parallelization can be controlled with:
+   cmake --build . --parallel $(nproc)
+   ~~~
```
Marina-Zubova, line 106: The command above is only working on Linux, so we need one for macOS as well:
```suggestion
**For macOS:**
~~~sh
cmake --build . --parallel $(sysctl -n hw.ncpu)
~~~
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-cmake --build . --parallel $(nproc)
-~~~
+#### Use Intel oneAPI DPC++/C++ Compiler with SYCL runtime
 
-#### Intel oneAPI DPC++/C++ Compiler with SYCL runtime
+1. Set up the environment for the compiler
 
-- Set up the environment for the compiler
+   Intel oneAPI DPC++/C++ Compiler uses the `setvars.sh` script to set all the
+   required variables. The command below assumes you installed to the default
```
Marina-Zubova:
```suggestion
   required variables. The command below assumes you installed the compiler to the default
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~
+
+   @note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
+
+3. Build the library
+   ~~~sh
```
Marina-Zubova, line 144: Provide the same description for this step as we have in the previous section (line 91). Provide the command for macOS as well

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-
-~~~bat
-mkdir build
-cd build
-
-cmake .. -G Ninja ^
+   ~~~
+
+3. Build the library
+   ~~~sh
```
Marina-Zubova, line 172: Provide the same description for this step as we have in the previous section (line 91). Provide the command for macOS as well

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   export CXX=g++
+   ~~~
+
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_AARCH64_USE_ACL=ON
+   ~~~
+
+3. Build the library
```
Marina-Zubova, line 200: Provide the same description for this step as we have in the previous section (line 91). Provide the command for macOS as well

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   cd build
+   cmake .. -G "Visual Studio 17 2022"
+   ~~~
+
+3. Build the library
+   ~~~bat
+   cmake --build . --config=Release
+   ~~~
+
+   @note Currently, the oneDNN build system has limited support for multi-config
```
Marina-Zubova: The note on the lines 204-213 should be clarified and probably rewritten. Maybe it should be separated into 2 different notes

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   generators. Build configuration is based on the `CMAKE_BUILD_TYPE` option
+   (`Release` by default), and CMake must be rerun from scratch every time
+   the build type changes to apply the new build configuration. You can choose
+   a specific build type with the `--config` option (the solution file supports
+   both `Debug` and `Release` builds), but it must refer to the same build type
+   (`Release`, `Debug`, etc.) as selected with the `CMAKE_BUILD_TYPE` option.
+
+   @note You can also open `oneDNN.sln` to build the project from the
```
Marina-Zubova: Clarify that this is another option for step 3 Build the library. Maybe we should move this note higher and make it step Option B instead of keeping it in the note.   
```suggestion
   @note Alternatively, you can open `oneDNN.sln` to build the project from the
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   cmake ..
+   ~~~
 
-- Build the library
+3. Build the library
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
+   # compilation. Parallelization can be controlled with:
+   cmake --build . --parallel $(nproc)
+   ~~~
```
Marina-Zubova, line 105: Proceed with 'Verify the Build' (relative link)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
+
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
+
```
Marina-Zubova, line 147: Proceed with 'Verify the Build' (relative link)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-cd build
-
-cmake .. -G Ninja ^
+   ~~~
+
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
+
```
Marina-Zubova, line 175: Proceed with 'Verify the Build' (relative link)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_AARCH64_USE_ACL=ON
+   ~~~
+
+3. Build the library
+   ~~~sh
+   cmake --build . --parallel $(nproc)
+   ~~~
+
```
Marina-Zubova, line 206: Proceed with 'Verify the Build' (relative link)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   generators. Build configuration is based on the `CMAKE_BUILD_TYPE` option
+   (`Release` by default), and CMake must be rerun from scratch every time
+   the build type changes to apply the new build configuration. You can choose
+   a specific build type with the `--config` option (the solution file supports
+   both `Debug` and `Release` builds), but it must refer to the same build type
+   (`Release`, `Debug`, etc.) as selected with the `CMAKE_BUILD_TYPE` option.
+
+   @note You can also open `oneDNN.sln` to build the project from the
+   Microsoft Visual Studio IDE.
+
```
Marina-Zubova: Proceed with 'Verify the Build' (relative link)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~
+   or open `Intel oneAPI Commmand Prompt` from start menu instead.
+
+2. Generate the build system
+   ~~~bat
+   mkdir build
+   cd build
+
+   cmake .. -G Ninja ^
          -DDNNL_CPU_RUNTIME=SYCL ^
```
Marina-Zubova: update the prefix
```suggestion
         -DONEDNN_CPU_RUNTIME=SYCL ^
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
          -DDNNL_GPU_RUNTIME=SYCL
```
Marina-Zubova:
```suggestion
         -DONEDNN_GPU_RUNTIME=SYCL
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+   ~~~sh
+   export ACL_ROOT_DIR=<path/to/ComputeLibrary>
+   export CC=gcc
+   export CXX=g++
+   ~~~
+
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_AARCH64_USE_ACL=ON
```
Marina-Zubova:
```suggestion
   cmake .. -DONEDNN_AARCH64_USE_ACL=ON
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
-- Generate the build system
-
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DDNNL_CPU_RUNTIME=SYCL \
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_CPU_RUNTIME=SYCL \
          -DDNNL_GPU_RUNTIME=SYCL
```
Marina-Zubova:
```suggestion
         -DONEDNN_GPU_RUNTIME=SYCL
```

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
 
-- Generate the build system
-
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DDNNL_CPU_RUNTIME=SYCL \
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DDNNL_CPU_RUNTIME=SYCL \
```
Marina-Zubova:
```suggestion
   cmake .. -DONEDNN_CPU_RUNTIME=SYCL \
```

Marina-Zubova (changes requested):
This is the first round of DX review, left comments and also suggestions where possible

doc/build/build.md
ranukund: I thought the same and wanted to tackle in a separate PR. We need enough context for setting up the compiler when we move the prerequisites.

doc/build/build.md
ranukund: This sentence is added under Build the Library option. From the right hand side navigation menu, it should be understandable what the next step is. The tabbed approach can be implemented to enhance the usability further which will reduce the page scroll time and the right hand navigation will make much more sense. Alternatively, we could number the steps as follows:

Step 1. Download the source code
Step 2: Build the Library 
Step 3: Validate the Build
Step 4: (Optional) Build Documentation
Step 5: Install the Library

doc/build/build.md
ranukund: As mentioned earlier, we can enhance usability by implementing tabbed approach.

doc/build/build.md
ranukund: As mentioned earlier, let's avoid repeating the same information at every stage.

ranukund:
> This is the first round of DX review, left comments and also suggestions where possible

A lot of comments are around the commands. Adding @spalicki for review.

doc/build/build.md
ranukund: Let's not repeat the same information at every stage. Let's wait for the tabbed approach to be implemented and the guide can be enhanced further with subsequent PRs.

doc/build/build.md
ranukund: Let's not repeat the same information at every stage. Let's wait for the tabbed approach to be implemented and the guide can be enhanced further with subsequent PRs.

doc/build/build.md
ranukund: Let's not repeat the same information at every stage. Let's wait for the tabbed approach to be implemented and the guide can be enhanced further with subsequent PRs.

doc/build/build.md
ranukund: We can have a bulleted list for enhanced clarity.

doc/build/build.md
ranukund: Let's not repeat the same information at every stage. Let's wait for the tabbed approach to be implemented and the guide can be enhanced further with subsequent PRs.

doc/build/build.md
spalicki: I do not think it is usual to have onDNN in home folder like that, also `~` would work on Linux/macOS and not Windows.

doc/build/build.md
spalicki: Again, not usual.

doc/build/build.md
spalicki: `cmake -B <path-to-build> [-S <path-to-source>] [<options>]` is a generic command `cmake ..` (skipping `-S` and `-B`) is one that overwhelming majority of people will use. If we need to use 1 we should stick with `cmake ..` and add `-S` and `-B` to `useful options` section.

doc/build/build.md
spalicki:
```
mkdir -p build ; cd build
cmake ..
```
Is the way to go.

doc/build/build.md
spalicki: This section defines CMake options not one defined by oneDNN, those are linked below. We should most likely then move the oneDNN options up front.

doc/build/build.md
spalicki: Those are linked below. But maybe we should define most common `ONEDNN_*` ones up top (`ONEDNN_CPU_RUNTIME`, `ONEDNN_GPU_RUNTIME`, `ONEDNN_TEST_SET`, etc.)

doc/build/build.md
```diff
@@ -8,264 +9,256 @@ or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
...
+2. Generate the build system
+   ~~~bat
+   mkdir build
+   cd build
+   cmake .. -G "Visual Studio 17 2022"
+   ~~~
+
+3. Build the library
+   ~~~bat
+   cmake --build . --config=Release
```
spalicki: If we want to be consistent, we can add:
```
cmake --build . --config=Release --parallel %NUMBER_OF_PROCESSORS%
```

spalicki:
> > This is the first round of DX review, left comments and also suggestions where possible
> 
> A lot of comments are around the commands. Adding @spalicki for review.

I have added my comments to existing review.

doc/build/build.md
Marina-Zubova: it can be replaced with any usual path to oneDNN build and source that oneDNN engineers provide. Do you have any suggestions?

doc/build/build.md
Marina-Zubova: my comment above where I was explaining the issue has been removed for some reason. The idea was that it's confusing to see 2 cmake commands here while you only need to run it once

doc/build/build.md
Marina-Zubova: To support developers, voting for listing of the most common ONEDNN cmake arguments with most common values. If more details are needed, developer can refer to Build option doc, but the most common arguments should be called out

doc/build/build.md
Marina-Zubova: not sure how tabs can resolve this, do we know how the document will look like after applying tabs? I'd like to understand better what will change

ranukund: Until tabs are implemented, we can add numbering for each step as mentioned above.

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+   (`Release` by default), and CMake must be rerun from scratch every time
+   the build type changes to apply the new build configuration. You can choose
+   a specific build type with the `--config` option (the solution file supports
+   both `Debug` and `Release` builds), but it must refer to the same build type
+   (`Release`, `Debug`, etc.) as selected with the `CMAKE_BUILD_TYPE` option.
+
+   @note Alternatively, you can open `oneDNN.sln` to build the project from the
+   Microsoft Visual Studio IDE.
+
+#### Use Intel oneAPI DPC++/C++ Compiler with SYCL Runtime
```
mzhukova, line 254: This section is misplaced somehow, as the result in the rendered document the instructions are displayed as continuation of steps for previous compiler and a new warning appears when building doc:
```
build/reference/rst/dev_guide_build.rst:242: WARNING: Unexpected section title.

Use Intel oneAPI DPC++/C++ Compiler with SYCL Runtime
+++++++++++++++++++++++++++++++++++++++++++++++++++++
```

please take a look..

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
 
-## Download the Source Code
+## Step 1. Download the Source Code
 
 Download [oneDNN source code](https://github.com/uxlfoundation/oneDNN/archive/main.zip)
 or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
 
 ~~~sh
 git clone https://github.com/uxlfoundation/oneDNN.git
+cd oneDNN
```
mzhukova, line 17: I suggest removing this `cd oneDNN` since below (in a section about generating build system) we would have `cd <path-to-source>`

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
 Build from Source {#dev_guide_build}
 ====================================
+You can install the oneDNN library using the source distribution.
```
mzhukova:
```suggestion
You can build and install the oneDNN library using the source distribution.
```

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
 ~~~sh
 git clone https://github.com/uxlfoundation/oneDNN.git
+cd oneDNN
 ~~~
 
-## Build the Library
+## Step 2. Build the Library
 
-### Set up the environment for the compiler
+### Prerequisites
```
mzhukova: (minor) all other section titles sound like an action, so maybe "Resolve Prerequisites"?

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
-~~~
-
-- Generate the build system
-
-~~~sh
-mkdir -p build ; cd build
-cmake .. -DCMAKE_SYSTEM_NAME=Linux \
+1. Set up the environment for the compiler
+
+Configure your operating system's
```
mzhukova: this list is currently rendered as 1, 1, 2 instead of 1, 2, 3. Could you please fix (I think the indentation is the root cause)?

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
 Download [oneDNN source code](https://github.com/uxlfoundation/oneDNN/archive/main.zip)
 or clone [the repository](https://github.com/uxlfoundation/oneDNN.git).
 
 ~~~sh
 git clone https://github.com/uxlfoundation/oneDNN.git
+cd oneDNN
 ~~~
 
-## Build the Library
+## Step 2. Build the Library
```
mzhukova, line 20: The new structure of "Step 2: Build the Library" section is confusing.
There are three sub-sections: "Prerequisites", "Build on Linux/macOS", and "Build on Windows". However, the "Prerequisites" section contains only a single sentence about resolving requirements, and the rest is generic instructions that are later tailored for different compilers in the Linux/macOS/Windows sections.

"Generating Build system" and "Building the library" sub-steps do not really belong under "Prerequisites".
Perhaps we could simply rename this section to something like ~Generic instructions, so that we get the following structure:
sub-section 1: setup environment -> generic step 1 -> generic step 2 -> generic step 3
sub-section 2: Linux/macOS specifics
sub-section 3: Windows specifics

Additionally, satisfying the [system requirements](https://github.com/uxlfoundation/oneDNN#requirements-for-building-from-source) (i.e., ensuring the supported OS, compiler, and CMake version) could be listed as "Step 0", since if the system does not meet these requirements, building the library is not possible. Then only compiler/OS-specific details could be included in the ~environment setup sub-section.

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+   compilation. Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
+   ~~~sh
+   # Some generators, like Unix Makefiles, might default to single-threaded
+   # compilation. Parallelization can be controlled with:
+   cmake --build . --parallel $(nproc)
+   ~~~
+
+   **For macOS:**
+     ~~~sh
```
mzhukova: This is not rendered correctly right now, please fix

mzhukova: also, I think the analogue to `nproc` would be `hw.logicalcpu`, but for simplicity, I would actually suggest doing something like `--parallel <njobs>` or similar (here and everywhere linux/macos-related below), so that we have common note for linux/macos

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake ..
+   ~~~
+
+3. Build the library
+
+   Generators such as Unix Makefiles might default to single-threaded
```
mzhukova: seems like unnecessary repetition, since it is already present as a comment in the cmake call below, so we should keep one or the other

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
+
+3. Build the library
+   ~~~bat
+   cmake --build .
+   ~~~
+
+## Step 4. Validate the Build
```
mzhukova: Validating a build also could be marked as optional..

I think one way of reorganizing this page would be to mark it as optional, and the other way is to maybe move "Step 6: Install the Library" to be Step 4 (actually 3, right?), and then have leave Building Documentation and Validating the Build sections at the end, without any numeration.

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+
+After building the library, you can run a predefined test set using: >>
```
mzhukova:
```suggestion
After building the library, you can run a predefined test set using:
```

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
-~~~bat
-cmake --build .
-~~~
-
-## Validate the Build
-
-After building the library, you can run a predefined test set using:
 ~~~sh
 ctest
 ~~~
```
mzhukova, line 302: I know that you didn't touch the line below, but could you please fix:
"The [`ONEDNN_TEST_SET`](https://uxlfoundation.github.io/oneDNN/dev_guide_build_options.html#onednn-test-set)
build option set during the build configuration determines determines the scope" so that the link displays nicely and remove "determines" duplication?

doc/build/build.md
```diff
@@ -1,271 +1,278 @@
...
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
+
+3. Build the library
+   ~~~bat
+   cmake --build .
+   ~~~
+
```
mzhukova, line 295: there is no Step 3...

doc/build/build.md
mzhukova: I agree with @spalicki, I think we should leave the simple / common one and add -B, -S to the list of useful options (especially since everywhere below we're using that one)

doc/build/build.md
mzhukova: There is no usual path, since this is building from source repo that could have been cloned to any random location on the system, I don't think that "current folder" and "one folder higher" could create any confusion tbh, so probably no need for such clarification at all..

doc/build/build.md
mzhukova: I think including all the oneDNN options here in addition to CMake ones would crowd page way too much, we have a Build Options page for a reason (it is really big table!)

doc/build/build.md
mzhukova: tbh doesn't make sense to me, the most common build configuration of the project is probably what defined by default values of all build options, so you actually don't need to specifically call them out, because this is what you get by not specifying anything, right..?

doc/build/build.md
vpirogov: Maybe for the sake of documentation we can assume that all actions happen in the folder where oneDNN was cloned?

doc/build/build.md
ranukund: I see that introducing Prerequisites can be problematic. I plan to remove the header and add just the sentence about ensuring the software dependencies are in place. I usually look at the rendered output before committing changes to a PR. However, this time around, the commit got accidentally pushed.

doc/build/build.md
ranukund: fixed in my next commit

doc/build/build.md
ranukund: I see this is confusing. As mentioned in my earlier comment. I plan to remove the header and add just the sentence about ensuring the software dependencies are in place.

doc/build/build.md
ranukund: - I understand Validating the Build is an optional step. But wouldn't the user like to validate the build right after building the library? Can we integrate the validation of the build inside the "Build the Library" step and call it out as optional?

  Step 2: Build the Library
          2.1 Build on Linux/macOS
          2.2 Build on Windows
          2.3 Validate the Build

- Build documentation is optional but is it tied to Installation of the library? The verbiage under the [install section]   (https://uxlfoundation.github.io/oneDNN/dev_guide_build.html#install-library) is as follows:

     Install the library, headers, and documentation

   cmake --build . --target install

ranukund:
> Still a few issues to address, and a bit of polishing would help. Before requesting a review next time, please take a moment to visually check the rendered pages. Thanks for improving the structure!

Thanks, Maria for your review. I usually look at the rendered output before pushing commits however, this time around the commit got accidentally pushed.

doc/build/build.md
ranukund: Fixed in my next commit

doc/build/build.md
ranukund: From my local build, the right hand side navigation menu looks as follows: 

image

doc/build/build.md
ranukund: I've addressed this, please have a look.

doc/build/build.md
ranukund: fixed in the next commit, thanks

doc/build/build.md
ranukund: fixed, thanks

doc/build/build.md
vpirogov: This looks like common practice:
* https://github.com/pytorch/pytorch?tab=readme-ov-file#get-the-pytorch-source
* https://www.tensorflow.org/install/source#download_the_tensorflow_source_code

doc/build/build.md
mzhukova: double checked after latest commit, but warning still exists and section is still misplaced

doc/build/build.md
mzhukova: I wonder if instead of having "Multi-threaded compilation is recommended for a faster build process. Use the --parallel option to specify the number of parallel jobs." in each sub-section for different compilers, we just have this note once in ~generic section for better readability? (but keep `--parallel` everywhere anyway)

doc/build/build.md
mzhukova: >But wouldn't the user like to validate the build right after building the library?

User might (or not) want to, but they *don't have to* do validation in order to build or install a library, that was my point. 
And therefore it probably makes sense to also mark it as optional or separate it from the other steps (if we're doing this to documentation step).

> Build documentation is optional but is it tied to Installation of the library?

Installation is a separate step that is basically copying stuff to a dedicated folder, so if you have reference documentation built already, then on installation step it would be copied to <install-folder>/share/doc/..., otherwise (if documentation was not built previously) doc folder would only include README, not the complete reference guide AFAIK.

doc/build/build.md
spalicki: I would leave it and assume that every other command is executed inside `oneDNN`, then remove the `cd <path-to-source>` below.

doc/build/build.md
spalicki: Not all just 3/4 most common ones CPU/GPU/TEST and then point to the whole table?

doc/build/build.md
ranukund: I'm not sure what's causing the issue. I need to investigate. I saw the warning still exists.

doc/build/build.md
spalicki: The default for `ONEDNN_GPU_RUNTIME` is `NONE`, so people might be confused why nothing is working on GPU when built with just `cmake ..`.

doc/build/build.md
mzhukova: that's fine too, I'm not attached to my particular suggestion, the point was to remove duplication, since it makes things confusing (`cd folder` then `cd some other folder`)

doc/build/build.md
mzhukova: right, but this option is explicitly set in "Intel oneAPI DPC++/C++ Compiler with SYCL runtime" section, where it makes sense, I'm not sure there is a need to clutter ~generic section

doc/build/build.md
mzhukova: oh sure, you just re-requested review, so I assumed everything got addressed

spalicki:
@ranukund If we want to make it easy to parse for new developers, how about we streamline the documentation reorganize the structure and remove all the duplicated commands and notes? Something similar to #3395

doc/build/build.md
Marina-Zubova: @spalicki @mzhukova  Correct me if I'm wrong, this section of build from source aims to provide the high-level overview of build from source process to provide expectations to developers and guide them to further sections. We have specific executable commands in further section based on the OS, compiler and runtime. But here we are providing just an overview, right? Then why would we have `cmake ..` here, isn't it a duplication of what we already have in further sections?

doc/build/build.md
Marina-Zubova: okay, let's address the prerequisites in a separate PR. It makes sense to still keep the step 1. Set up the environment for the compiler here for consistency. But instead of pointing to prerequisites (this will be moved in the future), we can inform about that on this step you may need to set environment variables specific to your compiler.

doc/build/build.md
Marina-Zubova: If we don't have usual installation and build paths, let's not mention it then. We do have 'cd oneDNN' command included (at the very top of the document), so assuming developer is in this folder

doc/build/build.md
Marina-Zubova: looks much better with the numbering!

doc/build/build.md
Marina-Zubova: the build from source doc describes only a few build from source scenarios, assuming developer can customize the commands for other cases. The list of supported CPU/GPU runtimes seems to be something specific to oneDNN, other libraries may have other supported runtimes (e g cuda). 
So, to support developers, it makes sense to call out the most common runtimes, clarify default runtimes, maybe list some other helpful oneDNN-specific arguments (like ONEDNN_TEST_SET). At the same time, let's try to keep it minimal to not take a lot of space

doc/build/build.md
ranukund: Do I have it right that there are two sets of options?

Category 1: Options that are **defined by CMake** such as CMAKE_BUILD_TYPE, CMAKE_INSTALL_PREFIX etc.
Category 2: Options that are **created and customized for oneDNN** such as ONEDNN_CPU_RUNTIME etc?

If yes, can we move category 1 to Build Options page to avoid cluttering the build from source guide?

mzhukova: CMake variables are not oneDNN build options (== oneDNN supported functionality), so I think we should leave it up to CMake documentation to list them. Here we just mention a few frequently used ones for user convenience.

We already provide a link to CMake documentation at the beginning of this section, but we could also provide link to CMake options (https://cmake.org/cmake/help/latest/manual/cmake.1.html#options) and CMake variables (https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html) if you think that is necessary..

doc/build/build.md
mzhukova: hi @Marina-Zubova, I think there is some kind of misunderstanding..

this is a somewhat generic command -- 
```
cmake -B <path-to-build> [-S <path-to-source>] [<options>]
```

the following one does _exactly the same_, and is presented as alternative simplified version based on assumptions wrt build path and src path (that is what comment right before the command says) -- 
```
mkdir -p build ; cd build
cmake .. [<options>]
```

Stefan and I suggested to keep only 2nd one and incorporate `-B, -S` into "a few useful options defined by CMake:..." list, that is following right after, to improve readability and since for most people the assumption wrt paths would be true.

Now, if by duplication of further sections you mean having the same steps repeated in sections specific to a compiler/OS, then it is expected. First we provide generic instructions: user could stop right there, check build option and decide for themselves how to proceed. In addition, for _some_ common scenarios, we follow up with sub-sections and explicit instructions, since we do setup compiler and generate build system differently (e.g., for DPC++ compiler w SYCP runtime we assume that you actually need to build with CPU/GPU SYCL runtime support, etc.), so the user could navigate right to the section that reflects their needs better.

I hope it makes more sense, please let me know if there is still confusion on this part.

ranukund: I think to simplify, we can have one version and not showcase both the alternatives, which might be confusing. Therefore, could we have the simplified version as follows, please advise:

## 2. Generate the build system
The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
Use the following command to generate a build system:
~~~sh
mkdir -p build ; cd build
cmake .. [<options>]
~~~
The following are a few useful options defined by CMake:

   - `-G <generator-name>` to specify build system generator (e.g. `"Visual Studio 17 2022"`,
    `Ninja`, `"Unix Makefiles"`).

   - `-DCMAKE_INSTALL_PREFIX=<path>` to control the library installation location.

   - `-DCMAKE_BUILD_TYPE=<build-type>` to select between build type (`Release`, `Debug`,
    `RelWithDebInfo`).

   - `-DCMAKE_PREFIX_PATH=<path>` to specify directories to be searched for the
    dependencies located at non-standard locations.

   See @ref dev_guide_build_options for detailed description of build-time
   configuration options defined by oneDNN.

doc/build/build.md
ranukund: fixed, thanks

doc/build/build.md
ranukund: Added.

doc/build/build.md
ranukund: Added.

doc/build/build.md
ranukund: fixed, @mzhukova could you please check?

doc/build/build.md
ranukund: Marked the "Validate the Build" as optional step.

ranukund:
Should I handle this PR in phases, cherry-pick the latest stable version, and publish the approved changes? In the next PR, I can address the remaining comments if required. @mzhukova @spalicki - please let me know what suits.

doc/build/build.md
```diff
@@ -1,276 +1,305 @@
...
+   # Set Intel oneAPI DPC++/C++ Compiler as default C and C++ compilers
+   export CC=icx
+   export CXX=icpx
+   ~~~
+
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DONEDNN_CPU_RUNTIME=SYCL \
+         -DONEDNN_GPU_RUNTIME=SYCL
```
mzhukova: (minor) fix alignment?

doc/build/build.md
```diff
@@ -294,26 +323,30 @@ When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI toolk
...
-## Install library
+## Step 5. Install the Library
 
-Install the library, headers, and documentation
-~~~sh
-cmake --build . --target install
-~~~
+Install the library, headers, and documentation.
+
+@note
```
mzhukova: (minor) seems like everywhere else we have a note after the actual command, not before

doc/build/build.md
mzhukova: I prefer this option, but the suggestion was to also add `-B`, `-S` to the "The following are a few useful options defined by CMake: ..." list.

doc/build/build.md
```diff
@@ -1,276 +1,305 @@
...
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   cmake -B <path-to-build> [-S <path-to-source>] [<options>]
+   ~~~
+   In most cases, both `-B` and `-S` options are skipped with the assumption that
+  `<path-to-build>` is the current folder and `<path-to-source>` is the one
+   higher in the directory tree:
+   ~~~sh
+   cd <path-to-source>
```
mzhukova: need to get rid of `cd` here since we decided to have `cd oneDNN` at the top (cloning) part

doc/build/build.md
mzhukova: closing this thread as duplication

doc/build/build.md
mzhukova: closing this thread as a duplication

mzhukova:
> Should I handle this PR in phases, cherry-pick the latest stable version, and publish the approved changes? In the next PR, I can address the remaining comments if required. @mzhukova @spalicki - please let me know what suits.

hi @ranukund, I think there are just a few threads + minor comments left, it would be great to have them resolved before merging so to not loose track and also to keep Build page update self-contained. Thanks!

doc/build/build.md
ranukund: I think that we cannot provide the -B or -S in the `<options>` placeholder in the command. In the list below, we are providing the options that can be used in the placeholder <options>. I've now added a note below the command for completeness. Please have a look.

doc/build/build.md
ranukund: The usability study participants missed the information about having administrator privileges and running using the sudo command when installing in the default directory. Therefore, had to move this content up before the command. I also think we need not use the note admonition.

ranukund:
> > Should I handle this PR in phases, cherry-pick the latest stable version, and publish the approved changes? In the next PR, I can address the remaining comments if required. @mzhukova @spalicki - please let me know what suits.
> 
> hi @ranukund, I think there are just a few threads + minor comments left, it would be great to have them resolved before merging so to not loose track and also to keep Build page update self-contained. Thanks!

Hi @mzhukova I've addressed your comments.

doc/build/build.md
ranukund: - What's the decision here? I like the idea of adding the note in the generic section to avoid repetation at every stage. 
- Should I append `--parallel %NUMBER_OF_PROCESSORS%` at line 236?

doc/build/build.md
Marina-Zubova: I'm fine with current solution, thanks!

doc/build/build.md
Marina-Zubova: Seems like we agreed on CMake arguments, but what about ONEDNN ones, are we going to provide the most common ones in this guide to support new users?

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
-~~~sh
-# Some generators, like Unix Makefiles, might default to single-threaded
-# compilation. Parallelization can be controlled with:
-cmake --build . --parallel $(nproc)
-~~~
+3. Build the library
 
-#### Intel oneAPI DPC++/C++ Compiler with SYCL runtime
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
```
Marina-Zubova: adding missing text
```suggestion
   Use the `--parallel` option to specify the number of parallel jobs.
   **For Linux:**
```

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
+   @note Open-source version of oneAPI DPC++ Compiler does not have the icx driver,
+   use clang/clang++ instead. Open-source version of oneAPI DPC++ Compiler may not
+   contain OpenCL runtime. In this case, you can use `OPENCLROOT` CMake option or
+   environment variable of the same name to specify path to the OpenCL runtime if
+   it is installed in a custom location.
+
+3. Build the library
+
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
```
Marina-Zubova:
```suggestion
   Use the `--parallel` option to specify the number of parallel jobs.
   **For Linux:**
```

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
-
-~~~bat
-cmake --build . --config=Release
-~~~
+   ~~~
+
+3. Build the library
+   
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
```
Marina-Zubova:
```suggestion
   Use the `--parallel` option to specify the number of parallel jobs.
   **Fir Linux:**
```

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
+2. Generate the build system
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. -DONEDNN_AARCH64_USE_ACL=ON
+   ~~~
+
+3. Build the library
+
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
```
Marina-Zubova:
```suggestion
   Use the `--parallel` option to specify the number of parallel jobs.
   **For Linux:**
```

doc/build/build.md
Marina-Zubova: will we be working on that in the next PR or in this one?

doc/build/build.md
Marina-Zubova: Adding more placeholders in the commands is not a good practice, we should avoid them where we can. 

Some usability study participants were not familiar with nproc variable. If we provide an incomplete command, they would need further assistance to make it working. Moreover, they may choose to ignore parallel argument because the command if incomplete, which my lead to system crush (issue that we fixed by adding nproc to the command).

doc/build/build.md
Marina-Zubova: If the note is overlooked in the generic section, it may lead to a critical issue while building the library. Seems like repetition is necessary here

doc/build/build.md
```diff
@@ -294,26 +321,29 @@ When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI toolk
...
 
-Install the library, headers, and documentation
 ~~~sh
 cmake --build . --target install
 ~~~
-The install directory is specified by the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)
-cmake variable. When installing in the default directory, the above command
-needs to be run with administrative privileges using `sudo` on Linux/Mac or a
-command prompt run as administrator on Windows.
+
```
Marina-Zubova, line 353: Add info on where to go next (Linking guide, Examples) - @ranukund  are we going to work on that in this PR or in the next one?

Marina-Zubova (changes requested):
Added a few more comments and minor suggestions. The document looks much better, we have a good progress!

doc/build/build.md
ranukund: We can address this in the next PR.

doc/build/build.md
ranukund: The suggestion here is to add the content around using the `-parallel` option in the generic section itself along with the adjacent text, and omit the repetition of adjacent text in each step. Therefore, in the generic section under step 3=, we can say:

Step 3. Build the library

CMake provides a unified method for building a project, independent of the generator or operating system used.

 Multi-threaded compilation is recommended for a faster build process.
 Use the `--parallel` option to specify the number of parallel jobs.

`cmake --build <path-to-build> --parallel<number of processors> [<options>]`

Full list of build options can be found [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project).

@mzhukova please confirm if this works.

Marina-Zubova: looks good to me, assuming we still keep similar messages in further specific sections

ranukund: Planning to omit this message in other sections, however, will retain the command such as, 

~~~sh
   cmake --build . --parallel $(nproc)
   ~~~

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
-~~~
+   The oneDNN build system is based on [CMake](https://cmake.org/cmake/help/latest/manual/cmake.1.html). 
+   Use the following command to generate a build system:
+   ~~~sh
+   mkdir -p build ; cd build
+   cmake .. [<options>]
+   ~~~
+   @note You can use `cmake -B <path-to-build> [-S <path-to-source>] [<options>]` to specify the following:
+   - -B <path-to-build>: Specify the path where the build files will be generated.
+   - -S <path-to-source>: Specify the path to the source directory containing the source files, dependencies, compiler options etc.
```
mzhukova:
```suggestion
   - `-B <path-to-build>`: Specify the path where the build files will be generated.
   - `-S <path-to-source>`: Specify the path to the source directory containing the source files, dependencies, compiler options etc.
```

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
+   CMake provides a unified method for building a project, independent of the
+   generator or operating system used.
 
-#### GCC targeting AArch64 on x64 host
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
 
-- Set up the environment for the compiler
+   ~~~sh
+   cmake --build <path-to-build> --parallel<numberofprocessors> [<options>]
```
mzhukova:
```suggestion
   cmake --build <path-to-build> --parallel <jobs> [<options>]
```
parallel doesn't necessarily have to go with number of processors, it could be any number, so I'd keep it generalized 
(I would actually prefer it that way everywhere below instead of NUMBER_OF_PROCESSORS and nproc, but not attached to it)

ranukund: addressed, thanks

ranukund:
@uxlfoundation/onednn-arch @uxlfoundation/onednn-doc I'll need one more approval to merge.

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
+   CMake provides a unified method for building a project, independent of the
+   generator or operating system used.
 
-#### GCC targeting AArch64 on x64 host
+   Multi-threaded compilation is recommended for a faster build process.
+   Use the `--parallel` option to specify the number of parallel jobs.
 
-- Set up the environment for the compiler
+   ~~~sh
+   cmake --build <path-to-build> --parallel<jobs> [<options>]
```
Marina-Zubova: missing space
```suggestion
   cmake --build <path-to-build> --parallel <jobs> [<options>]
```

doc/build/build.md
```diff
@@ -1,276 +1,303 @@
...
+
+#### Use Microsoft Visual C++ Compiler
+
+1. Set up the environment for the compiler
+
+   Microsoft Visual Studio uses the `VsDevCmd.bat` script to set all
+   required variables. The command below assumes you installed to the default
+   folder. If you customized the installation folder, `VsDevCmd.bat` is in your
+   custom folder.
+   ~~~bat
```
Marina-Zubova, line 222: The command should be visible on the page without a need to scroll right. @ranukund I suggest the following formatting, please check:
~~~sh
"C:\Program Files\Microsoft Visual Studio\2022\Professional\ ^
Common7\Tools\VsDevCmd.bat" ^
  -startdir=none ^
  -arch=x64 ^
  -host_arch=x64
~~~

doc/build/build.md
```diff
@@ -294,26 +321,29 @@ When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI toolk
 Make sure the correct oneDNN library is present in
```
Marina-Zubova: The call to action in this note is not clear enough, so it can create confusion or simply get ignored by new oneDNN users. To avoid that, let's work on improving the note in the next PR

ranukund: Let's work on this in the next PR

doc/build/build.md
```diff
@@ -289,31 +320,34 @@ Another useful option is `--output-on-failure`, which will print verbose output
 in case a test fails. Full set of options can be found [here](https://cmake.org/cmake/help/latest/manual/ctest.1.html).
 
 @warning
-When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI Toolkit,
-`LD_LIBRARY_PATH` is set to include the oneDNN library path from the installation.
-Make sure the correct oneDNN library is present in
-`LD_LIBRARY_PATH` by setting it explicitly if needed.
+When using the `/opt/intel/oneapi/setvars.sh` script from the Intel oneAPI toolkit,
```
mzhukova: IPLDT (linter) is currently unhappy about Toolkit -> toolkit change, if this one is indeed correct, I could update the hits.db after this PR is merged
@ranukund please let me know

ranukund: fixed, thanks. Changed this to Intel oneAPI Base Toolkit

PR #3381, opened 2025-05-30
LSTM regressions for OV framework

h-sadia:
MFDNN-13729: https://jira.devtools.intel.com/browse/MFDNN-13729

### DIFF
--- a/src/gpu/intel/ocl/rnn/utils.cpp
+++ b/src/gpu/intel/ocl/rnn/utils.cpp
@@ -183,9 +183,14 @@ void rnn_utils::init_rnn_conf(conf_t &rnn, const rnn_desc_t &rd,
 
         // For large enough k dimension, parallelization in external gemm
         // kernels is more performant.
-        int ideal_k_limit
-                = math::lcm((int)device_info.min_subgroup_size(), (int)rnn.sic);
-        dim_t k_limit = tail_dhc ? 50 : nstl::min(256, ideal_k_limit);
+        int min_k = nstl::min((int)rnn.sic, (int)rnn.slc);
+        int ideal_k = math::lcm((int)device_info.min_subgroup_size(), min_k);
+        if (rd.cell_kind == alg_kind::vanilla_lstm) {
+            min_k = (min_k <= 256) ? 160 : 256;
+        } else {
+            min_k = 256;
+        }
+        dim_t k_limit = tail_dhc ? 50 : nstl::min(min_k, ideal_k);
 
         // The fused gemm implementation assumes the dst channel dimension is
         // dense


src/gpu/intel/ocl/rnn/utils.cpp
```diff
@@ -183,9 +183,14 @@ void rnn_utils::init_rnn_conf(conf_t &rnn, const rnn_desc_t &rd,
 
         // For large enough k dimension, parallelization in external gemm
         // kernels is more performant.
-        int ideal_k_limit
-                = math::lcm((int)device_info.min_subgroup_size(), (int)rnn.sic);
-        dim_t k_limit = tail_dhc ? 50 : nstl::min(256, ideal_k_limit);
+        int min_k = nstl::min((int)rnn.sic, (int)rnn.slc);
+        int ideal_k = math::lcm((int)device_info.min_subgroup_size(), min_k);
+        if (rd.cell_kind == alg_kind::vanilla_lstm) {
+            min_k = (min_k <= 256) ? 160 : 256;
```
Simonsays095, line 189: nit: `min_k=256` will get pulled down to 160 instead of staying at 256.
```suggestion
            min_k = (min_k < 256) ? 160 : 256;
```

h-sadia: Yes, that is intended...performance issues with sic size 256.

Simonsays095: I'm not sure what `sic`/`slc` mean, but this logic seems odd to me. I can't reason why anything higher than 256 is good when set to 256, but if it's already at 256 then it needs to be lowered.
If your goal is to decrease `k_limit` for `vanilla_lstm` by making a cutoff where everything lower is set to `160`, maybe it needs to be higher than 256? Perhaps:
```suggestion
            const int k_cutoff = 512;
            min_k = (min_k < k_cutoff ) ? 160 : k_cutoff ;
```
This way `sic=256` gets pulled down as you want.

h-sadia: Not really. sic and slc are two iteration and layer counters we have for RNN. These numbers were gotten through experimentation, and are not random. So randomly throwing 512 in there is not a good technique here.

h-sadia: https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/ocl/rnn/grid.cl#L814 - This part of the code is where we are using k_limit. (#if CELL_COMP_ENABLED)

Simonsays095: This was just an example - instead of 512 you could do further experimentation to determine a better value. Alternatively - if you think the current implementation is better - could you explain why `sic=257->k_limit=256` but `sic=256->k_limit=160` is optimal?

h-sadia: Maybe I can word it better. A quick experimentation with various k_limits on DG2 machines will help you see that difference in time that comes with 160 (which was set earlier when cell_fusion was introduced by Roy for LSTM AND VANILLA RNN) to 256 and then to anything bigger than that gives regression for some LSTM cases. Its really about finding the sweet spot for cell_fusion - *a better value* is one that gives the most optimal result for all layers. There are multiple buffers that are involved in RNN from GEMM, post-gemm and pre gemm copy kernels, so its not a good idea to have a big k_limit. For regressions on DG2 this was the best range I could come up with. Frameworks are also getting their results back. Essentially 160 was the orignial k_limit set, I extended it to 256 for LBR_GRU, and since its harder for LSTM to sustain such a change I had to come up with this range.

rjoursler: @Simonsays095, the optimizations here are targeting scenarios where computation is limited by the rate at which GPU kernels can be dispatched to the GPU. This makes this a very difficult problem to solve in general as it implies the optimal implementation is a function of the CPU and GPU execution speed. What is going on here is that once the GEMM operations are large enough, the overhead from kernel dispatching is small enough to be "worthwhile", and we are finding this threshold via experimentation.

rjoursler (approved):

src/gpu/intel/ocl/rnn/utils.cpp
Simonsays095: I understand the basis of what's going on here and what's being optimized. Thank you for the history of this optimization @h-sadia, that does help clarify why these values are what they are.

My only concern is that this mapping `min_k= (min_k<= 256) ? 160 : 256` reduces `min_k` by at least 1 no matter what value it is, which made me think that there's likely some incorrect logic or room for improvement. I now understand that a more general form of this heuristic would be `min_k = (min_k <= cutoff) ? low : high`, so the two values of 256 correspond to two different heuristic quantities which happen to align. It's still surprising to me that these line up in this way, but it's just a nitpick in the end so I'm ok with it.

PR #3388, opened 2025-06-03
cpu: aarch64: extend brdgmm to support sve_128

karmeh01:
### Description
This PR extends the existing brdgmm depthwise convolution implementation to support sve_128, providing a faster implementation for 3D depthwise convolution and specific cases of 2D depthwise convolution.

### Speed-up in oneDNN:

**For 3D depthwise convolution:**
`OMP_NUM_THREADS=2 ./benchdnn --conv --mode=p --dt=f32 --alg=convolution_direct mb1_g128_ic128oc128_kd5kh1kw1_id32ih1iw1_od5oh1ow1_sd7sh1sw1_pd1ph0pw0_dd0dh0dw0n"dw_3d_strided_padding_1709"`:

gemm:ref - `avg(ms):0.0595099`
brdgmm_dw:sve_128 - `avg(ms):0.00197401`   **~30x speedup**     

**For 2D depthwise convolution:**
`OMP_NUM_THREADS=2 ./benchdnn --conv --dir=FWD_I --mode=p --dt=f32 --alg=convolution_direct <shape>`

Screenshot 2025-05-19 at 11 25 41

Shapes tested:
- `mb1_g128_ic128oc128_kh11kw1_ih55iw1_oh10ow1_sh5sw1_ph2pw0_dh0dw0n"dw_2d_strided_padding_1940_ph-bottom(-1)"`
- `mb1_g16_ic16oc16_kh11kw1_ih112iw1_oh108ow1_sh1sw1_ph6pw0_dh0dw0n"dw-conv_2d_unit-stride_padding_657"`
- `mb1_g512_ic512oc512_kh11kw1_ih6iw1_oh1ow1_sh1sw1_ph0pw0_dh0dw0n"dw_2d_unit-stride_no-padding_1031_ph-bottom(5)"`

### **Speed-up in Pytorch:** 

Depthwise convolution of shape 1x16x32x32x32
gemm:ref - 16.701 ms 
brdgmm_dw:sve_128 -  8.718 ms           **~1.91x faster**

Depthwise convolution of shape 2x32x64x64x64
gemm:ref - 659.357 ms 
brdgmm_dw:sve_128 - 293.193 ms           **~2.25x faster**

### Tests:
`./benchdnn --conv --batch=inputs/conv/test_conv_depthwise`

Modified `oneDNN/build/tests/benchdnn/inputs/conv/test_conv_depthwise` to add tests that check the brdgmm 2D depthwise convolution case


include/oneapi/dnnl/dnnl_types.h
```diff
@@ -737,6 +738,7 @@ typedef enum {
     dnnl_abCdef32c,
     dnnl_ABcd16a32b,
     dnnl_decbA8a,
+    dnnl_decbA4a,
```
dzarukin: New tags are not allowed in the middle of the list as they break ABI compatibility.
Please move the to the end of the list.

tests/benchdnn/inputs/conv/test_conv_depthwise
```diff
@@ -1,8 +1,8 @@
 # f32
 --reset
 --mb=2
---dir=FWD_D,BWD_D,BWD_WB --batch=shapes_mobilenet_dw
---dir=FWD_D,BWD_D,BWD_WB --batch=shapes_regression_dw
+--dir=FWD_D,BWD_D,BWD_WB,FWD_I --batch=shapes_mobilenet_dw
+--dir=FWD_D,BWD_D,BWD_WB,FWD_I --batch=shapes_regression_dw
```
dzarukin, line 5: `FWD_D` provides same level of coverage as `FWD_I`. Could you elaborate, please, why adding it?

karmeh01: brdgmm is used for 2D depthwise convolution when dir=FWD_I. With an incorrect implementation of brdgmm these tests still pass, so I added it to cover that case

dzarukin: Do you say that brdgmm is NOT used for `FWD_D`?

karmeh01: From what I can tell, it is **not** used for `FWD_D` for 2D, but is used for `FWD_D` for 3D. 

The logic is defined [here](https://github.com/uxlfoundation/oneDNN/blob/27b02be8a4ba735e0c464802c4f5253dbeaa616b/src/cpu/aarch64/jit_brdgmm_dw_conv.cpp#L179C5-L182C65)

jondea: If `FWD_D` is equivalent to `FWD_` for dw conv, @karmeh01 can we check if the tests pass if we change that check to just `is_fwd()`?

dzarukin: I don't know the exact-exact reason but it has something with changing the format of weights on training. The idea is that user specifies the exact format on forward_training and aligns it on their side with backward_weights, so that the tensor remains unchanged. Thus, `any_is_eligible` is `true` for inference only, where there's no such limitation: pass any, query the format, reorder once, reuse.

I guess I'm fine with testing both since x64 has same thing.

karmeh01: Changing the any_eligible condition to `const bool any_eligible = (cd.prop_kind == prop_kind::forward_inference || is_int8 || is_3d);` still passes all tests

karmeh01: > If `FWD_D` is equivalent to `FWD_` for dw conv, @karmeh01 can we check if the tests pass if we change that check to just `is_fwd()`?

Changing the any_eligible condition to `const bool any_eligible = (cd.prop_kind == prop_kind::forward || is_int8 || is_3d);` still passes all tests

dzarukin (approved):

jondea (approved):
LGTM! Thanks

aditew01 (approved):

PR #3391, opened 2025-06-04
graph: backend: dnnl: support conv + xnary + swish pattern

ElaineBao:
# Description

Fix MFDNN-10676.

### DIFF
--- a/src/graph/backend/dnnl/patterns/conv_post_ops.cpp
+++ b/src/graph/backend/dnnl/patterns/conv_post_ops.cpp
@@ -777,6 +777,55 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, fp_conv_post_ops)
         });
 
 /*
+              \   /
+              conv
+                |
+         unary/binary
+                |    \
+             sigmoid  |
+                |    /
+             multiply
+                |
+[unary/binary]*[0,MAX_REPETITION)
+                |
+*/
+DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(
+        dnnl, fp_conv_post_ops_swish_post_ops)
+        .set_priority(9.9f)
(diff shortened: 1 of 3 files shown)


src/graph/backend/dnnl/patterns/conv_post_ops.cpp
```diff
@@ -776,6 +776,55 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, fp_conv_post_ops)
...
+
+                    auto alt_graph = std::make_shared<pb_graph_t>();
+                    auto pop = alt_graph->append_alternation(
+                            get_unary_binary_ops());
+                    pop->allow_internal_inputs();
+                    alt_graph->create_input_port(0, pop, 0);
+                    alt_graph->create_output_port(0, pop, 0);
+
+                    pgraph->append_repetition(alt_graph, {0, 0}, 0,
+                            MAX_REPETITION,
```
rongzha1, line 821: do we need to limit post op here to MAX_REPETITION -2 to exclude xnary following conv and swish.

ElaineBao: As the maximum number of post operations supported in the library is 32, so I guess it's ok to use MAX_REPETITION here (which is 5)

rongzha1: got it. LGTM. 
better to align it with spec when we are free. 
spec: 0 to 4 Binary or Unary operations are supported in the epilogue subgraph https://uxlfoundation.github.io/oneDNN/dev_guide_graph_convolution_fusion_patterns.html

ShanSimu (approved):

dzarukin (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3393, opened 2025-06-04
rfcs: proposal for an asynchronous verbose mode

avmanerikar:
Link to rendered document: [[link](https://github.com/uxlfoundation/oneDNN/tree/amanerik/rfcs/async-verbose-mode/rfcs/20250527-async-verbose-mode)]
Link to PoC implementation: [[link](https://github.com/uxlfoundation/oneDNN/pull/5102)]

Addresses MFDNN-15108 and MFDNN-12088.

### DIFF
--- a/rfcs/20250527-async-verbose-mode/README.md
+++ b/rfcs/20250527-async-verbose-mode/README.md
@@ -1,0 +1,132 @@
+# Proposal for an Asynchronous Verbose Mode
+
+## Background and Motivation
+
+In the oneDNN verbose mode, the current approach to keeping track of kernel execution times relies on a host-side timing approach using `stream.wait()` calls. 
+While simple, this approach introduces the following drawbacks, especially in the context of GPU kernel profiling:
+- **Synchronization Overhead**: Measuring execution times on the host incurs synchronization costs when waiting for the GPU kernels to finish. For fast kernels, this overhead can also surpass the kernel execution time itself.
+- **Timing Bubbles**: Latency between sequential kernels further distorts the actual execution window, making it difficult to isolate pre-kernel timings.
+- **Blocking Semantics**: The use of `stream.wait()` is inherently blocking, which undermines performance in multi-stream or asynchronous execution environments.
+
+This results in a discrepancy between measured and true kernel times, limiting the utility of verbose logging data for accurate perfromance profiling.
+
+## Objectives
+The goal of this proposal is to introduce an *asynchronous verbose mode* that enables non-blocking, accurate tracking of kernel executions with minimal synchronization latencies. The new mechanism will be implemented with the following approach: 
+- Implement an event-polling mechanism that periodically checks for completed primitives without blocking execution.
+- Leverage device-measured timing data from SYCL/OpenCL events to ensure accurate profiling information.
+- Maintain compatibility with both regular execution and SYCL graph recording modes.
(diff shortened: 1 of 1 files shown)


avmanerikar:
The [PoC](https://github.com/uxlfoundation/oneDNN/pull/4187) implementation for asynchronous verbose profiling fails in the following scenario: 

The asynchronous callbacks responsible for calculating and printing profiling information rely on `stream_profiler_t` to access event execution times. When these callbacks are delayed beyond the point where the stream is destroyed, accessing the profiler becomes impossible and can potentially trigger a segmentation fault during timing computation.

The following example reproduces this scenario for OpenCL GPU runtimes:
```
ONEDNN_VERBOSE_USE_SYNC=0 ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn -v5 --engine=gpu --matmul --mode=R --repeats-per-prb=80000 128x128:128x128
```

The fix requires a non-blocking mechanism to retain the lifecycle of the profiler and its registered events until the callbacks complete verbose logging.

#### Candidate Solutions:
- **Pass a snapshot of profiler events in the callback payload** - The caveat is that this approach involves mutex-locked operations.
- **Defer stream destruction until callbacks are executed** - While this approach is blocking, the blocking occurs during stream destruction rather than during primitive execution.

vpirogov (approved):

rfcs/20250527-async-verbose-mode/README.md
```diff
@@ -0,0 +1,132 @@
...
+This design provides accurate execution timings, minimizes host-side interference, and keeps verbose logging compatible with both OpenCL and SYCL backends across all execution modes.
+
+
+## Proposal
+The new asynchronous verbose mechanism has the following design elements:
+
+### Event-Polling Based Architecture
+- A new `verbose_profiler_t` class manages a map of primitive execution events and their associated profiling data.
+- The profiler operates independently from the main execution thread, polling for event completion without blocking stream operations.
+- Each primitive is assigned a unique stamp, and its associated events are tracked until completion.
```
mzhukova: Is this aligned to the implementation?

rfcs/20250527-async-verbose-mode/README.md
```diff
@@ -0,0 +1,132 @@
...
+    // Graph recording compatibility (SYCL-specific)
+    status_t pause_recording();
+    status_t resume_recording();
+};
+```
+
+### New Verbose Profiler API
+A new verbose_profiler_t class provides the core asynchronous profiling functionality:
+```cpp
+struct verbose_profiler_t {
```
mzhukova, line 84: Since we're merging RFC post implementation, could you please update APIs to match what is in main currently? so that this RFC could be also used as reliable sort of documentation :)

mzhukova (commented):
LGTM, but could we please ensure this is aligned to what was implemented?

rfcs/20250527-async-verbose-mode/README.md
avmanerikar: Updated RFC to align with the implementation. Thanks!

rfcs/20250527-async-verbose-mode/README.md
avmanerikar: ditto

mzhukova (approved):

PR #3227, opened 2025-05-07
graph: backend: dnnl: relax restriction for compressed SDPA dispatch and improve validation

wzt1997:
## Description

Resolve MFDNN-12857. Graph API used to support compressed SDPA while group_num=1 with primitive ukernel only due to the capability and validation issue. The PR relaxed the restriction and fixed some issues in benchdnn graph driver:
1. Allow f16 mask for compressed SDPA
2. For input displacement with u8 weight, benchdnn graph will generate input with s8 weight. Currently, we reorder it back to u8 for more proper data range to avoid precision issue.
3. Improve the compressed sdpa test cases to better validate ukernel implementation.
4. Some enhancement on verbose.

PR #3396, opened 2025-06-06
graph: doc, interface, backend: support SDPA training

ElaineBao:
# Description

- This PR implements https://github.com/uxlfoundation/oneDNN/pull/3233

  - [x] extend SoftMaxBackward operations to support new data type combinations that used in SDPA training. Related doc, library implementation, benchdnn cases are added.
  - [x] extend SoftMax operation to support optional stats output. Related doc, library implementation are added.
  - [x] extend benchdnn graph to support the validation of SoftMax stats.
  - [x] support SDPA training backward pattern (the forward pattern is the same as inference). Related library reference implementation is added.
  - [x] document of the training pattern
  - [x] validate correctness of SDPA training forward and backward pattern with benchdnn graph.


doc/graph/operations/Softmax.md
```diff
@@ -38,13 +43,14 @@ constructing an operation.
...
-| Src  | Dst             |
-|:-----|:----------------|
-| f32  | f32, bf16, f16  |
-| bf16 | bf16            |
-| f16  | f16             |
+| Src  | Dst             | Stats |
+|:-----|:----------------|:------|
+| f32  | f32, bf16, f16  | f32   |
+| bf16 | bf16            | f32   |
+| f16  | f16             | f32   |
```
TaoLv, line 56: @LuFinch Could you please confirm that the stats tensor is always f32 in PyTorch?

@ElaineBao It would be better to clarify this in the RFC document as well. Thanks.

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -27,14 +27,17 @@ namespace graph {
...
+        const std::vector<logical_tensor_t> &outputs) {
     // The order of input logical tensors in inputs is not certain, we need
     // to record the input offset in a certain order of ops.
     CHECK_BOOL(record_input_offset(sg, inputs));
     dims src1_user_dims = ltw(inputs[graph_inport[mm1_src]]).vdims();
     ndims = src1_user_dims.size();
     VCHECK_SDP_DECOMP(ndims == 4 || ndims == 5, false,
             "Input dims should be 4 or 5, but got %zu", src1_user_dims.size());
+    VCHECK_SDP_DECOMP(
+            outputs.size() == 1, false, "Doesn't support SDPA training yet");
```
TaoLv: Better to make the error message consistent with the checks.
```suggestion
            outputs.size() == 1, false, "does not support multiple outputs");
```

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -169,10 +169,13 @@ status_t sdp_primitive_config_t::locate_io(std::shared_ptr<subgraph_t> &sg,
...
 status_t sdp_primitive_config_t::initial_check(
         const std::shared_ptr<subgraph_t> &sg,
-        const std::vector<logical_tensor_t> &inputs, bool v1_kernel) {
+        const std::vector<logical_tensor_t> &inputs,
+        const std::vector<logical_tensor_t> &outputs, bool v1_kernel) {
     // At least 3 inputs: Q, K, V
     VCHECK_SDP_PRIMITIVE(inputs.size() >= 3, status::invalid_arguments,
             "At least 3 inputs are required");
+    VCHECK_SDP_PRIMITIVE(outputs.size() == 1, status::unimplemented,
+            "Doesn't support SDPA training yet");
```
TaoLv: Ditto.

src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -698,6 +698,168 @@ static status_t select_handler(
...
+    if (no_stats) {
+        new_softmax_op->add_output(dst);
+        insert_empty_scratchpad(new_softmax_op);
+        rewriter.to_insert(new_softmax_op);
+        rewriter.to_remove(op);
+        return status::success;
+    }
+
+    auto f32_dst = dst;
+    if (f32_dst->get_logical_tensor().data_type == dnnl::impl::data_type::f32) {
```
TaoLv: I guess below should work:
```suggestion
    if (f32_dst->get_logical_tensor().data_type == data_type::f32) {
```

src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -698,6 +698,168 @@ static status_t select_handler(
...
+        dst->remove_consumer(*op, 0);
+        insert_empty_scratchpad(new_softmax_op);
+        rewriter.to_insert(new_softmax_op);
+        rewriter.to_remove(op);
+    } else {
+        logical_tensor_t softmax_op_out_lt
+                = empty_logical_tensor_with_default_id();
+        f32_dst = std::make_shared<value_t>(
+                *new_softmax_op, 0, softmax_op_out_lt, true);
+        f32_dst->set_data_type(dnnl::impl::data_type::f32);
```
TaoLv: Same here. I think it's not needed to spell out all the nested namespaces.

src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -698,6 +698,168 @@ static status_t select_handler(
     return status::success;
 }
 
+static status_t softmax_handler(
+        const std::shared_ptr<op_t> &op, subgraph_rewriter_t &rewriter) {
```
TaoLv, line 711: what's the purpose of the lowering function?

src/graph/backend/dnnl/patterns/sdp.cpp
```diff
@@ -177,6 +177,48 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, float_sdp_fusion_gpu)
             return std::make_shared<sdp_base_t<>>();
         });
 
+DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, float_sdp_backward_fusion)
```
TaoLv, line 180: Does that mean the inference fusion pattern will be reused for training forward?

src/graph/backend/dnnl/passes/lower.cpp
ElaineBao: the main purpose is to compute stats with multiple primitives. If stats output does not exist, it will use a single softmax primitive to compute.

src/graph/backend/dnnl/patterns/sdp.cpp
ElaineBao: Yes, correct

doc/graph/operations/Softmax.md
LuFinch: Confirmed that all the CUDA FlashAttention/EfficientAttention/CudnnAttention backends use fp32 for logsumexp.

tests/benchdnn/graph/custom_driver.cpp
```diff
@@ -269,6 +270,91 @@ int execute(const prb_t *prb, const args_t &args, res_t *res) {
...
+        auto &mem = entry.second;
+
+        ref_mem_map.emplace(exec_arg,
+                dnn_mem_t(mem.md_, dnnl_f32, tag::abx, ref_engine,
+                        /* prefill = */ false));
+        auto &ref_mem = ref_mem_map[exec_arg];
+
+        switch (exec_arg) {
+            case DNNL_ARG_SRC:
+                SAFE(::custom::fill_mem(mem, ref_mem, -8, 7), WARN);
```
wzt1997: How do we decide this data range?

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -203,6 +210,13 @@ int ref_partition_t::init_graph_mem(
 }
 
 void ref_partition_t::exec_ops(res_t *res) {
+    // check if there's softmax backward op in the partition,
+    // which will be a candidate for sdpa training backward pattern
+    bool has_softmax_backward = std::any_of(partition_ops_ref_.begin(),
```
wzt1997, line 232: May use `static`?

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -263,7 +288,7 @@ void ref_partition_t::exec_ops(res_t *res) {
                     || (parent0 == "Multiply" && parent1 == "MatMul");
         }
 
-        if (is_sdpa_pattern || is_gated_mlp_pattern) {
+        if (is_sdpa_pattern || is_sdpa_bwd_pattern || is_gated_mlp_pattern) {
```
wzt1997: The names are a little bit confusing: actually they are used to indicate whether the current op needs precision downcasting, but not the status of the whole pattern. Could you help improve it?

doc/graph/operations/Softmax.md
ElaineBao: > @ElaineBao It would be better to clarify this in the RFC document as well. Thanks.

Updated RFC. Thanks!

src/graph/backend/dnnl/passes/lower.cpp
ElaineBao: I change it to `impl::data::type::f32`, or it will be deduced as `dnnl::memory::data_type`

tests/benchdnn/graph/ref_partition.cpp
ElaineBao: updated, could you please take a look at the new names?

tests/benchdnn/graph/custom_driver.cpp
ElaineBao: good question, I change it to reuse softmax driver's data filling function.

tests/benchdnn/graph/ref_partition.cpp
ElaineBao: I guess we don't want to make it shared between different function calls

tests/benchdnn/graph/ref_partition.cpp
wzt1997: Great. Thanks.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/memory_planning.cpp
```diff
@@ -392,43 +392,51 @@ status_t memory_planner_t::assign_external_inputs_buffer(
...
-                        if (pair.out_idx_ != cur_val->get_offset()) continue;
-                        auto in_val = producer.get_input_value(pair.in_idx_);
-                        if (buffer_assignments_.at(in_val.get()) != orig_info
-                                || buffer_assignments_.at(in_val.get()).kind_
-                                        == external_input)
-                            continue;
-                        q.push(in_val.get());
+    for (const auto &op : sg->get_ops()) {
+        for (const auto &val : op->get_output_values()) {
+            for (size_t i = 0; i < outputs.size(); i++) {
```
xiang1guo, line 397: Why this change?

ElaineBao: Good question, it's because the graph's get_output_values() function only considers the explicit output, it doesn't consider intermediate output, in softmax's case, after lowering, softmax's dst is used to compute stats, it's an intermediate output then, but it's still the output of the whole graph.

LuFinch:
I have a POC branch which integrated SDPA training into Pytorch XPU with this PR, and verify that UTs and model accuracy benchmark could pass.

tests/benchdnn/softmax/softmax.hpp
```diff
@@ -193,6 +193,7 @@ std::vector<int> supported_exec_args(dir_t dir);
 int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
         dnnl_primitive_t prim, const prb_t *prb, res_t *res,
         dnnl_primitive_t prim_ref = nullptr);
+int fill_data_fwd(const prb_t *prb, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp);
```
dzarukin:
```suggestion
// Used in graph custom op to cover softmax with stats for sdpa training.
// TODO: remove when/if softmax primitive receives a public extension for extra
// stat output support.
int fill_data_fwd(const prb_t *prb, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp);
```

tests/benchdnn/graph/ref_primitive.cpp
```diff
@@ -22,6 +22,12 @@ namespace graph {
 ref_primitive_t::ref_primitive_t(const deserialized_op_t &op)
     : op_(op), kind_(opstr2kind(op_.kind_)), driver_(opkind2driver(kind_)) {
 
+    if (kind_ == dnnl::graph::op::kind::SoftMax && op_.out_lts_.size() > 1) {
```
dzarukin: Looks like it better become a part of `opkind2driver`. To pass an extra argument, I suggest to change the signature to
`driver_t opkind2driver(const deserialized_op_t &op)`, and get `kind` out of the `op`.
If it's impossible to change to `op`, then create a new overloaded version with it.
A new version would also inspect the number of outputs from the op.

This way it would be possible to support future deviations of in-house primitives versus their custom modifications rather than piling those up in here.

tests/benchdnn/graph/ref_primitive.cpp
```diff
@@ -214,6 +220,7 @@ void ref_primitive_t::check_correctness(
                     {DNNL_ARG_BIAS, BIA},
                     {DNNL_ARG_DIFF_BIAS, BIA},
                     {DNNL_ARG_DST, DST},
+                    {DNNL_ARG_DST_1, DST_ITER},
```
dzarukin: Better do `DST_1` instead of `DST_ITER` even if it requires some more changes to enable it.

tests/benchdnn/graph/ref_partition.hpp
```diff
@@ -114,7 +114,7 @@ class ref_partition_t {
     // keep the memory for each logical tensor
     // before the execution of each reference primitive,
     // replace the args with the memory from this map.
-    std::unordered_map<size_t, const dnn_mem_t &> lt_id_2_mems_;
+    std::map<size_t, const dnn_mem_t &> lt_id_2_mems_;
```
dzarukin: Would be nice to get a comment why sorting is needed and where it's relied on.

tests/benchdnn/graph/flex_rewrite.cpp
```diff
@@ -892,6 +891,25 @@ void flex_rewrite_t::infer_output_shape(
...
+                    size_t out1 = aop.out_lts_[1].id_;
+                    int64_t axis = 1;
+                    if (aop.attrs_.find("axis") != aop.attrs_.end()) {
+                        axis = aop.attrs_["axis"].s64_value_;
+                    }
+                    axis = axis >= 0 ? axis : gi[out0].size() + axis;
+                    logical_tensor::dims out1_dims = gi[out0];
+                    out1_dims[axis] = 1;
+                    gi[out1] = out1_dims;
+                }
```
dzarukin, line 925: Couple questions regarding this piece:
* Is it aligned with Zhitao's work not to override stuff from the JSON and use that if it's not specified?
* I thought `axis` is mandatory for softmax, though it looks like it's optional. If it's not, please update the section handling `axis` with "either process or fail" style. If it is, the default should be coming from a more authority source over just using `1` blindly as it may change in the future.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -339,11 +339,32 @@ partition_data_displacer_t::partition_data_displacer_t(
             set_seq_len_displace_args(child_sub_op, seq_len_q);
             break;
         }
+
+        // Fill proper data for softmax stats in sdpa backward graph.
+        while (aop.kind_ == "Subtract") {
```
dzarukin, line 345: You might want to secure this modification with recent `dg` update to contain a known `sdpa` pattern (if it made to main). If not, please add a TODO to use it once it's available.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -615,4 +642,63 @@ int partition_data_displacer_t::gen_causal_mask_filling(
...
+            int64_t idx = ou_in_offset + as * inner_size;
+            dst[idx] *= space_denom;
+            dst_space_max = MAX2(dst_space_max, dst[idx]);
+        }
+
+        // computes stats
+        int64_t stats_idx = ou * inner_size + in;
+        float stats_value = dst_space_max ? space_max - logf(space_denom) : 0.f;
+        m.set_elem(stats_idx, stats_value);
+    });
```
dzarukin, line 692: Could you clarify, please, why exactly these values are needed? It seems to me that it's only a backward path that gets executed here, thus, data from forward is not available and final values don't have to have anything in common with forward, unless I'm missing something...

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -295,11 +322,14 @@ void ref_partition_t::exec_ops(res_t *res) {
...
         dnnl_data_type_t dt = dnnl_data_type_undef;
-        if ((is_sdpa_pattern || is_gated_mlp_pattern)
+        if ((is_softmax_in_sdpa_pattern || is_matmul_in_sdpa_bwd_pattern
+                    || is_multiply_in_gated_mlp_pattern)
                 && need_unfusable_output_crop(op, dt)) {
             for (size_t i = 0; i < op.out_lts_.size(); i++) {
                 // There's no need to reorder data for undefined or f32 tensors.
                 if (dt == dnnl_data_type_undef || dt == dnnl_f32) continue;
+                // For SoftMax stats, it's f32, no need to reorder.
+                if (op.kind_ == "SoftMax" && i == 1) continue;
```
dzarukin: I'd suggest to make it as a part of `need_unfusable_output_crop`. It will probably require to move the function inside the loop and add an index to the signature.

tests/benchdnn/graph/custom_driver.cpp
```diff
@@ -269,6 +271,102 @@ int execute(const prb_t *prb, const args_t &args, res_t *res) {
 }
 } // namespace reshape
 
+namespace softmax {
```
dzarukin: It's a placeholder for the idea to reuse benchdnn's softmax prb discussed offline.

tests/benchdnn/graph/input_displacer.cpp
ElaineBao: The stats is used in the backward path to resume softmax's output by: `out = exp(in - stats)`, the sematics of softmax is that out should be ranged in [0,1]. If we use a randomly initialized stats, it can not fit into this range, and result in some NaN or correctness issues at last.....

tests/benchdnn/graph/flex_rewrite.cpp
ElaineBao: > * Is it aligned with Zhitao's work not to override stuff from the JSON and use that if it's not specified?

Just had some offline discussion with Zhitao, it should be aligned (the output is deduced from the input).

> * I thought `axis` is mandatory for softmax, though it looks like it's optional. If it's not, please update the section handling `axis` with "either process or fail" style. If it is, the default should be coming from a more authority source over just using `1` blindly as it may change in the future.

Referring to our [document](https://uxlfoundation.github.io/oneDNN/dev_guide_op_softmax.html), the axis is an optional attribute with a default value of 1. In most cases, the json file should have this attribute if it's dumped from the library (the library will fill the value for optional attribute). So we can take this `1` as a placeholder in most cases :)

tests/benchdnn/graph/input_displacer.cpp
ElaineBao: thanks for the reminder, as https://github.com/uxlfoundation/oneDNN/pull/3227 has not been merged, let me add a TODO here first.

tests/benchdnn/graph/ref_primitive.cpp
ElaineBao: I cannot change it to `driver_t opkind2driver(const deserialized_op_t &op)` because of circular dependency..
Therefore, I change it to `driver_t opkind2driver(const graph::op::kind &kind, bool check_num_outputs = false, size_t num_outputs = 1)`. Does it look good to you?

tests/benchdnn/graph/ref_partition.hpp
ElaineBao: Added a comment:
> We use a map instead of unordered_map here because, in SDPA training, softmax stats need to be recomputed from the softmax input, which depends on Q, K, optional scale, and optional mask. All these memories must be finalized before recomputing the softmax stats.

tests/benchdnn/graph/ref_partition.cpp
ElaineBao: updated, thanks for the suggestion!

tests/benchdnn/graph/ref_primitive.cpp
ElaineBao: updated

tests/benchdnn/graph/ref_primitive.cpp
dzarukin: I anticipate more exclusions from this rule, thus, I'd like to introduce an universal interface.

Is it possible to convert the function into a method of `deserialized_op_t` instead? It has unconverted `kind_` for that and will have number of extra outputs.

tests/benchdnn/graph/ref_partition.hpp
dzarukin: I'm sorry, but I'm not following why `map` solves this problem... IIRC, the key is a logical tensor ID, and if softmax ID is less in value than all LT ID's that it has dependencies on, there's no way to identify if the dependency is fulfilled. What am I missing?

tests/benchdnn/graph/flex_rewrite.cpp
dzarukin: Could you, please, leave a comment for it?

ElaineBao: comment added, thanks!

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -149,11 +149,22 @@ int ref_partition_t::init_ref(
...
+            if (lt_id_2_mems_.find(lt_id) == lt_id_2_mems_.end()) continue;
+            if (data_displacer.get_filling_type(lt_id)
+                    == filling_type_t::softmax_stats) {
+                res_t temp_res;
+                exec_ops(&temp_res);
+            }
+            const dnn_mem_t &mem = lt_id_2_mems_.at(lt_id);
+            SAFE_V(data_displacer.displace_input_data(
+                    lt_id, const_cast<dnn_mem_t &>(mem), lt_id_2_mems_, res));
+        }
```
ElaineBao, line 178: a placeholder for a discussion in another thread:

in SDPA training, softmax stats need to be recomputed from the softmax input, which depends on Q, K, optional scale, and optional mask. All these memories must be finalized before recomputing the softmax stats. So in order to make sure the filling of input_data is in topo order, traverse the graph op by op.

tests/benchdnn/graph/ref_partition.hpp
ElaineBao: I'm sorry, it's my fault, I was thinking of something like vector, first in first out... I use another way to make sure the filling of input_data is in topo order, refer to https://github.com/uxlfoundation/oneDNN/pull/3396#discussion_r2176942633

tests/benchdnn/graph/ref_primitive.cpp
ElaineBao: thanks for the valuable suggestion, I've converted the function into a method of `deserialized_op_t`.

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -31,11 +31,11 @@ SDPA graph, getting partition from the graph, and optimizing the kernels
 underneath. In general, an SDPA pattern is defined as a directional acyclic
 graph (DAG) using oneDNN Graph API.
 
-### Floating-point SDPA
+### Floating-point SDPA for Inference and Training Forward
```
TaoLv: I would suggest to separate inference and training into two sections. Keep the existing inference part unchanged and in the new training section, explicit mention that training forward pattern is the same (if it is) as inference pattern and refer readers to the inference section.

src/graph/backend/dnnl/passes/memory_planning.cpp
TaoLv: what do you mean by explicit output and intermediate output? from what you described here, it looks like we need to fix `get_output_values()` instead.

xiang1guo: I looked into this part before. I think here `explicit` means that the softmax output has no additional consumers, while `intermediate` indicates that the softmax output still has consumers — it is used to compute the stats. This is exactly the situation in Yixin’s PR right now: when softmax is lowered, its output is inserted into an additional subgraph to compute the stats.

I think the `get_ouput_values` function doesn’t need to be modified, as it already takes outer consumers into account. ([here](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/interface/graph.hpp#L237)), One possible option is to attach an additional `Wildcard` output to the softmax during lowering, indicating that it has outer consumers. This would cause the memory planner to treat its output as an output buffer, not sure if works :). And I think Yixin's change here also works.

BTW, if that the softmax primitive will likely support outputting stats in the future, this issue may not even exist down the line.

ElaineBao: Yes, `get_output_values` does check outer-consumers. In fact, for sdpa training forward graph, we do NOT need to consider this part, because softmax's dst is an intermediate output that is used in the following bmm. The reason I modify the code here is for a single softmax op (e.g. defined in benchdnn graph single op json file), which doesn't have consumers, so originally `get_output_values` can recognize its dst as an output value. But when dst is used to compute stats, the dst's consumer is not empty any more, thus the issue.

doc/graph/fusion_patterns/sdpa.md
ElaineBao: updated, thanks for the suggestion!

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -123,19 +167,22 @@ platforms follow the general description in @ref dev_guide_data_types.
...
-     as (N, H, S, D_qk) and V tensor with shape defined as (N, H, S, D_v).
-   - Optimized implementation is available for OpenMP runtime and Threadpool
+   - Optimized implementation for inference is available for 4D Q/K tensors with
+     shape defined as (N, H, S, D_qk) and V tensor with shape defined as
+     (N, H, S, D_v).
+   - Optimized implementation for inference is available for OpenMP runtime and Threadpool
      runtime on Intel Architecture Processors.
    - Specifically for OpenMP runtime, the optimized implementation requires `N *
      H > 2 * thread number` to get enough parallelism.
+   - Optimized implementation for training forward and backward is not available.
```
TaoLv: I feel it's not needed to mention that. Let's only state what are optimized. For others, the first bullet about primitive-based implementation covers that.

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -123,19 +167,22 @@ platforms follow the general description in @ref dev_guide_data_types.
...
-   - Optimized implementation is available for `f16` or `bf16` SDPA with `f32`
-     intermediate data type and `D <= 512` on Intel Graphics Products with
-     Intel(R) Xe Matrix Extensions (Intel(R) XMX) support.
+   - Optimized implementation for inference is available for 4D Q/K tensors with
+     shape defined as (N, H, S, D_qk) and V tensor with shape defined as (N, H,
+     S, D_v) where D_qk equals D_v.
+   - Optimized implementation for inference is available for `f16` or `bf16`
+     SDPA with `f32` intermediate data type and `D <= 512` on Intel Graphics
+     Products with Intel(R) Xe Matrix Extensions (Intel(R) XMX) support.
+   - Optimized implementation for training forward and backward is not available.
```
TaoLv: Ditto.

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -89,19 +89,63 @@ optional.
...
+
+oneDNN defines floating-point (f32, bf16, or f16) SDPA for training forward as
+follows. The blue nodes are required while the brown nodes are optional.
+
+![SDPA pattern](images/sdpa_forward.png)
+
+The only difference between the inference and training forward patterns is that,
+for training forward, the `Stats` output of the SoftMax operation is needed.
+See [SoftMax](@ref dev_guide_op_softmax) in Graph API for more details.
+
```
TaoLv, line 103: Let's mention that dropout is not supported currently.

src/graph/backend/dnnl/passes/memory_planning.cpp
TaoLv: I still don't understand. Is it caused by the lowering of softmax with stats output?

LuFinch (approved):
Look good from Pytorch integration side.

doc/graph/fusion_patterns/sdpa.md
ElaineBao: removed

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -89,19 +89,63 @@ optional.
 
    ![SDPA-Reorder](images/sdpa-reorder.png)
 
+### Floating-point SDPA for Training Forward
```
vpirogov: Terms 'forward' and 'backward' are adjectives and require a noun. Let's use something like 'forward propagation' and 'backpropagation'.

```suggestion
### Floating-point SDPA for Training Forward Propagation
```

doc/graph/fusion_patterns/sdpa.md
```diff
@@ -89,19 +89,63 @@ optional.
...
+oneDNN defines floating-point (f32, bf16, or f16) SDPA for training forward as
+follows. The blue nodes are required while the brown nodes are optional.
+
+![SDPA pattern](images/sdpa_forward.png)
+
+The only difference between the inference and training forward patterns is that,
+for training forward, the `Stats` output of the SoftMax operation is needed.
+See [SoftMax](@ref dev_guide_op_softmax) in Graph API for more details.
+
+### Floating-point SDPA for Training Backward
```
vpirogov:
```suggestion
### Floating-point SDPA for Training Backpropagation
```

doc/graph/fusion_patterns/sdpa.md
ElaineBao: Thanks for the suggestion! I've updated all relevant instances to use `forward propagation` and `backpropagation`.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/flex_rewrite.cpp
```diff
@@ -1393,51 +1412,50 @@ void flex_rewrite_t::dt_rewrite(deserialized_graph_t &dgraph) {
...
-        }
+            if (aop.empty()) {
+                BENCHDNN_PRINT(0,
+                        "graph: rewrite: ID `%zd` is not found in the graph\n",
+                        v.first);
+                SAFE_V(FAIL);
+            }
+            auto op_driver = aop.get_driver();
 
-        auto &aop = dgraph.get_op(v.first);
```
dzarukin, line 1420: I'm not sure I'm following this line stopped working for the introduced changes...

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -376,6 +399,12 @@ int partition_data_displacer_t::displace_input_data(
         SAFE(gen_causal_mask_filling(mem_replace, mem.md_, res), WARN);
     } else if (filling_type == filling_type_t::fixed_setting) {
         SAFE(gen_fixed_set_filling(mem_replace, mem.md_, fill_cfg, res), WARN);
+    } else if (filling_type == filling_type_t::softmax_stats) {
+        auto softmax_src_lt = &main_op.in_lts_[0];
```
dzarukin:
```suggestion
        const auto *softmax_src_lt = &main_op.in_lts_[0];
```

tests/benchdnn/graph/deserialize.hpp
```diff
@@ -79,6 +79,8 @@ struct deserialized_op_t {
 
     op create() const;
 
+    dnnl_driver_t get_driver() const;
```
dzarukin: Why changing name from `opkind2driver`? There's no `driver` object or value to get from the op.

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -87,6 +87,9 @@ int ref_partition_t::init_ref(
         SAFE(ref_prim->init_prb(res), WARN);
 
         SAFE_V(ref_prim->init_prim(::get_test_engine(), res));
+        // reset res to avoid a skipped state from init_prim() to affect the rest part.
+        res->state = UNTESTED;
+        res->reason.clear();
```
dzarukin: Please introduce a state `reuse_driver_for_ref_compute_` that would control when to reset the state and whennot.
It should likely rely on the prb and `has_stats` member in case of softmax.
```suggestion
        // TODO: try to make a general logic when to reset the state.
        if (reuse_driver_for_ref_compute_ && res->state == SKIPPED) {
            // reset res to avoid a skipped state from init_prim() to affect the rest part.
            res->state = UNTESTED;
            res->reason.clear();
        }
```

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -300,18 +334,19 @@ void ref_partition_t::exec_ops(res_t *res) {
...
         // A data type to where transform the data will also be provided by the
         // same function since there are corner cases.
-        dnnl_data_type_t dt = dnnl_data_type_undef;
-        if ((is_sdpa_pattern || is_gated_mlp_pattern)
-                && need_unfusable_output_crop(op, dt)) {
+        if (is_softmax_in_sdpa_pattern || is_matmul_in_sdpa_bwd_pattern
+                || is_multiply_in_gated_mlp_pattern) {
             for (size_t i = 0; i < op.out_lts_.size(); i++) {
+                dnnl_data_type_t dt = dnnl_data_type_undef;
+                need_unfusable_output_crop(op, i, dt);
```
dzarukin: The function designed to return a bool value, I'd still rely on it in the first place instead of a `dt` value.

tests/benchdnn/graph/ref_primitive.cpp
```diff
@@ -122,12 +207,10 @@ int ref_primitive_t::init_prim(
             args_ = args_t(mems_); \
             SAFE(execute_and_wait(fwd_prim_, args_, res), WARN); \
         } \
-        SAFE(create_primitive(prim_, ref_eng, ::driver::init_pd, prb, res, \
-                     prb->dir, \
-                     is_special_backward_op_ ? query_pd(fwd_prim_) : nullptr, \
-                     false, nullptr, force_f32_prim_dt, \
-                     /*is_graph_ref=*/true), \
-                WARN); \
+        create_primitive(prim_, ref_eng, ::driver::init_pd, prb, res, \
```
dzarukin: Please return `SAFE` since the original idea hasn't flown.

tests/benchdnn/graph/ref_primitive.hpp
```diff
@@ -18,9 +18,35 @@
...
+DECLARE_REF_PRIM_HELPER_FUNCS(eltwise);
+DECLARE_REF_PRIM_HELPER_FUNCS(gnorm);
+DECLARE_REF_PRIM_HELPER_FUNCS(lnorm);
+DECLARE_REF_PRIM_HELPER_FUNCS(matmul);
+DECLARE_REF_PRIM_HELPER_FUNCS(pool);
+DECLARE_REF_PRIM_HELPER_FUNCS(prelu);
+DECLARE_REF_PRIM_HELPER_FUNCS(reduction);
+DECLARE_REF_PRIM_HELPER_FUNCS(reorder);
+DECLARE_REF_PRIM_HELPER_FUNCS(resampling);
+DECLARE_REF_PRIM_HELPER_FUNCS(softmax);
```
dzarukin: It seems that header declarations can be stripped off. Less macros in the code the better.

tests/benchdnn/graph/utils.cpp
```diff
@@ -430,157 +430,6 @@ dnnl::graph::op::attr attrstr2kind(const std::string &attr_name) {
     return dnnl::graph::op::attr::undef;
 }
 
-class op_kind_hash_t {
-public:
-    std::size_t operator()(const dnnl::graph::op::kind &op_kind) const {
-        return std::hash<int>()(static_cast<int>(op_kind));
-    }
-};
```
dzarukin, line 438: Hash was required for the `unordered_map`, I don't see it was left. How does it compile now for moderately new clang?

tests/benchdnn/graph/input_displacer.hpp
```diff
@@ -58,7 +60,14 @@ class partition_data_displacer_t {
     partition_data_displacer_t() = default;
     partition_data_displacer_t(
             const deserialized_graph_t &dg, const dnnl::graph::partition &par);
-    int displace_input_data(size_t lt_id, dnn_mem_t &mem, res_t *res);
+    int displace_input_data(size_t lt_id, dnn_mem_t &mem,
+            const std::unordered_map<size_t, const dnn_mem_t &> &lt_id_2_mems,
+            res_t *res);
```
dzarukin, line 67: I don't think it's a sustainable approach to extend the interface for each future need which doesn't fit current signature. Instead let's add an internal `private` state and use it inside softmax filling (and other future needs):
```
    // TODO: rename to a generic stash name in the future if needed.
    // Note: potentially extendable to a vector or map to have multiple stashes.
    dnn_mem_t softmax_src_mem;
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -615,4 +644,63 @@ int partition_data_displacer_t::gen_causal_mask_filling(
...
+        for (int64_t as = 0; as < axis_size; ++as) {
+            int64_t idx = ou_in_offset + as * inner_size;
+            space_max = MAX2(space_max, src_mem.get_f32_elem(idx));
+        }
+
+        for (int64_t as = 0; as < axis_size; ++as) {
+            int64_t idx = ou_in_offset + as * inner_size;
+            float s = src_mem.get_f32_elem(idx);
+            float D = dst[idx] = expf(s - space_max);
+            space_denom += D;
```
dzarukin:
```suggestion
            float D = expf(s - space_max);
            dst.set_elem_f32(idx, D);
            space_denom += D;
```

tests/benchdnn/softmax/ref_softmax.cpp
```diff
@@ -66,6 +69,12 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
         }
 
         if (alg == SOFTMAX || alg == SOFTMAX_INF_AS_ZERO) {
+            if (prb->has_stats) {
+                int64_t stats_idx = ou * inner_size + in;
+                float stats_value
+                        = space_denom ? space_max + logf(space_denom) : 0;
+                stats.set_elem(stats_idx, stats_value);
```
dzarukin:
```suggestion
                stats.set_elem_f32(stats_idx, stats_value);
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -615,4 +644,63 @@ int partition_data_displacer_t::gen_causal_mask_filling(
...
+        }
+
+        space_denom = space_denom ? 1.f / space_denom : 1.f;
+        float dst_space_max = -FLT_MAX;
```
dzarukin: Looks like can be dropped in a favor of `space_denom = space_denom ? 1.f / space_denom : 0.f;`

tests/benchdnn/graph/deserialize.hpp
ElaineBao: The original idea is that after the change, it doesn't take a parameter of opkind. I changed it back to make it clear that it still converts from opkind

tests/benchdnn/graph/input_displacer.cpp
ElaineBao: I removed the computing for dst, it's not needed any more if dst_space_max is not needed.

tests/benchdnn/graph/input_displacer.hpp
ElaineBao: lt_id_2_mems is passed from [ref_partition.cpp](https://github.com/uxlfoundation/oneDNN/blob/e052ae2b1880c934553f9e5af3d46938dd082fc2/tests/benchdnn/graph/ref_partition.cpp#L166-L167) to this `partition_data_displacer_t` class. If we don't want the input `lt_id_2_mems`, we will need to provide two interfaces for ref_partition to get the softmax_src's id and set the corresponding mem. I feel it's not a sustainable way, too..

tests/benchdnn/graph/input_displacer.cpp
ElaineBao: updated according to the comments, thanks!

tests/benchdnn/graph/flex_rewrite.cpp
```diff
@@ -1393,51 +1412,51 @@ void flex_rewrite_t::dt_rewrite(deserialized_graph_t &dgraph) {
...
 
-        auto target_kind = opstr2kind(v.second);
-        if (target_kind == dnnl::graph::op::kind::LastSymbol) {
-            BENCHDNN_PRINT(0,
-                    "graph: rewrite: invalid target op kind %s is provided\n",
-                    v.second.c_str());
-            SAFE_V(FAIL);
-        }
+    for (auto &aop : dgraph.ops_) {
+        for (const auto &v : op_kind_map_) {
```
dzarukin: Nit: less unneeded spaces this way (no curls on the first and last lines)
```suggestion
    for_(auto &aop : dgraph.ops_)
    for (const auto &v : op_kind_map_) {
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -615,4 +644,55 @@ int partition_data_displacer_t::gen_causal_mask_filling(
...
+    int64_t outer_size {1}, inner_size {1}, axis_size {1};
+    for (size_t i = 0; i < axis; i++) {
+        outer_size *= softmax_src_shape[i];
+    }
+    for (size_t i = axis + 1; i < softmax_src_shape.size(); i++)
+        inner_size *= softmax_src_shape[i];
+    axis_size = softmax_src_shape[axis];
+
+    benchdnn_parallel_nd(outer_size, inner_size, [&](int64_t ou, int64_t in) {
+        float space_denom = 0.;
```
dzarukin:
```suggestion
        float space_denom = 0.f;
```

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -615,4 +644,55 @@ int partition_data_displacer_t::gen_causal_mask_filling(
...
+            float s = src_mem.get_f32_elem(idx);
+            space_denom += expf(s - space_max);
+        }
+
+        // computes stats w.r.t. the softmax input
+        // stats = max(input) + log(sum(exp(input - max(input))))
+        // consider inf as a zero value
+        int64_t stats_idx = ou * inner_size + in;
+        float stats_value = space_denom ? space_max + logf(space_denom) : 0.f;
+        m.set_elem(stats_idx, stats_value);
```
dzarukin:
```suggestion
        m.set_f32_elem(stats_idx, stats_value);
```

tests/benchdnn/graph/input_displacer.hpp
dzarukin: Please add TODO in here to improve the interface as discussed offline.

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -300,20 +346,21 @@ void ref_partition_t::exec_ops(res_t *res) {
...
         // A data type to where transform the data will also be provided by the
         // same function since there are corner cases.
-        dnnl_data_type_t dt = dnnl_data_type_undef;
-        if ((is_sdpa_pattern || is_gated_mlp_pattern)
-                && need_unfusable_output_crop(op, dt)) {
+        if (is_softmax_in_sdpa_pattern || is_matmul_in_sdpa_bwd_pattern
+                || is_multiply_in_gated_mlp_pattern) {
             for (size_t i = 0; i < op.out_lts_.size(); i++) {
+                dnnl_data_type_t dt = dnnl_data_type_undef;
+                bool need_crop = need_unfusable_output_crop(op, i, dt);
```
dzarukin: This should do.
```suggestion
                if (!need_unfusable_output_crop(op, i, dt)) continue;
```

dzarukin (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

wzt1997 (approved):

TaoLv:
Thank you all for the review and discussions. Now I'm merging this into main branch. It's quite a big code change. @ElaineBao please keep an eye on the CI/Nightly testing results and prepare the backport for rls-v3.9. Thanks.

PR #3397, opened 2025-06-06
src: gpu: intel: jit: gemm: treat dst s8/u8 as dy_quant

hidefromkgb:
This resolves [MFDNN-13621](https://jira.devtools.intel.com/browse/MFDNN-13621).


petercad (changes requested):
@hidefromkgb This check needs some refinement, I think. The "dynamic quantization" path should be chosen not based on the type of C, but rather on:
* A/B are both integer types (so we don't want to convert them to floating point type before computation)
* At least one of A/B has grouped scales, requiring periodic conversion of the s32 temporary accumulators to f32.

kealan-barbieri:
@petercad can you clarify what the difference is for the "dynamic quantization" path other than allowing additional quantization attributes used in those cases?

petercad:
> @petercad can you clarify what the difference is for the "dynamic quantization" path other than allowing additional quantization attributes used in those cases?

Looking over the code more carefully, you're right -- sorry for the extra noise.

@hidefromkgb, I think you could go ahead and remove the check on C type entirely.

petercad (approved):

hidefromkgb:
> @hidefromkgb, I think you could go ahead and remove the check on C type entirely.

I'd prefer to leave it in place to prohibit untested cases like INT8×INT8→FP8.

PR #3399, opened 2025-06-06
build: account for Compute Library minor version in CMake

Sqvid:
# Description

We need to account for the minor version when accepting or rejecting Compute Library (ACL) versions. This change allows setting the minimum supported ACL version in MAJOR.MINOR format.


Sqvid:
cc: @Ryo-not-rio

Ryo-not-rio (approved):

vpirogov (approved):

PR #3403, opened 2025-06-08
cpu: x64: matmul: fix blocking heuristics for l2 set issues

yair-obodovsky:
fix performance degradation related to issue:
https://jira.devtools.intel.com/browse/MFDNN-13324

### DIFF
--- a/src/cpu/x64/matmul/amx_blocking_heuristics.cpp
+++ b/src/cpu/x64/matmul/amx_blocking_heuristics.cpp
@@ -683,6 +683,28 @@ bool matmul_amx_blocking_params_macro_t::set_blocking_parameters() {
 
             if (repeat_loop_over_k && critical_l2_set_issues_a)
                 vertical_not_possible = true;
+
+            // The following consts are correct for all platforms supporting AMX
+            constexpr size_t l2_ways = 16;
+            constexpr size_t l2_ways_threshold = size_t(l2_ways * 0.75);
+            constexpr size_t l2_sets = 2048;
+            constexpr size_t l2_line_size = l2_sets * 64;
+
+            // If output matrix in vertical traversal suffers from set issues in
+            // the L2, then go horizontally.
+            // First, calculate the stride on the N dim within a page
+            size_t cache_stride = N * c_dt_sz % PAGE_4K;
+            if (cache_stride == 0) { cache_stride = PAGE_4K; }
+            // Second, calculate the number of ways the output matrix will use
+            // in each set. If higher than the threshold, useful cache lines in
+            // L2 will be evicted.
+            if (l2_line_size % cache_stride == 0) {
+                size_t num_ways_c
+                        = div_up(cache_stride * m_per_thread, l2_line_size);
+                if (num_ways_c >= l2_ways_threshold) {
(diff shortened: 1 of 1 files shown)


yair-obodovsky:
make test
set test_scope=NIGHTLY

yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
```diff
@@ -683,6 +683,29 @@ bool matmul_amx_blocking_params_macro_t::set_blocking_parameters() {
 
             if (repeat_loop_over_k && critical_l2_set_issues_a)
                 vertical_not_possible = true;
+
+            // The following consts are correct for all platforms supporting AMX
+            const size_t page_size = 4096;
+            const size_t l2_ways = 16;
+            const size_t l2_ways_threshold = size_t(l2_ways * 0.75);
```
densamoilov: For my understanding, how did you choose 0.75?

densamoilov (approved):

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
yair-obodovsky: We selected 0.75 for the L2 capacity based on empirical testing, where it consistently yielded strong performance across our validation sets. This value also aligns with prior configurations used in earlier versions of OneDNN before the adoption of the new heuristics, suggesting it serves as a robust and reliable baseline as can be seen in:matmul_amx_blocking_params_t::L2_threshold()

dzarukin: Maybe `constexpr` instead of regular values?

There's also a value for a page size: https://github.com/uxlfoundation/oneDNN/blob/26589027aa7aaffe03f5a7eafbeeaecc07a12f11/src/cpu/platform.hpp#L188-L191

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
```diff
@@ -683,6 +683,29 @@ bool matmul_amx_blocking_params_macro_t::set_blocking_parameters() {
...
+            const size_t l2_sets = 2048;
+            const size_t l2_line_size = l2_sets * 64;
+
+            // If D matrix in vertical traversal suffers from set issues in
```
tczeszun: dumb question: don't we need to check if D matrix is present? or this applies to both D and C?

tczeszun (approved):

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
yair-obodovsky: fixed.

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
yair-obodovsky: It applies to the final output matrix (C or D). comments were fixed

yair-obodovsky:
make test
set test_scope=NIGHTLY

PR #3412, opened 2025-06-11
Pryorgal/host scalar sdpa

pv-pterab-s:
THIS PR IS A DRAFT AND NOT YET READY FOR REVIEW

REBASE/SQUASH WILL OCCUR AFTER FULL CI PLATFORM TESTING

Implements major changes:
    1. Addition of host-side-scalar memory descriptors: indicate to primitive descriptors that a scalar input is stored in host memory and is to be passed as a scalar parameter - not pointer - to OpenCL kernels
    2. Modify SDPA to accept a host-side-scalar descriptor for scale input. On such a descriptor, pass scale as a scalar kernel parameter to the OpenCL kernel while maintaining old behavior for pre-existing descriptor types.

By passing `scale` (a single value) as a scalar kernel parameter as opposed to a device memory pointer, this PR avoids costly host/device memory transfers on SDPA invocation (as scale changes).

To utilize these additions, instantiate the SDPA primitive with the descriptor for scale set as `host_side_scalar`. Then, on SDPA execution, pass the `scale` value as a 1-element-sized host-side memory object (via an engine of type CPU). Example:

```
dnnl::engine gpu_engine = dnnl::engine(engine::kind::gpu, 0);
dnnl::stream gpu_stream = dnnl::stream(gpu_engine);
dnnl::engine cpu_engine = dnnl::engine(engine::kind::cpu, 0);
dnnl::stream cpu_stream = dnnl::stream(cpu_eng);

// Create a host-side-scalar memory descriptor for SDPA's primitive descriptor
dnnl_memory_desc_t tmp_scale_md;
dnnl_memory_desc_create_host_side_scalar(&tmp_scale_md, memory::data_type::f16);
memory::desc scale_md = memory::desc(tmp_scale_md);

// Create a host-side memory block of 1-element for SDPA's execution
out.m_scale = memory(memory::desc({1, 1, 1, 1, 1}, 
                             memory::format_tag data_type::f16,
                             p.qdt, abcde), cpu_engine);

// Provide host-side-scalar scale descriptor to SDPA: scale will be passed as scalar kernel parameter
sdpa::primitive_desc sdpa_prim_pd = sdpa::primitive_desc(gpu_engine ... scale_md ...);

std::unordered_map<int, memory> sdpa_args;
sdpa_args.insert({DNNL_ARG_QUERIES, m_query});
sdpa_args.insert({DNNL_ARG_KEYS, m_keys});
sdpa_args.insert({DNNL_ARG_VALUES, m_value});
sdpa_args.insert({DNNL_ARG_DST, m_output});

// Provide host-side memory to SDPA execute: 
sdpa_args.insert({DNNL_ARG_SCALE, m_scale});
sdpa_prim.execute(gpu_stream, sdpa_args);
```


src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -503,6 +503,10 @@ status_t micro_sdpa_t::init(impl::engine_t *engine) {
     kernel_ctx.define_int("SOFTMAX_INF_AS_ZERO",
             d->softmax_alg == alg_kind::softmax_accurate_inf_as_zero);
 
+    if (pd()->scale_md()->format_kind == format_kind::host_side_scalar) {
+        kernel_ctx.define_int("HOST_SIDE_SCALE", 1);
+    }
```
umar456:
```suggestion
    kernel_ctx.define_int("HOST_SIDE_SCALE", pd()->scale_md()->format_kind == format_kind::host_side_scalar);
```

src/graph/backend/dnnl/kernels/sdp_primitive_config.cpp
```diff
@@ -376,9 +376,9 @@ status_t sdp_primitive_config_t::init(std::shared_ptr<subgraph_t> &sg,
             ? alg_kind::softmax_accurate_inf_as_zero
             : alg_kind::softmax_accurate;
     CHECK(create_sdpa_pd(sdpa_pd_, p_engine.get(), md_q.get(), md_k.get(),
-            md_v.get(), md_dst.get(), md_mask.get(), scale_dt, invert_scale_,
-            kv_head_number_, mask_type_, softmax_alg, attr.get(), qk_attr.get(),
-            vs_attr.get()));
+            md_v.get(), md_dst.get(), md_mask.get(), dnnl::memory::desc().get(),
```
umar456: This is creating an empty descriptor which is incorrect. You want to create a 1 element descriptor and pass that in.

tests/gtests/internals/test_sdpa.cpp
```diff
@@ -60,6 +60,7 @@ struct sdpa_dims_t {
     quantize_type qtype;
     bool with_key_transposed;
     mask_type mask;
+    bool with_host_side_scale;
```
umar456: Can you make this an enum instead of a bool?

tests/gtests/internals/test_sdpa.cpp
```diff
@@ -267,7 +275,8 @@ memory double_and_resize(const memory::desc &desc, dnnl::engine &eng) {
     return memory(desc, eng, handle);
 }
 
-sdpa_tensors_t get_descriptors(dnnl::engine &eng, const sdpa_dims_t &p) {
+sdpa_tensors_t get_descriptors(
```
umar456, line 279: can you reorder the parameters so that the engines are together.

tests/gtests/internals/test_sdpa.cpp
```diff
@@ -382,6 +390,21 @@ sdpa_tensors_t get_descriptors(dnnl::engine &eng, const sdpa_dims_t &p) {
     out.m_output = double_and_resize(output_md, eng);
     out.m_output_quantized = double_and_resize(output_quantized_md, eng);
 
+    // Populate scale parameter host or device side per test configuration
+    if (p.with_host_side_scale) {
+        out.m_scale
+                = memory(memory::desc({1, 1, 1, 1, 1}, p.qdt, abcde), cpu_eng);
+        out.scale_md = memory::desc([&] {
```
umar456: Shouldn't you be passing this descriptor into the memory object above?

tests/gtests/internals/test_utils.hpp
```diff
@@ -112,10 +112,21 @@ void write_to_dnnl_memory(const T *handle, dnnl::memory &mem) {
...
+        if (mem.get_desc().get_data_type() != dnnl_f32
+                && std::is_same<T, float>::value) {
+            dnnl::memory mem_f32_mem(
+                    {mem.get_desc().get_dims(), dnnl::memory::data_type::f32,
+                            mem.get_desc().get_strides()},
+                    eng);
+            write_to_dnnl_memory<float>((const float *)handle, mem_f32_mem);
+            dnnl::reorder(mem_f32_mem, mem).execute(s, mem_f32_mem, mem);
+        } else {
+            uint8_t *dst = static_cast<uint8_t *>(mem.get_data_handle());
```
umar456, line 124: I think this may fail with the sycl backend. Check out [this line](https://github.com/uxlfoundation/oneDNN/pull/3412/files#diff-9a0131644476f7146a3ca757dbbb0b93b42d1169748427de765e799d8600923fR100) for a solution.

tests/gtests/internals/test_sdpa.cpp
```diff
@@ -885,83 +937,91 @@ INSTANTIATE_TEST_SUITE_P(llama_2_7b_chat,
...
+                    sdpa_dims_t{   1,    32,       32,   2048,   2048,    128,   128,     128, mdt::f16,  mdt::f16,   mdt::s8,    mdt::f16,    mdt::s8,    mdt::s8,   mdt::f16,    mdt::s8, mdt::f16, quantize_type::per_token_with_groups,  with_key_transposed,  mask_type::causal_tl, with_host_side_scale },
+                    sdpa_dims_t{   1,    32,       32,   2049,      1,    128,   128,     128, mdt::f16,  mdt::f16,   mdt::s8,    mdt::f16,    mdt::s8,    mdt::s8,   mdt::f16,    mdt::s8, mdt::f16, quantize_type::per_token_with_groups,  with_key_transposed,  mask_type::causal_tl, with_host_side_scale },
+                    sdpa_dims_t{   1,    32,       32,    384,    384,    128,   128,     128, mdt::f16,  mdt::f16,   mdt::s8,    mdt::f16,    mdt::s8,    mdt::s8,   mdt::f16,    mdt::s8, mdt::f16, quantize_type::per_token_with_groups,  with_key_transposed,  mask_type::causal_tl, with_device_side_scale },
```
umar456: You don't need to double all these tests to handle host/device side scales. You should just set all of these bigger tests to host_side_scales and make a new test suite (using INSTANTIATE_TEST_SUITE_P) to tests different scenarios with just host_side_scale and device_side_scale as the only variable.

src/common/sdpa_utils.hpp
```diff
@@ -165,8 +165,8 @@ static inline status_t create_sdpa_pd(
         std::shared_ptr<primitive_desc_t> &sdpa_pd_, engine_t *engine,
         const memory_desc_t *q_md, const memory_desc_t *k_md,
         const memory_desc_t *v_md, const memory_desc_t *dst_md,
-        const memory_desc_t *attn_mask_md, data_type_t scale_dt,
-        bool invert_scale, dim_t kv_head_number,
+        const memory_desc_t *attn_mask_md, const memory_desc_t *scale_md,
+        data_type_t scale_dt, bool invert_scale, dim_t kv_head_number,
```
umar456: You don't need scale_dt here. That information can now be passed in through scale_md.

src/common/sdpa_utils.hpp
```diff
@@ -175,7 +175,7 @@ static inline status_t create_sdpa_pd(
             kq_attr, vs_attr));
```
umar456, line 175: You should add checks in `sdpa_desc_check` for scale_md.

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -211,8 +211,12 @@ DECLARE_2D_TILE_RSELECT(a_scale_tile_type, SUBGROUP_SIZE, ugemm_vs_sg_tile_n, 1,
 __attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE))) kernel void
 micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
         const global VAL_DATA_T *V, global DST_DATA_T *A,
-        const global SCALE_DATA_T *scale_ptr, int d, int k, int q,
-        const global KEY_ATTR_SCALES_DATA_T *K_scales,
+#ifdef HOST_SIDE_SCALE
+        const SCALE_DATA_T scale_value,
```
umar456:
```suggestion
        const float scale_value,
```

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -341,10 +345,18 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
     if (k0end > 0) {
 #if WITH_ATTN_SCALE
 #if INVERT_SCALE
+#ifdef HOST_SIDE_SCALE
+        iscale = SCALES_TO_FLOAT(scale_value);
```
umar456:
```suggestion
        iscale = scale_value;
```
Can you do this conversion on the host side? That way you don't need to specialize the kernel if its a host side scalar.

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -341,10 +345,18 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
...
+#else
         iscale = SCALES_TO_FLOAT(*scale_ptr);
+#endif
         scale = native_recip(iscale);
+#else
+#ifdef HOST_SIDE_SCALE
+        scale = SCALES_TO_FLOAT(scale_value);
```
umar456:
```suggestion
        scale = scale_value;
```

src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -559,7 +563,18 @@ status_t micro_sdpa_t::execute(const exec_ctx_t &ctx) const {
...
     arg_list.append(dst);
-    arg_list.append(scale);
+
+    if (pd()->scale_md()->format_kind == format_kind::host_side_scalar) {
+        void *handle = scale.data_handle();
+        if (pd()->scale_md()->data_type == data_type::f16) {
+            arg_list.append(*(float16_t *)handle);
+        } else if (pd()->scale_md()->data_type == data_type::bf16) {
+            arg_list.append(*(ushort *)handle);
+        }
```
umar456:
```suggestion
        float scale_value = 0.f;
        switch(pd()->scale_md()->data_type) {
            case data_type::f16: scale_value = static_case<float>(*(float16_t *)handle); break;
            case data_type::bf16: scale_value = static_cast<float>(*(bfloat16_t*)handle); break;
            case data_type::f32: scale_value = *(float *)handle; break;
            default: break;
        }
        arg_list.append(scale_value);
       
```

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -2825,9 +2825,9 @@ struct sdpa_executable_t : public op_executable_t {
                 ? alg_kind::softmax_accurate_inf_as_zero
                 : alg_kind::softmax_accurate;
         status_t s = create_sdpa_pd(sdpa_pd_, p_engine.get(), md_q.get(),
-                md_k.get(), md_v.get(), md_dst.get(), md_mask.get(), scale_dt,
-                is_invert_scale_, kv_head_number, mask_type_, softmax_alg,
-                attr.get());
+                md_k.get(), md_v.get(), md_dst.get(), md_mask.get(),
+                dnnl::memory::desc().get(), scale_dt, is_invert_scale_,
```
umar456: Same as the previous comment. Create a new descriptor with 1 element and pass that in instead of creating an empty descriptor.

tests/gtests/internals/test_sdpa.cpp
umar456: We should also probably ad the memory::data_type for scale as well.

tests/gtests/internals/test_sdpa.cpp
```diff
@@ -382,6 +390,21 @@ sdpa_tensors_t get_descriptors(dnnl::engine &eng, const sdpa_dims_t &p) {
     out.m_output = double_and_resize(output_md, eng);
     out.m_output_quantized = double_and_resize(output_quantized_md, eng);
 
+    // Populate scale parameter host or device side per test configuration
```
umar456:
```suggestion
    // Create scale descriptor/memory for host or device side
```

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -341,10 +345,18 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
     if (k0end > 0) {
 #if WITH_ATTN_SCALE
 #if INVERT_SCALE
```
umar456: We should be able to do the inversion on the host side as well. Ideally the host side scale should be able invert before the kernel is called as opposed to performing this operation on each work-item.

umar456:
You should also update the verbose log here: https://github.com/uxlfoundation/oneDNN/blob/main/src/common/verbose.cpp#L1562

src/common/sdpa_utils.hpp
pv-pterab-s: Yeah this is on the list of items to be absorbed. This is why this isn't quite ready for review, yet

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: So, this goes into the story of the micro sdpa kernel being fragile. I'm very hesitant to deviate from the original kernel too much at all since it is so easy to incur random failures.

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: This is a good idea. This'll go in if the kernel turns out to remain stable with the larger change. I'm a little hesitant to make the change part of this PR since the PR is explicitly for host-side-scalars and optimization on top can come later in an organized way.

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: I'll check on getting rid of this. Good idea. Another place, though, where I'm afraid to change the kernel too much in order to avoid random failures.

src/gpu/intel/ocl/micro_sdpa.cpp
pv-pterab-s: Sure - it'll be shorter.

tests/gtests/internals/test_sdpa.cpp
pv-pterab-s: Following the pattern of `with_key_transposed`, here.

tests/gtests/internals/test_sdpa.cpp
pv-pterab-s: This is the trick with the host-side-scale trick. The descriptor `scale_md` is only an indicator to primitives that the argument is a scalar kernel argument. The memory is independent since a different type on the CPU engine to copy data over.

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: Another place where things get tricky with the stability of the micro sdpa kernel. I'll try w/o this, but we'll see how it goes.

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: Yeah - I'm going to hold off on this one. This PR is explicitly to achieve host-scalar support. It's a big lift across so many platforms, etc, so lets be conservative.

src/gpu/intel/ocl/micro_sdpa.cpp
pv-pterab-s: Almost worked out.

pv-pterab-s:
make test
enable os_win
disable test_device_cpu
enable test_device_gpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
enable build_graph
enable compiler_icx-previous
enable compiler_gnu9
enable compiler_clang14
enable compiler_vs2022
disable benchdnn_all
enable benchdnn_softmax
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
disable arch_gpu_xe-lp
disable arch_gpu_xe-lpg
disable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
disable arch_gpu_xe2-lpg

src/common/verbose.cpp
```diff
@@ -1593,6 +1593,11 @@ std::string init_info_sdpa(const engine_t *e, const pd_t *pd) {
         }
         delimiter = " ";
     }
+    if (pd->with_host_side_scale()) {
+        ss << delimiter << "scale:host";
+    } else {
+        ss << delimiter << "scale:device";
+    }
```
atkassen: This should go down with the attention mask stuff (around L1615).

pv-pterab-s:
make test
enable os_win
disable test_device_cpu
enable test_device_gpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
enable build_graph
enable compiler_icx-previous
enable compiler_gnu9
enable compiler_clang14
enable compiler_vs2022
disable benchdnn_all
enable benchdnn_softmax
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
disable arch_gpu_xe-lp
disable arch_gpu_xe-lpg
disable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
disable arch_gpu_xe2-lpg

include/oneapi/dnnl/dnnl.h
```diff
@@ -989,6 +989,21 @@ dnnl_status_t DNNL_API dnnl_memory_desc_create_with_packed_encoding(
...
+/// object. It can only be used to create a primitive descriptor to
+/// query the actual memory descriptor (similar to the format tag
+/// `any`).
+///
+/// @param memory_desc Output memory descriptor.
+/// @param data_type Elements data type.
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+/// @sa @ref dev_guide_sparsity
+dnnl_status_t DNNL_API dnnl_memory_desc_create_host_side_scalar(
```
dzarukin, line 1004: C++ counter-part?

src/common/memory_desc.cpp
```diff
@@ -31,6 +31,20 @@ using namespace dnnl::impl::utils;
 namespace dnnl {
 namespace impl {
 
+status_t memory_desc_init_host_side_scalar(
+        memory_desc_t &memory_desc, data_type_t data_type) {
+    // filter supported scalar datatype
+    // TODO: shall we let any datatype through, especially for sub byte types?
+    VCHECK_MEMORY(utils::one_of(data_type, data_type::f32, data_type::f16,
+                          data_type::bf16),
+            invalid_arguments, VERBOSE_MEM_DESC_CHECK_FAIL);
```
dzarukin: Unsupported data type instead?

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -335,6 +339,7 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
 #endif
     }
 
+#if HOST_SIDE_SCALE == 0
```
dzarukin:
```suggestion
#if !HOST_SIDE_SCALE
```

src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -503,6 +503,9 @@ status_t micro_sdpa_t::init(impl::engine_t *engine) {
     kernel_ctx.define_int("SOFTMAX_INF_AS_ZERO",
             d->softmax_alg == alg_kind::softmax_accurate_inf_as_zero);
 
+    kernel_ctx.define_int("HOST_SIDE_SCALE",
+            pd()->scale_md()->format_kind == format_kind::host_side_scalar);
```
dzarukin: Re-use `with_host_side_scale()`?

src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -559,7 +562,37 @@ status_t micro_sdpa_t::execute(const exec_ctx_t &ctx) const {
     arg_list.append(qry);
     arg_list.append(val);
     arg_list.append(dst);
-    arg_list.append(scale);
+
+    if (pd()->scale_md()->format_kind == format_kind::host_side_scalar) {
```
dzarukin: Ditto.

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -211,8 +211,12 @@ DECLARE_2D_TILE_RSELECT(a_scale_tile_type, SUBGROUP_SIZE, ugemm_vs_sg_tile_n, 1,
 __attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE))) kernel void
 micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
         const global VAL_DATA_T *V, global DST_DATA_T *A,
-        const global SCALE_DATA_T *scale_ptr, int d, int k, int q,
-        const global KEY_ATTR_SCALES_DATA_T *K_scales,
+#if HOST_SIDE_SCALE
+        const float scale, const float iscale,
+#else
+        const global SCALE_DATA_T *scale_ptr,
+#endif
```
dzarukin: It feels like i(nverted)scale can be computed down below to have the same number of arguments.
Once one for both cases, I'd suggest to introduce `SCALE_T` (or something along the way) that will be `float` or `global SCALE_DATA_T *` depending on whether it's a host side scale or not.

Macros in kernel signature is hard to deal with at caller and callee levels, better avoid it whenever possible.

src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -559,7 +562,37 @@ status_t micro_sdpa_t::execute(const exec_ctx_t &ctx) const {
...
+        const void *handle = scale.data_handle();
+
+        const float scale_value = [&]() {
+            switch (pd()->scale_md()->data_type) {
+                case data_type::f16: return float(*(float16_t *)handle);
+                case data_type::bf16: return float(*(bfloat16_t *)handle);
+                case data_type::f32: return *(float *)handle;
+                default: return 0.f;
```
dzarukin: `assert` here will help in the future.

src/gpu/intel/ocl/micro_sdpa.cpp
```diff
@@ -559,7 +562,37 @@ status_t micro_sdpa_t::execute(const exec_ctx_t &ctx) const {
...
+        if (pd()->with_attn_scale()) {
+            if (pd()->desc()->invert_scale) {
+                iscale = scale_value;
+                scale = 1.f / iscale;
+            } else {
+                scale = scale_value;
+                iscale = 1.f / scale;
+            }
+        }
+        scale *= 1.442695f;
```
dzarukin: Please use named `constexpr` instead.

dzarukin (commented):
What about benchdnn validation?

include/oneapi/dnnl/dnnl.h
```diff
@@ -1002,8 +1002,7 @@ dnnl_status_t DNNL_API dnnl_memory_desc_create_with_packed_encoding(
 ///     otherwise.
 /// @sa @ref dev_guide_sparsity
 dnnl_status_t DNNL_API dnnl_memory_desc_create_host_side_scalar(
-        dnnl_memory_desc_t *memory_desc,
-        dnnl_data_type_t data_type);
+        dnnl_memory_desc_t *memory_desc, dnnl_data_type_t data_type);
```
mgouicem, line 1005: Looks like fixup for first commit, could you just amend that one?

src/common/primitive_hashing.cpp
```diff
@@ -122,7 +122,8 @@ size_t get_md_hash(const memory_desc_t &md) {
     // format desc
     switch ((int)md.format_kind) {
         case format_kind::undef:
-        case format_kind::any: break;
+        case format_kind::any:
+        case format_kind::host_side_scalar: break;
```
mgouicem, line 126: Looks like fixup for first commit, could you just amend that one?

src/common/primitive_hashing.cpp
```diff
@@ -734,14 +735,14 @@ size_t get_desc_hash(const sdpa_desc_t &desc) {
     seed = hash_combine(seed, get_md_hash(desc.q_desc));
     seed = hash_combine(seed, get_md_hash(desc.k_desc));
     seed = hash_combine(seed, get_md_hash(desc.v_desc));
+    seed = hash_combine(seed, get_md_hash(desc.scale_desc));
```
mgouicem, line 738: Looks like fixup for first commit, could you just amend that one?

src/common/primitive_serialization.cpp
```diff
@@ -577,13 +577,13 @@ void serialize(serialization_stream_t &sstream, const sdpa_desc_t &desc) {
     serialize(sstream, desc.q_desc);
     serialize(sstream, desc.k_desc);
     serialize(sstream, desc.v_desc);
+    serialize(sstream, desc.scale_desc);
```
mgouicem, line 580: Looks like fixup for first commit, could you just amend that one?

src/common/memory_desc.cpp
pv-pterab-s: I think it best to allow any datatype though. The example primitive picking up host-side-scalar support (SDPA) pulls the datatype of the scale parameter from the associated descriptor. The information conveyed will be the same for other primitives - indicate a host-side-scalar and specify the datatype.

pv-pterab-s:
> What about benchdnn validation?

SDPA is an internal-only primitive, so the only checking is in the internal testing and not benchdnn.

src/common/verbose.cpp
pv-pterab-s: How's this...  Just asking to make sure I'm upholding the format, here. It is a little haphazard it seems...

```suggestion
    ss << "scale:" << (pd->with_host_side_scale() ? "host" : "device") << ",";

    if (pd->with_attn_mask()) {
        auto *md = pd->attn_mask_md();
        ss << "msk:" << (md->dims[2] == 1 ? 1 : 2) << 'd';

    } else if (pd->with_causal_mask()) {
        if (desc->mask_type == attn_mask_type::top_left)
            ss << "msk:causal:top_left";
        else
            ss << "msk:causal:bottom_right";
    }
```

include/oneapi/dnnl/dnnl.h
pv-pterab-s: Done.

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: See my attempt at commit: https://github.com/uxlfoundation/oneDNN/commit/fd3c5eb5c21a7102080ad017f35a9096b5c51c63 . So, this is one argument, but to me this obfuscates things and puts a division (to get iscale) back into the kernel. @umar456 and I went back and forth on this and pulled all compute back to the host. Less work device side. Also, `WITH_ATTN_MASK` is a macro in the kernel signature already, so the mess may have already been made. Lemme know!

dzarukin: If the scale is host scalar, it can be actually passed as pointer, thus, no need to have a different argument signature. And it can be passed already inverted so the kernel can always multiply on it unconditionally, this information is available at execution time and access to the scale is identical to the access from the kernel.

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -2967,6 +2969,23 @@ struct memory : public handle<dnnl_memory_t> {
...
 
+        /// Constructs a memory descriptor describing a host-side
+        /// scalar to be directly passed as an argument to device-side
+        /// kernels. Intended to avoid memory transfer overhead on
+        /// primitive execution. The resulting descriptor cannot be
+        /// utilized to generate a memory object.
+        ///
+        /// @param adata_type Data type of the scalar.
+        /// @returns A memory descriptor for host-side scalar input.
+        desc(data_type adata_type) {
```
dzarukin: I feel it's better to follow the sparse signature `static desc host_scalar(dt)` instead to have:
* A clear name;
* Untied hands if any extension will be needed;
* Confusion why just a data type ctor leads to a host-side memory and not anything else;
 
@mgouicem, any thoughts on this one?

dzarukin:
> > What about benchdnn validation?
> 
> SDPA is an internal-only primitive, so the only checking is in the internal testing and not benchdnn.

OK (just because there's no Graph extension for this feature in this PR), then, at least new API tests should appear in gtests folder.

include/oneapi/dnnl/dnnl.hpp
mgouicem: I agree it would make the intent clearer with a dedicated static method than with desc constructor.

src/common/primitive_hashing.cpp
pv-pterab-s: This one isn't - - rather specifically for SDPA and not general API

src/common/primitive_serialization.cpp
pv-pterab-s: This is another one specific to SDPA. This goes in the following commits SDPA specific and not in general API.

include/oneapi/dnnl/dnnl.hpp
pv-pterab-s: Done.

include/oneapi/dnnl/dnnl.h
densamoilov, line 1004: The [RFC ](https://github.com/uxlfoundation/oneDNN/pull/3236) hasn't been approved yet and from the discussion there it's not clear what option we are going with. Can you please give an update on the situation?

src/common/memory_desc.cpp
```diff
@@ -31,6 +31,15 @@ using namespace dnnl::impl::utils;
 namespace dnnl {
 namespace impl {
 
+status_t memory_desc_init_host_side_scalar(
+        memory_desc_t &memory_desc, data_type_t data_type) {
+    memory_desc.ndims = 1;
+    memory_desc.dims[0] = 1;
```
densamoilov, line 37: I think the scalar is dimensionless in our context, so defining a dimension for it seems unnecessary.

src/common/primitive_hashing.cpp
densamoilov, line 126: We should hash the data type. 

And also serialize it (serialization part it's missing now).

src/common/type_helpers.hpp
```diff
@@ -979,13 +979,13 @@ inline bool operator==(const sdpa_desc_t &lhs, const sdpa_desc_t &rhs) {
             && COMPARE_DESC_MEMBERS(q_desc)
             && COMPARE_DESC_MEMBERS(k_desc)
             && COMPARE_DESC_MEMBERS(v_desc)
+            && COMPARE_DESC_MEMBERS(scale_desc)
```
densamoilov, line 982: (in this file) `operator==` for memory desc should be updated to accommodate the new format kind,

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -2954,6 +2956,23 @@ struct memory : public handle<dnnl_memory_t> {
             return desc {md};
         }
 
+        /// Constructs a memory descriptor describing a host-side
```
densamoilov, line 2954: (random spot) 
How do we create a memory object for the new host scalar descriptor? Also, when we create it, in what format should the scalar be stored? Currently, we always rely on memory storage that is tied to an engine, but I assume the goal is for the scalar to be engine-agnostic.

Since the API doesn’t seem to be defined yet, one possible option could look like:
```cpp
template <typename T>
memory(const desc &md, T scalar) {
    dnnl_memory_t result;
    dnnl_status_t status = dnnl_memory_create_host_side_scalar(&result, md.get(), &scalar);
    // Or the following, if we don't want to introduce new C API:
    // dnnl_status_t status = dnnl_memory_create(&result, md.get(), nullptr, &scalar);
    // ...
}
```
```cpp
auto mem = dnnl::memory(memory::desc::host_side_scalar(data_type::f32), 13.5f);
```

But this assumes that we decide to create memory objects even for scalars. I think that was part of the discussion in the RFC that has hasn't been resolved yet.

src/gpu/intel/ocl/micro_sdpa.cl
dzarukin: Discussed offline.
Due to the CPU nature of the host scalar, it can't be passed inside the kernel as a pointer, only as a plain float value.
If the final decision is to go with two values, agreed on putting a comment to explain the benefits of it. Otherwise, the scale handling would be self-documented in the code.

src/gpu/intel/ocl/micro_sdpa.cl
```diff
@@ -225,11 +225,16 @@ DECLARE_2D_TILE_RSELECT(a_scale_tile_type, SUBGROUP_SIZE, ugemm_vs_sg_tile_n, 1,
...
+        int d, int k, int q, const global KEY_ATTR_SCALES_DATA_T *K_scales,
```
pv-pterab-s, line 237: DISCUSSION: There may be an issue with OpenVINO dealing with dual macros in this kernel signature. It has been suggested that we may be able to deal with the issue by relying on an USM approach: a pointer is passed to memory containing the scalar. 

On the other hand, this approach reduces compute inside the kernel and keeps typing of `scale` with convention. It also relies on passing a scalar to the kernel through absolutely basic OpenCL reducing any unexpected behavior due to driver errors, synchronization, etc.

include/oneapi/dnnl/dnnl.h
pv-pterab-s: This was option 2.

src/common/memory_desc.cpp
pv-pterab-s: I'm not sure about this one. It seems as though a dimensionless descriptor would create an edge case in some circumstances? Correct me, however, as I am not familiar with all the various uses of the descriptor throughout oneDNN.

pv-pterab-s: Per @mzhukova ,  design discussion to continue on associated rfc: https://github.com/uxlfoundation/oneDNN/blob/mgouicem/rfcs/host_scalar/rfcs/20250430-host-side-scalars/README.md

include/oneapi/dnnl/dnnl.hpp
pv-pterab-s: Per @mzhukova , design discussion to continue on associated rfc: https://github.com/uxlfoundation/oneDNN/blob/mgouicem/rfcs/host_scalar/rfcs/20250430-host-side-scalars/README.md

src/gpu/intel/ocl/micro_sdpa.cl
pv-pterab-s: Per @mzhukova , design discussion to continue on associated rfc: https://github.com/uxlfoundation/oneDNN/blob/mgouicem/rfcs/host_scalar/rfcs/20250430-host-side-scalars/README.md

include/oneapi/dnnl/dnnl.hpp
TaoLv: This also requires the scalar value to be stored in memory descriptor or memory as the address/lifetime of the scalar value may become invalid after the creation.

mgouicem: Yes correct, we need to store the value in memory object.

mzhukova: resolving this thread, as this is addressed in RFC :)

src/common/primitive_hashing.cpp
mzhukova: @densamoilov this would be incorporate in impl / POC https://github.com/uxlfoundation/oneDNN/pull/3506 probably, so I'm resolving this and the next thread and making it a note to myself to take care of it
Thanks for the input!

src/common/memory_desc.cpp
mzhukova: @densamoilov I agree with @pv-pterab-s, this would be the behavior if we create a memory object for scalar with current interface, right? we should also continue discussion in https://github.com/uxlfoundation/oneDNN/pull/3506 probably, so resolving this

TaoLv:
Closing this one.
- The API RFC #3236 has been implemented via #3506 by @mzhukova .
- The SDPA ukernel support has been implemented via #3909 by @skazakov1 .

PR #3413, opened 2025-06-11
graph: handle different engine instances in compiled partition cache

gyhintel:
Fix MFDNN-12121
In this PR,  we add the `reset_engine` method to reset the compiled partition's and kernel's engine. With this PR, the compiled partition and primitive cache can be hit for different engine cases (Same member functions and member variables)

### DIFF
--- a/src/graph/backend/dnnl/kernels/batch_norm.hpp
+++ b/src/graph/backend/dnnl/kernels/batch_norm.hpp
@@ -42,7 +42,6 @@ struct batch_norm_fwd_t : public kernel_base_t {
 private:
     allocator_t *g_alloc_ = nullptr;
 
-    std::shared_ptr<subgraph_t> subgraph_;
     memory_planner_t memory_planner_;
 
     std::function<std::shared_ptr<execution_args_set_t>()> resource_ctor_;
@@ -98,7 +97,6 @@ struct batch_norm_bwd_t : public kernel_base_t {
 private:
     allocator_t *g_alloc_ = nullptr;
 
-    std::shared_ptr<subgraph_t> subgraph_;
     memory_planner_t memory_planner_;
 
     std::function<std::shared_ptr<execution_args_set_t>()> resource_ctor_;
--- a/src/graph/backend/dnnl/kernels/layer_norm.hpp
+++ b/src/graph/backend/dnnl/kernels/layer_norm.hpp
@@ -43,7 +43,6 @@ struct layer_norm_fwd_t : public kernel_base_t {
 private:
     allocator_t *g_alloc_ = nullptr;
 
-    std::shared_ptr<subgraph_t> subgraph_;
     memory_planner_t memory_planner_;
 
     std::function<std::shared_ptr<execution_args_set_t>()> resource_ctor_;
@@ -99,7 +98,6 @@ struct layer_norm_bwd_t : public kernel_base_t {
 private:
     allocator_t *g_alloc_ = nullptr;
 
-    std::shared_ptr<subgraph_t> subgraph_;
     memory_planner_t memory_planner_;
 
     std::function<std::shared_ptr<execution_args_set_t>()> resource_ctor_;
--- a/src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
@@ -474,6 +474,31 @@ dnnl::primitive_attr mqa_decomp_config_t::make_primitive_attr(
     return attr;
 }
 
+#define DECLARE_RESET_ENGINE(primitive_name, primitive_type) \
+    { \
+        const auto desc_t = (primitive_name).get_primitive_desc()->impl(); \
+        dnnl_primitive_desc new_pd_t(desc_t, p_engine.get()); \
+        primitive_type::primitive_desc new_pd(&new_pd_t); \
+        (primitive_name) = primitive_type(new_pd); \
+    }
+
+impl::status_t mqa_decomp_config_t::reset_engine(const dnnl::engine &p_engine) {
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(1);
+#endif
+    DECLARE_RESET_ENGINE(sub_mm1_prim, matmul);
+    DECLARE_RESET_ENGINE(sub_mm2_prim, matmul);
+    DECLARE_RESET_ENGINE(sub_softmax_prim, softmax_forward);
+    sub_reorder0.reset_engine(p_engine);
+    sub_reorder1.reset_engine(p_engine);
+    sub_reorder2.reset_engine(p_engine);
+    sub_reorder3.reset_engine(p_engine);
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(nthr);
+#endif
+    return dnnl_success;
+}
+
 template status_t
 mqa_decomp_config_t::construct_params<false, dnnl::memory::data_type::f32>(
         std::shared_ptr<subgraph_t> &sg, registry_t &mqa_registry,
--- a/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
@@ -856,6 +856,32 @@ dnnl::primitive_attr sdp_decomp_config_t::make_primitive_attr(
     return attr;
 }
 
+#define DECLARE_RESET_ENGINE(primitive_name, primitive_type) \
+    { \
+        const auto desc_t = (primitive_name).get_primitive_desc()->impl(); \
+        dnnl_primitive_desc new_pd_t(desc_t, p_engine.get()); \
+        primitive_type::primitive_desc new_pd(&new_pd_t); \
+        (primitive_name) = primitive_type(new_pd); \
+    }
+
+impl::status_t sdp_decomp_config_t::reset_engine(const dnnl::engine &p_engine) {
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(1);
+#endif
+    DECLARE_RESET_ENGINE(sub_mm1_prim, matmul);
+    DECLARE_RESET_ENGINE(sub_softmax_prim, softmax_forward);
+    DECLARE_RESET_ENGINE(sub_mm2_prim, matmul);
+    if (has_select) DECLARE_RESET_ENGINE(sub_select_prim, binary);
+    sub_reorder0.reset_engine(p_engine);
+    sub_reorder1.reset_engine(p_engine);
+    sub_reorder2.reset_engine(p_engine);
+    sub_reorder3.reset_engine(p_engine);
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(nthr);
+#endif
+    return dnnl_success;
+}
+
 template status_t
 sdp_decomp_config_t::construct_params<false, dnnl::memory::data_type::f32>(
         std::shared_ptr<subgraph_t> &sg, registry_t &mqa_registry,
--- a/src/graph/backend/dnnl/kernels/sdp_primitive.hpp
+++ b/src/graph/backend/dnnl/kernels/sdp_primitive.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -45,7 +45,6 @@ struct sdp_primitive_kernel_t : public kernel_base_t {
 private:
     allocator_t *g_alloc_ = nullptr;
 
-    std::shared_ptr<subgraph_t> subgraph_;
     memory_planner_t memory_planner_;
     std::function<std::shared_ptr<execution_args_set_t>()> resource_ctor_;
 
@@ -97,6 +96,10 @@ public:
 
     DEF_KERNEL_METHOD_STR(sdp_primitive_kernel_t)
     DNNL_DISALLOW_COPY_AND_ASSIGN(sdp_primitive_kernel_t)
+    status_t reset_engine(const engine_t *g_engine) override {
+        UNUSED(g_engine);
+        return dnnl_success;
+    }
 };
 
 } // namespace dnnl_impl
--- a/tests/gtests/graph/unit/interface/test_compiled_partition.cpp
+++ b/tests/gtests/graph/unit/interface/test_compiled_partition.cpp
@@ -190,13 +190,26 @@ TEST(test_interface_compiled_partition, CacheEngine) {
         std::vector<const impl::graph::logical_tensor_t *> inputs {&input};
         std::vector<const impl::graph::logical_tensor_t *> outputs {&output};
         // Partition compilation
-        par.compile(cpcache, inputs, outputs, eng);
+        ASSERT_EQ(par.compile(cpcache, inputs, outputs, eng),
+                impl::status_t::dnnl_success);
+
+        // Partition execution
+        dnnl::stream strm(engine);
+        impl::stream_t *strm_t = strm.get();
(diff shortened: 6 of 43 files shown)


gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/sdp_primitive.hpp
```diff
@@ -97,6 +97,10 @@ struct sdp_primitive_kernel_t : public kernel_base_t {
 
     DEF_KERNEL_METHOD_STR(sdp_primitive_kernel_t)
     DNNL_DISALLOW_COPY_AND_ASSIGN(sdp_primitive_kernel_t)
+    status_t reset_engine(const engine_t *g_engine) override {
+        UNUSED(g_engine);
```
ElaineBao, line 100: why unused here?

gyhintel: For `sdp_primitive_kernel_t`, the executable object is `primitive_t`[L70](https://github.com/uxlfoundation/oneDNN/blob/9136964f8cc9e9c8a3ef4e3d38eee1baff68ccc4/src/graph/backend/dnnl/kernels/sdp_primitive_config.hpp#L70). For `primitive_t` object, there is no `engine` in this object. So we don't need to reset engine for it.

gyhintel:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/dnnl_partition_impl.cpp
```diff
@@ -113,7 +113,7 @@ status_t dnnl_partition_impl_t::compile(
         compiled_partition_t *compiled_partition,
         const std::vector<logical_tensor_t> &inputs,
         const std::vector<logical_tensor_t> &outputs,
-        const engine_t *g_engine) const {
+        engine_t *g_engine) const {
```
TaoLv: What does this mean? We do have a `const` engine in the public compile API. I assume that const-ness will propagate from API to internal implementations.

src/graph/backend/dnnl/kernels/batch_norm.hpp
```diff
@@ -91,6 +91,7 @@ struct batch_norm_fwd_t : public kernel_base_t {
 
     DEF_KERNEL_METHOD_STR(batch_norm_fwd_t)
     DNNL_DISALLOW_COPY_AND_ASSIGN(batch_norm_fwd_t)
+    KERNEL_RESET_ENGINE
```
TaoLv:
```suggestion
    DEF_KERNEL_METHOD_RESET_ENGINE
```
Just try to align with the name of `DEF_KERNEL_METHOD_STR`.

src/graph/backend/dnnl/dnnl_partition_impl.cpp
gyhintel: There is no `const` engine in the public compile C API. eg. [L415](https://github.com/uxlfoundation/oneDNN/blob/4d668a58bb00df820de33762f107d2df1d35bbd3/include/oneapi/dnnl/dnnl_graph.h#L415). The `const` engine is for the C++ API, [L1253](https://github.com/uxlfoundation/oneDNN/blob/4d668a58bb00df820de33762f107d2df1d35bbd3/include/oneapi/dnnl/dnnl_graph.hpp#L1253). The const-ness is for the engine handle, not an engine object.

src/graph/backend/dnnl/kernels/batch_norm.hpp
gyhintel: Changed, thanks!

src/graph/backend/dnnl/dnnl_partition_impl.cpp
TaoLv: Why do we need to remove `const` here? Also I see both `reset_engine(engine_t *)` and `reset_engine(const engine_t *)` across the PR.

gyhintel: > Why do we need to remove const here?

The key reason is that I removed the `const` of the `engine_` in the `compiled_partition_impl_t` [L378](https://github.com/uxlfoundation/oneDNN/blob/aa97c10e184655c5fd823b0ad28352c39bca210f/src/graph/interface/partition_impl.hpp#L378). In the new design mechanism, the `engine_` is mutable. So there is no const in `reset_engine` related to `compiled_partition_impl_t::engine_`, but others have `const` without relation to `compiled_partition_impl_t::engine_`.

xiang1guo:
Do we have any gtest case to test such scenario that compiled partition with different engines can cache hit?

src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
```diff
@@ -467,6 +467,31 @@ dnnl::primitive_attr mqa_decomp_config_t::make_primitive_attr(
     return attr;
 }
 
+#define DECLARE_RESET_ENGINE(primitive_name, primitive_type) \
```
rongzha1, line 477: same macro is used in sdp_decomp.hpp. Maybe we can move into to common place to avoid duplicate?

src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
```diff
@@ -467,6 +467,31 @@ dnnl::primitive_attr mqa_decomp_config_t::make_primitive_attr(
...
+    }
+
+impl::status_t mqa_decomp_config_t::reset_engine(const dnnl::engine &p_engine) {
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(1);
+#endif
+    DECLARE_RESET_ENGINE(sub_mm1_prim, matmul);
+    DECLARE_RESET_ENGINE(sub_mm2_prim, matmul);
+    DECLARE_RESET_ENGINE(sub_softmax_prim, softmax_forward);
+    sub_reorder0.reset_engine(p_engine);
```
rongzha1, line 492: why reorder can't use DECLARE_RESET_ENGINE?

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
```diff
@@ -822,6 +822,32 @@ dnnl::primitive_attr sdp_decomp_config_t::make_primitive_attr(
...
+    { \
+        const auto desc_t = (primitive_name).get_primitive_desc()->impl(); \
+        dnnl_primitive_desc new_pd_t(desc_t, p_engine.get()); \
+        primitive_type::primitive_desc new_pd(&new_pd_t); \
+        (primitive_name) = primitive_type(new_pd); \
+    }
+
+impl::status_t sdp_decomp_config_t::reset_engine(const dnnl::engine &p_engine) {
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_OMP
+    omp_set_num_threads(1);
```
rongzha1, line 869: why to set thread num as 1? If we want to avoid race conditions caused by multithreading, we should use locking instead of changing the number of threads inside the function.

gyhintel:
> Do we have any gtest case to test such scenario that compiled partition with different engines can cache hit?

Yes, there is a gtest case to test this scenario in `tests/gtests/graph/unit/interface/test_compiled_partition.cpp`.

src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
gyhintel: Yes, the same macro is used in both of these files. But the common heads of these files are not in the `src/graph/backend/dnnl/kernels` folder. They are in the `common` or `graph/backend/dnnl`  folders. I think putting the macro in these folders or creating a new header file in the `kernels` folder is not good.

src/graph/backend/dnnl/kernels/mqa_decomp_config.cpp
gyhintel: This reorder is not a primitive type. It is an internal type that includes a reorder primitive in [L44-L69](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/kernels/mqa_decomp_config.hpp#L44-L69)

src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp
gyhintel: Actually, it is not to avoid race conditions. Because we create primitive first with `omp_set_num_threads(1)`,  [L150](https://github.com/uxlfoundation/oneDNN/blob/972b9e1475bed71cfaa30c8b82ca3ea5c7838617/src/graph/backend/dnnl/kernels/sdp_decomp_config.cpp#L150). The thread_num is one of the primitive cache key info. So I set thread number 1 to hit the primitive cache.

xiang1guo:
Added allocator into partition key in last commit, please review again, thanks! @TaoLv @rongzha1 @ElaineBao  cc @gyhintel

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/dnnl_partition_impl.cpp
TaoLv: Just to confirm, i assume that this is the engine instance passed by user through API. Are you modifying it inside this function.

xiang1guo: The `compile` function here does not modify the `g_engine` variable itself; however, because constructing the `dnnl_compiled_partition_impl_t` object in this func ([code](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/backend/dnnl/dnnl_partition_impl.cpp#L167)) requires a non-const engine (used to set none-const member engine [here](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/interface/partition_impl.hpp#L280)), the function's parameter must also be non-const.

TaoLv: > because constructing the dnnl_compiled_partition_impl_t object in this func ([code](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/backend/dnnl/dnnl_partition_impl.cpp#L167)) requires a non-const engine (used to set none-const member engine [here](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/interface/partition_impl.hpp#L280)),

Is it because `dnnl_compiled_partition_impl_t` object will change the status of the non-const member engine, or just reset it to point to a new engine object?

xiang1guo: >> because constructing the dnnl_compiled_partition_impl_t object in this func ([code](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/backend/dnnl/dnnl_partition_impl.cpp#L167)) requires a non-const engine (used to set none-const member engine [here](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/interface/partition_impl.hpp#L280)),

>Is it because `dnnl_compiled_partition_impl_t` object will change the status of the non-const member engine, or just reset it to point to a new engine object?

The latter one, just reset it to point to a new engine object.

TaoLv: Then please consider to keep the `const` for the function parameters and use `const_cast` when it's necessary. From your description, we still guarantee the const-ness of the input engine object through our implementation.

xiang1guo: Thanks for your advice! I think that make sense, I added a new commit for you to review and added a description [here ](https://github.com/uxlfoundation/oneDNN/blob/gyh/main/cp_eigine/src/graph/backend/dnnl/dnnl_partition_impl.hpp#L57)to avoid possible issue in the future. Please review again~

ElaineBao (approved):

src/graph/backend/fake/fake_partition_impl.hpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2023 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
```
TaoLv: revert this one?

src/graph/backend/dnnl/dnnl_partition_impl.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2024 Intel Corporation
+ * Copyright 2024-2025 Intel Corporation
```
TaoLv: revert this one.

src/graph/backend/dnnl/kernels/layer_norm.hpp
```diff
@@ -152,6 +153,7 @@ struct layer_norm_bwd_t : public kernel_base_t {
 
     DEF_KERNEL_METHOD_STR(layer_norm_bwd_t)
     DNNL_DISALLOW_COPY_AND_ASSIGN(layer_norm_bwd_t)
+    DEF_KERNEL_METHOD_RESET_ENGINE
```
TaoLv: random place: is it possible to put this function into the base class?

xiang1guo: I feel that's doable, but we need to add `subgraph_` as a new member of `kernel_base_t` class, and implement a virtual function `reset_engine()`, for those kernels with different implementations, may need to override this function. Do you have any concerns about this?

tests/gtests/graph/unit/interface/test_compiled_partition.cpp
```diff
@@ -190,13 +190,26 @@ TEST(test_interface_compiled_partition, CacheEngine) {
...
     }
 
 #ifdef DNNL_GRAPH_DISABLE_COMPILED_PARTITION_CACHE
     ASSERT_EQ(get_compiled_partition_cache_size(), 0);
 #else
-    ASSERT_EQ(get_compiled_partition_cache_size(), static_cast<int>(batch_num));
+    if (get_test_engine_kind() == impl::graph::engine_kind::cpu) {
+        ASSERT_EQ(get_compiled_partition_cache_size(), 1);
+    } else if (get_test_engine_kind() == impl::graph::engine_kind::gpu) {
+        ASSERT_GE(get_compiled_partition_cache_size(), batch_num);
```
rongzha1, line 211: why gpu cache size is batch_num while cpu cache is 1 ?

xiang1guo: GPU cannot cache hit with a different engine because [here ](https://github.com/uxlfoundation/oneDNN/blob/main/src/xpu/sycl/engine_id.hpp#L46) uses engine `context_` and `device_` as part of cache key. This behavior is aligned with GPU primitive behavior.

src/graph/backend/dnnl/kernels/layer_norm.hpp
xiang1guo: I added one more last commit to reflect this suggestion, as I mentioned, one thing different is that I need to put `subgraph_` into base class to implement `reset_engine` function. 
I recall that a previous implementation in the SDPA decompose kernel, where a new `select_subgraph_` was introduced to support select operations in specific sdpa kernel. This means that different kernels might require multiple `subgraph_ object`, such a change may lead to potential confusion in the future, I guess. This is my concern of putting `subgraph_` into the base class. Please help to review~

TaoLv (approved):

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3414, opened 2025-06-11
cpu: aarch64: brgemm: Add support for int8 in brgemm kernel

kasturedeeksha:
# Description

This PR extends the BRGEMM (Batch-Reduce General Matrix Multiplication) kernel to support additional INT8 data types, enabling broader applicability for low-precision computations, particularly in deep learning workloads.

Supported Data Type Tags
The following source:weight:destination (src:wei:dst) combinations are now supported:

- s8:s8:f32
- u8:u8:f32
- u8:s8:f32

# Checklist

## General

1. make test output 

```
98% tests passed, 4 tests failed out of 224
 
Total Test time (real) = 1058.22 sec
 
The following tests FAILED:
        172 - test_graph_unit_dnnl_large_partition_cpu (Failed)
        195 - test_benchdnn_modeC_binary_ci_cpu (Failed)
        196 - test_benchdnn_modeC_binary_different_dt_ci_cpu (Failed)
        204 - test_benchdnn_modeC_graph_ci_cpu (Failed)
```

2. brgemm_test_all output 
command used : 

`./benchdnn --brgemm --batch=inputs/brgemm/test_brgemm_all`

Before 
```
tests:660480 passed:18496 skipped:641984 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total: 24.50s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 7.30s (30%); execute: 0.00s (0%); compute_ref: 4.30s (18%); compare: 5.34s (22%);
```
After
```   
tests:660480 passed:20480 skipped:640000 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total: 23.11s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 6.83s (30%); execute: 0.00s (0%); compute_ref: 3.82s (17%); compare: 4.40s (19%);
```
Yes

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2021-2023 Intel Corporation
-* Copyright 2024 FUJITSU LIMITED
+* Copyright 2024-2025 FUJITSU LIMITED
 * Copyright 2024 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
@@ -509,12 +509,20 @@ void jit_brgemm_kernel_t::cvt2ps(data_type_t type_in, const ZReg zmm_in,
                 mov(zmm_in.s, ktail_mask / T_m, z_tmp_1().s);
             break;
         case data_type::bf16: assert(!"unsupported data type\n"); break;
-        case data_type::s8: assert(!"unsupported data type\n"); break;
-        case data_type::u8: assert(!"unsupported data type\n"); break;
+        case data_type::s8:
+            LD_MUL_VL(ld1b, z_tmp_1().b, mask, addr, offset - base_offset, 1);
+            sxtb(z_tmp_1().d, mask / T_m, z_tmp_1().d);
+            if (store) // Merging
+                mov(zmm_in.s, ktail_mask / T_m, z_tmp_1().s);
+            break;
+        case data_type::u8:
+            LD_MUL_VL(ld1b, z_tmp_1().b, mask, addr, offset - base_offset, 1);
+            uxtb(z_tmp_1().s, mask / T_m, z_tmp_1().s);
+            if (store) // Merging
+                mov(zmm_in.s, ktail_mask / T_m, z_tmp_1().s);
+            break;
         default: assert(!"unsupported data type");
     }
-    if (!one_of(type_in, data_type::f32, data_type::bf16))
-        assert(!"unsupported data type\n");
 }
 
 void jit_brgemm_kernel_t::advance_ldb_post_op_regs() {
@@ -961,7 +969,9 @@ void jit_brgemm_kernel_t::store_accumulators_apply_post_ops(
         }
         for (int bd = 0; bd < bd_block; bd++) {
             auto zmm = accm(ld_block2, bd, ld);
-            if (dq2ps_required) { scvtf(zmm.s, P_ALL_ONE / T_m, zmm.s); }
+            if (dq2ps_required && !brg.with_scales) {
+                scvtf(zmm.s, P_ALL_ONE / T_m, zmm.s);
+            }
             if (brg.with_bias) { fadd(zmm.s, zmm.s, zmm_bias.s); }
         }
     }
@@ -989,7 +999,7 @@ void jit_brgemm_kernel_t::store_accumulators_apply_post_ops(
         auto vmm_zp_c = z_tmp_1();
         if (brg.zp_type_c == brgemm_broadcast_t::per_tensor) {
             add_imm(X_DEFAULT_ADDR, reg_aux_zp_c_values, 0, X_TMP_0);
-            ldr(z_tmp_2(), ptr(X_DEFAULT_ADDR));
+            ld1rw(z_tmp_2().s, P_ALL_ONE, ptr(X_DEFAULT_ADDR));
             scvtf(vmm_zp_c.s, k_mask / T_m, z_tmp_2().s);
         }
         for (int ld = 0; ld < ld_block2; ld++) {
@@ -1048,7 +1058,7 @@ void jit_brgemm_kernel_t::apply_compensation(
     if (!brg.req_cal_comp_pads && brg.zp_type_a != brgemm_broadcast_t::none) {
         auto vmm_zp_a_val = z_tmp_2();
         add_imm(X_DEFAULT_ADDR, X_SP, reg_zp_a_val_offs_, X_TMP_0);
-        ldr(reg_zp_a_val, ptr(X_DEFAULT_ADDR));
+        add_imm(reg_zp_a_val, X_SP, reg_zp_a_val_offs_, X_TMP_0);
         ldr(W_TMP_0, ptr(reg_zp_a_val));
         dup(vmm_zp_a_val.s, W_TMP_0);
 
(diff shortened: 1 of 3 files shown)


src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1254,14 +1270,21 @@ void jit_brgemm_kernel_t::set_A_B_matrices() {
...
-    else
+    else if (brg.is_int8 && isa_has_s8s8(brg.isa_impl)) {
+        if (brg.dt_a == data_type::u8 && brg.dt_b == data_type::u8)
+            udot(v1.s, v_b.b, v_a.b);
+        else if (brg.dt_a == data_type::s8 && brg.dt_b == data_type::s8)
+            sdot(v1.s, v_a.b, v_b.b);
+        else if (brg.dt_a == data_type::u8 && brg.dt_b == data_type::s8)
+            usdot(v1.s, v_a.b, v_b.b);
+        else if (brg.dt_a == data_type::s8 && brg.dt_b == data_type::u8)
+            assert(!"unsupported\n");
```
jondea, line 1290: Can we not just swap `v_a` and `v_b` in this case?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1254,14 +1270,21 @@ void jit_brgemm_kernel_t::set_A_B_matrices() {
     add(reg_aux_B, reg_aux_B, reg_b_offset);
 }
 
-void jit_brgemm_kernel_t::dot_product(ZReg v1, ZReg v2, ZReg v3) {
+void jit_brgemm_kernel_t::dot_product(ZReg v1, ZReg v_b, ZReg v_a) {
```
jondea: `v1`,`v2` and `v3` is unclear, although it is consistent. `v1`,`v_b` and `v_a` feels like the right direction, but is more confusing.
- Why is `b` first?
- What is `v1` in this case, if `a` and `b` reference the A and B matrices, then it probably makes sense to give `v1` a name with a similar theme, maybe `v_acc`?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1470,8 +1500,10 @@ void jit_brgemm_kernel_t::gemm_microkernel_sve512(int bd_block2,
                     const auto bd_by_load_bytes
                             = (bd >= bd_e - rows_by_load_bytes
                                     || brg.brgattr.wary_A_k_tail_read);
-                    broadcast(bcst(), A_offset(bd, rd),
-                            have_to_load_bytes && bd_by_load_bytes, brg.dt_a);
+                    int should_broadcast = static_cast<int>(
```
jondea: What does this mean? We have a variable called `should_broadcast` but then we call `broadcast` even if it is not true, are they two different kinds of broadcasting? If so it would be great to make that clear in the variable/function names

jondea (changes requested):
This looks generally great, thank you for this contribution.

Do you have a rough idea of the performance. For example, compared to F32, is it ~4x faster?

Also, what's the general idea of this algorithm? `SDOT` is more complicated than `FMLA` in that it reduces the elements from Int8 to S32. Do we handle this reduction using some kind of blocking? The other way is operating on the transpose and using FADDV, but I can't see that here. In short, it would be great to have a short overview of how this kernel works.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: This order is important, usdot is defined specifically for u8 × s8. If we swap the arguments, the result will be incorrect.
 
The core idea of this kernel is to compute int8 × int8 → int32 matrix multiplications efficiently using SDOT, which performs 4-way dot products and accumulates directly into int32 lanes.
The reduction (from int8 to int32) is handled implicitly by SDOT, so we don’t need explicit reductions like FADDV.
Each SDOT call reduces 4 elements of A and B and adds them to the accumulator. Because of this implicit accumulation, there's no need for extra reduction steps or transposing.
We apply blocking only on matrix A and B and not on destination.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: You are right I updated v1 to v_acc.
We used v_b and v_a because, in all calls to dot_product, the source matrix (A) is passed second and the weight matrix (B) is passed first. This matters because instructions like usdot require a fixed order, we can't swap v_a and v_b without changing the result.
One alternative is to switch the order in the function call so source comes first and weights second, and update the code inside.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: The broadcast function always runs but behaves differently based on the should_broadcast flag.
To avoid confusion from the variable name, I’ve reverted the code to the original inline form.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: > This order is important, usdot is defined specifically for u8 × s8. If we swap the arguments, the result will be incorrect.

Agreed, but the reason there is no `sudot` for vector-vector operands is that if it did exist, it would be such that
```
sudot(acc, a, b) == usdot(acc, b, a)
```
So we can support s8u8 by adding
```
usdot(v_acc.s, v_b.b, v_a.b)
```
to line 1286

On your second point, does this mean that the RHS has to be transposed? What format is the accumulator?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: > One alternative is to switch the order in the function call so source comes first and weights second, and update the code inside.

I'll leave this to you to decide which way makes more sense

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Oh, I thought you were referring only to the u8 × s8 case.
Yes, we can support s8u8 by using usdot with reversed operands, but we’ll also need to handle cases where zero-points are involved to ensure correctness.
However, when testing s8 × u8 with zero-points applied, the tests currently fail even before executing brgemm kernel which needs  to be analysed so this path is not included in this PR.

For second part : 
We are not doing an explicit transpose of the RHS, but we do apply a reorder and convert it into the AB16a16b4a format to align correctly for SDOT lane-wise dot products.
The accumulators are ZReg vector registers with int32 lane (Z.s), organized in row-major order based on M×N blocking.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Thanks! For now, we’ll keep it as is since it matches the operand order and is used in many places, but we can change it later if needed.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Ah okay, so the issue is zero point point handling, not the matmul kernel. Would it be possible to just exclude cases with zero points? Then it would be more explicit where the problem lies. If not, at least a comment somewhere explaining this would be great

Second part: thank you, this is a really useful explanation. Could this go as a comment somewhere in the code for discoverability?

kasturedeeksha: @jondea I’ve updated the code with comments.

jondea (approved):

kasturedeeksha:
Hi @vpirogov, can we get someone from the uxlfoundation side to review and approve this? Thanks!

renato-arantes (approved):

vpirogov:
@renato-arantes, you have an invitation to uxlfoundation organization in your inbox.

PR #3424, opened 2025-06-16
graph: backend: dnnl: backend refactor of adding fusion info attr

xiang1guo:
### Background
Preliminary work for MFDNN-13442. The main purpose of this PR is to bind the `fusion_info` as an attribute directly onto the op, rather than using a `fusion_info_key` to reference a container(currently a vector store in `fusion_info_mgr`) that manages the fusion info.

### Works
- [x] Add new `fusion_info` attr. 
- [x] Update all op's attr with new attr
- [x] Gtest update
- [x] remove `fusion_info_mgr_t`

### DIFF
--- a/src/graph/backend/dnnl/fusion_info.hpp
+++ b/src/graph/backend/dnnl/fusion_info.hpp
@@ -324,33 +324,6 @@ private:
     std::vector<std::shared_ptr<meta_op_t>> post_ops_;
 };
 
-// This class is used to manage all fusion infos in a subgraph. The
-// fusion_info_t can't be directly stored in op's attribute system, so we store
-// them in this manager class and then store the generated int64_t typed key in
-// op's attribute system. When using an ops' fusion info, we can use the fusion
-// info key to query it out from the manager.
-class fusion_info_mgr_t {
-public:
-    fusion_info_mgr_t(
-            graph::fpmath_t fpm_mode = {}, bool can_use_blocked_layout = false)
-        : fpmath_mode_(fpm_mode)
-        , can_use_blocked_layout_(can_use_blocked_layout) {}
-
-    // Disable assignment and copy
-    fusion_info_mgr_t(const fusion_info_mgr_t &) = delete;
-    fusion_info_mgr_t(fusion_info_mgr_t &&) = delete;
-    fusion_info_mgr_t &operator=(const fusion_info_mgr_t &) = delete;
-    fusion_info_mgr_t &operator=(fusion_info_mgr_t &&) = delete;
-
-    const fpmath_t &get_fpmath_mode() const { return fpmath_mode_; }
-    bool get_use_blocked_layout() const { return can_use_blocked_layout_; }
-
-private:
-    // specified floating-point math mode for all fusions
-    fpmath_t fpmath_mode_;
-    bool can_use_blocked_layout_;
-};
-
 // This function is used to make a dnnl::primitive_attr from the fusion info.
 // Note that the op and fusion_info arguments must be matched since a fusion
 // info make sense only when it belongs to a specific op.
--- a/src/graph/backend/dnnl/fusion_info.hpp
+++ b/src/graph/backend/dnnl/fusion_info.hpp
@@ -71,6 +71,13 @@ class fusion_info_t {
                 const std::vector<size_t> &extra_input_indices)
             : op_(op), unfused_input_indices_(extra_input_indices) {};
 
+        bool operator==(const meta_op_t &other) const {
+            return *op_ == *other.op_ && scale_ == other.scale_
+                    && zp_ == other.zp_
+                    && unfused_input_indices_ == other.unfused_input_indices_
+                    && is_post_sum_ == other.is_post_sum_;
+        }
+
         float get_scale() const { return scale_; }
         int32_t get_zp() const { return zp_; }
         const std::vector<size_t> &get_unfused_input_indices() const {
@@ -107,6 +114,56 @@ public:
             const op_ptr &op, const fusion_info_t &fusion_info);
 
     fusion_info_t() = default;
+
+    bool operator==(const fusion_info_t &other) const {
+        // input_zps_
+        if (input_zps_.size() != other.input_zps_.size()) return false;
+        for (const auto &pair : input_zps_) {
+            size_t index = pair.first;
+            const std::shared_ptr<meta_op_t> &meta_op = pair.second;
+            auto it = other.input_zps_.find(index);
+            if (it == other.input_zps_.end() || !(*meta_op == *it->second)) {
+                return false;
+            }
+        }
+
+        // output_zps_
+        if ((output_zps_ == nullptr) != (other.output_zps_ == nullptr))
+            return false;
+        if (output_zps_ && other.output_zps_
+                && !(*output_zps_ == *other.output_zps_))
+            return false;
+
+        // input_scales_
+        if (input_scales_.size() != other.input_scales_.size()) return false;
+        for (const auto &pair : input_scales_) {
+            size_t index = pair.first;
+            const std::shared_ptr<meta_op_t> &meta_op = pair.second;
+            auto it = other.input_scales_.find(index);
+            if (it == other.input_scales_.end() || !(*meta_op == *it->second)) {
+                return false;
+            }
+        }
+
+        // dst_scales_
+        if ((dst_scales_ == nullptr) != (other.dst_scales_ == nullptr))
+            return false;
+        if (dst_scales_ && other.dst_scales_
+                && !(*dst_scales_ == *other.dst_scales_))
+            return false;
+
+        // post_ops_
+        if (post_ops_.size() != other.post_ops_.size()) return false;
+        for (size_t i = 0; i < post_ops_.size(); ++i) {
+            if (!(*post_ops_[i] == *other.post_ops_[i])) { return false; }
+        }
+
+        return true;
+    }
+
+    bool operator!=(const fusion_info_t &other) const {
+        return !(*this == other);
+    }
 
     // used to modify the fused arg scales, like modifying it's axis after
     // inserting reshape op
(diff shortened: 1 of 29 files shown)


src/graph/backend/dnnl/fusion_info.hpp
```diff
@@ -108,6 +108,19 @@ class fusion_info_t {
 
     fusion_info_t() = default;
 
+    // TODO(GX): the operator here is mainly for making compiler happy, and it's
+    // not actually used becasue fusion_info_t is an internal attribute. We may
+    // need to extend the logic here in case of future necessarity.
+    bool operator==(const fusion_info_t &other) const {
```
ElaineBao, line 118: If it's not used, shall we mark it as = delete?

src/graph/backend/dnnl/fusion_info.hpp
```diff
@@ -267,49 +280,25 @@ class fusion_info_t {
...
 };
 
-// This class is used to manage all fusion infos in a subgraph. The
-// fusion_info_t can't be directly stored in op's attribute system, so we store
-// them in this manager class and then store the generated int64_t typed key in
-// op's attribute system. When using an ops' fusion info, we can use the fusion
-// info key to query it out from the manager.
-class fusion_info_mgr_t {
+// This class is used to store subgraph info such as fpmath_mode and flag of
+// using block_layout for subgraph.
```
ElaineBao: How about directly saving these info in the subgraph_t class?

src/graph/backend/dnnl/fusion_info.hpp
xiang1guo: We cannot mark it as delete because this is called in `attribute_value_t` for comparison purpose ([code](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/utils/attribute_value.hpp#L148)). Namely all attribute should define `==` operator. And the operator `==` in `attribute_value_t` is actually called [here ](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/interface/op.hpp#L110)in op.hpp for comparison. Since fusion_info_t is an internal attribute, it won't be used for comparison, thus I just implemented a rough one to make compiler happy(no compiation issue)

src/graph/backend/dnnl/fusion_info.hpp
xiang1guo: I was thinking about this, and I think it's doable, but I didn't do that because:
1. Those subgraph info is used in op_executable when creating desc, see [here](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/op_executable.cpp#L54). I feel passing a cleaner sungraph_info instead of full subgraph_t is clearer and simpler. 
2. This involves too much change, and the first thing I want to do is just split fusion_info from this mgr to serve for compressed SDPA support. 
cc @TaoLv  any suggestions?

TaoLv: I see there are only two extra flags: fpmath and use block layout. Is it possible to move them along with fusion info?  I know these are applied graph-wise. But eventually it's used by each op and each primitive. It also seems to simplify the logic around.

src/graph/backend/dnnl/fusion_info.hpp
TaoLv: what do we need to extend?

xiang1guo: Oh, the `extend` I mentioned refers to the future scenario where we need to compare internal attributes. In that case, it would be necessary to strictly compare all the member meta ops within the fusion info.

TaoLv: What's the potential effort if we want to complete the comparison function here?

xiang1guo: Not much, it just looks a bit redundant. I had already written it down, but I thought it was too messy, so I didn't put it in current PR. :)  I will add that later.

src/graph/backend/dnnl/fusion_info.hpp
xiang1guo: > I see there are only two extra flags: fpmath and use block layout. Is it possible to move them along with fusion info? I know these are applied graph-wise. But eventually it's used by each op and each primitive. It also seems to simplify the logic around.

ok, I will take a look at this suggestion.

src/graph/backend/dnnl/passes/insert_ops.cpp
```diff
@@ -829,16 +815,22 @@ impl::status_t insert_runtime_u8_to_s8_for_matmul(
...
+                fusion_info_t fusion_info;
+                fusion_info.set_zero_points(
+                        zps_op->shared_from_this(), true, 1);
+                cur_op->set_attr<fusion_info_t>(
+                        op_attr::fusion_info, fusion_info);
             }
 
-            fusion_info_t &fusion_info = mgr.get_mutable_info(key);
-            fusion_info.set_zero_points(zps_op->shared_from_this(), true, 1);
+            // fusion_info_t &fusion_info = mgr.get_mutable_info(key);
```
rongzha1: remove unused code

src/graph/backend/dnnl/fusion_info.hpp
xiang1guo: I have added two new commits: the first one is related to the operator `==`, and the second one involves cleaning up redundant code discovered while reviewing `subgraph_info_mgr`.

src/graph/backend/dnnl/passes/insert_ops.cpp
xiang1guo: Removed, thanks~

src/graph/backend/dnnl/fusion_info.hpp
xiang1guo: I choose to save `fpmath_t` and `block layout` info in subgraph currently and pass that info to `op_executable` and `layout_propagayor` independently, please refer to the last commit and review again. Thanks! 
BTW, bind these info to op `fusion_info` level can be a long-term plan and depends on real scenario usage as the change involves too much change now(transformation pass handling is not straightforward after I look into the implementation) :)

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/internal_attrs.hpp
```diff
@@ -69,6 +69,9 @@ const op_attr_t dst_zps = 0x10400;
 const op_attr_t src_zps = 0x10401;
 const op_attr_t permutation = 0x10402;
 
+// custom class
+const op_attr_t fusion_info = 0x10500;
```
TaoLv, line 72: Isn't it backend specific attribute?

xiang1guo: yes，will update the comment here.

TaoLv: Also fix the commit message a bit with proper scopes. Thanks.

src/graph/interface/op.hpp
```diff
@@ -248,6 +248,13 @@ struct dnnl_graph_op : public std::enable_shared_from_this<dnnl_graph_op> {
             return it->second.get<value_type>();
     }
 
+    template <typename value_type>
+    value_type &get_mutable_attr(op_attr_t name) {
```
TaoLv: I'm actually thinking about explicitly using the existing interfaces by: get_attr() -> modify it -> set_attr().

src/graph/interface/c_types_map.hpp
```diff
@@ -307,6 +307,7 @@ const attribute_kind_t i = 2;
 const attribute_kind_t is = 3;
 const attribute_kind_t s = 4;
 const attribute_kind_t b = 5;
+const attribute_kind_t c = 6; // custom attribute type
```
TaoLv: what does it mean? Is it "any" type?

src/graph/interface/op.hpp
xiang1guo: Ok, It was aiming to align with fusion info `get_mutable_scales` [func](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/fusion_info.hpp#L113), one of my concern is that this mainly used to update `fusion_info` attr, if use get_attr and set_attr, there will be some overhead of copy for `fusion_info` class.

src/graph/interface/c_types_map.hpp
xiang1guo: Yes, it means other custom types except the ones above. "c" stands for custom class.

src/graph/backend/dnnl/internal_attrs.hpp
xiang1guo: Updated comment, for the commit message, the first commit includes both the interface and the dnnl file, so I didn't add `dnnl` scope, any suggestion?

src/graph/interface/op.hpp
xiang1guo: Added into last 2 commits. If no other concern, I will merge into the previous commit and trigger ci, thanks!

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/internal_attrs.hpp
TaoLv: in that case, you may also want to drop the `backend` scope in the commit message i think. Previously we used something like `graph: {interface, backend}: xxx` but i guess this format will not pass the lint check now.

src/graph/interface/c_types_map.hpp
TaoLv: If it already supports "any" type, i'm thinking if we can drop the `f/fs/i/is/s/b`.
We will need to decide if we want to benefit from the flexibility of `any` type, or stick on the limited kinds for strong typing.

xiang1guo: I was thinking about that f/fs/i/is/s/b are still useful because it can help us to understand the real type of attribute in op_schema, I use `c` to represent customized class which developer implement. It helps to better understand the op schema.

src/graph/backend/dnnl/internal_attrs.hpp
xiang1guo: oh, ok, i missed `backend` is redudant.

src/graph/interface/c_types_map.hpp
TaoLv: With that, if we don't plan (or are not ready) to accept "any" types as op attributes, it would be better to keep attribute type limited and explicit. It will make the code easier to test, debug, and read.

xiang1guo: Not sure if I fully understand your point, I added last commit that check if the attr kind is `fusion_info_t`. Please help to review again, thanks!

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3428, opened 2025-06-16
src: remove unnecessary compute device info header dependencies

rjoursler:
The nGEN interface used for code generation relies on ngen::Product. To enable better interoperation within oneDNN, this PR modifies device_info_t to store the ngen::Product. In order to accomplish this, some implicit dependencies on `gpu/intel/compute/device_info.hpp` are broken to avoid nGEN dependencies outside of the `src/gpu/intel` directories.


third_party/ngen/ngen_core.hpp
```diff
@@ -17,6 +17,11 @@
 #ifndef NGEN_CORE_HPP
 #define NGEN_CORE_HPP
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wimplicit-int-conversion"
```
dzarukin, line 22: TODO to address them later?

third_party/ngen/ngen_opencl.hpp
```diff
@@ -63,10 +63,14 @@ class OpenCLCodeGenerator : public ELFCodeGenerator<hw>
     inline cl_kernel getKernel(cl_context context, cl_device_id device, const std::string &options = "-cl-std=CL2.0");
     bool binaryIsZebin() { return isZebin; }
 
+    static inline HW detectHW(cl_device_id device);
     static inline HW detectHW(cl_context context, cl_device_id device);
+    static inline Product detectHWInfo(cl_device_id device);
     static inline Product detectHWInfo(cl_context context, cl_device_id device);
 
 private:
+    static inline Product detectHWInfoImpl(cl_device_id device, cl_context context = nullptr);
```
dzarukin: Why not just make `detectHWInfo` with this signature?

third_party/ngen/ngen_core.hpp
rjoursler: These are likely not going to be addressed. The nGEN library does a lot of binary format style manipulation, so it was decided that these warnings are not worth the effort to maintain. In particular, many operations applied to `uint8_t` are implicitly promoted to `int`, which then requires static casts all over the place to convert back into `uint8_t`.

third_party/ngen/ngen_opencl.hpp
rjoursler: This is part of the nGEN API, and so there are some stability guarantees. In particular, as there is no user need for this interface (due to the existing detectHWInfo interface), I am keeping this new function out of the nGEN API. I used the name `detectHWInfoImpl` to make the distinction a bit more clear.

rjoursler:
>  Having an explanation why direct use of nGEN header is not possible given that a direct dependency is introduced would be helpful as well.

I added a description. The core issue is that the nGEN headers are only enabled within the src/gpu directory, so we get compilation errors if we add an nGEN dependency to device_info.hpp. We could alternatively just include nGEN headers everywhere in oneDNN (when applicable), but this PR keeps the current status quo as nGEN really should not be necessary as a dependency in `src/xpu` or `src/graph`,

third_party/ngen/ngen_core.hpp
dzarukin: Minor: might be good to put a note next to this change for the future reference.

third_party/ngen/ngen_opencl.hpp
dzarukin: Thanks for explaining. Resolving.

dzarukin:
@spalicki, Could you, please, take a look one this one?

third_party/ngen/ngen_core.hpp
rjoursler: I will see about adding comments to this end in the upstream nGEN repo, which will get downstreamed at a later date.

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

echeresh (approved):

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

rjoursler:
Closing this as partial merge was completed in #3579, so this needs significant rework to merge.

PR #3432, opened 2025-06-17
cpu: some improvements 

dzarukin:
This is some groundwork for future asynchronous runtime support:
* There are overflow issues caught by `-ftrapv` option when running test_conv_all for conv and ip.
* There's a rename for the variable which is shadowed by outer same variable. It get a problem when capture by copy is used.
* Make post-ops kernel apply scales when they are defined in attributes. It's possible to distinguish them since v3.0.
* Align scale_utils function interfaces to rely on same objects between calls for proper re-use of the code underneath.

(there's more to come).

### DIFF
--- a/src/cpu/scale_utils.cpp
+++ b/src/cpu/scale_utils.cpp
@@ -33,33 +33,23 @@ constexpr size_t scales_simd_w = 16;
 
 void book_precomputed_scales(memory_tracking::registrar_t &scratchpad,
         const scales_t &attr_scales, size_t wei_scale_count,
-        bool force_scales_book) {
-    using namespace dnnl::impl::memory_tracking::names;
-
-    const bool with_src_scales = !attr_scales.has_default_values(DNNL_ARG_SRC);
-    const bool with_wei_scales
-            = !attr_scales.has_default_values(DNNL_ARG_WEIGHTS);
-
-    if ((with_src_scales && with_wei_scales) || force_scales_book
-            || !attr_scales.has_default_data_type(DNNL_ARG_WEIGHTS)
-            || !attr_scales.get(DNNL_ARG_WEIGHTS).has_default_groups()) {
+        float scale_adjust_factor) {
+    if (req_copy_scales(attr_scales, scale_adjust_factor)) {
         const int wei_mask = attr_scales.get_mask(DNNL_ARG_WEIGHTS);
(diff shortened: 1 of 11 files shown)


dzarukin:
make test
disable test_device_gpu

src/cpu/x64/jit_avx512_core_bf16_dw_conv_kernel.cpp
```diff
@@ -584,7 +584,9 @@ void jit_avx512_dw_conv_fwd_kernel_bf16_t::generate() {
         mov(reg_tail_32, zmm_16b_mask >> oc_tail_shift);
         kmovw(k_oc_tail_mask, reg_tail_32);
         if (need_extended_mask) {
-            auto zmm_32b_mask = (1 << (oc_tail + jcp.ch_block)) - 1;
+            // Computing the mask the other way around overflows int32_t for
+            // `oc_tail + jcp.ch_block == 31`.
+            auto zmm_32b_mask = INT_MAX >> (31 - (oc_tail + jcp.ch_block));
```
mgouicem: (nit)
```suggestion
            auto zmm_32b_mask = (-1) >> (32 - (oc_tail + jcp.ch_block));
```

mgouicem (approved):

src/cpu/x64/jit_avx512_core_bf16_dw_conv_kernel.cpp
densamoilov: Can `(oc_tail + jcp.ch_block)` be 0?

dzarukin: It should be in a [16; 31] range. The problem appeared for `31` specifically when it overflows `INT_MAX`.

densamoilov: Ok, I just asked because shifting 32 bit would result in undefined behavior.

dzarukin: Applied `-1U`, thanks.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
disable test_device_gpu
enable benchdnn_nightly

PR #3437, opened 2025-06-17
ci: aarch64: remove patterns that are not in nightly from skip-list

Sqvid:
# Description

These test are no longer built for the AArch64 CPU target so are redundant in the skip list.
Relevant [nightly run here](https://github.com/uxlfoundation/oneDNN/actions/runs/15710173241/job/44266114472).

# Checklist

## General


Sqvid:
@Ryo-not-rio I cancelled the workflow since some of the runners are hanging. Could we restart it when the issue is resolved?

Sqvid:
Closing. Looks like this was based on a faulty assumption that TEST_SET_NIGHTLY was a proper superset of TEST_SET_CI.

PR #3440, opened 2025-06-18
cpu: x64: pool: enable u8 type in fwd pooling

jstachowintel:
[MFDNN-8916](https://jira.devtools.intel.com/browse/MFDNN-8916) int8 maxpool with binary mul post-ops run into ref path


src/cpu/x64/utils/jit_io_helper.cpp
```diff
@@ -921,6 +922,7 @@ void jit_io_helper_t<Vmm>::store_i8(
             store_i8_fn(src_xmm, src_vmm);
             host_->uni_vmovntps(dst_addr, src_xmm);
         } else {
+            host_->vcvtps2dq(src_vmm, src_vmm);
```
dzarukin: This is covered by `saturate`. Shouldn't be here.

src/cpu/x64/utils/jit_io_helper.cpp
```diff
@@ -898,6 +898,7 @@ void jit_io_helper_t<Vmm>::store_f8(
 template <typename Vmm>
 void jit_io_helper_t<Vmm>::store_i8(
         const Vmm &src_vmm, const Xbyak::Address &dst_addr) {
+
```
dzarukin: Would be nice to to include empty lines unless it's a common style for the file (which I don't see it is).

src/cpu/x64/utils/jit_io_helper.cpp
```diff
@@ -802,9 +802,9 @@ template <typename Vmm>
 void jit_io_helper_t<Vmm>::saturate(const Vmm &vmm) {
     assert(saturation_conf_.has_value() && "Config for saturation is not set.");
 
-    host_->saturate_cvt_f32(vmm,
-            Vmm(saturation_conf_->vreg_zero_saturation_idx_),
-            Vmm(saturation_conf_->vreg_saturation_ubound_idx_), data_type_);
+    host_->sub(host_->rsp, 64);
+    host_->uni_vmovups(host_->ptr[host_->rsp], vmm);
+    host_->add(host_->rsp, 64);
```
dzarukin: This is not how the stack works. The process must use the stack when it's accessible. By adding back it becomes inaccessible as it was before sub.

src/cpu/x64/jit_uni_pool_kernel.cpp
```diff
@@ -132,8 +132,14 @@ jit_uni_pool_kernel_t<isa>::jit_uni_pool_kernel_t(
     if (jpp.ind_dt == data_type::s32) dtypes.insert(data_type::f32);
     if (jpp.needs_f32_accum_for_bf16) dtypes.insert(data_type::f32);
 
+    typename io_mdt_helper::saturation_map_t saturation_confs;
+    if (dtypes.count(data_type::u8)) {
+        saturation_confs[data_type::u8]
+                = io::io_saturation_conf_t(0, 1, tmp_gpr);
```
dzarukin: `0` and `1` must be a named constants and play together with other vector registers in this implementation's picture. This way it's very sneaky and hard to find any potential register collision.

src/cpu/x64/jit_uni_pooling.cpp
```diff
@@ -1477,6 +1477,9 @@ template struct jit_uni_pooling_fwd_t<avx512_core, data_type::bf16>;
 template struct jit_uni_pooling_bwd_t<avx512_core, data_type::bf16>;
 template struct jit_uni_pooling_fwd_t<avx512_core_fp16, data_type::f16>;
 template struct jit_uni_pooling_bwd_t<avx512_core_fp16, data_type::f16>;
+template struct jit_uni_pooling_fwd_t<sse41, data_type::u8>;
+template struct jit_uni_pooling_fwd_t<avx2, data_type::u8>;
+template struct jit_uni_pooling_fwd_t<avx512_core, data_type::u8>;
```
dzarukin, line 1473: An extra task might be to get rid of data type template argument from the implementation which would save on same disk space.

src/cpu/x64/jit_uni_pooling.cpp
```diff
@@ -1477,6 +1477,9 @@ template struct jit_uni_pooling_fwd_t<avx512_core, data_type::bf16>;
 template struct jit_uni_pooling_bwd_t<avx512_core, data_type::bf16>;
 template struct jit_uni_pooling_fwd_t<avx512_core_fp16, data_type::f16>;
 template struct jit_uni_pooling_bwd_t<avx512_core_fp16, data_type::f16>;
+template struct jit_uni_pooling_fwd_t<sse41, data_type::u8>;
```
dzarukin: Please remove, this is not needed.

jstachowintel:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_pool
set test_scope=NIGHTLY

src/cpu/x64/utils/jit_io_helper.cpp
dzarukin: Looks like the changeset in this file can be dropped as there are just new lines.

src/cpu/x64/jit_uni_pool_kernel.hpp
```diff
@@ -91,6 +91,9 @@ struct jit_uni_pool_kernel_t : public jit_generator_t {
 
     Vmm vmm_k_offset = Vmm(1);
 
+    Vmm vmm_zero = Vmm(5);
+    Vmm vmm_saturation_ubound = Vmm(6);
```
dzarukin: Would be nice to check if bf16:x8 and x8:bf16 combination works on avx512_core to make sure the register collision is legit. If it is, would be great to document why it's legit.

jstachowintel: It has been created a separate [task](https://jira.devtools.intel.com/browse/MFDNN-14412) for mix types

src/cpu/x64/jit_uni_pool_kernel.cpp
```diff
@@ -785,10 +801,22 @@ void jit_uni_pool_kernel_t<isa>::apply_postops(int ur_bc, int ur_w, int c_block,
...
                     rhs_arg_params.vmm_tail_idx_.emplace(vmm_idx);
                 }
+
+                vmm_idxs.emplace(vmm_idx);
             }
         }
     }
-    postops_injector_->compute_vector_range(start_idx, end_idx, rhs_arg_params);
+
+    if (vmm_idxs.empty()) {
```
dzarukin, line 825: Could you, please, explain when this outcome is possible and why it's legit to emplace indices post-factum the code above said there's no available indices?

src/cpu/x64/jit_uni_pool_kernel.cpp
```diff
@@ -132,8 +133,22 @@ jit_uni_pool_kernel_t<isa>::jit_uni_pool_kernel_t(
...
     if (jpp.needs_f32_accum_for_bf16) dtypes.insert(data_type::f32);
 
-    io_ = io_mdt_helper(this, jpp.isa, dtypes, {}, io_tail_conf, io_bf16_conf,
-            {}, utils::nullopt, io_fp8_conf);
+    typename io_mdt_helper::saturation_map_t saturation_confs;
+    if (jpp.dst_dt == data_type::u8) {
+        uni_vxorps(vmm_zero, vmm_zero, vmm_zero);
+        mov(tmp_gpr, float2int(255.0f));
+        uni_vmovq(xmm_tmp, tmp_gpr);
+        uni_vbroadcastss(vmm_saturation_ubound, xmm_tmp);
```
dzarukin: It's is prohibited to inset instructions into the kernel before `preamble()` happens to the kernel.

src/cpu/x64/jit_uni_pool_kernel.hpp
```diff
@@ -61,11 +61,14 @@ struct jit_uni_pool_kernel_t : public jit_generator_t {
...
     using Vmm = typename cpu_isa_traits_t<isa>::Vmm;
 
+    int reserved_vmms = 0;
     int vmm_idx_upper_bound() const noexcept {
         return is_superset(isa, avx512_core) ? 31 : 15;
     }
 
-    int reg_idx(int idx) const noexcept { return vmm_idx_upper_bound() - idx; }
+    int reg_idx(int idx) const noexcept {
+        return vmm_idx_upper_bound() - idx - reserved_vmms;
```
dzarukin: Could you explain the choice of saturation indices to start from the end while the rest of the kernel uses indices starting from 0 for service registers?

I'd use 4 and 5 for that purpose and assert on `reg_idx` being greater than 5 for avx2 and 10 otherwise.

src/cpu/x64/jit_uni_pool_kernel.cpp
```diff
@@ -132,8 +133,22 @@ jit_uni_pool_kernel_t<isa>::jit_uni_pool_kernel_t(
...
+
+        io::io_saturation_conf_t io_saturation_conf(
+                vmm_zero.getIdx(), vmm_saturation_ubound.getIdx(), tmp_gpr);
+        saturation_confs.insert({data_type::u8, io_saturation_conf});
```
dzarukin: Why not just use dst_dt instead of hardcoded value?

src/cpu/x64/jit_uni_pool_kernel.cpp
```diff
@@ -551,6 +566,8 @@ inline void jit_uni_pool_kernel_t<isa>::store(const data_type_t dt,
         const bool is_c_tail_proccessing) {
     if (is_c_tail_proccessing && jpp.is_c_padded && jpp.with_postops)
         pad_with_zeros(idx);
+    if (utils::one_of(dt, data_type::u8, data_type::s8, data_type::s32))
+        io_.init_saturate_f32({dt});
```
dzarukin: No need in guarding this call. It does necessary checks internally.

src/cpu/x64/jit_uni_pool_kernel.hpp
jstachowintel: It was my first idea, but I agree that used vmms for saturation should be similar in each kernel

src/cpu/x64/jit_uni_pool_kernel.cpp
jstachowintel: `vmm_idxs` is only populated inside the `if (jpp.with_binary && !sse41_postops_disabled)` block, i.e. when we actually have binary post-ops and we need per-VMM RHS addressing. In the pure eltwise-only case (`with_postops == true` but `with_binary == false`), or when binary is disabled for the SSE41 high-half path, this block is skipped and `vmm_idxs` remains empty.

The fallback `if (vmm_idxs.empty()) { ... }` does not mean "there are no available indices"; it just means "no binary-specific configuration was added". In that situation we still want to apply post-ops to all accumulator VMMs, and eltwise does not require any RHS dynamic params, so we safely rebuild `vmm_idxs` with all `reg_ind(0, ...)` vectors and call `compute_vector_range` with an empty `rhs_arg_params`.

src/cpu/x64/jit_uni_pooling.cpp
dzarukin: I'm not sure why this got resolved while the change is still there...

dzarukin (commented):
Where did all the kernel changes go? Is it a bad rebase or no changes are required?

jstachowintel:
@dzarukin I had an issue with a rebase, but I have already resolved it.

jstachowintel:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_pool
set test_scope=NIGHTLY

jstachowintel:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_pool
set test_scope=NIGHTLY

jstachowintel:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all
enable benchdnn_pool
set test_scope=NIGHTLY

jstachowintel:
make test
set test_scope=NIGHTLY

jstachowintel:
make test
set test_scope=NIGHTLY

PR #3444, opened 2025-06-18
cpu: x64: enable fp8 support in reorder on NVL

xuxinzen:
Fixes MFDNN-13835

fp8 pooling on NVL will hit an assertion about "invalid isa or fp8 emulation not available" in jit_uni_reorder and then cause failure. Enabling fp8 support in reorder on NVL can fix this issue.

Example of failing case:
```c++
DNNL_MAX_CPU_ISA=AVX10_2_512 sde -nvl -- ./tests/benchdnn/benchdnn -v5 --pool --dt=f8_e4m3:f8_e4m3 --alg=avg_p ic32iw300ow151kw3sw2pw1
create: --pool --dt=f8_e4m3:f8_e4m3 --alg=avg_p ic32iw300ow151kw3sw2pw1
benchdnn: src/cpu/x64/jit_uni_reorder.cpp:626: dnnl::impl::cpu::x64::tr::jit_uni_reorder_kernel_f32_t::process_unroll_generic_step(int, const int*, const int*, const int*, const int*, const int*, bool)::<lambda(Xbyak::Xmm, const Xbyak::Operand&, dnnl::impl::data_type_t)>: Assertion `!"invalid isa or fp8 emulation not " "available"' failed.
Aborted (core dumped)
```

### DIFF
--- a/src/cpu/x64/jit_uni_reorder.cpp
+++ b/src/cpu/x64/jit_uni_reorder.cpp
@@ -203,9 +203,14 @@ struct jit_uni_reorder_kernel_f32_t : public kernel_t, public jit_generator_t {
                         mayiuse(avx512_core_fp16) || mayiuse(avx2_vnni_2))
                 && IMPLICATION(utils::one_of(f8_e5m2, p.itype, p.otype)
                                 || utils::one_of(f8_e4m3, p.itype, p.otype),
-                        mayiuse(avx512_core_amx))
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test
set test_scope=NIGHTLY
disable test_device_gpu

densamoilov (approved):

ankalinin (approved):

PR #3454, opened 2025-06-20
example: graph: sdpa: use host-side scale value

TaoLv:
- Modified the SDPA example to use host-side scale value on GPU.
- Added check in the backend to fallback the case to primitive based implementation. It can be removed once host-side scale is supported by ukernel SDPA.

### DIFF
--- a/examples/graph/sdpa.cpp
+++ b/examples/graph/sdpa.cpp
@@ -32,6 +32,7 @@ using namespace dnnl;
 
 using namespace dnnl::graph;
 using layout_type = logical_tensor::layout_type;
+using property_type = logical_tensor::property_type;
 using dim = logical_tensor::dim;
 using dims = logical_tensor::dims;
 
@@ -300,7 +301,9 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     bmm1.add_outputs({score});
 
     // scaled_score = score / scale
-    auto scale = logical_tensor(id++, dt, scale_sz, layout_type::strided);
+    const logical_tensor::data_type dt_scale = logical_tensor::data_type::f32;
+    auto scale = logical_tensor(id++, dt_scale, scale_sz, layout_type::strided,
+            property_type::host_scalar);
     auto scaled_score
             = logical_tensor(id++, dt_inter, score_sz, layout_type::strided);
     auto scale_div = op(id++, op::kind::Divide, "scale_div");
@@ -351,18 +354,9 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     compiled_partition cp = partitions[0].compile(
             {query, key, scale, mask, value}, {output}, eng);
 
-    // Create tensor objects
-    auto ts_query = tensor(query, eng);
-    auto ts_key = tensor(key, eng);
-    auto ts_scale = tensor(scale, eng);
-    auto ts_mask = tensor(mask, eng);
-    auto ts_value = tensor(value, eng);
-    auto ts_output = tensor(output, eng);
-
     // Allocate user data.
     std::vector<float> query_data(product(qv_sz));
     std::vector<float> key_data(product(k_sz));
-    std::vector<float> scale_data(product(scale_sz), std::sqrt(p.head_size));
     std::vector<float> mask_data(product(mask_sz));
     std::vector<float> value_data(product(k_sz));
     std::vector<float> output_data(product(qv_sz));
@@ -372,12 +366,21 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     fill_random(value_data);
     fill_mask(mask_data, static_cast<size_t>(p.seq_len));
 
+    // Create tensor objects
+    auto ts_query = tensor(query, eng);
+    auto ts_key = tensor(key, eng);
+    auto ts_mask = tensor(mask, eng);
+    auto ts_value = tensor(value, eng);
+    auto ts_output = tensor(output, eng);
+
     // Write data to tensor object's handle.
     write_to_dnnl_tensor(query_data.data(), ts_query);
     write_to_dnnl_tensor(key_data.data(), ts_key);
-    write_to_dnnl_tensor(scale_data.data(), ts_scale);
     write_to_dnnl_tensor(mask_data.data(), ts_mask);
     write_to_dnnl_tensor(value_data.data(), ts_value);
+
+    float scale_data = std::sqrt(p.head_size);
+    auto ts_scale = tensor::make_scalar_tensor(scale, &scale_data);
 
     // Warmup run.
     // Execute the compiled partition of sdpa.
--- a/include/oneapi/dnnl/dnnl_graph.hpp
+++ b/include/oneapi/dnnl/dnnl_graph.hpp
@@ -575,6 +575,8 @@ private:
 /// A tensor object
 class tensor : public tensor_handle {
 public:
+    using tensor_handle::handle;
+
     /// Default constructor. Constructs an empty object.
     tensor() = default;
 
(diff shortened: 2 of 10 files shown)


examples/graph/sdpa.cpp
```diff
@@ -277,6 +278,15 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
     // Create dnnl::stream.
     dnnl::stream strm(eng);
 
+// Create host engine for the host-side scale data. If oneDNN is built
+// without CPU runtime, we will not be able to create host engine and have
+// to use device-side scale on GPU.
+#if DNNL_CPU_RUNTIME != DNNL_RUNTIME_NONE
```
TaoLv: I'm not very sure about this part as it makes the example slightly not straightforward. Usually we don't do runtime check in example code. Call for comments.

src/graph/interface/tensor.hpp
```diff
@@ -93,6 +96,7 @@ struct dnnl_graph_tensor {
         // this field is valid when logical tensor's
         // property is host_scalar
         int32_t s32_value = 0;
+        float f32_value;
```
ElaineBao, line 99: init it with 0?

TaoLv: Union can have at most one member with default initialization. `float f32_value = 0.f;` will cause compile error.

examples/graph/sdpa.cpp
mgouicem: Agreed. On the primitive host-side scalar RFC, there is a discussion on exposing a method to create memory from host-side scalar without an engine and with the scalar directly. That would prevent the issue at hand here.

examples/graph/sdpa.cpp
```diff
@@ -372,13 +379,30 @@ void bench_sdpa(engine::kind ekind, logical_tensor::data_type dt,
...
     write_to_dnnl_tensor(key_data.data(), ts_key);
-    write_to_dnnl_tensor(scale_data.data(), ts_scale);
     write_to_dnnl_tensor(mask_data.data(), ts_mask);
     write_to_dnnl_tensor(value_data.data(), ts_value);
 
+#if DNNL_CPU_RUNTIME != DNNL_RUNTIME_NONE
+    float scale_data = std::sqrt(p.head_size);
+    auto ts_scale = ekind == engine::kind::cpu
+            ? tensor(scale, eng, &scale_data)
+            : tensor(scale, host_eng, &scale_data);
```
mgouicem: same as above. Maybe we could add a make_tensor function or tensor constructor that does not take an engine and takes a scalar directly?

TaoLv: Thank you for the suggestion, Mourad. Do I understand it correctly that the new function only works for host scalar tensor creation? Does the below definition look good to you?

```cpp
/// Creates a tensor object for host-side scalar value. The data type contained
/// in the logical tensor parameter will be used to interpret the scalar pointer.
/// The property type in the logical tensor must be `host_scalar`.
///
/// @param lt The logical tensor describing the host scalar
/// @param scalar The pointer to scalar value
/// @returns Created tensor object
tensor make_tensor(const logical_tensor &lt, void *scalar) {
        ...
        // C API will be reused with engine = nullptr. 
        dnnl_graph_tensor_create(&ts, lt, /* dnnl_engine_t = */ nullptr, scalar);
        ...
}
```

TaoLv: It seems the code still will diverge for CPU and GPU. Do we encourage CPU users to create tensor with engine or without engine?

TaoLv:
Hi @mgouicem @LuFinch I added a new `make_scalar_tensor(...)` API as discussed above and called it in the SDPA example. Please take a look again at the latest changes. Note that the example still needs CPU runtime to execute as in the backend we need to create dnnl::memory and reorder to copy the scalar value from host to device for reference execution. Thanks!

LuFinch:
@TaoLv I have tried with your new api `make_scalar_tensor(...)` and it works well. It LGTM and is more easy to use than before.

examples/graph/sdpa.cpp
mgouicem: @TaoLv could you clarify where the diversion comes from? Here this make_tensor is engine-less, so there should be no engine specific divergence at API level. 
However, it might diverge depending on library support and we would have to extend backend to take this new scalar tensor kind for CPU device.

include/oneapi/dnnl/dnnl_graph.hpp
```diff
@@ -647,6 +655,21 @@ class tensor : public tensor_handle {
...
+/// Creates a tensor object for host-side scalar value. The data type contained
+/// in the logical tensor parameter will be used to interpret the scalar
+/// pointer. The property type in the logical tensor must be `host_scalar`.
+///
+/// @param lt The logical tensor describing the host scalar
+/// @param scalar The pointer to scalar value
+/// @returns Created tensor object
+inline tensor make_scalar_tensor(const logical_tensor &lt, void *scalar) {
+    dnnl_graph_tensor_t t = nullptr;
+    error::wrap_c_api(dnnl_graph_tensor_create(&t, &(lt.data), nullptr, scalar),
```
mgouicem: (up to discussion) Would it be more error-proof to have a new C symbol as well?
For example, accepting nullptr for engine in dnnl_graph_tensor_create might make debugging harder on user side (they would get an error about host-side scalar or segfault if they pass nullptr as engine by mistake).

mgouicem:
@mzhukova @densamoilov for awareness

examples/graph/sdpa.cpp
TaoLv: Thanks, Mourad. I updated the example. Now the new interface `make_scalar_tensor` is also used for CPU as well. So we don't have diversion from the example code perspective.

TaoLv:
Updated the PR with below changes:
- Move C++ API `make_scalar_tensor` as a static method of tensor. Now `auto ts = tensor::make_scalar_tensor(lt, &scalar)`. It's still a `make` function, not a constructor so far.
- Add a new C API `dnnl_graph_tensor_create_host_scalar(...)` without engine parameter, as I see a new memory create API is added in #3506. 

@mgouicem @densamoilov @mzhukova , a few more clarifications:

- For host scalar memory/tensor, what do `get/set_data_handle()` and `get_engine()` mean for users?

include/oneapi/dnnl/dnnl_graph.hpp
TaoLv: Added a new C API `dnnl_graph_tensor_create_host_scalar()` and kept the original `dnnl_graph_tensor_create()` behavior unchanged in regard of the engine parameter.

mgouicem:
Thanks Tao, that looks great to me. Regarding your questions, I would recommend:
- get/set_data_handle(): return a nullptr for get, and return an invalid status for set. Though we might need a function to update the scalar value.
- get_engine(): return empty engine `engine()` for C++ API, nullptr for C API.

TaoLv:
@LuFinch  Do you see any requirement for updating the scalar value in a scalar tensor after it's created?

LuFinch:
@TaoLv  No. Never see such a scenario.

src/graph/backend/dnnl/common.cpp
```diff
@@ -164,6 +164,11 @@ dnnl::engine make_dnnl_engine(const engine_t &g_engine) {
     return engine;
 }
 
+// The function will throw expect if cpu runtime is none.
```
ElaineBao, line 167:
```suggestion
// The function will throw exception if cpu runtime is none.
```

src/graph/backend/dnnl/kernels/large_partition.cpp
```diff
@@ -170,9 +170,17 @@ void larger_partition_kernel_t::setup_pipeline(pass_pipeline_t &pipeline,
 void larger_partition_kernel_t::prepare_host_scalar_args(
         execution_args_set_t *res, const std::vector<tensor_t> &inputs) {
     for (const auto &host_scalar_info : res->get_host_scalar_infos()) {
-        auto mem = make_dnnl_memory(host_scalar_info.md,
-                make_dnnl_engine(
-                        *(inputs[host_scalar_info.input_idx].get_engine())),
+        const engine_t *eng_ptr
+                = inputs[host_scalar_info.input_idx].get_engine();
+        // For host scalar tensor, if it contains an egine, use it. Otherwise,
```
ElaineBao, line 175:
```suggestion
        // For host scalar tensor, if it contains an engine, use it. Otherwise,
```

src/graph/backend/dnnl/kernels/large_partition.cpp
```diff
@@ -170,9 +170,17 @@ void larger_partition_kernel_t::setup_pipeline(pass_pipeline_t &pipeline,
...
+        // create a host engine for it. This can be changed once dnnl::memory
+        // supports host scalar memory creation without engine and also supports
+        // reorder from a host scalar memory to a gpu device memory.
+#if DNNL_CPU_RUNTIME == DNNL_RUNTIME_NONE
+        assert(eng_ptr);
+#endif
```
ElaineBao, line 181: >  Note that the example still needs CPU runtime to execute as in the backend we need to create dnnl::memory and reorder to copy the scalar value from host to device for reference execution.

It's a bit inconsistent to me that the ukernel support doesn't require the CPU runtime, while the reference implementation does need it just for the reorder. If the primitive could provide an interface that allows the reorder to be created with src_engine = nullptr when the CPU runtime is none, that would be ideal!

TaoLv:
Closing as it's done in:
- #3528 
- #3958

PR #3455, opened 2025-06-20
build: risc-v: dynamic selection of -march for RV64

krishnasai-mcw:
# Configure RV64 march flags dynamically based on RVV intrinsics support

# Description

This PR adds support for dynamically detecting whether the current RISC-V toolchain supports RVV intrinsics (version 1.0) during CMake configuration. Based on the result, it sets the appropriate `-march` flag automatically:
- `-march=rv64gcv` if RVV intrinsics are supported
- `-march=rv64gc` otherwise
 
### Fix:
Leverages the existing CAN_COMPILE_RVV_INTRINSICS CMake variable — originally used to check RVV intrinsic support — to dynamically set the -march flag at configure time.
 
### Validation:
Validated using the riscv-gnu-toolchain version 14.2. The toolchain was built with support for the RISC-V Vector Extension (RVV) version 1.0, and the dynamic -march flag selection logic worked correctly during the CMake configuration and build.
 
This enables clean and portable builds of oneDNN across RISC-V toolchains with or without RVV support, without requiring manual changes to the build configuration.

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -121,6 +121,43 @@ if(DNNL_WITH_SYCL)
     endif()
 endif()
 
+if (DNNL_TARGET_ARCH STREQUAL "RV64")
+    # Check if the RVV Intrinsics can be compiled with the current toolchain and flags
+    include(CheckCXXSourceCompiles)
+    check_cxx_source_compiles("#if !defined(__riscv) || !defined(__riscv_v)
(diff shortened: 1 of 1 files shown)


vpirogov:
make test
disable test_device_cpu
disable test_device_gpu

vpirogov (approved):

PR #3456, opened 2025-06-20
api: graph: remove dnnl_graph_op_last_symbol

TaoLv:
Fixes MFDNN-13011

The op kind is not intended for public usage. But with it in the list, we have seen ABI compatibility issue every time when a new op kind is added to the library and API. Now moving it to internal.

### DIFF
--- a/src/graph/interface/c_types_map.hpp
+++ b/src/graph/interface/c_types_map.hpp
@@ -213,7 +213,8 @@ const op_kind_t Tanh = dnnl_graph_op_tanh;
 const op_kind_t TanhBackward = dnnl_graph_op_tanh_backward;
 const op_kind_t TypeCast = dnnl_graph_op_type_cast;
 const op_kind_t Wildcard = dnnl_graph_op_wildcard;
-const op_kind_t LastSymbol = dnnl_graph_op_last_symbol;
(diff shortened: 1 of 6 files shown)


src/graph/interface/c_types_map.hpp
```diff
@@ -213,7 +213,8 @@ const op_kind_t Tanh = dnnl_graph_op_tanh;
 const op_kind_t TanhBackward = dnnl_graph_op_tanh_backward;
 const op_kind_t TypeCast = dnnl_graph_op_type_cast;
 const op_kind_t Wildcard = dnnl_graph_op_wildcard;
-const op_kind_t LastSymbol = dnnl_graph_op_last_symbol;
+// Increase it once a new op is added.
+const op_kind_t LastSymbol = static_cast<op_kind_t>(87);
```
densamoilov, line 217: I'm a bit concerned that this approach is error prone. Have you considered adjusting ABI testing to avoid marking it as a legit issue? I think it could be possible to pre-process our headers to remove these sentinel-type symbols from them before running the test.

TaoLv:
The ABI compatibility check of `dnnl_graph_op_last_symbol` now is skipped on the infra side. This PR is not needed any more.

PR #3459, opened 2025-06-22
cpu: x64: matmul: Adjust heuristics for postops bound layers

yair-obodovsky:
Estimate the size of the post-operation for the blocking heuristics. Avoid splitting dimension K if bound on post-processing , and incorporate post-processing information into the core work partitioning algorithm.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
@@ -112,6 +112,7 @@ struct brgemm_matmul_conf_t {
     bool packed_sparse_weights;
     bool req_transpose_scales;
     bool with_wei_decompression;
+    int postops_inst_count;
     brgemm_broadcast_t src_zp_type;
     brgemm_broadcast_t wei_zp_type;
     brgemm_broadcast_t dst_zp_type;
--- a/src/cpu/x64/matmul/postops_estimator.hpp
+++ b/src/cpu/x64/matmul/postops_estimator.hpp
@@ -1,0 +1,114 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef CPU_X64_MATMUL_POSTOPS_ESTIMATOR_HPP
+#define CPU_X64_MATMUL_POSTOPS_ESTIMATOR_HPP
+
+#include "cpu/x64/injectors/jit_uni_binary_injector.hpp"
+#include "cpu/x64/injectors/jit_uni_eltwise_injector.hpp"
+#include "cpu/x64/injectors/jit_uni_postops_injector.hpp"
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace x64 {
+namespace matmul {
+
+// The postops estimator JIT-compiles post-operation code for a single
+// cache line result produced by the GEMM. The size of the generated code
+// serves as a metric to help blocking heuristics account for the impact
+// of post-operations on execution performance. Note: the generated code
+// is not executed.
+
+class postops_estimator_t : public jit_generator_t {
+public:
+    static status_t estimate_insts_per_cacheline(memory_desc_t &dst_md,
+            primitive_attr_t &attr, int &estimated_vec_insts) {
+        postops_estimator_t post_ops_gen(dst_md, attr);
+        status_t res = post_ops_gen.check_status();
+        if (res != status::success) return res;
+        post_ops_gen.generate();
+        res = post_ops_gen.check_status();
+        if (res != status::success) return res;
+
+        size_t code_size = post_ops_gen.getSize();
+        size_t estimated_vec_code_size = (size_t)nstl::max(
+                (int)code_size - (int)mean_none_vec_code_bytes, 0);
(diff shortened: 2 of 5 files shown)


src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,88 @@
...
+                po_injector_t::create(this,
+                        impl::cpu::x64::cpu_isa_t::avx512_core_amx,
+                        attr.post_ops_, bsp, esp));
+        if (st != status::success) {
+            assert(!"postops_injector creation failed");
+        }
+    }
+
+    const char *name() const final { return nullptr; }
+    const char *source_file() const final { return nullptr; }
```
densamoilov: Are these not expected to be called? If so, can you please add asserts there?

src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "cpu/x64/injectors/jit_uni_binary_injector.hpp"
+#include "cpu/x64/injectors/jit_uni_eltwise_injector.hpp"
+#include "cpu/x64/injectors/jit_uni_postops_injector.hpp"
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace x64 {
+namespace matmul {
+
+class postops_estimator_t : public jit_generator_t {
```
densamoilov, line 35: Please document the idea and implementation so that other developers can understand it.

src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,88 @@
...
+public:
+    static size_t estimate_insts_per_cacheline(
+            memory_desc_t &dst_md, primitive_attr_t &attr) {
+        postops_estimator_t post_ops_gen(dst_md, attr);
+        post_ops_gen.generate();
```
densamoilov, line 42: We always call this `generate` member indirectly via `create_kernel()` that returns a status, which is safer from the error handling perspective.

src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+        dnnl::impl::cpu::x64::eltwise_injector::static_params_t esp;
+        esp.preserve_vmm = false;
+        esp.preserve_p_table = false;
+
+        auto st = safe_ptr_assign(postops_injector_,
+                po_injector_t::create(this,
+                        impl::cpu::x64::cpu_isa_t::avx512_core_amx,
+                        attr.post_ops_, bsp, esp));
+        if (st != status::success) {
```
densamoilov, line 82: The lack of error handling could cause `estimate_insts_per_cacheline` to crash if `po_injector_t::create` fails. So, introducing `postops_estimator_t::create` that can return `nullptr` to indicate an error might be a better option to make the code less bug-prone.

yair-obodovsky:
make test
set test_scope=NIGHTLY

yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,88 @@
...
+                        get_all_strategies_supported_by_injector(),
+                rhs_sp, nullptr, nullptr);
+
+        dnnl::impl::cpu::x64::eltwise_injector::static_params_t esp;
+        esp.preserve_vmm = false;
+        esp.preserve_p_table = false;
+
+        auto st = safe_ptr_assign(postops_injector_,
+                po_injector_t::create(this,
+                        impl::cpu::x64::cpu_isa_t::avx512_core_amx,
```
densamoilov, line 80: btw, is using `avx512_core_amx` here future proof?

yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/postops_estimator.hpp
yair-obodovsky: done

yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/postops_estimator.hpp
yair-obodovsky: done

src/cpu/x64/matmul/postops_estimator.hpp
yair-obodovsky: added error handling and avoided calling create_kernel, because it registers the code (we don't want this)

src/cpu/x64/matmul/postops_estimator.hpp
yair-obodovsky: added error handling

src/cpu/x64/matmul/postops_estimator.hpp
yair-obodovsky: this code is correct only for for AMX for now. if we will want to use it for a different isa we will have to enhance it

src/cpu/x64/matmul/postops_estimator.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+// The postops estimator JIT-compiles post-operation code for a single
+// cache line result produced by the GEMM. The size of the generated code
+// serves as a metric to help blocking heuristics account for the impact
+// of post-operations on execution performance. Note: the generated code
+// is not executed.
+
+class postops_estimator_t : public jit_generator_t {
+public:
+    static dnnl_status_t estimate_insts_per_cacheline(memory_desc_t &dst_md,
```
densamoilov: (nit) `dnnl_status_t` is an external C type that is not supposed to be used internally. Internally we need to use the corresponding aliases: `dnnl_status_t` -> `status_t`, `dnnl_success` -> `status::success`.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -112,6 +112,7 @@ struct brgemm_matmul_conf_t {
     bool packed_sparse_weights;
     bool req_transpose_scales;
     bool with_wei_decompression;
+    int post_inst_count;
```
densamoilov: ?
```suggestion
    int po_inst_count;
```

PR #3463, opened 2025-06-23
[GPU] benchdnn: memory: remove rng_mem

yehudaorel:
Fixes - https://jira.devtools.intel.com/browse/MFDNN-13848

Remove rng memory fill in benchdnn due to build linking errors.


echeresh (approved):

dzarukin (commented):
The clean approach is to use:
`git revert <commit-to-revert>`, it will use a proper commit message which is easy to match to the original one.

And updating the description mentioning it's a pure revert is preferrable.

Issue #3181, opened 2025-04-29
Does oneDNN has locale limitations?


dan20210809:
Hi, oneDNN developers

I am using OpenVINO2025.1 to do dGPU (A770) inference, but I found if I set:
 std::locale::global(std::locale("chs"));
or
 std::locale::global(std::locale("En_US"));
or
 std::locale::global(std::locale(""));
The inference will crash on oneDNN, but If I removed the locale set, the inference will be fine.
This is part of the crash log:

> 1:153:23: note: expanded from macro '\
> CONCAT2'
> #define CONCAT2(a, b) CONCAt2(a, b)
>                       ^
> note: (skipping 7 expansions in backtrace; use -fmacro-backtrace-limit=0 to see all)
> 1:152:23: note: expanded from macro 'CONCAt2'
> #define CONCAt2(a, b) a##b
>                       ^
> <scratch space>:213:1: note: expanded from here
> PO_0_ELTWISE_SCALE
> ^
> <command line>:329:29: note: expanded from here
> #define  PO_0_ELTWISE_SCALE as_float(0x3f,800,000)
>                             ^
> 1:12553:18: warning: left operand of comma operator has no effect
> const off_t oh = GWS_GET_OH();
>                  ^~~~~~~~~~~~
> <command line>:27:21: note: expanded from here
> #define  GWS_GET_OH GWS0_GET_ID2
>                     ^
> 1:887:63: note: expanded from macro 'GWS0_GET_ID2'
> #define GWS0_GET_ID2() GWS0_OP2(GWS_GET_THREAD_ID(GWS0_IDX2), GWS0_STRIDE2, div_up(GWS0_DIM2, GWS0_BLOCK2)) / GWS0_VEC_SIZE2 * GWS0_VEC_SIZE2 * GWS0_BLOCK2
>                                                               ^~~~~~~~~~~~
> <command line>:145:23: note: expanded from here
> #define  GWS0_STRIDE2 4,608
>                       ^
> 1:884:48: note: expanded from macro 'GWS_OP_MOD'
> #define GWS_OP_MOD(idx, stride, max) ((idx) / (stride) % (max))
>                                                ^~~~~~
> ,src\gpu\intel\ocl\engine.cpp:174
> onednn_verbose,v1,primitive,error,ocl,errcode -11,CL_BUILD_PROGRAM_FAILURE,src\gpu\intel\ocl\engine.cpp:278,src\gpu\intel\ocl\engine.cpp:278

[...]

PR #3464, opened 2025-06-23
common: verbose: fix locale handling for verbose logging

avmanerikar:
# Description

Addresses MFDNN-13634.

The PR removes locale-related dependencies during verbose logging, occurring because of the use of `stringstream` instances for string operations. It introduces DNNL wrappers for the standard `std::stringstream` classes to ensure locale-independent string handling.

### DIFF
--- a/tests/gtests/graph/unit/backend/dnnl/test_pass.cpp
+++ b/tests/gtests/graph/unit/backend/dnnl/test_pass.cpp
@@ -5061,7 +5061,7 @@ TEST(test_pass, InputJsonIsValid) {
     auto pm = dnnl::impl::graph::pass::pass_manager_t(
             backend_ptr.get_pass_registry());
 
-    std::ostringstream valid_stream;
+    dnnl::impl::ostringstream_t valid_stream;
     std::string version = std::to_string(dnnl_version()->major) + "."
             + std::to_string(dnnl_version()->minor) + "."
             + std::to_string(dnnl_version()->patch);
@@ -5084,7 +5084,7 @@ TEST(test_pass, InputJsonIsValid) {
                  << "]\n"
                  << "}\n";
     std::string valid_str = valid_stream.str();
-    std::istringstream valid_is(valid_str);
+    dnnl::impl::istringstream_t valid_is(valid_str);
     pm.run_passes(agraph, &valid_is);
     ASSERT_EQ(agraph.get_num_partitions(), 2U);
 }
@@ -5118,7 +5118,7 @@ TEST(test_pass, InputJsonIsInvalidWithIncompleteHash) {
     auto pm = dnnl::impl::graph::pass::pass_manager_t(
             backend_ptr.get_pass_registry());
 
-    std::ostringstream invalid_stream;
+    dnnl::impl::ostringstream_t invalid_stream;
     std::string version = std::to_string(dnnl_version()->major) + "."
             + std::to_string(dnnl_version()->minor) + "."
             + std::to_string(dnnl_version()->patch);
@@ -5143,7 +5143,7 @@ TEST(test_pass, InputJsonIsInvalidWithIncompleteHash) {
                    << "]\n"
                    << "}\n";
     std::string invalid_str = invalid_stream.str();
-    std::istringstream invalid_is(invalid_str);
+    dnnl::impl::istringstream_t invalid_is(invalid_str);
     pm.run_passes(agraph, &invalid_is);
     ASSERT_EQ(agraph.get_num_partitions(), 1U);
 }
@@ -5177,7 +5177,7 @@ TEST(test_pass, InputJsonIsInvalidWithMissingFiled) {
     auto pm = dnnl::impl::graph::pass::pass_manager_t(
             backend_ptr.get_pass_registry());
 
-    std::ostringstream invalid_stream;
+    dnnl::impl::ostringstream_t invalid_stream;
     invalid_stream << "{\n"
                    << "\"passes\": [\n"
                    << "  {\n"
@@ -5195,7 +5195,7 @@ TEST(test_pass, InputJsonIsInvalidWithMissingFiled) {
                    << "]\n"
                    << "}\n";
     std::string invalid_str = invalid_stream.str();
-    std::istringstream invalid_is(invalid_str);
+    dnnl::impl::istringstream_t invalid_is(invalid_str);
     pm.run_passes(agraph, &invalid_is);
     ASSERT_EQ(agraph.get_num_partitions(), 1U);
 }
@@ -5229,7 +5229,7 @@ TEST(test_pass, InputJsonIsInvalidWithWrongFormat) {
     auto pm = dnnl::impl::graph::pass::pass_manager_t(
             backend_ptr.get_pass_registry());
 
-    std::ostringstream invalid_stream;
+    dnnl::impl::ostringstream_t invalid_stream;
     std::string version = std::to_string(dnnl_version()->major) + "."
             + std::to_string(dnnl_version()->minor) + "."
             + std::to_string(dnnl_version()->patch);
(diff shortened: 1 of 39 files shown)


avmanerikar:
make test

xuxinzen (approved):

spalicki:
Does it have to be manually set each time? Setting the global C++ locale to `std::locale::classic()` does not work?

avmanerikar:
> Does it have to be manually set each time? Setting the global C++ locale to `std::locale::classic()` does not work?

We would not want to alter any process-wide settings by changing the global C++ locale. The locale limitations are only for oneDNN.

spalicki (approved):

avmanerikar:
make test
set test_scope=NIGHTLY

atkassen:
Does https://github.com/uxlfoundation/oneDNN/blob/e17917251732693c50ad373472690c500e135ade/src/cpu/x64/jit_uni_reorder_utils.cpp#L610-L611 (and aarch64 counterpart) also need updating?

avmanerikar:
> Does
> 
> https://github.com/uxlfoundation/oneDNN/blob/e17917251732693c50ad373472690c500e135ade/src/cpu/x64/jit_uni_reorder_utils.cpp#L610-L611
> 
> (and aarch64 counterpart) also need updating?

No. The operation does not involve floating-point values and will be unaffected by locale changes.

atkassen:
> > Does
> > https://github.com/uxlfoundation/oneDNN/blob/e17917251732693c50ad373472690c500e135ade/src/cpu/x64/jit_uni_reorder_utils.cpp#L610-L611
> > 
> > (and aarch64 counterpart) also need updating?
> 
> No. The operation does not involve floating-point values and will be unaffected by locale changes.

#3181 didn't either, a hex integer literal was being formatted incorrectly.

avmanerikar:
> > > Does
> > > https://github.com/uxlfoundation/oneDNN/blob/e17917251732693c50ad373472690c500e135ade/src/cpu/x64/jit_uni_reorder_utils.cpp#L610-L611
> > > 
> > > (and aarch64 counterpart) also need updating?
> > 
> > 
> > No. The operation does not involve floating-point values and will be unaffected by locale changes.
> 
> #3181 didn't either, a hex integer literal was being formatted incorrectly.

Updated in the latest commit. Thanks!

avmanerikar:
@uxlfoundation/onednn-cpu-aarch64, can you please review the changes? Thanks!

src/common/gemm.cpp
```diff
@@ -73,6 +73,7 @@ std::string get_descriptor(dim_t M, dim_t N, dim_t K) {
         status = __VA_ARGS__; \
         double duration_ms = get_msec() - start_ms; \
         std::stringstream ss; \
+        ss.imbue(std::locale::classic()); \
```
rjoursler: Would it make sense to add a dnnl specific wrapper type which handles this, so that developers don't have to remember to use this in other places?

src/common/gemm.cpp
```diff
@@ -73,6 +73,7 @@ std::string get_descriptor(dim_t M, dim_t N, dim_t K) {
...
         ss << "cpu,gemm_api,,undef,"; \
```
rjoursler, line 76: Could we add some a small regression tests suite to ensure the oneDNN functionality works as expected under different locales?

src/common/gemm.cpp
jondea: Agreed, if oneDNN is making the decision that user locale should be ignored when printing, then we should ignore it centrally otherwise this bug will just keep creeping back in.

avmanerikar: Thanks for the suggestion. I updated the PR to reflect these changes. Instead of updating the `stringstream` instances, the PR now introduces a dnnl locale-safe wrapper to replace them.

src/common/gemm.cpp
avmanerikar: Currently, the PR replaces all the STL stream instances with the dnnl wrappers. Referring to the suggestions in MFDNN-13634, a more useful test will be to catch any direct uses of `stringstream` or similar calls via CI.

src/common/utils.hpp
```diff
@@ -687,6 +689,31 @@ int getenv_int_user(const char *name, int default_value = 0);
...
+// These are locale-invariant wrappers to define streaming objects for
+// string manipulation. Use these instead of the std library variants, namely,
+// std::stringstream, std::istringstream and std::ostringstream to ensure
+// locale-independent string behavior.
+struct dnnl_stringstream_t : public std::stringstream {
+    dnnl_stringstream_t() { this->imbue(std::locale::classic()); }
+    dnnl_stringstream_t(const std::string &str) : std::stringstream(str) {
+        this->imbue(std::locale::classic());
+    }
+};
```
atkassen: It'd be nice to keep the interfaces as close to the originals as possible while avoiding any further changes to the locale.

```suggestion
struct dnnl_stringstream_t : public std::stringstream {
    template <typename ... Args>
    dnnl_stringstream_t(Args &&...args) : std::stringstream(std::forward<Args>(args)...) {
        imbue(std::locale::classic());
    }

private:
    using stringstream::imbue;
};
```

avmanerikar: Done. Thanks!

TaoLv (approved):

avmanerikar:
make test
set test_scope=NIGHTLY

tests/gtests/graph/unit/backend/dnnl/test_pass.cpp
```diff
@@ -5143,7 +5143,8 @@ TEST(test_pass, InputJsonIsInvalidWithIncompleteHash) {
                    << "]\n"
                    << "}\n";
     std::string invalid_str = invalid_stream.str();
-    std::istringstream invalid_is(invalid_str);
+    dnnl::impl::dnnl_istringstream_t invalid_is;
+    invalid_is.str(invalid_str);
```
atkassen: nit: Here and below
```suggestion
    dnnl::impl::dnnl_istringstream_t invalid_is(invalid_str);
```

avmanerikar: Fixed. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

rjoursler (approved):

src/common/utils.hpp
```diff
@@ -687,6 +689,43 @@ int getenv_int_user(const char *name, int default_value = 0);
 // "DNNL_" (secondary) prefixes.
 std::string getenv_string_user(const char *name);
 
+// These are locale-invariant wrappers to define streaming objects for
+// string manipulation. Use these instead of the std library variants, namely,
+// std::stringstream, std::istringstream and std::ostringstream to ensure
+// locale-independent string behavior.
+struct dnnl_stringstream_t : public std::stringstream {
```
rjoursler: Comment: The `dnnl_` prefix is generally used for objects exposed by the external API. Could we drop the prefix to avoid confusing developers, especially as this is already contained in the `dnnl` namespace?

avmanerikar: Sure. But dropping the prefix dnnl_ may lead to confusion between the `stringstream_t` wrapper and the STL's `stringstream`. If there is an alternative preferred prefix, I'll update the terms accordingly.

rjoursler: IMO, the fact that we have the `_t` postfix is sufficient notification that this is a dnnl type. We already have a number of custom wrappers to handle C++11 interoperability (e.g [optional_t](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/optional.hpp), [enable_if](https://github.com/uxlfoundation/oneDNN/blob/bd26fbc77885afd36555306106af509db38c5707/src/common/utils.hpp#L110), etc), and historically, I haven't seen much confusion around these objects.

avmanerikar: Updated. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

src/gpu/intel/jit/gemm/generator/pieces/problem_utils.cpp
```diff
@@ -21,7 +21,7 @@
 GEMMSTONE_NAMESPACE_START
 
 using namespace ngen;
-
+using namespace dnnl::impl;
```
petercad: Hi @avmanerikar -- gemmstone code in the `src/gpu/intel/jit/gemm/{generator,include}` directories is an external component and should not use oneDNN-specific data types unless it's unavoidable.

avmanerikar: Updated. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

PR #3467, opened 2025-06-23
ci: aarch64: Fix nightly performance tests

Ryo-not-rio:
The nightly performance tests are failing due to benchdnn no longer accepting underscores for the matmul operator

### DIFF
--- a/.github/automation/performance/benchdnn_comparison.py
+++ b/.github/automation/performance/benchdnn_comparison.py
@@ -56,10 +56,15 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
     r2_ctime = defaultdict(list)
 
     for key, exec_time, ctime in r1:
+        # Older versions of benchdnn outputs underscores
+        # instead of hyphens for some ops leading to
+        # mismatches in problems with newer versions
+        key = key.replace("_", "-")
         r1_exec[key].append(float(exec_time))
         r1_ctime[key].append(float(ctime))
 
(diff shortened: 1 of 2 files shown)


.github/automation/performance/benchdnn_comparison.py
```diff
@@ -56,10 +56,12 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
     r2_ctime = defaultdict(list)
 
     for key, exec_time, ctime in r1:
+        key = key.replace("_", "-")
```
Sqvid, line 62: Could we please add a comment above these two lines with a TODO so that we can remove them once the base version is updated?

Sqvid (changes requested):

.github/automation/performance/benchdnn_comparison.py
Ryo-not-rio: I think we should keep this as is because we will need this script for the performance comparison pipeline that @gausah-arm is working on right now. And in that pipeline, people can use whatever base version they want

Sqvid: I see. No need to TODO it then but I think it still needs an explanatory comment that older versions need the hyphen rather than underscore. Without that context these lines seem very random.

Ryo-not-rio: Done

vpirogov (approved):

PR #3470, opened 2025-06-24
graph: backend: dnnl: add reduction dimention check for NV GPU

Jiexin-Zheng:
Fix MFDNN-13550

### DIFF
--- a/src/graph/backend/dnnl/passes/lower.cpp
+++ b/src/graph/backend/dnnl/passes/lower.cpp
@@ -187,6 +187,15 @@ static status_t batchnorm_fwd_handler(
 
 static status_t reduction_handler(
         const std::shared_ptr<op_t> &op, subgraph_rewriter_t &rewriter) {
+
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    auto src_lt = op->get_input_values()[0]->get_logical_tensor();
+    auto src_nelems = ltw(src_lt).nelems();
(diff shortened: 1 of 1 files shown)


src/graph/backend/dnnl/passes/lower.cpp
```diff
@@ -187,6 +187,20 @@ static status_t batchnorm_fwd_handler(
...
+
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+    auto product = [](const dnnl::memory::dims &dims) {
+        return std::accumulate(dims.begin(), dims.end(),
+                static_cast<dnnl::memory::dim>(1),
+                std::multiplies<dnnl::memory::dim>());
+    };
+    const auto src_lt = op->get_input_values()[0]->get_logical_tensor();
+    auto src_dims = product(ltw(src_lt).vdims());
```
rongzha1: seems `total_elems` is more accuracy than `src_dims` in this code.  

In this function, it means total elems should less than 65536, no like the comments "dimension exceeds 65535"

could you double check it?

TaoLv: We already have the same functionality by `ltw(src_lt).nelems()`. See https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/interface/logical_tensor.hpp#L171 .

Jiexin-Zheng: I have applied your suggestion, thanks~

xiang1guo (approved):

TaoLv:
@Jiexin-Zheng please rebase and test it. Thanks.

PR #3473, opened 2025-06-24
tests: benchdnn: graph: enable cpu mem_pool

rongzha1:
complete task of JIRA 13799.

enable a simple benchdnn graph cpu mem_pool as gpu mem_pool.

### DIFF
--- a/tests/benchdnn/graph/allocator.cpp
+++ b/tests/benchdnn/graph/allocator.cpp
@@ -65,34 +65,36 @@ bool check_memory_size(size_t size) {
 
 } // namespace
 
-#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_SYCL \
-        || DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
 void graph_mem_manager_t::clear_memory_pool() {
     mem_pool_.clear();
 }
-#endif
 
-void *graph_mem_manager_t::host_malloc_wrapper(
-        size_t size, size_t alignment) const {
-    CHECK_GRAPH_MEM_SIZE(need_mem_check_, size);
+bool graph_mem_manager_t::enable_host_mem_pool() {
+    return (!has_bench_mode_bit(mode_bit_t::corr)) && is_cpu();
+}
 
+void *graph_mem_manager_t::host_malloc_wrapper(size_t size, size_t alignment) {
     void *ptr = nullptr;
-    const size_t align = alignment == 0 ? default_alignment : alignment;
-#ifdef _WIN32
-    ptr = _aligned_malloc(size, align);
-    int rc = ((ptr) ? 0 : errno);
-#else
-    int rc = ::posix_memalign(&ptr, align, size);
-#endif /* _WIN32 */
-    return (rc == 0) ? ptr : nullptr;
+    if (enable_host_mem_pool()) {
+        bool need_alloc_new_mm = mem_pool_.check_allocated_mem(ptr, size);
+
+        if (need_alloc_new_mm) {
+            CHECK_GRAPH_MEM_SIZE(need_mem_check_, size);
+            ptr = mem_pool_.allocate(size, alignment);
+        }
+    } else {
+        CHECK_GRAPH_MEM_SIZE(need_mem_check_, size);
+        if (alignment == 0) alignment = default_alignment;
+        ptr = host_malloc(size, alignment);
+    }
+    return ptr;
 }
 
 void graph_mem_manager_t::host_free_wrapper(void *ptr) {
-#ifdef _WIN32
-    _aligned_free((void *)ptr);
-#else
-    ::free((void *)ptr);
-#endif /* _WIN32 */
+    if (enable_host_mem_pool())
+        mem_pool_.deallocate(ptr);
+    else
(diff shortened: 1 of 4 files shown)


rongzha1:
make test
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/allocator.cpp
```diff
@@ -87,14 +84,37 @@ void *graph_mem_manager_t::host_malloc_wrapper(
...
 #ifdef _WIN32
     _aligned_free((void *)ptr);
 #else
     ::free((void *)ptr);
 #endif /* _WIN32 */
 }
 
+void *graph_mem_manager_t::host_malloc_wrapper(size_t size, size_t alignment) {
+    void *ptr = nullptr;
+    if (!has_bench_mode_bit(mode_bit_t::corr) && is_cpu()) {
```
dzarukin: Would you find it beneficial to apply mem pool for all modes, not just performance?

If you don't, I suggest to have a single `private` call to wrap `!has_bench_mode_bit(mode_bit_t::corr) && is_cpu()` condition applied for malloc and free wrappers as a single switch when to enable/disable mem_pool.

rongzha1:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

wzt1997 (approved):

rongzha1:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/memory_pool.hpp
```diff
@@ -77,12 +77,28 @@ static void ocl_free(
 }
 #endif
 
+static void *host_malloc(size_t size, size_t alignment) {
```
dzarukin: I'm not sure why `static` appeared in all function declarations at the first place. This way each translation unit got their own copy of the function. It should be `inline` instead here and everywhere.

The second thing - allocator calls could reuse these calls to avoid code duplication since these functions are standalone.

tests/benchdnn/graph/memory_pool.hpp
```diff
@@ -102,6 +118,26 @@ class simple_memory_pool_t {
...
+            ptr = sh_ptr.get();
+            // record the map of mm size and its ptr for reuse
+            map_size_ptr_.emplace(size, sh_ptr);
+            is_free_ptr_[ptr] = false;
+        }
+        return ptr;
+    }
+
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_SYCL \
+        || DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
```
dzarukin: That's a little bit clumsy, let's split them properly:
```
#if OCL
...
#elif SYCL
...
#endif
```

tests/benchdnn/graph/allocator.cpp
rongzha1: got it. thanks for your suggest.

tests/benchdnn/graph/memory_pool.hpp
rongzha1: Good idea. changed as following:
1. replace static with inline.
2. reuse calls in allocator

tests/benchdnn/graph/memory_pool.hpp
rongzha1: changed in this way. Could you review it again？ Thanks

tests/benchdnn/graph/allocator.cpp
dzarukin: What about this part?
> Would you find it beneficial to apply mem pool for all modes, not just performance?

Edit:
I see the correctness mode is still restricted... Though if correctness gets this support, it would be better to rely on zmalloc (benchdnn's version of allocation) which guard's stuff, and I'm not sure if it would make sense to keep memory pool in that case, so, up to the team. Feel free to resolve.

tests/benchdnn/graph/memory_pool.hpp
```diff
@@ -77,12 +77,28 @@ static void ocl_free(
 }
 #endif
 
+inline void *host_malloc(size_t size, size_t alignment) {
```
dzarukin, line 80: Would you mind to update calls above from `static` to `inline` as well?

dzarukin (approved):

tests/benchdnn/graph/allocator.cpp
rongzha1: Thanks for your advice. 

This simple memory pool only reuses requests with the exact same memory size requirements, making it very suitable for performance mode scenarios where a fixed-size scratchpad is allocated each time.

In correct mode, the scratchpad memory is only allocated once, so from a theoretical standpoint, there is no significant advantage.

I would keep it in performance mode.

tests/benchdnn/graph/memory_pool.hpp
rongzha1: got it. just fixed them in the 2nd commit.

rongzha1:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3477, opened 2025-06-24
benchdnn: utils: parser: resolve UB with string search

dzarukin:
Fixes Memory sanitizer hits in Weekly, and Address sanitizer hits on Windows.

### DIFF
--- a/tests/benchdnn/utils/parser.hpp
+++ b/tests/benchdnn/utils/parser.hpp
@@ -40,12 +40,25 @@ extern const size_t eol;
 extern std::stringstream help_ss;
 
 namespace parser_utils {
+
 std::string get_pattern(const std::string &option_name, bool with_args = true);
+
 void add_option_to_help(const std::string &option,
         const std::string &help_message, bool with_args = true);
+
 int64_t stoll_safe(const std::string &s);
+
 float stof_safe(const std::string &s);
+
 attr_t::post_ops_t parse_attr_post_ops_func(const std::string &s);
+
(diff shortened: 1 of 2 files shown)


avmanerikar (approved):

dzarukin:
make test

tests/benchdnn/utils/parser.hpp
```diff
@@ -40,12 +40,25 @@ extern const size_t eol;
...
         const std::string &help_message, bool with_args = true);
+
 int64_t stoll_safe(const std::string &s);
+
 float stof_safe(const std::string &s);
+
 attr_t::post_ops_t parse_attr_post_ops_func(const std::string &s);
+
+// `option_str` is a string in a format `--option-name=`.
+inline bool option_matched(const std::string &option_str, const char *str) {
```
densamoilov, line 56: Is there a reason for this function to be defined in the header? All other functions in this namespace are defined in the cpp file.

densamoilov (approved):

tests/benchdnn/utils/parser.hpp
dzarukin: Yes, it's a follow-up on Eugene's finding that not inlined small functions (though it was a class method, but I extrapolated it) may cause noticeable overheads on medium and large batch files. This function is called 10-20 times per each descriptor, I decided to have it `inline`d. I'll follow up on other functions here just kept the change specific to the issue resolution.

densamoilov: ok, thanks.

PR #3480, opened 2025-06-25
generic: sycl: Add check on number of post-ops used

s-Nick:
# Description

Due to technical limitation the generic backend cannot support more than 5 post-ops. This PR implement a check during test to skip those tests instead of reporting unimplemented.

# Checklist

## General

The issue was presented to us as part of `deconv` testing, so I re-run that test suite to confirm it doesn't break anything. I attach here tests with this PR and master. These changes reduce the number of unimplemented without compromising correctness of other tests. The number of tests run is the same

- PR: [test-deconv-generic-intel-pvc.txt](https://github.com/user-attachments/files/20900577/test-deconv-generic-intel-pvc.txt)
- main: [test-deconv-main-generic-pvc.txt](https://github.com/user-attachments/files/20900582/test-deconv-main-generic-pvc.txt)

### DIFF
--- a/tests/benchdnn/binary/binary.cpp
+++ b/tests/benchdnn/binary/binary.cpp
@@ -137,9 +137,7 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     std::vector<dnnl_data_type_t> dts = {prb->sdt[0], prb->sdt[1], prb->ddt};
     skip_unimplemented_data_type(dts, prb->dir, res);
-    skip_unimplemented_arg_scale(prb->attr, res);
-    skip_unimplemented_binary_po(prb->attr, res);
-    skip_unimplemented_prelu_po(prb->attr, res, dnnl_binary);
+    skip_unimplemented_attr(prb->attr, res, dnnl_binary, dnnl_data_type_undef);
 
     if (is_gpu()) {
         // N.B: Adding this for gpu as cfg is not supported in POST-OPS
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -849,6 +849,29 @@ void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res) {
     }
 }
 
+void skip_unimplemented_po_quantity(const attr_t &attr, res_t *res) {
+    // Generic SYCL backend currently cannot support port more than 5 post-ops.
+    if (is_generic_gpu()) {
+        if (attr.post_ops.len() > 5) {
+            res->state = SKIPPED;
+            res->reason = skip_reason::case_not_supported;
+            return;
+        }
+    }
+}
+
+void skip_unimplemented_attr(const attr_t &attr, res_t *res,
+        dnnl_primitive_kind_t prim_kind, dnnl_data_type_t src_dt,
+        dnnl_data_type_t dst_t) {
+
+    skip_unimplemented_sum_po(attr, res, prim_kind, src_dt, dst_t);
+    skip_unimplemented_binary_po(attr, res);
+    skip_unimplemented_prelu_po(attr, res, prim_kind);
+    skip_unimplemented_arg_scale(attr, res);
+
+    skip_unimplemented_po_quantity(attr, res);
+}
+
 void skip_invalid_inplace(res_t *res, dnnl_data_type_t sdt,
         dnnl_data_type_t ddt, const std::string &stag,
         const std::string &dtag) {
--- a/tests/benchdnn/dnnl_common.hpp
+++ b/tests/benchdnn/dnnl_common.hpp
@@ -268,6 +268,10 @@ void skip_unimplemented_prelu_po(
 void skip_invalid_inplace(res_t *res, dnnl_data_type_t sdt,
         dnnl_data_type_t ddt, const std::string &stag, const std::string &dtag);
 void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res);
+void skip_unimplemented_po_quantity(const attr_t &attr, res_t *res);
+void skip_unimplemented_attr(const attr_t &attr, res_t *res,
+        dnnl_primitive_kind_t prim_kind, dnnl_data_type_t src_dt,
+        dnnl_data_type_t dst_t = dnnl_data_type_undef);
 
 template <typename prb_t>
 int check_caches(benchdnn_dnnl_wrapper_t<dnnl_primitive_t> &primw,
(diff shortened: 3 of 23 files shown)


tests/benchdnn/binary/binary.cpp
```diff
@@ -140,6 +140,7 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_arg_scale(prb->attr, res);
     skip_unimplemented_binary_po(prb->attr, res);
     skip_unimplemented_prelu_po(prb->attr, res, dnnl_binary);
+    if (is_generic_gpu()) { skip_unimplemented_num_po(prb->attr, res); }
```
Rbiessy: Is there a reason for not moving `is_generic_gpu()` inside `skip_unimplemented_num_po`? This skip is specific to the generic backend so would make sense to me.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -849,6 +849,17 @@ void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res) {
     }
 }
 
+void skip_unimplemented_num_po(const attr_t &attr, res_t *res) {
+// Generic SYCL backend currently cannot support port more than 5 post-ops.
+#if defined(DNNL_WITH_SYCL) && DNNL_GPU_VBENDOR == DNNL_VENDOR_GENERIC
```
AD2605:
```suggestion
#if defined(DNNL_WITH_SYCL) && DNNL_GPU_VENDOR == DNNL_VENDOR_GENERIC
```
Typo ?

dzarukin: I'd say it's time to introduce a common call for all attribute related checks:

```
void skip_unimplemented_attr(const attr_t &attr, res_t *res, dnnl_primitive_kind_t prim_kind) {
    ... (list of existing checks);
    a new check;
}
```

tests/benchdnn/dnnl_common.cpp
```diff
@@ -849,6 +849,17 @@ void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res) {
     }
 }
 
+void skip_unimplemented_num_po(const attr_t &attr, res_t *res) {
```
dzarukin:
```suggestion
void skip_unimplemented_po_quantity(const attr_t &attr, res_t *res) {
```

tests/benchdnn/dnnl_common.cpp
dzarukin: And please replace macros with a runtime check.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -849,6 +849,35 @@ void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res) {
...
+    }
+    if (prim_kind != dnnl_group_normalization) {
+        skip_unimplemented_binary_po(attr, res);
+        skip_unimplemented_prelu_po(attr, res, prim_kind);
+    }
+    if (prim_kind == dnnl_binary || prim_kind == dnnl_concat) {
+        skip_unimplemented_arg_scale(attr, res);
+    }
+
+    skip_unimplemented_po_quantity(attr, res);
```
AD2605: I am not sure if this function should be handling all the primitives, given it checks for the primitive kind and then their corresponding skips anyway. It would also make it easier to handle if we were to amend the list of unsupported configurations of a certain primitive.  

Similar to how we have `skip_unimplemented_prb` in each of the primitive's namespace, I think the correct way to go about this would be to implement a `check_unsupported_attr` in each of the namespace, which would then be called by `skip_unimplemented_prb`, making it much clearer which attributes the primitive does and does not support. 

Thoughts @dzarukin @sgeor255 @ShanoToni ?

tests/benchdnn/dnnl_common.hpp
```diff
@@ -268,6 +268,11 @@ void skip_unimplemented_prelu_po(
 void skip_invalid_inplace(res_t *res, dnnl_data_type_t sdt,
         dnnl_data_type_t ddt, const std::string &stag, const std::string &dtag);
 void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res);
+void skip_unimplemented_po_quantity(const attr_t &attr, res_t *res);
+void skip_unimplemented_attr(const attr_t &attr, res_t *res,
+        dnnl_primitive_kind_t prim_kind,
+        dnnl_data_type_t src_dt = dnnl_data_type_undef,
```
dzarukin: This default value is excessive.
```suggestion
        dnnl_data_type_t src_dt,
```

tests/benchdnn/dnnl_common.cpp
dzarukin: > I think the correct way to go about this would be to implement a check_unsupported_attr in each of the namespace

This kills the purpose of generalization. The whole point of the function is to add a common limitation for all drivers in a single spot.

Primitive specific limitations don't make their way into common functions.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -849,6 +849,35 @@ void skip_unimplemented_arg_scale(const attr_t &attr, res_t *res) {
...
+            return;
+        }
+    }
+}
+
+void skip_unimplemented_attr(const attr_t &attr, res_t *res,
+        dnnl_primitive_kind_t prim_kind, dnnl_data_type_t src_dt,
+        dnnl_data_type_t dst_t) {
+
+    if (prim_kind != dnnl_binary && prim_kind != dnnl_group_normalization) {
```
dzarukin: These `if`s shouldn't be here. It's just the unconditional list of function calls. Those functions should handle a configuration properly, thus, if there's no sum po, the function shouldn't do any checks.

tests/benchdnn/dnnl_common.cpp
s-Nick: Thank you for your review @dzarukin . Referring also to [this comment](https://github.com/uxlfoundation/oneDNN/pull/3480#discussion_r2172537795), could you clarify which changes are you suggesting? To me it seems I don't need to implement @AD2605  suggestion, but I need to remove the if-statements in this function and skip functions will deal with the postops properly. 

I added the if-statements because those skip functions weren't part of that specific primitive skips, but I am happy to remove them if not necessary.

dzarukin: The current change is fine with me. I've asked to remove `if`s because the function suppose to be universal for any driver. If some driver doesn't support certain attributes, it's fine, the correspondent function shouldn't do anything for them in that case.

s-Nick: Thank you for clarifying it. I addressed all your requests in latest commit.

openmodelzooci:

vpirogov:
Closing as stale.

PR #3487, opened 2025-06-26
graph: utils: pm: enhance matching of inputs

ElaineBao:
# Description

Enhance input matching logic in pattern matcher to support inputs without producers.

This update improves the pattern matcher's ability to handle inputs that do not have explicit producer operations. A typical case arises in SDPA training backward, where two parallel MatMul operations receive their `src0` inputs from the same SoftmaxBackward node. The only distinction between them lies in their src1 inputs, which are shared with the MatMul_QK's `src0` and `src1`, respectively. To correctly differentiate and match these two MatMul operations, the matcher must consider inputs even if they lack a producer.

### DIFF
--- a/tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
+++ b/tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
@@ -1564,6 +1564,109 @@ TEST(test_utils_pattern_matcher, SharedInputCase2) {
     ASSERT_EQ(fusion_ops2.size(), 0U);
 }
 
+TEST(test_utils_pattern_matcher, ParallelMatmulWithSharedInput) {
+    auto graphp = std::make_shared<pb_graph_t>();
+    /*  Pattern that captures two parallel MatMuls share inputs
+        with another MatMul
+           IN0    IN1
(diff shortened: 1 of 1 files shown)


ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/gtests/graph/unit/utils/test_pattern_matcher_cpu.cpp
```diff
@@ -1564,6 +1564,109 @@ TEST(test_utils_pattern_matcher, SharedInputCase2) {
     ASSERT_EQ(fusion_ops2.size(), 0U);
 }
 
+TEST(test_utils_pattern_matcher, ParallelMatmulWithSharedInput) {
```
Jiexin-Zheng, line 1567: Is it possible to add the original SDPA backward case into this PR for test?

ElaineBao: This PR doesn't support SDPA backward yet, I cannot add that case, but this UT can mimic the pattern of that case, so I think it's ok

PR #3496, opened 2025-06-27
generic: sycl: Skip all tests that use block format for generic backend

s-Nick:
# Description

Generic backend doesn't support block format. This PR implement a skip function to check if the format tag and skip the test in case of generic backend.

# Checklist

## General

### DIFF
--- a/tests/benchdnn/binary/binary.cpp
+++ b/tests/benchdnn/binary/binary.cpp
@@ -140,6 +140,10 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_arg_scale(prb->attr, res);
     skip_unimplemented_binary_po(prb->attr, res);
     skip_unimplemented_prelu_po(prb->attr, res, dnnl_binary);
+    for (const auto &tag : prb->stag) {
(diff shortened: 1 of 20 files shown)


s-Nick:
make test
disable test_device_cpu
enable build_vendor_generic
enable arch_gpu_xe-hpc
enable compiler_icx-oss

Rbiessy (approved):

vpirogov:
@dzarukin, does this change look right to you?

dzarukin:
> @dzarukin, does this change look right to you?

I was about to post a message that was not the agreement when the reorder change landed, but we may have more lists where specific blocked tags are called out, so I didn't.

However, this change won't help to decrease the amount of CI failures as we are not using blocked tags in ci lists, so I don't see why this is more important than changes that are exposed on day-to-day basis.

Rbiessy:
@uxlfoundation/onednn-arch we discussed this PR should be good to do.

vpirogov (approved):

openmodelzooci:

vpirogov:
Closing as stale.

PR #3214, opened 2025-05-05
benchdnn: inputs: matmul: add basic generated dequantization tests

rjoursler:
The matmul primitive has been consistently encountering bug escapes during GPU development (for example #3112, #3148, and #3195). This PR adds some very basic dequantization generation to synthdnn and regenerates harness_matmul_generated_ci to use the new tests help prevent such issues in the future.

PR #3498, opened 2025-06-27
xe: gemm: fixup per_oc zp detection

Simonsays095:
Addresses [MFDNN-13869](https://jira.devtools.intel.com/browse/MFDNN-13869). Quantized wei attributes were not appropriately detecting per_ocic cases that are equivalent to per_oc cases (which can be optimized further) due to dimension ordering issues. This is a quick fix to restore performance.

### DIFF
--- a/src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
+++ b/src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
@@ -23,6 +23,20 @@ namespace gpu {
 namespace intel {
 namespace jit {
 
+namespace {
+void quant_dims(
+        const memory_desc_t &md, const quant_entry_t &entry, dims_t &out) {
+    auto mask = entry.get_mask();
+    for (int i = 0; i < md.ndims; i++)
+        out[i] = md.dims[i] * ((mask >> i) & 1);
+    // Groups apply to the last 2 dims
+    if (!entry.has_default_groups()) {
+        out[md.ndims - 2] /= entry.get_group(0);
+        out[md.ndims - 1] /= entry.get_group(1);
+    }
+}
+} // anonymous namespace
+
 status_t jit_gemm_pd_t::init_post_ops() {
     using namespace primitive_kind;
     using namespace alg_kind;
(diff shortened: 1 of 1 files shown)


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

rjoursler:
Could you check this PR against the test set in #3214 to ensure there are no functional regressions in that test set?

Simonsays095:
> Could you check this PR against the test set in #3214 to ensure there are no functional regressions in that test set?

Tested against that test suite: main and this PR both have 39 failures, in the same layers.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

rjoursler (approved):

atkassen (approved):

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
```diff
@@ -239,32 +255,41 @@ void jit_gemm_pd_t::init_attrs() {
...
+    const auto c_zps = attr_zps.get(DNNL_ARG_C);
 
     const auto &scales = attr()->scales_;
-    const auto *src_scales = &attr()->scales_.get(DNNL_ARG_B);
-    const auto *wei_scales = &attr()->scales_.get(DNNL_ARG_A);
-    asc_dims_ = quant_entry_ndims(scales.get(DNNL_ARG_A), d->b_desc);
-    bsc_dims_ = quant_entry_ndims(scales.get(DNNL_ARG_B), d->a_desc);
-    wei_scales_group_k_ = wei_scales->get_group(0);
-    src_scales_group_k_ = src_scales->get_group(1);
+    const auto a_scales = scales.get(DNNL_ARG_A);
```
h-sadia, line 264: is a/b different than src/wei?

Simonsays095: A/B is the GEMM canonical way to refer to the matrices (AB=C). We may choose either src/wei for A (left matrix) depending on transpositions, matrix shapes, etc. So the preferred term to use inside of gemm code to be less ambiguous is A/B instead of src/wei.

Note: by definition, `DNNL_ARG_A == DNNL_ARG_WEI` and `DNNL_ARG_B == DNNL_ARG_SRC`, since we have to transpose and swap matrices to use gemmstone - it expects column major and we expect row major. Yet another reason to stick with one set of naming instead of switching back and forth.

kealan-barbieri (approved):

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
h-sadia: Right, I was just wondering if we could keep variable names synonymous through the code ... I see we use src/wei later on.

h-sadia (approved):

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
Simonsays095: The goal is to ultimately switch to A/B as quickly as possible and keep using those - it's just an ongoing project. I think we mainly just have a few class members that use src/wei naming, so we're almost there.

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
```diff
@@ -239,32 +255,41 @@ void jit_gemm_pd_t::init_attrs() {
...
+    cmask_b_ = b_zps.get_mask();
+    cmask_c_ = c_zps.get_mask();
+
+    // Swap descriptors to follow column major format.
+    ao_dims_ = quant_entry_ndims(a_zps, d->b_desc);
+    bo_dims_ = quant_entry_ndims(b_zps, d->a_desc);
+    asc_dims_ = quant_entry_ndims(a_scales, d->b_desc);
+    bsc_dims_ = quant_entry_ndims(b_scales, d->a_desc);
+
+    wei_scales_group_k_ = a_scales.get_group(0);
```
h-sadia, line 277: wei_scales_group_k can't be a_scales_group_k...? since you are already using a/b here and adding code. Its probably a minor refactor.

Simonsays095: I agree it's sort of confusing how it's named now, but renaming this variable will make naming disparities pop up elsewhere, since there are a few places left to change the name like this. There are also several variables that can be eliminated entirely (e.g. `src_po_sc_`, which is never set). Given the chance for typos during this refactor, I'll save it for a separate PR next week after the code freeze.

PR #3502, opened 2025-06-30
jit: ir: remove custom IR types from the core headers

echeresh:
PR is to get rid of centralized handling of IR objects. For non-core types we can use the IR object name for type information purposes - no need in updating the type id enumeration.

With that we will be able to extend the IR object system with custom IR objects without updating the core headers.

@rjoursler please take a look.

### DIFF
--- a/src/gpu/intel/jit/ir/core.cpp
+++ b/src/gpu/intel/jit/ir/core.cpp
@@ -424,7 +424,6 @@ void object_impl_t::_visit(ir_visitor_t &visitor) const {}
 DECL_TRAVERSE_LEAF(bool_imm_t)
 DECL_TRAVERSE_LEAF(const_var_t)
 DECL_TRAVERSE_LEAF(float_imm_t)
-DECL_TRAVERSE_LEAF(func_impl_t)
 DECL_TRAVERSE_LEAF(int_imm_t)
 DECL_TRAVERSE_LEAF(var_t)
 
@@ -492,16 +491,14 @@ void ir_visitor_t::_visit(const for_t &obj) {
 }
 
 object_t ir_mutator_t::_mutate(const func_call_t &obj) {
-    auto func = mutate(obj.func);
     auto args = mutate(obj.args);
 
-    if (func.is_same(obj.func) && ir_utils::is_same(args, obj.args)) return obj;
+    if (ir_utils::is_same(args, obj.args)) return obj;
 
-    return func_call_t::make(func, args, obj.attr);
+    return func_call_t::make(obj.func, args, obj.attr);
 }
 
 void ir_visitor_t::_visit(const func_call_t &obj) {
-    visit(obj.func);
     visit(obj.args);
 }
 
@@ -688,23 +685,6 @@ object_t ir_mutator_t::_mutate(const while_t &obj) {
 void ir_visitor_t::_visit(const while_t &obj) {
     visit(obj.cond);
     visit(obj.body);
-}
-
-// Catch missing mutates that are not expected to dispatch to the base
-// mutator
-object_t ir_mutator_t::_mutate(const nary_op_t &obj) {
-    gpu_error_not_expected() << "Can't handle type: nary_op_t";
-    return {};
-}
-void ir_visitor_t::_visit(const nary_op_t &obj) {
-    gpu_error_not_expected() << "Can't handle type: nary_op_t";
-}
-object_t ir_mutator_t::_mutate(const pexpr_t &obj) {
-    gpu_error_not_expected() << "Can't handle type: pexpr_t";
-    return {};
-}
-void ir_visitor_t::_visit(const pexpr_t &obj) {
-    gpu_error_not_expected() << "Can't handle type: pexpr_t";
 }
 
 } // namespace jit
--- a/src/gpu/intel/jit/ir/core.hpp
+++ b/src/gpu/intel/jit/ir/core.hpp
@@ -64,85 +64,22 @@
     HANDLE_IR_OBJECT(store_t) \
     HANDLE_IR_OBJECT(while_t)
 
-#define HANDLE_TRAVERSE_TARGETS() \
+#define HANDLE_CORE_IR_OBJECTS() \
     HANDLE_EXPR_IR_OBJECTS() \
-    HANDLE_STMT_IR_OBJECTS() \
-    HANDLE_IR_OBJECT(func_impl_t) \
-    HANDLE_IR_OBJECT(nary_op_t) \
-    HANDLE_IR_OBJECT(pexpr_t)
-
-#define HANDLE_ALL_IR_OBJECTS() \
-    HANDLE_EXPR_IR_OBJECTS() \
-    HANDLE_STMT_IR_OBJECTS() \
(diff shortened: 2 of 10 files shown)


src/gpu/intel/jit/ir/core.hpp
```diff
@@ -658,10 +595,72 @@ class object_t;
...
+
+    ir_type_id_t type_id = ir_type_id_t::undef;
+    const char *name = nullptr;
+    bool is_expr = false;
+    bool is_stmt = false;
+
+    bool operator==(const type_info_t &other) const {
+        if (type_id == ir_type_id_t::undef
+                && other.type_id == ir_type_id_t::undef)
+            return (name == other.name) || std::strcmp(name, other.name) == 0;
```
rjoursler: I don't think using a string comparison is a good idea here. Is there anything stopping us from just using [std::type_info](https://en.cppreference.com/w/cpp/types/type_info.html)?

src/gpu/intel/jit/ir/core.cpp
```diff
@@ -492,16 +491,14 @@ void ir_visitor_t::_visit(const for_t &obj) {
 }
 
 object_t ir_mutator_t::_mutate(const func_call_t &obj) {
-    auto func = mutate(obj.func);
```
rjoursler, line 495: The abstraction in the mutator and visitor look wrong to me. For a `func_call_t` object, I would expect that we should have the ability to inline more complex operation. For example during codegen, we should be able keep the code generation logic within the `func_impl_t` to avoid modifications to [this dispatch table](https://github.com/uxlfoundation/oneDNN/blob/d8d1ead904a1db0cf031bcfde88633dd6ae1ce58/src/gpu/intel/jit/codegen/codegen.cpp#L158-L207). I think we need to modify the functions from IR_DECLARE_TRAVERSERS for `func_call_t` so we can do something like:

```
    object_t _mutate(ir_mutator_t &mutator) const override {
        object_t(*m)(const func_call_t &) = mutator.try_get(*this);
        if(m) {
            return m(*this);
        }
       return mutator._mutate(*this); // Default case that only mutates arguments.
    }
```

echeresh: Sorry, I'm not sure I understand. Is the current code broken? The only thing that's changed in the PR related to this - `func_impl_t` objects can not be directly mutated anymore (however mutation can still be implemented through `func_call_t`) but we've never used this feature anyway...

rjoursler: You're right, I was overthinking the implication of this. I do think we need to add a virtual interface for `ir->ngen` so that we can do:
```
// ir_to_ngen_t visitor
void _visit(const func_call_t &obj) overrride {
    host_->comment(obj.line_str());
    auto scope = register_scope();
    obj.func.to_ngen(host_, eval(obj.args, scope));
}
```

echeresh: Yeah, it would be nice though `to_ngen()` in your snippet implies that IR functions would depend on codegen abstractions which is not good. But there still should be a way to rely on `func_impl_t` polymorphism here I guess.

src/gpu/intel/jit/ir/core.hpp
echeresh: Thanks, changed. I don't see overhead from `std::type_info`, and if no issues with compiler/OS - then we should be good to use it.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/ir/core.hpp
```diff
@@ -658,10 +594,70 @@ class object_t;
...
+struct type_info_t {
+    constexpr type_info_t(ir_type_id_t type_id, const std::type_info &info)
+        : type_id(type_id)
+        , info(&info)
+        , is_expr(jit::is_expr(type_id))
+        , is_stmt(jit::is_stmt(type_id)) {}
+
+    ir_type_id_t type_id = ir_type_id_t::undef;
+    const std::type_info *info = nullptr;
+    const char *name = nullptr;
```
rjoursler: Is this unused?

src/gpu/intel/jit/ir/core.hpp
```diff
@@ -658,10 +594,70 @@ class object_t;
 class ir_mutator_t;
 class ir_visitor_t;
 
+enum class ir_type_id_t : uint8_t {
```
rjoursler, line 600: Is this enum class necessary? It seems like it should be unnecessary if we are using std::type_info.

src/gpu/intel/jit/pass/simplify.cpp
```diff
@@ -725,20 +785,28 @@ class nary_op_visitor_t : public ir_visitor_t {
...
-    object_t _mutate(const nary_op_t &obj) override {
+    virtual object_t _mutate(const nary_op_t &obj) {
         auto args = mutate(obj.args);
         if (ir_utils::is_equal(args, obj.args)) return obj;
         return make_nary_op(obj.op_kind, args);
     }
 };
 
+inline object_t nary_op_t::_mutate(ir_mutator_t &mutator) const {
+    return static_cast<nary_op_mutator_t &>(mutator)._mutate(*this);
```
rjoursler: Use `utils::downcast` for the down conversion here and in other relevant places?

rjoursler (approved):

src/gpu/intel/jit/pass/simplify.cpp
echeresh: Thanks, changed.

src/gpu/intel/jit/ir/core.hpp
echeresh: I checked - without this enum I see ~5% creation time regression for convolution for 100 randomly sampled cases on average - tested a host with BMG. It's not big but I don't think it's worth it - we won't gain that much from simplifying this further.

src/gpu/intel/jit/ir/core.hpp
echeresh: Thanks, removed.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

PR #3505, opened 2025-06-30
cpu: remove zero-points macros

dzarukin:
This is a piece to support asynchronous runtime for Google in the future.

While the major problem lays in scales due to a local buffer of 16 elements that will be destroyed after task submission is done, it was decided to remove ZP macros as well since there's a potential misuse of their stack objects as well.


dzarukin:
make test
enable benchdnn_nightly
disable test_device_gpu

jondea:
Why can't the macro be rewritten to not allow misuse? Or replaced with a function which does the right thing?

dzarukin:
> Why can't the macro be rewritten to not allow misuse? Or replaced with a function which does the right thing?

The only reason it's a macro and not anything else is a stack variable that gets updated if the conditions for an attribute is met.
For zero-points it's possible to have a function that returns either a value or a pointer, but async runtime really doesn't like stack objects allocated before the parallel loop that's why they get eliminated.

The more problematic case is scales. You may find a buffer of 16 elements there which will be returned as a pointer. If it is wrapped into a function, it will be destroyed instantly once leaving the function which is the actual reason to have a macro - to have a pointer and that buffer on a stack alive. This buffer is exactly the reason of failures in asynchronous runtime which will be solved with scales elimination in the future PR.

Hope this answers your question.

dzarukin:
make test
disable test_device_gpu

jondea:
So in general, we are not allowed to pass stack objects from the execute function to the parallel lambda because of the way XLA handles the stack? Is there a way we can formalize this so that it doesn't happen again? Perhaps ban passing objects by reference into those lambdas?

dzarukin:
> So in general, we are not allowed to pass stack objects from the execute function to the parallel lambda because of the way XLA handles the stack? Is there a way we can formalize this so that it doesn't happen again? Perhaps ban passing objects by reference into those lambdas?

I'm not sure if "handling the stack" is the right term in here, but consider the lambda passed to parallel as a capsule (formal - "a task") that must contain everything by the time a thread starts working on it. Since it's asynchronous, stack objects may or may not be alive, thus, the library should prepare itself for the "may not" scenario.

Thus, POD-typed objects are handled by switching capturing from by reference to by copy in those lambdas. But non-POD-typed object is a different story. And arrays are the worst among of them, I'd say, thus, we are removing them from execute call, that's why this PR and some will follow by. What can be changed now and unrelated to asynchronous runtime itself will go first.

We are considering ways to formalize that but there's no paved path for this formalization yet. We will keep your team posted on that. And thank you for suggestions.

dzarukin:
Dropped a kernel debug commit as it causes issues on SYCL due to missing `get_binary` implementation.

dzarukin:
make test
disable test_device_gpu
enable benchdnn_nightly

densamoilov (approved):

PR #3509, opened 2025-07-01
generic: sycl: Skip f64 data type for generic backend

s-Nick:
# Description

This PR disable tests if data type is f64 since, due to current limitation, generic backend doesn't support f64 data type.  

# Checklist

## General

### DIFF
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -689,9 +689,23 @@ bool check_md_consistency_with_tag(
     return dnnl_memory_desc_equal(md_new_tag, md);
 }
 
+void skip_unimplemented_generic_f64(
+        const std::vector<dnnl_data_type_t> &v_dt, res_t *res) {
+    if (is_generic_gpu()) {
+        for (const auto &i_dt : v_dt) {
+            if (i_dt == dnnl_f64) {
+                res->state = SKIPPED;
+                res->reason = skip_reason::data_type_not_supported;
+                return;
+            }
+        }
+    }
+}
+
 void skip_unimplemented_data_type(
         const std::vector<dnnl_data_type_t> &v_dt, dir_t dir, res_t *res) {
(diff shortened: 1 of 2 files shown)


tests/benchdnn/conv/conv.cpp
```diff
@@ -382,6 +382,9 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
             prb->get_dt(SRC), prb->get_dt(DST));
     skip_unimplemented_binary_po(prb->attr, res);
     skip_unimplemented_prelu_po(prb->attr, res, dnnl_convolution);
+    skip_unimplemented_generic_f64({prb->get_dt(SRC), prb->get_dt(WEI),
+                                           prb->get_dt(BIA), prb->get_dt(DST)},
+            res);
```
s-Nick: I added this extra check to each primitive that, according to the doc, should support f64. It is not inside the `skip_unimplemented_data_type` function to keep the possibility of removing one check at the time once/if the f64 implementation for a given primitive will be available.

Rbiessy (approved):

tests/benchdnn/conv/conv.cpp
dzarukin: You don't have to remove it one at a time since the logic is built like this:
* Try create PD.
* If succeeded, go to filling memories, otherwise, check if unimplemented case is valid.
* If it's valid (skip_unimpl_functions), then report a skip, otherwise, report an error.

So there's no reason for a standalone function, please update the common one.

dzarukin (dismissed):

tests/benchdnn/conv/conv.cpp
s-Nick: Thank you for the explanation. I addressed your comments in last commit.

openmodelzooci:

vpirogov:
Closing as stale.

PR #3511, opened 2025-07-01
[GPU] Add switch to disable micro sdpa kernel for non-fused perf collection

h-sadia:
This PR is to address a need for SDPA performance testing. We want to collect non-fused vs fused performance collection data. Non-fused basically means running the graph primitive version on gpu, which requires the micro_sdpa kernel to be disabled.

### DIFF
--- a/src/gpu/gpu_sdpa_list.cpp
+++ b/src/gpu/gpu_sdpa_list.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


cmake/options.cmake
```diff
@@ -388,6 +388,8 @@ option(DNNL_ENABLE_STACK_CHECKER "enables stack checker that can be used to get
 option(DNNL_DISABLE_GPU_REF_KERNELS
         "builds oneDNN with only optimized kernels for GPU compute
         primitives" OFF)
+option(DNNL_DISABLE_MICRO_SDPA_KERNEL
+        "builds oneDNN without micro kernel for sdpa for perf collection" OFF)
```
dzarukin, line 392: Was an environment variable considered as an alternative version? It seems more flexible to me over a build option.

h-sadia: Adding a switch will be more streamlined with our 1perf tool because of the way the tool is built.

atkassen: Setting `_DNNL_GRAPH_SDPA_FORCE_PRIMITIVE=1` does what we want and requires much less work.

h-sadia: Can you please paste reference link here?

atkassen: It's undocumented, I'm sure, but you can find it [here](https://github.com/uxlfoundation/oneDNN/blob/dd1e6fe2a0b995fefec7f95aa3a507b747439e5a/src/graph/backend/dnnl/kernels/sdp.hpp#L117).

h-sadia: Right, in that case I will close this PR.

PR #3512, opened 2025-07-01
cpu: x64: enable fp8 conv with bf16 output

xuxinzen:
fp8 conv should support bf16 as output according to latest document.


xuxinzen:
make test

PR #3516, opened 2025-07-02
cpu: x64: fix correct when using int8 sat cvt

xuxinzen:
Fixes MFDNN-13883

### DIFF
--- a/src/cpu/x64/jit_generator.hpp
+++ b/src/cpu/x64/jit_generator.hpp
@@ -2081,10 +2081,16 @@ public:
 
     template <typename Vmm>
     void init_saturate_f32(Vmm vmm_lbound, Vmm vmm_ubound, Xbyak::Reg64 reg_tmp,
-            data_type_t idt, data_type_t odt, bool force_lbound = false) {
+            data_type_t idt, data_type_t odt, bool force_lbound = false,
+            const bool use_sat_cvt = false) {
+        // Only brgemm-based primitives support sat_cvt for now
+        // when converting data to int8 is needed later
+        // TODO: extend sat_cvt support to jit-based primitives
+        // and remove flag use_sat_cvt
         using namespace data_type;
         if (!((idt == f32) && utils::one_of(odt, u8, s8, s32))) return;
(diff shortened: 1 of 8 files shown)


xuxinzen:
make test
set test_scope=NIGHTLY

ankalinin (approved):

xuxinzen:
make test
set test_scope=NIGHTLY

tprimak (approved):

src/cpu/x64/jit_generator.hpp
```diff
@@ -2081,10 +2081,12 @@ class jit_generator_t : public Xbyak::MmapAllocator,
 
     template <typename Vmm>
     void init_saturate_f32(Vmm vmm_lbound, Vmm vmm_ubound, Xbyak::Reg64 reg_tmp,
-            data_type_t idt, data_type_t odt, bool force_lbound = false) {
+            data_type_t idt, data_type_t odt, bool force_lbound = false,
+            const bool use_sat_cvt = false) {
```
dzarukin, line 2085: I have a hard time to understand how the flag actually solves the issue and what the issue actually is. Neither the PR description, nor the tracker explains it. Could you, please, elaborate on the fix?

xuxinzen: After using saturating conversion instructions `VCVTPS2IBS` in `saturate_cvt_f32()`, the data is downconverted to 8-bit result at the lower 8-bit of the corresponding 32-bit element. And the upper 3 bytes are zeroed. So we need to use `vpmovusdb` instead of `vpmovsdb` to convert it to s8.
For now, only brgemm-based primitives cover the special handle of s8. To solve this failure for int8, we can either extend such support to jit-based primitives or add a flag to identify to when we can use sat_cvt instructions.

dzarukin: Thanks for the details. The issue is clear now, would be nice to have it in a comment for arguments explaining brgemm versus jit support and potential TODO to remove it as it's an interim solution.

PR #3518, opened 2025-07-02
jit: removes various ir limitations and fixes various bugs encountered in prototyping IR GEMM

rjoursler:


src/gpu/intel/jit/ir/core.cpp
```diff
@@ -277,7 +277,9 @@ type_t nary_op_type(op_kind_t op_kind, const std::vector<expr_t> &args) {
 }
 
 void ptr_t::normalize(expr_t &base, expr_t &off, op_kind_t op_kind) {
-    gpu_assert(base.type().is_ptr()) << "base is not a pointer: " << base;
+    if (!base.is<ptr_t>()) return;
```
echeresh: If we have subtraction, then `off` is not modified however should be `off` -> `-off` - this doesn't seem to be expected behavior.

src/gpu/intel/jit/ir/core.hpp
```diff
@@ -1152,6 +1159,10 @@ class binary_op_t : public expr_impl_t {
     expr_t b;
 
 private:
+    bool is_commutative() const {
+        return utils::one_of(op_kind, op_kind_t::_add, op_kind_t::_mul,
+                op_kind_t::_or, op_kind_t::_and);
```
echeresh: There is `is_commutative_op()`, please check:
https://github.com/uxlfoundation/oneDNN/blob/68f2f12d85c67ed27da700f4cb788a790437b463/src/gpu/intel/jit/ir/core.hpp#L1108

src/gpu/intel/jit/ir/message.hpp
```diff
@@ -117,9 +118,9 @@ struct block_2d_info_t {
     }
 
     // Encoded in header.
-    int surface_width = 0;
-    int surface_height = 0;
-    int surface_pitch = 0;
+    expr_t surface_width = {};
```
echeresh:
```suggestion
    expr_t surface_width;
```

src/gpu/intel/jit/ir/message.hpp
```diff
@@ -163,8 +164,9 @@ class send_t : public func_impl_t {
     }
 
     static func_t make_2d(const hw_t &hw, send_op_t op, const type_t &type,
-            int surface_width, int surface_height, int surface_pitch, int width,
-            int height, int count, bool vnni, bool transpose, bool zero_out,
+            expr_t surface_width, expr_t surface_height, expr_t surface_pitch,
```
echeresh, line 167:
```suggestion
            const expr_t &surface_width, const expr_t &surface_height, const expr_t &surface_pitch,
```

src/gpu/intel/jit/pass/simplify.cpp
```diff
@@ -547,10 +547,29 @@ expr_t simplify_rewrite_with_ternary(const expr_t &e, bool recursive) {
...
+            bool b_le_0 = cset.can_prove(b <= 0);
+
+            if (op_kind == op_kind_t::_ne && b_lt_0) return expr_t(true);
+            if (op_kind == op_kind_t::_ge && b_le_0) return expr_t(true);
+            if (op_kind == op_kind_t::_gt && b_lt_0) return expr_t(true);
+
+            if (op_kind == op_kind_t::_eq && b_lt_0) return expr_t(false);
+            if (op_kind == op_kind_t::_le && b_lt_0) return expr_t(false);
+            if (op_kind == op_kind_t::_lt && b_le_0) return expr_t(false);
+        }
```
echeresh: Do you have a branch to check the original scenario and test case?

I would try to avoid relying on mixed signed/unsigned combinations in comparison, e.g. see this C++ example which would contradict the new rules:

```$ cat 1.cpp
#include <cstdint>
#include <iostream>
int main() {
    uint32_t x = 1;
    int32_t y = -1;
    std::cout << (x < y) << std::endl; // line 570 would simplify that to false
}
$ g++ 1.cpp; ./a.out
1
```

src/gpu/intel/jit/ir/core.cpp
echeresh: Another comment: I see you have a bunch of changes to handle regular variables where pointers are expected.

Can we mark these changes as temporary? I think we should be able to switch to the new IR types such as `assign_t`, `var_ref_t`, etc and leave `ptr_t`, `load_t` and `store_t` working only with pointers.

src/gpu/intel/jit/pass/simplify.cpp
rjoursler: The branch `rjoursle/gemm_repro` can reproduce this issue. The specific issue I was attempting to resolve here can be reproduced on Xe2 with:
```
 DNNL_VERBOSE=debuginfo=200 enable_gemm_ir=1 ~/rjoursle/dnnl/build/tests/benchdnn/benchdnn --matmul --engine=gpu --dt=bf16:bf16:f32 --stag=ab --wtag=ab --dtag=ab 32x32:32x32
 ...
 Assertion to_cpp<int64_t>(off) % type.scalar().size() == 0 failed at /home/gta/rjoursle/dnnl/src/gpu/intel/jit/ir/core.cpp:317
Incompatible offset: 20
terminate called after throwing an instance of 'std::runtime_error'
  what():  Assertion to_cpp<int64_t>(off) % type.scalar().size() == 0 failed at /home/gta/rjoursle/dnnl/src/gpu/intel/jit/ir/core.cpp:317
Incompatible offset: 20
Aborted (core dumped)
```
The core issue is that when attempting to construct a message header we have the expression for the x argument
```
(k_idx.u32 + 96).s64
```

rjoursler: The issue I was trying to avoid can be reproduced from `rjoursle/gemm_repro` with the command:
```
DNNL_VERBOSE=debuginfo=200 gemm_ir_specialize=1 enable_gemm_ir=1 ~/rjoursle/dnnl/build/tests/benchdnn/benchdnn --matmul --engine=gpu --dt=bf16:bf16:f32 --stag=ab --wtag=ab --dtag=ab 32x32:32x32
```
In particular, we get the following simplification
```
while ((((m_idx <= (m + -16)) && (n_idx <= (n + -4))) && (k_idx <= (k + -128)))) {
-> 
 while ((((m_idx <= 16) && (n_idx <= 28)) && (k_idx <= -32))) {
```
where as I expect this while loop to be completely eliminated for this simplification. I can go ahead and drop this change though. To some extent, this is workaround for a lack simplification support for mutable variables.

src/gpu/intel/jit/ir/core.cpp
rjoursler: > Can we mark these changes as temporary?

I don't really see any benefit to this. For the most part, I am just removing assertions and everything works as expected.

src/gpu/intel/jit/ir/message.hpp
rjoursler: I switched to using `std::move` in the assignment instead, as the pattern of passing by value and moving generally enables more optimization options for the C++ compiler with r_values, while being equivalent to the above for l_values.

src/gpu/intel/jit/pass/simplify.cpp
echeresh: Thanks, yes, looks like it's similar behavior to my snippet: `unsigned < signed` comparison. I see two separate issues:

- This particular comparison: to make it compliant with type promotion rules we need to convert it to `unsigned < [positive] signed` to ensure it's valid or switch to signed integers
- Even when mixed-sign comparison is fixed, extra simplification is needed
    - However, I would just handle it manually in `is_inbounds()` if the RHS part is constant. It's simpler, we can skip it in place, without relying on simplification

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/pass/simplify.cpp
rjoursler: Agreed. I have extended `range_simplifier_t` so that we get the behavior I expect, but without the reliance on unsigned/signed comparisons.

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/ir/core.cpp
echeresh: Well, this is probably not a big deal but this adds ambiguity. Before this change we have a single usage scenario when load/store/ptr always take a pointer type. With that change we extend the set of valid states in IR to consider and handle.

As an example: it seems after this change it will be possible to modify a `let_t` variable via `store_t` which is undesired and not expected. Of course, it can be fixed by adding some checks but we could avoid dealing with that by sticking to the original behavior.

And, since the end goal is to rely on new IR/DSL abstractions, I'm not sure I understand the potential benefit of extending the semantics of the existing IR types.

rjoursler: > As an example: it seems after this change it will be possible to modify a let_t variable via store_t which is undesired and not expected.

Just for reference, my intention to handle this was to just add an assertion that the base type is either a byte pointer or is mutable once we have the type attributes patch merged. This will have a similar effect  as the current requirements since types are immutable by default. I am fine with reverting this if that seems insufficient though.

echeresh: To put it simply, we have two options:

1. Implementing the IR GEMM on top of the new IR/DSL types
2. Having something intermediate, based on the extensions from this PR

I think (1) is the final goal but it seems (2) would allow to make more incremental updates - since you already have a prototype based on that. So let's move forward with (2) but I would keep (1) in mind as the final goal.

echeresh (approved):

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/conv/pipeline.cpp
```diff
@@ -314,8 +314,8 @@ class compute_step_t {
         expr_t preload_var;
         expr_t mul_var;
 
-        bool is_preload() const { return !preload_var.is_empty(); }
-        bool is_mul() const { return !mul_var.is_empty(); }
+        bool is_preload() const { return bool(preload_var); }
+        bool is_mul() const { return bool(mul_var); }
```
kealan-barbieri, line 318: will coverity complain about any of the implicit conversions elsewhere?

kealan-barbieri (approved):

src/gpu/intel/jit/conv/pipeline.cpp
rjoursler: I don't believe so. The implicit conversions in those cases are well defined by the standard as contextual conversions. From [here](https://en.cppreference.com/w/cpp/language/implicit_conversion.html):

> In the following contexts, the type bool is expected and the implicit conversion is performed if the declaration bool t(e); is well-formed (that is, an explicit conversion function such as explicit T::operator bool() const; is considered). Such expression e is said to be contextually converted to bool.

> * the controlling expression of if, while, for;

> * the operands of the built-in logical operators !, && and ||;
> * the first operand of the conditional operator ?:;
> ...

src/gpu/intel/jit/ir/core.cpp
```diff
@@ -277,7 +277,7 @@ type_t nary_op_type(op_kind_t op_kind, const std::vector<expr_t> &args) {
 }
 
 void ptr_t::normalize(expr_t &base, expr_t &off, op_kind_t op_kind) {
-    gpu_assert(base.type().is_ptr()) << "base is not a pointer: " << base;
+    // Normalize (base + off1) + off2 -> base + (off1 + off2)
```
Simonsays095, line 280: If the `ptr_t` doesn't have a pointer base, what other valid base type could it have? It's unclear to me what this change enables us to do now.

src/gpu/intel/jit/ir/core.hpp
```diff
@@ -774,6 +774,8 @@ class object_t {
 
     bool is_empty() const { return !impl_; }
 
+    explicit operator bool() const { return !is_empty(); }
```
Simonsays095, line 777: Is it possible that this will cause confusion with immediates or provable statements? I.E. should we make a distinction between `object_t::is_empty`, `object_t::is_false`, and `object_t::is_null`?

src/gpu/intel/jit/ir/message.hpp
```diff
@@ -117,9 +118,9 @@ struct block_2d_info_t {
     }
 
     // Encoded in header.
-    int surface_width = 0;
-    int surface_height = 0;
-    int surface_pitch = 0;
+    expr_t surface_width;
+    expr_t surface_height;
+    expr_t surface_pitch;
```
Simonsays095, line 123: Is the idea here to allow these surface parameters to be created as part of the JIT compilation (i.e. stored in registers and mutated at runtime)? If so, in what cases do we need this?

Simonsays095 (commented):
Forgive my lack of understanding of IR concepts - hopefully my questions aren't too naive :)

src/gpu/intel/jit/ir/core.cpp
rjoursler: This effectively enables generating pointers from value types. In OpenCL syntax, this would be like:
```C
float x;
float4 y;

&x  // ptr_t(x, 0)
&y[2] // ptr_t(y, 2)
```

src/gpu/intel/jit/ir/core.hpp
rjoursler: No, the `object_t` type is effectively an intrusive pointer. Just by the design, `object_t::is_null` doesn't exist and is completely covered by `object_t::is_empty`. Since this is a pointer object, it generally makes sense to support common pointer idioms like:

```
if(ptr)
   ...
```
In regards to `object_t::is_false`, this information would need to be explicitly evaluated via `to_cpp<bool>(obj)` (after fist verifying the object is a compile time constant.

src/gpu/intel/jit/ir/message.hpp
rjoursler: This is needed for reusable kernels. Say we are targeting the GEMM A matrix with dimension `m` and `k`,  layout `ab` and  stride `lda`. When performing 2D loads from this matrix, we would set `surface_width = k`, `surface_height = m`, and `surface_pitch = lda` in the message header. Previously, this interface was not used for reusable kernels, and so it only supported hard-coded surface values.

src/gpu/intel/jit/ir/core.cpp
Simonsays095: In OpenCL syntax, you use the address-of operator to convert a value into an address/pointer - instead of letting `ptr_t` hold a value (and implicitly understanding this as a pointer to the value), could we explicitly construct a `ptr_t` from a value using a special constructor or factory (which would store a pointer to the value in the `ptr_t`)?

rjoursler: That is effectively how this works already.  Construction of `ptr_t` is done via [ptr_t::make](https://github.com/uxlfoundation/oneDNN/blob/7ec980b49b6e367d3be6c81508d265dd643c62e4/src/gpu/intel/jit/ir/core.hpp#L1501-L1503). For the example of `&y[2]`, we would call `ptr_t::make(y, 2)`. The assertions removed by this PR are just removing the current restriction that `base`, which is `y` in this example, is already a pointer object.

rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

PR #3390, opened 2025-06-03
xe: jit: gemm: downstream gemmstone

petercad:
Omnibus pulldown from upstream gemmstone repository. Highlights include:

* Long-overdue refactoring of register layouts into their own class, `RegisterLayout`
* Numerous copy planner optimizations/fixes:
  - New dedicated upconversions to bf16 that save a few cycles/register
  - More de-interleaving prior to complex conversion sequences
  - Fix some corner cases in hf8/e2m1/e3m0 downconversions where rounding was happening in the wrong direction
* Expanded support for C repacking ("dynamic quantization" scenarios). Now supports FMA and dot kernels.
* nf4 upconversion
* Bug fixes/cleanup in A/B dequantization
* Resolved some register spills in the catalog

PR #3519, opened 2025-07-02
xe: jit: gemm: reuse f16 compressed weights strategies for bf16

petercad:
Reuses f16 compressed weights strategies for bf16 (relies on #3390). With #3390, bf16 decompression register space requirements are now similar to f16, so this PR eliminates the dedicated bf16 strategies and updates the kernel selector to fall back to f16 strategies if a bf16 strategy is not found.

CI perf testing shows up to 5x improvement for some layers, and good E2E improvement on many models.

Splitting this PR from #3390 to allow time to fix some functional issues.


kealan-barbieri (approved):

Simonsays095:
This should be able to replace BBS (i.e. `--dt=bf16:bf16:f32`) kernels as well, right? Then the only bf16 kernels that remain are `[SB][SB]S` to cover (I think) `--dt=f32 --attr-fpmath=bf16` cases.

petercad:
@Simonsays095 Yes, that's right. I didn't do that in this PR because HHS and BBS have different power characteristics on PVC and I wasn't sure if removing BBS strategies would cause regressions as a result. But we should try it.

petercad:
Closing in favor of #3693.

PR #3517, opened 2025-07-02
[GPU] Fixup gemm attr

kealan-barbieri:
# Description

- Separate handling for edge case dim counts with non-trivial grouped attr.
- Fix scale batch offset computation
- Additional test cases
- Fix type initialization for zp in cases with `swap_ab`

Fixes # [MFDNN-13688](https://jira.devtools.intel.com/browse/MFDNN-13688)

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

### Bug fixes

- [x] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [x] Have you added relevant regression tests?

PR #3521, opened 2025-07-03
xe: gemm: refactor post ops and quantization

Simonsays095:
Post-ops for A/B/C each have their own handling code, potentially leading to bug escapes. This PR refactors post-op initialization to simplify the related logic, and so that A/B/C scales are handled in one block and can easily be extended for future features. Additionally refactors a few functions to be defined where they're used and to reduce type conversions.

The first commit is from #3498 and will be dropped upon merging (that PR, then this one).

### DIFF
--- a/src/gpu/intel/jit/gemm/gen_gemm.cpp
+++ b/src/gpu/intel/jit/gemm/gen_gemm.cpp
@@ -34,6 +34,16 @@ namespace gpu {
 namespace intel {
 namespace jit {
 
+namespace {
+dim_t quant_entry_group_prod(const quant_entry_t &attr) {
+    dim_t ret = 1;
+    if (attr.has_default_groups()) return ret;
+    for (int i = 0; i < 2; ++i)
+        ret *= attr.get_group(i);
+    return ret;
+}
+} // anonymous namespace
+
 status_t gen_gemm_t::launch_nocopy(const gemm_exec_ctx_t &ctx,
         compute::compute_stream_t *compute_stream, zero_pool_t *zero_pool,
         const memory_storage_t &a, const memory_storage_t &b,
@@ -145,14 +155,14 @@ status_t gen_gemm_t::launch_nocopy(const gemm_exec_ctx_t &ctx,
             if (problem->asPtrDims > 2) {
                 dim_t scale_stride = pd()->stride_scale(i, DNNL_ARG_A);
                 auto scale_stride_a = int32_t(scale_stride
-                        / pd()->quant_entry_group_prod(
+                        / quant_entry_group_prod(
                                 pd()->attr()->scales_.get(DNNL_ARG_A)));
                 arg_list.set(argn++, scale_stride_a);
             }
             if (problem->bsPtrDims > 2) {
                 dim_t scale_stride = pd()->stride_scale(i, DNNL_ARG_B);
                 auto scale_stride_b = int32_t(scale_stride
-                        / pd()->quant_entry_group_prod(
+                        / quant_entry_group_prod(
                                 pd()->attr()->scales_.get(DNNL_ARG_B)));
                 arg_list.set(argn++, scale_stride_b);
             }
--- a/src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
+++ b/src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
@@ -24,6 +24,8 @@ namespace intel {
 namespace jit {
 
 namespace {
+// convert a quant_entry_t and the base memory_desc_t into the dims_t for
+// the required quantization md.
 void quant_dims(
         const memory_desc_t &md, const quant_entry_t &entry, dims_t &out) {
     auto mask = entry.get_mask();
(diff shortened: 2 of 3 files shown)


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
```diff
@@ -23,6 +25,44 @@ namespace gpu {
...
+    bool full_dim = false;
+    for (int i = 0; i < md.ndims; ++i) {
+        if (qdims[i] > 1) {
+            count++;
+            full_dim = (qdims[i] == md.dims[i]);
+        }
+    }
+
+    // gemmstone doesn't support 1D grouped scales, these have to be sent as 2D
+    if (count == 1 && !full_dim) count = 2;
```
kealan-barbieri: There are nd cases with per-tensor scales affected by this as well, this probably needs to be something like `if(count < ndims && !full_dim) ++count`.

Simonsays095: I think this handles it as-is. Let me know if these aren't the cases you're talking about
> 0:PASSED (999 ms) __REPRO: --matmul --engine=gpu --allow-enum-tags-only=false --check-ref-impl=true --dt=s8:u4:f16 --stag=ab --wtag=ba --dtag=ab --bia-dt=f16 --bia_mask=4 --attr-scales=src:per_ocic:f16:1x128 1x2560:2560x10240
0:PASSED (891 ms) __REPRO: --matmul --engine=gpu --allow-enum-tags-only=false --check-ref-impl=true --dt=s8:u4:f16 --stag=abc --wtag=acb --dtag=abc --bia-dt=f16 --bia_mask=4 --attr-scales=src:per_ocic:f16:1x128 2x1x2560:1x2560x10240
0:PASSED (1143 ms) __REPRO: --matmul --engine=gpu --allow-enum-tags-only=false --check-ref-impl=true --dt=s8:u4:f16 --stag=abc --wtag=acb --dtag=abc --bia-dt=f16 --bia_mask=4 --attr-scales=src:per_ocic:f16:1x128 2x1x2560:2x2560x10240

Simonsays095: Actually I just found one that fails. Thanks, I'll fix the check:
> 0:FAILED (errors:10233 total:20480) (1159 ms) __REPRO: --matmul --engine=gpu --allow-enum-tags-only=false --check-ref-impl=true --dt=s8:u4:f16 --stag=abc --wtag=acb --dtag=abc --bia-dt=f16 --bia_mask=4 --attr-scales=src:per_tensor:f16:1x128 2x1x2560:2x2560x10240

kealan-barbieri: if you can please also check the cases added in https://github.com/uxlfoundation/oneDNN/pull/3517

Simonsays095: Main fails 4 tests, this PR only fails 1, and the remaining failure is fixed by #3517. Thanks Kealan!

src/gpu/intel/jit/gemm/gen_gemm.cpp
```diff
@@ -34,6 +34,16 @@ namespace gpu {
 namespace intel {
 namespace jit {
 
+namespace {
```
h-sadia, line 37: is there any other way of handling this rather than creating a standalone namespace?

Simonsays095: The anonymous namespace is to give the function internal linkage (similar to declaring the function as static). I just prefer this method over static functions, as it's more flexible. Right now, this isn't strictly necessary - it's just good practice.

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
```diff
@@ -23,6 +25,51 @@ namespace gpu {
 namespace intel {
 namespace jit {
 
+namespace {
```
h-sadia, line 28: Ditto for this, the code looks rather odd with a  standalone namespace for just a function. Are there any other possible ideas of converting this? @rjoursler @echeresh - can you please share thoughts? Maybe, more so for my clarity.

kealan-barbieri (approved):

src/gpu/intel/jit/gemm/jit_gemm_pd.cpp
Simonsays095: Here's a SO post explaining its use compared to declaring functions as static:
https://stackoverflow.com/questions/154469/why-should-you-prefer-unnamed-namespaces-over-static-functions

rjoursler: @h-sadia, using anonymous namespaces like this is fairly standard practice in C++. The main point is to prevent unexpected name clashes between unrelated files.

h-sadia (approved):

PR #3527, opened 2025-07-04
graph: interface: combine fpmath into compiled partition cache key

xiang1guo:
Found in MFDNN-13693, not about solving the MFDNN-13693 itself, but a potential issue.

Different fpmath mode may has different kernel implementation, shouldn't cache hit if fpmath is different. Also following the implementation of primitive hashing([here](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/primitive_hashing.cpp#L226))

### DIFF
--- a/tests/gtests/graph/unit/interface/test_compiled_partition.cpp
+++ b/tests/gtests/graph/unit/interface/test_compiled_partition.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/gtests/graph/unit/interface/test_compiled_partition.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2021-2024 Intel Corporation
+* Copyright 2021-2025 Intel Corporation
```
TaoLv, line 2: I feel we will need to test case to guarantee that cp cache will miss when fp math mode is changed.

xiang1guo: I added a new case to test different fpmath mode.

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3528, opened 2025-07-04
api: graph: fix scalar tensor creation and usage

TaoLv:
- Extract the new API changes from #3454 
    - Create scalar tensor with `tensor::make_scalar_tensor(logical_tensor, void*)` instead of the original constructor.
    - `scalar_tensor.get_data_handle()` returns nullptr.
    - `scalar_tensor.set_data_handle()` raises error.
- Benchdnn graph driver is changed accordingly to use the new method to create scalar tensor.
- Fix DNNL backend for creating `dnnl::memory` from a scalar tensor without engine.
- Add an example for SDPA with bottom-right causal mask which requires scalar tensor to define the patter. It's also a new feature for oneDNN v3.9, so I think we may need to backport this one as well.

### DIFF
--- a/examples/graph/sdpa_bottom_right_causal_mask.cpp
+++ b/examples/graph/sdpa_bottom_right_causal_mask.cpp
@@ -1,0 +1,363 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cassert>
+#include <chrono>
+#include <iomanip>
+#include <iostream>
+#include <memory>
+#include <random>
+#include <string>
+#include <vector>
+
+#include "oneapi/dnnl/dnnl.hpp"
+#include "oneapi/dnnl/dnnl_graph.hpp"
+
+#include "graph_example_utils.hpp"
+
+using namespace dnnl;
+
+using namespace dnnl::graph;
+using layout_type = logical_tensor::layout_type;
+using property_type = logical_tensor::property_type;
+using data_type = logical_tensor::data_type;
+using dim = logical_tensor::dim;
+using dims = logical_tensor::dims;
+
+struct sdpa_dims_t {
+    dim mb;
+    dim seq_len;
+    dim head_num;
+    dim head_size;
+    dim query_num;
+};
+
+static const int min_runs = 4;
+
+// this is changed from the fill_random() function in matmul_perf.cpp.
+void fill_random(std::vector<float> &out) {
+    static std::vector<float> random_data_f;
+    constexpr size_t nrand = 1037;
+
+    if (random_data_f.empty()) {
+        std::mt19937 generator;
+        std::uniform_real_distribution<float> dist_f(-1.0f, 1.0f);
+
+        random_data_f.resize(nrand);
+        for (auto &d : random_data_f)
+            d = dist_f(generator);
+    }
+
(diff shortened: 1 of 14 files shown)


include/oneapi/dnnl/dnnl_graph.h
```diff
@@ -193,6 +193,17 @@ dnnl_status_t DNNL_API dnnl_graph_tensor_create(dnnl_graph_tensor_t *tensor,
...
         void *handle);
 
+/// Creates a scalar tensor with logical tensor and scalar data handle.
+///
+/// @param tensor Output scalar tensor.
+/// @param logical_tensor Description for this tensor.
+/// @param handle Handle of the memory buffer to use as an underlying storage.
+/// @returns #dnnl_success on success or a status describing the error
+///     otherwise.
+dnnl_status_t DNNL_API dnnl_graph_tensor_create_host_scalar(
```
mgouicem: maybe remove mention of host from name to align with C++ API, and the fact that it is no more device related ?

include/oneapi/dnnl/dnnl_graph.hpp
```diff
@@ -607,6 +609,22 @@ class tensor : public tensor_handle {
...
         : tensor(lt, aengine, DNNL_MEMORY_ALLOCATE) {}
 
+    /// Creates a tensor object for host-side scalar value. The data type contained
+    /// in the logical tensor parameter will be used to interpret the scalar
+    /// pointer. The property type in the logical tensor must be `host_scalar`.
+    ///
+    /// @param lt The logical tensor describing the host scalar
+    /// @param scalar The pointer to scalar value
+    /// @returns Created tensor object
+    static tensor make_scalar_tensor(const logical_tensor &lt, void *scalar) {
```
mgouicem, line 619: (open) is logical tensor even needed here? Could we just take `dt`?
I guess we would either copy-initialize the logical_tensor within tensor with the logical tensor passed, or just initialize it with dt and host_scalar property directly.

include/oneapi/dnnl/dnnl_graph.h
TaoLv: Thank you, Mourad. I remove `host` from the name of this API.

include/oneapi/dnnl/dnnl_graph.hpp
TaoLv: Good question, Mourad. I chose to keep logical tensor here as we need not only dt and property, but also the ID in the logical tensor for internal topology representation.

mgouicem (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

examples/graph/sdpa_bottom_right_causal_mask.cpp
```diff
@@ -0,0 +1,360 @@
+
+/*******************************************************************************
```
ElaineBao, line 1: Do you want to update the sdpa doc to include this example?

examples/graph/sdpa_bottom_right_causal_mask.cpp
```diff
@@ -0,0 +1,360 @@
...
+    gen_index_col.add_outputs({index_col});
+
+    auto mask_ge_out = logical_tensor(
+            id++, data_type::boolean, score_sz, layout_type::strided);
+    auto mask_ge = op(id++, op::kind::GreaterEqual, "mask_ge");
+    mask_ge.add_inputs({mask_sub_out, index_col});
+    mask_ge.add_outputs({mask_ge_out});
+
+    auto neg_inf
+            = logical_tensor(id++, dt_inter, scale_sz, layout_type::strided);
```
ElaineBao, line 179: Also use host_scalar for negative infinity?

examples/graph/sdpa_bottom_right_causal_mask.cpp
TaoLv: Added.

examples/graph/sdpa_bottom_right_causal_mask.cpp
TaoLv: No. Let's add negative infinity support along with the scalar attention.

ElaineBao (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3529, opened 2025-07-04
cpu: aarch64: perf: increase 3d matmul shape for perf tests

Ryo-not-rio:
Increasing the size of 3d shapes for the nightly matmul performance tests as it is causing a false positive in the ci


Sqvid (approved):
LGTM

vpirogov (approved):
@Ryo-not-rio, I don't remember what rules benchdnn uses for relative path resolution, so it might be worth checking that it picks up `shapes_3d_perf` correctly.

PR #3530, opened 2025-07-04
ci: aarch64: Extend performance aarch64 workflow with user-triggered comparison mode

gausah-arm:
This PR enhances the aarch64 performance CI workflow with the ability to run user-defined comparison tests in addition to the existing nightly benchmarks.

Summary of changes:
Nightly Mode (default):

- When no input is provided, the workflow runs as it always has: running nightly performance tests using the latest known oneDNN and ACL baseline versions.
- Uses preconfigured batch files and benchmarks via bench_nightly_performance.sh.

Comparison Mode (opt-in via inputs):

- If a user supplies benchdnn_command and optional oneDNN/ACL hashes through workflow dispatch inputs:
- Clones and builds both baseline and new versions of oneDNN and ACL.
- Runs the user-provided benchdnn_command on both builds.
- Sets the user-provided number of threads.
- Collects and compares execution time and creation time (ctime) from multiple repetitions.
- Outputs CSV-format results and highlights regressions (with p-values) using benchdnn_comparison.py.

Optimizations:
Adds support for running each test multiple times (via REPS) to enable robust statistical checks.

Example benchdnn input to trigger comparison: 
--conv --mode=P --dir=FWD_D --dt=f32 --alg=direct mb1_ic64oc256_ih200oh200kh1sh1dh0ph0_iw267ow267kw1sw1dw0pw0

Outputs:

- base.txt and new.txt contains perf results with format:   Output template: %prb%,%-time%,%-ctime%
- Speed comparisons and regression checks are printed directly in GitHub Actions logs.


.github/automation/performance/benchdnn_comparison.py
```diff
@@ -104,41 +105,72 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
-            exec_failures.append(
-                f"{prb} exec: {r1_med_exec:.3g} → {r2_med_exec:.3g} "
-                f"(p={exec_ttest.pvalue:.3g})"
-            )
-        if ctime_regressed:
-            ctime_failures.append(
-                f"{prb} ctime: {r1_med_ctime:.3g} → {r2_med_ctime:.3g}"
-                f"(p={ctime_ttest.pvalue:.3g})"
+        if check:
+
```
Ryo-not-rio, line 124: nit: unnecessary space

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -104,41 +105,72 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        print_to_github_out(f"pass={not exec_failures}")
+
+        message = ""
+        if ctime_failures:
+            message += (
+                "\n----The following ctime regression tests failed:----\n"
+                + "\n".join(ctime_failures)
+                + "\n"
+            )
+            # print(message)
```
Ryo-not-rio: remove

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -104,41 +105,72 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+        description="Compare two benchdnn result files."
+    )
+    parser.add_argument("file1", help="Path to baseline result file")
+    parser.add_argument("file2", help="Path to new result file")
+    parser.add_argument(
+        "--check", action="store_true", help="Enable regression checks"
+    )
+    args = parser.parse_args()
+
+    compare_two_benchdnn(args.file1, args.file2, check=args.check)
```
Ryo-not-rio: missing newline at the end of file

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -68,7 +69,7 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
         r2_exec[key].append(float(exec_time))
         r2_ctime[key].append(float(ctime))
 
-    exec_failures, ctime_failures = [], []
+    exec_failures, ctime_failures, comparison_rows = [], [], []
```
Ryo-not-rio: no need to have a list for the comparisons, we can just print them as we go

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -104,41 +105,72 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+                    f"(p={exec_ttest.pvalue:.3g})"
+                )
+
+            if ctime_regressed:
+                ctime_failures.append(
+                    f"ctime: {r1_med_ctime:.3g} → {r2_med_ctime:.3g}"
+                    f"(p={ctime_ttest.pvalue:.3g})"
+                )
+        else:
+            comparison_rows.append(
```
Ryo-not-rio: instead of adding to a list, let's just print it here for simplicity

.github/workflows/performance-aarch64.yml
```diff
@@ -84,16 +116,57 @@ jobs:
...
+          ACL_ROOT_DIR: ${{ github.workspace }}/ComputeLibrary_New
+          ACL_VERSION: ${{ inputs.acl_new_hash }}
+          ACL_BUILD_TYPE: ${{ matrix.config.build }}
+          ACL_THREADING: ${{ matrix.config.threading }}
+
+      - name: Clone and Build ACL (nightly fallback)
+        if: ${{ inputs.acl_new_hash == '' }}
+        run: |
+            ACL_ACTION=clone ${{ github.workspace }}/oneDNN/.github/automation/aarch64/build_acl.sh
+            ACL_ACTION=build ${{ github.workspace }}/oneDNN/.github/automation/aarch64/build_acl.sh
```
Ryo-not-rio: Let's reuse the "Build ACL cache" workflow like we do with the current nightly so we're not re-building ACL every time. This workflow can be found in `aarch64-acl.yml`

.github/workflows/performance-aarch64.yml
```diff
@@ -125,17 +198,42 @@ jobs:
         env:
           ONEDNN_ACTION: build
 
-      - name: Checkout oneDNN base
+      - name: Checkout oneDNN base
```
Ryo-not-rio: remove extra spaces

.github/workflows/performance-aarch64.yml
```diff
@@ -104,13 +177,13 @@ jobs:
         uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
         with:
           key: ${{ steps.get_system_name.outputs.SystemName }}-acl-${{ matrix.config.toolset }}-${{ matrix.config.build }}-${{ steps.get_acl_commit_hash.outputs.ACLCommitHash }}
-          path: ${{ github.workspace }}/ComputeLibrary/build
+          path: ${{ env.ACL_CACHE_PATH }}
 
-      - name: Configure oneDNN
+      - name: Configure oneDNN  (comparison or nightly)
```
Ryo-not-rio: How come we have "Build oneDNN", "Build oneDNN new" and "Build oneDNN base"?

.github/workflows/performance-aarch64.yml
```diff
@@ -155,12 +253,70 @@ jobs:
...
+          echo "" > new.txt
+          for i in $(seq 1 "$REPS"); do
+            echo "Running new iteration $i..."
+            OMP_NUM_THREADS=${{ inputs.num_threads || 16 }} "$NEW_BINARY" "${FINAL_ARGS[@]}" >> new.txt
+          done
+
+          echo "base.txt"
+          cat base.txt || echo "base.txt is empty"
+          echo "new.txt"
+          cat new.txt || echo "new.txt is empty"
```
Ryo-not-rio: Let's put this in a bash script. it would be ideal if we can reuse the bash script for nightly, precommit and comparison pipelines

.github/workflows/performance-aarch64.yml
```diff
@@ -155,12 +253,70 @@ jobs:
...
+
+      - name: Print speed comparisons
+        if: ${{ inputs.benchdnn_command != '' }}
+        run: |
+          echo "Speed comparison output (CSV format):"
+          python3 ${{ github.workspace }}/oneDNN/.github/automation/performance/benchdnn_comparison.py base.txt new.txt
+
+      - name: Check for regressions
+        if: ${{ inputs.benchdnn_command != '' }}
```
Ryo-not-rio: This should be `inputs.benchdnn_command == ''`

.github/workflows/performance-aarch64.yml
```diff
@@ -169,4 +325,4 @@ jobs:
     name: "Nightly Performance AArch64"
     steps:
       - name: Print success
-        run: echo Success
+        run: echo Success
```
Ryo-not-rio: missing new line at EOF

.github/workflows/performance-aarch64.yml
```diff
@@ -155,12 +253,70 @@ jobs:
...
+
+          echo "base.txt"
+          cat base.txt || echo "base.txt is empty"
+          echo "new.txt"
+          cat new.txt || echo "new.txt is empty"
+
+      - name: Print speed comparisons
+        if: ${{ inputs.benchdnn_command != '' }}
+        run: |
+          echo "Speed comparison output (CSV format):"
```
Ryo-not-rio: nit: no need for this echo, let's not clutter the output too much

.github/workflows/performance-aarch64.yml
```diff
@@ -155,12 +253,70 @@ jobs:
...
+          python3 ${{ github.workspace }}/oneDNN/.github/automation/performance/benchdnn_comparison.py base.txt new.txt
+
+      - name: Check for regressions
+        if: ${{ inputs.benchdnn_command != '' }}
+        run: |
+          echo "Checking for regressions..."
```
Ryo-not-rio: ditto

.github/workflows/performance-aarch64.yml
```diff
@@ -155,12 +253,70 @@ jobs:
...
+          python3 ${{ github.workspace }}/oneDNN/.github/automation/performance/benchdnn_comparison.py base.txt new.txt --check | tee regression.log
+          cat regression.log
```
Ryo-not-rio: Instead of outputting to a file and printing the contents of file, we can just run the python command and let it print to stdout

Ryo-not-rio (changes requested):
Thanks for this new pipeline, I've left a few comments, mostly for tidiness

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -92,7 +98,7 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
             )
             continue
 
-        # A test fails if execution time:
+        # A test fails if either execution time or creation time:
```
Ryo-not-rio: This should just be execution time

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -103,42 +109,62 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+                    f"{prb} exec: {r1_med_exec:.3g} → {r2_med_exec:.3g} "
+                    f"(p={exec_ttest.pvalue:.3g})"
+                )
+
+            if ctime_regressed:
+                ctime_failures.append(
+                    f"ctime: {r1_med_ctime:.3g} → {r2_med_ctime:.3g}"
+                    f"(p={ctime_ttest.pvalue:.3g})"
+                )
+        if not check:
```
Ryo-not-rio, line 131: else

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -103,42 +109,62 @@ def compare_two_benchdnn(file1, file2, tolerance=0.05):
...
+    )
+    parser.add_argument("file1", help="Path to baseline result file")
+    parser.add_argument("file2", help="Path to new result file")
+    parser.add_argument(
+        "--check", action="store_true", help="Enable regression checks"
+    )
+    args = parser.parse_args()
+
+    compare_two_benchdnn(args.file1, args.file2, check=args.check)
+
```
Ryo-not-rio: Remove the space here

.github/automation/performance/run_benchdnn_compare.sh
```diff
@@ -0,0 +1,56 @@
...
+MODE='--mode=P'
+
+read -ra TOKENS <<< "$CMD_STR"
+FINAL_ARGS=()
+if [[ "$CMD_STR" != *"$MODE"* ]]; then
+  FINAL_ARGS=("${TOKENS[0]}" "${PERF}" "${MODE}" "${TOKENS[@]:1}")
+else
+  FINAL_ARGS=("${TOKENS[0]}" "${PERF}" "${TOKENS[@]:1}")
+fi
+
```
Ryo-not-rio: We can just have this line here: `FINAL_ARGS=("${TOKENS[0]}" "${PERF}" "${MODE}" "${TOKENS[@]:1}")`, no need for the if statement

.github/automation/performance/run_benchdnn_compare.sh
```diff
@@ -0,0 +1,56 @@
...
+for i in $(seq 1 "$REPS"); do
+  echo "Running base iteration $i..."
+  OMP_NUM_THREADS="$THREADS" "$BASE_BINARY" "${FINAL_ARGS[@]}" >> base.txt
+done
+
+for i in $(seq 1 "$REPS"); do
+  echo "Running new iteration $i..."
+  OMP_NUM_THREADS="$THREADS" "$NEW_BINARY" "${FINAL_ARGS[@]}" >> new.txt
+done
```
Ryo-not-rio: Please take a look at how tests are interleaved in bench_nightly_performance.sh. The tests should be interleaved to capture the aws variability and avoid false positives

.github/automation/performance/run_benchdnn_compare.sh
```diff
@@ -0,0 +1,56 @@
...
+
+echo "base.txt"
+cat base.txt || echo "base.txt is empty"
+echo "new.txt"
+cat new.txt || echo "new.txt is empty"
```
Ryo-not-rio: No need to echo these files here

.github/workflows/performance-aarch64.yml
```diff
@@ -19,7 +19,34 @@ name: "Performance AArch64"
...
+        required: false
+        type: string
+        description: 'Number of threads to use'
+      benchdnn_command:
+        required: false
+        type: string
+        description: 'benchdnn command to run'
+        
   workflow_call:
+
```
Ryo-not-rio: remove

.github/workflows/performance-aarch64.yml
```diff
@@ -89,11 +116,12 @@ jobs:
...
           ACL_ROOT_DIR: ${{ github.workspace }}/ComputeLibrary
-          ACL_VERSION: ${{ fromJson(steps.get-versions.outputs.output).dependencies.acl }}
-
+          ACL_VERSION: ${{ inputs.acl_new_hash || inputs.acl_base_hash || fromJson(steps.get-versions.outputs.output).dependencies.acl }}
+ 
       - name: Get ACL commit hash for cache key
         id: get_acl_commit_hash
-        run: (cd ${{ github.workspace }}/ComputeLibrary && echo "ACLCommitHash=$(git rev-parse --short HEAD)") >> $GITHUB_OUTPUT
+        run: |
+          cd ${{ github.workspace }}/ComputeLibrary && echo "ACLCommitHash=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
```
Ryo-not-rio: We need two different versions of ACL built - new and old. Currently, it looks like you're only building one version. You'll need to modify the aarch64-acl.yml to take an ACL hash as an input and build ACL for that hash

.github/workflows/performance-aarch64.yml
```diff
@@ -145,22 +180,42 @@ jobs:
...
+        run: .github/automation/aarch64/build.sh
         working-directory: ${{ github.workspace }}/oneDNN_base
         env:
           ONEDNN_ACTION: build
 
       - name: Run performance tests
         shell: bash
         run: |
-          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_nightly_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN/build/tests/benchdnn/benchdnn base.txt new.txt
+          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_pr_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn base.txt new.txt
```
Ryo-not-rio: this should be bench_nightly_performance.sh

.github/workflows/performance-aarch64.yml
```diff
@@ -170,3 +225,4 @@ jobs:
     steps:
       - name: Print success
         run: echo Success
+
```
Ryo-not-rio: Remove the space here

.github/workflows/performance-aarch64.yml
```diff
@@ -145,22 +180,42 @@ jobs:
...
         shell: bash
         run: |
-          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_nightly_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN/build/tests/benchdnn/benchdnn base.txt new.txt
+          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_pr_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn base.txt new.txt
         env:
           DYLD_LIBRARY_PATH: ${{ github.workspace }}/ComputeLibrary/build
 
       - name: Compare 16 threads performance test results
         run: |
           python ${{ github.workspace }}/oneDNN/.github/automation/performance/benchdnn_comparison.py base.txt new.txt
```
Ryo-not-rio: We should be running these steps only if the user has not provided a benchdnn command

Ryo-not-rio (changes requested):
We need two different versions of ACL built - new and old. Currently, it looks like you're only building one version. You'll need to modify the aarch64-acl.yml to take an ACL hash as an input and build ACL for that hash

.github/automation/performance/run_benchdnn_compare.sh
```diff
@@ -0,0 +1,49 @@
...
+THREADS=${4:-16}
+REPS=5
+PERF='--perf-template=%prb%,%-time%,%-ctime%'
+MODE='--mode=P'
+
+read -ra TOKENS <<< "$CMD_STR"
+FINAL_ARGS=()
+FINAL_ARGS=("${TOKENS[0]}" "${PERF}" "${MODE}" "${TOKENS[@]:1}")
+
+echo "Final benchdnn command:"
```
Ryo-not-rio: No need for this echo

.github/workflows/performance-aarch64.yml
```diff
@@ -19,8 +19,34 @@ name: "Performance AArch64"
...
+        type: string
+        description: 'Number of threads to use'
+      benchdnn_command:
+        required: false
+        type: string
+        description: 'benchdnn command to run'
+        
   workflow_call:
-
+
```
Ryo-not-rio: Remove

.github/workflows/performance-aarch64.yml
```diff
@@ -106,11 +140,30 @@ jobs:
           key: ${{ steps.get_system_name.outputs.SystemName }}-acl-${{ matrix.config.toolset }}-${{ matrix.config.build }}-${{ steps.get_acl_commit_hash.outputs.ACLCommitHash }}
           path: ${{ github.workspace }}/ComputeLibrary/build
```
Ryo-not-rio: You need to have two restore steps - one for base and one for new ACL. Have a look at the github actions caching action documentation to see how it works

.github/workflows/performance-aarch64.yml
```diff
@@ -89,11 +122,12 @@ jobs:
         env:
           ACL_ACTION: clone
           ACL_ROOT_DIR: ${{ github.workspace }}/ComputeLibrary
-          ACL_VERSION: ${{ fromJson(steps.get-versions.outputs.output).dependencies.acl }}
-
+          ACL_VERSION: ${{ inputs.acl_new_hash || inputs.acl_base_hash || fromJson(steps.get-versions.outputs.output).dependencies.acl }}
```
Ryo-not-rio: we can just use the ci.json version for the cloning step and git checkout later

.github/workflows/performance-aarch64.yml
```diff
@@ -145,22 +192,43 @@ jobs:
...
+
+      - name: Run user-provided benchdnn command on both builds
+        if: ${{ inputs.benchdnn_command != '' }}
+        shell: bash
+        run: |
+          bash ${{ github.workspace }}/oneDNN/.github/automation/performance/run_benchdnn_compare.sh \
+            "${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn" \
+            "${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn"  \
+            "${{ inputs.benchdnn_command }}"                                    \
+            "${{ inputs.num_threads }}"
```
Ryo-not-rio: We can just do OMP_NUM_THREADS=${{inputs.num_threads}}

.github/workflows/performance-aarch64.yml
```diff
@@ -145,22 +192,43 @@ jobs:
...
+          OMP_NUM_THREADS=16 bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_pr_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn base.txt new.txt
 
       - name: Compare 16 threads performance test results
+        if: ${{ inputs.benchdnn_command == '' }}
         run: |
           python ${{ github.workspace }}/oneDNN/.github/automation/performance/benchdnn_comparison.py base.txt new.txt
 
+      - name: Check for regressions
+        if: ${{ inputs.benchdnn_command == '' }}
+        continue-on-error: true
```
Ryo-not-rio: We don't want to continue on error if there is a regression

Ryo-not-rio (changes requested):

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done,

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Fixed, Done

.github/workflows/performance-aarch64.yml
gausah01: I have created a new bash script now- run_benchdnn_comparison.py. Done, Thanks!

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Done

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Done

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: I’ve refactored the workflow based on our discussion. We're now using two separate clones for the ACL base and new commits, as checking out the hashes independently after a single clone was leading to issues.

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/workflows/performance-aarch64.yml
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/benchdnn_comparison.py
gausah01: Done

.github/automation/performance/bench_nightly_performance.sh
```diff
@@ -19,37 +19,19 @@
...
-        "$1 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $3"
-        "$2 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $4"
-    )
+    "--matmul --batch=$INPUTS_DIR/matmul_nightly"
+    "--conv --batch=$INPUTS_DIR/conv_nightly"
+    "--eltwise --batch=$INPUTS_DIR/eltwise_nightly"
+    "--reorder --batch=$INPUTS_DIR/reorder_nightly"
+)
 
 N=5
```
Ryo-not-rio: sorry, we don't need `N` anymore, could you remove it from both here and pr_performance?

gausah01: Done

Ryo-not-rio:
You need to add `--check` ci-aarch64.yml as well

.github/automation/performance/run_benchdnn_compare.sh
```diff
@@ -0,0 +1,51 @@
...
+#   .github/automation/performance/run_benchdnn_compare.sh \
+#       ${BASE_BIN} ${NEW_BIN} ${BASE_OUT} ${NEW_OUT} ${CMD}
+
+set -euo pipefail
+
+BASE_BINARY=$1
+NEW_BINARY=$2
+BASE_OUT=$3
+NEW_OUT=$4
+CMD_STR=("${@:5}")
```
Sqvid: I assume you are trying to get the benchdnn driver right? Like `--reorder` or `--conv`? In which case maybe name this `BENCHDNN_DRIVER` or something a bit more descriptive?

Also why is `${@:5}` required here? Was there some edge-case with `$5`?

.github/automation/performance/bench_nightly_performance.sh
```diff
@@ -19,37 +19,17 @@
...
-        "$2 --eltwise --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/eltwise_nightly' >> $4"
-        "$1 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $3"
-        "$2 --reorder --mode=P $PERF_TEMPLATE --batch='$INPUTS_DIR/reorder_nightly' >> $4"
-    )
-
-N=5
+    "--matmul --batch=$INPUTS_DIR/matmul_nightly"
+    "--conv --batch=$INPUTS_DIR/conv_nightly"
+    "--eltwise --batch=$INPUTS_DIR/eltwise_nightly"
+    "--reorder --batch=$INPUTS_DIR/reorder_nightly"
```
Sqvid, line 29: There is a slightly tricky edge-case here if a user runs this locally and their `$INPUTS_DIR` has a space in it. Is this intended for local use? (I think that would be useful)

.github/automation/performance/bench_pr_performance.sh
```diff
@@ -19,36 +19,17 @@
...
-
-    for test in "${TESTS[@]}"
-    do
-        echo "Starting ${test}"
-        SECONDS=0
-        eval $test
-        duration=$SECONDS
-        echo "Completed in $((duration / 60)):$((duration % 60))"
-    done
+    $SCRIPT_DIR/run_benchdnn_compare.sh $1 $2 $3 $4 $test
```
Sqvid: `$1`, `$2`, `$3`, and `$4`should be double-quoted.

.github/automation/performance/run_benchdnn_compare.sh
gausah-arm: So, I believe using just $5 would only capture the first driver arg (like --conv) and miss all others like --dir=, --dt=, etc. while ${@:5} safely captures all remaining args and handles them correctly as an array, even if arguments include spaces or complex values.

Sqvid (changes requested):

.github/automation/performance/run_benchdnn_compare.sh
gausah01: So, I believe using just `$5` would only capture the first driver arg like `--reorder` or `--conv` and miss all the other args like `--dir= --dt= etc`. and `${@:5}` safely captures all the remaining args and handles them correctly as array even if the args contains spaces or complex values.

Sqvid: Ah I see please ignore that comment then. I misunderstood the intent here.

Sqvid: I tried it locally and I don't think it is quite parsing as you expect. It is indeed only picking up the driver as I initially thought. I might be doing something wrong though.

```bash
$ ./.github/automation/performance/bench_pr_performance.sh ./build/tests/benchdnn/benchdnn ./build/tests/benchdnn/benchdnn old.txt new.txt
CMD_STR: --matmul
CMD_STR: --conv
CMD_STR: --eltwise
CMD_STR: --reorder
```

Sqvid: You probably intended to quote `$test` in `bench_pr_performance.sh` and `bench_nightly_performance.sh`

Sqvid: Another possible cause of your parsing issues is that you are creating an array here since you have surrounded `${@:5}` with round brackets. Again, it is a little difficult to figure out what your original intent was. Based on the name of the variable I assume you just want a plain string, in which case please remove the parenthesis.

Sqvid (changes requested):

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Thanks for trying this out! I think for comparison workflow, where you are benchmarking a single operation, the script run_benchdnn_compare.sh expects something like: 
`bash run_benchdnn_compare.sh ./tests/benchdnn/benchdnn ./tests/benchdnn/benchdnn base.txt new.txt --conv --dir=FWD_D --dt=f32 --alg=direct mb1_ic64oc256_ih200oh200kh1sh1dh0ph0_iw267ow267kw1sw1dw0pw0`

So the last arguments starting from $5 are: `--conv --dir=FWD_D --dt=f32 --alg=direct mb1_ic64oc256_ih200oh200kh1sh1dh0ph0_iw267ow267kw1sw1dw0pw0
`

Sqvid: I did not run `run_benchdnn_compare.sh` directly. I ran `bench_pr_performance.sh`. I provided my invocation above.

gausah01: Ah sorry, please ignore that!

Ryo-not-rio: This is to get the entire benchdnn command string e.g. --matmul --dt=bf16 512x512:512x512, we need the @:5 to handle the whitespaces in the command

.github/automation/performance/bench_nightly_performance.sh
Ryo-not-rio: the `@5` mentioned above should be able to handle the whitespace in INPUTS_DIR as well

.github/automation/performance/run_benchdnn_compare.sh
Ryo-not-rio: I think discussing with Sid, the best thing would be to add a comment as to what the cmd_str is intended to be, or to make ${CMD_STR[0]} and ${CMD_STR[@]:1} into variables

Sqvid: And probably name it `BENCHDNN_ARGS`. A variable named `CMD_STR` not being a string is a little unclear :sweat_smile:

.github/automation/performance/bench_pr_performance.sh
gausah01: Done

.github/automation/performance/run_benchdnn_compare.sh
gausah01: Done

gausah01: I have now changed the variable name. Thanks for spotting out!

gausah01:
> You need to add `--check` ci-aarch64.yml as well

Done

Sqvid:
Thanks for addressing the comments. Could you please squash your commits?

Sqvid:
You also have trailing white-space in a few files. It would be nice if you could remove it since it creates diff-noise in future edits.

gausah01:
> Thanks for addressing the comments. Could you please squash your commits?

Done

gausah01:
> You also have trailing white-space in a few files. It would be nice if you could remove it since it creates diff-noise in future edits.

It's done now! Thanks for spotting this out.

Sqvid (approved):
LGTM, thanks!

Ryo-not-rio (approved):

PR #3539, opened 2025-07-07
tests: benchdnn: graph: support softmax stats correctness check in softmax driver

ElaineBao:
# Description

The original implementation of softmax stats correctness check in https://github.com/uxlfoundation/oneDNN/pull/3396 adds a new custom driver of softmax, which has a lot of repeated code as softmax driver. To address this concern, this PR aims to support softmax stats correctness check by extending softmax driver.


ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/softmax/softmax.hpp
```diff
@@ -103,7 +107,7 @@ struct prb_t : public prb_dims_t {
 
     dir_t dir;
     dnnl_data_type_t sdt, ddt;
-    std::string stag, dtag;
+    std::string stag, dtag, stat_tag;
```
dzarukin: IIUC, stat tensor always has axis dim as 1. Al relevant cases would have axis=1 and tag=axb, which makes it abcd... or axis=-1 and tag=abx, which would still make it abcd... So the tensor should always be dense and abc-straight.
Is my observation correct? If yes, I'd replace the tag with `bool use_stat_io`. Please don't forget to add a documentation what it is for and where it's used.

tests/benchdnn/softmax/softmax.cpp
```diff
@@ -301,6 +301,7 @@ std::vector<int> supported_exec_args(dir_t dir) {
     static const std::vector<int> exec_fwd_args = {
             DNNL_ARG_SRC,
             DNNL_ARG_DST,
+            DNNL_ARG_DST_1,
```
dzarukin: Please follow `benchdnn/pool/pool.cpp::supported_exec_args(...)` as plain inclusion of an extra argument will have consequences for a regular execution flow.

tests/benchdnn/softmax/ref_softmax.cpp
```diff
@@ -66,6 +70,11 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
         }
 
         if (alg == SOFTMAX || alg == SOFTMAX_INF_AS_ZERO) {
+            if (has_stats) {
+                int64_t stats_idx = ou * inner_size + in;
+                stats_ptr[stats_idx]
```
dzarukin:
```suggestion
                stats.set_elem(stats_idx, value);
```

tests/benchdnn/softmax/bench_softmax.cpp
```diff
@@ -34,16 +34,17 @@ void check_correctness(
...
     for_(const auto &i_attr : s.attributes)
     for_(const auto &i_ctx_init : s.ctx_init)
     for_(const auto &i_ctx_exe : s.ctx_exe)
     for (auto i_inplace : s.inplace) {
-        const prb_t prb(s.prb_dims, i_dir, i_sdt, i_ddt, i_stag, i_dtag, i_alg,
-                i_axis, i_mb, i_inplace, i_attr, i_ctx_init, i_ctx_exe,
-                s.impl_filter);
+        const prb_t prb(s.prb_dims, i_dir, i_sdt, i_ddt, i_stag, i_dtag,
+                i_stat_tag, i_alg, i_axis, i_mb, i_inplace, i_attr, i_ctx_init,
+                i_ctx_exe, s.impl_filter);
```
dzarukin, line 46: I don't think this is used on graph side, might be dropped.

tests/benchdnn/graph/ref_primitive.cpp
```diff
@@ -258,6 +280,12 @@ void ref_primitive_t::check_correctness(
         compare::compare_t cmp;
         attr_t attr;
         SWITCH_DRIVER(CASE_CHECK_CORRECTNESS, CASE_CUSTOM_CHECK_CORRECTNESS);
+        if (driver_ == dnnl_driver_t::softmax) {
```
dzarukin: I suggest to think of a more generic solution...
It feels like the case when `driver_ != custom && !prim_` should fit the general pattern when to trigger an extra logic here and at other places.
Then if it's softmax, then it might have additional details...

BTW, could you clarify why the threshold is attached to the log when the log itself is just a part of the value which is not important at the end of the backward chain...

ElaineBao: OK, let me try to put it into the setup_cmp(..) function of softmax driver.

Regarding the question why the threshold is attached to the log, it's because the computation of stats involves log, which has a larger difference than the original threshold of softmax. If using the original threshold of softmax, it cannot pass the correctness check..

tests/benchdnn/softmax/bench_softmax.cpp
ElaineBao: The c-tor of prb is changed to have a new stat_tag, in order to be used in graph side, for memory creation and reference computation. Do you think I can add a new c-tor instead of changing the original one, in this way, we don't need the change here.

tests/benchdnn/softmax/softmax.hpp
ElaineBao: I'm not sure if it will always be abcd, because the tag is determined by users in their inputs, that means the tag can be anything.. But I guess in most cases in SDPA, axis=-1, tag=abcd.

tests/benchdnn/softmax/softmax.cpp
ElaineBao: updated, thanks

tests/benchdnn/softmax/softmax.hpp
dzarukin: Graph API provides only two tags options - ABX and AXB, AFAIK, thus, I'm not following your comment...

tests/benchdnn/graph/ref_primitive.cpp
dzarukin: > OK, let me try to put it into the setup_cmp(..) function of softmax driver.

That's not what I meant. There are three pieces of code in this file all of them have the same condition, which, actually, should be true for a generally supported cases by Graph API. I suggested to change the code from a fixed unscalable logic to a generic expression which should be true for this and all future cases.

I have hard times to decipher the logic for a threshold. Could you, please, clarify, what is this trh used for? Which tensor does it relax?

tests/benchdnn/softmax/softmax.hpp
ElaineBao: Gotta, the stat_tag here was never used. Changed to a boolean flag, please take another look, thanks!

tests/benchdnn/graph/ref_primitive.cpp
ElaineBao: >  There are three pieces of code in this file all of them have the same condition, which, actually, should be true for a generally supported cases by Graph API. I suggested to change the code from a fixed unscalable logic to a generic expression which should be true for this and all future cases.

Thanks for the suggestion, changed to a more general way to handle the special case of memory and execution, please take another look, thanks!

tests/benchdnn/graph/ref_primitive.cpp
```diff
@@ -149,6 +149,7 @@ void ref_primitive_t::init_memory_args(const engine_t &ref_eng) {
             const ::driver::prb_t *prb = prb_wrapper_->get<::driver::prb_t>(); \
             ::init_memory_args(mems_, prb, prim_, \
                     ::driver::supported_exec_args(prb->dir), ref_eng); \
+            driver::memory_post_processing(op_, mems_, ref_eng); \
```
dzarukin: The correct way is likely to do something like this:
```
            if (prim_) { \
                ::init_memory_args(mems_, prb, prim_, \
                        ::driver::supported_exec_args(prb->dir), ref_eng); \
            } else { \
                /* handle an empty primitive through a driver's reference */ \
                driver::memory_post_processing(op_, mems_, ref_eng); \
            } \
```

If `init_memory_args` returned a status, that'd be preferred to use instead, but...

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -1987,9 +2002,24 @@ bool get_softmax_dir(const deserialized_op_t &base_op_ref, dir_t &dir) {
...
+
+bool get_softmax_stag_and_dtag(const deserialized_op_t &base_op_ref,
+        ::softmax::settings_t &op_setting) {
+    get_driver_tag_by_idx(
+            base_op_ref, op_setting.stag.front(), 0, /*from_output*/ false);
+    get_driver_tag_by_idx(
+            base_op_ref, op_setting.dtag.front(), 0, /*from_output*/ true);
+    if (base_op_ref.out_lts_.size() > 1) {
+        op_setting.has_stats.front() = true;
+    }
```
dzarukin: Maybe move this piece into the main function? Doesn't seem to belong to tags.

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -2038,6 +2066,27 @@ ::softmax::settings_t get_setting(
     return op_setting;
 }
 
+void memory_post_processing(const deserialized_op_t &base_op_ref,
+        dnn_mem_map_t &mem_map, const engine_t &ref_eng) {
+    if (base_op_ref.out_lts_.size() < 2) return;
```
dzarukin: With the comment above, this becomes an assert at best.

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -158,6 +158,34 @@ void extend_dims(::graph::deserialized_lt_t &lt, size_t ndims) {
...
+DECLARE_TEMPLATE_GENERAL_FUNC(eltwise);
+DECLARE_TEMPLATE_GENERAL_FUNC(gnorm);
+DECLARE_TEMPLATE_GENERAL_FUNC(lnorm);
+DECLARE_TEMPLATE_GENERAL_FUNC(matmul);
+DECLARE_TEMPLATE_GENERAL_FUNC(pool);
+DECLARE_TEMPLATE_GENERAL_FUNC(prelu);
+DECLARE_TEMPLATE_GENERAL_FUNC(reduction);
+DECLARE_TEMPLATE_GENERAL_FUNC(reorder);
+DECLARE_TEMPLATE_GENERAL_FUNC(resampling);
+
```
dzarukin:
```suggestion
DECLARE_TEMPLATE_GENERAL_FUNC(resampling);
// DECLARE_TEMPLATE_GENERAL_FUNC(softmax); // Has defined body below.
```

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -2038,6 +2066,27 @@ ::softmax::settings_t get_setting(
...
+    dims_t strides = base_op_ref.out_lts_[1].stride_;
+    mem_map.erase(DNNL_ARG_DST_1);
+    mem_map.emplace(DNNL_ARG_DST_1,
+            dnn_mem_t(shapes.size(), shapes.data(), dnnl_f32, strides.data(),
+                    ref_eng, /* prefill = */ true));
+}
+
+int execute(const ::softmax::prb_t *prb, dnnl_primitive_t prim,
+        const args_t &args, res_t *res) {
+    if (prb->has_stats) {
```
dzarukin: If at any time in the future the softmax primitive gets this support, the way it's written here will never make into it for faster execution. Let's try to write the code the way it relies on `prim_` presence to avoid the knowledge of this place when/if primitive gets updated.

tests/benchdnn/softmax/ref_softmax.cpp
```diff
@@ -28,6 +28,9 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
 
     float *dst_ptr = (float *)dst;
 
+    const dnn_mem_t &stats = args.find(DNNL_ARG_DST_1);
+    const bool has_stats = prb->has_stats;
```
dzarukin:
```suggestion
```

tests/benchdnn/softmax/ref_softmax.cpp
```diff
@@ -66,6 +69,12 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
         }
 
         if (alg == SOFTMAX || alg == SOFTMAX_INF_AS_ZERO) {
+            if (has_stats) {
```
dzarukin:
```suggestion
            if (prb->has_stats) {
```

tests/benchdnn/softmax/ref_softmax.cpp
```diff
@@ -28,6 +28,9 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
 
     float *dst_ptr = (float *)dst;
 
+    const dnn_mem_t &stats = args.find(DNNL_ARG_DST_1);
```
dzarukin, line 32:
```suggestion
    // Used for graph reference version only when runs SDPA fwd training.
    const dnn_mem_t &stats = args.find(DNNL_ARG_DST_1);
```

tests/benchdnn/softmax/softmax.hpp
```diff
@@ -74,15 +76,16 @@ struct prb_t : public prb_dims_t {
...
         SAFE_V(s.has_single_setup() ? OK : FAIL);
     }
 
     prb_t(const prb_dims_t &prb_dims, dir_t dir, dnnl_data_type_t sdt,
             dnnl_data_type_t ddt, const std::string &stag,
             const std::string &dtag, alg_t alg, int axis, int64_t mb,
             bool inplace, const attr_t &attr, const thr_ctx_t &ctx_init,
-            const thr_ctx_t &ctx_exe, const impl_filter_t &impl_filter)
+            const thr_ctx_t &ctx_exe, const impl_filter_t &impl_filter,
+            bool has_stats)
```
dzarukin: Nit: please put between the `axis` and `mb`. Last several entries are unified across all drivers and are used for the fast insert through sed when a massive driver update is needed. Breaking the pattern will make it harder to update all at once.

tests/benchdnn/graph/ref_primitive.cpp
ElaineBao: Previously the softmax prim will be created and used to init memory args for src and dst. After this change, softmax prim will not be created any more, all the memory args, including src, dst and stats will be init at driver::memory_post_processing function. Please take a look, thanks!

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -2025,19 +2054,55 @@ ::softmax::settings_t get_setting(
...
+    dims_t stats_strides = base_op_ref.out_lts_[1].stride_;
+    mem_map.emplace(DNNL_ARG_DST_1,
+            dnn_mem_t(stats_shapes.size(), stats_shapes.data(), dnnl_f32,
+                    stats_strides.data(), ref_eng, /* prefill = */ true));
+}
+
+int execute(const ::softmax::prb_t *prb, dnnl_primitive_t prim,
+        const args_t &args, res_t *res) {
+    if (!prim) {
+        ::softmax::compute_ref(prb, prb->dir, args);
```
ElaineBao: placeholder: should have some reorders before and after the computation

tests/benchdnn/softmax/softmax.cpp
```diff
@@ -230,6 +236,11 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
         res->reason = skip_reason::case_not_supported;
         return;
     }
+    if (prb->has_stats) {
+        res->state = SKIPPED;
+        res->reason = skip_reason::case_not_supported;
+        return;
+    }
```
dzarukin: Discussed offline: to be removed.

tests/benchdnn/graph/setting_handler.hpp
```diff
@@ -42,6 +42,11 @@ namespace graph {
     namespace driver { \
     ::driver::settings_t get_setting( \
             const deserialized_op_t &base_op_ref, res_t *res); \
+    void memory_post_processing(const ::driver::prb_t *prb, \
+            const deserialized_op_t &base_op_ref, dnn_mem_map_t &mem_map, \
+            const engine_t &ref_eng); \
+    int execute(const ::driver::prb_t *prb, dnnl_primitive_t prim, \
+            const args_t &args, res_t *res); \
```
dzarukin: Discussed offline: asked to move to ref_primitive.cpp

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -2025,19 +2054,79 @@ ::softmax::settings_t get_setting(
...
+        ::softmax::compute_ref(prb, prb->dir, ref_args);
+
+        // restore original memory format after ::softmax::compute_ref
+        SAFE_V(dst.reorder(dst_f32_abx));
+        SAFE_V(stats.reorder(stats_f32_abx));
+
+        res->state = EXECUTED;
+    } else {
+        SAFE(execute_and_wait(prim, args, res), WARN);
+    }
```
dzarukin: Discussed offline: suggested to move this `else` into a common part. `prim` arg can be stripped then.

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -158,6 +158,36 @@ void extend_dims(::graph::deserialized_lt_t &lt, size_t ndims) {
     }
 }
 
+#define DECLARE_TEMPLATE_GENERAL_FUNC(driver) \
+    void driver::memory_post_processing(const ::driver::prb_t *prb, \
```
dzarukin:
```suggestion
    void driver::init_memory_args_native(const ::driver::prb_t *prb, \
```

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -2025,19 +2054,79 @@ ::softmax::settings_t get_setting(
...
+
+int execute(const ::softmax::prb_t *prb, dnnl_primitive_t prim,
+        const args_t &args, res_t *res) {
+    if (!prim) {
+        assert(args.find(DNNL_ARG_DST_1).size() != 0);
+        // in order to run with ::softmax::compute_ref, we need to
+        // construct all memory with abx tag and f32 dt
+        args_t ref_args;
+        const dnn_mem_t &src = args.find(DNNL_ARG_SRC);
+        dnn_mem_t src_f32_abx(src, dnnl_f32, "abx", src.engine());
```
dzarukin:
```suggestion
        dnn_mem_t src_f32_abx(src, dnnl_f32, tag::abx, src.engine());
```

tests/benchdnn/graph/ref_partition.cpp
```diff
@@ -87,6 +87,9 @@ int ref_partition_t::init_ref(
         SAFE(ref_prim->init_prb(res), WARN);
 
         SAFE_V(ref_prim->init_prim(::get_test_engine(), res));
+        // reset res to avoid a skipped state from init_prim() to affect the rest part.
+        res->state = UNTESTED;
+        res->reason.clear();
```
dzarukin: Discussed offline: unimplemented should be emerged here and processed later the stack where it's known if there's a native driver support or not. If not, unimplemented should be returned to the user.

Adding a skip condition here would be good, too.

ElaineBao: Thanks for the comments, fixed. The unimplemented status will be returned [here](https://github.com/uxlfoundation/oneDNN/blob/f23b3fbd2411f14a64b3f3090b52e2a63a2a4cbb/tests/benchdnn/graph/ref_primitive.cpp#L27). 
The skip condition is already inside the init_prim function ([here](https://github.com/uxlfoundation/oneDNN/blob/f23b3fbd2411f14a64b3f3090b52e2a63a2a4cbb/tests/benchdnn/graph/ref_primitive.cpp#L217)), so I think there's no need to add a new one.

dzarukin: It would prevent from going further is SKIPPED status was issued.

PR #3542, opened 2025-07-07
ngen: rename ELFCodeGenerator::interface

rjoursler:
The windows.h header contains the following define:

```
#define interface                   struct
```

This adversely interacts with the ELFCodeGenerator::interface function and produces a compilation error. To avoid this interaction, the function is renamed.


rjoursler:
make test
disable test_device_cpu

echeresh (approved):

PR #3548, opened 2025-07-07
[GPU] various quantization fixes

Simonsays095:
Fixes several issues in quantization code, each with their own commit. Also renames variables used in gemm code to rely on A/B names instead of SRC/WEI whenever possible, to ensure variable naming consistency.

High-impact bugs fixed:
- In some cases, common scales/zero-points were emitting invalid instructions leading to accuracy issues. They now generate correctly.
- Some valid cases were getting skipped without reason, which now go to the optimized implementation.
- When reshaping GEMM layers to reduce dimensions, dst scales were not getting reshaped with the other scales.
- Reduced strictness of dst scales support via jit:gemm:any, which was overly-restrictive.
- Some `per_ocic` quantization cases with M=1 or N=1 were treated as 1D, although gemmstone requires these to be labeled as 2D.

Also included a build fix for generic builds, sequestering intel-specific code from these configurations.

### DIFF
--- a/src/gpu/intel/jit/gemm/gen_gemm.hpp
+++ b/src/gpu/intel/jit/gemm/gen_gemm.hpp
@@ -66,11 +66,28 @@ struct gen_gemm_t : public gpu_gemm_t {
                     VERBOSE_UNSUPPORTED_ATTR);
 
             auto &attr_zps = attr()->zero_points_;
+            auto &attr_scales = attr()->scales_;
 
             dev_info_ = compute_engine->device_info();
             arch_ = dev_info_->gpu_arch();
             int stepping = dev_info_->stepping_id();
             init_attrs();
+
+            // If we have both grouped scales and grouped zero-points, they must
+            // have the same group size
+            if (a_scales_2d() && a_zp_2d()) {
+                auto asc_group_k = attr_scales.get_group(DNNL_ARG_A, 0);
+                auto azp_group_k = attr_zps.get_group(DNNL_ARG_A, 0);
+                VDISPATCH_GEMM(
+                        asc_group_k == azp_group_k, VERBOSE_UNSUPPORTED_ZP_CFG);
+            }
+            if (b_scales_2d() && b_zp_2d()) {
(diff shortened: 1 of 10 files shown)


src/gpu/intel/jit/gemm/gen_gemm.hpp
```diff
@@ -55,17 +55,39 @@ struct gen_gemm_t : public gpu_gemm_t {
...
+            auto &attr_scales = attr()->scales_;
 
             dev_info_ = compute_engine->device_info();
             arch_ = dev_info_->gpu_arch();
             int stepping = dev_info_->stepping_id();
             init_attrs();
+
+            // If we have both grouped scales and grouped zero-points, they must
+            // have the same group size
+            if (a_scales_2d() && a_zp_2d()) {
```
kealan-barbieri, line 78: looks like this is duplicated in `xe_hp_systolic`, can it be checked in a shared function?

kealan-barbieri:
Were the invalid instructions in common scales/zp existing prior to the your refactors to attribute handling?

src/gpu/intel/jit/gemm/gen_gemm.hpp
Simonsays095: Actually, xe_hp_systolic doesn't support groups. I added this unnecessarily - removing from xe_hp_systolic and leaving this here.

Simonsays095:
> Were the invalid instructions in common scales/zp existing prior to the your refactors to attribute handling?

Yes. These failures are uncovered in #3214, and these changes reduce the failures from 39 down to just 1.

tmp.out
petercad, line 1: This file wasn't added on purpose, I guess.

Simonsays095: It was not, thank you.

kealan-barbieri (approved):

dzarukin (approved):

Simonsays095:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

PR #3549, opened 2025-07-07
[rls-v3.9 backport][GPU] various quantization fixes

Simonsays095:
Backport of #3548 to rls-v3.9.

PR #3551, opened 2025-07-07
gpu: add guard to vendor specific headers in sdpa list

syurkevi:
# Description

Vendor specific headers cause a compilation error in gpu_sdpa_list when building the generic backend. This PR adds a vendor specific guard to avoid the inclusion of these incomplete headers.

# Checklist

## General


syurkevi:
PR redundant, fix included in #3548 , #3549.

PR #3552, opened 2025-07-08
cpu: aarch64: Support for int8 in BRGEMM General and 1x1 Forward Convolution

kasturedeeksha:
# Description

This commit adds support for the int8 data type with the configuration u8:s8:f32 (src:wei:dst) in BRGEMM general and 1x1 forward convolution. It also enables bias_dt=undef. Both propagation kinds, FWD_B and FWD_I, will now execute BRGEMM convolution for the int8 data type.

# Checklist

## General

1. make test output
```
98% tests passed, 4 tests failed out of 224
 
Total Test time (real) = 1015.08 sec
 
The following tests FAILED:
        172 - test_graph_unit_dnnl_large_partition_cpu (Failed)
        195 - test_benchdnn_modeC_binary_ci_cpu (Failed)
        196 - test_benchdnn_modeC_binary_different_dt_ci_cpu (Failed)
        204 - test_benchdnn_modeC_graph_ci_cpu (Failed)
```
Yes


src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
```diff
@@ -167,14 +168,14 @@ void jit_uni_brgemm_conv_comp_pad_kernel_t<isa>::khw_loop(const int icb,
...
-    add_imm(reg_kh_l, param1, GET_OFF(kh_l), X_TMP_0);
+    ldr(reg_kh_l, ptr(param1, static_cast<uint32_t>(GET_OFF(kh_l))));
     mov(reg_aux_kh_in, reg_in);
 
     L_aligned(label_kh_loop);
     {
         cmp(reg_kh_l, 0);
         b(EQ, label_kh_end);
-        add_imm(reg_kw_l, param1, GET_OFF(kw_l), X_TMP_0);
+        ldr(reg_kw_l, ptr(param1, static_cast<uint32_t>(GET_OFF(kw_l))));
```
jondea: Do we need the static case if it's now in the macro?

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
```diff
@@ -167,14 +168,14 @@ void jit_uni_brgemm_conv_comp_pad_kernel_t<isa>::khw_loop(const int icb,
         const int mb_tail, const int n_block) {
     Xbyak_aarch64::Label label_kw_loop, label_kw_end, label_kh_loop,
             label_kh_end;
-    add_imm(reg_kh_l, param1, GET_OFF(kh_l), X_TMP_0);
+    ldr(reg_kh_l, ptr(param1, static_cast<uint32_t>(GET_OFF(kh_l))));
```
jondea: `add_imm` and `ldr` are very different, did this work before?

src/cpu/aarch64/jit_brgemm_conv_utils.cpp
```diff
@@ -828,7 +828,8 @@ bool brg_blocking_t::fast_check_oc_block() const {
     const auto rnd_oc = rnd_up(oc, acc_simd_w);
     auto res = false;
     if (oc_block == 64) {
-        res = (rnd_oc % oc_block == 0 && rnd_oc * wei_dsz < 192 * 4);
+        res = one_of(src_dt, u8, s8)
```
jondea, line 831: Could you put a comment explaining the logic for u8/s8? Do we just always use oc_block = 64? If so, I think the logic could be more explicit

src/cpu/aarch64/jit_brgemm_post_ops.hpp
```diff
@@ -457,8 +457,9 @@ struct jit_brgemm_kernel_post_ops : public jit_generator {
             case data_type::bf16: assert(!"unsupported data type\n"); break;
             default: assert(!"unsupported data type");
         }
-        if (!utils::one_of(type_in, data_type::f32, data_type::bf16))
-            assert(!"unsupported data type\n");
+        if (types::is_integral_dt(type_in)) {
```
jondea, line 460: If `type_in` was s8, u8 or u32, `is_integral_dt` would be true, but `scvtf` would produce garbage. Could we make this more specific to s32?

jondea (changes requested):
This generally looks great, thank you! I just have a few minor comments.

jondea:
Can we have some rough benchmark numbers? Also, what implementation does this replace?

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
```diff
@@ -196,9 +197,9 @@ void jit_uni_brgemm_conv_comp_pad_kernel_t<isa>::khw_loop(const int icb,
 
 template <cpu_isa_t isa>
 void jit_uni_brgemm_conv_comp_pad_kernel_t<isa>::load_params() {
-    add_imm(reg_in, param1, GET_OFF(ptr_in), X_TMP_0);
-    add_imm(reg_zp_comp_out, param1, GET_OFF(ptr_zp_out), X_TMP_1);
-    add_imm(reg_comp_out, param1, GET_OFF(ptr_cp_out), X_TMP_2);
+    ldr(reg_in, ptr(param1, (GET_OFF(ptr_in))));
+    ldr(reg_zp_comp_out, ptr(param1, (GET_OFF(ptr_zp_out))));
+    ldr(reg_comp_out, ptr(param1, (GET_OFF(ptr_cp_out))));
```
renato-arantes, line 202: You have changed from `add_imm` to `ldr` in several places. Could you explain the motivation behind this?

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
```diff
@@ -276,7 +277,7 @@ void jit_uni_brgemm_conv_comp_pad_kernel_t<isa>::generate() {
     const auto mb_tail = div_up(icb_tail, ic_step);
 
     Label label_kd_loop, label_loop_end;
-    add_imm(reg_kd_l, param1, GET_OFF(kd_l), X_TMP_0);
+    ldr(reg_kd_l, ptr(param1, static_cast<uint32_t>(GET_OFF(kd_l))));
```
renato-arantes: As @jondea mentioned, the cast is already in the macro. Why the static here?

src/cpu/aarch64/jit_brgemm_post_ops.hpp
kasturedeeksha: is_integral_dt is defined in src/common/type_helpers.hpp and returns true for s32, s8, u8, u4 and s4 so it will not be true for u32. 
Regarding s8 and u8, we will explicitly perform bit extension (sign-extension for s8, zero-extension for u8) to safely convert them to s32. This will ensure there is no garbage when using scvtf, as the data is always correctly prepared in 32-bit signed format.

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
kasturedeeksha: Since this kernel wasn’t actually being used earlier, this issue didn’t show up. Now that it’s used for int8, we found that with add_imm we were only computing addresses but not loading the actual values, which caused errors at runtime.

This change to ldr directly fixes that by properly loading the values from memory.

kasturedeeksha:
@jondea We don't have any other implementation for u8:s8:f32 other than reference gemm_s8u8s32:ref
machine : hpc7g.8xlarge
command used : 
**For INT8:** 
`./benchdnn --conv --mode=p --dt=u8:s8:f32 --batch=inputs/conv/set_conv_all`

> total perf: min(ms):297.863 **avg(ms):300.155**
> total: 666.01s; create_pd: 0.03s (0%); create_prim: 0.08s (0%); fill: 1.35s (0%); execute: 0.31s (0%);

`./benchdnn --conv --mode=p --dt=u8:s8:f32 --stag=axb --dtag=axb --batch=inputs/conv/set_conv_all`

> total perf: min(ms):296.136 **avg(ms):298.139**
> total: 665.88s; create_pd: 0.03s (0%); create_prim: 0.08s (0%); fill: 1.21s (0%); execute: 0.31s (0%);

**For FP32:**
`./benchdnn --conv --mode=p --batch=inputs/conv/set_conv_all`

> total perf: min(ms):1155.73 **avg(ms):1160.53**
> total: 668.24s; create_pd: 0.01s (0%); create_prim: 0.03s (0%); fill: 1.24s (0%); execute: 1.18s (0%);

`./benchdnn --conv --mode=p --stag=axb --dtag=axb --batch=inputs/conv/set_conv_all`

> total perf: min(ms):1165.66 **avg(ms):1171.5**
> total: 668.63s; create_pd: 0.06s (0%); create_prim: 0.09s (0%); fill: 1.26s (0%); execute: 1.19s (0%);

For both formats (with and without channel last) we are getting roughly **4x** performance from FP32, similarly with int8 ref we are getting around **45x** performance.

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
kasturedeeksha: You're right, I updated the code.

src/cpu/aarch64/jit_brgemm_conv_utils.cpp
kasturedeeksha: No this is just heuristic where we can consider oc_block = 64 for int8 dtype directly.
This function fast_check_oc_block() is just a quick filter used inside the blocking loop to skip bad block sizes early before doing further computations.
The actual blocking loop tries many oc_block values, starting from large down to small.
So even if oc_block = 64 passes this check, it may later decide that 32 or 16 is better based on estimated efficiency.

jondea: Understood, thank you. Anything you have discovered that you think is even mildly interesting about this heuristic would be appreciated as comments. It would be great to make some progress in cracking what it means!

src/cpu/aarch64/jit_brgemm_conv_comp_pad_kernel.cpp
jondea: Answered in: https://github.com/uxlfoundation/oneDNN/pull/3552#discussion_r2194547093

kasturedeeksha:
Please check https://github.com/uxlfoundation/oneDNN/pull/3591#issuecomment-3079188543

PR #3554, opened 2025-07-08
cpu: x64: matmul: weight decompression

inteldimitrius:
## Brgemm matmul weight decompression support
Implements [MFDNN-11993](https://jira.devtools.intel.com/browse/MFDNN-11993) and [MFDNN-13483](https://jira.devtools.intel.com/browse/MFDNN-13483). 
### Supported cases for weight decompression:

- [x] f32:{s4/u4/s8/u8}:f32 
- [x] bf16:{s4/u4/s8/u8}:bf16
- [x] f16:{s4/u4/s8/u8}:f16
- [x] Test coverage

Scales support: `common`, `per_oc/per_ocic`
Zero_points support: `common`, `per_oc/per_ocic`

### What was done:
1. Reworked scales per_oc/per_ocic usage with grouped masks without using scratchpad memory.
2. Enabled zero-points with `per_oc/per_ocic` including grouped support.
3. Enabled u4/s4/u8/s8 for zero-points and for weights (for the cases, that didn't support before)

The big part of the weight decompression support was implemented earlier, this PR includes refactoring and adding remaining cases. It was created `jit_brgemm_matmul_copy_b_base_t` base class implementing common methods among all copy B kernels containing `load_value` and `decompress_value` methods which support weight decompression feature.

### Support
Supported ISAs: **AVX512**, **AVX2**.
Tested cases for support (setting `ONEDNN_MAX_CPU_ISA=`):
```bash
$CMD --dt=f32:s4:f32 --wtag=ab,ba --attr-scales=,wei:per_ocic:f32:32x1 --attr-zero-points=,wei:per_ocic:u8:32x1 --attr-fpmath=strict:true --skip-impl=ref 1x1024:1024x4
$CMD --dt=f16:s4:f16 --wtag=ab,ba --attr-scales=,wei:per_ocic:f32:32x1 --attr-zero-points=,wei:per_ocic:u8:32x1 --attr-fpmath=f16:true --skip-impl=ref 1x1024:1024x4
$CMD --dt=bf16:s4:bf16 --wtag=ab,ba --attr-scales=,wei:per_ocic:f32:32x1 --attr-zero-points=,wei:per_ocic:u8:32x1 --attr-fpmath=bf16:true --skip-impl=ref 1x1024:1024x4
```
#### Support table (implementations showed are on SPR): 
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

</head>

<body link="#0563C1" vlink="#954F72">

Datatype   Case (including zp/scales) | AVX512_core | AVX10_1_512_AMX | AVX10_1_512_AMX_FP16 | AVX10_2_512_AMX_2 | AVX512_CORE_BF16
-- | -- | -- | -- | -- | --
f32:{int4/int8}:f32 | brg:avx512_core | brg:avx512_core | brg:avx512_core | brg:avx512_core | brg:avx512_core
f16:{int4/int8}:f16 | SKIPPED | brg:avx10_1_512 | brg:avx10_1_512 | brg:avx10_1_512 | SKIPPED
bf16:{int4/int8}:bf16 | SKIPPED | brg:avx10_1_512_amx | brg:avx10_1_512_amx | brg:avx10_1_512_amx | brg:avx512_core_bf16

</body>

</html>

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

</head>

<body link="#0563C1" vlink="#954F72">

Datatype   case (including zp & scales) | AVX2 | AVX2_VNNI | AVX2_VNNI_2
-- | -- | -- | --
f32:{int4/int8}:f32 | brg:avx2 | brg:avx2 | brg:avx2
f16:{int4/int8}:f16 | SKIPPED | SKIPPED | SKIPPED
bf16:{int4/int8}:bf16 | SKIPPED | SKIPPED | SKIPPED

</body>

</html>

### Performance report

[perf_report_mm_wei_decomp.pdf](https://github.com/user-attachments/files/22662387/perf_report_mm_wei_decomp.pdf)

`dmitriyo/matmul_weight_decompression` vs. `main` in context of copy routines without weight decompression.
`Spoiler`: Refactoring doesn't affect on non-(weight decompression) copy routines.
[gnr_72C_0.9Ghz_wwmm_wei_decomp_testing.xlsx](https://github.com/user-attachments/files/22948968/gnr_72C_0.9Ghz_wwmm_wei_decomp_testing.xlsx)
[spr_56C_2.2Ghz_wwmm_wei_decomp_testing.xlsx](https://github.com/user-attachments/files/22948969/spr_56C_2.2Ghz_wwmm_wei_decomp_testing.xlsx)

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
@@ -2118,7 +2118,7 @@ protected:
     * @param zmm Destination ZMM register where data will be inserted
     * @param ymm_half Source YMM register containing data for the upper half
     */
-    void copy_half_int4(const Zmm &zmm, const Ymm &ymm_half) {
+    void copy_half_reg(const Zmm &zmm, const Ymm &ymm_half) {
         vinserti64x4(zmm, zmm, ymm_half, 1);
     }
 
@@ -2127,13 +2127,88 @@ protected:
     * @param ymm Destination YMM register where data will be inserted
     * @param xmm_half Source XMM register containing data for the upper half
     */
-    void copy_half_int4(const Ymm &ymm, const Xmm &xmm_half) {
+    void copy_half_reg(const Ymm &ymm, const Xmm &xmm_half) {
         vinserti128(ymm, ymm, xmm_half, 1);
+    }
+
+    /** Restores vmm register containing permute indices for int4 processing.
+    * Due to lack of registers this register is used as temporary in `prepare_loaded_int4`.
+     */
+    void restore_vmm_permd(const Ymm &vmm_permd) {
+        alignas(64) static constexpr const uint32_t int4_permute_avx2[8]
+                = {0, 4, 1, 5, 2, 6, 3, 7};
+        const auto reg_tmp = r15;
+        mov(reg_tmp, reinterpret_cast<size_t>(int4_permute_avx2));
+        vmovdqa(vmm_permd, ptr[reg_tmp]);
+    }
+
+    /** Loaded in register(half of it) containing bytes (2 int4 values)
+    * The idea is duplicate each byte in two dwords
+    * using `copy_half_reg` and `vpermd`. Then shift bytes left or right
+    * depending on the position (odd/even) and signed/unsigned type.
+     */
+    template <typename Vmm>
+    void prepare_loaded_int4(
+            const Vmm &reg, const Vmm &vmm_permd, const bool is_signed) {
+        using Vmm_lower_t = typename vreg_traits_t<Vmm>::Vmm_lower_t;
+        const auto vmm_lower = Vmm_lower_t(reg.getIdx());
+
+        copy_half_reg(reg, vmm_lower);
+        vpermd(reg, vmm_permd, reg);
+        // Without masks int4 is going to use 2 tmp registers
+        // To perform masks using VPAND operator
+        if (!isa_has_masks(conf_->isa)) {
+            // TODO: Unify register usage over kernels
+            const auto mask_vmm = Vmm(conf_->transposed_B ? 13 : 0);
+            const auto tmp_vmm = vmm_permd;
+            // const auto tmp_vmm2 = Vmm(conf_->transposed_B ? 13: 4);
+            // f32 and transposed used the same register for regq_tmp
+            const auto reg_tmp = r15;
+            alignas(64) static constexpr const uint32_t odd_indices[8] = {
+                    0, 0xffffffff, 0, 0xffffffff, 0, 0xffffffff, 0, 0xffffffff};
+            alignas(64) static constexpr const uint32_t even_indices[8] = {
+                    0xffffffff, 0, 0xffffffff, 0, 0xffffffff, 0, 0xffffffff, 0};
+            // Process odd indices
+            mov(reg_tmp, reinterpret_cast<size_t>(even_indices));
+            vmovdqa(mask_vmm, ptr[reg_tmp]);
+            uni_vpand(tmp_vmm, reg, mask_vmm);
+            uni_vpslld(tmp_vmm, tmp_vmm, 28);
+            if (is_signed) {
+                vpsrad(tmp_vmm, tmp_vmm, 28);
+            } else {
+                vpsrld(tmp_vmm, tmp_vmm, 28);
+            }
+
+            // Process even indices
+            mov(reg_tmp, reinterpret_cast<size_t>(odd_indices));
+            vmovdqa(mask_vmm, ptr[reg_tmp]);
+            uni_vpand(reg, reg, mask_vmm);
+            if (is_signed) {
+                vpsrad(reg, reg, 4);
+            } else {
+                vpsrld(reg, reg, 4);
+            }
+            // Store result to desired reg
+            vpaddd(reg, reg, tmp_vmm);
+            // Clean tmp regs
+            vpxor(mask_vmm, mask_vmm, mask_vmm);
+            restore_vmm_permd(vmm_permd);
+        } else {
+            uni_vpslld(reg | k5555, reg, 28);
+            if (is_signed) {
+                vpsrad(reg | k5555, reg, 28);
+                vpsrad(reg | kAAAA, reg, 4);
+            } else {
+                vpsrld(reg | k5555, reg, 28);
+                vpsrld(reg | kAAAA, reg, 4);
+            }
+        }
     }
 
     /**
     * @brief Loads and converts data of various types into vector registers with appropriate handling
-    * 
+    * Integer types s4, u4, s8, u8, s32 will be loaded and converted to s32 during the loading.
+    * Floating point types f16, bf16, f32 will be loaded and converted to f32 during the loading.
     * @tparam Vmm Vector register type for computation
     * @param reg Destination vector register
     * @param op Source memory operand to load from
@@ -2187,28 +2262,18 @@ protected:
                 uni_vpmovsxbd(
                         maybe_mask(vmm_lower, is_tail, /* is_int4 = */ true),
                         op);
-                copy_half_int4(reg, vmm_lower);
-
-                vpermd(reg, vmm_permd, reg);
-                uni_vpslld(reg | k5555, reg, 28);
-                vpsrad(reg | k5555, reg, 28);
-                vpsrad(reg | kAAAA, reg, 4);
+                prepare_loaded_int4(reg, vmm_permd, /* is_signed = */ true);
                 break;
             case data_type::u4:
                 uni_vpmovzxbd(
                         maybe_mask(vmm_lower, is_tail, /* is_int4 = */ true),
                         op);
-                copy_half_int4(reg, vmm_lower);
-
-                vpermd(reg, vmm_permd, reg);
-                uni_vpslld(reg | k5555, reg, 28);
-                vpsrld(reg | k5555, reg, 28);
-                vpsrld(reg | kAAAA, reg, 4);
+                prepare_loaded_int4(reg, vmm_permd, /* is_signed = */ false);
                 break;
             case data_type::f4_e2m1:
             case data_type::f4_e3m0:
                 uni_vpmovzxbd(maybe_mask(vmm_lower, is_tail), op);
-                copy_half_int4(vmm_in, vmm_lower);
+                copy_half_reg(vmm_in, vmm_lower);
                 vpermd(vmm_in, vmm_permd, vmm_in);
                 uni_vpslld(vmm_in | k5555, vmm_in, 28);
                 vpsrld(vmm_in | k5555, vmm_in, 28);
@@ -2415,7 +2480,7 @@ protected:
                 const auto src_vmm_lower1 = Vmm_lower_t(reg2.getIdx());
                 vcvtps2phx(src_vmm_lower0, reg1);
                 vcvtps2phx(src_vmm_lower1, reg2);
-                vinsertf64x4(reg1, reg1, src_vmm_lower1, 1);
+                copy_half_reg(reg1, src_vmm_lower1);
                 break;
             }
             case data_type::f32:
@@ -3496,9 +3561,11 @@ void jit_brgemm_matmul_copy_b_bf16_t<Vmm>::copy_2x32(
         auto load_addr = maybe_EVEX_compress_addr(reg_src_load, offset);
         if (!isa_has_masks(conf_->isa)) {
             if (is_tail)
-                load_bytes(src_load, load_addr, columns_tail * tr_typesize);
+                load_bytes(src_load, load_addr,
+                        columns_tail * tr_typesize / elems_per_byte);
             else
                 uni_vmovups(src_load, load_addr);
+            load_value(src_reg, src_reg, vmm_permd, conf_->orig_wei_dt, false);
         } else {
             load_value(
                     src_reg, load_addr, vmm_permd, conf_->orig_wei_dt, is_tail);
@@ -3867,7 +3934,6 @@ struct jit_brgemm_matmul_copy_b_f32_t
         , req_apply_wei_scales_(conf->apply_scales_in_buffer_b)
         , typesize_in_(types::data_type_size(dt_in_))
         , src_elems_per_byte_(is_src_int4_ || is_src_f4_ ? 2 : 1)
-        , wei_scales_typesize_(conf_->wei_scales_dt_sz)
         , src_stride_(conf_->copy_B_wei_stride)
         , tr_src_stride_(conf_->LDB * typesize_out_) {}
 
@@ -3882,10 +3948,11 @@ private:
     using opmask_t = const Xbyak::Opmask;
     using Vmm_lower_t = typename vreg_traits_t<Vmm>::Vmm_lower_t;
 
+    static constexpr bool is_ymm_ = std::is_same<Vmm, Xbyak::Ymm>::value;
     const data_type_t dt_in_;
     const int simd_w_;
     const bool is_src_f4_, is_src_int4_, req_zp_b_shift_, req_apply_wei_scales_;
-    const size_t typesize_in_, src_elems_per_byte_, wei_scales_typesize_;
+    const size_t typesize_in_, src_elems_per_byte_;
     const size_t typesize_out_ = sizeof(float);
     dim_t src_stride_, tr_src_stride_;
 
@@ -3940,9 +4007,11 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::copy_16_x_n_block(
         const bool is_tail = ncolumns - n < simd_w_;
         auto addr = maybe_EVEX_compress_addr(reg_src,
                 (k * src_stride_ + n * typesize_in_) / src_elems_per_byte_);
-        if (is_tail && !isa_has_masks(conf_->isa))
-            vmaskmovps(src_vmm, ymm_tail_mask, addr);
-        else
+        if (is_tail && !isa_has_masks(conf_->isa)) {
+            load_bytes(src_vmm, addr,
+                    (ncolumns % simd_w_) * typesize_in_ / src_elems_per_byte_);
+            load_value(src_vmm, src_vmm, vmm_permd, conf_->orig_wei_dt, false);
+        } else
             load_value(src_vmm, addr, vmm_permd, conf_->orig_wei_dt, is_tail);
 
         decompress_reg(maybe_mask(src_vmm, is_tail), vmm_zp_b_shift,
@@ -3962,7 +4031,12 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::copy_16_x_n_block(
                 = one_of(zp_dt, data_type::s4, data_type::u4) ? 2 : 1;
         const auto offset = n * zp_dt_sz / elems_per_byte;
         const auto addr = maybe_EVEX_compress_addr(reg_zp_ptr, offset);
-        load_value(vmm_zp_b_shift, addr, vmm_permd, zp_dt, is_tail);
+        if (is_tail && !isa_has_masks(conf_->isa)) {
+            load_bytes(vmm_zp_b_shift, addr,
+                    (ncolumns % simd_w_) * zp_dt_sz / elems_per_byte);
+            load_value(vmm_zp_b_shift, vmm_zp_b_shift, vmm_permd, zp_dt, false);
+        } else
+            load_value(vmm_zp_b_shift, addr, vmm_permd, zp_dt, is_tail);
     };
 
     /**  Loads scales, when is_wei_scale_per_n is set.
@@ -3977,7 +4051,12 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::copy_16_x_n_block(
         const auto scales_dt_sz = types::data_type_size(scales_dt);
         const auto offset = n * scales_dt_sz;
         const auto addr = maybe_EVEX_compress_addr(reg_wei_scales, offset);
-        load_scale_value(vmm_wei_scales, addr, scales_dt, is_tail);
+        if (is_tail && !isa_has_masks(conf_->isa)) {
+            load_bytes(
+                    vmm_wei_scales, addr, (ncolumns % simd_w_) * scales_dt_sz);
+            load_scale_value(vmm_wei_scales, vmm_wei_scales, scales_dt, false);
+        } else
+            load_scale_value(vmm_wei_scales, addr, scales_dt, is_tail);
     };
 
     const int columns_tail = ncolumns % simd_w_;
@@ -3990,8 +4069,6 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::copy_16_x_n_block(
                         = (1 << (columns_tail / src_elems_per_byte_)) - 1;
                 kmovw(kTail_int4, tail_mask_4bit);
             }
-        } else {
-            init_f32_avx2_mask_ymm(ymm_tail_mask, reg_tmp, columns_tail);
         }
     }
 
@@ -4056,17 +4133,24 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::generate() {
     mov(reg_wei_scales, ptr[param1 + GET_OFF(wei_scales_ptr)]);
     mov(reg_zp_ptr, ptr[param1 + GET_OFF(zp_b_value_ptr)]);
     kmovw(kFFFF, 0xffff); // 1111111111111111
-    if (is_src_int4_ || is_src_f4_) {
-        alignas(64) static constexpr const uint32_t int4_permute[16]
-                = {0, 8, 1, 9, 2, 10, 3, 11, 4, 12, 5, 13, 6, 14, 7, 15};
-        mov(reg_tmp, reinterpret_cast<size_t>(int4_permute));
-        vmovdqa32(vmm_permd, ptr[reg_tmp]);
 
+    if (is_src_int4_ || is_src_f4_) {
         kmovw(kAAAA, 0xaaaa);
         kmovw(k5555, 0x5555);
+        if (is_superset(conf_->isa, avx512_core)) {
+            alignas(64) static constexpr const uint32_t int4_permute[16]
+                    = {0, 8, 1, 9, 2, 10, 3, 11, 4, 12, 5, 13, 6, 14, 7, 15};
+            mov(reg_tmp, reinterpret_cast<size_t>(int4_permute));
+            vmovdqa32(vmm_permd, ptr[reg_tmp]);
+        } else if (is_superset(conf_->isa, avx2)) {
+            alignas(64) static constexpr const uint32_t int4_permute_avx2[8]
+                    = {0, 4, 1, 5, 2, 6, 3, 7};
+            mov(reg_tmp, reinterpret_cast<size_t>(int4_permute_avx2));
+            vmovdqa(vmm_permd, ptr[reg_tmp]);
+        }
     }
-    if (is_src_f4_) {
 
+    if (is_src_f4_ && is_superset(conf_->isa, avx512_core)) {
         alignas(64) static constexpr const float f4_e2m1_table[16]
                 = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f, -.5f,
                         -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
@@ -4347,9 +4431,9 @@ private:
         const auto &scales_dt_sz = conf_->wei_scales_dt_sz;
         const auto offset = n * scales_dt_sz;
         const auto masked_vmm = maybe_mask(vmm_wei_scales, is_tail);
-        const auto addr = EVEX_compress_addr(
+        const auto addr = maybe_EVEX_compress_addr(
                 reg_wei_scales, offset, scales_dt == data_type::f16);
-        vpxord(vmm_wei_scales, vmm_wei_scales, vmm_wei_scales);
+        uni_vpxor(vmm_wei_scales, vmm_wei_scales, vmm_wei_scales);
         switch (scales_dt) {
             case data_type::f32: uni_vbroadcastss(vmm_wei_scales, addr); break;
             case data_type::bf16:
@@ -4383,6 +4467,7 @@ template <typename Vmm>
 void jit_brgemm_matmul_copy_b_transposed_t<Vmm>::init_tail_mask(
         const int columns_tail, const bool use_int4_mask) {
     assert(IMPLICATION(use_int4_mask, is_src_int4_));
+    assert(isa_has_masks(conf_->isa));
     if (columns_tail > 0) {
         const int dt_step = req_cvtps2xf16_ || use_fp16_instructions_
                         || use_bf16_instructions_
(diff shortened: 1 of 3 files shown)


src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2063,13 +2063,159 @@ void jit_brgemm_matmul_copy_a_transposed_int8_impl_t::generate() {
...
+        MAYBE_UNUSED(vmm_lower);
+
+        const bool is_xf16
+                = everyone_is(conf_->dst_dt, data_type::bf16, data_type::f16);
+
+        switch (dt) {
+            case data_type::f32: uni_vmovups(vmm_in, op); break;
+            case data_type::f16:
+                if (!is_xf16) {
+                    if (is_superset(conf_->isa, avx512_core_fp16)) {
```
xuxinzen: use `uni_vcvtph2psx(vmm_in, op)`

inteldimitrius:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2063,6 +2063,328 @@ void jit_brgemm_matmul_copy_a_transposed_int8_impl_t::generate() {
...
 template struct jit_brgemm_matmul_copy_a_transposed_impl_t<Ymm>;
 
+/** 
+ * @brief Base class for BRGEMM B matrix copy operations
+ * 
+ * This class contains common methods and properties for all copy B kernels.
+ * Now it consists of `load_value` and `decompress_value` and it's considered(TODO: )
+ * to contain all common methods for copy B kernels.
+ */
+struct jit_brgemm_matmul_copy_b_base_t : public jit_brgemm_matmul_copy_b_t,
```
densamoilov: I understand the idea, which looks reasonable to me but the naming seems a bit off. Using `jit_brgemm_matmul_copy_b_base_t` as a name for a derived class when the base class's name is `jit_brgemm_matmul_copy_b_t` looks backwards and illogical.

Maybe it would be a good idea to structure the hierarchy differently since we extend it anyway. 

We already have `jit_brgemm_matmul_copy_b_t` which is the "root class" for all `*copy_b*` classes. We could inherit it from `jit_generator_t` and move all the common interfaces to it making them protected. 
If we do so then all the type specific classes will only have to be derived from `jit_brgemm_matmul_copy_b_t`:
```cpp
template <typename Vmm>
struct jit_brgemm_matmul_copy_b_bf16_t : public jit_brgemm_matmul_copy_b_t{}
```

I see that you didn't use the new `jit_brgemm_matmul_copy_b_base_t` for the classes for integer data types. Does it mean that the new common interfaces are not fully common but rather type specific? 
If it's critical to separate the interfaces common to integral types from those common to floating-point types, we could introduce corresponding classes:(`jit_brgemm_matmul_copy_b_integral_t`/`jit_brgemm_matmul_copy_b_fp_t`), but I can't see a reason why we would really need it, but I might be missing something. 

Anyway, what do you think?

inteldimitrius: Hi, thanks for reviewing!
It's a good point with the naming, for me it's not critical how the "base" class will be named, I will move it to `jit_brgemm_matmul_copy_b_t`. 
The main point here is - all kernels have pretty the same logic which could organized better imho. 
And I did the first step towards it. In my vision ideal scenario is when the specific implementation contains only unique code. 

int8 cases doesn't really differs from others, but we do not support `s8:s4:s8` weight decompression **for now** and they remain untouched. I can do it this PR or in the next.

Would appreciate your ideas about it.

inteldimitrius:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2063,6 +2063,126 @@ void jit_brgemm_matmul_copy_a_transposed_int8_impl_t::generate() {
...
+    jit_brgemm_matmul_copy_b_base_t(const brgemm_matmul_conf_t *conf)
+        : jit_brgemm_matmul_copy_b_t(conf), jit_generator_t(jit_name()) {}
+
+    template <typename Vmm>
+    void apply_shift(const Vmm &input, const Vmm &zp, data_type_t src_dt) {
+
+        if (!conf_->has_zero_point_b || !conf_->with_wei_decompression) return;
+
+        switch (src_dt) {
+            case data_type::s8:
```
xuxinzen: I prefer to use `if...else` instead of `switch `for readability. The same as below.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2063,6 +2063,126 @@ void jit_brgemm_matmul_copy_a_transposed_int8_impl_t::generate() {
...
+            case data_type::f32:
+                // bf16 and f16 already converted into f32 while loading
+                break;
+            default: assert(!"Unsupported source data type for decompression");
+        }
+    }
+
+    template <typename Vmm>
+    void apply_scales(const Vmm &input, const Xbyak::Operand &scale_op) {
+        if (conf_->apply_scales_in_buffer_b) { vmulps(input, input, scale_op); }
```
xuxinzen: nit: use `uni_vmulps()`

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2063,6 +2063,126 @@ void jit_brgemm_matmul_copy_a_transposed_int8_impl_t::generate() {
...
+    * @brief Decompresses values by applying zero point shift, data type conversion, and scaling.
+    * @tparam Vmm Vector register type used for computation
+    * @param input Vector register containing input values to be decompressed
+    * @param zp Vector register containing zero point values
+    * @param scale_op Operand containing scales to be applied
+    * @param src_dt Source data type of the values being decompressed
+    * @param dst_dt Destination data type for the decompressed values
+    */
+    template <typename Vmm>
+    void decompress_value(const Vmm &input, const Vmm &zp,
```
xuxinzen: The three functions have the same name which is confusing to me. It's better rename those three function and use the first one as main body(like rename it as `decompress_value_body()`).

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2167,9 +2175,12 @@ struct jit_brgemm_matmul_copy_b_base_t : public jit_brgemm_matmul_copy_b_t,
                 }
             case data_type::bf16:
                 if (is_xf16) {
-                    // jit_brgemm_matmul_copy_b_bf16 requires
+                    // jit_brgemm_matmul_copy_b_transposed_t requires
                     // another instruction to load
-                    vmovdqu16(vmm_in, op);
+                    if (conf_->transposed_B)
+                        vmovdqu8(vmm_in, op);
```
xuxinzen: why use `vmovdqu8 `to load xf16 with transpose ?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -310,6 +313,10 @@ brgemm_matmul_conf_utils_t::brgemm_matmul_conf_utils_t(
               && one_of(bgmmc.dst_dt, bf16, f32))
     , f16_with_int_wei_dt(weights_decompression_support && bgmmc.src_dt == f16
               && one_of(bgmmc.dst_dt, f16, f32))
+    , f32_with_int_wei_dt(one_of(bgmmc.wei_dt, u8, s8, u4, s4)
+              && everyone_is(f32, bgmmc.src_dt, bgmmc.dst_dt)
+              && !one_of(attr.fpmath_.mode_, fpmath_mode::bf16,
```
xuxinzen: In the offline discussion, f32 with int weights should use `fpmath=strict:true`. `weights_decompression_support` should also be updated, like adding `IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict, bgmmc.src_dt == f32)`.  And the initialization of `f32_with_int_wei_dt` can be simplified.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -244,6 +244,7 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
                             | primitive_attr_t::skip_mask_t::scales_groups
                             | primitive_attr_t::skip_mask_t::
                                     zero_points_data_type
+                            | primitive_attr_t::skip_mask_t::zero_points_groups
```
xuxinzen, line 296: Support for zp with groups should be removed

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -279,29 +279,32 @@
...
 
---reset
 --skip-impl=ref
 --dt=f32:s4:f32,f32:u4:f32
---attr-scales=wei:per_ocic:f32:32x1,wei:per_tensor:f32:32x1
---attr-zero-points=,wei:per_ocic:s4:32x1,wei:common:-1:s4
+--wtag=any,ab,ba
+--attr-scales=,wei:common:0.5:f32,wei:per_ocic:f32:32x1
+--attr-zero-points=,wei:common:2
+# ,wei:per_oc:s4:4096x1 # PER_OC skips for all matmul implemetations
```
xuxinzen: remove this unnecessary tests

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -130,6 +130,8 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             && one_of(wei_dt, s8, u8, s4, u4) && one_of(dst_dt, bf16, f32);
     const bool is_f16_with_int_wei = src_dt == f16
             && one_of(wei_dt, s8, u8, s4, u4) && one_of(dst_dt, f16, f32);
+    const bool is_f32_with_int_wei
```
xuxinzen, line 174: random post: Please clarify the supported ISA. And I only see the extended support for f32 with int weights decompression in this PR. Do you also extend support xf16 with int weights on other ISAs?

tests/benchdnn/inputs/matmul/harness_matmul_decompression
inteldimitrius: Hi, since now there's a grouped ZP support, I have uncommented these lines.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Thanks, removed!

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: There was an idea of polymorphism overloading the same logic for different sets of arguments. I renamed adding details to the name. Thanks!

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Thanks, added!

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Not sure if it really increases readability, but added, thanks!

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Thanks, updated!

inteldimitrius:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: This is part of the feature request - support grouped ZPs and scales.

inteldimitrius:
make test
set test_scope=NIGHTLY

inteldimitrius:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Thanks, added!

inteldimitrius:
make test
set test_scope=NIGHTLY

inteldimitrius:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
enable arch_cpu_adl
enable arch_cpu_spr

inteldimitrius:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
enable arch_cpu_adl
enable arch_cpu_spr

densamoilov (approved):

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1418,6 +1419,29 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    const auto has_wei_zp = !wei_zp.has_default_values();
+
+    if (has_wei_zp) {
+        const auto wei_zp_mask = wei_zp.get_mask();
+        bgmmc.is_wei_zp_per_k = wei_zp_mask & (1 << (bgmmc.ndims - 2));
+        bgmmc.is_wei_zp_per_n = wei_zp_mask & (1 << (bgmmc.ndims - 1));
+        bgmmc.is_wei_zp_common = wei_zp_mask == 0;
+        VCONDCHECK_BG(wei_zp_mask == 0 || bgmmc.is_wei_zp_per_k
+                        || bgmmc.is_wei_zp_per_n,
+                VERBOSE_UNSUPPORTED_ZP_CFG);
```
xuxinzen, line 1443: Do we also support zp w/ group for int8 or only for weights decompression?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -136,6 +136,7 @@ struct brgemm_matmul_conf_t {
     data_type_t reduce_dt;
     data_type_t orig_src_dt;
     data_type_t orig_wei_dt;
+    data_type_t wei_zp_dt;
```
xuxinzen: Do we have any limitation on `wei_zp_dt` when wei_zp_dt == u4/s4? Like, the input of per_oc/per_ic should be divided by 2.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -4951,9 +5035,10 @@ void jit_brgemm_matmul_copy_b_cvt_bf16_t<Vmm>::copy_block(
 
         const bool is_n_tail = ncolumns - n < n_blk_step;
         const bool is_k_tail = nrows - k < k_blk_step;
-        get_scales(blk, k, n, is_n_tail, is_k_tail);
-        decompress_value(src_vmm0, src_vmm1, vmm_zp_b_val, vmm_scales0,
-                vmm_scales1, conf_->orig_wei_dt, conf_->wei_dt);
+        get_wei_scales(blk, k, n, is_n_tail, is_k_tail);
+        get_zero_points(n, is_k_tail);
```
xuxinzen: I saw the `get_zero_points()` only works for `wei_zp_per_n`, not sure if the input should be (n, is_k_tail) or (n, is_n_tail).
And how do we handle `wei_zp_per_k` here?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Hi, int8:int4:int8 is not supported so far. It's going to be delivered in separate PR. 
Other cases are supported with int4/int8 weights and int4/int8/int32 ZP both grouped and non-grouped.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -297,3 +297,36 @@
...
+2x40x256:2x256x64
+7x41x256:1x256x64
+3x96x512:3x512x64
+3x6x512:1x512x62
+
+--skip-impl=ref
+--dt=f32:s4:f32,f32:u4:f32
+--wtag=any,ab,ba
+--attr-scales=,wei:common:0.5:f32,wei:per_ocic:f32:32x1
+--attr-zero-points=,wei:common:2,wei:per_oc:s4:4096x1
```
xuxinzen: For the support of zero point data type, we do not support u4/s4 for common zero points. Not sure if we should support that for per_oc/per_ic/per_ocic

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -557,6 +557,7 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
             return dnnl::impl::utils::one_of(
                     t, dnnl_s4, dnnl_u4, dnnl_s8, dnnl_u8, dnnl_s32);
         };
+
```
xuxinzen: nit: remove unnecessary line

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -225,24 +227,24 @@ struct brgemm_matmul_conf_t {
...
     size_t wei_scales_dt_sz = 0;
-    dim_t wei_scales_k_group_size = 0;
+    bool is_wei_scale_per_n = false;
+    bool is_wei_scale_per_k = false;
+    bool is_wei_scale_common = false;
+    dim_t wei_scales_k_gsize = 0;
     data_type_t wei_scales_dt = data_type::undef;
-    bool extendable_k = false;
+
+    dim_t wei_zp_k_gsize; // per_ic
```
xuxinzen: nit: initialize the variable

dzarukin:
```suggestion
    dim_t wei_zp_k_gsize = 0;
```

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -225,24 +227,24 @@ struct brgemm_matmul_conf_t {
...
+    bool is_wei_zp_per_k = false;
+    bool is_wei_zp_per_n = false;
+    bool is_wei_zp_common = false;
+    data_type_t wei_zp_dt;
```
dzarukin:
```suggestion
    data_type_t wei_zp_dt = data_type::undef;
```

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -292,12 +295,14 @@ brgemm_matmul_conf_utils_t::brgemm_matmul_conf_utils_t(
...
-                      fpmath_mode::any)
+                      fpmath_mode::strict, fpmath_mode::any)
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::f16,
                       bgmmc.src_dt == f16)
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::bf16,
                       bgmmc.src_dt == bf16)
-              && attr.fpmath_.apply_to_int_)
+              && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict,
+                      bgmmc.src_dt == f32)
+              && (attr.fpmath_.apply_to_int_ || bgmmc.src_dt == f32))
```
dzarukin:
```suggestion
              && (attr.fpmath_.apply_to_int_))
```

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -310,6 +315,10 @@ brgemm_matmul_conf_utils_t::brgemm_matmul_conf_utils_t(
               && one_of(bgmmc.dst_dt, bf16, f32))
     , f16_with_int_wei_dt(weights_decompression_support && bgmmc.src_dt == f16
               && one_of(bgmmc.dst_dt, f16, f32))
+    , f32_with_int_wei_dt(weights_decompression_support
+              && everyone_is(f32, bgmmc.src_dt, bgmmc.dst_dt)
+              && !one_of(
+                      attr.fpmath_.mode_, fpmath_mode::bf16, fpmath_mode::f16))
```
dzarukin: It probably would be safer to list supported modes instead.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1378,22 +1398,27 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
-        VCONDCHECK_BG(wei_scales.get_mask() == 0 || bgmmc.is_wei_scale_per_n
+        VCONDCHECK_BG(bgmmc.is_wei_scale_common || bgmmc.is_wei_scale_per_n
                         || IMPLICATION(bgmmc.is_wei_scale_per_k,
                                 bgmmc.with_wei_decompression),
                 VERBOSE_UNSUPPORTED_SCALES_CFG);
+
+        // AVX2 supports f32 scales only
+        VCONDCHECK_BG(
+                IMPLICATION(is_superset(isa, avx2), bgmmc.wei_scales_dt == f32),
+                VERBOSE_UNSUPPORTED_SCALES_CFG);
```
dzarukin, line 1421: I'd say this should disable all decompression with bf16 and f16 because `is_superset(isa, avx2)` is true for any ISA brgemm_matmul supports.
The correct way is to check for exact match: `IMPLICATION(isa == avx2, bgmmc.wei_scales_dt == f32)`.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
dzarukin: Minor: please align names and variable orders with scales for easier future modifications.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
inteldimitrius: I think it should be supported since there's this check https://github.com/uxlfoundation/oneDNN/blob/main/src/common/primitive_attr.cpp#L628
We just hadn't any tests, which could break it.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1402,6 +1427,26 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    if (has_wei_zp) {
+        const auto wei_zp_mask = wei_zp.get_mask();
+        bgmmc.is_wei_zp_per_k = wei_zp_mask & (1 << (bgmmc.ndims - 2));
+        bgmmc.is_wei_zp_per_n = wei_zp_mask & (1 << (bgmmc.ndims - 1));
+        bgmmc.is_wei_zp_common = wei_zp_mask == 0;
+        VCONDCHECK_BG(wei_zp_mask == 0 || bgmmc.is_wei_zp_per_k
+                        || bgmmc.is_wei_zp_per_n,
+                VERBOSE_UNSUPPORTED_ZP_CFG);
+
+        if (bgmmc.is_wei_zp_per_k) bgmmc.wei_zp_k_gsize = wei_zp.get_group(0);
```
dzarukin: Move under `if` above?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1402,6 +1427,26 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+        // Check if K groups for scales and for zero points are identical
+        VCONDCHECK_BG(
+                IMPLICATION(bgmmc.is_wei_zp_per_k && bgmmc.is_wei_scale_per_k,
+                        bgmmc.wei_zp_k_gsize == bgmmc.wei_scales_k_gsize),
+                VERBOSE_UNSUPPORTED_ZP_CFG);
```
dzarukin, line 1449: Is it current implementation limitation? I saw the test case added with zp_g=4096 and scales_g=32, which looks like won't fit this condition...

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -204,18 +206,26 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
     auto check_attr_zero_points = [&]() -> bool {
         const auto &zp = attr()->zero_points_;
         static const std::vector<int> supported_args {
-                DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST};
+                DNNL_ARG_SRC, DNNL_ARG_DST};
```
dzarukin, line 255: Dropping weights from here means if weights zp are specified, the implementation returns `unimplemented`.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -204,18 +206,26 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
...
             if (!zp.has_default_values(arg)) {
                 const int mask = zp.get_mask(arg);
                 if (mask > 0) return false;
             }
         }
+        if (!zp.has_default_values(DNNL_ARG_WEIGHTS)) {
+            const auto mask = zp.get_mask(DNNL_ARG_WEIGHTS);
+            const auto dims = ndims();
+            const auto kn_mask = (1 << (dims - 1)) + (1 << (dims - 2));
+            const bool zp_over_batch = (mask ^ kn_mask);
```
dzarukin: XOR turned out to be very unreliable. I suggest to follow the changes from [this commit](https://github.com/uxlfoundation/oneDNN/commit/62e39afcaf1f7c706d9536541874c966e8b04005).

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: The only limitation for groupsize per_ocic - that it should be divisible by 32 by hardware limitations and this feature is implemented taking it into account.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Grouping over K is implemented a level above, calling copy kernel with specified `k_group_size` and tailing processing.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Because we use `fpmath_mode::strict` and it could **not** be `apply_to_int` by default, I found this solution is best for now.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Yes, it's current limitation.
Added case:
```
--attr-scales=,wei:common:0.5:f32,wei:per_ocic:f32:32x1
--attr-zero-points=,wei:common:2,wei:per_oc:s4:4096x1
```
ZP is **per_oc** and scales are **per_ocic**, this case totally fits the condition, the only `4096x1` is redundant(per_oc mask by default is `Nx1`), I will remove it thanks.

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius:
```cpp
auto check_attr_zero_points = [&]() -> bool {
        const auto &zp = attr()->zero_points_;
        static const std::vector<int> supported_args {
                DNNL_ARG_SRC, DNNL_ARG_DST};
        for (int arg : supported_args) {
            if (!zp.has_default_values(arg)) {
                const int mask = zp.get_mask(arg);
                if (mask > 0) return false;
            }
        }
        if (!zp.has_default_values(DNNL_ARG_WEIGHTS)) {
            const auto mask = zp.get_mask(DNNL_ARG_WEIGHTS);
            const auto dims = ndims();
            const auto kn_mask = (1 << (dims - 1)) + (1 << (dims - 2));
            const bool zp_over_batch = (mask ^ kn_mask);
            const bool mask_ok = (mask & ~kn_mask) == 0;
            return !(zp_over_batch && batch() > 1) && mask_ok;
        }
        return true;
    };
```
`DNNL_ARG_WEIGHTS` are validated outside the loop, since the validation is **not** the same for `src_zp` and `dst_zp`.

It should return true if masks are specified correctly for weights' zero points. I don't see the issue here, if weights zp are specified the computation finishes successfully. Could you please explain your point?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Thanks, fixed!

inteldimitrius:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: UPD: My mistake, fixed. Thanks!

inteldimitrius:
make test
set test_scope=NIGHTLY

inteldimitrius:
Rearranged commits and made it atomic, each commit refers to logic feature.

tests/benchdnn/matmul/matmul.cpp
inteldimitrius: It's already done, thanks!

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: Done, thanks!

inteldimitrius:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
enable arch_cpu_adl
enable arch_cpu_skx
enable arch_cpu_spr

inteldimitrius:
Last update before merge.
Default performance testing doesn't show any performance degradations (copy routines without any weight decompression)
in comparing to the latest main.
[gnr_72C_0.9Ghz_wwmm_wei_decomp_testing.xlsx](https://github.com/user-attachments/files/22948968/gnr_72C_0.9Ghz_wwmm_wei_decomp_testing.xlsx)
[spr_56C_2.2Ghz_wwmm_wei_decomp_testing.xlsx](https://github.com/user-attachments/files/22948969/spr_56C_2.2Ghz_wwmm_wei_decomp_testing.xlsx)

inteldimitrius:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
enable arch_cpu_adl
enable arch_cpu_skx
enable arch_cpu_spr

inteldimitrius:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
enable arch_cpu_adl
enable arch_cpu_skx
enable arch_cpu_spr

PR #3555, opened 2025-07-08
[SynthDNN] Enable synthetic per_tensor, per_ocic cases

kealan-barbieri:
# Description

- Add coverage for `per_ocic`, `per_tensor` scales.
- Add coverage for non-common src scales.

# Checklist

## General


scripts/synthdnn/matmul/sampler.py
```diff
@@ -68,12 +70,17 @@ def supported_scale_type():
...
         out = [None]
         for t in Dequantization.supported_scale_type():
             for p in Dequantization.supported_scale_policy(is_wei):
                 if p == "common":
                     out.append(f"{p}:0.25:{t}")
+                if p == "per_ocic" or p == "per_tensor":
+                    if k_dim > 32 and k_dim % 32 == 0:
+                        out.append(f"{p}:{t}:32x1")
+                    else:
+                        out.append(f"{p}:{t}:{k_dim}x1")
```
rjoursler: Would it be reasonable to do something more like
```
group_k = random.choice(get_group_divisors(k_dim));
out.append(f"{p}:{t}:{group_k}x1"
```

kealan-barbieri: yes, much better and more diverse cases, refactored.

PR #3558, opened 2025-07-09
xe: ocl: simplify post op calculation

rjoursler:
Simplifying the post-op calculation enables better optimization by the compiler. For example, this simplification eliminates register spills that occur in the reference group normalization kernel.


rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

kealan-barbieri (approved):

echeresh (approved):

PR #3563, opened 2025-07-09
[GPU] Fixup fp8 weights decomp handling

kealan-barbieri:
# Description

Use int8 upconversion strategies for fp8 upconversion rather than plain f16/bf16, fixes out of register issues as conversion overhead is similar.

# Checklist

## General

### DIFF
--- a/src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
+++ b/src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
@@ -560,7 +560,7 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
     add_mode_matches(fpmath_bf16, [](Type dt) -> const char * {
         if (dt == Type::f32) { return "[SB]"; }
         if (dt.isInt8() || dt.isInt4()) return "[OB]";
-        if (dt.isF8()) return "B";
+        if (dt.isF8()) return "[OB]";
         if (dt.isF4()) return "F";
         return nullptr;
     });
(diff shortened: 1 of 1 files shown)


src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp
```diff
@@ -560,15 +560,15 @@ status_t gen_gemm_nocopy_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
...
+        if (dt.isF8()) return "[OB]";
         if (dt.isF4()) return "F";
         return nullptr;
     });
 
     add_mode_matches(fpmath_f16, [](Type dt) -> const char * {
         if (dt == Type::f32) { return "[SH]"; }
         if (dt.isInt8() || dt.isInt4()) return "[OH]";
-        if (dt.isF8()) return "H";
+        if (dt.isF8()) return "[OH]";
```
echeresh, line 571: How does this work?

Before, with `H` - I assume there is some logic later on to convert it to something like `[QH]` to match `f8`. But what happens with `[OH]`? What should we have in the strategy to match it?

kealan-barbieri: I expected this to be handled the same, via https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/jit/gemm/gen_gemm_kernel.cpp#L604 and https://github.com/uxlfoundation/oneDNN/blob/d5a2b6306d2d05d5a84c6f6468a6dbcb6558d52c/src/gpu/intel/jit/gemm/include/gemmstone/problem.hpp#L293, however on review its actually not reusing the strategies with eg `[OH]` and is going directly to strategies with `Q`, will review and resubmit.

PR #3564, opened 2025-07-09
cache: fix complex primitives creation from the blob

dzarukin:
[MFDNN-13804](https://jira.devtools.intel.com/browse/MFDNN-13804)

All the important information may be found in the comments.

There's also a patch for benchdnn coming, that would allow to reproduce creation_from_blob issues reliably without OV reproducers, but it requires more love to hit production.


dzarukin:
make test

src/gpu/intel/compute/kernel.hpp
```diff
@@ -240,10 +244,31 @@ class kernel_t {
         return impl_->dump();
     }
 
+    // A `tag` may be provided by the user to differentiate the source of the
+    // kernel. In particular, it may come from the blob, or it could be
+    // properly generated.
+    void hash_dump(impl::engine_t *engine, const char *tag = nullptr) const {
+        if (get_verbose_dev_mode(verbose_t::debuginfo) >= 6) {
+            printf("[%s] %s -> %u\n", tag ? tag : "unlabeled", name().c_str(),
+                    get_hash(engine));
```
echeresh: To be able to grep to find the source of logging (and for more context):

```suggestion
            printf("kernel creation [%s] %s -> %u\n", tag ? tag : "unlabeled", name().c_str(),
                    get_hash(engine));
```

dzarukin: Sure, added, thanks.

dzarukin:
make test

TaoLv (approved):

src/gpu/intel/compute/kernel.hpp
```diff
@@ -240,10 +244,31 @@ class kernel_t {
...
+            fflush(stdout);
+        }
+    }
+
     std::string name() const { return impl_->name(); }
 
 private:
     std::shared_ptr<kernel_impl_t> impl_;
+
+    uint32_t get_hash(impl::engine_t *engine) const {
```
rjoursler: Why not return a 64-bit hash like most other hashing functions in the library?

src/common/primitive_desc_iface.cpp
```diff
@@ -95,7 +95,11 @@ status_t dnnl_primitive_desc::create_primitive_iface(
         const cache_blob_t &cache_blob) const {
     // Step 1: create impl::primitive_t or get it from primitive cache
     std::pair<std::shared_ptr<primitive_t>, cache_state_t> p;
-    auto status = impl()->create_primitive(p, engine(), cache_blob);
+    // Top level primitive can be fetched from the primitive cache since it's
+    // fetching it faster.
```
rjoursler, line 99: The same will hold true for any duplicated nested primitives. Could we just deduplicate the (non-nested) primitives in the cache blob instead? This would speed up object creation and also reduce the size of the cache blob.

echeresh: > The same will hold true for any duplicated nested primitives. Could we just deduplicate the (non-nested) primitives in the cache blob instead?

I think we could but looks like the practical impact would be really small. Do we have such cases with duplicated nested primitives at all?

rjoursler: > Do we have such cases with duplicated nested primitives at all?

This is a fairly common occurrence in the GPU RNN implementation which instantiates up to 12 GEMM kernels. 

> I think we could but looks like the practical impact would be really small.

Agreed, to some extent, it seems to me like nested primitive creation from cache blobs is more complicated than necessary. The fact that we encountered the bug fixed in this PR is just a symptom of this complexity. I am partially wondering if we can use an algorithmic change to simplify the process and which happens to run faster due to the reduced complexity.

echeresh: > This is a fairly common occurrence in the GPU RNN implementation which instantiates up to 12 GEMM kernels.

But can it be that a typical RNN case has no duplication between nested GEMM primitives? For example, the RNN case from the Jira has two GEMM primitives - both unique.

As an observation, persistent cache blob seems to duplicate kernels coming from the same program (by adding the whole program binary multiple times). My bet is that is the first thing to fix to reduce overhead and blob size as RNN uses a lot of small kernels from the same program. But this is really out of scope for this PR and Jira.

> The fact that we encountered the bug fixed in this PR is just a symptom of this complexity. 

Well, we have three separate cache mechanisms (kernel cache, primitive cache, persistent cache) + caching with nesting. All of this have to interact properly. It'd be more surprising if we wouldn't step into something like that.

rjoursler: > It'd be more surprising if we wouldn't step into something like that.

That is not quite how I see it. In this case, we are artificially coupling runtime caches and the cache blob by using a streaming data structure. We could avoid this by encoding some meta data in the cache blob so that we no longer rely on streaming the data. If we did this, cache blob deserialization could effectively become:
```C++
for(primitive_blob: decode_nested_primitive_header(cache_blob)) {
      create_primtive_from_blob(primitive_blob);
}
```
After this change, we cannot skip a piece of the `cache_blob` while deserializing it and the coupling with primitive/kernel cache is obviated.

echeresh: I think the details are important here. How to define meta data? Primitive blob IDs are large while OpenCL kernels do not have good IDs.

Streaming is conceptually simple, and I guess that was the reason to implement it.

rjoursler: > How to define meta data? Primitive blob IDs are large while OpenCL kernels do not have good IDs.

I don't think we need to worry about the concept of ID's, we just need to encode the size of the nested objects so that we can iterate over it in the cache blob.  The main issue that I see is that we need to somehow encode the logic for where to store the created primitive in the resulting object. If we just require all nested primitives be directly stored in a structure, we could do:
```
struct header_entry_t {
     entry_kind_t kind
     size_t struct_offset;
     size_t size;
};
struct nested_primitive_header_data {
    std::vector<header_entry> entries;
}
```
and then implement a simple iterator + dispatching to decode the cache blob in a type erased context using the information from the header.

src/gpu/intel/compute/kernel.hpp
dzarukin: Updated, thanks for a suggestion.

PR #3572, opened 2025-07-10
feat: add JIT asimd path for table-free eltwise algorithms

micyan01:
# Description

This commit introduces a JIT execution path for eltwise functions that do not require approximate method and table lookups, such as ReLU, Abs, Square, and so on.

- Adds asimd eltwise method for table-free algorithms
- Introduces asimd and sve-specific helper specializations for better ISA modularity
- Supports f32, bf16, and f16 data types

# performance report

### ✅ ReLU Latency Comparison (Shape = 4096x4096)

| Implementation        | FP32 Latency (ms) | Δ vs FP32 c7g sve main | FP16 Latency (ms) | Δ vs FP16 c7g sve main | BF16 Latency (ms) | Δ vs BF16 c7g sve main |
|------------------------|-------------------|------------------|--------------------|------------------|--------------------|------------------|
| **c7g sve main**       | 0.564             | 0.00%            | 0.961              | 0.00%            | 0.422              | 0.00%            |
| c7g sve feature        | 0.564             | +0.00%           | 0.962              | +0.10%           | 0.426              | +0.95%           |
| c7g jit:asimd          | 0.558             | −1.06%           | 0.414              | −56.92%          | 0.267              | −36.73%          |
| c6g asimd              | 0.858             | +52.13%          | 0.846              | −11.97%          | 0.453              | +7.35%           |


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2398,8 +2398,139 @@ void jit_uni_eltwise_injector_f32<isa>::register_table_entries() {
     }
 }
 
-// We only need sve_128 as the injector is fully vector length agnostic
+template <>
+void jit_uni_eltwise_injector_f32<asimd>::relu_zero_ns_compute_vector_fwd(
+        const TRegS &vmm_src) {
+    TRegS vmm_tmp(vmm_src.getIdx() + 3);
```
jondea: Can we use one of the aux register variables rather than constructing our own, which might clobber a register in the host primitive? Likewise in `relu_compute_vector_fwd`

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2398,8 +2398,139 @@ void jit_uni_eltwise_injector_f32<isa>::register_table_entries() {
...
+    h->fadd(vmm_src, vmm_src, vmm_aux0);
+    h->fminnm(vmm_src, vmm_src, vmm_tmp1);
+    h->fmaxnm(vmm_src, vmm_src, vmm_tmp0);
+}
+
+template <>
+void jit_uni_eltwise_injector_f32<asimd>::hardswish_compute_vector_fwd(
+        const TRegS &vmm_src) {
+    // result = x * hardsigmoid(x)
+    h->mov(VReg16B(vmm_aux1.getIdx()), VReg16B(vmm_src.getIdx()));
```
jondea: Can this be shortened to
```
h->mov(vmm_aux1, vmm_src);
```

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -229,10 +233,115 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
     std::unique_ptr<jit_uni_eltwise_injector_f32<isa>> eltwise_injector_;
 
     PReg p_tmp0 {4}; /* Index is temporal. */
-    PReg pg_s {5};
-    PReg pg_h {7};
+
+    // Template vector load/store interface
+    void load_vector(TRegS &dst, const XReg addr);
+    void store_vector(const XReg &addr, const TRegS src);
+    void unpack_bf16(TRegS &dst0, TRegS &dst1, TReg src0);
```
jondea: These use a mixture of type sized and register sized variables (`TRegS` and `Treg` respectively). Can we make this consistent?

Also, some comments here on exactly what pack/unpack means would be useful

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
micyan01: Yes, we need to leverage aux_vecs_count to automatically manage the number of auxiliary vector registers required. This helps avoid manual allocation conflicts and potential register overlap issues. The patch of the latest versio has fixed this problem.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
micyan01: we can't optimize more, because xbyak only support 16b/8b input.

src/cpu/aarch64/jit_uni_eltwise.cpp
micyan01: You’re right — unifying the interface makes sense. I’m currently using TReg as the unified input type, and internally I select between .h or .s as needed based on context.

jondea:
@dzarukin would you be able to take a look for @uxlfoundation/onednn-arch please? We still plan to remove the datatype template, but not in this PR

PR #3579, opened 2025-07-11
xe: jit: workaround hardware read suppression bug 

rjoursler:
Fixes [MFDNN-13917](https://jira.devtools.intel.com/browse/MFDNN-13917).

### DIFF
--- a/src/gpu/intel/compute/device_info.cpp
+++ b/src/gpu/intel/compute/device_info.cpp
@@ -22,6 +22,7 @@
 #include "gpu/intel/compute/device_info.hpp"
 #include "gpu/intel/jit/utils/ngen_type_bridge.hpp"
 #include "gpu/intel/utils.hpp"
+#include "ngen.hpp"
 
 #ifdef DNNL_WITH_SYCL
 #include "gpu/intel/sycl/engine.hpp"
@@ -63,9 +64,14 @@ uint64_t get_future_extensions(
     return extensions;
 }
 
+int device_info_t::stepping_id() const {
+    ngen::Product p = jit::get_ngen_product(*this);
+    return p.stepping;
+}
+
 bool device_info_t::is_integrated() const {
-    auto family = static_cast<ngen::ProductFamily>(gpu_product_family_);
-    return ngen::getPlatformType(family) == ngen::PlatformType::Integrated;
+    ngen::Product p = jit::get_ngen_product(*this);
+    return p.type == ngen::PlatformType::Integrated;
 }
 
 std::string device_info_t::get_cl_ext_options() const {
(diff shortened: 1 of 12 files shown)


rjoursler:
make test
disable test_device_cpu

dyoussif (approved):

rjoursler:
make test perf-gpu

src/gpu/intel/compute/device_info.cpp
```diff
@@ -296,8 +302,7 @@ status_t device_info_t::init_serialized_device_info(
     }
 
     serialized_device_info_.append(gpu_arch_);
-    serialized_device_info_.append(gpu_product_family_);
-    serialized_device_info_.append(stepping_id_);
```
kealan-barbieri, line 300: wouldnt it be simpler to just query using `ngen::product` for stepping and family rather than storing the anonymous struct?

src/gpu/intel/compute/device_info.cpp
```diff
@@ -379,6 +383,11 @@ void device_info_t::fixup_l3_cache_size() {
     }
 }
 
+static_assert(std::is_trivially_copyable<ngen::Product>(),
+        "ngen::Product cannot safely be copied into gpu_product_t");
+static_assert(sizeof(ngen::Product) == sizeof(compute::gpu_product_t),
```
kealan-barbieri, line 388: what is the additional utility we get from hanging on to the anonymous ngen product struct rather than just querying on demand for stepping and family?

kealan-barbieri (approved):

src/gpu/intel/compute/device_info.hpp
```diff
@@ -38,6 +38,12 @@ namespace compute {
 
 enum class gpu_arch_t { unknown, xe_lp, xe_hp, xe_hpg, xe_hpc, xe2, xe3 };
 
+// Memory for storing ngen::Product to avoid directly including nGEN because of
+// header dependencies outside of src/gpu/intel.
+struct alignas(int) gpu_product_t {
+    unsigned char data[12];
+};
```
echeresh, line 45: A comment for the future: I feel like we can add more isolation to `compute::device_info_t` to ensure its definition is not used outside of `gpu::intel` code. Then we can just use nGEN types here directly.

But IMO, this is not worth it until we touch some related code.

echeresh (approved):

src/gpu/intel/compute/device_info.hpp
rjoursler: I already did that work in #3428, it's just that PR is too complex for backporting and needs approvals from too many groups.

src/gpu/intel/compute/device_info.cpp
rjoursler: Querying for the device product can be an expensive operation, in the worst case, [it could require compiling an OpenCL kernel](https://github.com/uxlfoundation/oneDNN/blob/1c44732b492b130fec61d5401d65630310cd72cb/third_party/ngen/ngen_opencl.hpp#L279).

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

PR #3588, opened 2025-07-11
xe: ocl: avoid ref_gnorm fwd register spills

rjoursler:
Attempt to fix the eternal register spill warnings emitted by ref_gnorm_fwd.

### DIFF
--- a/src/gpu/intel/ref_group_normalization.cl
+++ b/src/gpu/intel/ref_group_normalization.cl
@@ -47,7 +47,7 @@
  * @param [in]      src_scale Input data scale [1]
  * @param [in]      dst_scale Output data scale [1]
  * @param [in]      eps       User defined constant
- * 
+ *
  * POST_OP_ARGS Expands to ', const __global POST_OP_TYPE *po_0_binary_arg' where POST_OP_TYPE
  * defined by post operation type API
  *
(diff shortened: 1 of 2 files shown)


rjoursler:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_gnorm

rjoursler:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_gnorm

rjoursler:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_gnorm

Simonsays095 (commented):
I think both solutions (converting to a runtime parameter, and limiting the unroll factor) would work independently. Why not just limit the unroll and avoid the overhead of converting to a runtime parameter?

Also - a comment by the loop indicating why the loop unroll has to be restricted would be nice. Maybe also indicate if another small (yet limited) loop unroll would also work. I'm not sure how quickly register spilling happens.

rjoursler:
> Why not just limit the unroll and avoid the overhead of converting to a runtime parameter?

This was my first solution, it turns out the unroll hint will get ignored on XeHPC and still trigger the register spill...

> I'm not sure how quickly register spilling happens.

I have seen spilling for any value larger than 1. In general as this is the outer loop, unrolling should largely be irrelevant on real word workloads.

atkassen:
Can you also run CI on Xe3? It tends to be even more prone to register spills.

rjoursler:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_gnorm
enable arch_gpu_xe3-lpg

atkassen (approved):

rjoursler:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_gnorm
enable arch_gpu_xe3-lpg

rjoursler:
@atkassen, you were right about Xe3 still producing warnings. In particular, there are some warnings against the bwd kernel which isn't touched by this PR. I don't intend to address those right now, but hopefully we can resolve them in the near future.

echeresh:
@rjoursler Maybe we should just implement filtering for such warnings - for reference kernels it's not an issue. Even if fixed, they can get back at some point - there is no much value to spend time on that, IMO.

PR #3591, opened 2025-07-14
Revert "cpu: aarch64: Support for int8 in BRGEMM General and 1x1 Forward Convolution"

Ryo-not-rio:
Reverts uxlfoundation/oneDNN#3552, which is causing a nightly unit test failure: https://github.com/uxlfoundation/oneDNN/actions/runs/16258443419/job/45898983634


Ryo-not-rio:
Waiting to merge this for a few days to see if we can get a fix this

kasturedeeksha:
Hi
I checked the testcase:

```
--conv --dt=u8:s8:f32 --skip-impl=ref,x64:gemm mb1ic64ih1iw33oc1oh1ow33kh1kw24ph0pw23n"l_pad_exceeds_ow_block"
```

in relation to the INT8 BRGEMM convolution PR:
[https://github.com/uxlfoundation/oneDNN/pull/3552](https://github.com/uxlfoundation/oneDNN/pull/3552)

**Findings:**

* Even before this PR, the testcase would fail when executed with the reference implementation (`gemm_s8u8s32:ref`). However, since the testcase explicitly skips the reference and `x64:gemm` implementations (`--skip-impl=ref,x64:gemm`), it was not executed previously and so did not surface.
* With this PR, the testcase is now executed but fails. The failure arises because the output is being compared against the `sve_jit_convolution` kernel, which itself produces inaccurate results. As a result, both the reference int8 and BRGEMM int8 kernels fails due to the comparison against this.

To verify this, I commented out the `sve_jit_convolution` kernel in cpu_convolution_list . After doing so, all INT8 testcases pass, as their outputs are then compared directly against the correct reference implementation.

This indicates that the root cause of the issue lies within the `jit_sve_convolution f32` kernel. Therefore, there is **no fix required in the INT8 convolution implementation** as part of this PR.


**Query:**
I also had a small question regarding the current testing approach:

* Why are we comparing the output of the INT8 kernel against the `sve_jit_convolution` or any other  implementation instead of directly against the `f32` reference?
* Additionally, are there tests in place that validate the `sve_jit_convolution` kernel output against the `f32` reference for this same testcase? 

In summary, the observed failures are due to inaccuracies in the jit_sve f32 kernel, and there is no issue with the INT8 convolution implementation itself. Thus, no changes are required for INT8 as part of this PR.

Thus to fix this issue we need to add fix in jit_sve_convolution

dzarukin:
Hi @kasturedeeksha,
> Why are we comparing the output of the INT8 kernel against the sve_jit_convolution or any other implementation instead of directly against the f32 reference?

For performance reasons, it's times faster per test case, you may imagine the impact on a large batch file.

> Additionally, are there tests in place that validate the sve_jit_convolution kernel output against the f32 reference for this same testcase?

It's likely not the case, otherwise, the failure would pop up. f32 problems are always validated with f32 reference, no primitives are used for that. You may extend the regression list to have one.

Ryo-not-rio:
@kasturedeeksha thank you for your analysis, we will have a look at jit_sve_convolution, we may just decide to disable it for u8:s8:f32 test cases

kasturedeeksha:
@dzarukin Thanks for the clarification. I think we should include the shapes used for other datatypes like INT8 testing in the 32-bit tests as well for better coverage and accurate testing of other datatypes.
@Ryo-not-rio Thanks for the update. Disabling jit_sve_convolution(f32) validation for u8:s8:f32 cases sounds like a good interim step till the issue is fixed. After doing so we will not require this fix https://github.com/uxlfoundation/oneDNN/pull/3597.

PR #3614, opened 2025-07-16
src: gpu: intel: sycl: do not wait on graph recording

hidefromkgb:
This addresses [MFDNN-13947](https://jira.devtools.intel.com/browse/MFDNN-13947).

### DIFF
--- a/src/gpu/generic/ref_sum.hpp
+++ b/src/gpu/generic/ref_sum.hpp
@@ -176,9 +176,6 @@ struct ref_sum_t : public gpu::primitive_t {
             nested_scratchpad_t ns(ctx, key_nested_multiple + i, reorders_[i]);
             r_ctx.set_scratchpad_grantor(ns.grantor());
             CHECK(reorders_[i]->execute(r_ctx));
-#ifndef DNNL_SYCL_CUDA
-            CHECK(ctx.stream()->wait());
-#endif
         }
 
         if (pd()->need_output_reorder()) {


src/xpu/sycl/stream_impl.hpp
```diff
@@ -49,8 +49,13 @@ class stream_impl_t : public impl::stream_impl_t {
...
     ::sycl::queue *queue() { return queue_.get(); }
 
+    bool recording() {
+        return queue()->ext_oneapi_get_state()
+                == ::sycl::ext::oneapi::experimental::queue_state::recording;
+    }
+
     status_t wait() {
-        queue()->wait_and_throw();
+        if (!recording()) queue()->wait_and_throw();
```
echeresh: I checked - I see an explicit `wait()` here which is likely the culprit:
https://github.com/uxlfoundation/oneDNN/blob/2bd6e963aa7086b612762e59ebcacabc2a388b68/src/gpu/generic/ref_sum.hpp#L180

I have two questions/comments:

1. Is it needed for correctness? If yes - we can't just skip it while recording. If not - we need to remove it from the implementation
2. Given the above, the diagnostics from the Jira are actually useful - it helps to detect `wait()` calls that are not needed (or maybe needed but need to be refactored for SYCL Graph support)

Given 2 - we don't need to skip `wait()` with SYCL Graph - it's useful. For 1 - we need to investigate further. At first glance, `wait()` can be just dropped - but need to test.

petercad: @echeresh, @hidefromkgb -- If I'm reading it right the wait in ref_sum is only there to ensure correctness for out-of-order queues. I think it can be replaced by a `compute_stream_t::barrier()`.

echeresh: @petercad I think this should be handled by internal tracking of events, e.g. here:
https://github.com/uxlfoundation/oneDNN/blob/94b83cc99e587a41dd4e78cd477a5ecaf99ceab5/src/gpu/intel/sycl/interop_kernel.cpp#L182

There are several similar multi-kernel primitives (e.g. RNN or GEMM + post-ops) - we generally support out-of-order queues without explicit waiting or barriers.

dzarukin: ref_sum was the one implementation that suffered from parallel creation under benchdnn on Linux due to a driver issue. It might not be relevant any longer given it's been 2 or 3 years.

hidefromkgb: [1]: It most likely isn't on Intel GPUs, certainly isn't on NVidia, but might be on AMD — but this is speculative, for I haven't run an exhaustive CI on it yet.
[2]: SYCL explicitly tells us not to do that:
> what():  wait cannot be called for a queue which is recording to a command graph.

hidefromkgb:
Added a test commit to cancel all previous changes and remove the `wait()` instead.
The custom CI run for this commit is [here](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f0669a27-87e7-f122-8e19-a4bf010d0e2d) — featuring all Intel and AMD GPUs.

echeresh (approved):

PR #3584, opened 2025-07-11
cpu: gnorm bwd fix

raistefintel:
This issue arose from discrepancies observed when comparing the source gradients produced by PyTorch and oneDNN during backward propagation in group normalization. It was identified that the previous implementation in oneDNN did not incorporate group statistics in the gradient calculations, leading to misalignment with PyTorch's outputs. With the recent updates, the implementation now correctly includes group statistics, ensuring that the outputs are consistent with those generated by PyTorch.

Fixes #3543

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [ ] Have you formatted the code using clang-format?

### Bug fixes

- [x] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [ ] Have you added relevant regression tests?

PR #3615, opened 2025-07-16
cpu, benchdnn: gnorm: corrected bwd compute formulae

### LINKED ISSUE #3543: group_normalization_backward produces incorrect gradients for diff_src
**Describe the bug**

The backward pass of `group_normalization` in oneDNN produces incorrect results for `diff_src`, even when `mean`, `variance`, `diff_dst`, `scale`, and `shift` are all correctly provided. This has been tested against both:

* PyTorch's `GroupNorm(...).backward()` output
* A manual implementation using the analytical backward formula

In both reference implementations, the computed `diff_src` matches perfectly. The oneDNN output differs significantly, especially when `scale` is not all-ones.

---

**To Reproduce**

1. Input: shape `[2, 6, 4]` (interpreted as NCH with H = 4)
2. `G = 3` groups (so group size = 2 channels per group)
3. Forward and backward performed using oneDNN v3.9
4. `group_normalization_forward` with `prop_kind::forward_training`, `use_scale_shift` flag
5. `group_normalization_backward` using same descriptors, providing:

   * `src`
   * `mean`, `variance` (from forward)
   * `scale_shift`
   * `diff_dst`
   * all memory descriptors aligned and correct
6. `diff_src` output diverges from reference implementation

---

**Expected behavior**

`diff_src` should match PyTorch’s `GroupNorm(...).backward()` output within numerical tolerance (1e-5 to 1e-6). It currently does not, even when using the same inputs, saved stats, and layout.

---

**Additional context**

* Removing `dnnl_use_global_stats` did not fix the issue
* `scale_diff` and `shift_diff` match PyTorch exactly — only `diff_src` is wrong
* The reference formula used was:

  $$
  \frac{\partial L}{\partial x} = \frac{\gamma}{\sqrt{\sigma^2 + \epsilon}} \cdot \left(
    dy - \frac{1}{M} \sum dy - \hat{x} \cdot \frac{1}{M} \sum (dy \cdot \hat{x})
  \right)
  $$

  where $M$ is the number of elements per group

---

**System Info**

* oneDNN Version: 3.8.1 (confirmed backward is implemented)
* OS: Windows 11 
* CPU: Intel Core i5 11th Gen
* Tested in both single-threaded and multithreaded modes

dzarukin:
Due to implementation difference caused by collecting statistics over spatial and channels in groups but scales over channels, the formulae are different from lnorm ones that were used.

TODO: would be nice to update formulae in the documentation...

This is a minor upgrade over #3584 (which I don't have an access to update de to a fork nature). Kept the original author's commit signature.

Closes #3543.

### DIFF
--- a/tests/benchdnn/gnorm/gnorm.cpp
+++ b/tests/benchdnn/gnorm/gnorm.cpp
@@ -500,6 +500,16 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
 
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_data_type({prb->dt[0], prb->dt[1]}, prb->dir, res);
+
+    if ((is_gpu() || is_generic_gpu()) && (prb->dir & FLAG_BWD)) {
+        BENCHDNN_PRINT(2,
+                "[SKIP][%s:%d]: The implementation doesn't return the correct "
+                "result.\n",
+                __FILE__, __LINE__);
+        res->state = SKIPPED;
+        res->reason = skip_reason::case_not_supported;
+        return;
+    }
 }
 
 void skip_invalid_prb(const prb_t *prb, res_t *res) {
--- a/tests/benchdnn/gnorm/ref_gnorm.cpp
+++ b/tests/benchdnn/gnorm/ref_gnorm.cpp
@@ -57,19 +57,23 @@ void compute_ref_fwd(const prb_t *prb, const args_t &args) {
         float sqrt_var = sqrtf(svar + prb->eps);
         float rcp_denom = 1.f / sqrt_var;
 
-        for_(int64_t c = prb->get_c_start(g); c < prb->get_c_start(g + 1); ++c)
-        for_(int64_t d = 0; d < D; ++d)
-        for_(int64_t h = 0; h < H; ++h)
-        for (int64_t w = 0; w < W; ++w) {
-            auto off = data_off(prb, n, c, d, h, w);
-            float x_hat = (src.get_f32_elem(off) - smean) * rcp_denom;
+        for (int64_t c = prb->get_c_start(g); c < prb->get_c_start(g + 1);
+                ++c) {
             float gamma = use_sc ? sc.get_f32_elem(c) : 1.f;
             float beta = use_sh ? sh.get_f32_elem(c) : 0.f;
-            float res = gamma * x_hat + beta;
-            const auto v_po_vals = prepare_po_vals(dst, args, v_po_masks, off);
-            res *= src_scale_val;
-            maybe_post_ops(prb->attr, res, 0.f, v_po_vals);
-            dst_ptr[off] = res * r_dst_scale_val;
+
+            for_(int64_t d = 0; d < D; ++d)
+            for_(int64_t h = 0; h < H; ++h)
+            for (int64_t w = 0; w < W; ++w) {
+                auto off = data_off(prb, n, c, d, h, w);
+                float x_hat = (src.get_f32_elem(off) - smean) * rcp_denom;
+                float res = gamma * x_hat + beta;
+                const auto v_po_vals
+                        = prepare_po_vals(dst, args, v_po_masks, off);
+                res *= src_scale_val;
+                maybe_post_ops(prb->attr, res, 0.f, v_po_vals);
+                dst_ptr[off] = res * r_dst_scale_val;
+            }
         }
     });
 }
@@ -97,10 +101,10 @@ void compute_ref_bwd(const prb_t *prb, const args_t &args) {
     const int C_PER_G = C / G;
     const float CSP = C_PER_G * D * H * W;
 
+    // Scale and shift are computed over a channel, thus, accumulate diff_dst
+    // values over the spatial only.
     benchdnn_parallel_nd(C, [&](int64_t c) {
         int64_t g = c / C_PER_G;
-
-        float gamma = use_sc ? sc.get_f32_elem(c) : 1.f;
 
         float d_gamma = 0;
         float d_beta = 0;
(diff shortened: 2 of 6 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_gnorm

tests/benchdnn/gnorm/ref_gnorm.cpp
```diff
@@ -123,26 +123,69 @@ void compute_ref_bwd(const prb_t *prb, const args_t &args) {
...
+    // scales and statistics are applied to the dst point. Since scales are
+    // applied per channel, a part of derivative coming from the mean and
+    // variance over different channels will be different, this is the primary
+    // reason why accumulation is needed.
+    //
+    // y(c) = [(x(c) - m) / v] * gamma(c) + beta;
+    // dy/dx = gamma(c) * [d(x(c) - m)/dx * v - (x(c) - m) * dv/dx] / v^2
+    // `sum_dd_scaled` covers `d(x(c) - m)/dx` part;
+    // `sum_dd_snorm` covers a part coming from `dv/dx`;
+    // Rest values are computed at `Apply gradients` part.
```
mzhukova:
```suggestion
    // y(c) = [(x(c) - m) / v] * gamma(c) + beta(c)
    // dy/dx = gamma(c) * [d(x(c) - m)/dx * v - (x(c) - m) * dv/dx] / v^2
    // `sum_dd_scaled` covers `d(x(c) - m)/dx` part
    // `sum_dd_snorm` covers a part coming from `dv/dx`
    // The rest of the values are computed at `Apply gradients` part
```

tests/benchdnn/gnorm/ref_gnorm.cpp
```diff
@@ -123,26 +123,69 @@ void compute_ref_bwd(const prb_t *prb, const args_t &args) {
...
+        if (!glob_stats) {
+            for_(int64_t c = prb->get_c_start(g); c < prb->get_c_start(g + 1);
+                    ++c)
             for_(int64_t d = 0; d < D; ++d)
             for_(int64_t h = 0; h < H; ++h)
             for (int64_t w = 0; w < W; ++w) {
                 auto off = data_off(prb, mb, c, d, h, w);
                 float dd = d_dst.get_f32_elem(off);
-                float ds = dd;
+                float gamma = use_sc ? sc.get_f32_elem(c) : 1.f;
```
mzhukova, line 162: seems like this could be moved to outer `for_(int64_t c = ...)` loop?

tests/benchdnn/gnorm/ref_gnorm.cpp
```diff
@@ -123,26 +123,69 @@ void compute_ref_bwd(const prb_t *prb, const args_t &args) {
...
+                float x_hat = (src.get_f32_elem(off) - smean) * rcp_denom;
 
-                if (!glob_stats) {
-                    float x_hat = (src.get_f32_elem(off) - smean) * d_gamma;
-                    ds -= (d_beta + x_hat * rcp_denom) / CSP;
-                }
+                float dd_scaled = dd * gamma;
+                sum_dd_scaled += dd_scaled;
+                sum_dd_snorm += dd_scaled * x_hat;
```
mzhukova: (minor question) why `_snorm`? I couldn't quite figure the naming out.. would something like `_scaled_x_hat` be a bit more clear?

tests/benchdnn/gnorm/gnorm.cpp
```diff
@@ -500,6 +500,16 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
...
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_data_type({prb->dt[0], prb->dt[1]}, prb->dir, res);
+
+    if ((is_gpu() || is_generic_gpu()) && (prb->dir & FLAG_BWD)) {
+        BENCHDNN_PRINT(2,
+                "[SKIP][%s:%d]: The implementation doesn't return the correct "
+                "result.\n",
+                __FILE__, __LINE__);
+        res->state = SKIPPED;
+        res->reason = skip_reason::data_type_not_supported;
```
mzhukova: maybe unsupported case or invalid case would be better match?

tests/benchdnn/gnorm/gnorm.cpp
```diff
@@ -500,6 +500,16 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
 
 void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
     skip_unimplemented_data_type({prb->dt[0], prb->dt[1]}, prb->dir, res);
+
+    if ((is_gpu() || is_generic_gpu()) && (prb->dir & FLAG_BWD)) {
```
mzhukova, line 504: just for my own understanding, what is the case when we would need "either `is_gpu()` or `is_generic_gpu()`" check?

dzarukin:
> I've also noticed a todo wrt documentation update, however AFAIK we don't put explicit formulas for backward for any of the normalizations, could you elaborate?

We don't but I think we should start doing it because it would be a big challenge to understand what the library does. But I didn't care more to put them...

> Should we also supply a documentation/release note saying that gnorm backward is incorrect before v3.10 and that the result would be significantly different?

It's incorrect only when scales are requested. Without scales it would be correct. I don't know if it's worth mentioning as we got the bug report just now, past two years the code was there.

mzhukova:
> > Should we also supply a documentation/release note saying that gnorm backward is incorrect before v3.10 and that the result would be significantly different?
> 
> It's incorrect only when scales are requested. Without scales it would be correct. I don't know if it's worth mentioning as we got the bug report just now, past two years the code was there.

Right, but user don't necessarily compare one implementation to another, however they could see a noticeable difference in a "convergence rate" or whatnot, so I was just wondering if it might be worth mentioning, maybe at least in the Known Limitations section for v3.9 (and then I guess the fix would be mentioned in v3.10 release notes as well)..?

tests/benchdnn/gnorm/gnorm.cpp
dzarukin: Thanks for the suggestions and catches, I've applied them all.

`is_gpu()` is considered a common ground of all GPUs but rather considered as Intel GPU, while generic is a special vendor free build. I anticipate Intel Gpu get an update and generic will be broken for a while, so used both.

It's easier to include each specific configuration in mind from the code perspective. This way the one removing one condition shouldn't think if they need to change one over the other.

mzhukova: got the idea, thanks! 
if you anticipate, that the gpu intel vendor would be fixed soon and not the others, should the check be even extended to something like `is_gpu || is_nvidia_gpu || is_amd_gpu || is_generic_gpu`?

dzarukin: Thanks for thinking ahead but there's no need in that because those backends don't support gnorm.

PR #3617, opened 2025-07-16
xe: gemm: correct ordering of kernels

Simonsays095:
Corrects the ordering of `OOS`/`OOI` (i.e. int8 src/wei and f32/int32 dst) in the kernel catalog. The kernel selector assumes the catalog is sorted, so this issue could have 2 bugged outcomes:
1. Incorrectly use f32-dst and int32-dst kernels interchangeably, resulting in accuracy failures (we may already check and skip this case)
2. Skip some valid kernels when searching the catalog.

Note: we should add a mechanism to ktool that detects this issue when adding kernels or appending catalogs, then gives a warning or just fixes it automatically.


atkassen (approved):

Simonsays095:
It turns out, the catalog doesn't sort by dst type, only A/B types. These kernels are sorted, and integer/fp accumulation is handled separately.

PR #3622, opened 2025-07-17
cpu: add cross platform ALWAYS_INLINE macro and apply it to load_float_value function

aobolensk:
# Description

Introduce a cross‑platform `ALWAYS_INLINE` macro in **`common/utils.hpp`** and apply it to `load_float_value()` to guarantee compiler inlining on all supported toolchains

In real‑world workloads the function was sometimes left un‑inlined, degrading network inference throughput by **up to 1.5x** on CPU backends

### Key changes
* `common/utils.hpp` - adds portable `ALWAYS_INLINE` definition that maps to  
  * `__forceinline` on MSVC / Intel C/C++ / Intel LLVM  
  * `inline __attribute__((always_inline))` on GCC / Clang  
  * falls back to plain `inline` elsewhere
* `ref_io_helper.hpp` - annotates `load_float_value()` with the new macro (the gist of the change)
* `s390x/helpers.h` - stops declaring its own always‑inline helper and includes the shared header instead.

This centralizes the inline directive, avoids duplicate macros, and restores expected performance.  
No functional behavior changes.

Fixes 166103 (performance data attached)

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes


aobolensk:
@alvoron FYI

vpirogov:
make test

vpirogov (approved):

spalicki (approved):

avmanerikar (approved):

jwakely:
The unconventional placement of `ALWAYS_INLINE` at https://github.com/uxlfoundation/oneDNN/blob/ee74ce2cb1b98c37f7a7119b09970504259af3a2/src/cpu/s390x/helpers.h#L79 is broken by this change.

Why is the macro being positioned after the return type, instead of before it?

It worked previously when the macro expanded to just `__attribute__((always_inline))` but it no longer works now that it also expands to include `inline`

aobolensk:
> The unconventional placement of `ALWAYS_INLINE` at
> 
> https://github.com/uxlfoundation/oneDNN/blob/ee74ce2cb1b98c37f7a7119b09970504259af3a2/src/cpu/s390x/helpers.h#L79
> 
> is broken by this change.
> Why is the macro being positioned after the return type, instead of before it?
> 
> It worked previously when the macro expanded to just `__attribute__((always_inline))` but it no longer works now that it also expands to include `inline`

You are right, thank you for noticing! I have submitted a patch to align the location in such places: https://github.com/uxlfoundation/oneDNN/pull/4364

PR #3629, opened 2025-07-18
fixups: kernel dump for SYCL and benchdnn matmul reference

dzarukin:
[MFDNN-13649](https://jira.devtools.intel.com/browse/MFDNN-13649)

Kernel dump error was reported by @rjoursler (thank you!) and the fix was picked up from https://github.com/uxlfoundation/oneDNN/pull/3180 (@rjoursler, thanks again!).

Checked that both backends now report hashes. They did, and I found a difference in generated kernels for OCL and SYCL backends for RNN problems for which a cache skew issue was reported.

### DIFF
--- a/src/gpu/intel/compute/kernel.hpp
+++ b/src/gpu/intel/compute/kernel.hpp
@@ -25,6 +25,7 @@
 #include <memory>
 #include <utility>
 
+#include "common/serialization.hpp"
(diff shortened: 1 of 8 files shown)


dzarukin:
make test

spalicki (approved):

echeresh (approved):

PR #3630, opened 2025-07-18
xe: concat: explicitly zero invalid data past cutoff

syurkevi:
# Description

This PR fixes a concat bug where unneeded data from the first input could be combined with valid data in the second input.
The combination would happen along the combined boundary between the two tensors. Now, any regions that could potentially have unused data are explicitly zeroed in the trailing border case. 

Jira: [MFDNN-13838](https://jira.devtools.intel.com/browse/MFDNN-13838)

### DIFF
--- a/src/gpu/intel/reusable_simple_concat.cl
+++ b/src/gpu/intel/reusable_simple_concat.cl
@@ -572,6 +572,8 @@ internal_padding_block_concat2(__global DATA_T *dst,
         int blocks_per_simd1 = SIMD / B0;
         COMPUTE_T bVal = AS_FULL(BLOCK_READ_UL(src1 + batch_offset1
                 + next_bii * B0)); // load next blocks across concat boundary
+
+        const COMPUTE_T zero = 0;
         if (leading_boundary_shift) {
             int block_scaled_leading_shift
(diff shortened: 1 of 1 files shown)


syurkevi:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_concat

atkassen (approved):

echeresh:
@syurkevi Looks like the issue wasn't reproducible with benchdnn based on the Jira, do you know why?

syurkevi:
> @syurkevi Looks like the issue wasn't reproducible with benchdnn based on the Jira, do you know why?

I did some digging, and still don't have an exact answer, but it seems like reads past the end of the tensor are always 0 in benchdnn which doesn't affect the bit operations, unlike the standalone test where that read can return non-zero data.

echeresh:
@syurkevi Hmm, actually benchdnn should guard the tensor by NaNs after its end so it's good at catching such kind of issues.

There is a chance it's an issue on the framework side - for all block layouts we require to pad them by zeros, e.g. in this case we have 88 channels with `nChw16c` layout so channels at range [88, 95] should be all zeros when passing to oneDNN.

PR #3634, opened 2025-07-18
cpu: postops: dt support check

pmanczak:
[MFDNN-8493](https://jira.devtools.intel.com/browse/MFDNN-8493)
The library was incorrectly accepting bf16/f16 data types in post-ops on the AVX2 architecture, where these data types are not natively supported. This was due to the absence of a check for supported post-ops data types.
I have implemented a function to verify the support of post-ops data types in [src/cpu/primitive_attr_postops.cpp](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/primitive_attr_postops.cpp)

I have integrated this function into the ref_post_ops_t::init function within the same file.
Since some implementations utilize jit_uni_postops_injector instead of ref_postops so I've also added necessary check in [cpu/x64/injectors/jit_uni_postops_injector.cpp](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/x64/injectors/jit_uni_postops_injector.cpp)

### DIFF
--- a/tests/gtests/graph/unit/backend/dnnl/test_bmm.cpp
+++ b/tests/gtests/graph/unit/backend/dnnl/test_bmm.cpp
@@ -647,6 +647,11 @@ TEST(test_bmm_execute_subgraph_int8, BmmDivX8x8bf16_CPU) {
     // gpu doesn't support mixed int8-bf16 matmul with runtime zero points
     SKIP_IF(engine->kind() == graph::engine_kind::gpu, "skip on gpu");
 
+    SKIP_IF(unsupported_data_type(static_cast<dnnl::memory::data_type>(
+                                          graph::data_type::bf16),
+                    graph::dnnl_impl::make_dnnl_engine(*engine)),
+            "Skip bf16 examples for systems that do not support avx512_core.");
+
     std::vector<std::string> dtypes = {"uint8", "int8"};
 
     std::vector<int64_t> src_shape = {1, 4, 16, 8};
@@ -796,6 +801,11 @@ TEST(test_bmm_execute_subgraph_int8, BmmDivX8x8bf16_CPU) {
 TEST(test_bmm_execute_subgraph_int8, BmmDivBlockedX8x8bf16_CPU) {
     graph::engine_t *engine = get_engine();
     graph::stream_t *strm = get_stream();
+
+    SKIP_IF(unsupported_data_type(static_cast<dnnl::memory::data_type>(
+                                          graph::data_type::bf16),
+                    graph::dnnl_impl::make_dnnl_engine(*engine)),
+            "Skip bf16 examples for systems that do not support avx512_core.");
 
     std::vector<std::string> dtypes = {"uint8", "int8"};
 
--- a/tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
+++ b/tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
@@ -5257,6 +5257,11 @@ TEST(test_convolution_execute_subgraph_int8, ConvolutionSwishU8s8u8MixBf16) {
     using dims = dnnl::impl::graph::dnnl_impl::dims;
     graph::engine_t *engine = get_engine();
 
+    SKIP_IF(unsupported_data_type(static_cast<dnnl::memory::data_type>(
+                                          graph::data_type::bf16),
+                    graph::dnnl_impl::make_dnnl_engine(*engine)),
+            "Skip bf16 examples for systems that do not support avx512_core.");
+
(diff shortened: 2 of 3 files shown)


tests/gtests/graph/unit/backend/dnnl/test_bmm.cpp
```diff
@@ -647,6 +647,12 @@ TEST(test_bmm_execute_subgraph_int8, BmmDivX8x8bf16_CPU) {
     // gpu doesn't support mixed int8-bf16 matmul with runtime zero points
     SKIP_IF(engine->kind() == graph::engine_kind::gpu, "skip on gpu");
 
+    static auto isa = dnnl_get_effective_cpu_isa();
+    SKIP_IF((isa < dnnl_cpu_isa_avx512_core)
+                    && engine->kind() == graph::engine_kind::cpu,
+            "Skip bf16 examples for systems that do not support "
+            "avx512_core.");
```
TaoLv: What's the problem here (and in other tests)? Do we get correctness issue or crash after adding bf16/f16 checks in kernels?

pmanczak: The exception with the description "could not create a primitive descriptor" is thrown because, as mentioned in the task, it is prohibited to use unsupported data types in post-operations.

src/cpu/primitive_attr_postops.cpp
```diff
@@ -149,6 +150,19 @@ float ref_binary_scalar_t::compute_scalar(
     return compute_binary_scalar(alg_, src0, src1, src2);
 }
 
+bool ref_binary_scalar_t::data_type_ok(
+        const post_ops_t::entry_t::binary_t &binary) {
```
dzarukin: Suggest to pass `entry_t` instead to rely on `is_binary_with_ternary_op()` for the generality of the code.
An assert that it's `is_binary()` wouldn't hurt as well.

src/cpu/primitive_attr_postops.cpp
```diff
@@ -250,9 +279,11 @@ dim_t get_binary_src_off(const memory_desc_t &src_md, const dim_t l_offset,
 
 status_t ref_post_ops_t::init(const memory_desc_t *dst_md) {
     if (!dst_md) return status::invalid_arguments;
+    if (!post_ops_ok(po_)) return status::unimplemented;
```
dzarukin: Any reason for this specific line to appear here?

This `init` is done after pd was given to the user, meaning it's too late to return `unimplemented`.

tests/gtests/graph/unit/backend/dnnl/test_bmm.cpp
dzarukin: There's an idiom in gtests to skip specific data types:
```
SKIP_IF(unsupported_data_type(xxx_dt), message);
```

I suggest to adopt it for Graph validation as well since there's no function for `graph::data_type`. To prevent much code movement, it can be used directly in places where bf16 is called, like this (assuming it's a new available function):
```
SKIP_IF(unsupported_data_type(graph::data_type::bf16), "The example relies on bf16 dt which is not available on this system");
```
Please check up with @TaoLv what's the best place for such function. Thanks.

src/cpu/primitive_attr_postops.cpp
```diff
@@ -173,8 +187,24 @@ float ref_eltwise_scalar_fwd_t::compute_scalar(float s) const {
...
+
+bool ref_sum_scalar_t::data_type_ok(const post_ops_t::entry_t &entry) {
+    const auto &sum = entry.sum;
+    return sum.dt == data_type::undef
+            || platform::has_data_type_support(sum.dt);
+}
+
 ref_post_ops_t::ref_post_ops_t(const post_ops_t &po, bool skip_sum)
-    : po_(po), skip_sum_(skip_sum) {
+    : po_(po), sum_po_(po, skip_sum) {
```
dzarukin:
```suggestion
    : po_(po), sum_po_(skip_sum) {
```

src/cpu/primitive_attr_postops.hpp
```diff
@@ -78,11 +92,13 @@ struct ref_post_ops_t {
         return po.has_default_values({binary, eltwise, prelu, sum});
     }
 
+    static bool post_ops_ok(const post_ops_t &po);
+
 private:
     const post_ops_t &po_;
     // some primitives for example gemm are able to perform sum postop itself,
     // in such cases executing sum should be skipped
```
dzarukin: Comment to be moved next to `skip_sum_` member.

dzarukin (approved):

PR #3637, opened 2025-07-18
cpu: Coverity fixes

tczeszun:
Fixes to all major issues detected by Coverity.

### DIFF
--- a/src/cpu/reorder/simple_sparse_reorder.hpp
+++ b/src/cpu/reorder/simple_sparse_reorder.hpp
@@ -101,7 +101,8 @@ struct simple_sparse_reorder_impl_t<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
         const auto nelems = output_d.nelems(true);
-        const auto tmp_output_sz = nelems * output_d.data_type_size();
+        const auto out_dt_sz = output_d.data_type_size();
+        const auto tmp_output_sz = nelems * (out_dt_sz > 0) * out_dt_sz;
         const auto nnz_per_blocks_sz
                 = nelems / output_d.blk_size() * sizeof(dim_t);
         return tmp_output_sz + nnz_per_blocks_sz;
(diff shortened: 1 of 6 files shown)


src/cpu/reorder/simple_sparse_reorder.hpp
```diff
@@ -101,7 +101,8 @@ struct simple_sparse_reorder_impl_t<SIMPLE_SPARSE_REORDER_TEMPL_CALL,
     static size_t get_scratchpad_size(const memory_desc_wrapper &input_d,
             const memory_desc_wrapper &output_d) {
         const auto nelems = output_d.nelems(true);
-        const auto tmp_output_sz = nelems * output_d.data_type_size();
+        const auto out_dt_sz = output_d.data_type_size();
+        const auto tmp_output_sz = nelems * (out_dt_sz > 0) * out_dt_sz;
```
dzarukin: I'd drop this and marked it as Intentional. `undef` data type shouldn't make to here.

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
```diff
@@ -713,7 +713,7 @@ bool matmul_amx_blocking_params_macro_t::set_blocking_parameters() {
             // Second, calculate the number of ways the output matrix will use
             // in each set. If higher than the threshold, useful cache lines in
             // L2 will be evicted.
-            if (l2_line_size % cache_stride == 0) {
+            if (l2_line_size % cache_stride == 0 && l2_line_size > 0) {
```
dzarukin: Probably better check for `0` at line 706-707 to be explicit.

src/cpu/reorder/simple_sparse_reorder.hpp
densamoilov: I second this. Most of our utility functions, such as `data_type_size()`, have a default return value to indicate an error, usually caused by invalid arguments. There's always an assert at the end of those functions.
The idea is that those errors are unrecoverable, if they occur, it means something went wrong during development and needs to be fixed. We don't expect these errors to happen during regular use, so we don't handle them explicitly.

tczeszun: done

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
tczeszun: I changed it to set `vertical_not_possible` and silently return, as this lambda is void, `vertical_not_possible` is something like a return flag in this case

dzarukin (approved):

tczeszun:
@densamoilov need your review, please look

densamoilov (approved):

PR #3638, opened 2025-07-18
xe: extend alignment check to print more information

echeresh:


echeresh:
make test

examples/gpu_opencl_interop.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019-2022 Intel Corporation
+* Copyright 2019-2025 Intel Corporation
```
atkassen, line 2: Not needed (and same for examples/sycl_interop_buffer.cpp)

atkassen (approved):

src/gpu/intel/compute/kernel.hpp
```diff
@@ -182,14 +182,16 @@ class kernel_impl_t {
...
         const int min_alignment = 64;
         auto addr = reinterpret_cast<uint64_t>(ptr);
         if (addr % min_alignment == 0) return status::success;
         // Reference kernels support element-wise alignment.
         if (name().find("ref_") == 0) return status::success;
         // Report an error otherwise.
-        VERROR(common, runtime, "found misaligned buffer: %p", ptr);
+        VERROR(common, runtime,
+                "found misaligned buffer: %p for kernel %s at index %d", ptr,
+                name().c_str(), arg_idx);
```
Simonsays095, line 194: Since this function is only called in 2 places currently, might it be better to call it with the `VCHECK` macro (or just move the `VERROR` to the calling code)? I'd like to avoid adding the `arg_idx` argument just for the verbose print statement.

echeresh: Can you please explain what is you concern?

> might it be better to call it with the VCHECK macro (or just move the VERROR to the calling code)

I think `VCHECK()` is mostly used for regular flow conditions (e.g. during implementation dispatch). `VERROR` seems to be a better choice here.

Moving `VERROR()` to the calling code would result in duplication which is not good but not a big deal either, however what do we get with this?

kealan-barbieri (approved):

src/gpu/intel/compute/kernel.hpp
Simonsays095: It's just a nitpick based on the fact that `check_alignment` does not logically depend on `arg_idx`. In my opinion it's worse to include a logically unnecessary argument than duplicating the verbose logging code. Consider a potential future where we use this function without knowing the index of the pointer in the kernel's arguments (e.g. we're checking if pointer alignment is valid separately from kernel execution) - we would need a refactor to reuse this function in that case.

echeresh: I see what you are saying. IMO, refactoring in this case wouldn't be a big deal, and there is also a chance we never need such a check without reporting.

I was thinking about:

- Similar `check_scalar_arguments()` which reports an error right in place, so `check_*()` functions check and report - we follow the same convention here
- Future L0 runtime support. With the suggestion to duplicate we end up with three places having the same message reporting code. Frankly, this doesn't look right to me vs just putting a single message (and potentially refactoring in the future as you noticed - e.g. by adding another message-free function and calling it from both places - but only if needed)

Simonsays095: I agree, it can be refactored easily later if needed.

PR #3643, opened 2025-07-21
Update ACL reorder API without introducing regressions

Ryo-not-rio:
# Description

The ACL reorder API has been updated to allow users to specify whether to transpose or not transpose for each reorder.
A previous PR https://github.com/uxlfoundation/oneDNN/pull/2992 had been merged but reverted due to some regressions. This PR enables the API but disables ACL reorders for those that show regressions.

### DIFF
--- a/src/cpu/aarch64/acl_reorder.cpp
+++ b/src/cpu/aarch64/acl_reorder.cpp
@@ -74,14 +74,6 @@ status_t acl_reorder_fwd_t::pd_t::create(reorder_pd_t **reorder_pd,
 
     VDISPATCH_REORDER_IC(ok, "unsupported datatype");
 
-    if (!attr->scales_.has_default_values(DNNL_ARG_DST)) {
-        int mask = attr->scales_.get_mask(DNNL_ARG_DST);
-        const memory_desc_wrapper input_d(src_md);
-        VDISPATCH_REORDER_IC(
-                !(input_d.has_runtime_dims_or_strides() && mask > 0),
-                "unsupported strides or masks");
-    }
-
     // Create and check primitive descriptor
     auto _pd = make_unique_pd<pd_t>(
             attr, src_engine->kind(), src_md, dst_engine->kind(), dst_md);
@@ -93,7 +85,7 @@ status_t acl_reorder_fwd_t::pd_t::create(reorder_pd_t **reorder_pd,
     // In case we have two or four dimensions, we can't have one of the
     // two first dimensions as 1. This is valid for f32->f32 and f32->bf16.
     VDISPATCH_REORDER_IC(dst_md->dims[0] != 1 && dst_md->dims[1] != 1,
-            "first two dimensions cannot be 1");
+            "first two dimensions of the reorder being 1 is not supported");
 
     auto src_tag = memory_desc_matches_one_of_tag(
             *src_md, format_tag::ab, format_tag::ba, format_tag::cdba);
@@ -112,9 +104,10 @@ status_t acl_reorder_fwd_t::pd_t::create(reorder_pd_t **reorder_pd,
         VDISPATCH_REORDER_IC(
                 memory_desc_matches_tag(*src_md, dnnl::impl::format_tag::cdba)
                         && (memory_desc_matches_one_of_tag(*dst_md,
-                                dnnl::impl::format_tag::Acdb4a,
-                                dnnl::impl::format_tag::Acdb8a)),
-                "unsupported memory format");
+                                    dnnl::impl::format_tag::Acdb4a,
+                                    dnnl::impl::format_tag::Acdb8a)
+                                != format_tag::undef),
+                VERBOSE_UNSUPPORTED_TAG);
         transpose = true;
     } else {
         int src_dense_idx = find_innermost_dense_idx(src_md);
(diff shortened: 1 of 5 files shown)


jondea (approved):

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,221 @@
...
+    using namespace acl_utils;
+
+    // ComputeLibrary reorders support f32->f32 and f32->bf16
+    bool ok = src_md->data_type == data_type::f32
+            && utils::one_of(dst_md->data_type, data_type::f32, data_type::bf16)
+            && attr->has_default_values();
+
+    VDISPATCH_REORDER_IC(ok, "unsupported datatype");
+
+    if (!attr->scales_.has_default_values(DNNL_ARG_DST)) {
```
dzarukin: Line 73 ensures there are no attributes. This is a dead `if` statement.

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,221 @@
...
+            attr, src_engine->kind(), src_md, dst_engine->kind(), dst_md);
+    if (_pd == nullptr) return status::out_of_memory;
+    VDISPATCH_REORDER_IC(
+            _pd->init(engine, src_engine, dst_engine) == status::success,
+            "pd initialization failed");
+
+    // In case we have two or four dimensions, we can't have one of the
+    // two first dimensions as 1. This is valid for f32->f32 and f32->bf16.
+    VDISPATCH_REORDER_IC(dst_md->dims[0] != 1 && dst_md->dims[1] != 1,
+            "first two dimensions cannot be 1");
```
dzarukin: Would be nice to see not "what" but "why" there can't be other values.

"cannot" in verbose for dispatch sounds not quite correct. "cannot" should be caught at API level checks, here it's rather "not supported".

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,221 @@
...
+            "Number of dimensions in src and dst do not match");
+    VDISPATCH_REORDER_IC((dst_md->ndims == 2 || dst_md->ndims == 4),
+            "ACL only supports 2D and 4D reorders");
+    // Check if a transpose is needed during the reorder
+    if (src_md->ndims == 4) {
+        VDISPATCH_REORDER_IC(
+                memory_desc_matches_tag(*src_md, dnnl::impl::format_tag::cdba)
+                        && (memory_desc_matches_one_of_tag(*dst_md,
+                                dnnl::impl::format_tag::Acdb4a,
+                                dnnl::impl::format_tag::Acdb8a)),
```
dzarukin: `memory_desc_matches_one_of_tag` returns a tag, thus, it's incorrect to check it as a `bool`.
At best, it should be `!= format_tag::undef` if that's what you meant.

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,221 @@
...
+
+    // Return unimplemented for non-transposed reorders for now
+    // as they are faster in JIT for most cases.
+    VDISPATCH_REORDER_IC(
+            transpose, "non-transposed reorders are not supported");
+
+    auto &dst_wf = _pd->app_.dst_wf;
+
+    VDISPATCH_REORDER_IC(
+            dst_blocking.inner_nblks <= 2, "unsupported blocking format");
```
dzarukin: Please check this out: https://github.com/uxlfoundation/oneDNN/blob/main/src/common/verbose_msg.hpp#L99

Maybe you'll find more hits with generalized messages fitting the context.

src/cpu/aarch64/acl_reorder.cpp
```diff
@@ -16,11 +16,221 @@
...
+                acl_tensor_shape_out = arm_compute::TensorShape(
+                        dst_md->padded_dims[0], dst_md->padded_dims[1]);
+            } else if ((src_tag == format_tag::ba && transpose)
+                    || (src_tag == format_tag::ab && !transpose)) {
+                acl_tensor_shape_in = arm_compute::TensorShape(
+                        src_md->dims[1], src_md->dims[0]);
+                acl_tensor_shape_out = arm_compute::TensorShape(
+                        dst_md->padded_dims[1], dst_md->padded_dims[0]);
+            } else {
+                return status::unimplemented;
```
dzarukin, line 167: verbose messaging?

dzarukin:
And please squash all commits into a single one, either manually, or when promoted.

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: done

src/cpu/aarch64/acl_reorder.cpp
Ryo-not-rio: removed

PR #3644, opened 2025-07-21
benchdnn: add feature to benchmark implementation lists

Ryo-not-rio:
## Context
There has been some recent cases where we've found that simply shuffling around the implementation list offers some significant performance benefits e.g. aarch64 cpu convolution. Currently, we have to benchmark each implementation manually to see if the implementation lists are up-to-date. This new feature allows users to run a benchdnn test through all available implementations and see if any new ordering should be considered.

## Feature
Adds a --bench-list option to benchdnn which benchmarks all available implementations and gives recommendations where necessary. To minimize variability when running the benchmarks, only performance differences of > 10% are considered significant. This feature also does not work for primitives where --skip-impl currently does not work as intended e.g. reorder, and also does not work for ops not using task_executor in benchdnn i.e. brgemm, graph and rnn.

Usage:
```
benchdnn --eltwise --bench-list --batch=test_eltwise_ci
```

Output:
```
--- New list recommendations ---
--mode=P --eltwise --test-list=true --alg=gelu_erf --alpha=0 --beta=0 512x512
Current list: jit:sve_128: 0.051514, acl: 0.027832, ref:any: 0.180664
New list: acl: 0.027832, jit:sve_128: 0.051514, ref:any: 0.180664
--mode=P --eltwise --test-list=true --alg=logistic --alpha=0 --beta=0 512x512
Current list: jit:sve_128: 0.049316, acl: 0.036133, ref:any: 0.181152
New list: acl: 0.036133, jit:sve_128: 0.049316, ref:any: 0.181152
```

### DIFF
--- a/tests/benchdnn/benchdnn.cpp
+++ b/tests/benchdnn/benchdnn.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2017-2025 Intel Corporation
+* Copyright 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -76,6 +77,7 @@ bool allow_enum_tags_only {true};
 int test_start {0};
 bool attr_same_pd_check {false};
 bool check_ref_impl {false};
+bool bench_list {false};
 
 execution_mode_t execution_mode {execution_mode_t::direct};
 
@@ -185,6 +187,9 @@ int main(int argc, char **argv) {
             benchdnn_stat.mistrusted, benchdnn_stat.unimplemented,
             benchdnn_stat.invalid_arguments, benchdnn_stat.failed,
             benchdnn_stat.listed);
+
+    if (bench_list) std::cout << benchdnn_stat.recommendation;
+
     if (has_bench_mode_bit(mode_bit_t::perf)) {
         const auto &perf_timer
                 = benchdnn_stat.ms.find(timer::names::perf_timer);
--- a/tests/benchdnn/binary/bench_binary.cpp
+++ b/tests/benchdnn/binary/bench_binary.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2019-2025 Intel Corporation
+* Copyright 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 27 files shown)


dzarukin:
Hi @Ryo-not-rio, could you, please, provide more details of the problem statement behind this PR?

From what I can see, you are trying to convert a benchmarking tool to a recommendation tool which doesn't seem to fit its purpose. I'd say such kind of statistics better handled through custom scripts over the tools, unless I'm missing something essential here.

Ryo-not-rio:
Hi @dzarukin,
The issue is that we've repeatedly come across cases where when we benchmark the impl. list the current order is suboptimal. Right now, the approach to find these is for someone to manually try to benchmark each primitive and identify cases where we can shuffle the list around, but this is a tedious process prone to errors. So we needed a tool to do this automatically and implementing it in benchdnn was the simplest approach. 

To be honest, I'm happy to keep this tool internal to us, but since this was a repeated problem we have had, I thought I'd see if other contributors would find this tool helpful. Would it make more sense to change the output format so it's more in line with it being a benchmarking tool? i.e. instead of `--- New list recommendations ---` we could call it `---Non-optimal impl. list benchmarks---`

dzarukin:
> Hi @dzarukin, The issue is that we've repeatedly come across cases where when we benchmark the impl. list the current order is suboptimal. Right now, the approach to find these is for someone to manually try to benchmark each primitive and identify cases where we can shuffle the list around, but this is a tedious process prone to errors. So we needed a tool to do this automatically and implementing it in benchdnn was the simplest approach.
> 
> To be honest, I'm happy to keep this tool internal to us, but since this was a repeated problem we have had, I thought I'd see if other contributors would find this tool helpful. Would it make more sense to change the output format so it's more in line with it being a benchmarking tool? i.e. instead of `--- New list recommendations ---` we could call it `---Non-optimal impl. list benchmarks---`

Given the space of variables for a single implementation within a single primitive, my concerns are the following:
* What is the reliability of the output for various shapes times datatypes times algorithms (eltwise) times format tags?
* What is the actionable decision coming from this output? The most problematic case is when best implementations are spread 50/50 among all tested cases.
* What about untested cases - will the distribution hold there as well to make a reasonable decision?

What works for x64 is there's a single best implementation on a target platform for a single primitive. If it doesn't perform better than the next implementation, it's likely an issue to address, or it's a non-applicable performance scenario for the best implementation which is nicely handled by the next one. Either there's a problem to investigate, or a separation of responsibilities (which makes the claim for best performance implementation _per problem domain_ true). Having two best implementations doesn't make much sense from the maintenance support on a long distance. From that perspective, I don't feel there's much value in what is proposed in this PR not to say about the complexity it brings to benchdnn internal flow.

Ryo-not-rio:
The reality of the implementations for aarch64 is that currently some implementations are best for some cases and others are better for other cases. While we are trying to consolidate some of these implementations, this is a long term goal and also some primitives are always gonna have multiple good implementations so we should not rely on this. To answer your concerns, the actionable outcome I see for this tool is as follows:
 - Run the tool on a comprehensive batch of test cases
 - If the result overwhelmingly favours an implementation, implement the new order
 - If the result is favourable for a small number of test cases, add a pass through in the higher implementation for those cases
 - If the result is a 50/50, leave it as is and improve either implementation to try to consolidate further

The usefulness of the tool is not in the actual recommendation but for identifying possible optimizations whether that is a reordering or whether that is adding a pass through to the implementations.

To address your other concerns:
- What is the reliability of the output for various shapes times datatypes times algorithms (eltwise) times format tags? -- This is up to the user running the benchmarks to ensure they are running a thorough set of benchmarks
- What about untested cases - will the distribution hold there as well to make a reasonable decision? -- again, this would be up to the user to run a thorough set of benchmarks. However, even without a thorough test set, the user could implement a pass through which would immediately improve the performance of primitives

rjoursler:
> To be honest, I'm happy to keep this tool internal to us, but since this was a repeated problem we have had, I thought I'd see if other contributors would find this tool helpful.

My 2 cents from developing Xe GPU kernels: the implementation list doesn't provide enough granularity. For the code I am working on, we generally have some kind of strategy for code generation + some heuristic to select the strategy associated with each implementation. Unless we are able to iterate over different strategies and use that to inform dispatching heuristics, we won't be able to do much with a tool like this. We have [built some tooling](https://github.com/uxlfoundation/oneDNN/tree/main/src/gpu/intel/jit/v2/conv/planner) to handle this for GPU convolution, but I think it would be a lot of work to enable a more generic tool to accomplish this task.

>  However, even without a thorough test set, the user could implement a pass through which would immediately improve the performance of primitives

FYI, this is highly reminiscent of a previously proposed [tuning API](https://github.com/uxlfoundation/oneDNN/tree/dyoussif/autotune/rfcs/20231129-tuning-api) which we never ended up implementing.

vpirogov:
Rearranging implementation lists based on a single point (or even a handful of points) is definitely a bad idea. 

Is benchdnn option to benchmark all available implementations useful on it's own?

Ryo-not-rio:
@vpirogov for aarch64 it's definitely useful just as a reference point for potential optimizations but it sounds like for x86 it isn't a good feature to have

PR #3653, opened 2025-07-22
More fixes: x64 brgemm matmul weights quantization and benchdnn memory limits

dzarukin:
[MFDNN-13868](https://jira.devtools.intel.com/browse/MFDNN-13868)

[MFDNN-13635](https://jira.devtools.intel.com/browse/MFDNN-13635)
After two commits from this PR the list reported in the PR has a single problem crashed:
```
--matmul --dt=bf16:u8:f32 --stag=bac --wtag=abc --dtag=abc --attr-scales=wei:per_ocic:bf16 --attr-zero-points=wei:common:1 --attr-fpmath=bf16:true --skip-impl=ref 1x240x1912:1x1912x101
```


dzarukin:
make test

inteldimitrius (approved):

src/cpu/scale_utils.cpp
```diff
@@ -80,7 +84,7 @@ const float *precompute_scales(const memory_tracking::grantor_t &scratchpad,
             = (wei_scale_per_ic ? IC : 1) * (wei_scale_per_oc ? OC : 1);
 
     const float *scales = nullptr;
-    if (req_copy_scales(attr_scales, scale_adjust_factor)) {
+    if (req_copy_scales(attr_scales, scale_adjust_factor) || req_transpose) {
```
densamoilov: Why not `req_copy_scales(attr_scales, scale_adjust_factor, req_transpose)`?

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -171,13 +171,21 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
                 && !asc.has_default_values(DNNL_ARG_WEIGHTS)
                 && asc.get_mask(DNNL_ARG_WEIGHTS) > 0) {
             // This case requires scratchpad
-            if (N() == DNNL_RUNTIME_DIM_VAL) ok = false;
+            if (N() == DNNL_RUNTIME_DIM_VAL) return false;
```
densamoilov: We use the single exit point principle in this function so we should stick with it.

src/cpu/scale_utils.cpp
dzarukin: That was the intent, slipped away in the first edition of the patch, thanks for spotting.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: Got it, updated.

spalicki (approved):

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1132,9 +1132,10 @@ int check_total_size(res_t *res, dnnl_primitive_t prim_ref) {
     // 0.75f is taken randomly and is subject to change in future.
     const double capacity_factor = 0.75;
     const double benchdnn_device_limit = capacity_factor * device_max_capacity;
+    // Note: there used to be a separate limit for combined memory pool, however
+    // it didn't work even at 0.80 point due to, likely, system memory
+    // requirements. Use same limit for combined cpu and pure cpu cases.
     const double benchdnn_cpu_limit = capacity_factor * cpu_device_capacity;
-    // Note: used to be 0.90 until large f64 conv cases on Xe2-LPG emerged.
-    const double benchdnn_combined_limit = 0.80 * cpu_device_capacity;
```
rjoursler, line 1137: Rather than just constantly reducing the capacity (and consequently skipping problems more aggressively), would it be reasonable to do a form of lazy evalutation? Essentially what I am thinking is
```
    if(limit > 0.95*capacity) {
            return skip problem; // Same logic as currently
    } else {
          auto ret = try_execute();
          if(ret == out_of_memory && limit > 0.8*capacity)
                return skip_problem; // Expected failure mode
          return ret;
    }
```

dzarukin: That would require building a completely different flow where the execution system has a communication channel to the estimation system and can figure out the solution.
So far, I tried to stay away from such complexity. If 0.75 will prove it's not sustainable again, then, I guess, we can consider going this way...

densamoilov (approved):

PR #3656, opened 2025-07-23
remove memory type check

mayuyuace:
In before code, I get a pointer from virtual memory, but sycl::get_pointer_type will return unknown if check this pointer.
In oneDNN, I want to remove this memory type check to avoid throwing runtime error directly.

```
#include <sycl/sycl.hpp>

using namespace sycl::ext::oneapi::experimental;

int main(){

    sycl::queue q(sycl::default_selector_v, sycl::property::queue::enable_profiling{});
    auto device = q.get_device();
    std::cout << "Running on device: " << device.get_info<sycl::info::device::name>() << std::endl;

    auto mini = granularity_mode::minimum;
    auto reco = granularity_mode::recommended;

    auto syclContext = sycl::context(q.get_device());
    auto syclDevice = sycl::device(q.get_device());

    size_t mini_mem_granularity = get_mem_granularity(syclDevice, syclContext, mini);
    size_t reco_mem_granularity = get_mem_granularity(syclDevice, syclContext, reco);

    std::cout << "Minimum memory granularity: " << mini_mem_granularity << " bytes" << std::endl;
    std::cout << "Recommended memory granularity: " << reco_mem_granularity << " bytes" << std::endl;

    uintptr_t test_malloc1 = reserve_virtual_mem(mini_mem_granularity, syclContext);

    auto p_type = sycl::get_pointer_type(reinterpret_cast<void*>(test_malloc1), syclContext);
    std::cout << "is_unknown?:" << (p_type == sycl::usm::alloc::unknown) << std::endl;

    return 0;
}
```

Output:
```
Running on device: Intel(R) Graphics [0xe211]
Minimum memory granularity: 65536 bytes
Recommended memory granularity: 65536 bytes
is_unknown?:1
```

### DIFF
--- a/src/xpu/sycl/capi/capi_memory.cpp
+++ b/src/xpu/sycl/capi/capi_memory.cpp
@@ -62,18 +62,6 @@ status_t dnnl_sycl_interop_memory_create_v2(memory_t **memory,
 
     if (is_usm) {
         for (int i = 0; i < nhandles; i++) {
-            if (handles[i] != DNNL_MEMORY_NONE
-                    && handles[i] != DNNL_MEMORY_ALLOCATE) {
-                const auto *sycl_engine_impl
-                        = utils::downcast<const xpu::sycl::engine_impl_t *>(
-                                engine->impl());
-                auto &sycl_ctx = sycl_engine_impl->context();
-                ::sycl::usm::alloc ptr_type
-                        = get_pointer_type(handles[i], sycl_ctx);
-                if (ptr_type == ::sycl::usm::alloc::unknown
-                        && !engine->mayiuse_system_memory_allocators())
-                    return status::invalid_arguments;
-            }
             size_t sz = dnnl_memory_desc_get_size_v2(md, i);
             mem_storages[i].reset(new xpu::sycl::usm_memory_storage_t(engine));
             if (!mem_storages[i]) return status::out_of_memory;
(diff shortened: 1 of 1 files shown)


dzarukin:
Hi @mayuyuace, it's quite unclear what you are trying to achieve... Could you, please, elaborate on a use case, provide a reproducer that is failing with oneDNN API and provide justification why you think it should work the way you think it should. Thanks.

densamoilov:
Hi @mayuyuace, the [specification ](https://github.com/intel/llvm/blob/8861f6bb86823e9af43db1b5faf291f319807f1e/sycl/doc/extensions/experimental/sycl_ext_oneapi_virtual_mem.asciidoc#physical-memory-representation)for `sycl_ext_oneapi_virtual_mem` extension states that you would need to map the memory to get a pointer that could be used as a USM pointer. That's one thing. The second thing is that the current behavior of the extension doesn't seem correct because even if I do map the pointer `sycl::get_pointer_type` will still return `unknown`, even though it should return `device` in this case.

I have contacted the team responsible for the extension and they agreed with what I described above. I suggest you to create an issue against https://github.com/intel/llvm so that they can prioritize it accordingly and so that you can track it.

From oneDNN perspective, I think we have to keep checking the pointer because if the memory kind specified by the user is USM then the pointer is expected to work with the interfaces for USM pointers. There is only one exception, if  `usm_system_allocations` device aspect is supported then we can allow a system pointer to bypass the check.

mayuyuace:
@dzarukin  @densamoilov 
GSD have received and fixed this bug in last driver version.
So this PR can be closed. 
Thanks for your reply.

PR #3659, opened 2025-07-23
xe: reduction: reduce register spills

Simonsays095:
Limits the maximum unroll of loops in the reference reduction kernel, to prevent register spills. Since the loop bounds are known at compile time, the compiler tries to aggressively unroll the loops (in a way that depends on the specific dimensions in the problem), so randomly some cases will spill registers and cause a warning.


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_reduction

kealan-barbieri (approved):

echeresh (approved):

PR #3665, opened 2025-07-24
doc: resolve a few warnings

mzhukova:
- transform-related functionality was incorrectly marked as brgemm group
- `*` is not handled very well, options are either using `\c` or simply `` (which results into nicer table displayed and all the reference to dnnl_ still work)

### DIFF
--- a/include/oneapi/dnnl/dnnl_ukernel_types.h
+++ b/include/oneapi/dnnl/dnnl_ukernel_types.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


mzhukova:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

avmanerikar (approved):

include/oneapi/dnnl/dnnl_ukernel_types.h
```diff
@@ -79,7 +84,7 @@ typedef struct dnnl_transform *dnnl_transform_t;
 /// A constant transform routine handle.
 typedef const struct dnnl_transform *const_dnnl_transform_t;
 
-/// @} dnnl_api_ukernel_brgemm
+/// @} dnnl_api_ukernel_transform
```
dzarukin, line 87: It was intentional as the functionality does nothing by itself without brgemm. Any specific reason to change that?

mzhukova: Yes, this is to resolve a warning as per PR title/commit message :)

if this is intentional, then the other option is to get rid of dnnl_api_ukernel_transform group altogether, I suppose. For now, it is used in https://github.com/uxlfoundation/oneDNN/blob/main/include/oneapi/dnnl/dnnl_ukernel.hpp#L400 for instance, so this commit also aligns to that usage.

dzarukin: Alright, the copy-paste approach failed at some point. Let's keep the new version and see how it goes. Thanks for checking it out.

dzarukin (approved):

PR #3669, opened 2025-07-24
xe: sdpa: test: additional synchronization in write for sycl runtime

umar456:
# Description

Additional synchronization was necessary for the SYCL runtime to function properly and avoid intermittent failures. 

Fixes: [MFDNN-13937](https://jira.devtools.intel.com/browse/MFDNN-13937)

### DIFF
--- a/tests/gtests/internals/test_utils.hpp
+++ b/tests/gtests/internals/test_utils.hpp
@@ -98,20 +98,12 @@ void write_to_dnnl_memory(const T *handle, dnnl::memory &mem, dnnl::engine &eng,
             dnnl::reorder(mem_u32_mem, mem).execute(s, mem_u32_mem, mem);
         } else {
             s.wait();
-#ifdef DNNL_WITH_SYCL
-            void *mapped_ptr = (uint8_t *)mem.get_data_handle();
-            if (!mapped_ptr)
-                throw std::runtime_error("get_data_handle returned nullptr.");
-            auto sycl_queue = dnnl::sycl_interop::get_queue(s);
-            sycl_queue.memcpy(mapped_ptr, handle, size).wait();
-#else
             void *mapped_ptr = mem.map_data();
             if (!mapped_ptr)
                 throw std::runtime_error(
                         "Failed to map memory in write_to_dnnl_memory");
             std::memcpy(mapped_ptr, handle, size);
             mem.unmap_data(mapped_ptr);
-#endif
         }
         return;
     }
--- a/tests/gtests/internals/test_utils.hpp
+++ b/tests/gtests/internals/test_utils.hpp
@@ -97,14 +97,19 @@ void write_to_dnnl_memory(const T *handle, dnnl::memory &mem, dnnl::engine &eng,
                     (const unsigned *)handle, mem_u32_mem, eng, s);
             dnnl::reorder(mem_u32_mem, mem).execute(s, mem_u32_mem, mem);
         } else {
+            s.wait();
 #ifdef DNNL_WITH_SYCL
             void *mapped_ptr = (uint8_t *)mem.get_data_handle();
+            if (!mapped_ptr)
+                throw std::runtime_error("get_data_handle returned nullptr.");
             auto sycl_queue = dnnl::sycl_interop::get_queue(s);
-            sycl_queue.memcpy(mapped_ptr, handle, size);
+            sycl_queue.memcpy(mapped_ptr, handle, size).wait();
 #else
-            s.wait();
             void *mapped_ptr = mem.map_data();
-            if (mapped_ptr) std::memcpy(mapped_ptr, handle, size);
+            if (!mapped_ptr)
+                throw std::runtime_error(
+                        "Failed to map memory in write_to_dnnl_memory");
+            std::memcpy(mapped_ptr, handle, size);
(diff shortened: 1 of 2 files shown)


dzarukin (approved):

tests/gtests/internals/test_utils.hpp
```diff
@@ -97,14 +97,19 @@ void write_to_dnnl_memory(const T *handle, dnnl::memory &mem, dnnl::engine &eng,
                     (const unsigned *)handle, mem_u32_mem, eng, s);
             dnnl::reorder(mem_u32_mem, mem).execute(s, mem_u32_mem, mem);
         } else {
+            s.wait();
 #ifdef DNNL_WITH_SYCL
             void *mapped_ptr = (uint8_t *)mem.get_data_handle();
+            if (!mapped_ptr)
```
echeresh: Is there a specific reason to have separate handling with SYCL?
map -> memcpy -> unmap should work with SYCL memory too, also `get_data_handle()` here won't work when `mem` uses `sycl::buffer` and not USM.

umar456: On the OpenCL runtime, the map operation will map the entire buffer to the returned pointer. In the SYCL runtime we need to index through each handle and by default we will only get the first bytes from the buffer. I am sure there is a way to handle it but it seemed easier to perform a sycl::memcpy for this operation than create something more elaborate to handle multiple handles and make sure we transfer the correct range to the device.

Its something I can change if you prefer to not have multiple paths for the two runtimes.

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

tests/gtests/internals/test_utils.hpp
echeresh: Sorry, I'm not sure I understand, To me it seems these two versions are functionally equal for SYCL USM buffers:

Version 1:

```c++
void *mapped_ptr = (uint8_t *)mem.get_data_handle();
sycl_queue.memcpy(mapped_ptr, handle, size).wait();
```

Version 2:

```c++
void *mapped_ptr = mem.map_data();
std::memcpy(mapped_ptr, handle, size);
mem.unmap_data(mapped_ptr);
```

Are they different? What is the difference?

umar456: Yes, they do the same thing, but the mem.map_data() function performs differently on the two runtimes. One the OpenCL runtime the map_data function will return one pointer for buffers of any size.

I was running into an issue with the SYCL runtime where I was only getting the first few mega/kilobytes of the buffers. I copied the approach from the examples here: https://github.com/uxlfoundation/oneDNN/blob/main/examples/example_utils.hpp#L258
And it seemed to have addressed the issues I was running into. I am not sure why we need the different cases for the two runtimes.

echeresh: > Yes, they do the same thing, but the mem.map_data() function performs differently on the two runtimes. One the OpenCL runtime the map_data function will return one pointer for buffers of any size.

I would expect `map_data()` to have the same behavior in SYCL/OpenCL. We rely on it heavily in benchdnn so it should be well-tested.

> I was only getting the first few mega/kilobytes of the buffers

This could be due a missing synchronization. `map_data()` by default uses the internal engine stream - so-called "service" stream so if the previous operation done on a different stream - we need explicit `wait()` to ensure `map_data()` gets the updated contents.

> I copied the approach from the examples here: https://github.com/uxlfoundation/oneDNN/blob/main/examples/example_utils.hpp#L258

It's a bit confusing, I agree. I remember in the examples it was implemented with more verbosity to demonstrate what specific API to use with SYCL and with OpenCL. Functionality-wise map/unmap would be fine in the examples too.

umar456: The initial implementation of that function was just the OpenCL portion but the SYCL tests were failing once the size of the buffers increased in size, even with excessive synchronization. I can try to reproduce it and make a JIRA issue.

echeresh: @umar456 Thanks! Yes, this doesn't sound like expected behavior. I would investigate to rule out any potential test issues.

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

tests/gtests/internals/test_utils.hpp
umar456: Looks like SYCL runtime is working fine with the map/unmap branch. I have removed that code from the branch and tests are passing.

PR #3670, opened 2025-07-25
graph: backend: dnnl: support sdpa training with a special case of kv_seq_len=1

ElaineBao:
# Description

When a special case has kv_seq_len = 1, the reduction along this dimension becomes unnecessary (and reduction primitive doesn't support identity operation), so this PR aims to support such case by removing reduction ops in the original implementation.

### DIFF
--- a/tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
+++ b/tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
@@ -14,6 +14,8 @@
 --reset --dt=f32,bf16,f16 --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
 --reset --dt=0:f32+1:f32+4:f32+7:f32+10:f32+13:f32+14:f32 --case=complex_fusion/mha/sdpa-plain-training-forward-bf16-f32.json
 --reset --case=complex_fusion/mha/sdpa-plain-training-backward-f32.json
+# kv_seq_len = 1
+--reset --dt=0:f32+1:f32+4:f32+7:f32+10:f32+13:f32+14:f32 --in-shapes=1:1x16x1x64+7:1x16x384x1+13:1x16x1x64 --case=complex_fusion/mha/sdpa-plain-training-forward-bf16-f32.json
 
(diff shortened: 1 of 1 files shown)


ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/utils.cpp
```diff
@@ -415,6 +415,17 @@ bool post_binary_fusible(
         if (ltw(fused_in).data_type() != ltw(bin_out).data_type()) return false;
     }
 
+    // Special check: subtract and divide binary post-op can only be fused
+    // when base_op is their first input's producer
+    if (static_cast<dnnl::algorithm>(
+                bin_op->get_attr<int64_t>(op_attr::alg_kind))
+                    == dnnl::algorithm::binary_sub
```
dzarukin, line 422: Minor: why using public values versus internal `alg_kind::XXX`?

ElaineBao: Thanks for the comment. Just to clarify — we don't have an internal `alg_kind::XXX` enum here. We're directly storing and retrieving `dnnl::algorithm:XXX` values as int64_t, see the reference code [here](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/passes/lower.cpp#L90) and [here](https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/backend/dnnl/passes/utils.cpp#L243). So this cast back to `dnnl::algorithm:XXX` is restoring the original value.

TaoLv: I think what Dima meant is below (though I'm not sure if it works):

```
const auto alg = static_cast<impl::alg_kind_t>(bin_op->get_attr<int64_t>(op_attr::alg_kind));
if (alg == impl::alg_kind::binary_sub) {
    // ...
}
```

ElaineBao: I see, thanks for the clarification. Both should work, but as originally it was casted from dnnl::algorithm, I think it's better to keep it as is.

PR #3677, opened 2025-07-28
cpu: x64: binary: enable per_w strategy

jstachowintel:
[MFDNN-8594](https://jira.devtools.intel.com/browse/MFDNN-8594) Support per_dim_2 broadcast pattern in binary

### DIFF
--- a/src/cpu/x64/jit_uni_binary.cpp
+++ b/src/cpu/x64/jit_uni_binary.cpp
@@ -29,6 +29,7 @@ namespace x64 {
 static bcast_set_t get_supported_postops_bcast_strategies() {
     return {broadcasting_strategy_t::scalar, broadcasting_strategy_t::per_oc,
             broadcasting_strategy_t::per_oc_spatial,
+            broadcasting_strategy_t::per_w,
             broadcasting_strategy_t::no_broadcast};
 }
 
@@ -169,6 +170,9 @@ status_t jit_uni_binary_t::pd_t::init(engine_t *engine) {
 
     conf_.postops_per_oc_broadcast_exists
             = binary_injector::any_binary_postop_rhs_per_oc_broadcast(
+                    po, src0_md_, get_supported_postops_bcast_strategies());
+    conf_.postops_per_w_broadcast_exists
+            = binary_injector::any_binary_postop_rhs_per_w_broadcast(
                     po, src0_md_, get_supported_postops_bcast_strategies());
     conf_.is_bf16 = conf_.dst_type == bf16;
     conf_.is_f16 = conf_.dst_type == f16;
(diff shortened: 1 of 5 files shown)


tczeszun:
added comment to jira, as it still crashes for some cases

tczeszun (approved):

tests/benchdnn/inputs/binary/test_binary_bcast4d
```diff
@@ -0,0 +1,6 @@
+--reset
```
tczeszun, line 1: If we add new files to tests, it only makes sense if they are run by our infra, so I suggest to rename file to `harness_binary_bcast4d` and add it to `test_binary_all` (then it will be run on nightly/weekly)
Also new line at the end is missing in this and in the other file.

src/cpu/x64/jit_uni_binary.cpp
```diff
@@ -110,6 +111,31 @@ static bool data_format_supported(
...
 
+// TODO: Temporary workaround for unsupported per_w binary post-op broadcasting patterns.
+// Remove this code and add proper kernel support for these cases.
+static bool is_per_w_skipped(const memory_desc_wrapper &src0_md,
+        const memory_desc_wrapper &src1_md) {
+    const dims_t &src0_dims = src0_md.dims();
+    const dims_t &src1_dims = src1_md.dims();
+    const int ndims = src0_md.ndims();
+
+    if (ndims != 4) return false;
```
dzarukin, line 122: Looking at the JIRA and the correspondent IM conversation, it seems that ndims=3 is the major case that didn't work, thus, this condition basically kills the whole necessity in rest code.

Besides that, I think it's quite clear that the problem was solved in a brute-force manner by tailoring specific places in the code to make some cases working instead of sharpening the common logic that would fit without breaking generality. The next broadcast policy will likely make it even more clumsy. Would be nice to handle the solution in a more flexible manner.

Thirdly, it feels that the solution is a workaround because the original ask as I understand it is to add yet another **post-op** case support, instead, the main kernel is modified. Post-op code has _almost_ nothing to do with the main kernel as it's applied over destination memory. At least, I really don't see why parts of main binary primitive kernel were modified while the generic post-op infrastructure should be instead.

jstachowintel: Initially, the task was to add support for per_w broadcasting strategy. That change went through, but it turned out that for certain 4D dimensions it results in a [segfault](https://jira.devtools.intel.com/browse/MFDNN-13940).
I wouldn’t say I approached it in a brute-force manner, since I proposed a shared solution for different dimension configurations.
I agree though that it’s not the most optimal approach, as I found two use cases where it doesn’t work. I was told, however, that these cases are not critical and we can revisit them later if needed.

Could you clarify what you mean by “this condition basically kills the whole necessity in the rest of the code”, given that all cases—except for the two configurations
src0: 1 x a x b x c, src1: 1 x a x 1 x 1
src0: 1 x a x b x c, src1: 1 x 1 x b x c
are handled by the same code path?

Additionally, the is_per_w_skipped function does not affect kernel behavior; it’s only a validation in pd_t::init that helps avoid incorrect results/regressions and unblock the current workloads.

The current changes resolve the JIRA issue. What I believe still needs to be done is to remove the is_per_w_skipped method (as noted in the TODO) and extend my changes accordingly.

Please let me know what you think is the right approach here:
should I apply these modifications directly in this PR, or should we create a new JIRA ticket to track the optimization? Any advice or suggestions would be greatly appreciated.

dzarukin: Something's telling me that if I run 3D instead of 4D problem reported in the original JIRA, it would still dispatch to reference.
And if formats of sources would be different for 4D cases, there are would be failures as well.

My understanding that the implementation itself is not very good to pile up features on top of it and it would be nice to re-implement it with a simpler kernel, and less sophisticated logic with those lame broadcast policies when it comes to parallel calls.

You may reach me out if you need more detailed guidance.

PR #3683, opened 2025-07-29
graph: backend: dnnl: add matmul + select fusion support

xiang1guo:
Support matmul + select post op fusion pattern.

- Limitation
  - Only support this fusion on CPU, GPU not support yet.
  - Only support this optimized kernel fusion when select condition with full dims with dst shape. MFDNN-14003 is created to track this.
  - Only support fusion when previous op's output to be connected to select's src0.

- Performance:
  - See MFDNN-14003. Optimzied kernel has speedup.


xiang1guo:
Close the PR for now as MFDNN-14003 is pushed forward to v3.11, reopen it when MFDNN-14003 ready.

PR #3685, opened 2025-07-29
[backport][v3.9] cmake: add standard c++ exception handling on Windows

spalicki:
Backport of: #3661


dzarukin (approved):

cmake/SDL.cmake
```diff
@@ -84,10 +84,10 @@ if(UNIX)
     endif()
 elseif(WIN32)
     if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-        append(ONEDNN_SDL_COMPILER_FLAGS "/GS /Gy /guard:cf /DYNAMICBASE /sdl")
+        append(ONEDNN_SDL_COMPILER_FLAGS "/EHsr /GS /Gy /guard:cf /DYNAMICBASE /sdl")
```
HamzaAB0, line 87: Is there a reason why MSVC isn't also EHs? There must be an overhead to runtime checks.

spalicki: ICX does not have "r" option, since it is based on Clang and Clang does it by default. GCC does not create those checks when using noexcept, but we do not support GCC on Windows so I just used Clang (and ICX) as baseline.

PR #3687, opened 2025-07-29
common: add missing namespace for verbose defines

rjoursler:
Avoids unexpected compilation failures when injecting logging interfaces into third-party source code.


rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

avmanerikar (approved):

PR #3689, opened 2025-07-29
[GPU] Enable ref GEMM with per tensor scale

kealan-barbieri:
# Description

Enable src `full_tensor` scales for reference matmul, fixing a set of unimplemented cases. 

Fixes # [MFDNN-14005](https://jira.devtools.intel.com/browse/MFDNN-14005)

# Checklist

## General


src/gpu/intel/ref_matmul.cpp
```diff
@@ -140,6 +141,7 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
                 : last_scale_dim * last_scale_stride;
         last_scale_stride = src_scale_strides[d];
         last_scale_dim = src_scale_dims[d];
+        if(src_scale_dims[d] == 1)srcc_scale_strides[d] = 0;
```
dzarukin: Typo.

kealan-barbieri: fixed, thanks.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/ref_matmul.cpp
```diff
@@ -108,6 +108,7 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
                 : last_scale_dim * last_scale_stride;
         last_scale_stride = wei_scale_strides[d];
         last_scale_dim = wei_scale_dims[d];
+        if (wei_scale_dims[d] == 1) wei_scale_strides[d] = 0;
```
Simonsays095, line 111: Why are we setting the stride to 0 in these cases? It seems like the strides are being correctly computed otherwise.

kealan-barbieri: Since the offset into scale is calculated as `scale_stride_0 * dim_iter_0 + ... + scale_stride_n * dim_iter_n ` when the actual dimension is trivial we need to make sure that no offset is added for it. In most cases this doesnt need special handling since when dimension is 1 then the `dim_iter` will always be 0 so `scale_stride_x` can be non-zero. This doesnt work for broadcast matmul dims, in which case the iter can be non-zero while referring to a dimension with size 1.

PR #3692, opened 2025-07-30
xe: various DSL fixes and improvents

rjoursler:
Fixes various issues found with the DSL after rebase `rjoursle/gemm_ir` onto main. Also includes a number of improvements made since @dyoussif started work on merging the initial DSL commit. This PR is preparation for merging a prototype IR based gemm implementation.

### DIFF
--- a/src/gpu/intel/jit/ir/core.hpp
+++ b/src/gpu/intel/jit/ir/core.hpp
@@ -211,6 +211,30 @@ inline type_attr_t &operator&=(type_attr_t &a, type_attr_t b) {
     return a = a & b;
 }
 
+inline type_kind_t get_kind(ngen::DataType t) {
+    switch (t) {
+        case ngen::DataType::uq: return type_kind_t::u64;
+        case ngen::DataType::q: return type_kind_t::s64;
+        case ngen::DataType::ud: return type_kind_t::u32;
+        case ngen::DataType::d: return type_kind_t::s32;
+        case ngen::DataType::uw: return type_kind_t::u16;
+        case ngen::DataType::w: return type_kind_t::s16;
+        case ngen::DataType::ub: return type_kind_t::u8;
+        case ngen::DataType::b: return type_kind_t::s8;
+        case ngen::DataType::u4: return type_kind_t::u4;
+        case ngen::DataType::s4: return type_kind_t::s4;
+
+        case ngen::DataType::df: return type_kind_t::f64;
+        case ngen::DataType::f: return type_kind_t::f32;
+        case ngen::DataType::tf32: return type_kind_t::tf32;
+        case ngen::DataType::hf: return type_kind_t::f16;
+        case ngen::DataType::bf: return type_kind_t::bf16;
+        case ngen::DataType::bf8: return type_kind_t::bf8;
+        case ngen::DataType::hf8: return type_kind_t::hf8;
+        default: return type_kind_t::undef;
+    }
+}
+
 class type_t {
 public:
     static type_t undef() { return type_t(type_kind_t::undef); }
@@ -402,6 +426,10 @@ public:
     type_t(type_kind_t kind, uint32_t elems = 1,
             type_attr_t attr = type_attr_t::undef)
         : kind_(kind), elems_(elems), attr_(attr) {}
+
+    type_t(ngen::DataType type, uint32_t elems = 1,
+            type_attr_t attr = type_attr_t::undef)
+        : type_t(get_kind(type), elems, attr) {}
 
     type_t(const std::string &s) : elems_(1) {
 #define CASE(x) \
--- a/src/gpu/intel/jit/ir/kernel_info.hpp
+++ b/src/gpu/intel/jit/ir/kernel_info.hpp
@@ -24,9 +24,9 @@
 
 #include "common/c_types_map.hpp"
 #include "common/primitive_exec_types.hpp"
-#include "common/serialization.hpp"
 #include "gpu/intel/gpu_primitive.hpp"
 #include "gpu/intel/jit/ir/kernel_desc.hpp"
+#include "ngen_interface.hpp"
 
 namespace dnnl {
 namespace impl {
@@ -74,6 +74,22 @@ private:
 
 class kernel_iface_t {
 public:
+    kernel_iface_t() = default;
+    kernel_iface_t(const ngen::InterfaceHandler &iface) {
+        for (unsigned int i = 0; i < iface.numAssignments(); i++) {
+            auto &a = iface.getAssignment(i);
+            if (a.exttype == ngen::ExternalArgumentType::Scalar) {
+                register_arg(a.name, type_t(a.type));
+            } else if (a.exttype == ngen::ExternalArgumentType::GlobalPtr) {
+                register_arg(a.name, type_t::byte_ptr(1, false));
+            } else if (a.exttype == ngen::ExternalArgumentType::LocalPtr) {
+                register_arg(a.name, type_t::byte_ptr(1, true));
+            } else {
+                gpu_assert(false) << "Unimplemented";
+            }
+        }
+    }
+
     int nargs() const { return int(args_.size()); }
     const expr_t &arg_var(int idx) const {
         gpu_assert(idx >= 0 && idx < nargs());
--- a/third_party/ngen/ngen_interface.hpp
+++ b/third_party/ngen/ngen_interface.hpp
@@ -149,7 +149,6 @@ public:
 
     static constexpr int noSurface = 0x80;        // Returned by getArgumentSurfaceIfExists in case of no surface assignment
 
-protected:
     struct Assignment {
         std::string name;
         DataType type;
@@ -163,6 +162,10 @@ protected:
         bool globalStatelessAccess() const { return (static_cast<int>(access) & static_cast<int>(GlobalAccessType::Stateless)); }
     };
 
+    const Assignment &getAssignment(int idx) const   { return assignments[idx]; }
+    size_t numAssignments() const   { return assignments.size(); }
+
+protected:
     HW hw;
 
     std::vector<Assignment> assignments;


src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -288,101 +303,157 @@ expr_t let(const std::string &name, const expr_t &value) {
     return default_ctx().let(name, value);
 }
 
+template <>
```
dyoussif: this looks to be a duplicate of the subsequent `if_`. only difference i can spot is if_body is missing a `const ` in front of it.

rjoursler: I have reworked this. The core issue is that the current template fails to find the right implementation for
```
stmt_t create_body();
...
if_(cond, create_body())
```
As the type of `create_body()` is not `const expr_t &`. Avoiding this and clang-tidy warnings about unnecessary copies ended up being a bit involved.

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/gpu_post_ops.hpp
```diff
@@ -226,54 +230,10 @@ struct relative_md_t {
...
-        }
+            const std::array<std::string, MAX_NDIMS> &strides, int ndims) const;
 
-        std::string offset_macro = " -D" + prefix
-                + "_RMD_OFF(x0,x1,x2,x3,x4,x5)="
-                + (offset.empty() ? "0" : offset) + "";
-
-        return offset_macro;
-    }
+    jit::expr_t get_offset(const std::vector<jit::expr_t> &dim_idxs,
```
echeresh, line 235: Can we move this into `src/gpu/intel/jit` directory? For example into `ir/tensor.hpp`?

It would be nice to keep all IR usage only coming from `intel/jit`.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -101,18 +108,22 @@ struct ctx_t {
 
     tensor_t def(const v2::layout_t &layout, const std::string &name,
             const expr_t &value = {}) {
-        auto t = type_t(
-                layout.type().kind(), layout.type().elems() * layout.elems());
+        // Tensors need to be grf-aligned for loading/storing
```
echeresh, line 111: Do you have an example for that? I'm wondering why this is not handled during layout construction.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -101,18 +108,22 @@ struct ctx_t {
...
+        // TODO: IR should be modified to enable loading small tensors (such as
+        // scalar values) without GRF alignment.
+        auto elems = std::max(layout.type().elems() * layout.elems(),
+                grf_size() / layout.type().scalar().size());
```
echeresh, line 115: Should it be round-up? E.g. if it's 17 dwords, we round up to 32

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -275,114 +300,159 @@ tensor_t def(const v2::layout_t &layout, const std::string &name,
...
-    assign(this->var, obj);
+lval_t &lval_t::operator=(expr_t obj) {
+    assign(this->var, std::move(obj));
     return *this;
 }
 
-expr_t let(type_t type, const std::string &name, const expr_t &value) {
-    return default_ctx().let(type, name, value);
+expr_t let(type_t type, const std::string &name, expr_t value) {
+    return default_ctx().let(type, name, std::move(value));
```
echeresh: Do we really need all these moves? I think we have a lot of simialr code without `std::move`. To me, looks worse to read.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -275,114 +300,159 @@ tensor_t def(const v2::layout_t &layout, const std::string &name,
...
 }
 
 enum class send_kind_t { load, prefetch, store };
 
 void scatter_send(const tensor_t &t, const global_tensor_t &g,
-        const transform_t &transform, send_kind_t &op_kind,
-        const icoord_t &base) {
-    gpu_warning() << "Scatter messages are not yet implemented.";
+        send_kind_t &op_kind, const icoord_t &base, const send_hint_t &hint) {
+    gpu_warning() << "Scatter messages are not yet implemented\n";
```
echeresh: Nit: `\n` is automatic.

```suggestion
    gpu_warning() << "Scatter messages are not yet implemented";
```

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -530,15 +609,62 @@ void mma(const tensor_t &C, const tensor_t &A, const tensor_t &B,
...
+                ret[bd[i].dim] *= bd[i].int_size();
+            else
+                break;
+        }
+
+        return ret;
+    }();
+
+    std::cout << "tile: " << tile;
+    std::cout << "matching_subtile: " << matching_subtile;
```
echeresh: Debugging prints

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -42,132 +33,30 @@ int grf_size();
...
-        oss << "buffer:    " << buf.str();
+        oss << "buffer:    " << buffer.str();
         oss << "layout: " << layout.str();
         return oss.str();
     }
+
     IR_DEFINE_DUMP()
-    expr_t buf;
+
+    expr_t buffer;
```
echeresh: Nit: this looks inconsistent with `buf` in `load_t`, `store_t` and `alloc_t`.

I would change all of them, or keep `buf`. Also, historically, IR naming has a lot abbreviations such as `ptr`, `off`, `attr` so I would keep `buf` here (otherwise we add inconsistency in general or more work to rename all such abbreviations).

src/gpu/intel/jit/v2/ir/tensor.cpp
```diff
@@ -897,7 +897,6 @@ void for_each(const tile_t &base_tile, const tile_t &_tile,
     auto tile = _tile;
     for (auto &d : tile) {
         gpu_assert(base_tile.has(d));
-        gpu_assert(base_tile[d] % tile[d] == 0);
```
echeresh, line 900: Do you have an example when uneven split is needed?

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: You can hit this issue on `rjoursle/gemm_ir` using common bias, in this case the tensor is `1x1` since every dimension is broadcasted: 
```
enable_gemm_ir=1 DNNL_VERBOSE=debuginfo=200 ~/dnnl/build_dpcpp/tests/benchdnn/benchdnn --matmul --engine=gpu:1 --dt=bf16:bf16:f32 --bia-dt=f32 --bia_mask=0 --stag=ab --wtag=ab --dtag=ab 32x32:32x32
```
In this case, the tensor was allocated in `r4.5` causing a blocked load of 4 bytes to not work.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: I could restrict this to just `if_t` and `while_t`, but otherwise these seem to be necessary. The core issue is that when templating `if_/while_` dsl functions, I ended up with:

```
stmt_t create_body();
...
if_(cond, create_body())
```
The template then matches to calling the function:

```
if_(const expr_t & cond, stmt_t stmt);
```
Clang-tidy then issues warnings for a performance-pessimizing-copy unless I implement interfaces for forwarding `stmt` into the final ir object.

src/gpu/intel/jit/v2/ir/tensor.cpp
rjoursler: Yeah, I hit this in multiple times. For example, some GPU strategies are blocked by 48 over the `k` dimensions. This resulted in loading a 2d tensor with type bf16 in a  `48k32n` tile. As the natural tile used for 2d loads is `32k32n`, we need to divide this into a `32k32n` and 16k32n` load. I also hit this same style of issue related to the dpas rcount, although I no longer recall the details there.

src/gpu/intel/gpu_post_ops.hpp
rjoursler: While I generally agree with keeping IR usage in `intel/jit`,  I think it makes more sense to make an exception in this case. In particular, since this is related to post-ops, I didn't want to put this into any of the IR related directories as we are looking to upstream them to gemmstone. Additionally, none of the other files in `intel/jit` seem to be well aligned. I did make sure that `gpu_post_ops.hpp` doesn't include any of the IR headers, so there won't be any incidental includes of IR functionality outside of `intel/jit` either.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Thanks, I see. Maybe just drop the versions without lambda? Are they used in the GEMM prototype? If someone ends up manipulating IR directly, then `append(if_t::make(...)))` is always an option.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Replying to myself, it seems only relevant for <= 64-bit allocations:
https://github.com/uxlfoundation/oneDNN/blob/e6010f5c8788c1f289d6245b722d388232de0e3d/src/gpu/intel/jit/codegen/codegen.cpp#L87-L88

So should be fine.

src/gpu/intel/gpu_post_ops.hpp
echeresh: > I didn't want to put this into any of the IR related directories as we are looking to upstream them to gemmstone.

I'm not a fan of the idea of upstreaming the whole `jit/ir` due to the scope of work - I feel it's just too much to untangle and then deal with supporting this in third-party dependencies without clear benefit. But let's discuss offline.

Otherwise, let's keep this usage as is, feel free to resolve my comment.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: Part of the issue is this is used by the [labmda implementation](https://github.com/uxlfoundation/oneDNN/blob/abc61dcc0642277ce16cb319838f9828ae45af62/src/gpu/intel/jit/dsl/dsl.hpp#L169). I guess I could just create a temporary variable to work around it, but it is a fairly unexpected limitation the DSL.

echeresh: But looks like this line can be changed to `append(if_t::make(...)))`. Just in general seems like DSL API shouldn't take `stmt_t` directly, it's up to IR.

src/gpu/intel/gpu_post_ops.hpp
rjoursler: > I'm not a fan of the idea of upstreaming the whole jit/ir due to the scope of work

Agreed, it's more that we haven't clearly defined what portion will be up streamed yet.

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -42,125 +33,22 @@ int grf_size();
...
-    kind_t kind = kind_t::none;
-    int pack_size = 0;
-    send_cache_hint_t cache_hint = send_cache_hint_t::undef;
-    std::array<pvar_t, 2> dims = {};
+struct send_hint_t {
+    send_cache_hint_t cache = send_cache_hint_t::undef;
 };
 
 struct tensor_t {
+    tensor_t sub_tensor(const v2::layout_t &l, const icoord_t &coord) const {
```
echeresh: A comment for the future: we may need `layout_t::map()` like interface to pass both a tile and a coordinate. Also the offset can be part of the layout to keep the buffer normalized (without offset). With that, it would be natural to align it with `layout_t::map()` both in naming and the argument list.

echeresh (approved):

src/gpu/intel/jit/dsl/dsl.hpp
rjoursler: >  Also the offset can be part of the layout to keep the buffer normalized (without offset).

Is there a motivating reason to keep the offset in the layout? This seems counter intuitive to me. For example, the IR GEMM we have multiple tensor copies merged into one tensor object. Because of this, I would expect the following to pass:

```
tensor0 = combined_tensor.map(tile,  0);
tensor1 = combined_tensor.map(tile,  1);
gpu_assert(tensor0.layout == tensor1.layout);
```
but if we put the offset into `layout`, this would fail.

echeresh: It's just the current convention in the layout. In your example, we can do `is_equal(other, /*compare_offset=*/false)`.
I'm fine with either way as long as it won't cause issues because of inconsistency.

Actually I see this interface is not what I would expect for similar `map()`, why is `l` layout coming from outside here, what is the scenario?

rjoursler: > why is l layout coming from outside here, what is the scenario?

This is mainly for simplicity. As mentioned above, the IR GEMM implementation we has multiple copies of a tensor with a known layout wrapped into one containing tensor object. Additionally, the implementation measures offsets over the `k` dimension. As such in this case, it is much simpler to generate a sub-tensor by providing the `k` offset and the known tensor layout, rather converting the known layout to its corresponding tile, converting the `k` offset into its tile idx, and rederiving the already known sub-layout associated with that tile.

echeresh: I think I need to look at the implementation. To me it sounds like it's all there, if you have logical `k` offset - you can easily extract the sub-layout via `map()`, you just need the relative `k` index and the tile sizes.

At first glance, looking at the GEMM implementation here:

https://github.com/uxlfoundation/oneDNN/blob/2cbc7374fb8b13579d528cedb471291947a0a17d/src/gpu/intel/jit/gemm/ir/builder.cpp#L210-L214

I think we only need two out of these 4 objects: `copy_layout` and `tile`. Other two are derived from these.

rjoursler: > I think we only need two out of these 4 objects: copy_layout and tile. Other two are derived from these.

I don't disagree, but in this case, `copy_layout` is [originally derived from `layout`](https://github.com/uxlfoundation/oneDNN/blob/2cbc7374fb8b13579d528cedb471291947a0a17d/src/gpu/intel/jit/gemm/ir/builder.cpp#L205-L206). While we can rederive `layout` from `copy_layout` + `tile`, it much less computational work to just save the already computed `layout` for use later.

echeresh: I agree, it's more computationally optimal, but it's just easier to judge on the logic if there are less entities and we derive things on the fly.

Say, here we have one storage which has a fixed layout (this layout is one entity). Now to the loop: here we have indices and tile sizes corresponding to the algorithm. With these three entities we can construct loads and multiplication.

If 1) we can avoid saving derived temporary objects and 2) it's not significantly slower, IMO, better to use a more compact style.

rjoursler: Alright, I will go ahead and drop this interface and switch to deriving these values.

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

PR #3695, opened 2025-07-30
xe: sdpa: Only perform mask type check when attn mask is enabled

umar456:
# Description

Fixes the attention mask type check. This check was performed even when the attention mask was not enabled so the mask type was undef so any check against the query type was false.


umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
enable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe2-lpg

atkassen (approved):

umar456:
make test perf-gpu
set primitive=sdpa

PR #3697, opened 2025-07-31
xe2: Update BMG row major M: 33-96 strategy (subdivide granularity)

CaoZhongZ:
# Description

This should conclude fp8 BMG tuning for M 1~128. Generally we make finer N stepping for better fitting of problem shapes. The tuning will make execution time almost horizontal when M<96.


CaoZhongZ:
make test perf-gpu
set primitive=gpu:gemm

CaoZhongZ:
@Simonsays095

src/gpu/intel/jit/gemm/selector/db/kernel.db
```diff
@@ -1216,8 +1216,10 @@ auto _CATALOG_ = kcatalog::toArray({
...
+{{'G', "gemm", {"Q", "H", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, 65, -1}, {-1, 96, -1}, {-1, 65, -1}, {-1, 96, -1}, {16, 16, 1}, "IAB"}, "at32x2+m64@64 am64+m64@64 aB wg 4x4 xaf vav sr br sb64 bm0 bk0 sys grf256 sm sn st np", {16, (LoopType) 255, 128, {(LoopType) 0, (LoopType) 1, (LoopType) 2}, {8192, 8192, 16777216}, {8192, 8192, 16777216}, {32, 24, 64}, {4, 4, 1}, 1, (WGType) 1, 1, 0, 2048, {16, 16, 4}, {true, true, true}}, {'W', 1, {128}}},
+{{'G', "gemm", {"Q", "H", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, 97, -1}, {-1, 128, -1}, {-1, 97, -1}, {-1, 128, -1}, {16, 16, 1}, "IAB"}, "at32x2+m64@64 am64+m64@64 aB wg 4x4 xaf vav sr br sb64 bm0 bk0 sys grf256 sm sn st np", {16, (LoopType) 255, 128, {(LoopType) 0, (LoopType) 1, (LoopType) 2}, {8192, 8192, 16777216}, {8192, 8192, 16777216}, {32, 32, 64}, {4, 4, 1}, 1, (WGType) 1, 1, 0, 2048, {16, 16, 4}, {true, true, true}}, {'W', 1, {128}}},
```
rjoursler: For this workload, what `M` range corresponds with the transition from memory-bound to compute bound? In particular, the lack of any k-parallelization in these strategies suggest to me that they will have poor load balancing behavior. What range of shapes were these validated on?

CaoZhongZ: Any form of K-parallel doesn't help when N is large. The tuned results out perform originals (with k-slicing) in double digits. Tuning involves shapes from 8b/14b/32b/70b llama/qianwen. These strategies are essential to deliver vLLM inference efficiency and reflected in perf KPI reports.

rjoursler: Just to demonstrate what I am seeing, I did a sweep of N values checking performance. The effect of poor load balancing is fairly significant until N  ~= 30,000 which is larger than most vLLM workloads. In my experience, we can normally find strategies using local k-parallelization to split the k-dimension by 2 for minimal overhead, which would significantly improve the load balancing. I have generall found splitting the k-dimension into more than 2 chunks is not generally beneficial though, as that significantly increases synchronization and warmup costs, so I think we can only increase the number of workgroups by a factor of 2 while maintaining top-end performance.

image

CaoZhongZ:
@Simonsays095 Sorry to drag you here because it's need two approvals :)

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
Looks like performance CI didn't trigger, so I restarted it. I can approve if we don't see any regressions. I had to modify the kernels in your last PR to exclude them on LNL (we didn't catch it because CI was broken), so I want to be extra sure this time :)

CaoZhongZ:
Oh, yes. The tuning results are affected other xe2 GPUs. So I guess the strategy has to be separated at SKU level?

Simonsays095:
> Oh, yes. The tuning results are affected other xe2 GPUs. So I guess the strategy has to be separated at SKU level?

@CaoZhongZ It's not clear yet. Performance CI runs your branch against the main branch in your fork. Please rebase both to latest main and we'll try performance tests again. The issue with the previous tests is that your main branch is quite a bit more out-of-date than the feature branch.

Simonsays095:
@CaoZhongZ Apparently you're not able to trigger CI. Please rebase the main branch in your fork on top of oneDNN main, and I'll trigger it for you. Just ping me when main has been rebased.

CaoZhongZ:
@Simonsays095  Done.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #3700, opened 2025-07-31
tests: benchnn: graph: address some technical debts

ElaineBao:
# Description

Address technical debts left from https://github.com/uxlfoundation/oneDNN/pull/3396:

- [x] MFDNN-13996: benchdnn: graph: recognize sdpa backward pattern before filling data for softmax stats
- [x] MFDNN-13998: benchdnn: graph: add a dedicate state for graph specific cases
- [x] MFDNN-13997: benchdnn: graph: improve function signature for displace_input_data

### DIFF
--- a/tests/benchdnn/graph/graph.cpp
+++ b/tests/benchdnn/graph/graph.cpp
@@ -636,9 +636,6 @@ int doit(const prb_t *prb, res_t *res) {
         SAFE(ref_partition.init_ref(graph_in_ports, res), WARN);
         if (res->state == SKIPPED) return OK;
 
-        SAFE(ref_partition.init_graph_mem(partition_mem_map_v[i], res), WARN);
-        if (res->state == SKIPPED) return OK;
-
         if (has_bench_mode_bit(mode_bit_t::corr)) {
             // correctness mode, run ref partition
             if (res->state == UNTESTED || res->state == EXECUTED
@@ -652,6 +649,13 @@ int doit(const prb_t *prb, res_t *res) {
                 return FAIL;
             }
         }
+
+        // To ensure data consistency between graph and reference paths,
+        // memory initialization and data displacement for reference primitives
+        // (performed in ref_partition.init_ref and ref_partition.exec_ops)
+        // must be completed before initializing graph memory.
+        SAFE(ref_partition.init_graph_mem(partition_mem_map_v[i], res), WARN);
+        if (res->state == SKIPPED) return OK;
 
         // unmap memory from host to device
         SAFE(map_unmap_partition_mem(
--- a/tests/benchdnn/graph/input_displacer.cpp
+++ b/tests/benchdnn/graph/input_displacer.cpp
@@ -457,7 +457,6 @@ partition_data_displacer_t::partition_data_displacer_t(
 }
 
 int partition_data_displacer_t::displace_input_data(size_t lt_id,
-        dnn_mem_t &mem,
         const std::unordered_map<size_t, const dnn_mem_t &> &lt_id_2_mems,
         res_t *res) {
     if (!dg_) {
@@ -470,6 +469,7 @@ int partition_data_displacer_t::displace_input_data(size_t lt_id,
         return OK;
     }
     const displace_args_t &d_args = displace_args_.at(lt_id);
+    dnn_mem_t &mem = const_cast<dnn_mem_t &>(lt_id_2_mems.at(lt_id));
     const auto &main_op = d_args.main_op_;
     const auto &main_op_offset = d_args.main_op_offset_;
     const auto &tensor = d_args.tensor_;
--- a/tests/benchdnn/graph/input_displacer.hpp
+++ b/tests/benchdnn/graph/input_displacer.hpp
@@ -61,9 +61,7 @@ public:
     partition_data_displacer_t() = default;
     partition_data_displacer_t(
             const deserialized_graph_t &dg, const dnnl::graph::partition &par);
-    // TODO: some arguments of displace_input_data() are redundant
-    // and can be removed in the future.
-    int displace_input_data(size_t lt_id, dnn_mem_t &mem,
+    int displace_input_data(size_t lt_id,
             const std::unordered_map<size_t, const dnn_mem_t &> &lt_id_2_mems,
             res_t *res);
     filling_type_t get_filling_type(size_t lt_id) const {
--- a/tests/benchdnn/graph/ref_partition.cpp
+++ b/tests/benchdnn/graph/ref_partition.cpp
@@ -150,22 +150,6 @@ int ref_partition_t::init_ref(
         }
     }
 
-    // displace data if needed, with topo order
-    for (const auto &par_op_ref : partition_ops_ref_) {
-        for (size_t i = 0; i < par_op_ref.get().in_lts_.size(); i++) {
-            size_t lt_id = par_op_ref.get().in_lts_[i].id_;
-            if (lt_id_2_mems_.find(lt_id) == lt_id_2_mems_.end()) continue;
-            if (data_displacer.get_filling_type(lt_id)
-                    == filling_type_t::softmax_stats) {
-                res_t temp_res;
-                exec_ops(&temp_res);
-            }
-            const dnn_mem_t &mem = lt_id_2_mems_.at(lt_id);
-            SAFE_V(data_displacer.displace_input_data(
-                    lt_id, const_cast<dnn_mem_t &>(mem), lt_id_2_mems_, res));
-        }
-    }
-
     return OK;
 }
 
@@ -217,6 +201,13 @@ int ref_partition_t::init_graph_mem(
 void ref_partition_t::exec_ops(res_t *res) {
     for (const auto &par_op_ref : partition_ops_ref_) {
         const auto &op = par_op_ref.get();
+        // displace data if needed, before executing the ref_prim
+        for (size_t i = 0; i < op.in_lts_.size(); i++) {
+            size_t lt_id = op.in_lts_[i].id_;
+            SAFE_V(data_displacer.displace_input_data(
+                    lt_id, lt_id_2_mems_, res));
+        }
+
         auto ref_prim = ref_prims_.at(op.id_);
         // check if the condition input of Select op is from the parent op.
         bool select_op_cond_has_parent


tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -222,196 +222,205 @@ partition_data_displacer_t::partition_data_displacer_t(
...
-                if (!parent_op->empty()) {
-                    if (aop_in_lt->get_data_type()
-                            != logical_tensor::data_type::boolean) {
-                        // This is the qk_data, need to know its offset to
-                        // properly fill condition for padding mask.
-                        qk_data_offset = i;
-                    } else {
-                        // This means it's implicit causal mask.
-                        select_cond_has_parent = true;
+        if (dg.get_recognized_pattern() == graph_recognized_pattern_t::sdpa) {
```
dzarukin: Let's move this new condition inside `while` to simplify the changeset.

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -694,6 +694,11 @@ void deserialized_graph_t::detect_recognized_patterns() {
         BENCHDNN_PRINT(3, "%s\n", "[INFO]:sdpa pattern is recognized");
         return;
     }
+    if (ops_.size() >= 8 && detect_sdpa_bwd_impl()) {
+        recognized_pattern_ = graph_recognized_pattern_t::sdpa_bwd;
+        BENCHDNN_PRINT(3, "%s\n", "[INFO]:sdpa bwd pattern is recognized");
```
dzarukin: Let's keep the code name for easier grepping.
```suggestion
        BENCHDNN_PRINT(3, "%s\n", "[INFO]: sdpa_bwd pattern is recognized");
```

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -772,6 +823,37 @@ const deserialized_op_t &deserialized_graph_t::get_post_op(
     return in_lt_2_ops_.at(out_id).front();
 }
 
+const std::vector<deserialized_op_t> &deserialized_graph_t::get_post_ops(
```
dzarukin: Maybe "child op" instead? Post-op has a fixed meaning and it's not the same as in this context.

tests/benchdnn/graph/utils.hpp
```diff
@@ -79,6 +79,7 @@ enum class dnnl_driver_t {
 enum class graph_recognized_pattern_t {
     sdpa,
     ordinary,
```
dzarukin, line 80: Let's put ordinary on the top of the list as a default value.
I'd probably rename sdpa into sdpa_fwd to explicitly differentiate.

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -772,6 +823,37 @@ const deserialized_op_t &deserialized_graph_t::get_post_op(
...
+        const deserialized_op_t &op) const {
+
+    if (op.out_lts_.empty()) {
+        BENCHDNN_PRINT(
+                0, "Error: Getting post op of op with id %zu failed\n", op.id_);
+        SAFE_V(FAIL);
+    }
```
dzarukin, line 852: Is it really an error not to have any children for a given op?
I'd say if it's not expect on a caller level, the check for empty output should move there, but here it must be a pretty valid case.

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -756,6 +741,72 @@ bool deserialized_graph_t::detect_sdpa_impl() const {
...
+
+    std::vector<std::reference_wrapper<const deserialized_op_t>> starter_ops;
+    for (const auto &aop : ops_) {
+        if (is_root_op(aop)) starter_ops.emplace_back(aop);
+    }
+
+    for (const auto &starter : starter_ops) {
+        // find the first MatMul
+        auto cur_op_ref = find_next_util(starter.get(), "MatMul", {});
+        if (cur_op_ref.empty()) continue;
```
dzarukin: I suggest to add BENCHDNN_PRINT(8, ...) describing those `continue` calls. Otherwise, when a pattern slightly changes in the future and will stop being recognized, it would require prints anyway and much time to figure out which piece doesn't let the recognition through.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -233,12 +235,16 @@ partition_data_displacer_t::partition_data_displacer_t(
             // influencing SoftMax input values.
             while (aop.kind_ == "Add" || aop.kind_ == "Select") {
                 auto *aop_out_lt = &aop.out_lts_[0];
-                auto *child_op = &dg_->get_op_by_in_lt(aop_out_lt->id_);
-                if (child_op->kind_ != "SoftMax") break;
+                // s32 add is a case for bottom-right causal mask
+                if (aop_out_lt->data_type_ == "s32") break;
```
dzarukin: Maybe append an explicit check for Add to avoid potential select s32 cases?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -357,6 +363,7 @@ partition_data_displacer_t::partition_data_displacer_t(
             // Fill proper data for bottom-right implicit casual mask
             while (aop.kind_ == "Add") {
                 auto *aop_out_lt = &aop.out_lts_[0];
+                if (aop_out_lt->data_type_ != "s32") break;
```
dzarukin: I'd add a comment, or refer to the one above for better documentation.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -424,24 +431,26 @@ partition_data_displacer_t::partition_data_displacer_t(
...
-            auto *child_exp_op = &dg_->get_op_by_in_lt(aop_out_lt->id_);
-            if (child_exp_op->kind_ != "Exp") break;
-
-            displace_args_.emplace(aop_in_lt->id_,
-                    displace_args_t {
-                            aop, 1, *aop_in_lt, filling_type_t::softmax_stats});
-            break;
+        if (dg.get_recognized_pattern()
+                == graph_recognized_pattern_t::sdpa_bwd) {
+            while (aop.kind_ == "Subtract") {
```
dzarukin: Same here, let's move a pattern recognized into `while`.

tests/benchdnn/utils/res.hpp
```diff
@@ -36,6 +36,7 @@ enum res_state_t {
     LISTED,
     INITIALIZED,
     EXECUTED,
+    DEFERRED,
```
dzarukin: Let's add a comment explaining this status here since it's not quite conventional judging by its name.

tests/benchdnn/graph/deserialize.cpp
ElaineBao: Sorry, the error message here is misleading, I'll update. The error here is the op doesn't have any output logical tensors, which is not possible except the End/Wildcard op.

tests/benchdnn/graph/input_displacer.cpp
ElaineBao: Fixed, thanks!

tests/benchdnn/common.cpp
```diff
@@ -109,6 +109,7 @@ void parse_result(res_t &res, const char *pstr) {
             break;
         case FAILED: is_failed = true; break;
         case SKIPPED: bs.skipped++; break;
+        case DEFERRED:
```
wzt1997, line 113: What‘s the difference between `DEFERRED` and `SKIPPED` or `UNIMPLEMENTED`?

ElaineBao: Sorry, I missed the comment.. 
`DEFERRED` means primitive has no support while graph might has support (I had a comment [here](https://github.com/uxlfoundation/oneDNN/pull/3700/files#diff-e079866fd9d2614924518a3839d2c90d4ec058e282cd510c48f8a2f6d15bd0f2R39)).
`SKIPPED` or `UNIMPLEMENTED` means neither primitive nor graph has support.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -773,6 +857,37 @@ const deserialized_op_t &deserialized_graph_t::get_post_op(
...
+    }
+
+    const auto out_id = op.out_lts_[0].id_;
+    static std::vector<deserialized_op_t> dummy;
+
+    if (in_lt_2_ops_.find(out_id) == in_lt_2_ops_.end()) return dummy;
+    return in_lt_2_ops_.at(out_id);
+}
+
+deserialized_op_t deserialized_graph_t::find_next_util(
```
dzarukin: Should it return a copy?
Or it may return a `const deserialized_op_t &` instead? If not, maybe a pointer at least?

tests/benchdnn/graph/deserialize.hpp
```diff
@@ -132,8 +132,16 @@ struct deserialized_graph_t {
     const deserialized_op_t &get_op_by_out_lt(size_t out_lt_id) const;
     // returns an op based on its input logical tensor ID.
     const deserialized_op_t &get_op_by_in_lt(size_t in_lt_id) const;
-    // returns the post op that consumes the first output of given op.
-    const deserialized_op_t &get_post_op(const deserialized_op_t &op) const;
+    // returns the first child op that consumes the first output of given op.
+    const deserialized_op_t &get_child_op(const deserialized_op_t &op) const;
+    // returns the child ops that consumes the first output of given op.
+    const std::vector<deserialized_op_t> &get_child_ops(
+            const deserialized_op_t &op) const;
```
dzarukin: Looks like the first interface can be merged in a latter one by adding to a second one a number of expected ops (`1` in case of the first interface) to avoid checks at call places. Less interfaces, the better.

tests/benchdnn/graph/graph.cpp
```diff
@@ -652,6 +650,9 @@ int doit(const prb_t *prb, res_t *res) {
             }
         }
 
+        SAFE(ref_partition.init_graph_mem(partition_mem_map_v[i], res), WARN);
```
dzarukin, line 657: I'd left a comment what dependencies between initialization and `init_graph_mem` are and why this order should be preserved.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -231,13 +231,21 @@ partition_data_displacer_t::partition_data_displacer_t(
         // This is done to avoid taking future tokens into account by
         // influencing SoftMax input values.
         while (aop.kind_ == "Add" || aop.kind_ == "Select") {
+            if (dg.get_recognized_pattern()
```
dzarukin, line 235:
```suggestion
            // TODO: consider adding `dg.has_known_patterns()` when the list gets bigger.
            if (dg.get_recognized_pattern()
```

tests/benchdnn/graph/deserialize.cpp
ElaineBao: Thanks for the comments! Fixed.

wzt1997 (approved):

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

xiang1guo (approved):

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -757,20 +753,126 @@ bool deserialized_graph_t::detect_sdpa_impl() const {
...
+        if (!allowed_skips.empty() && !allowed_skips.count(cur_op_ptr->kind_)) {
+            break;
+        }
+        const auto &child_ops = get_child_ops(*cur_op_ptr);
+        if (child_ops.empty()) break;
+        cur_op_ptr = &child_ops[0];
+    }
+
+    static deserialized_op_t dummy;
+    return (cur_op_ptr->kind_ == target_kind) ? *cur_op_ptr : dummy;
```
TaoLv, line 876: What's the intention of this function? It seems if the `start_op` has `target_kind`, the `start_op` will be returned directly, is it correct?

tests/benchdnn/graph/utils.hpp
```diff
@@ -77,8 +77,9 @@ enum class dnnl_driver_t {
 };
 
 enum class graph_recognized_pattern_t {
-    sdpa,
     ordinary,
+    sdpa_fwd,
```
TaoLv, line 81: Just to confirm, it's intended that we don't differentiate forward and inference here, right?

tests/benchdnn/graph/deserialize.cpp
ElaineBao: This function is to find the next op that has target op_kind, starting from the start_op. e.g. I want to find if there's a Softmax (target op) after MatMul (start_op) If start_op has the target op_kind, then it means it's the target op also, but we seldom have such situation in the current sdpa pattern.

tests/benchdnn/graph/deserialize.hpp
```diff
@@ -132,8 +132,14 @@ struct deserialized_graph_t {
...
     // returns an op based on its input logical tensor ID.
     const deserialized_op_t &get_op_by_in_lt(size_t in_lt_id) const;
-    // returns the post op that consumes the first output of given op.
-    const deserialized_op_t &get_post_op(const deserialized_op_t &op) const;
+    // returns the child ops that consumes the first output of given op.
+    const std::vector<deserialized_op_t> &get_child_ops(
+            const deserialized_op_t &op) const;
+    // traverse downward through the given graph until either a disallowed
+    // operation is encountered or an operation of the specified kind is reached.
+    const deserialized_op_t &find_next_util(const deserialized_op_t &start_op,
```
TaoLv: Typo?
```suggestion
    const deserialized_op_t &find_next_until(const deserialized_op_t &start_op,
```

tests/benchdnn/graph/deserialize.cpp
TaoLv: Better to describe this behavior in the comment of the function. I was thinking how I can find the BMM2 starting from BMM1, with this function.

tests/benchdnn/graph/utils.hpp
ElaineBao: yes, correct

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -756,20 +753,126 @@ bool deserialized_graph_t::detect_sdpa_impl() const {
...
     const auto out_id = op.out_lts_[0].id_;
-    static deserialized_op_t dummy;
+    static std::vector<deserialized_op_t> dummy;
 
     if (in_lt_2_ops_.find(out_id) == in_lt_2_ops_.end()) return dummy;
-    return in_lt_2_ops_.at(out_id).front();
+    return in_lt_2_ops_.at(out_id);
+}
+
+const deserialized_op_t &deserialized_graph_t::find_next_util(
```
rongzha1: maybe add comments "Traverse the graph by always following the 0th child op. "is more accurate since it didn't traverse all the children

tests/benchdnn/graph/deserialize.cpp
ElaineBao: Thanks for the comments, updated

tests/benchdnn/graph/deserialize.cpp
ElaineBao: updated, thanks

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

PR #3701, opened 2025-07-31
cpu: x64: brgemm: rework scales application

dzarukin:
The PR changes the fundamentals of scales application inside the brgemm kernel.

The motivation to do it is Asynchronous runtime support required by Google. It requires non-POD stack objects like [scales_buf[16]](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/cpu_primitive.hpp#L36) either to be moved inside the parallel task (a lambda function in a parallel call), or removed. Besides that, any functions operating over user memories called outside of a parallel task by a submitter thread, such as [`precompute_scales`](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/scale_utils.hpp#L41-L43) (which combines src_scale value and wei_scale values into a scratchpad memory) leads to an error in decomposed SDPA implementation under Graph API which submits matmul tasks, and the submitter overwrites scratchpad oscales buffer with new scales values leading to correctness issues if a previous task didn't finish before the next task was submitted.

To overcome all those issues it was decided to proceed with a cleaner way by splitting scales support in the brgemm and remove the need in `precompute_scales` overall. It led to:
* Renaming all `oscales` and `scales` to `wei_scales`.
* A necessity to flip `dst_scale` from multiplication to division as there's no more pre-processing of the value in a macro. All brgemm-based kernels (brgemm, brgemm_uker, brdgmm, brgemm_post_ops) were updated accordingly to apply `src_scales` first and then `wei_scales` and then `scale_adjust` on legacy platforms. 

The approach was fine and clear until matmul with weights decompression feature emerged.
It turned out that for grouped cases, such as `--attr-scales=wei:per_ocic:bf16:32x1`, `precompute_scales` would allocate a buffer of size `KxN` instead of using `num_groups x N`, (increasing the buffer size by the factor of group size), which completely removed the purpose of the feature to save memory - it would be cheaper and faster to use non-quantized weights directly as scales alone are of the size of non-quantized weights in such case.

Since `precompute_scales` was removed, the action was required - to update brgemm_matmul implementation and respected copy_B kernels to use wei_scales directly. The support was provided for cases when `K_blk` is divisible by `gK` or vice versa (the expectation is only such cases should be in the real-world applications) since the logic of copy kernels is tightly tied to K_blk and changing it was not intended. For cases when gK divides K_blk, an additional `for`-loop appeared to process a sub_K_blk and a kernel would be generated with a correspondent logic to pick up a single value or a vector over N scales values.

Perf points (measured on 48c SPR single socket latency mode with 4c per instance):
```
--mode=P --max-ms-per-prb=1000 --matmul --dt=bf16:u8:bf16 --stag=ab --wtag=ba --dtag=ab --attr-scales=wei:per_tensor:bf16:128x1 --attr-fpmath=bf16:true --skip-impl=ref 23x1024:1024x1024_n"rgat:mm7*7"
GFLOPs Before: 4.11566
GFLOPs After: 827.149

--mode=P --max-ms-per-prb=1000 --matmul --dt=bf16:u8:bf16 --stag=ab --wtag=abx --dtag=ab --attr-scales=wei:per_tensor:bf16:128x1 --attr-fpmath=bf16:true --skip-impl=ref 2016x4096:4096x32000_n"llama2_7b:gemm6
GFLOPs Before: 947.56
GFLOPs After: 1958.02
```

For a random group size the speed ups over a set of 205 popular matmul shapes is between 1.2 to 430x with average 67x and median 13x.

The case the remained unsupported (`K_blk` is not divisible by `gK`) leads to a huge perf drop as memory is copied on the fly per B block per thread instead of one time.

The original idea was to split changes to the kernel + convolutions and separate matmul changes, but it took too much unnecessary changes and correctness issues so I dropped the idea.


dzarukin:
make test
enable benchdnn_nightly
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

src/cpu/scale_utils.cpp
```diff
@@ -36,6 +36,9 @@ void book_precomputed_scales(memory_tracking::registrar_t &scratchpad,
         float scale_adjust_factor, bool req_transpose) {
     if (req_copy_scales(attr_scales, scale_adjust_factor, req_transpose)) {
         const int wei_mask = attr_scales.get_mask(DNNL_ARG_WEIGHTS);
+        // TODO: if this kernel stays for a while and used primarily in
+        // brgemm_matmul, then the total size must be adjusted by multiplying on
+        // `scales_dt_size` but be allocated in bytes instead.
```
densamoilov: Can you also add some background to help understand the proposed approach?

src/cpu/scale_utils.cpp
```diff
@@ -36,6 +36,9 @@ void book_precomputed_scales(memory_tracking::registrar_t &scratchpad,
         float scale_adjust_factor, bool req_transpose) {
     if (req_copy_scales(attr_scales, scale_adjust_factor, req_transpose)) {
         const int wei_mask = attr_scales.get_mask(DNNL_ARG_WEIGHTS);
+        // TODO: if this kernel stays for a while and used primarily in
+        // brgemm_matmul, then the total size must be adjusted by multiplying on
```
densamoilov:
```suggestion
        // brgemm_matmul, then the total size must be adjusted by multiplying by
```

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -740,13 +743,14 @@ int brgemm_cmp(const brgemm_desc_t &lhs, const brgemm_desc_t &rhs) {
...
     CMP_BRGEMM_FIELD(with_binary);
-    CMP_BRGEMM_FIELD(with_scales);
 
     CMP_BRGEMM_FIELD(zp_type_a);
     CMP_BRGEMM_FIELD(zp_type_b);
     CMP_BRGEMM_FIELD(zp_type_c);
 
     CMP_BRGEMM_FIELD(is_oc_scale);
+    CMP_BRGEMM_FIELD(with_src_scales);
+    CMP_BRGEMM_FIELD(with_wei_scales);
```
densamoilov, line 754: Are `dt_wei_scales` and `skip_scales` missing?

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -382,26 +389,70 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
...
             const bool mask_flag = substep_simd < simd_w_;
-            const bool scale_embdbcast = !brg.is_oc_scale;
-            if (IMPLICATION(mask_flag || scale_embdbcast,
-                        is_superset(brg.isa_impl, avx512_core))) {
-                const Vmm vmm_m = maybe_mask(vmm, mask_flag, false);
-                if (scale_embdbcast) {
-                    vmulps(vmm_m, vmm, ptr_b[reg_aux_scales]);
+            const Vmm vmm = accm(m_blocks, n_blocks, m, n, v_i);
+
+            // If there are src scales, conversion happened there.
```
densamoilov: ?
```suggestion
            // If there are src scales, conversion happens there.
```

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -382,26 +389,70 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
...
+            if (dq2ps_required && !brg.with_src_scales) vcvtdq2ps(vmm, vmm);
+
+            const Vmm vmm_wei_scales = vmm_tmp(0);
+            const auto addr
+                    = ptr[reg_aux_wei_scales + wei_scales_offset(n, v_i)];
+            if (is_single_scale) {
+                if (has_ptr_b_support) {
+                    // Same isa has masks support.
```
densamoilov: What does it mean?

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -423,7 +474,9 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
         }
         for (int m = 0; m < m_blocks; m++) {
             auto vmm = accm(m_blocks, n_blocks, m, n, v_i);
-            if (dq2ps_required && !brg.with_scales) vcvtdq2ps(vmm, vmm);
+            // If there are src or wei scales, conversion happened there.
```
densamoilov: ?
```suggestion
            // If there are src or wei scales, conversion happens there.
```

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1070,13 +1070,66 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
...
+            if (is_single_scale) {
+                // Single value is not anticipated to be of any other type.
+                assert(brg.dt_wei_scales == data_type::f32);
+                if (brg.with_src_scales) {
+                    // Src scales being set, need to multiply by their value.
+                    auto scales_bcast_ptr = EVEX_compress_addr(reg_scales,
+                            scales_offset(ldi->pos(ldb)), /* bcast = */ true);
+                    vmulps(zmm_scale_masked, zmm_scale, scales_bcast_ptr);
+                } else {
+                    // No src scales, need just to load a single value.
```
densamoilov:
```suggestion
                    // No src scales, just load a single value.
```

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1070,13 +1070,66 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
...
+            const bool is_single_scale = !brg.is_oc_scale;
+
+            const auto zmm_scale = zmm_scales(ldb);
+            const auto zmm_scale_masked = zmm_scales(ldb) | k_mask | T_z;
+
+            if (is_single_scale) {
+                // Single value is not anticipated to be of any other type.
+                assert(brg.dt_wei_scales == data_type::f32);
+                if (brg.with_src_scales) {
+                    // Src scales being set, need to multiply by their value.
```
densamoilov:
```suggestion
                    // Src scales are set, need to multiply by their value.
```

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1070,13 +1070,66 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
...
+                    uni_vpslld(zmm_wei_scale, zmm_wei_scale, 16);
+                    break;
+                case data_type::f16:
+                    vcvtph2ps(zmm_wei_scale_masked, scales_ptr);
+                    break;
+                default: assert(!"unsupported wei_scales data type");
+            }
+
+            if (brg.with_src_scales) {
+                // Src scales being set, need to multiply by their value.
```
densamoilov:
```suggestion
                // Src scales are set, need to multiply by their value.
```

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1070,13 +1070,66 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
...
+                vmulps(zmm_scale_masked, zmm_scale, zmm_wei_scale);
+            } else {
+                // No src scales, need just to load.
```
densamoilov: To load what? A vector of scales?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1301,27 +1313,96 @@ void jit_brgemm_kernel_t<Wmm>::store_accumulators_apply_post_ops(dim_t bd_block,
             = brg.beta == 1.f && IMPLICATION(brg.is_int8, brg.alpha == 1.0f);
     const bool dq2ps_required = brg.is_int8
             && IMPLICATION(alpha_or_beta_applicable, beta_uses_vadd);
+    const bool has_ptr_b_support = is_superset(brg.isa_impl, avx512_core);
+
+    // This flag tracks if conversion happened since it must be done once, but
+    // any scale application or bias requires it.
```
densamoilov:
```suggestion
    // This flag tracks whether the conversion has happened, since it must be done
    // only once despite all scale and bias applications requiring it.
```

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1192,18 +1161,43 @@ void brgemm_matmul_t<isa>::copy_b_chunk_in_buffer(
...
     ctx.dynamic_src_stride = brgmm_ctx.copy_B_wei_stride();
 
-    for (int gb = 0; gb < gemm_batch; gb++) {
-        const int k = k_start + gb * bgmmc.K_blk;
+    // For best performance, scales should be taken in copy kernels as is.
+    // The biggest challenge is grouped scales (`wei_scales_gK > 1` == true,
+    // (`bgmmc.is_wei_scale_per_k` == true) as `gK` starts interfering with
+    // `K_blk`. In case when `bgmmc.gK_and_K_blk_are_divisible` == true, it's
+    // easy to call a full `K_blk` piece, or a portion of it, and apply single
+    // scale group over. When doing a portion of B, it requires an additional
```
densamoilov: ?
```suggestion
    // `K_blk`. In case when `bgmmc.gK_and_K_blk_are_divisible` == true, it's
    // easy to process a full `K_blk` piece or a portion of it by applying a single
    // scale group. When doing a portion of B, it requires an additional
```

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1772,13 +1775,16 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
                 + gb * bgmmc_.buffer_a_gb_stride;
     }
 
-    char *get_buf_B_ptr(int ithr, int k_blk_idx, int n_blk_idx, int gb) const {
+    // `k` is only used when apply wei_decompression with a single scale over
+    // a sub-piece of a `K_blk`.
+    char *get_buf_B_ptr(
+            int ithr, int k_blk_idx, int n_blk_idx, int gb, int k = 0) const {
```
densamoilov, line 1809: What is the "k" here? An offset by K?

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1926,18 +1932,109 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
...
-                ? bgmmc_.is_oscale_per_k * k
-                        + (bgmmc_.is_oscale_per_n * n
-                                * (bgmmc_.is_oscale_per_k ? bgmmc_.K : 1))
-                : bgmmc_.is_oscale_per_n * n
-                        + (bgmmc_.is_oscale_per_k * k
-                                * (bgmmc_.is_oscale_per_n ? bgmmc_.N : 1));
-        return oscales_ptr_ + offset;
+    const void *get_src_scales_ptr() const { return src_scales_; }
+
+    const void *get_wei_scales_ptr(int n, int k = 0) const {
```
densamoilov, line 1970: Can you please document these new functions? :)

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2287,8 +2384,17 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
...
     const char *bias_ptr_;
-    const float *oscales_ptr_;
-    const float *dst_scales_ptr_;
+    const void *src_scales_;
+    const void *wei_scales_;
+    // This pointer is coming from scratchpad and is needed to expand (K/g)xN
+    // scales to KxN scales as copy_B kernels rely on the full register scales
+    // for weights decompression feature in case when K_blk is not divisible by
+    // the scales group size or vice versa.
+    // TODO: implement kernels to avoid the need in this pointer and the
```
densamoilov: Can you clarify what you mean by "kernels" here?

ankalinin:
I'm interested in the performance impact of using division instead of multiplication in kernels. Have you had a chance to test the performance for sizes with small K where postprocessing takes relatively large portion of the total computation time?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3897,14 +3954,67 @@ struct jit_brgemm_matmul_copy_b_transposed_t
...
-    const auto vmm = maybe_mask(vmm_in, is_tail);
-    const auto scales_addr = EVEX_compress_addr(reg_scales, offset);
-    vmulps(vmm, vmm, scales_addr);
+        const auto vmm = maybe_mask(vmm_in, is_tail);
+        vmulps(vmm, vmm_in, vmm_wei_scales);
+    } else {
+        // A broadcasted ahead-of-time scales case.
+
+        // This branch relies on the logic wei_scales are transposed outside
+        // before passing thier values in here. This is done in
```
densamoilov:
```suggestion
        // This branch assumes that `wei_scales` have been transposed outside
        // before passing their values in here. This is done in
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3897,14 +3954,67 @@ struct jit_brgemm_matmul_copy_b_transposed_t
...
+        // `get_wei_scales_ptr` function in brgemm_matmul_ctx_t.
+        //
+        // What's important that even when groups are specified, the amount of
```
densamoilov:
```suggestion
        // What's important is that even when groups are specified, the amount of
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3897,14 +3954,67 @@ struct jit_brgemm_matmul_copy_b_transposed_t
...
+        // memory allocated is KxN, thus, there's an over-use of memory, but
+        // such usage allows to simplify the kernel logic and just load weights
```
densamoilov:
```suggestion
        // such usage allows us to simplify the kernel logic and just load weights
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -4658,19 +4772,24 @@ struct jit_brgemm_matmul_copy_b_cvt_bf16_t : public jit_brgemm_matmul_copy_b_t,
         , jit_generator_t(jit_name())
         , typesize_(conf->b_dt_sz)
         , tr_typesize_(conf->tr_b_dt_sz)
-        , scales_typesize_(sizeof(float))
+        , wei_scales_typesize_(conf_->wei_scales_dt_sz)
         , is_src_int4_(one_of(conf->orig_wei_dt, data_type::s4, data_type::u4))
-        , typesize_scale_(is_src_int4_ ? 2 : 1)
-        , src_stride_((conf->LDB * k_blk_step * typesize_) / typesize_scale_)
+        , n_src_in_byte_(is_src_int4_ ? 2 : 1)
```
densamoilov:
```suggestion
        , src_elems_per_byte_(is_src_int4_ ? 2 : 1)
```

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -225,9 +226,20 @@ struct brgemm_matmul_conf_t {
     bool is_src_batch_layout_trivial = false;
     bool is_wei_batch_layout_trivial = false;
     bool is_dst_batch_layout_trivial = false;
-    bool is_oscale_per_n = false;
-    bool is_oscale_per_k = false;
+    bool is_wei_scale_per_n = false;
+    bool is_wei_scale_per_k = false;
+    bool req_transpose_scales = false;
     bool apply_scales_in_buffer_b = false;
+    // For generic cases, when groups are selected the way they can't divide a
```
densamoilov, line 233: Did we support the "generic cases" before this PR? If so, how did we handle them?

densamoilov (approved):

src/cpu/scale_utils.cpp
dzarukin: I plan to put this text:
```
        // Current infrastructure relies on the fact the precomputed scales
        // output will be float values. However, changes in this commit update
        // the expectation for matmul grouped scaling.
        // TODO: once precompute scales infrastructure remains to be used in
        // brgemm_matmul only (for x64), update the interface with requested
        // data type for scales, since it'll be no longer true that float values
        // are expected.
```
Does it help to understand now?

src/cpu/x64/brgemm/brgemm.cpp
dzarukin: Yes, thanks for the catch, updated.

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
dzarukin: Changed to past tense, thanks for noticing.

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
dzarukin: Updated the comment to this:
```
// No need to check for mask support separately, as masks
// are supported with the same isa that introduced address
// broadcast support.
```
Does it look better now?

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
dzarukin: Updated, thanks.

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
dzarukin: A vector of values, since it's per_oc case, yes. Updated the comment, thanks.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
dzarukin: Updated, thanks.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: `k` is like an iterator inside the block for K dimension.
`ithr` defines the offset per thread,
`k_blk_idx` defines the offset between portions of blocks over K.
`gb` defines an offset to a specific K_blk inside a portion of blocks.
`k` now defines an offset inside that specific K_blk, this is the smallest granularity now.

I can put this hint as a comment to the function if you'd like.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: Sounds good to me. Applied.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: Reformulated like this:
```
// TODO: implement the logic at calling copy routines spot to handle the
// unsupported scenario mentioned above to avoid the need in this pointer
// and the overhead around filling that big buffer.
```
Does it look better now?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
dzarukin: Applied the suggestions, thanks.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
dzarukin: They were and still are supported, both before and after handles it through a bigger scratchpad buffer for weights. The difference is preparing data happened in precompute_scales once per whole buffer, now it's handled by thread per correspondent B block which is significantly less efficient.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
dzarukin: Updated, thanks.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: Sure thing, please check it out.

dzarukin:
> I'm interested in the performance impact of using division instead of multiplication in kernels. Have you had a chance to test the performance for sizes with small K where postprocessing takes relatively large portion of the total computation time?

So, division does perform worse but only when AMX is used extensively and when amount of data doesn't fit L1 any longer.
Since the precompute_scales function is gone, there's no more extra memory and extra processing which took some portion of time. Let me explore the different sequence of instructions, and if it turns out not to be better, I'll get back to a version with scratchpad memory, but this time just for dst scales and per thread inside the parallel call. Thanks for raising a concern.

src/cpu/scale_utils.cpp
densamoilov: Yeah, it's definitely clearer now, thanks.

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
densamoilov: Yeah, I see it now, thanks.

src/cpu/x64/matmul/brgemm_matmul.cpp
densamoilov: I think a quick mention of what the `k` is would be enough. But feel free to document the rest of the parameters if you want.

src/cpu/x64/matmul/brgemm_matmul.cpp
densamoilov: Very nice, thanks!

src/cpu/x64/matmul/brgemm_matmul.cpp
densamoilov: Makes sense. Do you envision it as a separate kernel or an extension to the copy kernel?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
densamoilov: Ok, so the new approach negatively affects performance of these generic cases but we don't seem to have users for the feature on CPU yet. Is it correct?

dzarukin: This is my expectation until I'm proven wrong.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: With latest modifications the kernel should be fine, the way of calling an updated kernel with correspondent K_chunk and proper pointers will be the most tricky part.

src/cpu/x64/matmul/brgemm_matmul.cpp
dzarukin: Added doc for the whole function.

dzarukin:
> > I'm interested in the performance impact of using division instead of multiplication in kernels. Have you had a chance to test the performance for sizes with small K where postprocessing takes relatively large portion of the total computation time?
> 
> So, division does perform worse but only when AMX is used extensively and when amount of data doesn't fit L1 any longer. Since the precompute_scales function is gone, there's no more extra memory and extra processing which took some portion of time. Let me explore the different sequence of instructions, and if it turns out not to be better, I'll get back to a version with scratchpad memory, but this time just for dst scales and per thread inside the parallel call. Thanks for raising a concern.

Alright, I've restored multiplication on dst_scale, it required introducing extra scratchpad space, though very small - nthr * sizeof(float) - it restored the introduced regressions and from a limited validation, it seems it should do fine now.

dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_deconv

PR #3703, opened 2025-08-01
Add OHOS judgment in the cmakelist and add OHOS macro judgment in the .h file

SC404:
# Description

we are using oneDNN in OpenHarmony, so we would like to add OHOS judgment in the cmakelist and add OHOS macro judgment in the .h file
Fixes # (github issue)

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes


third_party/xbyak_aarch64/src/util_impl_linux.h
```diff
@@ -385,10 +385,14 @@ class CpuInfoLinux : public CpuInfo {
        */
       lastDataCacheLevel_ = 2; // It is assumed L1 and L2 cache exist.
 
+#ifndef __OHOS__
```
dzarukin, line 388: It seems that it would be more sustainable to check whether `sysconf` is supported instead (if this is the major thing this macro guards from...)

third_party/xbyak_aarch64/src/util_impl_linux.h
```diff
@@ -385,10 +385,14 @@ class CpuInfoLinux : public CpuInfo {
...
       cacheInfo_.levelCache[0].size[0] = sysconf(_SC_LEVEL1_ICACHE_SIZE); // L1, ICache
       cacheInfo_.levelCache[0].size[1] = sysconf(_SC_LEVEL1_DCACHE_SIZE); // L1, DCache
       cacheInfo_.levelCache[1].size[2] = sysconf(_SC_LEVEL2_CACHE_SIZE);  // L2, UCache
       cacheInfo_.levelCache[2].size[2] = sysconf(_SC_LEVEL3_CACHE_SIZE);  // L3, UCache
+#else
+      printf("not support\n");
```
jondea, line 394: Does OHOS have an alternative to sysconf we can put here? I think we need to think more about this case. If we build and run on AArch64, it will print "not support" and continue. If it is fatal, then I would expect it to throw an error here. If it is not fatal, I would expect it to set some semi-reasonable fallback values, or perhaps emit a meaningful and actionable warning.

Either way, I think we also need a comment explaining the impact of this path being taken.

jondea (changes requested):

third_party/xbyak_aarch64/src/util_impl_linux.h
vpirogov: Also these changes should probably go to upstream [xbyak_aarch64](https://github.com/fujitsu/xbyak_aarch64) first.

openmodelzooci:

vpirogov:
Closing as stale.

PR #3707, opened 2025-08-01
ci: aarch64: upgrade perf test baseline to v3.8.0

Ryo-not-rio:
All the current failing regression tests have been resolved, and the remaining failures have been deemed as intentional slow downs.


theComputeKid (approved):

Sqvid:
Should the baseline be a patch-release? If I am not mistake a patch-release need not have a linear history with `main` which makes it a bit harder to bisect.

Ryo-not-rio:
ah was not aware of that, I've changed it to 3.8

PR #1657, opened 2023-05-24
rfcs: extend vtune integration with pd_info

mgouicem:
This is a proposal to introduce a new ITT_TASK level, in order for users to get benchdnn reproducers from Vtune trace.
This allows for:
- better correlation of oneDNN execution calls with the rest of the application, as vtune generates call stack and nested tasks
- better usability, as the users will no more need to separately generate the ONEDNN_VERBOSE output to get benchdnn reproducers.

[rendered document](https://github.com/oneapi-src/oneDNN/blob/09c09d05a6b47efd90c851f7b7e32578a046f437/rfcs/20230524-vtune-pdinfo-task/README.md)

PR #3714, opened 2025-08-04
[WIP] common: vtune: update primitive representation in VTune

avmanerikar:
(WORK IN PROGRESS - DO NOT REVIEW) 

# Description

The PR reintroduces VTune integration improvements (RFC: #1657, MFDNN-10201) which verbose information to VTune profiling tasks. 

Addresses MFDNN-13682.

### DIFF
--- a/src/common/ittnotify.cpp
+++ b/src/common/ittnotify.cpp
@@ -43,20 +43,11 @@ bool get_itt(__itt_task_level level) {
 namespace {
 
 thread_local primitive_kind_t thread_primitive_kind;
+thread_local __itt_string_handle *thread_primitive_name;
 
-__itt_domain *itt_domain() {
-    static __itt_domain *d = __itt_domain_create("dnnl::primitive::execute");
-    return d;
-}
-
-} // namespace
-
-void primitive_task_start(primitive_kind_t kind) {
-    if (kind == primitive_kind::undefined) return;
-
-#define CASE(x) \
-    __itt_string_handle_create(dnnl_prim_kind2str(primitive_kind::x))
-    static __itt_string_handle *prim_kind_itt_strings[] = {
+__itt_domain *itt_domain(primitive_kind_t kind) {
+#define CASE(x) __itt_domain_create("dnnl.execute." STRINGIFY(x))
+    static __itt_domain *prim_kind_itt_domain[] = {
             CASE(undefined),
             CASE(reorder),
             CASE(shuffle),
@@ -78,28 +69,49 @@ void primitive_task_start(primitive_kind_t kind) {
             CASE(prelu),
             CASE(softmax),
             CASE(layer_normalization),
-            CASE(group_normalization),
-            CASE(sdpa),
     };
 #undef CASE
+
     int kind_idx = (int)kind;
     assert(kind_idx >= 0);
-    if (kind_idx < primitive_kind::internal_only_start) {
-        assert((size_t)kind_idx < sizeof(prim_kind_itt_strings)
-                        / sizeof(prim_kind_itt_strings[0]));
-        __itt_task_begin(itt_domain(), __itt_null, __itt_null,
-                prim_kind_itt_strings[kind_idx]);
-    }
+    assert((size_t)kind_idx
+            < sizeof(prim_kind_itt_domain) / sizeof(prim_kind_itt_domain[0]));
+
+    return prim_kind_itt_domain[kind_idx];
+}
+
+} // namespace
+
+void primitive_task_start(primitive_kind_t kind, const char *task_name) {
+    if (kind == primitive_kind::undefined) return;
+
+    __itt_string_handle *str_handle = __itt_string_handle_create(task_name);
+    __itt_task_begin(itt_domain(kind), __itt_null, __itt_null, str_handle);
     thread_primitive_kind = kind;
+    thread_primitive_name = str_handle;
+}
+
+void primitive_task_start(
+        primitive_kind_t kind, __itt_string_handle *str_handle) {
+    if (kind == primitive_kind::undefined) return;
+
+    __itt_task_begin(itt_domain(kind), __itt_null, __itt_null, str_handle);
+    thread_primitive_kind = kind;
+    thread_primitive_name = str_handle;
 }
 
 primitive_kind_t primitive_task_get_current_kind() {
     return thread_primitive_kind;
 }
 
+__itt_string_handle *primitive_task_get_current_name() {
+    return thread_primitive_name;
+}
+
 void primitive_task_end() {
     if (thread_primitive_kind != primitive_kind::undefined) {
-        __itt_task_end(itt_domain());
+        __itt_task_end(itt_domain(thread_primitive_kind));
+        thread_primitive_name = nullptr;
         thread_primitive_kind = primitive_kind::undefined;
     }
 }
--- a/src/common/primitive_iface.cpp
+++ b/src/common/primitive_iface.cpp
@@ -91,11 +91,11 @@ status_t primitive_execute(
         const primitive_iface_t *primitive_iface, exec_ctx_t &ctx) {
     auto stream = ctx.stream();
     status_t status = success;
+    auto pd = primitive_iface->pd();
 
 #if defined(DNNL_ENABLE_ITT_TASKS)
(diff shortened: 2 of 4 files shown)


avmanerikar:
make test

vpirogov:
Could you please attach a picture showing what this looks like in VTune?

avmanerikar:
> Could you please attach a picture showing what this looks like in VTune?

Sure - The visualization is added in the description.

avmanerikar:
make test

third_party/ittnotify/ittptmark64.asm
```diff
@@ -1,11 +1,12 @@
+COMMENT @
```
vpirogov, line 1: This looks like a typo resulting in a build fail.

In general with third party components I would expect that we keep verbatim copy of the source.

avmanerikar: Fixed by reverting to the copy in the source. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/ittnotify.cpp
```diff
@@ -85,22 +77,66 @@ void primitive_task_start(primitive_kind_t kind) {
...
+
+} // namespace
+
+void primitive_task_start(
+        primitive_kind_t kind, const char *pd_info, const char *primitive_op) {
+    if (kind == primitive_kind::undefined) return;
+
+    __itt_string_handle *str_handle = __itt_string_handle_create(pd_info);
+
+    __itt_task_begin(itt_domain(kind), __itt_null, __itt_null, str_handle);
```
mgouicem: Could you confirm that there is no extra overhead from using different domain for each primitive kind?
I believe we tried that in the past ([see this commit](https://github.com/uxlfoundation/oneDNN/commit/557f3f0ba9e366fac082427fff9d149eb0f75fce)), but we had to revert as this caused performance regressions, even when vtune collection was not ongoing.

src/common/ittnotify.cpp
```diff
@@ -85,22 +77,66 @@ void primitive_task_start(primitive_kind_t kind) {
...
+        return prim_kind_itt_domain[kind_idx];
     }
+
+    return prim_kind_itt_domain[0];
+}
+
+} // namespace
+
+void primitive_task_start(
+        primitive_kind_t kind, const char *pd_info, const char *primitive_op) {
```
mgouicem: why not get rid of this overload?

src/common/ittnotify.cpp
```diff
@@ -85,22 +77,66 @@ void primitive_task_start(primitive_kind_t kind) {
...
+    if (kind == primitive_kind::undefined) return;
+
+    __itt_string_handle *str_handle = __itt_string_handle_create(pd_info);
```
mgouicem: This string_handle_create will likely cause overheads to happen: IIRC vtune has a global lock that is acquired as part of string_handle_creation.

Why not use the metadata to register pd_info?

avmanerikar: This makes sense. I have updated the commits to flip the assignments of the task labels and metadata. This also takes care of `primitive_task_start()` overload and the multiple domain definitions. 
The new hierarchy in VTune representation is:
Domain (op: create/execute) -> Task (Primitive kind) -> Metadata (Primitive Info) -> Function -> Call Stack
Thanks for the suggestion!

src/common/ittnotify.cpp
avmanerikar: See the following [comment](https://github.com/uxlfoundation/oneDNN/pull/3714#discussion_r2501240815) for explanation. For the new hierarchy, domains definition is based not on primitive kind but the operation type (primitive creation or execution).

src/common/ittnotify.cpp
avmanerikar: See the following [comment](https://github.com/uxlfoundation/oneDNN/pull/3714#discussion_r2501240815) for explanation.

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/ittnotify.cpp
```diff
@@ -43,16 +43,34 @@ bool get_itt(__itt_task_level level) {
...
 thread_local primitive_kind_t thread_primitive_kind;
+thread_local const char *thread_primitive_info;
+thread_local const char *thread_primitive_op;
+__itt_string_handle *thread_primitive_meta_fmt
+        = __itt_string_handle_create("%s");
 
-__itt_domain *itt_domain() {
-    static __itt_domain *d = __itt_domain_create("dnnl::primitive::execute");
+__itt_domain *itt_domain(const char *prim_op) {
+    if (std::strcmp(prim_op, "exec") == 0) {
```
mgouicem: please use strncmp, same below.

src/common/ittnotify.cpp
```diff
@@ -43,16 +43,34 @@ bool get_itt(__itt_task_level level) {
...
+        return d_exec;
+    }
+
+    if (std::strcmp(prim_op, "create") == 0) {
+        static __itt_domain *d_create
+                = __itt_domain_create("dnnl::primitive::create");
+        return d_create;
+    }
+
+    static __itt_domain *d = __itt_domain_create(prim_op);
```
mgouicem: Is it expected that some cases fall here? If not, please at least put an assert so that it gets flagged when testing debug build.

src/common/ittnotify.cpp
```diff
@@ -43,16 +43,34 @@ bool get_itt(__itt_task_level level) {
...
     return d;
 }
 
 } // namespace
 
-void primitive_task_start(primitive_kind_t kind) {
+void primitive_task_start(
+        primitive_kind_t kind, const char *pd_info, const char *primitive_op) {
```
mgouicem: `primitive_op` name is a bit misleading, as this suggest it is related to primitive_kind somehow. Maybe use `itt_grouping` or `log_kind` to align with verbose naming?

src/common/primitive_iface.cpp
```diff
@@ -84,58 +91,60 @@ status_t primitive_create(primitive_iface_t **primitive_iface,
...
-                    = primitive_iface->pd()->impl()->invariant_bia_md();
+            const auto pd_bia_md = pd->impl()->invariant_bia_md();
             const auto bia_md = ctx.memory_mdw(DNNL_ARG_BIAS, pd_bia_md).md_;
-            const auto pd_dst_md
-                    = primitive_iface->pd()->impl()->invariant_dst_md();
+            const auto pd_dst_md = pd->impl()->invariant_dst_md();
             const auto dst_md = ctx.memory_mdw(DNNL_ARG_DST, pd_dst_md).md_;
 
-            std::string info = primitive_iface->pd()->info_with_runtime_dims(
+            std::string info = pd->info_with_runtime_dims(
```
mgouicem, line 144: maybe capture this info string with runtime dims in task start?

src/common/ittnotify.cpp
avmanerikar: I missed that - thanks for pointing it out. This has been updated in the latest change.

src/common/ittnotify.cpp
avmanerikar: Updated the name to `log_kind`. Thanks!

src/common/ittnotify.cpp
avmanerikar: Fixed. Thanks!

src/common/primitive_iface.cpp
avmanerikar: Using `info_with_runtime_dims()` to capture runtime dims info adds back the overhead in primitive execution.  With `pd->info()`, the string generation is carried out during primitive creation which avoids the overhead during primitive execution - but when generating the info string with runtime dim values, this will be not the case.

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
make test
set test_scope=NIGHTLY

mgouicem (approved):

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
make test linters

PR #3721, opened 2025-08-05
graph: backend: dnnl: fix utility functions

TaoLv:
- Use the utility functions from src/common and remove the duplicated ones.
- Change the return value of fmap to avoid a false warning with GCC 13.
- Compare 1.f and -1.f directly and remove compare_float() as it's not used elsewhere.


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

wzt1997 (approved):

PR #3723, opened 2025-08-05
cpu: x64: Added sprinkled prefetching

yair-obodovsky:
This feature introduces prefetching for the A and B matrices. Each BRGEMM operation prefetches data for the subsequent BRGEMM, provided they share the same dimensions, otherwise no prefetching occurs. Cache lines are evenly distributed across AMX instructions and strategically sprinkled throughout the entire BRGEMM. The optimization results in a 4% geomean performance improvement on applicable layers.

### DIFF
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -130,6 +130,7 @@ struct brgemm_prf_t {
     int dist1 {-1};
     int dist2 {-1};
     int distNTA {-1};
+    bool sprinkled {false};
 };
 
 struct brgemm_batch_element_t {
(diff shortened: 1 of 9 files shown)


yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -130,6 +130,7 @@ struct brgemm_prf_t {
     int dist1 {-1};
     int dist2 {-1};
     int distNTA {-1};
+    bool sprinkled {false};
```
ankalinin, line 133: I think this new member should be added brgemm_cmp function as well as all other members are compared.
Something like `CMP_BRGEMM_FIELD(brgattr.hint_prfA.sprinkled);`

yair-obodovsky: done

PR #3725, opened 2025-08-05
scripts: synthdnn: add scatter plot report functionality

rjoursler:
Adds the ability to generate scatter plots using synthdnn. This intended to improve analysis of PRs by by allowing users to plot data and quickly identify problematic regions such as sample plot below.

```
 ./synthdnn.py report --scatter m,flops,relative f16_data/main.log f16_data/tune.log
```

Figure_1


scripts/synthdnn/synthdnn.py
```diff
@@ -184,12 +183,39 @@ def matmul_main(args):
...
+    report_parser.add_argument(
+        "logs", nargs="*", help="path to benchdnn executable"
+    )
+    report_parser.add_argument(
+        "--scatter", action="append", help="x_idx,[y_idx],metric,[scale]"
+    )
+
+
+def report_main(args):
+    reports = ReportList(args.scatter)
```
atkassen, line 201: Defer
```suggestion
    from report.report_list import ReportList

    reports = ReportList(args.scatter)
```
to here to break a dependency on `matplotlib`/`numpy` when running `collect` mode.

scripts/synthdnn/matmul/primitive.py
```diff
@@ -253,5 +253,44 @@ def __init__(self, kind, dims):
...
+        elif index == "n":
+            return self.dims.n
+        elif index == "k":
+            return self.dims.k
+        elif index == "dt":
+            return self.kind.type
+        elif index == "kind":
+            return self.kind.benchdnn_str()
+
+        raise RuntimeError(f"Cannot get {index} from matmul primitive")
```
atkassen: `KeyError` is typical for `__getitem__`

spalicki:
@rjoursler, can you please add `environment.yml` (like in `doc/environment.yml`) to `scripts/synthdnn` with tested versions of all dependencies?

scripts/synthdnn/report/report_list.py
```diff
@@ -0,0 +1,77 @@
...
+        if self.has_plot:
+            plotter.initialize()
+
+    def add_line(self, line):
+        if line.startswith("sample"):
+            sample = Sample(line)
+            for r in self.reports:
+                r.add(sample)
+
+    def finalize(self):
```
avmanerikar, line 73: (nit) This is more of a suggestion with regard to displaying the scatter plots - it will be useful to instead save the results offline with `matplotlib.pyplot.savefig()` as `matplotlib.pyplot.show()` will exhibit blocking behavior.

rjoursler: This seems like a good extension to add in the future. At least for my current personal use, I haven't had much reason to enable this. The main issue is I am often viewing 3D plots, and matplotlib cannot save interactive 3d plots. I have thought about switching to [plotly](https://github.com/plotly/plotly.py), as that library can save interactive 3d plots to HTML.

scripts/synthdnn/report/report_list.py
```diff
@@ -0,0 +1,77 @@
...
+            self.has_plot = True
+
+        if self.has_plot:
+            plotter.initialize()
+
+    def add_line(self, line):
+        if line.startswith("sample"):
+            sample = Sample(line)
+            for r in self.reports:
+                r.add(sample)
```
Simonsays095, line 71: This organization makes it hard to perform data analysis when plotting. For example, suppose we want a report for the ratio of bandwidth for two data sets. Right now, the only real option is to perform the analysis separately and save a new file with syntax that synthdnn expects.

If we instead store data collated by input coordinates, we can do pointwise ratios immediately before adding to a plot (the ratio specifically could be defined via a different metric). We'd also have to separate data collection, data collation, and plotting steps, which is probably for the best anyway.

rjoursler: This is simply ingesting data, any data collation/analysis needs be handled the corresponding report. For example, the [Summary Bounds](https://github.com/uxlfoundation/oneDNN/blob/12d6d72ee9e239d6073dd83bd5207d8213fe9163/scripts/synthdnn/reporter.py#L79-L175) report I implemented at one point automatically determines sample based compute and memory efficiency, and automatically normalizes the data so we can compare efficiency between implementations.  There is definitely room for utility functionality to improve cross-sample analysis though.

PR #3732, opened 2025-08-07
cpu: aarch64: support for destination int8 in brgemm kernel

kasturedeeksha:
# Description
This PR extends the current int8 BRGEMM implementation to support int8 destination types (s8 and u8). Additionally, it introduces bias support for int8 data type.
# Checklist

## General

1. make test output 

```
98% tests passed, 4 tests failed out of 225

Total Test time (real) = 2299.95 sec

The following tests FAILED:
        173 - test_graph_unit_dnnl_large_partition_cpu (Failed)
        196 - test_benchdnn_modeC_binary_ci_cpu (Failed)
        197 - test_benchdnn_modeC_binary_different_dt_ci_cpu (Failed)
        205 - test_benchdnn_modeC_graph_ci_cpu (Failed)
```

2. brgemm_test_all output 
    command used : 

    `./benchdnn --brgemm --batch=inputs/brgemm/test_brgemm_all`

    Before 
    ```
    tests:660480 passed:66868 skipped:593312 mistrusted:300 unimplemented:0 invalid_arguments:0 failed:0 listed:0
    total: 242.10s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 97.71s (40%); execute: 0.00s (0%); compute_ref: 63.18s (26%); compare: 54.76s (23%);
    
    ```
    After
    ```   
    tests:660480 passed:178908 skipped:479456 mistrusted:2116 unimplemented:0 invalid_arguments:0 failed:0 listed:0
    total: 193.60s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 63.65s (33%); execute: 0.00s (0%); compute_ref: 58.19s (30%);     compare: 26.61s (14%);
    
    ```
   Yes


Sqvid:
Thanks for the patch. Can you supply performance numbers (before and after) for applicable shapes please, thanks!

kasturedeeksha:
Since destination int8 was not supported earlier, a direct performance comparison is not possible. However, comparing it with f32:f32:f32 shows a 4x boost. Also, testing with u8:s8:f32 gives similar performance.

On benchmarking with 896 shapes in benchdnn, using mode=p on a Graviton 3E machine with 32 cores, the average time (in ms) :
u8:s8:u8 : 0.0032 
u8:s8:f32 : 0.0031
f32:f32:f32 : 0.0129

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1020,7 +1032,20 @@ void jit_brgemm_kernel_t::store_accumulators_apply_post_ops(
...
+    auto zmm_lbound = z_tmp_1();
+    auto zmm_ubound = z_tmp_2();
+    if (dt_requires_saturation) {
+        init_saturate_f32(
+                zmm_lbound, zmm_ubound, reg_tmp_gpr, data_type::f32, brg.dt_d);
+        for (int bd = 0; bd < bd_block; bd++) {
+            for (int ld = 0; ld < ld_block2; ld++) {
+                auto zmm = accm(ld_block2, bd, ld);
+                saturate_f32(zmm, zmm_lbound, zmm_ubound, brg.dt_d, k_mask);
+                frintn(zmm.s, k_mask, zmm.s);
```
jondea: I think we want `frinti` so that it obeys the current rounding mode, rather than always nearest with ties to even (`frintn`).

https://developer.arm.com/documentation/ddi0596/2021-09/SVE-Instructions/FRINT-r---Floating-point-round-to-integral-value--predicated--

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1135,7 +1169,22 @@ void jit_brgemm_kernel_t::store_accumulators_without_post_ops(
             = brg.beta == 1.f && IMPLICATION(brg.is_int8, brg.alpha == 1.0f);
     const bool dt_requires_saturation = brg.is_int8
             && !IMPLICATION(alpha_or_beta_applicable, beta_uses_vadd);
-    if (dt_requires_saturation) { assert(!"unsupported\n"); }
+    auto zmm_lbound = z_tmp_1();
+    auto zmm_ubound = z_tmp_2();
+    assert(zmm_lbound.getIdx() != zmm_ubound.getIdx());
+    if (dt_requires_saturation) {
```
jondea, line 1175: Could you explain why we need to saturate and convert the accumulators to s32 when storing? I was under the impression that we keep our accumulators in f32 and only convert to s8 at the very end.

jondea (changes requested):
Sorry I didn't get around to this sooner. I have a couple of minor comments, otherwise looks great. Thank you!

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Thanks, I have updated the code to use frinti instead of frintn.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: I checked, and all tests pass even if we remove the saturation part here. However, since the AVX code flow already includes this step, and there was a placeholder for the saturation case in the SVE path, I decided to keep it for now to stay consistent with the existing implementation. If you prefer, this can be removed.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Thanks, same applies to src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: If it is still in the AVX code path, then I'd be inclined to leave it in, because I trust it is faithful to the oneDNN numerical model. But in general I feel like this is the kind of detail we should be trying to understand.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Done

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Yes you are right, Once we clearly understand its purpose and confirm it’s redundant for SVE, we can revisit removing it.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1041,8 +1066,17 @@ void jit_brgemm_kernel_t::store_accumulators_apply_post_ops(
                             4);
                     break;
                 case data_type::bf16: assert(!"unsupported\n"); break;
-                case data_type::s8: assert(!"unsupported\n"); break;
-                case data_type::u8: assert(!"unsupported\n"); break;
+                case data_type::s8:
+                    smin(zmm.s, 127);
```
aditew01: It's good to name the constants for better code readability. 
maybe define a macro: `#define S8_MAX 127` ?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1041,8 +1066,17 @@ void jit_brgemm_kernel_t::store_accumulators_apply_post_ops(
...
+                    smax(zmm.s, -128);
+                    ST_MUL_VL(st1b, zmm.s, k_mask, x_addr, offset - base_offset,
+                            1);
+                    break;
+                case data_type::u8:
+                    umin(zmm.s, 255);
```
aditew01: same as above.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Rather than defining a new constant, it looks like `numeric_limits<int8_t>` is used elsewhere in oneDNN.

kasturedeeksha: Updated please check

aditew01 (approved):
Thanks! LGTM.

Sqvid:
> Do we have sufficient testing for this?

@theComputeKid yes, verified.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
Sqvid: > If it is still in the AVX code path, then I'd be inclined to leave it in, because I trust it is faithful to the oneDNN numerical model.

@mgouicem, could you share any insights on this? Thank you.

theComputeKid (approved):

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
mgouicem: In general, we keep Matmul accumulators in s32 for integral computations (e.g. int8 inputs).
Those are then converted to f32 only in presence of post-ops or scales. In which case we convert to f32, apply post-ops/scales, and converted to dst datatype with saturation.

Are you using f32 accumulators for int8 computation here?

jondea: So for s8 inputs, accumulators are always in s32 and could overflow? Or do we need to check and saturate?
[The Nuances of int8 Computations doc page](https://uxlfoundation.github.io/oneDNN/dev_guide_int8_computations.html) doesn't mention saturation.

Practically speaking, even for s8s8s8 there will almost always be a scale, otherwise the output would saturate. In this case, should the f32 conversion only happen to the finished s32 matmul accumulators? As in just before post ops/dst type conversion?

> Are you using f32 accumulators for int8 computation here?

I'm not sure to be honest, I haven't dug into the numerics of this code yet. But I think I need to 😄

vpirogov: > So for s8 inputs, accumulators are always in s32 and could overflow? Or do we need to check and saturate?

We just updated [numerical behavior](https://uxlfoundation.github.io/oneDNN/dev_guide_data_types.html#general-numerical-behavior-of-the-onednn-library) section of the documentation. Integral data types are expected to saturate and not overflow.

kasturedeeksha: @jondea Yes  accumulation for int8 inputs is done in s32.

PR #3736, opened 2025-08-07
xpu: sycl: remove is_host_accessible() workaround

rjoursler:
This workaround is causing CPU RAM to be unexpectedly "double booked" by benchdnn. This workaround was introduced nearly 5 years ago, it is unlikely to be necessary at this point.


src/xpu/sycl/usm_memory_storage.hpp
```diff
@@ -69,26 +69,8 @@ class usm_memory_storage_t : public memory_storage_base_t {
...
-         * the main thread, causing the "sequential" data migration. This W/A
-         * is a little bit more preferable as it doesn't allocate extra memory
-         * on the host. However it didn't work well for benchdnn, though worked
-         * perfectly fine for gtests. As we weren't able to find the cause of
-         * this behavior we went with the approach above. Hopefully, the driver
-         * will be fixed and we can get rid of W/A altogether. */
         return utils::one_of(usm_kind_, ::sycl::usm::alloc::host,
-                // ::sycl::usm::alloc::shared, // W/A (see above)
-                ::sycl::usm::alloc::unknown);
+                ::sycl::usm::alloc::shared, ::sycl::usm::alloc::unknown);
```
echeresh, line 73: Random spot. I see benchdnn mentions this workaround:

https://github.com/uxlfoundation/oneDNN/blob/1496c4a0539489ab49b7adffa6577fff578d4ecb/tests/benchdnn/dnnl_common.cpp#L1318-L1323

This benchdnn comment probably needs adjustments. At the same time, it's simpler to keep `mapped_mem_factor` as is - the default is USM device memory which needs this extra adjustment.

@rjoursler Just curious, how did you step into that? We should use USM device memory by default in oneDNN so USM shared memory is not regularly tested.

echeresh (approved):

src/xpu/sycl/usm_memory_storage.hpp
rjoursler: I stepped into this while tracking down the issue in [MFDNN-14017](https://jira.devtools.intel.com/browse/MFDNN-14017). I was essentially messing with different configurations to gain a better understanding of what benchdnn was allocating (while avoiding OOM), and noticed the unnecessary memory use in this configuration.

rjoursler:
make test
disable test_device_cpu

PR #3739, opened 2025-08-08
fixup: cpu: x64: brgemm: handle src and wei scales inside the kernel

dzarukin:
Fixup for #3701 , forgot to restore brgemm driver's dst scale inversion for x64 and to test the driver.


dzarukin:
make test

mzhukova (approved):

PR #3740, opened 2025-08-11
cpu: x64: matmul: enable extendable K in amx blocking heuristics

yair-obodovsky:
Added extendable K support to AMX blocking heuristics, targeting layers where the K dimension doesn't align with cache line boundaries.

### DIFF
--- a/src/cpu/x64/matmul/amx_blocking_heuristics.cpp
+++ b/src/cpu/x64/matmul/amx_blocking_heuristics.cpp
@@ -93,11 +93,8 @@ size_t matmul_amx_blocking_params_t::L1_threshold() {
 bool matmul_amx_blocking_params_macro_t::is_supported(
         const brgemm_matmul_conf_t &bgmmc,
         const brgemm_matmul_conf_utils_t &bm_conf_utils) {
-    // TODO: enable extendable_k optimization
-    if (bgmmc.K < bgmmc.wei_k_blk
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul/amx_blocking_heuristics.cpp
```diff
@@ -93,11 +93,8 @@ size_t matmul_amx_blocking_params_t::L1_threshold() {
 bool matmul_amx_blocking_params_macro_t::is_supported(
         const brgemm_matmul_conf_t &bgmmc,
         const brgemm_matmul_conf_utils_t &bm_conf_utils) {
-    // TODO: enable extendable_k optimization
-    if (bgmmc.K < bgmmc.wei_k_blk
-            || bgmmc.K % data_type_vnni_granularity(bgmmc.wei_dt) != 0) {
-        return false;
-    }
+
```
densamoilov, line 96: Does this PR fix any known issues?

yair-obodovsky:
make test
set test_scope=NIGHTLY

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
yair-obodovsky: This PR extends support for new AMX blocking heuristics to the support layer with 'K' not aligned to VNNI granularity. Previously, such cases would fall back to the old heuristics.

ankalinin (approved):

densamoilov (approved):

PR #3745, opened 2025-08-11
xe: rnn, zeropad: fixup renaming

atkassen:
Fixes errors in [coverity builds](https://intel-ci.intel.com/f076d8b9-a66a-f1a0-abfe-a4bf010d0e2d) ( [MFDNN-14050](https://jira.devtools.intel.com/browse/MFDNN-14050)) and adds missing rename of zero pad pd.

### DIFF
--- a/src/gpu/intel/rnn/utils.cpp
+++ b/src/gpu/intel/rnn/utils.cpp
@@ -16,6 +16,8 @@
 
 #include "gpu/intel/rnn/utils.hpp"
 
+#include <CL/cl.h>
+
 #include "common/c_types_map.hpp"
 #include "common/math_utils.hpp"
 
--- a/src/gpu/intel/rnn/utils.cpp
+++ b/src/gpu/intel/rnn/utils.cpp
@@ -16,8 +16,6 @@
 
 #include "gpu/intel/rnn/utils.hpp"
 
-#include <CL/cl.h>
-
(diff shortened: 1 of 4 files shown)


atkassen:
make test
disable test_device_cpu

rjoursler (approved):

src/gpu/intel/rnn/utils.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include "gpu/intel/rnn/utils.hpp"
 
+#include <CL/cl.h>
```
echeresh: This is not critical, but just a general comment for the future: we should try to keep all dependencies on OpenCL headers only in `src/gpu/intel/ocl` subdirectory - this is to simplify future L0 support (as OpenCL headers may not be available at build time).

Fine to merge as is but something to keep in mind anticipating L0 support.

atkassen: I think this is worthwhile to handle now. I have pushed a commit with the changes. Please have a look.

I have left the OpenCL dependencies in intel/sycl alone for now.

echeresh: @atkassen Thanks, looks good to me!

atkassen:
make test
disable test_device_cpu

PR #3751, opened 2025-08-12
test: benchdnn: matmul: warn on legacy bia_dt option

Sqvid:
# Description

Uncomments a warning for when a user supplies the legacy `--bia_dt` option for `benchdnn --matmul`.

### DIFF
--- a/tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
+++ b/tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
@@ -8,7 +8,7 @@
 --dtag=any
 
 --dt=f32,bf16,f16
---bia_dt=undef,f32
+--bia-dt=undef,f32
 --bia_mask=2
 --attr-post-ops=,linear:2:1
 --batch=shapes_2d_ci
@@ -28,7 +28,7 @@
 --wtag=ab
 --dtag=ab
 --dt=f32,bf16,f16,u8:s8:s32
---bia_dt=undef,f32
+--bia-dt=undef,f32
 --bia_mask=2
 --attr-scales=
 --attr-zero-points=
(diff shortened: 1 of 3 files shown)


Sqvid:
cc: @dzarukin, @mgouicem

tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
```diff
@@ -8,7 +8,7 @@
 --dtag=any
 
 --dt=f32,bf16,f16
---bia_dt=undef,f32
+--bia-dt=undef,f32
```
mzhukova, line 11: (semi-random spot) it seems like there are a few more places in `tests/benchdnn/inputs/brgemm`, should we also update to not get a warning message or is not applicable?

Sqvid: [I don't think the `--bia_dt` has been deprecated yet](https://github.com/uxlfoundation/oneDNN/blob/a0ee8f7769758252c0ed3d5057bd75103b479ebd/tests/benchdnn/brgemm/bench_brgemm.cpp#L165) for the `--brgemm` driver. @dzarukin would be better placed to answer if there is a plan to do this in the future.

theComputeKid (approved):

tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
mzhukova: got it, thanks for clarification!

dzarukin: Looks like Akela missed the driver. I'll follow up on this one. Thanks for bringing my attention.

PR #3752, opened 2025-08-12
cpu: matmul: support f32f32x8 and bf16bf16x8

asztukow:
This PR adds support for f32:f32:u8, f32:f32:s8, bf16:bf16:u8 and bf16:bf16:s8 data types configurations in matmul.
Addresses [MFDNN-14030](https://jira.devtools.intel.com/browse/MFDNN-14030)

PR #3759, opened 2025-08-13
Enable f16 dst for s8/u8 inputs in matmul and matmul-based ip

densamoilov:
This PR is a continuation of #3752 and covers the following:
* Fixes a data type check in matmul that previously caused it to fall back to reference for `s8:s8:f16`/`u8:s8:f16` configurations.
* Enables inner product (ref and matmul-based) for `s8:s8:f16`/`u8:s8:f16` configurations.
* Fixes data type configuration: `s8:u8:f16` -> `u8:s8:f16` in the benchdnn input files because the weights data type is always expected to be s8 otherwise the case is skipped.


densamoilov:
make test

avmanerikar (approved):

densamoilov:
make test

vpirogov (approved):

PR #3762, opened 2025-08-13
doc: give doxygen a hint wrt nested groups

mzhukova:
Doxygen seems to have an issue parsing through nested groups and namespaces, but giving it a small hint helps.
Also explored option to update Doxygen, but this currently produces enormous amount of warnings.
This should take care of the 3 remaining doc warnings.


mzhukova:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

vpirogov (approved):

densamoilov (approved):

PR #3604, opened 2025-07-16
WIP: xe: jit: refactor layout_t to support pvar_t dimensions

echeresh:
PR for early feedback.

This is to prepare for DSL support. Two main changes:

- `jit::layout_t` is refactored to use `pvar_t` for dimension identification: [link](https://github.com/uxlfoundation/oneDNN/blob/echeresh/layout/src/gpu/intel/jit/ir/tensor.hpp#L217)
    - I had to bring the block structure from `gpu/intel/block_structure.hpp` at cost of some duplication
-  `pvar_t` is extended and improved to behave like `dim_idx_t` where indices are used as dimensions: [link](https://github.com/uxlfoundation/oneDNN/blob/echeresh/layout/src/gpu/intel/jit/ir/problem.hpp#L44)

PR #3765, opened 2025-08-14
GPU DSL extensions

echeresh:
PR extends the IR/DSL functionality and includes other minor changes.

Full list of changes:

- IR changes:
    - Introduced `assign_t`
    - Extended `shuffle_t` to support SIMD types (used by `dsl::extract` to extract i-th component from a SIMD expression)
    - Added left/right shift and xor operations support
    - Adjusted type deduction to take into account SIMD types
    - Extended `type_t` to support `type[elems]` as a shortcut for `type.with_elems(elems)`
- DSL changes
    - Added `dsl::ctx_t::new_ir_api()` to enable new IR semantics with `assign_t` and without alloc/let
        - We should fully migrate to the new semantics in the future and drop this flag
    - Migrated DSL to rely on `jit::layout_t` instead of v2 (which is possible after #3604)
    - Introduced `dsl::def_slm()` to allocate SLM tensors
    - Introduced `dsl::_for` for for loops
    - Introduced `dsl/decl.hpp` for shorter aliases for types and `expr_t`.

There are conflicts with #3729 so this PR should be merged after #3729.

### DIFF
--- a/src/gpu/intel/jit/dsl/decl.hpp
+++ b/src/gpu/intel/jit/dsl/decl.hpp
@@ -1,0 +1,51 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef GPU_INTEL_JIT_DSL_DECL_HPP
+#define GPU_INTEL_JIT_DSL_DECL_HPP
+
+#include "gpu/intel/jit/ir/core.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace gpu {
+namespace intel {
+namespace jit {
+namespace dsl {
+
+static type_t _bool = type_t::_bool();
+static type_t s8 = type_t::s8();
+static type_t u8 = type_t::u8();
+static type_t s16 = type_t::s16();
+static type_t u16 = type_t::u16();
+static type_t s32 = type_t::s32();
+static type_t u32 = type_t::u32();
+static type_t s64 = type_t::s64();
+static type_t u64 = type_t::u64();
+static type_t f32 = type_t::f32();
+static type_t f16 = type_t::f16();
+static type_t bf16 = type_t::bf16();
+
+using expr_t = jit::expr_t;
+
+} // namespace dsl
+} // namespace jit
+} // namespace intel
+} // namespace gpu
+} // namespace impl
+} // namespace dnnl
+
+#endif
--- a/src/gpu/intel/jit/dsl/dsl.cpp
+++ b/src/gpu/intel/jit/dsl/dsl.cpp
@@ -31,8 +31,12 @@ namespace jit {
 namespace dsl {
 
 struct ctx_t {
+    bool new_ir_api() const { return new_ir_api_; }
 
-    void declare_kernel(const kernel_iface_t &interface, ir_context_t &ctx) {
+    void declare_kernel(const kernel_iface_t &interface, ir_context_t &ctx,
+            bool new_ir_api = false) {
+        slm_byte_offset_ = 0;
+        new_ir_api_ = new_ir_api;
         gpu_assert(stmts_stack_.empty())
                 << "Invalid generation of a kernel within a kernel";
         interface_ = interface;
@@ -40,25 +44,39 @@ struct ctx_t {
 
         begin_scope();
 
-        for (int i = 0; i < interface.nargs(); i++) {
-            const auto &var = interface.arg_var(i);
-            if (var.type().is_ptr()) {
-                if (var.type().is_slm()) {
-                    append(alloc_t::make(var, 0, alloc_kind_t::slm, stmt_t {}));
-                } else {
-                    append(alloc_t::make(
-                            var, 0, alloc_kind_t::global, stmt_t {}));
-                }
-            } else {
-                append(let_t::make(var, {}, {}));
+        if (new_ir_api_) {
+            for (int i = 0; i < 3; i++) {
+                local_sizes_[i] = var_t::make(local_size_type(),
+                        std::string("local_size_") + "012"[i]);
+                local_ids_[i] = var_t::make(
+                        local_id_type(), std::string("local_id_") + "012"[i]);
+                group_ids_[i] = var_t::make(
+                        group_id_type(), std::string("group_id_") + "012"[i]);
             }
-        }
+        } else {
+            for (int i = 0; i < interface.nargs(); i++) {
+                const auto &var = interface.arg_var(i);
+                if (var.type().is_ptr()) {
+                    if (var.type().is_slm()) {
+                        append(alloc_t::make(
+                                var, 0, alloc_kind_t::slm, stmt_t {}));
+                    } else {
+                        append(alloc_t::make(
+                                var, 0, alloc_kind_t::global, stmt_t {}));
+                    }
+                } else {
+                    if (!new_ir_api_) append(let_t::make(var, {}, {}));
+                }
+            }
 
-        for (int i = 0; i < 3; i++) {
-            group_ids_[i] = let(type_t::u32(), ir_builder_t::tg_idx(i), {});
-            local_ids_[i] = let(type_t::u16(), ir_builder_t::local_id(i), {});
-            local_sizes_[i]
-                    = let(type_t::u16(), ir_builder_t::local_size(i), {});
+            for (int i = 0; i < 3; i++) {
+                group_ids_[i]
+                        = let(group_id_type(), ir_builder_t::tg_idx(i), {});
+                local_ids_[i]
+                        = let(local_id_type(), ir_builder_t::local_id(i), {});
+                local_sizes_[i] = let(
+                        local_size_type(), ir_builder_t::local_size(i), {});
+            }
         }
     }
 
@@ -101,7 +119,11 @@ struct ctx_t {
                 append(funcs::zero_out(alloc_var, type.size()));
             }
         } else {
-            append(let_t::make(alloc_var, value, {}));
+            if (new_ir_api_) {
+                if (!value.is_empty()) append(assign_t::make(alloc_var, value));
+            } else {
+                append(let_t::make(alloc_var, value, {}));
+            }
         }
         return lval_t(alloc_var.as<var_t>());
     }
@@ -130,6 +152,10 @@ struct ctx_t {
         return let(value.type(), name, value);
     }
 
+    int slm_byte_offset() const { return slm_byte_offset_; }
+
+    void reserve_slm(int bytes) { slm_byte_offset_ += bytes; }
+
     void assume(const expr_t &e) { ctx_->add_constraint(e); }
 
     void begin_scope() { stmts_stack_.emplace(); }
@@ -155,8 +181,12 @@ struct ctx_t {
     const ir_context_t *ir_ctx() const { return ctx_; }
 
 private:
+    type_t local_id_type() const { return u16; }
+    type_t group_id_type() const { return u32; }
+    type_t local_size_type() const { return u16; }
+
     expr_t var(type_t type, const std::string &name) {
-        return var_t::make(type, name);
+        return var_t::make(type, ctx_->create_tmp_name(name));
     }
 
     stmt_t to_stmt() {
@@ -207,6 +237,8 @@ private:
     std::array<expr_t, 3> group_ids_;
     std::array<expr_t, 3> local_ids_;
     std::array<expr_t, 3> local_sizes_;
+    bool new_ir_api_ = false;
+    int slm_byte_offset_ = 0;
 };
 
 ctx_t &default_ctx() {
@@ -224,8 +256,9 @@ int min_pitch_2d() {
     return block_2d_pitch_alignment(default_ctx().ir_ctx()->hw().to_ngen());
 }
 
-void declare_kernel(const kernel_iface_t &interface, ir_context_t &ctx) {
-    default_ctx().declare_kernel(interface, ctx);
+void declare_kernel(
+        const kernel_iface_t &interface, ir_context_t &ctx, bool new_ir_api) {
+    default_ctx().declare_kernel(interface, ctx, new_ir_api);
 }
 
 kernel_t end_kernel() {
@@ -276,6 +309,11 @@ const expr_t &local_size(int idx) {
     return default_ctx().local_size(idx);
 }
 
+expr_t subgroup_id(int idx) {
+    int simd = default_ctx().ir_ctx()->exec_cfg().simd();
+    return extract((local_id(idx) / simd), 0);
+}
+
 expr_t arg(const std::string &name, bool allow_empty) {
     return default_ctx().arg(name, allow_empty);
 }
@@ -285,6 +323,10 @@ lval_t def(type_t type, const std::string &name, const expr_t &value,
     return default_ctx().def(type, name, value, force_alloc);
 }
 
+lval_t def(const std::string &name, const type_t &type, const expr_t &value) {
+    return def(type, name, value);
+}
+
 lval_t def(const std::string &name, const expr_t &value) {
     return def(value.type(), name, value);
 }
@@ -292,6 +334,27 @@ lval_t def(const std::string &name, const expr_t &value) {
 tensor_t def(
         const layout_t &layout, const std::string &name, const expr_t &value) {
     return default_ctx().def(layout, name, value);
+}
+
+tensor_t def_slm(layout_t layout, const std::string &name) {
+    auto alloc_elems = into<int>(layout.size() / layout.type().size());
+    auto buf = def(name, layout.type().slm()[alloc_elems]);
+    int bytes = (to_cpp<int>(layout.offset()) + alloc_elems)
+            * layout.type().size();
+    auto off = utils::div_up(
+            default_ctx().slm_byte_offset(), layout.type().size());
+    layout.set_offset(off);
+    default_ctx().reserve_slm(bytes);
+    return tensor_t(buf, layout);
+}
+
+expr_t iif(
+        const expr_t &cond, const expr_t &true_expr, const expr_t &false_expr) {
+    return iif_t::make(cond, true_expr, false_expr);
+}
+
+expr_t extract(const expr_t &expr, int lane) {
+    return shuffle_t::make(expr, {lane});
 }
 
 lval_t &lval_t::operator=(const expr_t &obj) {
@@ -308,7 +371,11 @@ expr_t let(const std::string &name, const expr_t &value) {
 }
 
 void assign(const expr_t &var, const expr_t &value) {
-    append(store_t::make(var, 0, value));
+    if (default_ctx().new_ir_api()) {
+        append(assign_t::make(var, value));
+    } else {
+        append(store_t::make(var, 0, value));
+    }
 }
 
 enum class send_kind_t { load, prefetch, store };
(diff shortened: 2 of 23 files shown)


src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -194,6 +254,19 @@ void if_(const expr_t &cond, const F &if_body, const G &else_body) {
     }
 }
 
+template <typename F>
+void _for(const expr_t &var, const expr_t &bound, const expr_t &step,
```
rjoursler, line 269: It would be nice to keep the format aligned between `_for`, `if_` and `while_`.

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -136,15 +183,30 @@ class lval_t {
...
 expr_t let(type_t type, const std::string &name, const expr_t &value);
 expr_t let(const std::string &name, const expr_t &value);
+tensor_t def_slm(layout_t layout, const std::string &name);
+
+expr_t iif(
+        const expr_t &cond, const expr_t &true_expr, const expr_t &false_expr);
+expr_t extract(const expr_t &expr, int lane);
+
+void assign(const expr_t &var, const expr_t &value);
+void func_call(const func_t &func, const std::vector<expr_t> &args = {});
```
rjoursler: Can we replace this with normal C++ function call semantics? I am thinking we could have something like:
```
struct func_t {
    ir::object_t func
    operator(std::vector<expr_t> & args) {
          ....
     }
}
```

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -39,39 +40,81 @@ struct send_hint_t {
...
+        global_tensor_t ret = *this;
+        ret.coord = coord;
+        ret.tile = tile;
+        return ret;
+    }
+
+    pvar_t major_stride_dim() const {
+        gpu_assert(strides.size() == 2);
+        for (auto &d : strides) {
+            if (is_one(strides.at(d))) return d;
```
rjoursler: I feel like we are heading in the wrong direction with this interface. In particular,  it is unclear how it should work for indeterminate scenarios like an `nx1` or `1x1` tensors? I am worried we will hit this if we specialize values while debugging as this  could completely change program behavior due to selecting the specialized stride. Do you have an intended mechanism to avoid this ambiguity?

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -136,15 +183,30 @@ class lval_t {
     expr_t var;
 };
 
+expr_t subgroup_id(int idx = 0);
 expr_t arg(const std::string &name, bool allow_empty = false);
+// TODO: Unify def() API, keep three versions:
+// 1. def(name, type, value)
+// 2. def(name, type)
+// 3. def(name, value)
+// name goes first in all three for consistency.
```
rjoursler, line 204: Do Todo?

src/gpu/intel/jit/ir/core.hpp
```diff
@@ -498,6 +499,8 @@ class type_t {
 
     bool is_slm() const { return any(attr() & type_attr_t::slm); }
 
+    type_t operator[](int elems) const { return with_elems(elems); }
```
rjoursler, line 502: I am not a fan of this, `operator[]` as it breaks the idiom that this operator is for indexing into collection types. Can we come up with a different short-hand for `with_elems`? Some ideas are using `operator()` or maybe we can create an interface like `with(T t)` as shorthand for the commonly used `with_X(T t)` interfaces?

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -30,34 +30,52 @@ namespace jit {
 namespace dsl {
 
 struct ctx_t {
+    bool new_ir_api() const { return new_ir_api_; }
 
-    void declare_kernel(const kernel_iface_t &interface, ir_context_t &ctx) {
+    void declare_kernel(const kernel_iface_t &interface, ir_context_t &ctx,
+            bool new_ir_api = false) {
+        slm_byte_offset_ = 0;
+        new_ir_api_ = new_ir_api;
```
rjoursler, line 39: Why do we need this? At this point we barely have any compatibility concerns, so I don't see why we would merge compatibility functionality at this point.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -30,34 +30,52 @@ namespace jit {
...
-                    append(alloc_t::make(var, 0, alloc_kind_t::slm, stmt_t {}));
+        if (new_ir_api_) {
+            for (int i = 0; i < 3; i++) {
+                local_sizes_[i] = var_t::make(local_size_type(),
+                        std::string("local_size_") + "012"[i]);
+                local_ids_[i] = var_t::make(
+                        local_id_type(), std::string("local_id_") + "012"[i]);
+                group_ids_[i] = var_t::make(
+                        group_id_type(), std::string("group_id_") + "012"[i]);
+            }
```
rjoursler, line 55: How are we going to handle logging kernel interfaces under this new API?

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -89,7 +107,7 @@ struct ctx_t {
             bool force_alloc = false) {
         auto type = type_t(
                 _type.kind(), _type.elems(), _type.attr() | type_attr_t::mut);
-        auto alloc_var = var(type, name);
+        auto alloc_var = var(type, ctx_->create_tmp_name(name));
```
rjoursler: This looks wrong, why is it here?

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Well, we do :) #3729 relies on the current semantics.

I have an implementation relying on the new semantics, so my plan is to merge the current work and then follow-up with specific adjustments. We can do all at once but I think it's less productive.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Good question, we need to think about it. I would add a section with an ordered list of arguments in the beginning of a kernel output.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: This is to avoid name duplicates in nested DSL, e.g.

```c++
void func() {
    auto tmp = def("tmp", ...);
    foo();
}

void foo() {
    auto tmp = def("tmp", ...);
}
```

In other words this is to track names and adjust to make names unique.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: Yep, would be nice but `_for` is reserved in oneDNN:

https://github.com/uxlfoundation/oneDNN/blob/0ccff502dd3d794c5720cf1350fc499f71a8af13/src/common/z_magic.hpp#L20

src/gpu/intel/jit/ir/core.hpp
echeresh: I agree to some extent, I came up with this because:

- It's shorter: `u32[4]` vs `u32.with(4)`
- It's for DSL and `int[10]` in C++ is easy to parse/read - `int` array of 10 elements
- Indexing into a type doesn't make much sense
- This is a core API so it gets idiomatic to a developer once you see a couple spots (and it's a common operation so people can get used to it from the very beginning)

My main motivation was to come up with some unambiguous shortcuts for common DSL constructs:

- `type_t::u32()` -> `u32`
- `u32.with_elems(4)` -> `u32[4]`

I don't like `operator()` that much (mainly because there is no similarity with C++ array syntax) but I'm fine to change it if you think it is a better choice.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: Similarly to the `new_ir_api` - I suggest to keep all `def()` versions and resolve once all affected PRs land.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: Yeah, I had the same thought. I can move these functions out of `global_tensor_t` but on your questions:

- major/minor behavior is well-defined even for `1x1` or `nx1` tensors as long as stride names are fixed
    - major is the first that is one (in an ordered list)
    - minor is the one that is not major

Other than that, this functionality is needed to determine row/column-major tensors and to differentiate between major dimension (with must be dense) and minor (usually not dense but may be dense as well).

src/gpu/intel/jit/ir/core.hpp
rjoursler: Thanks, for making the connection with array type syntax, it make sense to me to keep it as is in this case.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: Yeah, but why do we need the names to be unique? Similar to C++, we shouldn't have name conflicts within a given scope. Because of this, I would expect we have something more like:
```
void func() {
    begin_scope();
    auto tmp = def("tmp", ...);
    foo();
    end_scope();
}

void foo() {
    begin_scope();
    auto tmp = def("tmp", ...);
    end_scope();
}
```
and the fact that variables in different scopes have the same name should be largely irrelevant.

echeresh: Right, the problem here is that `foo()` doesn't define its own scope, it's all one scope from the DSL perspective. If you suggest to add `begin_scope()`/`end_scope()` manually into every such a function - I don't think it's a good idea as this makes DSL usage harder vs it could be and hard to enforce. Tracking/adjusting names, on the other hand, doesn't look like a a big deal.

src/gpu/intel/jit/dsl/dsl.hpp
rjoursler: > major is the first that is one (in an ordered list)

Lets modify this to be the first non-trivial dimension with stride 1 or the first dimension in the trivial case. I believe that would address my main concerns.

src/gpu/intel/jit/dsl/dsl.hpp
rjoursler: That's what I figured, what about just renaming `if_ -> _if` and `while_ -> _while`? This would also then be aligned with `_bool` which you introduced as well.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: I mainly dislike the divergence from normal languages design, so would prefer to avoid if it can be done in a simple and robust manner.
 
> If you suggest to add begin_scope()/end_scope() manually into every such a function - I don't think it's a good idea as this makes DSL usage harder 

Agreed, what do you think of using an RAII pattern:
```
void foo() {
    auto guard = dsl::scope_guard();
    ....
}
```

echeresh: > I mainly dislike the divergence from normal languages design

I'm mostly thinking about usability, divergence is not an issue per se to me. Do you see any particular issues?

The guard is not much different, right? It still imposes extra requirements to use DSL. I anticipate that requiring this will add more build-run-debug-rebuild iterations during development without tangible benefits.

rjoursler: >  Do you see any particular issues?

I don't see any obvious issues and we can go ahead with this design, but I suspect this will lead to weird interactions in the future. It is just unclear to me if they will be legitimately problematic. For example, I could easily see us injecting optimized subroutines rather than generating them in place to avoid redundant work. If we did that, we would get different IR between those methods which seems like it could expose issues. I could also see this generating noise when diffing good/bad IR to root cause bugs, and consequently slowing down debugging.

Edit: Another (but likely rare possibility) is we could generate a name collision via explicitly creating `A_0` and `create_tmp_name` generating `A_0`.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: Sure, thanks, added a couple helpers in `func_t` and dropped `func_call()`.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: @rjoursler I agree, this is not perfect. I'm just thinking this way: what happens more often: either we create a new function/split existing while developing a kernel or we compare two IR outputs. My experience tells me that the former takes more time. Another observation (from my side): IR (being simpler and more stable) itself is less often at fault comparing with assembly. I can recall ~a dozen of issues that I spotted comparing assembly vs maybe a couple by comparing IR.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: Ok, can we at least modify `create_tmp_name` so that it only applies a suffix when a collision is detected, this way we don't generate a lot of unnecessary suffices? It seems like we also should be applying this to all DSL variables, maybe the name change should be moved into `ctx_t::var()`?

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: @rjoursler Thanks, I'll implement this logic but I suggest to drop it from this PR for now. This is needed for specific 2D loads so I'll move this closer to the usage.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Thanks, implemented both suggestions.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/pass/simplify.cpp
```diff
@@ -2032,6 +2035,8 @@ class const_fold_helper_t {
 
             CASE(op_kind_t::_and)
             CASE(op_kind_t::_or)
+            CASE(op_kind_t::_shl)
```
hidefromkgb, line 2039: Is the lack of XOR intended?

echeresh: No, just missing, added as well. Thanks!

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/ir/core.cpp
```diff
@@ -405,6 +408,7 @@ DEFINE_BINARY_OPERATOR(<=, op_kind_t::_le)
 
 DEFINE_BINARY_OPERATOR(&, op_kind_t::_and)
 DEFINE_BINARY_OPERATOR(|, op_kind_t::_or)
+DEFINE_BINARY_OPERATOR(^, op_kind_t::_xor)
```
kealan-barbieri, line 413: should the shifts be in here too?

src/gpu/intel/jit/dsl/decl.hpp
Simonsays095, line 1: What is the purpose for this file? Is it just to allow `jit::dsl::<type>` in addition to `jit::<type>`?

echeresh: Correct, the end goal is to eliminate direct usage of IR objects in DSL kernels so we introduce `dsl::` aliases.
Type aliases could be used in IR code too, so we may want to move them to `ir/decl.hpp` in the future.

src/gpu/intel/jit/ir/core.cpp
echeresh: Yes, but they were already defined, at line 401, before this PR.

src/gpu/intel/jit/dsl/decl.hpp
Simonsays095: Why do we want to eliminate usage of IR objects in DSL? And why do we use static variables instead of `using type_t = jit::type_t`? I'm worried that when data types get added in the future (+etc, for any other types we alias in this way), we'll have to touch code in several places to extend them to all aliases.

src/gpu/intel/jit/ir/ir.hpp
```diff
@@ -55,7 +55,7 @@ class ir_context_t {
 
     std::string create_tmp_name(const std::string &prefix = "tmp") {
         int &id = prefix_ids_[prefix];
-        auto name = prefix + "_" + std::to_string(id);
+        auto name = (id == 0 ? prefix : prefix + "_" + std::to_string(id));
```
rjoursler: There is an missed edge case with this implementation:
```
var_t("A") -> "A";
var_t("A") -> "A_1"
var_t("A_1") -> "A_1"
```
To completely avoid this, could we use a less common symbol reserved for internal use? I am thinking something like `#`, so this becomes:
```
var_t("A") -> "A";
var_t("A") -> "A#1"
var_t("A#1") -> gpu_assertion failure
```

src/gpu/intel/jit/dsl/decl.hpp
echeresh: > Why do we want to eliminate usage of IR objects in DSL?

This is for simplicity, once we have IR/DSL separation, and we mostly use DSL for implementing kernels, then relying on two namespaces is not needed while we can just have one. These aliases help with that. We have a related discussion with Roy, I'll share a link.

> And why do we use static variables instead of using type_t = jit::type_t?

This is because of the plan to put IR into `ir` namespace and DSL into `dsl` namespace. This type alias will look like:

```
namespace dsl {
using type_t = ir::type_t;
}
```

Then aliases for separate types must be part of `dsl` namespace anyway.

> we'll have to touch code in several places to extend them to all aliases.

I agree but luckily types are extended not that often. When this gets stable, it's roughly once per architecture (at best), so maybe not that of a big deal.

But in general, I have a rule that duplication is fine as long as it won't noticeably interfere with regular development work. Let's say, there is some duplication that causes us to do some repetitive work ~once per year. Well, it might be actually more work to fix this vs. just accept and live with it.

src/gpu/intel/jit/ir/ir.hpp
echeresh: > var_t("A_1") -> "A_1"

@rjoursler This case should get resolved to `A_1_1` because `A_1` is not unique after previous lines, isn't it? What is missing to make this happen?

rjoursler: It's because the prefix in the last case is "A_1", so the lookup fails to find the collision:
```
prefix_ids_ == {}
var_t("A") -> "A"; -> prefix_ids_ == {{A, 1}}
var_t("A") -> "A_1" -> prefix_ids_ == {{A, 2}}
var_t("A_1") -> "A_1" -> prefix_ids_ == {{A, 2}, {A_1, 1}};
```

echeresh: Thanks, got it now. Wouldn't it be better to adjust the check to look through all registered names?

I don't like the `#` adjustment as it makes it harder to use editors to search for the identifier, e.g. vim by default won't recognize a name containing `#` as a valid identifier.

rjoursler: That's fine with me ( I am also not set on the `#` symbol, I was just using as a sample for the technique). I was mainly thinking that having some internal only symbol might be a bit simpler as it avoids some ambiguity in creation order which slightly complicates the logic. For example:
```
// Order 1
var_t("A") -> "A"
var_t("A") -> "A_1"
var_t("A_1") -> "A_1_1"

// Order 2
var_t("A") -> "A"
var_t("A_1") -> "A_1"
var_t("A") -> "A_2"(?)
```

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/ir/ir.hpp
echeresh: Yeah, I think we can get back to this later if this gets important. My immediate reaction is that names, suffixed with `_<number>` are rare so I wouldn't look into that.

PR #3772, opened 2025-08-15
xe: jit: dsl: add forward declaration concept to DSL

rjoursler:
While seeing if I can simplify tensor management in #3729, it seems like it would be beneficial to replace `tensor_config_t` with a structure like:
```
tensor_state_t {
    global_tensor_t global_tensor;
    tensor_t tensor;
    icoord_t tensor_coord;
    send_hint_t hint;
}
```
This structure needs setup before the k-loop is generated where `tensor` allocated. As such we can use a method to forward declare variables before they are allocated or initialized. 

@echeresh I feel like the abstraction provided by `global_tensor_t` and `tensor_t` isn't quite meshing with GEMM IR's needs. I don't feel like I should need to be creating this `tensor_state_t` structure within the GEMM IR, and should instead be relying on a DSL native construct. Do you have any suggestions for how we should modify the DSL? The core issue is that `tensor` maps to multiple `global_tensor` tiles and that the layout of `tensor` is fixed by the strategy, resulting in having to manually manage the correspondence.

### DIFF
--- a/src/gpu/intel/jit/dsl/dsl.cpp
+++ b/src/gpu/intel/jit/dsl/dsl.cpp
@@ -107,12 +107,12 @@ struct ctx_t {
         return def(value.type(), name, value);
     }
 
-    tensor_t def(const v2::layout_t &layout, const std::string &name,
+    tensor_t def(const layout_t &layout, const std::string &name,
             const expr_t &value = {}) {
         // Tensors need to be grf-aligned for loading/storing
         // TODO: IR should be modified to enable loading small tensors (such as
         // scalar values) without GRF alignment.
-        auto elems = std::max(layout.type().elems() * layout.elems(),
+        auto elems = std::max(into<int>(layout.type().elems() * layout.elems()),
                 grf_size() / layout.type().scalar().size());
         auto t = type_t(layout.type().kind(), elems);
         return {def(t, name, value, true), layout};
@@ -318,14 +318,14 @@ void scatter_send(const tensor_t &t, const global_tensor_t &g,
 void block_send(const tensor_t &t, const global_tensor_t &g,
         send_kind_t &op_kind, const icoord_t &base, const send_hint_t &hint) {
     bool is_prefetch = t.buf.is_empty();
-    auto &operation_tile = is_prefetch ? g.tile : t.layout.int_dim_sizes();
+    auto &operation_tile = is_prefetch ? g.tile : t.layout.tile();
 
     pvar_t w_dim;
     tile_t tile;
     for (auto &var : operation_tile) {
         if (is_const(g.strides[var]) && to_cpp<dim_t>(g.strides[var]) == 1
-                && !t.layout.is_scalar()) {
-            tile[var] = t.layout.blocks()[0].int_size();
+                && t.layout.elems() != 1) {
+            tile[var] = t.layout.blocks()[0].block;
             gpu_assert(t.layout.blocks()[0].dim == var);
             w_dim = var;
         } else {
@@ -397,7 +397,7 @@ void block_2d_send(const conf_2d_t &conf, const tensor_t &t,
         const send_hint_t &hint) {
 
     bool is_prefetch = t.buf.is_empty();
-    auto &operation_tile = is_prefetch ? g.tile : t.layout.int_dim_sizes();
+    auto &operation_tile = is_prefetch ? g.tile : t.layout.tile();
 
     pvar_t w_dim = conf.w_dim;
     pvar_t h_dim;
@@ -449,7 +449,7 @@ void block_2d_send(const conf_2d_t &conf, const tensor_t &t,
 void send(const tensor_t &t, const global_tensor_t &g, send_kind_t op_kind,
         const icoord_t &base, const send_hint_t &hint) {
     bool is_prefetch = t.buf.is_empty();
-    auto &operation_tile = is_prefetch ? g.tile : t.layout.int_dim_sizes();
+    auto &operation_tile = is_prefetch ? g.tile : t.layout.tile();
     pvar_t w_dim;
     for (auto &var : operation_tile) {
         if (is_const(g.strides[var]) && to_cpp<dim_t>(g.strides[var]) == 1) {
@@ -466,8 +466,8 @@ void send(const tensor_t &t, const global_tensor_t &g, send_kind_t op_kind,
         auto conf = [&]() -> conf_2d_t {
             if (is_prefetch) { return {g.type, w_dim, 0, false, false, false}; }
             auto &l = t.layout;
-            int pack_dim = l.blocks()[0].int_size() * l.type().size() == 4;
-            int pack_size = l.blocks()[pack_dim].int_size();
+            int pack_dim = l.blocks()[0].block * l.type().size() == 4;
+            int pack_size = into<int>(l.blocks()[pack_dim].block);
             bool is_transpose_vnni = l.blocks()[pack_dim].dim != w_dim;
             bool is_vnni = pack_dim == 1 && !is_transpose_vnni;
             bool is_store = op_kind == send_kind_t::store;
@@ -481,7 +481,7 @@ void send(const tensor_t &t, const global_tensor_t &g, send_kind_t op_kind,
         }
     }
 
-    if (is_prefetch || t.layout.is_scalar()
+    if (is_prefetch || t.layout.elems() == 1
             || t.layout.blocks()[0].dim == w_dim) {
         block_send(t, g, op_kind, base, hint);
     } else {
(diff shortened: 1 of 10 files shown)


echeresh:
> @echeresh I feel like the abstraction provided by global_tensor_t and tensor_t isn't quite meshing with GEMM IR's needs. I don't feel like I should need to be creating this tensor_state_t structure within the GEMM IR, and should instead be relying on a DSL native construct. Do you have any suggestions for how we should modify the DSL? The core issue is that tensor maps to multiple global_tensor tiles and that the layout of tensor is fixed by the strategy, resulting in having to manually manage the correspondence.

I took a look at the implementation. Please check if my understanding of the algorithm is correct, I came up with the pseudo-code/DSL mix below. Does it look remotely correct? No iterators for now just to keep it simple.

I'm trying to understand why we need extra abstractions besides what we already have in the DSL.

```c++
coord_t a_coord = ...;
coord_t b_coord = ...;
_for(k_outer, 0, K, /*step=*/k_unroll_blk, [&]() {
	auto A = def("A", a_reg_layout(...)); // This layout is with A_copies.
	auto B = def("B", b_reg_layout(...)); // This layout is with B_copies.
	int a_k_blk = A.layout.dim(k_dim);
	int b_k_blk = B.layout.dim(k_dim);
	for (int k_inner = 0; k_inner < k_unroll_blk; k_inner += k_blk) { // This loop is fully unrolled.
		bool do_a_load = (k_inner % a_k_blk == 0);
		bool do_b_load = (k_inner % b_k_blk == 0);
		if (do_a_load) {
			auto tile = A.tile().with(k_dim, a_k_blk / A_copies);
			auto src = A_mem.sub(a_coord.with(k_dim, k_outer + k_inner), tile);
			auto dst = A.sub(icoord_t().with(k_dim, k_inner % a_k_blk), tile);
			do_load(src, dst, a_transform);
		}
		if (do_b_load) {
			auto tile = B.tile().with(k_dim, b_k_blk / B_copies);
			auto src = B_mem.sub(b_coord.with(k_dim, k_outer + k_inner), tile);
			auto dst = B.sub(icoord_t().with(k_dim, k_inner % b_k_blk), tile);
			do_load(src, dst, b_transform);
		}
		do_mma(C, A, B, k_inner);
	}
});
```

rjoursler:
Yeah, that is close for the non-pipelined case, although it looks like `{a,b}_k_block` isn't quite being used correctly. For the A tensor, computations should look more like:
```C++
	int a_k_blk = A.layout.dim(k_dim) / A_copies;
	for (int k_inner = 0; k_inner < k_unroll_blk; k_inner += k_blk) { // This loop is fully unrolled.
		bool do_a_load = (k_inner % a_k_blk == 0);
		if (do_a_load) {
			auto tile = A.tile().with(k_dim, a_k_blk);
			auto src = A_mem.sub(a_coord.with(k_dim, k_outer + k_inner), tile);
			auto dst = A.sub(icoord_t().with(k_dim,  k_inner % A.layout.dim(k_dim), tile);
	        }
```
 I am effectively working to consolidate the indexing logic related to the A/B/C loads and prefetches into a shared abstraction. The end goal would be to have something like:

```C++
// The actual strategy are an input to the loop builder as they change for edge case handling
A_state(GEMMStrategy.A, ...);
B_state(GEMMStrategy.B, ...);
for (int k_inner = 0; k_inner < k_unroll_blk; k_inner += k_blk) { // This loop is fully unrolled.
    if (A_state.do_load(k_inner)) { load(A_state, k_outer, k_inner); }
    if (B_state.do_load(k_inner)) { load(B_state, k_outer, k_inner); }
```

echeresh:
@rjoursler Thanks. But looks like the `load()` functions from your example can just take these parameters directly? Or do you  think this needs a separate abstraction? At the end the difference is a function taking either one "state" object vs say 3-4 arguments.

As a side comment: I noticed in your prototype that a lot of arithmetic, tile/layout manipulation and general setup happens at a higher level, I commented on that already [here](https://github.com/uxlfoundation/oneDNN/pull/3729#discussion_r2264159601). And as the result, at a lower level the implementation has small lambdas or functions taking one/two arguments that operate based off of some preliminarily prepared setup (either config, or some either structure). What I'm talking about is:

```c++
A_prefetch_config = prepare_prefetch_config(A);
A_prefetch_main_config = prepare_prefetch_main_config(A);
...
loop_iteration(A_prefetch_config, A_load_config);
```

However, if we could restructure this to move all this main/short, warmup, prefetch/load logic a few levels down to shift verbosity/complexity to a lower level - it could get simpler to understand. I would focus on that first and see what it will look like after that. I'm mentioning this here, because introducing a "state" is rather a move in the other direction.

Otherwise, I agree that it makes sense to combine concepts if we constantly use them together. But my suggestion would be to work towards a (simpler) version that hides complexity better, maybe with duplication. I guess after that we'll have only one problem to solve - spot repeating patterns and come up with corresponding abstractions.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -306,6 +306,21 @@ const expr_t &local_size(int idx) {
     return default_ctx().local_size(idx);
 }
 
+declaration_t declare(const std::string &name, type_t type) {
+    return declaration_t(name, type);
+}
+declaration_t declare(
+        const std::string &name, const layout_t &layout, bool is_mutable) {
```
echeresh, line 313: Can you please comment on what are potential use cases for  that?

rjoursler:
>  Or do you think this needs a separate abstraction?

I think we will want an abstraction to translate `GEMMStrategy` structures (mostly [this](https://github.com/uxlfoundation/oneDNN/blob/46c2a0c32e5e63b867edb12e4585674bd77a4162/src/gpu/intel/gemm/jit/include/gemmstone/strategy.hpp#L70-L89) + matrix k_blk and k_pipeline_offset) into its "DSL equivalent". There first two things here, (unrelated to this PR, for the third, see my response to your other comment) are:
`global_tensor_t` is inconsistent in its interaction with generation state:
```C++
struct global_tensor_t {
    expr_t buf; // Constant
    type_t type; // Constant
    expr_t base_offset; // Constant
    coord_t coord; // Varies with k-loop builder
    pvar_map_t<expr_t> strides; // Constant
    pvar_map_t<expr_t> sizes; // Constant
    tile_t tile; // Constant within calls to k-loop builder
```
Because of coord, we end up need to do something like one of the following:

    1. Mutate global_tensor_t object with current coord, this is what gemm_ir currently does.
    2. Introduce interface like `with()` from your sample
    3. Modify load/store/prefetch interface to accepting coord_t

And it is unclear to me which should be the DSL idiomatic solution (although 3 looks the most reasonable to me). 

The second thing I noticed is the new abstraction will have to represent some notion of multi-buffering. The current DSL interfaces do not provide an abstraction for multi-buffering. Since this is a common usage mode for tensors, I am wondering if it make sense to add some kind of abstractions to simplify multi-buffering.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: There are a couple of things I have encountered that are the reason for my proposal. The first is to enable some method to decouple variable specialization from the kernel interface such as [here](https://github.com/uxlfoundation/oneDNN/blob/69c8f96e2d8e0d9608ebc2d9aee1689ca495ebd1/src/gpu/intel/gemm/jit/gen_kernel.cpp#L1022-L1032). More changes would be required to get the "deduplicated" variables though.

The second is to decouple the matrix loading abstraction from k-loop building.  The key issue is that this abstraction needs inputs to construct a `global_tensor_t::tile` and `tensor_t::layout` as they vary between k-loop builder calls. I am considering just directly  storing these in a `global_tensor_t` and a `tensor_t`, rather than these sub-pieces. This should clarify how these are tied to their use, and hopefully remove some management logic for when these objects are used. This can induce some unnecessary coupling though, unless we cannot create the `tensor_t` object independently from the DSL code generation without forward declaration.

echeresh:
> 1. Mutate global_tensor_t object with current coord, this is what gemm_ir currently does.
> 2. Introduce interface like `with()` from your sample
> 3. Modify load/store/prefetch interface to accepting coord_t

1 and 2 look like reasonable options (2 here is a way to use 1 though, `with()` is used to construct coordinates), we discussed load/store taking a global tensor (for memory buffer) and a tensor (for register file buffer) including all coordinates and tile sizes as part of the tensor objects. Do you see any issues with that?

Anyway, to me this doesn't look too critical at this point. I think your main concern is duplication but once there is a state object, someone should still set the coordinate in it so I don't see why it's much different from creating a new `global_tensor_t` with an updated coordinate in it.

> The new abstraction will have to represent some notion of multi-buffering

Can we come up with an implementation based on the current abstractions? I suggest to not spend much time on that for now as we need to keep newer architectures in mind where buffering may need a different approach and/or abstractions.

Issue #3749, opened 2025-08-12
[PPC64] '-mmma' requiring instructions not guarded by __MMA__


spalicki:
After merging #3156 our automated conda build system started showing build issues where `__builtin_vsx_lxvp`, `__builtin_vsx_disassemble_pair` and similar instructions are directly used. Some code containing those is guarded with `__MMA__` e.g. entire `src/cpu/ppc64/ppc64_gemm_s8x8s32.cpp`, but code in `src/cpu/ppc64/ppc64_gemm_s8x8s32.hpp` is not:
```
 │ │ In file included from $SRC_DIR/src/cpu/ppc64/gemm/gemm_driver.cpp:38:
 │ │ $SRC_DIR/src/cpu/ppc64/ppc64_gemm_s8x8s32.hpp: In function 'int dnnl::impl::pack_N8_8bit_V2_lxvp_signed(dim_t, dim_t, const BufType*, dim_t, uint8_t*, bool) [with VecType = __vector(16) unsigned char; BufType = unsigned char]':
 │ │ $SRC_DIR/src/cpu/ppc64/ppc64_gemm_s8x8s32.hpp:2347:38: error: '__builtin_vsx_lxvp' requires the '-mmma' option
 │ │  2347 |             row1 = __builtin_vsx_lxvp(
 │ │       |                    ~~~~~~~~~~~~~~~~~~^
 │ │  2348 |                     0, reinterpret_cast<__vector_pair *>(&b[0]));
 │ │       |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
We need all code requiring `-mmma` build option guarded with `__MMA__`.

spalicki:
@Tiwari-Avanish Can you help with this issue?

vpirogov:
Related conda-forge build fail: https://github.com/conda-forge/onednn-feedstock/pull/86

vpirogov:
Does it make sense to temporarily throw in `-mmma` option to at least make the packages buildable?

spalicki:
> Does it make sense to temporarily throw in `-mmma` option to at least make the packages buildable?

Let me check if this would actually work, but according to GCC `The -mma option requires that the option -mcpu=power10 (or later) is enabled.`, which would break compatibility with any earlier CPU (older than late 2021).

vpirogov:
That's definitely not perfect. Having no conda-forge packages for oneDNN is even worse.

[...]

PR #3773, opened 2025-08-15
cmake: add "-mcpu=power10 -mmma" to cross-compilation flags

spalicki:
Required to fix build issues with new GEMM implementation for conda-forge release.

Workaround for #3749.

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -428,6 +428,15 @@ elseif(UNIX OR MINGW)
             # For native compilation tune for the host processor
             if (CMAKE_SYSTEM_PROCESSOR STREQUAL CMAKE_HOST_SYSTEM_PROCESSOR)
                 append(DEF_ARCH_OPT_FLAGS "-mcpu=native")
+            else()
+                # The new GEMM/reorder implementation does not fully guard for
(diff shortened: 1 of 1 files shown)


cmake/platform.cmake
```diff
@@ -428,6 +428,8 @@ elseif(UNIX OR MINGW)
             # For native compilation tune for the host processor
             if (CMAKE_SYSTEM_PROCESSOR STREQUAL CMAKE_HOST_SYSTEM_PROCESSOR)
                 append(DEF_ARCH_OPT_FLAGS "-mcpu=native")
+            else()
+                append(DEF_ARCH_OPT_FLAGS "-mcpu=power10 -mmma")
```
mzhukova, line 439: just for my own knowledge, could you elaborate what exactly is going wrong? 
also, could we add a comment here to explain the addition?

vpirogov: A comment would definitely help future IBM developers.

spalicki: New implementation added in #3156 does not fully guard for instructions added in newer ISAs (Power9/10).

mzhukova: thanks for the pointer! this make sense

PR #3748, opened 2025-08-12
graph: backend: dnnl: fuse select to optimize gemma sdpa decompose kernel

xiang1guo:
oneDNN GPU matmul support post select fusion when select's 1st input connected to previous matmul, enable such pattern in CPU sdpa decompose kernel. 

- [ ] Relax decompose kernel pattern check
- [ ] Fuse select into matmul when connection between mm1 and select matched
- [ ] Added 2 more cases for validation
- [ ] Performance speedup is 10x against legacy pattern.

PR #3783, opened 2025-08-19
graph: backend: dnnl: fix post select md creation

xiang1guo:
Segfault issue for Gemma sdpa case in decompose kernel, found in local test, missed part of changes for select, original CI didn't catch such issue, it was caused by code rebase in #3748

### DIFF
--- a/src/graph/backend/dnnl/fusion_info.cpp
+++ b/src/graph/backend/dnnl/fusion_info.cpp
@@ -298,6 +298,19 @@ dnnl::primitive_attr make_dnnl_primitive_attr(
                     sum_dt = dnnl::memory::data_type::s8;
                 }
                 dnnl_pops.append_sum(scale, zp, sum_dt);
+            } else if (alg == dnnl::algorithm::binary_select) {
+                // post-binary
+                VCHECK_FUSION_INFO(
+                        extra_inputs.size() == 2 && scale == 1.f && zp == 0,
+                        attr,
+                        "%s post-binary_select only has 2 extra inputs and "
+                        "doesn't support input scale and zp",
+                        op->get_name().c_str());
+                auto md1 = make_dnnl_memory_desc(psrc);
+                auto psrc_cond = op->get_input_value(extra_inputs[1])
+                                         ->get_logical_tensor();
+                auto md2 = make_dnnl_memory_desc(psrc_cond);
+                dnnl_pops.append_binary(alg, md1, md2);
             } else {
                 // post-binary
                 VCHECK_FUSION_INFO(


src/graph/backend/dnnl/fusion_info.cpp
```diff
@@ -298,6 +298,21 @@ dnnl::primitive_attr make_dnnl_primitive_attr(
...
+                // post-binary
+                VCHECK_FUSION_INFO(
+                        extra_inputs.size() == 2 && scale == 1.f && zp == 0,
+                        attr,
+                        "%s post-binary_select only has 2 extra inputs and "
+                        "doesn't support input scale and zp",
+                        op->get_name().c_str());
+                auto md1 = make_dnnl_memory_desc(psrc);
+                if (op->get_kind() == op_kind::dnnl_convolution)
+                    md1 = to_format_any(md1);
```
ElaineBao: Can you please elaborate why conv needs this?

xiang1guo: Just a copy from below code handling, the original logic comes from this [PR](https://github.com/intel-innersource/libraries.performance.math.onednn/pull/8040), but we don't have such a pattern(conv + binary select) now.

ElaineBao: Not sure whether that perf optimization is valid for conv+select, how about leaving this change to the future if it's really needed.

xiang1guo: Updated! Thanks~

ElaineBao (approved):

xiang1guo:
> Can you please explain why it escaped from the previous CI?

CI ran into large partition kernel. I tested locally with OMP_NUM_THREADS=4 and found the segfault issue with the decompose kernel.

TaoLv:
> > Can you please explain why it escaped from the previous CI?
> 
> CI ran into large partition kernel. I tested locally with OMP_NUM_THREADS=4 and found the segfault issue with the decompose kernel.

Since the previous PR (and this one) mainly targets decomposition kernel, to trigger the kernel, I would expect either a nightly scope job is triggered or eigen threadpool is enabled during pre-ci.

xiang1guo:
> > > Can you please explain why it escaped from the previous CI?
> > 
> > 
> > CI ran into large partition kernel. I tested locally with OMP_NUM_THREADS=4 and found the segfault issue with the decompose kernel.
> 
> Since the previous PR (and this one) mainly targets decomposition kernel, to trigger the kernel, I would expect either a nightly scope job is triggered or eigen threadpool is enabled during pre-ci.

Yes, I launched a nightly job, will also trigger eigen threadpool during pre-ci. Thanks!

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

PR #3731, opened 2025-08-07
cpu: aarch64: brgemm: Extend support for bf16 in BRGEMM Kernel

vishwascm:
# Description

This PR extends the BRGEMM (Batch-Reduce General Matrix Multiplication) kernel to support bf16 data type.

Major Code Changes:
• Support for data type bf16:bf16:f32 and bf16:bf16:bf16
• Support for bias data type bf16 and f32
Other minor changes added accordingly in some files.

Perf Timings (ms)  Compared with f32 for single core:
| Shapes | f32:f32:f32 | bf16:bf16:f32 | bf16:bf16:bf16 | 
|:----------:|:----------:|:----------:|:----------:|
|512x768:768x768 | 8.02 | 3.94 | 3.97 |
|197x3072:3072x768 | 30.93 | 7.07 | 7.08 |

## Checklist
General
[✓] Do all unit and benchdnn tests (make test and make test_benchdnn_*) pass locally for each commit? Yes
Test output is same with and without this commit.

### Gtest:
```./test_matmul```
[----------] Global test environment tear-down
[==========] 64 tests from 8 test suites ran. (22 ms total)
[  PASSED  ] 52 tests.
[  SKIPPED ] 12 tests, listed below:
[  SKIPPED ] TensorDims/attr_test_t.TestMatmulShouldCallSameImplementationWithAttributes/0
[  SKIPPED ] TensorDims/attr_test_t.TestMatmulShouldCallSameImplementationWithAttributes/1
[  SKIPPED ] TensorDims/attr_test_t.TestMatmulShouldCallSameImplementationWithAttributes/2
[  SKIPPED ] TensorDims/attr_test_t.TestMatmulShouldCallSameImplementationWithAttributes/3
[  SKIPPED ] TensorDims/attr_test_t.TestMatmulShouldCallSameImplementationWithAttributes/4
[  SKIPPED ] Generic_f16/iface.TestsMatMul/0
[  SKIPPED ] Generic_f16/iface.TestsMatMul/1
[  SKIPPED ] Generic_f16/iface.TestsMatMul/2
[  SKIPPED ] Generic_f16/iface.TestsMatMul/3
[  SKIPPED ] Generic_f16/iface.TestsMatMul/4
[  SKIPPED ] Generic_f16/iface.TestsMatMul/5
[  SKIPPED ] Generic_f16/iface.TestsMatMul/6

### ```make test```
98% tests passed, 4 tests failed out of 225

Total Test time (real) = 922.55 sec

[...]

PR #3786, opened 2025-08-19
cpu: aarch64: Add optimised path for MxK:Kx1 GEMV

karmeh01:
### Description
This PR adds an extra path to the brgemm matmul kernel to optimise for the MxK:Kx1 GEMV case.

### Speed-up in oneDNN
Compared to the existing implementation for wtag=ab and wtag=ba, improvements range from 1.25x to 5.5x for the shapes tested.
[FP32 results](https://github.com/user-attachments/files/21839236/brg_to_pass_tests_MxN.pdf)

### Tests
`./benchdnn --matmul --batch=inputs/matmul/test_matmul_all`

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021-2023 Intel Corporation
 * Copyright 2024-2025 FUJITSU LIMITED
-* Copyright 2024 Arm Ltd. and affiliates
+* Copyright 2024-2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -241,6 +241,7 @@ private:
 
     PReg ld_full_mask = PReg(2);
     PReg ld_tail_mask = PReg(3);
+    PReg gemv_tail_mask = PReg(4);
 
     ZReg accm(int ld_block, int bd, int ld) {
         return ZReg(max_effective_vregs - 1 - (bd * ld_block + ld));
@@ -306,6 +307,7 @@ private:
     void apply_alpha_beta(int bd_block, int ld_block, bool is_ld_tail);
     void apply_post_ops(int bd_block, int ld_block2, int ldb_and_bdb_offset,
             bool is_ld_tail);
+    void sum_into_one_lane(int bd_block, int ld_block2, bool is_ld_tail);
     void restore_A_B_matrices();
     void set_A_B_matrices();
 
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -314,7 +316,8 @@ struct jit_brgemm_kernel_t : public jit_generator {
 
     void dot_product(ZReg z1, ZReg z2, ZReg z3);
     void gemm_microkernel_sve512(int bd_block2, bool is_bdb_tail, int ld_block,
-            bool is_rd_tail, bool is_ld_tail, int vpad, int rows_for_rd_tail);
+            bool is_rd_tail, bool is_ld_tail, int vpad, int rows_for_rd_tail,
```
Sqvid: I think this function ought to be renamed if we are using it generically. There is [another open PR that does this](https://github.com/uxlfoundation/oneDNN/pull/3731) so you may be able to rebase depending on merge order.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1150,7 +1155,16 @@ void jit_brgemm_kernel_t::store_accumulators_without_post_ops(
                 base_offset = offset;
                 x_addr = reg_tmp_;
             }
-            ST_MUL_VL(st1w, zmm.s, mask, x_addr, offset - base_offset, 4);
+            if (brg.LDB == 1 && brg.beta == 0.f) {
```
Sqvid: probably makes sense to factor out `brg.LDB == 1` into ` bool is_gemv` since you use it several times after.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1364,7 +1411,7 @@ void jit_brgemm_kernel_t::compute_int8_compensation(int rd_loop, int bd_b,
 
 void jit_brgemm_kernel_t::gemm_microkernel_sve512(int bd_block2,
         bool is_bdb_tail, int ld_block2, bool is_rd_tail, bool is_ld_tail,
-        int vpad, int rows_for_rd_tail) {
+        int vpad, int rows_for_rd_tail, bool is_gemv) {
```
Sqvid: I wonder if signaling this via a `bool` will lock you in in the future? If I understand this correctly, you are optimising the case where the the RHS is a matrix and the LHS is a `Kx1` vector. This means that if you add an optimisation for `1xK x KxM` later, then a bool won't be rich enough to capture this right?

karmeh01: You’re right. Based on preliminary results from ongoing work, it seems unlikely that optimisations for the 1xK x KxM case will end up in this file. That said, if we did need to handle them here, I think there are three options: introduce an enum, add a new bool, or extend the current bool to cover both cases and then add another conditional further down in the code. Which of these approaches do you think makes the most sense?

Sqvid: I guess in the case where we are adding several GEMV-specific optimisations it might make sense to have a separate `gemv_microkernel()` method.

jondea:
This can be closed. This work went directly into #4351. Thanks again for this excellent work!

PR #3750, opened 2025-08-12
(new API) src: gpu: intel: add user-supplied precomputed zps to gemm

hidefromkgb:
This is the new-API version of #3222, with the API concept provided by the oneDNN Arch team.

PR #3794, opened 2025-08-20
api: introduce precomputed reductions primitive attribute

dzarukin:
An implementation for RFC #3136.


dzarukin:
make test

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -158,6 +159,28 @@ zero points factors are used.
...
+
+`ARG` specifies which memory argument will be modified. Supported values are:
+  - `src` corresponds to `DNNL_ARG_SRC`
+  - `wei` corresponds to `DNNL_ARG_WEIGHTS`
+
+`POLICY` has the same semantics and meaning as for `--attr-scales`. Supported
+values are:
+  - `per_tensor`
+
+`DATA_TYPE` specifies data type for zero points. Supported values are
```
densamoilov: Do you mention "zero points" here intentionally? If yes then maybe it would make sense to mention how those are related to the precomputed reductions?

include/oneapi/dnnl/dnnl.h
```diff
@@ -479,20 +479,48 @@ dnnl_status_t DNNL_API dnnl_primitive_attr_set_zero_points_mask(
...
+/// @param group_ndims Number of group dimensions.
+/// @param group_dims Precomputed reduction factors correspondence groups that
+///     define the correspondence between the tensor dimensions and the
+///     precomputed reductions array.
+///     The group dimensions should be only provided for each logical dimension
+///     that has the bit set correspondence mask @p mask set.
+/// @param data_type Precomputed reduction factors data_type.
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+dnnl_status_t DNNL_API dnnl_primitive_attr_set_precomputed_reductions(
```
densamoilov, line 552: It seems that we would need to add this new attribute to documentation: https://uxlfoundation.github.io/oneDNN/dev_guide_attributes.html

A tutorial example could be nice too.

tests/benchdnn/dnn_types.hpp
```diff
@@ -418,6 +473,9 @@ struct attr_t {
 
     void insert(const arg_scales_t &as) { this->scales = as; }
     void insert(const zero_points_t &zp) { this->zero_points = zp; }
+    void insert(const precomputed_reductions_t &pl) {
```
echeresh:
```suggestion
    void insert(const precomputed_reductions_t &pr) {
```

src/gpu/intel/matmul/ref.cpp
```diff
@@ -205,11 +208,40 @@ status_t ref_t::execute_ref(const exec_ctx_t &ctx) const {
     const dim_t src_zp_stride_k = src_zp_strides[a_d.ndims() - 1];
     const dim_t src_zp_stride_m = src_zp_strides[a_d.ndims() - 2];
 
+    const auto &attr_pr = pd()->attr()->precomputed_reductions_;
+    const int src_pr_mask = attr_pr.get_mask(DNNL_ARG_SRC);
+    const auto src_pr_group_k = attr_pr.get_group(DNNL_ARG_SRC, 1);
+    const auto src_pr_ngroups_k = K / src_pr_group_k;
+    // Identify src_pr dimensions as user may not pass them.
```
echeresh, line 215: I see that in `ref_matmul_int8.cpp` memory descriptor is access from the attribute:

```
CHECK(attr_pr.get(DNNL_ARG_SRC).get_md(src_pr_md, *src_d.md_));
```

Is it different here? Or another question: is it optional for the user to specify dimensions for precomputed reductions memory?

tests/benchdnn/doc/knobs_attr.md
dzarukin: Copy-paste error. Fixed, thanks.

tests/benchdnn/dnn_types.hpp
dzarukin: Bad renaming process. Thanks for the catch, updated.

src/gpu/intel/matmul/ref.cpp
dzarukin: To some extent it is. The problem is gpu reference version demands stride values explicitly and memory desc obtained with mask=0 will have strides `{1, 1}` for m and k, or k and n, which means `per_oc` policy, while the policy was `common`. That's the primary reason it doesn't work the same way as it does on CPU which utilizes a special function for that memory desc (which we can't pass to CL kernel).

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -2707,6 +2707,8 @@ typedef const struct dnnl_primitive *const_dnnl_primitive_t;
 #define DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) \
     (DNNL_ARG_ATTR_MULTIPLE_POST_OP_BASE * ((idx) + 1))
 
+#define DNNL_ARG_ATTR_PRECOMPUTED_REDUCTIONS 1048576
```
hidefromkgb: This collides with `DNNL_ARG_ATTR_MULTIPLE_POST_OP(31)`.
`1048576 << 1`, in its turn, prohibits adding further post-ops transparently if we ever want to.

I suggest to use 512 as the flag value, since `DNNL_ARG_ATTR_OUTPUT_SCALES = 513` is not being used for at least 2 major versions (2.x and 3.x), and isn't even in the codebase anymore, to say nothing of the docs.

include/oneapi/dnnl/dnnl.h
```diff
@@ -436,7 +436,7 @@ dnnl_status_t DNNL_API dnnl_primitive_attr_set_scales_mask(
 ///     The set i-th bit indicates that a dedicated scaling factor is used for
 ///     each index along that dimension. Set the mask to 0 to use a common
 ///     scaling factor for the whole output tensor.
-/// @param ndims Number of group dimensions.
+/// @param group_ndims Number of group dimensions.
```
hidefromkgb, line 439: [nitpick]
I believe `s/ndims/group_ndims/` here and below should be a separate commit within this PR.

include/oneapi/dnnl/dnnl.h
dzarukin: I tried my best and crafted to examples in quantization page explaining advanced quantization per se, and precomputed reductions on top of that. Feel free to review.

include/oneapi/dnnl/dnnl_types.h
dzarukin: Updated that part, thanks.

include/oneapi/dnnl/dnnl.h
dzarukin: I agree, and more micro style changes should be moved to dedicated commits, I'm lazy this time, sorry.

hidefromkgb: I already have those changes split, and I can give you a .diff; do you need it?

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -2704,7 +2707,8 @@ typedef const struct dnnl_primitive *const_dnnl_primitive_t;
 /// Starting point for a binary post operation.
 #define DNNL_ARG_ATTR_MULTIPLE_POST_OP_BASE 32768
 
-/// Arguments for a binary post operation. Up to 32 arguments are supported.
+/// Arguments for a binary post operation. Up to 31 arguments are supported,
```
hidefromkgb: [nitpick]
```suggestion
/// Arguments for a binary post operation. Up to 32 arguments are supported,
```

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -2680,6 +2680,9 @@ typedef const struct dnnl_primitive *const_dnnl_primitive_t;
 /// Dropout RNG seed value passed via a buffer.
 #define DNNL_ARG_ATTR_DROPOUT_SEED 511
 
+/// Precomputed reductions argument.
+#define DNNL_ARG_ATTR_PRECOMPUTED_REDUCTIONS 512
+
 /// Output scaling factors provided at execution time.
 /// Deprecated value.
 #define DNNL_ARG_ATTR_OUTPUT_SCALES 513
```
hidefromkgb, line 2690: `DNNL_ARG_ATTR_OUTPUT_SCALES` should be removed as it collides with `DNNL_ARG_ATTR_PRECOMPUTED_REDUCTIONS | DNNL_ARG_A`.

dzarukin: It can't be removed until the next major version.

hidefromkgb: OK, got it. Then I'd suggest to update the comment above it:
```
/// Deprecated value, collides with DNNL_ARG_ATTR_PRECOMPUTED_REDUCTIONS | DNNL_ARG_SRC, do not use.
```

include/oneapi/dnnl/dnnl.h
densamoilov: looks very nice, thanks!

doc/programming_model/attributes_quantization.md
```diff
@@ -351,6 +351,140 @@ channel scaling.
...
+   auto matmul_pd = dnnl::matmul::primitive_desc(
+           engine,
+           src_f16_any_md,
+           wei_s8_any_md,
+           dst_f16_any_md,
+           attr);   // the attributes describe the quantization flow
+// ...
+~~~
+
+#### Example 4: matmul with precomputed reductions and advanced quantization
```
mzhukova, line 400: (minor) new examples are technically under "dev_guide_attributes_quantization_scales / Argument Scaling", would it be more clear to create a new section for examples (esp since the idea to showcase new attr)?

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -8,8 +8,9 @@
     --attr-rounding-mode=ARG:MODE[+...]
     --attr-deterministic=BOOL
     --attr-dropout=PROBABILITY[:SEED[:TAG]]
-    --attr-scales=ARG:POLICY[:SCALE[:DATA_TYPE[:GROUPS]]][+...]
-    --attr-zero-points=ARG:POLICY[:ZEROPOINT[:DATA_TYPE[:GROUPS]]][+...]
+    --attr-scales=ARG:POLICY[:SCALE][:DATA_TYPE[:GROUPS]][+...]
+    --attr-zero-points=ARG:POLICY[:ZEROPOINT][:DATA_TYPE[:GROUPS]][+...]
+    --attr-precomputed-reductions=ARG:POLICY:DATA_TYPE:GROUPS[+...]
```
mzhukova, line 13: should this be accompanied by extending inputs..?

dzarukin: For a generally applicable feature - yes. In this case I refrain from extending them as its usability is very limited. Test cases will come from OV later once they integrate the functionality and update their verbose logs.

mzhukova: should we at least add a toy case to ensure benchdnn/verbose/etc. work as expected?

include/oneapi/dnnl/dnnl.h
dzarukin: I'll keep it as is as it's really minor.

include/oneapi/dnnl/dnnl_types.h
dzarukin: Added a note about collision.

include/oneapi/dnnl/dnnl_types.h
dzarukin: Restored and added a dev note for the future.

tests/benchdnn/doc/knobs_attr.md
dzarukin: It just works :) I'll let Andrey to follow up on this and future test cases.

doc/programming_model/attributes_quantization.md
dzarukin: I see your point, but it looks like it's already inconsistent since zero-points are there which are not scaling as well.
I'd say we may want to touch-up that section later with better wording and pointers and let current examples slide in.

hidefromkgb (approved):

tests/benchdnn/doc/knobs_attr.md
densamoilov: Would it make sense to add this option to the self check?

hidefromkgb: I'll add some tests in my #3750, for which this PR serves as basis.
Since this is purely a performance-oriented feature, it might be more logical to add tests to a PR that brings performance.

densamoilov: I wasn't actually talking about the library feature but rather about the new benchdnn attribute option that we seem to test in: https://github.com/uxlfoundation/oneDNN/blob/e58f35836d6d14b1e32030e24fe8ae8f1f4de4f7/tests/benchdnn/self/common.cpp#L49
So it's about testing benchdnn as a tool and not the library.

dzarukin: Added them, thanks for reminding.

dzarukin:
make test

PR #3795, opened 2025-08-20
graph: backend: dnnl: passes: get in/out after op kind check

TaoLv:
Query input and output logical tensors only after op kind is matched.


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3806, opened 2025-08-22
common: minor refactoring to resolve compiler warning

mzhukova:
Fixes [MFDNN-14098](https://jira.devtools.intel.com/browse/MFDNN-14098).

The issue is with older GCC compiler complaining on out-of-bounds access, that shouldn't be true, as there are checks for allowed sizes in place.
Extending checks just in case and also changing to more straightforward copy.
// possible alternative is to add ~ignore pragma

### DIFF
--- a/src/common/host_scalar_memory_storage.hpp
+++ b/src/common/host_scalar_memory_storage.hpp
@@ -17,6 +17,8 @@
 #ifndef COMMON_HOST_SCALAR_MEMORY_STORAGE_HPP
 #define COMMON_HOST_SCALAR_MEMORY_STORAGE_HPP
 
+#include <algorithm>
+#include <cstdint>
 #include <memory>
 #include <type_traits>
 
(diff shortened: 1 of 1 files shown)


mzhukova:
make test
enable compiler_gnu8
disable benchdnn_all
enable benchdnn_matmul

src/common/host_scalar_memory_storage.hpp
```diff
@@ -45,20 +47,32 @@ class host_scalar_memory_storage_t : public memory_storage_t {
...
+                || value_size != types::data_type_size(data_type_)
+                || value_size == 0 || value_size > max_scalar_size)
             return status::invalid_arguments;
 
-        std::memcpy(value, &scalar_storage_, value_size);
+        const auto *src
+                = reinterpret_cast<const std::uint8_t *>(&scalar_storage_);
+        auto *dst = static_cast<std::uint8_t *>(value);
+
+        std::copy(src, src + value_size, dst);
```
dzarukin, line 59: Probably could use a comment what was the original version of the code and what's the reason for the current one for historical context.

dzarukin (approved):

src/common/host_scalar_memory_storage.hpp
```diff
@@ -45,20 +47,34 @@ class host_scalar_memory_storage_t : public memory_storage_t {
 
     status_t get_scalar_value(void *value, size_t value_size) const {
         if (!is_initialized_ || value == nullptr
-                || value_size != types::data_type_size(data_type_))
+                || value_size != types::data_type_size(data_type_)
+                || value_size == 0 || value_size > max_scalar_size)
             return status::invalid_arguments;
 
-        std::memcpy(value, &scalar_storage_, value_size);
+        // Use std::copy with explicit byte pointers to help compiler verify bounds
```
densamoilov: I think `std::copy` cannot verify the bounds in this case but using it will force the compiler to avoid using `__builtin_memcpy` that triggers the the warning. 

Ideally, I think we should ignore the warning because it's GCC's buggy behavior. But if we don't want to deal with disabling the warning for this part of the code we could keep this fix as is and leave a comment that this `std::copy` has to stay here so that we don't remove it accidently in the future.

PR #3820, opened 2025-08-26
benchdnn: memory tracking: address potential OOM cases

dzarukin:
[MFDNN-14017](https://jira.devtools.intel.com/browse/MFDNN-14017) - this gets addressed by decreasing the allowed portion of memory from 0.75 to 0.7. There are comments that explain the problem.

[MFDNN-13670](https://jira.devtools.intel.com/browse/MFDNN-13670)
[MFDNN-11194](https://jira.devtools.intel.com/browse/MFDNN-11194) - this gets addressed by adding extra variable holding driver memory need.

Besides that, the PR moves memory allocations and tracking functionality into a separate file and extends tracking to mapped memories for easier identifying which spot looks ill.

### DIFF
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -475,6 +475,10 @@ void dnn_mem_t::map() const {
                 DNN_SAFE_V(dnnl_memory_unmap_data_v2(mem, mapped_ptrs_[i], i));
             DNN_SAFE_V(st);
         }
+        // Register a mapped memory entry if a different pointer from `data_`
+        // was returned.
+        if (IMPLICATION(!data_.empty(), data_[i] != mapped_ptrs_[i]))
+            zmalloc_registry().add_mapped(mapped_ptrs_[i], size(i));
     }
 }
 
@@ -486,6 +490,11 @@ void dnn_mem_t::unmap() const {
     auto mem = m_padded_ ? m_padded_ : m_;
     const int nhandles = query_md_num_handles(md_);
     for (int i = 0; i < nhandles; i++) {
+        // Unregister a mapped memory entry if a different pointer from `data_`
+        // was returned when mapped.
+        if (IMPLICATION(!data_.empty(), data_[i] != mapped_ptrs_[i]))
+            zmalloc_registry().remove_mapped(mapped_ptrs_[i], size(i));
+
         DNN_SAFE_V(dnnl_memory_unmap_data_v2(mem, mapped_ptrs_[i], i));
         mapped_ptrs_[i] = nullptr;
     }
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -239,8 +239,8 @@ int dnn_mem_t::reorder(const dnn_mem_t &rhs, const_dnnl_primitive_attr_t attr,
     return status;
 }
 
-size_t dnn_mem_t::size() const {
-    return dnnl_memory_desc_get_size(md_);
+size_t dnn_mem_t::size(int index) const {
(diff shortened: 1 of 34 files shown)


dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_rnn
disable test_device_cpu

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -475,6 +475,10 @@ void dnn_mem_t::map() const {
                 DNN_SAFE_V(dnnl_memory_unmap_data_v2(mem, mapped_ptrs_[i], i));
             DNN_SAFE_V(st);
         }
+        // Register a mapped memory entry if a different pointer from `data_`
+        // was returned.
+        if (IMPLICATION(!data_.empty(), data_[i] != mapped_ptrs_[i]))
+            zmalloc_registry().add_mapped(mapped_ptrs_[i], size(i));
```
rjoursler: nit: the name `zmalloc_registry` no longer seems applicable as we are registering memory not managed by `zmalloc`.

tests/benchdnn/utils/memory.cpp
```diff
@@ -44,24 +44,77 @@ void memory_registry_t::add(void *ptr, size_t size) {
...
+            ptr, smart_bytes(size).c_str(), smart_bytes(total_size_).c_str());
+    // Do not warn on overflow as it can be a temporary jump due to reorder or
+    // other internal memory manipulation.
+}
+
+void memory_registry_t::remove_mapped(void *ptr) {
+    std::lock_guard<std::mutex> g(m_);
+
+    // See `add_mapped` comment.
+    // Since double mapping may happen, double mapping may happen as well.
```
rjoursler: Missing prefix, it should be: "... double unmaping may happen as well."

tests/benchdnn/rnn/ref_rnn_bwd.cpp
```diff
@@ -240,7 +233,7 @@ void rnn_linear_bwd(const prb_t &prb, const args_t &args,
     AOC<float> diff_bias(diff_bias_, prb.n_layer, prb.n_dir(),
             prb.n_gates() + is_lbr, prb.dhc);
 
-    std::vector<float> ws_bwd_buffer;
+    float *ws_bwd_buffer = (float *)zmalloc(prb.get_ws_size(FLAG_BWD), 64);
```
rjoursler, line 236: Use `std::unique_ptr` to avoid the manual memory management.

rjoursler (approved):
Would it be reasonable to enable device memory tracking as well when `dnn_mem_t` objects are constructed?

tests/benchdnn/rnn/ref_rnn_bwd.cpp
dzarukin: Discussed offline: it will require extra manipulations such as custom deleter through zfree, which unlikely make the code simpler and more readable.

tests/benchdnn/utils/memory.cpp
dzarukin: Thanks for the catch, updated.

tests/benchdnn/dnnl_memory.cpp
dzarukin: True that. Update the name and made the object a singleton.

dzarukin:
> Would it be reasonable to enable device memory tracking as well when `dnn_mem_t` objects are constructed?

I've added the logic to track that as well. Thanks for a suggestion.

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

mgouicem (approved):

vpirogov (approved):

PR #3824, opened 2025-08-26
graph: backend: dnnl: binary chain + matmul + binary chain pattern

TaoLv:
Fix MFDNN-14020

- Extended the number of post-ops to 20.
- Added a new special pattern with binary ops chain + matmul + binary ops chain. Currently implemented through binary primitive and matmul primitive.
- Added corresponding test cases.

### DIFF
--- a/tests/benchdnn/inputs/graph/complex_fusion/bmb/binary-matmul-binary-f32.json
+++ b/tests/benchdnn/inputs/graph/complex_fusion/bmb/binary-matmul-binary-f32.json
@@ -1,0 +1,1488 @@
+{
+  "version": "3.10.0",
+  "engine_kind": "cpu",
+  "fpmath_mode": "strict",
+  "fpmath_mode_apply_to_int": "false",
+  "input_ports": [
+    0,
+    1,
+    2,
+    3,
+    4,
+    5,
+    6,
+    7,
+    8,
+    9,
+    10,
+    11,
+    12,
+    13,
+    14,
+    15,
+    16,
+    17,
+    18,
+    19,
+    20,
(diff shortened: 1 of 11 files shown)


tests/benchdnn/inputs/graph/complex_fusion/bmb/binary-matmul-binary-f32.json
```diff
@@ -0,0 +1,1488 @@
+{
+  "version": "3.10.0",
+  "engine_kind": "cpu",
```
xiang1guo, line 3: From the case, it seems that we use inter f32 to represent the accumulated dt. Since we don't pass any accumulation mode to the backend, does that mean both cases—with and without type casting(pure bf16)—will end up running the same kernel?

TaoLv: The pattern itself will not match the pure bf16 case. There is a decision function to check the dst data type of the first binary.

xiang1guo: oh, got it, I missed that decision function.

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3833, opened 2025-08-27
xe: conv: jit: fix zp_plan count missing zp_src

rjoursler:
Fixes performance regression identified in [MFDNN-14105](https://jira.devtools.intel.com/browse/MFDNN-14105).

The root cause is that register usage misestimation triggers selection of a strategy with no register space which can be used for CSE, resulting in a poorly performing generated kernel.


atkassen (approved):

rjoursler:
make test perf-gpu
set primitive=gpu:ir

rjoursler:
make test
disable test_device_cpu

kealan-barbieri (approved):

PR #3834, opened 2025-08-27
gpu: fix ocl gemm impl for src and wei zps

mzhukova:
I've encountered an issue with this implementation not producing correct results with src and wei zero points (dst works fine), the reason was that defines were not propagated and also src/wei order was swapped. 

I think there is no need in src/gpu/intel/gemm/ocl_attrs.h anymore as it is not used anywhere else.


kealan-barbieri (approved):

mzhukova:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul

mzhukova:
ping to @uxlfoundation/onednn-gpu-intel for 2nd review :)

echeresh (approved):

PR #3843, opened 2025-08-29
cpu: refactor scales for lnorm, softmax, binary and reorder

dzarukin:
Continuation of preliminary work needed to support Asynchronous runtime from Google.

### DIFF
--- a/src/cpu/ref_layer_normalization.cpp
+++ b/src/cpu/ref_layer_normalization.cpp
@@ -47,8 +47,15 @@ status_t ref_layer_normalization_fwd_t::execute_forward(
             : CTX_OUT_MEM(float *, DNNL_ARG_VARIANCE);
     auto dst = CTX_OUT_MEM(void *, DNNL_ARG_DST);
 
-    DEFINE_ARG_SCALES_BUFFER(src_scales, DNNL_ARG_SRC);
-    DEFINE_ARG_SCALES_BUFFER(dst_scales, DNNL_ARG_DST);
+    const float *src_scales
+            = CTX_IN_MEM(const float *, DNNL_ARG_ATTR_SCALES | DNNL_ARG_SRC);
+    const float *dst_scales
+            = CTX_IN_MEM(const float *, DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST);
+
+    const bool with_src_scales
+            = !pd()->attr()->scales_.has_default_values(DNNL_ARG_SRC);
+    const bool with_dst_scales
+            = !pd()->attr()->scales_.has_default_values(DNNL_ARG_DST);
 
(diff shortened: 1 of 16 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_softmax
enable benchdnn_binary
enable benchdnn_lnorm
enable benchdnn_reorder
disable test_device_gpu

src/cpu/ref_softmax.cpp
```diff
@@ -170,7 +177,7 @@ status_t ref_softmax_fwd_t::execute_forward_dense(const exec_ctx_t &ctx) const {
             } else if (pd()->is_logsoftmax()) {
                 val = d - space_denom;
             }
-            val *= src_scales[0];
+            if (with_src_scales) val *= src_scales[0];
```
mgouicem, line 180: Is the scale datatype checked anywhere? I would recommend to use load_float_value here.

mgouicem (commented):
Most of the changes here assume scales are in f32.
It is true when using DEFINE_ARG_SCALES_BUFFER macro, since it hides conversion to float when scale is common to tensor. But it is no more true when using the pointer to scales directly.

I would recommend:
- for jitted impl, to add check for scales datatype
- for ref, use load_float_value from io_helper.

dzarukin:
> Most of the changes here assume scales are in f32. It is true when using DEFINE_ARG_SCALES_BUFFER macro, since it hides conversion to float when scale is common to tensor. But it is no more true when using the pointer to scales directly.
> 
> I would recommend:
> 
> * for jitted impl, to add check for scales datatype
> * for ref, use load_float_value from io_helper.

It is still true, because data type support is tucked behind `skip_mask_t::scales_data_type`. Without it explicitly enabled only f32 scales are supported. The next one who is going to extend that support if ever needed, gonna apply those changes you mentioned above.

mgouicem (approved):

src/cpu/ref_layer_normalization.cpp
```diff
@@ -117,7 +124,7 @@ status_t ref_layer_normalization_fwd_t::execute_forward(
             args.dst_md = pd()->dst_md();
             ref_post_ops->execute(d, args);
 
-            d *= dst_scales[0];
+            if (with_dst_scales) d /= dst_scales[0];
```
densamoilov, line 127: (nit) I understand that performance of reference implementations is not important but previously we had 1 division and now we have N*C divisions. Maybe it's not a big deal but it could slow down validation for non-x64 configuration.

dzarukin: It might be true, though routines that collect single memory value by logical offset take way more time so I was not considering that division seriously. Besides, I'd like to avoid of adding scratchpad support to handle the division.

PR #3842, opened 2025-08-29
sdpa f16 accumulation

syurkevi:
# Description

This PR add f16 accumulation modes to the KQ and VS matmuls. The intermediate softmax, scaling and masking operations continue to be with f32.  The individual attributes of each matmul allow for all permutations of f16,f32 for the two matmuls. Currently, only systems using FMA variants of sdpa are enabled. Systolic f16 accumulate can be enabled if/once corresponding f16 accumulate matmul operations are enabled that are used for generating test data.

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

PR #3844, opened 2025-08-29
graph: support accumulation mode for matmul and sdpa

TaoLv:
- Add a new attribute `accumulation_mode` to matmul operation. It has the same definition for the supported values as the primitive attribute.
- Pass the attribute value to matmul when creating the primitive.
- <s>Currently GPU ukernel does not support the attribute, hence check it and return unimplemented.</s>
- Add test case in benchdnn. The op attribute rewriting logic is slightly enhanced to accept attribute which is not present in the json file.

<s>Mark this as a draft as:</s> Now it's ready for review:

- [x] I'm still working on a RFC doc for the new attribute. (update: #3857)
- [x] let #3842 to be merged firstly.

### DIFF
--- a/doc/graph/operations/MatMul.md
+++ b/doc/graph/operations/MatMul.md
@@ -29,6 +29,7 @@ the left side of dimensions (inserting `1`) to make sure two ranks matched.
 |:-------------------------------------------------------|:--------------------------------------------------------------------|:-----------|:----------------------|:---------------------|
 | [transpose_a](@ref dnnl::graph::op::attr::transpose_a) | Controls whether to transpose the last two dimensions of `src`.     |bool        | True, False (default) | Optional             |
 | [transpose_b](@ref dnnl::graph::op::attr::transpose_b) | Controls whether to transpose the last two dimensions of `weights`. |bool        | True, False (default) | Optional             |
+| [accumulation_mode](@ref dnnl::graph::op::attr::accumulation_mode) | Specifies the accumulation mode to be used for MatMul calculation | string | `strict` (default), `relaxed`, `any`, `f32`, `s32`, `f16`. | Optional |
 
 The above transpose attribute will not in effect when rank of an input tensor is
 less than 2. For example, in library implementation 1D tensor is unsqueezed
@@ -37,6 +38,10 @@ firstly before compilation. The rule is applied independently.
 - For \src tensor, the rule is defined like: `[d] -> [1, d]`.
 
 - For \weights tensor, the rule is defined like: `[d] -> [d, 1]`.
+
+For the attribute `accumulation_mode`, refer to
+[accumulation mode](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_accumulation_mode.html)
+for the definition for each supported value.
 
 ## Execution arguments
 


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/flex_rewrite.cpp
```diff
@@ -1285,14 +1285,15 @@ int flex_rewrite_t::op_attrs_rewrite(deserialized_graph_t &dgraph) {
...
             const auto &new_val = new_attr.second;
-            const auto &attr_type = temp_op.attrs_[attr_name].type_;
+            std::string attr_type = "undef";
+            if (temp_op.attrs_.count(attr_name)) {
+                attr_type = temp_op.attrs_[attr_name].type_;
+            } else {
+                attr_type = attrstr2type(attr_name);
+                temp_op.attrs_[attr_name].type_ = attr_type;
+            }
+
```
ElaineBao, line 1301: Could you please elaborate what does this code change try to resolve?

TaoLv: As explained by the commit message of this change, it enables benchdnn to support `--op-attrs=` rewriting even the target attribute is not present in the JSON file. Before this change, benchdnn graph will raise an error if the attribute is not defined on the op in the JSON file.
Take this PR as example, `accumulation_mode` is new and not part of the JSON files. Now with this change, we can test it through `--op-attrs=ID:accumulation_mode:f16`, without editing the JSON file to add the attribute.

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/flex_rewrite.cpp
dzarukin: You may want to put a comment explaining the behavior of injecting newly added attributes which don't present in previously created JSON files.

tests/benchdnn/graph/utils.cpp
```diff
@@ -430,6 +431,75 @@ dnnl::graph::op::attr attrstr2kind(const std::string &attr_name) {
     return dnnl::graph::op::attr::undef;
 }
 
+std::string attrstr2type(const std::string &attr_name) {
```
dzarukin: Maybe `const std::string &` since the output string is not modified?

tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
```diff
@@ -190,3 +190,6 @@
 --reset --dt=1:bf16+2:bf16+3:bf16+6:bf16+104:bf16 --in-shapes=1:2x10x2304x64*acbd+2:2x10x77x64*acbd+3:2x10x77x64*acbd --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
 --reset --dt=1:f16+2:f16+3:f16+6:f16+104:f16 --in-shapes=1:2x10x2304x64*acbd+2:2x10x77x64*acbd+3:2x10x77x64*acbd --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
 --reset --in-shapes=1:2x10x2304x64*acbd+2:2x10x77x64*acbd+3:2x10x77x64*acbd --case=complex_fusion/mha/sdpa-plain-implicit-causal-mask-fp32-bs1.json
+
+# accumulation mode
+--reset --op-attrs=0:accumulation_mode::strict,0:accumulation_mode:f16 --case=complex_fusion/mha/sdpa-plain-simplified-f16-f32.json
```
dzarukin: Would the fist case here just repeat what `--case=complex_fusion/mha/sdpa-plain-simplified-f16-f32.json` does?
If yes, is it desired to have a double run?

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/flex_rewrite.cpp
TaoLv: Thanks for the suggestion. Added a comment to the code.

tests/benchdnn/graph/utils.cpp
TaoLv: Fixed.

tests/benchdnn/inputs/graph/complex_fusion/harness_mha_all
TaoLv: I just removed the first case (the strict one).
I also tried to test rewriting both QK and VS's accumulation mode as f16. But it caused result mismatching on Xe-LP. Hence I did not add the case here.

PR #3849, opened 2025-08-30
tests: benchdnn: add capability to import buffers from files

hidefromkgb:
The direct predecessor to this patch was instrumental in resolving [MFDNN-14086](https://jira.devtools.intel.com/browse/MFDNN-14086).

Buffer import capability is a great feature to have when debugging real-world usecases in which the synthetic data generated by `benchdnn` is not always enough to reproduce the error.

### DIFF
--- a/tests/benchdnn/binary/binary.cpp
+++ b/tests/benchdnn/binary/binary.cpp
@@ -36,9 +36,10 @@
 namespace binary {
 
 int fill_mem(
-        const prb_t *prb, int input_idx, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp) {
+        int exec_arg, const prb_t *prb, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp) {
     const auto nelems = mem_fp.nelems();
     if (nelems == 0) return OK;
+    if (fill_from_file(exec_arg, mem_dt, mem_fp)) return OK;
 
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
@@ -58,7 +59,7 @@ int fill_mem(
 
     int min_val = static_cast<int>(MAX2(-8.f, lowest_dt(mem_dt.dt())));
     // Tenrary op supports a third input which can't be negative so far.
-    if (input_idx == 2) min_val = 0;
+    if (exec_arg == DNNL_ARG_SRC_2) min_val = 0;
 
     /* Do fixed partitioning to have same filling for any number of threads */
     static constexpr int64_t chunk_size = 64;
@@ -70,7 +71,7 @@ int fill_mem(
         // repeating patterns. We could use discard(idx_start) too but
         // it has a complexity in O(idx_start). We also add 1 to avoid
         // seeding with 0.
-        std::minstd_rand int_seed(idx_start + nelems * input_idx + 1);
+        std::minstd_rand int_seed(idx_start + nelems * exec_arg + 1);
         int_seed.discard(1);
 
         std::uniform_int_distribution<> gen(min_val, 8);
@@ -79,9 +80,9 @@ int fill_mem(
             float val = gen(int_seed);
             // Make floating-point values only for src0 as src1 filling can be
             // used in other drivers and preferred to be integer.
-            if (input_idx == 0) val *= 0.5f;
+            if (exec_arg == DNNL_ARG_SRC_0) val *= 0.5f;
             // Remove zeroes in src1 to avoid division by zero.
-            if (input_idx == 1 && val == 0.0f) val = 1.0f;
+            if (exec_arg == DNNL_ARG_SRC_1 && val == 0.0f) val = 1.0f;
             val = round_to_nearest_representable(mem_dt.dt(), val);
             mem_fp.set_f32_elem(idx, val);
         }
@@ -248,7 +249,7 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
 
         switch (exec_arg) {
             case DNNL_ARG_SRC_0:
-                SAFE(fill_mem(prb, 0, mem, ref_mem), WARN);
+                SAFE(fill_mem(exec_arg, prb, mem, ref_mem), WARN);
                 // Need a copy of source data for inplace mode for bitwise
                 // testing.
                 if (has_bench_mode_bit(mode_bit_t::bitwise) && prb->inplace) {
@@ -258,14 +259,14 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
                 }
                 break;
             case DNNL_ARG_SRC_1:
-                SAFE(fill_mem(prb, 1, mem, ref_mem), WARN);
+                SAFE(fill_mem(exec_arg, prb, mem, ref_mem), WARN);
                 break;
             case DNNL_ARG_SRC_2:
-                SAFE(fill_mem(prb, 2, mem, ref_mem), WARN);
+                SAFE(fill_mem(exec_arg, prb, mem, ref_mem), WARN);
                 break;
             case DNNL_ARG_DST:
                 if (prb->attr.post_ops.find(alg_t::SUM) >= 0) {
-                    SAFE(fill_mem(prb, 3, mem, ref_mem), WARN);
+                    SAFE(fill_mem(exec_arg, prb, mem, ref_mem), WARN);
 
                     // Bitwise mode for sum requires a copy due to data for
                     // post-op will be overwritten and it must be refreshed.
--- a/tests/benchdnn/bnorm/bnorm.cpp
+++ b/tests/benchdnn/bnorm/bnorm.cpp
@@ -38,6 +38,7 @@ namespace bnorm {
 
 int fill_mean(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
         dnn_mem_t &mem_dt) {
+    if (fill_from_file(DNNL_ARG_MEAN, mem_dt, mem_fp)) return OK;
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
         // Mean must be computed unless it is passed by user directly.
@@ -64,6 +65,7 @@ int fill_mean(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
 
 int fill_src(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
         dnn_mem_t &mem_dt, const dnn_mem_t &ref_mean, res_t *res) {
+    if (fill_from_file(DNNL_ARG_SRC, mem_dt, mem_fp)) return OK;
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
         return fill_random_real(mem_dt, mem_fp, res);
@@ -119,6 +121,7 @@ int fill_src(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
 int fill_variance(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
         dnn_mem_t &mem_dt, const dnn_mem_t &ref_src,
         const dnn_mem_t &ref_mean) {
+    if (fill_from_file(DNNL_ARG_VARIANCE, mem_dt, mem_fp)) return OK;
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
         // Variance must be computed unless it is passed by user directly.
@@ -166,6 +169,7 @@ int fill_src_add(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
         const dnn_mem_t &ref_mean) {
     const bool fill_src_add = prb->fuse_add_relu();
     if (!fill_src_add) return OK;
+    if (fill_from_file(DNNL_ARG_SRC_1, mem_dt, mem_fp)) return OK;
 
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
@@ -217,6 +221,7 @@ int fill_src_add(const prb_t *prb, const cfg_t &cfg, dnn_mem_t &mem_fp,
 int fill_scale(const prb_t *prb, dnn_mem_t &mem_fp, dnn_mem_t &mem_dt) {
     const bool use_sc = prb->use_sc();
     if (!use_sc) return OK;
+    if (fill_from_file(DNNL_ARG_SCALE, mem_dt, mem_fp)) return OK;
 
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
@@ -241,6 +246,7 @@ int fill_scale(const prb_t *prb, dnn_mem_t &mem_fp, dnn_mem_t &mem_dt) {
 int fill_shift(const prb_t *prb, dnn_mem_t &mem_fp, dnn_mem_t &mem_dt) {
     const bool use_sh = prb->use_sh();
     if (!use_sh) return OK;
+    if (fill_from_file(DNNL_ARG_SHIFT, mem_dt, mem_fp)) return OK;
 
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
@@ -307,6 +313,7 @@ int prepare_bwd(
 
     const auto nelems = mem_fp.nelems();
     if (nelems == 0) return OK;
+    if (fill_from_file(DNNL_ARG_DIFF_DST, mem_dt, mem_fp)) return OK;
 
     // Refer to modes documentation for filling principles.
     if (has_bench_mode_bit(mode_bit_t::bitwise)) {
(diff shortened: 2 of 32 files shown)


tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,39 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
     dnn_mem_t dummy;
     return fill_random_real(dummy, mem_ref, nullptr, fill_cfg, dnnl_memory);
 }
+
+int fill_from_file(int exec_arg, dnn_mem_t &mem, dnn_mem_t &mem_ref) {
+    static const char format[] = "File '%s' %s; buffer not imported.\n";
+    auto prefix = benchdnn_getenv_string("BENCHDNN_BUFFER_PREFIX", false);
+    if (prefix.empty()) return FAIL;
+
+    prefix += "." + std::to_string(exec_arg);
```
echeresh: Can we add `.bin` suffix here by default?

I have a branch dumping memory arguments:
https://github.com/uxlfoundation/oneDNN/tree/echeresh/memdump

This would help to make this feature self-contained in oneDNN/benchdnn: oneDNN dumps buffers and then benchdnn loads the buffers.

When adding memory-dumping as an official feature, it would be nice to align with `ONEDNN_JIT_DUMP` including `.bin` suffix: https://uxlfoundation.github.io/oneDNN/dev_guide_inspecting_jit.html

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,39 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+    auto file = fopen(prefix.c_str(), "rb");
+    if (!file) {
```
echeresh, line 383: I suggest to add a quick check on entry to  exit if argument is empty (size is zero).

For scratchpad and workspace we may not have files/data but `fill_from_file()` is still called and will print "cannot be opened" error.

tests/benchdnn/bnorm/bnorm.cpp
```diff
@@ -282,25 +286,33 @@ int prepare_fwd(const prb_t *prb, dnn_mem_map_t &mem_map,
...
 
     return OK;
 }
 
-int prepare_bwd(
-        const prb_t *prb, dnn_mem_map_t &mem_map, dnn_mem_map_t &ref_mem_map) {
+int prepare_bwd(const prb_t *prb, dnn_mem_map_t &mem_map,
+        dnn_mem_map_t &ref_mem_map,
+        const std::unordered_map<int, int> &file_fill_map) {
+    if (file_fill_map.at(DNNL_ARG_DIFF_DST) == OK) return OK;
```
echeresh: Backward bnorm cases (and maybe RNN?) require forward to be called first. How will it work with this feature?

- Does the user have to provide both sets of buffers?
- If yes, is argument overlapping possible between forward/backward?

If this is not trivial to support we can exclude batchnorm, RNN (and something else maybe?) from the initial support.

vpirogov:
Could you please elaborate on the usage model of this feature? As benchdnn correctness checks rely on certain properties of generated data for stability. How do we plan to test correctness with arbitrary buffers?

tests/benchdnn/utils/fill.cpp
hidefromkgb: You mean e.g. `/path/to/prefix.bin.41`?
Will do. I'm okay with that — and it'll also help adding plain-text import in the future (e.g. using a .txt.IDX extension).

tests/benchdnn/utils/fill.cpp
hidefromkgb: These errors are not fatal, the execution will continue even if some files are absent.
This is by design, just to let the developer know all of the buffers that participate.

tests/benchdnn/bnorm/bnorm.cpp
hidefromkgb: Yup, and pooling as well.
This is rather easy to support by appending an additional suffix for direction, e.g. `/path/to/prefix.fwd_i.bin.41`

hidefromkgb:
> Could you please elaborate on the usage model of this feature? As benchdnn correctness checks rely on certain properties of generated data for stability. How do we plan to test correctness with arbitrary buffers?

@vpirogov This is not for general CI, but rather to give developers a means to debug correctness issues filed by customers, and these are always different case by case, so some correctness issues (due e.g. to a different summation order in CPU vs. GPU, DPAS vs. VNNI, etc.) are inevitable — but there's a difference between 0.41 on CPU vs. 0.42 on GPU and e.g. 0.41 vs. –NaN.

tests/benchdnn/utils/fill.cpp
echeresh: I meant at the end: `/path/to/prefix.41.bin`

For JIT dump we put `.bin` prefix at the end - this is an extension to mark a binary file.

tests/benchdnn/utils/fill.cpp
echeresh: @hidefromkgb "cannot be opened" doesn't sound like a warning or information message. As a user I would likely consider this as "something is missing, I have to do something with it". If you want to keep it - I would suggest to rephrase it to communicate that it's OK for the buffer to be missing.

hidefromkgb: @echeresh I've got no issue with "is missing" instead of "cannot be opened". Will change.

vpirogov:
> This is not for general CI, but rather to give developers a means to debug correctness issues filed by customers, and these are always different case by case, so some correctness issues (due e.g. to a different summation order in CPU vs. GPU, DPAS vs. VNNI, etc.) are inevitable — but there's a difference between 0.41 on CPU vs. 0.42 on GPU and e.g. 0.41 vs. –NaN.

I get this part. What I'm saying is that benchdnn correctness check is not equipped to handle arbitrary data and you will likely be getting false positive "test fail" status even for correct results.

tests/benchdnn/bnorm/bnorm.cpp
hidefromkgb: @echeresh BTW aren't the buffer IDs different between different directions? I looked closer and I don't see clashes between buffer IDs across different directions.

hidefromkgb:
> I get this part. What I'm saying is that benchdnn correctness check is not equipped to handle arbitrary data and you will likely be getting false positive "test fail" status even for correct results.

Well, this is what `-v99` and developers' common sense are for. I don't think we need to fundamentally change how benchdnn works just to accommodate for this feature. All of us know that things like machine epsilon exist.

tests/benchdnn/utils/fill.cpp
hidefromkgb: OK. LGTM

tests/benchdnn/bnorm/bnorm.cpp
echeresh: @hidefromkgb That makes sense. So you expect the user to provide buffers from forward - can you please update documentation as well to specify that some cases (such as batchnorm) require forward buffers.

Also I think this should be a requirement, if forward buffers are missing, we can't run correctness checks as the buffers won't be coherent. So if missing forward buffers for such cases - should report an error.

echeresh: > @echeresh BTW aren't the buffer IDs different between different directions? I looked closer and I don't see clashes between buffer IDs across different directions.

What about scratchpads? Also it would be nice to consider future `ONEDNN_MEMORY_DUMP` support - we have to have unique prefixes between cases for that feature so it would be natural to take that into account for this benchdnn feature.

hidefromkgb: @echeresh Is there a usecase for scratchpad buffer initialization?
Don't get me wrong, directionality is trivial to support, but is it really necessary?

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

vpirogov:
> Well, this is what `-v99` and developers' common sense are for. I don't think we need to fundamentally change how benchdnn works just to accommodate for this feature. All of us know that things like machine epsilon exist.

Thanks, this explain the usage model. As you are introducing a feature available to all benchdnn users we need to add some warnings for users who are not as well versed in machine arithmetic as oneDNN developers.

hidefromkgb:
> Thanks, this explain the usage model. As you are introducing a feature available to all benchdnn users we need to add some warnings for users who are not as well versed in machine arithmetic as oneDNN developers.

@vpirogov Should we maybe hide this under `DNNL_DEV_MODE`? This would alleviate all such concerns.
Anyone who builds oneDNN with this compilation flag basically consents to knowing exactly what they're doing.

vpirogov:
> Should we maybe hide this under `DNNL_DEV_MODE`? This would alleviate all such concerns. Anyone who builds oneDNN with this compilation flag basically consents to know exactly what they're doing.

We have a few more options here:
* Add a warning indicating that false positives are likely
* Stop reporting pas/fail status and show statistics on data difference from reference
I'll defer decision on which approach to take to @dzarukin.

tests/benchdnn/bnorm/bnorm.cpp
echeresh: Yeah, are right about scratchpads. What about mean/variance between forward/backward batch normalization?

https://uxlfoundation.github.io/oneDNN/dev_guide_batch_normalization.html

(I guess here practically we can assume they are the same through between forward/backward)

Anyway, I suggest to take future support for memory dumping in oneDNN into account, say we can have the following files when dumping batch normalization:

```bash
# backward
dnnl_dump_mem.1.gpu.batch_normalization.17.bin
dnnl_dump_mem.1.gpu.batch_normalization.1.bin
dnnl_dump_mem.1.gpu.batch_normalization.49.bin
dnnl_dump_mem.1.gpu.batch_normalization.50.bin
dnnl_dump_mem.1.gpu.batch_normalization.64.bin

# forward
dnnl_dump_mem.3.gpu.batch_normalization.129.bin
dnnl_dump_mem.3.gpu.batch_normalization.145.bin
dnnl_dump_mem.3.gpu.batch_normalization.1.bin
dnnl_dump_mem.3.gpu.batch_normalization.49.bin
dnnl_dump_mem.3.gpu.batch_normalization.50.bin
dnnl_dump_mem.3.gpu.batch_normalization.64.bin
```

Then it could be as simple as specifying two separate prefixes for forward/backward.

hidefromkgb: @echeresh After a careful consideration I realized that we simply don't need to run both forward and backwards primitives at once if we have the necessary buffers ready — or fail if not given something that otherwise necessitates a forward run.

echeresh: @hidefromkgb Thanks for checking, if so - that's better. Since this is not the default scenario - can you please test it following the expected usage flow to ensure everything works as expected? If something pops up on the other hand, it's fine to have limitations, if needed.

tests/benchdnn/binary/binary.cpp
```diff
@@ -35,10 +35,11 @@
 
 namespace binary {
 
-int fill_mem(
-        const prb_t *prb, int input_idx, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp) {
+int fill_mem(int input_idx, const prb_t *prb, int exec_arg, dnn_mem_t &mem_dt,
```
dzarukin: Please replace `input_idx` completely with `exec_arg` here.

tests/benchdnn/concat/concat.cpp
```diff
@@ -68,6 +68,7 @@ int fill_src(int input_idx, dnnl_data_type_t dt, dnn_mem_t &mem_dt,
         dnn_mem_t &mem_fp) {
     const auto nelems = mem_fp.nelems();
     if (nelems == 0) return OK;
+    if (fill_from_file(input_idx, mem_dt, mem_fp)) return OK;
```
dzarukin: I'd rename `input_idx` with `exec_arg` here as well.

dzarukin:
make test

tests/benchdnn/sum/sum.cpp
```diff
@@ -68,6 +68,7 @@ dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {
 int fill_src(int input_idx, dnn_mem_t &mem_dt, dnn_mem_t &mem_fp) {
     const auto nelems = mem_dt.nelems();
     if (nelems == 0) return OK;
+    if (fill_from_file(input_idx, mem_dt, mem_fp)) return OK;
```
dzarukin: Ditto.

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,43 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+        BENCHDNN_PRINT(2, format, prefix.c_str(), "is missing");
+        return false;
+    }
+    fseek(file, 0, SEEK_END);
+    size_t total = 0, size = ftell(file);
+    if (mem.size() != size) {
+        BENCHDNN_PRINT(2, format, prefix.c_str(),
+                "differs in size from the buffer's memory descriptor");
+        return false;
```
dzarukin: I suggest to inform the user at 0 level about the ill-formed file and abort the execution.
```suggestion
        BENCHDNN_PRINT(0, format, prefix.c_str(),
                "differs in size from the buffer's memory descriptor");
        fclose(file);
        SAFE_V(FAIL);
```

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,43 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+
+bool fill_from_file(int exec_arg, dnn_mem_t &mem, dnn_mem_t &ref_mem) {
+    static const char format[] = "File '%s' %s; buffer not imported.\n";
+    auto prefix = buffer_prefix;
+    if (prefix.empty()) return false;
+
+    prefix += "." + std::to_string(exec_arg) + ".bin";
+    auto file = fopen(prefix.c_str(), "rb");
+    if (!file) {
+        BENCHDNN_PRINT(2, format, prefix.c_str(), "is missing");
```
dzarukin:
```suggestion
        BENCHDNN_PRINT(2, format, prefix.c_str(), "is not found");
```

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,43 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+        BENCHDNN_PRINT(2, format, prefix.c_str(),
+                "differs in size from the buffer's memory descriptor");
+        return false;
+    }
+    fseek(file, 0, SEEK_SET);
+    for (size_t read = ~0; read && (total < size); total += read)
+        read = fread(
+                static_cast<uint8_t *>(mem) + total, 1, size - total, file);
+    if (total != size) {
+        BENCHDNN_PRINT(2, format, prefix.c_str(), "cannot be read correctly");
```
dzarukin: Ditto as for ill-formed file. Read incorrectly - must provide guidance what could go wrong and abort.

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,43 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+        return false;
+    }
+    fclose(file);
+    if (has_bench_mode_bit(mode_bit_t::corr)) {
```
dzarukin: Is it possible to change this to simply `if (ref_mem) {`?

tests/benchdnn/utils/fill.cpp
```diff
@@ -364,3 +364,43 @@ int fill_random_real(dnn_mem_t &mem_ref, const fill_cfg_t &fill_cfg,
...
+        if (ref_mem.reorder(mem) != OK) {
+            BENCHDNN_PRINT(2, format, prefix.c_str(), "cannot be reordered");
```
dzarukin: I'd abort this as well as two previous cases.

tests/benchdnn/binary/binary.cpp
hidefromkgb: Done.

tests/benchdnn/concat/concat.cpp
hidefromkgb: Done.

tests/benchdnn/sum/sum.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

tests/benchdnn/utils/fill.cpp
hidefromkgb: Done.

hidefromkgb:
make test

PR #3850, opened 2025-08-30
cpu: risc-v: build: add Zvfh extension support detection

xiazhuozhao:
# Description

The Zvfh extension, an optional extension of the RISC-V V (Vector) extension which was ratified in June 2023, enables FP16 SIMD operations. Utilizing FP16 can boost neural network performance, as it potentially doubles the number of elements processed per vector instruction while also reducing memory bandwidth.

### How it Works

The implementation checks for Zvfh support after confirming the presence of the base Vector (V) extension. It uses `check_cxx_source_compiles` with the `-march=rv64gcv_zvfh` flag. If the toolchain supports Zvfh, the build system will:

- Set the appropriate `-march` flag to enable Zvfh for compilation.
- Define the `DNNL_RISCV_USE_ZVFH_INTRINSICS` preprocessor macro, allowing C++ code to conditionally compile Zvfh-specific intrinsic paths.

If Zvfh is not available, the build system falls back to the existing RVV-only configuration, ensuring backward compatibility and portability across different RISC-V platforms.

# Checklist

## General

- [√] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- Have you formatted the code using clang-format?
No, clang-format does not support formatting for .cmake files.

## Performance improvements

- Have you submitted performance data that demonstrates performance improvements?
No. To achieve a significant performance speedup, future work will be required to register the FP16 data type and implement dedicated RVV optimizations for it.

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -145,10 +145,35 @@ if (DNNL_TARGET_ARCH STREQUAL "RV64")
     set(CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS_SAVE})
     # Set CAN_COMPILE_RVV_INTRINSICS to TRUE / FALSE instead of 1 / "" (Undefined)
     if (CAN_COMPILE_RVV_INTRINSICS)
+        # RVV is supported, now check for Zvfh (which depends on V)
+        set(ARCH_SIMD_TEST_FLAGS "-march=rv64gcv_zvfh")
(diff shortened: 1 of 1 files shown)


dzarukin (approved):

xiazhuozhao:
Thanks a lot for the approval, @dzarukin! As this is my first PR to be approved in the project, it's a great encouragement for me.

PR #3856, opened 2025-09-02
rfcs: graph api: support verbose pipeline for SDPA debugging

ElaineBao:
# Description

MFDNN-13623.

rendered_version: [link](https://github.com/uxlfoundation/oneDNN/blob/yixin/rfc/verbose_pipeline/rfcs/20250902-graph-verbose-pipeline/README.md)

### DIFF
--- a/rfcs/20250902-graph-verbose-pipeline/README.md
+++ b/rfcs/20250902-graph-verbose-pipeline/README.md
@@ -129,7 +129,8 @@ default, so users do not need to manually configure it for graph debugging. By
 default, no graph JSON files are generated unless the
 `ONEDNN_GRAPH_DUMP` environment variable is set, so performance remains
 unaffected. When `ONEDNN_GRAPH_DUMP=subgraph` is specified, graph JSON files are
-dumped, and corresponding verbose lines are printed during partition compilation:
+dumped during partition compilation (calling the partition::compile API), and
+corresponding verbose lines are printed:
 
 ```shell
 onednn_verbose,v1,graph,info,serialize graph to a json file graph-100002-9333857264675079303.json
--- a/rfcs/20250902-graph-verbose-pipeline/README.md
+++ b/rfcs/20250902-graph-verbose-pipeline/README.md
@@ -1,0 +1,129 @@
+# RFC: Graph Verbose Pipeline for SDPA Debugging and Reproducibility
+
+## Introduction
+
+This RFC proposes enhancements to the oneDNN Graph verbose logging mechanism to
+better support PyTorch framework in reporting and debugging Scaled Dot-Product
+Attention (SDPA) issues. The primary goal is to restore the usability and
+reproducibility contract previously provided by primitive verbose logs, now for
+the Graph API.
+
+Currently, PyTorch leverages both the Graph API and Primitive API, but
+reproducing issues requires different workflows:
+
+- **Primitive API**: Users set the `ONEDNN_VERBOSE` environment variable to
+  generate verbose logs, convert those logs to benchdnn command lines using a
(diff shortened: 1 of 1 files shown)


dzarukin (commented):
I'd say that dumping JSON makes more sense for the targeted scenario.
The major caveat here is backward compatibility of dumped JSON with benchdnn. New features, esp. without a default value, or with different default values than it used to be, will likely lead to JSON files that will require additional handling on benchdnn side and all previously dumped JSON might come unusable.

It's likely possible to get around it by keeping a tool that can update JSON files kept internally (a-la benchdnn inputs) but it's still an extra tool though it should be more light weighted than a full scope verbose converter to JSON/benchdnn line. Keeping that logic inside benchdnn will unlikely be feasible on a long distance.

If you decide to follow this path, I'd suggest to invert the CMake logic same as for VERBOSE - users may opt-in for env var disabling instead of enabling, and env var used to dump files will follow dumping practices - global indexing for all graphs, standardized name, etc (what dnnl_jit_dump does).

ElaineBao:
> I'd say that dumping JSON makes more sense for the targeted scenario. The major caveat here is backward compatibility of dumped JSON with benchdnn. New features, esp. without a default value, or with different default values than it used to be, will likely lead to JSON files that will require additional handling on benchdnn side and all previously dumped JSON might come unusable.
> 
> It's likely possible to get around it by keeping a tool that can update JSON files kept internally (a-la benchdnn inputs) but it's still an extra tool though it should be more light weighted than a full scope verbose converter to JSON/benchdnn line. Keeping that logic inside benchdnn will unlikely be feasible on a long distance.
> 
> If you decide to follow this path, I'd suggest to invert the CMake logic same as for VERBOSE - users may opt-in for env var disabling instead of enabling, and env var used to dump files will follow dumping practices - global indexing for all graphs, standardized name, etc (what dnnl_jit_dump does).

Thanks @dzarukin , I’ve added a new option for JSON dumping to reflect your suggestions. Regarding backward compatibility and standardized dumping practices, since these are ongoing concerns, Do you think we can discuss them separately? Let me create a dedicated JIRA for that topic. (edit: created MFDNN-14153)

TaoLv:
@LuFinch @majing921201 Can you please take a look at this and check if it helps PyTorch integration? Thanks.

rfcs/20250902-graph-verbose-pipeline/README.md
```diff
@@ -0,0 +1,165 @@
...
+  post-processing may still be required for downstream tools
+
+### Option 3: Enable `ONEDNN_ENABLE_GRAPH_DUMP` by Default
+
+This option proposes enabling the `ONEDNN_ENABLE_GRAPH_DUMP` CMake option by
+default, so users do not need to manually configure it for graph debugging. By
+default, no graph JSON files are generated unless the
+`ONEDNN_GRAPH_DUMP` environment variable is set, so performance remains
+unaffected. When `ONEDNN_GRAPH_DUMP=subgraph` is specified, graph JSON files are
+dumped, and corresponding verbose lines are printed during partition compilation:
```
TaoLv: Better to clarify that, in this case, the JSON file is dumped during partition compilation (calling the API partition::compile(...)).

LuFinch:
Enabling graph dump LTGM.

rfcs/20250902-graph-verbose-pipeline/README.md
ElaineBao: thanks for the suggestion, done.

majing921201:
LGTM

TaoLv (approved):
Implemented as part of #3957

vpirogov (approved):

PR #3857, opened 2025-09-02
rfcs: graph api: add accumulation mode support

TaoLv:
- Rendered version: [link](https://github.com/uxlfoundation/oneDNN/blob/lvtao/rfcs/graph-acc-mode/rfcs/20250901-graph-api-acc-mode/README.md)
- Implementation: #3844

### DIFF
--- a/rfcs/20250901-graph-api-acc-mode/README.md
+++ b/rfcs/20250901-graph-api-acc-mode/README.md
@@ -1,0 +1,113 @@
+# Graph API: Add Accumulation Mode Support
+
+## Overview
+
+By default, oneDNN uses `f32` for floating point computation or `s32` for
+integer computation as the accumulation data types. However, on some platforms,
+using smaller accumulation data types (e.g., `f16`) can result in additional
+speed improvement.
+
+Introducing `f16` as the accumulation data type can achieve up to a 2x speedup
+for `f16` inference while maintaining a high level of accuracy. It is important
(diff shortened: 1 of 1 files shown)


rfcs/20250901-graph-api-acc-mode/README.md
```diff
@@ -0,0 +1,113 @@
...
+Please note that the proposal focuses on adding `accumulation_mode` attribute to
+MatMul, as required by SDPA pattern definition. In the future, the same
+attribute and value options also can be supported by other operations (eg.
+Convolution, Softmax, etc.) in the same way. Before that, setting
+`accumulation_mode` to other operations will lead to failures during the
+graph/operation creation.
+
+Please also note that, `f16` (and other) accumulation modes may not be supported
+on all platforms supported by oneDNN. If the accumulation mode is not
+supported/implemented on the target platform, partition compilation may fail.
```
dzarukin, line 111: It would be great to provide a verbose output for such scenarios if it's not yet implemented.

TaoLv: Yes, it's already implemented in the library. Attributes will be verified by the op schema. Any undefined attribute will trigger this check verbose: https://github.com/uxlfoundation/oneDNN/blob/main/src/graph/interface/op_schema.cpp#L315 .

PR #3859, opened 2025-09-02
ci: aarch64: Add a git bisect for unit test failures

Ryo-not-rio:
If a unit test fails in the nightly, automatically trigger a git bisect and output result as a table


Radu2k (approved):

.github/automation/aarch64/git_bisect.sh
```diff
@@ -0,0 +1,35 @@
...
+BUILD_DIR=$3
+CMD=${@:4}
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+ONEDNN_TEST_SET=NIGHTLY ONEDNN_BUILD_GRAPH=OFF ${SCRIPT_DIR}/build.sh --configure
+git bisect reset
+git bisect start
+git bisect good ${GOOD}
+git bisect bad ${BAD}
+git bisect run sh -c "make -j -C ${BUILD_DIR} && ${BUILD_DIR}/tests/benchdnn/benchdnn ${CMD}"
```
Sqvid: You probably want to specify a number for `-j`. The default is to not limit it. Or you could use `ninja` which will default to `nproc`.

.github/automation/aarch64/bisect_ctest.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python3
```
Sqvid: Best to do `#!/usr/bin/env python` so that it respects virtualenvs.

.github/automation/aarch64/git_bisect.sh
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+
+# Usage: ./git_bisect.sh GOOD_SHA BAD_SHA BUILD_DIR BENCHDNN_CMD
+set -o pipefail
+
+GOOD=$1
```
Sqvid, line 23: Let's verify the arguments were actually passed. The errors are not helpful otherwise.

.github/automation/aarch64/git_bisect.sh
```diff
@@ -0,0 +1,35 @@
...
+BAD=$2
+BUILD_DIR=$3
+CMD=${@:4}
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+ONEDNN_TEST_SET=NIGHTLY ONEDNN_BUILD_GRAPH=OFF ${SCRIPT_DIR}/build.sh --configure
```
Sqvid: The [correct option seems to be](https://github.com/uxlfoundation/oneDNN/blob/f4dd76bda1a5660b9ca8143a307539ac17c96164/.github/automation/aarch64/build.sh#L35) `configure` not `--configure`.

Sqvid (changes requested):
Thanks for the patch. In general this seems like a case where the `bash` script is just calling very few lines of code. Might it be better to just do it from the python script?

.github/automation/aarch64/git_bisect.sh
Ryo-not-rio: done

.github/automation/aarch64/git_bisect.sh
Ryo-not-rio: dine

.github/automation/aarch64/bisect_ctest.py
Ryo-not-rio: done

.github/automation/aarch64/git_bisect.sh
Ryo-not-rio: done

Sqvid:
> Thanks for the patch. In general this seems like a case where the bash script is just calling very few lines of code. Might it be better to just do it from the python script?

Thanks for the changes. I had left a comment earlier; let me know what you think.

Ryo-not-rio:
ah, sorry missed that message. I think it would be useful to keep the bash script as is, since we can just use that script locally for testing purposes as well

.github/automation/aarch64/git_bisect.sh
Sqvid: 🍴

.github/automation/aarch64/bisect_ctest.py
Sqvid: https://github.com/uxlfoundation/oneDNN/blob/09a09b4153db51dfc497897178f89e9fc00b491c/.github/automation/aarch64/bisect_ctest.py#L1

It's the same isn't it?

Sqvid:
>ah, sorry missed that message. I think it would be useful to keep the bash script as is, since we can just use that script locally for testing purposes as well

I am fine to keep them as separate scripts but python isn't any harder to run locally surely?

.github/automation/aarch64/git_bisect.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+
+# Usage: ./git_bisect.sh GOOD_SHA BAD_SHA BUILD_DIR BENCHDNN_CMD
+set -o pipefail
```
Sqvid: Probably a decent case for `set -uo pipefail`

Sqvid (changes requested):

.github/automation/aarch64/git_bisect.sh
Ryo-not-rio: done

Ryo-not-rio:
The python script we have now runs git bisect on a ctest log whereas the bash script runs a benchdnn command directly, so the usage will get confusing if we merge the bash script and python script together

.github/automation/aarch64/bisect_ctest.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python
```
Sqvid: `#!/usr/bin/env python`

.github/automation/aarch64/git_bisect.sh
```diff
@@ -0,0 +1,40 @@
...
+BUILD_DIR=$3
+CMD=${@:4}
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+if [ "$#" -ne 4 ]; then
+    echo "git_bisect.sh: Illegal number of parameters"
+    exit 1
+fi
+
+ONEDNN_TEST_SET=SMOKE ONEDNN_BUILD_GRAPH=OFF ONEDNN_ACTION=configure ONEDNN_THREADING=OMP ${SCRIPT_DIR}/build.sh
```
Sqvid: Why `ONEDNN_TEST_SET=SMOKE`? Are you just trying to speed up the build? There is `NO_CORR` if you dont want any tests.

Sqvid: Also what if the failure is a graph test? Don't you need `ONEDNN_BUILD_GRAPH=ON`?

Ryo-not-rio: done!

Sqvid: > Also what if the failure is a graph test? Don't you need ONEDNN_BUILD_GRAPH=ON?

.github/automation/aarch64/bisect_ctest.py
Ryo-not-rio: done

.github/automation/aarch64/git_bisect.sh
Ryo-not-rio: done

Sqvid:
Not sure why this triggers a review from @uxlfoundation/onednn-arch. @vpirogov can we merge this?

vpirogov:
> Not sure why this triggers a review from @uxlfoundation/onednn-arch. @vpirogov can we merge this?

Looks like this review was manually requested. It's not blocking promotion and can be removed if you don't need it.

PR #3864, opened 2025-09-02
xe: gemm: relax reshaping requirements

Simonsays095:
Addresses [MFDNN-14112](https://jira.devtools.intel.com/browse/MFDNN-14112). Recent changes try to preempt failing cases, but they were too strict. Instead, we rely on the return status of `memory_desc_reshape` directly to skip reshaping.

Bit of context: the `md_ok` check was added before the `safe_reshape` check was added. The latter is the better solution (allowing more cases through reshapes), making the `md_ok` checks obsolete.

### DIFF
--- a/src/gpu/intel/matmul/gemm.hpp
+++ b/src/gpu/intel/matmul/gemm.hpp
@@ -86,18 +86,6 @@ struct gemm_t : public primitive_t {
                 if (!allow_reshape || !(reshape_2d || reshape_3d))
                     return status::success;
 
-                // memory_desc_reshape does not support strided matrices. In these cases, we want
-                // to gracefully exit without reshaping
-                auto md_ok = [](const memory_desc_wrapper &mdw) -> bool {
-                    if (mdw.format_any()) return true;
-                    if (mdw.is_dense()) return true;
-                    return false;
-                };
-                bool ok = true;
-                ok = ok && md_ok(a_md);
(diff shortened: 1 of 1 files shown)


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -127,6 +115,7 @@ struct gemm_t : public primitive_t {
                 squash_dims(bia_dims, bias_md->dims, ndims, reshape_size);
 
                 // Cannot reshape if bias is broadcast across a subset of squashed dimensions
+                bool ok = true;
                 ok = ok
                         && IMPLICATION(with_bia,
                                 utils::one_of(bia_dims[0], 1, c_dims[0]));
```
atkassen: nit:
```suggestion
                bool ok = IMPLICATION(with_bia,
                        utils::one_of(bia_dims[0], 1, c_dims[0]));
```

rjoursler (approved):

src/gpu/intel/matmul/gemm.hpp
Simonsays095: Thanks, fixed.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #3865, opened 2025-09-02
xe: jit: enable parsing `static T T::parse()`

rjoursler:


rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv

src/gpu/intel/jit/utils/utils.hpp
```diff
@@ -785,8 +785,7 @@ struct has_parse_t {
 
 template <typename T>
 struct has_parse_t<T,
-        decltype(std::declval<T>().parse(std::declval<std::istream &>()),
-                void())> {
+        decltype(std::declval<T>().parse(std::declval<std::istream &>()))> {
```
echeresh: Random spot. I see that `parse()` interfaces mostly follow `void parse(istream&)` form. Can we make `type_t` aligned with the rest to have consistency? Do you see any issues with that?

```
 grep '([ :])parse.*stream' src -rIP
```

rjoursler: > Do you see any issues with that?

The current paradigm has a fairly major limitation as it requires structures to be default constructible. By switching to interfaces like `static T T::parse`, we remove this limitation. One additional (minor) benefit is this reduces the work required for parsing, as we do not have to default construct an object before subsequently overwriting it.

echeresh: @rjoursler I see, and I agree in general but since we already follow the other form in other places I suggest to try to make it consistent. For `type_t` in particular the requirement to be default constructible also comes from other places, ~~e.g. from `layout_t`~~ (apparently not `layout_t` but there are a lot of places relying on the default `type_t` constructor in the library).

To me, this is majorly a stylistic change so if you feel it's worth to change all similar `parse()` interfaces - I'm fine with that. But otherwise, I would suggest to align `type_t::parse()` - this is more in line with other code and less changes in general.

rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv

src/gpu/intel/jit/utils/utils.hpp
rjoursler: Sure, I can align with the other interfaces (`type_t` was originally non-standard anyway due to using a constructor for parsing). The main reason for the change is to prevent inconsistency if we end up with a non-default constructible object, but I agree this is fairly inconsequential concern at the moment.

echeresh (approved):

kealan-barbieri (approved):

PR #3869, opened 2025-09-03
xe: jit: introduce stable interface for object types

rjoursler:
Introduces a stable interface to `object_t`, `expr_t`, and `stmt_t`. To accomplish this, `ref_count_t` and `type_info_t` were moved to be private/hidden members of `object_impl_t`, a number of defines were replaced with CRTP injectors, SANITY_CHECK was removed (as it doesn't seem to provide much value IME), and some `object_` prefixes were replaced with an `object_t` namespace. This also contains the fix from #3865.

### DIFF
--- a/src/gpu/intel/jit/ir/include/object.hpp
+++ b/src/gpu/intel/jit/ir/include/object.hpp
@@ -99,10 +99,10 @@ protected:
     }
 
     struct info_t {
-        constexpr info_t(void *uid, bool is_expr, bool is_stmt)
+        constexpr info_t(const void *uid, bool is_expr, bool is_stmt)
             : uid(uid), is_expr(is_expr), is_stmt(is_stmt) {}
 
-        void *uid = nullptr;
+        const void *uid = nullptr;
         bool is_expr = false;
         bool is_stmt = false;
 
@@ -111,6 +111,12 @@ protected:
             return !operator==(other);
         }
     };
+
+    // std::type_info objects may differ between TUs. Deduplicate them in a
+    // single TU to ensure uniqueness. This is used to improve performance of
+    // object equality checks, as std::type_info comparisons may be relatively
+    // slow.
+    static const void *get_uid(const std::type_info &);
 
     impl_t(info_t info) : info_(info) {};
 
@@ -144,8 +150,9 @@ struct info_t {
 protected:
     friend class impl_t;
     static impl_t::info_t get_info() {
-        return impl_t::info_t(
-                (void *)get_info, is_expr_t<T>::value, is_stmt_t<T>::value);
+        static const void *uid = impl_t::get_uid(typeid(T));
+        return impl_t::info_t(static_cast<const void *>(uid),
+                is_expr_t<T>::value, is_stmt_t<T>::value);
     }
 
 private:
--- a/src/gpu/intel/jit/ir/object.cpp
+++ b/src/gpu/intel/jit/ir/object.cpp
@@ -14,6 +14,8 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include <typeindex>
+
 #include "gpu/intel/jit/ir/core.hpp"
 #include "gpu/intel/jit/utils/utils.hpp"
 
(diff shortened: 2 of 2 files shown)


rjoursler:
make test
disable test_device_cpu

third_party/ngen/ngen_gen12.hpp
```diff
@@ -725,13 +725,15 @@ static inline void encodeTernaryTypes(Instruction12 &i, D dst, S0 src0, S1 src1,
 template <typename S0, typename Tag>
 static inline void encodeTernarySrc0(Instruction12 &i, S0 src0, Tag tag)
 {
-    i.ternary.src0 = encodeTernaryOperand12<false>(src0, tag).bits;
-    i.ternary.src0Mods = src0.getMods();
+    auto i2 = i; /* separate variable to avoid gcc13 bug */
```
echeresh, line 728: Leaving a comment to not forget to port this to the nGEN repo.

src/gpu/intel/jit/ir/include/object.hpp
```diff
@@ -0,0 +1,365 @@
...
+
+template <typename T>
+struct info_t {
+    using self_type = T;
+
+protected:
+    friend class impl_t;
+    static impl_t::info_t get_info() {
+        return impl_t::info_t(
+                (void *)get_info, is_expr_t<T>::value, is_stmt_t<T>::value);
```
echeresh: I remember something like that was used in the first version but it caused issues due to MSVC folding multiple functions into one: https://github.com/uxlfoundation/oneDNN/commit/6383a4f1ad

I tried a toy example with MSVC - observing this folding even with default options: https://godbolt.org/z/44eaPWonG

Please take a look if it's an issue here. Looks like the trick with a static variable like in the commit could help.

echeresh (approved):

src/gpu/intel/jit/ir/include/object.hpp
rjoursler: Thanks for pointing this out, I had thought there was likely to be some issue here and intended to investigate further.  I went ahead and switched to a registry system which should provide effectively the same performance, while still keeping the id/impl decoupling, and avoiding all the pitfalls of pointers duplication/deduplication.

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/ir/object.cpp
```diff
@@ -0,0 +1,146 @@
...
+namespace jit {
+expr_t const_fold_non_recursive(const expr_t &expr);
+
+object_t object::impl_t::_mutate(ir_mutator_t &mutator) const {
+    return *this;
+}
+void object::impl_t::_visit(ir_visitor_t &visitor) const {}
+
+const void *object::impl_t::get_uid(const std::type_info &info) {
+    static std::unordered_map<std::type_index, const void *> type_registry;
```
echeresh, line 35: Why do we need registry? Can we simply return `&info` here?

rjoursler: We likely could, but there is a lot of nuance here that is seems better to avoid. From [cpp reference](https://en.cppreference.com/w/cpp/language/typeid.html):

> There is no guarantee that the same [std::type_info](https://en.cppreference.com/w/cpp/types/type_info.html) object will be referred to by all evaluations of the typeid expression on the same type, although they would compare equal, [std::type_info::hash_code](https://en.cppreference.com/w/cpp/types/type_info/hash_code.html) of those type_info objects would be identical, as would be their [std::type_index](https://en.cppreference.com/w/cpp/types/type_index.html).

Because of this, if `typeid()` were called multiple times for a given type we could get different results. This shouldn't be an issue for the current implementation (due to external linkage guarantees on namespaces scope classes and the use of static variables), but it would be easy to misuse `get_uid` and introduce a very subtle and hard to catch bug.

echeresh: Sorry, yes, I was rather wanted to ask why we can't keep the current code, operating directly with `std::type_info`:

https://github.com/uxlfoundation/oneDNN/blob/1c01e94128d25465739dd8c042145a0899580910/src/gpu/intel/jit/ir/core.hpp#L264

But now I see that the registry allows to eliminate a `std::type_info` comparison so it makes sense now.

PR #3870, opened 2025-09-03
xe: gemm: disable XeLP grouped scales

Simonsays095:
Addresses [MFDNN-13992](https://jira.devtools.intel.com/browse/MFDNN-13992) and [MFDNN-14044](https://jira.devtools.intel.com/browse/MFDNN-14044). Grouped scales increase register pressure and break XeLP kernels. In this PR, we just skip the optimized implementation for these cases.

Benchdnn is already aware that cases like this can fail, and will return "Case not supported" instead of "unimplemented" if not supported via reference.

### DIFF
--- a/src/gpu/intel/gemm/jit.hpp
+++ b/src/gpu/intel/gemm/jit.hpp
@@ -24,6 +24,7 @@
 #include "common/c_types_map.hpp"
 #include "common/gemm_utils.hpp"
 #include "common/utils.hpp"
+#include "gpu/intel/compute/device_info.hpp"
 #include "gpu/intel/compute/kernel.hpp"
 #include "gpu/intel/compute/zero_pool.hpp"
 #include "gpu/intel/gemm/jit/gen_kernel.hpp"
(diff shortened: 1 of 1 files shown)


src/gpu/intel/gemm/jit.hpp
```diff
@@ -232,6 +233,15 @@ struct gen_t : public primitive_t {
             VDISPATCH_GEMM(IMPLICATION(with_binary, arch_ >= arch_t::xe_hp),
                     VERBOSE_UNSUPPORTED_ARCH, "gpu");
 
+            // Grouped scales break pre-XeHPG kernels due to increased register pressure
```
kealan-barbieri, line 236: should we not just adjust strategies here? Or if we dont expect arbitrary attributes to be supported by strategies maybe we need a method of sorting them by support.

Simonsays095: In general, grouped quantization increases register pressure (depending on group size) and therefore any kernels in the catalog could break when using grouped quantization. Nowadays, we tune kernels (and test in CI) with this in mind - it's just these older kernels that have an issue.

I could adjust the strategy, but I don't think it's worth the effort and potential for bugs, for XeLP. I could be convinced otherwise, though. I don't generally like the approach of skipping cases to avoid broken behaviour either - but this was quick and easy for this older platform.

Simonsays095: Oh, one more note on support: these cases started breaking when we added CI tests for grouped weights decompression cases. These cases have always failed, but it was only recently exposed via #3255.

kealan-barbieri: Yeah I dont have any particular concern about these cases if XELP isnt a target platform for this feature, just that as you say this is now a universal potential problem for grouped quantization attributes. In the future it may be worth maintaining sets of strategies that can have eg larger unrolls if they are specifically disabled when grouped quant attrs are present.

Simonsays095: Yeah, we need some sort of either:
1. Fallback behaviour if a kernel fails to compile
2. Adjustable kernel parameters

Option 1 is "safer" as it just covers more cases, while option 2 has performance implications. Something like a strategy tag for "with grouped quantization" that we can enable/disable for the variants (or leave unset to allow either) would probably be easiest.

kealan-barbieri: Agreed, I was thinking a strategy tag would be the most obvious but its not clear if its warranted yet because most of these issues only seem to affect older architectures, thanks for fixing this one!

PR #3871, opened 2025-09-03
src: address static analyzers hits

dzarukin:
[MFDNN-14150](https://jira.devtools.intel.com/browse/MFDNN-14150)

This PR addresses latest Coverity hits (gpu_primitive.hpp and stream_submit_cpu_primitive.cpp) and and some less intelligent hits produced by clang-tidy while trying to find an option that would mimic Coverity behavior.
I failed with the latter task.

### DIFF
--- a/src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_convolution.cpp
+++ b/src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_convolution.cpp
@@ -60,7 +60,9 @@ jit_avx512_core_x8s8s32x_1x1_convolution_fwd_t::maybe_adjust_dw_oscales(
         const float *dw_wei_scales) const {
     if (!pd()->jcp_.with_dw_conv) return nullptr;
 
-    const auto &jcp_dw = pd()->jcp_dw_;
+    const auto *jcp_dw = pd()->jcp_dw_;
+    if (!jcp_dw) return nullptr;
+
     memory_tracking::grantor_t dw_scratchpad(
             scratchpad, memory_tracking::names::prefix_fusion);
     auto dw_loc_scales
@@ -193,7 +195,7 @@ void jit_avx512_core_x8s8s32x_1x1_convolution_fwd_t::execute_forward_thr(
             : jcp.nb_load_blocking_max;
 
     // Begin: declare Variables needed for dw conv.
-    const auto jcp_dw = pd()->jcp_dw_;
+    const auto *jcp_dw = pd()->jcp_dw_;
     const auto &dw_pd = pd()->dw_conv_pd_;
     memory_tracking::grantor_t dw_scratchpad(
             scratchpad, memory_tracking::names::prefix_fusion);
--- a/src/gpu/gpu_primitive.hpp
+++ b/src/gpu/gpu_primitive.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2024 Intel Corporation
+* Copyright 2024-2025 Intel Corporation
 *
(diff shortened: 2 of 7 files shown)


dzarukin:
make test

src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_convolution.cpp
```diff
@@ -61,6 +61,8 @@ jit_avx512_core_x8s8s32x_1x1_convolution_fwd_t::maybe_adjust_dw_oscales(
     if (!pd()->jcp_.with_dw_conv) return nullptr;
 
     const auto &jcp_dw = pd()->jcp_dw_;
+    if (!jcp_dw) return nullptr;
```
densamoilov, line 64: Can you please replace `const auto &jcp_dw = pd()->jcp_dw_;` with `const auto *jcp_dw = pd()->jcp_dw_;` to improve readability?

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -138,8 +138,12 @@ void brgemm_convolution_fwd_t<isa>::pd_t::init_batch(int icc,
                         brg_batch[b_idx].offset.B = (dim_t)B_addr - (dim_t)ptrB;
                     }
                     if (jcp.max_vpad != 0) {
-                        brg_batch[b_idx].vvpad.top = kw_top_vpads[kw];
-                        brg_batch[b_idx].vvpad.bottom = kw_bottom_vpads[kw];
+                        // It's assumed that buffers are always non-empty in
+                        // case of `jcp.max_vpad != 0`. Check to be extra safe.
+                        if (kw_top_vpads)
```
densamoilov: I think this is redundant because `jcp.max_vpad != 0` is the condition when the arrays are defined and it's explicitly checked here before using them. 

It's the same as:
```cpp
if (jcp.with_bias) {
    const data_t *__restrict bia_arr = bia_base + g * jcp.oc;
    // use bia_arr
}
```

densamoilov (approved):

src/gpu/gpu_primitive.hpp
```diff
@@ -74,6 +74,8 @@ struct primitive_t : public impl::primitive_t {
             std::shared_ptr<impl::primitive_t> &primitive,
             const std::shared_ptr<primitive_desc_t> &pd,
             impl::engine_t *engine) {
+        if (!pd) return status::invalid_arguments;
```
rjoursler, line 77: Just to check: did you consider modifying this function to take a `const primitive_desc_t &` rather than a shared pointer? The shared pointer doesn't appear to be used anywhere and this would just remove this invalid case from the interface (although it is not worthwhile to implement the necesssary refactoring).

dzarukin: It wouldn't help with static analyzer as it would require to dereference a pointer at a higher level, thus, only more checks instead of at this place.

src/cpu/x64/jit_avx512_core_x8s8s32x_1x1_convolution.cpp
dzarukin: Absolutely, adjusted that.

src/cpu/x64/jit_brgemm_conv.cpp
dzarukin: Sure thing, dropped this change and all those tentative (referred as less intelligent) checks and changes, just kept essential ones. I guess I just wanted to show what clang-tidy is really up to. :)

dzarukin:
make test

PR #3872, opened 2025-09-04
common: verbose: split macros for attribute checks during creation / execution

avmanerikar:
# Description

Verbose macros for attribute checks are called during both primitive creation and execution but the verbose level printed for both is `exec:check`. The PR splits the macros to print the correct logging levels.


avmanerikar:
make test

dzarukin (approved):

PR #3873, opened 2025-09-04
[GPU] Checks for double blocked format softmax/matmul

h-sadia:
This PR is adding checks to avoid failures for double blocked format for softmax/mamtul. Since Matmul gives input to softmax layers when used in larger context and we don't have any particular tests in benchdnn for this, we will just return unimplemented for such cases from init prim to avoid the false illusion. 
This is to avoid cases like: [MFDNN-13673](https://jira.devtools.intel.com/browse/MFDNN-13673)

### DIFF
--- a/src/gpu/intel/gemm/jit_xe_hp_systolic.cpp
+++ b/src/gpu/intel/gemm/jit_xe_hp_systolic.cpp
@@ -211,7 +211,7 @@ bool xe_hp_systolic_t::pd_t::use_nocopy() {
 
     bool xehpc = (arch >= compute::gpu_arch_t::xe_hpc);
 
-    if (any_prepacked_ || (packed_a_ && packed_b_)) return false;
+    if (any_prepacked_) return false;
 
     // Use no-copy for gemv/ger cases.
     if (d->m() <= 1 || d->n() <= 1 || d->k() <= 1) return true;
--- a/src/gpu/intel/softmax/simple.hpp
+++ b/src/gpu/intel/softmax/simple.hpp
@@ -49,6 +49,10 @@ struct simple_fwd_t : public primitive_t {
 
             using namespace data_type;
             using skip_mask_t = primitive_attr_t::skip_mask_t;
+            const int num_blocks = 2;
+            auto is_not_double_blk
+                    = src_md()->format_desc.blocking.inner_nblks != num_blocks;
+
             VDISPATCH_SOFTMAX(is_fwd(), VERBOSE_BAD_PROPKIND);
             VDISPATCH_SOFTMAX(
                     utils::one_of(src_dt, f64, f32, f16, bf16, u8, s8),
@@ -73,6 +77,7 @@ struct simple_fwd_t : public primitive_t {
             VDISPATCH_SOFTMAX(attr()->has_default_values(skip_mask_t::scales
                                       | skip_mask_t::post_ops),
                     VERBOSE_UNSUPPORTED_ATTR);
+            VDISPATCH_SOFTMAX(is_not_double_blk, VERBOSE_UNSUPPORTED_TAG);
             VDISPATCH_SOFTMAX(attr_scales_ok(), VERBOSE_UNSUPPORTED_SCALES_CFG);
             VDISPATCH_SOFTMAX(post_ops_ok(), VERBOSE_UNSUPPORTED_POSTOP);
             VDISPATCH_SOFTMAX_SC(


src/gpu/intel/softmax/simple.hpp
```diff
@@ -49,6 +49,10 @@ struct simple_fwd_t : public primitive_t {
 
             using namespace data_type;
             using skip_mask_t = primitive_attr_t::skip_mask_t;
+            const int num_blocks = 2;
+            auto is_double_blk
+                    = src_md()->format_desc.blocking.inner_nblks != num_blocks;
```
rjoursler, line 54:
```Suggestion
                    = src_md()->format_desc.blocking.inner_nblks >= num_blocks;
```

src/gpu/intel/softmax/simple.hpp
```diff
@@ -73,6 +77,7 @@ struct simple_fwd_t : public primitive_t {
             VDISPATCH_SOFTMAX(attr()->has_default_values(skip_mask_t::scales
                                       | skip_mask_t::post_ops),
                     VERBOSE_UNSUPPORTED_ATTR);
+            VDISPATCH_SOFTMAX(is_double_blk, VERBOSE_UNSUPPORTED_TAG);
```
rjoursler: Since this appears to be an actual use case, why don't we just modify the softmax implementation to use [OFF_MD](https://github.com/uxlfoundation/oneDNN/blob/f47f4e9261ec7038745849efb9418e6446f335db/src/gpu/intel/include/types.h#L945-L946) based offset calculations? I don't see any reason softmax should have custom (and apparently incorrect) logic for offset calculations.

rjoursler (commented):
Why is the root cause of this failure not also due to the different layouts from the sample problem? In the linked test case I see:
```
--stag=abCD32c16d --dtag=abcd
```
But there is one macro used for offset calculation ([here](https://github.com/uxlfoundation/oneDNN/blob/f47f4e9261ec7038745849efb9418e6446f335db/src/gpu/intel/softmax/simple.cl#L89) and [here](https://github.com/uxlfoundation/oneDNN/blob/f47f4e9261ec7038745849efb9418e6446f335db/src/gpu/intel/softmax/simple.cl#L131)).

h-sadia:
make test perf-gpu
set primitive=softmax matmul

src/gpu/intel/softmax/simple.hpp
h-sadia: hmm... Hi Roy. Thanks for pointing it out. Are you suggesting to rework this [OFF](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/softmax/simple.h#L37) macro here? Since we are changing the way we are are calculation the offset for data and padding depending on the softmax axis.

h-sadia:
> Why is the root cause of this failure not also due to the different layouts from the sample problem? In the linked test case I see:
> 
> ```
> --stag=abCD32c16d --dtag=abcd
> ```
> 
> But there is one macro used for offset calculation ([here](https://github.com/uxlfoundation/oneDNN/blob/f47f4e9261ec7038745849efb9418e6446f335db/src/gpu/intel/softmax/simple.cl#L89) and [here](https://github.com/uxlfoundation/oneDNN/blob/f47f4e9261ec7038745849efb9418e6446f335db/src/gpu/intel/softmax/simple.cl#L131)).

As far as I see, softmax tests in benchdnn do have different stag and dtag (with dtag being any), so I am not sure if we can pin the correction issue to using one macro. Ofcourse this could also be one of the problems. If we want to rewrite our macros for double_blocking, we can therefore check this furthermore. Does OFF_MD cater for double blocking?

src/gpu/intel/softmax/simple.hpp
rjoursler: Yeah, the general idea would be to swizzle inputs to OFF_MD so we can normalize which input corresponds to the axis dimension. We also likely need a version for both source and dst. For reference we can make calls like in [ref_prelu](https://github.com/uxlfoundation/oneDNN/blob/88617595be725e313fa07df5e70633f808cfa248/src/gpu/intel/prelu/ref.cl#L33-L35). In order to use those macros, all that should be required is to add a call to [def_memory_desc_info](https://github.com/uxlfoundation/oneDNN/blob/88617595be725e313fa07df5e70633f808cfa248/src/gpu/intel/prelu/ref.cpp#L89-L91) on the relevant buffers during kernel_ctx initialization.

> Does OFF_MD cater for double blocking?

Yes, it should support double blocking, which is largely why the [defintion](https://github.com/uxlfoundation/oneDNN/blob/88617595be725e313fa07df5e70633f808cfa248/src/gpu/intel/include/types.h#L913-L943) is so long.

h-sadia: We will still need to work the softmax axis part in this - so it won't be so straightforward. Basically, I can try to create a small use case for OFF_MD macro and see if what you are suggesting works for it and is worth the effort and time to extend to double blocking. We are not supporting double blocking formats (except maybe for PVC) anymore so need to see the trade off of time vs usage in these cases.

src/gpu/intel/softmax/simple.hpp
Simonsays095: nit: change name to `is_not_double_blk` or check for `== num_blocks` and update the `VDISPATCH_SOFTMAX` check.

src/gpu/intel/gemm/jit_xe_hp_systolic.cpp
```diff
@@ -211,7 +211,7 @@ bool xe_hp_systolic_t::pd_t::use_nocopy() {
 
     bool xehpc = (arch >= compute::gpu_arch_t::xe_hpc);
 
-    if (any_prepacked_ || (packed_a_ && packed_b_)) return false;
+    if (any_prepacked_) return false;
```
Simonsays095, line 214: Can you explain what this change is doing? My assumption was that `xe_hp_systolic_t` doesn't support double-packed inputs - is this just an obsolete check?

Simonsays095 (approved):

src/gpu/intel/gemm/jit_xe_hp_systolic.cpp
h-sadia: As we were still getting double blocked output for PVC in one of the framework tests, @petercad suggested this patch. We are disabling the double blocked output we received through matmul into softmax.

src/gpu/intel/softmax/simple.hpp
h-sadia: yeah, i was wondering about that too... let me fix the var name.

src/gpu/intel/gemm/jit_xe_hp_systolic.cpp
petercad: @Simonsays095 -- `xe_hp_systolic_t` supports any combination of blocked/plain formats for A/B/C. Here `packed_a_` means "A is 'any' format", so the old logic was "always use xe_hp_systolic_t if the user specifically is passing in a blocked format we support, or if they left one of A/B as any." The new logic removes the second part, so if the user passes any for A and/or B it's not treated specially, and we go through the regular set of size/transpose/alignment heuristics to determine whether to use xe_hp_systolic_t.

h-sadia:
The only regression for matmul observed are differences between run to run.

PR #3892, opened 2025-09-07
x64: cpu: matmul: Fuse copy of A in cases of unaligned K dimension

yair-obodovsky:
When the K dimension is not aligned to the cache line, loading matrix A can degrade the performance of highly optimized BRGEMMs by up to 2×. To prevent this, matrix A is copied to an aligned buffer on the fly, ensuring proper cache line alignment. This avoids misaligned memory accesses and reduces L1 bandwidth usage, resulting in improved overall performance.

### DIFF
--- a/src/cpu/x64/brgemm/brgemm.cpp
+++ b/src/cpu/x64/brgemm/brgemm.cpp
@@ -532,6 +532,17 @@ status_t brgemm_desc_set_attr(
         return status::unimplemented;
     }
 
+    // fused copy A already copies A for the BRGEMM. No need for partial copy.
+    if (brgattr.hint_fused_copy_a
+            && (brgattr.extendable_k || brgattr.wary_A_k_tail_read)) {
+        return status::unimplemented;
+    }
+
+    if (!IMPLICATION(brgattr.hint_fused_copy_a,
+                is_superset(brg->isa_impl, amx_tile))) {
+        return status::unimplemented;
+    }
+
     brg->prfA = brgattr.hint_prfA;
     brg->prfB = brgattr.hint_prfB;
     brg->prfC = brgattr.hint_prfC;
@@ -547,6 +558,8 @@ status_t brgemm_desc_set_attr(
     if (brgattr.hint_prefetching == brgemm_kernel_prefetching_t::brgemm_prf2
             && brg->prfC.dist2 < 0)
         brg->prfC.dist2 = 0;
+
+    if (brgattr.hint_fused_copy_a) brg->fused_copy_a = true;
 
     if (brg->is_fp8
             && !utils::one_of(true,
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -238,6 +238,8 @@ struct DNNL_API brgemm_attr_t {
     // static_offsets is a static array of fixed offsets used for
     // brgemm_static_offs batch kind
     const brgemm_batch_element_t *static_offsets;
+    // hint whether to use on the fly copy of A due to unaligned accesses
+    bool hint_fused_copy_a = false;
 };
 
 struct brgemm_desc_t {
@@ -254,6 +256,8 @@ struct brgemm_desc_t {
     int LDB = 0;
     int LDC = 0;
     int LDD = 0;
+
+    bool fused_copy_a = false;
     // we use two isa_ variables
     // isa_user to store the user provided isa value
     // isa_impl to store actual implementation. This can change until the kernel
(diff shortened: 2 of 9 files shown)


src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -238,6 +238,8 @@ struct DNNL_API brgemm_attr_t {
     // static_offsets is a static array of fixed offsets used for
     // brgemm_static_offs batch kind
     const brgemm_batch_element_t *static_offsets;
+    // hint whether to use on the fly copy of A due to unaligned accesses
+    bool hint_use_otf = false;
```
densamoilov: I think we need to make the name more meaningful because "one the fly" doesn't really tell us much. I also think we need to document implications, e.g. how `hint_use_otf = true` affects "copy a" scenarios, etc.

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -723,6 +736,21 @@ int jit_brgemm_amx_uker_base_t::skipped_bd_mask(int inp_bd) noexcept {
         return skipped_bd_mask_buffer_[inp_bd];
 }
 
+size_t jit_brgemm_amx_uker_base_t::A_offset_wsp(
```
densamoilov, line 739: What are the requirements to the shape of the wsp buffer for A?

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -2177,9 +2244,18 @@ void jit_brgemm_amx_uker_base_t::maybe_pre_process_data(brgemm_iteration_t &bi,
     if (buf_offt) sub(reg_buf, buf_offt);
 }
 
+void jit_brgemm_amx_uker_base_t::pre_process_k_tail_otf(
```
densamoilov: I have a general concern about the copy-inside-brgemm-kernel approach we recently introduced to support extendable K, which we now use to handle special K cases. With this approach, we have two levels of copy logic:
1. At the primitive level
2. At the brgemm kernel level

Because of that we need to ensure that both levels of copy logic are properly used. The reason we have the copy A/B kernels externally is to keep the copy logic separate from the brgemm kernel.

What prevents us from handling the brgemm kernel-level copy at the primitive level?

+@ankalinin

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -532,6 +532,17 @@ status_t brgemm_desc_set_attr(
         return status::unimplemented;
     }
 
+    // OTF already copies A for the BRGEMM. No need for partial copy.
```
densamoilov: (random spot) Can you please share some of the cases that are affected and performance data for those?

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
rinatrap: The kernel-level copy approach consistently outperforms separate copy logic, delivering an average performance gain of 3% across a broad set of synthetic benchmarks, with improvements reaching up to 20%. Notably, no performance regressions were observed

src/cpu/x64/brgemm/brgemm.cpp
rinatrap: To evaluate the impact of misaligned memory access on performance, each layer, defined by the parameters m, k, and n, was tested with K values ranging from K+4 to K+60, in increments of 4. This setup was designed to create intentional misalignment in accessing the A matrix.

The observed performance benefits of using fused copy A over the baseline implementation were:
- 1 core: +19% improvement
- 4 cores: +25% improvement
- 56 cores: +5% improvement

Here is the list of layer dimensions that were tested:
|   m   |   k   |   n   |
|-------|-------|-------|
|  128  |  128  |  128  |
|  256  |  256  |  256  |
|  512  |  512  |  512  |
| 1024  | 1024  | 1024  |
| 2048  | 2048  | 2048  |
| 4096  | 4096  | 4096  |
|  128  |  512  |  128  |
|  512  |  128  |  128  |
|  128  |  128  |  512  |
|  256  | 1024  | 1024  |
|  256  |  256  | 1024  |
| 1024  |  256  |  256  |
| 1024  | 8192  | 1024  |
| 1024  | 1024  | 8192  |
| 8192  | 1024  | 1024  |
| 1024  |  256  | 1024  |
| 1024  | 1024  |  256  |
| 2048  |  256  | 2048  |
|  256  | 2048  | 2048  |
| 2048  | 2048  |  256  |
| 2048  |  512  | 2048  |
|  512  | 2048  | 2048  |
| 2048  | 2048  |  512  |
| 4096  |  256  | 4096  |
|  256  | 4096  | 4096  |
| 4096  | 4096  |  256  |
| 4096  |  512  | 4096  |
|  512  | 4096  | 4096  |
| 4096  | 4096  |  512  |
| 4096  | 1024  | 4096  |
| 1024  | 4096  | 4096  |
| 4096  | 4096  | 1024  |

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
yair-obodovsky: added comment

src/cpu/x64/brgemm/brgemm_types.hpp
yair-obodovsky: done

ankalinin (approved):

tczeszun (approved):

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
densamoilov: I see, thank you for the explanation! Sounds good.

densamoilov (approved):

PR #3898, opened 2025-09-08
cpu: rv64: add support for rv64 eltwise feature with RISC-V Vector Extension

zhangjian29:
# Description

This PR introduces optimized element-wise operation kernels for RISC-V architectures using RVV (RISC-V Vector) intrinsics. The `rvv_eltwise` implementation delivers significant performance improvements for common neural network operations including ReLU, square, abs, sqrt, and more.
This initial version provides:

1. Eltwise operations utilizing RVV intrinsics, including ReLU, square, abs, sqrt, linear, clip, hardsigmoid, and hardswish.
2. Multiple data types including `f32`, `f16`, `s32`, `s8` and `u8`.
3. Supporting forward or backward executation flow, and dense or channel-padded memory layouts.

## Key Features

- **RVV-Eltwise Kernels**: Added `rvv_eltwise_fwd/bwd_kernel_*dtype` kernels for unified design pattern across operations for better maintainability.
- **Data Types**: Source and destination tensors must be one of `f32`, `f16` (Zvfh extension required), `s32`, `s8` and `u8`. The `alpha` and `beta` parameters are `f32` data type and they are used by related operations when required.
- **Memory Layouts**: Source and destination tensors must be dense or nCspBc format. Both forward and backward execution flows are supported.
- **Post-Ops**: Current implementation requires no post-ops.
- **Parallelization**: By adopting parallel methods provided by oneDNN API including `balance211`, `parallel_nd` and `parallel`, it leverages RVV-vectorize inner loops optimization and multi-core CPU performance.

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

## Performance improvements

- [x] Have you submitted performance data that demonstrates performance improvements?

All experiments are performed on the LRW platform with VLEN=128. We draw comparisons between baseline method of **scalar** implementation and our method of **RVV intrinsic** implementation, using `benchdnn` with input batch of `test_eltwise_all` (filtered reduntant test cases that `rvv_eltwise` doesn't support).

Results are as follows:

| Data Types | Scalar Runtime (s) | RVV Intrinsics </br> Runtime (s) | Improvements |
|------------|--------------------|----------------------------|--------------|
| f32        | 0.49               | 0.46                       | 6.12%       |
| s32        | 0.27               | 0.24                       | 11.11%       |
| s8         | 0.27               | 0.24                       | 11.11%       |
[...]

PR #3899, opened 2025-09-08
cpu: rv64: add support for rv64 binary feature with RISC-V Vector Extension

Ralap-Jee:
# Description

This PR introduces optimized binary operation kernels for RISC-V architectures using RVV (RISC-V Vector) intrinsics. The `rvv_binary` implementation delivers significant performance improvements for common neural network operations including ReLU, square, abs, sqrt, and more.
This initial version provides:

1. Implementing binary operations utilizing RVV intrinsics, including all operations that supported by `ref_binary`.
2. Multiple data types including `f32`, `f16`, `s32`, `s8` and `u8`.
3. Supporting dense and channel-padded memory layouts.

## Key Features

- **RVV-Binary Kernels**: Added `rvv_binary_kernel_*dtype` kernels for unified design pattern across binary operations for better maintainability.
- **Data Types**: Source and destination tensors must be one of `f32`, `f16` (Zvfh extension required), `s32`, `s8` and `u8`. The third source required by select operation must be `s8` type.
- **Memory Layouts**: Source and destination tensors must be dense or nCspBc format.
- **Post-Ops**: Current implementation requires no post-ops.
- **Parallelization**: By adopting parallel methods provided by oneDNN API including `balance211`, `parallel_nd` and `parallel`, it leverages RVV-vectorize inner loops optimization and multi-core CPU performance.

# Checklist

## General

## Performance improvements

All experiments are performed on the LRW platform with VLEN=128. We draw comparisons among 1st baseline method of **scalar** implementation , 2nd baseline method of **auto vectoration** by compiler,and our method of **RVV intrinsic** implementation, using `benchdnn` with input batch of `test_binary_all` (filtered reduntant test cases that `rvv_binary` doesn't support).

Results are as follows:

| Data Types | Scalar Runtime (s) | RVV Intrinsics </br> Runtime (s) | Improvements |
|------------|--------------------|----------------------------|--------------|
| f32        | 1.59               | 1.46                       | 8.18%        |
| s32        | 2.36               | 2.2                       | 6.78%       |
| s8         | 2.32               | 2.19                       | 5.6%       |
| u8         | 2.33               |2.18                       | 6.44%         |

| Data Types | Auto Vectora-</br>tion Runtime (s) | RVV Intrinsics </br> Runtime (s) | Improvements |
|------------|--------------------|----------------------------|--------------|
| f32        | 1.8               | 1.46                       | 18.89%        |
| s32        | 2.66               | 2.2                       | 17.29%       |
| s8         | 2.61               | 2.19                       | 16.09%       |
| u8         | 2.61               | 2.18                       | 16.48%         |

*Note that the runtimes are the total time minus rerfence and comparison time for precise comparisons.*

### New features

- [N/A] Have you published an RFC for the new feature?
- [N/A] Was the RFC approved?
- [N/A] Have you added relevant tests?

### Bug fixes

- [N/A] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [N/A] Have you added relevant regression tests?

- [N/A] Does RFC document follow the [template](https://github.com/uxlfoundation/oneDNN/blob/rfcs/rfcs/template.md#onednn-design-document-rfc)?
- [N/A] Have you added a link to the rendered document?

### DIFF
--- a/src/cpu/rv64/rvv_binary.cpp
+++ b/src/cpu/rv64/rvv_binary.cpp
@@ -1,0 +1,292 @@
+/******************************************************************************
+* Copyright 2025
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+******************************************************************************/
+
+#include <assert.h>
+#include <memory>
+#include <riscv_vector.h>
+
+#include "common/c_types_map.hpp"
+#include "common/dnnl_thread.hpp"
+#include "common/math_utils.hpp"
+#include "common/type_helpers.hpp"
+
+#include "cpu/primitive_attr_postops.hpp"
+#include "cpu/rv64/rvv_binary_kernels.hpp"
+
+#include "cpu/rv64/rvv_binary.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace rv64 {
+
+// Data type dispatch for RVV binary
+static inline void compute_binary_rvv(const alg_kind_t alg, const void *x,
+        const void *y, void *dst, const int8_t *c, const dim_t len,
+        const data_type_t dt) {
+    switch (dt) {
+        case data_type::f32:
+            rvv_binary_apply_f32(alg, static_cast<const float *>(x),
+                    static_cast<const float *>(y), static_cast<float *>(dst), c,
+                    len);
+            break;
+        case data_type::f16:
+            rvv_binary_apply_f16(alg, static_cast<const _Float16 *>(x),
+                    static_cast<const _Float16 *>(y),
+                    static_cast<_Float16 *>(dst), c, len);
+            break;
+        case data_type::s32:
+            rvv_binary_apply_s32(alg, static_cast<const int32_t *>(x),
+                    static_cast<const int32_t *>(y),
+                    static_cast<int32_t *>(dst), c, len);
+            break;
+        case data_type::s8:
+            rvv_binary_apply_s8(alg, static_cast<const int8_t *>(x),
+                    static_cast<const int8_t *>(y), static_cast<int8_t *>(dst),
+                    c, len);
+            break;
+        case data_type::u8:
+            rvv_binary_apply_u8(alg, static_cast<const uint8_t *>(x),
+                    static_cast<const uint8_t *>(y),
+                    static_cast<uint8_t *>(dst), c, len);
+            break;
+        default: assert(!"Unsupported data type for RVV binary");
+    }
+}
+
+template <data_type_t data_type>
+status_t rvv_binary_t<data_type>::execute_binary(const exec_ctx_t &ctx) const {
+    if (pd()->has_zero_dim_memory()) return status::success;
+
+    status_t status = status::success;
+    auto src0 = CTX_IN_MEM(
+            const typename prec_traits_t<data_type>::type *, DNNL_ARG_SRC_0);
+    auto src1 = CTX_IN_MEM(
+            const typename prec_traits_t<data_type>::type *, DNNL_ARG_SRC_1);
+    // src2 is optional (only for ternary ops like select); treat as s8 mask
+    const int8_t *src2_s8 = nullptr;
+    if (pd()->is_ternary_op()) {
+        src2_s8 = CTX_IN_MEM(const int8_t *, DNNL_ARG_SRC_2);
+    }
+    auto dst = CTX_OUT_CLEAN_MEM(
+            typename prec_traits_t<data_type>::type *, DNNL_ARG_DST, status);
+    CHECK(status);
+
+    const memory_desc_wrapper src0_d(pd()->src_md(0));
+    const memory_desc_wrapper src1_d(pd()->src_md(1));
+    const memory_desc_wrapper src2_d(pd()->src_md(2));
+    const memory_desc_wrapper dst_d(pd()->dst_md());
+
+    const auto nelems = dst_d.nelems(true);
+
+    const auto alg_kind = pd()->desc()->alg_kind;
+
+    if (pd()->use_dense_) {
+        const auto off0 = src0_d.offset0();
+        const auto off1 = src1_d.offset0();
+        const auto offd = dst_d.offset0();
+        const auto off2 = pd()->is_ternary_op() ? src2_d.offset0() : 0;
+
+        const auto *x_base = src0 + off0;
+        const auto *y_base = src1 + off1;
+        auto *d_base = dst + offd;
+        const int8_t *c_base
+                = pd()->is_ternary_op() ? (src2_s8 + off2) : nullptr;
+
+        parallel(0, [&](const int ithr, const int nthr) {
+            dim_t start = 0, end = 0;
+            balance211(nelems, nthr, ithr, start, end);
+            if (start == end) return;
+
+            const void *thr_x = static_cast<const void *>(x_base + start);
+            const void *thr_y = static_cast<const void *>(y_base + start);
+            void *thr_d = static_cast<void *>(d_base + start);
+            const int8_t *thr_c = c_base ? (c_base + start) : nullptr;
+            const dim_t len = end - start;
+
+            compute_binary_rvv(alg_kind, thr_x, thr_y, thr_d, thr_c, len,
+                    pd()->dst_md()->data_type);
+        });
+
+        return status::success;
+    }
+
+    if (pd()->use_nCspBc_padded_) {
+        const blocking_desc_t &blk = dst_d.blocking_desc();
+        const dim_t block = blk.inner_blks[0];
+
+        const dim_t MB = dst_d.dims()[0];
+        const dim_t C_PADDED = dst_d.padded_dims()[1] / block;
+        const dim_t C = dst_d.dims()[1] / block;
+        const dim_t tail = dst_d.dims()[1] % block;
+        const dim_t D = (dst_d.ndims() > 4 ? dst_d.dims()[2] : 1);
+        const dim_t H
+                = (dst_d.ndims() > 3 ? dst_d.dims()[dst_d.ndims() - 2] : 1);
+        const dim_t W
+                = (dst_d.ndims() > 2 ? dst_d.dims()[dst_d.ndims() - 1] : 1);
+        const dim_t SP = D * H * W;
+
+        parallel_nd(MB, C_PADDED, SP, [&](dim_t n, dim_t c, dim_t sp) {
+            // Decompose sp into D/H/W indices for proper broadcasting
+            dim_t dd = 0, hh = 0, ww = 0;
+            if (dst_d.ndims() > 4) {
+                const dim_t HW = H * W;
+                dd = sp / HW;
+                const dim_t hw_rem = sp % HW;
+                hh = hw_rem / W;
+                ww = hw_rem % W;
+            } else if (dst_d.ndims() > 3) { // 4D: NCHW
+                hh = sp / W;
+                ww = sp % W;
+            } else { // 3D or lower: treat as single spatial dim
+                ww = sp;
+            }
+
+            // Starting channel index of the current block
+            const dim_t c_start = c * block;
+
+            // Helper to fill logical positions per md with broadcasting
+            auto fill_pos = [&](const memory_desc_wrapper &md, dims_t &pos) {
+                for (int k = 0; k < md.ndims(); ++k)
+                    pos[k] = 0;
+                const int nd = md.ndims();
+                // N
+                pos[0] = (md.dims()[0] == 1) ? 0 : n;
+                // C (logical channel index)
+                pos[1] = (md.dims()[1] == 1) ? 0 : c_start;
+                if (nd > 4) {
+                    // D, H, W
+                    const dim_t Dm = md.dims()[2];
+                    const dim_t Hm = md.dims()[nd - 2];
+                    const dim_t Wm = md.dims()[nd - 1];
+                    pos[2] = (Dm == 1) ? 0 : dd;
+                    pos[nd - 2] = (Hm == 1) ? 0 : hh;
+                    pos[nd - 1] = (Wm == 1) ? 0 : ww;
+                } else if (nd > 3) {
+                    // H, W
+                    const dim_t Hm = md.dims()[nd - 2];
+                    const dim_t Wm = md.dims()[nd - 1];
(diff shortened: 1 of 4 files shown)


zhangfeiv0:
Out of curiosity, why is the auto-vectorized version slower than the scalar one?

zhangjian29:
> Out of curiosity, why is the auto-vectorized version slower than the scalar one?

Yes, it is interesting to see how auto-vectorization of compiler can fail. Based on our observation, the primary reason is:

- Creating test cases takes much more time than executing them, indicating that setting, loading and storing vectors too frequently makes the performance worse.

Deeper causes may include the compiler’s poor handling of vector tails, channel tails, redundant data type conversions, and similar issues. **When using intrinsics, we explicitly managed these these issues, which led to improved performance.**

Evidence supporting this can be found in Table 5.7 in page 59 of [*A comparison of auto-vectorization performance between GCC and LLVM for the RISC-V vector extension*](https://www.diva-portal.org/smash/get/diva2:1905944/FULLTEXT01.pdf).

In our experiments, detailed settings are:

1. Salar compilation with gcc 14.2:
```
cmake .. \
    -DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++" \
    -DCMAKE_SHARED_LINKER_FLAGS="-static-libstdc++" \
    -DDNNL_TARGET_ARCH="RV64" \
    -DARCH_SIMD_TEST_FLAGS="-march=rv64gc -mabi=lp64d" \
    -DDNNL_ARCH_OPT_FLAGS="-march=rv64gc -mabi=lp64d -O3" \
    -DCMAKE_CXX_FLAGS="-march=rv64gc -mabi=lp64d -O3" \
    -DCMAKE_C_FLAGS="-march=rv64gc -mabi=lp64d -O3" \
    -DCAN_COMPILE_RVV_INTRINSICS=OFF
```
2. Auto-vectorization compilation with gcc 14.2:
```
cmake .. \
    -DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++" \
    -DCMAKE_SHARED_LINKER_FLAGS="-static-libstdc++" \
    -DDNNL_TARGET_ARCH="RV64" \
    -DARCH_SIMD_TEST_FLAGS="-march=rv64gcv_zvfh -mabi=lp64d" \
    -DDNNL_ARCH_OPT_FLAGS="-march=rv64gcv_zvfh -mabi=lp64d -O3 -ftree-vectorize -fopt-info-vec" \
    -DCMAKE_CXX_FLAGS="-march=rv64gcv_zvfh -mabi=lp64d -O3 -ftree-vectorize" \
    -DCMAKE_C_FLAGS="-march=rv64gcv_zvfh -mabi=lp64d -O3 -ftree-vectorize" \
```

And tests are run with cmd: `./benchdnn --binary --batch=test_binary_perf_f32`

```
# test_binary_perf_f32
--reset

--inplace=true,false
--ddt=f32 --sdt=f32:f32

--alg=ADD,MUL,MAX,MIN,DIV,SUB,GE,GT,LE,LT,EQ,NE,SELECT
--batch=option_set_all
--batch=option_set_src0_bcast

--alg=ADD 
## scales
--attr-post-ops=
--attr-scales=, \
              src:common:0.125, \
              src1:common:2, \
              src:common:0.25+src1:common:0.5
--batch=option_set_all
--batch=option_set_src0_bcast
```

zhangjian29:
Hi @dzarukin ,

Apologies for the interruption. I’m reaching out to kindly request your professional advice and review of this PR #3899 when you have a moment. Our team has plans for ongoing contributions to the project and would greatly value your feedback. 

We would also appreciate it if you could take a look at the related PRs #3898, #3915, and #3933 at your convenience.

Thank you very much for your time and guidance!

Best regards,

Zhang Jian

src/cpu/rv64/rvv_binary_kernels.hpp
```diff
@@ -0,0 +1,727 @@
+/******************************************************************************
```
mgouicem, line 1: Any reason to put all these methods in hpp, instead of putting them in cpp directly?

src/cpu/rv64/rvv_binary.cpp
```diff
@@ -0,0 +1,292 @@
...
+                    = pd()->is_ternary_op() && src2_d.dims()[1] == 1;
+
+            // Small stack buffers for typical block sizes; fall back to heap if needed.
+            data_t x_tmp_stack[64];
+            data_t y_tmp_stack[64];
+            int8_t c_tmp_stack[64];
+
+            std::unique_ptr<data_t[]> x_tmp_heap;
+            std::unique_ptr<data_t[]> y_tmp_heap;
+            std::unique_ptr<int8_t[]> c_tmp_heap;
```
mgouicem: oneDNN typically does not allocate memory during execution.
For that, oneDNN exposes a mechanism called [scratchpad](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_scratchpad.html), which allows user to schedule memory allocations.
 
The internal mechanism to handle this is described [here](https://github.com/uxlfoundation/oneDNN/blob/382986ec76ca0cb199614ff470c8b52923347f63/src/common/memory_tracking.hpp#L34).

More generally, is it really beneficial to do the extra copy here? If not, you could prevent dispatching of this implementation if one of the inputs has broadcast semantic.

src/cpu/rv64/rvv_binary.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include "cpu/cpu_binary_pd.hpp"
+#include "cpu/platform.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace rv64 {
+
+template <impl::data_type_t date_type>
+struct rvv_binary_t : public primitive_t {
```
mgouicem, line 33: I would encourage to not use templatization here:
- it seems you are passing dt as an argument after execute function anyway,
- it would make the library size grow

src/cpu/rv64/rvv_binary_kernels.hpp
zhangjian29: We would like to seperate the primitive executation (in `rvv_binary.cpp`) and its intrinsic implmentation (in `kernel.hpp`) into two locations for clarity and better scalability. All the intrinsic methods are short and declared as `inline`, we believe the compiler can optimize them effectively. Is it necessary to consolidate all of these into `rvv_binary.cpp`?

src/cpu/rv64/rvv_binary.cpp
zhangjian29: Thank you for pointing that out. We have removed the dispatching logic and implementation for cases where one of the inputs has a broadcast semantic. Currently, only dense memory layout (`src0` and `src1` must have the same shape to avoid broadcasting) would be dispatched to this impl. Broadcast feature will be added soon.

src/cpu/rv64/rvv_binary.hpp
zhangjian29: I appreciate your professional advice. We have removed the templatization and now obtain the `data_type` within the struct. I have also reviewed all other similar sections in my other PRs to apply this correction.

zhangjian29:
Hi @mgouicem ,

Thanks for your time and thorough review!  Your professional feedback has been incredibly valuable, and I’ve learned a great deal from it. I’ve now implemented all of your suggestions, as detailed below:

## Key Changes

1. Removed bad memory allocation for broadcasting. The current implementation now only supports dense memory layouts (requiring `src0` and `src1` to have identical shapes). Broadcasting feature will be added in a future update.
2. Removed all the templatization and obtained the `data_type` within the struct. This reduces the library size and ensures consistency with other implementations.
3. Corrected `#ifdef DNNL_RISCV_USE_RVV_INTRINSICS` and dispatch logic in `cpu_binary_list.cpp`.

## Evaluation

I have rebuilt and retested using the RISC-V GNU toolchain version 14.2, verifying the functionality under the QEMU RISCV64 emulator. All CI tests passed successfully, with proper dispatching confirmed. Calls to the implemented binary can be traced by searching for `RISCV64GCV` in

- [test_binary_ci_rvv.log](https://github.com/user-attachments/files/22328198/test_binary_ci_rvv.log)

Additionally, I’ve reviewed and applied these changes across my other PRs.

Once again, thank you for your guidance. It has been immensely helpful!

src/cpu/rv64/rvv_binary_kernels.hpp
zhangjian29: Another reason for putting these intrinsics in `rvv_binary_kernels.hpp` is to enable their reuse by `rvv_postops.hpp` implementation, which will be proposed once this PR is successfully merged. This approach eliminates redundant implementation efforts.

zhangjian29:
Hi @dzarukin @vpirogov ,

I was wondering if you could help take a look at this PR. Thanks a lot!

## Key Changes from Last Review

1. Removed bad memory allocation for broadcasting. The current implementation now only supports dense memory layouts.
2. Removed all the templatization and obtained the data_type within the struct.
3. Corrected `#ifdef DNNL_RISCV_USE_RVV_INTRINSICS` and dispatch logic in `cpu_binary_list.cpp`.
4. Add `#if defined(DNNL_RISCV_USE_ZVFH_INTRINSICS)` guard to f16-related codes.

## Evaluation

All tests passed. Calls to the implemented binary can be traced by searching for `RISCV64GCV` in

- [test_binary_ci_rvv.log](https://github.com/user-attachments/files/22400661/test_binary_ci_rvv.log)

src/cpu/rv64/rvv_binary.cpp
```diff
@@ -0,0 +1,125 @@
...
+            rvv_binary_apply_s8(alg, static_cast<const int8_t *>(x),
+                    static_cast<const int8_t *>(y), static_cast<int8_t *>(dst),
+                    c, len);
+            break;
+        case data_type::u8:
+            rvv_binary_apply_u8(alg, static_cast<const uint8_t *>(x),
+                    static_cast<const uint8_t *>(y),
+                    static_cast<uint8_t *>(dst), c, len);
+            break;
+        default: assert(!"Unsupported data type for RVV binary");
```
dzarukin, line 53: This shouldn't be needed as typed pointers can be replaced with universal `void *` and at the spot where the pointer arithmetic should happen, it can be updated like this:
```
    auto ptr_for_a_task = static_cast<char *>(ptr_base) + work_amount * types::data_type_size(dt);
```
`dt` is available from a correspondent memory descriptor associated with a pointer. Doing it this way saves on maintainability.

src/cpu/platform.cpp
```diff
@@ -157,6 +157,8 @@ bool has_training_support(data_type_t data_type) {
 #endif
 #elif defined(DNNL_AARCH64_USE_ACL)
             return arm_compute::CPUInfo::get().has_bf16();
+#elif defined(DNNL_RISCV_USE_ZVFH_INTRINSICS)
+            return true;
```
dzarukin: It seems this change was meant to target `has_data_type_support` function instead. And it should declare support for f16 and bf16 based on the code logic.

src/cpu/rv64/rvv_binary.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+        status_t init(engine_t *engine) {
+            UNUSED(engine);
+            const data_type_t d_type = dst_md()->data_type;
+
+            VDISPATCH_BINARY(utils::everyone_is(d_type, src_md(0)->data_type,
+                                     src_md(1)->data_type)
+                            && platform::has_data_type_support(
+                                    src_md(0)->data_type),
+                    VERBOSE_UNSUPPORTED_DT);
```
dzarukin, line 46: Left a comment in a platform file. This should return unimplemented for bf16 and f16 as of now.

src/cpu/rv64/rvv_binary.cpp
zhangjian29: Thank you for pointing this out. The casting operations were unnecessary in that location. I have moved them to the sections where the pointers are directly involved in computations, as shown in the latest commit.

src/cpu/platform.cpp
zhangjian29: You are absolutely right. I have deleted those two lines and restored the `platform.cpp` file.

src/cpu/rv64/rvv_binary.hpp
zhangjian29: Yes, that's right. The `platform.cpp` file have been revised for this purpose.

zhangjian29:
Hi @dzarukin ,

Thank you for your time. I have implemented the changes based on your feedback, including the following:

## Key Changes

1. Moved the pointer casting logic in `rvv_binary.cpp` into the kernel, where the pointers are used for computations, to improve maintainability.
2. Removed the RVV `Zvfh` logic from `platform.cpp`. Support for the `f16` data type remains unimplemented at this time.

I have rerun the tests successfully to verify correctness. Looking forward to your further guidance. Thank you very much!

src/cpu/rv64/rvv_binary_kernels.hpp
```diff
@@ -0,0 +1,779 @@
...
+    }
+    auto eval = get_eval_u8(alg);
+    if (!eval) {
+        assert(!"[rvv_binary_apply_u8] unknown binary alg_kind");
+        return;
+    }
+    rvv_binary_kernel_u8(x, y, dst, c, len, eval, dt);
+}
+
+/* F16 integration with Zvfh extension */
```
mgouicem: Do I get it right that all the f16 code below is now deadcode?

Is there any plan to enable it? From Dima's early comment, the suggestion was only to:
- move the check for f16 support from `platform::has_training_support` to `platform::has_data_type_support`. 
- maybe add a runtime check for Zvfh extension availability if not all RISC-V CPU support it.

Is there any hurdle with the above?

zhangjian29: Hi, thanks for your detailed review. It is true that all the f16 code is deadcode now. According to what dzarukin suggested, `platform::has_data_type_support` should return unimplemented for bf16 and f16 as of now. I would delete the deadcode as for now and will add `f16` feature after this PR is merged in the future.

The way to enable `f16` support you mentioned is totally right. However, I think maybe we should pull another request to add runtime check for Zvfh extension. I want to focus on binary feature for this PR.

Thank you for your time!

zhangjian29:
Hi @mgouicem ,

> Hi, thanks for your detailed review. It is true that all the f16 code is deadcode now. According to what dzarukin suggested, platform::has_data_type_support should return unimplemented for bf16 and f16 as of now. I would delete the deadcode as for now and will add f16 feature after this PR is merged in the future.
> The way to enable f16 support you mentioned is totally right. However, I think maybe we should pull another request to add runtime check for Zvfh extension. I want to focus on binary feature for this PR.
> Thank you for your time!

As mentioned before, I have deleted f16-related deadcode. Rerun the tests with cmd:

```
ONEDNN_VERBOSE=1 ./benchdnn --binary --batch=inputs/binary/test_binary_ci
```

- [test_binary_ci.log](https://github.com/user-attachments/files/22613523/test_binary_ci.log)

All tests passed or skipped successfully.

zhangjian29:
Hi @vpirogov,

Sorry to bother. This PR requires one more review to merge. Could you please take a look at it when you have time? 

Thank you so much.

vpirogov:
Thank you for the contribution, @zhangjian29!

PR #3901, opened 2025-09-08
tests: gtests: fix win failure in mem api test

mzhukova:
- (only) win debug build is currently unhappy when comparing two strings in `EXPECT_EQ` (message thrown, that is `char *`, vs message expected, that is `std::string`), according to [gtests documentation](https://google.github.io/googletest/reference/assertions.html#c-strings) string objects could be used explicitly, so that we get proper string comparison vs pointer comparison in `EXPECT_EQ`
- locally PASSED + [relevant portion of weekly run](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f08d0f7e-83d5-f13c-8dcb-d4f5ef20c6a0)


mzhukova:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all

PR #3902, opened 2025-09-08
[GPU] Add gemm support for gs16 weights decompression.

dyoussif:
closes [MFDNN-13798](https://jira.devtools.intel.com/browse/MFDNN-13798)

### DIFF
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -157,14 +157,14 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
 
             // Due to hardware specifics, groups, when more than 1, should be
-            // multiple of 32.
+            // multiple of 16.
             VCHECK_MATMUL_UNIMPL(
                     IMPLICATION(wei_scale_group_k > 1 && wei_scale_group_k < K,
-                            wei_scale_group_k % 32 == 0),
+                            wei_scale_group_k % 16 == 0),
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
             VCHECK_MATMUL_UNIMPL(
                     IMPLICATION(wei_scale_group_n > 1 && wei_scale_group_n < N,
-                            wei_scale_group_n % 32 == 0),
+                            wei_scale_group_n % 16 == 0),
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
         }
 
@@ -240,14 +240,14 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     VERBOSE_UNSUPPORTED_ZP_CFG);
 
             // Due to hardware specifics, groups, when more than 1, should be
-            // multiple of 32.
+            // multiple of 16.
             VCHECK_MATMUL_UNIMPL(IMPLICATION(wei_zero_point_group_k > 1
                                                  && wei_zero_point_group_k < K,
-                                         wei_zero_point_group_k % 32 == 0),
+                                         wei_zero_point_group_k % 16 == 0),
                     VERBOSE_UNSUPPORTED_ZP_CFG);
             VCHECK_MATMUL_UNIMPL(IMPLICATION(wei_zero_point_group_n > 1
                                                  && wei_zero_point_group_n < N,
-                                         wei_zero_point_group_n % 32 == 0),
+                                         wei_zero_point_group_n % 16 == 0),
                     VERBOSE_UNSUPPORTED_ZP_CFG);
 
             if (utils::one_of(zp.get_data_type(DNNL_ARG_WEIGHTS), data_type::s4,
--- a/src/gpu/intel/gemm/jit.hpp
+++ b/src/gpu/intel/gemm/jit.hpp
@@ -358,38 +358,99 @@ struct gen_t : public primitive_t {
                     bsc_dims_, bo_dims_, bg_dims_, b_q2d_group_k(),
                     b_q2d_group_n(), has_gs(DNNL_ARG_B)};
 
-            VDISPATCH_GEMM_SC(
-                    kernel_desc_.select_kernel(arch_, stepping,
-                            dev_info_->eu_count(), has_systolic, is_integrated,
-                            mode, batch_dims(), eff_transa(), eff_transb(),
-                            eff_trans_bias(), swap_ab(), a_quant, b_quant,
-                            with_sround_, with_c_zero_points(), with_bias(),
-                            eff_sum_ab(), alpha(), beta(), eff_a_type(),
-                            eff_b_type(), desc()->c_type(), co_type, acc_type,
-                            eff_align_a(), eff_align_b(), align_c(), eff_m(),
-                            eff_n(), d->k(), eff_lda(), eff_ldb(), d->ldc(),
-                            d->batch(), std::move(gpu_post_ops)),
-                    VERBOSE_UNSUPPORTED_FEATURE,
-                    "matching kernel not found in catalog");
+            bool print_verbose = get_verbose(verbose_t::debuginfo) >= 5;
+            bool kernel_success = false;
+            auto entries = kernel_desc_.select_kernel(arch_, stepping,
+                    dev_info_->eu_count(), has_systolic, is_integrated, mode,
+                    batch_dims(), eff_transa(), eff_transb(), eff_trans_bias(),
+                    swap_ab(), a_quant, b_quant, with_sround_,
+                    with_c_zero_points(), with_bias(), eff_sum_ab(), alpha(),
+                    beta(), eff_a_type(), eff_b_type(), desc()->c_type(),
+                    co_type, acc_type, eff_align_a(), eff_align_b(), align_c(),
+                    eff_m(), eff_n(), d->k(), eff_lda(), eff_ldb(), d->ldc(),
+                    d->batch(), std::move(gpu_post_ops));
+            for (auto &entry : entries) {
+                kernel_desc_.set_entry(entry);
+                auto status = kernel_desc_.finalize();
+                // select_kernel can return a strategy that failed in the finalize call
+                bool valid = status == status::success;
+                if (!valid && print_verbose)
+                    dnnl::impl::verbose_printf(
+                            "info,gpu,gemm,skipping:%s,Strategy finalization "
+                            "failed.\n",
+                            kernel_desc_.entry().str().c_str());
+                // Global k-parallel kernels don't support post-ops or non-f32/s32
+                //   accumulation unless fusion is enabled.
+                if (kernel_desc_.driver_info()->kParallel()
+                        && !kernel_desc_.driver_info()->fusedPostOps()) {
+                    bool po_valid = !non_scale_po_
+                            && !(with_sum_ && with_c_scales())
+                            && utils::one_of(d->c_type(), f32, s32);
+                    if (!po_valid && print_verbose)
+                        dnnl::impl::verbose_printf(
+                                "info,gpu,gemm,skipping:%s,Invalid post op.\n",
+                                kernel_desc_.entry().str().c_str());
+                    valid &= po_valid;
+                }
+                // Limited post-op support for low-precision accumulation.
+                if (kernel_desc_.problem()->Tc.size() < 4) {
+                    valid &= !need_x32_acc;
+                    if (need_x32_acc && print_verbose)
+                        dnnl::impl::verbose_printf(
+                                "info,gpu,gemm,skipping:%s,Invalid post op.\n",
+                                kernel_desc_.entry().str().c_str());
+                }
+                // Ensure kernel can be run deterministically if required.
+                if (attr()->deterministic_) {
+                    bool deterministic
+                            = !kernel_desc_.driver_info()->nondeterministic();
+                    valid &= deterministic;
+                    if (!deterministic && print_verbose)
+                        dnnl::impl::verbose_printf(
+                                "info,gpu,gemm,skipping:%s,Non deterministic "
+                                "kernel.\n",
+                                kernel_desc_.entry().str().c_str());
+                }
 
-            // Global k-parallel kernels don't support post-ops or non-f32/s32
-            //   accumulation unless fusion is enabled.
-            auto info_ = kernel_desc_.driver_info();
-            if (info_->kParallel() && !info_->fusedPostOps()) {
-                VDISPATCH_GEMM(!non_scale_po_ && !(with_sum_ && with_c_scales())
-                                && utils::one_of(d->c_type(), f32, s32),
-                        VERBOSE_UNSUPPORTED_POSTOP);
+                if (valid) {
+                    auto try_create = [&]() {
+                        std::vector<compute::kernel_t> kernel_(1);
+                        auto *intel_engine
+                                = utils::downcast<intel::engine_t *>(engine);
+                        auto key = std::make_shared<
+                                trivial_key_container_t<dnnl::impl::gpu::intel::
+                                                gemm::jit::gen_nocopy_desc_t>>(
+                                kernel_desc_, intel_engine->engine_id());
+                        cache_state_t kernel_cache_status;
+                        auto kernel_name = "gemm_kernel";
+                        auto verbose
+                                = get_verbose(verbose_t::create_profile) >= 1;
+                        double start_ms = 0;
+                        if (verbose) start_ms = get_msec();
+                        status = get_cached_kernels<typename trivial_key_t<
+                                dnnl::impl::gpu::intel::gemm::jit::
+                                        gen_nocopy_desc_t>::value_type>(
+                                std::move(key), intel_engine, kernel_,
+                                {kernel_name}, kernel_cache_status);
+                        if (verbose && status == status::success) {
+                            double duration_ms = get_msec() - start_ms;
+                            const char *str
+                                    = cache_state2str(kernel_cache_status);
+                            VPROF(start_ms, primitive, create, str,
+                                    info(engine), duration_ms);
+                        }
+                        return status;
+                    };
+                    status = try_create();
+                    if (status == status::success) {
+                        kernel_success = true;
+                        break;
+                    }
+                }
             }
 
-            // Limited post-op support for low-precision accumulation.
-            if (kernel_desc_.problem()->Tc.size() < 4) {
-                VDISPATCH_GEMM(!need_x32_acc, VERBOSE_UNSUPPORTED_POSTOP);
-            }
-
-            // Ensure kernel can be run deterministically if required.
-            if (attr()->deterministic_)
-                VDISPATCH_GEMM(
-                        !info_->nondeterministic(), VERBOSE_DETERMINISTIC_FAIL);
+            VDISPATCH_GEMM(
+                    kernel_success, "matching kernel not found in catalog");
 
             init_scratchpad();
 
@@ -588,7 +649,6 @@ struct gen_t : public primitive_t {
 
     status_t init_nocopy(impl::engine_t *engine) {
         using namespace data_type;
-
         auto kd = pd()->kernel_desc();
 
         CHECK(create_kernel(engine, nocopy_kernel_, "gemm_kernel", *kd));
--- a/src/gpu/intel/gemm/jit/gen_kernel.cpp
+++ b/src/gpu/intel/gemm/jit/gen_kernel.cpp
@@ -24,6 +24,7 @@
 #include "gemmstone/strategy_parser.hpp"
 #include "gpu/intel/compute/device_info.hpp"
 #include "gpu/intel/gemm/jit/gen_kernel_db.hpp"
+#include "gpu/intel/gemm/jit/generator/pieces/compute_utils.hpp"
 #include "gpu/intel/gemm/jit/generator_dsl/builder.hpp"
 #include "gpu/intel/gemm/jit/generator_dsl/kernel_desc.hpp"
 #include "gpu/intel/jit/codegen/kernel.hpp"
@@ -410,15 +411,16 @@ status_t gen_desc_t::transfer_post_ops(
     return status::success;
 }
 
-status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
-        int stepping, int eu_count, bool has_systolic, bool is_integrated,
-        compute_mode mode, int batch_dims, bool trans_a, bool trans_b,
-        bool trans_co, bool swap_ab, const quant_params &a_quant,
-        const quant_params &b_quant, bool dst_sround, bool c_offset, bool bias,
-        sum_ab_t reduce_ab, float alpha, float beta, data_type_t a_type,
-        data_type_t b_type, data_type_t c_type, data_type_t co_type,
-        data_type_t acc_type, int align_a, int align_b, int align_c, dim_t m,
-        dim_t n, dim_t k, dim_t lda, dim_t ldb, dim_t ldc, dim_t batch,
+std::vector<const gemmstone::kcatalog::Entry *>
+gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
+        int eu_count, bool has_systolic, bool is_integrated, compute_mode mode,
+        int batch_dims, bool trans_a, bool trans_b, bool trans_co, bool swap_ab,
+        const quant_params &a_quant, const quant_params &b_quant,
+        bool dst_sround, bool c_offset, bool bias, sum_ab_t reduce_ab,
+        float alpha, float beta, data_type_t a_type, data_type_t b_type,
+        data_type_t c_type, data_type_t co_type, data_type_t acc_type,
+        int align_a, int align_b, int align_c, dim_t m, dim_t n, dim_t k,
+        dim_t lda, dim_t ldb, dim_t ldc, dim_t batch,
         gpu_post_ops_t &&post_ops) {
     using namespace ngen;
     using namespace kcatalog;
@@ -543,7 +545,8 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
     if (beta == 0.0f || beta == 1.0f) problem_.beta = beta;
 
     auto status = transfer_post_ops(std::move(post_ops), swap_ab);
-    if (status != status::success) return status;
+    if (status != status::success)
+        return std::vector<const gemmstone::kcatalog::Entry *>();
 
     if (c_offset || bias || reduce_ab != sum_ab::sum_none) {
         assert(!(c_offset && bias));
@@ -714,23 +717,24 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
         match_params.back().selector.precisions[2] = "I";
     }
 
-    EvaluateParams eval_params;
-
-    eval_params.sizes = base.sizes;
-    eval_params.alpha = alpha;
-    eval_params.beta = beta;
-    eval_params.postOps = !problem_.postOps.empty();
-    eval_params.cConvert = (acc_type != c_type);
-    eval_params.euCount = eu_count;
-    eval_params.batch = (batch_dims > 0);
-    eval_params.deterministic = (mode & mode_deterministic);
+    eval_params_.sizes = base.sizes;
+    eval_params_.alpha = alpha;
+    eval_params_.beta = beta;
+    eval_params_.postOps = !problem_.postOps.empty();
+    eval_params_.cConvert = (acc_type != c_type);
+    eval_params_.euCount = eu_count;
+    eval_params_.batch = (batch_dims > 0);
+    eval_params_.deterministic = (mode & mode_deterministic);
 
     SelectionObserver observer = entryObserver;
-    entry_ = select(catalog(), static_cast<int>(match_params.size()),
-            match_params.data(), eval_params, aux_params_, &observer);
+    tags_ = match_params[0].tags;
+    Ts_ = problem_.Ts;
+    beta_ = problem_.beta;
+    return select(catalog(), static_cast<int>(match_params.size()),
+            match_params.data(), eval_params_, aux_params_, &observer);
+}
 
-    if (!entry_) return status::unimplemented;
-
+status_t gen_nocopy_desc_t::finalize() {
     // Update A/B/C types from entry.
     Type Ta_new, Ta_ext_new, Tb_new, Tb_ext_new, Tc_new;
     parsePrecisions(entry_->selector.precisions[0], Ta_ext_new, Ta_new);
@@ -759,13 +763,16 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
     };
     use_tf32(problem_.Ta_ext, Ta_ext_new);
     use_tf32(problem_.Tb_ext, Tb_ext_new);
+    problem_.Ts = Ts_;
 
     if (problem_.Ts == Type::invalid) problem_.Ts = problem_.Tc;
 
     auto block_k = entry_->driverInfo.blocking[LoopK];
-    if (block_k > 0 && k > block_k && beta != 1.0f) problem_.beta = Scalar();
-
-    return finalize(match_params[0].tags);
+    problem_.beta = beta_;
+    if (block_k > 0 && k_ > block_k && eval_params_.beta != 1.0f)
+        problem_.beta = Scalar();
+    evaluate(*entry_, eval_params_, aux_params_);
+    return gen_desc_t::finalize(tags_.c_str());
 }
 
 status_t gen_xe_systolic_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
@@ -887,11 +894,12 @@ status_t gen_xe_systolic_kernel_desc_t::select_kernel(compute::gpu_arch_t arch,
     eval_params.batch = (batch_dims > 0);
 
     SelectionObserver observer = entryObserver;
-    entry_ = select(
+
+    auto entries = select(
             catalog(), match_params, eval_params, aux_params_, &observer);
 
-    if (!entry_) return status::unimplemented;
-
+    if (entries.size() < 1) return status::unimplemented;
+    entry_ = entries[0];
     return finalize(match_params.tags);
 }
 
--- a/src/gpu/intel/gemm/jit/gen_kernel.cpp
+++ b/src/gpu/intel/gemm/jit/gen_kernel.cpp
@@ -305,6 +305,16 @@ status_t gen_desc_t::finalize(const char *tags) {
     if (problem_.bOffset2D() || problem_.bScale2D())
         if (problem_.bqGroupK % strategy_.bqGroupKGranularity())
             return status::unimplemented;
+    if (problem_.aScale2D()
+            && problem_.aqGroupK
+                            % minOuterProductCount(hw_, problem_, strategy_)
+                    != 0)
+        return status::unimplemented;
+    if (problem_.bScale2D()
+            && problem_.bqGroupK
+                            % minOuterProductCount(hw_, problem_, strategy_)
+                    != 0)
+        return status::unimplemented;
 
     // If the M/N group size is equal to M or N, align up to a multiple of unroll size
     // XXX: Increase group size to a large value before aligning to increase reusability
--- a/src/gpu/intel/gemm/jit/gen_kernel.hpp
+++ b/src/gpu/intel/gemm/jit/gen_kernel.hpp
@@ -83,10 +83,14 @@ struct gen_desc_t {
     }
     compute::gpu_arch_t arch() const { return arch_; }
 
+    bool has_entry() { return entry_ != nullptr; }
+
     const gemmstone::kcatalog::Entry &entry() const {
         assert(entry_ != nullptr);
         return *entry_;
     };
+
+    void set_entry(const gemmstone::kcatalog::Entry *entry) { entry_ = entry; }
 
 protected:
     compute::gpu_arch_t arch_;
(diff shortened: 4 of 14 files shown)


dyoussif:
make test perf-gpu
set primitive=matmul ip

dyoussif:
make test perf-gpu
set primitive=matmul ip

src/common/matmul.cpp
```diff
@@ -155,11 +155,11 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
             // multiple of 32.
             VCHECK_MATMUL_UNIMPL(
                     IMPLICATION(wei_scale_group_k > 1 && wei_scale_group_k < K,
-                            wei_scale_group_k % 32 == 0),
+                            wei_scale_group_k % 16 == 0),
```
dzarukin, line 163: Please update comments along the way.

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
-            }
-
-            // Limited post-op support for low-precision accumulation.
-            if (kernel_desc_.problem()->Tc.size() < 4) {
-                VDISPATCH_GEMM(!need_x32_acc, VERBOSE_UNSUPPORTED_POSTOP);
+            bool valid = false;
+            status_t status;
+            int strategy_count = 1;
+            for (int i = 0; i < strategy_count; i++) {
+                auto po = gpu_post_ops;
```
dzarukin: Maybe `const auto &`? And maybe out of the `for` loop?

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
-                        VERBOSE_UNSUPPORTED_POSTOP);
-            }
-
-            // Limited post-op support for low-precision accumulation.
-            if (kernel_desc_.problem()->Tc.size() < 4) {
-                VDISPATCH_GEMM(!need_x32_acc, VERBOSE_UNSUPPORTED_POSTOP);
+            bool valid = false;
+            status_t status;
+            int strategy_count = 1;
+            for (int i = 0; i < strategy_count; i++) {
```
dzarukin: Making a counter based on a loop variable given that the status for selecting a kernel might not be successful seems like adding more debug work while the code might be simplified.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -104,9 +103,18 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
...
+        while (a % b != 0) {
+            auto tmp = b;
+            b = a % b;
+            a = tmp;
+        }
+        return b;
+    };
+
+    auto smallest_k_group = gcd(gcd(src_scale_group, wei_scale_group),
+            gcd(src_zp_group, wei_zp_group));
```
dzarukin: Why is `gcd` better than a plain `std::min`?

dyoussif: In the case where zp group size is not equal to scales group size, we would like to take the `gcd`. eg. for sizes 32 and 48, `smallest_k_group` should be 16 not 32 to calculate a correct result.

src/gpu/intel/gemm/jit.hpp
dyoussif: `select_kernel` expects an rvalue for `post_ops_t`.

src/gpu/intel/gemm/jit.hpp
dyoussif: The goal here is to try generating the kernel and if it fails to try another strategy. This is to avoid potential out of registers issues and was discussed in the tracker.

tests/benchdnn/matmul/ref_matmul.cpp
dzarukin: Alright, so it's a functional change...
Would you mind to create a function in benchdnn/utils/numeric.hpp:
```
template< class T >
T gcd(std::initializer_list<T> ilist) {
    ...
}
```
to match the `std::min` interface with the same functionality?

src/gpu/intel/gemm/jit.hpp
dzarukin: I'm fine with the goal, but it feels the contract in this for-loop can be more straightforward. At least, the updated logic may be documented.
I find it confusing that if status is unsuccessful, the loop counter still gets updated and further logic still applies. If the loop range was fixed at the beginning and try-fail approach implemented, that would help for readability.

src/gpu/intel/gemm/jit/generator/pieces/gemm.cxx
```diff
@@ -822,7 +822,7 @@ bool Generator<hw>::gemmBody(GEMMProblem problem, GEMMStrategy strategy, GEMMSta
         success = gemmBodyInternal(problem, strategy, state);
     } catch (...) {
         lastException = std::current_exception();
-    }
+     }
```
kealan-barbieri: leftover from earlier changes?

src/common/matmul.cpp
```diff
@@ -238,11 +238,11 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
             // multiple of 32.
             VCHECK_MATMUL_UNIMPL(IMPLICATION(wei_zero_point_group_k > 1
                                                  && wei_zero_point_group_k < K,
-                                         wei_zero_point_group_k % 32 == 0),
+                                         wei_zero_point_group_k % 16 == 0),
                     VERBOSE_UNSUPPORTED_ZP_CFG);
             VCHECK_MATMUL_UNIMPL(IMPLICATION(wei_zero_point_group_n > 1
                                                  && wei_zero_point_group_n < N,
-                                         wei_zero_point_group_n % 32 == 0),
+                                         wei_zero_point_group_n % 16 == 0),
```
kealan-barbieri, line 250: random place, we could use some testcases to cover gs16.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -177,7 +177,7 @@ const kcatalog::Entry *select1(const kcatalog::Catalog &catalog, int npatterns,
                 continue;
 
             double score = evaluate(*it, eparams, thisAux);
-
+            entries[score] = std::pair<const kcatalog::Entry *,EvaluateAuxOutput>( &*it, thisAux);
```
echeresh: I have a few observations and questions about this approach:

1. We may have collisions between scores (same score value for two entries). Pretty rare I guess but may complicate debugging
2. I see the comparison function also relies on one extra condition regarding fallback/alignment [here](
https://github.com/uxlfoundation/oneDNN/blob/77db184ea56135d584d9b0d07b8ed83f7510c964/src/gpu/intel/gemm/jit/selector/kernel_selector.cpp#L182). `std::map` wouldn't take that into account. Not a big deal probably but we may end up with a different ordering if I read code correctly
3. Condition [here](https://github.com/uxlfoundation/oneDNN/blob/a876168fb399b0573b04f3dd33a046ba8fd9bb7c/src/gpu/intel/gemm/jit/selector/kernel_selector.cpp#L201) has an extra check of `bestEntry`. How do we ensure this condition is met when entries are taken from the map?

As for the approach, I'm wondering what if we can change it to rely on the index instead vs storing entries and skipping the conditions above? Looks like it could be more straightforward:

- We try to select the first best kernel (pass index 0). If success - use it
- Try the second best kernel (pass index 1), etc
- ...
- Once we reach a null entry - it indicates no more strategies, meaning primitive creation is unsuccessful

This approach might be better as it naturally avoids issues 1 and 2 above. Additionally, we don't need to introduce `entries_` and a related iterator and can keep all the logic inside `select()`. There may be some overhead related to sorting in `select1()` but it should be relatively small - we are sorting a few dozens of numbers at worst.

@dyoussif @petercad What do you think?

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
+                }
+                if (status == status::success) break;
             }
-
-            // Ensure kernel can be run deterministically if required.
+            VDISPATCH_GEMM_SC(status, VERBOSE_UNSUPPORTED_FEATURE,
+                    "matching kernel not found in catalog");
             if (attr()->deterministic_)
                 VDISPATCH_GEMM(!kernel_desc_.driver_info()->nondeterministic(),
                         VERBOSE_DETERMINISTIC_FAIL);
```
Simonsays095: This should be inside the fallback loop as well.

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
+            VDISPATCH_GEMM(valid, VERBOSE_UNSUPPORTED_POSTOP);
```
Simonsays095: However the fallback loop is set up, once we create a kernel and exit the loop with `status = success`, we should be confident that the kernel will work. Why does the `valid` variable leak out of the loop like this, instead of being combined with status somehow within the loop?

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
+                    valid &= !kernel_desc_.driver_info()->nondeterministic();
+
+                compute::kernel_t kernel;
+                if (valid) {
+                    auto *intel_engine
+                            = utils::downcast<intel::engine_t *>(engine);
+                    dnnl::impl::gpu::intel::gemm::jit::gen_kernel_t kd(
+                            kernel_desc_);
+                    status = intel_engine->create_kernel(&kernel, &kd);
+                }
```
Simonsays095, line 449: If I understand this correctly, this will compile the kernel just to check if it can compile successfully, and then it discards the resulting kernel. This means we're compiling 2 kernels every time in the best case scenario (no fallbacks) and our creation time will double (excluding kernel cache use). Is it possible to use this result as our final kernel if successful?

src/gpu/intel/gemm/jit/include/gemmstone/kernel_selector.hpp
```diff
@@ -93,8 +93,8 @@ struct MatchParams : public MatchParamsBase
 
 using SelectionObserver = std::function<void (const kcatalog::Entry *entry, double score, EvaluateAuxOutput aux)>;
 
-const kcatalog::Entry *select(const kcatalog::Catalog &catalog, const MatchParams &pattern, const EvaluateParams &eparams, EvaluateAuxOutput &aux, SelectionObserver *observer = nullptr);
-const kcatalog::Entry *select(const kcatalog::Catalog &catalog, int npatterns, const MatchParams *patterns, const EvaluateParams &eparams, EvaluateAuxOutput &aux, SelectionObserver *observer = nullptr);
+const kcatalog::Entry *select(const kcatalog::Catalog &catalog, const MatchParams &pattern, const EvaluateParams &eparams, EvaluateAuxOutput &aux, std::map<double, std::pair<const kcatalog::Entry *, EvaluateAuxOutput>> &entries,  SelectionObserver *observer = nullptr);
+const kcatalog::Entry *select(const kcatalog::Catalog &catalog, int npatterns, const MatchParams *patterns, const EvaluateParams &eparams, EvaluateAuxOutput &aux, std::map<double, std::pair<const kcatalog::Entry *, EvaluateAuxOutput>> &entries,  SelectionObserver *observer = nullptr);
```
Simonsays095: There's an extra space.
```suggestion
const kcatalog::Entry *select(const kcatalog::Catalog &catalog, int npatterns, const MatchParams *patterns, const EvaluateParams &eparams, EvaluateAuxOutput &aux, std::map<double, std::pair<const kcatalog::Entry *, EvaluateAuxOutput>> &entries, SelectionObserver *observer = nullptr);
```

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
```diff
@@ -107,10 +107,10 @@ bool Generator<hw>::gemmMake2DQuantizationLayouts(bool isA, const GEMMProblem &p
         c = slmA ? state.ka_slm : cNoSLM;
         k = slmA ? strategy.unrollKSLM : cNoSLM;
         r = std::max(1, r / xqGroupMN);
-        c = state.kaq = std::max(1, c / xqGroupK);
-        state.kaqStride = std::max(1, k / xqGroupK);
+        c = state.kaq = std::max(1,c % xqGroupK == 0 ? c / xqGroupK : 1);
```
Simonsays095: Please update formatting of changes in this file - spaces around ternary operators are inconsistent, and there are a few double-spaces.

src/gpu/intel/gemm/jit.hpp
```diff
@@ -333,38 +333,55 @@ struct gen_t : public primitive_t {
...
+                if (status == status::success) break;
```
Simonsays095: Another thing: using fallback kernels could have performance implications - when this happens, we should log it so the user knows. My suggestion is to log each skipped kernel at verbosity level `debuginfo=1`.

src/common/matmul.cpp
dyoussif: added a couple test cases.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
dyoussif: I refactored the `select1` function so that the logic is more streamlined and takes an index argument. It should address points 1-3 above.

src/gpu/intel/gemm/jit.hpp
dyoussif: This check is at line 366

src/gpu/intel/gemm/jit.hpp
dyoussif: Thanks for pointing this out. The loop should now only exit if there are no more strategies or on successsful kernel creation.

src/gpu/intel/gemm/jit.hpp
dyoussif: Good idea! `kernel_t` is saved to the kernel descriptor upon successful creation and reused in  `create_generator`.

src/gpu/intel/gemm/jit.hpp
dyoussif: A fallback kernel should only be used in the case of kernel generation failure. It should not cause regressions for existing cases.

dyoussif:
make test perf-gpu
set primitive=matmul ip

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,68 +156,70 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
-        }
+    };
+    std::sort(entries.begin(), entries.end(), comp);
+    if (entries.size() > 0){
+    auto prevEntry = entries[0];
+    bool prevFallback = (prevEntry->restrictions.tags[0] == kcatalog::ReqAlignFallback);
+    int prevAlignA = std::max(prevEntry->restrictions.alignment[0], 4);
+    int prevAlignB = std::max(prevEntry->restrictions.alignment[1], 4);
+
+    for (int i=1; i< (int) entries.size(); i++){
```
echeresh: It would be nice to put this sorting logic under a compare function as well, e.g.:

```c++
auto less = [](Entry lhs, Entry rhs) {
    if (lessAligned(rhs, lhs) && isFallback(rhs)) return true;
    if (lessAligned(lhs, rhs) && isFallback(lhs)) return false;
    return score(lhs) < score(rhs);
};
```

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,68 +156,70 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
-                bestIsFallback = fallback;
-                aux = thisAux;
-            }
-
-            if (observer) (*observer)(&*it, score, aux);
-        }
+    };
+    std::sort(entries.begin(), entries.end(), comp);
+    if (entries.size() > 0){
+    auto prevEntry = entries[0];
```
echeresh: Please check formatting:
```suggestion
    if (entries.size() > 0) {
        auto prevEntry = entries[0];
```

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
echeresh: Thanks. This looks more straightforward.

One potential concern is sorting overhead (which I assume is not significant).
@petercad @Simonsays095 Any concerns on that?

src/gpu/intel/gemm/jit.hpp
Simonsays095: In that case, we no longer need this check outside the loop, right?

src/gpu/intel/gemm/jit.hpp
Simonsays095: Please remove this check (it's unnecessary due to the status check a few lines above) and move `valid` to be defined inside the loop. Also - we don't even need a status check after this loop. Instead just check for `kernel_desc.kernel_ == nullptr` to check if we have a valid kernel and just use the status inside the loop.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
Simonsays095: We currently have roughly <20 catalog entries for any given problem. Although it might become a problem in the future, I think it should be fine for now. Since this overhead occurs at initialization time, we are likely to spend much more time compiling the kernel than sorting the entries.

src/gpu/intel/gemm/jit.hpp
Simonsays095: I'm not worried about regressions currently, I'm thinking about the future. If a change is made that invalidates a performant kernel and it's skipped, or some feature is added and we're leaving performance on the table by skipping many kernels, things like that. It would just be handy to have it yell at us if there's an issue like this.

Even for this feature (gs16): if enabling it makes use skip all but the last kernel (which is possible, since these kernels were not designed with gs16 in mind), we could end up with a large performance gap between gs16 versus gs32. A verbose line like this would make that very obvious, without having to look at each selected kernel between the two runs and comparing the scores manually.

dyoussif:
make test perf-gpu
set primitive=matmul ip

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
dyoussif: done.

src/gpu/intel/gemm/jit.hpp
dyoussif: done.

src/gpu/intel/gemm/jit.hpp
dyoussif: I added some verbose prints in the loop to show why a particular implementation was skipped. In addition, upon failed kernel generation a verbose message should already be printed out from the try_catch stmts in the generator.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,68 +156,60 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
-            int alignA = std::max(it->restrictions.alignment[0], 4);
-            int alignB = std::max(it->restrictions.alignment[1], 4);
-
-            if (fallback && lessAligned(alignA, alignB, bestAlignA, bestAlignB))
-                continue;
-
-            double score = evaluate(*it, eparams, thisAux);
+             // Late tag checking. If late tags do not match, we skip entry.
+	    if (tagMatch(it->restrictions.tags, patterns[ipattern].lateTags))
+            entries.push_back(&*it);
```
Simonsays095: Needs another indent, and a tab is mixed in with the spaces.
```suggestion
            if (tagMatch(it->restrictions.tags, patterns[ipattern].lateTags))
                entries.push_back(&*it);
```

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,68 +156,60 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
-            }
+    };
+    std::sort(entries.begin(), entries.end(), less);
 
-            if (observer) (*observer)(&*it, score, aux);
-        }
+    if (entry_idx < (int) entries.size()){
+	    double score = evaluate(*entries[entry_idx], eparams, aux);
+            if (observer) (*observer)(entries[entry_idx], score, aux);
+	    return entries[entry_idx];
```
Simonsays095: I think it would be better to refactor this function to return the (sorted) vector of entries. Then we carry this vector back out to select_kernel (or even pd_t::init) and simply iterate through the vector, stopping at the first one that we deem to be valid.

For cases where we need a fallback, this would reduce the number of duplicate catalog searches, entry sorts, and performance model evaluations.

echeresh: @Simonsays095 So sounds like you would still consider repeated sorting being a performance concern :)

I think you are right that the vector has to be propagated up to `pd_t::init()` but e.g. `select_kernel()` includes some other code to initialize/adjust the problem structure and also some post-processing. Not sure how much work to split/handle these, probably @dyoussif can tell better.

Simonsays095: Of course, less sorting/searching is better. :) But I think sorting will be required for the fallback mechanism, and I'm not too worried about sorting 20 kernels once. Currently in this PR, we search the catalog, evaluate the performance of all selected kernels, sort the kernels, and try to compile once per fallback iteration.

By contrast, pulling the vector of entries out of this function means we could just do this once and the fallback overhead is just in the additional compilation attempt.

> some other code to initialize/adjust the problem structure and also some post-processing. Not sure how much work to split/handle these, probably @dyoussif can tell better.

In the PR currently, the index is only used to select a kernel from the final list. All of the other arguments are identical, so it should be easy to pull the index back out of this function (by returning the vector) with no side-effects. From there, we can do as we please to the vector. As for post-processing, the main machinery in the select function is when no kernels are found, to use type and hardware fallbacks. This all still applies when we return one entry versus the entire vector - I don't see any major stumbling blocks.

echeresh: @Simonsays095 Thanks, I think I was looking at `select_kernel()` logic here:

https://github.com/uxlfoundation/oneDNN/blob/ddcb9bebe093078a407dd3ba8d543717419d384e/src/gpu/intel/gemm/jit/gen_kernel.cpp#L714-L736

This function performs some problem adjustment based on the entry from what I see. Do you have suggestions how to decouple these?

dyoussif: It should be an edge case to fall back to another kernel, so having to call select kernel more than once should not happen often. Moreover, if we filter out strategies we basically have to  handle subsequent logic after `select` and call `finalize` for all candidate strategies in the default case. I would also anticipate the largest time would be spent in actually attempting to compile the kernel during the fallback. For simplicity, I would prefer to keep things as is and we can always revisit later.

Simonsays095: Any updates to `problem_` should be done only once we've selected a candidate kernel to compile. My suggestion is to move kernel override logic out of `finalize` into its own function, and move all problem modification like this into `finalize`. This function will then update the entry and problem as a final pass before compilation. Then the flow is:
```C++
std::vector<entry> entries = kernel_desc_.select_kernel(...);
for (auto &entry: entries) {
    // Check entry against attrs for validity, and log+continue if not
    ...
    // Once we'd like to try compilation
    kernel_desc_.entry_ = &entry;
    kernel_desc_.kernel_override(...); // New function for GEMM_KERNEL functionality
    kernel_desc_.finalize(...);.
    auto status = intel_engine->create_kernel(&kernel, &kernel_desc_);
}
VDISPATCH_GEMM(kernel_desc_.kernel != nullptr, 
                    VERBOSE_UNSUPPORTED_FEATURE,
                    "matching kernel not found in catalog");
```
If compilation fails, problem_ and entry_ may still have some variables updated. We can either create/modify temporary variables while compiling via `create_kernel` and copy into `kernel_desc_` if that's successful (similar to what's done in the PR now), or we can leave it dirty and overwrite those changes on the next fallback iteration (as in my pseudocode).

dyoussif:
make test perf-gpu
set primitive=matmul ip

dyoussif:
make test perf-gpu
set primitive=matmul ip

dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
dyoussif: I modified so that `select kernel` needs to be only called once.

src/gpu/intel/gemm/jit/generator/pieces/gemm_setup.cxx
```diff
@@ -1608,7 +1608,12 @@ bool Generator<hw>::gemmAccumulateCSetup(GEMMProblem &problem, GEMMStrategy &str
         bool lazyRepack = state.Ta_load.is4() && one_of(Ta, Type::f16, Type::bf16, Type::f32);    // Other cases are unimplemented
         if (lazyRepack)
             state.ka_repack = std::min(state.ka_repack, strategy.kb_load);
-        state.Ar_layout = RegisterLayout(hw, Ta, unrollM, state.ka_repack, state.A_layout.colMajor(), crosspackA, tileM_A, tileK_A, true, splitA);
+	int repackN = state.ka_repack;
```
echeresh: Please check tabs/spaces here and below.

src/gpu/intel/gemm/jit/generator/pieces/k_loop.cxx
```diff
@@ -954,8 +954,14 @@ void Generator<hw>::kLoop(KLoop type, const GEMMProblem &problem, GEMMStrategy &
             auto layoutCopy = layout;
             layoutCopy.unlinkFromMemory();
             sublayout = layoutCopy.slice(true, ha, ha + k_repack, false);
-            for (auto &l: Ar_sublayout)
-                l.offsetC += ha;
+	    if (Ar_sublayout.cols() > k_repack){
```
echeresh: Indentation doesn't look right.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,70 +156,56 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
-                bestScore = score;
-                bestIPattern = ipattern;
-                bestAlignA = alignA;
-                bestAlignB = alignB;
-                bestIsFallback = fallback;
-                aux = thisAux;
-            }
-
-            if (observer) (*observer)(&*it, score, aux);
+             // Late tag checking. If late tags do not match, we skip entry.
```
echeresh, line 165:
```suggestion
         // Late tag checking. If late tags do not match, we skip entry.
```

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -156,70 +156,56 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
+	              EvaluateAuxOutput thisAux;
+                      bool lhsFallback = (lhs->restrictions.tags[0] == kcatalog::ReqAlignFallback);
+                      int  lhsAlignA = std::max(lhs->restrictions.alignment[0], 4);
+                      int  lhsAlignB = std::max(lhs->restrictions.alignment[1], 4);
+                      bool rhsFallback = (rhs->restrictions.tags[0] == kcatalog::ReqAlignFallback);
+                      int  rhsAlignA = std::max(rhs->restrictions.alignment[0], 4);
+                      int  rhsAlignB = std::max(rhs->restrictions.alignment[1], 4);
+		      if (rhsFallback && lessAligned(rhsAlignA, rhsAlignB, lhsAlignA, lhsAlignB)) return true;
+		      if (lhsFallback && lessAligned(lhsAlignA, lhsAlignB, rhsAlignA, rhsAlignB)) return false;
+	              double lhs_score = evaluate(*lhs, eparams, thisAux);
```
echeresh: Please check formatting

tests/benchdnn/utils/numeric.hpp
```diff
@@ -36,5 +36,21 @@ bool is_integral_dt(dnnl_data_type_t dt);
...
+        if (b == 0) return a;
+        while (a % b != 0) {
+            auto tmp = b;
+            b = a % b;
+            a = tmp;
+        }
+        return b;
+    };
+    if (ilist.size() < 1) return 0;
+    T ret = *ilist.begin();
```
echeresh:
```suggestion
    T ret = 0;
```

src/gpu/intel/gemm/jit/gen_kernel.hpp
```diff
@@ -97,6 +107,7 @@ struct gen_desc_t {
     const gemmstone::kcatalog::Entry *entry_ = nullptr;
     gemmstone::EvaluateAuxOutput aux_params_;
     gemmstone::CommonDriverInfo driver_info_;
+    compute::kernel_t kernel_ = nullptr;
```
echeresh: Can we move it outside of the descriptor, directly to `gen_t::pd_t`?
Keeping the kernel object under a kernel descriptor is a little confusing.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -727,13 +743,15 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
...
     if (problem_.Ts == Type::invalid) problem_.Ts = problem_.Tc;
 
     auto block_k = entry_->driverInfo.blocking[LoopK];
-    if (block_k > 0 && k > block_k && beta != 1.0f) problem_.beta = Scalar();
-
-    return finalize(match_params[0].tags);
+    if (block_k > 0 && k_ > block_k && eval_params_.beta != 1.0f)
+        problem_.beta = Scalar();
+    evaluate(*entry_, eval_params_, aux_params_);
+    return gen_desc_t::finalize(const_cast<char *>(tags_.c_str()));
```
echeresh: Looks like `const_cast` is redundant here.

src/gpu/intel/gemm/jit/gen_kernel.hpp
```diff
@@ -83,11 +83,21 @@ struct gen_desc_t {
...
         assert(entry_ != nullptr);
         return *entry_;
     };
 
+    bool has_kernel() { return kernel_; }
+
+    void set_kernel(compute::kernel_t kernel) { kernel_ = std::move(kernel); }
+
+    void set_entry(const gemmstone::kcatalog::Entry *entry) {
+        entry_ = std::move(entry);
```
echeresh:
```suggestion
        entry_ = entry;
```

src/gpu/intel/gemm/jit/gen_kernel.hpp
echeresh: Default constructor is enough.

```suggestion
    compute::kernel_t kernel_;
```

src/common/matmul.cpp
```diff
@@ -152,14 +152,14 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
 
             // Due to hardware specifics, groups, when more than 1, should be
-            // multiple of 32.
+            // multiple of 16.
```
Simonsays095, line 160: (random spot) It looks like your first two commits were (accidentally?) merged. Since each one has a decent chance for bug introductions and they aren't related, could you please split them again?

src/gpu/intel/gemm/jit/gen_kernel.hpp
dyoussif: rewrote this to rely on the kernel cache.

tests/benchdnn/utils/numeric.hpp
dyoussif: `ret `is used below. don't think we want to set it to 0.

src/common/matmul.cpp
dyoussif: done.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -693,23 +706,23 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
...
-    eval_params.batch = (batch_dims > 0);
-    eval_params.deterministic = (mode & mode_deterministic);
+    eval_params_.sizes = base.sizes;
+    eval_params_.alpha = alpha;
+    eval_params_.beta = beta;
+    eval_params_.postOps = !problem_.postOps.empty();
+    eval_params_.cConvert = (acc_type != c_type);
+    eval_params_.euCount = eu_count;
+    eval_params_.batch = (batch_dims > 0);
+    eval_params_.deterministic = (mode & mode_deterministic);
```
Simonsays095, line 737: These changes can be reverted, to keep `eval_params` as a local  variable.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -693,23 +706,23 @@ status_t gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch,
...
 
     SelectionObserver observer = entryObserver;
-    entry_ = select(catalog(), static_cast<int>(match_params.size()),
-            match_params.data(), eval_params, aux_params_, &observer);
-
-    if (!entry_) return status::unimplemented;
+    tags_ = match_params[0].tags;
+    Ts_ = problem_.Ts;
```
Simonsays095, line 741: Do we still have to cache `Ts_` and `tags_` here, since `select_kernel` is only called once?

tests/benchdnn/utils/numeric.hpp
echeresh: It will work with `0` as well, see [here](https://ideone.com/F10lNO) as `gcd(0, a) = a`.

src/gpu/intel/gemm/jit/gen_kernel.cpp
dyoussif: This is used in the `finalize `call below.

src/gpu/intel/gemm/jit/gen_kernel.cpp
dyoussif: yes, we need these values for `finalize`.

src/gpu/intel/gemm/jit/gen_kernel.cpp
Simonsays095: I see - we have to call `evaluate` (requiring `eval_params_`) in `finalize` to reset `aux_params_` since it's "dirty" after calling `select`.

src/gpu/intel/gemm/jit/gen_kernel.cpp
Simonsays095: I see why we need `tags_` as `match_params` is on the stack and is destroyed at the end of the function. But `problem_.Ts` is available for use in `finalize` - not to mention the only thing we use `Ts_` for in `finalize` is to set `problem_.Ts` again.

dyoussif: `problem_.Ts`  can be overwritten in line 755,  so to avoid mixing values between iterations it is reset to the  original value.

Simonsays095: Got it, thanks for the explanation.

dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

dyoussif:
make test perf-gpu
set primitive=matmul ip

dyoussif:
make test perf-gpu
set primitive=matmul ip

byungilm:
Could I know a plan to merge?
(Is perf test on-going? or is it being fixed?)

PR #3904, opened 2025-09-08
cpu: convolutions: update the way scales are applied

dzarukin:
Continuation of #3701 and #3843 - updates remaining forward convolutions to handles scales inside the kernels instead of inside a macro.
Ref_matmul as a bonus.
This PR closes the scales rework for the scope of async runtime but doesn't close the topic completely as there're implementations left that still rely on the macro.

### DIFF
--- a/tests/benchdnn/inputs/matmul/harness_matmul_decompression
+++ b/tests/benchdnn/inputs/matmul/harness_matmul_decompression
@@ -89,8 +89,8 @@
 --skip-impl=ref
 --dt=bf16:s8:bf16,bf16:u8:bf16
 --wtag=any,abcd,abdc
---attr-scales=wei:per_tensor:bf16:128x1,wei:per_tensor:bf16:1x128
---attr-zero-points=wei:per_tensor:u8:128x1,wei:per_tensor:u8:1x128
+--attr-scales=wei:per_tensor:bf16:128x1
+--attr-zero-points=wei:per_tensor:u8:128x1
 --attr-fpmath=bf16:true
 2x3x5x512:2x3x512x1024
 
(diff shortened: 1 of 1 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
enable test_scope=NIGHTLY

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
```diff
@@ -1522,14 +1521,34 @@ void jit_avx512_core_amx_fwd_kernel_t::store_output_vector_int8(
...
     /* add bias and zero-point to zmm_accum */
     vcvtdq2ps(zmm_out, zmm_out);
+
     const Zmm zmm_out_msk = zmm_mask(zmm_out, mask_flag);
-    vmulps(zmm_out_msk, zmm_out,
-            EVEX_compress_addr(reg_ptr_scales, scale_offset));
+
+    if (jcp.with_src_scales) {
+        mov(reg_ptr_src_scales, ptr[param1 + GET_OFF(src_scales)]);
+        vmulps(zmm_out_msk, zmm_out,
```
ankalinin, line 1529: just a question: In the original version we read the scale pointer only once per kernel call and then use it. In the update we read the pointers to the scales for each output data vector. This may/should affect the performance, especially for cases with a large spatial/oc and small ic. Have you had a chance to check the performance?

ankalinin (approved):

mgouicem (approved):

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
mgouicem: Agree with Andrey's comment. Some perf data might be useful.
Another related question is: why not keep the scales product precomputation, but just do it in scratchpad or small heap allocation instead of stack? This should allow to not increase the number of mul instructions and access to scale inside kernel.

dzarukin: BRGeMM kernel data showed marginal gain by not using extra scratchpad memory (from precompute_scales) and running extra multiplications at least on SPR. I'll collect some data to make sure it's not regressing severely.

dzarukin: > Have you had a chance to check the performance?

I collected data on SPR from a single socket for amx and avx512_core_vnni implementations, the results are within 1-2% variation with some outliers that are not reproducible in a standalone run.

I'll put comments where applicable regarding managing registers in kernels and will fix perf issues in the future if anything comes (most likely from OV only).

dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv

PR #3907, opened 2025-09-09
cpu: x64: fix for test_internals segfault in debug build

spalicki:
`test_internals` on Windows ICX Debug builds (with at least 2 threads) crashes with Access Violation on write to `reg64_aux` in `jit_cvt_fp8_t::generate()`. Assigning it `abi_param3` instead of `abi_not_param1` fixes the issue.

Fixes: [MFDNN-13190](https://jira.devtools.intel.com/browse/MFDNN-13190).

### DIFF
--- a/src/cpu/x64/jit_avx512_core_fp8cvt.hpp
+++ b/src/cpu/x64/jit_avx512_core_fp8cvt.hpp
@@ -205,7 +205,7 @@ private:
     const Xbyak::Xmm xmm_aux4 = xmm5;
     const Xbyak::Xmm xmm_aux5 = xmm6;
     const Xbyak::Opmask kmask_aux = k1;
-    const Xbyak::Reg64 reg64_aux = abi_not_param1;
+    const Xbyak::Reg64 reg64_aux = abi_param3;
     const Xbyak::Reg64 reg64_out = abi_param1;
     const Xbyak::Reg64 reg64_inp = abi_param2;
     void generate() override;


src/cpu/x64/jit_avx512_core_fp8cvt.hpp
```diff
@@ -205,7 +205,7 @@ struct jit_cvt_fp8_t : public jit_generator_t {
     const Xbyak::Xmm xmm_aux4 = xmm5;
     const Xbyak::Xmm xmm_aux5 = xmm6;
     const Xbyak::Opmask kmask_aux = k1;
-    const Xbyak::Reg64 reg64_aux = abi_not_param1;
+    const Xbyak::Reg64 reg64_aux = abi_param3;
```
densamoilov, line 208: Do you know why it fails only in debug build?

spalicki: I do not. As far as I understand it, it should fail in all cases, but it does not.

spalicki:
make test
disable test_device_gpu
disable build_gpu_runtime_ocl
disable benchdnn_all

src/cpu/x64/jit_avx512_core_fp8cvt.hpp
spalicki: This issue only shows when using `RDI` as `reg64_aux` on Windows, other available registers are fine. It only affects debug builds (not even RELWITHMDD) and is not affected by optimization level. I do not know if Intel OpenMP is using RDI in parallel for on Windows Debug or some other component is at fault.

tczeszun: it's due to ABI 1st function param passing register is different between Windows and Linux. Not sure if it's a better way, but there's a hack we used years ago https://github.com/intel-innersource/libraries.performance.math.onednn/blob/main/src/cpu/x64/shuffle/jit_uni_shuffle_kernel.cpp#L451.

PR #3597, opened 2025-07-15
cpu: aarch64: Fix nightly test crash caused by PR #3552

renato-arantes:
# Description

This PR aims to create a workaround for a test failure caused by PR #3552 until a definitive solution is developed. To reproduce the test failure, you need to run:

`DNNL_VERBOSE=all OMP_NUM_THREADS=1 ./tests/benchdnn/benchdnn --conv --dt=u8:s8:f32 --skip-impl=ref,x64:gemm mb1ic64ih1iw33oc1oh1ow33kh1kw24ph0pw23n"l_pad_exceeds_ow_block"`

Fixes #3591 

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

### Bug fixes

- [x] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?

PR #3911, opened 2025-09-09
cpu: aarch64: support for dst u8 in brgemm general and 1x1 fwd conv

kasturedeeksha:
# Description

This commit adds support for the destination type u8 in BRGEMM general and 1x1 forward convolution.

It also addresses issue (https://github.com/uxlfoundation/oneDNN/pull/3591) since the problem was not in the BRGEMM convolution kernel but in the JIT SVE kernel, the temporary fix https://github.com/uxlfoundation/oneDNN/pull/3597 has been removed from BRGEMM and applied to the SVE JIT kernel instead.
For more details, please refer to https://github.com/uxlfoundation/oneDNN/pull/3591#issuecomment-3079188543.

Fixes # https://github.com/uxlfoundation/oneDNN/pull/3591

# Checklist

## General

  1. make test output 

```
99% tests passed, 1 tests failed out of 226
 
Total Test time (real) = 1318.15 sec
 
The following tests FAILED:
        206 - test_benchdnn_modeC_graph_ci_cpu (Failed)
```
- [x] Have you formatted the code using clang-format? Yes


src/cpu/cpu_convolution_list.cpp
```diff
@@ -673,6 +673,8 @@ const std::map<pk_dt_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map()
             CPU_INSTANCE_AARCH64(brgemm_1x1_convolution_fwd_t<sve_256>)
             CPU_INSTANCE_AARCH64(brgemm_convolution_fwd_t<sve_256>)
             CPU_INSTANCE_AARCH64(jit_sve_512_x8s8s32x_convolution_fwd_t<u8, f32>)
+            CPU_INSTANCE_AARCH64(brgemm_1x1_convolution_fwd_t<sve_256>)
+            CPU_INSTANCE_AARCH64(brgemm_convolution_fwd_t<sve_256>)
```
dzarukin: Looks same as ones on lines 673-674.

kasturedeeksha: Updated please check

dzarukin: I don't see any updates to the `pd_t::init` function, which really bothers me. It seems that if this impl gets into a list with s8 output, kaboom gonna happen. I'd kindly ask to address that and please add support according explicit statements in the init function over assumptions on what is supported.

kasturedeeksha: Added datatype check in pd_t::init to allow only supported cases (dst=u8/f32) for int8. These dtypes checks already exist in brgemm.cp : so even if dst=s8, no kernel is created, but now they’re explicitly enforced at the init level.

dzarukin: Thanks for adding those. Could you, please, point me to the spot:
> These dtypes checks already exist in brgemm.cp 

I haven't found those, though if it was extended it should show up in the changeset?

kasturedeeksha: https://github.com/uxlfoundation/oneDNN/blob/2e1cbe4256371177135a533d3781a7e0c15b8bfb/src/cpu/aarch64/brgemm/brgemm.cpp#L232
Here the dtype checks are in brgemm.cpp. The BRGEMM kernel support for int8 was added prior (https://github.com/uxlfoundation/oneDNN/pull/3732) to this PR, which is why it doesn’t appear in the current changeset.

dzarukin: Let's double check to make sure I understand correctly:
Given that you've added u8 dst support but it was marked as supported in the former check - does it mean the previous check was basically wrong and claimed support that wasn't provided?

kasturedeeksha: Not exactly. The BRGEMM kernel is independent of convolution and can support dtypes that convolution itself doesn’t.
BRGEMM is also used by other ops like MatMul. In the earlier change, support was added only at the BRGEMM kernel level, which is why the checks were present there.
In this PR we’re extending convolution to use that support. Since convolution calls into BRGEMM, anything not supported at the kernel level will still return unimplemented from BRGEMM. That said, I agree it’s good practice to add explicit checks in convolution as well, so the so the limits are clear and checked early.

dzarukin: I see, thanks for the clarification, that resolves my confusion.

dzarukin (approved):

aditew01 (approved):
@kasturedeeksha can you please open an issue to track this skip, if it's not there already. 

Temporary fix nightly crash for large l_pad failing test mb1ic64ih1iw33oc1oh1ow33kh1kw24ph0pw23n"l_pad_exceeds_ow_block"

kasturedeeksha:
@aditew01 I created issue to track this https://github.com/uxlfoundation/oneDNN/issues/3981

PR #3916, opened 2025-09-10
Enable MXFP inputs in matmul ref implementation

mgouicem:


src/common/matmul_pd.hpp
```diff
@@ -203,16 +203,19 @@ struct matmul_pd_t : public primitive_desc_t {
...
-                const bool is_decompression_or_dynquant
-                        = utils::one_of(weights_md(0)->data_type, data_type::s8,
-                                  data_type::u8, data_type::s4, data_type::u4)
-                        && IMPLICATION(
-                                !types::is_integral_dt(src_md()->data_type),
-                                attr()->fpmath_.apply_to_int_);
-                ok = ok
-                        && IMPLICATION((mask & wei_qmask_K()),
-                                is_decompression_or_dynquant);
+                // Mask over K dim is allowed for fp types or weights decompression only.
```
dzarukin, line 206: I'm not sure if it's intended but f32/bf16/f16 will have it allowed as well, though kernels don't support this case.

src/common/type_helpers.hpp
```diff
@@ -297,6 +297,37 @@ inline T digits(data_type_t data_type) {
 #undef CASE
 }
 
+inline float round_to_dt(data_type_t data_type, float val) {
```
dzarukin: I fail to find where it's used.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -136,41 +145,67 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
...
+                if (with_wei_decompression) {
+                    if (with_wei_zero_points) {
+                        const dim_t wei_zp_offset
+                                = matmul_helper_t::get_quant_off(
+                                        weights_dims_idx, ndims, wei_zp_mask,
+                                        wei_zp_group_k, wei_zp_group_n,
+                                        wei_zp_md);
+                        const auto wei_zp = io::load_int_value(
+                                wei_zp_dt, wei_zero_points, wei_zp_offset);
+                        w -= wei_zp;
```
dzarukin, line 172: A potential problem of subtracting int value from float value. Given the values for decompression it's unlikely but a comment here might help if it ever comes a case.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -192,20 +232,9 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
                 for_(int m = 2 * m_; m < std::min<int>(2 * (m_ + 1), M); m++)
                 for (int n = 2 * n_; n < std::min<int>(2 * (n_ + 1), N); n++) {
                     dims_t dst_dims_idx;
-                    // account for M, N dims for index calculations
-                    const size_t l_offset = mb * M * N + m * N + n;
-                    utils::l_dims_by_l_offset(
-                            dst_dims_idx, l_offset, dst_d.dims(), ndims);
+                    get_dst_dims_idx(dst_dims_idx, mb, m, n);
```
dzarukin, line 235: I'm not I see the benefit of moving this into a separate lambda given nothing changed in the code and there's no re-usage.

src/common/matmul_pd.hpp
mgouicem: It should not because of the checks in matmul.cpp, where zero_points and scales_group are explicitly allowed only for int, fp8 and fp4.

src/common/type_helpers.hpp
mgouicem: Thanks, this is leftover from rebase, removed.

src/cpu/matmul/ref_matmul.cpp
mgouicem: Good point, fixed.

src/cpu/matmul/ref_matmul.cpp
mgouicem: Mostly for readability and to align with ref_matmul_int (also it will be reused with dynamic quantization support).

dzarukin: Could you, please, drop a comment just in case the support above will be extended just to give a hint where extra changes might appear?

dzarukin (approved):

PR #3917, opened 2025-09-10
xe: jit: create stable interface for pvar_t, pvar_map_t, tile_t, coord_t, icoord_t

rjoursler:
Introduces a place for storing tensor related abstractions. The scope is currently limited to `pvar_t`, `pvar_map_t`, `tile_t`, `coord_t`, and `icoord_t`. In an upcoming PR, this will be extended to include `layout_t` (which is involved to extract from oneDNN due to dependencies on `tile_coord_t` and `memory_desc_t`), `dsl::tensor_t`, and `dsl::global_tensor_t`. As part of this, `pvar_t` was renamed to `idx_t` within the stable interface to improve clarity. The reasoning is that in the context of tensors, these correspond to indices in a basis vector. I am not convinced this is the best name to represent this concept (and am open to better ideas), but the other name that clearly represents this concept in my mind is `dimension_t/dim_t` which I avoided due to its similarity with `dim_t` usage within oneDNN.

### DIFF
--- a/src/gpu/intel/jit/dsl/tensor.hpp
+++ b/src/gpu/intel/jit/dsl/tensor.hpp
@@ -247,17 +247,6 @@ public:
 
     size_t get_hash() const { return ir_utils::get_hash(map_); }
 
-    void stringify(std::ostream &out) const {
-        if (is_empty()) {
-            out << "x";
-            return;
-        }
-        for (auto &d : *this) {
-            auto &value = operator[](d);
-            out << d << value;
-        }
-    }
-
     void parse(std::istream &in) {
         auto s = stream_parse<std::string>(in);
         if (s == "x") return;
--- a/src/gpu/intel/jit/ir/blocking.hpp
+++ b/src/gpu/intel/jit/ir/blocking.hpp
@@ -71,11 +71,11 @@ public:
     void stringify(std::ostream &out) const {
         out << "simd=" << simd_;
         out << " l=";
-        loop_.stringify(out);
+        out << loop_.str();
         out << " T=";
-        thread_group_.stringify(out);
+        out << thread_group_.str();
         out << " i=";
-        iter_.stringify(out);
+        out << iter_.str();
     }
 
     void parse(std::istream &in) {
@@ -469,7 +469,7 @@ public:
     void set_id(int id) { id_ = id; }
 
     void stringify(std::ostream &out) const {
-        blocking_.stringify(out);
+        out << blocking_.str();
         if (bufs_hint_ != -1) out << " bufs=" << bufs_hint_;
     }
 


src/gpu/intel/jit/dsl/tensor.hpp
```diff
@@ -0,0 +1,409 @@
...
+    std::string str() const { return name_.str(); }
+
+private:
+    class name_t {
+    public:
+        name_t() = default;
+        explicit name_t(uint32_t idx) { data_[0] = into<char>('a' + idx); }
+        explicit name_t(const std::string &s) {
+            gpu_assert(!s.empty() && s.length() <= max_len);
+            s.copy(data_, s.length());
```
echeresh, line 59: Looks like `data_` is not null-terminated but it's used with `strlen()` below.

src/gpu/intel/jit/dsl/tensor.hpp
```diff
@@ -0,0 +1,409 @@
...
+        };
+        std::string str() const { return data_; }
+
+    private:
+        uint64_t numeric_value() const {
+            uint64_t ret;
+            static_assert(sizeof(*this) == sizeof(ret),
+                    "name_t is intended to be small enough to fit in a single "
+                    "register");
+            memcpy(&ret, this, 8);
```
echeresh: Nit:
```suggestion
            memcpy(&ret, this, sizeof(ret));
```

src/gpu/intel/jit/ir/problem.hpp
```diff
@@ -137,353 +74,21 @@ extern pvar_t n;
...
 
 struct tile_coord_t {
     tile_t tile;
     coord_t coord;
+    bool is_valid;
 
-    tile_coord_t() = default;
+    tile_coord_t(bool is_valid = true) : is_valid(is_valid) {}
     tile_coord_t(const tile_t &tile, const coord_t &coord = {})
         : tile(tile), coord(coord) {}
```
echeresh: Missing `is_valid` initialization

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -389,7 +389,7 @@ void block_send(const tensor_t &t, const global_tensor_t &g,
     bool is_prefetch = t.buf.is_empty();
     auto &operation_tile = is_prefetch ? g.tile : t.layout.tile();
 
-    pvar_t w_dim;
+    idx_t w_dim;
```
echeresh: Looks like this has to be `w_idx`, correct?

It's fine to change later but it'd be nice to align naming in all such places eventually (or at least start with DSL files).

src/gpu/intel/jit/ir/problem.hpp
```diff
@@ -42,71 +43,7 @@ enum class tensor_kind_t {
...
-    static pvar_t from_index_var(const expr_t &ndex_var);
-    static pvar_t from_var(const expr_t &var);
-
-    char to_spatial() const;
-    int spatial_index() const;
-
-private:
-    pvar_name_t name_;
-};
+using pvar_t = dsl::idx_t;
```
echeresh, line 46: Just checking: is your plan to complete full migration to the new name in the future?
This alias seems redundant after renaming.

rjoursler: I was thinking we should just leave this as the legacy interface. In particular, I am concerned it would cause conflicts with some other random `idx_t` declarations within oneDNN, and as such require maintaining the `dsl` namespacing.

src/gpu/intel/jit/ir/problem.hpp
rjoursler: Good catch!

rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_pool
enable benchdnn_reduce
enable benchdnn_reorder

src/gpu/intel/jit/ir/blocking.hpp
```diff
@@ -71,11 +71,11 @@ class blocking_t {
     void stringify(std::ostream &out) const {
         out << "simd=" << simd_;
         out << " l=";
-        loop_.stringify(out);
+        out << loop_.str();
         out << " T=";
-        thread_group_.stringify(out);
+        thread_group_.str();
```
kealan-barbieri: should it be `out << thread_group_.str()`?

src/gpu/intel/jit/ir/problem.hpp
echeresh: I think the plan is to maintain and use `dsl` namespace. This has to be `...intel::jit::dsl::idx_t`, correct?

I see what you are saying. Would it be too bad if use all DSL abstractions with `dsl::` namespace? This would be similar to nGEN. To me it seems a longer name is better than an alias with a different name - less cognitive load after all.

rjoursler: > Would it be too bad if use all DSL abstractions with dsl:: namespace?

This is probably fine, it just strikes me as a little counterintuitive if implementations which are not following DSL idioms explicitly rely on DSL interfaces. In general, I have no issue with performing this refactoring as it make sense to do though.

rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

PR #3919, opened 2025-09-10
xe: gemm: improve LNL transitionary N strategy

Simonsays095:
Addresses [MFDNN-14123](https://jira.devtools.intel.com/browse/MFDNN-14123). LNL is not allowed to use kernels marked 'h' (i.e. `ReqNoIntegrated`) - in a previous PR, a kernel was split so that LNL doesn't use it from N=64-128, but no kernel was added to replace it. Additionally, the kernel over N=49-64 is unstable with strange (low) performance.

Fixed both issues by adding a new kernel that covers N=49-128 performantly, and simplifying BMG kernel selection.


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
@rjoursler Here's that plot. Due to data compression effects, I had to change your PR to use `--mode=P --mode-modifier=P --max-ms-per-prb=10` (i.e. fast-mode but with randomized data).
image

PR #3920, opened 2025-09-10
common: verbose: update empty tensor/parameter checks with tensor names

avmanerikar:
# Description

The PR contains minor updates to verbose checks to specify tensor/parameter names for empty tensor checks and bad parameter
checks respectively.

### DIFF
--- a/src/cpu/x64/jit_uni_binary.cpp
+++ b/src/cpu/x64/jit_uni_binary.cpp
@@ -137,7 +137,7 @@ status_t jit_uni_binary_t::pd_t::init(engine_t *engine) {
             VERBOSE_ISA_DT_MISMATCH);
     VDISPATCH_BINARY(
             set_default_params() == status::success, VERBOSE_UNSUPPORTED_TAG);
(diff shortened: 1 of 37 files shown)


src/cpu/x64/jit_uni_binary.cpp
```diff
@@ -137,7 +137,7 @@ status_t jit_uni_binary_t::pd_t::init(engine_t *engine) {
             VERBOSE_ISA_DT_MISMATCH);
     VDISPATCH_BINARY(
             set_default_params() == status::success, VERBOSE_UNSUPPORTED_TAG);
-    VDISPATCH_BINARY(!has_zero_dim_memory(), VERBOSE_EMPTY_TENSOR, "");
+    VDISPATCH_BINARY(!has_zero_dim_memory(), VERBOSE_EMPTY_TENSOR, "src");
```
mzhukova: not 100% sure, but could be src0?

avmanerikar: Good catch. Updated the commits. Thanks!

Simonsays095 (approved):

avmanerikar:
make test

PR #3921, opened 2025-09-10
[GPU] Minor updates to IR/DSL

echeresh:
PR includes many minor changes, some of them:

- Added proper checks to support IR shuffle with SIMD type
- [style] Unified `def` and `let` API - put `name` first
- [style] Use `sub()` instead of `map()` and use consistent order of `tile`/`coord` arguments
- Refactored `alloc_manager_t` to reuse it for the new IR API (without `alloc` and `let`)
- Updated `tensor_t::sub()` to rely on `ref_t` (variable reference) instead of `ptr_t` to preserve the size
- [DSL] Introduced group ID support
- [DSL] Introduced `def(name, layout, attr)` to support attributes such as SIMD or SLM
- Fixed layout related to dimension semantics to use with `split()` API in the future
- Fixed `simplify()` to skip some unsupported/unexpected types


src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -321,8 +321,8 @@ class layout_t {
 
     bool is_empty() const { return ndims_ == 0; }
     bool with_ndims() const { return ndims_ != dim_idx::invalid; }
-    dim_idx_t ndims() const {
-        gpu_assert(with_ndims());
+    dim_idx_t ndims(bool check_invalid = true) const {
+        if (check_invalid) gpu_assert(with_ndims());
```
rjoursler, line 325: Do you know what are the main blockers with removing `ndims()` from `layout_t`? It seems to me like this interface doesn't really fit with how this structure should work.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -52,7 +52,10 @@ struct ctx_t {
                         local_id_type(), ir_builder_t::local_id(i));
                 group_ids_[i] = var_t::make(
                         group_id_type(), ir_builder_t::group_id(i));
+                subgroup_ids_[i] = def("sg_id",
```
rjoursler: We are missing the dimension associated with the variable: 
```
"sg_id" + "012"[i]
```
Also, I don't think there is much benefit to shortening `subgroup -> sg`, that just seem like it obfuscates how the IR maps to the DSL interface.

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -195,21 +195,15 @@ class lval_t {
...
+lval_t def(const std::string &name, const type_t &type,
+        const expr_t &value = {}, bool force_alloc = false);
 lval_t def(const std::string &name, const expr_t &value);
-tensor_t def(const layout_t &layout, const std::string &name,
-        const expr_t &value = {});
-expr_t let(type_t type, const std::string &name, const expr_t &value);
+tensor_t def(const std::string &name, const layout_t &layout,
+        type::attr_t attr = {});
+tensor_t def(const std::string &name, layout_t layout, const expr_t &value,
+        type::attr_t attr = {});
```
rjoursler, line 202: These interfaces seems a be a bit convoluted with the addition of type attributes. We already have up to two different types floating around (`value.type()`, and `layout.type()`), why not just pass in the type attributes via those structures? Also with this change, it becomes unclear the distinction between `tensor_t` with a global type attribute and `global_tensor_t`. Maybe it would make sense to introduce something `local_tensor_t` for an SLM tensor or remove the distinction and just depend on type attributes? To some extent in my prototyping, I felt like the `global_tensor_t` abstraction doesn't seem quite coherent anyway as it ends up being a blending of builder state information (mainly via `coord`,  and to some extent `base_offset`) with tensor object information.

src/gpu/intel/jit/dsl/dsl.hpp
```diff
@@ -34,6 +34,7 @@ int grf_size();
 int min_align_2d();
 int min_pitch_2d();
 
+using type_t = dnnl::impl::gpu::intel::jit::type_t;
```
rjoursler: nit: this seems like it will need declared in `dsl/decl.hpp` due to the intended namespace changes when upstreaming.

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -71,19 +92,25 @@ struct ctx_t {
...
     }
 
     kernel_t end_kernel() {
         gpu_assert(stmts_stack_.size() == 1)
                 << "Invalid end of kernel, imbalanced scopes detected";
-        kernel_t ret {std::move(interface_), pop_scope(), ctx_->exec_cfg()};
+        auto body = pop_scope();
+        if (slm_byte_offset() > 0) {
+            auto slm_buf = var_t::make(u8[slm_byte_offset()].with_slm(), "slm");
+            auto slm_alloc = builtin_t::make("alloc")(slm_buf);
```
rjoursler, line 93: Would it make sense to modify `slm_byte_offset()` to be an `expr_t` an export it via `interface_` so that we can support dynamic slm sizing?

echeresh: For now we don't need dynamic SLM sizing - all SLM tensors by design have static sizes. But I can move SLM size to the interface if you think it's a better approach.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: > We already have up to two different types floating around (value.type(), and layout.type())

`value` is optional, right? As for `layout.type()` - my assumption was to not keep attributes in the layout and ensure that `layout.type()` is always scalar and no attributes type.

Why I think we should do that:

- `layout` is a logical description. We map coordinates to offset in elements. Whether the underlying memory is GRF, SLM or something else is not important for most layout operations
- On contrary, DSL/IR variables (their types in particular) define the underlying storage - GRF, SLM, etc. This is controlled by attributes

> it becomes unclear the distinction between tensor_t with a global type attribute and global_tensor_t

I think we can come up with an abstraction unifying these but operations on them don't overlap much. One obstacle is that `tensor_t` has static sizes while `global_tensor_t` has dynamic sizes.

Also, we don't have "global type attribute". I see how all of this can be introduced and supported but I don't see it being simpler.

As for attributes, I have at least three use cases:

- SIMD register tensor (and non-SIMD)
- Packed register tensor (and non-packed)
- SLM backed tensor

First two define 4 flavors that differ by `buf` type only.

The third case has one difference - its offset can be dynamic (e.g. for pipelining). This difference doesn't seem to be worth introducing a separate abstraction though.

src/gpu/intel/jit/ir/tensor.hpp
echeresh: The answer here is the legacy code :)
We have a lot of places relying on `ndims()` based loops. I ported many important places but there is more work left. I consider this as tech debt.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Thanks, fixed.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/dsl/dsl.hpp
rjoursler: > my assumption was to not keep attributes in the layout and ensure that layout.type() is always scalar and no attributes type.

I think your reasoning makes sense here (although I am not 100% convinced with respect to the packed attribute). We should probably add assertions to enforce it.

> One obstacle is that tensor_t has static sizes while global_tensor_t has dynamic sizes.

From what I have seen, I have been thinking the distinction between static sized and dynamic sized tensors likely makes more sense to target from an  DSL API perspective than the current interface. It also aligns well with the `coord_t/icoord_t` division.

> First two define 4 flavors that differ by buf type only.

Yeah, this overall concept makes sense, I am just trying to avoid ambiguity/complexity in the tensor API. How about we add a new enum kind or attribute directly targeting tensors? This should make the interface more clear and easier to document.

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: I don't see a reason to do this now, but I am general thinking it will likely be useful to extend the DSL with a kernel submission API.  The intention would be to simplify use with precomputed constants such as can be used for division and dynamic SLM size, and avoid all the invasive changes on implementations associated with using these constants.

src/gpu/intel/jit/dsl/dsl.hpp
echeresh: > Although I am not 100% convinced with respect to the packed attribute

Can you please elaborate? If your concern is attribute combination, we can add checks. Say, SLM and SIMD are not compatible - so we have to enforce it. This is similar to other combinations.

> From what I have seen, I have been thinking the distinction between static sized and dynamic sized tensors likely makes more sense to target from an DSL API perspective than the current interface. It also aligns well with the coord_t/icoord_t division.

If you have something specific in mind, could you please draft a short extension to the RFC when you have time? API draft and usage examples would help to discuss that in detail.

> Yeah, this overall concept makes sense, I am just trying to avoid ambiguity/complexity in the tensor API. How about we add a new enum kind or attribute directly targeting tensors? This should make the interface more clear and easier to document.

Can you please comment on your concerns? I see that the type attribute describes the tensor pretty well (except a pointer attribute - which is not applicable). So the new attribute/kind would duplicate most of `type::attr_t`.

rjoursler: > Can you please elaborate?  If your concern is attribute combination, we can add checks. 

No, the main thing that stands out to me is when we are measuring things in bytes, in particular, `layout_t::offset_in_bytes`. If the packed attribute is not part of `layout_t`, my concern is would create complexity when writing tensor operations (such as [here](https://github.com/uxlfoundation/oneDNN/blob/aa270a93d87c674794849212cc7ec8aaeb62b849/src/gpu/intel/jit/dsl/dsl.cpp#L617)).  While this may be a non-issue, it isn't obvious to me this design won't result in a bunch of extra case handling elsewhere.

> Can you please comment on your concerns? I see that the type attribute describes the tensor pretty well (except a pointer attribute - which is not applicable)

By my reckoning, about half of the current attributes aren't applicable (`attr_t::{ptr, mut}`). While I agree this is semi-duplicative, it is very minor amount. I have a two main issues from a stable API perspective:
*  I think this it is just not obvious what core application is as there are multiple types floating around. As such, when learning the interface, developers will need to find documentation (or implementation) to understand its use. If we have a tensor specific modifier for passing this information, this information would be easily discoverable by documenting the modifier.
* Adding new `type::attr_t`  is more complex (and non-local) as now we should add validation or documentation how this handled at this interface.

echeresh: > No, the main thing that stands out to me is when we are measuring things in bytes, in particular, layout_t::offset_in_bytes. If the packed attribute is not part of layout_t, my concern is would create complexity when writing tensor operations (such as [here](https://github.com/uxlfoundation/oneDNN/blob/aa270a93d87c674794849212cc7ec8aaeb62b849/src/gpu/intel/jit/dsl/dsl.cpp#L617)). While this may be a non-issue, it isn't obvious to me this design won't result in a bunch of extra case handling elsewhere.

Good point. This method is not compatible with logical/physical split between layout and type. To me, it sounds like the only good option is to drop `offset_in_bytes()` from the layout. It's fine to keep it in non-DSL code as non-member function though.

> By my reckoning, about half of the current attributes aren't applicable (attr_t::{ptr, mut}). While I agree this is semi-duplicative

1. To preserve semantics I think `mut` has to be added by default with `def()`. It'd be nice to align with variable behavior - to have one set of rules. Tensors (with their buffers) are usually mutable so this attribute is applicable.
2. Pointer is a special attribute, agree (but it also has special handling in some other places)

> I think this it is just not obvious what core application is as there are multiple types floating around. As such, when learning the interface, developers will need to find documentation (or implementation) to understand its use.

Well, at a certain angle the interface we discuss (that is `t = def(name, layout)`) is a wrapper for:

```c++
auto buf = def(name, layout.type().with_elems(layout.elems()));
t = tensor_t(buf, layout);
```

It's quite intuitive: the passed attribute is applied to the underlying tensor buffer (which is missing in the snippet). Learning the type attributes is fundamental to the DSL usage. This is one of a few things that pop up even in a simple SIMT/DSL example.

> Adding new type::attr_t is more complex (and non-local) as now we should add validation or documentation how this handled at this interface.

I agree but I don't see much difference vs introducing a new data type and ensuring it's properly supported in all expression operations. Here it's simpler though, we have one entry point where it's relatively easy to validate that the passed attribute is supported.

As for the new kind - I don't like this idea much as we force the users to learn more concepts vs building on top of the existing ones.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/jit/pass/simplify.cpp
```diff
@@ -671,13 +671,15 @@ class range_simplifier_t : public ir_mutator_t {
...
-        if (obj.op_kind == op_kind_t::_or) {
-            if (cset.can_prove(a) || cset.can_prove(b)) return expr_t(true);
-        } else if (obj.op_kind == op_kind_t::_and) {
-            auto not_a = try_not(a);
-            if (not_a && cset.can_prove(not_a)) return expr_t(false);
-            auto not_b = try_not(b);
-            if (not_b && cset.can_prove(not_b)) return expr_t(false);
+        if (obj.type.is_bool()) {
+            if (obj.op_kind == op_kind_t::_or) {
+                if (cset.can_prove(a) || cset.can_prove(b)) return expr_t(true);
```
hidefromkgb, line 676: Should we also add negations?
`if (can_prove(not_a) && can_prove(not_b)) return expr_t(false);`, here and below.

echeresh: Yep, looks missing, thanks. If you don't mind I suggest to add this if/when we'll need this in the future.
(just to avoid unnecessary changes and potential issues in the simplification flow)

PR #3925, opened 2025-09-11
[GPU] GEMM IP Fix

kealan-barbieri:
# Description

Previous kparallel restrictions were too strict, causing a change in primitive selection in nested GEMMs in RNN, failing IP case.

- explicitly disable unsupported sum+dst scale with kparallel
- add fallback for `threads_per_eu` query in rnn, close gap for `with_post_ops` GEMM

Fixes # [MFDNN-14155](https://jira.devtools.intel.com/browse/MFDNN-14155)

# Checklist

## General

### Bug fixes

### DIFF
--- a/src/gpu/intel/rnn/grid.cpp
+++ b/src/gpu/intel/rnn/grid.cpp
@@ -756,13 +756,13 @@ status_t simple_common_t<aprop>::pd_t::init(impl::engine_t *engine) {
         attr.deterministic_ = this->attr()->deterministic_;
         CHECK(dnnl::impl::create_gemm_pd(gemm_pd, engine, &a_md, &b_md, &c_md,
                 &glob_zero_md, c_dt, &attr));
-        if (threads_per_eu == 0)
-            CHECK(gemm_pd->query(
-                    query::preferred_gpu_threads_per_eu, 0, &threads_per_eu));
-        else if (get_verbose_dev_mode(verbose_t::debuginfo) > 1) {
+        bool verbose = get_verbose_dev_mode(verbose_t::debuginfo) > 1;
+        if (threads_per_eu == 0 || verbose) {
             auto t = 0;
-            CHECK(gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t));
-            if (t != threads_per_eu)
+            auto s = gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t);
+            if (threads_per_eu == 0)
+                threads_per_eu = (status::success != s) ? t : 128;
+            if (verbose && t != threads_per_eu)
                 verbose_printf("[WARNING] GEMM grf modes are inconsistent\n");
         }
         return status::success;


src/gpu/intel/gemm/with_post_ops.hpp
```diff
@@ -42,6 +42,9 @@ struct with_post_ops_t : public primitive_t {
         bool use_scratchpad() const {
             return use_scratchpad_with_post_op_worker;
         }
+        status_t query(query_t what, int idx, void *result) const override {
+            return pd_->query(what, idx, result);
+        }
```
rjoursler, line 47: Side comment: It looks like this implementation also needs to configure the OCL kernel to use the same number of GRFS as the gemm kernel [similar to what RNN does starting here](https://github.com/uxlfoundation/oneDNN/blob/aa270a93d87c674794849212cc7ec8aaeb62b849/src/gpu/intel/rnn/grid.cpp#L336).

src/gpu/intel/rnn/grid.cpp
```diff
@@ -756,13 +756,17 @@ status_t simple_common_t<aprop>::pd_t::init(impl::engine_t *engine) {
...
+                            &threads_per_eu))
+                threads_per_eu = 128;
+        } else if (get_verbose_dev_mode(verbose_t::debuginfo) > 1) {
             auto t = 0;
-            CHECK(gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t));
-            if (t != threads_per_eu)
+            if (status::success
+                            == gemm_pd->query(
+                                    query::preferred_gpu_threads_per_eu, 0, &t)
+                    && t != threads_per_eu)
```
rjoursler: I think we can simplify the logic here some:
```C++
auto t = 0;
auto status = gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t);
if(status == status::success) {
    if (threads_per_eu == 0) 
         threads_per_eu = t;
    else if(threads_per_eu != t)
         verbose_printf("[Warning] GEMM grf modes are inconsistent\n");
}
```

src/gpu/intel/gemm/with_post_ops.hpp
kealan-barbieri: looks like this is already being handled here: https://github.com/uxlfoundation/oneDNN/blob/aa270a93d87c674794849212cc7ec8aaeb62b849/src/gpu/intel/gemm/with_post_ops.hpp#L62

src/gpu/intel/rnn/grid.cpp
kealan-barbieri: Simplified this in a similar way preserving the verbose check and fallback value does it look okay?

src/gpu/intel/gemm/with_post_ops.hpp
rjoursler: You're right, I missed seeing that.

src/gpu/intel/rnn/grid.cpp
rjoursler: No, that check doesn't look quite right. In particular, we could get different behavior when calling with verbose as `thread_per_eu` could be different value. Also, as a general rule, I would recommend against calling `query` in the if statement as well. The reason is that `query` mutates program state (via setting t), and the pattern of mutating state in if statements can lead to subtle bugs due to evaluation order.

kealan-barbieri: got it, adjusted again, the only difference I can see here from the previous algorithm is it may throw more warnings than before due to the default value assignment. where previously it would only have warned once in a sequence of `threads_per_eu = 0` now it will throw a warning each time.

src/gpu/intel/rnn/grid.cpp
```diff
@@ -756,14 +756,14 @@ status_t simple_common_t<aprop>::pd_t::init(impl::engine_t *engine) {
...
+        bool verbose = get_verbose_dev_mode(verbose_t::debuginfo) > 1;
+        if (threads_per_eu == 0 || verbose) {
             auto t = 0;
-            CHECK(gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t));
-            if (t != threads_per_eu)
+            auto s = gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t);
+            if (verbose && t != threads_per_eu)
                 verbose_printf("[WARNING] GEMM grf modes are inconsistent\n");
+            if (status::success != s) t = 128;
+            threads_per_eu = t;
```
rjoursler: This needs to initialize `threads_per_eu` before the verbose logging, otherwise we will generate warnings when initially setting `threads_per_eu`. Also `threads_per_eu` should be guarded behind a check that it is uninitialized via checking `threads_per_eu == 0`.

kealan-barbieri: I see, moved above verbose log with guard.

kealan-barbieri:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_lstm
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

PR #3928, opened 2025-09-11
CPU : x64: Matmul: fp4 weights support 

amakarev:
Support fp4 weights for matmul primitive

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
@@ -3442,12 +3442,14 @@ struct jit_brgemm_matmul_copy_b_f32_t : public jit_brgemm_matmul_copy_b_t,
         , jit_generator_t(jit_name())
         , dt_in_(conf->orig_wei_dt)
         , simd_w_(vreg_traits_t<Vmm>::vlen / sizeof(float))
+        , is_src_f4_(one_of(
+                  conf->orig_wei_dt, data_type::f4_e2m1, data_type::f4_e3m0))
         , is_src_int4_(one_of(conf->orig_wei_dt, data_type::s4, data_type::u4))
         , req_zp_b_shift_(
                   conf->has_zero_point_b && conf->with_wei_decompression)
         , req_apply_wei_scales_(conf->apply_scales_in_buffer_b)
         , typesize_in_(types::data_type_size(dt_in_))
-        , src_elems_per_byte_(is_src_int4_ ? 2 : 1)
+        , src_elems_per_byte_(is_src_int4_ || is_src_f4_ ? 2 : 1)
         , wei_scales_typesize_(conf_->wei_scales_dt_sz)
         , src_stride_(conf_->copy_B_wei_stride)
         , tr_src_stride_(conf_->LDB * typesize_out_)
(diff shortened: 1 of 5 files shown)


src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3471,7 +3473,7 @@ struct jit_brgemm_matmul_copy_b_f32_t : public jit_brgemm_matmul_copy_b_t,
 
     const data_type_t dt_in_;
     const int simd_w_;
-    const bool is_src_int4_, req_zp_b_shift_, req_apply_wei_scales_;
+    const bool is_src_f4_, is_src_int4_, req_zp_b_shift_, req_apply_wei_scales_;
```
xuxinzen, line 3476: For my understanding, one byte contains two elements of fp4 which is the same as int4. 
So `src_elems_per_byte_` should be initialized as `2`, and you should also check all the offsets/addr are correct to avoid segfault.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3708,6 +3719,26 @@ void jit_brgemm_matmul_copy_b_f32_t<Vmm>::generate() {
         kmovw(kAAAA, 0xaaaa);
         kmovw(k5555, 0x5555);
     }
+    if (is_src_f4_) {
+        alignas(64) static constexpr const float f4_e2m1_table[16]
+                = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f, -.5f,
+                        -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
+        alignas(64) static constexpr const float f4_e3m0_table[16]
+                = {0.0f, .25f, .5f, 1.0f, 2.0f, 4.0f, 8.0f, 16.0f, -0.0f, -.25f,
+                        -.5f, -1.0f, -2.0f, -4.0f, -8.0f, -16.0f};
```
xuxinzen, line 3735: Why do we need to use these `float `values in this permute table? We use `vpermd` to do the permutation which data type is not matched.

amakarev: here the idea is a little different. since fp4 can contain only 16 values ​​we can simply enter them into the table and use the value of the elements being rearranged as an index `vpermd(vmm_in, vmm_in, vmm_permd);`

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: thanks

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: I probably rushed my answer, now I understand what you meant. I'll upload a commit with an updated version soon

amakarev: Still, my previous answer is where I’ll stand.

_For example here:
 vpermps(vmm_in, vmm_in, vmm_f4_lut); // LUT to convert fp4 to fp32
This last step is using the 4 LSB in each 32bit chunk as index and picks up an element from vmm_f4_lut with contents:
[0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0, -0.0, -0.5, -1.0, -1.5, -2.0, -3.0, -4.0, -6.0]_

ankalinin (approved):

xuxinzen:
Do we need to add CI tests on fp4? I only saw such tests on GPU

amakarev:
> Do we need to add CI tests on fp4? I only saw such tests on GPU

Sure. I'll add it.

amakarev:
> Do we need to add CI tests on fp4? I only saw such tests on GPU

seems its added with https://github.com/uxlfoundation/oneDNN/commit/4b679dd728d8ceb1634f5177cfe1c5baf649df92

PR #3930, opened 2025-09-11
[GPU][GEMM] Enforce 2d send pitch min

kealan-barbieri:
# Description

Spec requires pitch <= width, this doesnt currently cause functional fails on hardware but causes failures in simulation. 

# Checklist

## General

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/pieces/address_setup.cxx
+++ b/src/gpu/intel/gemm/jit/generator/pieces/address_setup.cxx
@@ -378,8 +378,10 @@ void Generator<hw>::setupAddr(Type T, const GRFRange &addr, const BO &ptr, const
                 auto pitch = bw * bcount * block.ebytes;
                 if (pitch < 64 || pitch & 0xF) hw_unsupported();
                 mov(1, addr[0].ud(4), pitch - 1);
-            } else
+            } else {
                 add(1, addr[0].ud(4), bld, -1);
+                max_<uint32_t>(1, addr[0].ud(4), addr[0].ud(4), addr[0].ud(2)); 
+            }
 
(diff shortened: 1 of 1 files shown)


src/gpu/intel/gemm/jit/generator/pieces/address_setup.cxx
```diff
@@ -378,8 +378,10 @@ void Generator<hw>::setupAddr(Type T, const GRFRange &addr, const BO &ptr, const
                 auto pitch = bw * bcount * block.ebytes;
                 if (pitch < 64 || pitch & 0xF) hw_unsupported();
                 mov(1, addr[0].ud(4), pitch - 1);
-            } else
+            } else {
                 add(1, addr[0].ud(4), bld, -1);
+                max_<uint32_t>(1, addr[0].ud(4), addr[0].ud(4), addr[0].ud(2)); 
+            }
```
petercad, line 384: Unfortunately this patch won't work -- results will be incorrect. What's happening here is that when the matrix base address is not 64b-aligned, gemmstone will align the base address down, then increase offset_x and width to compensate. This works because we never need negative edge OOB handling. But the result typically is that pitch < width. As mentioned in the PR, real HW is just fine with this, it's only a restriction in the spec.

I can think of one other option, but it's a little fiddly. If `ctz(pitch in bytes) <= ctz(base address)` (ctz = count trailing zeros), then we can increase height (and offset_y) to push the base address to a 64b boundary. This condition holds for basically all cases of interest, but calculating exactly how much to increase height by is a bit annoying.

kealan-barbieri: thank you, will look into a proper fix, as it sounds like this may be causing issues on prv-gpu.

PR #3932, opened 2025-09-12
graph: backend: dnnl: adopt host scalar memory api

TaoLv:
Now host scalar APIs are available for memory descriptor and memory. Change DNNL backend to adopt the new APIs and remove workarounds.


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/op_executable.hpp
```diff
@@ -505,12 +509,27 @@ struct host_scalar_executable_t : public op_executable_t {
...
 
-        auto prim = dnnl::reorder(src_mem, dst_mem);
-
-        auto e = dnnl::ocl_interop::execute(prim, stream, args, deps);
+        assert(deps.size() <= 1);
+        // Passing the empty event to memcpy below causes failure.
+        const bool empty = deps.empty() || deps[0] == nullptr;
+        const cl_uint num = empty ? 0 : static_cast<cl_uint>(deps.size());
+        const size_t size = src_mem.get_desc().get_size();
+        assert(size == 4);
```
ElaineBao: this is easy to break if we support new data types for host scalar, better to use sizeof(dtype)?

TaoLv: Updated.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #3935, opened 2025-09-12
rfcs: OCP MX dynamic quantization support

mgouicem:
This is a proposal to support MXFP datatype. In particular, to support dynamic quantization of outputs.

[Link](https://github.com/uxlfoundation/oneDNN/tree/mgouicem/rfcs/mxfp/rfcs/20250911-mxfp-support) for rendered version.

### DIFF
--- a/rfcs/20250911-mxfp-support/README.md
+++ b/rfcs/20250911-mxfp-support/README.md
@@ -31,8 +31,8 @@ How the scales are computed is defined by the MXFP spec. For a given
 1d block of size 32 (denoted $`X`$), its scale `S` and quantized values `Q`
 are defined as:
 
-$$ S = E8M0(amax(X)) / E8M0(MAX\_DST\_DT) $$
-$$ Q = dst\_dt(X / S) $$
+$$ S = E8M0(amax(X)) / E8M0(MAX\\_DST\\_DT) $$
+$$ Q = dst\\_dt(X / S) $$
 
 With `E8M0` a conversion to `E8M0` datatype, by rounding down to the
 closest power of two below the argument value, `MAX_DST_DT` the
--- a/rfcs/20250911-mxfp-support/README.md
+++ b/rfcs/20250911-mxfp-support/README.md
@@ -1,0 +1,169 @@
+# Open Compute Microscaling (MX) datatype support
+
+There is growing interest in using block scaling formats in recent
+LLMs. In an effort to standardize hardware support for those, the Open
+Compute Platform (OCP) Microscaling standard (MX) [^1] defines
+interchange formats as well as operations (dot-product and
+conversions).  In particular, MXFP4 is currently used in gpt-oss [^2]
+for weights compression.
+
+In this RFC, we go over what are the MX formats, and what oneDNN needs
+to support those.
+
+## MX formats and requirements to oneDNN
+
+Even though they are sometimes referred to as datatypes, MX formats
+are actually a quantized blocked format to represent tensors.
+
+Here are the elements oneDNN need to support for MX spec [^1] conformance:
+- block element types. All types documented in MXFP spec but `fp6` are
+  already supported by oneDNN API (namely, `s8`, `f8_e5m2`, `f8_e4m3`,
+  `f4_e2m1`).
+- block scaling with 1d groups of 32. This is supported by oneDNN
+  through `set_scales` API.
+- `e8m0` block scale datatype. This is supported in oneDNN.
+- support dynamic quantization of output. This is the missing piece
+  and main point of discussion of this RFC.
+
+## MX Dynamic quantization definition
+
+How the scales are computed is defined by the MXFP spec. For a given
+1d block of size 32 (denoted $`X`$), its scale `S` and quantized values `Q`
+are defined as:
+
+$$ S = E8M0(amax(X)) / E8M0(MAX\_DST\_DT) $$
+$$ Q = dst\_dt(X / S) $$
+
+With `E8M0` a conversion to `E8M0` datatype, by rounding down to the
+closest power of two below the argument value, `MAX_DST_DT` the
+maximum representable value in `dst` type, and `dst_dt` a conversion
+to `dst` type.
+
+Note that the division operation happens _after_ rounding down to the
+scale datatype (`E8M0`). This allows to implement this division as a
+subtraction of the exponent fields. However, this differs from other
+dynamic scaling formulas, for example:
+- traditional int8 dynamic quantization applies the division _before_
+  conversion to scale datatype
+- cuDNN [^3] and cublas [^4] formula compute the division in f32
+  _before_ the conversion to scale datatype.
+
+## Proposals
+
+First of all, this RFC focuses on MXFP formats supports but might
+cover some generalizations to allow extension to other formats and
+recipes.  We will also assume that scales and data are not interleaved
+in memory, similarly to current quantization support.
+
+### Option 1 (Recommended): Through a new attribute (`set_dynamic_scales`)
+
+Here we propose to expose a new `dynamic_scales` attribute, similar to
+`scales` attribute except that:
+- it would be applicable only to output memory,
+- it implies that scales are computed by oneDNN, and not provided by user,
+- it implies a new output memory to store the computed scales.
+
+In order to support MXFP formats, it needs to support groups of 32,
+and a way to express along which axis the grouping occurs. It also
+needs to support `E8M0` format for scales. We propose to use the same
+semantics as with `set_scales` method:
+- grouping factor can be made explicit with mask, ndims and group_dims
+  arguments,
+- `E8M0` scale datatype can be set explicitly with a scale data_type
+  argument.
+- regarding the scale computation formula, we would not expose a knob
+  to configure it, and support only MXFP spec formula.
+
+This will result in the following new entry-point in C API:
+```C
+dnnl_status_t DNNL_API dnnl_primitive_attr_set_dynamic_scales(
+        dnnl_primitive_attr_t attr, int arg, int mask, int n dims,
+        const dnnl_dims_t group_dims, dnnl_data_type_t data_type);
+```
+
+And the associated symbol in C++ API:
+```c++
+    void set_dynamic_scales(int arg, int mask, const memory::dims &groups,
+            memory::data_type data_type = memory::data_type::f32);
+```
+
+Finally, the existing `scales/zero_points` attributes and the new
+`dynamic_scales` attribute would be mutually exclusive. Hence only one
+quantization scheme would be accepted for a given memory argument.
+Also, this allows to reuse `DNNL_ARG_ATTR_SCALES` argument kind mask
+to pass dynamic scales buffer to the `execute` function.
+
+If not mutually exclusive, we would need to expose a new argument kind
+`DNNL_ARG_ATTR_DYNAMIC_SCALES` for users to specify the memory
+argument for writing computed scales.
+
+To summarize, here is an example of this new API usage with mxfp4:
+``` C++
+    // Elements type for all in/out puts are e2m1 for MXFP4
+    memory::desc a_md({M, K}, memory::data_type::f4_e2m1, {K, 1}); // M x K layout
+    memory::desc b_md({K, N}, memory::data_type::f4_e2m1, {K, 1}); // N x K layout
+    memory::desc c_md({M, N}, memory::data_type::f4_e2m1, {N, 1}); // M x N layout
+
+    // Create attributes and set static scales for inputs, 
+    // with type of e8m0 and groups of 32 along K. 
+    primitive_attr attr;
+    attr.set_scales(DNNL_ARG_SRC,
+            /* mask */ (1 << 0) + (1 << 1), {1, 32}, memory::data_type::e8m0);
+    attr.set_scales(DNNL_ARG_WEIGHTS,
+            /* mask */ (1 << 0) + (1 << 1), {32, 1}, memory::data_type::e8m0);
+
+    // Set dynamic scales for output, with type of e8m0 and groups of 32 along K.
+    attr.set_dynamic_scales(DNNL_ARG_DST,
+            /* mask */ (1 << 0) + (1 << 1), {1, 32}, memory::data_type::e8m0);
+
+    // Create a MatMul primitive descriptor
+    matmul::primitive_desc(eng, a_md, b_md, c_md, attr);
+```
+
+### Option 2: Through new datatypes 
+
+Another option could be to expose new datatype (e.g. `mxfp8_e4m3`,
+`mxfp8_e5m2`, `mxfp4_e2m1`, ...). Those would encode group element
+datatype, but also scale datatype, group size (32 in case of MX
+types), and axis along which scales apply (e.g. innermost physical
+dimension).
+
+A new argument kind `DNNL_ARG_MX_SCALES`, will be used to
+specify the memory argument for reading/writing scales.
+
+This option is not recommended for a couple of reasons:
+- it is not aligned with existing quantization support in oneDNN,
+  providing two very different ways to specify quantization of a
+  memory object.
+- it lacks flexibility: user cannot specify along which axis
+  quantization would happen (it would be only physical innermost
+  axis). Furthermore, any new combination of scale datatype / group
+  size / group element type would require a new datatype in oneDNN.
+
+### Testing
+
+benchdnn will be extended according to the proposal adopted:
+- if new attribute, a new `--attr-dynamic-scales` knob will be
+  exposed.
+- if new datatype, a new datatype will be available to be passed to
+  `--dt` knob.
+
+## Summary
+
+oneDNN already support MXFP inputs, as it supports base datatypes, as
+well as grouped scales of type e8m0. To support MXFP outputs, we
+recommend introducing a new attribute (set_dynamic_scales), which
+would allow the user to specify that oneDNN will compute the scaling
+factors and apply them upon output conversion to MXFP type. oneDNN
+will also provide a new argument kind for user to collect the scales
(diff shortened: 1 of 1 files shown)


mgouicem:
@Sqvid @theComputeKid

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+dimension).
+
+A new argument kind `DNNL_ARG_MX_SCALES`, will be used to
+specify the memory argument for reading/writting scales.
+
+This option is not recommended for a couple of reasons:
+- it is not aligned with existing quantization support in oneDNN,
+  providing two very different ways to specify quantization of a
+  memory object.
+- it lacks flexibility: user cannot specify along which axis
```
jondea, line 193: Isn't this lack of flexibility intentional though? The scale is shared between physically contiguous elements so that it can be efficiently operated on by vector instructions.

A related question that I can't seem to find the answer to: are the scales typically stored inline after the blocks of 4 bit numbers or in a separate buffer? I think this answer informs the decision

dzarukin: > A related question that I can't seem to find the answer to: are the scales typically stored inline after the blocks of 4 bit numbers or in a separate buffer?

If I read it correctly, the lines 52-53 have it:

> We will also assume that scales and data are not interleaved
> in memory, similarly to current quantization support.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+```c++
+    void set_dynamic_scales(int arg, int mask, const memory::dims &groups,
+            memory::data_type data_type = memory::data_type::f32);
+```
+
+Finally, the existing `scales/zero_points` attributes and the new
+`dynamic_scales` attribute would be mutually exclusive. Hence only one
+quantization scheme would be accepted for a given memory argument.
+
+A new argument kind `DNNL_ARG_ATTR_DYNAMIC_SCALES`, will be used to
```
dzarukin: I'm not sure what's the additional benefit this argument provides over the existing one...
The only purpose of the argument is to specify the entry in the user args. Given that (input) scales and dynamic (output) scales are mutually excluded, it's fine to fetch the memory over the same `DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST` and either read from it (if `set_scales` were used) or write into it (if `set_dynamic_scales` were used).

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+dnnl_status_t DNNL_API dnnl_primitive_attr_set_dynamic_scales(
+        dnnl_primitive_attr_t attr, int arg, int mask, int ndims,
+        const dnnl_dims_t group_dims, dnnl_data_type_t data_type);
+```
+
+And the associated symbol in C++ API:
+```c++
+    void set_dynamic_scales(int arg, int mask, const memory::dims &groups,
+            memory::data_type data_type = memory::data_type::f32);
+```
```
dzarukin, line 88: It would be nice to have an example of how it'll be used for an expected common scenario, like what mask and groups would look like.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+In order to support MXFP formats, it needs to support groups of 32,
+and a way to express along which axis the grouping occurs. It also
+needs to support E8M0 format for scales. We propose to use the same
+semantics as with `set_scales` method:
+- grouping factor can be made explicit with mask, ndims and group_dims
+  arguments,
+- E8M0 scale datatype can be set explicitely with a scale data_type
+  argument.
+- regarding the scale computation formula, we would not expose a knob
+  to configure it, and support only MXFP spec formula.
```
dzarukin, line 75: What's the plan to address this if the spec changes?

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+## MX Dynamic quantization definition
+
+How the scales are computed is defined by the MXFP spec. For a given
+1d block of size 32 (denoted $`X_i`$), its scale and quantized values
+are defined as:
+
+$$ S = E8M0(amax(X)) / E8M0(max(dst_dt)) $$
+
+With `E8M0` a conversion to E8M0 datatype, by rounding down to the
+closest power of two below the argument value.
```
dzarukin: Given that the API provides a way to configure scale data type, what's the general formula for a random data type?

rfcs/20250911-mxfp-support/README.md
jondea: Thanks, I missed that

mgouicem: > Isn't this lack of flexibility intentional though? The scale is shared between physically contiguous elements so that it can be efficiently operated on by vector instructions.

You are correct that it can be beneficial to expose only flavors that we can accelerate.
On the other hand:
- some backend might be able to accelerate the case where scale is not applied on contiguous data (e.g. if they have transpose hardware acceleration).
- the number of combination to expose might be large (cartesian product of scales type, block size, element type). We are already seeing that evolve beyond MX types with nvfp4 for example.
- different backends might have different flavors (e.g. block size of 32, 16, ...). So we might still end up exposing a large set of configurations, not all accelerated by each backends.

Also, using datatype to encode scaling might bring some ambiguities with existing scale attribute for inputs, both externally (mx input is actually a subset of what `set_scales` supports), and internally if we don't "unify" quantization parameters with input datatype.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
+# Open Compute Microscaling (MX) datatype suport
```
tkrupa-intel:
```suggestion
# Open Compute Microscaling (MX) datatype support
```

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+  conversion to scale datatype
+
+## Proposals
+
+First of all, this RFC focuses on MXFP formats supports but might
+cover some generalizations to allow extension to other formats and
+recipes.  We will also assume that scales and data are not interleaved
+in memory, similarly to current quantization support.
+
+### Opetion 1 (Recommended): Through a new attribute (`set_dynamic_scales`)
```
tkrupa-intel:
```suggestion
### Option 1 (Recommended): Through a new attribute (`set_dynamic_scales`)
```

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+benchdnn will be extended according to the proposal adopted:
+- if new attribute, a new `--attr-dynamic-scales` knob will be
+  exposed.
+- if new datatype, a new datatype will be available to be passed to
+  `--dt` knob.
+
+## Summary
+
+oneDNN already suport MXFP inputs, as it supports base datatypes, as
```
tkrupa-intel:
```suggestion
oneDNN already supports MXFP inputs, as it supports base datatypes, as
```

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,144 @@
...
+interchange formats as well as operations (dot-product and
+conversions).  In particular, MXFP4 is currently used in gpt-oss [^2]
+for weights compression.
+
+In this RFC, we go over what are the MX formats, and what oneDNN need
+to support those.
+
+## MX formats and requirements to oneDNN
+
+Even though though they are sometimes refered to as datatypes, MX
```
tkrupa-intel:
```suggestion
Even though though they are sometimes referred to as datatypes, MX
```

rfcs/20250911-mxfp-support/README.md
mgouicem: Good point. The main point here was to expose `DNNL_ARG_ATTR_DYNAMIC_SCALES` for "explicitness". But I agree that we can reuse `DNNL_ARG_ATTR_SCALES` if both scales and dynamic_scales are mutually exclusive. I clarified the RFC on that point.

rfcs/20250911-mxfp-support/README.md
mgouicem: Added.

rfcs/20250911-mxfp-support/README.md
mgouicem: We would likely have to extend the API with a knob to select formula to use (mx_v1, mx_v2, ...).

rfcs/20250911-mxfp-support/README.md
mgouicem: Anytime number of mantissa bits is greater than 0 (so anything other than e8m0), computing the division in f32 first, and convert to scale datatype second makes more sense. This is widely used convention in existing frameworks. However, we are never safe from new applications starting to compute those differently.

Note that MXFP does the conversion first because:
- it is limited to e8m0 scales
- it relies on the property that division in E8M0 is cheap (just a subtraction)

rfcs/20250911-mxfp-support/README.md
mgouicem: Applied, thanks

rfcs/20250911-mxfp-support/README.md
mgouicem: Applied, thanks

rfcs/20250911-mxfp-support/README.md
mgouicem: Applied, thanks

rfcs/20250911-mxfp-support/README.md
mgouicem: Applied, thanks

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,169 @@
...
+  and main point of discussion of this RFC.
+
+## MX Dynamic quantization definition
+
+How the scales are computed is defined by the MXFP spec. For a given
+1d block of size 32 (denoted $`X`$), its scale `S` and quantized values `Q`
+are defined as:
+
+$$ S = E8M0(amax(X)) / E8M0(MAX\_DST\_DT) $$
+$$ Q = dst\_dt(X / S) $$
```
TaoLv:
```suggestion
$$ S = E8M0(amax(X)) / E8M0(MAX\\_DST\\_DT) $$
$$ Q = dst\\_dt(X / S) $$
```
Just to fix a rendering problem on GitHub. Now it looks like:

$$ S = E8M0(amax(X)) / E8M0(MAX\\_DST\\_DT) $$
$$ Q = dst\\_dt(X / S) $$

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,169 @@
...
+```C
+dnnl_status_t DNNL_API dnnl_primitive_attr_set_dynamic_scales(
+        dnnl_primitive_attr_t attr, int arg, int mask, int n dims,
+        const dnnl_dims_t group_dims, dnnl_data_type_t data_type);
+```
+
+And the associated symbol in C++ API:
+```c++
+    void set_dynamic_scales(int arg, int mask, const memory::dims &groups,
+            memory::data_type data_type = memory::data_type::f32);
```
TaoLv: Is it the scale data type? Why defaults to f32? I see it's E8M0 scale described above.

mgouicem: Thanks, corrected.

rfcs/20250911-mxfp-support/README.md
mgouicem: I added an option with enum to cover different quantization flavors (including static). Let me know what you think.

rfcs/20250911-mxfp-support/README.md
jondea: That makes sense, thank you!

mgouicem:
Hi all, just pushed an update to the RFC. In a nutshell:
- added link to [POC PR](https://github.com/uxlfoundation/oneDNN/pull/3978)) for option 1.b (extend set_scales)
- updated recommendation to option 1.b (extend set_scales)

The main driver to now recommend extending set scales are:
- it allows to unify scales handling both for external API and internally, making different ways of handling scales explicitly mutually exclusive
- it should be more robust for extending to new quantization flavors (e.g. static quantization with floating-point zero-points, or dynamic quantization with division by scale before conversion).

Let me know if there are preferences or other opinions on this.
Thanks.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+        dnnl_primitive_attr_t attr, int arg, int mask, int ndims,
+        const dnnl_dims_t group_dims, dnnl_data_type_t data_type,
+        int is_on_host, dnnl_quantization_kind qkind);
+```
+
+In C++ API, we would introduce similar enum type and extension to `set_scales`
+
+```C++
+enum class quantization_kind {
+    undef = dnnl_quantization_kind_undef,
```
mzhukova, line 157: just wondering, what is the purpose of `..._undef` since `..._static` seem to be a default (== current behavior)?

mgouicem: This can be useful to differentiate between when scales are not set, and when they are explicitly set (in which case default is static_). So if we expose a getter, returning undef will be valid, whereas returning _static when scales are not set is kind of incorrect.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+### Option 1.b (recommended): extend scales attribute (`set_scales`)
+
+Same principle as option 1.a. except that instead of exposing new
+attribute, `set_scales` would be extended to support dynamic
+scaling. Main benefit here would be that:
+- the API itself would make it clear that static and dynamic
+quantization are mutually exclusive.
+- it would allow to extend to other quantization formulas like static
+  quantization with floating point zero-point as $`x_{f32} =
+  scale*x_q - zp`$ instead of $`x_{f32} = scale * (x_q - zp)`$, or
```
dzarukin, line 132: I'm not sure how important it is or correct it is, but I recall this:
```suggestion
  scale*x_q + zp`$ instead of $`x_{f32} = scale * (x_q - zp)`$, or
```

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+If not mutually exclusive, we would need to expose a new argument kind
+`DNNL_ARG_ATTR_DYNAMIC_SCALES` for users to specify the memory
+argument for writing computed scales.
+
+To summarize, here is an example of this new API usage with mxfp4:
+``` C++
+    // Elements type for all in/out puts are e2m1 for MXFP4
+    memory::desc a_md({M, K}, memory::data_type::f4_e2m1, {K, 1}); // M x K layout
+    memory::desc b_md({K, N}, memory::data_type::f4_e2m1, {K, 1}); // N x K layout
```
dzarukin, line 104:
```suggestion
    memory::desc b_md({K, N}, memory::data_type::f4_e2m1, {1, K}); // N x K layout
```

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+    attr.set_scales(DNNL_ARG_WEIGHTS,
+            /* mask */ (1 << 0) + (1 << 1), {32, 1}, memory::data_type::e8m0);
+
+    // Set dynamic scales for output, with type of e8m0 and groups of 32 along K.
+    attr.set_dynamic_scales(DNNL_ARG_DST,
+            /* mask */ (1 << 0) + (1 << 1), {1, 32}, memory::data_type::e8m0);
+
+    // Create a MatMul primitive descriptor
+    matmul::primitive_desc(eng, a_md, b_md, c_md, attr);
+```
```
dzarukin, line 121: Would be nice to see the same example with option 1b for everyone's clarify.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+        const dnnl_dims_t group_dims, dnnl_data_type_t data_type,
+        int is_on_host, dnnl_quantization_kind qkind);
+```
+
+In C++ API, we would introduce similar enum type and extension to `set_scales`
+
+```C++
+enum class quantization_kind {
+    undef = dnnl_quantization_kind_undef,
+    static_ = dnnl_quantization_kind_static,
```
dzarukin, line 158: ~~Quick search suggests a `PTQ` name standing for "post-training quantization". We may expand it to "static" quantization in the documentation section.~~

It seems you've solved this, you may want to update the text to reflect that.

rfcs/20250911-mxfp-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+    dynamic_mx = dnnl_quantization_kind_dynamic_mx,
```
dzarukin, line 159: I wonder if `dynamic_mx` is precise enough to describe the intent when it comes to fp4 data type and if it requires more words in the name to distinct it from a similar but different flavor.

mgouicem: It should be independent of elements datatype. Here it is about OCP MX spec definition of how scales are computed.

Sqvid:
I don't think there are any major comments on our end.
Out of interest, here is a link to some similar work that has gone into the TOSA specification. Any future work regarding TOSA and oneDNN interacting would be aided by similar numerical models.

https://git.mlplatform.org/tosa/specification.git/commit/?id=063846a75b9687ab01e58cb3538472bffb3a03b0

openmodelzooci:

vpirogov (approved):

vpirogov (approved):

PR #3937, opened 2025-09-12
xe: jit: ir: layout simplification part 1: the unconstructoring

rjoursler:
This PR largely focuses on removing `layout_t` constructors to enable simpler usage.

### DIFF
--- a/src/gpu/intel/conv/jit/plan.cpp
+++ b/src/gpu/intel/conv/jit/plan.cpp
@@ -920,7 +920,7 @@ void fma_plan_t::set_split(abc_kind_t abc, int factor) {
                     fma_kind_t::dpasw)) {
         auto blocks = a_layout.blocks();
         blocks.back().block /= factor;
-        auto layout = layout_t(a_layout.type(), a_layout.ndims(), 0, blocks);
+        auto layout = layout_t(a_layout.type(), blocks, 0, a_layout.ndims());
         m_blk = get_dpas_block_rcount(layout, 1);
     }
 }
@@ -1344,7 +1344,7 @@ struct fma_context_t {
             auto bmnk_layout
                     = mapper.map_to_bmnk(abc, bmnks, layout).retype(type);
             auto fma_layout = bmnk_layout.make_with_block(
-                    layout_t(type, bmnks.size(), 0, blocks));
+                    layout_t(type, blocks, 0, bmnks.size()));
             auto abc_layout
                     = mapper.map_from_bmnk(abc, bmnks, fma_layout, layout);
             if (cvt_f16) return abc_layout.retype(type_t::f16());
@@ -1363,7 +1363,7 @@ struct fma_context_t {
             auto bmnks = get_bmnk_kinds(abc, /*with_batch=*/true);
             auto bmnk_layout = mapper.map_to_bmnk(abc, bmnks, ret);
             auto fma_layout = bmnk_layout.make_with_block(
-                    layout_t(ret.type(), (int)bmnks.size(), 0, blocks));
+                    layout_t(ret.type(), blocks, 0, (int)bmnks.size()));
             auto abc_layout = mapper.map_from_bmnk(abc, bmnks, fma_layout, ret);
             if (layout.type().is_x8()) {
                 gpu_assert(abc_layout.type().is_s16());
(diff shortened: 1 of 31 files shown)


rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_pool
enable benchdnn_reduce
enable benchdnn_reorder

echeresh (approved):

rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_pool
enable benchdnn_reduce
enable benchdnn_reorder

src/gpu/intel/conv/jit/v2/kernel_desc.cpp
```diff
@@ -718,7 +718,8 @@ tensor_config_t get_tensor_config(
         if (!is_input && !is_output) continue;
         int key = h.key(t);
         tensor_cfg.add_tensor(t, key, is_input, is_output,
-                pd ? jit::layout_t(pd->arg_md(key)) : jit::layout_t());
+                pd ? jit::make_layout(*pd->arg_md(key), true)
+                   : jit::layout_t());
```
hidefromkgb, line 722: Should we maybe have a dedicated zero-argument `jit::make_layout()`?

src/gpu/intel/conv/jit/plan.cpp
```diff
@@ -1624,7 +1623,7 @@ layout_t to_reduce_layout(
         reduce_blocks.push_back(bb);
     }
     auto type = get_accumulation_type(cfg, layout.type(), layout.type());
-    return layout_t(type, reduce_ndims, 0, reduce_blocks).make_dense();
+    return layout_t(type, reduce_blocks, 0, reduce_ndims).make_dense();
```
hidefromkgb, line 1626: Is it possible to infer ndims from blocks?

src/gpu/intel/conv/jit/zp_plan.cpp
```diff
@@ -134,14 +132,14 @@ class split_dispatcher_t {
                     && ends_with(c_layout, c_g_idx_, b)) {
                 simd_dim_idx_ = 1;
                 simd_ = b;
-                simd_tile_ = get_simd_tile(c_layout);
+                simd_tile_ = {{simd_dim_idx_, simd_}};
```
hidefromkgb, line 135: Should we maybe keep `get_simd_tile` so that the code would look more uniform? See [L622](3937/files#diff-04c5b9ad835d21302cdd10b58dba049daad15506dc24e73c49f0a71f8e6b6089R622).

hidefromkgb (approved):

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -535,10 +526,6 @@ class layout_t {
         return oss.str();
     }
 
-    IR_DEFINE_DUMP()
```
hidefromkgb, line 538: What is the new way to dump a layout in GDB?

src/gpu/intel/conv/jit/plan.cpp
rjoursler: In the past, I have attempted to remove `reduce_ndims`, but it results in correctness errors. As such, this is necessary at the moment. Longer term, the `ndims` related functionality should be removed from `layout_t`.

src/gpu/intel/conv/jit/v2/kernel_desc.cpp
rjoursler: No, that just seems to introduce unnecessary indirection.

src/gpu/intel/conv/jit/zp_plan.cpp
rjoursler: With these modification, `c_layout` is not needed as an input to `get_simd_tile`, so it should be removed.  When this is done, we end up with a name conflict with the `simd_tile_t` [getter](https://github.com/uxlfoundation/oneDNN/blob/7db067420af9c53cbc704bc06d44a62221012ab0/src/gpu/intel/conv/jit/zp_plan.cpp#L160).

src/gpu/intel/jit/ir/tensor.hpp
rjoursler: In my common usage, I will just run
```
print layout.str()
```
We can definitely add back in the `dump()` interface later though. The main issue for this removal is to avoid the `IR_DEFINE_DUMP` macro in stable DSL interface.

echeresh: @rjoursler I think the main usage for `dump()` functionality was multi-line output which doesn't look nice with `print` in gdb. This includes `stmt_t`, `tensor_t`, etc. As a quick fix I would just migrate the macro (`DSL_DEFINE_DUMP`) - it doesn't seem to make the interface less stable, IMO.

rjoursler:
make test
disable test_device_cpu
disable test_device_gpu

PR #3938, opened 2025-09-13
x64: brgemm matmul: update fp8 dispatching

ankalinin:
I suppose fp8 matmul should not take into account the condition of small size. Since the lower implementations go through fp8 -> f16 conversion and f16 calculation.
Example: 
--matmul --dt=f8_e4m3:f8_e4m3:f8_e4m3 32768x13:13x512 goes to avx10_1_512_amx_fp16 instead of avx10_2_512_amx_2

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -1665,6 +1665,8 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
         const dim_t buffer_a_chunk_sz_limit = 126;
         is_small_shapes = is_small_shapes
                 && bgmmc.buffer_a_gb_stride <= buffer_a_chunk_sz_limit;
+    } else if (bm_conf_utils.is_f8() || bm_conf_utils.is_bf8()) {
+        is_small_shapes = false;
     } else {
         is_small_shapes = is_small_shapes && bgmmc.ndims < 3
                 && ((bgmmc.M == 1 && bgmmc.K == 256)


ankalinin:
make test
set test_scope=NIGHTLY

disable compiler_gnu9
disable compiler_icx-previous
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_cmake_RelWithMDd
disable test_device_gpu

disable benchdnn_all
enable benchdnn_matmul

disable arch_cpu_hsw
disable arch_cpu_slx

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1665,6 +1665,8 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
         const dim_t buffer_a_chunk_sz_limit = 126;
         is_small_shapes = is_small_shapes
                 && bgmmc.buffer_a_gb_stride <= buffer_a_chunk_sz_limit;
+    } else if (bm_conf_utils.is_f8() || bm_conf_utils.is_bf8()) {
+        is_small_shapes = false;
```
georgen117, line 1669: Since this will run on fp16 compute shouldn't the `is_small_shapes` follow the same logic as the is_f16 above?

ankalinin: I think no. We're making this `is_small_shapes = false` unconditional for fp8 now. In the future, when real hardware becomes available, we can test performance more thoroughly and change this condition if necessary.

PR #3939, opened 2025-09-14
rfcs: graph api: support Dropout operation

ElaineBao:
# Description

The RFC proposes adding a Dropout operation in oneDNN Graph.

Rendered version: [link](https://github.com/uxlfoundation/oneDNN/blob/yixin/rfc/dropout/rfcs/20250910-graph-api-dropout/README.md)

### DIFF
--- a/rfcs/20250910-graph-api-dropout/README.md
+++ b/rfcs/20250910-graph-api-dropout/README.md
@@ -1,0 +1,183 @@
+# Support Dropout Operation in Graph API
+
+## Background
+
+Dropout [[#1]][1] is a regularization technique widely adopted in deep learning
+models, including transformers and convolutional neural networks. It helps
+prevent overfitting by randomly zeroing out elements of the input tensor during
+training and scaling the outputs to maintain the expected sum. During inference,
+Dropout is typically bypassed, and the input is passed through unchanged.
+
+Mathematically, Dropout is defined as follows:
+
+**Forward (Training)**
+
+Given input tensor $X$ and dropout probability $p$:
+
+For each element $i$:
+
+> $$
+> M_i \sim \text{Bernoulli}(1-p)
+> $$
+> $$
+> Y_i = \frac{X_i \cdot M_i}{1-p}
+> $$
+
+where $M$ is the mask tensor (same shape as $X$), and $Y$ is the output tensor.
+
+**Forward (Inference)**
+
+Dropout is typically bypassed during inference:
+
+> $$
+> Y_i = X_i
+> $$
+
+**Backward**
+
+Given gradient tensor $dY$, mask tensor $M$, and dropout probability $p$:
+
+> $$
+> dX_i = \frac{dY_i \cdot M_i}{1-p}
+> $$
+
+where $dX$ is the gradient with respect to the input.
+
+## Adoption in Frameworks and Libraries
+
+**PyTorch**
+
+troch.nn.Dropout [[#2]][2] takes `p` (probability) and `inplace` as attributes,
+with one input and one output. Internally, the random number generator used to
+generate the dropout mask differs by device:
+
+- CPU: Uses torch.Generator backed by the C++ Mersenne Twister engine [[#3]][3],
+  which consumes a 64-bit seed.
+- CUDA / XPU: Uses a 64-bit Philox RNG state (seed + offset) obtained from the
+  device generator [[#4]][4][[#5]][5]. The offset is advanced after each kernel
+  launch to carve out disjoint subsequences of random numbers.
+
+**TensorFlow**
+
+tf.keras.layers.Dropout [[#6]][6] takes `rate` (probability), an optional `noise_shape`
+(shape for randomly generated keep/drop flags) and an optional seed as
+attributes, with one input and one output. Internally, the random number
+generator produces a 64-bit pair (key, counter) from the seed and passes
+them to a selected RNG algorithm to generate the mask.
+Supported algorithms [[#7]][7] include Philox, ThreeFry, or AUTO_SELECT
+(which lets TensorFlow choose the best algorithm for the current device;
+AUTO_SELECT may pick an algorithm other than Philox or ThreeFry).
+
+**ONNX**
+
+ONNX's Dropout operation [[#8]][8] takes `data`, `ratio` (probability) and `training_mode`
+as inputs, and `seed` as an attribute. Outputs include the dropped tensor and
+the dropout mask. ONNX does not specify the algorithm used to generate random numbers.
+
+**cuDNN**
+
+cuDNN Graph library does not provide a dedicated Dropout operation. Instead, it
+introduces an RNG operation, which can be used to implement Dropout with two
+pointwise Multiply operations.
+The RNG operation takes `seed`, `offset` as inputs, supports various
+distributions (Bernoulli, Uniform, Normal) by setting attributes, and outputs
+the generated tensor. The RNG is based on the Philox algorithm.
+
+**oneDNN**
+
+In oneDNN primitive, Dropout is implemented as a binary post-op applied to
+output tensor before other post-ops. It uses the Philox algorithm and takes
+`seed` (s32) and `probability` (f32) as inputs, and outputs a `mask` (u8).
+
+## Proposal
+
+Based on the investigation above, we propose to define the Dropout operation in
+the oneDNN Graph API as follows:
+
+**Operation Kind:**
+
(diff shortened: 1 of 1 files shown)


rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+- Dropout will use the Philox algorithm for random number generation.
+- Data types for `src`, `p`, `dst` and `mask` are consistent with the
+  current oneDNN primitive design.
+- The `mask` output is optional, allowing flexibility for scenarios where
+  the mask is not needed.
+- The `offset` input is introduced to ensure reproducibility and correct
+  partitioning across devices and threads, following best practices from PyTorch
+  and cuDNN.
+- Data types for `seed` and `offset` are aligned to PyTorch to ensure
```
TaoLv: How about the data types of probability and mask in PyTorch?

rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,183 @@
...
+  compatibility and support large models.
+- A dedicated DropoutBackward operation is not defined, as the backward
```
TaoLv: "not defined" or "not needed"?

rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+```cpp
+
+using namespace dnnl::graph;
+
+graph g = graph(engine::kind::gpu);
+
+logical_tensor src = logical_tensor(ID_SRC, data_type::f16, {256, 1000});
+logical_tensor seed = logical_tensor(ID_SEED, data_type::u64, {1});
+logical_tensor offset = logical_tensor(ID_SEED, data_type::u64, {1});
```
TaoLv:
```suggestion
logical_tensor offset = logical_tensor(ID_OFFSET, data_type::u64, {1});
```

rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,183 @@
...
+**Operation Kind:**
+
+- `Dropout` (C++), `dnnl_graph_op_dropout` (C)
+
+**Inputs:**
+
+- `src`: input tensor (`f32` / `bf16` / `f16`)
+- `seed`: random seed (`uint64_t`, single value)
+- `offset`: random offset (`uint64_t`, single value)
+- `p`: dropout probability (`f32`, single value)
```
TaoLv, line 106: Does that mean seed, offset, and p will be changed among different iterations?

rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,183 @@
...
+dropout.add_input(p);
+dropout.add_output(dst);
+dropout.add_output(mask);
+
+g.add_op(dropout);
+g.finalize();
+
+```
+
+**Open Question:**
```
TaoLv, line 170: Random place: please consider and document how it will be tested, especially about the random-ness part.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: I think it's not defined. We always have an option to define a dedicated DropoutBackward operation to make it more clear that forward uses Dropout, backward uses DropoutBackward, although underneath the math formula is the same.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: probability is float32,
for mask, the aten sdpa doesn't define an output of mask. the aten dropout has this output, it supports u8 and boolean dtype of mask.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: p and seed most likely will not change among different iterations, only offset will change

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: Thanks for the suggestion, will do.

rfcs/20250910-graph-api-dropout/README.md
TaoLv: Then why do we put p and seed as inputs, not attributes? Are they on device memory?

rfcs/20250910-graph-api-dropout/README.md
TaoLv: Since you explicitly mentioned that `Data types for seed and offset are aligned to PyTorch`. It makes me wondering if probability and mask are not aligned.

> the aten dropout has this output, it supports u8 and boolean dtype of mask.

Better to mention this in the RFC as well and why we choose `u8`, not `boolean`. What if we are asked to support `boolean` mask?

rfcs/20250910-graph-api-dropout/README.md
TaoLv: Thank you for the explain. It still sounds like "not needed". Probably we will not need to define two operations to implement the same math formula. Just like we don't define backward operation for MatMul.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: seed is still an input in framework side, it can be changed if user wants to, but this use scenario is not common. That's why seed is kept as an input.
for offset, I left it as an open question.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: ok, I'll mention it, thanks for the suggestion. The reason why mask is chosen as u8 is to facilitate integration between the graph API and primitive API, as primitive API already supports u8 mask. We can support boolean if it's requested in the future.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: ok, let me clarify it.

rfcs/20250910-graph-api-dropout/README.md
```diff
@@ -0,0 +1,197 @@
...
+## Adoption in Frameworks and Libraries
+
+**PyTorch**
+
+troch.nn.Dropout [[#2]][2] takes `p` (probability, f32) and `inplace` (bool) as attributes,
+with one input and one output. Internally, the random number generator used to
+generate the dropout mask (u8/bool) differs by device:
+
+- CPU: Uses torch.Generator backed by the C++ Mersenne Twister engine [[#3]][3],
+  which consumes a 64-bit seed.
```
mgouicem, line 55: This is interesting, does it mean that PyTorch deterministic mode is not supported on CPU by default? Or that dropout has to run sequentially?

rfcs/20250910-graph-api-dropout/README.md
mgouicem: If we stick with Philox, the generation is actually deterministic: for a given output element at index idx, the generated rng value is Philox(seed, offset + idx).

The thing to be careful with is that the flattened `idx` should be consistent between reference and optimized impl (physical offset to base ptr should do).

rfcs/20250910-graph-api-dropout/README.md
mgouicem: Extra things to consider here wrt input vs attribute: using attribute will also trigger extra jitting and increase cache occupancy. So unless we jit completely different code based on p and seed (which we don't), it might be beneficial to keep those as inputs.

If those are never on device, would it make sense to only support host_side_scalar for those?

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: Yes,  the random number generator used for Dropout CPU is the default PyTorch CPU RNG, which is non determinitic (evidence: [link](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/CPUGeneratorImpl.cpp#L57)), so If users run dropout (or any random op) on CPU in a single thread and set the seed, they can get deterministic results.
If they use multiple threads, the order of RNG calls can vary, so results may not be deterministic across runs, even with the same seed.

rfcs/20250910-graph-api-dropout/README.md
ElaineBao: Yes, I think supporting host scalar is enough for seed, offset and p.

TaoLv (approved):

vpirogov (approved):

PR #3943, opened 2025-09-15
ci: aarch64: update wiki from ci

Ryo-not-rio:
Add functionality to update the wiki from the nightly aarch64 ci with unit test and performance test results.
The generated wiki can be seen here: https://github.com/uxlfoundation/oneDNN/wiki/AArch64-status

### DIFF
--- a/.github/automation/aarch64/wiki_utils.py
+++ b/.github/automation/aarch64/wiki_utils.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python
+
+# *******************************************************************************
+# Copyright 2025 Arm Limited and affiliates.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+import argparse
+import ctest_utils
+import os
+
+PAGE_TITLE = "AArch64 Testing Status"
+
+
+class MdConverter:
+    # Recursive function to convert dict to md sections
+    def __unpack_md_dict(self, data, out_str="", level=1):
+        # base case
+        if not isinstance(data, dict):
+            return out_str + data + "\n"
+
+        for k, v in data.items():
+            out_str += "#" * level + " " + k + "\n"
+            out_str = self.__unpack_md_dict(v, out_str, level + 1)
+        return out_str
+
+    # Recursive function to convert md to dict
+    def __unpack_md(self, data, level=1):
+        prev = ""
+        inner = []
+        out = {}
+        for l in data:
+            if l[:level] == "#" * level and l[level] != "#":
+                if prev:
+                    out[prev[level + 1 :]] = self.__unpack_md(inner, level + 1)
+                    inner = []
+                prev = l.strip()
+            else:
+                inner.append(l)
+        if prev:
+            out[prev[level + 1 :]] = self.__unpack_md(inner, level + 1)
+        elif inner:
+            # base case
+            return "".join(inner)
+        return out
+
+    def dict2md(self, in_dict, out_file):
+        output = self.__unpack_md_dict(in_dict)
+        with open(out_file, "w") as f:
+            f.write(output)
+
+    def md2dict(self, in_file):
+        if not os.path.isfile(in_file):
+            return {}
+
+        with open(in_file) as f:
+            r = f.readlines()
+
+        out = self.__unpack_md(r)
+
+        return out
+
+
+def parse(file, title, subtitle, body):
+    """
+    Add a new section/subsection or an existing section/subsection
+    without overwriting existing section/subsections
+    """
+    converter = MdConverter()
+    d = converter.md2dict(file)
+    if PAGE_TITLE not in d:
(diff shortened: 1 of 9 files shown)


vpirogov:
Cool

.github/automation/aarch64/wiki_utils.py
```diff
@@ -0,0 +1,149 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+import argparse
+import ctest_utils
+import os
+
+PAGE_TITLE = "AArch64 Status"
```
vpirogov: (nit) Generated page has heading duplicating the page name:
image

I would also suggest renaming the page to "AArch64 test results" or something similar for clarity.

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -128,38 +130,47 @@ def compare_two_benchdnn(file1, file2, check=False):
...
+        if (
+            out_file is not None
+            and abs((r2_med_exec - r1_med_exec) / r1_med_exec) >= 0.05
+        ):
+            colour = "green" if r1_med_exec >= r2_med_exec * 1.05 else "red"
+            speedup_string = (
+                "$${\\color{"
+                + colour
+                + "}"
+                + f"{(r1_med_exec)/r2_med_exec:.3g}x"
```
jondea: In LaTeX, this x looks like a variable. You probably want `\times`

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -76,10 +76,12 @@ def compare_two_benchdnn(file1, file2, check=False):
...
 
     exec_failures, ctime_failures = [], []
-    if not check:
-        print(
-            "primitive,exec_base,exec_new,ctime_base,ctime_new,exec_diff,ctime_diff"
-        )
+    if out_file is not None:
+        with open(out_file, "w") as f:
+            f.write(
+                "| problem | base time(ms) | new time(ms) | speedup |\n"
```
jondea: Can we have something like `(>1 means faster)`.

Also, looking at the wiki, it's hard to understand what base and new are. Please can we have some way of labelling these? Either from the page title or the column headers?

jondea (changes requested):

.github/automation/aarch64/wiki_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def parse_perf(args):
+    with open(args.in_file) as f:
+        body = f.read()
+    parse(args.out_file, "Performance test results", args.title, body)
+
+
+def main():
+    parser = argparse.ArgumentParser(
```
jondea, line 113: Can you explain the choices behind the different formats please? There seems to be a lot of different formats and parsing between them

- We need xml because it is the only structured output format that ctest has
- We need to output markdown because that's what GitHub wiki/comments need
- Do we need to input (potentially styled) markdown? markdown is an output format, parsing it feels brittle
- Do we need JSON as another intermediate format? Do we actually have JSON, or are they just Python dicts?

.github/automation/performance/benchdnn_comparison.py
```diff
@@ -168,8 +179,8 @@ def compare_two_benchdnn(file1, file2, check=False):
     parser.add_argument("file1", help="Path to baseline result file")
     parser.add_argument("file2", help="Path to new result file")
     parser.add_argument(
-        "--check", action="store_true", help="Enable regression checks"
+        "--out-file", help="md file to output performance results to"
     )
     args = parser.parse_args()
 
-    compare_two_benchdnn(args.file1, args.file2, check=args.check)
+    compare_two_benchdnn(args.file1, args.file2, args.out_file)
```
jondea, line 206: Why do we no longer need the check functionality?

.github/automation/aarch64/wiki_utils.py
Ryo-not-rio: I've added comments explaining the first two points. We unfortunately need to input markdown since the c6g unit test, c7g unit test, and the perf test all write to the page at different times, so we neeed to make sure each of them do not overwrite the other. Thanks for pointing it out, we're only using dicts so I've renamed all mentions of json to dicts

.github/automation/performance/benchdnn_comparison.py
Ryo-not-rio: Just wanted to reduce complexity here. We run this file as the last step in the ci so we can just always run the check whether it is for a custom perf run or a nightly perf run

.github/automation/aarch64/wiki_utils.py
```diff
@@ -0,0 +1,150 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# *******************************************************************************
+import argparse
+import ctest_utils
+import os
+
+PAGE_TITLE = "AArch64 testing Status"
```
vpirogov: (nit) 
```suggestion
PAGE_TITLE = "AArch64 Testing Status"
```

.github/automation/aarch64/wiki_utils.py
jondea: Excellent, thank you. So we are only parsing so that we can make sure the page is in the right order? If so, that might be helpful to document somewhere as a short comment.

The other way to do this would be to make this the terminal node in the DAG and do all the markdown rendering at once from all raw artifacts (probably from CSVs), but it's probably marginal at this point.

Ryo-not-rio: Done!

I agree, but with using the matrix strategy in the workflow we'd have to use artefacts or some storage option to pass all the information around which just moves the complexity around elsewhere so probably not worth it at this point

PR #3946, opened 2025-09-15
brgemm: extend single scale support

dzarukin:
Fixes [MFDNN-14195](https://jira.devtools.intel.com/browse/MFDNN-14195)
* Update a test case scale value to be a pow2 value. Comparison adjustment must be really poor to make bad values to pass.

Fixes [MFDNN-14196](https://jira.devtools.intel.com/browse/MFDNN-14196)
* After [MFDNN-14141](https://jira.devtools.intel.com/browse/MFDNN-14141) was resolved, it hit a Graph case that mandates a single scale support for weights decompression case which wouldn't apply scales inside copy-kernel but will push it out to BRGeMM which wasn't known before. Thus, extending a single scale support for scenarios when src-scale is not defined (guess, till the next case).

Removed an executable mode from the file promoted some time ago.


dzarukin:
make test
disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_matmul
enable benchdnn_conv
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_deconv
set test_scope=NIGHTLY

PR #3947, opened 2025-09-16
[GPU][IR] Introduce and enforce formal pointer/reference rules

echeresh:
PR includes the following changes:

- Changed `ptr_t` behavior to take offset as the number of elements vs bytes before
    - This won't affect the existing IR-based kernels as all existing pointer operations are done over `byte` buffers
    - Switched the existing DSL GEMM implementation to use element offsets
- Added normalization for `ref_t` - to allow nesting
- Added checks/adjustment for pointer types: to ensure the number of elements is `1` (which is default)
- Added `expr_t::ptr()` to explicitly construct a pointer type expression from a variable
    - `operator[]` now can take either a shuffle expression, a pointer type or a variable/reference type
    - This is to disambiguate `var_expr[off]` - before this was `ptr_t` expression
        - However with `lval_t` we have a `ref_t` expression for the same operator: [link](https://github.com/uxlfoundation/oneDNN/blob/e621bdc3a21495ebf3ef786350c702d323e5c413/src/gpu/intel/jit/dsl/dsl.hpp#L164)
        - Now both are resolved to `ref_t` while `ptr_t` construction requires a different API: `var_expr.ptr(off)`
- Renamed `type_t::scalar()` -> `type_t::base()`
    - This is to reduce ambiguity as this method clears the attributes as well (in addition to making the type scalar)

### DIFF
--- a/src/gpu/intel/jit/dsl/dsl.cpp
+++ b/src/gpu/intel/jit/dsl/dsl.cpp
@@ -424,8 +424,7 @@ void block_send(const tensor_t &t, const global_tensor_t &g,
     auto type = g.type;
 
     v2::for_each(operation_tile, tile, [&](const icoord_t &coord) {
-        auto buffer
-                = is_prefetch ? expr_t() : t.buf[t.layout.offset<int>(coord)];
+        auto buffer = is_prefetch ? expr_t() : t.subbuf(coord);
         auto width = !w_idx.is_undef()
                 ? std::min(tile[w_idx], operation_tile[w_idx] - coord[w_idx])
                 : 1;
@@ -505,8 +504,7 @@ void block_2d_send(const conf_2d_t &conf, const tensor_t &t,
     auto tile = conf.get_tile({w_idx, h_idx});
 
     v2::for_each(operation_tile, tile, [&](const icoord_t &coord) {
-        auto buffer
-                = is_prefetch ? expr_t() : t.buf[t.layout.offset<int>(coord)];
+        auto buffer = is_prefetch ? expr_t() : t.subbuf(coord);
         int width = into<int>(
                 std::min(tile[w_idx], operation_tile[w_idx] - coord[w_idx]));
         int height = into<int>(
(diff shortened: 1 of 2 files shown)


src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -653,9 +673,9 @@ void mma(const tensor_t &C, const tensor_t &A, const tensor_t &B,
                     C.layout.type(), simd, A.layout.type(), a_stride,
                     B.layout.type(), b_stride);
 
-            auto a_off = A.layout.offset_in_bytes(base + coord);
-            auto b_off = B.layout.offset_in_bytes(base + coord);
-            auto c_off = C.layout.offset_in_bytes(base + coord);
+            auto a_off = A.layout.offset<int>(base + coord);
+            auto b_off = B.layout.offset<int>(base + coord);
+            auto c_off = C.layout.offset<int>(base + coord);
```
rjoursler: In one of my WIP branches (`rjoursle/external_ir_layout`), I am planning to introduce the following interface to `tensor_t` to avoid all these manual offset calculations.
```C++
struct tensor_t {
    ...
    expr_t subbuf(const icoord_t &coord) const {
        return buf[layout.offset<int>(coord)];
    }
...
};
```
Could you go ahead and add this interface (or some tweak of it) to your PR and remove explicit offset calculations from this file to avoid the inevitable merge conflict?

petercad:
@echeresh -- Do pointers to sub-byte types allow non-byte-aligned offsets?

echeresh:
> @echeresh -- Do pointers to sub-byte types allow non-byte-aligned offsets?

If you are asking about registers - then no, they do not. The current design follows a rule that the allowed access is only when the target region is aligned to the register boundary.

I hope all bit fiddling can be implemented at code lowering, outside of the IR/DSL - this should greatly reduce the IR/DSL implementation complexity. Do you see any issues with that?

For SLM and global memory it can be less strict though but it seems in practice we won't need that.

src/gpu/intel/jit/dsl/dsl.cpp
echeresh: Thanks, added a commit with `tensor_t::subbuf()`

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

rjoursler (approved):

PR #3945, opened 2025-09-15
xe: fix assertion with gcc 11.x

echeresh:
This is a fixup for https://github.com/uxlfoundation/oneDNN/pull/3917.

gcc 11.x compiled `ostream << dsl::idx_t` as `ostream << dsl::idx_t::operator uint32_t()` which is not correct as conversion assumes a one-letter index.

PR #3951, opened 2025-09-16
fixup: xe: fix assertion with gcc 11.x

echeresh:
Fixup for #3945 to fix dev mode build.


echeresh:
make test
disable test_device_cpu
enable test_device_gpu

PR #3952, opened 2025-09-16
[GPU] SDPA: fix ukernel bf16 to float conversion

skazakov1:
PR fixes https://jira.devtools.intel.com/browse/MFDNN-14197,

### DIFF
--- a/src/gpu/intel/sdpa/micro.cl
+++ b/src/gpu/intel/sdpa/micro.cl
@@ -14,6 +14,7 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include "gpu/intel/include/conversion.h"
 #include "gpu/intel/include/tile_ops.h"
 #include "gpu/intel/include/types_interop.h"
 #include "gpu/intel/sdpa/utils.h"
@@ -524,10 +525,10 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
 #if WITH_ATTN_SCALE
 #if WITH_HOST_SCALE
 #if INVERT_SCALE
-        iscale = SCALES_TO_FLOAT(scale_val);
+        iscale = into_float(scale_val);
         scale = native_recip(iscale);
 #else
-        scale = SCALES_TO_FLOAT(scale_val);
+        scale = into_float(scale_val);
         iscale = native_recip(scale);
 #endif
 #else
--- a/src/gpu/intel/include/math_utils.h
+++ b/src/gpu/intel/include/math_utils.h
@@ -396,8 +396,6 @@ float4  __attribute__((overloadable)) cvt_bf16_to_f32(ushort4  a) { return __bui
 float8  __attribute__((overloadable)) cvt_bf16_to_f32(ushort8  a) { return __builtin_IB_bftof_8 (as_short8 (a)); }
 float16 __attribute__((overloadable)) cvt_bf16_to_f32(ushort16 a) { return __builtin_IB_bftof_16(as_short16(a)); }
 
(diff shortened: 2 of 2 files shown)


skazakov1:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

Simonsays095:
Generally the `cvt_bf16_to_f32` function is used when we have type punning (old method), while the safer overload function `into_float` should be used when we don't use type punning (instead using the custom `bf16` struct, this is the new method). Do you know why we're mixing the two methods in this code?

skazakov1:
> Generally the `cvt_bf16_to_f32` function is used when we have type punning (old method), while the safer overload function `into_float` should be used when we don't use type punning (instead using the custom `bf16` struct, this is the new method). Do you know why we're mixing the two methods in this code?

yes: for scalar(!) kernel argument we have guard (chack_scalar_arguments(), https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/compute/kernel.hpp#L156) which failed if we do pass it as bf16 but get as ushort. extending cvt to bf16 argument seemed as the simplest way to avoid it.
in some sense it may be kind of temporal solution - I'm going to rework passing scale (when - ?) by passing TWO floats - scale and 1/scale what will remove div from kernel and bf16->float conversion.
And nevertheless I would keep this cvt_bf16_to_f32(bf16) for awhile due to "host-side scalar" activity is not completed.

Simonsays095:
Instead of adding this conversion, could we just replace the `SCALES_TO_FLOAT` with `into_float`? Since this code is only used with host-side scalars, and they're guaranteed to not use punning, I'd rather use the safer route.

Alternatively, we can define the scales type to always use punning and leave as-is - and eventually convert the kernel to use the custom types without punning all at once later.

skazakov1:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

skazakov1:
Thanks! Reworked.

echeresh (approved):

src/gpu/intel/sdpa/micro.cl
```diff
@@ -61,7 +62,11 @@ typedef ugemm_vs_c_type a_tile_type;
 #endif
 
 #ifdef SCALE_DT_BF16
+#if WITH_HOST_SCALE
+#define SCALES_TO_FLOAT into_float
+#else
 #define SCALES_TO_FLOAT cvt_bf16_to_f32
+#endif
```
Simonsays095: Sorry, I should have been more clear. I don't think we can unconditionally use `into_float`, since the pointer-based scales may still use punning. My suggestion is to do this inside the kernel:
```#if WITH_ATTN_SCALE
#if WITH_HOST_SCALE
#if INVERT_SCALE
        iscale = into_float(scale_val);
        scale = native_recip(iscale);
#else
        scale = into_float(scale_val);
        iscale = native_recip(scale);
#endif
#else
#if INVERT_SCALE
        iscale = SCALES_TO_FLOAT(*scale_ptr);
        scale = native_recip(iscale);
#else
        scale = SCALES_TO_FLOAT(*scale_ptr);
        iscale = native_recip(scale);
#endif
#endif
#endif
```

skazakov1: make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

skazakov1: Done

Simonsays095:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

PR #3954, opened 2025-09-16
Introduce GEMV code path to brgemm matmul

densamoilov:
This pull request introduces a GEMV code path for brgemm-matmul. The code path is currently enabled only for AVX2 and FP32 (see MFDNN-13362).

The primary motivation for this change is to close the performance gap between gemm-matmul and brgemm-matmul on AVX2, which is used for the FP32 data type. This improvement enables brgemm-matmul to become the default for AVX2 and FP32.

Performance data was collected on an Ultra 5 238V 2.10GHz (Lunar Lake) platform, using both Performance and Efficient cores. The inference dataset from the GNR performance validation was used. To avoid skewing results due to frequency boosting, the system was fully loaded during measurements by launching two instances of benchdnn simultaneously: one on four P-cores and the other on four E-cores. Threading type was GNU OpenMP.

This pull request also introduces a dedicated space for simple register-level operations that can be generalized and reused across different kernels. Currently, it includes one operation, `horizontal_add_ps`, which is reused in three places so far.

Based on the following performance data, brgemm-matmul already demonstrates overall better results than gemm-matmul. The main remaining performance gap was due to GEMV cases, which has now been addressed by the new GEMV code path.

While a few cases still show a gap larger than 5%, the majority are either on par or faster. Therefore, I suggest we move forward with making brgemm-matmul the default for AVX2 and address the remaining gaps in follow-up work.

In the table for GEMV case:
* gemm/brgemm (old) - brgemm without GEMV code path
* gemm/brgemm (new) - brgemm with GEMV code path

Color coding:
* Green: >=100%
* Yellow: >= 95% and < 100%
* Red: < 95%

<details> <summary> non-GEMV cases </summary>
image
image
</details>

<details> <summary> GEMV cases </summary>
image

</details>


tprimak:
@densamoilov, could you please include model-level summary as well?

src/cpu/x64/utils/jit_regops.hpp
```diff
@@ -0,0 +1,63 @@
...
+void horizontal_add_ps(
+        jit_generator_t *code, Xbyak::Zmm src, Xbyak::Zmm workspace);
+
+} // namespace regops
+} // namespace x64
+} // namespace cpu
+} // namespace impl
+} // namespace dnnl
+
+#endif
```
ankalinin, line 63: maybe just add method `horizontal_add_ps` to `jit_generator_t` ?

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -319,10 +319,16 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
...
-        CHECK(brgemm_desc_init(&brg, kernel_isa, bgmmc_.brg_type, bgmmc_.src_dt,
-                bgmmc_.wei_dt, false, false, brgemm_row_major, alpha, vbeta,
-                LDA, bgmmc_.LDB, bgmmc_.LDC, vM, vN, vK, nullptr,
-                bgmmc_.is_tf32));
+
+        if (bgmmc_.is_gemv)
+            CHECK(brgemv_desc_init(&brg, kernel_isa, bgmmc_.brg_type,
+                    bgmmc_.src_dt, bgmmc_.wei_dt, false, alpha, vbeta, LDA,
+                    bgmmc_.LDC, vM, vK));
+        else
```
ankalinin, line 327: Just a thought. Maybe we could hide the choice of a specific (brgemm or brgemv) implementation within a function `brgemv_desc_init`? Like we currently do for brgemm and brdgmm.

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -279,6 +279,27 @@ status_t brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
     return status::success;
 }
 
+status_t brgemv_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
+        brgemm_batch_kind_t type, impl::data_type_t dt_a,
+        impl::data_type_t dt_x, bool transA, float alpha, float beta, dim_t LDA,
```
georgen117, line 284: Just wondering why variable name dt_x was used instead of dt_b?
Does GEMV use a convention of `A * x = Y` instead of `A * B = C` which is typically used for GEMM?
Was this done to make it clear that the second input is a vector instead of a tensor?

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -279,6 +279,27 @@ status_t brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
...
+    // Only f32 is supported for now.
+    if (!utils::everyone_is(data_type::f32, dt_a, dt_x))
+        return status::unimplemented;
+
+    // y = x*A^t is not yet implemented.
+    if (transA) return status::unimplemented;
+
+    CHECK(brgemm_desc_init(brg, isa, type, dt_a, dt_x, false, false,
+            brgemm_row_major, alpha, beta, LDA, 1, INCY, M, 1, N, nullptr,
+            false));
```
georgen117, line 296: should this be 
```
CHECK(brgemm_desc_init(brg, isa, type, dt_a, dt_x, transA, false,
            brgemm_row_major, alpha, beta, LDA, 1, INCY, M, 1, N, nullptr,
            false));
```

src/cpu/x64/brgemm/brgemm.hpp
```diff
@@ -69,6 +69,36 @@ status_t DNNL_API brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
         dim_t LDC, dim_t M, dim_t N, dim_t K,
         const brgemm_strides_t *strides = nullptr, bool is_tf32 = false);
 
+/// Initializes a BRGEMV descriptor for batched matrix-vector multiplication.
+///
+/// @param brg     Output BRGEMM descriptor.
+/// @param dt_a    Data type of matrix A. Supported types: f32, u8, bf16, f16.
+/// @param dt_x    Data type of input vector x. Supported types: f32, s8, bf16,
+///                f16.
```
georgen117: There seems to be a disagreement between the documentation and the supported types.
Should these only list `f32` as supported since the code in `brgemm.cpp:L288` only indicates that `f32` is supported.

src/cpu/x64/brgemm/brgemm.hpp
```diff
@@ -69,6 +69,36 @@ status_t DNNL_API brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
...
+///     - If dt_a and dt_x are integer types (u8/s8), y has int32 data type.
+///     - If dt_a and dt_x are bf16, f16, or f32, y has f32 data type.
+/// @param transA  Specifies whether matrix A is transposed:
+///                'false' - A is not transposed,
+///                'true'  - A is transposed.
+/// @param alpha   Scalar multiplier alpha.
+/// @param beta    Scalar multiplier beta.
+/// @param LDA     Leading dimension of matrix A.
+///                Must be at least max(1, N).
+/// @param INCY    Stride between elements of output vector y.
```
georgen117, line 90: The `INCY` name is a little confusing. Just looking at the code it is unclear that it indicates a stride.  I assume this stands for increment Y.  If that is the case, I would suggest naming it `INC_Y` or `STRIDE_Y`.

Would it make sense to make the `INCY` a `drgemm_strides_t` to make it clear this a stride? The good reason for not making it a `drgemm_strides_t` I assume would be that `dim_t` is easier to work with. 

Nitpick: It would match the style to of `brgemm_desc_init` to place the stride after the dimension parameters, or are you thinking `INCY` should be grouped with the leading dimensions?

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -279,6 +279,27 @@ status_t brgemm_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
     return status::success;
 }
 
+status_t brgemv_desc_init(brgemm_desc_t *brg, cpu_isa_t isa,
+        brgemm_batch_kind_t type, impl::data_type_t dt_a,
+        impl::data_type_t dt_x, bool transA, float alpha, float beta, dim_t LDA,
+        dim_t INCY, dim_t M, dim_t N) {
+
```
georgen117, line 286: should there be a check added to verify that the size of vector `x` is the same as dim_t `N`? I don't know if this is the expected location for error checking or if that would belong somewhere else.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -415,6 +429,9 @@ struct jit_brgemm_kernel_t : public jit_base_brgemm_kernel_t {
     void gemm_microkernel(dim_t bd_block2, bool is_bdb_tail, dim_t ld_block,
             bool is_rd_tail, bool is_ld_tail, dim_t vpad,
             dim_t rows_for_rd_tail);
+    void gemv_microkernel(dim_t bd_block2, bool is_bdb_tail, dim_t ld_block,
+            bool is_rd_tail, bool is_ld_tail, dim_t vpad,
+            dim_t rows_for_rd_tail);
```
georgen117: This is likely a dumb question since I am less familiar with the code. Should this be named `gemv_microkernel_avx2` since it only works with avx2?  Is the kernel expected to be expanded to support other ISA so no subfix is expected.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2975,7 +3086,15 @@ void jit_brgemm_kernel_t<Wmm>::generate() {
...
-    if (!isa_has_masks(brg.isa_impl) && brg.ldb_tail > 0) {
+    if (brg.is_gemv && !isa_has_masks(brg.isa_impl) && brg.rdb_tail) {
+        L(avx_rd_tail_mask_);
+        for (dim_t i = 0; i < brg.rdb_tail; ++i)
+            dd(0xffffffff);
+        for (dim_t i = brg.rdb_tail; i < simd; ++i)
+            dd(0);
+    }
+
+    if (!isa_has_masks(brg.isa_impl) && brg.ldb_tail) {
```
georgen117: why was the line changed from `brg.ldb_tail > 0` to `brg.ldb_tail` It does not change how the code runs but I think the logic is clearer when using the original greater than

src/cpu/x64/utils/jit_regops.hpp
```diff
@@ -0,0 +1,63 @@
...
+namespace regops {
+
+/**
+ * Horizontally sums packed fp32 values in a 128-bit XMM register.
+ * @param code JIT code generator instance.
+ * @param src Source XMM register.i
+ * @param workspace Optional dummy register (ignored).
+ */
+void horizontal_add_ps(jit_generator_t *code, Xbyak::Xmm src,
+        Xbyak::Xmm workspace = Xbyak::Xmm {});
```
georgen117, line 35: personal preference would be to name this `reduce_add_ps` however this is not strong preference This is replacing the `horizontal_sum` and `reduce`. For that reason I am all for adding this utilis.

Should jit_getops.hpp/cpp be in its own PR?

src/cpu/x64/utils/jit_regops.hpp
```diff
@@ -0,0 +1,63 @@
...
+#ifndef CPU_X64_UTILS_JIT_REGOPS_HPP
+#define CPU_X64_UTILS_JIT_REGOPS_HPP
+
+#include "cpu/x64/jit_generator.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace x64 {
+namespace regops {
```
georgen117, line 26: Why name this `regops`?

densamoilov:
> @densamoilov, could you please include model-level summary as well?

@tprimak, I see some regressions caused by cases that are missing from the tables I included in the description. I'll look into them to figure out the reason.

densamoilov:
make test

densamoilov:
@tprimak, I enabled brgemm matmul only for GEMV cases supported via the new GEMV code path to avoid introducing regressions. The remaining gaps will be closed in separate pull-requests.

Below is performance data at the model level for ADL and SRF. Using brgemm with GEMV code path instead of GEMM doesn't introduce regressions (+/- 1-2% are irrelevant fluctuations).
image image

densamoilov:
make test

src/cpu/x64/matmul/brgemm_matmul.cpp
densamoilov: >  Like we currently do for brgemm and brdgmm.

Actually, we don't do that. We require the corresponding descriptors to be explicitly initialized by calling `brdgmm_desc_init` or `brgemm_desc_init`.

The main reason we have to explicitly choose which descriptor to initialize is that the initialization functions don't have enough information to make the decision for us. For example, at the matmul primitive level, we may want to create a brgemm for a tail part where N = 1. At this level, we know we want brgemm and not brgemv because the elements of the column are not contiguous in memory. However, the initialization functions don't have this information and therefore cannot make the correct decision.

src/cpu/x64/utils/jit_regops.hpp
densamoilov: I've thought about it, and technically we could put everything into `jit_generator_t` but I thought it might be better to use `jit_generator_t` only for individual instructions (i.e., the finest granularity) going forward, and to implement stateless operations that use multiple instructions separately as standalone functions with register-level interfaces. I thought we could benefit from structuring it that way, though maybe it's not a big deal.

src/cpu/x64/brgemm/brgemm.cpp
densamoilov: Yes, exactly, it's a typical notation for GEMV used in BLAS.

src/cpu/x64/brgemm/brgemm.hpp
densamoilov: It's the [classic notation](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-0/cblas-gemv.html) of GEMV, where INCY denotes a stride between elements of the output vector. Stride related parameters are grouped together in the brgemm interface so I just followed the convention.

src/cpu/x64/brgemm/brgemm.cpp
densamoilov: We don't have a separate parameter for the size of vector `x`. We have `M` and `N` parameters for matrix `A` and we infer the sizes of the vectors (`x` and `y`) from them so `N` is the size of vector `x`.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: It's the current state of affairs but this kernel will likely be enabled for avx512 later on. Usually, we use ISA in the name of a function when the function is designed to be used with that ISA only (e.g. maybe_set_avx_mask).

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: It's a leftover from some experiments, I'll revert it, thanks

src/cpu/x64/utils/jit_regops.hpp
densamoilov: I'd still prefer `horizontal_add_ps` because it's descriptive and more familiar due to the intrinsic with a similar name `_mm_hadd_ps`. We can switch to `reduce_ps` in the future if the need arises to support different reduction operations like sum, multiplication, division, etc.

densamoilov: > Should jit_getops.hpp/cpp be in its own PR?

Not necessarily, but we usually try to place such changes into different commits to reduce the noise.

src/cpu/x64/utils/jit_regops.hpp
densamoilov: Stands for "register operations", which refers to operations performed directly on registers. The interfaces must be at the registers level.

densamoilov:
> Would it make sense to separate the GEMV case into its own source files. Once brgemm.cpp has decided this is a GEMV case the rest of the GEMV path would be limited to only GEMV files. I think this would aid in long term maintenance.

@georgen117 , you are right. GEMV is a special case of GEMM, and that is exactly why we implement a GEMV code path within the existing GEMM kernel. This allows us to leverage the existing kernel infrastructure, which would otherwise have to be almost fully duplicated.

We could move the GEMV code into a separate file, but that would only be a spatial separation, not a structural one, because the GEMV microkernel still has to remain a member function of `jit_brgemm_kernel_t`. Additionally, the main kernel infrastructure would still reside in the GEMM file, and we would still need to handle branching between the GEMM and GEMV cases.

src/cpu/x64/brgemm/brgemm.cpp
densamoilov: yes, you are right, fixed

src/cpu/x64/brgemm/brgemm.hpp
densamoilov: Yeah, it's just f32 for now, fixed

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: Fixed

PR #3955, opened 2025-09-16
benchdnn: introduce implementation summary option support

dzarukin:
A new summary option providing the statistics of implementations used over the selected batch, enabled by default.
Designed to help to validate dispatching (that a new version doesn't make optimized version called less times than it should).
It's also helpful to understand the status quo in automated validation.

Feature preview:
image

### DIFF
--- a/src/cpu/x64/jit_uni_ncsp_convolution.hpp
+++ b/src/cpu/x64/jit_uni_ncsp_convolution.hpp
@@ -76,7 +76,7 @@ struct jit_uni_ncsp_convolution_fwd_t : public primitive_t {
         status_t init_matmul(engine_t *engine);
         reduction_helper_t reduction_helper_;
         bool is_matmul_ = false;
-        std::string name_ = "jit_uni_ncsp_convolution:";
+        std::string name_ = "jit_uni_ncsp:";
         void init_name() {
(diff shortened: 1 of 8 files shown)


amakarev:
I made [some changes](https://github.com/uxlfoundation/oneDNN/pull/3962) to ensure the table is displayed correctly even in small terminal sizes.
I'm not sure how useful or common this case is, so feel free to reject it if it's not relevant.

vpirogov (approved):

dzarukin:
> I made [some changes](https://github.com/uxlfoundation/oneDNN/pull/3962) to ensure the table is displayed correctly even in small terminal sizes. I'm not sure how useful or common this case is, so feel free to reject it if it's not relevant.

Thank you for taking care of small terminal windows. While it breaks the line and prints the tail on the next line, it may confuse the user a little bit with a new entry ending somewhere, and it also breaks the original alignment.

The general rule for reporting is to fit a laptop screen as that's generally a physical screen limit which devs can't go beyond.
Current output fits it and I'm very tempting to keep current change as is, however, I've left a TODO with the idea for small terminal windows and a piece of code to get the window size and an idea how to report the table in that case. Thanks again.

dzarukin:
make test

PR #3957, opened 2025-09-17
graph: src, tests, doc: enable graph dump by default

ElaineBao:
# Description

Implementation of option 3 in
- https://github.com/uxlfoundation/oneDNN/pull/3856

Performance comparison between w/o this change (graph dump disabled) and with this change (graph dump build option is enabled, but the env var is not set, meaning no json will be dumped):

compared ~70 cases in [harness_mha_ci](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/inputs/graph/complex_fusion/harness_mha_ci), most of them are within +-5% variance. Some corner cases can be rerun and meet this criteria. So the conclusion is: this PR will not have performance impact.

| %prb% | with_change %0time% | without_change %0time% | Δ (%) (- is better, + is worse) |
|-------|---------------------|------------------------|-------|
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/JAX-MHA-inf-fp32.json | 16.6222 | 16.7395 | -0.7% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/JAX-MHA-inf-fp32.json | 4.76679 | 4.65991 | +2.3% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/JAX-MHA-inf-fp32.json | 10.069 | 10.0166 | +0.5% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/JAX-MQA-inf-fp32.json | 7.53723 | 7.43765 | +1.3% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/JAX-MQA-inf-fp32.json | 2.27794 | 2.2197 | +2.6% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/JAX-MQA-inf-fp32.json | 3.81094 | 3.80145 | +0.2% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/MHA-GPT-inf-fp32-bs1.json | 0.587556 | 0.582546 | +0.9% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/MHA-GPT-inf-fp32-bs1.json | 0.718261 | 0.7803 | -8.0% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/MHA-GPT-inf-fp32-bs1.json | 0.624413 | 0.65197 | -4.2% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/MHA-bert_large-inf-fp32-bs1.json | 0.519551 | 0.59401 | -12.5% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/MHA-bert_large-inf-fp32-bs1.json | 0.42573 | 0.436145 | -2.4% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/MHA-bert_large-inf-fp32-bs1.json | 0.507759 | 0.512168 | -0.9% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/MHA-stable_diffusion-inf-fp32-bs1.json | 0.378318 | 0.437605 | -13.6% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/MHA-stable_diffusion-inf-fp32-bs1.json | 0.370395 | 0.366017 | +1.2% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/MHA-stable_diffusion-inf-fp32-bs1.json | 0.413213 | 0.405816 | +1.8% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json | 1.2373 | 1.33652 | -7.4% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json | 1.92972 | 1.95568 | -1.3% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/MHA-distill_bert-inf-fp32-bs1.json | 1.39691 | 1.51897 | -8.0% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.451484 | 0.436122 | +3.5% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.402384 | 0.394736 | +1.9% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.677041 | 0.67816 | -0.2% |
| --mode=P --graph --dt=f32 --op-kind=1:Multiply --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.534608 | 0.561263 | +5.9% |
| --mode=P --graph --dt=bf16 --op-kind=1:Multiply --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.540878 | 0.510381 | -5.9% |
| --mode=P --graph --dt=f16 --op-kind=1:Multiply --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.616135 | 0.62788 | -1.9% |
| --mode=P --graph --dt=f32 --op-kind=1:Divide --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.438017 | 0.41525 | +5.5% |
| --mode=P --graph --dt=bf16 --op-kind=1:Divide --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.385641 | 0.387307 | -0.4% |
| --mode=P --graph --dt=f16 --op-kind=1:Divide --case=complex_fusion/mha/sdpa-plain-simplified-f16.json | 0.476502 | 0.495837 | -3.9% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/sdpa-plain-wo-scale-f16-bs1.json | 0.430938 | 0.452047 | -4.7% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/sdpa-plain-wo-scale-f16-bs1.json | 0.378306 | 0.378202 | +0.03% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/sdpa-plain-wo-scale-f16-bs1.json | 0.540748 | 0.517701 | +4.5% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/GQA-fp16.json | 6.00854 | 6.23094 | -3.6% |
| --mode=P --graph --dt=bf16 --case=complex_fusion/mha/GQA-fp16.json | 1.89051 | 1.86772 | +1.2% |
| --mode=P --graph --dt=f16 --case=complex_fusion/mha/GQA-fp16.json | 3.63116 | 3.69625 | -1.8% |
| --mode=P --graph --dt=f32 --case=complex_fusion/mha/GQA-fp16-v2.json | 6.38605 | 6.18431 | +3.3% |
| ... | ... | ... | ... |

### DIFF
--- a/tests/gtests/graph/api/test_c_api_graph_dump.cpp
+++ b/tests/gtests/graph/api/test_c_api_graph_dump.cpp
@@ -19,8 +19,17 @@
 #include "oneapi/dnnl/dnnl_graph.h"
 
 TEST(CAPI, GraphDump) {
-    ASSERT_EQ(dnnl_graph_set_dump_mode("subgraph"), dnnl_success);
-    ASSERT_EQ(dnnl_graph_set_dump_mode("graph"), dnnl_success);
-    ASSERT_EQ(dnnl_graph_set_dump_mode(""), dnnl_success);
-    ASSERT_EQ(dnnl_graph_set_dump_mode("unknown"), dnnl_invalid_arguments);
+    ASSERT_EQ(dnnl_graph_set_dump_mode(dnnl_graph_dump_mode_subgraph),
+            dnnl_success);
+    ASSERT_EQ(
+            dnnl_graph_set_dump_mode(dnnl_graph_dump_mode_graph), dnnl_success);
+    ASSERT_EQ(dnnl_graph_set_dump_mode(static_cast<dnnl_graph_dump_mode_t>(
+                      dnnl_graph_dump_mode_graph
+                      | dnnl_graph_dump_mode_subgraph)),
+            dnnl_success);
+    ASSERT_EQ(
+            dnnl_graph_set_dump_mode(dnnl_graph_dump_mode_none), dnnl_success);
+    ASSERT_EQ(
+            dnnl_graph_set_dump_mode(static_cast<dnnl_graph_dump_mode_t>(0x4U)),
+            dnnl_invalid_arguments);
 }
--- a/tests/gtests/graph/api/test_cpp_api_graph_dump.cpp
+++ b/tests/gtests/graph/api/test_cpp_api_graph_dump.cpp
@@ -21,11 +21,17 @@
 #include "oneapi/dnnl/dnnl_graph.hpp"
 
 TEST(APIGraphDump, SetDumpMode) {
-    ASSERT_EQ(dnnl::graph::set_dump_mode("subgraph"),
+    using dnnl::graph::graph_dump_mode;
+
+    ASSERT_EQ(dnnl::graph::set_dump_mode(graph_dump_mode::subgraph),
             dnnl::graph::status::success);
-    ASSERT_EQ(
-            dnnl::graph::set_dump_mode("graph"), dnnl::graph::status::success);
-    ASSERT_EQ(dnnl::graph::set_dump_mode(""), dnnl::graph::status::success);
-    ASSERT_EQ(dnnl::graph::set_dump_mode("unknown"),
+    ASSERT_EQ(dnnl::graph::set_dump_mode(graph_dump_mode::graph),
+            dnnl::graph::status::success);
+    ASSERT_EQ(dnnl::graph::set_dump_mode(
+                      graph_dump_mode::graph | graph_dump_mode::subgraph),
+            dnnl::graph::status::success);
+    ASSERT_EQ(dnnl::graph::set_dump_mode(graph_dump_mode::none),
+            dnnl::graph::status::success);
+    ASSERT_EQ(dnnl::graph::set_dump_mode(static_cast<graph_dump_mode>(0x4U)),
             dnnl::graph::status::invalid_arguments);
 }
--- a/tests/gtests/graph/unit/utils/test_utils_cpu.cpp
+++ b/tests/gtests/graph/unit/utils/test_utils_cpu.cpp
@@ -105,38 +105,58 @@ TEST(test_utils_utils, GetGraphDumpSetByEnv) {
     custom_setenv("ONEDNN_GRAPH_DUMP", "subgraph", 1);
     ASSERT_EQ(dnnl::impl::getenv_string_user("GRAPH_DUMP"), "subgraph");
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::subgraph));
+            dnnl::impl::graph::graph_dump_mode_t::subgraph));
     ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::graph));
+            dnnl::impl::graph::graph_dump_mode_t::graph));
 
     custom_setenv("ONEDNN_GRAPH_DUMP", "graph,subgraph", 1);
     ASSERT_EQ(dnnl::impl::getenv_string_user("GRAPH_DUMP"), "graph,subgraph");
     // environment variable can only be set once
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::subgraph));
+            dnnl::impl::graph::graph_dump_mode_t::subgraph));
     ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::graph));
+            dnnl::impl::graph::graph_dump_mode_t::graph));
 }
 
 TEST(test_utils_utils, GetGraphDumpSetByAPI) {
-    ASSERT_EQ(dnnl::graph::set_dump_mode("subgraph"),
+    using dnnl::graph::graph_dump_mode;
+
+    ASSERT_EQ(dnnl::graph::set_dump_mode(graph_dump_mode::subgraph),
             dnnl::graph::status::success);
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::subgraph));
+            dnnl::impl::graph::graph_dump_mode_t::subgraph));
     ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::graph));
+            dnnl::impl::graph::graph_dump_mode_t::graph));
 
     // API can be used as many times as possible to change the mode
-    ASSERT_EQ(dnnl::graph::set_dump_mode("subgraph,graph"),
+    ASSERT_EQ(dnnl::graph::set_dump_mode(
+                      graph_dump_mode::subgraph | graph_dump_mode::graph),
             dnnl::graph::status::success);
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::subgraph));
+            dnnl::impl::graph::graph_dump_mode_t::subgraph));
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::graph));
+            dnnl::impl::graph::graph_dump_mode_t::graph));
 
-    ASSERT_EQ(dnnl::graph::set_dump_mode(""), dnnl::graph::status::success);
+    ASSERT_EQ(dnnl::graph::set_dump_mode(graph_dump_mode::none),
+            dnnl::graph::status::success);
     ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::none));
+            dnnl::impl::graph::graph_dump_mode_t::none));
     ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
-            dnnl::impl::graph::utils::graph_dump_mode_t::graph));
+            dnnl::impl::graph::graph_dump_mode_t::graph));
+
+    ASSERT_EQ(dnnl::graph::set_dump_mode(
+                      graph_dump_mode::graph | graph_dump_mode::none),
+            dnnl::graph::status::success);
+    ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
+            dnnl::impl::graph::graph_dump_mode_t::none));
+    ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
+            dnnl::impl::graph::graph_dump_mode_t::graph));
+
+    ASSERT_EQ(dnnl::graph::set_dump_mode(
+                      graph_dump_mode::graph & graph_dump_mode::none),
+            dnnl::graph::status::success);
+    ASSERT_TRUE(dnnl::impl::graph::utils::get_graph_dump_mode(
+            dnnl::impl::graph::graph_dump_mode_t::none));
+    ASSERT_FALSE(dnnl::impl::graph::utils::get_graph_dump_mode(
+            dnnl::impl::graph::graph_dump_mode_t::graph));
 }


TaoLv:
> compared ~70 cases in [harness_mha_ci](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/inputs/graph/complex_fusion/harness_mha_ci), most of them are within +-5% variance.

Please check more cases, especially smaller cases (op/pattern cases). Please also share an average number.

ElaineBao:
> > compared ~70 cases in [harness_mha_ci](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/inputs/graph/complex_fusion/harness_mha_ci), most of them are within +-5% variance.
> 
> Please check more cases, especially smaller cases (op/pattern cases). Please also share an average number.

Compared ~130 smaller cases (op/pattern cases) in [test_graph_ci](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/inputs/graph/test_graph_ci), most of them are within +-5% variance. Some corner cases can be rerun and meet this criteria. The avg number is -1.5% (with_change is slightly better, which is more likely to be variance).

src/graph/utils/verbose.hpp
```diff
@@ -51,6 +51,15 @@ struct partition_info_t {
     std::once_flag initialization_flag_;
 };
 
+enum graph_dump_mode_t : uint8_t {
+    graph_dump_none = 0,
+    graph_dump_subgraph = 1 << 0,
+    graph_dump_graph = 1 << 1,
+    graph_dump_pattern = 1 << 2,
+};
```
TaoLv: I suggest to remove the duplications in the names. 

```c++
enum class dump_mode_t : uint8_t {
    none = 0,
    subgraph = 1 << 0,
    graph = 1 << 1,
    pattern = 1 << 2,
};
```

src/graph/utils/verbose.cpp
```diff
@@ -241,6 +241,34 @@ void partition_info_t::init(const engine_t *engine,
     });
 }
 
+bool get_graph_dump_mode(uint8_t mode) {
```
TaoLv: Then,
```suggestion
bool get_graph_dump_mode(dump_mode_t mode) {
```

src/graph/utils/verbose.cpp
```diff
@@ -241,6 +241,34 @@ void partition_info_t::init(const engine_t *engine,
...
+            else if (token == "graph")
+                m |= graph_dump_graph;
+            else if (token == "pattern")
+                m |= graph_dump_pattern;
+            start = end + 1;
+        }
+        return m;
+    }();
+    return (modes & mode) != 0;
+#endif
```
TaoLv, line 287: It means the env var will be read and parsed at the first time this function is called. After that, there is no way to change the mode. I want to confirm that this aligns with how `ONEDNN_VERBOSE` is read and saved in the library. @avmanerikar  Could you please take a look? Thanks.

src/graph/utils/verbose.hpp
ElaineBao: fixed, thanks

src/graph/utils/verbose.cpp
dzarukin: The env var is read once at get_verbose time, that's true. However, the user can update the behavior in a programmable way through [`dnnl_set_verbose` ](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/verbose.cpp#L1813). Graph API would need to provide an identical call.

src/graph/backend/dnnl/dnnl_backend.hpp
```diff
@@ -111,24 +111,25 @@ class dnnl_backend_t : public backend_t {
...
-        std::string pass_config_json = "dnnl_graph_passes.json";
-        std::ifstream fs(pass_config_json.c_str());
-        if (fs) {
-            verbose_printf(
-                    "graph,info,pattern,load,%s\n", pass_config_json.c_str());
-        } else {
-            if (getenv_int_user("GRAPH_DUMP", 0) > 0
-                    || graph::utils::check_verbose_string_user(
-                            "GRAPH_DUMP", "pattern")) {
+#ifdef DNNL_DISABLE_GRAPH_DUMP
```
dzarukin: Given that `get_graph_dump_mode` is available at all times, why introducing extra macros at upper levels?

The idea behind a build time variable is to limit control over the application through env variables (which shouldn't be respected if it was opt-out) but not through a programmable way (which is to be added?).

ElaineBao: Hi @dzarukin , the reason I introduced this macro is that I want to make it align with DISABLE_VERBOSE, which also completely removes the code at build time. Are you suggesting the macro is unnecessary at all places or just here?

dzarukin: I found that saving negligible amount of space not worth the maintain effort for the `defined(DNNL_DISABLE_GRAPH_DUMP)` case. BTW, have you tried to use it in a build? Is it successful so far?

ElaineBao: > I found that saving negligible amount of space not worth the maintain effort for the defined(DNNL_DISABLE_GRAPH_DUMP) case.

Understood, I'm going to remove the definition of DNNL_DISABLE_GRAPH_DUMP.

> BTW, have you tried to use it in a build? Is it successful so far?

My local test shows it's ok.

src/graph/utils/verbose.cpp
ElaineBao: Thanks for the suggestion, added `dnnl_graph_set_dump_mode` API.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

include/oneapi/dnnl/dnnl_graph.hpp
```diff
@@ -1637,6 +1637,12 @@ inline size_t get_constant_tensor_cache_capacity(engine::kind kind) {
 
 /// @} dnnl_graph_api_constant_tensor_cache
 
+/// @copydoc dnnl_graph_set_dump_mode()
+inline status set_dump_mode(const std::string &modes = "") {
```
TaoLv: 1. What's the purpose of the default value of modes?
2. Why do we return `status` rather than raising an exception like other APIs?

src/graph/utils/verbose.cpp
```diff
@@ -241,7 +241,71 @@ void partition_info_t::init(const engine_t *engine,
...
+        // Assumes that all threads see the same environment
+        static std::string env = getenv_string_user("GRAPH_DUMP");
+        graph_dump_modes.set(parse_graph_dump_mode(env));
+    }
+
+    uint8_t saved_mode = graph_dump_modes.get();
+    if (saved_mode == 0) {
+        return saved_mode == mode;
+    } else {
+        return (graph_dump_modes.get() & mode) != 0;
```
TaoLv:
```suggestion
        return (saved_mode & mode) != 0;
```

Why call `get()` again and why check `saved_mode == 0`?

src/graph/utils/verbose.cpp
```diff
@@ -241,7 +241,71 @@ void partition_info_t::init(const engine_t *engine,
...
+}
+
 } // namespace utils
 } // namespace graph
 } // namespace impl
 } // namespace dnnl
+
+dnnl::impl::graph::status_t dnnl_graph_set_dump_mode(const char *modes) {
+#ifdef DNNL_DISABLE_GRAPH_DUMP
+    return dnnl::impl::graph::status::success;
```
TaoLv: Shouldn't it return a failure if graph dump feature is disabled?

src/graph/utils/verbose.hpp
```diff
@@ -51,6 +51,15 @@ struct partition_info_t {
     std::once_flag initialization_flag_;
 };
 
+enum graph_dump_mode_t : uint8_t {
```
TaoLv:
```suggestion
enum class graph_dump_mode_t : uint8_t {
```

src/graph/utils/verbose.cpp
```diff
@@ -241,7 +241,71 @@ void partition_info_t::init(const engine_t *engine,
...
+#else
+    std::string mode_str = modes ? std::string(modes) : "";
```
TaoLv: I would suggest to check `if (modes == nullptr)` before wrapping it into an empty string. You will not be able to tell if an empty mode_str is passed by the user, or created here by `""`.
BTW, is it intended to take a nullptr and an empty string as the same thing here?

include/oneapi/dnnl/dnnl_graph.hpp
ElaineBao: 1. It's to provide a default dump mode (which is to disable dump), it can be removed if you have concern about it.
2. It's to follow the common practice of [set_verbose](https://github.com/uxlfoundation/oneDNN/blob/main/include/oneapi/dnnl/dnnl.hpp#L13916) API, and I also notice some graph APIs, e.g. add op, return status.

src/graph/utils/verbose.cpp
ElaineBao: thanks for pointing out, indeed another get() is not needed.

The reason why checking saved_mode == 0 is that dump_mode=none is also a valid case, but using 0 & 0 will return false.

src/graph/utils/verbose.cpp
ElaineBao: Thanks for the reminder, I separate these two cases.

src/graph/utils/verbose.cpp
ElaineBao: returns invalid_arguments if graph dump feature is disabled.

include/oneapi/dnnl/dnnl_graph.hpp
```diff
@@ -1637,6 +1637,12 @@ inline size_t get_constant_tensor_cache_capacity(engine::kind kind) {
 
 /// @} dnnl_graph_api_constant_tensor_cache
 
+/// @copydoc dnnl_graph_set_dump_mode()
+inline status set_dump_mode(const std::string &modes) {
```
TaoLv: How about expose the `graph_dump_mode_t` to users through this API, so we don't need to parse the string?

ElaineBao: We eventually need to parse the string because we need to support the parse of environment variable

TaoLv: I can understand that we have to parse the string of environment variables and it leads to some performance overhead. That's why we choose to parse them once and save the modes inside the library. I think the purpose of this API is to address the limitation of env var, as it can be called during runtime multiple times with a lighter overhead.

Another consideration is that with explicit modes defined in `graph_dump_mode_t`, the user code will be more robust as the compiler will check the value and type during code compilation. But if we accept `char *` we will have to handle the illegal values or addresses passed to the library. We can also define `disable dump` as a mode explicitly, rather than using nullptr or empty string to mimic it.

TaoLv: BTW, how about the thread safety of this global function?

TaoLv: Do we also need to support the corresponding get function?

src/graph/utils/verbose.cpp
```diff
@@ -241,7 +241,75 @@ void partition_info_t::init(const engine_t *engine,
...
+    while ((end = user_opt.find(',', start)) != std::string::npos) {
+        std::string token = user_opt.substr(start, end - start);
+        if (token == "subgraph")
+            m |= static_cast<uint8_t>(graph_dump_mode_t::subgraph);
+        else if (token == "graph")
+            m |= static_cast<uint8_t>(graph_dump_mode_t::graph);
+        else if (token == "pattern")
+            m |= static_cast<uint8_t>(graph_dump_mode_t::pattern);
+        else
+            m = static_cast<uint8_t>(graph_dump_mode_t::none);
```
TaoLv, line 264: nit: Is the `static_cast` really needed?

include/oneapi/dnnl/dnnl_graph.hpp
ElaineBao: > expose the graph_dump_mode_t to users through this API

ok, I'll do that.

> how about the thread safety of this global function?

There's no thread safety guaranteed for this global function. IIUC, set_dnnl_verbose API doesn't guarantee that also. 

> Do we also need to support the corresponding get function?

I think get function is not very useful, and there's no get_dnnl_verbose API also, so we can keep the set function for now and add get function if there's user requests in the future.

src/graph/utils/verbose.cpp
ElaineBao: yes, there will be type mismatch between uint8_t and graph_dump_mode_t if I remove the cast.

include/oneapi/dnnl/dnnl_graph.hpp
ElaineBao: exposed the graph_dump_mode_t to users through this API, please take another look, thanks!

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -14180,8 +14180,6 @@ inline void primitive::execute(const stream &astream,
 
 /// @endcond
 
-#undef DNNL_DEFINE_BITMASK_OPS
```
TaoLv, line 14204: Undefining the macro here will prevent us from using the macro in dnnl_graph.hpp, when both dnnl.hpp and dnnl_graph.hpp are included. See an example here: https://github.com/uxlfoundation/oneDNN/blob/ceb0d6c8c5685ee9af1eb6c07a3508248e886673/examples/graph/sdpa.cpp#L26-L27

TaoLv (commented):
@dzarukin @vpirogov  The programming API for graph dump and library header files were updated since your approvals. Could you please take another look on them?

src/graph/utils/verbose.hpp
```diff
@@ -51,6 +53,15 @@ struct partition_info_t {
     std::once_flag initialization_flag_;
 };
 
+enum class graph_dump_mode_t : uint8_t {
+    none = dnnl_graph_dump_mode_none,
+    subgraph = dnnl_graph_dump_mode_subgraph,
+    graph = dnnl_graph_dump_mode_graph,
+    pattern = 0x4U,
+};
```
dzarukin: This should go through `c_types_map.hpp` and the public header must be removed from here. Extra internal values should be handled there as well.

dzarukin (approved):

src/graph/utils/verbose.hpp
ElaineBao: Fixed, thanks

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3958, opened 2025-09-17
graph: support f32 host scalar and use it in the gqa example

TaoLv:
- Support f32 host scalar in logical tensor and tensor. Previously only s32 was supported and used to pass sequence length.
- Updated the existing GQA example to define attention scale as a host scalar.

### DIFF
--- a/src/graph/backend/dnnl/kernels/sdp_primitive.cpp
+++ b/src/graph/backend/dnnl/kernels/sdp_primitive.cpp
@@ -84,7 +84,6 @@ status_t sdp_primitive_kernel_t<quantized>::compile_impl(
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_zero_points);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_zero_points);
     }
-    BACKEND_DNNL_ADD_PASS(pipeline, binary_canonicalization);
     BACKEND_DNNL_ADD_PASS(pipeline, insert_permute_for_matmul);
     if (quantized) {
         BACKEND_DNNL_ADD_PASS(pipeline, remove_quant_data_with_no_effect);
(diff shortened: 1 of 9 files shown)


xiang1guo (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/kernels/sdp_primitive.cpp
```diff
@@ -84,7 +84,6 @@ status_t sdp_primitive_kernel_t<quantized>::compile_impl(
         BACKEND_DNNL_ADD_PASS(pipeline, convert_to_runtime_dst_zero_points);
         BACKEND_DNNL_ADD_PASS(pipeline, fuse_dst_zero_points);
     }
-    BACKEND_DNNL_ADD_PASS(pipeline, binary_canonicalization);
```
ElaineBao, line 87: could you please elaborate why this line is deleted?

TaoLv: This pass was inherited from large partition kernel where it's used to convert scale input from a scalar value to a 4D value (with shape (1, 1, 1, 1)) as we rely on binary primitive to apply scale and the primitive requires ndims to match for src0 and src1.

It's not needed for sdp_primitive kernel because scale is passed to the primitive directly and no need to convert it to 4D.

src/graph/interface/logical_tensor.cpp
```diff
@@ -189,9 +189,12 @@ status_t DNNL_API dnnl_graph_logical_tensor_init(
...
     }
 
-    // currently only supports s32 host scalar.
-    if (ptype == property_type::host_scalar && dtype != data_type::s32) {
-        return status::invalid_arguments;
+    // currently only supports s32 and f32 host scalar.
+    if (ptype == property_type::host_scalar) {
+        if (!utils::one_of(dtype, data_type::s32, data_type::f32)
+                || ndims != 0) {
+            return status::invalid_arguments;
```
dzarukin: Minor: using verbose output for errors could be helpful.

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #3966, opened 2025-09-18
xe: layout simplification part 2

rjoursler:
Continuation of the work starting in #3937 and includes that patches in #3947 to avoid potential merge conflicts. This PR focuses on removing interfaces used for physical layout manipulation, legacy interfaces which no longer make sense given layout_t's structure, along with some interfaces with low utilization. Additionally, this fixes some weird properties of the structure due to how hashing and equality checks work. After this is merged, there is likely one more round of simplification required before layout_t can be moved in to the DSL tensor.hpp file.


src/gpu/intel/conv/jit/normalization.cpp
```diff
@@ -109,26 +108,26 @@ layout_t normalize_layout(const layout_t &_layout, bool with_groups,
     return layout;
 }
 
-std::vector<dim_t> normalize_dims(std::vector<dim_t> &dims, bool with_groups,
-        dim_t groups, bool is_dw, const std::array<int, 3> &dhw_map,
-        bool add_groups, bool is_wei) {
+tile_t normalize_tile(std::vector<dim_t> &dims, bool with_groups, dim_t groups,
```
atkassen, line 111: Based on the name, it seems like this function should also take a `(const) tile_t&`.

src/gpu/intel/jit/codegen/reorder.cpp
```diff
@@ -654,25 +654,24 @@ layout_t reorder_impl_t::make_compact_layout(
...
+        const dim_t inner_stride = [&]() {
+            dim_t ret = stride_t::max();
+            for (auto &b : blocks) {
+                if (b.stride < ret) ret = b.stride;
+            }
+            return ret;
+        }();
+        std::vector<layout_block_t> new_blocks = blocks;
+        for (size_t i = 0; i < blocks.size(); i++) {
+            new_blocks[i].stride = new_blocks[i].stride * stride / inner_stride;
```
atkassen: You could probably go even further:
```suggestion
        if (!layout.nblocks()) return layout;
        auto blocks = layout.blocks();
        dim_t inner_stride = stride_t::max();
        for (auto &b : blocks)
            inner_stride = std::min<dim_t>(inner_stride, b.stride);
        for (auto &b : blocks)
            b.stride = b.stride * stride / inner_stride;
```

src/gpu/intel/jit/ir/gemm_schedule.cpp
```diff
@@ -73,7 +73,7 @@ layout_t bmnk_block_mapper_t::map_from_bmnk(abc_kind_t abc_kind,
         const std::vector<bmnk_kind_t> &bmnk_kinds,
         const layout_t &bmnk_layout) const {
     gpu_assert(bmnk_layout.ndims() <= 3);
-    gpu_assert(bmnk_layout.has_zero_offset());
+    gpu_assert(bmnk_layout.offset().is_equal(expr_t(0)));
```
echeresh:
```suggestion
    gpu_assert(is_zero(bmnk_layout.offset()));
```

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -406,22 +375,10 @@ class layout_t {
...
-        dim_t ret = 1;
-        int nblocks = (int)dim_blocks.size();
-        int lo = 0;
-        int hi = skip_outer ? nblocks - 1 : nblocks;
-        if (inner_only) hi = std::min(hi, 1);
-        for (int i = lo; i < hi; i++)
-            ret *= dim_blocks[i];
-        return ret;
-    }
+    const layout_block_t &operator[](size_t idx) const { return blocks_[idx]; }
```
echeresh, line 380: I think we should change `nblocks()` to return `size_t` (or vice versa, `int idx` here) - to align.

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -458,9 +417,9 @@ class layout_t {
                 && self_blocks[i].stride == other_blocks[i].stride
                 && other_blocks[i].block % self_blocks[i].block == 0);
     }
-    bool operator>=(const layout_t &other) const { return other <= *this; }
 
-    bool is_equal(const layout_t &other, bool compare_offset = true) const {
+    bool is_isomorphic(
```
echeresh: Nit: isomorphic may be a bit broad, e.g. "isomorphism" may assume we don't care about whether dimensions (`pvar_t` objects) match between layouts

Another option: use something referring to "equivalent", e.g. `is_equiv()`
NumPy has [array_equiv](https://numpy.org/doc/2.1/reference/generated/numpy.array_equiv.html#numpy-array-equiv)

@rjoursler @atkassen @Simonsays095 What do you think?

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -522,52 +473,17 @@ class layout_t {
         if (is_empty()) return "(nil)";
         ostringstream_t oss;
         oss << desc_str();
-        if (!has_zero_offset()) oss << " offset: " << offset_;
+        if (!offset().is_equal(expr_t(0))) oss << " offset: " << offset_;
```
echeresh:
```suggestion
        if (!is_zero(offset())) oss << " offset: " << offset_;
```

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -1012,6 +808,78 @@ class layout_t {
...
+        max_block_size = std::max(max_block_size, b.block * (dim_t)b.stride);
+    }
+    dim_t max_elems = std::max(max_off + 1, max_block_size);
+    return utils::rnd_up(
+            max_elems * layout.type().size() / layout.type().packing(),
+            alignment);
+}
+
+template <typename T = expr_t>
+T offset_in_bytes(const layout_t &layout, const coord_t &coord = {},
```
echeresh: To align with `size_bytes()`:
```suggestion
T offset_bytes(const layout_t &layout, const coord_t &coord = {},
```

src/gpu/intel/utils.hpp
```diff
@@ -25,7 +25,7 @@
     VCHECK(common, create, check, runtime, stat, msg, ##__VA_ARGS__);
 
 // Uncomment this when aborting on ir_assert is desired:
-// #define GPU_ABORT_ON_ERROR
+#define GPU_ABORT_ON_ERROR
```
echeresh: Leaving a comment to not forget to revert this change back.

src/gpu/intel/jit/ir/tensor.hpp
```diff
@@ -326,46 +333,19 @@ class layout_t {
...
-        for (auto &eb : enumerated_blocks()) {
-            auto &b = eb.second;
+        for (auto &b : blocks()) {
             if (!_args.has(b.dim)) continue;
             auto &idx = _args[b.dim];
             if (is_zero(idx)) continue;
 
             // Do not use modulus for outermost blocks.
-            auto i = is_outermost(eb) ? idx : (idx % b.block);
+            auto i = is_outermost(b) ? idx : (idx % b.block);
```
echeresh, line 348: Should we rename `layout_block_t::block` to `layout_block_t::size`?
This is unrelated to the PR, just wanted to bring this up since you are working on stabilizing the API.

rjoursler: I agree and renaming `layout_block_t::{block -> size}` and additionally `layout_block_t::{dim->idx}` are both changes I would like to see as well.

src/gpu/intel/conv/jit/normalization.cpp
rjoursler: While I generally agree, there is a bit more complexity that I don't think should be handled by this PR. In particular, a `layout_t` cannot be constructed from a `tile_t` as there is no inherent order. As such, this would need rewritten to avoid the use of `normalize_layout`, which is not a straightforward refactoring.

src/gpu/intel/utils.hpp
rjoursler: Thanks, I was under the belief that this was already reverted, so it was definitely forgotten.

src/gpu/intel/jit/ir/tensor.hpp
rjoursler: My main issue with `is_equiv` is it is unclear what notion of equivalency is used (for example, the broad interpretation of an isomorhpism is just as much an equivalency relationship).  How about `norm_equal` which directly references the `normalization` routine?

echeresh: `is_norm_equal()` may be misinterpreted as something related to the "norm"
`is_equal_normalized()` may be a more straightforward option

atkassen: > My main issue with `is_equiv` is it is unclear what notion of equivalency is used

You could make it explicit:
```suggestion
    bool is_equiv(
            const layout_t &other, bool compare_offset = true,
            const std::function<layout_t(const layout_t &)> &representative
            = [](const layout_t &l) { return l.normalize(); }) const {
        // Compare equivalence class representatives
        return representative(*this).is_strictly_equal(
                representative(other), compare_offset);
    }
```

Simonsays095: Alternatively, you could just remove `is_equal` and use `layoutA.normalize().is_strictly_equal(layoutB.normalize())` in all cases where this is currently used, which only appears to be a couple of cases. It's a bit more verbose, but this conversation seems to indicate that verbosity is probably welcome.

I would prefer this to the optional `representative` function idea, since you would be able to use `is_equiv` expecting one idea of equivalent and having the code use another, due to the default argument.

rjoursler:
make test
disable test_device_cpu
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_pool
enable benchdnn_reduce
enable benchdnn_reorder

PR #3972, opened 2025-09-19
cmake: enable stdlib debug mode for debug builds

rjoursler:
In [MFDNN-14221](https://jira.devtools.intel.com/browse/MFDNN-14221), we caught a vector OOB access with the MSVC debug C++ library. This PR enables similar checks for Debug and ONEDNN_DEV_MODE builds on Linux to improved cross-platform debugging.

As an example for the aforementioned bug caught by this:
```Bash
$  ctest -R gpu-cnn-inference-f32-cpp --output-on-failure
Test project /nfs/pdx/home/rjoursle/dnnl/build
    Start 6: gpu-cnn-inference-f32-cpp
1/1 Test #6: gpu-cnn-inference-f32-cpp ........Subprocess aborted***Exception:   1.60 sec
/usr/include/c++/13/bits/stl_vector.h:1128: constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = long int; _Alloc = std::allocator<long int>; reference = long int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   1.65 sec

The following tests FAILED:
          6 - gpu-cnn-inference-f32-cpp (Subprocess aborted)
```

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -290,6 +290,15 @@ if(MSVC)
         append(CMAKE_CCXX_FLAGS "-Wno-unknown-warning-option")
     endif()
 elseif(UNIX OR MINGW)
+
+    # Enable debug checks for common C++ standard libraries
+    if(DNNL_DEV_MODE OR CMAKE_BUILD_TYPE STREQUAL "Debug")
+        # Enable debug checks for libstdc++
+        append(CMAKE_CCXX_FLAGS "-D_GLIBCXX_ASSERTIONS")
+        # Enable debug checks for libc++
+        append(CMAKE_CCXX_FLAGS "-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE")
+    endif()
+
     if(DNNL_WITH_SYCL OR CMAKE_BASE_NAME STREQUAL "icx" OR CMAKE_BASE_NAME STREQUAL "icpx")
         # When using Debug build mode CMake adds "-g" option without "-O0"
         # causing the warning. This probably happens because clang/gcc compilers


cmake/platform.cmake
```diff
@@ -290,6 +290,15 @@ if(MSVC)
         append(CMAKE_CCXX_FLAGS "-Wno-unknown-warning-option")
     endif()
 elseif(UNIX OR MINGW)
+
+    # Enable debug checks for common C++ standard libraries
+    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+        # Enable debug checks for libstdc++
+        append(CMAKE_CCXX_FLAGS "-D_GLIBCXX_DEBUG")
```
vpirogov: According to libstdc++ documentation this flag breaks ABI for some classess:
> Note that this flag changes the sizes and behavior of standard class templates such as std::vector, and therefore you can only link code compiled with debug mode and code compiled without debug mode if no instantiation of a container is passed between the two translation units.

Will this create issues for debug builds of applications using oneDNN?

rjoursler: Yeah, this does seem like it could cause problems. Would it be reasonable to guard this behind `ONEDNN_DEV_MODE` which is not generally used by applications?

vpirogov: Sounds reasonable.

echeresh: @rjoursler Can you please also check SYCL build with this option? Due to ABI differences this may cause issues.

rjoursler: I looked into this a bit more as ABI issues seem overkill for the OOB read I was originally addressing. It looks like there is a macro [_GLIBCXX_ASSERTIONS](https://gcc.gnu.org/wiki/LibstdcxxDebugMode) which enables many debug checks, but which is ABI compatible. As such I switched this over to use _GLIBCXX_ASSERTIONS.

Additionally, it looks like the _LIBCPP_DEBUG macro is being [removed](https://discourse.llvm.org/t/rfc-removing-the-legacy-debug-mode-from-libc/71026), and is being replaced with some [hardening controls](https://libcxx.llvm.org/Hardening.html, so libc++ was switched to using these controls instead.

vpirogov: This makes sense. `_GLIBCXX_ASSERTIONS` was [enabled by default](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=112808) in GCC 15.

rjoursler:
make test

cmake/platform.cmake
```diff
@@ -290,6 +290,15 @@ if(MSVC)
         append(CMAKE_CCXX_FLAGS "-Wno-unknown-warning-option")
     endif()
 elseif(UNIX OR MINGW)
+
+    # Enable debug checks for common C++ standard libraries
+    if(DNNL_DEV_MODE OR CMAKE_BUILD_TYPE STREQUAL "Debug")
```
dzarukin, line 295: `OR`? Maybe `AND`?

rjoursler: Within the intel GPU code, we [generally enable assertions](https://github.com/uxlfoundation/oneDNN/blob/df10e27efda3de959a1818c29dfbfff5ccbd4246/src/gpu/intel/utils.hpp#L105-L115) within `DNNL_DEV_MODE` as this enables significantly faster validation in benchdnn due to the higher optimization level. This is aligning with that convention. I can go ahead and drop the condition on `DNNL_DEV_MODE` if it seems unwarranted though.

dzarukin: Ah, so it's intended for DEV_MODE + Release and any Debug giving the options are now don't bring the incompatible behavior... Thanks for clarifications.

PR #3973, opened 2025-09-19
fixup: fixup: benchdnn: utils: memory: add device memory tracking

dzarukin:
The logic of skipping remove_device shouldn't be used for CPU SYCL.

Fixes [MFDNN-14229](https://jira.devtools.intel.com/browse/MFDNN-14229)


dzarukin:
make test

echeresh (approved):

PR #3975, opened 2025-09-19
xe: sdpa: add separate configs for f16 accumulate

syurkevi:
# Description

F16 accumulate uses additional registers for conversion and in some cases good tile sizes for FMA with f32 accumulate will not be optimal for f16 accumulate. For example:
```sh
# f32 acc mode is faster
--mode=P --graph --engine=gpu --allow-enum-tags-only=false --dt=2:f32+5:f32 --in-shapes=0:1x16x1024x80*acbd+1:1x16x1024x80*acbd+9:1x16x1024x80*acbd --case=complex_fusion/mha/sdpa-plain-wo-mask-f16.json,2.69719,2.92919
# than f16 acc mode
--mode=P --graph --engine=gpu --allow-enum-tags-only=false --dt=2:f32+5:f32 --in-shapes=0:1x16x1024x80*acbd+1:1x16x1024x80*acbd+9:1x16x1024x80*acbd --op-attrs=3:accumulation_mode:f16+11:accumulation_mode:f16 --case=complex_fusion/mha/sdpa-plain-wo-mask-f16.json,2.81094,3.08006

# with proposed config f16 accumulate is faster as expected
# SDPA_CONFIG=8,16,16,16,8,4,8,4
--mode=P --graph --engine=gpu --allow-enum-tags-only=false --dt=2:f32+5:f32 --in-shapes=0:1x16x1024x80*acbd+1:1x16x1024x80*acbd+9:1x16x1024x80*acbd --op-attrs=3:accumulation_mode:f16+11:accumulation_mode:f16 --case=complex_fusion/mha/sdpa-plain-wo-mask-f16.json,2.48906,2.70803
# but config isn't applicable to f32 accumulate
--mode=P --graph --engine=gpu --allow-enum-tags-only=false --dt=2:f32+5:f32 --in-shapes=0:1x16x1024x80*acbd+1:1x16x1024x80*acbd+9:1x16x1024x80*acbd --case=complex_fusion/mha/sdpa-plain-wo-mask-f16.json,3.06865,3.36723
```

This PR adds f16 accumulate as a property for SDPA configurations.

### DIFF
--- a/src/gpu/intel/sdpa/configs.cpp
+++ b/src/gpu/intel/sdpa/configs.cpp
@@ -53,7 +53,8 @@ std::string to_string(const config_query_t &q) {
       << ((bool)(q.prop & property::quantized) ? " quant" : "")
       << ((bool)(q.prop & property::integrated) ? " [integ]" : "")
       << ((bool)(q.prop & property::fma) ? " fma" : "")
-      << ((bool)(q.prop & property::f32) ? " [f32]" : "");
+      << ((bool)(q.prop & property::f32) ? " [f32]" : "")
+      << ((bool)(q.prop & property::f16_accumulate) ? " [f16_acc]" : "");
     return s.str();
 }
 std::string to_string(const config_criteria_t &c) {
@@ -64,7 +65,8 @@ std::string to_string(const config_criteria_t &c) {
       << ((bool)(c.prop & property::quantized) ? " quant" : "")
       << ((bool)(c.prop & property::integrated) ? " integ" : "")
       << ((bool)(c.prop & property::fma) ? " fma" : "")
-      << ((bool)(c.prop & property::f32) ? " f32" : "");
+      << ((bool)(c.prop & property::f32) ? " f32" : "")
+      << ((bool)(c.prop & property::f16_accumulate) ? " f16_acc" : "");
     return s.str();
 }
 std::string to_string(const config_t &c) {
@@ -97,6 +99,11 @@ bool operator==(const config_record_t &key, const config_query_t &query) {
                     && (((key.criteria.prop & property::f32) == property::none)
                             || ((query.prop & property::f32)
                                     == (key.criteria.prop & property::f32)))
+                    && (((key.criteria.prop & property::f16_accumulate)
+                                == property::none)
+                            || ((query.prop & property::f16_accumulate)
+                                    == (key.criteria.prop
+                                            & property::f16_accumulate)))
                     && (((key.criteria.prop & property::integrated)
                                 == property::none)
                             || ((query.prop & property::integrated)
@@ -115,6 +122,7 @@ bool operator<(const config_criteria_t &lhs, const config_criteria_t &rhs) {
         if ((int)(crit.prop & property::integrated)) { set_fields++; }
         if ((int)(crit.prop & property::fma)) { set_fields++; }
         if ((int)(crit.prop & property::f32)) { set_fields++; }
+        if ((int)(crit.prop & property::f16_accumulate)) { set_fields++; }
         return set_fields;
     };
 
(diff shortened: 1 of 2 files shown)


syurkevi:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

syurkevi:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

src/gpu/intel/sdpa/configs.cpp
```diff
@@ -115,6 +122,7 @@ bool operator<(const config_criteria_t &lhs, const config_criteria_t &rhs) {
         if ((int)(crit.prop & property::integrated)) { set_fields++; }
         if ((int)(crit.prop & property::fma)) { set_fields++; }
         if ((int)(crit.prop & property::f32)) { set_fields++; }
+        if ((int)(crit.prop & property::f16_accumulate)) { set_fields++; }
```
Simonsays095: nit: I think we could improve the `num_set_fields` function so that we don't have to change this code if properties are added - using a helper `popcount` function to count the number of set flags.
```
template <typename Enum>
int popcount(Enum e) {
    using U = std::underlying_type_t<Enum>;
    U x = static_cast<U>(e);
    int count = 0;
    while (x) {
        x &= (x - 1); // clear lowest set bit
        ++count;
    }
    return count;
}
```
We could also use macros to utilize compiler intrinsics (there's an implementation of this in ngen, as `popcnt`) instead of this portable definition.

src/gpu/intel/sdpa/configs.cpp
```diff
@@ -232,10 +246,11 @@ static std::vector<config_record_t> sorted_configs = []() {
         {{compute::gpu_arch_t::xe_hpg, 128,      quantized | second_token}, {16, 16, 16, 8, 16, 2, 8, 4}},
 
         {{compute::gpu_arch_t::xe_hpg, 128, fma},                {8, 16, 16, 16, 8, 4, 8, 4}},
-        {{compute::gpu_arch_t::xe_hpg, 128, 448, fma},           {8, 16, 16, 16, 8, 2, 8, 2}},
+        {{compute::gpu_arch_t::xe_hpg, 128, 448, fma},           {8, 16, 16, 16, 8, 4, 8, 4}},
+        {{compute::gpu_arch_t::xe_hpg, 128, 448, fma | f16_accumulate}, {8, 16, 16, 16, 8, 2, 8, 2}},
         {{compute::gpu_arch_t::xe_hpg, 128, fma | second_token}, {32, 16, 32, 8, 16, 2, 8, 4}},
 
-        {{compute::gpu_arch_t::xe_hpg, 128, fma | f32},                            { 8, 32,  8, 32, 32, 1, 32, 1 }},
+        {{compute::gpu_arch_t::xe_hpg, 128, fma | f32},                            { 8, 16, 32,  8,  8, 2,  4, 4 }},
```
Simonsays095, line 258: Just to make sure I understand these config changes: In addition to adding the `f16_accumulate` flag to one config, several others are changed. Are these increases in the work group tile size (via increased unrolls or work group size), just because they'll no longer be used by f16 accumulate problems, and so there's a bit more register wiggle room? Trying to understand if this is 2 separate changes mashed together, or if this is related to the new property.

syurkevi: These are technically two separate changes. The previous non-f16 accumulate configs would give better performance on linux but would cause correctness issues on windows. I'll bring these out into a separate commit...

syurkevi: Have placed the revert in a separate PR. The other configs are just tuned for best performance for each case(f16 acc or not).

syurkevi:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg

PR #3977, opened 2025-09-22
cpu: aarch64: fix remaining modernize-use clang-tidy failures

Anndrey24:
# Description

This commit fixes the rest of the "modernize-use-*" clang-tidy failures on the AArch64 path.  
More specifically, it addresses these checks:
 - modernize-use-using
 - modernize-use-override
 - modernize-use-equals-default


jondea (approved):
Great, thanks!

PR #3978, opened 2025-09-22
Extend set_scales with quantization formula

mgouicem:
This implements option 1b option of RFC.

In a nutshell:
- set_scales is extended to accept an extra parameter for user to specify the quantization formula. Two are proposed here: static (with scale/zp provided by user and `x_f32 = scale * (x_q - zp)`) and dynamic_mx (with scale computed by oneDNN using OCP MX formula, and applied as `x_f32 = scale * x_q`)
- here we can use set_scale_v2 as no release was made with set_scale_v2 yet.
- matmul CPU ref path is extended to support dynamic quantization of destination for MXFP types (f8, f4)
- benchdnn is extended with extra knob in scales to specify formula for scale.

TODO:
- [x] documentation (both library and benchdnn)
- [x] maybe come up with better name than `quantization_kind::static_` as it conflict with static keyword.

### DIFF
--- a/scripts/verbose_converter/src/benchdnn_generator.py
+++ b/scripts/verbose_converter/src/benchdnn_generator.py
@@ -195,6 +195,8 @@ class Converter(metaclass=ConverterMeta):
             # Set policy to "host_scalar" if is_host_scalar is True
             if param.is_host_scalar:
                 policy = "host_scalar"
+            if param.quantization_mode == "dynamic_mx":
+                policy = "mx"
             result = f"{arg}:{policy}"
             if policy == "common" or policy == "host_scalar":
                 result += f":{def_value}"
--- a/scripts/verbose_converter/src/ir.py
+++ b/scripts/verbose_converter/src/ir.py
@@ -307,6 +307,7 @@ class QuantizationParam(Mapping):
     mask: int = 0
     groups: str = ""
     is_host_scalar: bool = False
+    quantization_mode: str = ""
 
     def __str__(self):
         if self.groups:
--- a/scripts/verbose_converter/src/parse.py
+++ b/scripts/verbose_converter/src/parse.py
@@ -411,7 +411,7 @@ class ParserImpl:
     @staticmethod
     def parse_quantization_param(spec, read_value, param_type):
         # Old style: mask[:[value[*]|*]]
-        # New style: mask[:data_type[:groups]]
+        # New style: mask[:data_type[:host_scalar[:groups[:quantization_mode]]]]
         param = param_type()
         param.mask = spec.read_uint()
         if spec.read_literal(":"):
@@ -429,6 +429,9 @@ class ParserImpl:
                         param.is_host_scalar = True
                     else:
                         param.groups = groups_or_host_flag
+                    if spec.read_literal(":"):
+                        param.quantization_mode = spec.read_str()
+
         return param
 
     # v2.7 and below


include/oneapi/dnnl/dnnl.h
```diff
@@ -481,6 +481,34 @@ dnnl_status_t DNNL_API dnnl_primitive_attr_set_scales_v2(
...
+///     each index along that dimension. Set the mask to 0 to use a common
+///     scaling factor for the whole output tensor.
+/// @param ndims Number of group dimensions.
+/// @param group_dims Scaling factors correspondence groups that define the
+///     correspondence between the tensor dimensions and the scales array.
+///     The group dimensions should only be provided for each logical dimension
+///     that has correspondence mask @p mask set.
+/// @param data_type Scaling factors data_type.
+/// @param is_on_host Indicates whether the scale is a host-side scalar.
+/// @param qkind Quantization mode, can be static or dynamic with specific
```
densamoilov: with specific...?

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -320,6 +320,27 @@ inline dnnl_rounding_mode_t convert_to_c(rounding_mode mode) {
     return static_cast<dnnl_rounding_mode_t>(mode);
 }
 
+/// Quantization kind
+enum class quantization_kind {
```
densamoilov: If we define it as a "mode" and prefer to use this term in most cases then it would make sense to name it that way as we did it for `scratchpad_mode`, `rounding_mode`, `fpmath_mode`, `accumulation_mode`.

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -320,6 +320,27 @@ inline dnnl_rounding_mode_t convert_to_c(rounding_mode mode) {
...
+    /// used for unspecified quantization kind
+    undef = dnnl_quantization_kind_undef,
+    /// static quantization mode: quantization parameter is computed
+    /// ahead of time and passed to oneDNN as an input.
+    static_ = dnnl_quantization_kind_static,
```
densamoilov: Would it be worth having a little redundancy to avoid using a trailing underscore in the public API?
Maybe use prefix `quant` for all names?

include/oneapi/dnnl/dnnl.hpp
mgouicem: We use both kind and mode in lib. From quick review it seemed to me the following was applied:
- when defining specific nature of an object, we use kind (alg_kind, engine_kind, format_kind, prop_kind)
- when it is less specific to a single object but affects its behavior more generally we use mode (rounding_mode, fpmath_mode, ...).

For this particular case, I used `kind` as it defines the natures of what is contained in that specific scales buffer (similar to alg_kind). Had it been a single setting for the primitive that would govern all scales passed, maybe mode would be more appropriate.

I am not too hung up on the choice of words here though, if mode seems more adequate, I can change that.

include/oneapi/dnnl/dnnl.h
mgouicem: fixed, thanks!

include/oneapi/dnnl/dnnl.hpp
mgouicem: I specified `static_sazp` for `scale after zero-point`, as we use the formula $` x_{f32} = scale * (x_{quant} - zp)`$.
This leave open using `static_sbzp` for `scale before zero-point` for the formula $` x_{f32} = scale * x_{quant} - zp`$ (like tiny-gemm and some torchAO paths).

include/oneapi/dnnl/dnnl.h
dzarukin: We usually put values explicitly with kinds, thus, this row might still need an update with latest changes.

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -320,6 +320,27 @@ inline dnnl_rounding_mode_t convert_to_c(rounding_mode mode) {
     return static_cast<dnnl_rounding_mode_t>(mode);
 }
 
+/// Quantization kind
+enum class quantization_kind {
+    /// used for unspecified quantization kind
+    undef = dnnl_quantization_kind_undef,
+    /// static quantization mode: quantization parameter is computed
+    /// ahead of time and passed to oneDNN as an input.
```
dzarukin, line 328: You might want to copy the text from C API.

src/common/matmul_pd.hpp
```diff
@@ -233,7 +233,8 @@ struct matmul_pd_t : public primitive_desc_t {
                         && utils::one_of(mask, 0, dst_qmask_N(),
                                 dst_qmask_M() + dst_qmask_N());
                 ok = ok
-                        && IMPLICATION(!scales.get(arg).has_default_groups(),
+                        && IMPLICATION(!scales.get(arg).has_default_groups()
+                                        && !scales.get(arg).is_dynamic(),
```
dzarukin: Could you explain, please, why dynamic scales don't require a check for groups?

Besides, I'm not sure why there are groups over M in the static case...

src/common/primitive_attr_quant.hpp
```diff
@@ -95,6 +97,10 @@ struct quant_entry_t : public c_compatible {
         return group_dims_[d];
     }
     bool is_host_scalar() const { return is_host_scalar_; }
+    quantization_kind_t get_quantization_kind() const { return qkind_; }
+    bool is_dynamic() const {
+        return qkind_ == dnnl_quantization_kind_dynamic_mx;
```
dzarukin:
```suggestion
        return qkind_ == quantization_kind::dynamic_mx;
```

src/common/primitive_exec_types.cpp
```diff
@@ -71,7 +71,8 @@ status_t cvt_primitive_args(const primitive_desc_t *pd, int nargs,
                 args[arg] = {mem, false};
                 n_outputs++;
                 extra_outputs += (arg == DNNL_ARG_SCRATCHPAD)
-                        || (arg == DNNL_ARG_ATTR_DROPOUT_MASK);
+                        || (arg == DNNL_ARG_ATTR_DROPOUT_MASK)
+                        || (arg == (DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST));
```
dzarukin: Maybe straight `(arg & DNNL_ARG_ATTR_SCALES)` to save for `DNNL_ARG_DIFF_SRC` or what may come...

src/common/primitive_attr_quant.hpp
```diff
@@ -157,6 +163,7 @@ struct quant_entry_t : public c_compatible {
     int group_ndims_ = 0;
     dims_t group_dims_ {};
     bool is_host_scalar_ = false;
+    quantization_kind_t qkind_ = quantization_kind::undef;
```
dzarukin: `operator==` wasn't updated for both objects.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -55,6 +55,9 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
     DEFINE_ARG_SCALES_BUFFER(src_scales, DNNL_ARG_SRC);
     DEFINE_ARG_SCALES_BUFFER(wei_scales, DNNL_ARG_WEIGHTS);
     DEFINE_ARG_SCALES_BUFFER(dst_scales, DNNL_ARG_DST);
+    // No need to zero-pad dynamic scales as they should not be blocked
+    auto dst_dynamic_scales
+            = CTX_OUT_MEM(float *, DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST);
```
dzarukin, line 79: This might go away after rebase.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -130,6 +135,15 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
...
 
     const bool with_dst_scales = !attr_scales.has_default_values(DNNL_ARG_DST);
+    const auto &dst_scale_dt = attr_scales.get_data_type(DNNL_ARG_DST);
+    const auto dst_scale_mask = attr_scales.get_mask(DNNL_ARG_DST);
+    const auto dst_scale_group_m = attr_scales.get_group(DNNL_ARG_DST, 0);
+    const auto dst_scale_group_n = attr_scales.get_group(DNNL_ARG_DST, 1);
+    memory_desc_t dst_scales_md {};
+    CHECK(attr_scales.get(DNNL_ARG_DST).get_md(dst_scales_md, *dst_d.md_));
+    VCHECK_ATTR(IMPLICATION(attr_scales.get(DNNL_ARG_DST).is_dynamic(),
+            dst_dynamic_scales != nullptr));
```
dzarukin: I'd say we generally don't check for empty pointers.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -221,41 +235,118 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
...
+            memory_tracking::names::key_matmul_dst_in_acc_dt);
+
+    // Note 1: If dst type is < 8 bits, we cannot split a byte during
+    //         store or we get a race condition. To simplify logic, we limit
+    //         parallelization on M and N by a factor of 2.
+    // Note 2: dynamic quantization requires to compute a reduction
+    //         along dst groups, so we split parallelization accordingly
+    dim_t M_chunk = std::max<dim_t>(2, dst_scale_group_m);
+    dim_t N_chunk = std::max<dim_t>(2, dst_scale_group_n);
+    parallel_nd(batch, utils::div_up(M, M_chunk), utils::div_up(N, N_chunk),
```
dzarukin: Let's introduce `M_chunks` and `N_chunks` and rename `M_chunk` to `M_chunk_size`, so that lambda doesn't shrink to 40 symbols, it should lead to better readability.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -221,41 +235,118 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
...
+                            args.l_offset = offset;
+                            args.dst_md = pd()->dst_md();
+                            ref_post_ops->execute(d, args);
+                        }
+                        if (attr_scales.get(DNNL_ARG_DST).is_dynamic()) {
+                            max_dst_group = std::max(max_dst_group, ::fabsf(d));
+                            io::store_float_value(
+                                    data_type::f32, d, temp_dst, dst_off);
+                        } else {
+                            if (with_dst_scales) d *= dst_scales[0];
```
dzarukin: This will require a division after rebase.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -221,41 +235,118 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
...
+                                d = math::stochastic_round_fwd(d, dst_off,
+                                        rnd_seed[0], dst_d.data_type());
+                            io::store_float_value(
+                                    dst_d.data_type(), d, dst, dst_off);
+                            utils::dim_iterator(
+                                    dst_d.dims(), dst_dims_idx, batch_ndims);
+                        }
+                    }
+
+                    if (attr_scales.get(DNNL_ARG_DST).is_dynamic()) {
```
dzarukin: Is it supposed to e a maximum within the processed group or within the tensor?

src/cpu/matmul/ref_matmul.hpp
```diff
@@ -121,10 +121,23 @@ struct ref_matmul_t : public primitive_t {
...
+
             return status::success;
         }
 
     private:
+        void init_scratchpad() {
+            using namespace memory_tracking::names;
+            auto scratchpad = scratchpad_registry().registrar();
+            if (attr()->scales_.get(DNNL_ARG_DST).is_dynamic()) {
+                // TODO: ideally, we should only use nthr * group size.
```
dzarukin: Would you try to address it?

src/cpu/reorder/cpu_reorder.cpp
```diff
@@ -40,6 +40,7 @@ regular_impl_list_map() {
             {{f4_e3m0, data_type::undef, 0}, &regular_fp4_impl_list_map()},
             {{f8_e5m2, data_type::undef, 0}, &regular_fp8_impl_list_map()},
             {{f8_e4m3, data_type::undef, 0}, &regular_fp8_impl_list_map()},
+            {{e8m0, f32, 0}, &regular_fp8_impl_list_map()},
```
dzarukin, line 43: Any specific reason to have just f32 and not any data type instead?

tests/benchdnn/dnn_types.cpp
```diff
@@ -187,6 +189,7 @@ int attr_t::get_default_mask(policy_t policy, int ndims) {
         case PER_OC:
         case PER_DIM_1: return (1 << 1);
         case PER_OCIC:
+        case MX:
```
dzarukin: I think this should use `PER_TENSOR` policy instead to cover 3D, 4D and 5D matmul tensors as well.

tests/benchdnn/dnn_types.cpp
```diff
@@ -231,6 +241,7 @@ int attr_t::policy2mask(int arg, policy_t policy, int ndims,
         switch (policy) {
             case PER_DIM_1:
             case PER_OC: return (1 << (ndims - 1));
+            case MX:
```
dzarukin: Ditto.

tests/benchdnn/dnn_types.cpp
```diff
@@ -1268,12 +1280,7 @@ dnnl_primitive_attr_t create_dnnl_attr(
             if (as.is_def(arg_name)) continue;
 
             const auto &e = arg.second;
-            if (e.policy == policy_t::HOST_SCALAR) {
-                DNN_SAFE_V(dnnl_primitive_attr_set_scales_v2(dnnl_attr,
-                        arg_name, 0 /* mask */, 0 /* ndims */, nullptr, e.dt,
-                        true /* is_on_host = true */));
-                continue;
-            }
+            bool is_on_host = e.is_host_scalar();
```
dzarukin: It's rather `int` for C API. Suggest to align the type with the API call.

tests/benchdnn/dnn_types.hpp
```diff
@@ -224,6 +226,7 @@ struct attr_t {
             }
 
             bool is_host_scalar() const { return policy == HOST_SCALAR; }
+            bool is_mx() const { return policy == MX; }
```
dzarukin, line 229: Maybe `is_dynamic` to align with the library internals?

tests/benchdnn/dnn_types.hpp
```diff
@@ -81,6 +82,7 @@ struct attr_t {
     static int policy2mask(int arg, policy_t policy, int ndims = -1,
             dnnl_primitive_kind_t prim_kind = dnnl_undefined_primitive,
             bool has_groups = false);
+    static dnnl_quantization_kind_t policy2quantization_kind(policy_t policy);
```
dzarukin: It doesn't look like it should be `static` so far.

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -8,8 +8,8 @@
     --attr-rounding-mode=ARG:MODE[+...]
     --attr-deterministic=BOOL
     --attr-dropout=PROBABILITY[:SEED[:TAG]]
-    --attr-scales=ARG:POLICY[:SCALE][:DATA_TYPE[:GROUPS]][+...]
-    --attr-zero-points=ARG:POLICY[:ZEROPOINT][:DATA_TYPE[:GROUPS]][+...]
+    --attr-scales=ARG:POLICY[:SCALE[:DATA_TYPE[:GROUPS]]][+...]
+    --attr-zero-points=ARG:POLICY[:ZEROPOINT[:DATA_TYPE[:GROUPS]]][+...]
```
dzarukin: The original position was to reflect that `:SCALE` can't be present with `:GROUPS`, but probably, it makes more sense to introduce two lines:
```
--attr-scales=ARG:POLICY[:SCALE[:DATA_TYPE]][+...]
              ARG:POLICY[:DATA_TYPE[:GROUPS]][+...]
```
and same for ZPs.

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -18,8 +18,6 @@
 --attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1
 --batch=shapes_4bit
 
-# --dt=f4_e2m1:f4_e2m1:f32,f4_e2m1:f4_e2m1:bf16
-# --attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1
-#--batch=shapes_4bit
-
-
+--dt=f4_e2m1
+--attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1+dst:mx:e8m0:1x32
```
dzarukin:
```suggestion
--attr-scales=src:per_tensor:e8m0:1x32+wei:per_tensor:e8m0:32x1+dst:mx:e8m0:1x32
```

tests/benchdnn/inputs/matmul/shapes_mx
```diff
@@ -0,0 +1,8 @@
+24x32:32x64
+96x96:96x64
+14x96:96x32
+2048x1024:1024x512_n"DLRM:5*1"
+2048x256:256x128_n"NCF:1*1"
+2048x128:128x64_n"NCF:2*1"
+896x1024:1024x4096_n"RNN-T:Encoder_cell1_Hidden*11"
+896x320:320x1280_n"RNN-T:Prediction_Input*12"
```
dzarukin, line 8: Please stick in a 4D shape.

tests/benchdnn/inputs/matmul/test_matmul_fp8
```diff
@@ -86,3 +86,11 @@
 8x2664:2664x256
 3x4128:4128x128
 
+## MXFP8
+--dt=f8_e4m3:f8_e4m3:f32,f8_e4m3:f8_e4m3:bf16
+--attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1
```
dzarukin: Ditto.

tests/benchdnn/inputs/matmul/test_matmul_fp8
```diff
@@ -86,3 +86,11 @@
...
+--batch=shapes_mx
+
+--dt=f8_e4m3
+--attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1+dst:mx:e8m0:1x32
```
dzarukin: Ditto.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -577,10 +577,11 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
         if (!prb->attr.scales.is_def(DNNL_ARG_DST)
-                && prb->attr.scales.get(DNNL_ARG_DST).policy
-                        != attr_t::COMMON) {
+                && prb->attr.scales.get(DNNL_ARG_DST).policy != attr_t::COMMON
+                && prb->attr.scales.get(DNNL_ARG_DST).policy != attr_t::MX) {
             BENCHDNN_PRINT(2,
-                    "[SKIP][%s:%d]: Only Common dst scales are supported "
+                    "[SKIP][%s:%d]: Only Common and MX dst scales are "
+                    "supported "
                     "on CPU.\n",
```
dzarukin: Please fuse into a previous line.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -1082,6 +1083,8 @@ std::vector<data_kind_t> get_kinds_to_check(const prb_t *prb) {
     //       e.g. get_kinds_to_check_default_case
     std::vector<data_kind_t> check_kinds = {DST};
     if (!prb->attr.dropout.is_def()) check_kinds.push_back(DROPOUT_MASK);
+    if (prb->attr.scales.get(DNNL_ARG_DST).is_mx())
```
dzarukin: I'd check for `is_def()` first just for extra safety in case a default changes or something.

tests/benchdnn/utils/data_kind.cpp
```diff
@@ -57,6 +57,7 @@ static data_kind_entry_t data_kind_table[] = {
                         DNNL_ARG_WEIGHTS_PROJECTION}},
         {DAT_TOTAL, {DNNL_ARG_SCRATCHPAD}},
         {DROPOUT_MASK, {DNNL_ARG_ATTR_DROPOUT_MASK}},
+        {DST_SCALES, {DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST}},
```
dzarukin, line 59:
```suggestion
        {DROPOUT_MASK, {DNNL_ARG_ATTR_DROPOUT_MASK}},
        {DST_SCALES, {DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST}},
        // Always keep this entry the last in the list.
        {DAT_TOTAL, {DNNL_ARG_SCRATCHPAD}},
```

src/cpu/matmul/ref_matmul.cpp
mgouicem: Maximum for the processed group. This is why it is computed inside the loop nest.

src/cpu/reorder/cpu_reorder.cpp
mgouicem: `e8m0 -> f32` is needed for testing (checking computed scales).
If I was to add others, would need to add ci coverage, and I do not know of use cases for those other flavors.

tests/benchdnn/dnn_types.hpp
mgouicem: I switched it the other way around as `is_mx` is more specific.

tests/benchdnn/dnn_types.hpp
mgouicem: I followed existing style of not requiring an object for those conversions.
I am fine either way to be honest, but why make this one different than policy2mask?

tests/benchdnn/utils/data_kind.cpp
mgouicem: Applied, but why scratchpad?

src/common/matmul_pd.hpp
mgouicem: I honestly don't know what the group requirement comes from wrt to static quantization.

In any case, the above restriction does not fly with mx quantization, as it requires group of 1 along N, whereas mx quantization of dst activations requires group of 32 along N.

mgouicem: I added a simple check on divisibilty as it is common to existing check and mx requirement.
I removed the requirement for group along n to be equal to 1 though.

tests/benchdnn/matmul/matmul.cpp
mgouicem: Done.

src/cpu/matmul/ref_matmul.cpp
mgouicem: I kept it simply for differentiating cases where dst_scales are inputs (const and CTX_IN_MEM) or outputs (non-const and CTX_OUT_MEM)

src/cpu/matmul/ref_matmul.hpp
mgouicem: Done.

tests/benchdnn/doc/knobs_attr.md
mgouicem: Good point, I split it in two to make it clearer

tests/benchdnn/inputs/matmul/shapes_mx
mgouicem: done

include/oneapi/dnnl/dnnl.hpp
densamoilov: Yes, we do use both, and I see your point about using kind, especially since it relates to the contents of the scales buffer. But to me, this feels more like a setting that controls _how_ quantization behaves, not a fundamentally different type of object.

We tend to use `*_mode` for behavioral settings (`rounding_mode`, `fpmath_mode`, etc.), while `*_kind` is more about defining _what_ something is, like `alg_kind` or `engine_kind`.

That said, the line isn't always clear. For example, `alg_kind` includes average and max pooling, which are different algorithms, yes, but within the context of pooling, they could easily be seen as just different modes. So whether we use "kind" or "mode" often depends on perspective (context).

In this case, since the enum controls how quantization is applied (statically or dynamically), I think `quantization_mode` might better reflect that and align with existing naming. But I’m not strongly opposed if you prefer to keep kind.

include/oneapi/dnnl/dnnl.hpp
densamoilov: Does `sazp` stand for `Sharpness-Aware Zero-shot Quantization`? 🤔

include/oneapi/dnnl/dnnl.h
```diff
@@ -481,6 +481,35 @@ dnnl_status_t DNNL_API dnnl_primitive_attr_set_scales_v2(
         const dnnl_dims_t group_dims, dnnl_data_type_t data_type,
         int is_on_host);
 
+/// Sets primitive attributes scaling factors for primitive operations for a
```
densamoilov, line 484: (random spot) Is it still POC or it's already production ready code?

densamoilov (approved):

tests/benchdnn/utils/data_kind.cpp
dzarukin: There was no `DNNL_ARG_UNDEF` back in the day, you may update that for `DAT_TOTAL`. And thanks for taking care of that.

tests/benchdnn/dnn_types.hpp
dzarukin: policy2mask was converted to `static` to cover usages outside of attributes. A new function doesn't seem to do be used outside of attrs, and then it could reuse some new methods introduced to avoid code duplication.

include/oneapi/dnnl/dnnl.hpp
mgouicem: ok quantization_mode it is :)

include/oneapi/dnnl/dnnl.hpp
mgouicem: I aligned the two.

src/common/primitive_attr.cpp
```diff
@@ -577,16 +577,26 @@ status_t dnnl_primitive_attr_set_scales_mask(
...
+status_t dnnl_primitive_attr_set_scales_v3(primitive_attr_t *attr, int arg,
+        int mask, int group_ndims, const dims_t group_dims,
+        data_type_t data_type, int is_on_host, quantization_mode_t qmode) {
     using namespace data_type;
     VCHECK_ATTR(attr, VERBOSE_NULL_ARG);
     VCHECK_ATTR(arg >= 0, VERBOSE_BAD_PARAM, "arg");
+    VCHECK_ATTR(utils::one_of(qmode, quantization_mode::static_sazp,
+                        quantization_mode::dynamic_mx),
+            VERBOSE_BAD_PARAM, "qmode");
     VCHECK_ATTR(
```
mzhukova, line 600: for dynamic quantization, should we also check for 32 group size, allowed dt (and maybe that arg is output)?

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -348,3 +350,10 @@ Run a matmul problem with weights decompression where weights `bf16` scales and
              --attr-scales=wei:per_ocic:bf16:2x1 \
              --attr-zero-points=wei:per_ocic:s8:2x1 6x4:4x5
 ```
+
+Run a matmul problem with MXFP4 semantics:
+``` sh
+  ./benchdnn --matmul --dt=f4_e2m1 \
+             --attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1 \
+             --attr-dynamic-scales=dst:per_ocic:e8m0:1x32 32x32:32x64
```
mzhukova: should it be `--attr-scales=dst:mx:e8m0:1x32` here instead of `--attr-dynamic-scales`?

tests/benchdnn/doc/knobs_attr.md
mzhukova, line 88: if I followed the implementation correctly, we probably need a description of new `mx` policy here?

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -15,11 +15,13 @@
 
 ## MXFP4
 --dt=f4_e2m1:f4_e2m1:f32,f4_e2m1:f4_e2m1:bf16
---attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1
---batch=shapes_4bit
-
-# --dt=f4_e2m1:f4_e2m1:f32,f4_e2m1:f4_e2m1:bf16
-# --attr-scales=src:per_ocic:e8m0:1x32+wei:per_ocic:e8m0:32x1
-#--batch=shapes_4bit
+--attr-scales=src:per_tensor:e8m0:1x32+wei:per_tensor:e8m0:32x1
```
mzhukova, line 18: is this intentionally missing mx policy? (even though the comment above says MXFP4..?) just checking :)

src/common/primitive_attr.cpp
mzhukova: also, could you extend `set_scales` iface gtests with some of these new scenarios?

doc/programming_model/attributes_quantization.md
```diff
@@ -492,22 +517,3 @@ impossible to apply them on-the-fly without potential accuracy loss.
            attr);   // the attributes describe the quantization flow
 // ...
 ~~~
```
mzhukova, line 515: do you plan to also extent either doc or examples (I think it would be better!) with MXFP8 example?

tests/benchdnn/inputs/matmul/test_matmul_fp4
mgouicem: For these test cases, inputs are in "MXFP format", but outputs are bf16/f32.
This is why dst does not use MX policy.

src/common/primitive_attr.cpp
mgouicem: > for dynamic quantization, should we also check for 32 group size, allowed dt (and maybe that arg is output)?

It is checked in `src/common/matmul.cpp` but not in common part. This is because we don't know which layout is needed for grouping (1x32 vs 32x1) unless we know how argument will be used.

> also, could you extend set_scales iface gtests with some of these new scenarios?

Done, thanks.

tests/benchdnn/doc/knobs_attr.md
mgouicem: Fixed, thanks

tests/benchdnn/doc/knobs_attr.md
mgouicem: Added, thanks.

src/common/primitive_attr.cpp
mzhukova: oh, I see, thanks for the clarification!

tests/benchdnn/inputs/matmul/test_matmul_fp4
mzhukova: got it :)

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
-  (e.g. one scale per tensor, one scale per channel, ...).
+oneDNN support two main categories of quantization:
+- static quantization with scales only (symmetric) or scales and
+  zero-points (asymmetric), where scales are applied after zero-point.
+- dynamic quantization compliant with the Open Compute Project (OCP)
+  Microscaling (MX) [formats
+  specification](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf).
+
+To support quantization, primitives should be created and executed as
+follow:
```
vpirogov:
```suggestion
To support quantization, primitives should be created and executed as
follows:
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+
+- during primitive descriptor creation, if one or multiple
+  inputs/outputs use low precision datatype (e.g. `s8` or `fp8_e4m3`),
+  then the primitive will behave as a quantized operation.
```
vpirogov: Edits for clarity.

```suggestion
- during primitive descriptor creation source, weights, or destination
  memory descriptors use low precision datatype (e.g. `s8` or `fp8_e4m3`).
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+- still during primitive descriptor creation, the dimensionality
+  (using masks), group size and data type of the scaling factors and
+  zero-point should be provided using primitive attributes.
```
vpirogov:
```suggestion
- during primitive descriptor creation group size, data types, and broadcasting masks of
   the scaling factors and zero-point are provided using primitive attributes.
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
 - finally, during primitive execution, the user must provide the
   actual quantization parameters as arguments to the execute function.
```
vpirogov:
```suggestion
- during primitive execution the actual quantization parameters
  are provided as arguments to the execute function.
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+optimized is listed in each primitive documentation page.
 
 This guide does not cover how the appropriate scaling factor can be found.
 Refer to the materials in the [Introduction](@ref dgaq_intro).
 
-### Numerical behavior
+### Static quantization
+
+The only supported formula for static quantization currently supported
+by oneDNN is with scales applied after zero-point as follow:
```
vpirogov:
```suggestion
by oneDNN is with scales applied after zero-point as follows:
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+\f]
+
+where \f$x_{f32}\f$ and \f$x_{quant}\f$ are the non-quantized and
+quantized representation of \f$x\f$ respectively, \f$scale_{x}\f$ is a
+*scaling factor* in a floating-point format, \f$zp_{x}\f$ is a *zero
+point* (typically in integral format), and \f$[:]\f$ is used to denote
+elementwise application of the formula to the arrays. 
+
+In this model, oneDNN assumes that quantization parameters are
+inputs provided by the user the library does not compute those scaling
```
vpirogov:
```suggestion
inputs provided by the user and the library does not compute those scaling
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+
+\f[
+x_{f32}[:] = scale_{x} \cdot x_{quant}[:] 
+\f]
+
+where \f$x_{f32}\f$ and \f$x_{quant}\f$ are the non-quantized and
+quantized representation of \f$x\f$ respectively, and \f$scale_{x}\f$ is a
+*scaling factor*:
+- in e8m0 format,
+- computed for each group of size 32 (see set_scales group),
```
vpirogov: Link to `set_scales`?

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,105 @@ Related materials:
...
+- and computed as the largest power-of-two less than or equal to the
+  maximum abolute value of the group divided by the largest
```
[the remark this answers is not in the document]
vpirogov: Good bot.

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
+Refer to the materials in the [Introduction](@ref dgaq_intro).
+
+### Static quantization
+
+The only supported formula for static quantization currently supported
+by oneDNN is with scales applied after zero-point as follows:
+
 \f[
-    x_{f32}[:] = scale_{x} \cdot (x_{int8}[:] - zp_{x})
+x_{f32}[:] = scale_{x} \cdot (x_{quant}[:] - zp_{x})
```
vpirogov:
```suggestion
x_{f32}[:] = scale_{x} \cdot (x_{quant}[:] - zp_{x})
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
+supports only a subset of quantization parameter masks, group sizes
+and data type combinations. Which combination is supported and
+optimized is listed in each primitive documentation page.
+
+This guide does not cover how the appropriate scaling factor can be found.
+Refer to the materials in the [Introduction](@ref dgaq_intro).
+
+### Static quantization
+
+The only supported formula for static quantization currently supported
```
vpirogov:
```suggestion
The only formula for static quantization currently supported
```

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
-for activation.
 
-This guide does not cover how the appropriate scaling factor can be found.
-Refer to the materials in the [Introduction](@ref dgaq_intro).
+### Dynamic quantization
+
+The only formula for dynamic quantization currently supported by
+oneDNN is with scales computed following the [Open Compute Project
+Microscaling specification version
+1](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf),
```
vpirogov: This link does not render correctly presumably due to line breaks.
image

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
 
 ## Quantization Model
 
-The primary quantization model that the library assumes is the following:
+oneDNN support two main categories of quantization:
+- static quantization with scales only (symmetric) or scales and
+  zero-points (asymmetric), where scales are applied after zero-point.
+- dynamic quantization compliant with the Open Compute Project (OCP)
+  Microscaling (MX) [formats
+  specification](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf).
```
dzarukin: This link might not be working as well.

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
 
-Primitive implementations are allowed to convert int8 inputs to wider
-datatypes (e.g. int16 or int32), as those conversions do not impact
-accuracy.
+where \f$x_{f32}\f$ and \f$x_{quant}\f$ are the non-quantized and
+quantized representation of \f$x\f$ respectively, and \f$scale_{x}\f$ is a
+*scaling factor*:
+- in e8m0 format,
+- computed for each group of size 32 (see [set_scales](@ref
+  dnnl::primitive_attr::set_scales)),
```
vpirogov: This link does not render correctly presumably because of line breaks.

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
+- and computed as the largest power-of-two less than or equal to the
+  maximum absolute value of the group divided by the largest
+  power-of-two representable in the \f$x_{quant}\f$ data type
+  (e.g. \f$E8M0(amax(x_quant[:])) / E8M0(MAX_X_QUANT_DT) \f$).
```
vpirogov: This one fails to render in a very odd way

image

doc/programming_model/attributes_quantization.md
```diff
@@ -21,62 +21,106 @@ Related materials:
...
-@ref dev_guide_attributes_post_ops_eltwise with clip algorithm.
+The scales are applied in single precision floating point data type
+(#dnnl::memory::data_type::f32) before downconversion to the
+destination datatype. When converting to integral datatypes,
+implementations typically saturate, whereas for floating-point
+datatypes, underflow/overflow can occur. To force saturation in
+floating-point datatypes use @ref
+dev_guide_attributes_post_ops_eltwise with clip algorithm. Rounding
+happens according to [rounding mode attribute](@ref
+dev_guide_attributes_rounding_mode).
```
vpirogov: Does not render correctly presumably due to line break.

image

include/oneapi/dnnl/dnnl.hpp
dzarukin: I still don't see a `sazp` description for C++ API. Maybe a bad rebase?

src/common/primitive_attr_quant.hpp
```diff
@@ -92,9 +94,18 @@ struct quant_entry_t : public c_compatible {
         // But if they were, any out of bound access would return `0` and likely
         // lead to a division by zero which is fast to catch.
         if (d >= group_ndims_) return 0;
+        // we allow negative indexes to address from last to first
+        if ((d < 0) && (d + group_ndims_ >= 0))
```
dzarukin: If second condition doesn't hold, it'll end up with accessing group_dims_[-x] which is simply a crash. 
It either a `|| cond` appended to the previous line, or an additional if-else branch to return 0 in case of wrong `d`.

BTW, why it's really needed given that there's no more than 2 groups used in the library? If you have an extension in mind, all places should be fixed with negative access, otherwise it's just gonna break in a different spot.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -35,6 +35,21 @@ namespace impl {
...
 namespace matmul {
 
+void ref_matmul_t::pd_t::init_scratchpad() {
+    using namespace memory_tracking::names;
+    nthr_ = dnnl_get_max_threads();
+    auto scratchpad = scratchpad_registry().registrar();
+    auto dst_scales = attr()->scales_.get(DNNL_ARG_DST);
+    if (dst_scales.is_mx()) {
+        const memory_desc_wrapper dst_d(dst_md());
+        dim_t group_size = dst_scales.get_group_size();
```
dzarukin, line 46: I'd say the implementation of this interface doesn't reflect the intent. In case of grouping over M happens, with or without grouping over N, then the code below won't work just because it assumes a plain indexing and striding (`offset`).

I suggest to keep aligned with limitations specified in API (meaning straightforwardly rely on grouping over N) to keep it simple and get rid of this method,

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -35,6 +35,21 @@ namespace impl {
...
+        dim_t ntasks = dst_d.nelems() / group_size;
+        nthr_ = std::min<dim_t>(nthr_, ntasks);
```
dzarukin: This change doesn't look right. The number of threads shouldn't depend on number of tasks.

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -58,6 +73,9 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
             = CTX_IN_MEM(const void *, DNNL_ARG_ATTR_SCALES | DNNL_ARG_WEIGHTS);
     const void *dst_scales
             = CTX_IN_MEM(const void *, DNNL_ARG_ATTR_SCALES | DNNL_ARG_DST);
+    // No need to zero-pad dynamic scales as they should not be blocked
```
dzarukin:
```suggestion
    // No need to zero-pad dynamic scales as they should have plain format.
```

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -132,6 +152,11 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
 
     const bool with_dst_scales = !attr_scales.has_default_values(DNNL_ARG_DST);
     const auto dst_scale_dt = attr_scales.get_data_type(DNNL_ARG_DST);
+    const auto dst_scale_mask = attr_scales.get_mask(DNNL_ARG_DST);
+    const auto dst_scale_group_m = attr_scales.get_group(DNNL_ARG_DST, 0);
```
dzarukin: 0 and 1 here? Not -1 and -2 introduced? :)

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -15,11 +15,13 @@
...
+--batch=shapes_mx
 
+--dt=f4_e2m1
+--attr-scales=src:per_tensor:e8m0:1x32+wei:per_tensor:e8m0:32x1+dst:mx:e8m0:1x32
+--batch=shapes_mx
 
+# Adding a 4d shapes outside of shapes_mx as it is incompatible with tags tested above
```
dzarukin: Minor: using `abx` on line 5 should solve the problem.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -121,90 +122,163 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
...
+                int src_zp = has_src_single_zp ? src_zps.get_elem(0) : 0;
+                int wei_zp = has_wei_single_zp ? wei_zps.get_elem(0) : 0;
+                float src_scale = has_src_single_scale
+                        ? src_scales.get_f32_elem(0)
+                        : 1.f;
+                float wei_scale = has_wei_single_scale
+                        ? wei_scales.get_f32_elem(0)
+                        : 1.f;
+                float *tmp_dst = (float *)malloc(
+                        dst_M_group * dst_N_group * sizeof(float));
```
dzarukin: Please avoid allocations inside reference, it hurts debugging OOM cases later. Given that reference always has f32 data type, I'd say original dst can be re-used for that purpose.

doc/programming_model/attributes_quantization.md
vpirogov: This one works fine. Doxygen [gets upset](https://www.markdownguide.org/basic-syntax/#link-best-practices) when link includes linebreaks. The best way to deal with that is using [Reference Style Links](https://www.markdownguide.org/basic-syntax/#reference-style-links)

src/cpu/matmul/ref_matmul.cpp
mgouicem: Not sure I understand your point here. How `offset` is used is untouched by this PR, and `offset` is not used for addressing scratchpad.

tests/benchdnn/matmul/ref_matmul.cpp
mgouicem: Unfortunately, this would not be compatible with using sum post-op, which uses dst as input.
Note that the amount of memory here is typically small (group size per thread) and allocated on host.

src/cpu/matmul/ref_matmul.cpp
mgouicem: This counts the number of active threads and is necessary to allocate less scratchpad on systems with large number of cores.

src/common/primitive_attr_quant.hpp
mgouicem: > If second condition doesn't hold, it'll end up with accessing group_dims_[-x] which is simply a crash.
It either a || cond appended to the previous line, or an additional if-else branch to return 0 in case of wrong d.

Good point, fixed.

> BTW, why it's really needed given that there's no more than 2 groups used in the library? If you have an extension in mind, all places should be fixed with negative access, otherwise it's just gonna break in a different spot.

Currently, API support for groups is clunky: we allow user to pass a mask of all ones, yet not provide a group dimension for each of those masked dimension. For example, 4d memory with `0xf` mask should require `group_ndims=4`, yet we accept `group_ndims=2` and assume it applies to last 2 dims.

What I tried to do here is a first step to allow normal use of the API, which is group_ndims should be equal to number of bit set in mask. This implies using -1/-2 to map groups to N/M in matmul (simpler than pushing logic inside impls).

mgouicem:
make docs

src/cpu/matmul/ref_matmul.cpp
mgouicem: Note that this is aligned with standard flow of parallel_ext:
- we call adjust_num_threads [from all instances](https://github.com/uxlfoundation/oneDNN/blob/edb3380221a7e9aec76e281bb13a2b0c1ac118fa/src/common/dnnl_thread.hpp#L533)
- adjust_num_threads uses the minimum between ntasks and max_num_threads [here](https://github.com/uxlfoundation/oneDNN/blob/edb3380221a7e9aec76e281bb13a2b0c1ac118fa/src/common/dnnl_thread.hpp#L270)

src/cpu/matmul/ref_matmul.cpp
dzarukin: `get_group_size` does the product of all groups while I'm not sure whether the code below would support non-unit grouping over M in dst scales. If it does, then the comment is irrelevant. If it doesn't, then I don't see the purpose of the groups product. I assumed the latter, but I could be wrong there.

src/cpu/matmul/ref_matmul.cpp
dzarukin: Adjusting `nthr_` is not mandatory to make scratchpad using less, a differen variable could be used for that.

Adjusting `nthr_` means the number of threads will likely be changed for OMP (with the next or prev layer) leading to re-creating of the global thread pool, which we avoid by keeping that value equal to `dnnl_get_max_threads()`. Dealing with less tasks is the job for parallel call or lambda to do.

src/common/primitive_attr_quant.hpp
dzarukin: I understand the purpose, but my preference is to fix it everywhere in the library with a one gulp to avoid mix-n-match which usually leads to worse consequences and longer debugging.

tests/benchdnn/matmul/ref_matmul.cpp
dzarukin: The value is written after sum post-op is applied. What am I missing?

src/cpu/matmul/ref_matmul.cpp
mgouicem: > I'm not sure whether the code below would support non-unit grouping over M in dst scales. 

It does support grouping over M or N or both. It is not tested though as I restricted fp matmul dispatching to mxfp groups only.
If I disable the check test cases as follow pass:
```
0:PASSED (483 ms) __REPRO: --matmul --allow-enum-tags-only=false --dt=f4_e2m1:f4_e2m1:f4_e2m1 --stag=abcd --wtag=abcd --dtag=abcd --bia-dt=f32 --attr-scales=src:per_tensor:e8m0:1x32+dst:mx:e8m0:1x32+wei:per_tensor:e8m0:32x16 6x2x14x96:6x2x96x32
```

dzarukin: `--attr-scales=...+dst:mx:e8m0:1x32+...` - this one doesn't have groups over M.

tests/benchdnn/matmul/ref_matmul.cpp
mgouicem: Indeed. I removed the allocation and use dst for temporary storage.
Thanks.

src/cpu/matmul/ref_matmul.cpp
mgouicem: I think the OMP overhead of managing multiple teams of thread is the least of our concern if ref_matmul is dispatched :).
Nevertheless, I adjusted the logic to use `min(default number of threads, work amount)` from creation time.

src/common/primitive_attr_quant.hpp
mgouicem: I fixed the occurences in ref_matmul.cpp. Would love to avoid scope creep on this PR and not stall this PR further, as other devs rely on this PR.
I will open another PR with broader changes, but right now it would be nice to get this piece going.

src/cpu/matmul/ref_matmul.cpp
mgouicem: Same:
```
0:PASSED (323 ms) __REPRO: --matmul --allow-enum-tags-only=false --dt=f4_e2m1:f4_e2m1:f4_e2m1 --stag=abcd --wtag=abcd --dtag=abcd --bia-dt=f32 --attr-scales=src:per_tensor:e8m0:1x32+dst:mx:e8m0:16x32+wei:per_tensor:e8m0:32x1 --ctx-init=2:auto:auto --ctx-exe=2:auto:auto 6x2x16x96:6x2x96x32
```

The code right now is generic and does not assume a particular grouping: it just loops over groups along M/N for dst. The kernel to compute each element in dst loops on groups along M/K for src and K/N for weights.

dzarukin: Alright, thanks for checking it. Resolving then.

tests/benchdnn/inputs/matmul/test_matmul_fp8
```diff
@@ -86,3 +86,15 @@
 8x2664:2664x256
 3x4128:4128x128
 
+## MXFP8
+--dt=f8_e4m3:f8_e4m3:f32,f8_e4m3:f8_e4m3:bf16
+--attr-scales=src:per_tensor:e8m0:1x32+wei:per_tensor:e8m0:32x1
+--batch=shapes_mx
```
dzarukin, line 94: Since you added 4D shape into shapes_mx, you need to fix tags on line 83 now and then you could remove lines 98-100. Or tests will fail due to tag mismatch.

dzarukin (approved):

tests/benchdnn/inputs/matmul/test_matmul_fp8
mgouicem: Thanks for the catch, updated tags, and removed explicit 4d shape from here.

mgouicem:
make docs

PR #3979, opened 2025-09-22
cpu: aarch64: Add support for dst s8 and s32 in brgemm int8 conv

kasturedeeksha:
# Description

This PR adds support for the s8 and s32 destination type in BRGEMM general and 1x1 forward convolution. It also fixes redundant s32 → f32 conversions that were happening in certain cases.

Supported int8 configurations
src : s8 / u8
wei : s8
dst : s8 / u8 / s32 / f32

# Checklist

## General

  1. make test output 

```
99% tests passed, 1 tests failed out of 226
 
Total Test time (real) = 1318.15 sec
 
The following tests FAILED:
        206 - test_benchdnn_modeC_graph_ci_cpu (Failed)
```

2. ./benchdnn --conv --batch=inputs/conv/test_conv_int8
```
tests:22609 passed:17839 skipped:3494 mistrusted:1276 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total: 374.92s; create_pd: 18.86s (5%); create_prim: 21.02s (6%); fill: 50.78s (14%); execute: 30.00s (8%); compute_ref: 104.78s (28%); compare: 124.41s (33%);
```

- [x] Have you formatted the code using clang-format?Yes


src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -518,14 +518,7 @@ void jit_brgemm_kernel_t::cvt2ps(data_type_t type_in, const ZReg zmm_in,
...
         case data_type::s8:
-            if (mask_flag)
-                set_preg(P_TMP.b, brg.ldb_tail, X_TMP_0, X_TMP_1);
-            else
-                set_preg(P_TMP.b, brg.ld_block, X_TMP_0, X_TMP_1);
-            add_imm(X_DEFAULT_ADDR, addr, offset - base_offset, X_TMP_0);
-            ld1b(z_tmp_1().b, P_TMP / T_z, ptr(X_DEFAULT_ADDR));
-            sunpklo(z_tmp_1().h, z_tmp_1().b);
-            sunpklo(z_tmp_1().s, z_tmp_1().h);
+            LD_MUL_VL(ld1sb, z_tmp_1().s, mask, addr, offset - base_offset, 1);
```
dzarukin, line 501: Minor: given the amount of recent contributions to brgemm, I'd highly recommend to turn this macro into a function of `jit_generator` or a standalone function that takes `jit_generator` by pointer to form a collection.

Macro is undebuggable and not reusable given there are already 4 version of it across aarch64 codebase.

Sqvid:
With asserts enabled `test_conv_int8` crashes due to your previous change in this file: https://github.com/uxlfoundation/oneDNN/pull/3911.
This change fixes that crash by replacing the assert but it should probably have checked for support earlier in `brgemm_desc_set_postops()`. Can we review the post-op support checks as part of this PR please? Thanks.

```
$ % ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8 --attr-scales=src:common:0.25+dst:common:2+wei:per_oc --attr-zero-points=src:common:135+dst:common:1 --attr-post-ops=sum:1:0:s8 ic3ih224oc64oh224kh3ph1n"vgg_19:conv1_1"
benchdnn: oneDNN/src/cpu/aarch64/jit_brgemm_post_ops.hpp:452: void dnnl::impl::cpu::aarch64::jit_brgemm_kernel_post_ops<isa>::cvt2ps(dnnl::impl::data_type_t, Xbyak_aarch64::ZReg, const Xbyak_aarch64::AdrNoOfs&, bool, bool, Xbyak_aarch64::PReg) [with dnnl::impl::cpu::aarch64::cpu_isa_t isa = dnnl::impl::cpu::aarch64::sve_256; dnnl::impl::data_type_t = dnnl_data_type_t]: Assertion `!"unsupported data type\n"' failed.
zsh: abort (core dumped)  ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8
```

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: In our case, the macro is also used to simplify calls to load/store instructions like ld1b, ld1w, etc. These instructions have several overloads and passing them directly as parameters not only leads to compilation errors due to mismatched definitions but also makes the code more complex.
For debugging purposes, the computation part has now been moved into inline functions inside `jit_generator.hpp` and a single macro is also defined there for load store each instead of in each file.

kasturedeeksha:
> With asserts enabled `test_conv_int8` crashes due to your previous change in this file: #3911. This change fixes that crash by replacing the assert but it should probably have checked for support earlier in `brgemm_desc_set_postops()`. Can we review the post-op support checks as part of this PR please? Thanks.
> 
> ```
> $ % ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8 --attr-scales=src:common:0.25+dst:common:2+wei:per_oc --attr-zero-points=src:common:135+dst:common:1 --attr-post-ops=sum:1:0:s8 ic3ih224oc64oh224kh3ph1n"vgg_19:conv1_1"
> benchdnn: oneDNN/src/cpu/aarch64/jit_brgemm_post_ops.hpp:452: void dnnl::impl::cpu::aarch64::jit_brgemm_kernel_post_ops<isa>::cvt2ps(dnnl::impl::data_type_t, Xbyak_aarch64::ZReg, const Xbyak_aarch64::AdrNoOfs&, bool, bool, Xbyak_aarch64::PReg) [with dnnl::impl::cpu::aarch64::cpu_isa_t isa = dnnl::impl::cpu::aarch64::sve_256; dnnl::impl::data_type_t = dnnl_data_type_t]: Assertion `!"unsupported data type\n"' failed.
> zsh: abort (core dumped)  ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8
> ```

please refer 
https://github.com/uxlfoundation/oneDNN/pull/4010#issuecomment-3338341139
Note that `brgemm_desc_set_postops` works for post-ops handled within the brgemm kernel, but this issue occurs when the post-op is handled in the post-op kernel, which is why the check is added in pd_t::init.

src/cpu/aarch64/jit_brgemm_conv.cpp
```diff
@@ -328,6 +328,16 @@ status_t brgemm_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
...
 
+    //check if postop sum is supported
+    if (attr()->post_ops_.find(primitive_kind::sum) != -1) {
+        const auto &sum_po
+                = attr()->post_ops_
+                          .entry_[attr()->post_ops_.find(primitive_kind::sum)];
+        if (!one_of(sum_po.sum.dt, data_type::f32, data_type::s32,
+                    data_type::u8, data_type::s8))
+            return status::unimplemented;
+    }
```
Sqvid: Thank you for the fix!

There is already some similar logic for the `sum_idx` further down so it would be great to remove any duplication between these checks.

https://github.com/uxlfoundation/oneDNN/blob/143b694b9531b4f9a1dd0188d94ed8fb1f671f81/src/cpu/aarch64/jit_brgemm_conv.cpp#L482-L484

src/cpu/aarch64/jit_brgemm_conv.cpp
```diff
@@ -328,6 +328,16 @@ status_t brgemm_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
             && impl::is_dense_format_kind({src_md(), weights_md(), dst_md()});
     if (!ok) return status::unimplemented;
 
+    //check if postop sum is supported
+    if (attr()->post_ops_.find(primitive_kind::sum) != -1) {
+        const auto &sum_po
+                = attr()->post_ops_
+                          .entry_[attr()->post_ops_.find(primitive_kind::sum)];
+        if (!one_of(sum_po.sum.dt, data_type::f32, data_type::s32,
```
Sqvid: Why is the data-type check only for the `sum` post-op? Isn't the supported list of data-types in `src/cpu/aarch64/jit_brgemm_post_ops.hpp` applicable to all `alg_kinds`? Apologies if I have missed something.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -39,6 +39,27 @@
     const char *name() const override { return STRINGIFY(jit_name); } \
     const char *source_file() const override { return __FILE__; }
 
+#define LD_MUL_VL(mn, op, mask, addr, off, size) \
```
Sqvid, line 42: I know these macros were only moved by your patch but it would be nice if we could use this opportunity to use the `do { ... } while(0)` idiom to make it less bugprone. We plan to enforce clang-tidy checks for this soon.

Here's an example: https://github.com/uxlfoundation/oneDNN/pull/4006

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -39,6 +39,27 @@
...
+    { \
+        if (use_mul_vl(off, size, cpu_sveLen)) \
+            mn(op, mask / T_z, \
```
Sqvid: `mask` -> `(mask)` to prevent buggy macro usage.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -39,6 +39,27 @@
...
+                    ptr(addr, compute_off_mul_vl(off, size, cpu_sveLen), \
+                            MUL_VL)); \
+        else \
+            mn(op, mask / T_z, \
```
Sqvid: `(mask)`

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -39,6 +39,27 @@
...
+                    ptr(addr_off(addr, off, X_DEFAULT_ADDR, X_TMP_0))); \
+    }
+
+#define ST_MUL_VL(mn, op, mask, addr, off, size) \
+    { \
```
Sqvid: `do { ... } while (0)`

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
dzarukin: > not only leads to compilation errors due to mismatched definitions...

I might be wrong by C++ templating usually handles that pretty well if one instruction corresponds to a single usage criterion. I'm not sure I see the point of being it more complex without a code example of now versus would be.

src/cpu/aarch64/jit_brgemm_conv.cpp
kasturedeeksha: Thanks for the suggestion! I’ve updated the code to remove the duplication and align it with the existing sum_idx logic.

src/cpu/aarch64/jit_brgemm_conv.cpp
kasturedeeksha: In `jit_brgemm_post_ops.hpp`, the cvt2ps conversion is used for both sum and bias, and earlier assertion checks were failing specifically due to cvt2ps. I did not find any explicit supported data-type list or validation mechanism for the other post-ops in the file, which is why this check was added specifically for sum.
Also I updated the same for 1x1 conv as well.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
kasturedeeksha: Generalizing instructions like ld1w, ld1b, ld1h, etc., into a single template or function parameter is difficult because each instruction has multiple overloads based on register types, predicate forms, and addressing modes.

Even for a single instruction ld1w have around **14** different valid configurations.

- ld1w can use different register types (ZRegS, ZRegD, etc.),
- different addressing modes (AdrScImm, AdrNoOfs, etc).

Trying to pass the instruction itself as an argument would lead to ambiguity, because the compiler cannot deduce which overload to pick.

A fully generalized solution would require encoding all possible instruction variants and their combinations of registers, predicates, and addressing modes in one template, which quickly becomes very complex and hard to maintain.

src/cpu/aarch64/jit_brgemm_conv.cpp
```diff
@@ -483,6 +483,14 @@ status_t brgemm_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
     const int sum_idx = p.find(primitive_kind::sum);
     with_sum = (sum_idx != -1);
 
+    // Check if postop sum datatype is supported
+    if (with_sum) {
+        const auto &sum_po = p.entry_[sum_idx];
+        if (!one_of(sum_po.sum.dt, data_type::undef, data_type::f32,
+                    data_type::s32, data_type::u8, data_type::s8))
+            return status::unimplemented;
```
Sqvid, line 491: It would be nice to get som VERBOSE output here but I won't block the PR.

Sqvid (approved):
LGTM! I'll let @jondea have a look and merge it if he's happy.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
dzarukin: I might be missing something, but I always thought that xbyak handles all that providing all the possible flavors the hardware suggests and the developer only calls what's available and wraps over custom functions to simplify the usage.

Since I'm not aarch64 code maintainer you don't have to follow my suggestions but from whole oneDNN source code base supervision point of view I can confidently claim that macros in a long term are more of a problems than solvers. Different linters, analyzers and debuggers have issues to process them correctly. The bigger the macro, the more terrible experience it provides.

If you could deal with that, I assure you it will bring more benefit for the maintenance team. And with that I'm resolving the discussion. Thanks.

PR #3980, opened 2025-09-22
tests: benchdnn: graph: skip unsupported acc mode on cpu

xiang1guo:
We met several unimplemented test cases in both SRF and SPR. It turned out that primitive only supports `strict` and `relaxed` accumulation mode on CPU, so we skipped it in the benchdnn graph.

The reason the unimplemented status didn't occur on older CPU platforms is that the testing case with the fp16 data type was skipped on those platforms.

### DIFF
--- a/tests/benchdnn/graph/graph.cpp
+++ b/tests/benchdnn/graph/graph.cpp
@@ -463,6 +463,31 @@ int skip_unimplemented_ops(const dnnl::graph::partition &partition,
     return OK;
 }
 
+// Currently, all primitives support "strict" and "relaxed" accumulation modes
+// on CPU.
+int skip_unimplemented_accumulation_mode(
+        const dnnl::graph::partition &partition, const deserialized_graph_t &dg,
+        res_t *res) {
+    const auto &eng = get_graph_engine();
+    bool is_cpu = eng.get_kind() == dnnl::engine::kind::cpu;
+    if (!is_cpu) return OK;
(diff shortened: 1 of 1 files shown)


xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/graph.cpp
```diff
@@ -463,6 +463,31 @@ int skip_unimplemented_ops(const dnnl::graph::partition &partition,
...
+// on CPU.
+int skip_unimplemented_accumulation_mode(
+        const dnnl::graph::partition &partition, const deserialized_graph_t &dg,
+        res_t *res) {
+    const auto &eng = get_graph_engine();
+    bool is_cpu = eng.get_kind() == dnnl::engine::kind::cpu;
+    if (!is_cpu) return OK;
+    const std::vector<size_t> &partition_op_ids = partition.get_ops();
+    for (const size_t op_id : partition_op_ids) {
+        std::string acc_mode {};
```
dzarukin:
```suggestion
        std::string acc_mode;
```

dzarukin (approved):

tests/benchdnn/graph/graph.cpp
xiang1guo: Thanks for the review~ updated.

xiang1guo:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3984, opened 2025-09-22
[GPU] sdpa: remove natve_recip for host scalar scale

skazakov1:
Minor optimization. PR removes scale div operation from kernel if scale is host scalar. Couple % of perf gain for really short cases.

### DIFF
--- a/src/gpu/intel/sdpa/micro.cl
+++ b/src/gpu/intel/sdpa/micro.cl
@@ -372,7 +372,7 @@ __attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE))) kernel void
 micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
         const global VAL_DATA_T *V, global DST_DATA_T *A,
 #if WITH_HOST_SCALE
-        SCALE_DATA_T scale_val,
+        float scalar_scale, float inv_scalar_scale,
 #else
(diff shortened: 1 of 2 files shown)


skazakov1:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph

src/gpu/intel/sdpa/micro.cl
```diff
@@ -525,11 +525,11 @@ micro_sdpa(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
 #if WITH_ATTN_SCALE
 #if WITH_HOST_SCALE
 #if INVERT_SCALE
-        iscale = into_float(scale_val);
-        scale = native_recip(iscale);
+        iscale = scalar_scale;
```
Simonsays095, line 528: If we're passing both the scalar itself and its inverse, why do we define `INVERT_SCALE`? We should just be passing the correct scale, inverting on the host side if required before passing into the kernel.

I understand this cannot be mirrored on the device-side scales implementation, so I'm ok if this isn't done - although I do think it would simplify the kernel + workflow a bit. :)

skazakov1: > this cannot be mirrored on the device-side scales implementation

yes, mostly because of this point. Thanks

PR #3985, opened 2025-09-22
common: remove exec_ctx member from grantor

dzarukin:
This PR removes grantor dependency on exec_ctx which appeared with adding CPU SYCL support. There are a bunch of comments explaining dependencies and purposes.

`exec_ctx_t` must be removed from the grantor to unblock PImpl implementation that will be introduced for `exec_ctx_t` to support Async runtime, otherwise, there'll be double capture of the same context leading to a memory leak.

### DIFF
--- a/src/common/memory_tracking.hpp
+++ b/src/common/memory_tracking.hpp
@@ -426,7 +426,7 @@ struct registry_t {
 
     registrar_t registrar();
     grantor_t grantor(const memory_storage_t *mem_storage,
-            const exec_ctx_t &exec_ctx) const;
+            const void *base_mem_storage_host_ptr) const;
 
     template <typename return_type>
     class common_iterator_t {
@@ -512,16 +512,16 @@ protected:
 struct grantor_t {
     grantor_t(const registry_t &registry,
             const memory_storage_t *base_mem_storage,
-            const exec_ctx_t &exec_ctx)
+            const void *base_mem_storage_host_ptr)
         : registry_(registry)
         , prefix_(0)
         , base_mem_storage_(base_mem_storage)
-        , exec_ctx_(&exec_ctx) {}
+        , base_mem_storage_host_ptr_(base_mem_storage_host_ptr) {}
     grantor_t(const grantor_t &parent, const key_t &prefix)
         : registry_(parent.registry_)
         , prefix_(make_prefix(parent.prefix_, prefix))
         , base_mem_storage_(parent.base_mem_storage_)
-        , exec_ctx_(parent.exec_ctx_) {}
+        , base_mem_storage_host_ptr_(parent.base_mem_storage_host_ptr_) {}
 
     template <typename T = void>
     T *get(const key_t &key, size_t *size = nullptr) const {
@@ -534,7 +534,7 @@ struct grantor_t {
         if (size) *size = e.size;
         if (e.size == 0) return nullptr;
 
-        char *host_storage_ptr = get_host_storage_ptr(base_mem_storage_);
+        char *host_storage_ptr = host_ptr(base_mem_storage_);
         char *base_ptr = host_storage_ptr + base_mem_storage_->base_offset();
         return (T *)e.compute_ptr(base_ptr);
     }
@@ -554,7 +554,7 @@ struct grantor_t {
             // passed to nested primitives. It's required to keep host mapping
             // working. It's working because handles in memory storages are keys
             // in mapping.
-            char *host_storage_ptr = get_host_storage_ptr(base_mem_storage_);
+            char *host_storage_ptr = host_ptr(base_mem_storage_);
             char *base_ptr
                     = host_storage_ptr + base_mem_storage_->base_offset();
             char *aligned_ptr = (char *)e.compute_ptr(base_ptr);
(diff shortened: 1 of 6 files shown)


dzarukin:
make test

src/common/primitive_exec_types.hpp
```diff
@@ -98,7 +98,13 @@ struct exec_ctx_t {
...
     void *host_ptr(int arg, bool do_zeropad = false, status_t *status = nullptr,
             int index = 0) const;
-    void *host_ptr(const memory_storage_t *mem_storage) const;
+    // Returns a mapped pointer for a provided `mem_storage`. In case
+    // `memory_mapping_` wasn't registered, returns a `mem_storage` handle if
+    // `use_mem_storage_handle` is set to true. If `use_mem_storage_handle` is
+    // set to `false`, a `nullptr` will be returned. It's needed when a
+    // distinction of mapped pointer versus abstract 'host_ptr' is required.
+    void *host_ptr(const memory_storage_t *mem_storage,
+            bool use_mem_storage_handle = true) const;
```
densamoilov: `use_mem_storage_handle` may not be the best name here because it's not clear what flow it assumes. Given that our intention is to control the fallback behavior I'd recommend something like `allow_handle_fallback`.

Another option is to flip the logic and call it `require_host_ptr`, default it to false (which covers the usual case), and just set it to true in the execute function where the special case comes up.

src/common/primitive_iface.cpp
```diff
@@ -329,8 +329,17 @@ status_t dnnl_primitive::execute(exec_ctx_t &ctx) const {
...
-    auto scratchpad_grantor
-            = primitive_->pd()->scratchpad_registry().grantor(mem_storage, ctx);
+    // Obtain a scratchpad memory storage host ptr from the context.
+    // `use_mem_storage_handle = false` guarantees a nullptr will be returned
+    // if the usage model doesn't assume mapping.
+    // It's the only pointer that a grantor object requires to provide to
+    // sub-storages for nested primitives.
+    const void *mapped_mem_storage_ptr = mem_storage
+            ? ctx.host_ptr(mem_storage, /* use_mem_storage_handle = */ false)
+            : nullptr;
```
densamoilov: (nit) `host_ptr` has a check for `nullptr`. 
```suggestion
    const void *mapped_mem_storage_ptr = ctx.host_ptr(mem_storage,
            /* use_mem_storage_handle = */ false);
```

src/common/primitive_exec_types.cpp
```diff
@@ -151,7 +152,7 @@ void *exec_ctx_t::host_ptr(const memory_storage_t *mem_storage) const {
                 + mem_storage->base_offset();
     } else {
         assert(mem_storage->is_host_accessible());
```
densamoilov: I think we need to delete this assert because `exec_ctx_t::host_ptr` is called for any `mem_storage`. 

The `grantor_t::host_ptr` function is called only if `is_cpu_engine(base_mem_storage_)` is true or from `T *get(const key_t &key, size_t *size = nullptr) const`, which is a CPU only function (if I'm not mistaken).

src/common/memory_tracking.hpp
```diff
@@ -511,16 +511,16 @@ struct registrar_t {
 struct grantor_t {
     grantor_t(const registry_t &registry,
             const memory_storage_t *base_mem_storage,
-            const exec_ctx_t &exec_ctx)
+            const void *base_mem_storage_host_ptr)
```
densamoilov, line 515: If i get it right then if `base_mem_storage_host_ptr` is not `nullptr` then `base_mem_storage` is not needed?
So basically we always need only one of them. Could we have two `grantor_t` constructors 1) for mem storage 2) for mapped mem storage?

src/common/primitive_exec_types.hpp
dzarukin: I switched to the inverse logic version, thanks for a suggestion.

src/common/primitive_iface.cpp
dzarukin: Good point, removed the external check.

src/common/primitive_exec_types.cpp
dzarukin: I guess you meant the one in grantor. That one I removed. I left the one in exec_ctx as that logic remains unchanged.

src/common/memory_tracking.hpp
dzarukin: `base_mem_storage_host_ptr` is used in pair with mem_storage for base_offset, so I guess we can't. Even if we could, that would require an empty pointer and based on the host_ptr for scratchpad value do a branching for grantor construction which looks uglier.

src/common/primitive_exec_types.cpp
densamoilov: No, I meant this one because it's called from `ctx.host_ptr(mem_storage, /* require_host_ptr = */ true);` where `mem_storage` is not guaranteed to be host accessible for the cases other than SYCL CPU.

src/common/memory_tracking.hpp
densamoilov: Ok, I see. So if  `base_mem_storage` is `nullptr` then `base_mem_storage_host_ptr` is guaranteed to bu `nullptr` also?

And if `base_mem_storage` is not `nullptr` then `base_mem_storage_host_ptr` is not `nullptr` only for SYCL CPU?

src/common/primitive_exec_types.cpp
dzarukin: OK, gotcha, indeed it shots on GPU, dunno how CI missed that. Updated, removed the one from context, restored the one in grantor.

src/common/memory_tracking.hpp
dzarukin: True for both. I added an assert to catch a nullptr - nullptr expectation.

densamoilov (approved):

mgouicem (approved):

PR #3987, opened 2025-09-23
benchdnn: graph: pass acc mode to ref path and more test cases

TaoLv:
1. Pass the accumulation mode attribute to matmul driver in reference path.
2. Add test cases where both QK and VS are in f16 accumulation mode. Previously only QK in f16 accumulation.


xiang1guo (approved):

dzarukin (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #3991, opened 2025-09-23
[GPU] common: verbose: extend dispatch diagnostics to init_conf failures

avmanerikar:
# Description

The PR adds verbose dispatch diagnostics to `init_conf()` implementations instead of printing a generic message for failures during GPU primitive configuration initializations. 

Addresses MFDNN-14234.

### DIFF
--- a/src/common/resampling_pd.hpp
+++ b/src/common/resampling_pd.hpp
@@ -34,6 +34,10 @@
     VCHECK(primitive, create, dispatch, resampling, (f), "%s," msg, \
             this->info(engine), ##__VA_ARGS__)
 
+#define VDISPATCH_RESAMPLING_IC(cond, msg, ...) \
+    VCONDCHECK(primitive, create, dispatch, resampling, (cond), \
+            status::unimplemented, msg, ##__VA_ARGS__)
+
 namespace dnnl {
 namespace impl {
 
--- a/src/gpu/intel/bnorm/reusable.cpp
+++ b/src/gpu/intel/bnorm/reusable.cpp
@@ -110,14 +110,23 @@ static status_t init_calculate_stats_conf(reusable_params_t &conf,
     }
     compute::reusable_dispatch_config_t calc_stat_dispatch_config(
             intel_engine, dim_ids);
-    CHECK(calc_stat_dispatch_config.register_buffer(src_buf));
-    CHECK(calc_stat_dispatch_config.register_buffer(dst_buf));
-    CHECK(calc_stat_dispatch_config.define_dim_index(
-            "IC_DIM", dims::ic, rt_conf.ic));
+    VDISPATCH_BNORM_IC(calc_stat_dispatch_config.register_buffer(src_buf)
+                    == status::success,
+            "failed to register src buffer");
+    VDISPATCH_BNORM_IC(calc_stat_dispatch_config.register_buffer(dst_buf)
+                    == status::success,
+            "failed to register dst buffer");
+    VDISPATCH_BNORM_IC(calc_stat_dispatch_config.define_dim_index(
+                               "IC_DIM", dims::ic, rt_conf.ic)
+                    == status::success,
+            "cannot define dim index for dispatch config");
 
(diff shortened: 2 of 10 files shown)


avmanerikar:
make test

src/gpu/intel/concat/simple.hpp
```diff
@@ -93,9 +93,8 @@ struct simple_t : public primitive_t {
                     attr()->has_default_values(), VERBOSE_UNSUPPORTED_ATTR);
             VDISPATCH_CONCAT_SC(set_default_params(), VERBOSE_UNSUPPORTED_TAG);
 
-            VDISPATCH_CONCAT_SC(init_conf(engine),
-                    VERBOSE_PRIMITIVE_CREATION_FAIL,
-                    "reusable_concat init_conf");
+            VDISPATCH_CONCAT_SC(init_conf(engine), VERBOSE_INIT_CONF_FAIL,
+                    "reusable_concat");
```
Simonsays095: This seems to be inconsistent with the other uses of `VERBOSE_INIT_CONF_FAIL` which use the primitive name, i.e. `concat`. Is there a reason why we're using the primitive name instead of the implementation name for these messages? I think it would make more sense for `VDISPATCH_CONCAT_SC(init_conf(engine), VERBOSE_INIT_CONF_FAIL);` to give the error message "failed to initialize configuration for <impl_name> implementation"

avmanerikar: Makes sense. I have updated the message to print the implementation name instead. Thanks!

Simonsays095: Sorry to adjust my request, but I realize that the dispatch macro already prints out the implementation name. Printing it out again here is likely unneeded - is `failed to initialize configuration` verbose enough? I think it's better than `init_conf()` but I'm not sure it's what you're trying to go for.

avmanerikar: On second thought, adding any message for `init_conf()` calls seems unnecessary. The dispatch checks added within the `init_conf` functions cover all exits so an additional message about configuration failures would be overkill. This is thus better replaced with a simple `CHECK` instead.

Simonsays095: That works for me :) As long as each exit point in `init_conf` is covered by a `VDISPATCH` message, I completely agree.

src/gpu/intel/resampling/vectorized.hpp
```diff
@@ -42,8 +42,7 @@ struct vectorized_bwd_t : public primitive_t {
                     set_default_params(), VERBOSE_UNSUPPORTED_TAG);
             VDISPATCH_RESAMPLING(
                     attr()->has_default_values(), VERBOSE_UNSUPPORTED_ATTR);
-
-            VDISPATCH_RESAMPLING_SC(init_conf(engine), "init_conf()");
+            CHECK(init_conf(engine));
```
Simonsays095, line 45: A few implementations are missing `VDISPATCH` macros within `init_conf`:
- resampling/vectorized.hpp
- reorder/generic.cpp
- reorder/custom.cpp
- reduction/reusable_ref.cpp (there are a few `CHECK` macros)
- lnorm/reusable_vectorized.cpp (some `VDEBUGINFO` and `CHECK` should be swapped out for `VDISPATCH`)
- ip/conv.cpp
- bnorm/reusable.cpp

avmanerikar: Done. Thanks!

avmanerikar:
make test

Simonsays095 (approved):

avmanerikar:
make test
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_bnorm
enable benchdnn_binary
enable benchdnn_concat
enable benchdnn_conv
enable benchdnn_ip
enable benchdnn_lrn
enable benchdnn_matmul
enable benchdnn_reduction

PR #3993, opened 2025-09-23
common: exec_ctx_t: move the implementation to PImpl approach

dzarukin:
Closes [MFDNN-14203](https://jira.devtools.intel.com/browse/MFDNN-14203)

This is a preparational work for asynchronous runtime. `exec_ctx_t` object is [created](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/primitive_iface.cpp#L204) on stack inside [`dnnl_primitive_execute`](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/primitive_iface.cpp#L190) and will be destroyed upon submitter thread exits the function, thus, its lifetime must be extended by the library internals and properly managed to avoid memory leaks and wrong accesses.

The idea is to use a [PImpl](https://en.cppreference.com/w/cpp/language/pimpl.html) technique and put all the members of `exec_ctx_t` under shared pointer. Any time a call to its methods occurs, a dereference of shared pointer is registered which increases a `use_count` of the shared pointer which tracks the necessity of `exec_ctx_t` destruction. Since the main problem are tasks (lambda functions) issued by the `parallel` call, they will capture a context by copy when its needed, thus, copying it and increasing the implementation count, again, letting it longer until a task gets destroyed.

Same story happens with a [`grantor_t`](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/memory_tracking.hpp#L511) that provides a pieces of scratchpad memories to the implementation - it is [allocated](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/primitive_iface.cpp#L332-L333) on the stack and its pointer was saved inside the context. If its lifetime is not managed, it will be nullified (several lines lower that construction) and destroyed becoming unusable at the moment of task (asynchronous) execution. To overcome this problem, `registrar_t`, which [provides](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/memory_tracking.hpp#L427-L428) `grantor_t`, now allocates it on the heap and then sets it to the context which acquires ownership of it by wrapping it into a `unique_ptr` forcing the identical lifetime of the grantor and the context which is also true for synchronous runtimes.

Design and assistance credits go to @densamoilov.

PR #3985 must go first as it removes a back dependency of grantor on `exec_ctx_t` which would, otherwise, lead to a self-capture and further to a memory leak as the object never get destroyed since a shared pointer holds a unique pointer which hold a shared one.


dzarukin:
make test

dzarukin:
make test

src/common/memory_tracking.cpp
```diff
@@ -41,13 +41,29 @@ const void *registry_t::entry_t::compute_ptr(const void *base_ptr) const {
     return (const void *)aligned_ptr;
 }
 
-grantor_t registry_t::grantor(const memory_storage_t *mem_storage,
+// This call allocates a grantor on the heap, and then the ownership is
+// transferred to an `exec_ctx_t` object.
```
densamoilov: (nit) For those who are out of context, the phrase `and then the ownership is transferred to an exec_ctx_t object` won’t make much sense, because that _may_ happen on the callee side and is not even guaranteed (technically, we could use this function in a different way). I think documenting that this function returns a pointer to a grantor allocated on the heap would make it clearer.

src/common/primitive_exec_types.hpp
```diff
@@ -73,29 +71,48 @@ using exec_args_t = std::unordered_map<int, memory_arg_t>;
...
-    stream_t *stream() const { return stream_; }
-    const exec_args_t &args() const { return args_; }
+    // A main version when only a stream and args are available.
+    exec_ctx_t(stream_t *stream, exec_args_t &&args = {});
+
+    // A version used by `nested_scratchpad_t` objects. Underlying
+    // implementation copies a master context members except args and grantor.
+    exec_ctx_t(const exec_ctx_t &other, exec_args_t &&args);
+
+    // See `exec_ctx_impl_t` setters comment.
```
densamoilov: (nit) I think it would make more sense to document the interfaces of `exec_ctx_t` because those are what developers actually use and are interested in (far more often than their implementation counterparts).
Documenting interfaces of the implementation could make sense if they did something unusual but in our case it's just a redirection.

src/common/primitive_exec_types.hpp
```diff
@@ -73,29 +71,48 @@ using exec_args_t = std::unordered_map<int, memory_arg_t>;
...
-        , resource_mapper_(other.resource_mapper_) {}
+    // Doesn't work without a stream and args.
+    exec_ctx_t() = delete;
 
-    stream_t *stream() const { return stream_; }
-    const exec_args_t &args() const { return args_; }
+    // A main version when only a stream and args are available.
+    exec_ctx_t(stream_t *stream, exec_args_t &&args = {});
+
+    // A version used by `nested_scratchpad_t` objects. Underlying
```
densamoilov: I think this could be misleading, because the `nested_scratchpad_t` object doesn't appear to use this constructor. If I understand correctly, `nested_scratchpad_t` returns a grantor, which is then passed to the `exec_ctx_t` object created via this constructor.

In general, I think we should refrain from tying interfaces that have clear, general semantics to specific use cases, because nothing prevents us from using those interfaces in other scenarios. If we want to clarify why or where this interface is useful, we can add a separate note in the description.

src/common/primitive_exec_types.hpp
```diff
@@ -73,29 +71,48 @@ using exec_args_t = std::unordered_map<int, memory_arg_t>;
...
+    // implementation copies a master context members except args and grantor.
```
densamoilov: (nit) I think it would make more sense to drop the mention of "implementation" here, 
because whether it's a classic body function or one defined in the pimpl object 
makes no difference from the public interface perspective.

(nit) One more thing, mentioning "master context" is very contextual, while the constructor 
itself is not. We could probably just say that the constructor copies the contents 
of the `other` object, except for `args` and `grantor`.

src/common/primitive_exec_types.hpp
```diff
@@ -104,6 +121,75 @@ struct exec_ctx_t {
...
+
+    exec_ctx_impl_t(stream_t *stream, exec_args_t &&args);
+
+    exec_ctx_impl_t(stream_t *stream, exec_args_t &&args,
+            const std::unordered_map<void *, void *> &memory_mapping,
+            const resource_mapper_t *resource_mapper);
+
+    ~exec_ctx_impl_t();
+
+    // Copying `args` is restricted, must be moved instead.
```
densamoilov: I think, in general, we want to enforce working with `exec_ctx_impl_t` via a shared 
pointer, and that's why it makes sense semantically to delete copy capabilities 
for `exec_ctx_impl_t` (to avoid unintended copies).

src/common/primitive_exec_types.hpp
```diff
@@ -104,6 +121,75 @@ struct exec_ctx_t {
...
+    void set_memory_mapping(void *handle, void *host_ptr);
+
+    // ... the resource mapper is declared private and can be used only inside
+    // primitive_t methods, such as `primitive_t::execute`.
+    void set_resource_mapper(const resource_mapper_t *resource_mapper) {
+        resource_mapper_ = resource_mapper;
+    }
+    // ... a grantor has a dependency on a host pointer which is available only
+    // after mapping is fully set at `primitive_t::execute` call.
+    void set_scratchpad_grantor(
```
densamoilov, line 199: I would also suggest clarifying the behavior related to ownership. For example, `set_scratchpad_grantor` takes ownership of the passed object, while `set_resource_mapper` does not.

I see that you added a comment clarifying this, but it resides where the data members are declared, which is the very core of the class implementation. This forces developers to look into the implementation details when all they really need is clarity about the ownership semantics of the interface. This also further indicates that such documentation would be better placed at the `exec_ctx_t` level.

src/common/primitive_exec_types.cpp
```diff
@@ -92,18 +92,49 @@ status_t cvt_primitive_args(const primitive_desc_t *pd, int nargs,
     return success;
 }
 
-memory_t *exec_ctx_t::input(int arg) const {
-    if (args_.count(arg) != 1) return nullptr;
-    const auto ma = args_.at(arg);
-    assert(ma.is_const);
-    return ma.mem;
+exec_ctx_t::exec_ctx_t(stream_t *stream, exec_args_t &&args)
+    : impl_(std::make_shared<exec_ctx_impl_t>(stream, std::move(args))) {}
```
densamoilov: To ensure stronger encapsulation, we could make the constructors of `exec_ctx_impl_t` private and declare the `exec_ctx_t` class as a friend of `exec_ctx_impl_t`, so that we can tightly control how instances of `exec_ctx_impl_t` are created.

src/common/primitive.cpp
densamoilov: My biggest concern about the changes is `nested_scratchpad_t` because it doesn’t seem to make sense anymore, yet we keep it while the semantics are unclear (and probably ill-defined).

- `nested_scratchpad_t` has a state, which is `grantor_`, but it doesn't own it. At the time of creation of a `nested_scratchpad_t` object, `grantor_` has no owner. The problem is that we query `grantor_` using `nested_scratchpad_t::grantor()` and take ownership of it, but we don’t change the state of `nested_scratchpad_t`. Therefore, we end up with two objects: one owns `grantor_` and one doesn’t, but both have access to it. This creates ambiguous ownership relations that can cause issues if we don’t follow the exact pattern that was kept in mind when this was designed.

- Practical issues:  
  - When we call the nested primitive and exit immediately due to the non-blocking call, we essentially destroy the `nested_scratchpad_t` object that accesses `grantor_` in its destructor. The problem is that it’s not deterministic whether the submitted nested primitive has completed yet, and its `exec_ctx_t` object hasn't been destroyed yet. Therefore, `grantor_` may no longer be valid at the time of `nested_scratchpad_t` destruction, which is undefined behavior and can cause sporadic crashes.  
  - Before these changes, `nested_scratchpad_t` served the purpose of owning `grantor_` until the related nested primitive was complete. Now, it seems that having the `nested_scratchpad_t` class doesn’t make sense anymore because we could easily replace it with a function that creates `grantor_`. The only reason it’s still there is to implement `scratchpad_debug` capabilities, but this was based on the contract that `nested_scratchpad_t` owns `grantor_`. That contract has now been broken.

I think we need to come up with a way to support scratchpad debug capabilities. There are a few options: one could be extending the `grantor_t` class so it handles protection and unprotection. Another option is to add a callback to `exec_ctx_t` to unprotect the grantor.

If this is a time-sensitive issue, we could disable the `DNNL_ENABLE_MEM_DEBUG` capabilities for the thunk runtime to unblock progress, but we need to address it properly anyway.

src/common/memory_tracking.cpp
dzarukin: Updated with the following text:
```
// This call returns a pointer to a grantor allocated on the heap. It's user's
// responsibility to free it.
```

src/common/primitive_exec_types.hpp
dzarukin: I tend to agree with your comment and dropped this one without any further notes.

src/common/primitive_exec_types.hpp
dzarukin: Sounds reasonable to me, moved these comments to `exec_ctx_t` level.

src/common/primitive_exec_types.hpp
dzarukin: They are deleted right within next two lines, aren't they?

src/common/primitive_exec_types.hpp
dzarukin: Moved the comment to the interface and left a note in impl where to look for details. Thanks.

src/common/primitive_exec_types.cpp
dzarukin: Applied the suggestion for private ctors, thanks.

src/common/primitive.cpp
dzarukin: Will address this point into a separate commit not to mess up with rest changes.
Destruction is not a problem as grantor is allocated by `nested_scratchpad_t` but not explicitly freed at destructor, which is not clean, for sure, but, at least, not leaking or crashing.

dzarukin: Replaced `nested_scratchpad_t` with standalone `crete_nested_grantor` function. Slight change in a signature to avoid unnecessary header dependencies.

Wrapped a storage into a shared_ptr (thanks for the help. BTW!) to handle copy of the grantor (though I moved all existing ones to const & by clang-tidy recommendations).

dzarukin:
make test

src/common/primitive_exec_types.hpp
densamoilov: I was referring to the comment that I thought served as a justification for the ctors being deleted. It sounds confusing because the actual reason is different.

src/common/memory_tracking.cpp
```diff
@@ -72,6 +110,19 @@ bool grantor_t::is_cpu_engine(const memory_storage_t *mem_storage) const {
...
+        const registry_t &nested_registry) {
+    // Note: `release()` from `std::unique_ptr` returns a pointer to allocated
+    // sub-storage. `get()` would lead to `std::unique_ptr`'s underlying
+    // storage destruction leading to the loss of information about a storage.
+    // Once the info is lost, the storage can't jump through a vtable into
+    // proper implementations of its base class virtual methods and crashes.
+    return new grantor_t(nested_registry,
+            master_grantor.get_memory_storage(key).release(),
+            master_grantor.get_base_mem_storage_host_ptr(),
+            /* delete_storage = */ true);
```
densamoilov: (nit) To provide some context which implies "delete_storage" but gives us more information (about ownership).

```suggestion
            /* take_storage_ownership = */ true);
```

src/common/memory_tracking.cpp
```diff
@@ -72,6 +110,19 @@ bool grantor_t::is_cpu_engine(const memory_storage_t *mem_storage) const {
     return false;
 }
 
+grantor_t *create_nested_grantor(const grantor_t &master_grantor, int key,
+        const registry_t &nested_registry) {
+    // Note: `release()` from `std::unique_ptr` returns a pointer to allocated
+    // sub-storage. `get()` would lead to `std::unique_ptr`'s underlying
+    // storage destruction leading to the loss of information about a storage.
+    // Once the info is lost, the storage can't jump through a vtable into
+    // proper implementations of its base class virtual methods and crashes.
```
densamoilov: I think this explanation makes things more convoluted than they should be :)

I would rather remove the part that describes one of the possible consequences of a double-release issue.
Also, explaining the difference between `release()` and `get()` seems somewhat redundant.

src/common/memory_tracking.hpp
```diff
@@ -534,59 +539,63 @@ struct grantor_t {
...
-protected:
+private:
     const registry_t &registry_;
     const key_t prefix_;
-    const memory_storage_t *base_mem_storage_ = nullptr;
+    // `std::shared_ptr` is a way for clear copy semantics of nested grantors
+    // since it involves an ownership of the underlying storage.
+    // User-defined copy constructor would unlikely achieve a "clear" state,
+    // while prohibiting copy operations is impossible for asynchronous
+    // threadpool runtime.
```
densamoilov:
```suggestion
    // The `grantor_t` class is designed to be copyable for convenience.
    // When `base_mem_storage_` is owned, we need to ensure proper lifetime
    // management. Therefore, the copy semantics of `grantor_t` are implemented as
    // reference semantics.
```

densamoilov (approved):

src/common/primitive_exec_types.hpp
dzarukin: Removed this one.

src/common/memory_tracking.cpp
dzarukin: Applied the proposed name, thanks.

src/common/memory_tracking.cpp
dzarukin: Replaced with 
```
// Take the ownership of a nested storage through a `release()` call.
```

src/common/memory_tracking.hpp
dzarukin: Thanks for the wording, applied.

PR #3994, opened 2025-09-24
ci: aarch64: change nightly to test with assert

Ryo-not-rio:
Update nightly to build with RelWithAssert to catch any assertion errors e.g. https://github.com/uxlfoundation/oneDNN/actions/runs/17953932420

### DIFF
--- a/.github/workflows/aarch64-acl.yml
+++ b/.github/workflows/aarch64-acl.yml
@@ -41,6 +41,7 @@ jobs:
           { name: MacOS, label: macos-14, threading: SEQ, toolset: clang, build: RelWithAssert },
           { name: cb100, label: ubuntu-24.04-arm, threading: OMP, toolset: gcc, build: RelWithAssert },
           { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: clang, build: RelWithAssert },
+          { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: gcc, build: RelWithAssert },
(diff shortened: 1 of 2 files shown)


Sqvid (approved):
LGTM. thanks!

.github/workflows/aarch64-acl.yml
```diff
@@ -41,6 +41,7 @@ jobs:
           { name: MacOS, label: macos-14, threading: SEQ, toolset: clang, build: RelWithAssert },
           { name: cb100, label: ubuntu-24.04-arm, threading: OMP, toolset: gcc, build: RelWithAssert },
           { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: clang, build: RelWithAssert },
+          { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: gcc, build: RelWithAssert },
           { name: c6g, label: ah-ubuntu_22_04-c6g_2x-50, threading: OMP, toolset: gcc, build: Release }
```
jondea, line 45: Why do we still need `Release` if we have `RelWithAssert`?

Sqvid: I think it's to make sure the performance regression pipeline runs in release mode

Ryo-not-rio: We need the release version of ACL for the precommit CI

Ryo-not-rio: We run both performance tests and some precommit builds in release mode, so we need release mode ACL for those runs

PR #3997, opened 2025-09-24
cpu: aarch64: fix various clang-tidy failures

Anndrey24:
# Description

This commit fixes the failures associated with the following `clang-tidy` checks on the AArch64 path:
 - `bugprone-macro-parentheses`
 - `performance-unnecessary-value-param`
 - `google-default-arguments`


src/cpu/aarch64/jit_uni_dw_conv_kernel_f32.hpp
```diff
@@ -40,7 +41,7 @@ struct jit_uni_dw_conv_fwd_kernel_f32_t : public jit_generator {
     DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_uni_dw_conv_fwd_kernel_f32_t)
 
     jit_uni_dw_conv_fwd_kernel_f32_t(jit_conv_conf_t ajcp)
-        : jcp(ajcp), eltwise_injector_(nullptr) {
+        : jcp(std::move(ajcp)), eltwise_injector_(nullptr) {
```
Sqvid, line 44: Which check is this change from? I assume `performance-unnecessary-value-param`?

Sqvid: There's no need for the `std::move` here. Just change the argument to be of type `const jit_conv_conf_t &` and remove the `move`. Once that's done you wont need the `<utility>` header either.

Anndrey24: I'll do that, thanks!

aditew01 (approved):

PR #4001, opened 2025-09-25
Xzeng/brgemm zp comp pad amx

xuxinzen:
Fixes MFDNN-14036, MFDNN-14037, MFDNN-14134, MFDNN-12878

Removing heuristic on brgconv w/zp that using large buffer can fix the performance drop from v3.7 to v3.9. But the performance issue reported in MFDNN-12878 is back after that.

Reducing buffer for compensation padding can also help to improve the performance for the shapes that using large buffer(MFDNN-12878).

WIP: some shapes that using relo cause perf degradation on AMX comparing with VNNI. Working on adding heuristic to disable relo for such shapes.

WIP: adding performance data


xuxinzen:
make test

xuxinzen:
make test

tprimak (approved):

xuxinzen:
make test
set test_scope=NIGHTLY

PR #4006, opened 2025-09-25
cpu: aarch64: ensure LDR_ASSERT and STR_ASSERT are expanded fully

keanucz:
# Description

In this one-line if statement only the first line of the `LDR_ASSERT` macro would be conditionally handled & not the second.

This fix ensures that upon macro expansion the second line is also only ran if use_global_stats() is false by encapsulating `LDR_ASSERT` in `do { ... } while(0)` (and similarly with `STR_ASSERT` to prevent a similar bug from occurring).

This was flagged after running clang-tidy with the bugprone-multiple-statement-macro check.

### DIFF
--- a/src/cpu/aarch64/jit_uni_batch_normalization.cpp
+++ b/src/cpu/aarch64/jit_uni_batch_normalization.cpp
@@ -37,11 +37,15 @@
 
 #define IDX(a) static_cast<uint32_t>((a).getIdx())
 #define LDR_ASSERT(r, addr, offt) \
-    assert((offt) < 256); \
-    ldr(r, ptr(addr, (int)(offt)));
+    do { \
+        assert((offt) < 256); \
+        ldr(r, ptr(addr, (int)(offt))); \
+    } while (0)
 #define STR_ASSERT(r, addr, offt) \
-    assert((offt) < 256); \
-    str(r, ptr(addr, (int)(offt)));
+    do { \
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -2001,8 +2001,10 @@ struct jit_bnorm_t : public jit_generator {
         }
         if (jbp_->is_nspc_) {
             // comeback
-            if (!pd_->use_global_stats())
+            if (!pd_->use_global_stats()) {
                 LDR_ASSERT(reg_src, sp, (int)stack_off_src);
```
Sqvid, line 2009: I am a fan of braced-`if`s in general, but it would be nice if the macros themselves used the `do { ... } while(0)` idiom to guard against this happening in the future.

Sqvid (changes requested):

src/cpu/aarch64/jit_uni_batch_normalization.cpp
keanucz: Good point, I'll implement a fix to do that instead

Sqvid (approved):

Sqvid:
LGTM, thanks for the patch!

renato-arantes (approved):

PR #4007, opened 2025-09-25
cpu: aarch64: add bf16 support to brdgmm_dw_convolution

Ryo-not-rio:
speeds up on c8g:
```
--mode=P --conv --dir=FWD_I --dt=bf16:bf16:bf16 g32ic32ih202iw202oc32oh202ow202kh9kw1ph4pw0ndw_regression:conv4
```
16 threads: 149x
4 threads: 150x

PyTorch mobilenet:
7.78x


src/cpu/aarch64/brgemm/jit_brdgmm_kernel.hpp
```diff
@@ -116,9 +117,7 @@ struct jit_brdgmm_kernel_base_t : public jit_generator {
     int tail_length() { return n_block1_tail() % simd_w_; }
     bool is_fma_embd() { return brg.is_f32; }
     bool is_fast_vnni_int8() { return is_fast_vnni_int8(brg); }
-    int vnni_substep() {
-        return brg.isa_impl == sve_256 && (brg.is_bf16 || brg.is_f16) ? 2 : 1;
-    }
+    int vnni_substep() { return brg.isa_impl == sve_256 && brg.is_f16 ? 2 : 1; }
```
Sqvid, line 120: is `f16` even allowed here? I only see support for `f32`, `int8`, and `bf16` (now).

Ryo-not-rio: It's a left over from the x86 code but I'm guessing we might wanna keep it for when we do add f16 support

jondea: This might not be the right PR to address this, but anything that explicitly refers to just one sve vector length like this is likely to cause bugs.

src/cpu/cpu_convolution_list.cpp
```diff
@@ -220,6 +220,8 @@ const std::map<pk_dt_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map()
             CPU_INSTANCE_AVX2(brgemm_1x1_convolution_fwd_t<avx2_vnni_2>)
             CPU_INSTANCE_AVX2(brgemm_convolution_fwd_t<avx2_vnni_2>)
             CPU_INSTANCE_AARCH64(jit_uni_dw_convolution_fwd_t<sve_256, bf16, bf16>)
+            CPU_INSTANCE_AARCH64(brdgmm_dw_convolution_fwd_t<sve_256>)
```
jondea, line 223: Renato or Aditya may know better, but I think we might want to put these above jit uni. If your brdgmm is properly using BFDOT, then it will be ~2x faster

Ryo-not-rio: I benchmarked it and depending on the problem, jit_uni was faster, and where brdgemm was faster, it wasn't faster by a lot so I've put it here to not cause regressions

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -455,7 +468,16 @@ void jit_brdgmm_kernel_base_t::store_accumulators_apply_post_ops(
...
-                case data_type::bf16: assert(!"unsupported data type\n"); break;
+                case data_type::bf16:
+                    if (mask_flag) {
+                        bfcvt(vmm.h, k_mask / T_m, vmm.s);
+                        st1h(vmm.s, k_mask / T_m, addr);
+                    } else {
+                        bfcvt(vmm.h, P_ALL_ONE / T_m, vmm.s);
+                        st1h(vmm.s, P_ALL_ONE / T_m, addr);
+                    }
+
```
aditew01:
```suggestion
                    const auto mask = mask_flag ? k_mask/T_m : P_ALL_ONE/T_m;
                    bfcvt(vmm.h, mask, vmm.s);
                    st1h(vmm.s, mask, addr);

```

we can hoist the mask var for other dtype as well.

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -536,7 +558,10 @@ void jit_brdgmm_kernel_base_t::load_a(
         } else
             ld1w(vmma.s, P_ALL_ONE / T_z, addr);
     } else if (brg.is_bf16) {
-        assert(!"unsupported\n");
+        if (mask_flag) {
+            ld1h(vmma.s, k_mask / T_z, addr);
+        } else
+            ld1h(vmma.s, P_ALL_ONE / T_z, addr);
```
aditew01: use a common mask variable instead of the if / else block

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -217,9 +218,21 @@ void jit_brdgmm_kernel_base_t::cvt2ps(data_type_t type_in, const ZReg vmm_in,
...
+                    ld1h(vmm_tmp(0).s, k_mask / T_z, op);
+                    lsl(vmm_tmp(0).s, vmm_tmp(0).s, 16);
+                    mov(vmm_in.s, k_mask / T_m, vmm_tmp(0).s);
+                } else { //Zeroing
+                    ld1h(vmm_in.s, k_mask / T_z, op);
+                    lsl(vmm_in.s, vmm_in.s, 16);
+                }
+            } else {
+                ld1h(vmm_in.s, P_ALL_ONE / T_z, op);
+                lsl(vmm_in.s, vmm_in.s, 16);
```
aditew01, line 226: maybe refactor this one as well ?

src/cpu/cpu_convolution_list.cpp
jondea: I see, how is it relative to f32?

Ryo-not-rio: The problem listed in the PR goes to jit_uni and is about 22.9% faster than f32

PR #4008, opened 2025-09-25
ci: aarch64: add clang-tidy workflow for AArch64

Sqvid:
# Description

Adds a clang-tidy run for files in `src/cpu/aarch64`

### DIFF
--- a/src/cpu/aarch64/brgemm/brgemm.cpp
+++ b/src/cpu/aarch64/brgemm/brgemm.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2020-2025 Intel Corporation
 * Copyright 2023-2025 FUJITSU LIMITED
-* Copyright 2024 Arm Ltd. and affiliates
+* Copyright 2024-2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -461,7 +461,7 @@ status_t brgemm_init_tiles(const brgemm_t &brg, char palette[64]) {
 
 namespace {
 template <typename T>
-static inline int sign(T v) {
+inline int sign(T v) {
     return (v > 0) ? 1 : ((v < 0) ? -1 : 0);
 }
 
--- a/src/cpu/aarch64/jit_generator.hpp
+++ b/src/cpu/aarch64/jit_generator.hpp
@@ -84,7 +84,7 @@ constexpr Xbyak_aarch64::Operand::Code abi_save_gpr_regs[]
                 Xbyak_aarch64::Operand::X27, Xbyak_aarch64::Operand::X28};
 
 // See "Procedure Call Standsard for the ARM 64-bit Architecture (AArch64)"
-static const Xbyak_aarch64::XReg abi_param1(Xbyak_aarch64::Operand::X0),
+const Xbyak_aarch64::XReg abi_param1(Xbyak_aarch64::Operand::X0),
         abi_param2(Xbyak_aarch64::Operand::X1),
         abi_param3(Xbyak_aarch64::Operand::X2),
         abi_param4(Xbyak_aarch64::Operand::X3),
--- a/src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
@@ -65,7 +65,6 @@ struct jit_brgemm_matmul_copy_a_impl_t : public jit_brgemm_matmul_copy_a_t,
 
     jit_brgemm_matmul_copy_a_impl_t(const brgemm_matmul_conf_t *conf)
         : jit_brgemm_matmul_copy_a_t(conf)
-        , jit_generator_t()
         , typesize_(conf_->a_dt_sz)
         , tr_typesize_(conf_->tr_a_dt_sz)
         , vnni_granularity_(data_type_vnni_granularity(conf_->src_dt))
@@ -366,7 +365,6 @@ struct jit_brgemm_matmul_copy_b_int8_t : public jit_brgemm_matmul_copy_b_t,
 
     jit_brgemm_matmul_copy_b_int8_t(const brgemm_matmul_conf_t *conf)
         : jit_brgemm_matmul_copy_b_t(conf)
-        , jit_generator_t()
         , src_stride_(conf->wei_tag == format_tag::acbd
                           ? conf->copy_B_wei_stride
                           : conf->N * sizeof(int8_t))
@@ -655,7 +653,6 @@ struct jit_brgemm_matmul_copy_b_transposed_t
 
     jit_brgemm_matmul_copy_b_transposed_t(const brgemm_matmul_conf_t *conf)
         : jit_brgemm_matmul_copy_b_t(conf)
-        , jit_generator_t()
         , typesize_(conf_->b_dt_sz)
         , tr_typesize_(conf_->tr_b_dt_sz)
         , vnni_granularity_(data_type_vnni_granularity(conf_->wei_dt))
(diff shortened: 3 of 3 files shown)


Sqvid:
I think the pipeline is in a decent state as of right now. However, I will open it for review and merge it once we've fixed all the clang-tidy failures we are aware of.

Sqvid:
I think this is ready for review now.

cc: @vpirogov @Ryo-not-rio

.github/workflows/clang-tidy.yml
```diff
@@ -32,36 +51,85 @@ concurrency:
...
-              echo "Skipped $file as it's not built in x64 OpenMP/OpenCL configuration."
-            fi
-          done
+          # Match .cpp, .hpp, and .h files
+          analysis_list=$(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.[ch](pp|xx)?$');
+          if [[ -z "${analysis_list}" ]]; then
+            echo "No source files to check."
+          else
+            printf "File list:\n$analysis_list\n"
+            run-clang-tidy -format -config-file=./.clang-tidy -p ./build $analysis_list 2>&1 | tee -a clang-tidy.log
```
Sqvid: Switched out the `for` loop for the `run-clang-tidy` script. It should be able to collate checks across files and follow `#includes` more intelligently.

.github/workflows/clang-tidy.yml
```diff
@@ -32,36 +51,85 @@ concurrency:
...
-          do
-            if grep -q "$file" "build/compile_commands.json"; then
-              echo -e "\nAnalyzing $file"
-              clang-tidy -p build $file 2>&1 | tee -a clang-tidy.log
-            else
-              echo "Skipped $file as it's not built in x64 OpenMP/OpenCL configuration."
-            fi
-          done
+          # Match .cpp, .hpp, and .h files
+          analysis_list=$(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.[ch](pp|xx)?$');
```
Sqvid: The regex is for all the C/C++ source/header extensions that I am currently aware of in the library.

.github/workflows/clang-tidy.yml
HamzaAB0: We should add the -j flag for faster processing.

Sqvid: Cool, I didn't know about this. Done.

Sqvid:
I think I have tested this no source-changes, changes in path-specific code with no errors, changes in path-specific code with errors, and changes in common-code with errors.

Here is an example of both runners catching a common failure https://github.com/uxlfoundation/oneDNN/actions/runs/18164461524?pr=4008

Ryo-not-rio (approved):

src/cpu/aarch64/brgemm/brgemm.cpp
```diff
@@ -461,7 +461,7 @@ status_t brgemm_init_tiles(const brgemm_t &brg, char palette[64]) {
 
 namespace {
 template <typename T>
-static inline int sign(T v) {
+inline int sign(T v) {
```
jondea, line 464: What's the rationale on these ones where we remove static?

jondea (approved):
LGTM, I just have 1 question

src/cpu/aarch64/brgemm/brgemm.cpp
dzarukin: It's tucked behind `unnamed namespace`, which makes this function `static` by meaning, no need for extra redundant qualifier.

.github/automation/aarch64/build.sh
```diff
@@ -33,17 +33,32 @@ ONEDNN_TEST_SET=${ONEDNN_TEST_SET:-"SMOKE"}
...
-        -Bbuild -S. \
-        -DDNNL_AARCH64_USE_ACL=ON \
-        -DONEDNN_BUILD_GRAPH=$ONEDNN_BUILD_GRAPH \
-        -DDNNL_CPU_RUNTIME=$ONEDNN_THREADING \
-        -DONEDNN_WERROR=ON \
-        -DDNNL_BUILD_FOR_CI=ON \
-        -DONEDNN_TEST_SET=$ONEDNN_TEST_SET \
-        -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE
-    set +x
+    if [[ "$GITHUB_JOB" == "pr-clang-tidy" ]]; then
```
vpirogov, line 36: (nit) This looks fragile. Can we handle this via script parameters instead?

.github/workflows/clang-tidy.yml
```diff
@@ -32,36 +51,87 @@ concurrency:
...
+      - name: Read ci.json versions file
+        id: get-versions
+        run: |
+          content=`cat ${{ github.workspace }}/oneDNN/.github/automation/aarch64/ci.json`
+          content="${content//[$'\t\r\n$ ']}"
+          echo "output=$content" >> $GITHUB_OUTPUT
+
+      - if: ${{ matrix.config.arch == 'AArch64' }}
+        name: Clone ACL
+        run: ${{ github.workspace }}/oneDNN/.github/automation/aarch64/build_acl.sh
```
vpirogov, line 86: Can we get ACL build from cache instead?

.github/workflows/clang-tidy.yml
```diff
@@ -32,36 +51,87 @@ concurrency:
...
-          for file in $(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.cpp$');
-          do
-            if grep -q "$file" "build/compile_commands.json"; then
-              echo -e "\nAnalyzing $file"
-              clang-tidy -p build $file 2>&1 | tee -a clang-tidy.log
-            else
-              echo "Skipped $file as it's not built in x64 OpenMP/OpenCL configuration."
-            fi
-          done
+          analysis_list="$(git diff --name-only ${{ github.event.pull_request.base.sha }} | grep -E '\.[ch](pp|xx)?$')"
```
vpirogov: clang-tidy is not designed to handle header files directly. These should be checked only in the context of cpp files that include them.

.github/workflows/clang-tidy.yml
```diff
@@ -32,36 +51,87 @@ concurrency:
...
+          if [[ -z "$analysis_list" ]]; then
+            echo "No source files to check."
+            exit 0
+          else
+            printf "File list:\n$analysis_list\n"
+            run-clang-tidy -j $(nproc) -format -config-file=./.clang-tidy -p ./build $analysis_list 2>&1 | tee -a clang-tidy.log
```
vpirogov: `-config-file=./.clang-tidy` is not needed, as clang-format will pick up closest .clang-tidy from the hierarchy. oneDNN has multiple `.clang-tidy` files in the source tree.

.github/workflows/clang-tidy.yml
Sqvid: The current method is faster than cache retrieval. Our cache workflow also requires us to clone ACL (to check which hash we are supposed to retrieve). And the current method is basically a clone + creating 3 files.

Since we are only configuring oneDNN (not compiling) there is no reason to retrieve a compiled ACL from cache (we only need the headers + our 3 fake files).

https://github.com/uxlfoundation/oneDNN/pull/4008/files#diff-2b7b9696b464e236f8066c1c1d33c6c91c95160f813e2b3fe59ddf0e83057ebfR101-R103

.github/automation/aarch64/build.sh
Sqvid: Don't mind changing it, but what do you mean by script parameters in this context? I modelled the current method after x64:

https://github.com/uxlfoundation/oneDNN/blob/491235ed4b75ec31e7d6d7f0045ad2ed52e30bf8/.github/automation/x64/build_linters.sh#L9-L23

.github/workflows/clang-tidy.yml
Sqvid: Hmm I tried making this change locally but it throws errors from trying to read clang-tidy configs outside oneDNN (i.e, the ComputeLibrary repository). I have uploaded the change regardless so we can see if it causes issues in the CI.

.github/workflows/clang-tidy.yml
Sqvid: Thanks! Have made the change.

.github/workflows/clang-tidy.yml
Sqvid: Seems to work fine. I guess something is strange with my local environment. Thanks for the suggestion!

.github/workflows/clang-tidy.yml
vpirogov: Got it, thanks.

PR #3838, opened 2025-08-28
cpu: aarch64: Extend bf16 support for BRGEMM convolution and 1x1 convolution

vishwascm:
# Description

This PR extends support to BF16 datatype for the BRGEMM (Batch-Reduce General Matrix Multiplication) convolution kernels. (BRGEMM convolution and 1x1 convolution)

Major Code Changes:
• Support for bf16 in BRGEMM forward convolution (bf16:bf16:f32 and bf16:bf16:bf16)
• Support for bf16 in BRGEMM 1x1 forward convolution (bf16:bf16:f32 and bf16:bf16:bf16)
• Support for bf16 in BRGEMM Post Ops
Other minor changes added accordingly in some files.

Perf Timings (ms)  Compared with f32 for single core:
#### brgconv:sve_256
| Shapes | f32:f32:f32 | bf16:bf16:f32 | bf16:bf16:bf16 | 
|:----------:|:----------:|:----------:|:----------:|
|g2mb2_ic32oc32_iw27ow26kw2sw1dw0pw0 |  0.00240  |  0.00176 | 0.00169 |
|mb1_ic512oc512_ih7oh7kh3sh1dh0ph1_iw7ow7kw3sw1dw0pw1| 2.510 | 1.232 | 1.225 |

#### brgconv_1x1:sve_256
| Shapes | f32:f32:f32 | bf16:bf16:f32 | bf16:bf16:bf16 | 
|:----------:|:----------:|:----------:|:----------:|
|g1mb50_ic256oc512_ih56oh28kh1sh2dh0ph0_iw56ow28kw1sw2dw0pw0 |  131.478  |  66.870 | 66.4103 |
|g1mb50_ic512oc256_ih28oh14kh1sh2dh0ph0_iw28ow14kw1sw2dw0pw0| 33.984 | 16.763 | 16.733 |

# Checklist

## General

- [ ✓] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [ ✓] Have you formatted the code using clang-format?

### make test
99% tests passed, 1 tests failed out of 226

Total Test time (real) = 2292.72 sec

The following tests FAILED:
        206 - test_benchdnn_modeC_graph_ci_cpu (Failed)
Errors while running CTest
Output from these tests are in: /home/vishwas/oss/viswas/oneDNN/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
[...]

PR #4009, opened 2025-09-25
cpu: aarch64: implement brgdeconv non-strided

michalowski-arm:
This change implements non-strided brgdeconv on aarch64 based on existing x64 implementation. Since it uses brgconv and brgconv_1x1 for deconvolution, once #3838 is merged this impl will support bf16 and once #3979 is merged -- s8, u8, s32 will also be supported.

All unit and benchdnn tests pass locally and we get a broader coverage in the 10 relevant units tests (with the two above PRs used):
```
deconvolution tests:
	cpu-primitives-deconvolution-cpp
	test_deconvolution
	test_benchdnn_modeC_deconv_all_cpu
	test_benchdnn_modeC_deconv_all_f32_nxc_cpu
	test_benchdnn_modeC_deconv_bfloat16_cpu
	test_benchdnn_modeC_deconv_bfloat16_nxc_cpu
	test_benchdnn_modeC_deconv_bfloat16_ymm_cpu
	test_benchdnn_modeC_deconv_float16_nxc_cpu
	test_benchdnn_modeC_deconv_fp8_nxc_cpu
	test_benchdnn_modeC_deconv_int8_cpu
	
baseline:
    total tests: 16829  passed: 2177  skipped: 14652  failed: 0
    
with this change:
    total: 16829  passed: 5953  skipped: 10876  failed: 0
```

Performance-wise, brgdeconv covers a lot of problems not covered by existing ACL implementation but for covered shapes in `inputs/deconv/shapes_2d` and with `dt=f32`, brgdeconv is about 1.08x faster on average (c8g instance, 64 threads), but has quite a variance with min being 0.77x and max being  3.94x.

### DIFF
--- a/src/cpu/aarch64/jit_brgemm_deconv.cpp
+++ b/src/cpu/aarch64/jit_brgemm_deconv.cpp
@@ -1,0 +1,203 @@
+/*******************************************************************************
+* Copyright 2022-2025 Intel Corporation
+* Copyright 2025 Arm Ltd. and affiliates
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "common/dnnl_thread.hpp"
+#include "common/nstl.hpp"
+#include "common/primitive_desc_iterator.hpp"
+#include "common/type_helpers.hpp"
+#include "cpu_isa_traits.hpp"
+
+#include "cpu/aarch64/jit_brgemm_deconv.hpp"
+
+#include "cpu/aarch64/jit_brgemm_1x1_conv.hpp"
+#include "cpu/aarch64/jit_brgemm_conv.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+namespace {
+status_t fwd_conv_desc_create(const deconvolution_desc_t *fwd_deconv_d,
+        convolution_desc_t *fwd_conv_d) {
+    const memory_desc_t &fwd_weights_md = fwd_deconv_d->weights_desc;
+    // create a fwd convolution descriptor with padding adjusted
+    // to the perspective of backward propagation, namely:
+    // - left padding replaced by left overflow
+    // - right padding replaced by right overflow
+    const int ndims_spatial = fwd_deconv_d->dst_desc.ndims - 2;
+    dims_t overflow_l;
+    dims_t overflow_r;
+    dim_t ks = 1;
+    for (int i = 0; i < ndims_spatial; i++) {
+        VDISPATCH_DECONVOLUTION_IC(fwd_deconv_d->strides[i] == 1,
+                VERBOSE_UNSUPPORTED_FEATURE,
+                "only unit strides are allowed for bwd-to-fwd conversion");
+        const dim_t K
+                = fwd_weights_md.dims[fwd_weights_md.ndims - ndims_spatial + i];
+        ks *= K;
+        const dim_t D = fwd_deconv_d->dilates[i];
+        const dim_t PL = fwd_deconv_d->padding[0][i]; // left padding
+        const dim_t PR = fwd_deconv_d->padding[1][i]; // right padding
+        constexpr dim_t S = 1;
+        // the following relations hold for unit stride only
+        overflow_l[i] = ((K - 1) * (D + 1) - PL) / S;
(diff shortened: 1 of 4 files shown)


src/cpu/aarch64/jit_brgemm_deconv.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+        while (++it != it.end()) {
+            conv_pd_ = *it;
+            // try 1x1 fwd convolution
+            if (check_embedded_impl_init<
+                        typename brgemm_1x1_convolution_fwd_t<isa>::pd_t>(it)
+                    == status::success) {
+                break;
+            }
+            // try non-1x1 fwd convolution with invert weights' spatial indices
```
jondea: We don't appear to be inverting anything

src/cpu/aarch64/jit_brgemm_deconv.hpp
```diff
@@ -0,0 +1,117 @@
...
+#include "cpu/aarch64/cpu_isa_traits.hpp"
+
+#include "cpu/cpu_deconvolution_pd.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+template <cpu_isa_t isa>
```
jondea, line 35: Do we need to template this? As far as I can tell, the wrapper code has no xbyak. And even if the underlying conv impls are templated, I think we can handle them as their base classes. We definitely don't need the isa in the impl name, because they don't add anything

src/cpu/aarch64/jit_brgemm_deconv.cpp
michalowski-arm: fixed

src/cpu/aarch64/jit_brgemm_deconv.hpp
michalowski-arm: I think this needs at minimum an `sve` template, since the underlying impls used only support sve. But beyond that, yes, it's not explicitly needed. That said, I tried removing the templating entirely and it created a big, new set of issues. I'd rather deal with them later in a follow up PR, if we agree on a different approach; Especially since it doesn't impact the functioning of this deconv impl and is inline with existing brgemm kernels for now.

src/cpu/cpu_deconvolution_list.cpp
```diff
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2019-2025 Intel Corporation
 * Copyright 2022 FUJITSU LIMITED
-* Copyright 2022, 2024 Arm Ltd. and affiliates
+* Copyright 2022-2025 Arm Ltd. and affiliates
```
Ryo-not-rio: this should be 2022, 2024-2025

Ryo-not-rio:
Do we have tests for this already in the ci? If not could you add them?

michalowski-arm:
> Do we have tests for this already in the ci? If not could you add them?

Yea, we already have the tests: `test_benchdnn_modeC_deconv_ci_cpu`, `test_deconvolution` and `cpu-primitives-deconvolution-cpp` (and more in the nightly set)

src/cpu/cpu_deconvolution_list.cpp
michalowski-arm: done

src/cpu/aarch64/jit_brgemm_deconv.hpp
jondea: No problem, thanks for trying. This is a code quality issue, happy to address later

jondea: About the support question, I think you could filter out `asimd` with a runtime check. I suspect you can type erase the underlying impls and just deal with base pointers.

src/cpu/aarch64/jit_brgemm_deconv.cpp
```diff
@@ -0,0 +1,203 @@
...
+    const int ndims_spatial = fwd_deconv_d->dst_desc.ndims - 2;
+    for (int i = 0; i < ndims_spatial; i++) {
+        if (fwd_deconv_d->strides[i] != 1) {
+            has_strides_ = true;
+            break;
+        }
+    }
+
+    if (has_strides_) {
+        return status::unimplemented;
```
aditew01, line 137: adding a verbose here would be easier spot.

src/cpu/cpu_deconvolution_list.cpp
```diff
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2019-2025 Intel Corporation
 * Copyright 2022 FUJITSU LIMITED
-* Copyright 2022, 2024 Arm Ltd. and affiliates
+* Copyright 2022, 2024-2025 Arm Ltd. and affiliates
```
mgouicem, line 4: (nit) 
```suggestion
* Copyright 2022-2025 Arm Ltd. and affiliates
```

mgouicem (approved):

src/cpu/cpu_deconvolution_list.cpp
Sqvid: I think our convention is a little different. We use the hyphen only for contiguous years. A comma separates edit-years with gaps.

PR #4010, opened 2025-09-25
revert: "cpu: aarch64: support for dst u8 in brgemm..."

Sqvid:
# Description

This reverts commit 8ac246e5a338ab274fea4f5ca433836e5391aefc.

Reverts: #3911

Reverting as this causes an assert crash in nightly.

# Checklist

## General

### Bug fixes

- [X] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?


Sqvid:
cc: @kasturedeeksha 

I had left a [comment about this on your latest PR](https://github.com/uxlfoundation/oneDNN/pull/3979#issuecomment-3323142054). I'll keep the PR open for ~24 hours but will revert it after that as we need nightly running. Thanks.

Sqvid:
Failing case
```sh
$ ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8 --attr-scales=src:common:0.25+dst:common:2+wei:per_oc --attr-zero-points=src:common:135+dst:common:1 --attr-post-ops=sum:1:0:s8 ic3ih224oc64oh224kh3ph1n"vgg_19:conv1_1"
```

dzarukin:
> Failing case
> 
> ```shell
> $ ./build/tests/benchdnn/benchdnn --conv --dt=u8:s8:u8 --attr-scales=src:common:0.25+dst:common:2+wei:per_oc --attr-zero-points=src:common:135+dst:common:1 --attr-post-ops=sum:1:0:s8 ic3ih224oc64oh224kh3ph1n"vgg_19:conv1_1"
> ```

I'd probably try to give some hints to the author before performing a revert. It's usually helpful to find a minimal reproducer to identify the problem. Have you tried to remove attributes from the case? It might be a simple check to add since the new functionality might not have it supported. If it's just data type and shape failing, that would indicate a problem in the output or something if identical s8 dst_dt works fine.

Providing failing dots also helps to understand what might go wrong.

It's ultimately not my business but reverts also has cost (some maintainers review on things out and then further in) and I'm not sure treating a single failing test case in a testing process that suppose to catch issues and have issues until they fixed is a comparable price. I don't advocate to change anything now but it might be a good topic to discuss in the next sync meeting.

kasturedeeksha:
I checked this case.
This failure is due to the post-op sum, which is in s8 and uses the cvt2ps function. 
Since s8 was not supported there, it caused an assertion error.

In the PR, this has been fixed: https://github.com/uxlfoundation/oneDNN/pull/3979.
I also added a check so that unsupported cases now return unimplemented beforehand.

Sqvid:
@dzarukin Thanks for the perspective. Yes we definitely try to wait it out if it's a small fix. It is just a week with many noisy changes to those files (our brgemm changes + clang-tidy work) so I am a bit more keen to have the nightly running with 1000s of lines of changes merged in this week.

> I'd probably try to give some hints to the author before performing a revert.

Yes I think I pointed out that the datatype for post_ops needed to be checked in the primitive `init()` it's just the details are scattered between these two PRs.

> I don't advocate to change anything now but it might be a good topic to discuss in the next sync meeting.

Yes I agree, it would be good to have a more standardised protocol of what state the main branch is expected to be in. Thanks!

Sqvid:
@kasturedeeksha Thanks for the reply, Yes that change looks good (though I have left a few comments). Let's proceed with the fix there. I'll close this PR for now.

PR #3942, opened 2025-09-15
cpu: gemm_x8s8s32x_conv: fix post_op application

Sqvid:
# Description
Fixed an issue where minibatch was not used when calculating the offset
used by the `ref_post_ops kernel` for binary post-ops.

Also added a check for supported post-op `POLICY`/`MASK` values in
`gemm_x8s8s32x_convolution_utils::post_ops_ok()`

Fixes: #3803

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

### Bug fixes

- [x] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [ ] Have you added relevant regression tests?

PR #4020, opened 2025-09-26
cpu: x64: fix virtual interface

densamoilov:
Fixes virtual `jit_pp_ker_t::operator()` interface (the main one was updated in #3942).


densamoilov:
The linter issue is unrelated and caused by a missing copyright from `scripts/synthdnn/environment.yml`. I updated the ignore list for the copyright checker to skip all YML files in the project (will be visible in the following pull-requests).

PR #4021, opened 2025-09-26
cpu, conv: fix L2 blocking for AVX2

georgen117:
[MFDNN-14078](https://jira.devtools.intel.com/browse/MFDNN-14078)
Removed the hard coded 2 MB L2 blocking when AVX2 isa is used for the brgemm conv. Replaced it with value read from cpuid.

For MTL and PTL systems the hard coded value was too large and resulted in poor performance for some input shapes.

MTL system:
```
sdpmkl170254
$ lscpu
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          46 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   22
  On-line CPU(s) list:    0-21
Vendor ID:                GenuineIntel
  Model name:             Intel(R) Core(TM) Ultra 7 1003H
    CPU family:           6
    Model:                170
    Thread(s) per core:   2
    Core(s) per socket:   16
    Socket(s):            1
    Stepping:             4
    CPU(s) scaling MHz:   14%
    CPU max MHz:          4800.0000
    CPU min MHz:          400.0000
    BogoMIPS:             5990.40
```
###main
```
./benchdnn --fix-times-per-prb=10 --mode=p --cold-cache=all --conv --reset --allow-enum-tags-only=0 --engine=cpu --dir=FWD_I --alg=direct --dt=f32:f32:f32 --bia-dt=f32 --stag=acdb --wtag=any --dtag=acdb --attr-post-ops=eltwise_relu --attr-scratchpad=user mb1_ic1024oc512_ih46oh46kh3sh1dh0ph1_iw60ow60kw3sw1dw0pw1
Output template: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,brg_conv_fwd:avx2,,--mode=P --fix-times-per-prb=10 --conv --allow-enum-tags-only=false --cold-cache=all --dir=FWD_I --bia-dt=f32 --stag=acdb --dtag=acdb --attr-post-ops=relu --attr-scratchpad=user mb1ic1024ih46iw60oc512oh46ow60kh3kw3ph1pw1,25.3839,1.52271,63.353,400.674,73.9081,343.453
| brg_conv_fwd:avx2 : 1 (100%)                             |
total perf: min(ms):63.353 avg(ms):73.9081
total: 0.89s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 0.03s (3%); execute: 0.07s (8%);
```
###PR
```
./benchdnn --fix-times-per-prb=10 --mode=p --cold-cache=all --conv --reset --allow-enum-tags-only=0 --engine=cpu --dir=FWD_I --alg=direct --dt=f32:f32:f32 --bia-dt=f32 --stag=acdb --wtag=any --dtag=acdb --attr-post-ops=eltwise_relu --attr-scratchpad=user mb1_ic1024oc512_ih46oh46kh3sh1dh0ph1_iw60ow60kw3sw1dw0pw1
Output template: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,brg_conv_fwd:avx2,,--mode=P --fix-times-per-prb=10 --conv --allow-enum-tags-only=false --cold-cache=all --dir=FWD_I --bia-dt=f32 --stag=acdb --dtag=acdb --attr-post-ops=relu --attr-scratchpad=user mb1ic1024ih46iw60oc512oh46ow60kh3kw3ph1pw1,25.3839,108.092,41.6938,608.817,44.1903,574.423
| brg_conv_fwd:avx2 : 1 (100%)                             |
total perf: min(ms):41.6938 avg(ms):44.1903
total: 0.75s; create_pd: 0.08s (11%); create_prim: 0.03s (4%); fill: 0.07s (9%); execute: 0.05s (7%);
```
Results of performance testing on ADL system. (all shapes that showed 100% were hidden)
image

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_utils.cpp
@@ -1619,13 +1619,16 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
 
     // take L1 as 7/8 of the real size for L1
     brg_blocking_t::L1 = (platform::get_per_core_cache_size(1) * 7) / 8;
-    brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
-    // here is hard-coded L2 size for avx2 performance cores
-    // TODO: get L2 size from the platform
-    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2))
-        brg_blocking_t::L2 = 2 * 1024 * 1024;
-    // take L2 as 3/4 of the real size for L2
-    brg_blocking_t::L2 = (brg_blocking_t::L2 * 3) / 4;
+    // The L2 cache size appears to return smaller than expected for blocking
+    // purposes. When using avx2, the full size of L2 is used for blocking
+    // for other isa 3/4 of the size is used for blocking
+    // TODO : investigate the L2 cache size reporting issue
+    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2)) {
+        brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
+    } else {
+        // take L2 as 3/4 of the real size for L2
+        brg_blocking_t::L2 = (platform::get_per_core_cache_size(2) * 3) / 4;
+    }
 
     // These are rough estimates of the latency (relative) of access to various
     // cache levels. This is enough for an estimation of data access cost.


georgen117:
make test

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -1619,13 +1619,16 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
...
-        brg_blocking_t::L2 = 2 * 1024 * 1024;
-    // take L2 as 3/4 of the real size for L2
-    brg_blocking_t::L2 = (brg_blocking_t::L2 * 3) / 4;
+    // The L2 cache size appears to return smaller than expected for blocking
+    // purposes. When using avx2, the full size of L2 is used for blocking
+    // for other isa 3/4 of the size is used for blocking
+    // TODO : investigate the L2 cache size reporting issue
+    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2)) {
+        brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
+    } else {
```
ankalinin, line 1628: Could you please elaborate why we don't apply 3/4 for the avx2?
As far as I remember, we apply this empirical coefficient to L2 size for all architectures, setting aside a quarter  of cache for other expenses, such as code etc.

georgen117: @ankalinin I talked with you about the logic of using the hardcode value that was removed was because the `get_per_core_cache_size` was under reporting the usable cache size. For that reason a hardcoded value was used.

Based on the original idea that the hardcode value was there for a reason, I tested using both the `cache size` and `3/4 cache size`.

The `cache size` value performed the same as or better than the `3/4 cache size` for all shapes on the ADL system.  The same held true for testing 3 specific shapes that were identified in the MFDNN-14078 on MTL and PTL systems. They all performed the same or better using `cache size` instead of `3/4 cache size`.

georgen117: @ankalinin I am trying to do my due diligence to test the performance of the change. If there is something I am missing that I should be doing, I am 100% open to suggestions or directions to improve the process.

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -1619,13 +1619,16 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
...
     // take L1 as 7/8 of the real size for L1
     brg_blocking_t::L1 = (platform::get_per_core_cache_size(1) * 7) / 8;
-    brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
-    // here is hard-coded L2 size for avx2 performance cores
-    // TODO: get L2 size from the platform
-    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2))
-        brg_blocking_t::L2 = 2 * 1024 * 1024;
-    // take L2 as 3/4 of the real size for L2
-    brg_blocking_t::L2 = (brg_blocking_t::L2 * 3) / 4;
+    // The L2 cache size appears to return smaller than expected for blocking
```
densamoilov, line 1622: (random spot) Some of the cases from the performance data you shared have significantly degraded (up to 33%). We can't merge this change until the regressions are addressed.

georgen117: Apparently, I was reading the numbers backwards. I interpreted the 133% from the measured time as a gain not a loss. I interpreted the green coloring in the spreadsheet as a gain. Thanks @densamoilov for pointing out my misunderstanding. I wouldn't have pushed this pr if I have been reading the spreadsheet correctly.  I am going to close the PR and do some more investigation.

georgen117:
Closing due to misunderstanding the performance data spreadsheet.

src/cpu/x64/jit_brgemm_conv_utils.cpp
densamoilov: Using ~75% of L2 is quite common, as it leaves some headroom for other operations. So it's better to stick with that.

IMO, the hardcoded cache size likely had a cumulative effect, since it probably influenced how we tuned the implementation for different problems. As a result, all tuning parameters that depend on the L2 cache size may have been selected incorrectly. They relied on a hardcoded value so when the value changes the tuning parameters likely need to be updated as well to recalibrate the tuning logic.

Regarding `get_per_core_cache_size`, I think we should consider making it more robust so that it returns the correct cache size, where "correct" depends on how we define it. Currently, I believe we expect it to return the L2 cache size per physical core (ignoring hyper-threading).

For hybrid architectures, it currently returns incorrect sizes because it doesn't accurately reflect how the cache is organized.
For example, in Lunar Lake:
* The actual L2 cache per physical P-core is 2560 KB.
* However, `get_per_core_cache_size` returns 320 KB because it assumes the cache is shared across 8 cores, while in reality it's private to each of the four P-cores.

L1 cache sizes are also reported incorrectly, and the same goes for E-cores.

I think we should fix `get_per_core_cache_size` first, because otherwise we’d just be replacing one incorrect L2 value (hardcoded) with another inaccurate one. That would lead to further incorrect tuning of the parameters, which we want to avoid. Once it’s fixed, we can adjust the tuning parameters based on the correct cache sizes to close any performance gap, if one exists.

georgen117: @densamoilov using this thread to talk about the best fix.

I agree with you that the ~75% of L2 is the correct fix. Now that I understand the performance results. The ~75% gives slightly better performance than the code in the PR.

The hybrid architectures, currently return the incorrect size because the **Extended Topology Enumeration Leaf (Initial EAX Value = 0BH, ECX ≥ 0)** or **V2 Extended Topology Enumeration Leaf (Initial EAX Value = 1FH, ECX ≥ 0)** does not return the actual number of cores sharing a cache for the hybrid systems. At least I was unable to get the correct value with a lot of experimentation. 

The only solution I could find to get an accurate value for the cache was to use OS specific systems calls.

**On windows:**
the `GetLogicalProcessorsInformationEx` function from `windows.h` can be used to get the system cache size as well ask which cores are sharing cache. 

**On Linux**
the cache layout of the specific system can be read and parced from the `/sys/devices/system/cpu`

**On Mac** (don't know if this is a target for oneDNN) and was not tested by me
the cache layout can be obtained using `getSysctlInt` from `the sys/sysctl.h`

Right now oneDNN uses `xbyak_utils.h` to get the cache size and the cores sharing a cache. For L1 cache it gives accurate results. For non-hybrid systems it also seems to give accurate results. The issue seems to be that the **Extended Topology Enumeration** leaf does not return the actual value of the system. The documentation from the Intel Software Development manual notes the following for the **Extended Topology Enumeration** leaf:

From Intel SDE Vol 2A. Table 3-17. Information Returned by CPUID Instruction:

> Software must not use the value of EBX[15:0] to enumerate processor topology of the system. The 
> value is only intended for display and diagnostic purposes. The actual number of logical processors available to 
> BIOS/OS/Applications may be different from the value of EBX[15:0], depending on software and platform
> hardware configurations. 

Which seems to indicate if we want to get the actual cache values we likely need to use OS specific solutions.

PR #3948, opened 2025-09-16
cpu: riscv: pool: add post_ops

zhangfeiv0:
# Description

This PR introduces post-ops support to the RISC-V Vector (RVV) optimized NCHW pooling forward (MaxPooling, AvgPoolingIncludePadding, AvgPoolingExcludePadding).

PR #4029, opened 2025-09-30
Add rvv postops binary op support via a primitive-based approach

zhangjian29:
# Description

This PR adds binary operation support for `rvv_postops.hpp` using a primitive-based approach, following the implementation style of `acl_post_ops` (aarch64) and `jit_brgemm_post_ops` (x64). The binary operations are implemented by integrating with `rvv_binary` introduced in #3899.

# WIP

This PR is based on the unmerged `add-rvv-binary-integration` branch (simply include the header file `rvv_binary.hpp`) and will be rebased after #3899 is merged.
Current state: Ready for preliminary review.

# Refactoring `rvv_postops`

## Background

The current `rvv_postops.hpp` implementation uses an `apply` method during execution, which accepts a `vfloat32m1_t` input and `vl` and returns a `vfloat32m1_t` output. However, this kind of fine-grained approach:

- Limits the vectorization of other postop sources
- Requires redundant vectoization logic when applying postops
- Restricts data type support extensibility

Since only two implementations currently use this postop approach (merged: rvv_matmul, unmerged: #3948), making changes at this early stage requires minimal effort.

## Motivation

To address these limitations, we propose a primitive-based implementation following the `acl_post_ops`(AArch64) and `jit_brgemm_post_ops`(x64) patterns for better maintainability.

## Key Changes

- Added primitive-style `init` and `execute` methods (aligned with `acl_postops.hpp`)
- Enabled binary operation support via `rvv_binary.hp`p integration (future `rvv_binary` changes won't affect this PR)
- Kept the original `post_ops_ok` and `apply` methods unchanged in a fused way, making other implementation used original version still correct.
- Supported post-ops: 
    1. `relu` (via original methods)
    2. Any `binary` operations (via new primitive methods): `add`, `div`, `max`, `min`, `mul`, `sub`, `ge,` `gt`, `le`, `lt`, `eq`, `ne`, and the ternary `select`.
- Demonstrated functionality by adding primitive-style binary postop support to `rvv_nchw_pooling`

## Future Plans

1. Rebase after #3899 is merged
2. Update postop implementation in `rvv_matmul` and other related code
3. Remove the original `post_ops_ok` and `apply` methods safely
4. Add support for `eltwise` and `sum` operations
5. Extend data type supports.

# Checklist

## General

## Performance improvements

We test the changes using the RISC-V GNU toolchain(14.2) and verify the functionality under the QEMU RISCV64 emulator with cmd:

```
ONEDNN_VERBOSE=1 ./benchdnn --pool --dir=FWD_I --attr-post-ops=add:f32,div:f32,mul:f32,max:f32,min:f32,sub:f32,ge:f32,gt:f32,le:f32,lt:f32,eq:f32,ne:f32,select:f32,linear:0.5:-1,relu --batch=shapes_basic
```

All tests passed with proper dispatching confirmed (`relu` and `linear` are dispatched to `ref` while others are to `RISCV64GCV` implementation). Calls to the implemented `rvv_nchw_pooling` with all kinds of postops can be traced by searching for `RISCV64GCV` in:

- [test_pool_postops.log](https://github.com/user-attachments/files/22611822/test_pool_postops.log)


zhangjian29:
Hi @mgouicem ,

Could you please take a look at this PR when you have time? Thank you so much!

src/cpu/rv64/rvv_nchw_pooling.cpp
```diff
@@ -314,6 +315,15 @@ status_t riscv_nchw_pooling_fwd_t::execute_forward(
         return status::unimplemented;
     }
 
+    const post_ops_t &post_ops = pd()->attr()->post_ops_;
+    if (post_ops.len() == 1 && post_ops.entry_[0].is_binary()) {
+        engine_t *engine = ctx.stream()->engine();
+        post_ops_t local_po = post_ops;
+        rvv_postops_t postops;
+        CHECK(postops.init(engine, local_po, *pd()->dst_md()));
```
mgouicem: This init step should not happen in pooling execute function, but within the pooling init method. There are two reasons for that:
- avoid primitive creation overhead at each execute call
- allow dispatching to work properly if for some reason unimplemented is returned in postops.init

src/cpu/rv64/rvv_postops.hpp
```diff
@@ -90,7 +90,7 @@ struct rvv_postops_t {
         if (po.len() > 1) return false;
 
         const auto &e = po.entry_[0];
-        if (e.is_binary()) return true;
+        if (e.is_binary()) return false;
```
mgouicem: maybe just delete this line? the check below should properly return false in case of binary.

src/cpu/rv64/rvv_nchw_pooling.cpp
zhangjian29: Thank you for your help. Learned a lot from your feedback.

zhangjian29:
Thank you so much for your feedback @mgouicem. Ready for review now.

src/cpu/rv64/rvv_nchw_pooling.hpp
```diff
@@ -71,8 +79,17 @@ struct riscv_nchw_pooling_fwd_t : public primitive_t {
                     VERBOSE_UNSUPPORTED_FEATURE,
                     "kernel width exceeds maximum");
 
+            if (!attr()->post_ops_.has_default_values()) {
+                const auto &po = attr()->post_ops_;
+                if (po.len() == 1 && po.entry_[0].is_binary()) {
+                    post_ops_t local_po = po;
```
mgouicem: (nit) why do you need this local copy of `po`? Won't the copy happen anyway as part of init?

src/cpu/rv64/rvv_nchw_pooling.hpp
```diff
@@ -71,8 +79,17 @@ struct riscv_nchw_pooling_fwd_t : public primitive_t {
...
+                    CHECK(postops_.init(engine, local_po, *dst_md()));
+                }
+            }
             return status::success;
         }
+
+        rvv_postops_t postops_;
```
mgouicem, line 91: (nit) maybe use a pointer here to avoid allocation when not needed?

src/cpu/rv64/rvv_postops.cpp
```diff
@@ -0,0 +1,67 @@
...
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace rv64 {
+
+status_t rvv_postops_t::execute(
+        const exec_ctx_t &ctx, void *src, void *dst) const {
+    int post_op_index = post_op_start_index_;
+
+    VINFO(primitive, exec, check, cpu,
```
mgouicem: The check verbose level is typically used for creation time checks, as those help user understand why some implementations failed to dispatch.

These prints here seem to be debugging prints at execution time, so maybe use debuginfo level instead. You can also rely on the macros available in src/common/verbose.hpp (in particular VDEBUGINFO), that should help with proper formatting.

src/cpu/rv64/rvv_nchw_pooling.hpp
zhangjian29: because `rvv_postops_t::init` mutates its `post_ops_t` argument and requires a non-const lvalue reference. `attr()->post_ops_` is accessed as const, and we don’t want to (and cannot) mutate it, so we pass a mutable copy.

src/cpu/rv64/rvv_nchw_pooling.hpp
zhangjian29: A pointer won’t save allocations here because `rvv_postops_t` by-value does not allocate unless init is called. The vector is empty and no heap work happens until we build post-op primitives, which is already gated behind the `binary` check.

src/cpu/rv64/rvv_postops.cpp
zhangjian29: Thanks for pointing that out. Removing them now.

src/cpu/rv64/rvv_nchw_pooling.hpp
mgouicem: I am not sure I understand here: why not prevent `rvv_postops_t::init` from mutating its argument by making that argument const and copy it inside the function (you are already doing it line 55 with the copy to `po_`), you can then call set_default_format on that copy. This would avoid the need for spurious copies.

zhangjian29: You are absolutely right. I've made the changes. Appreciate your patience and valuable suggestions.

Issue #2175, opened 2024-10-21
Bug in memory_desc_init_by_tag: Incorrect Differentiation Between Memory Tags abcd and acbd


taoye9:
# Summary

oneDNN deduces memory tag from input tensor shape using this function:

`status_t status = memory_desc_init_by_tag(md_gold, md.ndims, md.dims, md.data_type, tag);` 

we found when the input memory description is for 4d fp32 tensor of shape `3x1x3x3`, the returned status are  successful for both of memory tag acbd and abcd.  that is `md.dims = [3, 1, 3, 3]`

This bug causes issues: https://github.com/oneapi-src/oneDNN/issues/2008. 

# Version
oneDNN v3.7.0 (commit fca8b85b7fc3373b169fc066e9e6a00374e15d9c)

# Environment
* CPU: Neoverse-V1 C7g.4xlarge
* OS version: Ubuntu 20.04
* Compiler version: gcc-10
* CMake version: 3.22.1
* git hash: 73ae041f2b5e956b58bd58ca5d121af839af076a

# Steps to reproduce

oneDNN/build/tests/benchdnn/benchdnn --engine=cpu --matmul --wtag=acbd --dtag=abcd 1x1x2x3:3x1x3x3

# Observed behavior
The cmd invokes function  `memory_desc_matches_one_of_tag(B_md, plain_tensor_layout_tag, transposed_tensor_layout_tag, acbd, adbc);` to determine memory tag of weight but get abcd instead of acbd.

# Expected behavior
the returned memory tag of weight should be acbd.

shu1chen:
Hello @taoye9, I was trying your steps to see if it's reproducible on a x86_64 machine and got this result:
```
$ ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn --engine=cpu --matmul --wtag=acbd --dtag=abcd 1x1x2x3:3x1x3x3
...
onednn_verbose,v1,primitive,exec,cpu,matmul,gemm:jit:f32,undef,src:f32:a:blocked:abcd::f0 wei:f32::blocked:acbd::f0 dst:f32::blocked:abcd::f0,,,1x1x2x3:3x1x3x3,0.791992
```
The weight shows acbd on x86_64, while it shows abcd on aarch64. Am I understanding right?

taoye9:
Hi, @shu1chen, thanks for reply. 

[...]

PR #4035, opened 2025-09-30
cpu: aarch64: memory stride detection update

henry-gar:
There were issues were memory tag formats that shouldn't have been accepted were accepted by gemm:acl. The reason behind this is discussed here https://github.com/uxlfoundation/oneDNN/issues/2175#issuecomment-2431603405.

For example, the case:

--matmul --stag=adbc --wtag=adbc --dtag=acbd 14x6x1x253:1x1x253x1_n"dot_prod_w_B_full_bcast"

Now correctly skips gemm:acl and falls to gemm:jit, whereas before it failed.

# Checklist

## General

- [Y] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [Y] Have you formatted the code using clang-format?

## Performance improvements

- [Y] Have you submitted performance data that demonstrates performance improvements?

### Bug fixes

- [Y] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [Y] Have you added relevant regression tests?

### DIFF
--- a/src/cpu/aarch64/matmul/acl_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/acl_matmul_utils.cpp
@@ -16,12 +16,33 @@
 
 #include "cpu/aarch64/matmul/acl_matmul_utils.hpp"
 #include "cpu/aarch64/acl_utils.hpp"
+#include "cpu/matmul/gemm_based_common.hpp"
 #include "cpu/matmul/matmul_utils.hpp"
 
 namespace dnnl {
 namespace impl {
 namespace cpu {
 namespace aarch64 {
+
+bool batch_dims_have_default_order(const memory_desc_wrapper &mdw) {
+    assert(mdw.is_blocking_desc());
+
+    if (mdw.ndims() <= 2) { return true; }
+
+    const auto &dims = mdw.dims();
+    const auto ndims = mdw.ndims();
+    const auto &strides = mdw.strides();
+
+    int prod = dims[ndims - 1] * dims[ndims - 2];
+
+    for (int i = ndims - 3; i >= 0; --i) {
+        if (strides[i] != prod) { return false; }
+
+        prod *= dims[i];
+    }
+
+    return true;
+}
 
 namespace acl_matmul_utils {
 
@@ -41,6 +62,10 @@ status_t init_conf_matmul(acl_matmul_conf_t &amp, memory_desc_t &src_md,
     const dim_t dst_batch = helper.batch();
     const dim_t src_batch = helper.src_batch();
     const dim_t wei_batch = helper.wei_batch();
+
+    ACL_CHECK_SUPPORT(
+            src_d.ndims() > 4 || wei_d.ndims() > 4 || dst_d.ndims() > 4,
+            "ACL does not support more than 4 dimensions");
 
     // We can only broadcast on one of src or wei at once
     // ACL supports broadcast for 3D shapes, and 4D shapes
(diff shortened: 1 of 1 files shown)


Sqvid:
@Ryo-not-rio, @jondea  could you have a look?

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
```diff
@@ -72,26 +97,31 @@ status_t init_conf_matmul(acl_matmul_conf_t &amp, memory_desc_t &src_md,
...
+    ACL_CHECK_SUPPORT(!matmul::gemm_based::check_gemm_input_format(src_md),
+            "at least one innermost dimension must be contiguous");
+    ACL_CHECK_SUPPORT(!matmul::gemm_based::check_gemm_output_format(dst_md),
+            "innermost dst dimension must be contiguous");
+    if (!IsFixedFormat) {
+        ACL_CHECK_SUPPORT(!matmul::gemm_based::check_gemm_input_format(wei_md),
+                "at least one innermost dimension must be contiguous");
     }
 
+    //added check to ensure dense tensors
```
Ryo-not-rio, line 110: (nit) `// added check to ensure dense tensors`

Ryo-not-rio:
This looks good to me, but we might wanna run the full performance tests to make sure we're not accidentally making things go to jit even when supported by acl. Could you run the `.github/automation/performance/bench_nightly_performance.sh` script on 16 threads and then run `.github/automation/performance/benchdnn_comparison.py`? Feel free to comment out the non-matmul tests

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
```diff
@@ -42,6 +63,10 @@ status_t init_conf_matmul(acl_matmul_conf_t &amp, memory_desc_t &src_md,
     const dim_t src_batch = helper.src_batch();
     const dim_t wei_batch = helper.wei_batch();
 
+    ACL_CHECK_SUPPORT(
```
jondea, line 66: There's now a generic macro for this which improves the logging, something like VDISPATCH

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
```diff
@@ -16,13 +16,34 @@
...
 #include "cpu/aarch64/acl_utils.hpp"
+#include "cpu/matmul/gemm_based_common.hpp"
 #include "cpu/matmul/matmul_utils.hpp"
 
 namespace dnnl {
 namespace impl {
 namespace cpu {
 namespace aarch64 {
 
+bool is_batch_ordered(const memory_desc_wrapper &mdw) {
```
jondea: I think we could be a little more precise with the naming here, perhaps something like `batch_dims_have_default_order`

Sqvid: @jondea @henry-gar 
Interestingly, there is a similar function I just found in the x64 path. Maybe we can borrow the terminology of "trivial" so that it is consistent in the library?

https://github.com/uxlfoundation/oneDNN/blob/42f4a366fc0f78e9f9b94814104e40db5733341d/src/cpu/x64/matmul/brgemm_matmul_utils.cpp#L193-L214

jondea: Possibly, although it does look slightly different. I actually think `default_order` is more specific than `layout_trivial`, but happy to go with either!

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
henry-gar: Changed.

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
henry-gar: VDISPATCH_MATMUL can only be used in acl_matmul.cpp; therefore ACL_CHECK_SUPPORT is the only option.

src/cpu/aarch64/matmul/acl_matmul_utils.cpp
henry-gar: Changed to batch_dims_have_default_order.

henry-gar:
I've run the bench_nightly_performance.sh script comparing current main and my changes - no exec time slowdowns.

I also ran test_matmul_all and saw no changes to the number of passing tests or which impl each case uses.

CC: @Ryo-not-rio @jondea

Sqvid:
@henry-gar Since this is a bug-fix, would you be able to open a backport PR to 3.10?

PR #4036, opened 2025-09-30
cpu: x64: matmul: updated threading scheme for reduction

yair-obodovsky:
Match the reduction threading scheme to that of GEMM section.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul.cpp
@@ -1057,131 +1057,161 @@ void brgemm_matmul_t<isa>::maybe_reduce_partial_results_and_apply_postops(
 
         const auto &bgmmc = pd()->get_brgemm_matmul_conf();
         const int nthr_k = brgmm_ctx.get_num_threads_for_k();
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1057,131 +1057,161 @@ void brgemm_matmul_t<isa>::maybe_reduce_partial_results_and_apply_postops(
 
         const auto &bgmmc = pd()->get_brgemm_matmul_conf();
         const int nthr_k = brgmm_ctx.get_num_threads_for_k();
-        const int ithr_bmn = brgmm_ctx.get_thread_idx_for_bmn(ithr);
+        const int ithr_bmn = brgmm_ctx.get_thread_idx_for_bmn_gemm(ithr);
```
densamoilov, line 1060: (random spot) Do you expect any changes in performance due to this change?

rinatrap: Overall, in the geomean of the performance CI testing there is no difference. However, in selected layers we see up to 20-30% gain.

tprimak (approved):

PR #4039, opened 2025-09-30
cpu: aarch64: brdgmm conv int8 support

kasturedeeksha:
# Description

This PR adds support for int8 in BRDGMM Depthwise Convolution with destination type f32.
src : s8/u8
wei : s8
dst : f32

# Checklist

## General

  1. make test output 

```
Total Test time (real) = 752.16 sec
 
The following tests FAILED:
        207 - test_benchdnn_modeC_graph_ci_cpu (Failed)
Errors while running CTest
```
- [x] Have you formatted the code using clang-format? Yes


src/cpu/aarch64/brgemm/jit_brdgmm_kernel.hpp
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2021-2023 Intel Corporation
-* Copyright 2024 FUJITSU LIMITED
+* Copyright 2024-2025 FUJITSU LIMITED
```
Ryo-not-rio: remove?

Ryo-not-rio:
Thank you for adding this. Could you post some performance numbers?

src/cpu/cpu_convolution_list.cpp
```diff
@@ -525,9 +525,8 @@ const std::map<pk_dt_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map()
             CPU_INSTANCE_AVX2(jit_uni_x8s8s32x_convolution_fwd_t<avx2>)
             CPU_INSTANCE_SSE41(jit_uni_x8s8s32x_1x1_convolution_fwd_t<sse41>)
             CPU_INSTANCE_SSE41(jit_uni_x8s8s32x_convolution_fwd_t<sse41>)
-            CPU_INSTANCE_AARCH64(brgemm_1x1_convolution_fwd_t<sve_256>)
-            CPU_INSTANCE_AARCH64(brgemm_convolution_fwd_t<sve_256>)
             CPU_INSTANCE_AARCH64(jit_sve_512_x8s8s32x_convolution_fwd_t<s8, f32>)
+            CPU_INSTANCE_AARCH64(brdgmm_dw_convolution_fwd_t<sve_256>)
```
jondea, line 546: Given that this is a very specialised kernel, do we not want it higher than any generic ones?

src/cpu/aarch64/jit_brdgmm_dw_conv.cpp
```diff
@@ -102,10 +102,8 @@ status_t brdgmm_dw_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
     // TODO: support s8s8 conv
```
jondea: Can this TODO go?

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -589,7 +607,18 @@ void jit_brdgmm_kernel_base_t::brdgmm_microkernel(int m_blocks, int n_blocks,
...
+                udot(vmm_acc.s, vmmb.b, vmma.b);
+            else if ((brg.dt_a == data_type::s8 && brg.dt_b == data_type::s8)
+                    && isa_has_s8s8(brg.isa_impl))
+                sdot(vmm_acc.s, vmma.b, vmmb.b);
+            else if ((brg.dt_a == data_type::u8 && brg.dt_b == data_type::s8)
+                    && isa_has_s8s8(brg.isa_impl))
+                usdot(vmm_acc.s, vmma.b, vmmb.b);
+            else if ((brg.dt_a == data_type::s8 && brg.dt_b == data_type::u8)
+                    && isa_has_s8s8(brg.isa_impl))
+                assert(!"unsupported\n");
```
jondea, line 621: As far as I can tell, [`UDOT` and `SDOT` are supported by all SVE architectures](https://developer.arm.com/documentation/ddi0602/2022-03/SVE-Instructions/SDOT--vectors---Signed-integer-dot-product-), but [`USDOT` requires `FEAT_I8MM`](https://developer.arm.com/documentation/ddi0602/2022-03/SVE-Instructions/USDOT--vectors---Unsigned-by-signed-integer-dot-product-). But either way, these checks also need to happen in the impl init as well, we don't want to hit these asserts in normal operation.

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -554,10 +572,10 @@ void jit_brdgmm_kernel_base_t::load_b(
             B_offset(n_i) + is_tail_block * v_i * simd_w_ * brg.typesize_B,
             X_TMP_0);
     const auto addr = ptr(X_DEFAULT_ADDR);
-    if (brg.is_f32) {
+    if (brg.dt_b == data_type::f32) {
         ld1w(vmmb.s, P_ALL_ONE / T_z, addr);
     } else if (brg.is_int8) {
```
jondea: If these `ld1`s are loading from b, can we just check the type of b in all the if statements?

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.hpp
kasturedeeksha: Done

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
kasturedeeksha: Yes, you're right
I checked and currently there isn’t any flag or variable in `cpu_isa_traits` or anywhere in oneDNN that performs this check. Do you have any suggestion on how we can detect `FEAT_I8MM` support?

kasturedeeksha:
> Thank you for adding this. Could you post some performance numbers?

For performance measurement, I used benchdnn in performance mode (--mode=p) on an AWS Graviton3E (64-core) system.
Command used :
``./benchdnn --conv --dt=u8:s8:f32 --mode=p --batch=inputs/conv/set_conv_dw``
For brdgmm :

**With brdgmm INT8 Support :**
tests:10431 passed:10431 skipped:0 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total perf: min(ms):63.6934 **avg(ms):80.1569**
total: 31496.28s; create_pd: 0.42s (0%); create_prim: 2.22s (0%); fill: 3.29s (0%); execute: 0.68s (0%);

**Without brdgmm INT8 Support :**
tests:10431 passed:10431 skipped:0 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
total perf: min(ms):120821 **avg(ms):122032**
total: 31670.56s; create_pd: 0.37s (0%); create_prim: 0.57s (0%); fill: 2.58s (0%); execute: 123.53s (0%);

Thus performance boost is **~1500** compared to reference int8 kernel.

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
kasturedeeksha: Done

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
jondea: Ah, yes good point, we need to add support in xbyak really for detecting this. In the meantime, Graviton 3/4 both have I8MM, so I think we can use sve as a proxy for now

jondea (approved):

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -554,11 +570,11 @@ void jit_brdgmm_kernel_base_t::load_b(
...
             X_TMP_0);
     const auto addr = ptr(X_DEFAULT_ADDR);
-    if (brg.is_f32) {
+    if (brg.dt_b == data_type::f32) {
         ld1w(vmmb.s, P_ALL_ONE / T_z, addr);
-    } else if (brg.is_int8) {
-        assert(!"unsupported\n");
-    } else if (brg.is_bf16) {
+    } else if (brg.dt_b == data_type::s8) {
+        ld1sb(vmmb.s, P_ALL_ONE / T_z, addr); //dtype is s8
```
Ryo-not-rio: (nit) this comment seems unnecessary

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -589,7 +605,18 @@ void jit_brdgmm_kernel_base_t::brdgmm_microkernel(int m_blocks, int n_blocks,
         } else if (brg.is_bf16) {
             bfdot(vmm_acc.s, vmmb.h, vmma.h);
         } else if (brg.is_int8) {
-            assert(!"unsupported\n");
+            if ((brg.dt_a == data_type::u8 && brg.dt_b == data_type::u8)
+                    && isa_has_s8s8(brg.isa_impl))
```
Ryo-not-rio: we should move `isa_has_s8s8` into the same if as `brg.is_int8` to avoid the duplication

kasturedeeksha: Updated, please check

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -540,7 +556,7 @@ void jit_brdgmm_kernel_base_t::load_a(
         ld1h(vmma.s, mask, addr);
         lsl(vmma.s, vmma.s, 16);
     } else if (brg.is_int8) {
-        assert(!"unsupported\n");
+        ld1b(vmma.s, mask, addr);
     }
```
Ryo-not-rio, line 562:
```suggestion
    } else assert(!"unsupported data type\n")
```

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -560,8 +576,8 @@ void jit_brdgmm_kernel_base_t::load_b(
     } else if (brg.dt_b == data_type::bf16) {
         ld1h(vmmb.s, P_ALL_ONE / T_z, addr);
         lsl(vmmb.s, vmmb.s, 16);
-    } else if (brg.is_int8) {
-        assert(!"unsupported\n");
+    } else if (brg.dt_b == data_type::s8) {
+        ld1sb(vmmb.s, P_ALL_ONE / T_z, addr);
     }
```
Ryo-not-rio, line 585: ditto

src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp
kasturedeeksha: Updated

kasturedeeksha:
> Could we enable this for sve_128 too? We should avoid diverging sve256 and sve128 further as much as possible. Also looks like the ci is flagging a conv regression with this patch. Could you try rebasing on top of main and see if the regression remains?

For now, we don’t have the bandwidth to extend this patch to SVE128, but we might be able to revisit SVE128 support in the future if we get the bandwidth.
Regarding the CI failure, this regression is occurring in the SVE JIT 1x1 convolution path and isn’t related to this patch. I’ve already rebased on main, and the issue still persists. I also checked other recent PRs, and the same case is failing there as well.

kasturedeeksha:
Hi @dzarukin @vpirogov, requesting your review on this PR. Thank you.

PR #3693, opened 2025-07-30
xe: gemm: reuse f16 strategies for bf16 problems

Simonsays095:
Addresses [MFDNN-13984](https://jira.devtools.intel.com/browse/MFDNN-13984), and follows up and modernizes #3519. Since bf16 and f16 have similar register space requirements, we can reuse existing f16 kernels for bf16 problems. This cuts down on tuning requirements in the future, and adds support for some cases on older hardware which have only been tuned for f16-f8 cases (which means they now support bf16-f8 cases).

I've removed all bf16 kernels except for ones where both src and wei are bf16, due to [this](https://github.com/uxlfoundation/oneDNN/pull/3519#issuecomment-3033049775) comment. If there are regressions, I'll reintroduce some of these kernels and we can transfer then to be listed as f16 later.

PR #4047, opened 2025-09-30
xe: gemm: jit: fix m-grouped bf16 scales

petercad:
Closes MFDNN-14254. #3693 exposed a gemmstone bug where an illegal region could be generated during m-grouped bf16 scales (currently used only in SDPA V quantization).

Native bf16/f32 mixed arithmetic only supports packed bf16 operands, so the f32 scales needed to be repacked to allow unit stride bf16 access for VNNI-format bf16 data.


petercad:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul

kealan-barbieri (approved):

Issue #4055, opened 2025-10-02
brgconv segfaults with log post-op for some shapes


michalowski-arm:
# Summary
brgconv:sve_256 (and brgconv:sve_128 in #4041) segfault when the following combinations of post-ops are being used `relu+log,gelu_tanh+log,log+gelu_tanh+log+gelu_erf` with `g1ic64ih1000oc64oh1000kh3ph128dh127`

However, it fails with 0s instead of NaNs when using `log+relu` and passes the case when using `gelu_erf+relu`.

This may not be an exhaustive list of failing cases and it also does not fail for all shapes.

# Version
oneDNN v3.10.0 (commit 717575bb)

# Environment
- c7g.16xlarge and c8g.16xlarge instances
- Ubuntu 24.04
- gcc 13.3
- cmake 3.28.3

# Steps to reproduce
Build oneDNN without ACL support and run:
`./tests/benchdnn/benchdnn --mode=c -v5 --conv --dt=f32 --attr-post-ops=relu+log g1ic64ih1000oc64oh1000kh3ph128dh127`
or any other combination of post ops above

When `dt=bf16` the same issues can be observed.

# Observed behavior
segfault, correctness issue (0s instead of NaNs)

# Expected behavior
The tests pass

Sqvid:
This is the smallest shape I could find for the `relu+log` segfault:
```
$ ./build/tests/benchdnn/benchdnn --conv --impl='brgconv' --dt=f32 --attr-post-ops=relu+log g1mb1ic32ih128iw128oc32oh128ow128kh2kw2sh1sw1ph0pw0dh0dw0
zsh: segmentation fault (core dumped)
```
@Shreyas-fuj @kasturedeeksha 
I might try to have a look at this but your expertise with these kernels would be very valuable as well.

Sqvid:
`tanh` failures are tracked here: https://github.com/uxlfoundation/oneDNN/issues/4089

Sqvid:
> However, it fails with 0s instead of NaNs when using log+relu

Not a bug, see https://github.com/uxlfoundation/oneDNN/pull/4085#issuecomment-3385848769

PR #4063, opened 2025-10-03
cpu: aarch64: fix segfault in eltwise_log injector

Sqvid:
# Description

Fixes a segfault that could happen because of a use-after-free bug in `eltwise_injector`.

Fixes: #4055

# Checklist

## General

### Bug fixes

- [X] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -21,6 +21,7 @@
 #include "common/nstl.hpp"
 #include "common/utils.hpp"
 
+#include <cstdint>
 #include "cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp"
 
 #define IDX(a) static_cast<uint32_t>((a).getIdx())
@@ -1002,22 +1003,43 @@ void jit_uni_eltwise_injector_t<isa>::log_compute_vector_fwd(
     h->dup(t1, wt0);
     h->fcmeq(mask, p_all, t4, t1);
     h->sel(t0, mask, t1, t0);
-
     h->b(exitL);
+
     h->L(tbl1L);
+    // Caution: This pointer can be invalidated if the underlying
+    // Xbyak_aarch64::CodeArray is reallocated. We take note of the current size
+    // so we can jump back to the appropriate spot in the reallocated CodeArray
+    // if needed.
     const float *tbl1Addr = (const float *)h->getCurr();
-    for (size_t i = 0; i < tblN; i++) {
+    auto tbl_max_size = h->getMaxSize();
+    const auto tbl_orig_size = h->getSize();
+    for (size_t i = 0; i < tblN; ++i) {
         fi fi;
         fi.i = (127 << 23) | (i << (23 - tblL));
         fi.f = std::sqrt(2) / fi.f;
         h->dd(fi.i);
     }
     h->L(tbl2L);
-    for (size_t i = 0; i < tblN; i++) {
+    for (size_t i = 0; i < tblN; ++i) {
         fi fi;
+
+        // If the table has been reallocated due to a previous instruction then
+        // jump back to the corresponding location in the new table. This will
+        // essentially be the current position minus the number of instructions
+        // that have been added since we initialised the original pointer.
+        if (tbl_max_size != h->getMaxSize()) {
+            constexpr int instr_size = sizeof(uint32_t);
+            auto tbl_curr_size = h->getSize();
+            auto correction = (tbl_curr_size - tbl_orig_size) / instr_size;
+
+            tbl1Addr = (const float *)h->getCurr() - correction;
+            tbl_max_size = h->getMaxSize();
+        }
+
         fi.f = std::log(tbl1Addr[i]);
         h->dd(fi.i);
     }
+
     h->L(exitL);
 }
 template <cpu_isa_t isa>
--- a/src/cpu/aarch64/jit_brgemm_1x1_conv.cpp
+++ b/src/cpu/aarch64/jit_brgemm_1x1_conv.cpp
@@ -100,9 +100,7 @@ status_t brgemm_1x1_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
                     && one_of(entry.eltwise.alg,
                             // these fail due to label offset being too large
                             alg_kind::eltwise_tanh, alg_kind::eltwise_gelu_tanh,
-                            alg_kind::eltwise_gelu_erf,
-                            // this po segfaults TODO: check issues with f32
-                            alg_kind::eltwise_log);
+                            alg_kind::eltwise_gelu_erf);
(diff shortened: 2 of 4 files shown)


Sqvid:
@michalowski-arm Could you provide the `REPRO` lines that we were skipping before?

src/cpu/aarch64/jit_brgemm_1x1_conv.cpp
```diff
@@ -100,9 +100,7 @@ status_t brgemm_1x1_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
                     && one_of(entry.eltwise.alg,
                             // these fail due to label offset being too large
                             alg_kind::eltwise_tanh, alg_kind::eltwise_gelu_tanh,
-                            alg_kind::eltwise_gelu_erf,
```
jondea, line 103: Why are we re-enabling gelu_erf if this patch is about log?

Sqvid: ~You're right. It looks like I mixed up a couple of trials here. I'll change it to just re-enabling `log`.~

EDIT: I think it is just a formatting thing. `gelu_erf` is till in the list, it just had the comma at the end removed.

I think our skipping logic here isn't very accurate though. For instance `eltwise_tanh` seems to pass for me:

```
./build/tests/benchdnn/benchdnn --conv --impl='brgconv' --canonical=true --dt=bf16 --attr-post-ops=eltwise_tanh g1ic64ih1000oc64oh1000kh3ph128dh127
0:PASSED (1284 ms) __REPRO: --mode=C --mode-modifier= --max-ms-per-prb=3000 --fix-times-per-prb=0 --conv --canonical=true --engine=cpu --fast-ref=true --mem-check=true --allow-enum-tags-only=true --cpu-isa-hints=none --attr-same-pd-check=false --check-ref-impl=false --cold-cache= --buffer-prefix= --execution-mode=direct --dir=FWD_B --dt=bf16:bf16:bf16 --stag=any --wtag=any --dtag=any --strides=:: --alg=direct --attr-post-ops=tanh --ctx-init=auto:auto:auto --ctx-exe=auto:auto:auto --impl=brgconv g1mb2ic64ih1000iw1000oc64oh1000ow1000kh3kw3sh1sw1ph128pw128dh127dw127
| brgconv:sve_128 : 1 (100%)                               |
total: 1.29s; create_pd: 0.01s (0%); create_prim: 0.01s (0%); fill: 0.41s (31%); execute: 0.05s (4%); compute_ref: 0.09s (7%); compare: 0.42s (33%);
```

Let's wait to see what @michalowski-arm says.

michalowski-arm: Yeah, tanh on its own passes but in combination with pretty much any other post-op I tested it failed. But then again gelu_tanh fails even on its own with the same error. Ultimately, I meant this skip list as only a placeholder since I'm hoping we can fix these failures quickly and really I'm not certain that there isn't a case where tanh fails on its own. If we want to be more precise though, we could single out tanh as being allowed if it is the only post-op.

Sqvid: Was this failing in any tests or production situations that we are aware of? I was not a reviewer on the original PR but if there isn't a known failure in the wild, then I would prefer to just remove this entire `is_failing_po` block since it isn't really the right approach to skip bugs here in my opinion.

michalowski-arm: Yea the failing post ops are listed here because I ran into failures in the nightly test set, tanh on its own wasn't one of them but since I narrowed down the sum+tanh failure to an issue with tanh implementation specifically I felt it was safer to skip it completely for now as well. Note that we didn't run into this issue before because for f32 the shapes involved need to be slightly larger to cause a failure, so we were just "lucky" and avoided the failures.

Sqvid: I've opened an issue here to track the `tanh` failures https://github.com/uxlfoundation/oneDNN/issues/4089

Ryo-not-rio:
I don't think relu+log is covered in our testset, could you please add it?

Sqvid:
> I don't think relu+log is covered in our testset, could you please add it?

Done.

Sqvid:
@dzarukin are you okay with the testing changes?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -1002,22 +1003,43 @@ void jit_uni_eltwise_injector_t<isa>::log_compute_vector_fwd(
...
     h->fcmeq(mask, p_all, t4, t1);
     h->sel(t0, mask, t1, t0);
-
     h->b(exitL);
+
     h->L(tbl1L);
+    // Caution: This pointer can be invalidated if the underlying
+    // Xbyak_aarch64::CodeArray is reallocated. We take note of the current size
+    // so we can jump back to the appropriate spot in the reallocated CodeArray
+    // if needed.
```
dzarukin, line 1012: I guess this issue exactly illustrates why all the tables that x64 manages are going to the bottom of the code. I wonder if it's possible to follow the model with `prepare_table` call right after `postambule`, because this solution doesn't look scalable to me with multiple algorithms involved.

tests/benchdnn/inputs/conv/harness_conv_regression_general
```diff
@@ -420,3 +420,9 @@ mb1_g50ic50oc50_ih28oh26_kh3sh1ph0_n"dw_post-op_ngroups-tail"
 # test big oc tail processing in brgemm kernel
 --reset
 --dir=BWD_D --dt=f16:f16:f16 mb1ic1iw2oc1387ow2kw2pw0_n"brgemm_big_oc_tail"
+
+# Test chained post-ops on brgemm kernels
+# See github.com/uxlfoundation/oneDNN issue #4055, and #4089
+--reset
+--impl=brg --dt=bf16 --attr-post-ops=relu+log
+g1mb1ic32ih128iw128oc32oh128ow128kh2kw2sh1sw1ph0pw0dh0dw0_n"brgemm_chained_post_ops"
```
dzarukin: Could you, please, form a single line for a case?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Sqvid: Yes, we have made a not to clean up the `log` injector. It is the only one currently relying on querying the `CodeArray` address manually like this.
https://github.com/uxlfoundation/oneDNN/pull/4063#pullrequestreview-3310156311

tests/benchdnn/inputs/conv/harness_conv_regression_general
Sqvid: Done.

PR #4064, opened 2025-10-03
cpu: aarch64: remove d_type compiletime arg from JIT

keanucz:
This removes the compiletime d_type arg from the eltwise kernel and fwd/bwd layer code and instead derives it from the src/dest memory descriptors instead.

### DIFF
--- a/src/cpu/aarch64/jit_uni_eltwise.cpp
+++ b/src/cpu/aarch64/jit_uni_eltwise.cpp
@@ -260,7 +260,7 @@ private:
 
     /**
      * @brief Unpack two BFloat16 values from a single register into two target registers.
-    * 
+    *
     * ASIMD (NEON) version: unpack in low/high half order.
     * SVE version: unpack in even/odd element order.
     *
@@ -271,7 +271,7 @@ private:
 
     /**
      * @brief Pack two BFloat16 values from two registers into a single register.
-     * 
+     *
      * ASIMD (NEON) version: pack from low/high half order.
      * SVE version: pack from even/odd element order.
      *
@@ -282,7 +282,7 @@ private:
 
     /**
      * @brief Unpack two IEEE‑754 FP16 values from a single register into two target registers.
-     * 
+     *
      * ASIMD (NEON) version: unpack in low/high half order.
      * SVE version: unpack in even/odd element order.
      *
@@ -293,7 +293,7 @@ private:
 
     /**
      * @brief Pack two IEEE‑754 FP16 values from two registers into a single register.
-     * 
+     *
      * ASIMD (NEON) version: pack from low/high half order.
      * SVE version: pack from even/odd element order.
      *
@@ -396,15 +396,15 @@ inline void jit_uni_kernel_t<cpu_isa_t::sve_128>::pack_fp16(
 }
 } // namespace
 
-template <cpu_isa_t isa, data_type_t d_type>
-status_t jit_uni_eltwise_fwd_t<isa, d_type>::pd_t::init(engine_t *engine) {
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_fwd_t<isa>::pd_t::init(engine_t *engine) {
     using namespace alg_kind;
 
     const memory_desc_wrapper src_d(src_md());
 
     bool ok = mayiuse(isa) && is_fwd()
             && utils::everyone_is(
-                    d_type, src_md()->data_type, dst_md()->data_type)
+                    src_md()->data_type, dst_md()->data_type)
             && !has_zero_dim_memory() && src_d.is_dense(true)
             && eltwise_injector::is_supported(isa, desc_.alg_kind)
             // refer to a comment in jit_uni_kernel why this is needed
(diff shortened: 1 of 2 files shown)


src/cpu/cpu_eltwise_list.cpp
```diff
@@ -66,12 +66,12 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
...
-            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<sve_128, f16>)
-            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd, f32>)
-            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd, bf16>)
-            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd, f16>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<sve_128>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<sve_128>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<sve_128>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd>)
+            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd>)
```
dzarukin, line 75:
```suggestion
            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<sve_128>)
            CPU_INSTANCE_AARCH64(jit_uni_eltwise_fwd_t<asimd>)
```

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -480,22 +484,23 @@ status_t jit_uni_eltwise_bwd_t<isa, d_type>::pd_t::init(engine_t *engine) {
...
     CHECK(safe_ptr_assign(kernel_, new jit_uni_kernel_t<isa>(pd())));
     return kernel_->create_kernel();
 }
 
-template <cpu_isa_t isa, data_type_t d_type>
-status_t jit_uni_eltwise_bwd_t<isa, d_type>::execute(
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_bwd_t<isa>::execute(
         const exec_ctx_t &ctx) const {
+    const data_type_t data_t = pd()->use_dst() ? pd()->dst_md()->data_type : pd()->src_md()->data_type;
```
dzarukin: This is a wrong use. This should be removed, pointers should be used as `char` or `void` with later cast to char to perform pointer arithmetic to shift them.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -414,24 +414,28 @@ status_t jit_uni_eltwise_fwd_t<isa, d_type>::pd_t::init(engine_t *engine) {
...
-template <cpu_isa_t isa, data_type_t d_type>
-status_t jit_uni_eltwise_fwd_t<isa, d_type>::execute(
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_fwd_t<isa>::execute(
         const exec_ctx_t &ctx) const {
-    auto src = CTX_IN_MEM(const data_t *, DNNL_ARG_SRC);
-    auto dst = CTX_OUT_MEM(data_t *, DNNL_ARG_DST);
+
+    const data_type_t src_dt = pd()->src_md()->data_type;
+    const data_type_t dst_dt = pd()->dst_md()->data_type;
```
dzarukin: Ditto.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -396,15 +396,15 @@ inline void jit_uni_kernel_t<cpu_isa_t::sve_128>::pack_fp16(
...
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_fwd_t<isa>::pd_t::init(engine_t *engine) {
     using namespace alg_kind;
 
     const memory_desc_wrapper src_d(src_md());
 
     bool ok = mayiuse(isa) && is_fwd()
             && utils::everyone_is(
-                    d_type, src_md()->data_type, dst_md()->data_type)
+                    src_md()->data_type, dst_md()->data_type)
```
dzarukin:
```suggestion
                    src_md()->data_type == dst_md()->data_type
```

src/cpu/aarch64/jit_uni_eltwise.cpp
Sqvid: I was suggesting `uint8_t` over `char`. I don't think it makes a practical technical difference but let us know if there is a preference for consistency.

dzarukin: I personally use `void` to make conscious decisions about further casting if/when it's needed, but it doesn't have to be like that. I'm fine with either type as long as this practice stretches to all implementations for the sake of consistency.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -421,49 +424,72 @@ inline void jit_uni_kernel_t<cpu_isa_t::sve_128>::pack_fp16(
...
-    auto dst = CTX_OUT_MEM(data_t *, DNNL_ARG_DST);
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_fwd_t<isa>::execute(const exec_ctx_t &ctx) const {
+
+    const data_type_t src_dt = pd()->src_md()->data_type;
+
+    const size_t elem_size = types::data_type_size(src_dt);
+
+    // data_type_size returns 0 for data_type::undef
+    if (elem_size == 0) return status::unimplemented;
```
Sqvid: The `execute()` function is too late to return `status::unimplemented`. 
Looking at the definition of `types::data_type_size`, I am not even sure `0` is a possible value.

If want to prevent unsupported data_types then just make sure you have all possibilities covered in `pd_t::init()` using your `VDISPATCH_ELTWISE` checks.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -421,49 +424,72 @@ inline void jit_uni_kernel_t<cpu_isa_t::sve_128>::pack_fp16(
...
 
-template <cpu_isa_t isa, data_type_t d_type>
-status_t jit_uni_eltwise_fwd_t<isa, d_type>::execute(
-        const exec_ctx_t &ctx) const {
-    auto src = CTX_IN_MEM(const data_t *, DNNL_ARG_SRC);
-    auto dst = CTX_OUT_MEM(data_t *, DNNL_ARG_DST);
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_fwd_t<isa>::execute(const exec_ctx_t &ctx) const {
+
+    const data_type_t src_dt = pd()->src_md()->data_type;
```
Sqvid: nit: readability: I'd probably move this further down, closer to where you first actually make use of it.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -484,56 +510,81 @@ status_t jit_uni_eltwise_fwd_t<isa, d_type>::execute(
...
-template <cpu_isa_t isa, data_type_t d_type>
-status_t jit_uni_eltwise_bwd_t<isa, d_type>::execute(
-        const exec_ctx_t &ctx) const {
-    auto src = pd()->use_dst() ? CTX_IN_MEM(const data_t *, DNNL_ARG_DST)
-                               : CTX_IN_MEM(const data_t *, DNNL_ARG_SRC);
-    auto diff_dst = CTX_IN_MEM(const data_t *, DNNL_ARG_DIFF_DST);
-    auto diff_src = CTX_OUT_MEM(data_t *, DNNL_ARG_DIFF_SRC);
+template <cpu_isa_t isa>
+status_t jit_uni_eltwise_bwd_t<isa>::execute(const exec_ctx_t &ctx) const {
+    const data_type_t data_dt = pd()->use_dst() ? pd()->dst_md()->data_type
```
Sqvid: ditto, move closer to first use.

Sqvid (changes requested):

Sqvid (approved):
LGTM, thanks for the patch

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -19,11 +19,14 @@
 #include "common/bfloat16.hpp"
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
+#include "common/eltwise_pd.hpp"
```
dzarukin: This shouldn't be needed, it's included inside `#include "cpu/cpu_eltwise_pd.hpp"` and that is included in a header file
`jit_uni_eltwise.hpp`.

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -19,11 +19,14 @@
...
 #include "common/nstl.hpp"
+#include "common/type_helpers.hpp"
 #include "common/utils.hpp"
 
 #include "cpu/aarch64/jit_generator.hpp"
 
+#include <cstdint>
```
dzarukin: Usually system headers have a different grouping, preferably at the bottom of all other headers included.

PR #4066, opened 2025-10-03
ci: only run clang-format on changed files

Sqvid:
# Description

Avoids checking the entire repository every time a change is made. The old behavior is still preserved for anyone relying on it but I have made it faster by using `find -exec` rather than a `for` loop.

### DIFF
--- a/.github/automation/clang-format.sh
+++ b/.github/automation/clang-format.sh
@@ -1,7 +1,8 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
 #===============================================================================
 # Copyright 2019-2020 Intel Corporation
+# Copyright 2025 Arm Ltd. and affiliates
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -26,7 +27,18 @@ fi
 
 echo "Starting format check..."
 
-for filename in $(find "$(pwd)" -type f | grep -P ".*\.(c|cpp|h|hpp|cl)$"); do ${CLANG_FORMAT} -style=file -i $filename; done
+src_regex='.*\.(c|h|cpp|hpp|cxx|hxx|cl)$'
+
+# Treat each argument as an input file. If called with no arguments check the
+# whole repo.
+if [[ $# -gt 0 ]]; then
+    base_sha=$1
+    file_list=$(git diff --name-only $base_sha | grep -E "$src_regex")
+    echo "Checking: $file_list"
+    for filename in $file_list; do ${CLANG_FORMAT} -style=file -i $filename; done
+else
+    find "$(pwd)" -type f -regextype posix-egrep -regex "$src_regex" -exec "$CLANG_FORMAT" -style=file -i {} \+
+fi
 
 RETURN_CODE=0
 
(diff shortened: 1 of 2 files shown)


.github/automation/clang-format.sh
```diff
@@ -1,7 +1,8 @@
-#! /bin/bash
+#! /usr/bin/env bash
```
vpirogov: Space after shebang must go. I noticed earlier that we have it in a lot of places.

```suggestion
#!/usr/bin/env bash
```

.github/automation/clang-format.sh
```diff
@@ -16,7 +17,7 @@
 # limitations under the License.
 #===============================================================================
 
-CLANG_FORMAT=clang-format-11
+CLANG_FORMAT="${CLANG_FORMAT:-"clang-format-11"}"
```
vpirogov: The plan here was to use clang-format-11 unconditionally for consistency. Why do we want to take a value from the environment?

.github/automation/clang-format.sh
vpirogov: Actually it is allowed...

.github/automation/clang-format.sh
Sqvid: Yes you're right it probably doesn't make too much sense here. Removed it.

.github/automation/clang-format.sh
Sqvid: I prefer it without the space as well, so I've removed it now.

.github/automation/clang-format.sh
```diff
@@ -26,7 +27,18 @@ fi
 
 echo "Starting format check..."
 
-for filename in $(find "$(pwd)" -type f | grep -P ".*\.(c|cpp|h|hpp|cl)$"); do ${CLANG_FORMAT} -style=file -i $filename; done
+src_regex='.*\.(c|h|cpp|hpp|cxx|hxx|cl)$'
```
Sqvid, line 30: The patch is ready for review. One thing to note is that I have added `.hxx|.cxx` to the regex pattern. This means some [files which were not checked before](https://github.com/uxlfoundation/oneDNN/blob/c8cd40f2502b99092f8d0dd60412516b816518a8/src/gpu/intel/gemm/jit/generator/pieces/atomic_fusions.cxx), might be now.

It might be a good idea to standardise `C` extensions to `.c|.h` and `C++` to `.cpp|.hpp` to reduce complexity for workflows like this, `clang-tidy`, etc.

atkassen: src/gpu/intel/gemm/jit/generator has [its own .clang-format](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/gemm/jit/generator/.clang-format) which simply disables formatting.

Sqvid: @atkassen, ah that makes sense, thanks. Is there any reason for the different file extensions? Also let me know if you would like `.cxx|.hxx` removed from the format pattern.

atkassen: It looks like the .cxx files are used like headers, so they are not separate compilation units. I don't know exactly what distinguishes them from .hxx files. I think it's fine to keep them in the pattern in any case.

Sqvid:
@vpirogov are you happy with the changes?

PR #4075, opened 2025-10-06
ci: aarch64: Fix git bisect false positives and wiki failure

Ryo-not-rio:
- add a fix for the git bisect conflicting with the wiki causing the wiki update to fail
- add a fix for git bisect picking up on non-failing tests due to output truncation
- fix edge case where success status hash was being updated when one of the ci matrices succeeded

### DIFF
--- a/.github/automation/aarch64/test.sh
+++ b/.github/automation/aarch64/test.sh
@@ -22,6 +22,8 @@
 set -o errexit -o pipefail -o noclobber
 
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+OUTPUT_XML=$1
+OUTPUT_LOG=$2
 
 # Defines MP, CC, CXX and OS.
 source ${SCRIPT_DIR}/common.sh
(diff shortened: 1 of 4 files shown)


.github/automation/aarch64/test.sh
```diff
@@ -32,5 +32,5 @@ if [[ "$ONEDNN_THREADING" == "SEQ" ]]; then
 fi
 
 set -x
-ctest --no-tests=error --output-on-failure -E $("${SCRIPT_DIR}"/skipped-tests.sh) --output-junit $1
+ctest --no-tests=error --output-on-failure -E $("${SCRIPT_DIR}"/skipped-tests.sh) --output-junit $1 -O $2; cat $2
```
Sqvid: Assign `$1`, and `$2` to a well-named variables so it is obvious what they contain.

Why do you cat `$2` at the end? Doesn't `ctest` write to `stdout` even if you provide `-O`?

Ryo-not-rio: Done, if you look at the output here: https://github.com/uxlfoundation/oneDNN/actions/runs/18282199126/job/52048351701 you can see that there's only one output, using `cat`

Sqvid: Not sure why the behavior seems different on my system but is the content of the file passed to `-O <file>` supposed to be the same as what would normally be printed to `stdout`? In other words are your trying to redirect the output to a file and print it? In which case you probably just want to [pipe it into `tee <file>`](https://man7.org/linux/man-pages/man1/tee.1.html).

Ryo-not-rio: done

Sqvid (approved):

PR #4077, opened 2025-10-06
cpu: aarch64: skip fp16 to fp32 upcast for simple eltwise JIT ops

Anndrey24:
# Description

This commit allows certain simple FP16 elementwise ops to run directly in FP16, where their results are numerically consistent across floating-point formats, thus bypassing the current upcast/downcast to FP32. These operations are ABS, non-leaky ReLU, SQUARE, and SQRT.

Additionally, for ABS and non-leaky ReLU on BF16 inputs, we skip the upcast by reinterpreting the inputs as FP16, since these ops depend only on the sign bit, which is in the same position in both encodings.

### Benchdnn performance results for FP16:
- ABS:
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=abs --alpha=0 --beta=0 --dt=f16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):1.93433 avg(ms):1.95283`
	- JIT direct in fp16 (new): `total perf: min(ms):0.248291 avg(ms):0.258242`
	- Speedup: **≈7.56x**
- ReLU (alpha = 0):
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=relu --alpha=0 --beta=0 --dt=f16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):1.93701 avg(ms):1.95328`
	- JIT direct in fp16 (new): `total perf: min(ms):0.249023 avg(ms):0.258845`
	- Speedup: **≈7.55x**
- SQUARE:
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=square --alpha=0 --beta=0 --dt=f16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):2.04297 avg(ms):2.05354`
	- JIT direct in fp16 (new): `total perf: min(ms):0.251709 avg(ms):0.26124`
	- Speedup: **≈7.86x**
- SQRT:
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=sqrt --alpha=0 --beta=0 --dt=f16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):2.54736 avg(ms):2.56354`
	- JIT direct in fp16 (new): `total perf: min(ms):1.02441 avg(ms):1.03189`
	- Speedup: **≈2.48x**
### Benchdnn performance results for BF16:
- ABS:
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=abs --alpha=0 --beta=0 --dt=bf16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):0.818848 avg(ms):0.900838`
	- JIT direct in fp16 (new): `total perf: min(ms):0.249512 avg(ms):0.25974`
	- Speedup: **≈3.47x**
- ReLU (alpha = 0):
	- Command: `OMP_NUM_THREADS=1 ./benchdnn --engine=cpu --mode=p --eltwise --alg=relu --alpha=0 --beta=0 --dt=bf16 1024x4096`
	- JIT upcast to fp32 (old): `total perf: min(ms):0.818359 avg(ms):0.907584`
	- JIT direct in fp16 (new): `total perf: min(ms):0.253174 avg(ms):0.272422`
	- Speedup: **≈3.33x**


src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -58,6 +58,28 @@ struct jit_uni_eltwise_kernel_t : public jit_generator_t {
     bool is_bf16() const { return data_type() == data_type::bf16; }
     bool is_f16() const { return data_type() == data_type::f16; }
     int dtype_size() const { return types::data_type_size(data_type()); }
+    // ABS and non-leaky RELU depend only on sign bit, so FP16 reinterpretation of BF16 value is safe
+    bool is_bf16_compatible_alg() const {
```
jondea: Given that they are only called once, I'm not sure we add much value putting these into separate functions. One code smell here is that `is_bf16_compatible_alg` isn't clear, it would probably be clearer as something like `can_compute_bf16_as_f16`, but this is complicated and unlikely to be needed anywehere else, implying that the logic should probably just live in `is_f16_compatible_alg`

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -111,7 +134,11 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
         // perspective and will complicate the compute logic significantly.
 
         load_vector(vmm_src.s, reg_src);
-        if (is_bf16()) {
+        if (is_f16_compute_safe()) {
+            // For FP16-compatible algorithms, we can keep the data in 16-bit format
```
jondea: nit: try to be consistent with the words used for the same technical concept, here we have
- `f16_compute_safe`
- FP16-compatible algorithms

src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -58,6 +58,28 @@ struct jit_uni_eltwise_kernel_t : public jit_generator_t {
...
+    bool is_f16_compatible_alg() const {
+        const auto &desc = *pd_->desc();
+        return pd_->is_fwd()
+                && ((desc.alg_kind == alg_kind::eltwise_relu
+                            && desc.alpha == 0.0f)
+                        || desc.alg_kind == alg_kind::eltwise_abs
+                        || desc.alg_kind == alg_kind::eltwise_square
+                        || desc.alg_kind == alg_kind::eltwise_sqrt);
+    }
+    bool is_f16_compute_safe() const {
```
jondea: This feels a bit wooly, `safe` isn't usually used with respect to numerics, and I think we are glossing over some complexity wrt `bf16`. Something like `can_compute_as_f16` would make the casting/reintepretting of bf16 clear.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
```diff
@@ -215,12 +222,16 @@ struct jit_uni_eltwise_injector_f32_t {
     void exp_compute_vector_fwd(const TRegS &vmm_src);
     void relu_compute_vector_fwd(const TRegS &vmm_src);
     void relu_zero_ns_compute_vector_fwd(const TRegS &vmm_src);
+    void relu_zero_ns_compute_vector_fwd(const TRegH &vmm_src);
     void elu_compute_vector_fwd(const TRegS &vmm_src);
     void tanh_compute_vector_fwd(const TRegS &vmm_src);
     void tanh_polynomial_approx_compute_vector_fwd(const TRegS &vmm_src);
     void square_compute_vector_fwd(const TRegS &vmm_src);
+    void square_compute_vector_fwd(const TRegH &vmm_src);
```
jondea: I like the way you have used overloading here, although if we had lots more of these, we could consider using templates, because the logic is identical

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -1600,114 +1624,142 @@ size_t jit_uni_eltwise_injector_f32_t<isa>::aux_vecs_count() {
...
-                    break;
-                case eltwise_hardsigmoid:
-                    hardsigmoid_compute_vector_bwd(TRegS(idx));
-                    break;
-                default: assert(!"unsupported eltwise algorithm");
-            }
-        }
+        };
+
+        // Pick the right register type (h/s) according to the data type
```
jondea: This is a lot of logic to wrap for just a few exceptions. Can we do the if inside the switch statement? Or change the functions to accept generic `TReg` and do the datatype switching inside the `*_compute_vector_*` functions? I think the second option would more easily allow for extending this concept for the more complicated functions, e.g. exp or tanh

jondea (changes requested):
This is excellent, thank you! I particularly appreciate the detailed justification in the PR summary. My comments are just minor code quality improvements.

src/cpu/aarch64/jit_uni_eltwise.cpp
Anndrey24: Agreed! I have now moved all this logic inside a single `can_compute_as_f16()` function.

src/cpu/aarch64/jit_uni_eltwise.cpp
Anndrey24: Good point. I have changed the immediately surrounding comments to use the F16/F32 notation too.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: That's definitely a much more elegant solution! I went with the second version you suggested.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
Anndrey24: I'd be happy to make `d_type` a template parameter instead of a member variable, to split up the dtype-related logic that's now inside the single `*_compute_vector_*` functions. However, with these "F16-native" implementations being just fortunate edge cases, I doubt we'll have many others added in the future anyway, unless F16 becomes accepted for accumulation/intermediary results.

jondea (approved):
Great stuff, LGTM

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
jondea: Makes sense, I like the way it is now!

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2411,24 +2423,29 @@ void jit_uni_eltwise_injector_f32_t<isa>::register_table_entries() {
...
 template <>
-void jit_uni_eltwise_injector_f32_t<asimd>::relu_zero_ns_compute_vector_fwd(
-        const TRegS &vmm_src) {
-    h->movi(vmm_aux0, 0);
-    h->fmaxnm(vmm_src, vmm_src, vmm_aux0);
+void jit_uni_eltwise_injector_t<asimd>::relu_zero_ns_compute_vector_fwd(
+        const TReg &vmm_src) {
+    if (d_type_ == data_type::f16) {
+        h->movi(TRegH(IDX(vmm_aux0)), 0);
+        h->fmaxnm(vmm_src.h, vmm_src.h, TRegH(IDX(vmm_aux0)));
```
aditew01, line 2440:
```suggestion
        h->movi(vmm_aux0.h, 0);
        h->fmaxnm(vmm_src.h, vmm_src.h, vmm_aux0.h);
```
Do you think we could simplify it to this, or would this break?

aditew01 (approved):
Left a minor comment. 
Thanks for the changes, refactoringand the verbose comments.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I believe this only works on TReg (i.e. VReg or ZReg) objects, whereas all the `vmm_aux*` variables were declared as TRegS unfortunately. 

I could create a `vmm_aux_h` variable instead if you think that's clearer than the current version:
```suggestion
        const auto &vmm_aux_h = TRegH(IDX(vmm_aux1));
        h->movi(vmm_aux_h, 0);
        h->fmaxnm(vmm_src.h, vmm_src.h, vmm_aux_h);
```

aditew01: Thanks for the clarification. The refactoring can go as a separate PR.

PR #4078, opened 2025-10-06
doc: restructure examples doc page

mzhukova:
Examples & Tutorials page restructuring to include more details, group by category and also include all the individual primitives samples.
Also, this PR should take care of [DOC-13375](https://jira.devtools.intel.com/browse/DOC-13375).

Updated examples doxygen documentation to follow the next structure (so that the examples description could be re-used and not duplicated):

This produces separate page w/a annotations, but with just code (existed before):
```
/// @example convolution.cpp 
/// > Annotated version: @ref convolution_example_cpp 
```

This produces page with a short description, that is only created so that it could be referenced from different documents (**new**):
```
/// @page convolution_example_cpp_brief
/// @brief This C++ API example demonstrates how to create and execute a
/// [Convolution](@ref dev_guide_convolution) primitive in forward propagation
/// mode in two configurations - with and without groups.
```
Since this produces new pages, the update to `orphans.rst` is needed.

This produces the main page that should be referenced in `examples.md`, etc. (existed, but changed to use `\copybrief'):
```
/// @page convolution_example_cpp Convolution Primitive Example
/// \copybrief convolution_example_cpp_brief
///
/// Key optimizations included in this example:
```

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -1,39 +1,176 @@
-Examples and Tutorials {#dev_guide_examples}
-============================================
+# Examples and Tutorials {#dev_guide_examples}
 
-Functional API
---------------
+This page provides an overview of oneDNN examples organized by functionality and use case.
 
-| Topic           | Engine  | C++ API                                | C API                       |
-|:----------------|:--------|:---------------------------------------|:----------------------------|
-| Tutorials       | CPU/GPU | @ref getting_started_cpp               |                             |
-|                 | CPU/GPU | @ref memory_format_propagation_cpp     |                             |
-|                 | CPU/GPU | @ref cross_engine_reorder_cpp          | @ref cross_engine_reorder_c |
-|                 | CPU/GPU | @ref sycl_interop_buffer_cpp           |                             |
-|                 | GPU     | @ref gpu_opencl_interop_cpp            |                             |
-|                 | CPU/GPU | @ref bnorm_u8_via_binary_postops_cpp   |                             |
-|                 | CPU/GPU | @ref matmul_with_host_scalar_scale_cpp |                             |
-| Performance     | CPU/GPU | @ref performance_profiling_cpp         |                             |
-|                 | CPU/GPU | @ref matmul_perf_cpp                   |                             |
-|                 | CPU/GPU | @ref bnorm_u8_via_binary_postops_cpp   |                             |
-| f32 inference   | CPU/GPU | @ref cnn_inference_f32_cpp             | @ref cnn_inference_f32_c    |
-|                 | CPU     | @ref cpu_rnn_inference_f32_cpp         |                             |
-| int8 inference  | CPU/GPU | @ref cnn_inference_int8_cpp            |                             |
-|                 | CPU     | @ref cpu_rnn_inference_int8_cpp        |                             |
-| f32 training    | CPU/GPU | @ref cnn_training_f32_cpp              |                             |
-|                 | CPU     |                                        | @ref cpu_cnn_training_f32_c |
-|                 | CPU/GPU | @ref rnn_training_f32_cpp              |                             |
-| bf16 training   | CPU/GPU | @ref cnn_training_bf16_cpp             |                             |
-| f8 quantization | CPU/GPU | @ref matmul_f8_quantization_cpp        |                             |
+## Functional API Examples
 
-Graph API
----------
+The Functional API provides access to individual oneDNN primitives.
 
-| Topic          | Engine  | Example Name                              |
-|:---------------|:--------|:------------------------------------------|
-| Tutorials      | CPU     | @ref graph_cpu_getting_started_cpp        |
-|                | CPU     | @ref graph_cpu_inference_int8_cpp         |
-|                | CPU/GPU | @ref graph_sycl_getting_started_cpp       |
-|                | CPU     | @ref graph_cpu_single_op_partition_cpp    |
-|                | GPU     | @ref graph_sycl_single_op_partition_cpp   |
-|                | GPU     | @ref graph_gpu_opencl_getting_started_cpp |
+### Fundamental Concepts and API Basics
+
+| Example | Description |
+|:--------|:------------|
+| @ref getting_started_cpp | \copybrief getting_started_cpp_brief |
+| @ref memory_format_propagation_cpp | \copybrief memory_format_propagation_cpp_brief |
+| @ref cross_engine_reorder_cpp | \copybrief cross_engine_reorder_cpp_brief |
+
+### Interoperability with External Runtimes
+
+| Example | Description |
+|:--------|:------------|
+| @ref sycl_interop_buffer_cpp | \copybrief sycl_interop_buffer_cpp_brief |
+| @ref sycl_interop_usm_cpp | \copybrief sycl_interop_usm_cpp_brief |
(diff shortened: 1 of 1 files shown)


mzhukova:
make test
disable test_device_cpu
disable test_device_gpu
enable run_documentation

doc/examples.md
```diff
@@ -1,39 +1,139 @@
...
+   - [Fundamental Concepts and API Basics](#fundamental-concepts-and-api-basics)
+   - [Interoperability with External Runtimes](#interoperability-with-external-runtimes)
+   - [Matrix Multiplication with Different oneDNN Features](#matrix-multiplication-with-different-onednn-features)
+   - [Inference and Training](#inference-and-training)
+   - [Recurrent Neural Networks](#recurrent-neural-networks)
+   - [Performance Analysis](#performance-analysis)
+   - [Microkernel Interface (uKernel)](#microkernel-interface-ukernel)
+   - [Individual Primitives](#individual-primitives)
+   - [C API Examples](#c-api-examples)
+- [Graph API Examples](#graph-api-examples)
```
vpirogov: Sphinx will show the table of contents automatically in the right pane.

doc/examples.md
```diff
@@ -1,39 +1,139 @@
...
+
+### Performance Analysis
+
+A few techniques for performance measurements:
+- @ref performance_profiling_cpp - Performance analysis based on `ONEDNN_VERBOSE` usage
+- @ref matmul_perf_cpp - Simple Matrix multiplication performance testing
+
+### Microkernel Interface (uKernel)
+
+- @ref cpu_brgemm_example_cpp - Batch-Reduced GEMM (BRGEMM) microkernel implementation on CPU
```
vpirogov: Maybe move this into it's own top level section?

doc/examples.md
mzhukova: Sure, I could remove to not duplicate

doc/examples.md
mzhukova: Would something like that be better? (so that ukernel is on the top level but with extra words to signify that it is still Functional API formally):
- Functional API
  - main bulk of examples
- Functional API: Microkernel Interface
- Functional API: C examples
- Graph API

mzhukova: ping @vpirogov :)

doc/examples.md
```diff
@@ -1,39 +1,127 @@
...
-|                | GPU     | @ref graph_gpu_opencl_getting_started_cpp |
+This page provides an overview of oneDNN examples organized by functionality and use case.
+
+## Functional API Examples
+
+The Functional API provides access to individual oneDNN primitives.
+
+### Fundamental Concepts and API Basics
+
+- @ref getting_started_cpp - Basic API workflow and primitive execution
```
avmanerikar: (Random spot) (nit): 
Any particular reason for not using a tabular representation? It could still be useful to have the description and engine placed in separate columns.

doc/examples.md
```diff
@@ -1,39 +1,127 @@
...
+- @ref vanilla_rnn_example_cpp - Basic RNN: Fundamental recurrent cell implementation
+- @ref lstm_example_cpp - LSTM: Long Short-Term Memory with gating mechanisms
+- @ref lbr_gru_example_cpp - LBR-GRU: Linear-Before-Reset Gated Recurrent Unit
+- @ref augru_example_cpp - AU-GRU: Attention-Update-Gate RNN variant
+
+### Performance Analysis
+
+A few techniques for performance measurements:
+- @ref performance_profiling_cpp - Performance analysis based on `ONEDNN_VERBOSE` usage
+- @ref matmul_perf_cpp - Simple Matrix multiplication performance testing
```
avmanerikar: (nit): Maybe switch the order, placing the simple test case before the one using oneDNN trace output?

doc/examples.md
```diff
@@ -1,39 +1,127 @@
...
+- @ref layer_normalization_example_cpp - Layer normalization
+- @ref lrn_example_cpp - Local Response Normalization
+
+Activation Functions:
+- @ref eltwise_example_cpp - Element-wise operations (ReLU)
+- @ref prelu_example_cpp - ReLU activation
+- @ref softmax_example_cpp - Softmax activation
+
+Tensor Operations:
+- @ref binary_example_cpp - Binary operations (add, multiply, etc.)
```
avmanerikar: (nit): Not sure if it is required, but it might be useful to specify if the example uses post-ops (like with the binary primitive).

doc/examples.md
mzhukova: Mainly for readability.
- A huge bulk of examples is available / make sense for both engines, so engine info is kind of redundant. I explicitly mentioned engine in "description" if it is needed.
- Also, tabular form making it harder to manage, as you could end up with long lines that would mess it up..

doc/examples.md
mzhukova: Sure, I could do that :)

doc/examples.md
mzhukova: I'll take a look and try to add more details as you suggested!

doc/examples.md
vpirogov: This structure makes sense. One small thing is that we are positioning microkernel API as an entity as the same level as Functional and Graph.

mzhukova: Sure, updated

doc/examples.md
```diff
@@ -1,39 +1,127 @@
...
+Memory Transformations:
+- @ref reorder_example_cpp - Memory layout transformations
+
+## Functional API: C API Examples
+
+- @ref cross_engine_reorder_c - Cross-engine data movement
+- @ref cnn_inference_f32_c - CNN inference implementation
+- @ref cpu_cnn_training_f32_c - CNN training with gradients on CPU
+
+## Microkernel Interface (uKernel)
```
vpirogov: This is a niche feature, so I'd suggest moving it to the end of the list.

```suggestion
## Microkernel (uKernel) API Examples
```

doc/examples.md
```diff
@@ -1,39 +1,127 @@
...
-|:---------------|:--------|:------------------------------------------|
-| Tutorials      | CPU     | @ref graph_cpu_getting_started_cpp        |
-|                | CPU     | @ref graph_cpu_inference_int8_cpp         |
-|                | CPU/GPU | @ref graph_sycl_getting_started_cpp       |
-|                | CPU     | @ref graph_cpu_single_op_partition_cpp    |
-|                | GPU     | @ref graph_sycl_single_op_partition_cpp   |
-|                | GPU     | @ref graph_gpu_opencl_getting_started_cpp |
+This page provides an overview of oneDNN examples organized by functionality and use case.
+
+## Functional API Examples
```
vpirogov, line 5: Shouldn't top level headers start with `#` not `##`?

mzhukova: I think (plus based on rendered view) the top `Examples and Tutorials` is considered having `#`.

vpirogov: Looks like our README.md is formatted incorrectly :)

mzhukova:
make test
disable test_device_cpu
disable test_device_gpu
enable run_documentation

doc/examples.md
mzhukova: I've updated to re-use brief description that already exists in examples, so that we should only update one place.

doc/examples.md
mzhukova: Fixed

PR #4082, opened 2025-10-07
tests: fix clang-tidy failures

Sqvid:
# Description

Looks like some test files might have been previously missed when making the library clang-tidy compliant. This PR fixes the failures I am aware of.

# Checklist

## General


Sqvid:
cc: @vpirogov

Sqvid:
> Thanks for taking care of those. They were not covered because those files are not used in CI.

Thanks for the explanation, I thought the `-p build` option we pass would have caught these. I discovered them because I added `-DCMAKE_CXX_CLANG_TIDY=clang-tidy-18` to my local build options so that I could lint commits before submitting them.

PR #4085, opened 2025-10-07
benchdnn: ref_post_op: eltwise: fix NAN propagation

dzarukin:
Fixes [MFDNN-14239](https://jira.devtools.intel.com/browse/MFDNN-14239).

### DIFF
--- a/tests/benchdnn/dnn_types.cpp
+++ b/tests/benchdnn/dnn_types.cpp
@@ -1723,8 +1723,14 @@ void maybe_zero_point(const attr_t &attr, float &d, const int32_t *zero_points,
 }
 
 float compute_eltwise_fwd(pk_t kind, float src, float alpha, float beta) {
-    // don't compute on nan, propagate it
-    if (std::isnan(src)) return NAN;
+    if (std::isnan(src)) {
+        // Don't compute if source is NAN, propagate it, unless a special case.
+
+        // pow(NAN, 0) = 1 according to IEEE Std 754-2008.
+        const bool is_pow0 = kind == pk_t::POW && beta == 0;
+        const bool propagate_nan = !is_pow0;
+        if (propagate_nan) { return NAN; }
+    }
 
     using namespace dnnl::impl::math;
 


tests/benchdnn/dnn_types.cpp
```diff
@@ -1723,8 +1723,14 @@ void maybe_zero_point(const attr_t &attr, float &d, const int32_t *zero_points,
 }
 
 float compute_eltwise_fwd(pk_t kind, float src, float alpha, float beta) {
-    // don't compute on nan, propagate it
-    if (std::isnan(src)) return NAN;
+    if (std::isnan(src)) {
```
mzhukova, line 1726: is it possible to get here to other edge cases: like `src` being `inf` or `beta` being `NAN`/`inf`?

also, does it make sense to do an early exit (like return 1) if `is_pow0` ..?

mzhukova (approved):

tests/benchdnn/dnn_types.cpp
dzarukin: The reason to handle NaNs in here is the behavior of NaN processing is implementation-defined (when it comes to `input < something` or `input > something`)  while should be pretty specific and benchdnn enforces that. Based on the algorithms supported, this is the first corner case met, since 2019.

If user asked for beta=NaN (which doesn't have any sense from an eltwise operation) they get what they get.

`inf` is handled properly in kernels and standard math functions, no need to handle it.

> does it make sense to do an early exit (like return 1) if is_pow0 ..?

This function just relays the input from benchdnn to the library's CPU implementation with the only exception of NaN handling, such branch should exist inside a `pow` function and it'll be reached within two extra calls.

Sqvid:
What is the intended numerical model with the the following? There similar cases that can "fail" (on `NaN` vs `0`) on both `x64`, and `AArch64`:

```
create: --conv --attr-post-ops=log+relu g1ic1ih2oc1oh2kh1ph0
oneDNN implementation: jit_dw:avx512_core
[COMPARE][DST]: zero_trust%=85.00% extra=
[CHECK_MEM]: zmalloc (0x5f3e73563fe0) of size 32 B, CPU total alloc: 1.167969 KB
[   0][DST][0:0:0:0] exp_f32:      4.8752 exp:      4.8752 got:      4.8752 diff:1.43051e-06 rdiff:2.93426e-07
[   1][DST][0:0:0:1] exp_f32:      6.1203 exp:      6.1203 got:      6.1203 diff:9.53674e-07 rdiff:1.55822e-07
[   2][DST][0:0:1:0] exp_f32:     6.30079 exp:     6.30079 got:     6.30078 diff:3.33786e-06 rdiff:5.29753e-07
[   3][DST][0:0:1:1] exp_f32:         nan exp:         nan got:           0 diff:     nan rdiff:     nan
[   4][DST][1:0:0:0] exp_f32:     3.13549 exp:     3.13549 got:     3.13549 diff:       0 rdiff:       0
[   5][DST][1:0:0:1] exp_f32:         nan exp:         nan got:           0 diff:     nan rdiff:     nan
[   6][DST][1:0:1:0] exp_f32:         nan exp:         nan got:           0 diff:     nan rdiff:     nan
[   7][DST][1:0:1:1] exp_f32:     5.68017 exp:     5.68017 got:     5.68017 diff:2.86102e-06 rdiff:5.03686e-07
[COMPARE_STATS][DST]: trh=0 err_max_diff:     nan err_max_rdiff:     nan all_max_diff:3.33786e-06 all_max_rdiff:5.29753e-07
[COMPARE_TRUST][DST]: z:38% (>85%) (z: 3, total: 8)
0:FAILED (errors:3 total:8) (7 ms) __REPRO: --conv --attr-post-ops=log+relu g1ic1ih2oc1oh2kh1ph0
| jit_dw:avx512_core : 1 (100%)                            |
0:FAILED (errors:3 total:8) (7 ms) __REPRO: --conv --attr-post-ops=log+relu g1ic1ih2oc1oh2kh1ph0
============================
total: 0.01s; create_pd: 0.00s (7%); create_prim: 0.00s (6%); fill: 0.00s (10%); execute: 0.00s (56%); compute_ref: 0.00s (1%); compare: 0.00s (11%);
```

mgouicem:
@Sqvid oneDNN does not guarantee proper NaN propagation see [this documentation note](https://uxlfoundation.github.io/oneDNN/dev_guide_c_and_cpp_apis.html#input-validation-notes).
Even if we have limited validation of NaN propagation, we still do try to propagate NaN properly. But as you found out, we have common "optimizations" like relu that return `0` when `alpha=0`, instead of computing `alpha*src`.

In the case of this PR, it is not fixing library behavior, but instead modifies testing to avoid false-positives.

As a side note, strict NaN propagation is a requirement only for low-precision training where NaNs do arise and trigger rescaling. For inference, either saturation is used for low-precision computation, or trained parameters tend to avoid NaN apparition.

Sqvid:
@mgouicem Thanks for the clarification, much appreciated.

tests/benchdnn/dnn_types.cpp
mgouicem: @dzarukin The fix seems fine for the issue at hand.
However, it is still not clear to me why we need the extra NaN check here.
Why not just let reference impls from math_utils act as numerical reference?
If that math_utils impls do not have the proper numerical behavior wrt our documentation, then we should fix those anyway (e.g. with a similar NaN check early exit).

> The reason to handle NaNs in here is the behavior of NaN processing is implementation-defined (when it comes to input < something or input > something)

Well this enforces a particular behavior that is currently not specified in [our documentation ](https://uxlfoundation.github.io/oneDNN/dev_guide_eltwise.html): technically, all comparisons there are false, and min/max operators is underspecified.

dzarukin: The reason for the check here is CPU/GPU had (probably still have) different implementation for some ops like clip, where a comparison like `s < a` versus `s >= b` could appear. Depending on a predicate the answer would be different since the comparison with NaN is always false. It leads to answer mismatch, thus, a hard check for NaN in benchdnn to be exact on NaN behavior.

The PR fixes a legit case when NaN shouldn't be reported as NaN by benchdnn.

Fixes towards following the documentation should be done separately and planned accordingly.

PR #4086, opened 2025-10-07
cpu: x64: correct initialization of comp_owb

xuxinzen:
Fixes MFDNN-14284, MFDNN-14286


xuxinzen:
make test

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -1187,18 +1187,26 @@ status_t brgemm_convolution_fwd_t<isa>::init(engine_t *engine) {
...
             const auto ow_e = nstl::min(ow_b + jcp.ow_block, OW);
+            const auto ow_l = ow_e - ow_b;
 
             if (owb == 0) {
                 comp_owb[owb] = 0;
                 continue;
             }
 
-            for_(int i = 0; i < comp_ow_l; i++)
-            for (int ow = ow_b; ow < ow_e; ow++) {
```
dzarukin, line 1197: Looks like this could be updated to `ow <= ow_e` and `if` condition below could be updated to not do `- 1` and that's it.

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -1187,18 +1187,26 @@ status_t brgemm_convolution_fwd_t<isa>::init(engine_t *engine) {
         for (int owb = 0; owb < jcp.nb_ow; owb++) {
             const auto ow_b = owb * jcp.ow_block;
             const auto ow_e = nstl::min(ow_b + jcp.ow_block, OW);
+            const auto ow_l = ow_e - ow_b;
```
dzarukin: This will trigger an unused variable in release mode.

xuxinzen:
make test
set test_scope=NIGHTLY

src/cpu/x64/jit_brgemm_conv.cpp
xuxinzen: thanks, updated.

xuxinzen:
make test

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -1187,19 +1187,23 @@ status_t brgemm_convolution_fwd_t<isa>::init(engine_t *engine) {
...
+            MAYBE_UNUSED(comp_ow_f);
 
             if (owb == 0) {
                 comp_owb[owb] = 0;
                 continue;
             }
 
-            for_(int i = 0; i < comp_ow_l; i++)
-            for (int ow = ow_b; ow < ow_e; ow++) {
+            for_(int i = 0; i < comp_ow_f; i++)
```
dzarukin, line 1198: It would be very helpful to get some general words on what the loop does and give some hints why the former version was not correct (like what's important in the latest version).

PR #4087, opened 2025-10-07
tests: prepare Eigen_threadpool implementation for new Eigen versions

dzarukin:
Eigen released a new version that it's recommended to use since it enhanced ThreadPool implementation in there and it should work faster with future asynchronous runtime.

This PR prepares a ground to update existing implementation, which is considered synchronous runtime, from Eigen v3.3.7 to Eigen v5.0.0 (they switched to semantic versioning in this release). It's sort of 3.5.0.0, but they'll be dropping this "world" version 3.

The plan is to promote this change, then move Infrastructure to Eigen v5.0 and release async runtime over Eigen v5.0 and force v5.0 for both synchronous and asynchronous versions.

### DIFF
--- a/cmake/Threadpool.cmake
+++ b/cmake/Threadpool.cmake
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2020-2021 Intel Corporation
+# Copyright 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dzarukin:
make test

cmake/Threadpool.cmake
```diff
@@ -33,10 +33,10 @@ if("${DNNL_CPU_THREADING_RUNTIME}" STREQUAL "THREADPOOL")
     endif()
 
     if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN")
-        find_package(Eigen3 REQUIRED 3.3 NO_MODULE)
+        find_package(Eigen3 3.3...<5.1 REQUIRED NO_MODULE)
```
vpirogov, line 36: Shoudn't we be safe up to 6.0 with semantic versioning?

vpirogov:
Any concerns with `-std=c++14` requirements? Eigen [also declares](https://gitlab.com/libeigen/eigen/-/releases/5.0.0) `-std=c++17` for future releases.

dzarukin:
> Any concerns with `-std=c++14` requirements? Eigen [also declares](https://gitlab.com/libeigen/eigen/-/releases/5.0.0) `-std=c++17` for future releases.

It's C++11 compatible (according to our testing), C++14 was rolling over requirement since Google used that for pretty long time and they simply didn't test with C++11, I presume.
I'd say same thing for C++17 requirement just because Google switched to that as well quite a while ago.

But... We will need Eigen to be built with C++17 anyway because of abseil-cpp piece, so... no concerns, it'll be followed.

cmake/Threadpool.cmake
dzarukin: I decided to stick to what was verified to be working. Future updates can completely switch to version 5 as minimal required.

PR #4090, opened 2025-10-08
aarch64: conv: enable bf16 brdgemm depthwise conv with f32 bias

Ryo-not-rio:
brdgemm already does bias addition in f32 so we can enable this safely


Ryo-not-rio:
ctest -R depthwise passes on c8g locally

src/cpu/aarch64/jit_brdgmm_dw_conv.cpp
```diff
@@ -115,7 +115,9 @@ status_t brdgmm_dw_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
             && mayiuse(isa)
             && IMPLICATION(is_int8,
                     one_of(bia_type, data_type::undef, f32, s32, s8, u8))
-            && IMPLICATION(!is_int8,
+            && IMPLICATION(
+                    is_bf16, one_of(bia_type, data_type::undef, bf16, f32))
+            && IMPLICATION(!is_int8 && !is_bf16,
```
jondea, line 120: Why is the `!is_bf16` necessary? Are they not mutually exclusive?

src/cpu/aarch64/jit_brdgmm_dw_conv.cpp
```diff
@@ -115,7 +115,9 @@ status_t brdgmm_dw_convolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
             && mayiuse(isa)
             && IMPLICATION(is_int8,
                     one_of(bia_type, data_type::undef, f32, s32, s8, u8))
-            && IMPLICATION(!is_int8,
+            && IMPLICATION(
+                    is_bf16, one_of(bia_type, data_type::undef, bf16, f32))
```
michalowski-arm, line 119: So --bia-dt=bf16 is also allowed here? Does it work fine?

Ryo-not-rio: yep, --bia-dt=bf16 already works and the nightly passes

PR #4095, opened 2025-10-08
xe: gemm: jit: use stable sort in getEntries

dyoussif:
closes [MFDNN-14274](https://jira.devtools.intel.com/browse/MFDNN-14274).
In cases where two entries have the same score, maintain pre-existing ordering.


src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -182,7 +182,7 @@ const std::vector<const kcatalog::Entry *> getEntries(const kcatalog::Catalog &c
                       return lhs_score < rhs_score;
 
     };
-    std::sort(entries.begin(), entries.end(), less);
+    std::stable_sort(entries.begin(), entries.end(), less);
```
petercad: `stable_sort` was introduced in C++17, so I don't know if we can use it here.

atkassen (approved):

dyoussif:
make test perf-gpu
set primitive=matmul ip

dyoussif:
make test perf-gpu
set primitive=matmul ip

PR #4097, opened 2025-10-09
brgemm matmul: avx2: f32: closing performance gap for non-gemv cases

densamoilov:
This PR includes the following changes:

- Updates brgemm matmul weight layouts for AVX2 FP32 to match those used for AVX512 FP32, as they offer better performance. Previously, special layouts were used only for AVX2 FP32 (e.g., BA8a24b), which negatively impacted performance due to:
    - 24 = 3 vmm = 1.5 cache lines - inefficient access pattern.
    - Many shapes do not divide evenly by 24, requiring extra handling for tail cases.
- Enables dynamic selection of the best M block size (previously hardcoded to 256), which eliminates unnecessary tail handling and improves work decomposition across threads.
- Performs minor refactoring in a few areas.

Performance data (brgemm matmul vs gemm matmul):
image
image

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -621,20 +621,26 @@ status_t brgemm_matmul_conf_utils_t::set_B_flags(memory_desc_t &B_md) const {
 
 format_tag_t brgemm_matmul_conf_utils_t::pick_blocked_B_layout(
         int n_blk) const {
+
     if (bgmmc.ndims > 3) return format_tag::undef;
-    if (this->is_int8() || this->is_f8()) switch (n_blk) {
+
+    if (is_int8() || is_f8()) {
+        switch (n_blk) {
             case 64: return bgmmc.ndims == 3 ? aCB16b64c4b : BA16a64b4a;
             case 48: return bgmmc.ndims == 3 ? aCB16b48c4b : BA16a48b4a;
             case 32: return bgmmc.ndims == 3 ? aCB16b32c4b : BA16a32b4a;
             case 16: return bgmmc.ndims == 3 ? aCB16b16c4b : BA16a16b4a;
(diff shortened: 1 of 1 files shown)


densamoilov:
make test

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -278,6 +277,32 @@ struct brgemm_matmul_conf_utils_t {
...
+     * @param use_orig_wei_dt If true, use the original weight data type to
+     *   determine block size. If false, use the compute data type type.
+     * @return The total K dimension block size.
+     */
+    int get_wei_k_blk(bool use_orig_wei_dt = false) const {
+        // No blocking is used in GEMV mode.
+        if (bgmmc.is_gemv) return 1;
+
+        // Fixed outer block size.
+        const int k_outer_block = 16;
```
inteldimitrius, line 296: Minor: does it consider avx2?

inteldimitrius: I meant should it be aligned with `vmm length / 32`?

densamoilov: The K (a) block is not contiguous in memory because we don't load along the K dimension (except in the GEMV code path), so the number of elements that can fit in a register doesn't really apply here. I believe we used that number (in `data_type_vnni_simd_elems`) as a proxy to identify the K (a) block size. However, in reality, the K (a) block size is determined inside `pick_blocked_B_layout`.

If we take a look at the removed `data_type_vnni_simd_elems` function, we can see that for all cases except AVX2 FP32, it returned 16. This is another indication that we used that value as a proxy.

I also removed `BA8a24b`, which was introduced and used specifically for AVX2 FP32, although `BA16a64b` (and its versions with smaller N (b) blocks) outperforms it in the vast majority of cases.

PR #4098, opened 2025-10-09
graph: address coverity warnings

TaoLv:
MFDNN-14281

Will backport these to rls-v3.10 later.

### DIFF
--- a/tests/benchdnn/graph/input_displacer.cpp
+++ b/tests/benchdnn/graph/input_displacer.cpp
@@ -555,12 +555,13 @@ int partition_data_displacer_t::displace_input_data(size_t lt_id,
             case ::graph::op::kind::StaticTranspose: {
                 ::std::vector<int64_t> order;
                 op.get_attr_s64_vector(order, "order");
-                size_t ndims = order.size();
-                ::std::vector<int64_t> new_order(ndims, 0);
+                const size_t ndims = order.size();
+                op.attrs_["order"].s64_vector_
+                        = ::std::vector<int64_t>(ndims, 0);
                 for (size_t i = 0; i < ndims; i++) {
-                    new_order[(order[i] + ndims) % ndims] = i;
+                    op.attrs_["order"].s64_vector_[(order[i] + ndims) % ndims]
+                            = i;
                 }
-                op.attrs_["order"].s64_vector_ = new_order;
                 break;
             }
             case ::graph::op::kind::TypeCast:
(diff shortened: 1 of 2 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/layout_propagator.cpp
```diff
@@ -1689,7 +1689,7 @@ status_t layout_propagator_for_gen_index(std::shared_ptr<op_t> &op,
                 src_md.get_data_type(), dnnl::memory::format_tag::abcd);
         insert_reorder_before(op, 0, plain_src_md, p_engine, pd_cache, fpmath,
                 use_block_layout, rewriter);
-        src_md = plain_src_md;
+        src_md = std::move(plain_src_md);
```
dzarukin: Given there's no error handling, why not using `src_md` directly?

That's the question for all std::move changes in the PR.

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -666,8 +666,10 @@ int partition_data_displacer_t::gen_compressed_sdpa_filling(
 
     auto ref_prim_ptr = init_ref_prim_and_fill_data(op, res);
     if (!ref_prim_ptr) {
-        if (res->state == INVALID_ARGUMENTS) return FAIL;
-        if (res->state == SKIPPED || res->state == UNIMPLEMENTED) return OK;
+        if (res->state == SKIPPED || res->state == UNIMPLEMENTED)
+            return OK;
+        else
+            return FAIL;
```
dzarukin, line 676: There's DEFERRED status relevant for graph, I'd check if it's intended to fail here. If yes, then  `else` can be dropped.

TaoLv: The purpose of the change here was to return OK or FAIL if ref_prim_ptr is null, so we will not dereference the null pointer on L675 below which is complained by Coverity.

@ElaineBao Can you please suggest how to handle DEFERRED here?

ElaineBao: Below code should work:
``` 
       if (res->state == SKIPPED || res->state == UNIMPLEMENTED || res->state == DEFERRED)
            return OK;
```
And it's better to explicitly set the prim to nullptr in the function of `init_ref_prim_and_fill_data` when state is DEFERRED:
```
::std::shared_ptr<ref_primitive_t> init_ref_prim_and_fill_data(...) {
    //................

    ref_prim->init_memory_args(::get_test_engine());
    ref_prim->init_ref_memory_args(::get_test_engine(), res);
    if (res->state == SKIPPED || res->state == UNIMPLEMENTED || res->state == DEFERRED) return nullptr;
    return ref_prim;
}
```

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #4101, opened 2025-10-09
cpu: aarch64: skip f16 to f32 upcast for clip and clip_v2 eltwise

Anndrey24:
# Description

Following #4077, this commit also allows the `clip` and `clip_v2` operators to be computed directly in f16, skipping the upcast and downcast to/from f32.  
To enable this, the alpha and beta parameters are downcast to f16 before being written to memory.

### Benchdnn performance results (c7g instance):
 - Command: `./benchdnn --engine=cpu --mode=p --eltwise --alg=clip --alpha=0.1 --beta=0.2 --dt=f16 1024x4096`
 - Threads = 1:
	- JIT upcast to fp32 (old): `total perf: min(ms):2.0835 avg(ms):2.09306`
	- JIT direct in fp16 (new): `total perf: min(ms):0.246826 avg(ms):0.254858`
	- Speedup: **≈8.21x**
 - Threads = 8:
	- JIT upcast to fp32 (old): `total perf: min(ms):0.26001 avg(ms):0.263965`
	- JIT direct in fp16 (new): `total perf: min(ms):0.0461426 avg(ms):0.0487945`
	- Speedup: **≈5.41x**
 - Threads = 16:
	- JIT upcast to fp32 (old): `total perf: min(ms):0.131592 avg(ms):0.133722`
	- JIT direct in fp16 (new): `total perf: min(ms):0.0212402 avg(ms):0.0232317`
	- Speedup: **≈5.76x**

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -1779,7 +1779,7 @@ void jit_uni_eltwise_injector_t<isa>::compute_vector_range(
 namespace {
 // Downcast f32 -> f16, then pack that 16-bit pattern twice into a uint32: [f16 | f16]
 inline uint32_t pack_f16_twice_from_f32(float s) {
-    float16_t h = s;
+    float16_t h = float16_t(s);
     uint16_t h_bits = h.raw;
     return (uint32_t(h_bits) << 16) | uint32_t(h_bits);
 }


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -1751,6 +1754,15 @@ void jit_uni_eltwise_injector_t<isa>::compute_vector_range(
     injector_postamble();
 }
 
+namespace {
+// Downcast f32 -> f16, then pack that 16-bit pattern twice into a uint32: [f16 | f16]
+inline uint32_t pack_f16_twice_from_f32(float s) {
+    float16_t h = s;
```
jondea: I'd prefer an explicit cast/round here to mark that rounding may happen

Anndrey24:
> Also, how did you check that this was numerically sound?

Since I don't believe there are any benchdnn test cases with alpha and beta values that round to different numbers when cast to f32 vs f16 (i.e. they are not exactly representable in either datatype), I've tried manually checking for this case using:
```
./benchdnn --mode=c --eltwise --alg=clip \
--alpha=-0.1,-0.2,-0.3,-0.4,-0.6,-0.7,-0.8,-0.9 \
--beta=0.1,0.2,0.3,0.4,0.6,0.7,0.8,0.9 \
--dt=f16 1024x4096
```
Should one of these pairs be added as a separate test case in the CI or nightly test sets? (they have been useful to me in disproving that a similar F16-native approach could be used for computing _leaky relu_, _hardsigmoid_, and _hardswish_)

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: Okay, this apears to be using the built in rounding function and there isn't a nice explicit one we can use, happy to leave as is https://github.com/uxlfoundation/oneDNN/blob/5f36333c572affc877462671e2af809e151c43dd/src/common/float16.hpp#L52

jondea:
Great work, thanks. I'm not sure it's necessary for this PR, but it's something to consider.

@mgouicem was the choice of only FP exact numbers intentional for the eltwise alpha/beta in the existing tests? Even if not, I think it could be argued that alpha/beta should be f16/bf16 numbers in such cases. This would still leave the f16/bf16 eltwise <1ulp in the case where alpha/beta is not exactly representable in f16/bf16.

mgouicem:
> Since I don't believe there are any benchdnn test cases with alpha and beta values that round to different numbers when cast to f32 vs f16 (i.e. they are not exactly representable in either datatype)

IMHO, it is not about a value not exactly representable in both datatypes. However, you might have issues when rounding alpha/beta to f16, to preserve outputs in `[alpha, beta]` interval.

Here is a corner case example:
- choose a midpoint between two fp16 values, calle it `m`
- pick beta an fp32 as `m + epsilon` with `epsilon < 1/2 * ulp_fp16`
- if you round beta to fp16, it will return next_fp16(m), and now you can have outputs > beta, even though `clip_v2(s, alpha, beta)` returns a value in `[alpha, beta]`

To avoid above issue, you have to round beta down, and alpha up. This will preserve the property that the result of clip is within `[alpha, beta]`


> @mgouicem was the choice of only FP exact numbers intentional for the eltwise alpha/beta in the existing tests?

Yes as FP exact numbers help avoid unexpected rounding, e.g. during parsing, and also help with bound checking as we fill inputs with exact numbers as well.

> Even if not, I think it could be argued that alpha/beta should be f16/bf16 numbers in such cases. This would still leave the f16/bf16 eltwise <1ulp in the case where alpha/beta is not exactly representable in f16/bf16.

I believe the alpha/beta we pick are typically exact in fp16 as well (e.g. small negative powers of two like 0.5, 0.25, ...).

Anndrey24:
> To avoid above issue, you have to round beta down, and alpha up. This will preserve the property that the result of clip is within `[alpha, beta]`

Thank you for pointing this out! I completely agree that this edge case goes against the real-valued guarantee of clip that output values must be in the `[alpha, beta]` interval. However, as far as I can tell, the same behaviour would also occur with our current _upcast-compute-downcast_ f16 implementation.

Given the same `beta = m + epsilon` (e.g. closer to 1.0 than 0.9990234375 -- two consecutive f16 values), inputs greater than `beta` would be clamped to `beta` in f32, but then get rounded to the nearest f16 value anyway (in this case greater than `beta`) in the subsequent downcast to f16 using `FCVT` (default rounding method being "_Round to Nearest_" [mode](https://developer.arm.com/documentation/ddi0601/2025-09/AArch64-Registers/FPCR--Floating-point-Control-Register#:~:text=Round%20to%20Nearest%20(RN)%20mode.)).

In practice, this means that the final f16 outputs aren’t strictly within the real-valued `[alpha, beta]` interval, but rather within`[float16(alpha), float16(beta)]` (i.e. the same range that would result if `alpha` and `beta` were rounded to f16 before the computation, dependent on rounding mode).

As an example, I stepped through the _upcast–compute-downcast_ JIT kernel with `alpha = 10,001` (representable in f32 as _0x461c4400_, but not in f16, where the nearest representable values are `10,000` (_0x70e2_) and `10,008` (_0x70e3_)). Inputs of zero are clamped to `alpha = 10,001` in f32, and then downcast to `10,000` (_0x70e2_) in f16,  smaller than the real-valued `alpha`.

I'm not sure whether this is the intended behaviour, but I don't believe this rounding-to-nearest edge case is introduced in this _compute-in-f16_ version alone. Rather, it was already present in the existing implementation, too.

I can look into performing the rounding of alpha (up) and beta (down) as you suggested, if that's how it's expected to behave.

mgouicem:
That is a fair point. I agree this behavior precedes this PR. Let's not block this PR as it is aligned with current behavior, and revise that behavior if we get any user request.

> I can look into performing the rounding of alpha (up) and beta (down) as you suggested, if that's how it's expected to behave.

I don't think it is necessary for now, as we would need all impls to align. Thanks.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
mgouicem: (nit) You can use the [constructor ](https://github.com/uxlfoundation/oneDNN/blob/5f36333c572affc877462671e2af809e151c43dd/src/common/float16.hpp#L37)`float16_t(s)` to make it explicit.

jondea (approved):

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Thank you, done!

Sqvid:
> Should one of these pairs be added as a separate test case in the CI or nightly test sets? (they have been useful to me in disproving that a similar F16-native approach could be used for computing leaky relu, hardsigmoid, and hardswish)

I think it might be worth adding one of these cases to `tests/benchdnn/inputs/eltwise/option_set_all_algs`.

PR #4105, opened 2025-10-10
benchdnn: inputs: graph: replace multiply with add to pass accuracy check

TaoLv:
Fixes MFDNN-14192

Replace binary-mul with binary-add to mitigate an f16 accuracy issue on a windows platform.


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4114, opened 2025-10-11
third_party: add abseil-cpp (tag: 20250814.1) as a future dependency

dzarukin:
Will be used in Eigen threadpool implementation coming in the next PR and for the asynchronous implementation coming somewhere after.

It's split to simplify the review process of changes that are actually relevant.
The only file that is not taken from the upstream is README.md which is welcome to be read.


third_party/abseil-cpp/README.md
```diff
@@ -0,0 +1,13 @@
+This is a copy-exact of all sources and CMake files from
```
mzhukova, line 1:
```suggestion
This is an exact copy of all sources and CMake files from
```
(or maybe just "clone")

third_party/abseil-cpp/README.md
```diff
@@ -0,0 +1,13 @@
...
+[abseil-cpp](git@github.com:abseil/abseil-cpp.git). The software is used to
+support oneDNN validation of Eigen Threadpool runtime.
+
+The snippet is taken from `lts_2025_08_14` branch, commit `d38452e1` which
+denotes `tag: 20250814.1`.
+
+The management of the dependency on abseil-cpp is done through
+tests/CMakeLists.txt.
```
mzhukova, line 9:
```suggestion
`tests/CMakeLists.txt`.
```

third_party/abseil-cpp/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Note: in case sources are not compiling, especially with GCC, which support is
```
mzhukova, line 11: minor suggestion -- if it is not going to be rendered into any sort of web documentation (and stays as .md), maybe you could use 
```
> [!NOTE] 
> blahblah
```
, so that it is displayed as a note :)

dzarukin:
Closed in a favor of building a package and distribute it the same way other software for validation is distributed.

PR #4118, opened 2025-10-12
cpu: x64: matmul: add a robust check for fallback compatibility

densamoilov:
Fixes MFDNN-14255.

This PR adds a robust fallback compatibility check. When the BRGEMM-based matmul decides to fall back to the GEMM-based matmul, we must ensure that the GEMM implementation can handle the problem, otherwise, we risk falling back to the reference implementation.


densamoilov:
make test

densamoilov:
make test

PR #4119, opened 2025-10-13
examples: graph: update implicit causal mask example to use host scalar

TaoLv:
Now the example uses host scalar to define:

- Attention scale
- Sequence length (part of implicit causal mask)
- -infinity (part of implicit causal mask)


TaoLv:
make test
disable benchdnn_all

TaoLv:
make test
disable benchdnn_all

PR #4125, opened 2025-10-13
ci: aarch64: Fix permission issues and input data handling

Ryo-not-rio:
- fixed default permissions as write-all is being flagged in the OSSF scorecard
- fixed unsafe input data handling

### DIFF
--- a/.github/workflows/aarch64-acl.yml
+++ b/.github/workflows/aarch64-acl.yml
@@ -28,24 +28,8 @@ on:
 
   workflow_dispatch:
 
-# Declare default permissions as read only.
-# We have to declare each permission individually as
-# setting this to read-all causes issues with the id-tokens permission
 permissions:
-  actions: read
-  checks: read
   contents: read
-  deployments: read
-  discussions: read
-  issues: read
-  packages: read
-  pages: read
-  pull-requests: read
-  repository-projects: read
-  security-events: read
-  statuses: read
-  attestations: read
-  models: read
   id-token: none
 
 jobs:
--- a/.github/workflows/performance-aarch64.yml
+++ b/.github/workflows/performance-aarch64.yml
@@ -52,23 +52,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-performance
   cancel-in-progress: true
 
-# Declare default permissions as read only.
 # Setting this to read-all causes issues with the id-tokens permission
 permissions:
-  actions: read
-  checks: read
   contents: read
-  deployments: read
-  discussions: read
-  issues: read
-  packages: read
-  pages: read
-  pull-requests: read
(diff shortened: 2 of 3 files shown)


.github/workflows/performance-aarch64.yml
```diff
@@ -222,8 +240,10 @@ jobs:
         if: ${{ inputs.benchdnn_command == '' }}
         shell: bash
         run: |
-          OMP_NUM_THREADS=${{ inputs.num_threads || 16 }} bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_nightly_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn base.txt new.txt
+          bash ${{ github.workspace }}/oneDNN/.github/automation/performance/bench_nightly_performance.sh ${{ github.workspace }}/oneDNN_base/build/tests/benchdnn/benchdnn ${{ github.workspace }}/oneDNN_new/build/tests/benchdnn/benchdnn base.txt new.txt
```
vpirogov, line 228: These changes look unrelated. Please separate these into a different commit.

Ryo-not-rio: Done and I've updated the PR message

.github/workflows/aarch64-acl.yml
```diff
@@ -29,7 +29,24 @@ on:
...
+  issues: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  security-events: read
+  statuses: read
+  attestations: read
+  models: read
+  id-token: none
```
vpirogov: Could you please elaborate on issues with id-token?

Do you have a sense of what permissions are actually needed for this workflow?

Ryo-not-rio: It's a bug to do with how github actions parses nested workflows. If a parent workflow of a nested workflow specifies `permissions: read-all` and the nested workflow specifies anything other than `read-all`, it causes a syntax error since the parent workflow sets `id-token: read` but the user only has options to set `id-token` to `none` or `write` in the nested workflow. An example of this failure can be seen here: https://github.com/uxlfoundation/oneDNN/actions/runs/18458580154

I'm not entirely sure but we could figure out the minimum required while we're at it if you would prefer

vpirogov: Sorry, went down to a rabbit hole of `id-token` semantics.

Would something like this be sufficient? Current list is clearly unnecessary.
```suggestion
  contents: read
  id-token: none
```

Ryo-not-rio: Done!

PR #4134, opened 2025-10-14
cpu: x64: matmul fix wei_k_blk query

densamoilov:
Fixup for #4097

Fixes MFDNN-14302. 

`wei_k_blk` is not always defined for `orig_wei_dt` therefore I dropped querying it except for 1 case for which it's needed and always defined (bf32 case).


densamoilov:
make test

PR #4135, opened 2025-10-14
src, tests: align synchronous Eigen threadpool implementation

dzarukin:
Aligns the implementation with the one from TensorFlow legacy runtime and adjusts the internals as they should be for truly synchronous threadpool.

This action introduces a dependency on abseil-cpp in Eigen threadpool implementation.

Testing binaries received an extra lib dependency across different CMake files. It will be re-used for asynchronous runtime.

### DIFF
--- a/cmake/Threadpool.cmake
+++ b/cmake/Threadpool.cmake
@@ -33,10 +33,16 @@ if("${DNNL_CPU_THREADING_RUNTIME}" STREQUAL "THREADPOOL")
     endif()
 
     if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN")
-        find_package(Eigen3 3.3...<5.1 REQUIRED NO_MODULE)
+        find_package(Eigen3 5.0 REQUIRED NO_MODULE)
         if(Eigen3_FOUND)
             list(APPEND EXTRA_STATIC_LIBS Eigen3::Eigen)
-            message(STATUS "Threadpool testing: Eigen (${PACKAGE_PREFIX_DIR})")
+            message(STATUS "Found Eigen: ${PACKAGE_PREFIX_DIR}")
+        endif()
+
+        find_package(absl REQUIRED CONFIG)
+        if(absl_FOUND)
+            list(APPEND EXTRA_STATIC_LIBS absl::synchronization)
+            message(STATUS "Found abseil-cpp: ${PACKAGE_PREFIX_DIR}")
         endif()
     endif()
 
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2018-2024 Intel Corporation
+# Copyright 2018-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -47,7 +47,7 @@ endfunction()
 #   arg4 -- (optional) list of extra library dependencies
 function(register_exe name srcs test)
     add_executable(${name} ${srcs})
-    target_link_libraries(${name} ${LIB_PACKAGE_NAME} ${EXTRA_SHARED_LIBS} ${ARGV3})
+    target_link_libraries(${name} ${LIB_PACKAGE_NAME} ${EXTRA_SHARED_LIBS} ${EXTRA_STATIC_LIBS} ${ARGV3})
     if("x${test}" STREQUAL "xtest")
         add_dnnl_test(${name} ${name})
         maybe_configure_windows_test(${name} TEST)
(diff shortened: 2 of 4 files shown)


dzarukin:
Hi @Sqvid, I want to get your attention on this PR. It changes the implementation of Eigen threadpool in validation and makes in actually synchronous, meaning there'll be no synchronization inside the library and [this spot](https://github.com/openxla/xla/blob/main/xla/tsl/util/onednn_threadpool.h#L141-L142) in the original TF threadpool implementation will lead to crashes due to the wrong order of decreasing counter and doing jobs (compare to the same spot in this PR which copies the code almost exactly).

I got insights this is a product of ACL integration. Once this PR lands oneDNN from main can't be used for legacy TF runtime until the threadpool implementation gets updated.

This is just for your information, feel free to pass this information to interested parties. Thanks.

tests/CMakeLists.txt
```diff
@@ -93,6 +93,15 @@ endif()
...
 set(TEST_THREAD ${CMAKE_CURRENT_SOURCE_DIR}/test_thread.cpp)
 
+set(LIBTHREADPOOL)
+if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN")
+    find_package(absl REQUIRED CONFIG)
+    if(absl_FOUND)
+        list(APPEND LIBTHREADPOOL absl::synchronization)
+        message(STATUS "Found abseil-cpp: ${PACKAGE_PREFIX_DIR}")
+    endif()
+endif()
```
mgouicem: Maybe put this alongside Eigen findpackage call in cmake/Threadpool.cmake?

Sqvid:
@dzarukin, thank you for the heads up. I have brought this to the attention of our team. Much appreciated.

tests/CMakeLists.txt
cfRod: >I got insights this is a product of ACL integration. Once this PR lands oneDNN from main can't be used for legacy TF runtime until the threadpool implementation gets updated.


This is an issue not just for ACL right but also non-ACL paths?

dzarukin: Should be true, yes, the thing is the team we are working with are aware of that and have modifications in private repo for that path (along with async runtime support for XLA), but for your team it wouldn't be the case, so I wanted to drag your attention.

In best case scenario if no plans to touch it, threadpool impl will get updated by v3.11 update (hopefully) and won't appear at all.

dzarukin: > Maybe put this alongside Eigen findpackage call in cmake/Threadpool.cmake?

Moved, and switched to EXTRA_STATIC_LIBS instead, and added that to register_exe as a single spot. Thanks.

spalicki (approved):

tests/CMakeLists.txt
cfRod: Hi Dmitry, 
>  have modifications in private repo for that path 

I am assuming these are modifications to [this spot](https://github.com/openxla/xla/blob/main/xla/tsl/util/onednn_threadpool.h#L141-L142) that will make TensorFlow compatible with the changes in this PR. I am not sure why that would be different for ACL path?

From my understanding,  for oneDNN-ACL primitive we dont have async runtime support yet and we will rely on [[1](https://github.com/openxla/xla/commit/5db7c22b1563586b5c359b18426fc4da17c1f66c), [2](https://github.com/openxla/xla/commit/0fb65c3111e4e6e27665381241c09ea0c0cd4f16)] to continue leveraging onednn in XLA,  until we have async support in the ACL path.

For TensorFlow, we would expect onednn threadpool to ensure synchronization - I am assuming that we  will have separate logic for both TF and XLA within onednn_threadpool.h to support both threadpool. I might be incorrect about this assumption!

dzarukin: The problem is [parallel](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/dnnl_thread.hpp#L329-L332) call delivers synchronization as of now until this PR gets merged. **Synchronous** TF legacy threadpool is marked as [_asynchronous_](https://github.com/openxla/xla/blob/main/xla/tsl/util/onednn_threadpool.h#L91), and that's how it plays together and remain synchronized.

On our end we change parallel to stop doing any synchronization, and this responsibility is expected to be provided by a threadpool implementation.

From the support perspective, there's no difference for x64 or aarch64/ACL path, it's just there's gonna be a time lag between updating legacy threadpool on TF side by Intel TF team to become truly synchronous and oneDNN main branch state where synchronization is removed.
During that time window ACL is exposed to the bug I mentioned as Intel TF doesn't do anything to support ACL (AFAIK). If you don't do active development on a main branch, sitting tight and waiting for changes to arrive would be fine, no actions required in that case. If you do develop on top of main, you want those lines swapped on your end to have it working properly.

I just wanted to warn you in case you are doing active development over oneDNN main within TF assuming you have your own testing infrastructure and procedures so that expected crushes don't stun your development.

densamoilov (approved):

PR #4137, opened 2025-10-14
graph: backend: dnnl: fix softmax fusion patterns

TaoLv:
Adjust the order of epilogue ops.

Now the optional TypeCast is between softmax and unary/binary post-ops. Before the change, it was after the unary/binary post-ops.

### DIFF
--- a/src/graph/backend/dnnl/patterns/softmax_post_ops.cpp
+++ b/src/graph/backend/dnnl/patterns/softmax_post_ops.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2022-2024 Intel Corporation
+* Copyright 2022-2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


rongzha1:
could you explain more about this change?  is this change used for softmax-->f32-->tc-->xf16-->post-ops? If so, seems we can't fuse them: according to spec, post ops should works on f32.

TaoLv:
> could you explain more about this change? is this change used for softmax-->f32-->tc-->xf16-->post-ops? If so, seems we can't fuse them: according to spec, post ops should works on f32.

Actually it's for: xf16 -> softmax -> xf16 -> TypeCast -> f32 -> post-ops -> 

TypeCast is optional in cases softmax generates f32 output.

I'm thinking if we still need another TypeCast after post-ops in cases the user wants xf16 output. Please suggest.

rongzha1:
> Actually it's for: xf16 -> softmax -> xf16 -> TypeCast -> f32 -> post-ops ->

according to this pattern, it is not align with bf16 in f32 out softmax primitive + post ops.

in this pattern, first get xf16 softmax output, then cast it to f32.  This kind of cast will cause precision loss compared to f32 output directly. 

So this pattern is not compatible with our current philosophy. What do you think?

TaoLv:
You are right. I will revisit the changes.

PR #4138, opened 2025-10-14
Coverity failures fix

mgouicem:


tests/gtests/api/test_memory_desc.cpp
```diff
@@ -118,10 +118,12 @@ TEST(c_api_host_scalar_md, TestHostScalarInvalidType) {
     ASSERT_EQ(dnnl_memory_desc_create_host_scalar(
                       &scalar_md, (dnnl_data_type_t)999),
             dnnl_invalid_arguments);
+    EXPECT_EQ(scalar_md, nullptr);
```
mzhukova, line 121: just wondering what was the Coverity complaint here? 👀

dzarukin: Unchecked variable/object, it suspected we missed it based on code in previous tests.

mgouicem: Yes, It assumes that scalar_md was allocated but not freed. Adding the check reassures it (and kind of makes sense).

mzhukova (approved):

PR #4145, opened 2025-10-14
xe: jit: gemm: fix stream-k slab count calculation (MFDNN-14012)

petercad:
Addresses MFDNN-14012.

There was a bug in the code calculating the number of stream-k slabs that could lead to suboptimal L3 cache usage.


petercad:
make test perf-gpu
set primitive=gpu:gemm

atkassen (approved):

skazakov1 (approved):

PR #4150, opened 2025-10-15
cpu: aarch64: jit_reorder: fix dispatch issue for direct copy

davsva01:
# Description

Adds checks in jit:uni reorder to ensure direct copy problems gets dispatched to direct copy implementation. Fixes from 1a0fb18.

Improves performance by ~1.5x running cmd: `OMP_NUM_THREADS=16 ./tests/benchdnn/benchdnn --reorder --sdt=f32 --ddt=f32 --stag=ab --dtag=ab --mode=p 4096x4096`

Ran with nightly CI tests: `100% tests passed, 0 tests failed out of 243`

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/jit_uni_reorder.cpp
+++ b/src/cpu/aarch64/jit_uni_reorder.cpp
@@ -55,6 +55,10 @@ namespace aarch64 {
 
 namespace tr {
 
+static inline bool is_direct_copy(const prb_t &prb) {
+    return prb.ndims == 1 && prb.nodes[0].is == 1 && prb.nodes[0].os == 1;
+}
+
 static bool prb_has_small_strides(const prb_t &prb) {
     constexpr ptrdiff_t max_stride = (1LL << 31) - 1;
     for (int d = 0; d < prb.ndims; ++d) {
@@ -2606,7 +2610,7 @@ static void prb_block_for_cache(tr::prb_t &prb) {
 
     const bool cache_blocking_needed
             = stride_cache_friendly || requires_inner_blocking;
-    if (!cache_blocking_needed) return;
+    if (!cache_blocking_needed || is_direct_copy(prb)) return;
 
     int unit_input_stride_idx = -1;
     for (auto idx = 0; idx < prb.ndims; ++idx) {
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2690,7 +2694,10 @@ static void prb_thread_kernel_balance(
      * size_drv_min = C0 + FC * (nthr > 1 ? 1 : 0) + VC * (nthr - 1)
      * where FC and VC are fixed and variable costs respectively.
      * Though for now, the below heuristic seems to be good enough */
-    const size_t size_drv_thr = (nthr > 1) ? 16 * nthr : 1;
+    // Note: direct copy needs only as many kernels as nthr.
+    const size_t size_drv_thr = is_direct_copy(prb) ? nthr
```
jondea, line 2698: Maybe this is just me, but I find nested ternaries incomprehensible. In this case it also separates the explanation in the comment from the relevant logic

jondea (changes requested):
Minor comment, otherwise looks like a lovely fix

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -55,6 +55,10 @@ namespace aarch64 {
 
 namespace tr {
 
+static bool is_direct_copy(const prb_t &prb) {
```
Sqvid: Do you expect this to have more general use? If not it might make sense to just define a `bool` in `prb_has_small_strides()`.

If you keep it here I might add `inline`.

davsva01: I don't think `prb_has_small_strides()` might be the right place for that check, but I've added `inline` which makes sense for this. Thanks!

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: Thanks Jonathan. I'll keep it as is now partly to keep it in line with x64 impl. while I go through differences between the two implementations, hope that's alright.

jondea (approved):

Sqvid:
@davsva01 can you open a backport against `v3.10`?

Sqvid:
Great patch. This scales well too. On 64 threads:
Before:
```
total perf: min(ms):0.102295 avg(ms):0.107472
```

After:
```
total perf: min(ms):0.048584 avg(ms):0.0634066
```

PR #4157, opened 2025-10-16
ci: aarch64: make comparison script friendlier for local testing

Ryo-not-rio:
# Description

Only run ttest if there are 3 or more test runs. This makes it easier to use the comparison script for local testing


Radu2k:
LGTM, please wait for the CI to validate the changes.

PR #4178, opened 2025-10-21
cpu: x64: matmul: fix memory leak

densamoilov:
Fixes memory leak.


densamoilov:
make test
disable test_device_gpu

PR #4182, opened 2025-10-21
tests: add asynchronous Eigen threadpool runtime implementation

dzarukin:
Introduces future asynchronous Eigen threadpool runtime.

This action introduces a dependency on custom XLA package in async Eigen threadpool implementation.


cmake/options.cmake
```diff
@@ -267,7 +267,7 @@ set(_DNNL_TEST_THREADPOOL_IMPL "STANDALONE" CACHE STRING
     "specifies which threadpool implementation to use when
     DNNL_CPU_RUNTIME=THREADPOOL is selected. Valid values: STANDALONE, EIGEN,
     TBB")
```
mzhukova:
```suggestion
    TBB, EIGEN_ASYNC")
```

tests/test_thread.cpp
```diff
@@ -191,6 +191,86 @@ class threadpool_t : public dnnl::threadpool_interop::threadpool_iface {
...
+#include <cstdint>
+#include <functional>
+
+#include "Eigen/ThreadPool"
+
+#include "xla/backends/cpu/runtime/work_queue.h"
+#include "xla/tsl/concurrency/async_value_ref.h"
+#include "xla/tsl/concurrency/chain.h"
+
+#define EIGEN_USE_THREADS
```
mzhukova: looks like  `EIGEN_USE_THREADS` is defined twice

tests/test_thread.cpp
```diff
@@ -191,6 +191,86 @@ class threadpool_t : public dnnl::threadpool_interop::threadpool_iface {
...
+                    [fn, n](size_t i) { fn(static_cast<int>(i), n); });
+        };
+
+        done_event_ = done_event_.FlatMap(parallelize);
+    }
+    void wait() override {
+        // While performing asynchronous execution, wait() method is needed to
+        // notify the user that the output is ready. oneDNN will not call wait()
+        // inside the library to avoid deadlock.
+        tsl::BlockUntilReady(done_event_);
```
mzhukova, line 267: could this thing return an error? and if so, do we need to handle it somehow in `wait()`?

tests/test_thread.cpp
```diff
@@ -191,6 +191,86 @@ class threadpool_t : public dnnl::threadpool_interop::threadpool_iface {
...
+
+#define EIGEN_USE_THREADS
+#include "unsupported/Eigen/CXX11/Tensor"
+
+namespace dnnl {
+namespace testing {
+
+static tsl::AsyncValueRef<tsl::Chain> OkDoneEventSingleton() {
+    static tsl::AsyncValueOwningRef<tsl::Chain> *singleton = [] {
+        auto *storage = new tsl::internal::AsyncValueStorage<tsl::Chain>();
```
mzhukova: does it need clean up too..?

mzhukova: just wondering, who does freeing of the `storage`?

spalicki (commented):
I think we should also add some entries in the README: https://github.com/uxlfoundation/oneDNN?tab=readme-ov-file#runtime-dependencies, especially for absl and xla dependencies and marking Eigen 5.0 as the only supported one. 
Also, some addition here: https://github.com/uxlfoundation/oneDNN/blob/main/doc/build/build_options.md#threadpool

tests/test_thread.cpp
dzarukin: This is a declaration:
```
template <typename T>
void BlockUntilReady(const AsyncValueRef<T>& ref);
```

I'd be surprised if functions blocking the execution return status because it either succeeded and the program goes further, or deadlocked and it doesn't.

tests/test_thread.cpp
dzarukin: > Objects with static storage duration are automatically deallocated when the program exits. Their destructors are called in the reverse order of their construction. This means that you typically do not need to explicitly call delete or free on static objects, as the C++ runtime handles their cleanup.

cmake/options.cmake
dzarukin: Thanks, added the missed one.

tests/test_thread.cpp
dzarukin: Thanks for spotting, updated the section.

dzarukin:
> I think we should also add some entries in the README: https://github.com/uxlfoundation/oneDNN?tab=readme-ov-file#runtime-dependencies, especially for absl and xla dependencies and marking Eigen 5.0 as the only supported one.

Introduced dependencies are not a part of the library build process, those are solely belong to validation dependencies, thus, I'm not updating this section, no new deps for the library.

> Also, some addition here: https://github.com/uxlfoundation/oneDNN/blob/main/doc/build/build_options.md#threadpool

Good call, I've updated this section to the best guidance I could provide on async part.

doc/build/build_options.md
```diff
@@ -236,29 +236,40 @@ oneDNN has functional limitations if built with TBB:
...
 ~~~sh
 $ cmake -DONEDNN_CPU_RUNTIME=THREADPOOL ..
 ~~~
 
+Threadpool threading support has the same limitations as TBB plus more:
+* As threadpools are attached to streams which are only passed during primitive
+  execution, work decomposition is performed statically at primitive creation
+  time. At the primitive execution time, the threadpool is responsible for
+  balancing the static decomposition from the previous item across available
+  worker threads.
```
vpirogov:
```suggestion
  balancing this decomposition across available worker threads.
```

doc/build/build_options.md
```diff
@@ -236,29 +236,40 @@ oneDNN has functional limitations if built with TBB:
...
+The `TBB` requires passing `TBBROOT` for CMake to find a package.
+
+The `EIGEN` requires Eigen 5.0 or higher and Abseil-CPP packages to be
+discoverable by CMake.
 
+The `EIGEN_ASYNC` has same requirements as `EIGEN` and additionally requires
+openXLA threadpool package, however, additional actions might be required to
+compile tests since this threadpool implementation relies on internal openXLA
+headers. Additional actions will likely require header files placement to be
+followed as they reside inside openXLA project.
```
vpirogov: The last sentence does not seem to add much.

```suggestion
headers.
```

tests/test_thread.cpp
mzhukova: sorry I wasn't clear, I didn't necessarily meant that the function `BlockUntilReady` should return an error code, but maybe there is extra info returned by event that might need to be monitored/checked?

tests/test_thread.cpp
densamoilov: I'm not sure what `AsyncValueOwningRef` does but it seems that it just calls a destructor of the storage type explicitly, without releasing the allocated buffer (see [here](https://github.com/mansnils/tensorflow/blob/b0164f014fd4f1b5af2c7b578aa7687198c5d92e/tensorflow/tsl/concurrency/async_value_ref.h#L435)). Does it mean that `storage` will leak?

One more thing, `singleton` is a pointer so somebody will have to release it explicitly, but I couldn't find it in this PR. Can you please clarify this?

cmake/Threadpool.cmake
```diff
@@ -32,7 +32,8 @@ if("${DNNL_CPU_THREADING_RUNTIME}" STREQUAL "THREADPOOL")
         endif()
     endif()
 
-    if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN")
+    if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN" OR
+            "${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN_ASYNC")
```
densamoilov: (nit)
```suggestion
if(_DNNL_TEST_THREADPOOL_IMPL MATCHES "^(EIGEN|EIGEN_ASYNC)$")
```

cmake/Threadpool.cmake
```diff
@@ -44,6 +45,12 @@ if("${DNNL_CPU_THREADING_RUNTIME}" STREQUAL "THREADPOOL")
             list(APPEND EXTRA_STATIC_LIBS absl::synchronization)
```
densamoilov, line 46: Do we use it for both EIGEN and EIGEN_ASYNC?

cmake/Threadpool.cmake
```diff
@@ -32,7 +32,8 @@ if("${DNNL_CPU_THREADING_RUNTIME}" STREQUAL "THREADPOOL")
         endif()
     endif()
 
-    if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN")
+    if("${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN" OR
+            "${_DNNL_TEST_THREADPOOL_IMPL}" STREQUAL "EIGEN_ASYNC")
         find_package(Eigen3 5.0 REQUIRED NO_MODULE)
```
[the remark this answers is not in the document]
dzarukin: Applied proper scoping in CMake code, thanks.

doc/build/build_options.md
```diff
@@ -236,29 +236,40 @@ oneDNN has functional limitations if built with TBB:
...
 The `_ONEDNN_TEST_THREADPOOL_IMPL` CMake variable controls which of the three
-threadpool implementations would be used for testing: `STANDALONE`, `TBB`, or
-`EIGEN`. The latter two require also passing `TBBROOT` or `Eigen3_DIR` paths
-to CMake. For example:
+threadpool implementations would be used for testing: `STANDALONE`, `TBB`,
+`EIGEN`, `EIGEN_ASYNC`.
+
+The `TBB` requires passing `TBBROOT` for CMake to find a package.
+
+The `EIGEN` requires Eigen 5.0 or higher and Abseil-CPP packages to be
```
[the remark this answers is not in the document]
dzarukin: There was a bug in CMake, doc part remains correct.

doc/build/build_options.md
dzarukin: Applied, thanks.

doc/build/build_options.md
dzarukin: Applied, thanks.

cmake/Threadpool.cmake
dzarukin: Applied, thanks.

cmake/Threadpool.cmake
dzarukin: Abseil - yes, but XLA is not shared, updated that CMake part.

tests/test_thread.cpp
dzarukin: I've wrapped both allocations into unique_ptr, hope it addresses your concerns.

tests/test_thread.cpp
dzarukin: AFAIK, it's sufficient. The runtime fills `done_event_` with the info when all tasks are completed and `BlockUntilReady` releases the submitter to go further after that.

densamoilov (approved):

tests/test_thread.cpp
densamoilov: yeah, I think it should fix the leak

PR #4184, opened 2025-10-22
gpu: conv: jit: add early slm check to skip ngen jitting

ZackyLake:
# Description

Convolution's jit kernel could fail due to SLM usage exceeding limit:
https://github.com/uxlfoundation/oneDNN/blob/05b4b09de057482a0b324cf8938476400418905b/src/gpu/intel/jit/codegen/codegen.cpp#L1752-L1758

The common usage of SLM in conv is for reduce of k slicing:
https://github.com/uxlfoundation/oneDNN/blob/05b4b09de057482a0b324cf8938476400418905b/src/gpu/intel/conv/jit/ir_builder.cpp#L513-L517

And this size is known with the config, no need to spend time in IR generation.

This PR calculates this SLM usage and perform early check, should save some compilation time wasting on unsuitable config.

Code could be better by abstracting logic into functions rather than copy-pasting, but that introduce a bigger change...

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes

### DIFF
--- a/src/gpu/intel/conv/jit.cpp
+++ b/src/gpu/intel/conv/jit.cpp
@@ -162,6 +162,7 @@ public:
                 init_nd_ranges(primitive, cfg);
                 auto &kernel_infos = data.kernel_infos;
 
+                std::vector<int> conv_tg_sizes;
                 // This absolutely HAS to be executed first if present,
                 // since it adds its own version mark to the cache blob
                 for (int i = 0; i < int(kernel_infos.size()); i++)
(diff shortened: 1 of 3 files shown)


echeresh:
@ZackyLake Thanks for the PR! Do you observe the related overhead in real workloads?

I agree with your comment - we can fail much earlier in case of insufficient SLM. However, I suggest two adjustments:

1. Extend `is_slm_limit_ok()` check to estimate the total SLM size (not just k-slicing related). Also it makes sense to move it out of `tiler.cpp` file (as it's not tiler-specific)
2. Drop the check [here](https://github.com/uxlfoundation/oneDNN/blob/ba879ca8ea3e7707511f5af27a1b9b1cab166abd/src/gpu/intel/jit/codegen/codegen.cpp#L1751-L1759) and add assertions validating real SLM size based on the estimated size at step 1

If this is a general improvement - I suggest to focus on making the check more complete (see bullets 1 and 2 above). But if this is a quick fix to address some E2E overhead - then I'm fine to merge this as is with some minor adjustments (like rename `is_slm_limit_ok()` and move it out of `tiler.cpp` to emphasize that it's only k-slicing related check, etc).

src/gpu/intel/conv/jit.cpp
```diff
@@ -162,15 +162,29 @@ class gen_t {
...
+                        const auto &local_range = nd_ranges_[i].local_range();
+                        const auto tg_size = ir_utils::safe_divide(
+                                into<int>(local_range.nelems()), cfg.simd());
+                        conv_tg_sizes.emplace_back(tg_size);
                     }
 
+                // Perform early check to ensure no exceeding SLM size
+                if (!std::all_of(conv_tg_sizes.begin(), conv_tg_sizes.end(),
+                            [&](const int tg_size) {
+                                return tiler->is_slm_limit_ok(cfg, tg_size);
```
rjoursler, line 184: @ZackyLake thanks for the PR. How significant of an impact on performance are you seeing from this change?

vpirogov:
Closing as stale.

PR #4187, opened 2025-10-22
[GPU] a non-blocking, profiler-based verbose mode for execution time tracking

avmanerikar:
# Description

The PR implements a non-blocking verbose profiling to record and print primitive execution times without relying on blocking, host-side `stream.wait()` calls. The implementation is aimed at triaging two issues associated with the previous blocked approach and highlighted in MFDNN-12088: 
(i) timing inaccuracies arising from host-to-device synchronization overheads and 
(ii) performance discrepancies related to blocking `stream.wait()` calls.

Following the discussions in the RFC (PR #3393) and the PoC implementation (PR #3055), the PR implements the non-blocking verbose by leveraging the [`stream_profiler`](https://github.com/uxlfoundation/oneDNN/blob/e69577c6336a815feb351a6da0604bc56cbd07cc/src/xpu/stream_profiler.hpp#L34) functionality to obtain device measured times for primitive execution and using asynchronous callback to print profiling info in a non-blocking manner. The implementation proposes the following changes to the verbose profiling:
- enable stream profiling capabilities when `DNNL_VERBOSE=profile_exec` for GPU streams.
- calculate primitive execution times by aggregating the timing data obtained from the stream profiler at the end of primtive execution.
- use asynchronous callbacks to print verbose profiling info upon output event completion.
 
Addresses MFDNN-12088.

### Considerations
- The non-blocking verbose mode is enabled for OpenCL and SYCL GPU streams only, defaulting to a blocked implementation for CPU streams and other runtimes.
- The non-blocking verbose mode is not supported for out-of-order queues and falls back to the blocked mode for such cases.
- The asynchronous mode can be user-disabled using `DNNL_DISABLE_ASYNC_VERBOSE` to default to a blocking implementation for all cases.

### Summary of Changes:
 - commit ([0af87f0](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/0af87f0ac17247695e5eddbf6f2b460313ced9a5) ... [a780588](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/a780588e4eacb50d9a431bbcd1964f655ba7f386)): defines API to enable asynchronous mode for supported runtimes when verbose profiling is enabled (`ONEDNN_VERBOSE=profile_exec` or `ONEDNN_VERBOSE=1`). 
Adds a runtime knob `ONEDNN_VERBOSE_USE_SYNC` to force disable the asynchronous mode during verbose logging.

 - commit ([89b0693](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/89b0693eb90a58bdc2b590220108be34ad908ca5) ... [e51ee1e](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/e51ee1eef62b610549e10355e5d721da44d1abdb)): Adds support for asynchronous profiling for OpenCL GPU runtimes. Defines logic for execution tracking via OpenCL event callbacks and computing device measured exec times.

 - commit ([c4b5c0d](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/c4b5c0deede0d2ade0ecc1b47a7e41297340e58d) ... [7d6bd9b](https://github.com/uxlfoundation/oneDNN/pull/4187/changes/7d6bd9b0303456f6b28bbbf1d77f8796ba952e45)): Adds support for asynchronous profiling for SYCL GPU runtimes. Defines similar logic for execution tracking using `host_task`.

Thanks @echeresh for the suggested optimizations. 

#### Reproducer: 
```
DNNL_VERBOSE=profile_exec ./examples/primitives-matmul-cpp gpu
```

#### With `ONEDNN_VERBOSE_USE_SYNC=0`:
```
onednn_verbose,v1,info,oneDNN v3.13.0 (commit 0fadac14800ba792a9d991b93b4917509eada63b)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
onednn_verbose,v1,info,cpu,isa:Intel AVX10.1 and Intel AMX with bfloat16 and 8-bit integer support
onednn_verbose,v1,info,gpu,runtime:OpenCL
onednn_verbose,v1,info,gpu,engine,opencl device count:2 
onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,gpu,engine,1,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,primitive,exec,gpu:0,matmul,jit:gemm:any,undef,src:f32::blocked:abc::f0 wei:f32::blocked:abc::f0 bia:f32::blocked:abc::f0_mask4 dst:f32::blocked:abc::f0,attr-post-ops:eltwise_relu,,3x128x256:3x256x512,0.01904
Example passed on GPU.
```

####  With `ONEDNN_VERBOSE_USE_SYNC=1`:
```
onednn_verbose,v1,info,oneDNN v3.13.0 (commit 0fadac14800ba792a9d991b93b4917509eada63b)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
onednn_verbose,v1,info,cpu,isa:Intel AVX10.1 and Intel AMX with bfloat16 and 8-bit integer support
onednn_verbose,v1,info,gpu,runtime:OpenCL
onednn_verbose,v1,info,gpu,engine,opencl device count:2 
onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,gpu,engine,1,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,primitive,exec,gpu:0,matmul,jit:gemm:any,undef,src:f32::blocked:abc::f0 wei:f32::blocked:abc::f0 bia:f32::blocked:abc::f0_mask4 dst:f32::blocked:abc::f0,attr-post-ops:eltwise_relu,,3x128x256:3x256x512,0.473145
Example passed on GPU.
```

Related RFC: [[link](https://github.com/uxlfoundation/oneDNN/tree/amanerik/rfcs/async-verbose-mode/rfcs/20250527-async-verbose-mode)]
Related PoC and discussion: [[link](https://github.com/uxlfoundation/oneDNN/pull/3055)]

### RFC Checklist:
- [x] Have you published an RFC for the new feature?
- [ ] Was the RFC approved?

### DIFF
--- a/doc/performance_considerations/verbose.md
+++ b/doc/performance_considerations/verbose.md
@@ -119,6 +119,59 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
 uses ONEDNN_VERBOSE output to tune oneDNN code to align with
 [best practices](@ref dev_guide_inference).
 
+#### Asynchronous Verbose Mode for GPU engines
+
+When oneDNN verbose mode is enabled for primitive execution profiling, the 
+execution time for synchronous runtimes is calculated based on the wall time 
+measured before and after primitive execution. 
+For GPU runtimes, which are generally asynchronous, using synchronizing 
+methodology can lead to additional overhead due to host-to-device stream 
+synchronization on entry and on exit in the `dnnl::primitive::execute()` call.
+
+To ensure accurate tracking of timing information for GPU runtimes, the verbose 
+mode uses a non-blocking approach which prints device-measured 
+times for primitive execution instead of relying on the wall time 
+measurements. Asynchronous profiling is currently supported only for OpenCL 
+and SYCL runtimes on Intel GPUs and can be optionally disabled by using 
+`ONEDNN_VERBOSE_USE_SYNC=1` environment variable:
+
+##### Example Output (with the asynchronous mode enabled):
+~~~sh
+# DNNL_VERBOSE_USE_SYNC is 0 by default
+DNNL_VERBOSE=profile_exec ./examples/primitives-matmul-cpp gpu
+~~~
+
+~~~sh
+onednn_verbose,v1,info,oneDNN v3.13.0 (commit 0fadac14800ba792a9d991b93b4917509eada63b)
+onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
+onednn_verbose,v1,info,cpu,isa:Intel AVX10.1 and Intel AMX with bfloat16 and 8-bit integer support
+onednn_verbose,v1,info,gpu,runtime:OpenCL
+onednn_verbose,v1,info,gpu,engine,opencl device count:2 
+onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
+onednn_verbose,v1,info,gpu,engine,1,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
+onednn_verbose,v1,primitive,info,template:operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
+onednn_verbose,v1,primitive,exec,gpu:0,matmul,jit:gemm:any,undef,src:f32::blocked:abc::f0 wei:f32::blocked:abc::f0 bia:f32::blocked:abc::f0_mask4 dst:f32::blocked:abc::f0,attr-post-ops:eltwise_relu,,3x128x256:3x256x512,0.01904
+Example passed on GPU.
+~~~
+
+##### Example Output (with the asynchronous mode disabled):
+~~~sh
+DNNL_VERBOSE=profile_exec DNNL_VERBOSE_USE_SYNC=1 ./examples/primitives-matmul-cpp gpu
+~~~
+
+~~~sh
+onednn_verbose,v1,info,oneDNN v3.13.0 (commit 0fadac14800ba792a9d991b93b4917509eada63b)
+onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
+onednn_verbose,v1,info,cpu,isa:Intel AVX10.1 and Intel AMX with bfloat16 and 8-bit integer support
+onednn_verbose,v1,info,gpu,runtime:OpenCL
+onednn_verbose,v1,info,gpu,engine,opencl device count:2 
+onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
+onednn_verbose,v1,info,gpu,engine,1,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
+onednn_verbose,v1,primitive,info,template:operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
+onednn_verbose,v1,primitive,exec,gpu:0,matmul,jit:gemm:any,undef,src:f32::blocked:abc::f0 wei:f32::blocked:abc::f0 bia:f32::blocked:abc::f0_mask4 dst:f32::blocked:abc::f0,attr-post-ops:eltwise_relu,,3x128x256:3x256x512,0.473145
+Example passed on GPU.
+~~~
+
 ### Understanding why a given implementation is dispatched
 
 When performance is lower than expected, it is usually likely due to
@@ -231,12 +284,6 @@ where:
    memory is dense, the field will be empty.
 5. `extra_flags` is unspecified information that is intended for development
    purposes.
-
-@note
-When oneDNN verbose mode is enabled with GPU engines, oneDNN adds extra stream
-synchronization on entry and on exit in the dnnl::primitive::execute() call.
-The execution time is calculated based on wall time measured before and after
-primitive execution.
 
 @note
 When oneDNN verbose mode is enabled for builds with
--- a/src/common/primitive_iface.cpp
+++ b/src/common/primitive_iface.cpp
@@ -147,6 +147,7 @@ status_t primitive_execute(
 
     if (get_verbose(verbose_t::exec_profile,
                 prim_kind2_comp_kind(pd->impl()->kind()))) {
+        std::string pd_info;
         bool block_on_wait = true;
 #if DNNL_CPU_RUNTIME == DNNL_RUNTIME_THREADPOOL
         dnnl::threadpool_interop::threadpool_iface *tp;
@@ -156,12 +157,7 @@ status_t primitive_execute(
                         & dnnl::threadpool_interop::threadpool_iface::
                                 ASYNCHRONOUS);
 #endif
-        if (block_on_wait) stream->wait();
-        double start_ms = get_msec();
-        status = stream->enqueue_primitive(primitive_iface, ctx);
-        if (block_on_wait) stream->wait();
 
-        double duration_ms = get_msec() - start_ms;
         if (pd->impl()->has_runtime_dims_or_strides()) {
             // Take out mds from `ctx` here to avoid primitive_desc dependency
             // on `exec_ctx_t` type.
@@ -174,14 +170,26 @@ status_t primitive_execute(
             const auto bia_md = ctx.memory_mdw(DNNL_ARG_BIAS, pd_bia_md).md_;
             const auto pd_dst_md = pd->impl()->invariant_dst_md();
             const auto dst_md = ctx.memory_mdw(DNNL_ARG_DST, pd_dst_md).md_;
-
-            std::string info = pd->info_with_runtime_dims(
+            pd_info = pd->info_with_runtime_dims(
                     src_md, wei_md, bia_md, dst_md);
-            VPROF(start_ms, primitive, exec, VERBOSE_profile, info.c_str(),
+        } else {
+            pd_info = primitive_iface->pd()->info();
+        }
+
+        if (!stream->is_verbose_profiler_enabled()) {
+            if (block_on_wait) stream->wait();
+            double start_ms = get_msec();
+            status = stream->enqueue_primitive(primitive_iface, ctx);
+            if (block_on_wait) stream->wait();
+            double duration_ms = get_msec() - start_ms;
+            VPROF(start_ms, primitive, exec, VERBOSE_profile, pd_info.c_str(),
                     duration_ms);
         } else {
-            VPROF(start_ms, primitive, exec, VERBOSE_profile, pd->info(),
-                    duration_ms);
+            // For OpenCL/SYCL GPU streams, the verbose logs print device-
+            // measured execution times in a non-blocking manner.
+            double start_ms = get_msec();
+            status = stream->enqueue_primitive(primitive_iface, ctx);
+            CHECK(stream->run_verbose_profiler(pd_info, start_ms));
         }
     } else {
         status = stream->enqueue_primitive(primitive_iface, ctx);
--- a/src/common/stream.cpp
+++ b/src/common/stream.cpp
@@ -49,7 +49,21 @@ status_t dnnl_stream_create(
         return status::unimplemented;
     }
 
-    return engine->create_stream(stream, flags);
+    // Enables profiling capabilities to allow the verbose mode to print
+    // profiling info using device measured times.
+    // The verbose profiler state is fixed at stream initialization and does not
+    // respond to runtime changes made via set_dnnl_verbose().
+    // TODO: allow runtime control of the asynchronous verbose mode via
+    // set_dnnl_verbose()
+    const bool enable_verbose_profiler = engine->kind() == engine_kind::gpu
+            && get_verbose(verbose_t::exec_profile) && !get_verbose_sync_mode();
+
+    if (enable_verbose_profiler) { flags |= stream_flags::profiling; }
+
+    CHECK(engine->create_stream(stream, flags));
+    CHECK((*stream)->init_verbose_profiler(enable_verbose_profiler));
+
+    return status::success;
 }
 
 status_t dnnl_stream_get_engine(const stream_t *stream, engine_t **engine) {
--- a/src/common/stream.hpp
+++ b/src/common/stream.hpp
@@ -54,6 +54,10 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
     virtual dnnl::impl::status_t reset_profiling() {
         if (!is_profiling_enabled())
             return dnnl::impl::status::invalid_arguments;
+        if (is_verbose_profiler_enabled())
+            VWARN(common, stream,
+                    "resetting profiler during DNNL_VERBOSE=profile_exec can "
+                    "can cause incorrect timing logs.");
         return dnnl::impl::status::unimplemented;
     }
 
@@ -73,6 +77,21 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
 
     bool is_profiling_enabled() const { return impl_->is_profiling_enabled(); }
 
+    dnnl::impl::status_t init_verbose_profiler(bool enable_profiler) {
+        use_verbose_profiler_ = enable_profiler
+                && impl_->supports_verbose_profiling(engine_->kind());
+        return dnnl::impl::status::success;
+    }
+
+    bool is_verbose_profiler_enabled() const { return use_verbose_profiler_; }
+
+    virtual dnnl::impl::status_t run_verbose_profiler(
+            std::string &pd_info, double start_ms) const {
+        if (!is_verbose_profiler_enabled())
+            return dnnl::impl::status::invalid_arguments;
+        return dnnl::impl::status::unimplemented;
+    }
+
     virtual dnnl::impl::status_t zero_pad(const dnnl::impl::memory_t *memory,
             const dnnl::impl::exec_ctx_t &ctx);
 
@@ -91,6 +110,7 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
 protected:
     dnnl::impl::engine_t *engine_;
     std::unique_ptr<dnnl::impl::stream_impl_t> impl_;
+    bool use_verbose_profiler_;
 };
 
 #endif
--- a/src/common/stream.hpp
+++ b/src/common/stream.hpp
@@ -28,7 +28,7 @@
 
 struct dnnl_stream : public dnnl::impl::c_compatible {
     dnnl_stream(dnnl::impl::engine_t *engine, dnnl::impl::stream_impl_t *impl)
-        : engine_(engine), impl_(impl) {}
+        : engine_(engine), impl_(impl), use_verbose_profiler_(false) {}
     virtual ~dnnl_stream() = default;
 
     /** returns stream's engine */
@@ -57,7 +57,7 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
         if (is_verbose_profiler_enabled())
             VWARN(common, stream,
                     "resetting profiler during DNNL_VERBOSE=profile_exec can "
-                    "can cause incorrect timing logs.");
+                    "cause incorrect timing logs.");
         return dnnl::impl::status::unimplemented;
     }
 
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -313,6 +313,21 @@ bool get_verbose_timestamp() {
     return verbose_timestamp.get();
 }
 
+static setting_t<bool> verbose_sync_mode {false};
+bool get_verbose_sync_mode() {
+#if defined(DISABLE_VERBOSE)
+    return false;
+#endif
+    if (verbose.get() == 0) return false;
+
+    if (!verbose_sync_mode.initialized()) {
+        static bool val
+                = getenv_int_user("VERBOSE_USE_SYNC", verbose_sync_mode.get());
+        verbose_sync_mode.set(val);
+    }
+    return verbose_sync_mode.get();
+}
+
 std::ostream &operator<<(std::ostream &ss, engine_kind_t eng_kind) {
     ss << dnnl_engine_kind2str(eng_kind);
     return ss;
--- a/src/gpu/intel/ocl/kernel.cpp
+++ b/src/gpu/intel/ocl/kernel.cpp
@@ -233,6 +233,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
     }
 
     if (stream.is_profiling_enabled()) {
+        // updates the execution dependency for the primitive operation.
+        // The verbose profiler relies on ctx deps to compute primitive
+        // duration.
+        if (stream.is_verbose_profiler_enabled()) {
+            xpu::ocl::event_t::from(out_dep).events = {event};
+        }
+
         ocl_stream->profiler().register_event(
                 utils::make_unique<xpu::ocl::event_t>(std::move(event)));
     }
--- a/src/gpu/intel/ocl/stream.cpp
+++ b/src/gpu/intel/ocl/stream.cpp
@@ -120,6 +120,87 @@ void stream_t::after_exec_hook() {
     if (is_profiling_enabled()) profiler_->stop_profiling();
 }
 
+status_t stream_t::run_verbose_profiler(
+        std::string &pd_info, double start_ms) const {
+
+    // utilize the verbose profiler only for profile_exec verbose levels.
+    if (!is_verbose_profiler_enabled()) return status::invalid_arguments;
+
+    // failsafe for primitive executions without any enqueued kernels
+    auto &deps = xpu::ocl::event_t::from(ctx().get_deps());
+    if (deps.size() < 1) {
+        double duration_ms = get_msec() - start_ms;
+        VPROF(start_ms, primitive, exec, VERBOSE_profile, pd_info.c_str(),
+                duration_ms);
+        return status::success;
+    }
+
+    // captured output event acts as the anchor to track primitive execution
+    // print profiling info asynchronously
+    cl_event out_evt = deps[0].get();
+
+    if (!profiler_->stamp()) {
+        VWARN(primitive, exec,
+                "%s, profiler error: failed to record events in context",
+                pd_info.c_str());
+        VPROF(start_ms, primitive, exec, VERBOSE_profile, pd_info.c_str(), 0.f);
+        return status::success;
+    }
+
+    auto *ocl_profiler
+            = utils::downcast<xpu::ocl::stream_profiler_t *>(profiler_.get());
+
+    // The verbose callback uses a snapshot of queued primitive events
+    // to compute execution timing in a thread-safe manner.
+    std::vector<cl_event> evt_snap;
+    CHECK(ocl_profiler->extract_primitive_events(evt_snap));
+
+    struct payload_t {
+        double start;
+        xpu::ocl::stream_profiler_t *prof;
+        std::string info_str;
+        std::vector<cl_event> evt_snap;
+    };
+
+    std::unique_ptr<payload_t> payload(new payload_t());
+    payload->prof = ocl_profiler;
+    payload->info_str = pd_info;
+    payload->start = start_ms;
+    payload->evt_snap = std::move(evt_snap);
+
+    // The prompt ensures the verbose headers are printed if they aren't
+    // already. Printing the headers asynchronously during the callback can
+    // result in access failures when printing engine-specific info.
+    verbose_printf(verbose_t::exec_profile, "\r");
+
+    cl_int err = xpu::ocl::clSetEventCallback(
+            out_evt, CL_COMPLETE, [](cl_event ev, cl_int, void *user) {
+        std::unique_ptr<payload_t> hold(static_cast<payload_t *>(user));
+        double duration_ms = 0.0;
+
+        if (hold->prof) {
+            hold->prof->get_aggregate_exec_timing(duration_ms, hold->evt_snap);
+        } else {
+            VWARN(primitive, exec, "%s, profiler error: profiler absent",
+                    hold->info_str.c_str());
+        }
+        VPROF(hold->start, primitive, exec, VERBOSE_profile,
+                hold->info_str.c_str(), duration_ms);
+    }, payload.get());
+
+    if (err != CL_SUCCESS) {
+        VWARN(primitive, exec,
+                "%s, profiler error: failed to set event callback",
+                pd_info.c_str());
+        VPROF(start_ms, primitive, exec, VERBOSE_profile, pd_info.c_str(), 0.f);
+        return status::success;
+    }
+
+    (void)payload.release();
+
+    return status::success;
+}
+
 status_t stream_t::copy(const memory_storage_t &src,
         const memory_storage_t &dst, size_t size, const xpu::event_t &deps,
         xpu::event_t &out_dep) {
--- a/src/gpu/intel/ocl/stream.cpp
+++ b/src/gpu/intel/ocl/stream.cpp
@@ -173,6 +173,8 @@ status_t stream_t::run_verbose_profiler(
     // result in access failures when printing engine-specific info.
     verbose_printf(verbose_t::exec_profile, "\r");
 
+    ocl_profiler->start_async_callback_tracking();
+
     cl_int err = xpu::ocl::clSetEventCallback(
             out_evt, CL_COMPLETE, [](cl_event ev, cl_int, void *user) {
         std::unique_ptr<payload_t> hold(static_cast<payload_t *>(user));
@@ -186,9 +188,12 @@ status_t stream_t::run_verbose_profiler(
         }
         VPROF(hold->start, primitive, exec, VERBOSE_profile,
                 hold->info_str.c_str(), duration_ms);
+
+        if (hold->prof) { hold->prof->end_async_callback_tracking(); }
     }, payload.get());
 
     if (err != CL_SUCCESS) {
+        ocl_profiler->end_async_callback_tracking();
         VWARN(primitive, exec,
                 "%s, profiler error: failed to set event callback",
                 pd_info.c_str());
--- a/src/xpu/ocl/stream_impl.cpp
+++ b/src/xpu/ocl/stream_impl.cpp
@@ -178,6 +178,13 @@ status_t stream_impl_t::copy(impl::stream_t *stream,
     }
 
     if (is_profiling_enabled()) {
+        // updates the execution dependency for the copy operation.
+        // The verbose profiler relies on ctx deps to compute primitive
+        // duration.
+        if (stream->is_verbose_profiler_enabled()) {
+            xpu::ocl::event_t::from(out_dep).events = {out_event};
+        }
+
         auto ocl_event = utils::make_unique<xpu::ocl::event_t>(
                 std::vector<xpu::ocl::wrapper_t<cl_event>> {out_event});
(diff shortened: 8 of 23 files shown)


echeresh:
> Because benchdnn also uses the stream_profiler functionality in perf mode, the verbose profiler is disabled when benchdnn is run in perf mode to record aggregate timing info.

There is a similar conflict with profiling API: https://uxlfoundation.github.io/oneDNN/dev_guide_experimental.html#onednn-experimental-profiling making usage of verbose functionality and profiling API exclusive.

I think this could be avoided by introducing a new `stream_profiler_t` object (dedicated only for verbose-related timing) and extending `get_profiling_data()` and similar APIs. This seems to be a better option to me as we will be able to keep benchmarking, verbose functionality and profiling API working independently, without extra cross-conditions.

@avmanerikar Do you see any issues with that?

avmanerikar:
> @avmanerikar Do you see any issues with that?

The trouble with having multiple profiler objects is that they will be tracking the same information and utilizing the same APIs throughout the stream lifetime.
The conflicting behavior was due to the verbose profiling data being reset after primitive execution leaving an empty slate for profiling queries post-execution.
I have adjusted the design to removing the resetting action which allows it to be used with the benchdnn perf mode and the experimental profiler.

echeresh:
> I have adjusted the design to removing the resetting action which allows it to be used with the benchdnn perf mode and the experimental profiler.

@avmanerikar Thanks for the update. Is concurrent access from the event callback and from [profiling API](https://uxlfoundation.github.io/oneDNN/dev_guide_experimental.html#onednn-experimental-profiling) a concern here?

avmanerikar:
> @avmanerikar Thanks for the update. Is concurrent access from the event callback and from [profiling API](https://uxlfoundation.github.io/oneDNN/dev_guide_experimental.html#onednn-experimental-profiling) a concern here?

This would be a concern if the profiler is reset by the experimental API while the verbose profiler is yet to print the timing data. But since the verbose profiler calculates and prints the timing data during the course of primitive execution, this is unlikely to happen.

echeresh:
> This would be a concern if the profiler is reset by the experimental API while the verbose profiler is yet to print the timing data. But since the verbose profiler calculates and prints the timing data during the course of primitive execution, this is unlikely to happen.

That makes sense. Alas, such rare issues are usually the hardest to debug. Can we somehow make this scenario to "never happen" instead of "unlikely to happen"?

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

avmanerikar:
> That makes sense. Alas, such rare issues are usually the hardest to debug. Can we somehow make this scenario to "never happen" instead of "unlikely to happen"?

Apologies for the late reply. I have updated the API so that the stream profiler is aware of asynchronous logging for such scenarios and manages the profiler lifetime accordingly. Also, a warning is printed to notify the user when the profiler is reset in verbose profiling mode (this will be the case when the experimental profiler is enabled and manually reset).

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

dzarukin:
We have a series of variables with `DNNL_VERBOSE` prefix:
```
verbose.cpp:  302:                      = getenv_int_user("VERBOSE_TIMESTAMP", verbose_timestamp.get());
logging.cpp:  28:           : logfile_path_(getenv_string_user("VERBOSE_LOGFILE"))
logging.cpp:  30:           , console_flag_(getenv_int_user("VERBOSE_LOG_WITH_CONSOLE", 0)) {
logging.cpp:  42:               logfile_size_ = getenv_int_user("VERBOSE_LOGFILE_SIZE");
logging.cpp:  45:               num_logfiles_ = getenv_int_user("VERBOSE_NUM_LOGFILES");
```
Suggesting to stick to the same prefix and rename `DNNL_DISABLE_ASYNC_VERBOSE` into `DNNL_VERBOSE_USE_SYNC` with `0` by default (inverse the logic, might require some code changes).

doc/performance_considerations/verbose.md
```diff
@@ -119,6 +119,26 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
...
+
+To ensure accurate tracking of timing information on GPU engines, the verbose 
+mode uses a non-blocking, asynchronous approach which prints device-measured 
+times for primitive execution instead of relying on wall time 
+measurements. Asynchronous profiling is currently supported only for OpenCL 
+and SYCL GPU runtimes and can be optionally disabled using 
+`ONEDNN_DISABLE_ASYNC_VERBOSE=1`:
+
+~~~sh
+ONEDNN_VERBOSE=profile_exec ONEDNN_DISABLE_ASYNC_VERBOSE=1 ./benchdnn --matmul --engine=gpu 64x256:256x64
```
dzarukin: It probably would look better if there's a comparison in the output from the same machine with async enabled/disabled.

src/common/stream.cpp
```diff
@@ -49,6 +49,13 @@ status_t dnnl_stream_create(
         return status::unimplemented;
     }
 
+    // enables profiling capabilities to allow the verbose mode to print
+    // profiling info using device measured times
+    const bool enable_verbose_profiler = engine->kind() == engine_kind::gpu
+            && get_verbose(verbose_t::exec_profile)
+            && (getenv_int_user("DISABLE_ASYNC_VERBOSE", 0) == 0);
```
dzarukin: Please use a single function, like `get_verbose` to obtain the variable value a single time and at a single place.

src/common/stream.cpp
```diff
@@ -49,6 +49,13 @@ status_t dnnl_stream_create(
         return status::unimplemented;
     }
 
+    // enables profiling capabilities to allow the verbose mode to print
+    // profiling info using device measured times
+    const bool enable_verbose_profiler = engine->kind() == engine_kind::gpu
```
dzarukin, line 58: It seems this won't work if the application relies on `dnnl_set_verbose` API set after the stream was created.
I'm not sure if it's an important case to support at this point but putting a note/TODO in the comment wouldn't hurt to document this limitation (if it turns true, may check manually through benchdnn).

src/xpu/stream_profiler.hpp
```diff
@@ -58,6 +58,13 @@ struct stream_profiler_t {
             uint64_t *data) const
             = 0;
 
+    virtual const xpu::event_t *peek_last_event() const {
```
dzarukin: Doesn't look like this method must be `virtual`.

src/xpu/stream_profiler.hpp
```diff
@@ -58,6 +58,13 @@ struct stream_profiler_t {
...
+        if (events_.empty()) return nullptr;
+        const auto &tracked = events_.back();
+        if (!tracked.event) return nullptr;
+        return tracked.event.get();
```
dzarukin: It seems the event can handle its emptiness themselves:
```
return events_.back().event.get();
```

src/xpu/ocl/stream_impl.cpp
```diff
@@ -178,6 +178,10 @@ status_t stream_impl_t::copy(impl::stream_t *stream,
     }
 
     if (is_profiling_enabled()) {
+        if (stream->is_verbose_profiler_enabled()) {
+            xpu::ocl::event_t::from(out_dep).events = {out_event};
+        }
```
dzarukin, line 186: Could you, please, explain several things:
1) **Why** does it update `out_dep`?
2) Why does it update `out_dep` **before** `stream_profiler`?
3) If **before** part is not important, can it be merged with line 190.

All comments helping read this would be appreciated if appeared in the code.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,130 @@ void stream_t::after_exec_hook() {
     if (is_profiling_enabled()) profiler_->stop_profiling();
 }
 
+status_t stream_t::set_verbose_profiler(int &last_entry, bool &use_profiler) {
+
+    // This is to prepares the stream profiler for current primitive
```
dzarukin: Minor: typo;

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,130 @@ void stream_t::after_exec_hook() {
...
+    int num_entries;
+    get_profiling_data(profiling_data_kind::time, &num_entries, nullptr);
+    if (!num_entries) return status::success;
+
+    last_entry = num_entries;
+
+    const xpu::event_t *last_evt = profiler_->peek_last_event();
+    if (!last_evt) return status::success;
+
+    auto *ocl_ev_wrap = utils::downcast<const xpu::ocl::event_t *>(last_evt);
```
dzarukin: Nit: there's a `from` interface to make this cast.

src/common/primitive_iface.cpp
```diff
@@ -140,14 +136,33 @@ status_t primitive_execute(
...
+            pd_info = primitive_iface->pd()->info();
+        }
+
+        bool use_stream_profiler = false;
+        int profiler_num_events = 0;
+
+        if (stream->is_verbose_profiler_enabled()) {
+            CHECK(stream->set_verbose_profiler(
+                    profiler_num_events, use_stream_profiler));
+        }
```
dzarukin: If I'm reading it right, the very first primitive executed won't have any dependencies/events, will keep `use_stream_profiler = false` (based on the logic from `set_verbose_profiler` and goes to a `blocked` version with synchronizations?

I'd also suggest  consider handling errors somehow differently from the current approach of returning `success` for situations that don't seem to be successful...

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,130 @@ void stream_t::after_exec_hook() {
...
+        duration_ms *= 1e-6;
+
+        VPROF(hold->start, primitive, exec, VERBOSE_profile,
+                hold->info_str.c_str(), duration_ms);
+
+        clReleaseEvent(ev);
+    }, payload);
+
+    if (err != CL_SUCCESS) {
+        delete payload;
```
dzarukin: Maybe put `payload` into `std::unique_ptr` at the construction time to avoid dealing with `delete`?

src/common/stream.hpp
```diff
@@ -73,6 +78,25 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
 
     bool is_profiling_enabled() const { return impl_->is_profiling_enabled(); }
 
+    bool is_verbose_profiler_enabled() const {
+        return impl_->is_profiling_enabled()
+                && dnnl::impl::get_verbose(dnnl::impl::verbose_t::exec_profile);
+    }
+
+    virtual dnnl::impl::status_t set_verbose_profiler(
+            int &last_entry, bool &use_profiler) {
```
dzarukin: Looks like this can be done `const` as I can't figure what it's **setting** inside the stream. If this indeed the case, please consider renaming the method.

doc/performance_considerations/verbose.md
avmanerikar: Done. Thanks!

src/common/stream.cpp
avmanerikar: `dnnl_stream::is_verbose_profiler_enabled()` can be used to query the verbose profiler state, but it is only available after the stream has been created. In the above case, the decision has to be made earlier and depends on how the stream is created (directly or from an existing OpenCL/SYCL queue), so the checks cannot be fully centralized into a single helper at this point.

src/common/stream.cpp
avmanerikar: added TODO. Thanks!

src/xpu/stream_profiler.hpp
avmanerikar: Fixed. Thanks!

src/xpu/stream_profiler.hpp
avmanerikar: Method has been modified in the latest update.

src/common/primitive_iface.cpp
avmanerikar: I have simplified the function `init_verbose_profiler()` (I changed the name from `set_verbose_profiler` for clarity) and added comments to describe the flow of operation. The function is altered to track the last event logged by the profiler instead of the total number of events. (Using `num_events` is unreliable when there are multiple primitives executive on the stream and the profiling data is accessed asynchronously).

src/gpu/intel/ocl/stream.cpp
avmanerikar: Fixed. Thanks!

src/common/stream.hpp
avmanerikar: The method has been modified in the latest update.

src/xpu/ocl/stream_impl.cpp
avmanerikar: >Why does it update out_dep?

`out_dep` is the execution dependency stored in the context (`ctx().get_deps()`). Setting it to `out_event` publishes the copy's completion event so the async verbose profiler can use it as the end anchor for timing.

>Why does it update `out_dep` before stream_profiler?

It's done first to keep the act of publishing the dependency separated from the profiling action and also to avoid any risk of using `out_event` after it has been moved into profiler storage.

avmanerikar:
> Suggesting to stick to the same prefix and rename `DNNL_DISABLE_ASYNC_VERBOSE` into `DNNL_VERBOSE_USE_SYNC` with `0` by default (inverse the logic, might require some code changes).

Done. The variable name is changed to `DNNL_VERBOSE_USE_SYNC`. The documentation has also been updated to that effect. 
(Another suggestion for the name could also be `DNNL_VERBOSE_PRINT_DEVICE_TIMINGS` since the mode logs device-measured times instead of wall time)

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

src/common/stream.cpp
dzarukin: I'm sorry but I'm not sure I'm following the explanation...

According to oneDNN practices and model usage, the env variable must be read only once. To support this flow, a `setting_t` class must be used (same as `verbose`). Currently, if the env variable value changes during the run, the behavior would be inconsistent. A single function I mentioned would be checking if the proposed class is initialized and read the value of env variable once on the contrary of current solution. It doesn't seem to me to contradict with the flow you've described.

src/xpu/ocl/stream_impl.cpp
dzarukin: It's nice to document such details in a comment in all relevant places as them usually lost with time. Could you please add some?

avmanerikar: Done. Thanks!

src/common/stream.cpp
avmanerikar: My bad. I misinterpreted the question - the update has been added.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,137 @@ void stream_t::after_exec_hook() {
...
+                        "%s, profiling error: could not find stamp for output "
+                        "event timing may be inaccurate",
+                        hold->info_str.c_str());
+            }
+        }
+        double duration_ms = 0.0;
+
+        // aggregate execution times are calculated from the start and end times
+        // of the first and last queued events for the primitive respectively
+        if (end_stamp > hold->lastev) {
```
dzarukin: Seems like this condition is (should be) handled inside the function itself. Applies to the other spot, too.

src/xpu/sycl/stream_profiler.cpp
```diff
@@ -70,6 +70,38 @@ status_t stream_profiler_t::get_info(profiling_data_kind_t data_kind,
     return xpu::stream_profiler_t::get_info_impl(stamp2entry, data_kind, data);
 }
 
+status_t stream_profiler_t::get_aggregate_exec_timing(
+        uint64_t start, uint64_t end, double &duration_ms) const {
+    using namespace ::sycl::info;
+    duration_ms = 0.0;
+    if (end < start) return status::success;
```
dzarukin: Maybe `<=` ?

src/common/verbose.cpp
```diff
@@ -305,6 +305,21 @@ bool get_verbose_timestamp() {
     return verbose_timestamp.get();
 }
 
+static setting_t<bool> verbose_async_mode {true};
+bool get_verbose_async_mode() {
+#if defined(DISABLE_VERBOSE)
+    return false;
+#endif
+    if (verbose.get() == 0) return true;
```
dzarukin: Could you, please, explain the logic behind `true` when verbose is off?

It looks like to me that `disable_verbose_profiler` (or as I asked to rename it to `enable_verbose_profiler`) would instruct the stream to collect data in this case which seems to be as unnecessary overhead.

src/xpu/ocl/stream_impl.hpp
```diff
@@ -82,9 +82,25 @@ class stream_impl_t : public impl::stream_impl_t {
...
+                               : stream_flags::in_order;
+
+        // Profiling capabilities are enabled on the stream to allow
+        // printing the info in verbose profiling mode
+        // The verbose profiler state is fixed at stream creation and does not
+        // respond to runtime changes made via set_dnnl_verbose().
+        // TODO: allow runtime control of the asynchronous verbose mode via
+        // set_dnnl_verbose()
+        const bool disable_verbose_profiler
+                = is_ooo_queue || !get_verbose_async_mode();
```
dzarukin: Is it possible to invert the variable to become positive?

```
const bool enable_verbose_profiler
        = !is_ooo_queue && get_verbose_async_mode();
```

doc/performance_considerations/verbose.md
```diff
@@ -119,6 +119,60 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
 uses ONEDNN_VERBOSE output to tune oneDNN code to align with
 [best practices](@ref dev_guide_inference).
 
+#### Asynchronous Verbose Mode for GPU engines
+
+When oneDNN verbose mode is enabled for primitive execution profiling, the 
+execution time is calculated based on wall time measured before and after
```
dzarukin:
```suggestion
execution time for synchronous runtimes is calculated based on the wall time measured before and after
```

doc/performance_considerations/verbose.md
```diff
@@ -119,6 +119,60 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
...
+primitive execution. 
+For GPU engines, this can add additional overhead due to host-to-device
```
dzarukin:
```suggestion
For GPU runtimes, which are generally asynchronous, using synchronizing methodology can lead to additional overhead due to host-to-device 
```

doc/performance_considerations/verbose.md
```diff
@@ -119,6 +119,60 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
...
+For GPU engines, this can add additional overhead due to host-to-device 
+stream synchronization on entry and on exit in the `dnnl::primitive::execute()` 
+call.
+
+To ensure accurate tracking of timing information on GPU engines, the verbose 
+mode uses a non-blocking, asynchronous approach which prints device-measured 
+times for primitive execution instead of relying on wall time 
+measurements. Asynchronous profiling is currently supported only for OpenCL 
+and SYCL GPU runtimes and can be optionally disabled using 
+`ONEDNN_VERBOSE_USE_SYNC=1`:
```
dzarukin:
```suggestion
To ensure accurate tracking of timing information for GPU runtimes, the verbose 
mode uses a non-blocking approach which prints device-measured 
times for primitive execution instead of relying on the wall time 
measurements. Asynchronous profiling is currently supported only for OpenCL 
and SYCL GPU runtimes and can be optionally disabled by using 
`ONEDNN_VERBOSE_USE_SYNC=1` environment variable:
```

src/common/primitive_iface.cpp
```diff
@@ -140,14 +136,30 @@ status_t primitive_execute(
...
+            if (block_on_wait) stream->wait();
+            double start_ms = get_msec();
+            status = stream->enqueue_primitive(primitive_iface, ctx);
+            if (block_on_wait) stream->wait();
+            double duration_ms = get_msec() - start_ms;
+            VPROF(start_ms, primitive, exec, VERBOSE_profile, pd_info.c_str(),
                     duration_ms);
+        } else {
+            // For OpenCL/SYCL GPU streams, the verbose logs print device-
+            // measured exeuction times in an asynchronous, non-blocking manner.
```
dzarukin:
```suggestion
            // measured execution times in a non-blocking manner.
```

src/gpu/intel/ocl/stream.cpp
avmanerikar: Yes it does. This was indeed overkill.

src/xpu/sycl/stream_profiler.cpp
avmanerikar: Fixed.

src/xpu/ocl/stream_impl.hpp
avmanerikar: Done.

src/common/verbose.cpp
avmanerikar: The asynchronous mode is expected to be enabled by default whenever verbose mode is enabled - with verbose disabled, its state will not matter. Changing it to `false` as it is a more natural setting.

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

doc/performance_considerations/verbose.md
```diff
@@ -119,6 +119,60 @@ Please see the profiling example [here](@ref performance_profiling_cpp), as it
...
+To ensure accurate tracking of timing information for GPU runtimes, the verbose 
+mode uses a non-blocking approach which prints device-measured 
+times for primitive execution instead of relying on the wall time 
+measurements. Asynchronous profiling is currently supported only for OpenCL 
+and SYCL GPU runtimes and can be optionally disabled by using 
+`ONEDNN_VERBOSE_USE_SYNC=1` environment variable:
+
+##### Example Output (with the asynchronous mode enabled):
+~~~sh
+DNNL_VERBOSE=profile_exec DNNL_VERBOSE_USE_SYNC=0 ./examples/primitives-matmul-cpp gpu
```
echeresh: Nit: can drop `DNNL_VERBOSE_USE_SYNC=0` and add comment to additionally demonstrate what is the default behavior.

```suggestion
# DNNL_VERBOSE_USE_SYNC is 0 by default
DNNL_VERBOSE=profile_exec ./examples/primitives-matmul-cpp gpu
```

src/common/verbose.cpp
```diff
@@ -312,6 +312,21 @@ bool get_verbose_timestamp() {
...
+static setting_t<bool> verbose_async_mode {true};
+bool get_verbose_async_mode() {
+#if defined(DISABLE_VERBOSE)
+    return false;
+#endif
+    if (verbose.get() == 0) return false;
+
+    if (!verbose_async_mode.initialized()) {
+        static bool val
+                = getenv_int_user("VERBOSE_USE_SYNC", verbose_async_mode.get());
```
echeresh: Should it be flipped?

```suggestion
                = getenv_int_user("VERBOSE_USE_SYNC", !verbose_async_mode.get());
```

src/common/stream.hpp
```diff
@@ -73,6 +77,25 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
 
     bool is_profiling_enabled() const { return impl_->is_profiling_enabled(); }
 
+    bool is_verbose_profiler_enabled() const {
+        return impl_->is_profiling_enabled()
+                && dnnl::impl::get_verbose(dnnl::impl::verbose_t::exec_profile);
```
echeresh: This condition doesn't include `get_verbose_async_mode()` which is also checked in `src/common/stream.cpp`. I think these should be aligned.

But looking at @dzarukin comment, I think a better option might be to track a separate flag in the stream - something like `bool with_verbose_profiler_`. Why: if verbose is updated at run time then `get_verbose()` result here can be different from `get_verbose()` during stream initialization. To make behavior more consistent, we can do query at stream initialization and avoid consequent queries to `get_verbose()`.

src/common/verbose.cpp
echeresh: As a more general comment: I would align environment variable with the flag name. Doing sync <-> async mapping between these makes things harder.

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,135 @@ void stream_t::after_exec_hook() {
     if (is_profiling_enabled()) profiler_->stop_profiling();
 }
 
+status_t stream_t::init_verbose_profiler(uint64_t &last_entry) const {
+
+    // Initialization prepares the stream profiler to collect profiling
+    // data for current primitive execution without affecting any ongoing
+    // events. This includes
```
echeresh: Should it be completed - the comment ends abruptly.

doc/performance_considerations/verbose.md
avmanerikar: Applied. Thanks!

src/common/verbose.cpp
avmanerikar: Applied. Thanks!

src/gpu/intel/ocl/stream.cpp
avmanerikar: Updated in the latest changes.

src/common/stream.cpp
```diff
@@ -49,6 +49,16 @@ status_t dnnl_stream_create(
         return status::unimplemented;
     }
 
+    // Enables profiling capabilities to allow the verbose mode to print
+    // profiling info using device measured times.
+    // The verbose profiler state is fixed at stream creation and does not
+    // respond to runtime changes made via set_dnnl_verbose().
+    // TODO: allow runtime control of the asynchronous verbose mode via
+    // set_dnnl_verbose()
+    const bool enable_verbose_profiler = engine->kind() == engine_kind::gpu
```
echeresh, line 58: Looks like this has to be enabled only for Intel runtimes, e.g. generic SYCL doesn't support profiling:

https://github.com/uxlfoundation/oneDNN/blob/23eb67647053c1360123dafa488a627d319e4aee/src/gpu/generic/sycl/stream.cpp#L33-L35

src/gpu/intel/ocl/stream.cpp
```diff
@@ -120,6 +120,99 @@ void stream_t::after_exec_hook() {
...
 }
 
+status_t stream_t::run_verbose_profiler(
+        std::string &pd_info, double start_ms) const {
+
+    // utilize the verbose profiler only for profile_exec verbose levels.
+    if (!is_verbose_profiler_enabled()) return status::invalid_arguments;
+
+    // Captured output event acts as the anchor to track primitive execution
+    cl_event out_evt = get_output_event();
```
echeresh: Please double check this place, some primitives don't execute kernels, e.g.:

```
$ ONEDNN_VERBOSE=1 ./build/tests/benchdnn/benchdnn -v5 --engine=gpu --eltwise 0
run: --eltwise --engine=gpu --alg=relu --alpha=0 --beta=0 0
benchdnn: .../src/xpu/ocl/stream_impl.cpp:285: const dnnl::impl::xpu::ocl::wrapper_t<_cl_event*>& dnnl::impl::xpu::ocl::stream_impl_t::get_output_event() const: Assertion `deps.size() == 1' failed.
Aborted (core dumped)

```

src/xpu/ocl/stream_impl.hpp
```diff
@@ -82,9 +82,25 @@ class stream_impl_t : public impl::stream_impl_t {
...
+        // printing the info in verbose profiling mode
+        // The verbose profiler state is fixed at stream creation and does not
+        // respond to runtime changes made via set_dnnl_verbose().
+        // TODO: allow runtime control of the asynchronous verbose mode via
+        // set_dnnl_verbose()
+        const bool enable_verbose_profiler
+                = !is_ooo_queue && !get_verbose_sync_mode();
+
+        if (get_verbose(verbose_t::exec_profile) && enable_verbose_profiler) {
+            props |= CL_QUEUE_PROFILING_ENABLE;
```
echeresh: `props` is the queried properties so we shouldn't update them.

Also please consider the scenario when the user-passed queue doesn't have the profiling flag. What happens in this case?

src/common/stream.hpp
avmanerikar: Done . Thanks!

src/common/stream.cpp
avmanerikar: Applied. The stream initialization now checks for the corresponding runtimes before enabling the verbose mode.

src/xpu/ocl/stream_impl.hpp
avmanerikar: Updated in the latest change. In the current scenario, this is avoided as the verbose mode is not enabled during flag initialization but during stream initialization.

src/gpu/intel/ocl/stream.cpp
avmanerikar: Corrected. Thanks!

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

src/xpu/ocl/stream_profiler.cpp
```diff
@@ -75,6 +75,50 @@ status_t stream_profiler_t::get_info(profiling_data_kind_t data_kind,
...
+    return status::success;
+}
+
+status_t stream_profiler_t::extract_primitive_events(
+        std::vector<cl_event> &evt_snap) {
+    evt_snap.clear();
+
+    // During the course of primitive execution, the current stamp marks the
+    // events enqueued for the current primitive.
+    for (auto it = events_.rbegin(); it != events_.rend(); ++it) {
```
echeresh, line 108: `events_` can be updated from a different thread, if the same stream is used concurrently for other submissions.

In that case this is a race condition.

src/gpu/intel/ocl/kernel.cpp
```diff
@@ -233,6 +233,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
...
 
     if (stream.is_profiling_enabled()) {
+        // updates the execution dependency for the primitive operation.
+        // The verbose profiler relies on ctx deps to compute primitive
+        // duration.
+        if (stream.is_verbose_profiler_enabled()) {
+            xpu::ocl::event_t::from(out_dep).events = {event};
+        }
+
         ocl_stream->profiler().register_event(
```
echeresh, line 243: Random place: I didn't find if `events_` are ever cleared during the stream lifetime. It might be a problem for long runs with enabled verbose - we need some logic to clear up synced events to limit memory usage growth.

src/xpu/ocl/stream_profiler.cpp
```diff
@@ -75,6 +75,50 @@ status_t stream_profiler_t::get_info(profiling_data_kind_t data_kind,
...
+
+    // computation of execution timings is self-contained because of the event
+    // snapshot and immune to concurrent updates to profiler event list
+    for (auto ev : evt_snap) {
+        cl_ulong evbeg, evend;
+        OCL_CHECK(xpu::ocl::clGetEventProfilingInfo(ev,
+                CL_PROFILING_COMMAND_START, sizeof(evbeg), &evbeg, nullptr));
+        OCL_CHECK(xpu::ocl::clGetEventProfilingInfo(
+                ev, CL_PROFILING_COMMAND_END, sizeof(evend), &evend, nullptr));
+        agg_start = agg_start == 0 ? evbeg : std::min(agg_start, evbeg);
```
echeresh: `evbeg` can be theoretically a zero..

I would assign `agg_start = UINT64_MAX` and do `agg_start = std::min(evbeg, agg_start)` for simplicity

src/common/stream.hpp
```diff
@@ -91,6 +110,7 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
 protected:
     dnnl::impl::engine_t *engine_;
     std::unique_ptr<dnnl::impl::stream_impl_t> impl_;
+    bool use_verbose_profiler_;
```
echeresh, line 113: Just in case: to avoid risk of uninitialized access:

```suggestion
    bool use_verbose_profiler_ = false;
```

src/common/stream.hpp
```diff
@@ -54,6 +54,10 @@ struct dnnl_stream : public dnnl::impl::c_compatible {
     virtual dnnl::impl::status_t reset_profiling() {
         if (!is_profiling_enabled())
             return dnnl::impl::status::invalid_arguments;
+        if (is_verbose_profiler_enabled())
+            VWARN(common, stream,
+                    "resetting profiler during DNNL_VERBOSE=profile_exec can "
+                    "can cause incorrect timing logs.");
```
echeresh:
```suggestion
                    "cause incorrect timing logs.");
```

src/common/stream.hpp
avmanerikar: Applied. Thanks!

src/gpu/intel/ocl/kernel.cpp
avmanerikar: It is possible to clear `events_` as part of `before_exec_hook()` since the callbacks operate on the event snapshot and are not affected by changes to events_. But doing so affects concurrent use of the profiler with the benchdnn perf mode and with the experimental profiler. In both cases, the profiling info is collected post-execution and will require all the `events_` be retained.

src/xpu/ocl/stream_profiler.cpp
avmanerikar: Fixed with `std::lock_guard<std::recursive_mutex>` for the traversal - the lock duration is minimal and since the snapshot is extracted before the callback is set, it avoids blocking during GPU execution.

src/xpu/ocl/stream_profiler.cpp
avmanerikar: Applied. Thanks!

src/common/stream.hpp
avmanerikar: Applied. Thanks!

avmanerikar:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

avmanerikar:
Closing in favor of  alternative polling-based approach in PR #5102 - the method relies on a periodic event polling to record and log profiling info. The motivation to using this approach is to have a unified implementation for the asynchronous mode that does not rely on runtime-specific callback APIs.

The implementation addresses points that the callback-based approach fails to triage:
- Event-polling does not rely on runtime-specific callback API (required for supporting L0 runtime).
- The callback method does not triage [MFDNN-14331](https://jira.devtools.intel.com/browse/MFDNN-14331) for SYCL graphs. 

##### Failing reproducer for SYCL Graph:
```
DNNL_VERBOSE=1 .\tests\benchdnn\benchdnn.exe --eltwise --engine=gpu --execution-mode=graph 1
```
- Extension to CPU Threadpool is difficult with callback-based method.

PR #4188, opened 2025-10-22
cpu: x64: prepare some implementations for asynchronous runtime

dzarukin:
Changes breakdown:
* avx2 conv sets the proper order of operations. When lambdas to be changed to capture by copy, it becomes important.
* Making some objects `const &` to avoid their copies during function calls.
* Initializing palettes per thread and inside a thread. Relying on a single configuration done in a submitter thread is prohibited.
* Matmul changes are styling to minimize the change related to brgmm_ctx, which would be wrapped into a `shared_ptr` since it keeps tons of local stuff and to re-use the rest stuff without modification, a new pointer will be dereferenced inside the lambda.


dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_matmul

PR #4189, opened 2025-10-22
cpu: x64: fix Coverity issues

xuxinzen:
Fixes Coverity issue 7592466, 7895938

### DIFF
--- a/src/cpu/x64/jit_brgemm_inner_product.cpp
+++ b/src/cpu/x64/jit_brgemm_inner_product.cpp
@@ -306,7 +306,8 @@ status_t brgemm_inner_product_fwd_t<isa>::execute_forward(
             brgemm_palettes_.maybe_tile_configure(
                     is_amx, prev_ker_idx, brg_ker_ic_tail_idx);
 
-            int ic_block = gemm_batch * ic_blocks_per_batch;
+            const dim_t ic_block
+                    = static_cast<dim_t>(gemm_batch) * ic_blocks_per_batch;
             addr_batch[0].ptr.A = src
                     + blk_off(src_d, n, ic + ic_block * jbgp.ic_block, kd, kh,
                             kw);
--- a/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
@@ -5469,7 +5469,7 @@ void jit_brgemm_matmul_copy_b_cvt_bf16_t<Vmm>::get_zero_points(
         const auto zp_dt_sz = types::data_type_size(zp_dt);
         const auto elems_per_byte
                 = one_of(zp_dt, data_type::s4, data_type::u4) ? 2 : 1;
-        return n_idx * zp_dt_sz / elems_per_byte;
+        return static_cast<size_t>(n_idx) * zp_dt_sz / elems_per_byte;
     };
 
     const auto addr = maybe_EVEX_compress_addr(reg_wei_zp, base_offset(n));


src/cpu/x64/jit_brgemm_inner_product.cpp
```diff
@@ -306,7 +306,7 @@ status_t brgemm_inner_product_fwd_t<isa>::execute_forward(
             brgemm_palettes_.maybe_tile_configure(
                     is_amx, prev_ker_idx, brg_ker_ic_tail_idx);
 
-            int ic_block = gemm_batch * ic_blocks_per_batch;
+            const dim_t ic_block = gemm_batch * ic_blocks_per_batch;
```
asimonov1: I'm not sure about the context. If both `gemm_batch` and `ic_blocks_per_batch` are `int`s and can be "large" then this does not help (as then `(gemm_batch * ic_blocks_per_batch)` is `int`). But if they are "small" then this is ok.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -610,6 +610,10 @@ dim_t jit_brgemm_kernel_t<Wmm>::bd_compensation_offset(
 template <typename Wmm>
 dim_t jit_brgemm_kernel_t<Wmm>::wei_scales_offset(
         dim_t ld, bool is_tail) const noexcept {
+    if (types::data_type_size(brg.dt_wei_scales) <= 0) {
+        assert("Unknown wei scales data type");
```
asimonov1: String literal is the same as always `true` here.

xuxinzen: I think you're referring to matmul_amx_blocking_params_macro_t::skip_extendable_k(), and updated it.

src/cpu/x64/jit_brgemm_inner_product.cpp
xuxinzen: Updated it to `const dim_t ic_block = static_cast<dim_t>(gemm_batch) * ic_blocks_per_batch;`

xuxinzen:
make test

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
asimonov1: If you want to inform about the error then it must be (as one option)
```suggestion
        assert(!"Unknown wei scales data type");
```

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -610,6 +610,10 @@ dim_t jit_brgemm_kernel_t<Wmm>::bd_compensation_offset(
 template <typename Wmm>
 dim_t jit_brgemm_kernel_t<Wmm>::wei_scales_offset(
         dim_t ld, bool is_tail) const noexcept {
+    if (types::data_type_size(brg.dt_wei_scales) <= 0) {
```
densamoilov: I'd rather classify the Coverity issue as "intentional" because the `data_type_size` function and other similar functions are expected to receive correct and well-defined arguments since they cannot communicate an error. They may return unexpected values (such as 0, -1, or something else), but only if they are used incorrectly. Such scenarios are not legitimate, if they occur, it indicates that something went wrong during the development stage. There is no way to handle or recover from this kind of error (for example, if the weight data type is incorrect).

I’d recommend always classifying such issues as "intentional".

xuxinzen: Ok, I agreed and updated.

PR #4190, opened 2025-10-22
xe: sycl: l0: use SYCL query for EU count

petercad:
Move to the native SYCL max_compute_units query for detecting EU count. As of 2025.2, this query uses the L0 EU count extension API to "properly" count EUs. "Properly" is in quotes here because some drivers have bugs in this API, in particular older Xe2 drivers.

As an upside, we get proper EU counting for 56EU LNL and some PTL SKUs with current drivers.


petercad:
make test
disable test_device_cpu
disable build_gpu_runtime_ocl

kealan-barbieri (approved):

echeresh (approved):

petercad:
make test perf-gpu
disable build_gpu_runtime_ocl

PR #4194, opened 2025-10-22
Georgena/hybrid cache topology

georgen117:
## Description

This PR fixes inaccurate per-core cache size reporting on hybrid x86 CPUs
(Alder Lake, Raptor Lake, Meteor Lake, Lunar Lake, Nova Lake) by adding an
x64-specific platform layer that uses `Xbyak::util::CpuTopology` instead of
CPUID leaf 4.

### Problem

The legacy `get_per_core_cache_size(int level)` uses CPUID leaf 4 via
`Xbyak::util::Cpu`. On hybrid systems the CPUID sharing-count fields are buggy
or ambiguous, producing values that do not match what any core type actually
sees. Measured examples:

| Platform | Cache | Legacy (CPUID) | New (topology) | Issue |
|----------|-------|----------------|----------------|-------|
| ARL | L2 | 512 KB | 1024 KB (lp_core) | ~6x CPUID undercount |
| PTL | L2 | 384 KB | 1024 KB (lp_core) | ~8x CPUID undercount |
| PTL | L1d | 24 KB | 32 KB (lp_core) | CPUID halves L1d even with no HT |
| LNL | L2 | 320 KB | 1024 KB (lp_core) | ~8x CPUID undercount |
| NVL | L2 | 512 KB | 1024 KB (lp_core) | ~4x CPUID undercount |
| MTL | L2 | 512 KB | 2048 KB (p_core) | ~4x CPUID undercount |

image

A hard-coded `jit_brgemm_conv_utils.cpp` L2 override of 2 MB for AVX2 was
masking part of this on those platforms. Changing the hard-coded L2 override addresses
[MFDNN-14078](https://jira.devtools.intel.com/browse/MFDNN-14078)

### Solution

A new x64-specific module (`src/cpu/x64/platform.cpp/.hpp`) uses
`Xbyak::util::CpuTopology` and distinguishes three core classes:

| Core class | Description |
|------------|-------------|
| P-core     | Intel Performance core |
| LP-core    | Efficient core sharing an L3 cache |
| LPE-core   | Efficient core with no L3 (e.g. Meteor Lake SoC tile) |

On hybrid systems the implementation now always returns a conservative cache
budget:

- `min` policy in code: `min(p_core, lp_core[, lpe_core])` per cache level.
- no runtime policy override.
- no policy enum exposed through the x64 platform API.

`min` is the default because over-committing cache is more costly than
under-utilising it: a blocking heuristic sized for P-cores would overflow the
smaller E-core caches when the workload runs there. Empirical comparison during
development (`p_core` vs `min` on NVL/PTL P-core-pinned runs) showed mixed
per-shape results with no stable rule, confirming `min` as the safer general
default.

### Performance validation (A/B, hardcode vs min)

Convolution data was collected to validate that changing the hardcoded L2 cache size to
a size based on detected cache is the right behavior. A/B data collection was used to
help remove thermal and run-order issues, i.e. multiple runs in different call orders
to avoid spikes caused by thermal throttling.

Ratio = hardcode / min
- If ratio > 1.0, hardcode took more time than min, so min is faster.
- If ratio < 1.0, hardcode took less time than min, so hardcode is faster.
- If ratio ~= 1.0, they are effectively neutral.

image

The MFDNN-14078 shape in standalone testing saw runtime increase from 41.7 ms to
98.6 ms (about 2.36x). However, with A/B testing, the MFDNN-14078 result shrank to
only slightly better than hardcoded performance (about 1.01x).

**Note:** data was collected with `--cold-cache=all` data was recollected without the `--cold-cache`
this resulted in small differences in performance but no outliers that would significantly change the
conclusion based on the `--cold-cache=all` results.

**Performance impact:** first call on a hybrid system costs ~25-30 ms to
initialise `CpuTopology` (magic static, paid once). On non-hybrid systems this
cost is bypassed entirely -- the non-hybrid fast path returns the CPUID value
immediately.

The default `get_per_core_cache_size(int level)` in `src/cpu/platform.cpp`
delegates to `x64::platform::get_per_core_cache_size(level)` on x64 and falls
back to the CPUID path on all other architectures.

All cache debuginfo logging for x64 now lives in `src/cpu/x64/platform.cpp`
(`print_cache_debuginfo_once` and `print_hybrid_cache_debuginfo_once`) behind
`ONEDNN_VERBOSE=debuginfo=1`.

### Files changed

| File | Change |
|------|--------|
| `src/cpu/x64/platform.hpp` | New x64 cache-topology API declaration for `get_per_core_cache_size(int level)` and hybrid helpers. |
| `src/cpu/x64/platform.cpp` | Implementation using `Xbyak::util::CpuTopology`. macOS guard (falls back to legacy). Thread-safe lazy init via C++11 magic statics. Non-hybrid fast path avoids `CpuTopology` init. Includes x64 cache debuginfo printing. |
| `src/cpu/platform.cpp` | Delegates `get_per_core_cache_size(int level)` to `x64::platform::get_per_core_cache_size(level)` on x64; AARCH64 path preserved. |
| `src/cpu/x64/jit_brgemm_conv_utils.cpp` | Removes hard-coded AVX2 L2 = 2 MB override. L2 is now always `get_per_core_cache_size(2) * 3/4`. |

### Known limitations

- `Xbyak::util::CpuTopology` is not available on macOS; all macOS paths fall
  back to legacy CPUID behaviour.
- All-E-core systems (e.g. Sierra Forest) are treated as non-hybrid.

### DIFF
--- a/src/cpu/platform.cpp
+++ b/src/cpu/platform.cpp
@@ -32,6 +32,7 @@
 
 #if DNNL_X64
 #include "cpu/x64/cpu_isa_traits.hpp"
+#include "cpu/x64/platform.hpp"
 #elif DNNL_AARCH64
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 #if defined(DNNL_AARCH64_USE_ACL)
@@ -269,16 +270,9 @@ unsigned get_per_core_cache_size(int level) {
             default: return 0U;
         }
     };
-
 #if DNNL_X64
-    using namespace x64;
-    if (cpu().getDataCacheLevels() == 0) return guess(level);
-
-    if (level > 0 && (unsigned)level <= cpu().getDataCacheLevels()) {
-        unsigned l = level - 1;
-        return cpu().getDataCacheSize(l) / cpu().getCoresSharingDataCache(l);
-    } else
-        return 0;
+    if (x64::cpu().getDataCacheLevels() == 0) return guess(level);
+    return x64::platform::get_per_core_cache_size(level);
 #elif DNNL_AARCH64
     const auto num_caches
             = static_cast<int>(aarch64::cpu().getLastDataCacheLevel());
--- a/src/cpu/x64/jit_brgemm_conv_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_utils.cpp
@@ -1655,13 +1655,8 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
 
     // take L1 as 7/8 of the real size for L1
     brg_blocking_t::L1 = (platform::get_per_core_cache_size(1) * 7) / 8;
-    brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
-    // here is hard-coded L2 size for avx2 performance cores
-    // TODO: get L2 size from the platform
-    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2))
-        brg_blocking_t::L2 = 2 * 1024 * 1024;
     // take L2 as 3/4 of the real size for L2
-    brg_blocking_t::L2 = (brg_blocking_t::L2 * 3) / 4;
+    brg_blocking_t::L2 = (platform::get_per_core_cache_size(2) * 3) / 4;
 
     // These are rough estimates of the latency (relative) of access to various
     // cache levels. This is enough for an estimation of data access cost.
--- a/src/cpu/x64/platform.cpp
+++ b/src/cpu/x64/platform.cpp
@@ -1,0 +1,367 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <algorithm>
+#include <atomic>
+#include <cstdio>
+#include "common/verbose.hpp"
+#include "cpu/x64/cpu_isa_traits.hpp"
+#include "cpu/x64/platform.hpp"
+#include "xbyak/xbyak_util.h"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace x64 {
+namespace platform {
+
+namespace {
+
+#ifndef __APPLE__
+// Global CPU topology instance, initialized on first use
+// This is thread-safe due to C++11 magic statics
+// NOTE: xbyak::util::CpuTopology is not supported on macOS.
+struct cpu_topology_cache_t {
+    Xbyak::util::CpuTopology topology;
+    cpu_topology_cache_t() : topology(cpu()) {}
+};
+
+// Thread-safe lazy initialization using C++11 magic statics
+cpu_topology_cache_t &get_topology_cache() {
+    static cpu_topology_cache_t cache;
+    return cache;
+}
+
+// Map platform cache level (1=L1d, 2=L2, 3=L3) to Xbyak cache type.
+Xbyak::util::CacheType convert_cache_level(int level) {
+    switch (level) {
+        case 1: return Xbyak::util::L1d;
+        case 2: return Xbyak::util::L2;
+        case 3: return Xbyak::util::L3;
+        default: return Xbyak::util::CACHE_UNKNOWN;
+    }
+}
+
+// Filter used when searching for Efficient cores to distinguish between
+// regular E-cores (those sharing an L3) and LP E-cores (no L3 present).
+// For non-Efficient core types this filter is ignored.
+enum class l3_filter_t {
+    any, // no L3 filtering
+    with_l3, // only cores that have a non-zero L3 (regular E-cores)
+    without_l3, // only cores with L3 size == 0 (LP E-cores)
+};
+
+// Find the index of a representative logical CPU of the given core type.
+// For Efficient cores, l3_filter distinguishes regular E-cores (with L3)
+// from LP E-cores (without L3) and optionally restricts matches by L3 presence.
+// Returns the first matching CPU index, or SIZE_MAX if none is found.
+size_t find_representative_cpu(Xbyak::util::CoreType target_type,
+        l3_filter_t l3_filter = l3_filter_t::any) {
+    const auto &topo = get_topology_cache().topology;
+    size_t num_cpus = topo.getLogicalCpuNum();
+
+    for (size_t i = 0; i < num_cpus; i++) {
+        const auto &logical_cpu = topo.getLogicalCpu(i);
+        if (logical_cpu.coreType != target_type) continue;
+
+        if (target_type == Xbyak::util::Efficient
+                && l3_filter != l3_filter_t::any) {
+            const auto &l3 = topo.getCache(i, Xbyak::util::L3);
+            bool has_l3 = (l3.size > 0);
+            if (l3_filter == l3_filter_t::with_l3 && !has_l3) continue;
+            if (l3_filter == l3_filter_t::without_l3 && has_l3) continue;
+        }
+        return i;
+    }
+    return SIZE_MAX;
+}
+
+// Calculate per-core cache size for a specific cache level and CPU index.
+// Matches the legacy getCoresSharingDataCache semantics: divides by the number
+// of PHYSICAL cores sharing the cache, not logical CPUs.  The legacy Xbyak Cpu
+// path (CPUID leaf 4) uses L1d sharing count as the SMT width and divides it
+// out of every level's logical-CPU sharing count.  We replicate that here so
+// that non-hybrid results are identical to the legacy path.
+uint32_t calculate_per_core_cache(size_t cpu_index, int level) {
+    const auto &topo = get_topology_cache().topology;
+
+    Xbyak::util::CacheType cache_type = convert_cache_level(level);
+    if (cache_type == Xbyak::util::CACHE_UNKNOWN) { return 0; }
+
+    const auto &cache = topo.getCache(cpu_index, cache_type);
+    if (cache.size == 0) { return 0; }
+
+    // Number of logical CPUs (threads) sharing this cache instance.
+    size_t sharing_logical = cache.getSharedCpuNum();
+    if (sharing_logical == 0) sharing_logical = 1;
+
+    // SMT width = logical CPUs sharing L1d (L1 is always private to one
+    // physical core, so this count equals the number of HT threads per core).
+    size_t smt_width
+            = topo.getCache(cpu_index, Xbyak::util::L1d).getSharedCpuNum();
+    if (smt_width == 0) smt_width = 1;
+
+    // Physical cores sharing this cache (mirrors legacy smt_width division).
+    size_t sharing_cores = std::max(sharing_logical / smt_width, size_t(1));
+
+    return cache.size / sharing_cores;
+}
+
+struct cache_level_info_t {
+    uint32_t total_kb; // total physical cache capacity in KB
+    size_t sharing_cores; // number of physical cores sharing this instance
+    size_t smt_width; // logical threads per physical core (HT width)
+};
+
+cache_level_info_t get_cache_level_info(size_t cpu_index, int level) {
+    const auto &topo = get_topology_cache().topology;
+    Xbyak::util::CacheType cache_type = convert_cache_level(level);
+    if (cache_type == Xbyak::util::CACHE_UNKNOWN) return {0, 1, 1};
+    const auto &cache = topo.getCache(cpu_index, cache_type);
+    if (cache.size == 0) return {0, 1, 1};
+    size_t sharing_logical = cache.getSharedCpuNum();
+    if (sharing_logical == 0) sharing_logical = 1;
+    size_t smt = topo.getCache(cpu_index, Xbyak::util::L1d).getSharedCpuNum();
+    if (smt == 0) smt = 1;
+    size_t sharing_cores = std::max(sharing_logical / smt, size_t(1));
+    return {cache.size / 1024, sharing_cores, smt};
+}
+
+// Print hybrid per-core cache topology once per process at debuginfo=1 or higher.
+// One line per core type showing physical cache sizes.
+// Shared levels annotated: L{N}:{total}KB({N}cores,{per_core}KB/core)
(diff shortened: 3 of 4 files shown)


src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
...
+            return core_type::p_core;
+        default:
+            return core_type::p_core;
+    }
+#endif
+    return core_type::default_core;
+}
+
+#if DNNL_X64
+// Global cache topology (lazy-initialized)
```
dzarukin: It's lazy, when it's in a function with static storage duration. This is just a permanent global object, static analyzer doesn't like those.

src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
 #endif
 }
 
+bool is_hybrid() {
```
dzarukin: Random spot: I would highly suggest to consider create a platform set of files in x64 and move all that stuff in there. This is a common spot and overloading it with details seems unnecessary given the width of changes.

Additionally, it's probably fine to provide `behavior_t` options (though I'd change a name), but in the core (pun, ha-ha) it seems that only min value is needed after all. I see the following structure:

```
// cpu/platform.cpp
int get_per_core_cache_size(int level) {
#if DNNL_X64
    return x64::get_per_core_cache_size(level); // Note a single argument.
#else
    return what_was_before;
#endif
}

// cpu/x64/platform.cpp
int get_per_core_cache_size(int level, behavior_t btype = behavior::min) {
    ... (all the code you have);
}
```
This way calling through a platform makes a behavior pretty specific, while, if choice is needed, an x64:: version should be called directly with specified `btype`.

All introduced types, expect behavior_t, should be hidden from the external user as an implementation detail of cache detection.

src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
...
+static std::once_flag g_cache_topology_once_flag;
+void init_cache_topology() {
+    std::call_once(g_cache_topology_once_flag, []() {
+#if defined(_WIN32)
+    init_cache_topology_windows(global_cache_topology);
+#elif defined(__linux__)
+    init_cache_topology_linux(global_cache_topology);
+#else
+    // Fallback: use CPUID-based method
+    init_cache_topology_cpuid(global_cache_topology);
```
dzarukin: How long does it take run new method over legacy method?

src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
...
+
+    // Calculate effective cache size per core (considering sharing)
+    auto pcore_size = pcore_cache.size;
+    if (pcore_cache.num_sharing_cores > 0) {
+        pcore_size /= pcore_cache.num_sharing_cores;
+    }
+
+    if (!global_cache_topology.is_hybrid) {
+        // For Non-hybrid system, all cores are assumed to be p-core
+        return pcore_size;
```
dzarukin: What about SRF where all cores are e-cores?

src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
...
+
+    // All char arrays are sized much larger than needed to avoid overflow.
+    // TODO: investigate reducing size of path char buffers. for example
+    //       `cpu_path` is 256 bytes but only needs to be about 30 bytes.
+    //       The `online_path` only needs to be about 40 bytes, but it was
+    //       increased to 512 bytes.  Similar larger numbers were used for
+    //       other paths associated with cache information.
+
+    // Read CPU information from /sys/devices/system/cpu/
+    for (int cpu = 0; cpu < 256; cpu++) { // Check up to 256 CPUs
```
dzarukin: Isn't this number quarriable?

src/cpu/platform.cpp
```diff
@@ -350,6 +351,839 @@ size_t get_timestamp() {
...
+    },
+    // is_hybrid
+    false
+};
+
+
+
+//------------------Start CPUID cache topology code-----------------------------
+
+void populate_cache_map_cpuid(
```
densamoilov: Do we ensure sure thread safety?

src/cpu/platform.cpp
georgen117: I need to remove the global object. I think I will move this into a function and use actual lazy-initialization. Unless you have a better suggestion.

src/cpu/platform.cpp
georgen117: Thanks for the suggestion. I placed the code in `cpu/platform.cpp` because I was not really clear where to place the files. Moving it into `cpu/x64/platform.cpp` makes since.

src/cpu/platform.cpp
georgen117: The functions are guarded by std::once found in the `init_cache_topology()` function. However, there is nothing in each of the OS specific functions that will ensure thread safety.

src/cpu/platform.cpp
georgen117: Yes this should be quarriable. I can get the max processor from the CPUID.1A leaf. I will see if the number of CPUs is placed somewhere in the system layout.

src/cpu/platform.cpp
georgen117: I will have to collect that information.

src/cpu/platform.cpp
georgen117: I still need to test it but on a system that has all E-cores it will treat them as P-cores and then it will copy all the P-core values to the E-core slot. The cache size numbers should still be correct.

georgen117: Any system that is not hybrid the cache size returned will be that of `cpu(0)`.

src/cpu/platform.cpp
georgen117: logic has been moved into Xbyak_utils and it now queries the OS for the number of cores.

src/cpu/platform.cpp
georgen117: With the switch of the code to using `Xbyak::CpuTopologyCache` the thread safety is ensured by C++11's magic statics.

src/cpu/platform.cpp
georgen117: The comment is on old code  I did take the time to measure the time:

| Path | First-call cost | Subsequent calls |
|---|---|---|
| New `CpuTopology` init | ~25ms - 30ms | negligible (cached static) |
| Legacy (CPUID) | ~7µs - 10µs (existing `xbyak::Cpu` already cached) | negligible (existing `xbyak::Cpu` already cached) |

Due to this measurement, I did a small reorganization of the code if the system is not hybrid, then the code will automatically default to the legacy CPUID path.  This avoids the cost of initializing the `CpuTopology` Unfortantly this cost is unavoidable if we want to get accurate cache size on hybrid systems and the cost is one time on init.

georgen117:
make test

ankalinin (approved):

src/common/verbose.cpp
```diff
@@ -109,6 +109,47 @@ void print_header() noexcept {
                 dnnl_runtime2str(dnnl_version()->cpu_runtime),
                 dnnl_get_max_threads());
         verbose_printf("info,cpu,isa:%s\n", cpu::platform::get_isa_info());
+        if (cpu::platform::is_hybrid_cpu()) {
+            verbose_printf(
+                    "info,cpu,per_core_cache,pcore:L1d:%uKB,L2:%uKB,L3:%uKB\n",
```
densamoilov: I'm not sure how we expect this information to be used. If we already know what CPU is used we can find this information.
I'm also not sure how useful it would be to have this information in verbose in general. Given that you have the most experience working with it, any use cases?

src/common/verbose.cpp
```diff
@@ -109,6 +109,47 @@ void print_header() noexcept {
...
+            verbose_printf("info,cpu,per_core_cache:L1d:%uKB,L2:%uKB,L3:%uKB\n",
+                    cpu::platform::get_per_core_cache_size(1) / 1024,
+                    cpu::platform::get_per_core_cache_size(2) / 1024,
+                    cpu::platform::get_per_core_cache_size(3) / 1024);
+        }
+        // If ONEDNN_CACHE_POLICY is set, all call sites use the same override
+        // policy. Emit a cache_budget line showing the effective values so a
+        // reader of the log can see exactly what the blocking heuristics used.
+        {
+            const std::string policy = getenv_string_user("CACHE_POLICY");
```
densamoilov: We need to document environment variables. 

I actually think that it'd probably be better if we put it under `debuginfo`, assuming this info is generally useful.

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -1655,13 +1655,8 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
...
     // take L1 as 7/8 of the real size for L1
     brg_blocking_t::L1 = (platform::get_per_core_cache_size(1) * 7) / 8;
-    brg_blocking_t::L2 = platform::get_per_core_cache_size(2);
-    // here is hard-coded L2 size for avx2 performance cores
-    // TODO: get L2 size from the platform
-    if (one_of(isa, avx2, avx2_vnni, avx2_vnni_2))
-        brg_blocking_t::L2 = 2 * 1024 * 1024;
     // take L2 as 3/4 of the real size for L2
-    brg_blocking_t::L2 = (brg_blocking_t::L2 * 3) / 4;
+    brg_blocking_t::L2 = (platform::get_per_core_cache_size(2) * 3) / 4;
```
densamoilov, line 1659: The PR doesn't seem to have perf data. I remember the problem was that this change affected performance for some cases. Do you have perf data to verify that it doesn't cause any regressions?

src/cpu/x64/platform.cpp
```diff
@@ -0,0 +1,236 @@
...
+// Inner implementation: resolves sizing_policy to a cache size with no env-var override.
+// Called by both get_per_core_cache_size (which may apply the override first)
+// and by topology-info helpers (get_per_core_cache_size_pcore etc.) that must
+// always return true topology values regardless of the active env-var override.
+static unsigned get_per_core_cache_size_for_policy(
+        int level, cache_sizing_policy_t sizing_policy) {
+    // Validate level
+    if (level < 1 || level > 3) { return 0; }
+
+#ifdef __APPLE__
```
densamoilov, line 331: I'm not sure we need to handle the APPLE case for x64 platforms

@vpirogov, do we?

src/cpu/x64/platform.hpp
```diff
@@ -0,0 +1,82 @@
...
+// Use Xbyak_utils cache topology methods to determine the per-core cache size.
+//
+// This avoids using older CPUID-based methods which can result in inaccurate
+// values on hybrid CPUs.
+//
+// The cache_sizing_policy_t argument specifies the behavior of the query on hybrid CPUs.
+// On non-hybrid CPUs, the cache_sizing_policy_t argument is ignored and the function
+// returns the per-core cache size as normal.
+//
+// - if cache_sizing_policy_t is p_core/lp_core/lpe_core, the function returns the per-core cache
```
densamoilov: What is the rationale behind having the sizing policies? Why do we need this level of flexibility?

src/common/verbose.cpp
georgen117: This was really useful when validating this PR. However, I don't know if there is a wider use for this information. Removing the verbose logging would simplify the PR.

src/common/verbose.cpp
georgen117: I will remove the environment variable. The environment variable was added to validate the code making it possible to change the behavior without recompiling the code.

Along with the removal of the verbose print statments the PR will be further simplified.

src/cpu/x64/jit_brgemm_conv_utils.cpp
georgen117: I know that this fixes the performance reported for  [MFDNN-14078](https://jira.devtools.intel.com/browse/MFDNN-14078).

I will need to do a quick check of the data I collected to give you more concrete numbers for other shapes. I may need to collect some number with the older hardcoded L2 value.

src/cpu/x64/platform.hpp
georgen117: The original thought was that some shapes that were compute bound instead of memory bound would be able to choose the p_core cache_sizing_policy_t and get an increase in performance. I looked for a heuristic that could benefit from changing to a specific cache_size_policy_t and I couldn't come up with one.  I can remove the cache_sizing_policy_t all together and have the code default to the `min` value for hybrid cores and default the legacy (CPUID leaf 4) behavior for non-hybrid systems.

That would further simplify the code. It could always be re-introduced if I can successfully find a case that the performance can be improved by using the `cache_sizing_policy_t`

georgen117 (commented):
@densamoilov after reading your reviews I can simplify this PR considerably.  This was over complicated because I was investigating a lot of possible configureations.  I ended up on using CPUID leaf 4 for non-hybrid systems and the `per_core_cache_size(min(p_core, lp_core[, lpe_core])` if I remove the envronment varable the is only two paths that are decided by the existence of the hybrid CPUID flag. 

I will simplify the PR.

src/cpu/platform.cpp
georgen117: Just to make sure I did test the code on a SRF system and it behaved just like a non-hybrid system which is what I expected.

src/common/verbose.cpp
georgen117: I removed all the verbose_printf out of the verbose.cpp function and moved it into the cpu/x64/platform.cpp under debuginfo logging.

Since a log of the functions placed in cpu/platform.hpp were only placed there to make them avalible to the verbose print I was able to remove all those functions reducing the diff from this PR considerably.

src/common/verbose.cpp
georgen117: I removed the CACHE_POLICY environment variable.

src/cpu/x64/platform.hpp
georgen117: when I removed the `CACHE_POLICY` environment variable I also removed the `cache_sizing_policy_t` env variable since the environment variable was the only way to change the policy.

src/cpu/x64/platform.cpp
vpirogov: Apple dropped support for x64 in macOS 29. While macOS 28 has 3 more years of life I don't think we need to care for anything other than breaking the build.

src/cpu/x64/jit_brgemm_conv_utils.cpp
georgen117: see the **Performance validation (A/B, hardcode vs min)** section in the PR description for numbers.  I used A/B testing for the performance gathering. The  difference between the hardcoded value and the min value was minimal. I tried to collect the same shape data as was collected for PR#2420 which introduced the hardcoded value. I did the best I could with the data found in `tests/benchdnn/inputs/conv/`

densamoilov (approved):

densamoilov:
@georgen117, thank you for the quick edits. LGTM!

georgen117:
make test

PR #4209, opened 2025-10-24
cpu: x64: matmul: check src md instead of src tag

densamoilov:
Multiple tags can correspond to the same physical layout when unit dimensions are present; therefore, we cannot rely on tags in cases where it matters. The newly introduced `is_canonical()` function can be used to perform certain checks in a robust way.

Fixes MFDNN-14321.

### DIFF
--- a/src/common/memory_desc_wrapper.hpp
+++ b/src/common/memory_desc_wrapper.hpp
@@ -77,6 +77,64 @@ struct memory_desc_wrapper : public c_compatible {
         return format_kind() == format_kind::host_scalar;
     }
 
+    /**
+     * Checks whether the memory descriptor represents a canonical (row-major)
+     * layout.
+     *
+     * A canonical layout is defined as one where the strides follow the
+     * standard row-major ordering of the remaining dimensions, ignoring any
+     * dimensions of size 1.
+     *
+     * @note Format tags (e.g., "abc", "acb") may correspond to different
+     *   stride arrays but the same physical layout when unit dimensions are
+     *   present. Therefore, checking for canonical layout is more reliable than
+     *   matching a specific tag.
+     *
+     * Example:
+     *   dims    = [10, 12, 1]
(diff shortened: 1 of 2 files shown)


ankalinin (approved):

src/common/memory_desc_wrapper.hpp
```diff
@@ -77,6 +77,64 @@ struct memory_desc_wrapper : public c_compatible {
...
+            if (dims()[i] != 1) {
+                normalized_dims[normalized_ndims] = dims()[i];
+                normalized_strides[normalized_ndims] = strides()[i];
+                normalized_ndims++;
+            }
+        }
+
+        if (normalized_ndims == 0) return true;
+
+        int expected_stride = 1;
```
densamoilov: (reminder for myself) use dim_t instead of int

src/common/memory_desc_wrapper.hpp
```diff
@@ -77,6 +77,64 @@ struct memory_desc_wrapper : public c_compatible {
...
+        int normalized_ndims = 0;
+        for (int i = 0; i < ndims(); ++i) {
+            if (dims()[i] != 1) {
+                normalized_dims[normalized_ndims] = dims()[i];
+                normalized_strides[normalized_ndims] = strides()[i];
+                normalized_ndims++;
+            }
+        }
+
+        if (normalized_ndims == 0) return true;
```
avmanerikar, line 128: (nit): 
```suggestion
        if (!normalized_ndims) return true;
```

densamoilov: I actually think it's not the right place to use the idiom because the idiom implies that the checked variable is conceptually boolean (e.g. valid vs invalid for pointers) but the number of dimensions is not conceptually boolean and 0 dimensions is actually a valid number, which we use just to do a quick return. I’ll keep the explicit check so that the intent is clearer.

PR #4220, opened 2025-10-27
[GPU] fixes for opencl primitives

dyoussif:
1) Improve subgroup handling
2) couple other minor fixes

### DIFF
--- a/src/gpu/intel/softmax/xe.cl
+++ b/src/gpu/intel/softmax/xe.cl
@@ -219,7 +219,8 @@ xe_softmax_fwd(__global SRC_DATA_T *src, __global DST_DATA_T *dst,
     const int channel_block = channel_id / CHANNELS;
     const int channel_in_block = channel_id % CHANNELS;
 
-    int data_off = mb * CHANNELS * SOFTMAX_AXIS_SIZE + channel_offset
+    int data_off = mb * CHANNELS_PADDED * SOFTMAX_AXIS_SIZE
+            + channel_block * SOFTMAX_AXIS_SIZE * CHANNELS + channel_offset
             + channel_in_block;
     const int buf_reads = THREAD_BUF_SIZE;
 #else


dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

echeresh (approved):

dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_softmax
enable benchdnn_reorder
enable benchdnn_zeropad
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

dyoussif:
make test perf-gpu
set primitive=softmax reorder

src/gpu/intel/reorder/ref.cpp
```diff
@@ -128,10 +128,12 @@ void ref_t::pd_t::init_scratchpad() {
...
     auto &src = CTX_IN_STORAGE(DNNL_ARG_FROM);
     auto &dst = CTX_OUT_STORAGE(DNNL_ARG_TO);
-    auto tmp = ctx.get_scratchpad_grantor().get_memory_storage(
-            memory_tracking::names::key_reorder_space);
-
     const auto &conf = pd()->conf;
+    auto tmp = conf.subbyte_pack
+            ? ctx.get_scratchpad_grantor().get_memory_storage(
+                    memory_tracking::names::key_reorder_space)
+            : nullptr;
```
Simonsays095, line 138: Instead of setting to `nullptr`, set to `dst` and use unconditionally below.
```suggestion
    auto tmp = conf.subbyte_pack
            ? ctx.get_scratchpad_grantor().get_memory_storage(
                    memory_tracking::names::key_reorder_space)
            : &dst;
```

src/gpu/intel/softmax/simple.hpp
```diff
@@ -70,7 +70,7 @@ struct simple_fwd_t : public primitive_t {
                                       intel_engine->mayiuse(
                                               compute::device_ext_t::khr_fp64)),
                     VERBOSE_UNSUPPORTED_DT_CFG);
-            VDISPATCH_SOFTMAX(intel_engine->mayiuse_sub_group(subgroup_size),
+            VDISPATCH_SOFTMAX(get_subgroup_size(intel_engine, subgroup_size),
```
Simonsays095, line 73: Since we're really just checking for compatibility here, why not just use `device_info_t::max_subgroup_size(arch)` and exclude this check (and all other related checks, while removing the `get_subgroup_size` function)?

src/gpu/intel/softmax/xe.hpp
```diff
@@ -297,6 +302,9 @@ struct xe_bwd_t : public primitive_t {
                     != format_tag::undef);
             is_blk = (diff_src_d.matches_one_of_tag(nCw16c, nChw16c, nCdhw16c)
                     != format_tag::undef);
+            VDISPATCH_SOFTMAX(
+                    (is_blk && subgroup_size == 16) || diff_src_d.is_plain(),
+                    VERBOSE_UNSUPPORTED_TAG);
```
Simonsays095, line 308: Instead of choosing another implementation when `is_blk` and subgroup_size is too large, could we instead just "snap" the subgroup size to 16 if it's in one of these formats?

src/gpu/intel/zeropad/simple.hpp
```diff
@@ -59,9 +52,15 @@ struct simple_t : public primitive_t {
...
         kernel_subg16_ = kernels[1];
         kernel_subg16_mask_and_clear_dt_1b_ = kernels[2];
+        auto *compute_engine = utils::downcast<intel::engine_t *>(engine);
+
+        can_use_subg16_ = compute_engine->mayiuse_sub_group(16);
 
-        if (!kernel_ || !kernel_subg16_ || !kernel_subg16_mask_and_clear_dt_1b_)
+        if (!kernel_) return status::runtime_error;
+        if (can_use_subg16_
+                && (!kernel_subg16_ || !kernel_subg16_mask_and_clear_dt_1b_))
```
Simonsays095, line 62: I believe kernel compilation will fail if we try to compile a sgs16 kernel when it's not supported by the hardware. In these cases where `!can_use_subg16`, I think we should use a define to guard and remove these kernels via the preprocessor, leaving `simple_zero_pad` available for use as a fallback.

dyoussif:
make test perf-gpu
set primitive=softmax reorder

dyoussif:
make test perf-gpu
set primitive=softmax reorder

dyoussif:
make test perf-gpu
set primitive=softmax reorder

dyoussif:
make test perf-gpu
set primitive=softmax reorder

dyoussif:
make test perf-gpu
set primitive=softmax reorder

src/gpu/intel/reorder/ref.cpp
dyoussif: `dst` is a unique ptr so i don't think that will work.

src/gpu/intel/softmax/simple.hpp
dyoussif: the `max_subgroup_size` may not be the most performant one. We want to maintain existing logic regarding subgroup selection.

src/gpu/intel/zeropad/simple.hpp
dyoussif: If sgs16 kernel compilation fails, we use `kernel_`.

src/gpu/intel/softmax/xe.hpp
dyoussif: I think this kernel is specifically intended only for `subgroup_size` 16, since `subgroup_size` is hardcoded. This is just  to be explicit about it.

src/gpu/intel/zeropad/simple.hpp
Simonsays095: All 3 of these kernels are compiled simultaneously, so if one of the sgs16 kernels fails to compile, we also will fail to compile `simple_zero_pad` which is used for `kernel_`. I think we just rely on the fact that all currently-supported hardware supports sgs16 and never hit this path.

src/gpu/intel/softmax/xe.hpp
Simonsays095: If it requires sgs16, then we should just check `mayiuse_sub_group` instead of calling `get_subgroup_size`, to keep behaviour consistent. Then we're guaranteed to use sgs16 always.

PR #4221, opened 2025-10-27
gpu: intel: compute: add "-cl-intel-greater-than-4GB-buffer-required"

spalicki:
If required memory allocation is >4GB and driver does not normally allow it, compile modules in stateless addressing model.
Fix for [MFDNN-14249](https://jira.devtools.intel.com/browse/MFDNN-14249)

### DIFF
--- a/src/gpu/intel/compute/dispatch_reusable.hpp
+++ b/src/gpu/intel/compute/dispatch_reusable.hpp
@@ -245,7 +245,8 @@ struct dispatch_compile_params_t {
     subgroup_data_t subgroup;
     int32_t num_terms = 0;
     bool use_int32_offset = false;
-    uint8_t padding[3] = {0};
+    bool require_large_buffers = true;
+    uint8_t padding[2] = {0};
     gws_indexing_term_t::compile_params_t terms[MAX_INDEXING_TERMS]
             = {{gws_op_t::SOLO, 0}};
 
@@ -525,6 +526,7 @@ public:
 
         // Save buffer information
         dim_t max_buffer_size = 0;
+        uint64_t max_buffer_size_bytes = 0;
         compile_params.num_buffers = buffers.size();
         for (size_t buf_idx = 0; buf_idx < buffers.size(); buf_idx++) {
             const named_buffer_t &buffer = buffers[buf_idx];
@@ -546,9 +548,14 @@ public:
 
             // Check buffer sizes to see if we can use int32_t offsets
             max_buffer_size = std::max(max_buffer_size, buffer.nelems(true));
+            max_buffer_size_bytes = std::max(max_buffer_size_bytes,
+                    buffer.nelems(true)
+                            * types::data_type_size(buffer.data_type));
         }
 
         compile_params.use_int32_offset = max_buffer_size <= INT32_MAX;
+        compile_params.require_large_buffers
+                = max_buffer_size_bytes > UINT32_MAX;
         compile_params.subgroup = subgroup;
 
         // Set runtime params
--- a/src/gpu/intel/compute/kernel_ctx.hpp
+++ b/src/gpu/intel/compute/kernel_ctx.hpp
@@ -52,13 +52,7 @@ public:
         for (auto &opt : option_set_)
             oss << " " << opt;
 
-        if (use_int32_offset_) {
-            oss << " -DUSE_INT32_OFFSET";
-        } else {
-            // TODO: Determine if specialization for buffers between 2GB and 4GB
-            // is worthwhile
-            oss << " -cl-intel-greater-than-4GB-buffer-required";
-        }
+        if (use_int32_offset_) { oss << " -DUSE_INT32_OFFSET"; }
 
         for (auto &int_var : int_var_map_) {
             // C tokens are parsed as (-)(integer_literal). As abs(INT_MIN) >
@@ -85,6 +79,7 @@ public:
 
     void register_buffer_size(size_t size) {
         if (size > INT_MAX) use_int32_offset(false);
+        if (size > UINT32_MAX) require_large_buffers(true);
     }
 
     void register_buffer_size(const memory_desc_wrapper &mdw) {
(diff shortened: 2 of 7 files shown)


src/gpu/intel/compute/dispatch_reusable.hpp
```diff
@@ -245,6 +245,7 @@ struct dispatch_compile_params_t {
     subgroup_data_t subgroup;
     int32_t num_terms = 0;
     bool use_int32_offset = false;
+    bool allow_huge_buffers = false;
```
rjoursler: nit: The default value should be `true` so that the performance is opt-in instead of functional correctness being opt-in.

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -97,6 +95,10 @@ class kernel_ctx_t {
     // operations when native 64-bit operations are unsupported.
     void use_int32_offset(bool value) { use_int32_offset_ = value; }
 
+    // Enable '-cl-intel-greater-than-4GB-buffer-required' flag
+    // to correctly process big buffers.
+    void allow_huge_buffers(bool value) { allow_huge_buffers_ = value; }
```
rjoursler: nit: rename `allow_huge_buffer` to `support_large_buffers` for improved clarity and to align with naming elsewhere related to the 4GB buffers.

src/gpu/intel/compute/kernel_ctx.hpp
rjoursler, line 88: These `register_buffer_size` interfaces need updated so that they call `allow_huge_buffers` when it is needed.

spalicki: Done.

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: I changed it to `require_huge_buffers` and `has_huge_buffers`. Huge like huge pages, since it is over the OCL limit, but I can rename it to large.

src/gpu/intel/compute/dispatch_reusable.hpp
spalicki: I do not know if I understand you correctly, but it is `false` at default because `-cl-intel-greater-than-4GB-buffer-required` has negative performance impact on its own, so we only want to enable it when needed.

rjoursler: The design here is that the default provides functional correctness. This is to avoid people unfamiliar with these details from accidentally committing an implementation which provide incorrect results on large buffers. We can then explicitly opt-in to the better optimized path when it provides benefit.

src/gpu/intel/compute/kernel_ctx.hpp
rjoursler: The reason for using `large` is that is already the naming convention used throughout oneDNN:
```
$ git grep large_buffer
src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp:    const bool req_zp_large_buffer = jcp.src_zero_point
src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp:    VDISPATCH_CONV_IC(!req_zp_large_buffer, VERBOSE_IMPL_HEURISTIC_FAIL,
src/cpu/x64/jit_uni_pool_kernel.cpp:bool jit_uni_pool_kernel_t<isa>::has_large_buffers(const pooling_pd_t *ppd) {
src/cpu/x64/jit_uni_pool_kernel.cpp:    VDISPATCH_POOLING_IC(!has_large_buffers(ppd), VERBOSE_SHAPE_RESTRICTION);
src/cpu/x64/jit_uni_pool_kernel.hpp:    static bool has_large_buffers(const pooling_pd_t *ppd);
src/gpu/intel/conv/jit_v2.cpp:bool has_large_buffers(const pd_t *pd) {
src/gpu/intel/conv/jit_v2.cpp:        if (has_large_buffers(pd)) return status::unimplemented;
tests/benchdnn/inputs/eltwise/harness_eltwise_large_buffer:--batch=shapes_large_buffer
tests/benchdnn/inputs/eltwise/harness_eltwise_large_buffer:--batch=shapes_large_buffer
tests/benchdnn/inputs/eltwise/test_eltwise_gpu:--batch=harness_eltwise_large_buffer
```

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -52,13 +52,7 @@ class kernel_ctx_t {
...
             oss << " " << opt;
 
-        if (use_int32_offset_) {
-            oss << " -DUSE_INT32_OFFSET";
-        } else {
-            // TODO: Determine if specialization for buffers between 2GB and 4GB
-            // is worthwhile
-            oss << " -cl-intel-greater-than-4GB-buffer-required";
-        }
+        if (use_int32_offset_) { oss << " -DUSE_INT32_OFFSET"; }
```
rjoursler, line 55: It looks like the logic to generate `-cl-intel-greater-than-4GB-buffer-required` was dropped.

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -216,6 +214,7 @@ class kernel_ctx_t {
     std::set<std::string> option_set_;
     std::unordered_map<std::string, std::string> custom_headers_;
     bool use_int32_offset_ = true;
+    bool require_huge_buffers_ = true;
```
rjoursler: This needs to default to `false` for `register_buffers_size` to work correctly.

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: It was moved to engine, since it is not needed on all platforms and require additional check.

src/gpu/intel/compute/dispatch_reusable.hpp
spalicki: Done.

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: Done.

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: Done.

rjoursler (approved):

Simonsays095 (approved):

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

PR #4224, opened 2025-10-28
graph: backend: dnnl: split op_executable.hpp/cpp

TaoLv:
This is to split the large monolithic op_executable.hpp/cpp into multiple hpp/cpp paris based on the executable classes in them. The purpose is to improve the readability and maintainability of the source code as well as save the build time of the backend. Currently it's simply cut-and-paste the code.


src/graph/backend/dnnl/executables/conv.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+    return {pd, false};
+}
+
+arg_indices_t conv_fwd_executable_t::get_arg_indices(const op_t *op) {
+    return get_arg_indices_for_conv_and_matmul(op);
+}
+
+// deconv_fwd_executable_t implementations
+deconv_fwd_executable_t::desc_t deconv_fwd_executable_t::create_desc(
```
ElaineBao: conv and deconv are in a single file, is it intended?

src/graph/backend/dnnl/executables/eltwise.hpp
```diff
@@ -0,0 +1,320 @@
...
+                = create_desc(op, p_engine, pd_cache, fpmath, use_block_layout);
+        prim_ = dnnl::binary(desc);
+
+        if (op->has_attr(op_attr::with_sum))
+            with_sum_ = op->get_attr<bool>(op_attr::with_sum);
+    }
+
+    void execute(const stream &stream,
+            const std::unordered_map<int, memory> &args) const override {
+        if (is_dummy_) {
```
ElaineBao, line 136: do you think some of the function implementations can be moved to cpp file?

src/graph/backend/dnnl/executables/conv.cpp
TaoLv: Actually Copilot decided that. I just accepted the idea. :)

src/graph/backend/dnnl/executables/eltwise.hpp
TaoLv: It's a good idea. I was thinking if we want to do it in this PR or another one. As I mentioned in the description, the current change is basically cut-and-paste. So the original op_executable.hpp -> executables/\*.hpp and op_executable.cpp -> executables/\*.cpp. We definitely can move more code from hpp to cpp in the future.

src/graph/backend/dnnl/executables/conv.cpp
TaoLv: Now they are split in the updated commit.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #4231, opened 2025-10-28
xe: make includes of gemmstone headers path-agnostic

echeresh:
This is to allow using a standalone (or third-party) gemmstone repository for building.

### DIFF
--- a/src/gpu/intel/jit/codegen/reorder.hpp
+++ b/src/gpu/intel/jit/codegen/reorder.hpp
@@ -20,7 +20,7 @@
 #include <functional>
 
 #include "common/utils.hpp"
-#include "gpu/intel/gemm/jit/generator/pieces/copy_plan.hpp"
+#include "gemmstone/../../generator/pieces/copy_plan.hpp"
 #include "gpu/intel/jit/codegen/operand.hpp"
 #include "gpu/intel/jit/codegen/register_scope.hpp"
 #include "gpu/intel/jit/ir/reorder.hpp"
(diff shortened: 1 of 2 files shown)


echeresh:
make test
disable device_cpu

kealan-barbieri (approved):

src/gpu/intel/jit/codegen/reorder.hpp
```diff
@@ -20,7 +20,7 @@
 #include <functional>
 
 #include "common/utils.hpp"
-#include "gpu/intel/gemm/jit/generator/pieces/copy_plan.hpp"
+#include "gemmstone/../../generator/pieces/copy_plan.hpp"
```
rjoursler, line 23: How much work would it be to move the necessary interfaces from `copy_plan.hpp` into the external headers? Relying on the internal headers like this is not a good design choice.

echeresh: I think @atkassen can comment better. Yes, without a proper interface this is not aligned with the current design.

atkassen: If we're happy with `#include "gemmstone/generator.hpp"` instead, I think it should suffice: https://github.com/uxlfoundation/oneDNN/blob/59ec03608ab9239e0c25d66530d0de12a9e89991/src/gpu/intel/gemm/jit/include/gemmstone/generator.hpp#L37

rjoursler: Yeah, looks like this just snowballs, so we can leave it as is. This definitely needs reworked at some point, `gemmstone/generator.hpp` shouldn't have includes outside of the `include` directory either.

echeresh:
make test
disable device_cpu

PR #4237, opened 2025-10-29
cpu: aarch64: jit_reorder: cache blocking 4/8 inner blocks

davsva01:
# Description

Adds improved blocking for cache for simple reorders of inner blocking of 4 or 8. The src is now traversed row-wise instead of column-wise which improves cache locality for these reorders. 

Example of benchmark cmd:
`OMP_NUM_THREADS=16 benchdnn --reorder --allow-enum-tags-only=0 --sdt=f32 --ddt=f32 --stag=cdba --dtag=Acdb8a --mode=p 512x512x3x3`

Performance improvements for jit:uni (16 threads, --sdt=f32 --ddt=f32):

| stag  | dtag | shape | perf improvement |
| ------------- | ------------- | ------------- | ------------- |
| cdba | Acdb8a | 512x512x3x3 | 2.97x |
| ab | Ba8b | 4096x4096 | 2.69x |
| ab | Ba4b | 4096x4096 | 3.27x |
| ba | Ab8a | 4096x4096 | 2.76x |
| ba  | Ab4a | 4096x4096 | 3.30x |

# Checklist

## General

Nightly CI tested.

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/jit_uni_reorder.cpp
+++ b/src/cpu/aarch64/jit_uni_reorder.cpp
@@ -2589,6 +2589,37 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
 } // namespace tr
 
 static void prb_block_for_cache(tr::prb_t &prb) {
+    // Performance improvements when doing simple inner blocking of 8 or 4
+    // This covers ab->Ba8b, ab->Ba4b, ba->Ab8a, ba->Ab4a and cdba->Acdb8a
+    // Split middle node, then swap to improve cache locality
+    // Before split+swap we traverse src column-wise 8 row elements at a time from top to bottom
+    // After split+swap in src we traverse split_countx8 row-wise for inner_blk=8, and split_countx4 for inner_blk=4
+    if (prb.ndims == 3) {
+        const int inner_blk = prb.nodes[0].n;
+        if ((inner_blk == 8 || inner_blk == 4) && prb.nodes[0].is == 1
+                && prb.nodes[0].os == 1 && prb.nodes[1].os == inner_blk
+                && prb.nodes[2].is == inner_blk) {
+
+            // Try finding value to split on
+            // prb.nodes[1].n > 32 to ensure we only split if enough rows for split to have cache benefit
+            size_t split_value = 0;
+            for (size_t d = 8; d >= 2; --d) {
+                if (prb.nodes[1].n % d == 0 && prb.nodes[1].n != d
+                        && prb.nodes[1].n > 32) {
+                    split_value = d;
+                    break;
+                }
+            }
+
+            // Split on found split_value, if any
+            if (split_value) {
+                prb_node_split(prb, 1, split_value);
+                prb_node_swap(prb, 3, 2);
+                prb_node_dependency(prb);
+            }
+        }
+    }
+
     /* If strides for 0th and 1st nodes are cache friendly
      * then one can altogether do away with blocking ! */
     static constexpr int num_elems_thr = 16;


src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2589,6 +2589,39 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
 } // namespace tr
 
 static void prb_block_for_cache(tr::prb_t &prb) {
+    // Performance improvements when doing simple inner blocking of 8 or 4
+    // This covers ab->Ba8b, ab->Ba4b, ba->Ab8a, ba->Ab4a and cdba->Acdb8a
+    // Split middle node, then swap to improve cache locality
+    // Before split+swap we traverse src column-wise 8 row elements at a time from top to bottom
+    // After split+swap in src we traverse split_countx8 row-wise for inner_blk=8, and split_countx4 for inner_blk=4
+    if (prb.ndims == 3) {
```
jondea, line 2597: Could this be `>=3`?

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2589,6 +2589,39 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
...
+        const int inner_blk = prb.nodes[0].n;
+        if ((inner_blk == 8 || inner_blk == 4) && prb.nodes[0].is == 1
+                && prb.nodes[0].os == 1 && prb.nodes[1].os == inner_blk
+                && prb.nodes[2].is == inner_blk
+                && utils::one_of(prb.itype, data_type::f32, data_type::s32,
```
jondea: Why does the type matter? Would it be more general and specific to use datatype _size_?

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2589,6 +2589,39 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
...
+                        data_type::bf16, data_type::f16)
+                && utils::one_of(prb.otype, data_type::f32, data_type::s32,
+                        data_type::bf16, data_type::f16)) {
+
+            // Try finding value to split on
+            size_t split_value = 0;
+            for (int d = 8; d >= 3; --d) {
```
jondea: Why stop at 3?

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2589,6 +2589,39 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
...
 
 static void prb_block_for_cache(tr::prb_t &prb) {
+    // Performance improvements when doing simple inner blocking of 8 or 4
+    // This covers ab->Ba8b, ab->Ba4b, ba->Ab8a, ba->Ab4a and cdba->Acdb8a
+    // Split middle node, then swap to improve cache locality
+    // Before split+swap we traverse src column-wise 8 row elements at a time from top to bottom
+    // After split+swap in src we traverse split_countx8 row-wise for inner_blk=8, and split_countx4 for inner_blk=4
+    if (prb.ndims == 3) {
+        const int inner_blk = prb.nodes[0].n;
+        if ((inner_blk == 8 || inner_blk == 4) && prb.nodes[0].is == 1
```
jondea, line 2599: Why only 8 or 4?

src/cpu/aarch64/jit_uni_reorder.cpp
```diff
@@ -2589,6 +2589,39 @@ kernel_t *kernel_t::create(const kernel_t::desc_t &desc) {
 } // namespace tr
 
 static void prb_block_for_cache(tr::prb_t &prb) {
+    // Performance improvements when doing simple inner blocking of 8 or 4
+    // This covers ab->Ba8b, ab->Ba4b, ba->Ab8a, ba->Ab4a and cdba->Acdb8a
+    // Split middle node, then swap to improve cache locality
+    // Before split+swap we traverse src column-wise 8 row elements at a time from top to bottom
+    // After split+swap in src we traverse split_countx8 row-wise for inner_blk=8, and split_countx4 for inner_blk=4
```
jondea, line 2596: So we get 4 or 8 rows at once, but get those blocks along the row?

jondea (changes requested):
This is great work! I just have mainly clarifying questions and nits. Just bear in mind that this stuff is elegant, but quite opaque, so any clarifying comments you can add would be greatly appreciated.

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: No not really, the dimensions are tied to the type of problem rather than the src and dst memory formats. It defines moving x elements from src to dst and how to index for each jump in src and dst.

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: It didn't, I've removed that check now :)

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: I've changed this to 2 now, the performance still holds when we split on 2.

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: For larger inner blocks we already use a larger portion of the cache line. I think we can revisit larger inner blocks for smaller datatypes in the future if that's something we want to accelerate.

src/cpu/aarch64/jit_uni_reorder.cpp
davsva01: Yes that's exactly it :)

src/cpu/aarch64/jit_uni_reorder.cpp
jondea: Ah, it's gone completely, I didn't expect that, but great!

src/cpu/aarch64/jit_uni_reorder.cpp
jondea: I guess I'd then ask, why stop at 2? But 2 looks a lot less arbitrary than 3, so I'll let you off

PR #4244, opened 2025-10-29
xe: jit: further refactor to enable upstreaming IR/DSL to gemmstone

rjoursler:
After these changes, we can fairly easily upstream the necessary `jit/ir` and `jit/dsl` files to Gemmstone.

### DIFF
--- a/src/gpu/intel/jit/dsl/dsl.cpp
+++ b/src/gpu/intel/jit/dsl/dsl.cpp
@@ -145,14 +145,37 @@ struct ctx_t {
 
     tensor_t def(const std::string &name, const layout_t &layout,
             type::attr_t attr, const expr_t &value = {}) {
+        auto &back = layout.blocks().back();
+        auto size = into<int>(back.size * int64_t(back.stride));
+
+        // Padding allocations due to strides with overlapping dimensions have
+        // unclear semantics, disallow their use.
+        gpu_assert([&]() {
+            int64_t max_off = 0;
+            for (auto &b : layout.blocks()) {
+                max_off += (b.size - 1) * int64_t(b.stride);
+            }
+            return max_off < size;
+        }());
+
+        gpu_assert(is_zero(layout.offset()));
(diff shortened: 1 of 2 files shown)


rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/dsl/dsl.cpp
```diff
@@ -349,7 +349,7 @@ tensor_t def(const std::string &name, layout_t layout, const expr_t &value,
         type::attr_t attr) {
     if (any(attr & type::attr_t::slm)) {
         gpu_assert(value.is_empty());
-        auto alloc_elems = into<int>(size_bytes(layout) / layout.type().size());
+        auto alloc_elems = into<int>(layout.elems());
```
echeresh: This is not correct for strided layouts, right? We either need an assertion somewhere in the DSL to check that any allocated SLM layout is dense (do we have it already?) or use `size_bytes()` style calculation.

src/gpu/intel/jit/ir/send_builder.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef GPU_INTEL_JIT_IR_MESSAGE_HPP
 #define GPU_INTEL_JIT_IR_MESSAGE_HPP
 
+#include "gpu/intel/jit/ir/send.hpp"
```
echeresh, line 20: Having both `message.hpp` and `send.hpp` looks confusing to me as message/send words can be used interchangeably.

Maybe rename `message.hpp` -> `send_builder.hpp` (or something else with `send` to align terminology)?

src/gpu/intel/jit/dsl/dsl.cpp
rjoursler: Good point, this had slipped my mind while I was refactoring it.

rjoursler:
make test
disable test_device_cpu

atkassen (approved):

PR #4259, opened 2025-10-31
ci: aarch64: improve effectiveness of t-test

murste01:
# Description

This change reduces the time limit for CI benchdnn problems to 100ms, and increases the number of iterations to 50. This should make regression testing up to ~3x faster, and make our t-tests more effective by taking more samples.

The choice of 100ms and 50 iterations was found experimentally. We may be able to reduce the time limit in future, but I think it's best that we choose a conservative number until this has been battle-tested.

### DIFF
--- a/.github/automation/performance/benchdnn_comparison.py
+++ b/.github/automation/performance/benchdnn_comparison.py
@@ -151,7 +151,7 @@ def compare_two_benchdnn(file1, file2, out_file=None):
 
         if out_file is not None and (exec_regressed or exec_improved):
             prb_params = [x.replace("--", "") for x in prb.split(" ")]
-            prb_params = [prb_params[1]] + [
+            prb_params = [prb_params[2]] + [
                 x for x in prb_params if ("dt=" in x) or ("alg=" in x)
             ]  # filter out the problem and data types
             prb_str = (


Ryo-not-rio (changes requested):
Could you upload this to a branch in oneDNN (not a fork) so that we can test the nightly on it? Also the precommit ci is flagging a new regresion with this change: `----The following exec time regression tests failed:----
--mode=P --max-ms-per-prb=100 --conv --dir=FWD-D mb1ic64ih200iw267oc256oh200ow267kh1kw1ph0pw0 exec: 5.29 → 13.6 (p=1.72e-259)`. This might be a false positive that's being caused by this change

Ryo-not-rio:
Running nightly here: https://github.com/uxlfoundation/oneDNN/actions/runs/19066389368

Ryo-not-rio:
It looks like the wiki table is a bit broken with this change - see c7g: https://github.com/uxlfoundation/oneDNN/wiki/AArch64-status

vpirogov:
On a side note, I wonder why most of these wiki updates is attributed to my account...
image

murste01:
Thanks @Ryo-not-rio for running this change on a branch; it looks like the regressions flagged in that pipeline shouldn't be:

```
--mode=P --max-ms-per-prb=100 --eltwise --alg=relu-dst --alpha=0 --beta=0 384x384 exec: 0.00244 → 0.00244 (p=3.64e-14)
--mode=P --max-ms-per-prb=100 --eltwise --tag=axb --alg=relu-dst --alpha=0 --beta=0 384x384 exec: 0.00244 → 0.00244 (p=2.69e-14)
```

I can take a look at this ~soon.

Sqvid:
> On a side note, I wonder why most of these wiki updates is attributed to my account...

@vpirogov I just assumed you were extremely passionate about our performance metrics.

murste01:
@Ryo-not-rio Was that table breakage due to #4273?

It looks like that nightly run reported exec regressions incorrectly:

```
 --mode=P --max-ms-per-prb=100 --eltwise --alg=relu-dst --alpha=0 --beta=0 384x384 exec: 0.00244 → 0.00244 (p=3.64e-14)
--mode=P --max-ms-per-prb=100 --eltwise --tag=axb --alg=relu-dst --alpha=0 --beta=0 384x384 exec: 0.00244 → 0.00244 (p=2.69e-14)
```

Neither appears to have actually regressed? Unless this is just due to how they are being reported.

Edit: I see these are median values, which aren't necessarily representative of the perf distribution.

Ryo-not-rio:
You'll see the table breakages locally if you run `benchdnn_comparison.py` with the `--out-file` flag, the wiki here updates every day so your changes have been overwritten 😢 the prb column in the table is formatted incorrectly with this change, instead of showing the problem, it's just showing `--max-ms-per-prb=`

Ryo-not-rio:
@vpirogov Looks like the nightly runs are attributed to you so that must be why it looks like you're making so many wiki changes 😄

murste01:
It looks like the nightly regression is always the same problem. I've rebased to see if this is due to an unlucky base commit.

murste01:
Rebased 🤞

Ryo-not-rio:
nightly triggered: https://github.com/uxlfoundation/oneDNN/actions/runs/19462607307

PR #4263, opened 2025-10-31
cpu: x64: matmul: brgemm: add zp per_oc support for dst and wei

pmanczak:
[MFDNN-4310](https://jira.devtools.intel.com/browse/MFDNN-4310)
Enabled per_oc zero-point support for destination and weights.
For weights, enforcing n_blk=1 was required

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul.cpp
@@ -251,14 +251,18 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
 
     auto check_attr_zero_points = [&]() -> bool {
         const auto &zp = attr()->zero_points_;
-        static const std::vector<int> supported_args {
-                DNNL_ARG_SRC, DNNL_ARG_DST};
-        for (int arg : supported_args) {
-            if (!zp.has_default_values(arg)) {
-                const int mask = zp.get_mask(arg);
-                if (mask > 0) return false;
-            }
+
+        if (!zp.has_default_values(DNNL_ARG_SRC)) {
+            const int mask = zp.get_mask(DNNL_ARG_SRC);
+            if (mask > 0) return false;
         }
+
+        if (!zp.has_default_values(DNNL_ARG_DST)) {
+            const int mask = zp.get_mask(DNNL_ARG_DST);
+            const int dst_n_mask = dst_qmask_N();
+            if (mask != 0 && mask != dst_n_mask) return false;
+        }
+
         if (!zp.has_default_values(DNNL_ARG_WEIGHTS)) {
             const auto mask = zp.get_mask(DNNL_ARG_WEIGHTS);
             const auto kn_mask = wei_qmask_N() + wei_qmask_K();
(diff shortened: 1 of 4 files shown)


src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -190,6 +190,7 @@ struct brgemm_matmul_conf_t {
     dim_t s8s8_comp_b_str;
     dim_t s8s8_comp_n_str;
     bool has_zero_point_a, has_zero_point_b, has_zero_point_c;
+    bool has_zero_point_c_per_oc, has_zero_point_b_per_oc;
```
dzarukin, line 193: There's no need in these variables, they must be used through `bgmmc.wei_zp_type` and `bgmmc.dst_zp_type` instead.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1860,6 +1879,10 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     // This is the only implementation that support the packed_sparse_weights
     // case therefore there is no fallback for it.
     is_small_shapes = is_small_shapes && !bgmmc.packed_sparse_weights;
+
+    // For per_n zp compensation calculation requires N_blk=1 which only BRGEMM supports
```
dzarukin, line 1883: Comments must be 80-symbols in length same as code, please adjust.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1860,6 +1879,10 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    const bool requires_brgemm = bgmmc.wei_zp_type == brgemm_broadcast_t::per_n;
+    is_small_shapes = is_small_shapes && !requires_brgemm;
```
dzarukin, line 1885: Why is this needed?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1723,7 +1733,16 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     VCHECK_BG(compute_blocking_heuristic(bgmmc, bm_conf_utils),
             VERBOSE_BLOCKING_FAIL, "");
 
-    if (bgmmc.wei_n_blk > bgmmc.N_blk && bgmmc.N != bgmmc.N_blk) {
+    // For per_n wei zero points, force N_blk=1 to ensure each N gets its own compensation value
+    if (bgmmc.wei_zp_type == brgemm_broadcast_t::per_n && !bgmmc.is_runtime_N) {
+        bgmmc.N_blk = 1;
```
dzarukin, line 1738: This doesn't make any sense performance-wise, it should be approached somewhat differently.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -690,9 +690,19 @@ format_tag_t brgemm_matmul_conf_utils_t::pick_blocked_B_layout(
...
-            ? brgemm_broadcast_t::none
-            : brgemm_broadcast_t::per_tensor;
+    if (attr.zero_points_.has_default_values(arg)) {
+        return brgemm_broadcast_t::none;
+    }
+
+    const int mask = attr.zero_points_.get_mask(arg);
+    if (mask == 0) {
+        return brgemm_broadcast_t::per_tensor;
+    } else if (mask == 2
```
dzarukin, line 700: mask=2 only covers 2D cases. What about 3D and 4D cases, have you tested those?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -396,10 +396,13 @@ void jit_brgemm_matmul_copy_a_impl_t<Vmm>::copy_K_loop(
         }
 
         // step 3: multiply by zp_b_val
-        mov(reg_zp_b_neg_val_ptr, ptr[param1 + GET_OFF(zp_b_neg_val_ptr)]);
-        const auto vmm_zp_b_neg_val = get_vmm_comp_acc(is_ymm_ ? 2 : 1);
-        uni_vbroadcastss(vmm_zp_b_neg_val, ptr[reg_zp_b_neg_val_ptr]);
-        uni_vpmulld(get_vmm_comp_acc(0), get_vmm_comp_acc(0), vmm_zp_b_neg_val);
+        if (conf_->wei_zp_type != brgemm_broadcast_t::per_n) {
```
dzarukin, line 399: It should rather specify which zp_type it supports. Same applies to the code changed below.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2471,8 +2544,11 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
     int32_t zero_point_a_negative_val_;
     int32_t zero_point_mixed_ab_compensation_component_;
     int32_t zero_point_c_val_;
-    int32_t wei_zp_neg_val_;
+    int32_t wei_zp_neg_val_, wei_zp_val_;
     const void *wei_zp_ptr_;
+    const void *zero_point_c_per_oc_ptr_;
+    const void *zero_point_b_per_oc_ptr_;
+    std::vector<int32_t> zero_point_b_per_oc_buf_;
```
dzarukin, line 2551: This is not allowed, all memory operations must follow the scratchpad design. Please re-do.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: In addition there's already a block of attributes related to ZPs for B: 
```cpp
   // Zero points
    dim_t wei_zp_k_gsize = 0;
    bool is_wei_zp_per_k = false;
    bool is_wei_zp_per_n = false;
```
For the reference: https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/x64/matmul/brgemm_matmul_utils.hpp#L246

PR #4266, opened 2025-10-31
cpu: x64: matmul: enable gemv code path for m=1

densamoilov:
This PR enables the M=1 case when tensor B is transposed by leveraging the recently introduced GEMV code path for the N=1 case.
The idea is to swap the original tensors A and B, effectively turning it into the N=1 case.

One important point is the extended support for bias in the brgemv interface. It now supports a non-broadcast way of applying bias to handle the swapped case. Since we use the N=1 code path and swap M and N, the destination vector for the swapped case has a length of M, and the bias vector also has a length of M. Therefore, we need to apply the bias elementwise.

Performance data shows that the performance of the autogenerated GEMM and this new case is on par.

image

### DIFF
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -362,6 +362,7 @@ struct brgemm_desc_t {
     bool is_runtime_ldd = false;
 
     bool is_gemv = false;
+    bool treat_y_as_row = false;
 
     static constexpr int MAX_VPAD = 100;
     static constexpr int AMX_TILES_NUM = 8;
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
@@ -251,6 +251,11 @@ struct brgemm_matmul_conf_t {
     data_type_t wei_zp_dt = data_type::undef;
 
     bool is_gemv = false;
+    // Currently, it's only used to enable the N=1 code path for M=1, when B
+    // is transposed.
+    // TODO: Generalize when a new code path to support M=1, when B is plain
+    // is added.
+    bool gemv_swap_a_b = false;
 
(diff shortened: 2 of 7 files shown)


densamoilov:
make test

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -362,6 +362,7 @@ struct brgemm_desc_t {
     bool is_runtime_ldd = false;
 
     bool is_gemv = false;
+    bool broadcast_bias = true;
```
ankalinin: May be better to add this parameter to the brgattr along with all the other parameters?

densamoilov: I'm not sure about it. The brgattr seems to control the basic kernel behavior while bias is the post-op parameter. So if we start controlling post-op behavior with brgattr then we will end up with two abstractions controlling post-ops behavior.

The parameter is also gemv specific and cannot be defined for non-gemv cases. So I thought it would be better to have it in the BRGEMV interface, and use it only if post-ops have bias and ignore otherwise.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -251,6 +251,11 @@ struct brgemm_matmul_conf_t {
     data_type_t wei_zp_dt = data_type::undef;
 
     bool is_gemv = false;
+    // Currently, it's only used to enable the N=1 code path for M=1, when B
+    // is transposed.
+    // TODO: Generalize when a new code path to support M=1, when B is plain
+    // is added.
+    bool swap_a_b = false;
```
dzarukin: Minor: maybe for clarity and smaller scope name it `gemv_swap_a_b`?

densamoilov: ok, let's make it gemv specific for now

PR #4267, opened 2025-10-31
xe: jit: remove non-determinism between tool chains in blocking generator

rjoursler:
While investigating the regression identified in [MFDNN-14350](https://jira.devtools.intel.com/browse/MFDNN-14350), I found two configurations resulted in the same score according to the performance model. Due to non-determinism in how these configurations are generated, different configurations can be chosen for that problem. This PR removes the non-determinism (although it happens to pick the bad case for MFDNN-14350, meaning more work will be required).


rjoursler:
make test
disable test_device_cpu

rjoursler:
make test perf-gpu

rjoursler:
make test perf-gpu
set primitive=gpu:conv

rjoursler:
make test perf-gpu
set primitive=gpu:conv

rjoursler:
make test
disable test_device_cpu

PR #4277, opened 2025-11-04
common: update memory_arg_t abstraction to prolong memory lifetime

dzarukin:
`memory_arg_t` starts tracking ref_count available in `memory_t` object to make sure that memory objects allocated on stack for nested primitives are lifetime aligned with places use them asynchronously.

### DIFF
--- a/src/common/primitive_exec_types.cpp
+++ b/src/common/primitive_exec_types.cpp
@@ -21,8 +21,59 @@
 #include "common/memory_tracking.hpp"
 #include "common/primitive_desc.hpp"
 
+#include <cassert>
+
 namespace dnnl {
 namespace impl {
+
+memory_arg_t::memory_arg_t(memory_t *mem, bool is_const)
+    : mem_(mem), is_const_(is_const) {
+    if (mem_) mem_->retain();
+}
+
+memory_arg_t::~memory_arg_t() {
+    if (mem_) mem_->release();
+}
+
+memory_arg_t::memory_arg_t(const memory_arg_t &other)
+    : mem_(other.mem_), is_const_(other.is_const_) {
+    if (mem_) mem_->retain();
+}
+
+memory_arg_t &memory_arg_t::operator=(const memory_arg_t &other) {
+    if (this == &other) return *this;
+
+    if (mem_) memory_deleter_t {}(mem_);
+
+    mem_ = other.mem_;
+    is_const_ = other.is_const_;
+    if (mem_) mem_->retain();
+
+    return *this;
+}
+
+memory_arg_t::memory_arg_t(memory_arg_t &&other)
+    : mem_(other.mem_), is_const_(other.is_const_) {
+    other.mem_ = nullptr;
+    other.is_const_ = false;
+}
+
+memory_arg_t &memory_arg_t::operator=(memory_arg_t &&other) {
+    if (this == &other) {
+        assert(!"self move assign is not expected");
+        return *this;
+    }
+
+    if (mem_) memory_deleter_t {}(mem_);
+
+    mem_ = other.mem_;
+    is_const_ = other.is_const_;
+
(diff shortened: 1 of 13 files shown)


dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

src/common/primitive_exec_types.cpp
```diff
@@ -24,6 +24,44 @@
...
+
+    return *this;
+}
+
+memory_arg_t::memory_arg_t(memory_arg_t &&other) {
+    *this = std::move(other);
+}
+
+memory_arg_t &memory_arg_t::operator=(memory_arg_t &&other) {
+    if (this == &other) return *this;
```
densamoilov: self-move usually indicates a bug (moving from itself?) so it would be better to use `assert` to catch such scenarios as those would be bugs at the higher level.

src/common/primitive_exec_types.cpp
```diff
@@ -24,6 +24,44 @@
...
+    mem_ = other.mem_;
+    is_const_ = other.is_const_;
+    if (mem_) mem_->retain();
+
+    return *this;
+}
+
+memory_arg_t::memory_arg_t(memory_arg_t &&other) {
+    *this = std::move(other);
+}
```
densamoilov: By the time operator= is called, the object has already been created, so it’s usually advisable to use an initializer list instead of the move assignment operator.

```suggestion
memory_arg_t::memory_arg_t(memory_arg_t &&other)
    : mem_(other.mem_), is_const_(other.is_const_) {
        other.mem_ = nullptr; 
    }
```

src/common/primitive_exec_types.cpp
```diff
@@ -24,6 +24,44 @@
...
+
+memory_arg_t &memory_arg_t::operator=(memory_arg_t &&other) {
+    if (this == &other) return *this;
+
+    // Note: swapping memory objects lets dtor to handle original `mem_`.
+    std::swap(mem_, other.mem_);
+    is_const_ = other.is_const_;
```
densamoilov: Maybe handle the logic explicitly to keep it simple? Something along the lines:

```cpp
    memory_deleter_t{}(mem_);

    mem_= other.mem_;
    is_const_= other.is_const_;

    other.mem_ = nullptr;

    return *this;
```

src/common/primitive_exec_types.cpp
```diff
@@ -24,6 +24,44 @@
...
+
+memory_arg_t::memory_arg_t(const memory_arg_t &other)
+    : mem_(other.mem_), is_const_(other.is_const_) {
+    if (mem_) mem_->retain();
+}
+
+memory_arg_t &memory_arg_t::operator=(const memory_arg_t &other) {
+    if (this == &other) return *this;
+
+    mem_ = other.mem_;
```
densamoilov, line 48: leak?

src/common/primitive_exec_types.cpp
dzarukin: Good to know about common practices. Added an assert, thanks.

src/common/primitive_exec_types.cpp
dzarukin: Makes sense, updated move ctor implementation.

src/common/primitive_exec_types.cpp
dzarukin: Added a cleanup of original memory, thanks for spotting. It seems we don't have such cases as original testing didn't highlight any leaks.

src/common/primitive_exec_types.cpp
dzarukin: Used the proposed version.

densamoilov (approved):

echeresh (approved):

PR #4278, opened 2025-11-04
tests: benchdnn: catch and print SYCL graph execution exceptions

spalicki:
Catch and print SYCL graph execution exception for quicker debugging.

### DIFF
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -389,13 +389,20 @@ int execute_and_wait(perf_function_t &exec_func, const dnnl_engine_t &engine,
                 {sycl::ext::oneapi::experimental::property::graph::
                                 assume_buffer_outlives_graph {}}};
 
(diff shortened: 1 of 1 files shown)


tests/benchdnn/dnnl_common.cpp
```diff
@@ -389,13 +389,20 @@ int execute_and_wait(perf_function_t &exec_func, const dnnl_engine_t &engine,
...
+        try {
+            graph.begin_recording(queue);
+            status = exec_func(stream, dnnl_args);
+            graph.end_recording(queue);
+            DNN_SAFE(dnnl_stream_wait(stream), CRIT);
+
+            auto exec = graph.finalize();
+            queue.ext_oneapi_graph(exec).wait();
+        } catch (const sycl::exception &e) {
+            BENCHDNN_PRINT(2, "%s %s\n",
```
dzarukin: Zero should be better, it's not an expected outcome.

avmanerikar (approved):

PR #4281, opened 2025-11-04
cpu: rv64: postops: add rvv postops eltwise op support

zhangjian29:
# Description

This PR adds **eltwise operation support** for `rvv_postops.hpp` using a primitive-based approach according to previous proposal #4029.

## Motivation

Following the successful integration of primitive-based `binary` post-ops into `rvv_postops.hpp`, our next step is to introduce `eltwise` type post-ops using the same primitive-based paradigm. This is crucial for establishing a general `post_ops` implementation for the RISC-V 64-bit architecture at this early stage.

## Key Changes

- Provided `eltwise` type of post-ops by integrating with `rvv_eltwise` in `init` and `execute` methods.
- Kept the original `post_ops_ok` and `apply` methods unchanged in a fused way, making other implementation used original version still correct.
- Added comments about the original `post_ops_ok` and `apply` methods for other future implementations.
- Currently supported post-ops: 
    1. `relu` without `alpha/beta` (via original methods)
    2. Any `eltwise` operations the `rvv_eltwise` supported: `relu` with `alpha/beta`, `square`, `abs`, `sqrt`, `linear`, `clip`, `hardsigmoid`, and `hardswish`.
    3. Any `binary` operations: `add`, `div`, `max`, `min`, `mul`, `sub`, `ge,` `gt`, `le`, `lt`, `eq`, `ne`, and the ternary `select`.
- Demonstrated functionality by adding primitive-style eltwise postop support to `rvv_nchw_pooling`

## Future Plans (on-going)

1. Update post-ops support for `rvv_matmul` in this primitive-based approach.
2. Update post-ops support for `rvv_gemm_convolution` in this primitive-based approach.
3. Remove the original `post_ops_ok` and `apply` methods safely.
4. Add support for `sum` operations
5. Extend data type supports.

# Checklist

## General

- [test_postops_eltwise_ctest.log](https://github.com/user-attachments/files/23328620/test_postops_eltwise_ctest.log)

## Performance improvements

We test the changes using the RISC-V GNU toolchain(14.2) and verify the functionality under the sg2044 platform with cmd:

```
./tests/benchdnn/benchdnn --pool --dir=FWD_I --attr-post-ops=relu,square,abs,sqrt,linear:0.5:-1,clip,hardsigmoid,hardswish,add:f32,div:f32,mul:f32,max:f32,min:f32,sub:f32,ge:f32,gt:f32,le:f32,lt:f32,eq:f32,ne:f32,select:f32 --batch=../inputs/pool/shapes_basic
```

All tests passed with proper dispatching confirmed.

- **Before the changes:** [test_postops_without_eltwise.log](https://github.com/user-attachments/files/23328646/test_postops_without_eltwise.log)

- **After the changes:** [test_postops_with_eltwise.log](https://github.com/user-attachments/files/23328649/test_postops_with_eltwise.log)

### DIFF
--- a/src/cpu/rv64/rvv_postops.hpp
+++ b/src/cpu/rv64/rvv_postops.hpp
@@ -57,6 +57,9 @@ struct rvv_postops_t {
         post_op_primitives_.clear();
         po_ = local_post_ops;
 
+        const int num_post_ops = local_post_ops.len() - post_op_start_index_;
+        if (num_post_ops > 0) post_op_primitives_.reserve(num_post_ops);
+
         for (int i = post_op_start_index_; i < local_post_ops.len(); i++) {
             auto &po = local_post_ops.entry_[i];
 


src/cpu/rv64/rvv_postops.hpp
```diff
@@ -79,6 +80,27 @@ struct rvv_postops_t {
...
+                auto empty_attr = dnnl_primitive_attr();
+                primitive_desc_iterator_t it(engine,
+                        reinterpret_cast<const op_desc_t *>(&po_desc),
+                        &empty_attr, nullptr);
+                if (++it == it.end()) return status::unimplemented;
+
+                std::shared_ptr<primitive_desc_t> elt_pd = *it;
+                std::shared_ptr<primitive_t> elt_prim;
+                CHECK(elt_pd->create_primitive(elt_prim, engine));
+                post_op_primitives_.push_back(elt_prim);
```
dzarukin, line 105: This would be bad for performance if there are more than 1 post-op in the chain. I suggest to reserve `len()` elements in the vector before getting into the for loop.

Besides, I'm not sure this is the right approach. Optimized version want to do such fusions on registers, not on memory which is not different from the case when the user does it not applying post-ops but applying plain primitives. For non-optimized approach there's already `ref_post_ops_t` struct that mostly handles everything for you. Feel free to inspect existing reference implementations with this struct and let me know you opinion on this one.

zhangjian29: Fixed.

zhangjian29:
> Besides, I'm not sure this is the right approach. Optimized version want to do such fusions on registers, not on memory which is not different from the case when the user does it not applying post-ops but applying plain primitives. For non-optimized approach there's already ref_post_ops_t struct that mostly handles everything for you. Feel free to inspect existing reference implementations with this struct and let me know you opinion on this one.

@dzarukin Thank you for your insightful suggestion. We propose leveraging the `rvv_eltwise` intrinsic kernel to support a broader range of eltwise post-ops, where the optimal code reuse strategy is to follow the primitive creation and execution flow. While this approach incurs some runtime overhead compared to the existing inlined `apply` method for `eltwise_relu` in most cases, it offers a more comprehensive solution that supports additional post-ops with better performance than `ref_post_ops`.

To validate this approach, we conducted experiments on `rvv_matmul` with three different implementations for `eltwise_relu` post-op support, demonstrating acceptable performance trade-offs:

### Performance Comparison (time in ms)

| Benchmark | Inlined `apply` (ms) | `rvv_postops` primitive (ms) |`ref_post_ops` primitive (ms) |
|-----------|-----------|-----------|-----------|
| shapes_converted_ip_inf_lb_wd | 196.45 |215.94 (+10%) | 236.88 (+21%) |
| shapes_converted_ip_inf_lb_googlenet | 279.26 | 289.40 (+4%) | 333.20 (+19%) |
| shapes_converted_ip_inf_lb_resnet | 120.49 | 130.47 (+8%) | 140.00 (+16%) |
| shapes_transformer | 139.21 | 149.82 (+8%) | 500.21 (+260%) |
| shapes_converted_ip_inf_lb_vgg16 | 4056.24 | 4044.09 (**-0.3%**) | 5416.81 (+34%) |
| shapes_converted_ip_inf_lb_ncf | 38.12 | 39.80 (+4%) | 77.51 (+103%) |
| shapes_converted_ip_inf_lb_maskrcnn | 3821.15 | 4143.82 (+8%) | 4319.36 (+13%) |
| shapes_converted_ip_inf_lb_rnn_t | 4765.07 | 5421.32 (+14%) | 5924.83 (+24%) |
| shapes_converted_ip_inf_lb_dlrm | 1380.79 | 1263.95 (**-8%**) | 1732.53 (+25%) |

Moving forward, we plan to maintain both the inlined `apply` method and the intrinsic primitive-based approach in `rvv_postops` to serve different design purposes and provide flexibility for future optimizations.

dzarukin (approved):

PR #4282, opened 2025-11-04
xe: jit: break codegen dependency on func_impl_t

rjoursler:
Adds a new interface  which is intended to be used for code generation with custom IR objects. The main benefit here is to break the codegen dependency on oneDNN post-ops which is required for upstreaming.


rjoursler:
Closing as this is being moved into #4324.

PR #4284, opened 2025-11-04
converter: update ukernel to provide a working benchdnn line

dzarukin:
[MFDNN-14361](https://jira.devtools.intel.com/browse/MFDNN-14361)

ukernel juggles with wtag, strides and ld options when it comes to ukernel API. Updated the routine to provide relevant options.

### DIFF
--- a/scripts/verbose_converter/tests/benchdnn_test.py
+++ b/scripts/verbose_converter/tests/benchdnn_test.py
@@ -92,7 +92,9 @@ def filter_verbose(verbose: IO[str], driver: str, filter_event: str):
                 prop_kind = parts[7]
             except IndexError:
                 continue
-            if component != "primitive" or event not in filter_event:
+            if component != "primitive" and component != "ukernel":
+                continue
+            if event not in filter_event:
                 continue
             if get_driver(primitive) != driver:
                 continue


scripts/verbose_converter/src/benchdnn_generator.py
```diff
@@ -467,6 +467,43 @@ def aux(self):
...
+                    strides.append(md.strides)
+                    lds.append("")
+                else:
+                    tags.append(f"--{arg[0]}tag={tag}")
+                    strides.append("")
+                    lds.append("")
+            else:
+                ld_letter = "a"
+                if arg == "dst":
+                    ld_letter = "d"
```
atkassen: Unused?

scripts/verbose_converter/src/benchdnn_generator.py
```diff
@@ -458,7 +458,7 @@ def tags(self):
         return f"--stag={src0_tag}:{src1_tag} --dtag={dst_tag}"
 
 
-class BRGEMMConverter(MultiDataTypeMixin, Converter):
+class BRGEMMConverter(StridesMixin, MultiDataTypeMixin, Converter):
```
atkassen: Not necessarily needed, since you're not using `StridesMixin.tags`.

spalicki (approved):

scripts/verbose_converter/src/benchdnn_generator.py
dzarukin: Leftover from the first attempt, thanks for noticing, removed.

scripts/verbose_converter/src/benchdnn_generator.py
dzarukin: I see, removed then, thanks.

PR #4287, opened 2025-11-05
[Backport] gpu: intel: gemm: jit: fix n range for xe2/xe3 gemm kernel (rls-v3.9)

h-sadia:
# Description

Fixing gemm n- range for the FHS kernel. As performance with this range is better compared to existing range.

The range matches that of rls-v3.10 which was merged earlier.

Look at results in #4288 for more detail.


h-sadia:
make test perf-gpu
set primitive=gpu:gemm

syurkevi (approved):

PR #4298, opened 2025-11-07
Xe2: fix BMG qwen3 fp8 regression

Yejing-Lai:
After this #3693 ,  BMG FP8 will not fallback to PVC strategies when N>128. It caused the regression for qwen3 for some specific shapes.
This PR aims to add some strategies to fix these regressions.


CaoZhongZ:
@Simonsays095

Simonsays095:
Can you please add some performance data for a few specific layers that these changes should affect?

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Yejing-Lai:
> Can you please add some performance data for a few specific layers that these changes should affect?

I mainly improved the case on qwe3-30B and qwen3-32B. Following table with avg_time from benchdnn --mode=p: 
image

PR #4300, opened 2025-11-07
[docs] x64: matmul: Update weight decompression docs & examples

inteldimitrius:
# Matmul weight decompression
Docs and examples related to matmul weight decompression updated.

### DIFF
--- a/doc/primitives/matmul.md
+++ b/doc/primitives/matmul.md
@@ -98,12 +98,11 @@ types for source, destination, weights, and bias tensors:
 | Source           | Weights                                | Destination                      | Bias                        |
 |:-----------------|:---------------------------------------|:---------------------------------|:----------------------------|
 | f64              | f64                                    | f64                              | f64, f32, f16, bf16, s8, u8 |
-| f32              | f32                                    | f32                              | f32, bf16, f16, u8, s8      |
+| f32              | f32, u8, s8, u4, s4                    | f32                              | f32, bf16, f16, u8, s8      |
 | f16              | f16, u8, s8, u4, s4                    | f16, u8, s8                      | f32                         |
-| f16              | f16, u8, s8                            | f32                              | f32, f16                    |
+| f16              | f16, u8, s8, u4, s4                    | f32, f16                         | f32, f16                    |
 | bf16             | bf16, u8, s8, u4, s4                   | f32, bf16                        | f32, bf16                   |
-| f32, bf16, f16   | u8, s8                                 | f32, bf16, f16                   | f32, bf16, f16              |
-| f32, bf16, f16   | u8, s8                                 | f32, bf16, f16                   | f32, bf16, f16              |
+| f32, bf16, f16   | u8, s8, u4, s4                         | f32, bf16, f16                   | f32, bf16, f16              |
 | bf16, f16        | f8_e5m2, f8_e4m3, f4_e2m1, f4_e3m0     | f32, f16, bf16                   | f32, bf16, f16              |
 | f8_e5m2, f8_e4m3 | f8_e5m2, f8_e4m3                       | f32, f16, bf16, f8_e5m2, f8_e4m3 | f32, bf16, f16              |
 | f4_e2m1, f4_e3m0 | f4_e2m1, f4_e3m0                       | f32, f16, bf16, f4_e2m1, f4_e3m0 | f32, bf16, f16              |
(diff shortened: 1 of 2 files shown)


doc/primitives/matmul.md
```diff
@@ -156,9 +155,10 @@ The following attributes and post-ops are supported:
 | Post-op   | [Prelu](@ref dnnl::post_ops::append_prelu)                     | Applies an @ref dnnl_api_prelu operation to the result                        |                                     |
 
 The following masks are supported by the primitive:
-- 0, which applies one scale / zero point value to an entire tensor, and
-- 2, which applies a scale value per column along the
-  `n`dimension for `DNNL_ARG_WEIGHTS`.
+- 0, which applies one scale / zero point value to an entire tensor
+- 1, which applies a scale / zero point values along `k`-dimension
```
mzhukova, line 159: Should we add that this description is also related to weights arg? 

Also, in the ~decompression example you're using a combination K and N mask `(1 << 0) + (1 << 1)` that is not described here, should there be a some sort of a note that multi-dimensional granularity is also supported?

examples/tutorials/matmul/weights_decompression_matmul.cpp
```diff
@@ -96,15 +96,15 @@ matmul::primitive_desc matmul_pd_create(
...
     primitive_attr attr;
-    // Set scales with multiple scales along K and N dimensions and with groups along K.
+    // Set scales with multiple values along K and N dimensions and with groups along K.
     attr.set_scales(DNNL_ARG_WEIGHTS,
             /* mask */ (1 << 0) + (1 << 1), {G, 1}, memory::data_type::f32);
-    // Set a single zero point with s8 data type.
-    attr.set_zero_points(
-            DNNL_ARG_WEIGHTS, /* mask */ 0, {}, memory::data_type::s8);
+    // Set zero points with multiple values along K and N dimensions and with groups along K.
+    attr.set_zero_points(DNNL_ARG_WEIGHTS, /* mask */ (1 << 0) + (1 << 1),
```
mzhukova, line 103: Do we not need to also change a memory object for zero points somewhere in the example..?

examples/tutorials/matmul/weights_decompression_matmul.cpp
```diff
@@ -96,15 +96,15 @@ matmul::primitive_desc matmul_pd_create(
...
+            {G, 1}, memory::data_type::s8);
     // Set fpmath mode with `apply_to_int=true` to apply fpmath mode behavior to
     // integral primitives (in this example, matmul).
-    attr.set_fpmath_mode(fpmath_mode::bf16, true);
+    attr.set_fpmath_mode(fpmath_mode::strict, true);
```
mzhukova, line 107: Could you explain this change?

dzarukin: I think Dmitry had in mind f32:int8:f32 case where fpmath mode must be strict. I guess, the right approach is to decide on the fpmath mode based on the src data type.

doc/primitives/matmul.md
inteldimitrius: Masks `1` & `2` are supported only for weights arg, I have added a note to  `1` in the following push. Thanks :)
However, a granularity is supported only for `k`-dimension for both scales and zero points and in the example the second dimension is always equals to one (`{G, 1}`) and will throw an error otherwise. 
I have added a comment in regards to it.

examples/tutorials/matmul/weights_decompression_matmul.cpp
inteldimitrius: Yes, that's true. Updated!

examples/tutorials/matmul/weights_decompression_matmul.cpp
inteldimitrius: Yes, @dzarukin is right, in the example f32:int8:f32 datatype combination is used and it requires fpmath_mode=strict.

examples/tutorials/matmul/weights_decompression_matmul.cpp
mzhukova: Hmm, it seems like the example passed before, even though dimensions of mem obj for zp were not in sync with this setting.. Do you have an understanding why it worked? 👀

examples/tutorials/matmul/weights_decompression_matmul.cpp
mzhukova: Oh I see, thanks for the clarification :)

examples/tutorials/matmul/weights_decompression_matmul.cpp
inteldimitrius: That's a very good point. I briefly looked at it and it seems runtime dimensions do not verified with primitive descriptor inputs at least for matmul CPU primtive. 

It doesn't really related to this PR, however, this check definitely makes sense.

mzhukova: Right, it is not related..  Could you please submit bug/short reproducer, so that we could take care of it eventually? (feel free to resolve this thread afterwards).

inteldimitrius: Hi, the task was created, so I'm resolving this thread,

vpirogov (approved):

PR #4302, opened 2025-11-07
graph: sdpa: fix GQA check that limits shapes to 4D tensors

umar456:
This PR fixes an issue where we were limiting the input query shape to 4D tensors which is not valid in GQA workloads.


umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
disable test_device_cpu
enable test_device_gpu

PR #4308, opened 2025-11-10
gpu: intel: compute: fix setting 4gb flag

spalicki:
Named buffers data type was not always set which led to incorrect buffer size calculation:
1. Added missing data types
2. Used memory wrapper `size()` method to calculate buffer size
3. Fixed `USE_INT32_OFFSET` option setting, which should be set based on number of elements and not size in bytes

Fixes: [MFDNN-14379](https://jira.devtools.intel.com/browse/MFDNN-14379)

### DIFF
--- a/src/gpu/intel/reduction/atomic.cpp
+++ b/src/gpu/intel/reduction/atomic.cpp
@@ -234,6 +234,7 @@ status_t atomic_conf_t::init_dispatcher(
             dims::subgroup,
     };
     compute::named_buffer_t src("SRC");
+    src.data_type = conf.src_type;
     std::array<dim_t, 6> sizes = {
             outer_block.block,
             conf.global_acc,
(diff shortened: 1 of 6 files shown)


src/gpu/intel/reduction/atomic.cpp
```diff
@@ -234,6 +234,7 @@ status_t atomic_conf_t::init_dispatcher(
             dims::subgroup,
     };
     compute::named_buffer_t src("SRC");
+    src.data_type = conf.src_type;
```
dzarukin, line 237: Probably forcing data type through a ctor would be less error prone solution.

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

PR #4311, opened 2025-11-11
common: math_utils: fix is_prime() for n <= 5

Sqvid:
# Description

The previous implementation would return `false` for 2, 3, and 5

# Checklist

## General

### DIFF
--- a/src/common/math_utils.hpp
+++ b/src/common/math_utils.hpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2017 Intel Corporation
+* Copyright 2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


Sqvid:
cc: @dzarukin

src/common/math_utils.hpp
```diff
@@ -36,12 +37,16 @@ template <typename T>
 inline bool is_prime(T n) {
     static_assert(std::is_integral<T>::value == true, "Not an integral type");
 
-    if (n <= 1 || n % 2 == 0 || n % 3 == 0 || n % 5 == 0) return false;
+    if (n <= 1) { return false; }
+
+    if (n == 2 || n == 3) { return true; }
+
+    if (n % 2 == 0 || n % 3 == 0) { return false; }
```
dzarukin: You might want to keep 5 for performance considerations as in the initial version.

dzarukin (approved):

src/common/math_utils.hpp
Sqvid: Done.

PR #4315, opened 2025-11-12
cpu: aarch64: matmul: enable brgemm matmul bf16

michalowski-arm:
# Description

This enables brgemm matmul bf16 for aarch64 for 2-3 dimensional shapes; below are some benchdnn time perf numbers for bf16 matmul on c8g with 32 threads (in ms, lower is better):

| shape | brg:sve_128 | gemm:acl | gemm:jit |
| ------ | ------------ | ---------- | -------- |
| 16x16:16x16                 | 0.00387 | 0.0251  | 0.00433 |
| 64x64:64x64               | 0.00394 | 0.0386| 0.0430   |
| 128x128:128x128         | 0.00518 | 0.0384| 0.316     |
| 512x512:512x512         | 0.0803   | 0.0801 | 10.7        |
| 1024x1024:1024x1024| 0.625     | 0.393   | 72.9       |
| 4096x4096:4096x4096 | 41.6 | 23.6 | 4408 |

All tests (from the nightly test set) ran using `ctest -R matmul` were successful on both c7g and c8g instances.

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul.cpp
@@ -154,7 +154,8 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
     VDISPATCH_MATMUL(is_dense_format_kind(), VERBOSE_NONTRIVIAL_STRIDE);
     VDISPATCH_MATMUL(mayiuse(isa), VERBOSE_UNSUPPORTED_ISA);
     VDISPATCH_MATMUL(problem_dt_correct, VERBOSE_UNSUPPORTED_DT);
-    VDISPATCH_MATMUL(!(is_bf16 && !mayiuse_bf16()), VERBOSE_UNSUPPORTED_ISA);
+    VDISPATCH_MATMUL(
+            IMPLICATION(is_bf16, mayiuse_bf16()), VERBOSE_UNSUPPORTED_ISA);
     VDISPATCH_MATMUL(!has_zero_dim_memory(), VERBOSE_EMPTY_TENSOR, "");
     VDISPATCH_MATMUL(
             no_dynamic_strides_for_B_and_C, VERBOSE_RUNTIMEDIM_UNSUPPORTED);
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -223,6 +223,14 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(
     } else {
         switch (bgmmc.wei_dt) {
             case f32:
+                // If the B memory descriptor matches both the transposed
+                // and plain version, we chose to treat it as "plain" since
+                // that saves us the extra time and scratchpad memory we
+                // would need for an unnecessary transpose.
+                if (memory_desc_wrapper(B_md).is_canonical()) {
+                    bgmmc.wei_tag = plain_tensor_layout_tag;
+                    return status::success;
+                }
                 if (blocked_B_layouts_allowed) {
                     bgmmc.wei_tag = memory_desc_matches_one_of_tag(B_md,
                             plain_tensor_layout_tag, blocked_64n_B_layout_tag,


Sqvid:
> 1024x1024:1024x1024: brg = 0.625, acl = 0.393

Any idea why this shape seems to behave so differently from the rest? It makes me wonder how the acl impl would stack up if we split 4096x4096 into 4 subtiles.

michalowski-arm:
> > 1024x1024:1024x1024: brg = 0.625, acl = 0.393
> 
> Any idea why this shape seems to behave so differently from the rest?

If you mean why does brgemm become slower than acl:gemm here, it's because brgemm currently uses bfdot instruction instead of bfmmla. It seems that makes the difference from 512:512:512 onwards.

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -211,27 +221,32 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(
...
     } else {
-        bgmmc.wei_tag = blocked_B_layouts_allowed
-                ? memory_desc_matches_one_of_tag(B_md, plain_tensor_layout_tag,
-                        transposed_tensor_layout_tag, blocked_64n_B_layout_tag,
-                        blocked_48n_B_layout_tag, blocked_32n_B_layout_tag,
-                        blocked_16n_B_layout_tag)
-                : memory_desc_matches_one_of_tag(B_md, plain_tensor_layout_tag,
-                        transposed_tensor_layout_tag, acbd, adbc);
-
-        // If the B memory descriptor matches both the transposed and plain
```
Sqvid, line 222: This optimisation still makes sense to have doesn't it? Though we should probably use the new [`is_canonical()`](https://github.com/uxlfoundation/oneDNN/blob/6ecff7da7282c40b996cfee7d4ecaeb3b9230480/src/common/memory_desc_wrapper.hpp#L80-L113) method.

Sqvid:
> If you mean why does brgemm become slower than acl:gemm here, it's because brgemm currently uses bfdot instruction instead of bfmmla. It seems that makes the difference from 512:512:512 onwards.

That makes sense. Can we put in a heuristic to dispatch based on size+bf16 then (along with a comment)?

src/cpu/aarch64/matmul/brgemm_matmul.cpp
```diff
@@ -151,6 +151,7 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
     VDISPATCH_MATMUL(is_dense_format_kind(), VERBOSE_NONTRIVIAL_STRIDE);
     VDISPATCH_MATMUL(mayiuse(isa), VERBOSE_UNSUPPORTED_ISA);
     VDISPATCH_MATMUL(problem_dt_correct, VERBOSE_UNSUPPORTED_DT);
+    VDISPATCH_MATMUL(!(is_bf16 && !mayiuse_bf16()), VERBOSE_UNSUPPORTED_ISA);
```
Ryo-not-rio: could we simplify this to `!is_bf16 || maiuse_bf16()`?

Sqvid: Or even, better `IMPLICATION(is_bf16, mayiuse_bf16())`

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
michalowski-arm: as far as I understand `is_canonical()`, this should now have identical effect?

jondea:
> > If you mean why does brgemm become slower than acl:gemm here, it's because brgemm currently uses bfdot instruction instead of bfmmla. It seems that makes the difference from 512:512:512 onwards.
> 
> That makes sense. Can we put in a heuristic to dispatch based on size+bf16 then (along with a comment)?

We need to tread cautiously around heuristics around this, because I think `jit_bf16_matmul` is better than brgemm matmul, it just doesn't support many memory formats. So the picture around this is quite complicated and fast moving, it may not be worth the complexity of a heuristic until the dust settles

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -352,7 +384,7 @@ brgemm_broadcast_t get_zp_type(const primitive_attr_t &attr, int arg) {
             : brgemm_broadcast_t::per_tensor;
 }
 
-struct matmul_sve512_blocking_params_t {
+struct matmul_brgemm_blocking_params_t {
```
jondea, line 387: Congratulations on finding a word which is the same length as `sve512` so that you don't have to refactor anything ;)

dzarukin:
A friendly reminder that merge commits are not allowed in oneDNN.

@vpirogov, it feels that we can enforce this policy through actions, there is a more or less reliable way for our dev environment.

vpirogov:
> A friendly reminder that merge commits are not allowed in oneDNN.
> 
> @vpirogov, it feels that we can enforce this policy through actions, there is a more or less reliable way for our dev environment.

This policy is already enforced by branch protection rules. You are right that there's no reminder though. Maybe we can include it in commit message checker?

michalowski-arm:
> A friendly reminder that merge commits are not allowed in oneDNN.
> 
Yep, thanks for the reminder!

dzarukin:
> This policy is already enforced by branch protection rules. You are right that there's no reminder though. Maybe we can include it in commit message checker?

Good suggestion, added it here: https://github.com/uxlfoundation/oneDNN/pull/4492

src/cpu/matmul/cpu_matmul_list.cpp
```diff
@@ -65,8 +65,9 @@ constexpr impl_list_item_t impl_list[] = REG_MATMUL_P({
         CPU_INSTANCE_AARCH64_ACL(acl_lowp_matmul_t)
         CPU_INSTANCE_AARCH64_ACL(acl_matmul_t)
         CPU_INSTANCE_AARCH64(jit_bf16_matmul_t)
-        CPU_INSTANCE_AARCH64(brgemm_matmul_t<sve_256>)
         CPU_INSTANCE_AARCH64(jit_int8_matmul_t<sve_256>)
+        CPU_INSTANCE_AARCH64(brgemm_matmul_t<sve_256>)
```
dzarukin, line 69: I have no idea if it's intended (this part I couldn't see clarified in the PR description) but changing implementations in the list might require more perf data in the PR than it does provide (though I don't have a clear understanding if they have intersecting functional support).

Another thing that slightly disturbs me is that optimizations for bf16 are done for brgemm_matmul but jit_bf16 still stands higher in the list. What's the catch here?

michalowski-arm: jit_bf16 uses a mult instruction with twice the throughput than that of brgemm, but the implementation is currently limited to sve_256 or higher and bf16:bf16:f32 only.

PR #4316, opened 2025-11-12
aarch64: ci: report mean and standard error of the mean for regressions

murste01:
This changes the reported exec and ctime metrics in regressions from medians to mean+/-standard-error-of-the-mean. The addition of standard-error-of-the-mean means that regressions where the mean or median are identical will be easier to reason about.


murste01:
Thanks both, I'll need to rely on one of you to merge.

PR #4015, opened 2025-09-26
src: gpu: intel: gemm: jit: enable dual zero points via f16 upconversion

hidefromkgb:
This addresses [MFDNN-13918](https://jira.devtools.intel.com/browse/MFDNN-13918).

PR #4318, opened 2025-11-13
(v3.10 backport) src: gpu: intel: gemm: jit: enable dual zero points via f16 upconversion

hidefromkgb:
This is a backport for #4015.


hidefromkgb:
Cancelled.

PR #4319, opened 2025-11-13
graph: backend: dnnl: fix sanitizer reported failures

ElaineBao:
# Description

- [x] Fix MFDNN-14383: host_scalar_executable_t::execute_ocl leaks cl_event memory
- [x] Fix MFDNN-14370: use-of-uninitialized-value in thread_local_cache_t

### DIFF
--- a/src/graph/backend/dnnl/thread_local_cache.hpp
+++ b/src/graph/backend/dnnl/thread_local_cache.hpp
@@ -171,7 +171,7 @@ public:
 private:
     class global_cache_type_t {
     public:
-        global_cache_type_t() : counter_(1) {}
+        global_cache_type_t() : mutex_(), data_(), counter_(1) {}
         ~global_cache_type_t() = default;
         std::mutex &mutex() { return mutex_; }
         std::unordered_map<size_t, std::vector<std::shared_ptr<T>>> &data() {
@@ -181,14 +181,13 @@ private:
         static global_cache_type_t *get_global_cache() {
             // A global table to store cached values in ALL threads. This global
             // table takes the ownership of cached values
-            try {
-                static auto global_cache = std::shared_ptr<global_cache_type_t>(
-                        new global_cache_type_t {},
-                        [](global_cache_type_t *ptr) {
-                            return ptr->release();
-                        });
-                return global_cache.get();
-            } catch (...) { return nullptr; }
+            static std::once_flag flag;
+            static global_cache_type_t *global_cache = nullptr;
+
+            std::call_once(
+                    flag, []() { global_cache = new global_cache_type_t {}; });
+
+            return global_cache;
         }
 
         // This function increments the reference count


ElaineBao:
make test

src/graph/backend/dnnl/thread_local_cache.hpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2021 Intel Corporation
+ * Copyright 2021-2025 Intel Corporation
```
TaoLv: oneDNN just switched to single copyright year. So no need to update it now.

src/graph/backend/dnnl/thread_local_cache.hpp
```diff
@@ -181,14 +181,16 @@ class thread_local_cache_t {
...
-                            return ptr->release();
-                        });
-                return global_cache.get();
-            } catch (...) { return nullptr; }
+            static std::once_flag flag;
+            static global_cache_type_t *global_cache = nullptr;
+
+            std::call_once(flag, []() {
+                try {
+                    global_cache = new global_cache_type_t {};
```
TaoLv: This is suspicious as there is no explicit delete for it. Isn't there a leak at the exit of the program?

ElaineBao: It will be deleted in the release() function.

The original issue is that the shared_ptr with custom deleter is creating a problematic interaction with the manual reference counting during static initialization/destruction. So the solution is to avoid using shared_ptr+custom deleter, and rely on the manual reference counting in retain()/release() to manage its lifetime.

ElaineBao:
make test

src/graph/backend/dnnl/thread_local_cache.hpp
TaoLv: I think this is a good chance to take another look at the cache design here.
IIUC, this `global_cache` is technically a singleton to store all the cached objects of all threads during the lifetime of a program. So do we really need the `retain/release` functions and the complexity here? We just need to: 1. construct it at the first access; 2. use the instance by ptr in every threads; 3. destruct it at the library exit (be careful with the release order.).  Please correct me if I'm wrong. Thanks.

Edit: Or do we want to have an explicit lifetime control for singleton object, not till the end of the program?

ElaineBao: OK, let me take a closer look at the cache design

rongzha1: > It will be deleted in the release() function.
> 
> The original issue is that the shared_ptr with custom deleter is creating a problematic interaction with the manual reference counting during static initialization/destruction. So the solution is to avoid using shared_ptr+custom deleter, and rely on the manual reference counting in retain()/release() to manage its lifetime.

I found primtive engine also use this mechanism, why it can pass sanitizer?

```
engine_t *get_service_engine() {
    static std::unique_ptr<engine_t, engine_deleter_t> cpu_engine;
    static std::once_flag initialized;
    std::call_once(initialized, [&]() {
        engine_t *cpu_engine_ptr;
        cpu::cpu_engine_factory_t f;
        auto status = f.engine_create(&cpu_engine_ptr, 0);
        assert(status == status::success);
        MAYBE_UNUSED(status);
        cpu_engine.reset(cpu_engine_ptr);
    });
    return cpu_engine.get();
}
```

src/graph/backend/dnnl/thread_local_cache.hpp
```diff
@@ -181,14 +181,16 @@ class thread_local_cache_t {
...
-                        [](global_cache_type_t *ptr) {
-                            return ptr->release();
-                        });
-                return global_cache.get();
-            } catch (...) { return nullptr; }
+            static std::once_flag flag;
+            static global_cache_type_t *global_cache = nullptr;
+
+            std::call_once(flag, []() {
+                try {
```
TaoLv: Given the shared_ptr is removed, i think we don't need the try-catch anymore.

TaoLv (approved):

src/graph/backend/dnnl/thread_local_cache.hpp
ElaineBao: the `new` operation may also throw, e.g. bad_alloc, in that case try/cache can help

TaoLv: IIRC, the previous try-catch was added to suppress Coverity warnings about shared_ptr usage. I suppose with changes to plain new operator, Coverity should be good with that. I also checked around, it seems oneDNN does not catch the exception thrown by new.

ElaineBao: gotta, updated

PR #4321, opened 2025-11-13
ci: aarch64: set perf fail-fast to false

michalowski-arm:


Ryo-not-rio:
thanks for this change! working nightly run can be seen in https://github.com/uxlfoundation/oneDNN/actions/runs/19333631473

PR #4324, opened 2025-11-13
xe: jit: refactor to enable upstreaming IR/DSL to gemmstone part 3

rjoursler:


rjoursler:
make test
disable test_device_cpu

src/gpu/intel/jit/ir/include/kernel.hpp
```diff
@@ -84,12 +91,34 @@ class options_t {
         : hw_(hw), regs_(regs), simd_(simd) {}
 
     const hw_t &hw() const { return hw_; }
+
+    // Set number of GRFs. This can be used to avoid context switch overhead on
+    // Xe and Xe2 architectures.
```
echeresh: Should be moved down to `set_regs()`?

src/gpu/intel/jit/ir/include/kernel.hpp
```diff
@@ -84,12 +91,34 @@ class options_t {
...
     int regs() const { return regs_; }
-    int simd() const { return simd_; }
-    int grf_size() const { return hw_.grf_size(); }
     void set_regs(int regs) { regs_ = regs; }
+
+    int simd() const { return simd_; }
     void set_simd(int simd) { simd_ = simd; }
 
+    // Override default dpas kernel annotation. This helps avoid overhead on
+    // XeHPG when switching between kernels with and without dpas support.
```
echeresh, line 105: Ditto, is this comment intended for the setter?

src/gpu/intel/jit/codegen/reorder.cpp
```diff
@@ -632,9 +657,19 @@ bool reorder_impl_t::layouts_compatible(
...
-            layout.with(type), stride * layout.type().size() / type.size());
+    auto blocks = layout.blocks();
+    if (layout.type().size() >= type.size()) {
+        gpu_assert(layout.type().size() % type.size() == 0);
+        for (auto &b : blocks)
+            b.stride *= layout.type().size() / type.size();
+    } else {
+        gpu_assert(type.size() % layout.type().size() == 0);
+        for (auto &b : blocks) {
+            assume(b.stride % (type.size() / layout.type().size()) == 0);
```
echeresh:
```suggestion
            gpu_assert(b.stride % (type.size() / layout.type().size()) == 0);
```

src/gpu/intel/jit/codegen/kernel.hpp
```diff
@@ -46,36 +43,13 @@ namespace gpu {
...
-
 class expr_binding_t {
 public:
     expr_binding_t(ngen::HW hw) : hw_(hw) {}
 
     ~expr_binding_t() {
-        if (!cpp_compat::uncaught_exceptions()) {
-            gpu_assert(expr2dst_.empty()) << "Detected missing unbind_dst().";
-        }
+        gpu_assert(expr2dst_.empty()) << "Detected missing unbind_dst().";
```
echeresh: What is the reason to drop `uncaught_exceptions()` guard?

src/gpu/intel/jit/codegen/kernel_ext.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+extern ngen::NEOInterfaceHandler generate_ngen_interface(
+        const kernel::iface_t &kernel_iface, const kernel::options_t &options,
+        const stmt_t &kernel_body);
+
+template <typename GeneratorT>
+void convert_ir_to_ngen(const stmt_t &body, GeneratorT &host,
+        const walk_order_t *kernel_grid_walk_order = nullptr);
+
+extern template void convert_ir_to_ngen<ngen_code_generator_t<ngen::HW::XeLP>>(
```
echeresh: Nit: better to use `REG_XXX_ISA` to align these declarations with the corresponding definitions.

src/gpu/intel/jit/codegen/extensions.cpp
```diff
@@ -0,0 +1,162 @@
...
+        const object_t &obj, codegen_extension_interface_t &ext_iface) {
+    auto host = ext_iface.root_code_generator();
+
+// NOLINTBEGIN(bugprone-macro-parentheses)
+#define EXT_STMT(ngen_generator_t) \
+    if (host.info == typeid(ngen_generator_t)) \
+        handler(*static_cast<ngen_generator_t *>(host.ptr), obj, ext_iface);
+    // NOLINTEND(bugprone-macro-parentheses)
+
+    EXT_STMTS
```
echeresh: Can we reuse the existing `GPU_HW_SWITCH`? Looks like it can work here (and handle `NGEN_ASM` in a separate `if`).

src/gpu/intel/jit/codegen/extensions.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    gpu_error_not_expected() << "Unknown object " << obj;
+}
+
+void extension_handler(
+        const object_t &obj, codegen_extension_interface_t &ext_iface) {
+    auto host = ext_iface.root_code_generator();
+
+// NOLINTBEGIN(bugprone-macro-parentheses)
+#define EXT_STMT(ngen_generator_t) \
```
echeresh: Nit: missing `undef`.

src/gpu/intel/jit/codegen/codegen_extensions.hpp
```diff
@@ -13,24 +13,35 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 *******************************************************************************/
+#ifndef GPU_INTEL_JIT_CODEGEN_CODEGEN_EXTENSIONS_HPP
```
echeresh: Can we drop the prefix in the file name? or is it because of some name clashing?
`codegen/codegen_extensions.hpp` -> `codegen/extensions.hpp`

src/gpu/intel/jit/codegen/codegen.cpp
```diff
@@ -80,6 +77,20 @@ class ir_to_ngen_t : public ir_visitor_t {
...
+    host_t root_code_generator() const override {
+        return {static_cast<typename ngen_generator_t::RootCodeGenerator *>(
+                        host_),
+                typeid(typename ngen_generator_t::RootCodeGenerator)};
+    }
+    const kernel::options_t &options() const override {
+        return host_->options();
+    }
+    reg_allocator_t &allocator() override { return host_->ra(); }
+    std::vector<ngen_operand_t> evaluate(const std::vector<expr_t> &exprs,
```
echeresh, line 92: Looks like naming is inconsistent with `evaluate`/`eval` - can we align?

src/gpu/intel/jit/codegen/codegen_extensions.hpp
```diff
@@ -13,24 +13,35 @@
...
 namespace dnnl {
 namespace impl {
 namespace gpu {
 namespace intel {
-namespace reorder {
 namespace jit {
 
-using namespace intel::jit;
-void normalize(layout_t &a, layout_t &b);
+struct codegen_extension_interface_t {
```
echeresh: Nit: `interface` -> `iface` could be more aligned with `expr_iface_t`, `kernel::iface_t`.

src/gpu/intel/jit/codegen/kernel.hpp
```diff
@@ -88,13 +62,13 @@ class expr_binding_t {
     }
 
     void bind_dst(const expr_t &expr, const ngen_operand_t &operand) {
-        gpu_assert(expr);
+        gpu_assert(bool(expr));
```
echeresh: Is there some reason to change that? (I'd expect it work fine due to `explicit operator bool()` in `object_t`).

src/gpu/intel/jit/ir/include/kernel.hpp
rjoursler: My main goal here is to just document what each option does and its main purposes, not necessarily the specific setter/getter. I can go ahead an swap the ordering of the setters/getters to keep everything more coherent.

src/gpu/intel/jit/codegen/kernel.hpp
rjoursler: This was unintentional, thanks for noticing it.

src/gpu/intel/jit/codegen/extensions.cpp
rjoursler: No, switch statements require an integer type and to my knowledge, it isn't possible to convert `type_info` to a unique integer which could be used in a switch statement.

src/gpu/intel/jit/codegen/kernel.hpp
rjoursler: This was a consequence of this call to `gpu_assert` being refactored into a function call, where implicit bool conversion is not valid. There is no particular reason this needed refactored though, it is mostly a consequence of the script I have been using to test upstreaming changes. I can go ahead and drop it.

src/gpu/intel/jit/codegen/extensions.cpp
echeresh: I see, it may need a bit more, can we add `hw` to `host_t` for proper dispatching?

Then we could do `GPU_HW_SWITCH(host.hw)` + separate `typeid(ngen::AsmCodeGenerator)` check. Looks like it's worth it to have less platform-dependent places in the library.

rjoursler: Sure, we can do that.

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test perf-gpu

src/gpu/intel/jit/codegen/codegen.cpp
```diff
@@ -1612,6 +1612,8 @@ setup_flags_t get_setup_flags(const stmt_t &s) {
     return visitor.flags;
 }
 
+// This interface is relied on by the oneDNN ir_kernel_t extensions. Do not
+// modify without propagating this change.
```
Simonsays095, line 1616: Is this comment (and the one below) needed? I assume upstreaming to gemmstone would have oneDNN-specific places like this guarded so that they're not included elsewhere - which should make it obvious that we use it in oneDNN. My understanding is that this function is used for legacy support in oneDNN, and should not be used in the DSL path - and should be entirely unnecessary to include for any other users of gemmstone.

src/gpu/intel/jit/codegen/codegen.cpp
```diff
@@ -831,9 +827,8 @@ class ir_to_ngen_t final : public codegen_extension_iface_t,
 template <typename ngen_generator_t>
 class expr_evaluator_t : public ir_visitor_t {
 public:
-    expr_evaluator_t(ngen_generator_t *host, const expr_binding_t &expr_binding,
-            ngen_register_scope_t &scope)
-        : host_(host), expr_binding_(expr_binding), scope_(scope) {}
+    expr_evaluator_t(ngen_generator_t *host, ngen_register_scope_t &scope)
+        : host_(host), expr_binding_(host->expr_binding()), scope_(scope) {}
```
Simonsays095, line 827: Like in the previous commit, does it make sense to use `host_->expr_binding()` directly instead of caching a pointer in the constructor?

third_party/ngen/ngen.hpp
```diff
@@ -96,7 +97,13 @@ class BinaryCodeGenerator
...
     static constexpr HW hardware = hw;
     static constexpr HW getHardware() { return hardware; }
-    void cancelAutoSWSB() { cancelAutoSWSB_ = true; }
+    void enableCancelAutoSWSB() {cancelAutoSWSB_.reset(new std::atomic<bool>(false));}
+    void cancelAutoSWSB() {
+#ifdef NGEN_SAFE
+        if(cancelAutoSWSB_ == nullptr) throw invalid_object_exception();
+#endif
+        if(cancelAutoSWSB_) *cancelAutoSWSB_ = true;
+    }
```
Simonsays095, line 106: Since this changes the default behaviour of `cancelAutoSWSB` from enabled to disabled, this requires changes to my benchstone refactoring PR. Please either add those changes to my existing gemmstone PR, or remind me to make those changes on that PR once this is merged to ngen.

src/gpu/intel/jit/codegen/codegen.cpp
rjoursler: This template functionality is currently not visible through any headers, which is the reason for the comment. Longer term we are going to want to continue refining oneDNN's interaction with the upstream to avoid API breaks, but my current goal is restricted to enabling the DSL. There are going to be a lot of warts left to treat after the initial merge as it is not worth treating them right now.

src/gpu/intel/jit/codegen/codegen.cpp
rjoursler: So this is actually creating a copy and it is needed to be functionally correct. My first implementation did exactly as you suggested and failed functional correctness tests.

PR #4276, opened 2025-11-03
xe: gemm: add 1D block load kernels

Simonsays095:
Fixes [MFDNN-14247](https://jira.devtools.intel.com/browse/MFDNN-14247). Adds some int4/int8 dynamic quantization kernels with 4-byte alignment to cover low-alignment cases performantly.

Example (before and after):
> Output template: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
> perf,gpu,jit:gemm:any,,--mode=P --matmul --engine=gpu --dt=f16:s8:f16 --strides=3420x1:1x3420:1280x1 --bia-dt=f16 --attr-scales=wei:per_oc:f16 --attr-post-ops=add:f16:3:ab --attr-scratchpad=user --attr-fpmath=f16:true 9048x3420:3420x1280,79.217,1165.27,11.8633,6677.47,11.919,6646.28
> perf,gpu,jit:gemm:any,,--mode=P --matmul --engine=gpu --dt=f16:s8:f16 --strides=3420x1:1x3420:1280x1 --bia-dt=f16 --attr-scales=wei:per_oc:f16 --attr-post-ops=add:f16:3:ab --attr-scratchpad=user --attr-fpmath=f16:true 9048x3420:3420x1280,79.217,568.56,0.831145,95310.7,1.06991,74040.9

Edit: Added a fix for some correctness failures in the new kernels, resulting from faulty indexing logic used in dequantization paths.

PR #4328, opened 2025-11-14
xe: conv: jit: add heuristic to handle model tie

rjoursler:
Fixes [MFDNN-14365](https://jira.devtools.intel.com/browse/MFDNN-14365).

### DIFF
--- a/src/gpu/intel/conv/jit/tiler.cpp
+++ b/src/gpu/intel/conv/jit/tiler.cpp
@@ -1218,6 +1218,31 @@ void sort_by_model_scores(params_generator_t &params_gen, const config_t &cfg,
     }
     params_gen.sort(0, params_gen.configs(),
             [&](const blocking_params_t &p) { return -eff_scores.at(p.id()); });
+
+    // Heuristics when model tie is detected
+    auto &params_vec = params_gen.params_vec();
+    auto &p_best = params_vec[0];
+    for (auto &p_next : params_gen.params_vec()) {
+        if (&p_best == &p_next) continue;
+        if (eff_scores.at(p_best.id()) != eff_scores.at(p_next.id())) break;
+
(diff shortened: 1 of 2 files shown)


rjoursler:
make test perf-gpu
set primitive=gpu:conv

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/conv/jit/tiler.cpp
```diff
@@ -1218,6 +1218,31 @@ void sort_by_model_scores(params_generator_t &params_gen, const config_t &cfg,
     }
     params_gen.sort(0, params_gen.configs(),
```
echeresh, line 1219: Nit: for higher robustness we may want to use `std::stable_sort` underneath to ensure `params_vec_` order is consistent across all platforms/runtimes.

Now sorting is done via `std::sort` which doesn't specify the order of equal elements in the resulting sequence. Because of that we may have different configs selected on Linux/Windows (due to different `std::sort` behavior).

Just thought it's worth mentioning since you've been looking into that.

echeresh (approved):

src/gpu/intel/conv/jit/tiler.cpp
rjoursler: This should already be handled by [#4276](https://github.com/uxlfoundation/oneDNN/pull/4267) by the check against ID [here]( https://github.com/uxlfoundation/oneDNN/blob/4ad8a7d8e1d725993bfb9775c14b4cd8e8b9274b/src/gpu/intel/jit/ir/blocking.hpp#L580) (note: `std::stable_sort` is not available in C++11, so I couldn't directly use it).

kealan-barbieri (approved):

PR #4338, opened 2025-11-18
cpu: pool: enable different dt on src and dst in ref

asztukow:
Addresses MFDNN-7685.

This PR adds support for different data types on source and destination in ref pooling. 
Support in JIT implementation is WIP and will be provided in separate PR.

### DIFF
--- a/src/cpu/ref_pooling.cpp
+++ b/src/cpu/ref_pooling.cpp
@@ -43,13 +43,11 @@ static inline dim_t get_offset(const memory_desc_wrapper &mdw, dim_t n, dim_t c,
 
 using namespace nstl;
 
-template <data_type_t data_type, data_type_t acc_type>
-status_t ref_pooling_fwd_t<data_type, acc_type>::execute_forward(
-        const exec_ctx_t &ctx) const {
+status_t ref_pooling_fwd_t::execute_forward(const exec_ctx_t &ctx) const {
 
     status_t status = status::success;
-    auto src = CTX_IN_MEM(const data_t *, DNNL_ARG_SRC);
-    auto dst = CTX_OUT_CLEAN_MEM(data_t *, DNNL_ARG_DST, status);
+    auto src = CTX_IN_MEM(const void *, DNNL_ARG_SRC);
+    auto dst = CTX_OUT_CLEAN_MEM(void *, DNNL_ARG_DST, status);
     CHECK(status);
     auto ws = CTX_OUT_CLEAN_MEM(unsigned char *, DNNL_ARG_WORKSPACE, status);
     CHECK(status);
@@ -111,7 +109,7 @@ status_t ref_pooling_fwd_t<data_type, acc_type>::execute_forward(
                     if (iw < 0 || iw >= IW) continue;
 
                     const auto off = get_offset(src_d, mb, oc, id, ih, iw);
-                    auto s = src[off];
+                    float s = io::load_float_value(src_d.data_type(), src, off);
                     if (s > d) {
                         d = s;
                         set_ws(mb, oc, od, oh, ow, (kd * KH + kh) * KW + kw);
@@ -134,7 +132,7 @@ status_t ref_pooling_fwd_t<data_type, acc_type>::execute_forward(
                     if (iw < 0 || iw >= IW) continue;
 
                     const auto off = get_offset(src_d, mb, oc, id, ih, iw);
-                    d += src[off];
+                    d += io::load_float_value(src_d.data_type(), src, off);
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


src/common/type_helpers.hpp
```diff
@@ -457,7 +457,9 @@ inline data_type_t default_accum_data_type(
     // we allow to use f32 accumulation type only when the
     // accumulation chain is small. Otherwise, strict should be set to
     // true
-    if (one_of(src_dt, s8, u8, u4, s4) && (dst_dt != f32 || strict)) return s32;
+    if (one_of(src_dt, s8, u8, u4, s4)
+            && ((dst_dt != f32 && dst_dt != f16) || strict))
+        return s32;
```
dzarukin: I'd expect `strict` should cover all your needs. When do you see it's not happening?

src/cpu/ref_pooling.cpp
```diff
@@ -171,8 +169,26 @@ status_t ref_pooling_fwd_t<data_type, acc_type>::execute_forward(
...
+                return nstl::numeric_limits<bfloat16_t>::lowest();
+            case data_type::f8_e5m2:
+                return nstl::numeric_limits<float8_e5m2_t>::lowest();
+            case data_type::f8_e4m3:
+                return nstl::numeric_limits<float8_e4m3_t>::lowest();
+            default: return 0.f;
+        }
+    };
+
+    float base_res = is_max_pool ? get_lowest_value(dst_d.data_type()) : 0.f;
```
dzarukin: There's `inline T lowest_value(data_type_t data_type)` interface to do exactly same thing.

src/cpu/ref_pooling.hpp
```diff
@@ -42,17 +41,38 @@ struct ref_pooling_fwd_t : public primitive_t {
...
             VDISPATCH_POOLING(is_fwd(), VERBOSE_BAD_PROPKIND);
-            VDISPATCH_POOLING(utils::everyone_is(data_type, src_md()->data_type,
-                                      dst_md()->data_type),
-                    VERBOSE_UNSUPPORTED_DT);
-            VDISPATCH_POOLING(desc()->accum_data_type == acc_type,
-                    VERBOSE_UNSUPPORTED_DT);
+            VDISPATCH_POOLING(
+                    IMPLICATION(src_md()->data_type != dst_md()->data_type,
+                            desc()->prop_kind == prop_kind::forward_inference),
+                    VERBOSE_BAD_PROPKIND);
```
dzarukin: I'm not sure about this condition... In theory, one could train with src = bf16 but ask for dst = f32 for the next layer as it might be sensitive for precision. I don't see why it can't be supported regardless of the prop kind.

In this sense, backward should get a change to support different data types as well.

src/cpu/ref_pooling.hpp
```diff
@@ -42,17 +41,38 @@ struct ref_pooling_fwd_t : public primitive_t {
...
+                                    dst_md()->data_type, s8, u8, f16, f32)),
+                    VERBOSE_INCONSISTENT_DT, "src_data_t", "dst_data_t");
+            VDISPATCH_POOLING(
+                    IMPLICATION(src_md()->data_type == f16,
+                            utils::one_of(dst_md()->data_type, s8, u8, f16)),
+                    VERBOSE_INCONSISTENT_DT, "src_data_t", "dst_data_t");
+            VDISPATCH_POOLING(
+                    IMPLICATION(src_md()->data_type == f32,
+                            utils::one_of(dst_md()->data_type, s8, u8, f32)),
+                    VERBOSE_INCONSISTENT_DT, "src_data_t", "dst_data_t");
```
dzarukin: All these 4 conditions seems unnecessary. The code should handle pretty much any data type combination.

src/common/type_helpers.hpp
asztukow: According to the [table](https://github.com/uxlfoundation/oneDNN/blob/main/doc/primitives/pooling.md#data-type-support), accumulation type for cases like int8:f16 should be f32, not s32. Without this change, returned type is s32.

dzarukin: There's additional `|| strict` condition which must trigger for pooling disrespectful of dst data type value.

I'm not even sure what dst_dt does here as it's always possible to convert to any data type as long it's supported on a system...

src/cpu/ref_pooling.hpp
asztukow: Removed

src/cpu/ref_pooling.hpp
asztukow: Removed

src/cpu/ref_pooling.cpp
```diff
@@ -171,8 +169,10 @@ status_t ref_pooling_fwd_t<data_type, acc_type>::execute_forward(
 
     const bool is_max_pool = alg == alg_kind::pooling_max;
 
-    float base_res
-            = is_max_pool ? (float)numeric_limits<data_t>::lowest() : 0.f;
+    float base_res = is_max_pool
+            ? dnnl::impl::types::lowest_value<float>(dst_d.data_type())
```
dzarukin:
```suggestion
            ? types::lowest_value<float>(dst_d.data_type())
```

dzarukin (approved):

asztukow:
make test
set test_scope=NIGHTLY

Issue #4313, opened 2025-11-11
PyTorch test failures for s8/u8->f32 brgemm convolutions on SVE256


Ryo-not-rio:
# Summary
The following tests:
```
test/inductor/test_cpu_cpp_wrapper.py TestCppWrapper.test_qconv2d_add_cpu_cpp_wrapper
test/inductor/test_cpu_cpp_wrapper.py TestCppWrapper.test_qconv2d_cpu_cpp_wrapper
test/inductor/test_cpu_cpp_wrapper.py TestCppWrapper.test_qconv2d_add_relu_cpu_cpp_wrapper
```
are failing when upgrading oneDNN to 3.10 in PyTorch

We have narrowed down the cause to https://github.com/uxlfoundation/oneDNN/pull/3979 and have raised a PR to mitigate this: https://github.com/uxlfoundation/oneDNN/pull/4312. However, we need a deeper look into the cause and a proper fix.

cc: @kasturedeeksha 

# Version
3.10

# Steps to reproduce
Upgrade PyTorch oneDNN to 3.10, then run the listed tests

# Observed behavior
Test fails

# Expected behavior
Test passes.

kasturedeeksha:
I checked this issue in detail and here are the insights 
All the failing tests hit the same runtime error:

`RuntimeError: could not create a primitive descriptor for the reorder primitive.`

On running `test test/inductor/test_cpu_cpp_wrapper.py TestCppWrapper.test_qconv2d_add_cpu_cpp_wrapper` with `ONEDNN_VERBOSE=all`
I see two convolution shapes for the u8/s8 → s32 configuration.

1. **mb1_ic3oc6_ih8oh6kh3sh1dh0ph0_iw8ow6kw3sw1dw0pw0** 
    This is getting executed using brgconv:sve_256 as seen in logs
 
```
onednn_verbose,v1,primitive,create:cache_miss,cpu,convolution,brgconv:sve_256,forward_inference,src:u8::blocked:acdb::f0 wei:s8:ap:blocked:AcdB8a4b::f8:zpm1 bia:undef::undef::: dst:s32::blocked:acdb::f0,attr-scratchpad:user attr-scales:src0:0:f32 attr-zero-points:src0:0:s32,alg:convolution_direct,mb1_ic3oc6_ih8oh6kh3sh1dh0ph0_iw8ow6kw3sw1dw0pw0,0.968018
```

[...]

PR #4340, opened 2025-11-18
cpu: aarch64: fix reorder

aditew01:
Fixes: #4313


aditew01:
Not intended to create huge changes. will repoen,

PR #4335, opened 2025-11-17
xe3: jit: gemm: workaround for HW DPAS read suppression bug

petercad:
Closes MFDNN-14292. Intermittent failures were seen for some Xe3 matmul layers, e.g.

```
--dt=bf16:bf16:f32 --stag=abc --wtag=acb --dtag=abc 1024x1x144:1024x144x1
```

This was root-caused to a complex HW bug in the DPAS read suppression logic when physical register allocations from different threads alias. The workaround here is to avoid dpas with uninitialized src1/src2 -- even though the kernel does not use the results of such dpas instructions.

PR #4343, opened 2025-11-18
xe3: jit: gemm: workaround for HW DPAS read suppression bug

petercad:
Backport of #4335 to `rls-v3.10`.


petercad:
make test
disable test_device_cpu
disable test_device_gpu

PR #4354, opened 2025-11-19
gpu: intel: ocl: allow unlimited allocations

spalicki:
Allow allocation of buffers bigger than `CL_DEVICE_MAX_MEM_ALLOC_SIZE` limited by `CL_DEVICE_GLOBAL_MEM_SIZE`.

### DIFF
--- a/src/gpu/intel/compute/dispatch_reusable.hpp
+++ b/src/gpu/intel/compute/dispatch_reusable.hpp
@@ -245,8 +245,7 @@ struct dispatch_compile_params_t {
     subgroup_data_t subgroup;
     int32_t num_terms = 0;
     bool use_int32_offset = false;
-    bool require_large_buffers = true;
-    uint8_t padding[2] = {0};
+    uint8_t padding[3] = {0};
     gws_indexing_term_t::compile_params_t terms[MAX_INDEXING_TERMS]
             = {{gws_op_t::SOLO, 0}};
 
@@ -532,7 +531,6 @@ public:
 
         // Save buffer information
         dim_t max_buffer_size = 0;
-        uint64_t max_buffer_size_bytes = 0;
         compile_params.num_buffers = buffers.size();
         for (size_t buf_idx = 0; buf_idx < buffers.size(); buf_idx++) {
             const named_buffer_t &buffer = buffers[buf_idx];
@@ -553,15 +551,10 @@ public:
             compile_params.buffer_types[buf_idx] = buffer.data_type;
 
             // Check buffer sizes to see if we can use int32_t offsets
-            // or do we need to use stateless addressing model
             max_buffer_size = std::max(max_buffer_size, buffer.nelems(true));
-            max_buffer_size_bytes = std::max(
-                    max_buffer_size_bytes, buffer.size(0, true, true));
         }
 
         compile_params.use_int32_offset = max_buffer_size <= INT32_MAX;
-        compile_params.require_large_buffers
-                = max_buffer_size_bytes > UINT32_MAX;
         compile_params.subgroup = subgroup;
 
         // Set runtime params
--- a/src/gpu/intel/concat/xe.cpp
+++ b/src/gpu/intel/concat/xe.cpp
@@ -122,12 +122,12 @@ status_t xe_t::pd_t::init_conf(impl::engine_t *engine) {
     conf.n = pd->n_inputs();
     conf.concat_axis = pd->concat_dim();
 
-    dim_t max_bytes = dst_mdw.size();
+    dim_t max_elems = dst_mdw.nelems();
     int concat_axis_end = 0;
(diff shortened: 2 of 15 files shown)


spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

src/gpu/intel/ocl/usm_utils.cpp
```diff
@@ -57,11 +57,17 @@ void *malloc_host(impl::engine_t *engine, size_t size) {
             size_t, cl_uint, cl_int *);
 
     if (size == 0) return nullptr;
+    bool large_buffer = size
+            > utils::downcast<const xpu::ocl::engine_impl_t *>(engine->impl())
+                      ->max_allocation_size();
+    static cl_bitfield properties[]
```
dzarukin: Minor: I'm not sure `static` is good for `properties` object that's passed to CL calls especially in case it will need to be extended in the future. Also current code doesn't provide a scalable way of expanding it either, probably a TODO with that note could help to look into that direction when/if needed.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1028,9 +1028,9 @@ size_t get_cpu_ram_size() {
 }
 #endif
 
-int get_gpu_ram_sizes(size_t &ram_size, size_t &max_alloc_size) {
+int get_gpu_ram_size(size_t &ram_size) {
```
dzarukin, line 1031: Did you have a chance to verify the change works with all 4 supported memory kinds times correctness and fast performance mode where different approach used for memory object management?

As a part of this question also: should [this call](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/dnnl_memory.cpp#L543) be updated?

src/gpu/intel/compute/dispatch_reusable.hpp
```diff
@@ -553,10 +554,10 @@ class reusable_dispatch_t {
             compile_params.buffer_types[buf_idx] = buffer.data_type;
 
             // Check buffer sizes to see if we can use int32_t offsets
-            // or do we need to use stateless addressing model
+            // and do we need to use stateless addressing model
             max_buffer_size = std::max(max_buffer_size, buffer.nelems(true));
-            max_buffer_size_bytes = std::max(
-                    max_buffer_size_bytes, buffer.size(0, true, true));
+            max_buffer_size_bytes = std::max(max_buffer_size_bytes,
+                    buffer.size(0, true, true) * num_work_items);
```
atkassen: What is the issue that this change is addressing?

tests/benchdnn/dnnl_common.cpp
spalicki: ~~Well, it seems that `CL_MEM_ALLOW_UNRESTRICTED_SIZE_INTEL` is truly unrestricted, allowing to allocate any size even over the GPU VRAM. So, I am going to have to add some additional guard.~~
NVM it looks like driver bug. 

The flag according to documentation only applies to `clCreateBuffer`, `clCreateBufferWithProperties`, `clCreateBufferWithPropertiesINTEL`, `clSVMAlloc`, `clSharedMemAllocINTEL`, `clDeviceMemAllocINTEL` and `clHostMemAllocINTEL`.

src/gpu/intel/compute/dispatch_reusable.hpp
spalicki: I will extract this to another PR, since there seems to be some driver issues when running this change. The problem this like is addressing is that when we are compiling kernels in stateless addressing model, we need to take entire buffer into consideration (>4GB) not just a work item.

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

atkassen:
make test perf-gpu

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

src/gpu/intel/concat/xe.cpp
```diff
@@ -122,12 +122,12 @@ status_t xe_t::pd_t::init_conf(impl::engine_t *engine) {
     conf.n = pd->n_inputs();
     conf.concat_axis = pd->concat_dim();
 
-    dim_t max_bytes = dst_mdw.size();
+    dim_t max_bytes = dst_mdw.nelems();
```
atkassen: nit: rename to something like `max_elems`

spalicki:
make test perf-gpu

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
set test_scope=NIGHTLY

PR #4368, opened 2025-11-21
xe: gemm: reduce evaluate calls

Simonsays095:
Reduces the number of `evaluate` calls during GEMM primitive creation, by calling it once and storing the score with the entry, to be used later in sorting. Also moves a final evaluate call out one layer, so it's not called between HW fallback iterations.

Finally, reintroduces observer functionality, to log the score for each valid kernel when `ONEDNN_VERBOSE=debuginfo=5` or greater.

### DIFF
--- a/src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
+++ b/src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
@@ -158,38 +158,51 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
     return (alignA1 <= alignA2) && (alignB1 <= alignB2) && (alignA1 + alignB1 < alignB1 + alignB2);
 }
 
+struct EntryData {
+    EntryData(const kcatalog::Entry *entry_, double score_) : entry(entry_), score(score_) {}
+    const kcatalog::Entry *entry;
+    double score;
+};
+
 // Inner kernel selection logic.
 // Choose the best entry, if any, matching one of the given patterns.
 const std::vector<const kcatalog::Entry *> getEntries(const kcatalog::Catalog &catalog, int npatterns, const MatchParams *patterns, const EvaluateParams &eparams, EvaluateAuxOutput &aux,  SelectionObserver * observer)
 {
-    std::vector<const kcatalog::Entry *> entries;
     // TODO: omit evaluation if only one match, if aux output not needed.
+    std::vector<EntryData> keys;
     for (int ipattern = 0; ipattern < npatterns; ipattern++) {
         for (auto it = match(catalog, patterns[ipattern]); it; it++) {
-             // Late tag checking. If late tags do not match, we skip entry.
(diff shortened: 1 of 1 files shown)


src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -281,6 +293,9 @@ const std::vector<const kcatalog::Entry *> select(const kcatalog::Catalog &catal
         }
     } while (hw);
 
+    if (entries.size() > 0)
+	    evaluate(*entries[0], eparams, aux);
```
kealan-barbieri, line 298: can we drop this if we're not catching the score?

Simonsays095: This call isn't for the score, it's to set the auxiliary parameters. I agree this is an odd place to do it - I looked around a bit and it seems like this is handled inconsistently (we should call evaluate before every finalize call, or we should save and return a separate set of aux params with each evaluate call). I'll make this better.

Simonsays095: Actually, changing this behaviour may result in performance and/or correctness changes, so I'll do it in another PR. The current changes just reduce the number of identical evaluate calls with no functional changes.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -161,17 +162,27 @@ bool lessAligned(int alignA1, int alignB1, int alignA2, int alignB2)
...
         }
     }
-    auto less = [&](const kcatalog::Entry * lhs, const kcatalog::Entry * rhs){
-                      EvaluateAuxOutput thisAux;
+
+    auto less = [&](const sort_key &lhs_key, const sort_key &rhs_key){
+                      auto lhs = lhs_key.first;
+                      auto rhs = rhs_key.first;
+                      auto lhs_score = lhs_key.second;
+                      auto rhs_score = rhs_key.second;
```
rjoursler: nit: Rather than extracting all these variables, we may as well replace `sort_key` with a struct to improve readability:
```
struct sort_key {
    const kCatalog::entry * entry;
    double score;
}
```

rjoursler (approved):

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
Simonsays095: Agreed. And in the next PR, we can add the aux params to the struct and return a vector of these instead of a vector of entries. That eliminates the repacking at the end of the function too.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
enable arch_gpu_xe2-lpg
disable benchdnn_all
enable benchdnn_ip
enable benchdnn_matmul
enable benchdnn_rnn

PR #4376, opened 2025-11-24
cpu: aarch64: add ASIMD exp JIT implementation

Anndrey24:
# Description

This commit introduces an ASIMD `exp` eltwise JIT implementation that supports the f32 data type, based on the `expf`/`expf_1u` implementations from Arm Optimized Routines (AOR).

In addition to computing the result using a range-reduced, scaled polynomial approximation (`exp(x) = 2^n·p(r)`), the algorithm also provides a split-scale special case to prevent overflow/underflow in intermediate products.

Placing the special-case handling logic at the end of the kernel and branching over it for the fast case slightly outperforms placing it at the start of the kernel and always branching over it. Both versions are faster than computing the special-case results every single iteration (i.e. no branching).

Also, ASIMD load replicate calls for table constants are replaced with simpler vector register load instructions whenever the values have already been broadcast to the vector length before being written to memory.

### F32
Command: `./benchdnn --mode=p --eltwise --alg=exp 1024x4096`
 
| Threads | ref:any (ms) | jit:asimd (ms) | Speedup |
|---------|--------------|----------------|---------|
| 1       | 98.5983      |  8.0453        | 12.26x  |
| 4       | 23.9298      |  2.0207          | 11.84x  |
| 16      |  5.9765      |  0.5104        | 11.71x  |
| 64      |  1.5084      |  0.1404        |  10.75x  |

### F16
Command: `./benchdnn --mode=p --eltwise --alg=exp --dt=f16 1024x4096`
 
| Threads | ref:any (ms) | jit:asimd (ms) | Speedup |
|---------|--------------|----------------|---------|
| 1       | 124.423      |  6.81038        | 18.27x  |
| 4       | 31.2355      |  1.70964        | 18.27x  |
| 16      |  7.82124      |  0.431423        | 18.13x  |
| 64      |  1.98035      |  0.117871        |  16.80x  |

cc @fadara01

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2019 Intel Corporation
 * Copyright 2021-2024 FUJITSU LIMITED
-* Copyright 2022, 2025 Arm Ltd. and affiliates
+* Copyright 2019-2025 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -56,9 +56,10 @@ bool is_supported(cpu_isa_t isa, alg_kind_t alg) {
         if (isa == asimd) {
             using namespace alg_kind;
             return utils::one_of(alg, eltwise_relu, eltwise_square, eltwise_abs,
-                    eltwise_sqrt, eltwise_linear, eltwise_hardsigmoid,
-                    eltwise_hardswish, eltwise_clip, eltwise_clip_v2,
-                    eltwise_round, eltwise_clip_v2_use_dst_for_bwd);
+                    eltwise_sqrt, eltwise_linear, eltwise_exp,
+                    eltwise_hardsigmoid, eltwise_hardswish, eltwise_clip,
+                    eltwise_clip_v2, eltwise_round,
+                    eltwise_clip_v2_use_dst_for_bwd);
         } else {
             return is_alg_supported(alg);
         }
@@ -1590,7 +1591,7 @@ size_t jit_uni_eltwise_injector_t<isa>::aux_vecs_count() {
             case eltwise_logistic_use_dst_for_bwd:
             case eltwise_logistic: return 5; /* = exp + 1 */
             case eltwise_exp_use_dst_for_bwd:
-            case eltwise_exp: return 4;
+            case eltwise_exp: return (isa == asimd) ? 5 : 4;
             case eltwise_gelu_tanh: return 9; /* = tanh */
             case eltwise_swish: return 6; /* = logistic */
             case eltwise_log: return 6;
@@ -1855,13 +1856,26 @@ void jit_uni_eltwise_injector_t<isa>::register_table_entries() {
             {exp_pol, {0x3efffee3, true}}, // p2 = 0.499991506f
             {exp_pol, {0x3e2aad40, true}}, // p3 = 0.166676521f
             {exp_pol, {0x3d2b9d0d, true}}, // p4 = 0.0418978221f
-            {exp_pol, {0x3c07cfce, true}} // p5 = 0.00828929059f
+            {exp_pol, {0x3c07cfce, true}}, // p5 = 0.00828929059f
+            {exp_pol_asimd, {0x3c072010, true}}, // c0 =  0x1.0e4020p-7
+            {exp_pol_asimd, {0x3d2b9f17, true}}, // c1 =  0x1.573e2ep-5
+            {exp_pol_asimd, {0x3e2aaf33, true}}, // c2 =  0x1.555e66p-3
+            {exp_pol_asimd, {0x3efffedb, true}}, // c3 =  0x1.fffdb6p-2
+            {exp_pol_asimd, {0x3f7ffff6, true}}, // c4 =  0x1.ffffecp-1
     };
     // exp(x) constants2
     static const table_t exp_consts2 {
             {exp_coeff1, {0x3f31721c, true}},
             {exp_coeff2, {0x3e772df2, true}},
             {exp_not_mask17, {~((1u << 17) - 1), true}},
+            {exp_exponent_bias,
+                    {0x3f800000, true}}, // 2^7 - 1 shifted to exponent position
+            {exp_neg_ln2_hi, {0xbf317200, true}}, // -0x1.62e400p-1f (hi)
+            {exp_neg_ln2_lo, {0xb5bfbe8e, true}}, // -0x1.7f7d1cp-20f (lo)
+            {exp_special_bound, {0x42fc0000, true}}, // 126.0f
+            {exp_scale_thresh, {0x43400000, true}}, // 192.0f
+            {exp_special_offset, {0x82000000, true}},
+            {exp_special_bias, {0x7f000000, true}},
     };
 
     // mish(x) constants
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2475,6 +2488,137 @@ size_t jit_uni_eltwise_injector_t<sve_128>::get_vec_len() {
...
+    const auto &t0 = VReg4S(vmm_src.getIdx());
+    const auto &t1 = VReg4S(vmm_aux0.getIdx());
+    const auto &t2 = VReg4S(vmm_aux1.getIdx());
+    const auto &t3 = VReg4S(vmm_aux2.getIdx());
+    const auto &t4 = VReg4S(vmm_aux3.getIdx());
+    const auto &t5 = VReg4S(vmm_aux4.getIdx());
+    const auto &t_tmp = VReg4S(vmm_tmp.getIdx());
+
+    // z = x * inv_ln2
+    const auto &vInvLn2 = table_val(exp_log2ef, t_tmp);
```
jondea: I think we should probably fit the variable names to the oneDNN coding conventions. This should be automatable.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
```diff
@@ -280,8 +281,16 @@ struct jit_uni_eltwise_injector_t {
         exp_ln_flt_max_f, // logf(FLT_MAX) - max normal value
         exp_ln_flt_min_f, // logf(FLT_MIN) - min normal value
         exp_pol, // see correspondent table for float values
+        exp_pol_asimd, // see correspondent table for float values
```
Sqvid: nit: typo: `s/correspondent/corresponding/`

fadara01:
>It's odd that fp16 is faster, naively I would expect it should be slower because it is the same but with extra unpacking/packing.

I find this odd too

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
```diff
@@ -280,8 +281,16 @@ struct jit_uni_eltwise_injector_t {
...
         exp_coeff1, // 0.6931473921 (0x3f31721c)
         exp_coeff2, // 0.2413862043 (0x3e772df2)
+        exp_exponent_bias, // 1.0f (0x3f800000)
+        exp_neg_ln2_hi, // -0x1.62e400p-1f (0xbf317200)
+        exp_neg_ln2_lo, // -0x1.7f7d1cp-20f (0xb5bfbe8e)
+        exp_special_bound, // 126.0f
+        exp_scale_thresh, // 192.0f
+        exp_special_offset, // 0x83000000
+        exp_special_bias, // 0x7f000000
```
Sqvid: I know these comments are pre-existing but I am not convinced by the value of just duplicating the comments in the table (not to mention these could go out of sync). I think better would be to explain what these mean where that's useful.

jondea:
Whoops, sorry, misclick!

Anndrey24:
> It's odd that fp16 is faster, naively I would expect it should be slower because it is the same but with extra unpacking/packing.

I believe f16 (and bf16) are faster because the f32 case has more input vectors taking the special-case path. In fact, f16/bf16 inputs populated by `--mode=p` might not fall outside of the fast-path-supported range (`|x| < 126.5 / ln 2 ≈ 87.6831f`) at all: https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/utils/fill.cpp#L110-L112

After commenting out the special-case logic, I get the following results for `shape=1024x4096` on a c6g, which are more in line with what we'd expect:
| Threads | f32 (ms) | f16 (ms) | bf16 (ms) |
|---------|----------|----------|-----------|
| 1       | 6.92     | 7.36     | 7.08      |

Likewise for removing the fast path, leaving only the special case path:
| Threads | f32 (ms) | f16 (ms) | bf16 (ms) |
|---------|----------|----------|-----------|
| 1       | 9.35     | 10.20    | 9.66      |


> expf_1u might be too accurate, have you looked into something accurate within 2 ULPs?

No, I haven't. I'll try implementing it and see if the correctness tests still pass!


> Also, do you know roughly how fast it is relative to SVE?

For threads = 1, shape = 1024x4096:
### c7g
| Data type  | jit:sve (ms) | jit:asimd (ms) | Speedup (SVE vs ASIMD)|
|---------|----------|----------|-----------|
| f32       | 3.874     | 6.554    | 1.69x      |
| bf16     | 4.244     | 4.99    | 1.18x      |
| f16       | 5.145     | 5.208    | 1.01x      |

### c8g
| Data type | jit:sve (ms) | jit:asimd (ms) | Speedup (SVE vs ASIMD)|
|---------|----------|----------|-----------|
| f32       | 3.099     | 5.223    | 1.69x      |
| bf16       | 7.546     | 3.814    | 0.51x      |
| f16       | 9.638     | 4.069    | 0.42x      |

Two observations:
 1. `jit:asimd` is slower for f32 because of taking the special-case path so often compared to bf16 and f16.
 2. `jit:sve` becomes surprisingly slow for bf16 and f16 on a c8g. I'm not really sure why that is, but this behaviour can be seen with other eltwise operations too (e.g. hardswish, leaky relu).

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Oops, I've renamed the variables according to the convention now.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
Anndrey24: Agreed! I've now changed the comments for entries introduced by this PR, while leaving the pre-existing ones untouched. Would you prefer I take a look at the rest as well?

Anndrey24:
The accuracy of AOR's `expf` seems to be good enough for oneDNN (it passes the benchdnn correctness tests), while being around 15% faster than the previous `expf_1u`-based implementation. 
I've updated the PR description with the new results. 

Thanks @fadara01 !

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp
Sqvid: No this is great thanks

fadara01:
>The accuracy of AOR's expf seems to be good enough for oneDNN (it passes the benchdnn correctness tests), while being around 15% faster than the previous expf_1u-based implementation.
I've updated the PR description with the new results.
Thanks @fadara01 !

Great to hear that! Awesome work :)

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2475,6 +2489,144 @@ size_t jit_uni_eltwise_injector_t<sve_128>::get_vec_len() {
     return get_sve_length();
 }
 
+template <>
+void jit_uni_eltwise_injector_t<asimd>::exp_compute_vector_fwd(
+        const TRegS &vmm_src) {
+    /*
+    * Based on the expf/expf_1u implementations from Arm Optimized Routines (AOR)
```
fadara01: should we remove this `expf_1u` from the comment given the new algorithm is not accurate within 1 ULP?
... and mention the new accuracy ~ 2 ULPs

fadara01:
>jit:sve becomes surprisingly slow for bf16 and f16 on a c8g. I'm not really sure why that is, but this behaviour can be seen with other eltwise operations too (e.g. hardswish, leaky relu).

Outside the scope of this PR, but it'd be nice to get a better understanding as to why this is the case

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I only left that comment in because the final part of the special case logic more closely resembles `expf_1u` than `expf`. Since the former was faster, while not impacting the result of the correctness tests, I just kept that implementation:

- `expf_1u`: computes 2 values (r0 and r1) which it blends into the output vector through a mask, without also computing the fast path result ([code](https://github.com/ARM-software/optimized-routines/blob/master/math/aarch64/advsimd/expf_1u.c#L46-L47))
- `expf`: computes the same 2 values (r1 and r2) + the fast path result (r0), blending the 3 into the output vector using 2 masks ([code](https://github.com/ARM-software/optimized-routines/blob/master/math/aarch64/advsimd/expf.c#L50-L51)) to "avoid double rounding in the subnormal range."

I see how it might be a little misleading, however, so I would be down to remove it from the comment if you think that's clearer!

Anndrey24: Oh, I've just now noticed your edit. I'll add the new accuracy as part of the comment.

PR #4377, opened 2025-11-24
cpu: aarch64: conv: optimize brgemm

Ryo-not-rio:
~5% speedup for bf16 SVE256
~11\~13% speedup for bf16 SVE128

e.g.
```
--mode=P --conv --dt=bf16:bf16:bf16 --impl=brg g1mb1ic32iw84350oc32ow84349kw2pw0n"wavenet:conv1d-1"
```

### DIFF
--- a/src/cpu/aarch64/brgemm/brgemm_utils.cpp
+++ b/src/cpu/aarch64/brgemm/brgemm_utils.cpp
@@ -204,7 +204,7 @@ status_t brgemm_blocking(brgemm_t *brg) {
     // reduce 'ld_block2' to allow a larger 'bd_block'
     const int max_vpad = nstl::max(
             brg->brgattr.max_top_vpad, brg->brgattr.max_bottom_vpad);
-    if (is_superset(brg->isa_impl, sve_256) && max_bcast_block < max_vpad) {
+    if (max_bcast_block < max_vpad) {
         adj_ld_block2 = calculate_ldb_params(brg, 2);
         max_bcast_block = calculate_max_bcast_block(brg, adj_ld_block2);
     }
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -222,6 +222,19 @@ private:
     PReg ld_full_mask = PReg(2);
     PReg ld_tail_mask = PReg(3);
 
+    void add_vl_or_imm(XReg dst, XReg src, int offset) {
+        // If offset is a multiple of the vector length and
+        // offset / vector_length is compatible with addvl
+        // use the addvl instruction. Refer to https://developer.arm.com/documentation/ddi0596/2021-03/SVE-Instructions/ADDVL--Add-multiple-of-vector-register-size-to-scalar-register-
+        // for details.
+        if ((offset % cpu_sveLen == 0)
+                && (offset / static_cast<int>(cpu_sveLen) >= -32)
+                && (offset / cpu_sveLen <= 31))
+            addvl(dst, src, offset / cpu_sveLen);
+        else
+            add_imm(dst, src, offset, X_TMP_0);
+    }
+
     ZReg accm(int ld_block, int bd, int ld) const {
         return ZReg(max_effective_vregs - 1 - (bd * ld_block + ld));
     }
@@ -264,10 +277,6 @@ private:
             bool mask_flag, bool store, PReg ktail_mask, const int offset,
             const int base_offset); //for only memory operand
 
-    void advance_ldb_post_op_regs();
-    void restore_ldb_post_op_regs(int ld_block2);
-    void advance_bdb_post_op_regs(int adj_bd_block);
-    void restore_bdb_post_op_regs(int bd_block2);
     void ldb_regs_shift(int ld_block2, bool is_tail = false);
     void advance_bd_block2_post_op_regs(int bd_block2);
 
@@ -490,8 +499,7 @@ void jit_brgemm_kernel_t::cvt2ps(data_type_t type_in, const ZReg zmm_in,
             break;
         }
         case data_type::bf16: {
-            add_imm(X_TMP_1, addr, offset - base_offset, X_TMP_0);
-            ld1h(z_tmp_1().s, mask, ptr(X_TMP_1));
+            LD_MUL_VL(ld1h, z_tmp_1().s, mask, addr, offset - base_offset, 2);
             lsl(z_tmp_1().s, z_tmp_1().s, 16);
             if (store) //Merging
                 mov(zmm_in.s, ktail_mask / T_m, z_tmp_1().s);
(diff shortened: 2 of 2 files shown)


src/cpu/aarch64/brgemm/brgemm_utils.cpp
```diff
@@ -204,7 +204,7 @@ status_t brgemm_blocking(brgemm_t *brg) {
     // reduce 'ld_block2' to allow a larger 'bd_block'
     const int max_vpad = nstl::max(
             brg->brgattr.max_top_vpad, brg->brgattr.max_bottom_vpad);
-    if (is_superset(brg->isa_impl, sve_256) && max_bcast_block < max_vpad) {
```
renato-arantes, line 207: Why is the `is_superset` no longer needed?

Ryo-not-rio: because we should be doing this for all supported isas i.e. sve128, 256, 512

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1881,10 +1811,14 @@ void jit_brgemm_kernel_t::bdb_loop() {
                                 bd_blocks_for_rd_tail <= 1 ? 0
                                                            : rows_for_rd_tail,
                                 skip_accumulation);
-                        sub(reg_bdb_loop, reg_bdb_loop, 1);
-                        cmp_imm(reg_bdb_loop, rows_for_rd_tail ? 1 : 0,
-                                X_TMP_0);
-                        b(GT, bdb_loop_label);
+                        if (rows_for_rd_tail) {
+                            sub(reg_bdb_loop, reg_bdb_loop, 1);
```
Sqvid, line 1825: is there any difference between these branches? I think [`subs` just does a comparison and sets the flags](https://developer.arm.com/documentation/ddi0602/2025-03/Base-Instructions/SUBS--immediate---Subtract-immediate-value--setting-flags-?lang=en), which is what you seem to be doing manually in the first branch. In fact `cmp` is just a alias of `subs`.

> Compare (immediate)
>
>This instruction subtracts an optionally-shifted immediate value from a register value. It updates the condition flags based on the result, and discards the result.
>
>This is an alias of [SUBS (immediate)](https://developer.arm.com/documentation/ddi0602/2025-03/Base-Instructions/SUBS--immediate---Subtract-immediate-value--setting-flags-?lang=en).

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1447,21 +1384,13 @@ void jit_brgemm_kernel_t::gemm_microkernel(int bd_block2, bool is_bdb_tail,
...
-                    ld1rw(z1.s, P_ALL_ONE / T_z, ptr(X_DEFAULT_ADDR));
-                }
-            } else if (one_of(dt, data_type::s8, data_type::u8,
-                               data_type::bf16)) {
-                add_imm(X_DEFAULT_ADDR, reg_aux_A, offset, X_TMP_0);
-                ld1rw(z1.s, P_ALL_ONE / T_z, ptr(X_DEFAULT_ADDR));
-            } else if (dt == data_type::f16) {
-                assert(!"unsupported\n");
+            auto offset_ = offset - base_offset;
+            if (offset_ > 252 || offset_ % 4 != 0) {
```
Sqvid, line 1395: nit: Let's try to avoid the magic numbers to make it more self-documenting (even though the "magic" isn't super obscure here). Maybe define `dt_size = sizeof(float);` and then instead for `252` you want `vl_size - dt_size)` for instance (or whatever else the intended meaning is).

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1554,7 +1484,8 @@ void jit_brgemm_kernel_t::gemm_microkernel(int bd_block2, bool is_bdb_tail,
                 for (int ld = 0; ld < ld_block2; ld++) {
                     auto zmm = accm(ld_block2, bd, ld);
                     if (is_emdbd) {
-                        if (A_offset(bd, rd) < (1 << 6)) {
+                        if (A_offset(bd, rd) < 252
```
Sqvid: Same as below (magic numbers)

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
Ryo-not-rio: The second branch is one less instruction because we can set the flag and subtract in one instruction. This is very marginally better than doing, `sub, cmp, b`. We can only do it in the second branch because we're comparing if greater than 0.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
Ryo-not-rio: done

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
Ryo-not-rio: done

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -222,6 +222,18 @@ struct jit_brgemm_kernel_t : public jit_generator_t {
     PReg ld_full_mask = PReg(2);
     PReg ld_tail_mask = PReg(3);
 
+    void add_vl_or_imm(XReg dst, XReg src, int offset) {
+        // If offset is a multiple of the vector length and
+        // offset / vector_length is compatible with addvl
+        // use the addvl instruction. Refer to https://developer.arm.com/documentation/ddi0596/2021-03/SVE-Instructions/ADDVL--Add-multiple-of-vector-register-size-to-scalar-register-
+        // for details.
+        if ((offset % cpu_sveLen == 0) && (offset / (int)cpu_sveLen >= -32)
```
Sqvid: nit: prefer `static_cast<int>` to `old-style-cast`

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1447,21 +1388,15 @@ void jit_brgemm_kernel_t::gemm_microkernel(int bd_block2, bool is_bdb_tail,
...
-                assert(!"unsupported\n");
+            auto offset_ = offset - base_offset;
+            // The ld1rw immediate must be <=252 and a multiple of 4
+            // refer to https://developer.arm.com/documentation/ddi0602/2025-09/SVE-Instructions/LD1RW--Load-and-broadcast-unsigned-word-to-vector-
+            if (offset_ > 252 || offset_ % 4 != 0) {
+                add_imm(X_TMP_2, X_TMP_2, offset_, X_TMP_0);
+                base_offset += offset_;
+                offset_ = 0;
             }
+            ld1rw(z1.s, P_ALL_ONE / T_z, ptr(X_TMP_2, (int32_t)offset_));
```
Sqvid: nit: `static_cast<int32_t>`

Ryo-not-rio: done

PR #4388, opened 2025-11-26
xe: move stable IR interfaces

rjoursler:
First part in aligning oneDNN file structure with the intended upstreaming to Gemmstone. To reduce scope, this PR focuses on moving `ir/include` and its dependencies. I plan to implement two additional PR's in this series, one moving DSL to Gemmstone and one moving code generation.

### DIFF
--- a/src/gpu/intel/gemm/jit/dsl/ir/ir.hpp
+++ b/src/gpu/intel/gemm/jit/dsl/ir/ir.hpp
@@ -75,7 +75,7 @@ expr_t make_buffer(const std::string &name);
 
 class buffer_manager_t {
 public:
-    struct entry_t {
+    struct entry_t : public stringify_t<entry_t> {
         entry_t() = default;
         entry_t(const expr_t &buf, int size) : buf(buf), size(size) {}
 
@@ -685,7 +685,7 @@ struct linear_transform_t {
 // - lhs is a variable
 // - rhs is an integer constant
 // - op is a comparison operation
-class relation_t {
+class relation_t : public stringify_t<relation_t> {
 public:
     relation_t(const expr_t &expr) : expr_(normalize(expr)) {}
 
@@ -728,7 +728,7 @@ private:
 // Equality for modulus: (var % mod) == 0, where:
 // - var is a variable
 // - mod is an integer constant
-class modulus_info_t {
+class modulus_info_t : public stringify_t<modulus_info_t> {
 public:
     modulus_info_t(const expr_t &expr) : expr_(expr) {}
 
@@ -827,7 +827,7 @@ private:
 };
 
 // TODO: Add integers check (only integers can be constrained).
-class constraint_set_t {
+class constraint_set_t : public stringify_t<constraint_set_t> {
 public:
     const object_map_t<expr_t, std::vector<relation_t>> &relations() const {
         return relations_;
--- a/src/gpu/intel/gemm/jit/dsl/utils/profiler.hpp
+++ b/src/gpu/intel/gemm/jit/dsl/utils/profiler.hpp
@@ -91,7 +91,7 @@ static double get_msec() {
 // All pointers need to be valid when stop() is called, at which point stamp
 // names are copied into long term storage.
 
-struct profiler_t {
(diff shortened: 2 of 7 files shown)


src/gpu/intel/post_ops.hpp
```diff
@@ -188,6 +191,14 @@ struct relative_md_t {
     };
 
     relative_md_t() = default;
+    relative_md_t(int md_broadcast_mask, int ndims, data_type_t dt,
+            const ndim_normalizer_t &ndim_normalizer)
+        : dt(dt), broadcast_mask(uint16_t(~0 << ndims)), inner_dim(0) {
```
echeresh: Looks like it's UB, as we left-shift a negative value.

```suggestion
        : dt(dt), broadcast_mask(uint16_t(0xFFFFu << ndims)), inner_dim(0) {
```

rjoursler:
make test
disable test_device_cpu

dzarukin (approved):

src/gpu/intel/conv/jit/key.hpp
```diff
@@ -67,7 +63,7 @@ class key_t {
     std::string str(bool csv = false) const;
     static std::vector<std::string> csv_keys();
 
-    IR_DEFINE_DUMP()
+    void dump() const { printf("%s\n", str().c_str()); }
```
echeresh: Can we keep a special macro for that?

This definition needs attributes to work under gdb, see an example below.

```bash
$ cat 1.cpp
#include <cstdio>

struct Foo {
    void dump1() const {
        printf("Foo\n");
    }

    __attribute__((noinline))
    __attribute__((used))
    void dump2() const {
        printf("Foo\n");
    }
};
int main() {
    Foo foo;
    return 0;
}

...
(gdb) call foo.dump1()
Cannot evaluate function -- may be inlined
(gdb) call foo.dump2()
Foo
```

src/gpu/intel/conv/jit/plan.cpp
```diff
@@ -154,21 +154,21 @@ void init_fwd(const config_t &cfg_, gemm_schedule_t &gemm_schedule,
...
 
     // Initialize views.
-    auto mb = var_t::make(type_t::s32(), "mb");
-    auto ic = var_t::make(type_t::s32(), "ic");
-    auto oc = var_t::make(type_t::s32(), "oc");
-    auto kd = var_t::make(type_t::s32(), "kd");
-    auto kh = var_t::make(type_t::s32(), "kh");
-    auto kw = var_t::make(type_t::s32(), "kw");
-    auto g = var_t::make(type_t::s32(), "g");
+    auto mb = var_t::make(dsl::type_t::s32(), "mb");
```
echeresh, line 157: As an option we can introduce aliases for common DSL types inside gpu::intel (such as type_t or layout_t, etc) to reduce boilerplate.

src/gpu/intel/conv/jit/key.hpp
rjoursler: Definitely, I went ahead and moved IR_DEFINE_DUMP into `intel/utils.hpp` (and renamed it XE_DEFINE_DUMP) as this seems like is is not IR specific. I am trying a different solution within the Gemmstone code though, just to limit macros by using CRTP as it additionally use it to resolve some issue with `operator<<` being ambiguous (although for some reason adding the CRTP structure to `type_t` is causing `one_of` to be ambiguous....).

src/gpu/intel/conv/jit/plan.cpp
rjoursler: My current plan is to generally add:
```
namespace dsl = gemmstone::dsl;
namespace ir = gemmstone::dsl::ir;
```
and leave further simplification up to specific implementations. I am not convinced we should broadly be overriding these as we already have some duplicate naming (especially in the conv::v2 model code) and to generally avoid inconsistency in places where there is a naming clash on specific names.

rjoursler:
make test
disable test_device_cpu

rjoursler:
make test
disable test_device_cpu

PR #4389, opened 2025-11-26
gpu: jit: gemm: convert hpc kernel to xe2/xe3 strategy

h-sadia:


h-sadia:
make test perf-gpu
set primitive=gpu:gemm

PR #4403, opened 2025-12-01
cpu: aarch64: refactor jit_uni_reorder.* into smaller compilation units

Sqvid:
# Description

Refactor of our jitted reorder code. The previous structure dumped all reorder-related code in `jit_uni_reorder.cpp`. This patch is a first attempt at breaking up that structure into smaller, more maintainable files.

1. Moved the all reorder code to `cpu/aarch64/reorder/`
2. All `prb_t`-related magic has been moved to `jit_uni_reorder_utils.*`
3. `jit_{uni,blk}_reorder.hpp` only contains the definition of the drivers `jit_{uni,blk}_reorder_t`
4. `jit_{uni,blk}_reorder.cpp` only contains the implementations of the drivers
5. The kernel definitions have been moved to `jit_{uni,blk}_reorder_kernel.hpp`
6. The kernels are then implemented by `jit_{uni,blk}_reorder_kernel.cpp`

# Checklist

## General


Sqvid:
Tested that the there is no change to the implementations selected with `--reorder --batch=test_reorder_all`. Performance pipeline also shows no regressions.

Sqvid:
The added lines are due to the fact that `class` definitions and impls are now separate.

So if we had:

**code.hpp**:
```cpp
class Code {
    void member() { doSomething(); }
};
```
We now have:

**code.hpp**:
```cpp
class Code {
    void member();
};
```
**code.cpp**:
```cpp
void Code::member() { doSomething(); }
```

PR #4404, opened 2025-12-02
graph: backend: dnnl, tests: benchdnn: support sdpa / gqa training with gradients for mask

ElaineBao:
# Description

MFDNN-14010.

- [x] Use an End op to represent an intermediate output (which has consumers within the partition as well as outside the partition)
- [x] Add patterns and related implementation for sdpa/gqa training backward w.r.t gradients for mask
- [x] Add test cases, modify benchdnn to support such cases' validation.


src/graph/interface/op_def.hpp
```diff
@@ -410,7 +410,8 @@ DNNL_GRAPH_OP_SCHEMA(End, 1,
                 .set_type_constraints("T",
                         {data_type::f32, data_type::f16, data_type::bf16,
                                 data_type::s8, data_type::u8, data_type::s32,
-                                data_type::undef}))
+                                data_type::undef})
+                .set_shape_inference_function(infer_dummy_output_shape))
```
TaoLv, line 414: Why do we need this? By definition, there is no output for an End op.

ElaineBao: The graph's infer shape function will iterate every op in the graph, we can certainly check if one op has output or not to avoid adding a dummy infer shape function for End op here, but I think it's not a better choice because in that way, every op will have to check that condition.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/dnnl_partition_impl.cpp
```diff
@@ -63,6 +63,13 @@ void dnnl_partition_impl_t::init_inputs_outputs() {
     }
 
     for (auto &cur_op : ops_) {
+        // Handle End op: directly mark its input as output
+        if (cur_op->get_kind() == graph::op_kind::End) {
+            auto end_input = cur_op->get_input_value(0);
+            outputs_.push_back(end_input->get_logical_tensor());
```
TaoLv:
```suggestion
            outputs_.push_back(end_input->get_input_logical_tensor(0));
```

src/graph/backend/dnnl/layout_propagator.cpp
```diff
@@ -1668,6 +1668,23 @@ status_t layout_propagator_for_gen_index(std::shared_ptr<op_t> &op,
     return status;
 }
 
+status_t layout_propagator_for_identity(std::shared_ptr<op_t> &op,
+        const dnnl::engine &p_engine, pd_cache_t &pd_cache,
+        const fpmath_t &fpmath, bool use_block_layout,
+        subgraph_rewriter_t &rewriter) {
+    logical_tensor_t dst_lt = op->get_input_value(0)->get_logical_tensor();
```
TaoLv:
```suggestion
    logical_tensor_t dst_lt = op->get_input_logical_tensor(0);
```

src/graph/backend/dnnl/layout_propagator.cpp
```diff
@@ -1668,6 +1668,23 @@ status_t layout_propagator_for_gen_index(std::shared_ptr<op_t> &op,
...
+    VCHECK_LAYOUT_PROPAGATOR(!ltw(dst_lt).is_any(), status::invalid_arguments,
+            "layout of identity dst can't be any");
+    auto dst_md = make_dnnl_memory_desc(dst_lt);
+    auto src_md = make_dnnl_memory_desc(
+            op->get_input_value(0)->get_logical_tensor());
```
TaoLv:
```suggestion
    auto src_md = make_dnnl_memory_desc(op->get_input_logical_tensor(0));
```

TaoLv (commented):
Any changes required for document? What's the main difference in the fusion graph?

src/graph/backend/dnnl/patterns/sdp.cpp
```diff
@@ -253,6 +253,90 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, float_sdp_backward_fusion)
                     pgraph->create_input_port(2, matmul_dv, 1);
                     pgraph->create_input_port(2, matmul_v_do, 0);
                 })
+        .set_attr<FCreatePattern>("FCreatePattern",
```
dzarukin, line 256: Just curious why a new pattern and not incorporating into existing one?

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -822,9 +822,16 @@ bool deserialized_graph_t::detect_sdpa_bwd_impl() const {
...
 
         // find MatMul for dQ or dV
-        cur_op_ref = get_child_ops(cur_op_refs[softmax_bwd_idx])[0];
+        cur_op_refs = get_child_ops(cur_op_refs[softmax_bwd_idx]);
+        // consider the case w/ & w/o gradients w.r.t. mask.
+        if (cur_op_refs.size() > 2) continue;
+        size_t matmul_idx = 0;
+        if (cur_op_refs.size() == 2 && cur_op_refs[0].kind_ == "End") {
+            matmul_idx = 1;
+        }
```
dzarukin, line 829: Would it be better if:
```
while (cur_op_refs[matmul_idx].kind_ != "MatMul") {
    matmul_idx++;
}
```

dzarukin (approved):

src/graph/backend/dnnl/dnnl_partition_impl.cpp
ElaineBao: updated, thanks

ElaineBao:
> Any changes required for document? What's the main difference in the fusion graph?

Added document changes, please take another look @TaoLv

src/graph/backend/dnnl/patterns/sdp.cpp
ElaineBao: It's because incorporating into existing one requires the pattern matcher's optional subgraph supports two consumers, which is currently not supported

tests/benchdnn/graph/deserialize.cpp
ElaineBao: Not really, becuase matmul_idx (maybe it's not a good name) doesn't mean the consumer must be matmul (below find_next_until indicates that), it means matmul is in the chain of that consumer. On the other hand, the End op is determined.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4405, opened 2025-12-02
cpu: aarch64: conv: optimize sve_1x1 kernel

Ryo-not-rio:
- Removed unnecessary loads
- Updated to use LD_MUL_VL and ST_MUL_VL where possible
- Move add_vl_or_imm to jit_generator.hpp and change add_imm to add_vl_or_imm where appropriate
- Updated to NOT use explicit broadcast in most cases
- Fix potential invalid immediate parameter for ld1rw in brgemm_kernel

Speedups on 16 threads:
c7g: ~1.11
c8g: ~1.11x ~ 1.89x

### DIFF
--- a/src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
+++ b/src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
@@ -98,17 +98,17 @@ void jit_sve_1x1_conv_kernel_t<isa_>::bcast_loop(int load_loop_blk) {
             if (i + 1 == num_substeps) L(large_tail);
             reduce_loop(load_loop_blk, jcp.ur, i, false);
             if (i < num_substeps - 1) {
-                add_imm(aux1_reg_bcast_data, aux1_reg_bcast_data,
+                add_vl_or_imm(aux1_reg_bcast_data, aux1_reg_bcast_data,
                         jcp.bcast_loop_bcast_substep, reg_tmp_imm);
-                add_imm(aux_reg_output_data, aux_reg_output_data,
+                add_vl_or_imm(aux_reg_output_data, aux_reg_output_data,
                         jcp.bcast_loop_output_substep, reg_tmp_imm);
             } else {
-                add_imm(aux1_reg_bcast_data, aux1_reg_bcast_data,
+                add_vl_or_imm(aux1_reg_bcast_data, aux1_reg_bcast_data,
                         jcp.bcast_loop_bcast_step
                                 - (num_substeps - 1)
                                         * jcp.bcast_loop_bcast_substep,
                         reg_tmp_imm);
-                add_imm(aux_reg_output_data, aux_reg_output_data,
+                add_vl_or_imm(aux_reg_output_data, aux_reg_output_data,
                         jcp.bcast_loop_output_step
                                 - (num_substeps - 1)
                                         * jcp.bcast_loop_output_substep,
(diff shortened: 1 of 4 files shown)


Sqvid:
> - Apply 1x1 conv optimizations
> - Fix potential bug in brgemm_conv kernel

Could you provide a high-level description of what optimisations you made, and what bug you fixed?

Ryo-not-rio:
All nightly unit tests are passing and a lot of conv regressions are resolved with this patch: https://github.com/uxlfoundation/oneDNN/actions/runs/19906852997

src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
```diff
@@ -376,41 +382,47 @@ void jit_sve_1x1_conv_kernel_t<isa_>::reduce_loop(
...
 
             for (int i_ur = 0; i_ur < ur; ++i_ur) {
+                int offset = bcast_offset(i_reduce, i_ur);
+                int offset_ = offset - base_offset;
+
                 if (jcp.expl_bcast && load_loop_blk > 1) {
-                    ldr(W_TMP_0,
-                            ptr(bcast_ptr(i_reduce, i_ur, false, X_DEFAULT_ADDR,
-                                    X_TMP_1)));
+                    if (offset_ < -256 || offset_ >= 255) {
```
jondea: Some short comments would be useful I think here

src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
```diff
@@ -925,7 +937,7 @@ status_t jit_sve_1x1_conv_kernel_t<isa_>::init_conf(jit_1x1_conv_conf_t &jcp,
             max_regs = 9; // max # of ur_w
             min_regs = 6; // min # of ur_w
             size_threshold = 14;
-            jcp.expl_bcast = true;
+            jcp.expl_bcast = false;
```
jondea: What does this mean? Why can we do it? Again, a comment explaining what this variable means would be useful

src/cpu/aarch64/jit_sve_1x1_conv_kernel.hpp
```diff
@@ -109,21 +111,6 @@ struct jit_sve_1x1_conv_kernel_t : public jit_generator_t {
     constexpr static int reg_abi_param1_backup = 2 * reg64_size_;
     constexpr static int stack_space_needed = 3 * reg64_size_;
 
-    template <typename T>
-    Xbyak_aarch64::XReg EVEX_compress_addr(const Xbyak_aarch64::XReg &addr,
```
jondea, line 113: What is this thing? Why can we get rid of it?

src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
Ryo-not-rio: done

src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
Ryo-not-rio: This is a flag that decides whether we use `ld1rw` or `ldr`, it seems like an optimization flag and from benchmarking all the 1x1 shapes, I only see speedups from setting it to false. It's not clear to me why it was set to true in the first place but it's not causing unit test failures or regressions

src/cpu/aarch64/jit_sve_1x1_conv_kernel.hpp
Ryo-not-rio: This was used in the pointer retrieval functions which I changed to return offsets instead so it is no longer used

src/cpu/aarch64/jit_sve_1x1_conv_kernel.cpp
```diff
@@ -925,7 +941,7 @@ status_t jit_sve_1x1_conv_kernel_t<isa_>::init_conf(jit_1x1_conv_conf_t &jcp,
             max_regs = 9; // max # of ur_w
             min_regs = 6; // min # of ur_w
             size_threshold = 14;
-            jcp.expl_bcast = true;
+            jcp.expl_bcast = false; // ld1rw is faster for msot cases
```
renato-arantes: Typo: msot instead of most.

PR #4406, opened 2025-12-02
Enable external brgemm API on aarch64

michalowski-arm:
This change is an initial implementation of the external brgemm API for aarch64, mostly directly based on the existing x64 code. For now, only supports f32 is supported, with support for int8 and bf16 (and further improvements) to come in later patches building on this baseline implementation.

To test this change, enable `DNNL_EXPERIMENTAL_UKERNEL` before compiling, then run benchdnn with `--brgemm` as usual. As an alternative to benchdnn, you can use `examples/ukernels/cpu_brgemm.cpp`, though it has to be changed slightly to use f32 instead of int8 and not set AB scales.

### DIFF
--- a/.github/automation/aarch64/build.sh
+++ b/.github/automation/aarch64/build.sh
@@ -31,6 +31,7 @@ export ACL_ROOT_DIR=${ACL_ROOT_DIR:-"${PWD}/ComputeLibrary"}
 CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-"Release"}
 ONEDNN_TEST_SET=${ONEDNN_TEST_SET:-"SMOKE"}
 ONEDNN_BUILD_GRAPH=${ONEDNN_BUILD_GRAPH:-"ON"}
+ONEDNN_EXPERIMENTAL_UKERNEL=${ONEDNN_EXPERIMENTAL_UKERNEL:-"ON"}
 
 if [[ "$ONEDNN_ACTION" == "configure" ]]; then
     if [[ "$GITHUB_JOB" == "pr-clang-tidy" ]]; then
@@ -44,7 +45,8 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
             -DDNNL_BUILD_FOR_CI=ON \
             -DONEDNN_TEST_SET=NO_CORR \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DDNNL_EXPERIMENTAL_UKERNEL=ON
         set +x
     else
         set -x
@@ -58,7 +60,8 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
             -DONEDNN_TEST_SET=$ONEDNN_TEST_SET \
             -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE \
             -DCMAKE_SKIP_BUILD_RPATH=FALSE \
-            -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON
+            -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON \
+            -DDNNL_EXPERIMENTAL_UKERNEL=$ONEDNN_EXPERIMENTAL_UKERNEL
         set +x
     fi
 elif [[ "$ONEDNN_ACTION" == "build" ]]; then
--- a/tests/benchdnn/brgemm/brgemm.cpp
+++ b/tests/benchdnn/brgemm/brgemm.cpp
@@ -407,11 +407,20 @@ int init_kernel(kernel_args_t &kernel_args) {
 
     dnnl_status_t st = dnnl_success;
     auto &brgemm = kernel_args.brgemm_;
-    DNN_SAFE(
-            dnnl_brgemm_create(&brgemm, prb->m, prb->n, prb->k, prb->batch_size,
-                    prb->get_lda(), prb->get_ldb(), prb->get_ldc(),
-                    prb->src_dt(), prb->wei_dt(), prb->acc_dt()),
-            WARN);
+    // brgemm ukernel returns unimplemented for int8 and bf16
+    // on aarch64 at the moment, for now we'll treat those cases
+    // as SKIPPED rather than UNTESTED FAILED
+    st = dnnl_brgemm_create(&brgemm, prb->m, prb->n, prb->k, prb->batch_size,
+            prb->get_lda(), prb->get_ldb(), prb->get_ldc(), prb->src_dt(),
+            prb->wei_dt(), prb->acc_dt());
+#if defined(brg_aarch64)
+    if (st == dnnl_unimplemented) {
+        SAFE(check_dnnl_status(st, prb, res), WARN);
+        if (res->state == SKIPPED) return OK;
+    }
+#endif
+    DNN_SAFE(st, WARN);
+
     // Only `beta` equal to `0.f` and `1.f` works.
     DNN_SAFE(dnnl_brgemm_set_add_C(brgemm, static_cast<int>(prb->beta)), WARN);
     DNN_SAFE(dnnl_brgemm_set_post_ops(


src/cpu/aarch64/brgemm/brgemm_types.hpp
```diff
@@ -290,10 +292,25 @@ struct brgemm_t {
         return sz;
     }
 
-    bool is_b_data_layout_vnni() { return true; }
+    // A class version of the `static` version of the function.
+    // Note: used in benchdnn only, not used inside the library.
+    bool is_b_data_layout_vnni() const { return is_b_data_layout_vnni(dt_b); }
```
jondea, line 303: Would it be possible to change vnni to something correct for AArch64? pack_32? DOT?

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -989,6 +989,67 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     return status::success;
 }
 
+status_t init_conf(brgemm_matmul_conf_t &conf, dim_t batch, dim_t M, dim_t K,
```
jondea, line 992: Why does adding the external brgemm interface mean we add a new function for the internal brgemm_matmul?

jondea:
Also, can we make sure this is tested in CI somehow?

src/cpu/aarch64/brgemm/brgemm_types.hpp
michalowski-arm: It seemed to me there's a couple different uses of vnni that would work best with different names. I was gonna change that in a separate PR but I could include it here if you think that works better?

jondea: As long as it's on your radar, it's up to you

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
michalowski-arm: That `init_conf` is used by the transform, which is used for packing of B. We could move this logic elsewhere, I was just focusing on following the x64 approach as closely as possible for this initial version.

michalowski-arm:
> Also, can we make sure this is tested in CI somehow?

As it is right now, building with `DNNL_EXPERIMENTAL_UKERNEL` is going to use the external brgemm api for `./benchdnn --brgemm`, so it would probably need to be a separate, new pipeline (if we want to keep testing the internal brgemm api). Alternatively, we could maybe not replace internal brgemm benchdnn with the ukernel version on aarch64, and just write a suite of tests similar to `examples/ukernels/cpu_brgemm.cpp` for the CI. Either way, it's not gonna be as easy as just enabling the ukernel in the CI.

jondea:
I'd lean towards just enabling `DNNL_EXPERIMENTAL_UKERNEL` where we can. My expectation is that when you call `benchdnn --something` you are directly testing external things. Testing the internal brgemm interface feels like a fallback rather than an intentional thing. The internal brgemm will continue to be tested via the external API, and more directly where it is actually used e.g. `brgemm_matmal` and `brgconv` etc.

.github/automation/aarch64/build.sh
```diff
@@ -44,7 +45,8 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
             -DDNNL_BUILD_FOR_CI=ON \
             -DONEDNN_TEST_SET=NO_CORR \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DDNNL_EXPERIMENTAL_UKERNEL=ON
```
jondea, line 49: Should we be using `ONEDNN_EXPERIMENTAL_UKERNEL`?

michalowski-arm: It's the way to enable the brgemm ukernel api, or is your question whether we should always compile with the external api enabled and not have a flag for that?

jondea: Sorry I wasn't clear. I mean to say that we have a bash variable `ONEDNN_EXPERIMENTAL_UKERNEL` which you have used below, but here you have hardcoded it to be on.

michalowski-arm: This first configuration is only used in the "pr-clang-tidy" pipeline to get the compile_commands.json for clang. We could change it to also use the var but my understanding is it wouldn't matter here for the result of clang-tidy, and we are also not building onednn from this config. So I'd leave it as it just since it aligns with the rest of this script.

jondea (approved):
LGTM

renato-arantes:
Now that the API is enabled, I expected the example `examples/ukernels/cpu_brgemm.cpp` to be functional. Still, when running it after enabling `-DDNNL_EXPERIMENTAL_UKERNEL=ON`, it returns the error `Kernel is not supported on this platform` after querying `brgemm::get_B_pack_type`. Is this behaviour expected?

michalowski-arm:
> Now that the API is enabled, I expected the example `examples/ukernels/cpu_brgemm.cpp` to be functional. Still, when running it after enabling `-DDNNL_EXPERIMENTAL_UKERNEL=ON`, it returns the error `Kernel is not supported on this platform` after querying `brgemm::get_B_pack_type`. Is this behaviour expected?

Yea, that's expected for now. This PR only enables f32 path for the moment, with bf16 and int8 to follow later. The `cpu_brgemm.cpp` example is running an int8 brgemm, so that particular kernel is not supported. It can quite easily be modified to run fine you change dt to f32 and also remove setting of A/B scales (I think binary post-op also has to be disabled too, that's an unrelated limitation of brgemm on aarch64).

michalowski-arm:
@dzarukin we need a review from the onednn-arch team, could you have a look please? Thanks!

src/cpu/aarch64/matmul/brgemm_matmul.cpp
```diff
@@ -196,7 +196,7 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
 
         int idx = get_brg_kernel_idx(i_bs, i_init, i_M, i_N, i_K);
         if (idx < 0) continue;
-        brgemm_t &brg = brg_descs_[idx];
+        brgemm_desc_t &brg = brg_descs_[idx];
```
dzarukin, line 199: Minor: there're some leftovers of renaming in the commit after renaming happened which would trigger build errors on renaming commit. In case of triaging it helps to have healthy commits in terms of build and run, but it's up to you, just a general observation. Feel free to resolve.

src/cpu/aarch64/ukernel/transform.hpp
```diff
@@ -0,0 +1,92 @@
...
+    std::string verbose_info_;
+};
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+namespace ukernel {
+
+status_t dnnl_transform_create(dnnl_transform **transform, dim_t K, dim_t N,
```
dzarukin, line 71: QQ ( as I'm not familiar with aarch64 hardware features): is transform needed for B for the same purpose as it's needed for x64? Does the hardware have instructions to deal with plain B formats instantly without moving the data when it comes to xf16 or int8 data types?

src/cpu/aarch64/ukernel/transform.cpp
```diff
@@ -0,0 +1,211 @@
...
+namespace cpu {
+namespace aarch64 {
+namespace ukernel {
+
+status_t dnnl_transform_create(transform_t **transform, dim_t K, dim_t N,
+        pack_type_t in_pack_type, dim_t in_ld, dim_t out_ld, data_type_t in_dt,
+        data_type_t out_dt) {
+    if (transform == nullptr) return status::invalid_arguments;
+    VCHECK_TRANSFORM(utils::one_of(out_ld, 16, 32, 48, 64),
+            "Transform routine supports only \'out_ld\' of 16, 32, 48, or 64.");
```
dzarukin, line 175: Does this limitation hold for aarch64, too?

src/cpu/aarch64/ukernel/brgemm.cpp
```diff
@@ -0,0 +1,426 @@
...
+    if (get_verbose(verbose_t::exec_profile, component_t::ukernel)) {
+        create_verbose_info();
+    }
+    return status::success;
+}
+
+status_t brgemm_t::execute(const void *A_ptr, const void *B_ptr,
+        const dim_t *A_B_offsets, void *C_ptr, void *scratchpad_ptr) const {
+    const auto batch_size = brgemm_desc_.brgattr.max_bs;
+    std::vector<brgemm_batch_element_t> v_batch_element(batch_size);
```
dzarukin, line 181: We've recently updated this part. I recommend to apply the patch for aarch64 backend as well.

src/cpu/aarch64/ukernel/transform.hpp
michalowski-arm: on aarch64, bf16 and int8 fast multiply instructions require the "vnni" style packing, for f16 plain format is fine

src/cpu/aarch64/ukernel/transform.cpp
michalowski-arm: No, seems I missed this check. I'll fix it in the follow-up PR

src/cpu/aarch64/ukernel/transform.hpp
jondea: The BF/SDOT instructions which we currently use in brgemm use the vnni style packing format. You may be thinking of the BF/SMMLA instructions which require a different LHS and RHS packing (either ahead of time or on the fly).

PR #4411, opened 2025-12-03
cpu: aarch64: reorder: add SVE-256 fp32:bf16 ab->BA8b4a kernel

Sqvid:
# Description

Add an optimised kernel for f32->bf16 ab->BA8b4a reorders.

# Checklist

## General

## Performance improvements

cc: @davsva01 @jondea

### DIFF
--- a/src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
+++ b/src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
@@ -260,6 +260,125 @@ void jit_uni_reorder_kernel_f32_t::step(int off, int prev_i_off, int prev_o_off,
             step_size);
 }
 
+bool jit_uni_reorder_kernel_f32_t::can_do_tr4x8() {
+    using namespace data_type;
+
+    // The kernel is specialised for f32 -> bf16 reorders.
+    //
+    // This process relies on swapping the two innermost dimensions.
+    // Therefore, the input stride in the second node and output stride in
+    // first node have to be equal to 1.
+    return mayiuse(sve_256) && prb_.ndims >= 2
+            && (prb_.itype == f32 && prb_.otype == bf16) && prb_.n(0) == 4
+            && prb_.n(1) == 8 && utils::everyone_is(1, prb_.os(0), prb_.is(1))
+            && !prb_.is_tail_present
+            && prb_.src_scale_type == scale_type_t::NONE
+            && prb_.dst_scale_type == scale_type_t::NONE && prb_.beta == 0.f
+            && !compensation_needed_;
+}
+
+bool jit_uni_reorder_kernel_f32_t::process_unroll_tr4x8(
+        const int ndims, const int len) {
+    if (!can_do_tr4x8()) return false;
+
+    const int step_size = prb_.n(0) * prb_.n(1);
+    int i_off = 0, o_off = 0;
+    for (int off = 0; off < len; off += step_size) {
+        step(off, i_off, o_off, i_off, o_off, step_size);
+        tr4x8_sve256(i_off, o_off);
+    }
+
+    return true;
+}
+
+void jit_uni_reorder_kernel_f32_t::tr4x8_sve256(int i_off, int o_off) {
+    using namespace data_type;
+
+    auto z0 = ZRegS(0);
+    auto z1 = ZRegS(1);
+    auto z2 = ZRegS(2);
+    auto z3 = ZRegS(3);
+
+    assert(x_tmp_vec.size() >= 4);
+    auto x_tmp_0 = x_tmp_vec[0];
+    auto x_tmp_1 = x_tmp_vec[1];
+    auto x_tmp_2 = x_tmp_vec[2];
+    auto x_tmp_3 = x_tmp_vec[3];
+
+    // Load
+    auto in_ptr_diff = itype_sz_ * prb_.is(0);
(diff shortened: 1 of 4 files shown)


Sqvid:
### Speedups:

```bash
# Before
$ OMP_NUM_THREADS=16 ./build-main/tests/benchdnn/benchdnn --reorder --mode=P --allow-enum-tags-only=0 --sdt=f32 --ddt=bf16 --stag=ab --dtag=BA8b4a 14336x4096
| acl : 1 (100%)                                           |
total perf: min(ms):2.50562 avg(ms):2.53653

# After
$ OMP_NUM_THREADS=16 ./build-pr/tests/benchdnn/benchdnn --reorder --mode=P --allow-enum-tags-only=0 --sdt=f32 --ddt=bf16 --stag=ab --dtag=BA8b4a 14336x4096
| jit:uni : 1 (100%)                                       |
total perf: min(ms):1.60986 avg(ms):1.65996
```

src/cpu/aarch64/reorder/jit_uni_reorder_utils.cpp
```diff
@@ -680,6 +680,13 @@ void prb_block_for_cache(prb_t &prb) {
                               && prb.nodes[1].n > num_elems_thr))
             && !prb.is_tail_present;
 
+    // TODO: Find a way to associate the caching logic its kernel.
+    // The issue is that this swap logic is separated from the tr4x8 kernel that
+    // it is relevant to.
+    if (prb.ndims == 4 && prb.n(0) == 4 && prb.n(1) == 8) {
+        tr::prb_node_move(prb, 2, 3);
```
davsva01, line 696: Could you please add a bit more information here on why this is required? A small example on what the prb looks like before and after would make it a lot more clear I think.

src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
```diff
@@ -260,6 +260,88 @@ void jit_uni_reorder_kernel_f32_t::step(int off, int prev_i_off, int prev_o_off,
...
+    add_imm(x_tmp_1, x_tmp_0, 1 * in_ptr_diff, X_DEFAULT_ADDR);
+    add_imm(x_tmp_2, x_tmp_0, 2 * in_ptr_diff, X_DEFAULT_ADDR);
+    add_imm(x_tmp_3, x_tmp_0, 3 * in_ptr_diff, X_DEFAULT_ADDR);
+
+    ld1w(z0, P_ALL_ONE, ptr(x_tmp_0));
+    ld1w(z1, P_ALL_ONE, ptr(x_tmp_1));
+    ld1w(z2, P_ALL_ONE, ptr(x_tmp_2));
+    ld1w(z3, P_ALL_ONE, ptr(x_tmp_3));
+
+    // Transpose
```
davsva01, line 320: Some comments here on how this transpose works would be fantastic, especially some of the less obvious lines like bfcvt vs bfcvtnt.

src/cpu/reorder/cpu_reorder_regular_f32_bf16.cpp
```diff
@@ -36,8 +36,8 @@ const impl_list_map_t &regular_f32_bf16_impl_list_map() {
             DNNL_NON_X64_ONLY(REG_SR_BIDIR(f32, any, bf16, nChw16c))
             DNNL_NON_X64_ONLY(REG_SR_BIDIR(f32, any, bf16, nCdhw16c))
 
-            DNNL_AARCH64_ACL_ONLY(CPU_REORDER_INSTANCE(aarch64::acl_reorder_fwd_t))
             DNNL_AARCH64_ONLY(CPU_REORDER_INSTANCE(aarch64::jit_uni_reorder_t))
+            DNNL_AARCH64_ACL_ONLY(CPU_REORDER_INSTANCE(aarch64::acl_reorder_fwd_t))
```
davsva01, line 39: Does this not mean that ACL reorder will never be called, even for cases with tails etc? Should we not just remove it from this list in that case?

Sqvid: That is almost certainly true. However I plan to leave the removal of `acl_reorder_t` to a follow up refactor PR so it can be tested on its own.

src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
Sqvid: Done.

src/cpu/aarch64/reorder/jit_uni_reorder_utils.cpp
Sqvid: Done.

src/cpu/reorder/cpu_reorder_regular_f32_bf16.cpp
Sqvid: I've opted to be even more conservative with the dispatch logic in this patch and just skip this particular case in `acl_reorder_t`

Ryo-not-rio (changes requested):
Looks like there's a regression, though it's ba -> BA8b4a. Did you change any codepaths for that reorder? + What's the plan for the SVE-128 path? No need to add it here since the code looks VLS but do we have a plan to add that as well?

Sqvid:
`BA8b4a` is not used on VL-128 so no plans to accelerate it there. That said, causing regression there was a defect that I have hopefully fixed with my latest update. Thanks for catching that.

PR #4412, opened 2025-12-03
rfcs: grouped gemm support proposal

mzhukova:
- Direct [design doc link](https://github.com/uxlfoundation/oneDNN/tree/mzhukova/rfcs/grouped-gemm-support/rfcs/20251203-grouped-gemm-support#readme).
- POC (WIP) branch for [Proposal 1](https://github.com/uxlfoundation/oneDNN/tree/mzhukova/main/poc-grouped-mem).
- RFC is for [MFDNN-14187](https://jira.devtools.intel.com/browse/MFDNN-14187), [MFDNN-14034](https://jira.devtools.intel.com/browse/MFDNN-14034) requests.

### DIFF
--- a/rfcs/20251203-grouped-gemm-support/README.md
+++ b/rfcs/20251203-grouped-gemm-support/README.md
@@ -39,6 +39,12 @@ And depending on the framework design, this information may only be available on
 See also:
 - [Useful resource to learn about MoE in PyTorch with examples of MoE workflow and challenges](https://pytorchconference.sched.com/event/27QE0/pytorch-apis-for-high-performance-moe-training-and-inference-daniel-vega-myhre-ke-wen-natalia-gimelshein-meta)
 
+In addition, grouped GEMM is also used for backward pass of MoE layers,
+where we have:
+
+- Backward pass wrt activations: `grad_src[i] = grad_output[i] * weights[i]^T`
+- Backward pass wrt weights: `grad_weights[i] = src[i]^T * grad_output[i]`
+
 ## Frameworks Implementations
 
 ### PyTorch Grouped Scaled MM
@@ -69,8 +75,8 @@ for drop-in replacement with various quantization schemes.
 Memory layout:
 ```
 mat_a:  [Expert0: M0xK | Expert1: M1xK | Expert2: M2xK]
-offs:    ^              ^              ^
-         0              M0             M0 + M1
+offs:                   ^              ^
+                        M0             M0 + M1
 Weights: [num_experts x K x N]
 ```
 
@@ -125,19 +131,22 @@ Representation is:
 Contiguous memory: [Expert0 | Expert1 | Expert2 | Expert3]
 Block sizes:       [M0 x K  | M1 x K  | M2 x K  | M3 x K ]
 
-Offsets array is for tracking where each expert starts:
-offsets = [0, M0, M0+M1, M0+M1+M2, M0+M1+M2+M3]
+Offsets array is for tracking where each grouped GEMM should start/end:
+offsets = [M0, M0+M1, M0+M1+M2, M0+M1+M2+M3]
 
 In the case of M_i equals zero, the next expert starts at the same offset, similar to CSR format row pointers.
-Example: if M1 = 0, then offsets = [0, M0, M0, M0 + M2, M0 + M2 + M3]
+Example: if M1 = 0, then offsets = [M0, M0, M0 + M2, M0 + M2 + M3]
 ```
+
+Note, that in contrast to CSR format, where offsets typically starts with zero,
+here offsets start from the size of the first block (M0), that is aligned with PyTorch design.
 
 The grouped memory descriptor defines:
 - Block count: `num_experts` (known at creation)
 - Total elements: `num_input_tokens x TOP_K x K` (known at creation)
 - Data types: values and offsets (known at creation)
 - Values buffer: Contiguous data sum(Mi x K) (provided at execution)
-- Offsets buffer: `[0, M0, M0 + M1, ...]` size `num_experts + 1` (at execution)
+- Offsets buffer: `[M0, M0 + M1, ...]` size `num_experts` (at execution)
 
 This representation is analogous to CSR sparse format:
 - CSR uses row pointers and column indices to describe sparse structure, buffers are resolved
@@ -167,7 +176,7 @@ static memory::desc grouped(
 
 The descriptor specifies a memory object with 2 buffers:
 - Buffer 0: Values, contiguous data organized as groups
-- Buffer 1: Offsets, cumulative row offsets of size `group_count + 1`
+- Buffer 1: Offsets, cumulative row offsets of size `group_count`
 
 Example:
 ```cpp
@@ -199,7 +208,7 @@ Memory objects are created with 2 buffer pointers:
 ```cpp
 // Buffers are coming from Framework side (could be device memory)
 float* values;       // total_tokens * K contiguous data
-int32_t* offsets;    // num_experts + 1 offsets
+int32_t* offsets;    // num_experts offsets
 
 // Create memory object with 2 buffers
 memory src_mem(src_md, eng, {values, offsets});


mzhukova:
FYI @umar456 and @echeresh

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+    const dims &grouped_dims,        // Group dimensions (clarification below)
+    data_type offsets_dt = s32,      // Offset buffer dtype (default int32)
+    const dims &astrides = {});      // Optional strides (default row-major)
+```
+
+Suggestion is to allow specifying `grouped_dims` in two ways for convenience:
+- **Shared dimensions**: `{M, K}` where M can be RUNTIME_DIM_VAL (size = 2)
+    - Useful for device-side runtime dimension M
+- **Per-group dimensions**: `{M0, K, M1, K, ...}` (size = group_count * 2)
+    - Useful when all dimensions are known on the host in advance
```
echeresh: A few questions:
1. In case of this version, do the user have to pass `offsets` during memory creation? (must pass the same? or can keep empty?)
2. Do we have known scenarios when M sizes are available on the host? If no, we could start with device-only offsets to reduce the scope, and extend later if needed

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+    dim group_count,                 // Group size or number of experts (all, not just active ones)
+    const dims &grouped_dims,        // Group dimensions (clarification below)
+    data_type offsets_dt = s32,      // Offset buffer dtype (default int32)
+    const dims &astrides = {});      // Optional strides (default row-major)
+```
+
+Suggestion is to allow specifying `grouped_dims` in two ways for convenience:
+- **Shared dimensions**: `{M, K}` where M can be RUNTIME_DIM_VAL (size = 2)
+    - Useful for device-side runtime dimension M
+- **Per-group dimensions**: `{M0, K, M1, K, ...}` (size = group_count * 2)
```
echeresh: Is grouping limited to `M` only? Maybe we can omit `K` sizes then?

Since the 1st option assumes adding a generic group memory descriptor, we can generalize it a bit to allow any number of dimensions in the descriptor and any dimension (but exactly one) to be grouped.

```c++
// Regular strided descriptor
auto md = memory::desc(dims, type, strides);

// Grouped descriptor: assign one dimension as "grouped" and specify the number of parts it is split into.
auto md_grouped = memory::desc::grouped(dims, type, strides, group_dim_idx, group_count);
```

Also this interface would avoid `RUNTIME_DIM_VAL` usage.

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+#### Memory Object Creation
+
+Memory objects are created with 2 buffer pointers:
+
+```cpp
+// Buffers are coming from Framework side (could be device memory)
+float* values;       // total_tokens * K contiguous data
+int32_t* offsets;    // num_experts + 1 offsets
```
echeresh: Looks like `offs` semantics here is different by the above PyTorch link: https://github.com/pytorch/pytorch/blob/bc6a4863c7246a6493d16d4ea6eee71ec07c6a09/torch/nn/functional.py#L6672

Can it be a problem? (e.g. for oneDNN usage PyTorch implementation would have to reorganize offsets via a separate kernel).

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+float* input_data;
+float* output_data;
+float* weights_data;
+
+// Offsets data coming from routing kernel
+int32_t* device_offsets;
+
+// Create memory objects (resolve values and offsets buffers)
+memory src_mem(src_md, eng, {input_data, device_offsets});
+memory dst_mem(dst_md, eng, {output_data, device_offsets});
```
echeresh, line 270: An observation: group GEMM operation requires the same offsets for src and dst - because offsets map to the corresponding `M` sizes. With "grouped" descriptors and separate memory per src/dst, we have to validate or assume that both src/dst offsets are the same. This gets a bit weird as:

- We can't really validate if they are the same as both offsets are device side
- We can specifically require from the user to pass the same buffer or pointer for both src/dst offsets but then such API is slightly redundant - what's the point to duplicate it in both memory objects

This is mostly an implication of the more general API via descriptors. It's generally fine and can be documented and handled but just wanted to bring this to your attention - maybe there are suggestions how to handle this better.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: > In case of this version, do the user have to pass offsets during memory creation? (must pass the same? or can keep empty?)

Good point, current expectation was that they must be the same.

> Do we have known scenarios when M sizes are available on the host? If no, we could start with device-only offsets to reduce the scope, and extend later if needed

Current OV implementation does that, but ideally this shouldn't be the case (since we don't want unnecessary device/host sync point).

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: > Is grouping limited to M only? Maybe we can omit K sizes then?

You mean avoiding the `grouped_dims` altogether and having `K` only as whole concatenated memory dim?

> auto md_grouped = memory::desc::grouped(dims, type, strides, group_dim_idx, group_count);

This looks doable :) I'll think about it 

> Also this interface would avoid RUNTIME_DIM_VAL usage

Is there any particular reasons to avoid `RUNTIME_DIM_VAL` concept? The idea was to have a placeholder in case we would need CPU support (or at least host routing)

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Could you elaborate?
I think marking the end or the start of the group should be roughly the same.. (i.e., we use either offs[i] or offs[i+1]), but maybe I misunderstood your point.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: I agree, that is somewhat redundant, but I thought it would be easy enough/cheap to just have the same pointer for offs (as in this snippet). 
I guess I don't have a good suggestion yet for Proposal 1, but it is not a problem for 2 and 3..

rfcs/20251203-grouped-gemm-support/README.md
echeresh: > You mean avoiding the grouped_dims altogether and having K only as whole concatenated memory dim?

My question was about why we need to repeat `K` here: `{M0, K, M1, K, ...}`, why not just `{M0, M1, ...}`?

> Is there any particular reasons to avoid RUNTIME_DIM_VAL concept? The idea was to have a placeholder in case we would need CPU support (or at least host routing)

I'm not sure what it adds on top of `offsets` inside memory...
Typical usage of `DNNL_RUNTIME_DIM_VAL` is to create a primitive supporting universal sizes and then pass fixed sizes during execution. In our case we have:

- For GPU: kernel needs device-side sizes - they are passed anyway, no need retrieving them via `RUNTIME_DIM_VAL`
- For CPU: kernel can directly access them from `offsets` as well - it's regular access for CPU

So I view `DNNL_RUNTIME_DIM_VAL` for this particular usage as just a tag/placeholder to indicate "dynamic" size. However, the original mechanism besides `DNNL_RUNTIME_DIM_VAL` is not needed as we have offsets passed inside `offsets` array at execution.

rfcs/20251203-grouped-gemm-support/README.md
echeresh: I see two mismatches at first glance:

1. RFC requires `0` at the beginning while PyTorch interface doesn't have it. To put it there, someone should allocate a new array and copy the PyTorch style array and prepend a zero
2. By the link: "``offs[-1]`` must be strictly less than the total length of that operand's sliced dimension" which seems like `offs` is expected to be like `[M0 - 1, M0 + M1 - 1, ..., M_total - 1]` (and not `[M0, M0 + M1, ..., M_total]`)

I suspect I'm misinterpreting 2 (looks weird) but that's how I understood the sentence

mzhukova: I see the confusion, I'll double check and get back to you.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: I see you point, it seems like having this information is both unnecessary and confusing, so I'll think about getting rid of block dims fully :)

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+```
+Contiguous memory: [Expert0 | Expert1 | Expert2 | Expert3]
+Block sizes:       [M0 x K  | M1 x K  | M2 x K  | M3 x K ]
+
+Offsets array is for tracking where each expert starts:
+offsets = [0, M0, M0+M1, M0+M1+M2, M0+M1+M2+M3]
+
+Note: Why "sparse"? M_i can be zero if an expert receives no tokens.
+In that case, the next expert starts at the same offset, similar to CSR format row pointers.
+Example: if M1 = 0, then offsets = [0, M0, M0, M0 + M2, M0 + M2 + M3]
```
echeresh: This may be a bit of a stretch :) But I guess it depends on how we view it.

I would say the main property of the grouped memory descriptor is that we split one dimension between groups however there are no implicit zeros in this structure - all data is there, just split dynamically.

Maybe a cleaner option is to introduce a new `memory::format_kind` vs extending `sparse_encoding`. But it would be nice to here from others.

mzhukova: Actually it is perceived in this way in some of the MoE-related articles, you essentially construct a block-diagonal sparse matrix of size `total_M` x `K * num_experts`.
And then it is about how you could represent this thing in oneDNN, which certainly has similarity to CSR, where we describe overall tensor shape, and then resolve individual non-zero elements mappings, but for block-diagonal one doesn't need as much data as for CSR (e.g., no need for column indices for diagonal matrix, so you end up with analogue to row pointers and values itself).

If it is too confusing, I could change it to a new format kind, although it would mean more work on dev side for sure.

Agreed, let's leave this thread open to hear from @uxlfoundation/onednn-arch.

echeresh: > Actually it is perceived in this way in some of the MoE-related articles, you essentially construct a block-diagonal sparse matrix of size total_M x K * num_experts.

I see what you are saying - we can mathematically represent group GEMM as a regular multiplication operation of two block-diagonal matrices. I guess what doesn't look right to me is:

- This block-diagonal sparse GEMM view, even though existing in theory, is not how API/RFC approaches that. The target operation is group GEMM (many groups, share same N/K and per-group M), and not sparse GEMM
- A generic memory descriptor (ideally) should have independent semantics. So if we decide to treat it as a sparse structure, it'd be more natural to define and document it that way - stating that it describes a block-diagonal tensor (which is defined as a "group" tensor now).

mzhukova:
++ @e-ddykim

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+receive many tokens while others receive few or none (distribution could be expected to be very unbalanced).
+The number of tokens per expert (`M` dimension) is only known after routing completes.
+
+(**) Regrouping involves gathering all tokens assigned to each expert
+and creating a contiguous block of memory for each expert's input.
+
+Grouped GEMM processes variable-sized batches (subsets of tokens) per expert,
+which is a key difference from batched GEMM operations that assume uniform batch sizes across groups.
+
+For each expert `i` (dimensions are: `src[Mi x K]`, `weights[K x N]`, `dst[Mi x N]`) we need to compute:
```
PershingSquare, line 24: The inner dimension is not fixed in the backward pass during MoE training.

Consider,

$\mathbf{d\mathbf{W}_i} = \mathbf{S}_i^T \cdot \mathbf{d\mathbf{O}_i}$

With dimensions,

$\[\mathbf{K} \times \mathbf{N}\] = \[ \mathbf{K} \times \mathbf{M}_i \] \cdot \[ \mathbf{M}_i \times \mathbf{N}\]$

I think the grouped gemm interface needs to support heterogeneous $M$, $N$ and $K$, mirroring the PyTorch API.

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+    scale_b,                         # Scaling factors for mat_b
+    scale_recipe_b,                  # Scaling type for mat_b
+    swizzle_a=None,                  # Swizzle pattern for scale_a
+    swizzle_b=None,                  # Swizzle pattern for scale_b
+    bias=None,                       # Optional bias tensor
+    offs=None,                       # Device-side offsets for source tensor
+    output_dtype=torch.bfloat16,     # Output data type
+    contraction_dim=None,            # Which dims are K in matmul
+    use_fast_accum=False             # Nvidia GPUs specifics
+)
```
mgouicem: This API is internal right? Do you know how does it plug to PyTorch external API?
If this gets called only from Inductor pattern matcher, do we have to align to this internal API or is it ok to deviate?

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+
+The grouped memory descriptor is created similarly to CSR descriptors:
+
+```cpp
+static memory::desc grouped(
+    const dims &adims,               // Overall tensor dims [total_M, K]
+    data_type adata_type,            // Elements data type
+    dim group_count,                 // Group size or number of experts (all, not just active ones)
+    const dims &grouped_dims,        // Group dimensions (clarification below)
+    data_type offsets_dt = s32,      // Offset buffer dtype (default int32)
```
mgouicem: Given that this is an address offset, maybe it would be safer to make it s64 ? (this is comming with dropout attribute extension PR).

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Request was originally for variable `M` only, but if we need to generalize it, I believe something along the lines of  @echeresh suggestion should work -- 

```
// Grouped descriptor: assign one dimension as "grouped" and specify the number of parts it is split into.
auto md_grouped = memory::desc::grouped(dims, type, strides, group_dim_idx, group_count);
```

So, essentially from Proposal 1 we drop describing block and just mark required tensor dim to signify that offsets for this dim will be provided during execution. I'll see what else needs updating to allow for this combination and get back to you.

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+int* expert_id;                 // Dynamic size, IDs of experts with tokens
+int* experts_info_start_idx;    // Start offset for each expert's tokens
+int* tokens_lens_per_expert;    // Number of tokens per expert
+float* weight_data;             // [Expert0_weights | Expert1_weights | ...]
+```
+
+- Offsets and sizes have dynamic size based on active ((!) not total) experts (therefore extra array is needed),
+  that is a bit different from the PyTorch/proposed oneDNN design, but compatible.
+- Current version provides offsets on host,
+  but device-side offsets expected for the future GPU-based routing.
```
e-ddykim: This PR (https://github.com/openvinotoolkit/openvino/pull/33073) makes `moe_mask_gen` to run on GPU, and offsets to be on usm_device memory.

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+
+Memory objects are created with 2 buffer pointers:
+
+```cpp
+// Buffers are coming from Framework side (could be device memory)
+float* values;       // total_tokens * K contiguous data
+int32_t* offsets;    // num_experts + 1 offsets
+
+// Create memory object with 2 buffers
+memory src_mem(src_md, eng, {values, offsets});
```
dzarukin, line 214: It's just got to me that with this approach a destination memory object demands two pointers, where `values` would be used as output to store values and `offsets` would be used as input to determine how to distribute those `values` unless I'm missing something. If this is the case, it'll be a unique outstanding behavior besides the fact that pointers for dst memory are considered non-const, thus, may mean that offsets might be changed by the library.

mzhukova: Yeah, that's correct understanding, you could see dst mem creation and usage right below in the "Primitive Execution" section for instance. Offsets are not to be changed by the library, they are filled and passed from the user's side in order to describe/resolve the memory.

Is there any particular issue that you see or is the point just to make sure that we clarify about offsets being input parameter?

dzarukin: The concept of `device_offsets` buffer having a sense of "input" for output memory is misleading and is not common in the industry.

The proper way would be to provide that buffer from the library instead which would truly mean the "output" memory for both buffers. In fact, if there's an abstract extra input parameter for the operation which would change offsets for destination in some way, then it would make more sense to provide those offsets associated with the output data.
And in that sense, input offsets and output offsets should be separate buffers.

However, in the case of interest they are the same, so thinking of something that would prevent of dumping the same buffer that user passed as input is preferred to save on unnecessary wasted bytes.

rfcs/20251203-grouped-gemm-support/README.md
```diff
@@ -0,0 +1,604 @@
...
+```cpp
+auto scales_md = memory::desc({num_experts, N}, dt::f32, tag::ab);
+// Memory: [E0: N scales | E1: N scales | ...]
+```
+
+- Block-wise: grouped descriptor
+```cpp
+// For src [total_tokens x K] with block_size=32 along K
+int block_size = 32;
+auto scales_md = memory::desc::grouped(
```
echeresh: One more observation: these grouped descriptors don't fit well the current API organization.

- `dnnl_primitive_attr_set_{scales,zero_points}` don't accept memory descriptors. Instead, they are specified via masks/dimensions and implicitly assume `abx` layout
- These memory descriptors will be only available at execution time, from memory objects. However, the primitive should have full information about the structure of scales/zero-points at primitive creation time so they are not practically useful

The only important part may be the `offsets` but by design these `offsets` (when present) must be identical to the "A" offsets. So, the question is do we really need to require to pass them for individual scales and zero-points?

It seems like it's all because we lack a "centralized" approach to offset handling in this option. Group GEMM by itself requires structural coupling of its arguments (i.e. group sizes must be identical for participating tensors) but with individual memory descriptors it's unenforceable.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Resolving since API in Proposal 1 got updated/simplified and it addresses this thread.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Resolving since API in Proposal 1 got updated/simplified and it addresses this thread.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Updated API in Proposal 1 to allow for K dim to be variable.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Ping to @uxlfoundation/onednn-arch wrt extending `sparse_encoding` / making it have more generic name in the next major release (also added this note into the RFC).

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Updated! Thanks for the info

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: As per offline discussion, changed to plain memory descriptors and added a bit more info wrt setting some of the commonly requested patterns, please let me if this could be resolved.

dzarukin: The proposed option seems to tie scales to original memory arguments' memory descriptor - thus, for int8 src case if DNNL_ARG_SRC is supplied with a regular `format_kind::blocked`, scales are expected with dimensions {M, K/g} and applied according to existing documentation, but if it is supplied with new `format_kind`, scales are expected with dimensions {total_tokens, K/g} and applied relying on a buffer from DNNL_ARG_SRC exec argument. And that happens with identical settings to the attributs passed to a primitive. The proposed behavior introduces the complexity which was not there before - besides variability in behavior it appends additional runtime cross argument dependency.

mzhukova: @dzarukin Would you prefer having grouped memory descriptor for the concatenated scales buffer and something like an additional `is_grouped` argument to the setter function, or do you have any other recommendation in mind?

(one of the drawbacks of having a grouped descriptor here is that there is no way we could check that `offsets` are compatible, e.g. between src and src scales, except in the kernel itself.. and really there is no indication that we would ever need different offsets for the scales)

rfcs/20251203-grouped-gemm-support/README.md
dzarukin: Based on the discussions it still looks to me that a path of memory descriptor rather creates more disputes than answers to a specific technical ask.

There's a proposal to introduce a new experimental primitive API as an option 4 (or 3, if there's any desire to keep the pointer-based API). Why?
* Straightforward documentation tied to execution flow - takes all main execution arguments and, if present, quantization arguments, apply offsets according to X/Y/Z rules.
* Single offsets execution argument (same as for attributes).
* It doesn't have to deal with limitations coming from conventional API and implementations - no intrusive checks across the board.
* Significantly easier to modify the interface for future needs if any.

It's cleaner and more consolidated and isolated in every aspect with the main downside of implementation price.

rfcs/20251203-grouped-gemm-support/README.md
dzarukin: I shared my proposal in [the other comment](https://github.com/uxlfoundation/oneDNN/pull/4412#discussion_r2692263294).

The drawback seems to be coming from the new descriptor kind and associating two buffers with it while all memory objects are expected to share the same second buffer which unavoidably comes with a verification problem if this buffer is identical for all memory objects.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Thanks for the feedback.

On proposal 3 (pointer arrays), since it was mentioned -- this was mainly considered for potentially cleaner API, but it introduces overhead for recalculating pointers after each routing step, and also seems like current frameworks concatenate tokens into buffers, so this doesn't match the reality..

On a new primitive approach, could you clarify -- are inputs suggested to be raw pointers or memory objects? I would assume memory objects and plain descriptors (like 2D src, 3D weights, 2D dst -- all not grouped), correct?

If so, the operation type (grouped GEMM vs regular matmul) is communicated via primitive name, which is clean, and the remaining piece is communicating individual sizes (== offsets) for matrices in a group, so that we know how to actually perform this operation.

Consideration here: for MoE backward pass, we would need two grouped GEMMs -- 
- grad_input = grad_output * weights^T, so offsets apply to grad_output and grad_input (and therefore it is a similar pattern as forward)
- grad_weights = input^T * grad_output, so offsets are applied to input and grad_output, and here I guess we need to be careful and specify to which arguments offsets should be applied to in order to understand what operation to perform (or we need an implicit assumption, that when 2D tensors are involved, we do this kind of thing "input^T * grad_output" always..?)
Basically that's also what contributed into the offsets as an attribute idea in Proposal 2 (similar to how scales are tied to specific arguments). And then the question would be, why new primitive and not just Proposal 2?

From my side, additional pros of grouped mem (or maybe just reiterating already mentioned ones):
- descriptor can easily accommodate other needs (e.g., padding, strides, max M per expert, maybe even pointers array vs concatenated memory), so that it is all hidden under "grouped memory" abstraction
- grouped memory "modifies" matmul similarly to how CSR does, in my mind it is about how you access elements to multiply and not the fundamentally different operation to perform 
- offsets do describe memory layout (how rows are organized in a buffer), not operation, so mem descriptor seems like a good place for it

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: I wonder if changing grouped mem to -- 
```
auto src_md = memory::desc::grouped(
    {num_experts, total_M, K},  // 3D shape
    dt::f32,
    variable_dim = 1)
```
from --  
```
auto src_md = memory::desc::grouped(
    {total_M, K},  // 2D shape
    dt::f32,
    variable_dim = 0,
    num_experts)
```
would resolve the concern, this way we would have `src` represented as sort of 3D (similar to concatenated `wei`) with extra flag wrt what dimension is variable (`variable_dim=1`), and with number of experts being 0-dim. I suppose this way, setting scales should be more explicit / aligned with current way of doing things.

TaoLv:
@mzhukova  Besides MOE optimization, do we know any other scenario where grouped gemm is used?

mzhukova:
> @mzhukova Besides MOE optimization, do we know any other scenario where grouped gemm is used?

@TaoLv I saw HPC applications mentioned (in the sense that grouped gemm is considered a generalization of batched one), but there were no specifics. No other ML scenarios came up when I was researching this topic..

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: I've added [this](https://github.com/uxlfoundation/oneDNN/pull/4412#discussion_r2709543607) to PR description, maybe after reviewing the implementation and example, we could decide whether the update is needed and I'll update this RFC afterwards.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Included this concern into PR description, we could discuss there and I'll update RFC/impl afterwards if needed.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Closing as resolved for now.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Closing as resolved for now.

rfcs/20251203-grouped-gemm-support/README.md
mzhukova: Sorry for the delayed reply., here is the clarification that I got:
> The offs tensor has value [3,7,7,10], which means mat1[0:3, :] will matmul with mat2[0] and then store to output[0:3, :];  mat1[3:7, :] will matmul with mat2[1] and then store to output[3:7, :];  mat1[7:7, :] will matmul with mat2[2] and then store to output[7:7, :](this means skip expert 2);  mat1[7:10, :] will matmul with mat2[3] and then store to output[7:10, :].

So I dropped `0` at the beginning, and I believe the rest is correct. You could also see comment in [here](https://jira.devtools.intel.com/browse/MFDNN-14187?focusedId=28592416&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28592416).

mzhukova:
@echeresh and/or @e-ddykim could you please approve, so that I could merge the current RFC?
feature is available in the main under experimental option and all future work (and changes if any) will be covered by a separate RFC if needed

PR #4413, opened 2025-12-04
test: update system memory test(s)

narenbagria:
This PR updates the system memory test(s). Now the OCL test will conditionally throw an exception based on device capability, similar to the SYCL test.

### DIFF
--- a/tests/gtests/ocl/api/test_memory_usm.cpp
+++ b/tests/gtests/ocl/api/test_memory_usm.cpp
@@ -24,6 +24,7 @@
 
 #include <cstdint>
 #include <vector>
+#include <CL/cl_ext.h>
 
 using namespace dnnl::impl::xpu::ocl;
 
@@ -204,8 +205,26 @@ TEST(ocl_memory_usm_test, ErrorMakeMemoryUsingSystemMemory) {
     memory::desc mem_d({n}, memory::data_type::f32, memory::format_tag::x);
 
     std::vector<float> system_buf(n);
+    bool system_memory_supported = false;
 
-    if (eng.get()->mayiuse_system_memory_allocators()) {
+#ifdef cl_intel_unified_shared_memory
+    cl_device_id device = ocl_interop::get_device(eng);
+    cl_device_unified_shared_memory_capabilities_intel
+            system_memory_capabilities_intel
+            = 0;
+
+    ASSERT_EQ(
+            clGetDeviceInfo(device,
+                    CL_DEVICE_SHARED_SYSTEM_MEM_CAPABILITIES_INTEL,
+                    sizeof(cl_device_unified_shared_memory_capabilities_intel),
+                    &system_memory_capabilities_intel, nullptr),
+            CL_SUCCESS);
+
+    system_memory_supported = system_memory_capabilities_intel
+            & CL_UNIFIED_SHARED_MEMORY_ACCESS_INTEL;
+#endif
+
+    if (system_memory_supported) {
         EXPECT_NO_THROW(
                 memory mem = ocl_interop::make_memory(mem_d, eng,
                         ocl_interop::memory_kind::usm, system_buf.data()));


narenbagria:
https://jira.devtools.intel.com/browse/NEO-16889
https://gtax-postsi-fm.intel.com/#/jobs/87809820/task_results/1614035944

With shared system USM / system memory support enabled: NEOReadDebugKeys=1 **EnableSharedSystemUsmSupport=1**
 ./test_api_ocl -engine=gpu --gtest_filter=ocl_memory_usm_test.ErrorMakeMemoryUsingSystemMemory
 We observed that the above OCL test fails. After comparing it with its SYCL counterpart, we believe that the OCL test should follow a similar approach.

narenbagria:
SYCL test: EXPECT the test to throw an exception **only if** the device **doesn't** have shared system USM capabilities:
https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/sycl/api/test_memory_usm.cpp#L223-L230
```cpp
    if (device.has(::sycl::aspect::usm_system_allocations)) {
        memory mem = sycl_interop::make_memory(
                mem_d, eng, sycl_interop::memory_kind::usm, system_buf.data());
    } else {
        EXPECT_THROW(memory mem = sycl_interop::make_memory(mem_d, eng,
                             sycl_interop::memory_kind::usm, system_buf.data()),
                dnnl::error);
    }
```

While OCL test would always throw an exception:
https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/ocl/api/test_memory_usm.cpp#L207-L209
```cpp
    std::vector<float> system_buf(n);
    EXPECT_THROW(memory mem = ocl_interop::make_memory(mem_d, eng,
                         ocl_interop::memory_kind::usm, system_buf.data()),
            dnnl::error);
```

This PR updates the test logic to throw an exception **only if** device **doesn't** support shared system USM capabilities:
https://github.com/uxlfoundation/oneDNN/blob/0edad49b49a0aa6551ee9bd76c1a0724939df4e4/tests/gtests/ocl/api/test_memory_usm.cpp#L221-L230

narenbagria:
Hi @vpirogov, @dzarukin, @umar456 , please share your feedback/review. Thank you!

tests/gtests/ocl/api/test_memory_usm.cpp
```diff
@@ -204,9 +206,28 @@ TEST(ocl_memory_usm_test, ErrorMakeMemoryUsingSystemMemory) {
...
-                         ocl_interop::memory_kind::usm, system_buf.data()),
-            dnnl::error);
+
+    cl_device_id device = ocl_interop::get_device(eng);
+    cl_device_unified_shared_memory_capabilities_intel
+            system_memory_capabilities_intel
+            = 0;
+    EXPECT_EQ(
+            clGetDeviceInfo(device,
+                    CL_DEVICE_SHARED_SYSTEM_MEM_CAPABILITIES_INTEL,
```
umar456: Because this is an intel specific extension of OpenCL you probably want to limit this check when the cl_intel_unified_shared_memory extension is supported.

tests/gtests/ocl/api/test_memory_usm.cpp
```diff
@@ -204,9 +206,28 @@ TEST(ocl_memory_usm_test, ErrorMakeMemoryUsingSystemMemory) {
...
+                    sizeof(cl_device_unified_shared_memory_capabilities_intel),
+                    &system_memory_capabilities_intel, nullptr),
+            CL_SUCCESS);
+
+    if (system_memory_capabilities_intel
+            & CL_UNIFIED_SHARED_MEMORY_ACCESS_INTEL) {
+        EXPECT_NO_THROW(
```
umar456, line 228: Because this is a test that expects an error. I think it would be better to skip this test using GTEST_SKIP() with a message instead of letting it pass.

A separate test can be made to ensure that the make_memory function passes with system memory to ensure it is doing what we expect and to ensure that you can access/modify it from both the host and device like we do with the tests here: https://github.com/uxlfoundation/oneDNN/pull/4413/files#diff-55acc8a8523e17b86920641fff0b562c022538143ff4a8752a6388484b483730R308

dzarukin: Well, test for SYCL runtime is doing same, I don't see a compelling reason for OCL test do it differently. If we want to distinguish those two different situations in the future, we can do that, but it better be done in a different PR.

tests/gtests/ocl/api/test_memory_usm.cpp
dzarukin: If the call doesn't explode or doesn't return negative status if the extension is not present, it should be fine to keep this version. The question is if this query is aligned with the oneDNN supported OpenCL version meaning the oldest oneDNN supported OpenCL API has these identifiers and types.

umar456: I was thinking that the extension(function and enum declaration) wouldn't be available on other platforms so we would run into compilation errors. I am not sure if we are targeting a specific version of the OpenCL library.

vpirogov: Extensions are carefully managed by OpenCL standard. See https://github.com/KhronosGroup/OpenCL-Headers/blob/6137cfbbc7938cd43069d45c622022572fb87113/CL/cl_ext.h#L3205

narenbagria: Hi @umar456, @dzarukin, @vpirogov, 
Thanks for pointing out, yes, without proper compile time guards, for older OCL impls we might run into issues.
I've updated the test to use engine->mayiuse_system_memory_allocators() directly instead of querying the device as the compute logic already has proper #ifdef guards and we can avoid code duplication by directly using this
https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/ocl/device_info.cpp#L167-L178
```cpp
#ifdef cl_intel_unified_shared_memory
    cl_device_unified_shared_memory_capabilities_intel
            system_memory_capabilities_intel
            = 0;
    err = clGetDeviceInfo(device,
            CL_DEVICE_SHARED_SYSTEM_MEM_CAPABILITIES_INTEL,
            sizeof(cl_device_unified_shared_memory_capabilities_intel),
            &system_memory_capabilities_intel, nullptr);
    OCL_CHECK(err);
    mayiuse_system_memory_allocators_ = system_memory_capabilities_intel
            & CL_UNIFIED_SHARED_MEMORY_ACCESS_INTEL;
#endif
````
Please let me know if we would like to switch back to older style rather than relying on mayiuse_system_memory_allocators

Thank you!

**Comment by narenbagria** [comment]
Updated the PR. Thanks!

File: tests/gtests/ocl/api/test_memory_usm.cpp (hunk 1 above)
Comment (narenbagria) [reply tests/gtests/ocl/api/test_memory_usm.cpp]: Hi All, 

commit 2: fixes the commit 1 by adding the missing ifdef guard
https://github.com/uxlfoundation/oneDNN/pull/4413/commits/b53e2c7ead4bd488496449c1995eea483c69fb4e

commit 3/latest: removes the identifier logic from the test altogether to use  eng->mayiuse_system_memory_allocators() instead, relying on the existing capability detection in device_info.cpp
https://github.com/uxlfoundation/oneDNN/pull/4413/commits/e2ac9b53b48a00fabe3c9ba407b9738544079140

Please review and let me know whether we prefer to keep commit 3 or revert it and proceed with commit 2 approach.

Thank you!

File: tests/gtests/ocl/api/test_memory_usm.cpp [hunk 3]
```diff
@@ -204,9 +204,16 @@ TEST(ocl_memory_usm_test, ErrorMakeMemoryUsingSystemMemory) {
     memory::desc mem_d({n}, memory::data_type::f32, memory::format_tag::x);
 
     std::vector<float> system_buf(n);
-    EXPECT_THROW(memory mem = ocl_interop::make_memory(mem_d, eng,
-                         ocl_interop::memory_kind::usm, system_buf.data()),
-            dnnl::error);
+
+    if (eng.get()->mayiuse_system_memory_allocators()) {
```
dzarukin: This is not a public symbol, it can't be used.

narenbagria: @dzarukin , updated by reverting commit 3, please review.

narenbagria: https://github.com/uxlfoundation/oneDNN/pull/4413/commits/51e9be81873c1928ca89afa3824d45d220bc3252

narenbagria:
Hi @umar456, @dzarukin, @vpirogov, please review/approve. Thanks!

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu

PR #4418, opened 2025-12-04
only support finite p in lp norm

mzhukova:
Small change to address [MFDNN-4095](https://jira.devtools.intel.com/browse/MFDNN-4095), that is to disallow `p = inf` for now.

### DIFF
--- a/tests/gtests/test_reduction.cpp
+++ b/tests/gtests/test_reduction.cpp
@@ -17,6 +17,8 @@
 #include "dnnl_test_common.hpp"
 #include "gtest/gtest.h"
 
+#include <limits>
+
 #include "oneapi/dnnl/dnnl.hpp"
 
 namespace dnnl {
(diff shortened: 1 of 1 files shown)


mzhukova:
make test
disable benchdnn_all

tests/gtests/test_reduction.cpp
```diff
@@ -153,6 +155,26 @@ static auto expected_failures = []() {
...
+            // not supported p = infinity for norm_lp_power_p_max
+            reduction_test_params_t {tag::nchw, tag::nchw,
+                    algorithm::reduction_norm_lp_power_p_max,
+                    std::numeric_limits<float>::infinity(), 0.0f, {1, 8, 4, 4},
+                    {1, 8, 4, 4}, true, dnnl_invalid_arguments},
+            // not supported p = infinity for norm_lp_power_p_sum
+            reduction_test_params_t {tag::nchw, tag::nchw,
+                    algorithm::reduction_norm_lp_power_p_sum,
+                    std::numeric_limits<float>::infinity(), 0.0f, {1, 8, 4, 4},
+                    {1, 8, 4, 4}, true, dnnl_invalid_arguments},
```
avmanerikar, line 177: (nit): Do we need a test to check for p < 1.0 too?

mzhukova: there is one above that covers it --

```
        reduction_test_params_t {tag::nchw, tag::nchw,
                    algorithm::reduction_norm_lp_max, ->0.5f<-, 0.0f, {1, 8, 4, 4},
                    {1, 8, 4, 4}, true, dnnl_invalid_arguments},
```

avmanerikar: It only covers for the `reduction_norm_lp_max` algorithm - but that should be sufficient I guess.

PR #4420, opened 2025-12-04
all: update clang-format to v18 and apply changes

dzarukin:
PR updates a clang-format binary in automation checks from v11 to v18 and applies all the changes it invokes.

Two motivation points:
1) Decrease the pressure on the community by searching for an obsolete clang-format binary (as they are not entirely compatible starting from clang-format-14).
2) Introduce `LambdaBodyIndentation : OuterScope` to align lambda code on a fixed number of spaces and not move it every time a lambda declaration changes (like number of argument increases, etc.). This introduction will happen in a separate PR.

Couple notes on changes that might be considered as "bad":
1) It updated the way to count indentation in conditions. Used to start counting from `?` and `:`, now it from the following statement, or just adds two more spaces.
2) there are some updates to pure virtual functions and default template values, it's because they addressed the issue with `BreakBeforeBinaryOperators` which didn't consider those cases as relevant, now it does. It's possible to update it back by setting a value of `BreakBeforeBinaryOperators` from `All` to `NonAssignment`, meaning that all `=` changes will stop breaking on a new line, but that to be discussed and applied as a separate change if agreed on.
3) There're some changes to the macro bodies which declare/define functions/methods when the type is not considered as `const T &obj`  but instead as a binary operator `const T & obj`. I was not able to fight it back. If somebody has some time and expertise to do that, be my guest.
4) Some
```
if (c) statement
else {
    statement
}
```
were converted into
```
if (c) statement
else { statement }
```
which might be addressed in the future version with the current config values, or can be adjusted by changing the code (adding braces for `if`) or by changing settings (no short statements) but the latter has a big impact on the codebase.

### DIFF
--- a/.github/workflows/pr-linter.yml
+++ b/.github/workflows/pr-linter.yml
@@ -55,14 +55,14 @@ jobs:
 
   pr-clang-format:
     name: Clang-Format
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0
       - name: Install clang-format
-        run: sudo apt update && sudo apt install -y "clang-format-11"
+        run: sudo apt update && sudo apt install -y "clang-format-18"
       - name: Check code formatting
         run: .github/automation/clang-format.sh ${{ github.event.pull_request.base.sha }}
 
--- a/src/common/engine.hpp
+++ b/src/common/engine.hpp
@@ -113,23 +113,27 @@ struct dnnl_engine : public dnnl::impl::c_compatible {
      * a NULL-terminated list */
     virtual const dnnl::impl::impl_list_item_t *get_reorder_implementation_list(
             const dnnl::impl::memory_desc_t *src_md,
-            const dnnl::impl::memory_desc_t *dst_md) const = 0;
+            const dnnl::impl::memory_desc_t *dst_md) const
+            = 0;
 
     /** return the list of concat implementations. engine guarantees to return
      * a NULL-terminated list */
     virtual const dnnl::impl::impl_list_item_t *
-    get_concat_implementation_list() const = 0;
+    get_concat_implementation_list() const
+            = 0;
 
     /** return the list of sum implementations. engine guarantees to return
      * a NULL-terminated list */
     virtual const dnnl::impl::impl_list_item_t *
-    get_sum_implementation_list() const = 0;
+    get_sum_implementation_list() const
+            = 0;
 
     /** return the list of implementations for a given descriptor.
      * engine guarantees to return a NULL-terminated list */
 
     virtual const dnnl::impl::impl_list_item_t *get_implementation_list(
-            const dnnl::impl::op_desc_t *desc) const = 0;
+            const dnnl::impl::op_desc_t *desc) const
+            = 0;
 
     virtual dnnl::impl::status_t serialize_device(
             dnnl::impl::serialization_stream_t &sstream) const {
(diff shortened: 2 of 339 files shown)


dzarukin:
make test

src/gpu/intel/gemm/jit/dsl/ir/core.cpp
```diff
@@ -295,7 +295,9 @@ stmt_t stmt_seq_t::make(const std::vector<stmt_t> &_vec) {
 }
 
 #define DECL_TRAVERSE_LEAF(name) \
-    object_t ir_mutator_t::_mutate(const name &obj) { return obj; } \
+    object_t ir_mutator_t::_mutate(const name &obj) { \
+        return obj; \
+    } \
```
rjoursler, line 300: Comment: I generally prefer formatting which avoids line breaks when there isn't a column overflow or some other beneficial alignment condition. As such, I prefer the more condensed previous version. A couple of consequences of this:
* My preference is to switch BreakBeforeBinaryOperators from All to NonAssignment
* I would be fine with just switching AllowShortFunctionsOnASingleLine to All as well.

Beyond that, the formatting changes generally look good to me.

rjoursler (approved):

src/gpu/intel/gemm/jit/dsl/ir/core.cpp
dzarukin: Noted, thanks for your opinion.

Sqvid:
Thanks @dzarukin, all good from our end but I'm not sure if this needs an update (though it was not version 11 anyway)

https://github.com/uxlfoundation/oneDNN/blob/69d50b08008b89a4d8db8f8a27898cceb90ae5e0/.github/azure/ci-x64.yml#L78

vpirogov:
> I'm not sure if this needs an update (though it was not version 11 anyway)
> 
> https://github.com/uxlfoundation/oneDNN/blob/69d50b08008b89a4d8db8f8a27898cceb90ae5e0/.github/azure/ci-x64.yml#L78

This one is defining clang version for x64 CI in Azure. It does not affect tidy or clang-format.

vpirogov (approved):

.github/workflows/pr-linter.yml
```diff
@@ -55,14 +55,14 @@ jobs:
 
   pr-clang-format:
     name: Clang-Format
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
spalicki, line 58: latest is 24.04 (latest LTS) or 25.10? It would be better to just specify the version number.

src/common/engine.hpp
```diff
@@ -113,23 +113,27 @@ struct dnnl_engine : public dnnl::impl::c_compatible {
      * a NULL-terminated list */
     virtual const dnnl::impl::impl_list_item_t *get_reorder_implementation_list(
             const dnnl::impl::memory_desc_t *src_md,
-            const dnnl::impl::memory_desc_t *dst_md) const = 0;
+            const dnnl::impl::memory_desc_t *dst_md) const
+            = 0;
```
spalicki, line 117: Yes, I agree new line break just for `= 0;` is bad.

.github/workflows/pr-linter.yml
dzarukin: That's what is used universally across the script so I sticked to the same value. The version was there to let clang-format-11 be downloaded as they dropped its support from latest distributions.

spalicki: In both cases (LTS or not) latest will automatically change to 26.04 in 5 months (approximately), which might cause some compatibility issues. I think it is better to stick to numbered version like 24.04 (in all scripts), so the underlying OS does not magically change and break something. I can create a dedicated PR for that, and we can move the discussion there.

dzarukin: It would be great to handle it separately, thanks.

PR #4421, opened 2025-12-04
cpu: x64: matmul: improve perf for shapes w/small K for f32 matmul

xuxinzen:
Partially fixes MFDNN-13919

Using large offset in writing data to memory causes latency in performance, adding `prefetchw` helps to close the perf gap for the shapes with large N. With this patch, most of the perf drop can be fixed except for some shapes(K == 32) according to customer support team. The perf gap is less than ~10% for the whole model.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -1846,7 +1846,7 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
             && is_superset(bgmmc.isa, avx512_core)) {
         // Dispatch the shapes with small K to gemm for better performance
         // The heuristic values are empirical
-        const bool small_K = bgmmc.N <= 14528
+        const bool small_K = bgmmc.N <= 14528 && bgmmc.K == 32
                 && ((bgmmc.M <= 768 && bgmmc.K <= 128)
                         || bgmmc.K * bgmmc.M <= 49152);
         // We need to exclude certain shapes as brgemm matmul performs better


xuxinzen:
make test

ankalinin (approved):

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -866,7 +866,10 @@ float compute_blocking_heuristic_avx512(brgemm_matmul_conf_t &bgmmc,
         matmul_avx512_blocking_params_t &best_blocking) {
     const int nthr = bgmmc.nthr;
 
-    const int max_m_blk = nstl::min(256, matmul.M);
+    const bool need_large_m_blk = bgmmc.ndims == 2 && bgmmc.N <= 14528
+            && ((bgmmc.M <= 768 && bgmmc.K <= 128)
+                    || bgmmc.K * bgmmc.M <= 49152);
+    const int max_m_blk = nstl::min(need_large_m_blk ? 512 : 256, matmul.M);
```
densamoilov, line 873: This new condition will affect all data types, not only fp32. Are we sure that it won't introduce regressions for non-fp32 data types?

xuxinzen: The upper condition limits this heuristic only apply to [f32](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/x64/matmul/brgemm_matmul_utils.cpp#L1841).

xuxinzen: Updated the limitation on `compute_blocking_heuristic_avx512`

densamoilov (approved):

PR #3770, opened 2025-08-14
Update nGEN

echeresh:
PR updates nGEN to fully align with the upstream repository.

For future upstreaming, I created `scripts/update_ngen.sh` to automate this process to keep manual interaction at minimum.

PR #4423, opened 2025-12-05
third_party: ngen: restore PVC WAR bug workaround

echeresh:
Jira: [MFDNN-14449](https://jira.devtools.intel.com/browse/MFDNN-14449)
Introduced with nGEN update in https://github.com/uxlfoundation/oneDNN/pull/3770.
Thanks to @petercad for help with investigation.


echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_matmul
enable arch_gpu_xe-hpc

rjoursler:
Since we have hit this same bug many times, can we add a PVC specific test to try and catch nGEN regressions here?

echeresh:
> Since we have hit this same bug many times, can we add a PVC specific test to try and catch nGEN regressions here?

Good point, added a regression test here: #4429
As discussed, it's an ad-hoc/convolution-specific test for now. Will think about an nGEN-specific test in the future if needed.

PR #4424, opened 2025-12-05
gpu: intel: ocl: add a size guard to stateless addressing flag

spalicki:
Adding guards for stateless addressing model:
1. OpenCL GPU runtime only since SYCL does not allow allocating buffers bigger than 4GB
2. Primitive uses buffers bigger than 4 GB
3. Device must support 64bit addresses

Also, readding a check for max allowed allocation on SYCL runtime, since it leads to some tests failing instead of being skipped.

Fixes: [MFDNN-14454](https://jira.devtools.intel.com/browse/MFDNN-14454) and [MFDNN-14459](https://jira.devtools.intel.com/browse/MFDNN-14459).

### DIFF
--- a/src/common/primitive_desc.hpp
+++ b/src/common/primitive_desc.hpp
@@ -260,12 +260,12 @@ struct primitive_desc_t : public c_compatible {
     virtual const memory_desc_t *invariant_bia_md() const {
         return invariant_wei_md(1);
     }
-    virtual const memory_desc_t *invariant_dst_md() const {
+    virtual const memory_desc_t *invariant_dst_md(int index = 0) const {
         const auto prop_kind = get_prop_kind();
         return utils::one_of(prop_kind, prop_kind::backward_data,
                        prop_kind::backward_weights, prop_kind::backward)
-                ? diff_dst_md()
-                : dst_md();
+                ? diff_dst_md(index)
+                : dst_md(index);
     }
 
     virtual format_kind_t invariant_src_user_format_kind(int index = 0) const {
@@ -470,6 +470,28 @@ struct primitive_desc_t : public c_compatible {
 
     int pd_iterator_offset() const { return pd_iterator_offset_; }
     int skip_idx() const { return skip_idx_; }
+
+    bool has_large_buffers() const {
+        auto is_large = [](const memory_desc_t *md) {
+            return memory_desc_wrapper(md).size() > UINT32_MAX;
+        };
+
+        for (int i = 0; i < n_inputs(); i++) {
+            if (is_large(invariant_src_md(i))) return true;
+        }
+        if (is_large(invariant_wei_md())) return true;
+        for (int i = 0; i < n_outputs(); i++) {
+            if (is_large(invariant_dst_md(i))) return true;
+        }
+        const auto &post_ops = attr()->post_ops_;
+        for (int i = 0; i < post_ops.len(); i++) {
+            const auto &e = post_ops.entry_[i];
+            if (e.is_binary()) {
+                if (is_large(&(e.binary.src1_desc))) return true;
+            }
+        }
+        return false;
+    }
 
 protected:
     primitive_attr_t attr_;
--- a/src/gpu/intel/compute/kernel_ctx.hpp
+++ b/src/gpu/intel/compute/kernel_ctx.hpp
@@ -85,6 +85,13 @@ public:
     // operations when native 64-bit operations are unsupported.
     void use_int32_offset(bool value) { use_int32_offset_ = value; }
 
+    void require_stateless_addressing(bool value) {
+        require_stateless_addressing_ = value;
+    }
+    bool require_stateless_addressing() const {
+        return require_stateless_addressing_;
+    }
+
     void define_int(const char *variable, int64_t value) {
         set_macro(variable, value, int_var_map_);
     }
@@ -157,8 +164,9 @@ public:
     bool has_custom_headers() const { return !custom_headers_.empty(); }
 
 private:
-    void register_buffer_size(dim_t nelems) {
-        if (nelems > INT_MAX) use_int32_offset(false);
+    void register_buffer_size(dim_t nelems, size_t size) {
+        if (nelems > INT32_MAX) use_int32_offset(false);
+        if (size > UINT32_MAX) require_stateless_addressing(true);
     }
 
     void set_default_options(const primitive_attr_t *attr) {
@@ -208,6 +216,7 @@ private:
     std::set<std::string> option_set_;
     std::unordered_map<std::string, std::string> custom_headers_;
     bool use_int32_offset_ = true;
+    bool require_stateless_addressing_ = false;
 };
 
 } // namespace compute
--- a/src/gpu/intel/rnn/config.hpp
+++ b/src/gpu/intel/rnn/config.hpp
@@ -181,6 +181,8 @@ struct ocl_conf_t {
     bool copy_diff_dst_layer = false;
     bool copy_diff_src_layer;
     bool deterministic = false;
+    bool require_stateless_addressing = true;
+    uint8_t pad2[7] = {};
     struct comp_conf_t {
         bool is_enabled = false;
         bool compute_gemm_layer = false;
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -1102,21 +1102,17 @@ int get_gpu_ram_sizes(size_t &ram_size, size_t &max_alloc_size) {
 
 #if DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
     auto eng = dnnl::engine(get_test_engine(), true);
-    cl_int status = CL_SUCCESS;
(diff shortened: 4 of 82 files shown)


spalicki:
make test perf-gpu

spalicki:
make test perf-gpu

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
set test_scope=NIGHTLY

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
set test_scope=NIGHTLY

spalicki:
make test perf-gpu

spalicki:
make test perf-gpu

spalicki:
make test perf-gpu

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
set test_scope=NIGHTLY

atkassen:
make test perf-gpu

src/common/primitive_desc.hpp
```diff
@@ -260,12 +260,12 @@ struct primitive_desc_t : public c_compatible {
     virtual const memory_desc_t *invariant_bia_md() const {
         return invariant_wei_md(1);
     }
-    virtual const memory_desc_t *invariant_dst_md() const {
+    virtual const memory_desc_t *invariant_dst_md(int index = 0) const {
```
dzarukin, line 263: `invariant_dst_md` also presents in other files inside CPU folder, their declarations and bodies must be updated as well.

src/common/primitive_desc.hpp
```diff
@@ -471,6 +471,28 @@ struct primitive_desc_t : public c_compatible {
     int pd_iterator_offset() const { return pd_iterator_offset_; }
     int skip_idx() const { return skip_idx_; }
 
+    bool has_large_buffers() const {
```
dzarukin, line 474: If the `true`/`false` is only thing you need, I'd just return once `is_large` returns `true` to skip unnecessary checks for further tensors.

src/common/primitive_desc.hpp
```diff
@@ -471,6 +471,28 @@ struct primitive_desc_t : public c_compatible {
...
+        uint32_t num_large_buffers = 0;
+
+        auto is_large = [](const memory_desc_t *md) {
+            return static_cast<uint64_t>(memory_desc_wrapper(md).size())
```
dzarukin: `size()` returns `size_t` which is basically `uint64_t`. Why is this cast needed?

src/common/primitive_desc.hpp
```diff
@@ -471,6 +471,28 @@ struct primitive_desc_t : public c_compatible {
...
+                    > UINT32_MAX;
+        };
+
+        for (int i = 0; i < n_inputs(); i++)
+            num_large_buffers += is_large(invariant_src_md(i));
+        num_large_buffers += is_large(invariant_wei_md());
```
dzarukin: Input may refer to weights also.
```suggestion
        for (int i = 0; i < n_inputs(); i++)
            num_large_buffers += is_large(invariant_src_md(i));
            num_large_buffers += is_large(invariant_wei_md(i));
        }
```

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -85,6 +85,13 @@ class kernel_ctx_t {
     // operations when native 64-bit operations are unsupported.
     void use_int32_offset(bool value) { use_int32_offset_ = value; }
 
+    void require_stateless_addressing(bool value) {
+        require_stateless_addressing_ = value;
+    }
+    bool require_stateless_addressing() const {
+        return require_stateless_addressing_;
+    }
```
dzarukin, line 93: This is a more traditional convention.
```suggestion
    void set_stateless_addressing(bool value) {
        stateless_addressing_ = value;
    }
    bool get_stateless_addressing() const {
        return stateless_addressing_;
    }
```

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -208,6 +216,7 @@ class kernel_ctx_t {
     std::set<std::string> option_set_;
     std::unordered_map<std::string, std::string> custom_headers_;
     bool use_int32_offset_ = true;
+    bool require_stateless_addressing_ = false;
```
dzarukin, line 219: Minor: I'd leave a comment what this switch is responsible for and what are implications.

src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -157,8 +164,9 @@ class kernel_ctx_t {
     bool has_custom_headers() const { return !custom_headers_.empty(); }
 
 private:
-    void register_buffer_size(dim_t nelems) {
-        if (nelems > INT_MAX) use_int32_offset(false);
+    void register_buffer_size(dim_t nelems, uint64_t size) {
```
dzarukin: Can `size_t` be used here instead?

src/gpu/intel/rnn/config.hpp
```diff
@@ -181,6 +181,8 @@ struct ocl_conf_t {
     bool copy_diff_dst_layer = false;
     bool copy_diff_src_layer;
     bool deterministic = false;
+    bool require_stateless_addressing = true;
+    uint8_t pad2[7] = {};
```
dzarukin, line 185: I've seen some places used `3` instead. What does it depend on?

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1096,30 +1096,36 @@ size_t get_cpu_ram_size() {
 }
 #endif
 
-int get_gpu_ram_size(size_t &ram_size) {
+int get_gpu_ram_sizes(size_t &ram_size, size_t &max_alloc_size) {
```
dzarukin, line 1099: Nit: would be great if you could use a separate revert commit for the check so it's visible if there's anything added on top of it.

src/gpu/intel/rnn/config.hpp
atkassen: We pad as necessary so the total size of the members matches the size of the struct. In this case, the alignment of the struct is 8B, so adding the extra `bool` adds 1B to the total size of the members, but 8B to the size of the struct, so pad with an additional 7B. Other structs may be 4B aligned, so the padding may be different in that case, or we have a different number/kind of sub-alignment-sized types, etc.

dzarukin: Thanks for clarifications!

rjoursler: To add a little bit more, this padding is so we can use [serilization_stream_t::append](https://github.com/uxlfoundation/oneDNN/blob/22bdd09d01c844e24b39bb1e9dc956e40dbed080/src/common/serialization.hpp#L96-L106), which allows us to avoid maintaining hashing and equality operator required for kernel cache keys.

rjoursler (approved):

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: The comment is in engine.cpp where this is used, but I can copy it here.

src/gpu/intel/compute/kernel_ctx.hpp
spalicki: Yes, it just did not make sense from language point of view. I assumed `require_stateless_addressing` is more descriptive of what we are trying to get/set.

src/common/primitive_desc.hpp
spalicki: But I do not think that both are of `n_inputs()` size, weights(0) is usually, well weights, while weights(1) is bias (if used).

src/common/primitive_desc.hpp
spalicki: macOS and aarch64 compiler complained when there was no explicit cast, but I can remove it if just checking for a bool.

src/common/primitive_desc.hpp
spalicki: That is why I added default 0, so all old usage is the same.

tests/benchdnn/dnnl_common.cpp
spalicki: Previous commit was a bit bigger so I would be reverting and then readding a lot of code.

dzarukin: For a full scope - sure. Partial revert of specifically benchdnn changes would do.

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

tests/benchdnn/dnnl_common.cpp
spalicki: Done.

dzarukin: Thank you!

src/common/primitive_desc.hpp
dzarukin: Got it, thanks for explaining.

dzarukin (approved):

PR #4427, opened 2025-12-05
ukernel: use thread-local, cache-line aligned buffers in execute functions

ankalinin:
Optimize ukernel multithreaded performance through cache-line aligned memory management:

- Introduction of a thread-local, cache-line aligned buffer for batch elements, which replaces the previous use of a standard vector. This helps prevent cache line sharing across threads and reduces memory allocation overhead.
- Documentation is also updated to reflect the importance of input buffers memory alignment for performance.

It is obvious that the performance improvement is most noticeable for small sizes. For 1-core case it is 4-5%. For very small sizes up to 10%
For larger sizes, the overhead from the execute functions code is negligible. I have not data for multi-thread cases

### DIFF
--- a/src/cpu/x64/ukernel/brgemm.cpp
+++ b/src/cpu/x64/ukernel/brgemm.cpp
@@ -152,8 +152,23 @@ status_t brgemm_t::get_B_pack_type(
     return status::success;
 }
 
+namespace {
+// Cache line size used for alignment
+constexpr size_t k_cache_line = 64;
+} // namespace
+
 size_t brgemm_t::get_scratchpad_size() const {
-    return brgemm_desc_.get_wsp_buffer_size();
+    const size_t wsp_size = brgemm_desc_.get_wsp_buffer_size();
+    // Align workspace end to cache line
+    const size_t wsp_size_aligned = utils::rnd_up(wsp_size, k_cache_line);
+
+    const size_t batch_element_size
+            = brgemm_desc_.brgattr.max_bs * sizeof(brgemm_batch_element_t);
+    // Align batch element size to cache line
+    const size_t batch_element_size_aligned
+            = utils::rnd_up(batch_element_size, k_cache_line);
+
+    return wsp_size_aligned + batch_element_size_aligned;
 }
 
 bool brgemm_t::is_execute_postops_valid() const {
@@ -186,8 +201,19 @@ status_t brgemm_t::generate() {
 
 status_t brgemm_t::execute(const void *A_ptr, const void *B_ptr,
         const dim_t *A_B_offsets, void *C_ptr, void *scratchpad_ptr) const {
+
+    if (reinterpret_cast<uintptr_t>(scratchpad_ptr) % k_cache_line != 0
+            && get_verbose(verbose_t::exec_profile, component_t::ukernel))
+        VWARN(primitive, ukernel, "Scratchpad is not cache-line aligned");
+
     const auto batch_size = brgemm_desc_.brgattr.max_bs;
-    std::vector<brgemm_batch_element_t> v_batch_element(batch_size);
+
+    // Batch elements at the end of aligned workspace
+    const size_t wsp_size = brgemm_desc_.get_wsp_buffer_size();
+    const size_t wsp_size_aligned = utils::rnd_up(wsp_size, k_cache_line);
+    auto *v_batch_element = reinterpret_cast<brgemm_batch_element_t *>(
+            reinterpret_cast<char *>(scratchpad_ptr) + wsp_size_aligned);
+
     for (int i = 0; i < batch_size; i++) {
         v_batch_element[i].offset.A = A_B_offsets[2 * i];
         v_batch_element[i].offset.B = A_B_offsets[2 * i + 1];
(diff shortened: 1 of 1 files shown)


ankalinin:
make test
set test_scope=NIGHTLY

disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022

disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

disable benchdnn_all
enable benchdnn_brgemm
enable benchdnn_conv
enable benchdnn_matmul
enable benchdnn_rnn

ankalinin:
Pre-commit testing with various compilers is [here](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f0d2d405-c520-f1e5-9bdd-a4bf010d0e2d)

src/cpu/x64/ukernel/brgemm.cpp
```diff
@@ -184,10 +184,46 @@ status_t brgemm_t::generate() {
...
+private:
+    static constexpr size_t cache_line_size_ = 64;
+    void *ptr_ = nullptr;
+    size_t capacity_ = 0;
+};
+
 status_t brgemm_t::execute(const void *A_ptr, const void *B_ptr,
         const dim_t *A_B_offsets, void *C_ptr, void *scratchpad_ptr) const {
     const auto batch_size = brgemm_desc_.brgattr.max_bs;
-    std::vector<brgemm_batch_element_t> v_batch_element(batch_size);
```
dzarukin, line 190: It's great that you identified a bad spot to improve ukernel performance though it seems to me doing that memory management stuff for the user might not be the best option.

I'd probably explore the option with scratchpad memory increase - the size of `brgemm_batch_element_t` is known, batch_size is known at create time - thus, we can ask for extra scratchpad memory including proper alignment and cut a piece here and pass the tail for the kernel itself.

This way the user would control the memory in the application-compliant way and would have the same effect as `batch_element_buffer_t` provides.
Could you, please, check if the proposed option delivers identical performance improvement?

ankalinin: I'm not sure we can use a scratchpad allocated at creation time. Generally, the execution function of the same brgemm_t object can be called from different threads simultaneously.
Update in [src/cpu/x64/ukernel/brgemm.cpp](https://github.com/uxlfoundation/oneDNN/pull/4427/files/485f3bf08d5a51fa50db35d2d28dd640d0074a08#diff-c34c805727a3008dafdd875fd96cb8c2403c40a2fcc0d16e6fab553c131ee237) doesn't impose any new requirements on the user in terms of memory management.
Changes in doc is only a recommendation for allocating buffers passed to the brgemm API.

dzarukin: I was thinking about updating [this method](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/x64/ukernel/brgemm.cpp#L155-L157) to increase the size of scratchpad required from the user and use a piece of it coming to execute through `scratchpad_ptr`.
This way memory is allocated by the user at the time they allocate it under the way the allocate it. It would be thread local as execute portions of memory belongs to threads and this combined should lead to the same effects that you are trying to achieve with a dedicated class.

ankalinin: Ah, I see.. 
Agree. Done.

src/cpu/x64/ukernel/brgemm.cpp
```diff
@@ -186,8 +201,17 @@ status_t brgemm_t::generate() {
 
 status_t brgemm_t::execute(const void *A_ptr, const void *B_ptr,
         const dim_t *A_B_offsets, void *C_ptr, void *scratchpad_ptr) const {
+    // Assert scratchpad is cache-line aligned
+    assert(reinterpret_cast<uintptr_t>(scratchpad_ptr) % k_cache_line == 0);
```
dzarukin: `assert` might spoil user's experience in case they don't have it aligned. Probably a verbose output with warning like [this one](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/jit_utils/jit_utils.cpp#L100-L101) would be a better option.

src/cpu/x64/ukernel/brgemm.cpp
```diff
@@ -228,8 +252,17 @@ status_t brgemm_t::execute(const void *A_ptr, const void *B_ptr,
         }
     }
 
+    // Assert scratchpad is cache-line aligned
+    assert(reinterpret_cast<uintptr_t>(scratchpad_ptr) % k_cache_line == 0);
```
dzarukin: Ditto.

src/cpu/x64/ukernel/brgemm.cpp
ankalinin: Updated. Is it ok?
`  if (reinterpret_cast<uintptr_t>(scratchpad_ptr) % k_cache_line != 0
            && get_verbose(verbose_t::exec_profile, component_t::ukernel))
        VWARN(primitive, ukernel, "Scratchpad is not cache-line aligned");`

src/cpu/x64/ukernel/brgemm.cpp
ankalinin: Ditto

ankalinin:
make test
set test_scope=NIGHTLY

disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022

disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

disable benchdnn_all
enable benchdnn_brgemm

src/cpu/x64/ukernel/brgemm.cpp
dzarukin: It will do, though I was hoping a warn level is set by default so users could see that message without additional actions, but I guess exec_profile is good enough to see the message when users profile it. Thanks for changing it.

PR #4351, opened 2025-11-19
cpu: aarch64: Enable 1xK and Kx1 gemv.

renato-arantes:
# Description

This PR improves the GEMV 1xK:KxN and MxK:Kx1 matmul paths for BRGeMM. It is an extension of the work initially developed by @karmeh01 on the PR #3786. The novelty of this PR is enabling the 1xK:KxN path into BRGeMM.

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?

All the NIGHTLY tests are passing when the command below is executed: 

`ctest -j$(nproc) -E $(../.github/automation/aarch64/skipped-tests.sh)`

- [x] Have you formatted the code using clang-format?

## Performance improvements

- [x] Have you submitted performance data that demonstrates performance improvements?

Here is the benchmark for VLLM with the relative (gemv / acl) comparison between GEMV and ACL. A value less than 1 indicates that GEMV is better than ACL. This PR provides significant performance improvements compared to ACL.

```
threads               1     8     16    32    64
DESC                                            
1x64:64x197         0.50  0.56  0.60  0.43  0.50
1x128:128x128       0.48  0.56  0.62  0.58  0.47
1x197:197x64        0.55  0.47  0.38  0.28  0.08
1x256:256x256       0.46  0.57  0.65  0.47  0.51
1x576:576x64        0.51  0.50  0.46  0.30  0.09
1x768:768x768       0.72  0.69  0.52  0.40  0.48
1x768:768x3072      0.86  0.82  0.66  0.50  0.51
1x768:768x2304      0.82  0.76  0.61  0.49  0.73
1x1152:1152x128     0.57  0.62  0.59  0.46  0.81
1x2304:2304x256     0.65  0.65  0.63  0.44  0.48
1x3072:3072x768     0.83  0.78  0.65  0.43  0.49
1x3584:3584x14336   0.89  0.85  0.84  0.85  0.87
1x3584:3584x3584    1.02  0.81  0.79  0.59  0.26
1x3584:3584x8192    0.90  0.85  0.88  0.74  0.59
1x4096:4096x8       0.71  0.59  0.59  0.36  0.12
1x4096:4096x6144    0.90  0.89  1.02  0.73  0.59
1x4096:4096x4096    0.94  0.93  0.73  0.70  0.53
1x4096:4096x14336   0.87  0.80  0.88  0.86  0.61
1x4096:4096x512     0.86  0.71  0.66  0.44  0.55
[...]

Issue #4396, opened 2025-11-28
brgemm_matmul segfaults with multiple threads when broadcasting dims


Sqvid:
# Summary
`brgemm_matmul` segfaults with multiple threads when broadcasting dims.

# Version
`main`: 976bf2d4eb61582c1655e69208ff8173a93d8b45

# Environment
oneDNN includes hardware-specific optimizations and may behave
differently on depending on the compiler and build environment. Include
the following information to help reproduce the issue:
* CPU: `x64` and `AArch64` 
* OS version: Linux 6.14
* git hash: 976bf2d4eb61582c1655e69208ff8173a93d8b45

# Steps to reproduce
On x64:
```sh
$ ONEDNN_VERBOSE=profile_create,profile_exec OMP_NUM_THREADS=2 ./build/tests/benchdnn/benchdnn --matmul --mode=R --stag=abcd --dtag=abcd 2x1x40x20:1x1x20x40
onednn_verbose,v1,info,oneDNN v3.11.0 (commit 976bf2d4eb61582c1655e69208ff8173a93d8b45)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:2
onednn_verbose,v1,info,cpu,isa:Intel AVX-512 with Intel DL Boost
onednn_verbose,v1,info,gpu,runtime:none
onednn_verbose,v1,info,graph,backend,0:dnnl_backend
onednn_verbose,v1,primitive,info,template:operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
onednn_verbose,v1,graph,info,template:operation,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,implementation,backend,exec_time
onednn_verbose,v1,primitive,create:cache_miss,cpu,matmul,brg_matmul:avx512_core,undef,src:f32::blocked:abcd::f0 wei:f32:a:blocked:abcd::f0 dst:f32::blocked:abcd::f0,,,2x1x40x20:1x1x20x40,0.2771
onednn_verbose,v1,primitive,create:cache_hit,cpu,matmul,brg_matmul:avx512_core,undef,src:f32::blocked:abcd::f0 wei:f32:a:blocked:abcd::f0 dst:f32::blocked:abcd::f0,,,2x1x40x20:1x1x20x40,0.00195312
[...]

PR #4434, opened 2025-12-08
cpu: aarch64: brgemm: Fix merge conflict issues and regression

Ryo-not-rio:
- Some optimizations were accidentally removed by https://github.com/uxlfoundation/oneDNN/pull/4351 when merge conflicts were being resolved.
- Also re-enabled matmuls that were being skipped due to a bug as the fix has been merged in.
- This addresses the matmul regressions in the nightly

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -1495,9 +1495,10 @@ void jit_brgemm_kernel_t::gemm_microkernel(int bd_block2, bool is_bdb_tail,
             rd_loop = brg.rd_block;
     }
 
+    unsigned long base_offset = 0;
     mov(X_TMP_2, reg_aux_A);
-    auto broadcast = [=](const ZReg &z1, size_t offset, bool is_tail,
-                             data_type_t dt) {
+    auto broadcast = [=, &base_offset](const ZReg &z1, size_t offset,
+                             bool is_tail, data_type_t dt) {
         if (is_tail) {
             eor(z1.d, z1.d, z1.d);
             auto xmm_tmp = z_tmp_1();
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1513,28 +1514,21 @@ void jit_brgemm_kernel_t::gemm_microkernel(int bd_block2, bool is_bdb_tail,
...
-                    if (offset < (1 << 6)) {
-                        ld1rw(z1.s, P_ALL_ONE / T_z,
-                                ptr(reg_aux_A, (int32_t)offset));
-                    } else {
-                        add_imm(X_DEFAULT_ADDR, reg_aux_A, offset, X_TMP_0);
-                        ld1rw(z1.s, P_ALL_ONE / T_z, ptr(X_DEFAULT_ADDR));
-                    }
-                }
-            } else if (one_of(dt, data_type::s8, data_type::u8,
+            if (dt == data_type::f32 && brg.is_gemv) {
```
jondea, line 1517: Are we sure these conditions are correct? Should this not be just `brg.is_gemv`? Note that `f32` is now in the else-if too.

Ryo-not-rio: @renato-arantes probably knows better what this if condition should be, I just updated it so it doesn't cause any different code behaviours

jondea: I'll let him take a look, but I'm happy otherwise

src/cpu/aarch64/matmul/brgemm_matmul.cpp
```diff
@@ -93,9 +93,6 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
     const auto wei_dt = weights_md_.data_type;
     const auto dst_dt = dst_md_.data_type;
 
-    // skip unsupported shapes until issue caused by PR #4126 is sorted
-    if (src_md_.ndims > 2) { return status::unimplemented; }
```
renato-arantes, line 97: Do you know if this has already been fixed? If we use ACL for the backend, it will be ok. But if we use oneDNN, a test like this will fail: ``DNNL_VERBOSE=all ./tests/benchdnn/benchdnn --matmul --stag=abcd --dtag=abcd 2x1x8x2:1x1x2x8``

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
renato-arantes: That seems good to me, since you fused the ``if/else`` statements.

src/cpu/aarch64/matmul/brgemm_matmul.cpp
Ryo-not-rio: I had read the comment incorrectly, it hasn't been fixed. I have updated the condition to be for only 4d matmuls

Sqvid: It is being tracked here https://github.com/uxlfoundation/oneDNN/issues/4396

PR #4438, opened 2025-12-09
all: introduce dynamic_fp quantization mode

dzarukin:
[MFDNN-14462](https://jira.devtools.intel.com/browse/MFDNN-14462)

### DIFF
--- a/doc/programming_model/attributes_quantization.md
+++ b/doc/programming_model/attributes_quantization.md
@@ -30,7 +30,7 @@ To support quantization, primitives should be created and executed as
 follows:
 - During primitive descriptor creation source, weights or destination
   memory descriptors use low precision datatype (e.g., `s8` or
-  `fp8_e4m3`).
+  `f8_e4m3`).
 - During primitive descriptor creation group size, data types, and
   broadcasting masks of the scaling factors and zero-point are
   provided using primitive attributes.
@@ -73,9 +73,9 @@ time.
 
 ### Dynamic Quantization
 
-The only formula for dynamic quantization currently supported by
-oneDNN is with scales computed following the
-[OCP MX Formats Specification][mx-spec], namely:
+
+oneDNN supports two dynamic quantization modes, for scales only (no
+zero-point) following the formula:
 
 \f[
 x_{f32}[:] = scale_{x} \cdot x_{quant}[:]
@@ -83,13 +83,22 @@ x_{f32}[:] = scale_{x} \cdot x_{quant}[:]
 
 where \f$x_{f32}\f$ and \f$x_{quant}\f$ are the non-quantized and
 quantized representation of \f$x\f$ respectively, and \f$scale_{x}\f$ is a
-scaling factor:
-- in `e8m0` format,
-- computed for each group of size `32`,
-- and computed as the largest power-of-two less than or equal to the
-  maximum absolute value of the group divided by the largest
-  power-of-two representable in the \f$x_{quant}\f$ data type,
-  e.g., \f$E8M0(amax(x_{quant}[:])) / E8M0(MAX\_QUANT\_DT) \f$.
+scaling factor.
+
+When using `quantization_mode::dynamic_mx`, \f$scale_{x}\f$ is computed
+following the [OCP MX Formats Specification][mx-spec], namely \f$scale_{x}\f$:
+- has `e8m0` datatype,
+- is computed for each group of size `32`,
+- is computed as the largest power-of-two less than or equal to the maximum
+  absolute value of the group divided by the largest power-of-two representable
+  in the \f$x_{quant}\f$ datatype, e.g.,
+  \f$E8M0(amax(x_{quant}[:])) / E8M0(MAX\_QUANT\_DT)\f$.
+
+When using `quantization_mode::dynamic_fp`, \f$scale_{x}\f$ is computed in
+`f32` first and then converted to a scale datatype, namely \f$scale_{x}\f$:
+- has `f8_e4m3` datatype,
+- is computed for each group of size `16`,
+- is computed as \f$SCALE\_DT(amax(x_{quant}[:]) / MAX\_QUANT\_DT)\f$.
 
 ## General Numerical Behavior Notes
 
@@ -174,7 +183,7 @@ Key parameters of the scaling API methods are summarized below:
 | `groups` | `{}`, `{G}`, `{G1,G2,...}` | Block quantization: none, single-size, multi-dimensional blocks |
 | `data_type` | `f32`, `bf16`, `f16`, `f8_e5m2`, `f8_e4m3`, `e8m0` | Scaling factor data type |
 | `is_on_host` | `true`/`false` | Host vs device memory location of scaling factor |
-| `qmode` | `static_sazp`, `dynamic_mx` | Quantization mode: static with scales and zero-points, dynamic (MXFP8 compatible) |
+| `qmode` | `static_sazp`, `dynamic_mx`, `dynamic_fp` | Quantization mode: static with scales and zero-points, dynamic (MXFP8 compatible), dynamic (NVFP4 compatible) |
 
 (*) Support for quantization options varies based on individual primitive and
 target hardware. Refer to primitives documentation for the details.
--- a/include/oneapi/dnnl/dnnl.hpp
+++ b/include/oneapi/dnnl/dnnl.hpp
@@ -331,6 +331,10 @@ enum class quantization_mode {
     /// parameter is computed by oneDNN following the OCP MX spec
     /// formula and written as an output.
     dynamic_mx = dnnl_quantization_mode_dynamic_mx,
+    /// dynamic quantization mode where quantization parameter is computed by
+    /// oneDNN as \f$scale\_dt(amax(X) / max(dst\_dt))\f$ in `f32` then
+    /// converted to a scale type and written as an output.
+    dynamic_fp = dnnl_quantization_mode_dynamic_fp,
 };
 
 /// Converts a quantization kind enum value from C++ API to C API type.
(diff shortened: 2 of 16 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/ref_matmul.cpp
atkassen: e4m3 has a zero, so we have to guard that case here...

(I also wonder if this assignment can be dropped)

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -263,7 +273,7 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
                                 = dst_m.get_idx(dst_off, dst_zp_mask);
                         dst_zp = dst_zps.get_elem(dst_zp_idx);
                     }
-                    if (has_dst_scale && !has_dst_mx) {
+                    if (has_dst_scale && !has_dst_dynamic) {
                         dst_scale = 1.f
                                 / dst_scales.get_f32_elem(
```
atkassen: ... and here...

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -346,7 +357,6 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
                                 dst_dynamic_scales, dst_scale_off);
                         // we pre-invert the scale to apply it as multiply for the group
                         dst_group_scale = 1.f / dst_group_scale;
```
atkassen: ... and here.

tests/benchdnn/matmul/ref_matmul.cpp
dzarukin: @vpirogov found out that torchAO clamps the value to be a minimal normal e4m3 value so 0 won't pop up. I've updated the place in ref impl and in benchdnn to copy the approach.

The assignment is here for dynamic scaling and this value will be multiplied later on in a common path.

tests/benchdnn/matmul/ref_matmul.cpp
dzarukin: This spot is for regular scales, zeros are not expected here.

src/cpu/matmul/ref_matmul.cpp
dzarukin: Updated here as well, see the comment above.

doc/programming_model/attributes_quantization.md
```diff
@@ -73,23 +73,32 @@ time.
...
+following the [OCP MX Formats Specification][mx-spec], namely \f$scale_{x}\f$:
+- has `e8m0` datatype,
+- is computed for each group of size `32`,
+- is computed as the largest power-of-two less than or equal to the maximum
+  absolute value of the group divided by the largest power-of-two representable
+  in the \f$x_{quant}\f$ datatype, e.g.,
+  \f$E8M0(amax(x_{quant}[:])) / E8M0(MAX\_QUANT\_DT)\f$.
+
+When using `quantization_mode::dynamic_fp`, \f$scale_{x}\f$ is computed in
+`fp32` first and then converted to a scale datatype, namely \f$scale_{x}\f$:
```
mzhukova:
```suggestion
`f32` first and then converted to a scale datatype, namely \f$scale_{x}\f$:
```

doc/programming_model/attributes_quantization.md
```diff
@@ -73,23 +73,32 @@ time.
...
+- has `e4m3` datatype,
```
mzhukova:
```suggestion
- has `f8_e4m3` datatype,
```

doc/programming_model/attributes_quantization.md
```diff
@@ -73,23 +73,32 @@ time.
...
+When using `quantization_mode::dynamic_mx`, \f$scale_{x}\f$ is computed
+following the [OCP MX Formats Specification][mx-spec], namely \f$scale_{x}\f$:
+- has `e8m0` datatype,
+- is computed for each group of size `32`,
+- is computed as the largest power-of-two less than or equal to the maximum
+  absolute value of the group divided by the largest power-of-two representable
+  in the \f$x_{quant}\f$ datatype, e.g.,
+  \f$E8M0(amax(x_{quant}[:])) / E8M0(MAX\_QUANT\_DT)\f$.
+
+When using `quantization_mode::dynamic_fp`, \f$scale_{x}\f$ is computed in
```
mzhukova, line 97: Similarly to mx, could we reference some sort of spec here with full details wrt data type?

doc/programming_model/attributes_quantization.md
mzhukova, line 26: Also need to update this part

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -331,6 +331,10 @@ enum class quantization_mode {
     /// parameter is computed by oneDNN following the OCP MX spec
     /// formula and written as an output.
     dynamic_mx = dnnl_quantization_mode_dynamic_mx,
+    /// dynamic quantization mode where quantization parameter is
+    /// computed by oneDNN as \f$ scale_dt(amax(X) / max(dst_dt))\f$
```
mzhukova: seems like the formula is not rendered correctly
also fp32 -> f32

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -2449,6 +2449,10 @@ typedef enum {
     /// parameter is computed by oneDNN following the OCP MX spec
     /// formula and written as an output.
     dnnl_quantization_mode_dynamic_mx,
+    /// dynamic quantization mode where quantization parameter is
+    /// computed by oneDNN as \f$ scale_dt(amax(X) / max(dst_dt))\f$
```
mzhukova: (same here)

src/common/matmul.cpp
```diff
@@ -207,6 +207,21 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     sc.get_data_type(DNNL_ARG_DST) == data_type::e8m0,
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
         }
+
+        // For dynamic_fp scaling, only NVFP4 flavor is supported.
+        if (sc.get(DNNL_ARG_DST).is_dynamic_fp()) {
+            // only group size of 16
```
mzhukova, line 213: seems like doc mentions group size of 32 also being supported, should we adjust it there?

doc/programming_model/attributes_quantization.md
mzhukova, line 285: Also, need to update this to either make it more generic to cover both dynamic modes (preferable) or maybe remove "(Dynamic Quantization)" to avoid confusion.

src/common/matmul.cpp
```diff
@@ -207,6 +207,21 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
...
+            VCHECK_MATMUL_UNIMPL(sc.get_mask(DNNL_ARG_DST) == full_tensor_mask,
+                    VERBOSE_UNSUPPORTED_SCALES_CFG);
+            VCHECK_MATMUL_UNIMPL(sc.get_group(DNNL_ARG_DST, -1) == 16
```
mzhukova, line 216: I wonder if we could maybe introduce something like `get_spec_group_size` and/or `get_spec_dtype` and hide all these 16/32 and e8m0/f8_e4m3, this way there would only be a single check for `is_dynamic` (which could be helpful if more supported combinations will appear)

mzhukova (commented):
I think we need a gtests extension and probably an example to go with this feature addition

doc/programming_model/attributes_quantization.md
dzarukin: The update pending GPU support and spec published.

doc/programming_model/attributes_quantization.md
dzarukin: Updated, thanks.

doc/programming_model/attributes_quantization.md
dzarukin: Updated, thanks.

doc/programming_model/attributes_quantization.md
dzarukin: Doesn't look like there's any published yet. Once it becomes available, it'll be added.

src/common/matmul.cpp
dzarukin: I dropped 32 from the doc part, thanks for noticing.

src/common/matmul.cpp
dzarukin: Until there's a consolidation with other primitives needed, I'm not sure what a named check through function would additionally achieve... Combinations are not uniform, may have several data types and multiple group sizes allowed.

doc/programming_model/attributes_quantization.md
dzarukin: The update pending GPU support and spec published.

include/oneapi/dnnl/dnnl.hpp
dzarukin: Updated, thanks for spotting.

src/common/matmul.cpp
mzhukova: Less duplication therefore better readability, and opportunity to extend support in a single place (check function),
I meant something like:
```
get_expected_group_size() {
    if (is_mx()) return 32;
    if (is_dynamic_fp()) return 16;
    return 1;
}
```

But I'm not attached if you believe that this is not unnecessary overhead.

doc/programming_model/attributes_quantization.md
mzhukova: Umm, that's just a short summary for "Quantization Model" section, so if we're adding into "Dynamic Quantization" sub-section, naturally it make sense to capture it here as well, we could point to the spec later

mzhukova (approved):

src/common/matmul.cpp
dzarukin: I'd get back to it once all recipes are officially supported and published so it's clear how to consolidate this check in a compact and meaningful way. Thanks for the idea.

PR #4439, opened 2025-12-09
fixup: asynchronous runtime support: fix the C++20 build

dzarukin:
[MFDNN-14455](https://jira.devtools.intel.com/browse/MFDNN-14455)
[MFDNN-14417](https://jira.devtools.intel.com/browse/MFDNN-14417) - code adjustment for this one.

Referencing to `this` must be done in an explicit way in C++20 which is in contradiction with C++17.
Decided to just patch places up if new come up as building macros infrastructure to support both standards seems to be a slight overkill.

Note: thanks to clang-format update, this PR takes not 6K (3K+, 3K-) lines of moved lambdas but just 300.

### DIFF
--- a/src/cpu/x64/jit_uni_binary_kernel.cpp
+++ b/src/cpu/x64/jit_uni_binary_kernel.cpp
@@ -92,7 +92,7 @@ size_t binary_kernel_t::get_tail_size() const {
 
 template <cpu_isa_t isa, typename Vmm>
 jit_uni_binary_kernel_t<isa, Vmm>::jit_uni_binary_kernel_t(
-        const binary_pd_t *pd, const jit_binary_conf_t conf, bool tail_kernel)
+        const binary_pd_t *pd, const jit_binary_conf_t &conf, bool tail_kernel)
     : binary_kernel_t(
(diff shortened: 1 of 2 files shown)


src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -1814,8 +1816,8 @@ void brgemm_convolution_fwd_t<isa>::maybe_conv_weights(const exec_ctx_t &ctx,
         const auto out_kh_offs
                 = nb_rd * jcp.oc_block * wei_dsz * jcp.vnni_block;
 
-        parallel_nd(
-                jcp.ngroups, jcp.nb_oc, KH, [=](dim_t g, dim_t ocb, dim_t kh) {
+        parallel_nd(jcp.ngroups, jcp.nb_oc, KH,
+                [= WA_THIS_COPY_CAPTURE](dim_t g, dim_t ocb, dim_t kh) {
```
densamoilov: (nit)
It's more of a compatibility feature than a workaround so maybe name it accordingly?

dzarukin: Renamed to `COMPAT_THIS_CAPTURE`, thanks.

PR #4441, opened 2025-12-09
cpu: aarch64: add ASIMD softmax JIT implementation

Anndrey24:
# Description
This commit introduces an f32 ASIMD `softmax` JIT implementation using the `exp` eltwise injector added in #4376, while also improving performance for the existing `sve_*` implementations (primarily by increasing the unrolling factor `unroll_regs_` and skipping the multiplication with default dequantization / requantization factors `src_scales` / `dst_scales`). For all implementations, the `exp` function is also effectively inlined by setting `save_state_= false`.

As the previous softmax implementation heavily relied on predicated instructions, `jit_softmax_base_t` was refactored to only include common logic for SVE and non-SVE implementations alike. At the same time, two different derived constructs were added to handle ISA-specific work: `jit_softmax_sve_t` and `jit_softmax_asimd_t`.

In addition, the JIT eltwise injector was changed to support storing/loading preserved vectors on non-SVE targets.

### c6g

| Shape     | Threads | jit:asimd (ms) | acl (ms) | Speedup |
| --------- | ------- | -------------- | -------- | ------- |
| 1539x387  | 1       | 1.21689        | 1.5615   | 1.28    |
| 1539x387  | 4       | 0.306583       | 0.394197 | 1.29    |
| 1539x387  | 16      | 0.078976       | 0.103172 | 1.31    |
| 1539x387  | 64      | 0.02816        | 0.04522  | 1.61    |
| 1024x4096 | 1       | 8.12552        | 10.4083  | 1.28    |
| 1024x4096 | 4       | 2.05314        | 2.62449  | 1.28    |
| 1024x4096 | 16      | 0.526042       | 0.678114 | 1.29    |
| 1024x4096 | 64      | 0.13881        | 0.182793 | 1.32    |
| 4096x4096 | 1       | 32.5925        | 41.3373  | 1.27    |
| 4096x4096 | 4       | 8.19186        | 10.3651  | 1.27    |
| 4096x4096 | 16      | 2.0928         | 2.66398  | 1.27    |
| 4096x4096 | 64      | 0.734764       | 0.937735 | 1.28    |

### c7g

| Shape     | Threads | jit:sve_256 (after) | jit:sve_256 (before) | Speedup |
| --------- | ------- | ------------------- | -------------------- | ------- |
| 1539x387  | 1       | 0.496183             | 0.748606             | 1.51    |
| 1539x387  | 4       | 0.125604            | 0.189787             | 1.51    |
| 1539x387  | 16      | 0.0327991            | 0.049228             | 1.50    |
| 1539x387  | 64      | 0.0168958            | 0.021218             | 1.26    |
| 1024x4096 | 1       | 3.30655              | 5.12185              | 1.55    |
| 1024x4096 | 4       | 0.816933            | 1.30929              | 1.60    |
| 1024x4096 | 16      | 0.201434            | 0.329952             | 1.64    |
| 1024x4096 | 64      | 0.0638843            | 0.108232             | 1.69    |
| 4096x4096 | 1       | 14.0951             | 20.4236              | 1.45    |
| 4096x4096 | 4       | 3.59208             | 5.56156              | 1.55    |
| 4096x4096 | 16      | 0.976691              | 1.43602              | 1.47    |
| 4096x4096 | 64      | 0.319601            | 0.432615             | 1.35    |

### c8g

| Shape     | Threads | jit:sve_128 (after) | jit:sve_128 (before) | Speedup |
| --------- | ------- | ------------------- | -------------------- | ------- |
| 1539x387  | 1       | 0.611942            | 0.863312             | 1.41    |
| 1539x387  | 4       | 0.154307            | 0.217245             | 1.41    |
| 1539x387  | 16      | 0.0403765            | 0.055711             | 1.38    |
| 1539x387  | 64      | 0.0169364            | 0.023519             | 1.39    |
| 1024x4096 | 1       | 4.19283             | 6.07039              | 1.45    |
| 1024x4096 | 4       | 1.06277             | 1.50691              | 1.42    |
| 1024x4096 | 16      | 0.267653            | 0.367653             | 1.37    |
| 1024x4096 | 64      | 0.0835895            | 0.130347             | 1.56    |
| 4096x4096 | 1       | 18.3172             | 24.4886              | 1.34    |
| 4096x4096 | 4       | 4.66065            | 6.25783              | 1.34    |
| 4096x4096 | 16      | 1.20816             | 1.58102              | 1.31    |
| 4096x4096 | 64      | 0.30616            | 0.478697             | 1.56    |

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -479,8 +479,8 @@ void jit_uni_eltwise_injector_t<isa>::exp_compute_vector_fwd(
     h->and_(ZRegD(t2.getIdx()), ZRegD(t0.getIdx()),
             ZRegD(IDX(table_val(exp_not_mask17, z_tmp))));
     h->fsub(t2, t0, t2);
-    h->movprfx(t0, p_all, ZRegS(IDX(table_val(exp_coeff2, z_tmp))));
-    h->fmad(t0, p_all, t2, ZRegS(IDX(table_val(exp_coeff1, z_tmp))));
+    h->fmad(ZRegS(IDX(table_val(exp_coeff2, t0))), p_all, t2,
+            ZRegS(IDX(table_val(exp_coeff1, z_tmp))));
     h->fmad(t0, p_all, t2, ZRegS(IDX(table_val(one, z_tmp))));
     h->fmul(t0, t1, t0);
 }
--- a/src/cpu/aarch64/jit_uni_softmax.cpp
+++ b/src/cpu/aarch64/jit_uni_softmax.cpp
@@ -102,6 +102,7 @@ struct jit_softmax_base_t : public jit_generator_t {
     bool axis_is_blocked_;
     bool need_scratchpad_;
 
+    size_t vlen_ = 0;
     size_t simd_w_ = 0;
     size_t unroll_regs_ = pd_->is_fwd() ? 16 : 4;
 
@@ -117,7 +118,6 @@ struct jit_softmax_base_t : public jit_generator_t {
 
     enum class op_t : unsigned { max, sum };
 
-    template <cpu_isa_t isa>
     void load_common_params();
     void compute_predefined_variables();
     size_t compute_process_n_elems(const memory_desc_wrapper &mdw) const;
@@ -164,13 +164,15 @@ size_t jit_softmax_base_t::compute_axis_stride(
     return compute_process_n_elems(mdw) * mdw.data_type_size();
 }
 
-template <cpu_isa_t isa>
 void jit_softmax_base_t::load_common_params() {
-    using TReg = typename cpu_isa_traits<isa>::TReg;
-
-    fmov(TReg(vone_idx).s, 1.0);
     mov(W_TMP_0, float2int(-FLT_MAX));
-    dup(TReg(vneg_flt_max_idx).s, W_TMP_0);
+    if (vlen_ > cpu_isa_traits<sve_128>::vlen) {
+        fmov(ZReg(vone_idx).s, 1.0);
+        dup(ZReg(vneg_flt_max_idx).s, W_TMP_0);
+    } else {
+        fmov(VReg(vone_idx).s, 1.0);
+        dup(VReg(vneg_flt_max_idx).s, W_TMP_0);
+    }
 
 #define PARAM_OFF(x) offsetof(call_params_t, x)
 #define PARAM_LOAD(reg, var) \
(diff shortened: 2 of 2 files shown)


michalowski-arm:
As this change is pretty big, do you think it would be possible to neatly split it into two commits: one for the sve optimizations and one for the asimd impl? The sve changes should even maybe be a separate PR.

Anndrey24:
I've now split up the changes into 3 separate commits: 
 1. `cpu: aarch64: refactor jit_uni_softmax`: 
     - keeps ISA-agnostic logic in `jit_softmax_base_t`, while all SVE-specific code is moved into a new construct `jit_softmax_sve_t`. 
     - most of the changes are due to indentation differences.
 2. `cpu: aarch64: add ASIMD softmax JIT implementation`: 
     - adds ASIMD kernel, but also improves SVE kernels as the unroll factor change is done directly in the common base struct `jit_softmax_base_t`.
 3. `cpu: aarch64: improve SVE JIT softmax performance`: 
     - adapts some of the ASIMD performance gains for the SVE kernels too, in particular SVE 128 as they share the same vector length.

I will move the final commit to a follow-up PR if you think that's best. I've only left all 3 together for now as the c7g/c8g speedups would be less noticeable at a glance with the SVE improvements in commits 2 and 3 split up, compared to being altogether in a single table like this.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2506,13 +2536,24 @@ void jit_uni_eltwise_injector_t<asimd>::exp_compute_vector_fwd(
...
 
+    Xbyak_aarch64::Label L_done, L_special;
     const auto &t0 = VReg4S(vmm_src.getIdx());
     const auto &t1 = VReg4S(vmm_aux0.getIdx());
     const auto &t2 = VReg4S(vmm_aux1.getIdx());
     const auto &t3 = VReg4S(vmm_aux2.getIdx());
     const auto &t4 = VReg4S(vmm_aux3.getIdx());
     const auto &t_tmp = VReg4S(vmm_tmp.getIdx());
 
+    const float special_bound_input = 126.5f * logf(2.0f);
```
jondea: Can this variable have a more specific name?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2565,15 +2606,17 @@ void jit_uni_eltwise_injector_t<asimd>::exp_compute_vector_fwd(
...
-    // mask_special = (|n| > 126)
-    Xbyak_aarch64::Label L_done, L_special;
-    const auto &v_mask_special = v_c3;
-    const auto &v_special_bound = table_val(exp_special_bound, t_tmp); // 126.0f
-    h->facgt(v_mask_special, v_n, v_special_bound);
-    h->addp(DReg(t_tmp.getIdx()), VReg2D(v_mask_special.getIdx()));
-    h->fmov(h->X_TMP_0, DReg(t_tmp.getIdx()));
-    h->cbnz(h->X_TMP_0, L_special);
+    if (need_special_case) {
+        // Check if any lane needs special-case handling
```
jondea, line 2609: Does this include NaN and Inf? I think a couple of comments explaining the flow for them would be useful

jondea (commented):
This looks really good, thank you! Couple of comments, I will have another look over and will probably have more.

dzarukin (approved):

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: > // i.e. if any input |x| ≥ 126.5 / ln 2 ≈ 87.6831f
> // Includes ±Inf, but not NaNs, which propagate through the fast path

I've added some comments in the latest commit to clarify this a bit.

src/cpu/aarch64/jit_uni_softmax.cpp
```diff
@@ -119,320 +115,447 @@ struct jit_softmax_base_t : public jit_generator_t {
...
+    , src_d_(pd_->is_fwd() ? pd_->src_md() : pd_->diff_src_md())
+    , dst_d_(pd_->dst_md())
+    , diff_dst_d_(pd_->diff_dst_md())
+    , need_scratchpad_(
+              utils::one_of(dst_d_.data_type(), data_type::u8, data_type::s8))
+    , simd_w_(vlen / sizeof(float)) // bf16 works on ymms
+{}
+
+// SVE JIT softmax generator
+template <cpu_isa_t isa>
```
jondea: If we are making two classes: `jit_softmax_sve_t` and `jit_softmax_asimd_t`, then I think we should get rid of this isa paramater. This means we could replace all the TReg stuff with concrete ZReg. The isa template parameter + TReg machinery is only useful if you want a single class to handle `asimd` and `sve`. It would remove a lot of things

src/cpu/aarch64/jit_uni_softmax.cpp
```diff
@@ -460,212 +583,354 @@ struct jit_softmax_base_t : public jit_generator_t {
...
+}
+
+void jit_softmax_asimd_t::generate() {
+    assert(pd_->is_fwd() && is_softmax_);
+
+    exp_injector_.reset(
+            new jit_uni_eltwise_injector_t<asimd>(this, alg_kind::eltwise_exp,
+                    0.0f, 0.0f, 1.0f, true, reg_exp_injector_table,
+                    injector_mask, injector_tmp, true, false, false));
+    exp_injector_->set_input_range(-INFINITY, 0.f);
```
jondea, line 838: Can we be more aggressive and ignore tiny values, and avoid any special case handling? Specifically any that will end up less than the precision of our accumulation datatype after exp. Softmax subtracts the max, so we know that there will be at least one value such that `exp(x) = 1`, therefore we can ignore any values such that `1 + eps == 1`

Anndrey24: This is a very interesting idea, although I don't think it would make much of an impact in this case, as the softmax path doesn't trigger/generate the exp injector’s special‑case handling at all.
>   const float special_case_input_threshold = 126.5f * logf(2.0f); // ~87.6831f
>   const float ln_flt_min = logf(FLT_MIN); // ~-87.3365f
>   bool need_clamp = min_input_ < ln_flt_min;
 >  bool need_special_case = max_input_ >= special_case_input_threshold;

In theory, special-case handling is only needed when the input range crosses the threshold ±87.6831 (i.e. `min_input_ < -87.6831` or `max_input_ >= 87.6831`):
 - On the positive side, since the highest possible input coming from softmax is 0, it means that the positive‑side special case (max_input_ >= 87.6831) is irrelevant.
 
 - On the negative side, as we go down from x = 0, we reach the natural log of FLT_MIN (`ln_flt_min = -87.3365`) before the special case input threshold (-87.6831), so the existing `h->fmax(t0, t0, table_val(exp_ln_flt_min_f, t4));` clamp already covers the relevant underflow range.

The special‑case path is only required when `max_input_ >= 87.6831` (e.g. standalone eltwise) because `ln(FLT_MAX) = 88.7228` lies slightly beyond that threshold, leaving a value range that another clamp could not handle by itself.

jondea: Excellent explanation, thank you!

src/cpu/aarch64/jit_uni_softmax.cpp
Anndrey24: Agreed!

Anndrey24:
I've also updated the results table at the top with the current SVE speedups.

jondea:
Excellent work, LGTM

> I've also updated the results table at the top with the current SVE speedups.

What was the change that sped it up even more?

jondea:
This is definitely optional for this current change, but after your changes, we could also remove the templating in the top level classes too

Anndrey24:
> What was the change that sped it up even more?

It was partly removing 2 unnecessary `movprfx` instructions in the exp injector, especially the [first one](https://github.com/uxlfoundation/oneDNN/pull/4441/changes#diff-2d98834a10af34edea40cb9565e9ff882d180ca2af9308e5fb088e880ee4efd5L442), as it fixed a register dependency chain issue on `t1` caused by merging predication in the following `frintm` instruction. 

Then, with this dependency issue fixed, `sve_256` is also 10-20% faster using the `accumulate_vsum` [unrolling method](https://github.com/uxlfoundation/oneDNN/pull/4441/changes#diff-c56431171a59f7de0ee88b757594f3d312b2f814026d7bd96ce32f3cc1dfaa35R496) I had initially only added for `sve_128`, so this path is now used for both of them (before the fix, `sve_256` had a massive slowdown using the new path):

- Current: 16x(Load + Sub_Max) ->16xExp_inj -> 16x(Add_to_Sum + Store)
vs
- Before:   16x(Load + Sub_Max + Exp_inj + Add_to_Sum + Store)

PR #4444, opened 2025-12-10
common: vtune: ITT task ID Integration for primitive tracking in VTune

avmanerikar:
# Description

Implements [MFDNN-14473](https://jira.devtools.intel.com/browse/MFDNN-14473).
The PRs introduces ITT ID tagging for primitive tasks for better representation in VTune. See MFDNN-14473 for full description.

### DIFF
--- a/src/common/primitive_iface.cpp
+++ b/src/common/primitive_iface.cpp
@@ -67,9 +67,12 @@ status_t primitive_create(primitive_iface_t **primitive_iface,
 
 #if defined(DNNL_ENABLE_ITT_TASKS)
     const bool enable_itt = itt::get_itt(itt::__itt_task_level_low);
-    if (enable_itt)
-        itt::primitive_task_start(primitive_desc_iface->impl()->kind(),
-                primitive_desc_iface->info(), VERBOSE_create);
+    if (enable_itt) {
+        const char *pd_info = primitive_desc_iface->info();
+        auto task_id = itt::make_itt_id(pd_info, get_msec());
+        itt::primitive_task_start(primitive_desc_iface->impl()->kind(), pd_info,
+                VERBOSE_create, task_id);
+    }
 #endif
 
     if (get_verbose(verbose_t::create_profile,
@@ -107,8 +110,14 @@ status_t primitive_execute(
 
 #if defined(DNNL_ENABLE_ITT_TASKS)
     const bool enable_itt = itt::get_itt(itt::__itt_task_level_low);
-    if (enable_itt)
-        itt::primitive_task_start(pd->impl()->kind(), pd->info(), VERBOSE_exec);
+    if (enable_itt) {
+        // ITT task IDs are uniquely assigned using pd->info() and
+        // timestamps - the timestamps helps distinguish between different
+        // instances of the same configuration
+        auto task_id = itt::make_itt_id(pd->info(), get_msec());
+        itt::primitive_task_start(
+                pd->impl()->kind(), pd->info(), VERBOSE_exec, task_id);
+    }
 #endif
 
     if (get_verbose(verbose_t::exec_profile,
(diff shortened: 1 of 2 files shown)


avmanerikar:
make test
set test_scope=NIGHTLY

src/common/ittnotify.cpp
```diff
@@ -45,6 +45,7 @@ namespace {
 thread_local primitive_kind_t thread_primitive_kind;
 thread_local const char *thread_primitive_info;
 thread_local const char *thread_primitive_log_kind;
+thread_local __itt_id thread_primitive_task_id = __itt_null;
```
dzarukin: Please verify that a build with `DNNL_ENABLE_ITT_TASKS` disabled also compiles. I don't see we utilized ITT types directly anywhere in other places.

dzarukin: Besides, a default value presented only for a single object is inconsistent with previous practice.

src/common/ittnotify.cpp
```diff
@@ -126,12 +130,19 @@ const char *primitive_task_get_current_log_kind() {
...
+}
+
 void primitive_task_end(const char *log_kind) {
     if (thread_primitive_kind != primitive_kind::undefined) {
         __itt_task_end(itt_domain(log_kind));
         thread_primitive_kind = primitive_kind::undefined;
         thread_primitive_info = nullptr;
         thread_primitive_log_kind = nullptr;
+
+        __itt_id_destroy(itt_domain(log_kind), thread_primitive_task_id);
```
dzarukin, line 151: Is the order between `__itt_id_destroy` and `__itt_task_end` fixed or sporadic? What does documentation say about it?

src/common/primitive_iface.cpp
```diff
@@ -67,9 +67,13 @@ status_t primitive_create(primitive_iface_t **primitive_iface,
 
 #if defined(DNNL_ENABLE_ITT_TASKS)
     const bool enable_itt = itt::get_itt(itt::__itt_task_level_low);
-    if (enable_itt)
-        itt::primitive_task_start(primitive_desc_iface->impl()->kind(),
-                primitive_desc_iface->info(), VERBOSE_create);
+    if (enable_itt) {
+        const char *pd_info = primitive_desc_iface->info();
+        __itt_id task_id = __itt_id_make(
```
dzarukin: Looks like this call must be wrapped as all others when it comes to ITT.

src/common/ittnotify.cpp
avmanerikar: Agreed. The explicit declaration is removed and the build is verified with `DNNL_ENABLE_ITT_TASKS` disabled (The `__itt_id` variable `thread_primitive_task_id` is only present under `DNNL_ENABLE_ITT_TASKS`)

src/common/ittnotify.cpp
avmanerikar: `__itt_id_destroy()` must come after `__itt_task_end()` as it can break task-ID association in Vtune. 
The task references the ID with `__itt_task_begin()` and VTune expects the ID to remain valid for the entire task lifetime.

src/common/primitive_iface.cpp
avmanerikar: Fixed. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/ittnotify.cpp
dzarukin: Would be nice to have a comment for this detail so that the order remains as expected.

dzarukin (approved):

src/common/ittnotify.cpp
avmanerikar: Done. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

PR #4447, opened 2025-12-10
[GPU] Implement C matrix interleaving strategy

Simonsays095:
Addresses [MFDNN-14154](https://jira.devtools.intel.com/browse/MFDNN-14154).

Implements a new strategy for GEMM which assigns C elements to threads in a cache-line-aware method. The result is an increase in the effective leading dimension, which decreases the average number of cache line loads per thread for unaligned C matrices.

Also updates the kernel evaluator to differentiate between identical kernels with and without the new strategy parameter (`cti`). As a side-effect, this also updates evaluation of all kernels when the problem is unaligned, and this results in some additional performance enhancements. Performance CI actually doesn't show improvements from the new strategy (it's only used in one kernel currently), but it does show improvements from this updated kernel selection.

Current performance example (PTL) - unaligned case then aligned case:
> Output template: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,gpu,jit:gemm:any,,--mode=F --matmul --engine=gpu --dt=f16:f16:f16 --strides=262144x1024x1:1024x1x1024:38827264x151669x1 --attr-post-ops=mul:f16:0 64x256x1024:1x1024x151669,5089.17,87.0127,193.075,26358.5,195.964,25969.9
perf,gpu,jit:gemm:any,,--mode=F --matmul --engine=gpu --dt=f16:f16:f16 --strides=262144x1024x1:1024x1x1024:38830080x151680x1 --attr-post-ops=mul:f16:0 64x256x1024:1x1024x151669,5089.17,87.0263,102.241,49776.2,102.568,49617.4

With the PR:
> Output template: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,gpu,jit:gemm:any,,--mode=F --matmul --engine=gpu --dt=f16:f16:f16 --strides=262144x1024x1:1024x1x1024:38827264x151669x1 --attr-post-ops=mul:f16:0 64x256x1024:1x1024x151669,5089.17,86.6478,101.544,50117.9,102.026,49881.3
perf,gpu,jit:gemm:any,,--mode=F --matmul --engine=gpu --dt=f16:f16:f16 --strides=262144x1024x1:1024x1x1024:38830080x151680x1 --attr-post-ops=mul:f16:0 64x256x1024:1x1024x151669,5089.17,86.698,102.393,49702.2,102.565,49619.2


src/gpu/intel/gemm/jit/generator/pieces/driver_info.cxx
```diff
@@ -83,6 +83,7 @@ CommonDriverInfo Generator<hw>::driverInfo(GEMMProblem problem, const GEMMStrate
     if (problem.needsAGroupSums())                                info.flags |= FlagAGroupSums;
     if (problem.needsBGroupSums())                                info.flags |= FlagBGroupSums;
     info.flags |= (strategy.fillGoal << FlagShiftFillGoal) & FlagMaskFillGoal;
```
kealan-barbieri, line 85: do we need to cast here as well?

src/gpu/intel/gemm/jit/generator/pieces/gemm.cxx
```diff
@@ -464,6 +500,52 @@ void Generator<hw>::gemm(GEMMProblem &problem, GEMMStrategy &strategy, GEMMState
...
+    if (strategy.cInterleaveChunk > 1) {
+        // Idea: The B and C matrices are reinterpreted as being made of X strided submatrices (X = strategy.cInterleaveChunk), with the columns interleaved.
+        //    Due to interleaving, the leading dimension of each of these matrices is X * ldc. When each submatrix
+        //    has an ld that's cache-line-aligned, we can shift these matrices along the rows to a cache line boundary
+        //    and write to C without cache line contention. Implementation is as follows:
+        // Important notes:
+        // 1. Work groups are divided among the submatrices, so #work groups in a row must be padded to a multiple of X.
+        // 2. Due to i0 shifting, in unaligned cases we need additional rows of work groups to handle extra remainders
+        // 3. Only works for TNN layout, due to assumptions about leading dimensions
+        if (problem.A.layout != MatrixLayout::T) stub();
```
kealan-barbieri, line 515: Can we check this condition during strategy selection? or possibly earlier during generation if youre concerned about blocking illegal manual configuration.

src/gpu/intel/gemm/jit/generator/pieces/driver_info.cxx
Simonsays095: Nope - currently only `cInterleaveChunk` crosses into the 64-bit territory. Without the cast below, the left shift could truncate a few of the MSB since (int << uint64_t -> int). In every other flag (so far) we don't have to worry about this.

src/gpu/intel/gemm/jit/generator/pieces/gemm.cxx
Simonsays095: Yeah, the idea is to prevent the ability to compile these kernels which will definitely have bad outcomes. As such, we should never have an invalid kernel of this kind in the catalog (it would never compile). The main motivator for this is multiderby (which doesn't support cti yet, but will) giving bad kernels - so we'll just kill the compilation and have it move on in that case.

As for the location of this check - it's actually fairly early, before `gemmSubkernel` and any k-loops. I want the stubs to be near the comments, which I wanted to be near the actual implementation. I think it's suitable here, but I'm open to other ideas if you have any.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/generator/pieces/math_helpers.cxx
```diff
@@ -176,6 +176,16 @@ void Generator<hw>::divDown(const ngen::Subregister &dst, const ngen::Subregiste
     }
 }
 
+// Divide an unsigned value by a constant, rounding up.
+template <HW hw>
+template <typename DT>
+void Generator<hw>::divUp(const Subregister &dst, const Subregister &src, uint32_t divisor, const CommonStrategy &strategy, CommonState &state)
+{
+    add<DT>(1, dst, src, divisor - 1);
```
rjoursler: What is the cost of using an algorithm more like the following to avoid the potential for integer overflow?
```
template <typename T, typename U>
inline decltype(std::declval<T>()/std::declval<U>()) div_up(T a, U b) {
    return (a / b) + (a % b != 0);
}
```
If the cost is minimal, I think it would be better to use the algorithm which is always correct.

Simonsays095: a/b is just `divDown` which is already done in this implementation, so the cost is just that of `a % b != 0` compared to `a + b - 1`. In the worst case, mod by a constant resolves to mul + shr (using multiplicative inverses), and we have to check against zero and then conditionally add. I think this change would be a total of 3 extra instructions (mul + shr + cmp).

I don't know if this is expensive (probably not), but other div_up implementations (A and B in registers) use this algorithm.

rjoursler: > I don't know if this is expensive (probably not), but other div_up implementations (A and B in registers) use this algorithm.

The other one should probably be fixed too. The reason I am suggesting this is because I have definitely hit bugs due to this overflow when implementing large buffer support. If we don't do the functionally correct transform, there will definitely be some hard to catch buggy cases.

src/gpu/intel/gemm/jit/generator/pieces/gemm.cxx
kealan-barbieri: that makes sense, thanks.

src/gpu/intel/gemm/jit/generator/pieces/math_helpers.cxx
Simonsays095: Currently don't have a subroutine for `mod` when everything is in registers so we'd need to add an implementation - and this would definitely have performance implications when the divisor isn't a compile-time constant (it might be ok if we can use multiplicative inverse trickery - I'm not sure).

I can fix this mod-by-constant version to be correct, but I'm going to leave the in-registers version for now and we'll have to tackle that one separately.

Simonsays095: @rjoursler Added the mod-by-constant implementation (plus a boolean to select between the two algorithms, if the dividend is known to be safe from overflows). I was able to shave off an extra instruction (the cmp), so the opportunity cost of the exact algorithm is one flag allocation and one mod, which is 2 instructions at worst.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4452, opened 2025-12-11
[GPU] initial support of host-side scalar zero-points in JIT GEMM

skazakov1:
PR addresses [MFDNN-14149](https://jira.devtools.intel.com/browse/MFDNN-14149)
This initial implementation is based on gemmstone's support of immediate scalar zero-points for A and B matrixes that should be passed to the kernel by int32_t "abo" argument that contains A and B zp value as 16-bit signed int.

### DIFF
--- a/src/gpu/intel/gemm/host_scalars.hpp
+++ b/src/gpu/intel/gemm/host_scalars.hpp
@@ -28,7 +28,7 @@ namespace gemm {
 // Get value of host side scalar from storage
 
 template <typename ScalarType, typename ResultType>
-status_t get_scalar_value_as(ResultType &scalar_value,
+status_t get_scalar_value(ResultType &scalar_value,
         const host_scalar_memory_storage_t *scale_storage) {
     ScalarType value = 0;
     status_t status = scale_storage->get_scalar_value(&value, sizeof(value));
@@ -40,36 +40,36 @@ status_t get_scalar_value_as(ResultType &scalar_value,
 }
 
 inline status_t maybe_get_host_scalar_value(
-        const memory_storage_t &scale_storage, float &scalar_value) {
+        const memory_storage_t &mem_storage, float &scalar_value) {
     using namespace data_type;
     const host_scalar_memory_storage_t *scalar_storage
             = utils::downcast<const host_scalar_memory_storage_t *>(
-                    &scale_storage);
+                    &mem_storage);
     status_t status = status::success;
 
     switch ((int)scalar_storage->data_type()) {
         case f32:
-            status = get_scalar_value_as<float, float>(
+            status = get_scalar_value<float, float>(
                     scalar_value, scalar_storage);
             break;
         case f16:
-            status = get_scalar_value_as<float16_t, float>(
+            status = get_scalar_value<float16_t, float>(
                     scalar_value, scalar_storage);
             break;
         case bf16:
-            status = get_scalar_value_as<bfloat16_t, float>(
+            status = get_scalar_value<bfloat16_t, float>(
                     scalar_value, scalar_storage);
             break;
         case s32:
-            status = get_scalar_value_as<int32_t, float>(
+            status = get_scalar_value<int32_t, float>(
                     scalar_value, scalar_storage);
             break;
         case s8:
-            status = get_scalar_value_as<int8_t, float>(
+            status = get_scalar_value<int8_t, float>(
                     scalar_value, scalar_storage);
             break;
         case u8:
-            status = get_scalar_value_as<uint8_t, float>(
+            status = get_scalar_value<uint8_t, float>(
                     scalar_value, scalar_storage);
             break;
         default:
@@ -81,24 +81,24 @@ inline status_t maybe_get_host_scalar_value(
 }
 
 inline status_t maybe_get_host_scalar_value(
-        const memory_storage_t &scale_storage, int &scalar_value) {
+        const memory_storage_t &mem_storage, int &scalar_value) {
     using namespace data_type;
     const host_scalar_memory_storage_t *scalar_storage
             = utils::downcast<const host_scalar_memory_storage_t *>(
-                    &scale_storage);
+                    &mem_storage);
     status_t status = status::success;
 
     switch ((int)scalar_storage->data_type()) {
         case s32:
-            status = get_scalar_value_as<int32_t, int>(
+            status = get_scalar_value<int32_t, int>(
                     scalar_value, scalar_storage);
             break;
         case s8:
-            status = get_scalar_value_as<int8_t, int>(
+            status = get_scalar_value<int8_t, int>(
                     scalar_value, scalar_storage);
             break;
         case u8:
-            status = get_scalar_value_as<uint8_t, int>(
+            status = get_scalar_value<uint8_t, int>(
                     scalar_value, scalar_storage);
             break;
         default:


skazakov1:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
```diff
@@ -165,6 +165,8 @@ struct GEMMProblem : public CommonProblem {
     MatrixAddressing Ag, Bg;                        // Addressing information for A/B group sums.
 
     bool checkBeta0 = true;                         // If true, check for beta = 0 and handle specially.
+    bool ao_hostscalar = false;
+    bool bo_hostscalar = false;
```
Simonsays095: I believe gemmstone implements host scales/zero-points as (for example, A zero-points): `problem.aoPtrDims = -1 && problem.aOffset = ABOffset::Calc`, meaning we don't need these variables. If you need to check this case frequently, we can instead add an accessor to the problem class that does this logic.

If you then pass the offsets as `ao` instead of `aoPtr`, we can separate A and B offsets. This should make the implementation much simpler.

Simonsays095: To be clear:

Device-side A zero-points:
- `problem.aoPtrDims = {0, 1 , 2}` (common, per-channel, grouped)
- `problem.aOffset = ABOffset::Calc`
- offsets buffer pointer passed as `aoPtr` argument

Host-side A zero-points:
- `problem.aoPtrDims = -1` (implies one scalar argument passed from host)
- `problem.aOffset = ABOffset::Calc` (unchanged)
- offset passed as `ao` argument

skazakov1: Thanks, Simon! Will try it.

src/gpu/intel/gemm/host_scalars.hpp
```diff
@@ -25,22 +25,89 @@ namespace gpu {
...
+        const memory_storage_t &scale_storage, float &scalar_value) {
+    using namespace data_type;
+    const host_scalar_memory_storage_t *scalar_storage
+            = utils::downcast<const host_scalar_memory_storage_t *>(
+                    &scale_storage);
+    status_t status = status::success;
+
+    switch ((int)scalar_storage->data_type()) {
+        case f32:
+            status = get_scalar_value_as<float, float>(
```
Simonsays095: Since `scalar_value` is a float, I think the compiler should be able to deduce the 2nd template parameter and you can drop it. I also recommend you rename `get_scalar_value_as` to `get_scalar_value` to avoid confusion around `get_scalar_value_as<float>` actually returning something other than float (i.e. the template parameter is the input type, not the output type).

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
```diff
@@ -41,8 +41,9 @@ bool Generator<hw>::gemmMake2DQuantizationLayouts(bool isA, const GEMMProblem &p
     bool xoTo2D = !xo2D && (isA ? problem.aOffset == ABOffset::Calc && (problem.earlyDequantizeA() || lateOffset)
                                 : problem.bOffset == ABOffset::Calc && (problem.earlyDequantizeB() || lateOffset));
     bool cColMajor = isRegisterColMajor(problem.Tc_ext, problem.C, strategy.C);
+    bool xoHostSacalar = isA ? problem.aOffsetHostScalar() : problem.bOffsetHostScalar() ;
 
-    if (!xo2D && !xoTo2D && !xs2D && !xg2D) return true;
+    if ((!xo2D && !xoTo2D && !xs2D && !xg2D) || xoHostSacalar) return true;
```
Simonsays095: I don't think this is necessary any more.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -487,10 +487,12 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
...
-    if (a_quant.zp_ndims >= 0) problem_.aOffset = ABOffset::Calc;
-    if (b_quant.zp_ndims >= 0) problem_.bOffset = ABOffset::Calc;
-    problem_.aoPtrDims = a_quant.zp_ndims;
-    problem_.boPtrDims = b_quant.zp_ndims;
+    if (a_quant.zp_ndims >= 0 || a_quant.zp_hostscalar)
+        problem_.aOffset = ABOffset::Calc;
+    if (b_quant.zp_ndims >= 0 || a_quant.zp_hostscalar)
+        problem_.bOffset = ABOffset::Calc;
+    problem_.aoPtrDims = a_quant.zp_hostscalar ? -1 : a_quant.zp_ndims;
+    problem_.boPtrDims = b_quant.zp_hostscalar ? -1 : b_quant.zp_ndims;
```
Simonsays095, line 495: I recommend changing `quant_entry_ndims` to return -1 for host-side scalar, then adding the variable `has_zp` to `quant_params_t`, which would make this logic more straightforward.

```
problem_.aOffset = a_quant.has_zp ? ABOffset::Calc : ABOffset::None;
problem_.bOffset = b_quant.has_zp ? ABOffset::Calc : ABOffset::None;
problem_.aoPtrDims = a_quant.zp_ndims;
problem_.boPtrDims = b_quant.zp_ndims;
```

src/gpu/intel/gemm/host_scalars.hpp
skazakov1: Renamed, thanks. As to dropping the 2nd parameter... not sure I've got it - scalar_value may be int and float, pls see two 
maybe_get_host_scalar_value(...) in this file. One used for host scalar scales (float), another for zp (int)

src/gpu/intel/gemm/jit/gen_kernel.cpp
skazakov1: Good idea but it brings changing with_a_zero_points() that will be inconsistent with other methods where dims < 0 means default (None, absent feature) values, for example:
    bool with_a_scales() const { return (asc_dims_ >= 0); }
    bool with_b_scales() const { return (bsc_dims_ >= 0); }
    bool with_c_scales() const {
        return !attr()->scales_.has_default_values(DNNL_ARG_DST);
    }

    bool with_a_zero_points() const { return (ao_dims_ >= 0); }
    bool with_b_zero_points() const { return (bo_dims_ >= 0); }
    bool with_c_zero_points() const {
        return !attr()->zero_points_.has_default_values(DNNL_ARG_DST);
    }

    bool with_a_group_sums() const { return (ag_dims_ >= 0); }
    bool with_b_group_sums() const { return (bg_dims_ >= 0); }
    
I dont think that host scalar ZP (in current implementation) is worth breaking this common approach

skazakov1:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

src/gpu/intel/gemm/host_scalars.hpp
Simonsays095: I don't mean to drop the templating on that parameter, you just don't have to list it explicitly here. i.e. call it like
`status = get_scalar_value<float>(scalar_value, scalar_storage);`
Since `scalar_value` is either a float (in this function) or an int (below) it'll get dispatched to the correct implementation.

src/gpu/intel/gemm/jit/gen_kernel.cpp
Simonsays095: I see. At some point we should refactor to store all quantization information directly in `quant_params` when calling `init_attr` and have these accessors/queries use this. Then adding `has_zp` to the struct could simplify these functions as well, to `bool with_a_zero_points() const { return a_quant.has_zp; }`. But I'm ok with a later refactor, since that affects other cases unrelated to this PR.

Simonsays095 (approved):

kealan-barbieri (approved):

echeresh (approved):

PR #4453, opened 2025-12-12
cpu: rv64: add support for f32 layer normalization using RVV

krishnasai-mcw:
# Description

This PR introduces optimized layer normalization primitive for RV64 architectures using RVV (RISC-V Vector) intrinsics.

## Features and Limitations

**Supported:**
* Data type: f32
* Input: plain and dense layouts, with the last dimension contiguous
* Stats modes:
  - Global stats (mean and variance provided externally)
  - Calculate stats (kernel computes mean and variance internally)
* Scale & shift
* Forward propagation - FWD_I & FWD_D

**Not supported:**
* Post-ops
* RMS layer normalization
* Backward propagation

## Implementation Details

The implementation uses an LMUL=1 configuration with 4-way loop unrolling to maximize throughput while maintaining register pressure. For statistics calculation, the kernel uses double-precision (f64) accumulators via widening operations to maintain numerical accuracy across large reduction operations. The normalization pass applies the standard formula `((x - mean) / std) × scale + shift` using fused multiply-accumulate instructions to efficiently compute the final output. 

# Checklist

## General

BenchDNN test log: [test_lnorm_all.log](https://github.com/user-attachments/files/24125961/test_lnorm_all.log)

## Performance improvements

### Performance Results

**Test Platform:** Banana Pi F3 

|**Test Case**| **Vector** | **Scalar** | **Speedup**|
|----|----|----|----|
| option_set_all | 86.46 | 1055.51 | 12.21 |
| option_set_fwks_ext_gpu | 62.36 | 765.01 | 12.27 |
| shapes_ci | 2.67 | 29.36 | 11.00 |

### DIFF
--- a/src/cpu/rv64/rvv_layer_normalization.cpp
+++ b/src/cpu/rv64/rvv_layer_normalization.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019 Intel Corporation
+* Copyright 2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
--- a/src/cpu/rv64/rvv_layer_normalization.hpp
+++ b/src/cpu/rv64/rvv_layer_normalization.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2019 Intel Corporation
+* Copyright 2025 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.


zhangjian29:
Hi @krishnasai-mcw ,

I noticed that you're using `f64` data type. Do you think it is necessary to add a check to ensure the platform supports `f64`?

src/cpu/rv64/rvv_layer_normalization.cpp
```diff
@@ -0,0 +1,242 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+#include <math.h>
+#include "common/dnnl_thread.hpp"
+#include "common/memory_desc_wrapper.hpp"
+#include "cpu/rv64/rvv_layer_normalization.hpp"
+#include <riscv_vector.h>
```
zhangfeiv0: (nit) grouping oneDNN header and "system" header separately and putting them in alphabetical order

krishnasai-mcw: Thanks for the suggestion! I've fixed the headers as you recommended.

krishnasai-mcw:
> Hi @krishnasai-mcw ,
> 
> I noticed that you're using `f64` data type. Do you think it is necessary to add a check to ensure the platform supports `f64`?

Hi @zhangjian29 ,
No extra check for f64 is needed. The RISC-V G extension itself includes the D extension, which provides full double-precision (f64) support. In our CMake setup, we already target G by default, ensuring f64 is supported by default.

Thanks

zhangfeiv0 (approved):

src/cpu/rv64/rvv_layer_normalization.cpp
```diff
@@ -0,0 +1,244 @@
+/*******************************************************************************
+* Copyright 2019 Intel Corporation
```
dzarukin: Please update the copyright year to 2025 or 2026.

dzarukin (approved):

PR #4206, opened 2025-10-24
xe: ocl: switch from constant memory to functions

echeresh:
See internal repo for more details.

PR #4454, opened 2025-12-12
xe: move alg_kind_t out of kernel_ctx

rjoursler:
Replaces alg_kind_t defines with an interop header. This is intended to reduce the excessive noise this introduces when analyzing the OpenCL compilation defines.

### DIFF
--- a/src/gpu/intel/include/dnnl_interop.h
+++ b/src/gpu/intel/include/dnnl_interop.h
@@ -1,0 +1,145 @@
+/*******************************************************************************
+ * Copyright 2025 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


rjoursler:
make test
disable test_device_cpu

src/gpu/intel/include/dnnl_interop.h
```diff
@@ -0,0 +1,132 @@
...
+ *******************************************************************************/
+#ifndef GPU_INTEL_INCLUDE_DNNL_INTEROP_H
+#define GPU_INTEL_INCLUDE_DNNL_INTEROP_H
+
+// Intended for use in both OpenCL C and C++ code. This is largely intended to
+// enable validation of the portion of the oneDNN C API that is required within
+// OpenCL kernels.
+
+#ifdef __OPENCL_VERSION__
+#define CONSTANT __constant
```
echeresh: Can we use macros here? See https://github.com/uxlfoundation/oneDNN/pull/4206 for more details, there is an issue with constant memory for embargo configurations.

I'm wondering if we can simply convert the whole header to use macros for both host/device and then put this header into a separate cpp file (e.g. `dnnl_interop_check.cpp`) to prevent macro pollution.

echeresh (approved):

src/gpu/intel/include/dnnl_interop.h
rjoursler: Sure, we can remove the use of `__constant`. I went ahead and switched this to just using an enum (mainly because this is a bit less work to implement on the current PR and allows using unnamed namespaces to control linking).

rjoursler:
make test
disable test_device_cpu

Issue #21, opened 2017-01-24
Issue when "make install" 


curlyt:
hi ,

I followed the steps to install MKLDNN. However, in the last step when I want to "make install". It showed the Error msg as below. 

Install the project...
-- Install configuration: "Release"
-- Installing: /usr/local/lib/libmklml_intel.so
CMake Error at cmake_install.cmake:36 (FILE):
  file INSTALL cannot copy file
  "/home/curly/mkl-dnn/external/mklml_lnx_2017.0.1.20161005/lib/libmklml_intel.so"
  to "/usr/local/lib/libmklml_intel.so".

can anyone helps me?

rsdubtso:
Did you specify CMAKE_INSTALL_PREFIX? If not then the default install prefix /usr/local which is usually writable only by the root user. Please point CMAKE_INSTALL_PREFIX to a directory writable by your user. For example
`cmake -DCMAKE_INSTALL_PREFIX=$HOME/mkldnn`.

curlyt:
It works, thank you!

PR #4457, opened 2025-12-13
cpu: rv64: jit: use updated Xbyak features to detect Zvfh

xiazhuozhao:
## Summary
This PR simplifies `src/cpu/rv64/cpu_isa_traits.hpp` by removing the ad-hoc `/proc/cpuinfo` parsing logic used for `Zvfh` extension detection.
Instead, we now leverage the updated `Xbyak_riscv` backend which introduces robust ISA detection capabilities.

## Dependency / Status 
This PR depends on upstream changes in `xbyak_riscv`.
Upstream PR: [herumi/xbyak\_riscv\#21](https://github.com/herumi/xbyak_riscv/pull/21) .
<del>This PR should remain Draft until the upstream PR is merged and the submodule is updated.</del>
<del>The dependent PR is about to be merged (already merged in dev branch), so this PR is opened to get review feedback.</del>
The dependent PR have been already merged.

## Motivation & Improvements
The new detection mechanism via `Xbyak_riscv` offers significant advantages over the current local implementation:
1.  `riscv_hwprobe`:
      * Current: Relies exclusively on legacy string parsing of `/proc/cpuinfo`, which can be fragile and format-dependent.
      * New: Prioritizes the `riscv_hwprobe` syscall (introduced in Linux 6.4). This is the official, kernel-recommended interface for querying RISC-V extensions, offering higher reliability and lower overhead than file I/O.
2.  Fallback:
      * The `Xbyak` implementation includes a safe fallback mechanism. If the kernel is older (\< 6.4) and `hwprobe` is unavailable, it gracefully reverts to parsing `/proc/cpuinfo`.
3.  Separation of Concerns:
      * Delegating ISA detection to the assembler backend (Xbyak) removes duplicated logic from oneDNN. This reduces technical debt and ensures that any future fixes to ISA detection in Xbyak automatically benefit oneDNN.

## Key Changes
  * Removed `detect_zvfh_procfs()` helper function and `<fstream>` dependencies.
  * Updated `Riscv64Cpu` constructor to check `Xbyak_riscv::CPU::getInstance().hasExtension(RISCVExtension::Zvfh)`.
  * Unified logic: both base Vector (`V`) and `Zvfh` now use the same source of truth.

### DIFF
--- a/third_party/xbyak_riscv/README.md
+++ b/third_party/xbyak_riscv/README.md
@@ -1,3 +1,3 @@
 This code is from [Xbyak_riscv](https://github.com/herumi/xbyak_riscv)
 
-tag: 1.01
+tag: 1.02


third_party/xbyak_riscv/xbyak_riscv_util.hpp
```diff
@@ -51,14 +53,58 @@ namespace Xbyak_riscv {
...
     C = COMPAT_HWCAP_ISA_C,
-    V = COMPAT_HWCAP_ISA_V
+    V = COMPAT_HWCAP_ISA_V,
+
+    // 26+: Extended Z-extensions
+    // Adding new extensions here is safe and conflict-free
+    Zvfh = 1ULL << 26,
+    Zvbb = 1ULL << 27,
+    Zvbc = 1ULL << 28,
+    Zvkg = 1ULL << 29
```
zhangjian29, line 107: It is not recommended to directly modify the third-party code, as this would create difficulties during future upgrades. I think we could move the entire detection logic to ·cpu_isa_traits· and use it locally. What is your opinion on this approach?

third_party/xbyak_riscv/xbyak_riscv_util.hpp
```diff
@@ -96,10 +145,38 @@ class CPU final {
...
+            uint64_t v = requests[0].value;
+            // Update V support from hwprobe if present
+            if (v & RISCV_HWPROBE_IMA_V) hwcapFeatures |= static_cast<uint64_t>(RISCVExtension::V);
+
+            // Detect Z-extensions using the table
+            const struct {
+                RISCVExtension id;
+                uint64_t hwprobe_bit; // Bit in RISCV_HWPROBE_KEY_IMA_EXT_0
+            } table[] = {
+                { RISCVExtension::Zvfh, RISCV_HWPROBE_EXT_ZVFH },
```
zhangjian29, line 165: Is this approach able to detect `zvfhmin`? If there are some hardwares support `zvfhmin` only but doesn't support the entire `zvfh` (e.g., the `__riscv_vfncvt_f_f_w_f16m1` and `__riscv_vfwcvt_f_f_v_f32m2` intrinsics used in `rvv_nhwc_pooling`), this detection might fail.

third_party/xbyak_riscv/xbyak_riscv_util.hpp
xiazhuozhao: > It is not recommended to directly modify the third-party code, as this would create difficulties during future upgrades. I think we could move the entire detection logic to ·cpu_isa_traits· and use it locally. What is your opinion on this approach?

Hello, they are the [changes](https://github.com/herumi/xbyak_riscv/compare/7579888..10ecd96) that have already been merged into xbyak_riscv.

third_party/xbyak_riscv/xbyak_riscv_util.hpp
xiazhuozhao: > Is this approach able to detect `zvfhmin`? If there are some hardwares support `zvfhmin` only but doesn't support the entire `zvfh` (e.g., the `__riscv_vfncvt_f_f_w_f16m1` and `__riscv_vfwcvt_f_f_v_f32m2` intrinsics used in `rvv_nhwc_pooling`), this detection might fail.

Hi, currently xbyak_riscv has not implemented support for Zvfhmin detection yet.

According to the [kernel definition](https://github.com/torvalds/linux/blob/master/arch/riscv/include/uapi/asm/hwprobe.h#L57), Zvfh and Zvfhmin are defined as two separate constants (masks). Therefore, devices that only support Zvfhmin will not trigger the current detection (it will return false).

zhangfeiv0 (approved):

dzarukin (approved):

PR #4472, opened 2025-12-18
xe: gemm: fixup cinterleave strategy

Simonsays095:
Fixes [MFDNN-14491](https://jira.devtools.intel.com/browse/MFDNN-14491). One of the last changes I made when implementing the feature was fixing some indexing logic: originally I shifted B and C offsets to ignore any columns that came before the thread's `j0` column, setting `j0=0` afterward. The issue is when we use cooperation between threads, we need to offset each thread from the work group's initial j0, `wgJ0`. This logic doesn't handle this case correctly.

The fix was to ignore columns before the work group, setting `wgJ0=0` and `j0<-j0-wgJ0`. But I forgot to divide the nonzero `j0` by the interleave chunk size. This just adds the final `divUp` when we need to use `wgJ0`, otherwise goes back to the original simpler logic.


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_rnn
enable benchdnn_ip
set test_scope=NIGHTLY

Simonsays095:
make test linters

PR #4474, opened 2025-12-19
xe: enable resuable_dispatcher_t to handle all buffer layouts

rjoursler:
This PR modifies the generation and encoding used by `reusable_dispatcher_t` to significantly increase flexibility. This increased flexibility is then used to make the reference eltwise kernel 100% completely reusable. The key changes in this PR:
* Switch to directly encoding expressions for calculations - this enables better compression of the expressions, so that more buffers can be registered, to encode expressions unrelated to buffer offsets (i.e. `gws_overflow` and `gws_in_padding`), and to optimize expensive computations (i.e. the addition `idiv`).
* Switch buffer offsets computation to be in terms of the outer dimensions i.e. `sum(outer_dim_idx * outer_dim_stride) + offset(get_inner_dim())`. This enables properly inclining constants associated with blocked layouts, which avoids expensive divisions.

Beyond that, there are a few other (but more minor) optimizations at play
* Switch named buffer encoding to a bitset. This reduces the overall structure size as we do not need to store buffer names and normalizes the structure layout which could prevent non-determinism due to ordering.
* Use 32-bit values in the runtime params when possible. This reduces the necessary data transfer to the kernel, and also avoids emulated 64-bit arithmetic.

Edit: The eltwise kernel PR was dropped for now due to merge conflicts. I will look to merging it in a future PR.

### DIFF
--- a/src/gpu/intel/block_structure.hpp
+++ b/src/gpu/intel/block_structure.hpp
@@ -110,20 +110,14 @@ struct block_layout_t {
 
     size_t size() const { return num_blocks; }
     bool empty() const { return num_blocks == 0; }
-    const block_t &front() const {
-        gpu_assert(num_blocks > 0);
-        return blocks[0];
-    }
-    block_t &front() {
-        gpu_assert(num_blocks > 0);
-        return blocks[0];
-    }
+    const block_t &front() const { return blocks[0]; }
+    block_t &front() { return blocks[0]; }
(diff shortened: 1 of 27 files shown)


rjoursler:
make test perf-gpu

rjoursler:
make test perf-gpu

rjoursler:
make test
disable test_device_cpu

PR #4475, opened 2025-12-22
cpu: aarch64: Add sve isa enum with no vlen

jondea:
# Description

Add `sve` to the `cpu_isa_t` enum for AArch64, so that we can express that an implementation is vector length agnostic (VLA at _kernel generation_, not yet using VLA instructions, but we can work towards that). It is semantically different to `sve_128` but almost functionally equivalent, the exception being that trying to access the vlen at compile time is a compilation error. This forces the implementation to use a runtime value for the vlen, elliminating a potential logical error. Add/modify a few helper functions which allow kernel generators to access runtime value generically in the correct way.

The functionality is demonstrated by its use in jit_uni_eltwise.

Broadly speaking, there should be no functional changes here, this is more a code quality/dev improvement.

Also removes sve_384, because [SVE length must be a power of 2](https://developer.arm.com/documentation/102476/0101/Introducing-SVE) 

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/jit_uni_eltwise.cpp
+++ b/src/cpu/aarch64/jit_uni_eltwise.cpp
@@ -216,7 +216,7 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
 
         add_imm(reg_src, reg_src, dtype_size(), X_TMP_0);
         add_imm(reg_dst, reg_dst, dtype_size(), X_TMP_0);
-        add_imm(reg_diff_dst, reg_diff_dst, dtype_size(), X_TMP_0);
+        if (!is_fwd) add_imm(reg_diff_dst, reg_diff_dst, dtype_size(), X_TMP_0);
         subs(reg_work_amount, reg_work_amount, 1);
 
         b(remainder_loop_start);


src/cpu/aarch64/jit_uni_eltwise.cpp
```diff
@@ -214,7 +216,7 @@ struct jit_uni_kernel_t : public jit_uni_eltwise_kernel_t {
 
         add_imm(reg_src, reg_src, dtype_size(), X_TMP_0);
         add_imm(reg_dst, reg_dst, dtype_size(), X_TMP_0);
-        add_imm(reg_diff_dst, reg_diff_dst, dtype_size(), X_TMP_0);
+        if (!is_fwd) add_imm(reg_diff_dst, reg_diff_dst, dtype_size(), X_TMP_0);
```
renato-arantes, line 219: Why is this `add_imm` now having to be in no forward mode?

jondea: `diff_dst` is only used in `bwd`, so I was just correcting a logical error unrelated to the rest of the PR. I should have put this in a separate commit and explained why. Good spot!

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -153,6 +153,19 @@ struct cpu_isa_traits<asimd> {
...
+    using TRegD = Xbyak_aarch64::ZRegD;
+    static constexpr int n_vregs = 32;
+    static constexpr dnnl_cpu_isa_t user_option_val
+            = static_cast<dnnl_cpu_isa_t>(dnnl_cpu_isa_sve);
+    static constexpr const char *user_option_env = "sve";
+};
+
 #define CPU_ISA_SVE(bits, shift) \
     template <> \
     struct cpu_isa_traits<sve_##bits> { \
```
dzarukin: In case inheritance from `sve` is introduced, some rows from below can be removed from a macro.

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -238,14 +260,7 @@ inline bool mayiuse_bf16() {
...
 inline int isa_num_vregs(cpu_isa_t isa) {
-    if (isa == sve_512)
-        return cpu_isa_traits<sve_512>::n_vregs;
-    else if (isa == sve_256)
-        return cpu_isa_traits<sve_256>::n_vregs;
-    else if (isa == sve_128)
-        return cpu_isa_traits<sve_128>::n_vregs;
-    else
-        return 0;
+    return isa == isa_undef ? 0 : 32;
```
dzarukin: It still better to rely on `cpu_isa_traits` instead. It works until some ISA has a different from 32 value.

dzarukin (approved):

renato-arantes (approved):

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -47,33 +47,33 @@ namespace aarch64 {
...
-    /// AARCH64 SVE 384 bits
-    dnnl_cpu_isa_sve_384 = 0xf,
-    /// AARCH64 SVE 512 bits
+    /// AArch64 SVE Vector Length Agnostic
+    dnnl_cpu_isa_sve = 0x3,
+    /// AArch64 SVE 128 bits
+    dnnl_cpu_isa_sve_128 = 0x7,
+    /// AArch64 SVE 256 bits
+    dnnl_cpu_isa_sve_256 = 0xf,
+    /// AArch64 SVE 512 bits
```
Sqvid: Changing the `enum` values will break the ABI right? Also isn't removing `vlen-384` technically backwards incompatible? (even if no one is using it)

jondea: As far as I can tell, this enum is internal only, so shouldn't affect the API/ABI

jondea: Nope! You are correct, the comment above this enum is wrong. Even though there is no _API_ for `dnnl_cpu_isa_t` on AArch64, these do actually affect the ABI via `get_effective_cpu_isa`, that's why the unit tests are failing. Interestingly, the unit tests also look wrong as they are comparing x64 and AArch64 isa types (e.g. https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp#L3841).

I think the fix for this is just leave `sve` as an internal only enum value.

src/cpu/aarch64/cpu_isa_traits.hpp
jondea: Good idea, done!

src/cpu/aarch64/cpu_isa_traits.hpp
jondea: Fair point, done.

src/cpu/aarch64/cpu_isa_traits.hpp
jondea: I have changed this patch to be more conservative, I don't think it changes any aspect of the external API/ABI now.

src/cpu/aarch64/jit_uni_eltwise.cpp
jondea: I have put this in a separate commit (I didn't think it was worth making a spearate PR for).

PR #4478, opened 2025-12-22
GEMM: support 2d grouped scales

kealan-barbieri:
# Description

Enable scales with non-trivial M/N group dim in existing scales use cases. Fix for improper bf16 special path usage. 

Fixes # [MFDNN-14021](https://jira.devtools.intel.com/browse/MFDNN-14021)

# Checklist

## General

### DIFF
--- a/src/gpu/intel/matmul/gemm.hpp
+++ b/src/gpu/intel/matmul/gemm.hpp
@@ -65,13 +65,22 @@ struct gemm_t : public primitive_t {
             bool with_bia = bias_md->ndims > 0;
             auto orig_dims = a_md->ndims;
 
+            // TODO: Enable reshaped 2D groups.
+            auto src_scales = attr()->scales_.get(DNNL_ARG_SRC);
+            int src_scale_group_ndims = 0;
+            if (!src_scales.has_default_groups())
+                for (int i = 0; i < 2; ++i) {
+                    if (src_scales.get_group(i) > 1) ++src_scale_group_ndims;
+                }
+
             auto maybe_reshape = [&]() -> status_t {
                 int batch_b_dims = 1;
                 for (int i = 0; i < b_md->ndims - 2; i++) {
                     batch_b_dims *= b_md->dims[i];
                 }
                 // for batch dim can map broadcast to 2d: eg. 4x1x4096:1x4096x16 -> 4x4096:4096x16
-                bool reshape_2d = (batch_b_dims == 1 && b_md->ndims > 2);
+                bool reshape_2d = (batch_b_dims == 1 && b_md->ndims > 2
+                        && src_scale_group_ndims < 1);
                 bool reshape_3d = (a_md->ndims > 3);
                 bool allow_reshape
                         = gpu_utils::dev_getenv("GEMM_ALLOW_RESHAPE", true);


src/gpu/intel/gemm/jit.hpp
```diff
@@ -222,16 +222,14 @@ struct gen_t : public primitive_t {
                         VERBOSE_SHAPE_RESTRICTION);
             }
 
-            VDISPATCH_GEMM(scales_ok(), VERBOSE_UNSUPPORTED_SCALES_CFG);
+            if (!attr()->scales_.has_default_values()) {
+                VDISPATCH_GEMM(scales_ok(), VERBOSE_UNSUPPORTED_SCALES_CFG);
```
dzarukin: Minor suggestion:

If you have some time for improvement, the better way is to move dispatching messages inside `scales_ok`, keep top-level call with `CHECK(...)` and move additional checks inside, such as a check for non-empty attributes.

```
    CHECK(scales_ok());
```
```
    status_t scales_ok(...) {
        if (attr()->scales_.has_default_values()) return status::success;

        VDISPATCH_GEMM(cond1, VERBOSE_UNSUPPORTED_SCALES_CFG);
        ....
    }
```

src/gpu/intel/matmul/ref.cpp
```diff
@@ -132,6 +132,7 @@ status_t ref_t::execute_ref(const exec_ctx_t &ctx) const {
             = !attr_scales.get(DNNL_ARG_SRC).has_default_groups()
             ? attr_scales.get_group(DNNL_ARG_SRC, 1)
             : (src_scale_per_k ? 1 : K);
+    const auto src_scale_group_m = attr_scales.get_group(DNNL_ARG_SRC, 1);
```
dzarukin: I might be wrong but `M` in positive number numeration is supposed to be `0`, not `1` as in (M, K) order for SRC.

The suggestion though is to move to negative values and make K group rely on `-1` value (meaning last group, as in python notation), and M group rely on `-2` value (meaning pre-last group).

tests/benchdnn/inputs/matmul/shapes_scale_2d
```diff
@@ -0,0 +1,14 @@
...
+--attr-scales=src:per_tensor:f8_e5m2:1x128+wei:per_tensor:f8_e5m2:128x128,src:per_tensor:f16:32x32+wei:per_tensor:f8_e5m2:32x32
+2x2048x1024:2x1024x512_n"DLRM:5*1"
+2x2048x256:2x256x128_n"NCF:1*1"
+2x2048x128:1x128x128_n"NCF:2*1"
+2x896x1024:2x1024x4096_n"RNN-T:Encoder_cell1_Hidden*11"
+2x896x320:1x320x1280_n"RNN-T:Prediction_Input*12"
+2048x1024:1024x512_n"DLRM:5*1"
+2048x256:256x128_n"NCF:1*1"
+2048x128:128x128_n"NCF:2*1"
+896x1024:1024x4096_n"RNN-T:Encoder_cell1_Hidden*11"
```
dzarukin, line 8: Would it be reasonable to fuse these cases into fp8/fp4 input files instead of creating another one?

kealan-barbieri: done, split the shapes off for reuse as well since they have different requirements than existing shape sets.

src/gpu/intel/matmul/ref.cpp
kealan-barbieri: youre correct, I've fixed this index but I'd rather leave the indexing value range as is.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit.hpp
kealan-barbieri: updated as you suggested, thanks!

src/gpu/intel/gemm/jit.hpp
```diff
@@ -353,14 +351,23 @@ struct gen_t : public primitive_t {
...
+                        ao_dims_, ag_dims_, a_q2d_group_k(), 0, a_q2d_group_m(),
+                        has_gs(DNNL_ARG_A), false, a_zp_hostscalar()};
+            } else {
+                a_quant = {a_scales_type_, ao_type, ag_type, asc_dims_,
+                        ao_dims_, ag_dims_, a_q2d_group_k(), a_q2d_group_m(), 0,
+                        has_gs(DNNL_ARG_A), false, a_zp_hostscalar()};
+                b_quant = {b_scales_type_, bo_type, bg_type, bsc_dims_,
+                        bo_dims_, bg_dims_, b_q2d_group_k(), 0, b_q2d_group_n(),
+                        has_gs(DNNL_ARG_B), false, b_zp_hostscalar()};
+            }
```
Simonsays095, line 366: Might be a bit of a larger refactor, but instead of swapping here and passing into `select_kernel`, could we instead not swap anything before this function, and then call `problem.transpose()` if we're swapping? We'd have to "unswap" the `eff_*` arguments (either by never swapping them in the first place, or by storing to local variables and using `std::swap`).

If you'd rather not make that sweeping of a change, then this would work better with `std::swap(a_quant, b_quant);` within `select_kernel` for now, and bigger changes can come later.

Ideally, swapping would only involve calling `problem.transpose()` and a then few `std::swap` of input buffers/variables at execution time.

src/gpu/intel/gemm/jit/generator/pieces/state.hpp
```diff
@@ -203,7 +203,7 @@ struct GEMMState : public CommonState {
         ngen::Subregister ldao, ldbo, ldco;                 // d
         ngen::Subregister ldaScale, ldbScale, ldcScale;     // d
         ngen::Subregister ldag, ldbg;                       // d
-        ngen::Subregister ldaq, ldbq, ldcq;                 // d
+        ngen::Subregister ldaqk, ldaqm, ldbqk, ldbqn, ldcq; // d
```
Simonsays095: `ldaqm` and `ldbqn` are unused

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
```diff
@@ -312,7 +312,7 @@ void Generator<hw>::gemmDequantizeOperation(bool doA, Type T, Type Tq, BinaryOp
 
     bool common = (qlayout.rows() * qlayout.cols()) == 1;
 
-    bool bfSpecialPath = (T == Type::bf16) && (Tq == Type::f32) && (xqGroupMN > 1) && layout.hasFullCrosspack(2);
+    bool bfSpecialPath = (T == Type::bf16) && (Tq == Type::f32) && (xqGroupMN > 1) && layout.hasFullCrosspack(2) && problem.Ta.isInteger();
```
Simonsays095: Why does this extra condition pop up from the use of 2D groups? If this is actually required, why doesn't `Tb` have a similar condition?

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -398,22 +401,6 @@ bool pd_t::scales_ok() {
         // Nontrivial groups are only supported across one GEMM dimension.
```
Simonsays095: This comment is no longer true, right?

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
kealan-barbieri: This is actually unrelated to 2d groups, I put it in a separate commit for that reason, I encountered this issue in testing the 2d group cases, but it affects others. 
This condition will use the `bfSpecialPath` when `Ta` is fp4/fp8 prior to this change, which is not the intended use of the special path and doesnt work.

src/gpu/intel/gemm/jit/pd.cpp
kealan-barbieri: absolutely, dropped thanks.

src/gpu/intel/gemm/jit/generator/pieces/state.hpp
kealan-barbieri: youre right, dropped throughout.

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
Simonsays095: I worry about this change, because we don't use `Ta`/`Tb` directly anywhere else in this function. This function assumes the data in `layout` has type `T` and the quantization data in `qlayout` has type `Tq`, which would make the actual tensor type `Ta` irrelevant.
What path does `fp4`/`fp8` take to get here? If the data is pre-upconverted to `T==bf16` before this function then I'd expect it to still work. If it's not pre-upconverted, then I'd expect `T==fp4` or `T==fp8` - both of which wouldn't take this path.

src/gpu/intel/gemm/jit/generator/pieces/state.hpp
Simonsays095: Please also rename `ldaqk` and `ldbqk` back to `ldaq` and `ldbq`. Quantization data has a single leading dimension, independent of any of the matrices dimensions.

kealan-barbieri: makes sense, fixed throughout.

src/gpu/intel/gemm/jit.hpp
kealan-barbieri: This is definitely meant as a first step, I'm definitely open to more thorough refactors as future work. In my view we should never swap/unswap "eff" arguments since such functions should always return the correct (gemmstone) arg data.

src/gpu/intel/gemm/jit/generator/pieces/quantization.cxx
kealan-barbieri: I looked into this more thoroughly as I dont have the failing case that inspired it on hand and not only cant I find it, I cant identify any cases that hit this condition at all. I think it should be handled separately so I've dropped this commit, I suspect this may be dead code.

src/gpu/intel/gemm/jit.hpp
kealan-barbieri: Unfortunately the difference here is not as simple as swap(a_quant, b_quant) because some of these attributes handle swap internally and some do not.  I agree we should work towards a single swap call to handle these, I'll add a TODO here.

src/gpu/intel/gemm/jit/generator/pieces/state.hpp
```diff
@@ -203,7 +203,7 @@ struct GEMMState : public CommonState {
         ngen::Subregister ldao, ldbo, ldco;                 // d
         ngen::Subregister ldaScale, ldbScale, ldcScale;     // d
         ngen::Subregister ldag, ldbg;                       // d
-        ngen::Subregister ldaq, ldbq, ldcq;                 // d
+        ngen::Subregister ldaq, ldbq, ldcq; // d
```
Simonsays095: nit: should probably undo this change, for uniform formatting.

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -65,13 +65,21 @@ struct gemm_t : public primitive_t {
...
+
             auto maybe_reshape = [&]() -> status_t {
                 int batch_b_dims = 1;
                 for (int i = 0; i < b_md->ndims - 2; i++) {
                     batch_b_dims *= b_md->dims[i];
                 }
                 // for batch dim can map broadcast to 2d: eg. 4x1x4096:1x4096x16 -> 4x4096:4096x16
-                bool reshape_2d = (batch_b_dims == 1 && b_md->ndims > 2);
+                bool reshape_2d = (batch_b_dims == 1 && b_md->ndims > 2
+                        && src_scale_group_ndims < 1);
```
Simonsays095, line 83: This limits some cases where we could reshape, but wouldn't. I'm ok with it for now since it won't cause regressions (although a comment would be nice for the future). I think the real issue here is unmasked batch dimensions. For example:
- `--attr-scales=src:per_ocic:f16:16x16 4x32x4096:1x4096x16` collapses to
- `--attr-scales=src:per_ocic:f16:16x16 128x4096:4096x16`

The 3D version has Ascales of size `1x2x256` while the 2D version has Ascales of size `8x256` and we've quadrupled the size (i.e. we can't broadcast quantization data in the same way we can with A/B).

If you're up for improving this case now, I would prefer that we check `original_qmd.dims -> squash_dims` versus `final_qmd.dims`, and if any dimensions change then we reject the reshaping. Or we can add this as a condition to `bcast_not_ok` since that's the real issue.

Simonsays095 (approved):

src/gpu/intel/matmul/gemm.hpp
kealan-barbieri: I added this to account for the removal of the 2d-groups check in `scales_ok`. Prior to that removal, any case that had its group dims modified such that more than one group dim was >1 during a reshape would be rejected and go to ref, which happens in multiple cases eg:

```
 --> ../../dnn-master/build/tests/benchdnn/benchdnn -v5 --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic
:f16:1x32+wei:per_ocic:f16:32x1 128x128:128x128
create: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 128x128:128x128
oneDNN implementation: jit:gemm:any
[IMPL_FILTER] All implementations were skipped!
[IMPL_FILTER] All implementations were skipped!
run: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 128x128:128x128
0:PASSED (395 ms) __REPRO: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 128x128:128x128
| jit:gemm:any : 1 (100%)                                  |
total: 0.40s; create_pd: 0.04s (10%); create_prim: 0.00s (0%); fill: 0.02s (6%); execute: 0.00s (0%); compute_ref: 0.00s (0%); compare: 0.01s (3%);
kealanba@jflmkl811316 /nfs/site/proj/mkl/hal9000/kealanba/dnn-ws/dnn-habana/build  (kealanba/double_block)
 --> ../../dnn-master/build/tests/benchdnn/benchdnn -v5 --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 2x128x128:1x128x128
create: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 2x128x128:1x128x128
oneDNN implementation: ocl:ref:any
[IMPL_FILTER] All implementations were skipped!
[IMPL_FILTER] All implementations were skipped!
run: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 2x128x128:1x128x128
0:PASSED (805 ms) __REPRO: --matmul --engine=gpu --dt=s8:u4:bf16 --attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 2x128x128:1x128x128
| ocl:ref:any : 1 (100%)                                   |
total: 0.81s; create_pd: 0.00s (0%); create_prim: 0.43s (53%); fill: 0.03s (4%); execute: 0.01s (2%); compute_ref: 0.00s (1%); compare: 0.04s (5%);
```

These cases are actually non-functional and fail without this modification, even in the absence of "true" 2d groups as above.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit/generator/pieces/state.hpp
kealan-barbieri: fixed, thanks.

src/gpu/intel/matmul/gemm.hpp
Simonsays095: This is what I'm talking about - the broadcast of the B matrix doesn't work with the quantization matrices
- original: `--attr-scales=src:per_ocic:f16:1x32+wei:per_ocic:f16:32x1 2x128x128:1x128x128`
  - src quant dims: 1x128x4
  - wei quant dims: 1x4x128
- reshaped: `--attr-scales=src:per_ocic:f16:2x32+wei:per_ocic:f16:32x1 256x128:128x128`
  - src quant dims: 128x4
  - wei quant dims: 4x128

The src quantization with groups of `2x32` has a different access pattern (although the same size in every dimension) as the original. But I'm going to resolve for now. We can return to this if we're missing performance in one of these cases in the future.

kealan-barbieri: yeah the reshape is correct, its just not currently supported correctly and prior to removing the condition in `scales_ok` it would dispatch to ref due to the `2x32` src scale. I'll add a TODO above the added condition to note the limitation but as far as I can tell this should only improve perf by dispatching a grouped GEMM to jit rather than ref.

tests/benchdnn/inputs/matmul/shapes_2d_attr
```diff
@@ -0,0 +1,9 @@
+2x2048x1024:2x1024x512_n"DLRM:5*1"
+2x2048x256:2x256x128_n"NCF:1*1"
+2x2048x128:1x128x128_n"NCF:2*1"
+2x896x1024:2x1024x4096_n"RNN-T:Encoder_cell1_Hidden*11"
+2x896x320:1x320x1280_n"RNN-T:Prediction_Input*12"
```
dzarukin: 320 doesn't divide by 128 to the best of my knowledge. Does benchdnn skip as invalid or as unimplemented?

kealan-barbieri: it skips as invalid, will drop the case just to reduce noise.

PR #4482, opened 2025-12-23
gpu: intel: gemmstone: guard #include <source_location>

spalicki:
MSVC analysis generates a warning when including this file while compiling with C++ older than C++20. Guarding for a file existence is not enough and requires check for C++20 standard version (`__cplusplus >= 202002L || _MSVC_LANG >= 202002L` ) or a feature presence (via `__cpp_lib_source_location` macro).

### DIFF
--- a/src/gpu/intel/gemm/jit/include/internal/utils.hpp
+++ b/src/gpu/intel/gemm/jit/include/internal/utils.hpp
@@ -21,7 +21,7 @@
 #include <string>
 #include <sstream>
 
-#if defined(__has_include) && __has_include("source_location")
+#if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
 #include <source_location>
 #endif
 


src/gpu/intel/gemm/jit/include/internal/utils.hpp
```diff
@@ -21,9 +21,12 @@
 #include <string>
 #include <sstream>
 
+#if (defined(_MSVC_LANG) && (_MSVC_LANG >= 202002L))
+ || (defined(__cplusplus) && (__cplusplus >= 202002L))
```
echeresh: I actually see this check further down in the file when `std::source_location` is used:

```c++
#if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
```

BTW, I think `__cpp_lib_source_location ` based check should work for the header as well.

Can we put it for the header as well to have these two places aligned?

spalicki:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

kealan-barbieri (approved):

PR #4485, opened 2025-12-26
tests: benchdnn: inputs: graph: using small shape to pass on win platform

rongzha1:
fix JIRA: MFDNN-14495

In the SDPA training pattern, softmax has two outputs, which causes it to fall back to the ::softmax::compute_ref kernel. This reference kernel uses a naive algorithm and has a different implementation from the JIT kernel, which can lead to small numerical discrepancies.

To resolve this failure, reduce the input shape size.


rongzha1:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

dzarukin (commented):
Would it be better to update it through the options in a list over changing the case directly?

rongzha1:
> Would it be better to update it through the options in a list over changing the case directly?

good idea! will change

rongzha1:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4466, opened 2025-12-17
cpu: rv64: cmake: fix gcc ice bug when specifying arch opt flags

zhangjian29:
# Description

This PR fixes the gcc14.2 internal compiler error (ICE) bug that occurs when specifying `-DDNNL_ARCH_OPT_FLAGS="-march=rv64gcv"` to disable `zvfh` extension. 

I attempted to disable `zvfh` extension by adding `-DDNNL_ARCH_OPT_FLAGS="-march=rv64gcv"` in the cmake options. However, I found two bugs:
- In `platform.cmake`, the test that sets `DNNL_RISCV_USE_ZVFH_INTRINSICS` to `true` passed *before* `DDNNL_ARCH_OPT_FLAGS` overried the `CMAKE_CCXX_FLAGS`.
- `rvv_nhwc_pooling.hpp` uses `DNNL_RISCV_USE_ZVFH_INTRINSICS` flag for dispatch, but this flag can be undefined.

These issues together cause a GCC ICE during compilation.

<details>
<summary>See Log</summary>

```
zhangjian@localhost:~/oneDNN/build_rvv $ cmake .. -DDNNL_TARGET_ARCH="RV64" -DDNNL_ARCH_OPT_FLAGS="-march=rv64gcv"
-- CMAKE_BUILD_TYPE is unset, defaulting to Release
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/zhangjian/tools/tool-gcc14.2/bin/riscv64-unknown-linux-gnu-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/zhangjian/tools/tool-gcc14.2/bin/riscv64-unknown-linux-gnu-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- DNNL_TARGET_ARCH: RV64
-- DNNL_LIBRARY_NAME: dnnl
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
[...]

PR #4486, opened 2025-12-26
cpu: rv64: pooling: add f16 support for nchw pooling

xiazhuozhao:
# Description

This PR introduces f16 (half-precision) support for the NCHW pooling primitive on RISC-V 64-bit architecture, utilizing the zvfh vector extension.

# Checklist

## General

## Performance improvements

[bench_with_f16_nchw_pool.log](https://github.com/user-attachments/files/24352434/bench_with_f16_nchw_pool.log)
[bench_without_f16_nchw_pool.log](https://github.com/user-attachments/files/24352435/bench_without_f16_nchw_pool.log)
[spedup_data.csv](https://github.com/user-attachments/files/24352517/spedup_data.csv)
Average speedup of 1.52x

### DIFF
--- a/src/cpu/rv64/rvv_nchw_pooling.hpp
+++ b/src/cpu/rv64/rvv_nchw_pooling.hpp
@@ -20,6 +20,7 @@
 
 #include "common/primitive.hpp"
 #include "cpu/cpu_pooling_pd.hpp"
+#include "cpu/rv64/cpu_isa_traits.hpp"
 #include "cpu/rv64/rvv_postops.hpp"
 
 namespace dnnl {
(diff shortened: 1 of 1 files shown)


src/cpu/rv64/rvv_nchw_pooling.hpp
```diff
@@ -50,10 +50,24 @@ struct riscv_nchw_pooling_fwd_t : public primitive_t {
...
-                    && dst_md()->data_type == d_type;
-            VDISPATCH_POOLING(types_ok, VERBOSE_UNSUPPORTED_DT);
+            const bool is_f16 = src_md()->data_type == data_type::f16;
+            VDISPATCH_POOLING(utils::one_of(src_md()->data_type, data_type::f32,
+                                      data_type::f16),
+                    VERBOSE_UNSUPPORTED_DT);
+            VDISPATCH_POOLING(src_md()->data_type == dst_md()->data_type,
+                    VERBOSE_UNSUPPORTED_DT);
+            if (is_f16) {
+                VDISPATCH_POOLING(DNNL_RISCV_USE_ZVFH_INTRINSICS,
```
zhangjian29: I think we could use `mayiuse(zvfh)` instead of `DNNL_RISCV_USE_ZVFH_INTRINSICS` flag to avoid potential gcc ice. See [here](https://github.com/uxlfoundation/oneDNN/pull/4466/changes#diff-afa3298fdc3557c3ccac5c9cce88e91eaeb0892d055fc3b1b8781e2d56f4692e).

xiazhuozhao: > I think we could use `mayiuse(zvfh)` instead of `DNNL_RISCV_USE_ZVFH_INTRINSICS` flag to avoid potential gcc ice. See [here](https://github.com/uxlfoundation/oneDNN/pull/4466/changes#diff-afa3298fdc3557c3ccac5c9cce88e91eaeb0892d055fc3b1b8781e2d56f4692e).

Thanks for the review! I've updated the code to use `mayiuse(zvfh)` instead of the macro as suggested to avoid the potential GCC ICE. I also added the necessary header. Good catch!

PR #4322, opened 2025-11-13
cpu: rv64: add runtime Zvfh detection and platform support

xiazhuozhao:
# Description

While `DNNL_RISCV_USE_ZVFH_INTRINSICS` is a compile-time variable that authorizes the compiler to generate `Zvfh` instructions, runtime detection of the `Zvfh` extension is also required to ensure portability of the compiled binary.

This runtime check is necessary for the oneDNN dispatching mechanism—specifically the `platform::has_data_type_support()` utility and the primitive descriptor's `init()` function—to verify the hardware's actual capabilities before attempting to execute specialized f16 vector code.

If this check fails at runtime (e.g., when running on older hardware that does not support `Zvfh`), oneDNN will fall back to a generic reference implementation, preventing an Illegal Instruction crash.

The principle of this check is as follows:

`V` Detection: The `getauxval(AT_HWCAP)` system call is used to check the hardware capability bitmask. The V extension is confirmed if the `(1UL << ('V' - 'A'))` bit is set.

`Zvfh` Detection: Since `Zvfh` is a multi-letter Z extension, it is not defined in the `AT_HWCAP` bitmask. Therefore, if the `V` extension is present, the implementation falls back to parsing the `/proc/cpuinfo` file. It reads the isa string and searches for the `_zvfh` token to confirm the extension is available.

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

PR #4487, opened 2025-12-26
cpu: riscv: add ISCAS copyright to new files

xiazhuozhao:
Add ISCAS copyright headers to files created by the Institute of Software, Chinese Academy of Sciences.

#4322 for cpu/rv64/cpu_isa_traits.*
#4300 for cpu/rv64/rvv_gemm_reorder*
#3852 for cpu/rv64/rvv_gemm_convolution*
#3785 for cpu/rv64/gemm/*

### DIFF
--- a/.github/.licenserc.yml
+++ b/.github/.licenserc.yml
@@ -10,7 +10,8 @@ header:
       |KNS Group LLC \(YADRO\)
       |Alanna Tempest
       |YANDEX LLC
-      |ZTE Corporation)\.?\s*)+
+      |ZTE Corporation
+      |Institute of Software, Chinese Academy of Sciences)\.?\s*)+
       ?(?:SPDX-License-Identifier: Apache-2\.0\s+)?Licensed under the Apache License, Version 2\.0 \(the .License.\);
       you may not use this file except in compliance with the License\.
       You may obtain a copy of the License at
--- a/src/cpu/rv64/cpu_isa_traits.cpp
+++ b/src/cpu/rv64/cpu_isa_traits.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2019 Intel Corporation
+* Copyright 2025 Institute of Software, Chinese Academy of Sciences
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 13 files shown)


.github/.licenserc.yml
```diff
@@ -10,7 +10,8 @@ header:
       |KNS Group LLC \(YADRO\)
       |Alanna Tempest
       |YANDEX LLC
-      |ZTE Corporation)\.?\s*)+
```
zhangjian29, line 13: Any reason to revise this line?

.github/.licenserc.yml
```diff
@@ -10,7 +10,8 @@ header:
...
+      |ZTE Corporation
+      |Institute of Software, Chinese Academy of Sciences)\.?\s*)+
```
zhangjian29, line 14: Maybe add check of short name "ISCAS"?

zhangjian29: Not quite sure about this.

.github/.licenserc.yml
xiazhuozhao: > Any reason to revise this line?

I'm not modifying the ZTE entry content. This is just how Git displays the change in the regex structure.

Previously, the ZTE line contained the closing regex pattern \).?\s*)+. I had to cut that pattern and move it to the new ISCAS line to make it the valid end of the group. Git sees this removal as a modification to the whole line, but logic-wise, it's just passing the closing tag to the next item.

.github/.licenserc.yml
xiazhuozhao: > Maybe add check of short name "ISCAS"?

The decision to use the full English name was reached after discussion with @zhangfeiv0, who is my mentor and an employee at the Institute.

Unlike other organizations, simply using the abbreviation (ISCAS) might be ambiguous. 

Image

Image

Image

Image

Image

However, if deemed necessary, we can wait for more community feedback to evaluate which naming convention makes the most sense.

zhangjian29: Thank you for your detailed information. I agreed with you.

.github/.licenserc.yml
zhangjian29: Got it. Thanks.

PR #4491, opened 2026-01-02
cpu: riscv: softmax: add f16 support

xiazhuozhao:
# Description

This PR adds support for FP16 (half-precision) data type to the RISC-V Vector (RVV) Softmax primitive.
Added compute_softmax_f16_rvv which utilizes Zvfh intrinsics. To ensure precision, the kernel performs accumulation and exponential/logarithm calculations in f32 (using widening conversions __riscv_vfwcvt) and converts back to f16 for the final output.

# Checklist

## General

## Performance improvements

Performance was evaluated on SG2044 (RISC-V) using 16 cores (pinned). The comparison is between the new RVV f16 implementation and the previous behavior (fallback to reference).

```
./benchdnn --mode=P --softmax --batch=test_softmax_float16
```
(Note: Only collected FWD_D cases)

[speedup_ratio.csv](https://github.com/user-attachments/files/24470702/speedup_ratio.csv)
[with_f16_softmax.csv](https://github.com/user-attachments/files/24470704/with_f16_softmax.csv)
[without_f16_softmax.csv](https://github.com/user-attachments/files/24470705/without_f16_softmax.csv)

Average Speedup: ~23.40x

### DIFF
--- a/src/cpu/rv64/rvv_softmax.cpp
+++ b/src/cpu/rv64/rvv_softmax.cpp
@@ -101,21 +101,26 @@ void compute_softmax_f16_rvv(const dnnl::impl::float16_t *src,
             i += (dim_t)vl;
         }
     } else {
+        float *tmp_dst = new float[len];
         float sum_exp = 0.f;
         for (dim_t i = 0; i < len; ++i) {
-            sum_exp += expf((float)src[i] - max_val);
+            float e = expf((float)src[i] - max_val);
+            tmp_dst[i] = e;
+            sum_exp += e;
         }
         const float inv_sum = 1.0f / sum_exp;
 
         for (dim_t i = 0; i < len;) {
             size_t vl = __riscv_vsetvl_e16m1((size_t)(len - i));
 
-            for (size_t j = 0; j < vl; ++j) {
-                float e = expf((float)src[i + j] - max_val);
-                dst[i + j] = (dnnl::impl::float16_t)(e * inv_sum);
-            }
+            vfloat32m2_t v_f32 = __riscv_vle32_v_f32m2(tmp_dst + i, vl);
+            vfloat32m2_t v_res = __riscv_vfmul_vf_f32m2(v_f32, inv_sum, vl);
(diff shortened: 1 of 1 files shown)


src/cpu/rv64/rvv_softmax.cpp
```diff
@@ -67,6 +70,56 @@ void compute_softmax_f32_rvv(
...
+        float sum_exp = 0.f;
+        for (dim_t i = 0; i < len; ++i) {
+            sum_exp += expf((float)src[i] - max_val);
+        }
+        const float inv_sum = 1.0f / sum_exp;
+
+        for (dim_t i = 0; i < len;) {
+            size_t vl = __riscv_vsetvl_e16m1((size_t)(len - i));
+
+            for (size_t j = 0; j < vl; ++j) {
```
zhangfeiv0: Why isn't vectorization applied here as it is with f32?

xiazhuozhao: > Why isn't vectorization applied here as it is with f32?

Thank you! I initially avoided vectorization here because of the required temporary `f32` buffer, as I was concerned about its performance overhead. I've now implemented the fully vectorized version and am currently benchmarking the impact. The results will be posted shortly.

xiazhuozhao: > Why isn't vectorization applied here as it is with f32?

Thanks for the suggestion. I have updated the code to include the fully vectorized path for the non-logsoftmax case (see commit 04e8c86). I benchmarked the performance to verify the impact of the temporary `f32` buffer. Comparing the fully vectorized implementation against the previous version (commit ca39a4c), the results show an average speedup of 1.29x.

Attached are the detailed benchmark logs:
[f16_softmax_with_tmp_04e8c86.csv](https://github.com/user-attachments/files/24621590/f16_softmax_with_tmp_04e8c86.csv)
[f16_softmax_without_tmp_ca39a4c.csv](https://github.com/user-attachments/files/24621591/f16_softmax_without_tmp_ca39a4c.csv)
[speedup.csv](https://github.com/user-attachments/files/24621592/speedup.csv)


The PR is ready for re-review. Thanks!

PR #4492, opened 2026-01-04
commit-msg: add verbosity for leading spaces

dzarukin:
PR adds an extra message to remove confusion when there's a leading space but the scoping is correct and also applies `black` formatting.

### DIFF
--- a/.github/automation/commit-msg-check.py
+++ b/.github/automation/commit-msg-check.py
@@ -38,6 +38,10 @@ def __scopeCheck(msg: str):
             )
             return False
 
+        if re.match("^Merge ", msg):
+            print(f"{status} FAILED: Merge commits are not allowed")
+            return False
+
         print(
             f"{status} FAILED: Commit message must follow the format "
             "<scope>:[ <scope>:] <short description>"


.github/automation/commit-msg-check.py
```diff
@@ -30,14 +31,23 @@
 def __scopeCheck(msg: str):
     status = "Message scope: "
 
-    if not re.match('^[a-z0-9_]+(, [a-z0-9_]+)*: ', msg):
-        print(f"{status} FAILED: Commit message must follow the format "
-               "<scope>:[ <scope>:] <short description>")
+    if not re.match("^[a-z0-9_]+(, [a-z0-9_]+)*: ", msg):
+        if re.match("^[ ]+", msg):
```
Sqvid: `"^\s+"` is probably preferred to catch tabs or weird unicode issues.

.github/automation/commit-msg-check.py
```diff
@@ -47,14 +57,17 @@ def __numCharacterCheck(msg: str):
     else:
         # Fixup commits usually include the full name of the commit they are
         # fixing, which adds 6 more symbols to the message. Let them in.
-        if re.match('^fixup: ', msg):
+        if re.match("^fixup: ", msg):
```
Sqvid, line 65: Should `revert:` have the same accommodation?

Sqvid (changes requested):

.github/automation/commit-msg-check.py
dzarukin: Sure thing, updated. Thanks.

.github/automation/commit-msg-check.py
dzarukin: I've added revert in the allowed list, thanks for a suggestion.

PR #4495, opened 2026-01-06
xe: jit: gemm: remove ir_context_t dependency from dsl interfaces

rjoursler:


rjoursler:
make test
disable test_device_cpu

rjoursler:
This PR validated the concept, future work on it will be handled in the upstream as not oneDNN changes were required.

PR #4496, opened 2026-01-06
Fixup recent indirect OpenCL and DSL separation changes

dzarukin:
While working on [MFDNN-14496](https://jira.devtools.intel.com/browse/MFDNN-14496), I met a compilation and linking issues.

Compilation issue roots to the fact that cl.h and cl_ext.h have different guarding macros and can't be mixed together.

Linking issue coming from direct linking to OpenCL which was removed straight before the DSL separation promotion.

### DIFF
--- a/third_party/ngen/ngen_opencl.hpp
+++ b/third_party/ngen/ngen_opencl.hpp
@@ -21,6 +21,9 @@
 
 #ifndef __OPENCL_CL_H
 #include <CL/cl.h>
+#endif
+
+#ifndef OPENCL_CL_EXT_H_
 #include <CL/cl_ext.h>
 #endif
 


src/gpu/intel/gemm/jit/dsl/runtime.cpp
```diff
@@ -90,24 +90,26 @@ dsl::hw_t get_hardware(cl_device_id device, cl_context context) {
 
     cl_int err;
     cl_uint eu_count = 0;
-    err = clGetDeviceInfo(device, CL_DEVICE_MAX_COMPUTE_UNITS, sizeof(eu_count),
-            &eu_count, nullptr);
+    err = ngen::dynamic::clGetDeviceInfo(device, CL_DEVICE_MAX_COMPUTE_UNITS,
```
spalicki, line 93: If this is our code, then `xpu::ocl::` instead of `ngen::dynamic::`

dzarukin: There's no access to `xpu` here, the only dependency dsl has is ngen, at least that was the motivation of the change after all.

echeresh (approved):

PR #4505, opened 2026-01-07
xe: grp_gemm: Add microkernel based implementation of grouped GEMM

umar456:
This pull request introduces support for a new "grouped micro GEMM" (General Matrix Multiply) implementation, along with corresponding kernel code, and enhancements to the testing utilities to support grouped memory.


tests/gtests/test_grouped_gemm.cpp
```diff
@@ -0,0 +1,168 @@
+
+#include "dnnl_test_common.hpp"
```
mzhukova: what is missing from benchdnn, so that this test is required..?

umar456: Nothing is missing. I want more control over how the data is initialized so that its easier to test/debug new configurations. I will probably remove this in the future.

mzhukova: > I want more control over how the data is initialized

Could you clarify, so that I could possibly make it better/easier?

> I will probably remove this in the future

I agree, we would need this removed prior merging if it is covered by benchdnn..

umar456: It's difficult to initialize the input data with known values to identify incorrect results. For example I want to find out why I am getting an incorrect value at a certain location in the destination buffer. With my own tests I can make my inputs to be known values to debug. I don't know of a way to do that with benchdnn.

Simonsays095: Does [--buffer-prefix](https://github.com/intel-innersource/libraries.performance.math.onednn/blob/main/tests/benchdnn/doc/knobs_common.md#--buffer-prefix) work?

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
enable benchdnn_gemm
disable test_device_cpu
enable test_device_gpu

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
```diff
@@ -263,8 +263,7 @@ std::string generateShim(const Package &package, HostLanguage language,
                  << "\n"
                     "typedef struct {\n"
                     "    "
-                 << ename;
-            if (vlen > 1) shim << vlen;
+                 << ename << vlen;
```
Simonsays095: Won't this make types like `long1` and `float1`?

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -109,6 +109,35 @@ Protocol makeProtocol(const GEMMOptions &o) {
     return {"ugemm", arguments(o), settings()};
 }
 
+InterfaceHandler generateInterface(Core hw, bool localA, bool localB,
```
Simonsays095: Since the interface is entirely defined by value stored in `GEMMOptions` (plus a hardware), can we make this a class method there?

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,174 @@
...
+        ugemm_grouped_c_type_block1, ugemm_grouped_c_type_nblock0,
+        ugemm_grouped_c_type_nblock1, CONVERT_DATA_T)
+
+#endif
+
+#if WITH_SRC_ATTR_SCALES
+#define SRC_ATTR_SCALE_ARGS , src_attr_scales
+#else
+#define SRC_ATTR_SCALE_ARGS
+#endif
```
Simonsays095: Would something like this work?
```suggestion
#define OPTIONAL(enabled, argName) CONCAT2(OPTIONAL_, enabled)(argName)
#define OPTIONAL_0(argName)
#define OPTIONAL_1(argName) , argName
```
And then below have `OPTIONAL(WITH_SRC_ATTR_SCALES, src_attr_scales)`? The idea being that a list of macros in the signature is hard to parse. This way is a bit more verbose about what's happening, while still avoiding lots of inline `#ifdef`.

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,174 @@
...
+        tile_load(&bias_in_tile, bias, n, m, 0, sg_i0, sg_j0);
+        tile_copy_reblock(bias_in_tile, &bias_tile);
+    }
+#endif
+    tile_binary(c_tile, bias_tile, binary_add);
+#endif
+
+#if DST_DT_F32
+    tile_store(c_tile, dst, n, m, lddst, sg_i0, sg_j0);
+    //tile_store_t_block2d(c_tile, dst, n, m, lddst, sg_j0, sg_i0);
```
Simonsays095: I guess this is a todo? What's stopping us from using this currently?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+            || !pd()->attr()->zero_points_.has_default_values(DNNL_ARG_SRC)) {
+        kernel_ctx_.define_int("SRC_GROUP_SIZE", pd()->src_group_sizes_[1]);
+    }
+    if (!pd()->attr()->scales_.has_default_values(DNNL_ARG_WEIGHTS)
+            || !pd()->attr()->zero_points_.has_default_values(
+                    DNNL_ARG_WEIGHTS)) {
+        kernel_ctx_.define_int("WEI_GROUP_SIZE", pd()->wei_group_sizes_[1]);
+    }
+    kernel_ctx_.define_int("SRC_ELEMS_PER_BYTE", elems_per_byte(src_dt));
+    kernel_ctx_.define_int("WEI_ELEMS_PER_BYTE", elems_per_byte(wei_dt));
```
Simonsays095: With this, you don't need to define `elems_per_byte`.
```suggestion
    kernel_ctx_.define_int("SRC_ELEMS_PER_BYTE", types::bytes_to_elements(src_dt, 1));
    kernel_ctx_.define_int("WEI_ELEMS_PER_BYTE", types::bytes_to_elements(wei_dt, 1));
```

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+                : ctx.input(DNNL_ARG_SRC | DNNL_ARG_ATTR_ZERO_POINTS)->md();
+        ldsrcq = static_cast<int>(src_quant_md->format_desc.blocking
+                                          .strides[src_quant_md->ndims - 1]);
+    }
+    if (with_wei_scales || with_wei_zero_points) {
+        const memory_desc_t *wei_quant_md = with_wei_scales
+                ? ctx.input(DNNL_ARG_WEIGHTS | DNNL_ARG_ATTR_SCALES)->md()
+                : ctx.input(DNNL_ARG_WEIGHTS | DNNL_ARG_ATTR_ZERO_POINTS)->md();
+        ldweiq = static_cast<int>(wei_quant_md->format_desc.blocking
+                                          .strides[wei_quant_md->ndims - 2]);
```
Simonsays095, line 495: Why do src/wei quantization LDs use different dimensions? Don't they have to use the abx format, which means the leading dimension is always the stride of `ndims - 1`?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+    }
+    int m_all = static_cast<int>(dst_md->dims[dst_md->ndims - 2]);
+    int n = static_cast<int>(dst_md->dims[dst_md->ndims - 1]);
+    int k = static_cast<int>(src_md->dims[src_md->ndims - 1]);
+    //printf("m_all=%d, n=%d, k=%d ldsrcq: %d ldweiq: %d\n", m_all, n, k, ldsrcq,
+    //        ldweiq);
```
Simonsays095: Remove the comment, please.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+            //CHECK_BOOL(false,
+            //         "gemm microkernel generation failure with message: %s",
+            //         ex.what());
+            return status::unimplemented;
+        }
+    }
+    //printf("Gemm: tile(%d %d)  unroll(%d, %d) sg_per_wg=(%d, %d)\n",
+    //        gemm.getSetting("wg_tile_m"), gemm.getSetting("wg_tile_n"),
+    //        gemm.getSetting("sg_tile_m"), gemm.getSetting("sg_tile_n"),
+    //        gemm.getSetting("sg_per_wg_m"), gemm.getSetting("sg_per_wg_n"));
```
Simonsays095: Please remove these comments.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+    auto internalTypes = [](Type extType) -> Type {
+        switch (extType) {
+            case Type::u4:
+            case Type::s4: return Type::s8;
+            default: return extType;
+        }
+    };
+
+    problem.Ta = internalTypes(problem.Ta_ext);
+    problem.Tb = internalTypes(problem.Tb_ext);
```
Simonsays095: Why do we require internal upconversion of int4 types? We don't do this for regular GEMM.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+    if (opts.scaleB || opts.offsetB) {
+        const quant_entry_t &src_scales = opts.scaleB
+                ? pd()->attr()->scales_.get(DNNL_ARG_SRC)
+                : pd()->attr()->zero_points_.get(DNNL_ARG_SRC);
+        memory_desc_t md;
+        const memory_desc_t &src_md = *pd()->src_md();
+        src_scales.get_md(md, src_md);
+        problem.bqGroupN = pd()->src_group_sizes_[0];
+        problem.bqGroupK = utils::rnd_up_pow2(pd()->src_group_sizes_[1]);
```
Simonsays095:
```suggestion
        const quant_entry_t &src_quant = opts.scaleB
                ? pd()->attr()->scales_.get(DNNL_ARG_SRC)
                : pd()->attr()->zero_points_.get(DNNL_ARG_SRC);
        memory_desc_t md;
        const memory_desc_t &src_md = *pd()->src_md();
        src_quant.get_md(md, src_md);
        problem.bqGroupN = pd()->src_group_sizes_[0];
        problem.bqGroupK = utils::rnd_up_pow2(pd()->src_group_sizes_[1]);
```

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+        gemm = microkernel::selectGEMM(opts, hw_info, sizes, problem);
+    } catch (const std::runtime_error &ex) {
+        std::vector<StrategyRequirement> reqs;
+        reqs.push_back(StrategyRequirement::UnrollM == 16);
+        reqs.push_back(StrategyRequirement::UnrollN
+                == utils::rnd_up_pow2(std::min<int>(pd()->M(), 64)));
+        reqs.push_back(StrategyRequirement::WGM == 2);
+        reqs.push_back(StrategyRequirement::WGN
+                == utils::rnd_up_pow2(std::max<int>(
+                        2, std::min<int>(pd()->M() / reqs[1].value, 8))));
```
Simonsays095: It looks to me like this is a last-ditch effort to get the heuristics-based strategy selection to work, is that right? Could we improve the heuristics internally, instead of forcing it here?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,482 @@
...
+                                          .strides[wei_quant_md->ndims - 2]);
+    }
+    int m_all = static_cast<int>(dst_md->dims[dst_md->ndims - 2]);
+    int n = static_cast<int>(dst_md->dims[dst_md->ndims - 1]);
+    int k = static_cast<int>(src_md->dims[src_md->ndims - 1]);
+    //printf("m_all=%d, n=%d, k=%d ldsrcq: %d ldweiq: %d\n", m_all, n, k, ldsrcq,
+    //        ldweiq);
+
+    int ldsrc = static_cast<int>(src_md->dims[src_md->ndims - 1]);
+    int lddst = static_cast<int>(dst_md->dims[dst_md->ndims - 1]);
```
Simonsays095, line 502: It looks like we assume abx format for src/dst. Can we add a check for this to `init`?

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
syurkevi: The indexing we currently have in tile_ops.hpp hardcodes tile_access with two [][]. Without float1 this will result in a compilation error.

Simonsays095: I don't think `float1` is a valid type according to [the spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-vector-data-types).
> Supported values of n are 2, 3, 4, 8, and 16 for all vector data types.

Two questions:
1. Why do we have an Nx1 tile, is this from strategies with `unroll[MN]=1`?
2. How do we handle this in SDPA?

If we really need this, we can possibly define it as `float name[N]` instead of `floatN name` - i.e. arrays instead of vectors.

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,43 @@
...
+__attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE))) kernel void
+grouped_micro_gemm(const global half *src, int ldsrc, const global half *wei, int ldwei,
+                   global float *dst, int lddst,
+                   const global int *src_offsets, const global int *dst_offsets, int n, int k) {
+    unsigned long batch = get_group_id(2);
+    int2 src_offset = *(global int2*)(src_offsets+batch);
+    src += src_offset.x * ldsrc; //A_offsets[batch];
+
+    int m = (int)(src_offset.y - src_offset.x);
+    //int n = n_array[batch];
```
h-sadia: Wouldn't it be better if we put the commented code with debug_print macros? Or I guess this the code in between commits... are you planning to clean up the commits once you are done?

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,43 @@
...
+#include "gpu/intel/sdpa/utils.h"
+
+#include "gemm_grouped.h"
+
+__attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE))) kernel void
+grouped_micro_gemm(const global half *src, int ldsrc, const global half *wei, int ldwei,
+                   global float *dst, int lddst,
+                   const global int *src_offsets, const global int *dst_offsets, int n, int k) {
+    unsigned long batch = get_group_id(2);
+    int2 src_offset = *(global int2*)(src_offsets+batch);
```
h-sadia: do we plan to make int2 in to a typedef?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,282 @@
...
+    const auto &dst_grouped = dst_d.sparse_desc().grouped_desc;
+
+    // Validate matching number of groups
+    VDISPATCH_MATMUL(src_grouped.ngroups == dst_grouped.ngroups,
+            VERBOSE_INCONSISTENT_NDIMS_WITH_VALS, "src ngroups", "dst ngroups",
+            (int)src_grouped.ngroups, (int)dst_grouped.ngroups);
+
+    ngroups_ = src_grouped.ngroups;
+    // only supported dt for now
+    VDISPATCH_MATMUL(utils::one_of(src_dt_, f32, f16)
```
h-sadia: alternatively you could also do:
`utils::everyone_is(f16, src_dt_, wei_dt_, dst_type) || utils::everyone_is(f32, src_dt_, wei_dt_, dst_type)`

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,282 @@
...
+                    && src_dt_ == wei_dt_, //&& src_dt_ == dst_type,
+            VERBOSE_UNSUPPORTED_DT_CFG);
+
+    // Check offsets are int32
+    VDISPATCH_MATMUL(
+            src_d.metadata_type(0) == s32 && dst_d.metadata_type(0) == s32,
```
h-sadia: maybe, utils::everyone_is works here as well?

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -7,9 +7,20 @@
 
 #include "gemm_grouped.h"
 
+#if DST_DT_F32 != 1
```
h-sadia: Is this for f16 and bf16 datatypes both?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -163,9 +165,12 @@ status_t grouped_micro_gemm_t::pd_t::init(impl::engine_t *engine) {
...
     // only supported dt for now
-    VDISPATCH_MATMUL(utils::one_of(src_dt_, f32, f16)
-                    && src_dt_ == wei_dt_, //&& src_dt_ == dst_type,
-            VERBOSE_UNSUPPORTED_DT_CFG);
+    VDISPATCH_MATMUL(
+            utils::one_of(src_dt_, f32, f16, bf16), VERBOSE_UNSUPPORTED_DT_CFG);
+    VDISPATCH_MATMUL(
+            utils::one_of(wei_dt_, f32, f16, bf16), VERBOSE_UNSUPPORTED_DT_CFG);
+    VDISPATCH_MATMUL(
+            utils::one_of(dst_dt_, f32, f16, bf16), VERBOSE_UNSUPPORTED_DT_CFG);
```
h-sadia: will these be intermixed in a test case, or are we strictly supporting f32 only; f16 only; bf16 only?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -360,40 +415,52 @@ status_t grouped_micro_gemm_t::execute(const exec_ctx_t &ctx) const {
     const auto &bias_data = CTX_IN_STORAGE(DNNL_ARG_BIAS);
 
     auto pd_ = pd();
-    const auto *src_md = ctx.input(DNNL_ARG_SRC)->md();
-    const auto *wei_md = pd()->weights_md();
-    const auto *dst_md = ctx.output(DNNL_ARG_DST)->md();
-    const memory_desc_t *src_scales_md = nullptr;
-    const memory_desc_t *wei_scales_md = nullptr;
+    const memory_desc_t *src_md = ctx.input(DNNL_ARG_SRC)->md();
+    const memory_desc_t *wei_md = pd()->weights_md();
```
h-sadia, line 473: are weights not part of ctx.input? or are these being tiled for MoE?

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -120,53 +120,55 @@ grouped_micro_gemm(const global SRC_DATA_T *src, int ldsrc,
...
 #if WITH_WEI_ATTR_ZP
-    wei_attr_zp += batch * ldweiq * (k / WEI_GROUP_SIZE);
+    wei_attr_zp += batch * n * (k / WEI_GROUP_SIZE);
 #endif
 
-    ugemm_grouped_c_type c_tile = ugemm_grouped(
-                                                wei, ldwei, src, ldsrc, m, n, k, wg_i0, wg_j0, 0, sg_i, sg_j, slm
-                                                SRC_ATTR_SCALE_ARGS SRC_ATTR_ZP_ARGS SRC_ATTR_LD_ARGS
-                                                WEI_ATTR_SCALE_ARGS WEI_ATTR_ZP_ARGS WEI_ATTR_LD_ARGS);
+    ugemm_grouped_c_type c_tile = ugemm_grouped(wei, ldwei, src, ldsrc, n, m, k,
```
h-sadia, line 197: ldwei -> wei_strides?

Nit: Generally, I have noticed we have lda, ldb, ldc in our gemm/matmul/rnn code. Having ldsrc, ldwei, lddst although is more user friendly - but breaks the similarity with the rest of the library.

h-sadia:
Just wondering how will the scheduling work for grouped GEMM? Are we using any persistent gemm kernels? Would be nice to have the link to RFC in the description as well. Thanks!

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
umar456: We run into this issue when we are searching for configs for SDPA, especially for smaller sizes. This allows us to use a wider variety of configurations when performing these searches.

We define float1 in [generic_vector_ops.h](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/include/generic_vector_ops.h#L22).

The tiles use both the vector dimension as well as the array indexing to define a tile. See: https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/include/tile_ops.h#L647

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: I noticed that there were several microkernel strategies that performed that conversion as part of the strategy(https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/gemm/jit/selector/db/kernel.db#L662-L671) but it seems like these strategies are not part of the ukernel catalog. I will remove this.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: Ideally we will never need to reach this point because we can expand the catalog to avoid this fallback.

The heuristic based approach still relies on these requirements to setup the work group tile shape. This is another layer to make the size as big as possible. If we were to move it internally the SDPA implementation may become oversized.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
Simonsays095: There are 3 cases where we require internal upconversions currently:
1. Grouped zero-points: e.g. `[FO]OS`
2. int8 with grouped scales and zero-points: e.g. `[OH][OH]S`
3. f32 with fpmath-mode=bf16: e.g. `[SB][SB]S`

In all of these cases, there are equivalents without upconversion that are used when the relevant attributes aren't used. We'll probably need to have similar logic here eventually, but only when those attributes require it.

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
Simonsays095: `shim.cpp` is a generic interface for designing microkernels, I don't think it should rely on definitions in oneDNN code in `generic_vector_ops.h`. Instead of changing this, can we change the implementation to define the tiles as `float1` instead of `float`?

Edit: It looks like `DECLARE_2D_TILE` creates a vector type, then creates the tile as an array of these vector types. It should never create a tile that isn't an array of vectors - in what cases do we face compilation issues if this change is excluded?

umar456: I have addressed this issue in my other PR #4655. Basically I replaced the struct generation with the DECLARE_2D_TILE that is defined in the tile_ops.h header. The shim was already using the DECLARE_2D_TILE_OPS macro and DECLARE_2D_TILE defines the OPS along with the type.

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
umar456: Addressed in #4655

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: These are grouped inputs so they have the same types.

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
enable benchdnn_gemm
disable test_device_cpu
enable test_device_gpu

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -383,13 +401,19 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
         if (systolic)
             s.ka_load = (problem.A.layout == MatrixLayout::T) ? (64 / problem.Ta_ext) : 16;
         s.slmA = true;
-
     } else if (problem.A.layout == MatrixLayout::T) {
         s.A.accessType = AccessType::Block2DTranspose;
-        s.ka_load = 64 / problem.Ta_ext;
+        s.ka_load = 64.f / ceil(( 1.f * problem.Ta_ext ) +
+                                (problem.aOffset2D() ? (1.f * problem.Tao) : 0));
```
syurkevi, line 434: This is targeting fp4/s4? A comment would be helpful.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: The Weights are usually fixed and do not change between calls. We don't need to use runtime shapes for this tensor.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: The kernel can use any combination of these types

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: They are different. The src input is coming in as a grouped memory tensor which has a different layout. Weights are regular memory objects. I have updated it because we only support one type of src layout at the moment(per-token).

src/gpu/intel/matmul/grouped_micro_gemm.cl
umar456: It requires stronger restrictions of the layout of src and wei. This is something I can add in a future PR.

umar456:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,245 @@
...
+grouped_micro_gemm(const global SRC_DATA_T *src, int ldsrc,
+        const global WEI_DATA_T *wei, long4 wei_strides, global DST_DATA_T *dst,
+        int lddst, const global int *src_offsets, const global int *dst_offsets,
+        const global SRC_SCALES_DATA_T *src_attr_scales,
+        const global SRC_ZP_DATA_T *src_attr_zp, const int ldsrcq,
+        const global WEI_SCALES_DATA_T *wei_attr_scales,
+        const global WEI_ZP_DATA_T *wei_attr_zp, const int ldweiq, int n, int k,
+        const global BIA_DATA_T *bias) {
+#if ugemm_grouped_slm_size > 0
+    local char slm[ugemm_grouped_slm_size];
```
syurkevi: slm[ MAX(ugemm_grouped_slm_size, 1) ];

src/gpu/intel/include/tile_ops.h
```diff
@@ -275,6 +275,17 @@ DEF_BLOCK2D_LOAD_STORE(float, uint, 8, 16, u32_m8k16v1, 16, 8)
                 = __builtin_convertvector(t.x[i], __typeof__(t_new.x[i])); \
     } while (0)
 
+#define tile_convert(t, t_new, f) \
```
syurkevi: nit: `conversion_func` to match reblock further down

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,561 @@
...
+status_t grouped_micro_gemm_t::init_microkernels(impl::engine_t *engine) {
+    using namespace jit;
+    using namespace gemmstone;
+    using namespace gemmstone::microkernel;
+    using gemm::jit::convert_dnnl_to_kernel_type;
+
+    assert(engine->kind() == engine_kind::gpu);
+    auto *intel_engine = utils::downcast<intel::engine_t *>(engine);
+    auto *dev_info = intel_engine->device_info();
+    //arch_ = dev_info->gpu_arch();
```
syurkevi: Needed? Defaults to unknown_

umar456:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

umar456:
make test perf-gpu
set primitive=matmul sdpa

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,561 @@
...
+    lws[0] *= sg_per_wg_m;
+    lws[1] *= sg_per_wg_n;
+    lws[2] *= sg_per_wg_k;
+
+    compute::range_t gws = lws;
+    // Swap wg_tile_[mn]_ for col-major vs row-major representations
+    gws[0] *= utils::div_up(n, wg_tile_m);
+    gws[1] *= utils::div_up(m_all, wg_tile_n);
+    gws[2] *= num_groups;
+    //std::cout << "LWS: " << lws.str() << std::endl;
```
syurkevi: stray comment

src/gpu/intel/matmul/grouped_micro_gemm.hpp
```diff
@@ -0,0 +1,101 @@
...
+        compute::gpu_arch_t arch_ = compute::gpu_arch_t::unknown;
+    };
+    status_t init(impl::engine_t *engine) override;
+    status_t init_microkernels(impl::engine_t *engine);
+
+    status_t execute(const exec_ctx_t &ctx) const override;
+
+    const pd_t *pd() const { return (const pd_t *)primitive_t::pd().get(); }
+
+    gemmstone::microkernel::Package gemm;
```
syurkevi: `gemm_` private member would be easier to ctrl+f

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+    problem.Ta = problem.Ta_ext;
+    problem.Tb = problem.Tb_ext;
+
+    problem.A.setAlignment(
+            alignmentForLD(static_cast<int>(gemm_desc_t::get_ld(*wei_mdw.md_))
+                    * problem.Ta_ext));
+    problem.B.setAlignment(
+            alignmentForLD(static_cast<int>(pd()->K()) * problem.Tb_ext));
+    problem.C.setAlignment(problem.Tc.size());
```
syurkevi, line 80: Does setting C alignment help w/performance?  Why not ldc? What is the internal default?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -0,0 +1,561 @@
...
+        problem.aqGroupK = utils::rnd_up_pow2(pd()->wei_group_sizes_[1]);
+    }
+
+    if (opts.scaleB || opts.offsetB) {
+        problem.bqGroupN = pd()->src_group_sizes_[0];
+        problem.bqGroupK = static_cast<int>(
+                utils::rnd_up_pow2(pd()->src_group_sizes_[1]));
+    }
+
+    // internal conversions do not work well when both A and B are integers
```
syurkevi, line 144: 👍

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,245 @@
...
+#define WEI_SCALE_ARGS \
+    OPTIONAL(AND(WITH_WEI_SCALES, WEI_SCALES_GROUPED), wei_attr_scales)
+#define WEI_ZP_ARGS OPTIONAL(WITH_WEI_ZP, wei_attr_zp)
+#define WEI_LD_ARGS OPTIONAL(OR(WITH_WEI_ZP, WEI_SCALES_GROUPED), ldweiq)
+
+void store_results(ugemm_grouped_c_type *tile, global DST_DATA_T *ptr, int n,
+        int m, int lddst, int sg_i0, int sg_j0) {
+#if DST_DT_F32
+    tile_store(*tile, ptr, n, m, lddst, sg_i0, sg_j0);
+    //tile_store_t_block2d(c_tile, dst, n, m, lddst, sg_j0, sg_i0);
```
syurkevi, line 76: Stray 2d block stores

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -0,0 +1,245 @@
...
+DECLARE_2D_TILE(c_tile_type_dst, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_grouped_c_type_block0, ugemm_grouped_c_type_block1,
+        ugemm_grouped_c_type_nblock0, ugemm_grouped_c_type_nblock1)
+#endif
+
+#define OPTIONAL(enabled, argName) CONCAT2(OPTIONAL_, enabled)(argName)
+#define OPTIONAL_0(argName)
+#define OPTIONAL_1(argName) , argName
+
+#define OR_RESULT00 0
```
syurkevi: Can we move these to include/utils.h, these would really clean up sdpa!

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: I don't think it matters because its a register. Alignment would be necessary for global loads.

syurkevi (approved):

umar456:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul

umar456:
make test perf-gpu
set primitive=matmul sdpa

umar456:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
enable arch_gpu_xe-hpc
disable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
disable arch_gpu_xe-lp
disable arch_gpu_xe-lpg
disable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

umar456:
make test perf-gpu
set primitive=sdpa

PR #4506, opened 2026-01-07
fixup: api: extend dropout attribute with offset host_scalars and s64

dzarukin:
[MFDNN-14517](https://jira.devtools.intel.com/browse/MFDNN-14517)

Basically a revert of "optimization" for POD types.

For some reason (I haven't dug up enough to discover) copying underlying `nibble_t` for int4 types leads to double free. I guess it somehow connects to `Implicitly creates objects at dest.` point described in std::memcpy documentation.

### DIFF
--- a/tests/benchdnn/dnnl_memory.cpp
+++ b/tests/benchdnn/dnnl_memory.cpp
@@ -177,18 +177,19 @@ int execute_reorder(const dnn_mem_t &src, dnn_mem_t &dst,
         if (dnnl_memory_desc_equal(src.md_, dst.md_)) {
             // If fail to create reorder pd, use plain data copy for identical
             // mds.
+            //
+            // For unknown reason using memcpy with int4 data types leads to
+            // the double corruption error. Stick for per element copy for now.
             BENCHDNN_PRINT(2, "%s\n", "[REORDER] Fallback to plain copy.");
             const int64_t chunk_size = 64;
             const int64_t n_chunks = div_up(src.nelems(), chunk_size);
-            const int64_t size_dt = src.sizeof_dt();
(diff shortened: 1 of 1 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -177,18 +177,19 @@ int execute_reorder(const dnn_mem_t &src, dnn_mem_t &dst,
...
+            //
+            // For unknown reason using memcpy with int4 data types leads to
+            // the double corruption error. Stick for per element copy for now.
             BENCHDNN_PRINT(2, "%s\n", "[REORDER] Fallback to plain copy.");
             const int64_t chunk_size = 64;
             const int64_t n_chunks = div_up(src.nelems(), chunk_size);
-            const int64_t size_dt = src.sizeof_dt();
             benchdnn_parallel_nd(n_chunks, [&](int64_t idx_chunk) {
                 int64_t idx_start = idx_chunk * chunk_size;
                 int64_t idx_end = MIN2(idx_start + chunk_size, src.nelems());
```
mzhukova, line 188: just for my own education, is this due to 2 elements per byte, but `nelems` returning actual number of elements? (so we should maybe divide by 2 somewhere..)

mzhukova (approved):

tests/benchdnn/dnnl_memory.cpp
dzarukin: That's one thing, but even if it's written correctly, the type that is copied is not POD but a custom struct, and it somehow messes up with allocation.

PR #4507, opened 2026-01-07
xe: disable error_stream_t while in an exception

rjoursler:
Fixes [MFDNN-14519](https://jira.devtools.intel.com/browse/MFDNN-14519) which is caused by an assertion triggered during an exception.

### DIFF
--- a/src/gpu/intel/gemm/jit/dsl/utils/utils.hpp
+++ b/src/gpu/intel/gemm/jit/dsl/utils/utils.hpp
@@ -101,7 +101,7 @@ private:
     ostringstream_t out_;
 };
 
-#if !defined(NDEBUG) || defined(GEMMSTONE_ASSERTIONS)
+#if !defined(NDEBUG) || GEMMSTONE_ASSERTIONS
 #define dsl_assert(cond) \
     !(cond) && gemmstone::dsl::error_stream_t(__FILE__, __LINE__, #cond)
 #else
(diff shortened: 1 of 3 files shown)


rjoursler:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv

src/gpu/intel/gemm/jit/dsl/utils/utils.hpp
```diff
@@ -65,48 +65,40 @@ std::unique_ptr<T> make_unique(Args &&...args) {
...
+#if __cplusplus < 201703L || (defined(_MSVC_LANG) && _MSVC_LANG < 201703L)
+        if (std::uncaught_exception()) {
+#else
+        if (std::uncaught_exceptions()) {
+#endif
+            return;
+        }
 
-        printf("%s\n", data_->out.str().c_str());
+        printf("%s\n", out_.str().c_str());
```
Simonsays095, line 92: Is there any harm in keeping the print, but just excluding the `throw std::runtime_error`?

echeresh:
@rjoursler The Jira description looks like `dsl_assert()` is enabled in the release build. Should we fix this as well?

src/gpu/intel/gemm/jit/dsl/utils/utils.hpp
rjoursler: I don't think it makes sense to keep this print. Generally, the reason for nested exceptions is an assertion hit in a destructor during stack unwinding. Since we already know an error occurred, there is no reason to believe structure invariants are correct. As such, this log is likely to be irrelevant.

rjoursler:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv

PR #4508, opened 2026-01-08
cmake: look for OpenCL only once

spalicki:
This change fixes 2 small issues:
1. Tests and examples will both look for OpenCL which is redundant
2. When rerunning CMake it might pickup cached `OpenCL_INCLUDE_DIR` from tests/examples and use it in main library.


spalicki:
make test
disable test_device_cpu
disable test_device_gpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

examples/CMakeLists.txt
```diff
@@ -24,12 +24,16 @@ append(CMAKE_CXX_FLAGS "${CMAKE_EXAMPLE_CCXX_FLAGS}")
 
 # Link the examples to the OpenCL runtime
 if(DNNL_GPU_RUNTIME STREQUAL "OCL")
-    find_package(OpenCL REQUIRED)
-    if(OpenCL_FOUND)
-        add_definitions(-DCL_TARGET_OPENCL_VERSION=120)
-        list(APPEND EXTRA_SHARED_LIBS OpenCL::OpenCL)
+    add_definitions(-DCL_TARGET_OPENCL_VERSION=120)
+    if(OpenCL_LIBRARY)
+        list(APPEND EXTRA_SHARED_LIBS ${OPENCL_LIBRARY})
```
dzarukin, line 29: Would it be helpful if `message(STATUS "Using user provided OpenCL library from '${OpenCL_LIBRARY}'")` is added here?

dzarukin (approved):

examples/CMakeLists.txt
spalicki: Yes, the problem is that this variable can be set as cached variable, ENV variable and -D variable. I know how to test for the first 2 but not the 3rd.

openmodelzooci:

vpirogov:
These changes are not relevant anymore.

PR #4509, opened 2026-01-08
graph: support sdpa/gqa training forward and backward with dropout

ElaineBao:
# Description

MFDNN-14009.


ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,61 @@
+Dropout {#dev_guide_op_dropout}
+============================
```
TaoLv:
```suggestion
===============================
```

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,61 @@
...
+
+## General
+
+Dropout operation is a regularization technique that randomly zeroes elements of
+the input tensor during training, scaling the outputs to maintain the expected
+sum. It is defined as:
```
TaoLv: What's the "expected sum"? Isn't it the Expectation of src and dst?

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,61 @@
...
+  dst_i = \frac{src_i \cdot mask_i}{1-p}
+  \f]
+  where \f$mask\f$ is a tensor of the same shape as \f$src\f$.
+
+- **Forward (Inference):**
+  \f[
+  dst_i = src_i
+  \f]
+
+- **Backward:**
```
TaoLv: Is it needed to define the dropout backward op and move this definition to it?

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,61 @@
...
+  given gradient \f$diff\_dst\f$ and \f$mask\f$:
+  \f[
+  diff\_src_i = \frac{diff\_dst_i \cdot mask_i}{1-p}
+  \f]
+
+The Dropout operation uses the Philox algorithm for random number generation,
```
TaoLv: Add a reference link for the Philox algorithm?

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,61 @@
...
+when constructing an operation.
+
+### Inputs
+
+| Index | Argument Name | Description                | Required or Optional |
+|:------|:--------------|:---------------------------|:---------------------|
+| 0     | `src`         | Input tensor               | Required             |
+| 1     | `seed`        | Random seed                | Required             |
+| 2     | `offset`      | Random offset              | Required             |
+| 3     | `probability` | Dropout probability        | Required             |
```
TaoLv, line 45: Is it intended to have seed/offset/prob as arguments which are not available during compilation and kernel generation?

TaoLv: Also are they single value tensors or scalar tensors?

src/graph/interface/logical_tensor.cpp
```diff
@@ -237,10 +237,10 @@ status_t DNNL_API dnnl_graph_logical_tensor_init_with_dims(
     if (ptype == property_type::host_scalar) {
         VCHECK_LOGICAL_TENSOR(ndims == 0, status::invalid_arguments,
                 "host scalar tensor should have zero ndims");
-        VCHECK_LOGICAL_TENSOR(
-                utils::one_of(dtype, data_type::s32, data_type::f32),
+        VCHECK_LOGICAL_TENSOR(utils::one_of(dtype, data_type::s32,
+                                      data_type::f32, data_type::s64),
                 status::invalid_arguments,
-                "host scalar tensor should have s32 or f32 dtype");
+                "host scalar tensor should have s32, f32 or s64 dtype");
```
TaoLv, line 243: Is this to support host scalar seed/offset/proba?

doc/graph/operations/Dropout.md
ElaineBao: expected sum means after scaling, the expected value of the sum of the output is almost equal to the sum of the input

doc/graph/operations/Dropout.md
ElaineBao: That is also an option, but previously the discussion in RFC tends to add one op, since they do the same math.

doc/graph/operations/Dropout.md
ElaineBao: Yes, previous discussion ([link](https://github.com/uxlfoundation/oneDNN/pull/3939#discussion_r2358118741)) concludes those should be inputs.
They can be single value tensors and scalar tensors from the document, but the current implementation in graph only supports host scalar.

src/graph/interface/logical_tensor.cpp
ElaineBao: this is to support host scalar seed and offset, as they are s64. prob is f32.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

doc/graph/operations/Dropout.md
TaoLv: Can you please add these requirements to the document?

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,63 @@
...
+  \f]
+
+- **Backward:**
+  given gradient \f$diff\_dst\f$ and \f$mask\f$:
+  \f[
+  diff\_src_i = \frac{diff\_dst_i \cdot mask_i}{1-p}
+  \f]
+
+The Dropout operation uses the [Philox
+algorithm](http://www.thesalmons.org/john/random123/papers/random123sc11.pdf)
```
TaoLv: I guess the link will not be correctly rendered.

src/graph/backend/dnnl/passes/memory_planning.cpp
```diff
@@ -44,6 +44,23 @@ namespace dnnl_impl {
...
+        return dnnl::memory(md, (float)0);
+    } else if (dt == dnnl::memory::data_type::s32) {
+        return dnnl::memory(md, (int32_t)0);
+    } else if (dt == dnnl::memory::data_type::u8) {
+        return dnnl::memory(md, (uint8_t)0);
+    } else if (dt == dnnl::memory::data_type::s8) {
+        return dnnl::memory(md, (int8_t)0);
+    } else {
+        return dnnl::memory(md, 0);
+    }
```
TaoLv: I think you can also use the macro `DNNL_HOST_SCALAR_TYPE_SWITCH` here, like:

```
DNNL_HOST_SCALAR_TYPE_SWITCH(md.get_data_type(), DType, {
                return dnnl::memory(md, static_cast<DType>(0));
            });
```

src/graph/backend/dnnl/passes/memory_planning.cpp
```diff
@@ -160,7 +177,7 @@ std::shared_ptr<execution_args_set_t> execution_args_set_t::clone() const {
         memory cloned_mem;
         if (val_mem.second.get_desc().get_format_kind()
                 == dnnl::memory::format_kind::host_scalar) {
-            cloned_mem = dnnl::memory(val_mem.second.get_desc(), 0);
+            cloned_mem = make_dnnl_memory_scalar(val_mem.second.get_desc());
```
TaoLv: what's the problem here?

src/graph/backend/dnnl/passes/memory_planning.cpp
TaoLv: why do we need u8/s8 here?

src/graph/backend/dnnl/patterns/matmul_post_ops.cpp
```diff
@@ -90,6 +90,20 @@ DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, fp_matmul_post_ops)
             return std::make_shared<float_matmul>();
         });
 
+DNNL_BACKEND_REGISTER_PATTERN_MATCHER_PASS(dnnl, fp_matmul_dropout)
```
TaoLv, line 93: These eltwise/matmul + dropout fusions are not required by SDPA training, right?

src/graph/backend/dnnl/passes/memory_planning.cpp
ElaineBao: we need to make sure the memory size for different dtypes are correct.

src/graph/backend/dnnl/passes/memory_planning.cpp
ElaineBao: we don't need u8/s8, they are redundant. removed and updated to use the macro.

src/graph/backend/dnnl/patterns/matmul_post_ops.cpp
ElaineBao: yes, correct, they are just added for coverage and correctness.

doc/graph/operations/Dropout.md
ElaineBao: removed the link to align with primitive document.

doc/graph/operations/Dropout.md
ElaineBao: added implementation limitation section.

doc/graph/operations/Dropout.md
```diff
@@ -0,0 +1,68 @@
...
+|:------|:------|:-----|:------|:-------|:------------|
+| f32   | f32   | u8   | u64   | u64    | f32         |
+| bf16  | bf16  | u8   | u64   | u64    | f32         |
+| f16   | f16   | u8   | u64   | u64    | f32         |
+
+## Implementation Limitations
+
+1. The `seed`, `offset`, and `probability` inputs only support logical tensors
+   with property type set to `host_scalar`.
+2. The optional `mask` output is not supported.
```
TaoLv: Can you please explain more about the `mask` output here? We even don't support it in the reference implementation of dropout? If it's not requested and we don't have plan to support it in short term, can we remove it from the API definition?

src/graph/backend/dnnl/patterns/matmul_post_ops.cpp
TaoLv: What do you mean by "correctness"? If they are not must to have in this PR, I'm thinking if we can move them to a follow-up PR. Just try to make this PR more focused on SDPA and moderate the change size for better review.

doc/graph/operations/Dropout.md
TaoLv: Can you please share any reference link about this description? Can we align the description with dropout attribute: https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_dropout.html ?

src/graph/backend/dnnl/patterns/matmul_post_ops.cpp
ElaineBao: to make sure the matmul/eltwise primitive+attribute can work correctly after integrating to graph API, which is a preliminary for sdpa

doc/graph/operations/Dropout.md
ElaineBao: We haven't supported mask output because it's currently not used in sdpa training - for backward, the same seed and offset are used to guarantee the reproductivity,
The reason I didn't remove it is that primitive supports the mask output, so implementation-wise it's doable. And previously the proposed definition in the RFC has this output.

TaoLv: If so, I suggest:
- Support the optional mask output through the single dropout op implementation. With that we can remove this note from the document.
- Add a check and verbose log in SDPA pattern if mask output is not supported there.

ElaineBao: There's no single dropout op implementation, as primitive supports it as a primitive attribute. So if support, it means supporting by matmul/eltwise+dropout (with mask output), which is not the focus of this PR. 

If it's really a concern, I think a separate PR is better. BTW, why do you think it is a concern?

TaoLv: It looks like we are adding/exposing a feature (dropout mask output) without any request and implementation.

ElaineBao: ok, let me just remove it, we can add if it's needed in the future.

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

tests/benchdnn/graph/deserialize.cpp
```diff
@@ -871,7 +879,9 @@ const deserialized_op_t &deserialized_graph_t::find_next_until(
         const std::unordered_set<std::string> &allowed_skips) const {
     const deserialized_op_t *cur_op_ptr = &start_op;
     while (!cur_op_ptr->empty() && cur_op_ptr->kind_ != target_kind) {
-        if (!allowed_skips.empty() && !allowed_skips.count(cur_op_ptr->kind_)) {
+        if (allowed_skips.empty()
+                || (!allowed_skips.empty()
+                        && !allowed_skips.count(cur_op_ptr->kind_))) {
```
dzarukin:
```suggestion
        if (IMPLICATION(!allowed_skips.empty(), !allowed_skips.count(cur_op_ptr->kind_))) {
```

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -14,8 +14,10 @@
 * limitations under the License.
 *******************************************************************************/
 
-#include "setting_handler.hpp"
+#include <random>
+
 #include "graph/utils.hpp"
+#include "setting_handler.hpp"
```
dzarukin:
```suggestion
#include "graph/utils.hpp"
#include "graph/setting_handler.hpp"

#include <random>
```

Putting system headers below helps to identify through compilation if there're dependencies on <random> in headers above.

tests/benchdnn/graph/setting_handler.cpp
```diff
@@ -91,6 +93,35 @@ bool get_graph_attr(const deserialized_op_t &base_op_ref,
...
+bool get_graph_attr(
+        const deserialized_op_t &base_op_ref, attr_t::dropout_t &arg_dropout) {
+
+    if (base_op_ref.kind_ != "Dropout") return true;
+
+    // Generate a random seed
+    std::minstd_rand seed1(1);
+    std::uniform_int_distribution<> gen1(1, 100);
+    int64_t seed = gen1(seed1);
+    arg_dropout.seed = seed;
```
dzarukin: I'd stick to a fixed number of 123456789 or 987654321. The less random it is, the easier debugging will be.

Same for offset. Maybe even keep it 0 at all...

Same for p. This can actually help with better accuracy dropping more and keeping less non-zeros, I'd post a note on potential filling control based on that value.

ElaineBao: changed to fixed seed/offset/p. thanks!

ElaineBao:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4514, opened 2026-01-08
[GPU] Add dropout to jit path for matmul

h-sadia:
This is the last part of enabling dropout as request from GRAPH API. This will enable SDPA on the fast path.

### DIFF
--- a/src/gpu/intel/gemm/host_scalars.hpp
+++ b/src/gpu/intel/gemm/host_scalars.hpp
@@ -109,6 +109,30 @@ inline status_t maybe_get_host_scalar_value(
     return status;
 }
 
+inline status_t maybe_get_host_scalar_value(
+        const memory_storage_t &mem_storage, int64_t &scalar_value) {
+    using namespace data_type;
+    const host_scalar_memory_storage_t *scalar_storage
+            = utils::downcast<const host_scalar_memory_storage_t *>(
+                    &mem_storage);
+    status_t status = status::success;
+
+    switch ((int)scalar_storage->data_type()) {
+        case s64:
+            status = get_scalar_value<int64_t, int64_t>(
+                    scalar_value, scalar_storage);
+            break;
+        case s32:
+            status = get_scalar_value<int32_t, int64_t>(
+                    scalar_value, scalar_storage);
+            break;
+        default:
+            assert(!"Support for requested data type is missing for host-side "
+                    "scalars");
+            status = status::invalid_arguments;
+    }
+    return status;
+}
 } // namespace gemm
 } // namespace intel
 } // namespace gpu


src/gpu/intel/gemm/with_post_ops.cpp
```diff
@@ -215,6 +238,10 @@ status_t with_post_ops_t::pd_t::init_kernel_ctx(
     const bool with_wei_scales
             = !attr_scales.has_default_values(DNNL_ARG_WEIGHTS);
     const bool with_dst_scales = !attr_scales.has_default_values(DNNL_ARG_DST);
+    const bool with_dropout = !attr()->dropout_.has_defaul_values();
```
Simonsays095:
```suggestion
    const bool with_dropout = !attr()->dropout_.has_default_values();
```

src/gpu/intel/gemm/with_post_ops.cpp
```diff
@@ -234,6 +261,10 @@ status_t with_post_ops_t::pd_t::init_kernel_ctx(
     kernel_ctx.define_int("A_SCALES", with_src_scales);
     kernel_ctx.define_int("B_SCALES", with_wei_scales);
     kernel_ctx.define_int("C_SCALES", with_dst_scales);
+    kernel_ctx.define_int("WITH_DROPOUT", with_dropout);
+    kernel_ctx.define_int("USE_OFFSET", use_offset);
+    kernel_ctx.define_int("USE_HOST_SCALARS", use_host_scalars);
+    kernel_ctx.define_int("HAS_OUTPUT_MASK", has_output_mask);
```
Simonsays095: Can you name these `DROPOUT_*` so it's clearer in the kernel what's being controlled by these macros? Specifically, `USE_HOST_SCALARS` is confusing as there are other scalars that may be supplied on the host or device side.

src/gpu/intel/gemm/with_post_ops.cpp
```diff
@@ -304,6 +335,12 @@ status_t with_post_ops_t::execute(const exec_ctx_t &ctx) const {
     arg_list.set(idx++,
             pd()->attr()->scales_.get_mask(DNNL_ARG_WEIGHTS) > 0 ? 1 : 0);
     arg_list.set(idx, GEMM_CTX_ARG_STORAGE(c_zero_point));
+    if (pd()->with_dropout) {
+        idx++;
+        arg_list.set(idx++, GEMM_CTX_ARG_STORAGE(dropout_seed));
+        arg_list.set(idx++, GEMM_CTX_ARG_STORAGE(dropout_offset));
+        arg_list.set(idx, GEMM_CTX_ARG_STORAGE(dropout_prob));
```
Simonsays095: I think these need to be set in `gpu/intel/matmul/gemm.cpp` similarly to the other `gemm:exec_args_t` values.

src/gpu/intel/gemm/with_post_ops.cpp
h-sadia: They all actually come withing the WITH_DROPOUT macro, so it is going to be pretty clear. But I added the DROPOUT_* to HOST_SCALARS for clarity

src/common/gemm_pd.hpp
```diff
@@ -36,6 +36,10 @@ namespace impl {
     VCHECK(primitive, create, dispatch, gemm, (f), "%s," msg, \
             this->info(engine), ##__VA_ARGS__)
 
+#define VDISPATCH_GEMM_IC(cond, msg, ...) \
+    VCONDCHECK(primitive, create, dispatch, gemm, (cond), \
+            status::unimplemented, msg, ##__VA_ARGS__)
```
Simonsays095, line 41: Why not just use `VDISPATCH_GEMM`?

h-sadia: Just following the pattern that Dmitry has set for all primitives - eltwise, softmax, matmul

kealan-barbieri (approved):

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc

src/common/gemm_pd.hpp
Simonsays095: I just don't know how the verbose line for `VDISPATCH_GEMM_IC` and `VDISPATCH_GEMM` look. I guess `IC` stands for "internal check"? And it's missing some extra information from `this->info()`?

dzarukin: It's not "missing", but intentionally avoiding it because at the level of calling these verbose calls there's no access to `pd()` or `this`.

src/gpu/intel/eltwise/xe.hpp
```diff
@@ -71,6 +71,8 @@ struct xe_fwd_t : public primitive_t {
                     VERBOSE_INCONSISTENT_DT, "src", "dst");
             VDISPATCH_ELTWISE(
                     attr()->has_default_values(), VERBOSE_UNSUPPORTED_ATTR);
```
dzarukin, line 73: This call should cover dropout, too. If it doesn't, it should be adjusted instead. Same applies to softmax changes.

src/gpu/intel/gemm/host_scalars.hpp
```diff
@@ -109,6 +109,31 @@ inline status_t maybe_get_host_scalar_value(
...
+    switch ((int)scalar_storage->data_type()) {
+        case s64:
+            status = get_scalar_value<int64_t, int64_t>(
+                    scalar_value, scalar_storage);
+            break;
+        case s32:
+            status = get_scalar_value<int32_t, int64_t>(
+                    scalar_value, scalar_storage);
+            break;
+            break;
```
dzarukin: Double break?

src/gpu/intel/gemm/with_post_ops.cl
```diff
@@ -103,5 +114,25 @@ __kernel void gemm_post_ops(__global SRC_DATA_T *src,
...
+    long dropout_offset = DROPOUT_USE_OFFSET ? dropout_offset_buf[0] : 0;
+    float dropout_p = dropout_p_buf[0];
+#endif
+    uint dropout_threshold = get_dropout_threshold(dropout_p);
+    float dropout_inv_q = (dropout_p != 1.f) ? 1.f / (1.f - dropout_p) : 0.f;
+#if DROPOUT_USE_OFFSET
+    uint res = philox_4x32_s64(
+            data_idx, (ulong)dropout_seed, (ulong)dropout_offset);
+#else
+    uint res = philox_4x32((uint)data_idx, (uint)dropout_seed);
```
dzarukin, line 129: What if there's `long` seed and no offset?

src/gpu/intel/gemm/host_scalars.hpp
h-sadia: typo

src/gpu/intel/eltwise/xe.hpp
h-sadia: you mean the has_default_values() function?

src/gpu/intel/gemm/with_post_ops.cl
h-sadia: it follows this path of else, and in the philox_4x32 I am using what we have on the CPU side (what Mourad had) where he is  caculating the seed in long, and offset and calling the philox_4_32_s64 function. Putting the offset as 0 causes correction issues, so offset does use some value based here: https://github.com/uxlfoundation/oneDNN/blob/main/src/common/math_utils.hpp#L508 
And the code for ref_dropout is similar as well, based on this: https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/primitive_attr_postops.cpp#L304

PR #4516, opened 2026-01-09
cpu: aarch64: remove redundant embd_bcast flag

jondea:
This flag appears to have been copied verbatim from x64, and it doesn't seem to apply to AArch64. Specifically, on x64, turning the flag on enables the use of fewer instructions. On AArch64, it uses _more_ instructions. The comment about when it is turned on also doesn't make sense, vector length is orthogonal to features.

Removing it to make it easier to add new functionality to `jit_brgemm_kernel_t`.

# Checklist

## General


jondea:
@Shreyas-fuj @vineelabhinav could you take a quick look at this PR please? Particulary how it relates to the performance on the A64FX.

Shreyas-fuj:
Sorry for the late response, I think, as mentioned above, the flag embd_bcast was maintained as is while porting from x86. The changes should not affect the performance of A64FX as well since there is no such parallel instruction to x86 in arm.

Issue #4089, opened 2025-10-08
Possible to generate xbyak_aarch64 exceptions with large kernels


Sqvid:
# Summary
This is a follow-on from #4055. There exist several cases in the code where it is possible for very large kernels to be generated; this in turn can cause `Xbyak_aarch64` to throw an exception when one tries to place a `Label` with a large jump address.

For example consider the following:
```
$ ./build/tests/benchdnn/benchdnn --conv --impl='brgconv:sve_128' --canonical=true --dt=bf16 --attr-post-ops=gelu_tanh g1ic64ih1000oc64oh1000kh3ph128dh127         

bad err=15 in Xbyak::Error
terminate called after throwing an instance of 'Xbyak_aarch64::Error'
  what():  illegal immediate parameter (range error)
zsh: abort (core dumped)
```
This gets thrown when underlying `CodeArray` grows large and you [try to place a label at the end](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp#L1759).
This is because the [`Xbyak_aarch64::LabelManager` tries to calculate](https://github.com/fujitsu/xbyak_aarch64/blob/59c6a7e7368cd2dbb07c46abe83239ace4d58849/xbyak_aarch64/xbyak_aarch64_label.h#L89) program-counter relative address, and if [the value of the immediate value is too large](https://github.com/fujitsu/xbyak_aarch64/blob/59c6a7e7368cd2dbb07c46abe83239ace4d58849/src/xbyak_aarch64_impl.h#L325) then the instruction is malformed.

Since this error can pop up in a variety of scenarios, throws an uncaught exception, and is unrelated to any particular kernel, I think we really need to find a good way to address it.

cc: @vpirogov, @dzarukin, @jondea, @Shreyas-fuj 

# Environment
oneDNN includes hardware-specific optimizations and may behave
differently on depending on the compiler and build environment. Include
the following information to help reproduce the issue:
* CPU make and model: Neoverse V1 (flags: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm paca pacg dcpodp svei8mm svebf16 i8mm bf16 dgh rng)
* git hash: bcc0ca00842739f54006d93917d807ea68616bc8

[...]

PR #4517, opened 2026-01-09
cpu: aarch64: remove branch target alignment

jondea:
Remove branch target alignment from brgemm functions, reducing code size by ~10% and improving performance on Graviton 3/4 by ~0.5% (this required many repeated measurements to be confident of results). Branch alignment can affect performance, but it does not need to be the default, and it should be machine and kernel specific and evidence based.

As another point of reference, arm_gemm in ComputeLibrary only uses nops in a single kernel `a64_sgemm_asimd_8x12_a53`, written for a core which we do not specifically target (and only then sparingly).

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2024-2025 FUJITSU LIMITED
-* Copyright 2024-2025 Arm Ltd. and affiliates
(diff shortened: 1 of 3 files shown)


Sqvid (approved):
Looks great, thanks! Did you see any impact wrt. https://github.com/uxlfoundation/oneDNN/issues/4089?

jondea:
@Sqvid I had a quick look at #4089 and it doesn't look like it has any effect. Specifically, I took your reproducer for `sve_256` and found the boundary of oh/ow where it happened, and it was the same with an without this commit. There may be another dimension where this does help though. It certainly _could_ help.

PR #4520, opened 2026-01-09
cmake: fix setting OCL version for host compilers

spalicki:
Small fix for SYCL builds with host compiler different than ICX.


spalicki:
make test
disable test_device_cpu
disable test_device_gpu

PR #4522, opened 2026-01-12
Fix gtests after async runtime

dzarukin:
[MFDNN-14549](https://jira.devtools.intel.com/browse/MFDNN-14549)

### DIFF
--- a/tests/gtests/test_sum.cpp
+++ b/tests/gtests/test_sum.cpp
@@ -89,10 +89,17 @@ private:
         const auto &dst_d = dst.get_desc();
         const auto dst_dims = dst_d.get_dims();
 
+        // Note: mapped pointers must be kept alive for XeLP and XeHPC platforms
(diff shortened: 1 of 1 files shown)


dzarukin:
make test

avmanerikar (approved):

tests/gtests/api/test_memory_creation.cpp
```diff
@@ -68,10 +68,13 @@ class memory_creation_test_t
 
         // Map-unmap section
         {
+            // Note: fill before mapping, otherwise, double-mapped (one here,
+            // one inside `fill_data` leads to hang in SYCL CPU runtime.
```
mzhukova:
```suggestion
            // Note: fill before mapping; otherwise, double mapping (one here,
            // one inside `fill_data`) leads to hang in SYCL CPU runtime.
```

mzhukova (approved):

PR #4529, opened 2026-01-12
benchdnn: cold cache improvement

dzarukin:
Jira: [MFDNN-13816](https://jira.devtools.intel.com/browse/MFDNN-13816)

LNL may report higher BW than HW provides for certain shapes (the last number is expected to be <= 108):
```
Output template: %prb%,%-time%,%-Gflops%,%-Gbw%
--mode=P --matmul --engine=gpu --cold-cache=all --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab --attr-scratchpad=user 1024x8:8x4096,0.026458,2536.43,320.15
```
The issue lies in the hardware behavior when in certain situations L3 cache won't be flushed and input memories will reside in there even for a cold-cache mode.
The resolution is to introduce a flushing kernel that is submitted after each execution for trigger that L3 cache flushing. The credit for idea. library implementation and pivoting of PoC goes to @echeresh.

Note: even though the flushing kernel does stabilize result significantly, it may happen there's a one or two hardware "shots" where the number would still be higher as reported above which is caused by unknown effect at this moment.
The work around those "shots" is to drop several fastest results from the timer collection
Finally, flushing kernel costs time but with better stability time which allows to decrease the sample size.
Three changes combined should address the "not-cold-enough" cold-cache issue.

```
Output template: %prb%,%-time%,%-Gflops%,%-Gbw%
--mode=F --matmul --engine=gpu --cold-cache=all --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab 1024x8:8x4096,0.08,838.861,105.882
--mode=P --matmul --engine=gpu --cold-cache=all --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab --attr-scratchpad=user 1024x8:8x4096,0.078437,855.577,107.991
```

### DIFF
--- a/tests/benchdnn/utils/cold_cache.cpp
+++ b/tests/benchdnn/utils/cold_cache.cpp
@@ -300,6 +300,11 @@ void cold_cache_t::flush_cache(dnnl_stream_t stream) const {
     std::call_once(flag, [&]() {
         SAFE_V(get_gpu_cache_size(flush_size));
         if (flush_size == 0) flush_size = (32 << 20); // 32 MB fallback
+        flush_size = std::min(flush_size * 3, (size_t)(400 * (1 << 20)));
+        // Allow overriding via CACHE_SIZE env variable (value in bytes).
+        const auto env_str = benchdnn_getenv_string("CACHE_SIZE");
+        if (!env_str.empty())
+            flush_size = strtoull(env_str.c_str(), nullptr, 0);
         const dnnl_dims_t dims = {(dnnl_dim_t)flush_size};
         flush_cache_memory() = dnn_mem_t(
                 1, dims, dnnl_s8, "a", get_test_engine(), /* prefill = */ true);


dzarukin:
make test perf-gpu
set primitive=reorder sum concat binary conv deconv pool

tests/benchdnn/utils/timer.cpp
```diff
@@ -92,6 +96,55 @@ void timer_t::stamp(int add_times) {
...
+            static_cast<size_t>(std::floor(outlier_percent * times_)));
+    std::vector<double> deltas(deltas_size);
+
+    std::string v;
+
+    // The major magnitude is when `x_i = x_{i+1} * 1.1`.
+    constexpr double magnitude_threshold = 1.1;
+    size_t major_magnitude = SIZE_MAX;
+    for (size_t i = 0; i < deltas.size(); i++) {
+        deltas[i] = ms_vec_[i + 1] / ms_vec_[i];
```
echeresh: Do we need to check bounds? Say, if `ms_vec_.size() == 1` and `deltas_size == 1`, this goes out of bounds.

echeresh: Another comment: do we really need `deltas` vector? Looks like its values are not used outside of the loop.

tests/benchdnn/utils/timer.cpp
```diff
@@ -92,6 +96,55 @@ void timer_t::stamp(int add_times) {
...
+            major_magnitude = i;
+            if (verbose >= 4) {
+                v += std::to_string(i) + ":" + std::to_string(ms_vec_[i + 1])
+                        + "/" + std::to_string(ms_vec_[i]) + "="
+                        + std::to_string(deltas[i]) + "; ";
+            }
+        }
+    }
+
+    if (major_magnitude < deltas.size()) {
```
echeresh:
```suggestion
    if (major_magnitude < deltas_size) {
```

tests/benchdnn/utils/timer.cpp
dzarukin: Thanks for catching, I removed deltas vector completely and removed unnecessary division.

tests/benchdnn/utils/timer.cpp
dzarukin: Applied, thanks.

dzarukin:
@Simonsays095 , just to let you know, I've put the same caching mechanism as for fast_fill kernel. Thanks for reminding.

dzarukin:
@echeresh, I've updated cold-cache parsing, re-directed "wei" mode to "all" and allowed bool input. Please take another look.

dzarukin:
make test

dzarukin:
make test perf-gpu
set primitive=reorder conv deconv pool matmul

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test
set test_scope=NIGHTLY

tests/benchdnn/utils/timer.cpp
```diff
@@ -92,6 +96,59 @@ void timer_t::stamp(int add_times) {
...
+            }
+        }
+    }
+
+    if (major_magnitude < deltas_size) {
+        for (size_t i = 0; i < major_magnitude; i++) {
+            ms_[mode_t::avg] -= ms_vec_[i];
+            ms_[mode_t::sum] -= ms_vec_[i];
+        }
+        ms_[mode_t::min] = ms_vec_[major_magnitude + 1];
```
echeresh, line 144:
```suggestion
        ms_[mode_t::min] = ms_vec_[major_magnitude];
```

tests/benchdnn/dnnl_common.cpp
```diff
@@ -681,28 +687,53 @@ inline int measure_perf_aggregate(timer::timer_t &t,
...
+            // the time needed for measurements.
+            //
+            // Do this before cold-cache init as it has reorders which update
+            // profiling results.
+            if (use_single_cycle) {
+                std::vector<uint64_t> v_nsecs, v_cycles;
+                SAFE(get_gpu_profiling_info(v_stream[j], v_nsecs, v_cycles, 1),
+                        CRIT);
+                double ms = v_nsecs[0] / 1e6;
+                double ms_warmup = std::min(ms, 0.0);
```
echeresh: Looks like `ms_warmup` is always zero, why do we need this `min`?

tests/benchdnn/utils/parser.cpp
```diff
@@ -567,8 +590,7 @@ cold_cache_input_t str2cold_cache_input(const std::string &s) {
         SAFE_V(FAIL);
```
echeresh: `parser_utils::parse_bool(mode_str)` returns either `true` or `false` - so we hit line 582 or 584. After that, this `else` is unreachable - so instead of the message can just add `assert(!"not expected");`.

src/gpu/intel/flush_cache.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    compute::range_t gws = {bytes / 64, 1, 1};
+    compute::range_t lws = {256, 1, 1};
+    compute::nd_range_t nd_range(gws, lws);
+    compute::kernel_arg_list_t arg_list;
+    arg_list.set(0, *data->memory_storage());
+
+    // Disable profiling when flushing the cache.
+    intel_stream->set_profiling(false);
+    CHECK(kernel.parallel_for(*stream, nd_range, arg_list,
```
echeresh: `set_profiling(true)` is not called when `parallel_for()` fails - this leaves the profiler in an unexpected state. It'd be nice to restore profiling or at least add an assertion - just in case, to catch a broken state earlier.

tests/benchdnn/utils/cold_cache.hpp
```diff
@@ -69,6 +67,8 @@ extern cold_cache_input_t cold_cache_input;
 
 const cold_cache_input_t &default_cold_cache_input();
 
+dnn_mem_t &flush_cache_memory();
+
 std::ostream &operator<<(std::ostream &s, cold_cache_mode_t cold_cache_mode);
```
echeresh: Should be dropped? I see that the definition was removed.

dzarukin:
make test

tests/benchdnn/utils/timer.cpp
dzarukin: This one looks correct to me, but the for loop had wrong exit condition not including the `magnitutude` idx itself. Updated that one. Thanks for spotting.

tests/benchdnn/dnnl_common.cpp
dzarukin: Indeed, removed `ms`, kept `ms_warmup` only without `min`. Thanks.

tests/benchdnn/utils/parser.cpp
dzarukin: I've update the function keeping bool parsing in the `else` branch, otherwise previous valid values might be recognized as bad ones. Thanks for bringing that up.

src/gpu/intel/flush_cache.cpp
dzarukin: That's a catch. I've replaced the logic to get the status and enable profiling unconditionally. Thanks.

tests/benchdnn/utils/cold_cache.hpp
dzarukin: Yes, removed, thanks.

TaoLv (approved):

tests/benchdnn/utils/timer.cpp
```diff
@@ -92,6 +96,59 @@ void timer_t::stamp(int add_times) {
     stop(add_times, ticks_now() - ticks_start_, ms_now() - ms_start_);
 }
 
+void timer_t::filter_collection() {
```
mzhukova, line 99: this doesn't do anything to ticks_ seems like, maybe document as intentional if no need to update those?

tests/benchdnn/utils/parser.cpp
```diff
@@ -670,7 +673,22 @@ cold_cache_input_t str2cold_cache_input(const std::string &s) {
     return c;
 }
 
-} // namespace parser_utils
+bool parse_bool(const std::string &s) {
```
mzhukova, line 676: I've noticed that `str2bool` ignored case (used `strcasecmp`), but this uses direct comparison, could this lead to any issues?

src/gpu/intel/flush_cache.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    return st;
+}
+
+} // namespace intel
+} // namespace gpu
+} // namespace impl
+} // namespace dnnl
+
+extern "C" dnnl::impl::status_t DNNL_API dnnl_impl_gpu_flush_cache(
```
mzhukova, line 80: is having DNNL_API makes it user visible? (does it mean we should document it and care about its API stability...?)

tests/benchdnn/utils/timer.cpp
dzarukin: `ticks_` (or frequency) is outdated feature (as it's impossible to guarantee its stability from the benchmark perspective), it's supposed to be removed (I'll do it sooner).

src/gpu/intel/flush_cache.cpp
dzarukin: It makes it visible from the library and it's possible to call it, yes.
We have quite a lot of such APIs designed for testing and we don't "officially" document them:
```
$ grep -r DNNL_API src/ | grep -v graph | wc -l
123
```
Some of them can be converted back if a better internals testing system is in-place, but some of them can't (like this one). There was as idea to make them half-official with a dedicated testing header, but lack of resources...

mzhukova: I see, thanks for the clarification

tests/benchdnn/utils/parser.cpp
dzarukin: No reason to break it, sure, thanks for spotting, updated.

dzarukin:
make test perf-gpu
set primitive=matmul

echeresh:
make test perf-gpu
set primitive=gpu:gemm

echeresh:
make test perf-gpu
set primitive=gpu:gemm

echeresh:
make test perf-gpu
set primitive=gpu:gemm

echeresh:
make test perf-gpu
set primitive=gpu:gemm

echeresh:
make test perf-gpu
set primitive=gpu:gemm
set tool_branch=echeresh/less-reruns

echeresh:
make test perf-gpu
set primitive=gpu:gemm
set tool_branch=echeresh/less-reruns

echeresh:
make test perf-gpu
set primitive=gpu:gemm
set tool_branch=echeresh/less-reruns

echeresh:
make test perf-gpu
set primitive=gpu:gemm
set tool_branch=echeresh/less-reruns

dzarukin:
A new flush kernel causes significant side effects, closing this PR.

PR #4532, opened 2026-01-13
[backport v3.11] Enable k-parallel GEMM microkernels

Simonsays095:
Part of [MFDNN-14526](https://jira.devtools.intel.com/browse/MFDNN-14526).

Enables microkernel support for k-parallel kernels, and adds a few such kernels to the full-kernel-like (MMR) catalog. Additionally adds verbosity + development features mirroring some features in the regular GEMM pathway.

Main branch PR will come immediately following this merge.

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/microkernel_provider.cpp
+++ b/src/gpu/intel/gemm/jit/generator/microkernel_provider.cpp
@@ -209,6 +209,25 @@ Package selectGEMMMicrokernel(GEMMProtocol protocol, HWInformation hwInfo, SizeP
 
     strategy.preflight(hw, problem);
 
+    if (getVerbose(gemmstone::GEMMVerbose::DebugInfo) >= 10) {
+        if (entry)
+            std::cout << "Selected microkernel catalog entry: " << entry->str() << std::endl;
+        else
+            std::cout << "Microkernel generated heuristically" << std::endl;
+    }
+    if (getVerbose(gemmstone::GEMMVerbose::DebugInfo) >= 2) {
+        std::string result = problem.toString();
+        result.append(" ");
+        result.append(std::to_string(strategy.unroll[LoopM]));
+        result.append(" ");
+        result.append(std::to_string(strategy.unroll[LoopN]));
+        result.append(" ");
+        result.append(problem.scalarsToString());
+        result.append(" ");
+        result.append(unparseStrategy(hw, problem, strategy));
+        std::cout << "Actual kernel: " << result << std::endl;
+    }
+
     /* Set up arguments for microkernel */
     InterfaceHandler interface(hw);
 


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/generator/microkernel_provider.cpp
```diff
@@ -144,10 +145,22 @@ Package selectGEMMMicrokernel(GEMMProtocol protocol, HWInformation hwInfo, SizeP
...
-    auto entries = select(catalog, 1, &matchParams, evalParams, auxParams);
+    auto entries = select(catalog, 1, &matchParams, evalParams, auxParams, observer);
     if (entries.size() > 0)
 	    entry = entries[0];
 
+    kcatalog::Entry modifiedEntry; // Oustide of the if statement to expand lifetime
+    if (entry) {
+        char *newStrat = getenv("MICRO_STRATEGY");
+        if (newStrat) {
+            std::cout << "Updating strategy to " << newStrat << std::endl;
```
umar456: Not a fan of writing to std::out for these logs. It would be preferable to have observer like functions to do this. For this function you can do that using either the observer function or the adjustStrategy function.

Simonsays095: I'm not really a fan of it, either, but this matches what we do in the copy planner already, and we don't really have any other infrastructure for it. The observer is built for printing out catalog entries (the function takes a catalog entry as an argument), and isn't much better of a solution in my opinion. Of course we could design a second function to pass as an optional argument for this information as well, but I don't like the idea of having tons of optional arguments for every bit of information we may want to print.
`adjustStrategy` is used in both microkernel and regular kernel pathways, and we already have a oneDNN-side approach for overriding and printing out kernels which this would clash with (see [gen_kernel.cpp#L128](https://github.com/Simonsays095/oneDNN/blob/1f6399befbbc7544e730a41912bc2003d372893e/src/gpu/intel/gemm/jit/gen_kernel.cpp#L128)). I would prefer to handle this similarly, on the calling code side, but afaik there's no way for the calling code to access/modify the strategy - it's completely opaque in the microkernel interface.

I think the best solution is ultimately to improve the gemmstone verbose structure to either enable passing data back as a string to calling code, add a full-featured verbose line format, and/or allow for ways to inject verbose logic. But that'll have to wait for another day.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

umar456 (approved):

PR #4533, opened 2026-01-13
Enable bf16 ukernel path on aarch64

michalowski-arm:
This builds on #4406, enabling bf16:bf16:bf16 and bf16:bf16:f32 paths for the ukernel on aarch64 and adding an appropriate transform utility for the weights.

Tested with the nightly test set and passing everything fine.

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2023-2024 FUJITSU LIMITED
-* Copyright 2024-2025 Arm Ltd. and affiliates
+* Copyright 2024-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -501,6 +501,169 @@ void jit_brgemm_matmul_copy_b_int8_t<isa>::generate() {
...
+
+    jit_brgemm_matmul_copy_b_bf16_t(const brgemm_matmul_conf_t *conf)
+        : jit_brgemm_matmul_copy_b_t(conf)
+        , typesize_((size_t)conf->b_dt_sz)
+        , k_blk_step_(data_type_vnni_granularity(data_type::bf16)) // = 2
+        , n_blk_step_(get_sve_length() / (k_blk_step_ * typesize_))
+        , tr_store_step(k_blk_step_ * typesize_)
+        , src_stride_(conf->wei_tag == format_tag::acbd
+                          ? conf->copy_B_wei_stride
+                          : conf->N * typesize_)
```
dzarukin, line 516: I'm not sure about this particular spot.
I'd be surprised to see 4D format tag in ukernel setup. Assuming this is never the case, then it means B is always expected in `ab` format, but the support for `ba` wasn't disabled.
Could you check if `wtag=ba` is working properly? If somehow yes, could you, please, explain me why it's working correctly?

openmodelzooci:

PR #4535, opened 2026-01-13
cpu: platform: aarch64: query cores even without DNNL_AARCH64_USE_ACL

Sqvid:
# Description

The code use to query the number of cores does not actually depend on Arm ComputeLibrary being built, allowing us to remove this limitation.

# Checklist

## General

### DIFF
--- a/src/cpu/platform.cpp
+++ b/src/cpu/platform.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2020 Intel Corporation
 * Copyright 2020-2024 FUJITSU LIMITED
-* Copyright 2022-2024 Arm Ltd. and affiliates
+* Copyright 2022-2024, 2026 Arm Ltd. and affiliates
 *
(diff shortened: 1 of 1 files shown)


jondea (approved):

Sqvid:
@dzarukin I don't think there is any way this would break your build but would you like to restart your Azure runner for completeness?

> ##[error]We stopped hearing from agent Azure Pipelines 6. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610

dzarukin:
Nope, I'm good. Thanks for asking.

vpirogov:
@Sqvid, these are public runners. We don't have any control over these :)

Sqvid:
Thanks, i'll keep that in mind! I just saw there's a "Sign In" button and I had no idea what power lay behind it.

PR #4536, opened 2026-01-13
cpu: platform: aarch64: query and calculate cache_per_core

Sqvid:
# Description
Query the actual cache levels on the processor rather than returning a guess value.

There are several instances of heuristics depending on this function. So I will do some extra testing before we merge.

# Checklist

## General

## Performance improvements


openmodelzooci:

PR #4541, opened 2026-01-13
[GPU] Refactor, fixup quant bcast check

kealan-barbieri:
# Description

Fixes ref dispatching for batched GEMMs with per_tensor quant attributes. 

- Add quant reshape check for zp
- Add exception to quant reshape restriction for `per_tensor` mask
- refactor check into `maybe_reshape`
- rename bcast variable
- addred regression testcase

Fixes # [MFDNN-14541](https://jira.devtools.intel.com/browse/MFDNN-14541)

# Checklist

## General

### Bug fixes


kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -295,6 +286,36 @@ struct gemm_t : public primitive_t {
...
+                            || entry.get_mask() == full_tensor_mask)
+                        return true;
+
+                    int quant_group_ndims = 0;
+                    for (int i = 0; i < 2; ++i) {
+                        if (entry.get_group(i) > 1) ++quant_group_ndims;
+                    }
+                    if (quant_group_ndims > 0) return false;
+                    return true;
+                };
```
Simonsays095: Can we just have similar logic to how we handle the bias broadcasts, instead of another control flow? I think this logic is incorrect too - it shouldn't depend on `DNNL_ARG_WEIGHTS`, the number of dimensions we're squashing, or even the number of grouped dimensions. We just need all or none of the dims to be broadcast just like bia dims (with one exception, below).

The exception: we can use the group size to "absorb" the last merged dim being broadcast if needed. This is what `squash_quant` is doing when it updates the group size. See the example above in the source code. I think this transformation is valid when the last merged dim has a quantization tensor size of 1 (i.e. is broadcast).

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/matmul/gemm.hpp
kealan-barbieri: Pushed a change that takes into account the bcast dims as we discussed offline. Preserving the group dim counts as the bias/dim squash restrictions do not apply in cases with only `per_oc`/`common` quant. The earlier checks re `DNNL_ARG_WEIGHTS` are just early outs where its not possible to have hit an invalid condition.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/matmul/gemm.hpp
kealan-barbieri: Further discussed offline, dropped arg specific condition as it doesnt affect result, replaced dim count with mask check so condition is applied to ungrouped per_ocic scales, added comments.

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -295,6 +286,42 @@ struct gemm_t : public primitive_t {
...
+                                b_md->dims, b_dims);
+                bcast_ok = bcast_ok
+                        && safe_bcast_quant(
+                                gemm_attr.zero_points_.get(DNNL_ARG_SRC),
+                                a_md->dims, a_dims);
+                bcast_ok = bcast_ok
+                        && safe_bcast_quant(
+                                gemm_attr.zero_points_.get(DNNL_ARG_WEIGHTS),
+                                b_md->dims, b_dims);
+                if (!bcast_ok) return status::success;
```
dzarukin, line 323: Is it a subject for immediate dropping this implementation? If yes, does it make sense to enable verbose_dispatch mechanism, otherwise, it would be harder to debug which condition out of 4 led to a problem.

tests/benchdnn/inputs/matmul/test_matmul_gpu
```diff
@@ -88,6 +88,13 @@
 --attr-scales=src:common:0.25+wei:per_oc+dst:common:2  --attr-scratchpad=user
 24576x1024:1024x4096
 
+--reset
+--dt=s8:u8:f16 --stag=abc --wtag=acb --dtag=abc
+--attr-scales=src:per_tensor:f16:1x128+wei:per_oc:f16
+--attr-zero-points=wei:per_oc:u8
+--attr-precomputed-reductions=src:per_tensor:s32:1x128
+256x1x512:1x512x1024
```
dzarukin: Suggest to post this into `harness_matmul_precomp_reductions` instead.

src/gpu/intel/matmul/gemm.hpp
Simonsays095: This implementation is more of a bridge, connecting `matmul` to `gemm` through a reshaping/normalization process. If this function returns `status::success`, then we try to find a gemm implementation that works. In this case, the return here just says "don't reshape, but try to use gemm". Only if this returns `status::unimplemented` or some other error status do we skip the implementation and we do use `VDISPATCH_MATMUL_SC` in that case.

So I don't think `VDISPATCH` makes sense for these cases, although maybe something like `VDEBUGINFO` would help clarify what the particular cause is for a failed reshape.

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -295,6 +286,42 @@ struct gemm_t : public primitive_t {
                     return status::success;
                 };
 
+                // Quantization attributes are unsupported in Gemmstone with any
+                // non-trivial modification of groups during reshape.
+                // TODO: Enable reshaped grouped quant.
```
Simonsays095: nit: I actually think any 2D quantization is supported by gemmstone (you recently enabled this, right?). I even think we'll still modify the group sizes with these changes due to the adjust_quant function (again, look at the example above: we'll pass on group sizes of 4x128 to gemmstone).
These limitations below get at the fundamental idea of equivalence between different representations of the underlying tensors. If a problem breaks any of these rules, then the reshaped problem corresponds to a different gemm problem entirely. I actually don't think that there's a TODO left, except to maybe refactor as we've discussed offline, but that's not a priority.

h-sadia:
Am I missing the maybe_reshape check here? Or is it somewhere in the changes that I dont see?

kealan-barbieri:
make test perf-gpu
set primitive=matmul

src/gpu/intel/matmul/gemm.hpp
kealan-barbieri: Gemmstone does now allow 2d groups and we do pass on 2d groups in some cases still, I can clarify this further by stating that "trivial" here means a group that is broadcast or full-dim size. 

I dont think that cases that are rejected by this check are not functionally equivalent, eg if we passed them to ref the results would be the same as un-reshaped jit:gemm, Gemmstone just doesnt handle them properly.

tests/benchdnn/inputs/matmul/test_matmul_gpu
kealan-barbieri: Relocated, thanks.

kealan-barbieri:
@h-sadia `maybe_reshape` called here https://github.com/uxlfoundation/oneDNN/blob/130206a383a8a0767b46bd63020abb3b26663d8e/src/gpu/intel/matmul/gemm.hpp#L332

dzarukin (approved):

Simonsays095:
make test perf-gpu
set primitive=matmul

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -66,21 +66,13 @@ struct gemm_t : public primitive_t {
             auto orig_dims = a_md->ndims;
 
             // TODO: Enable reshaped 2D groups.
```
Simonsays095: This TODO is also out-of-date now.

kealan-barbieri: dropped, thanks.

Simonsays095:
make test perf-gpu
set primitive=matmul

kealan-barbieri:
perf check for pre- #4478 [HERE](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f0f27a67-c438-f1a4-aa53-a4bf010d0e2d)

PR #4542, opened 2026-01-14
Maintainers: year end cleanup

vpirogov:
Two main changes:
* Removed maintainers who changed assignments and no longer contribute to oneDNN.
* Removed review groups for gpu-amd, gpu-nvidia, and gpu-generic from CODEOWNERS, as these do not have maintainers anymore.


vpirogov:
> Do we need to update the teams as well, eg. https://github.com/orgs/uxlfoundation/teams/onednn-graph ?

Yes, good catch.

PR #4546, opened 2026-01-15
benchdnn: inputs: work around false-positive hits

dzarukin:
Work around for [MFDNN-14550](https://jira.devtools.intel.com/browse/MFDNN-14550) and [MFDNN-14498](https://jira.devtools.intel.com/browse/MFDNN-14498).

Infrastructure scripts were modified and now pick up VS2022 properly (they were picking up VS2019 previously). This action led to change in RNG implementation between VS and leading to false positives caused by minor difference in DST and DST_ITER leading to rounding "minus 1" issue between ref and library, leading to off-norm caused by a single point (rest are exact), which doesn't give a credit for current system handling thresholds.

I've left a comment what potentially can be done but I didn't do it to save time for more important things as RNN is not considered important functionality and instead decreased shapes.

### DIFF
--- a/tests/benchdnn/inputs/rnn/harness_gru_regression
+++ b/tests/benchdnn/inputs/rnn/harness_gru_regression
@@ -1,4 +1,2 @@
 # int8 SIC != SLC
---reset
---trivial-strides=true --prop=FWD_I --alg=VANILLA_GRU --activation=UNDEF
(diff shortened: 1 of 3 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_rnn
set test_scope=NIGHTLY

PR #4550, opened 2026-01-15
xe: untangle microkernel <-> gemmstone interdependency

rjoursler:


rjoursler:
make test
disable test_device_cpu

src/gpu/intel/gemm/jit/include/gemmstone/microkernel_provider.hpp
```diff
@@ -30,8 +33,30 @@ struct HWInformation {
...
+    explicit GEMMOptions(int flags)
+        : localA(flags & (1 << 0))
+        , localB(flags & (1 << 1))
+        , addToC(flags & (1 << 2))
+        , slmPtr(flags & (1 << 3))
+        , offsetA(flags & (1 << 4))
+        , offsetB(flags & (1 << 5))
+        , scaleA(flags & (1 << 6))
+        , scaleB(flags & (1 << 7)) {}
+};
```
Simonsays095: I don't think this flag representation is used anywhere any more.

src/gpu/intel/gemm/jit/include/gemmstone/microkernel_selector.hpp
```diff
@@ -20,8 +20,11 @@
 #include "gemmstone/config.hpp"
 #include "gemmstone/kernel_selector.hpp"
```
Simonsays095, line 21: While we're changing names of these files by relocating them, can we change the name of this file to `microkernel_selector.hpp` to mirror the `kernel_selector.hpp` file?

src/gpu/intel/microkernels/gemmstone_interop.cpp
Simonsays095: The ordering of these arguments must match those specified via `interface.newArgument` in `microkernel_provider.cpp`. This creates a hard dependency between user code and gemmstone version, and the likelihood of this breaking on update. Can we move this file (and the .hpp) into gemmstone as well, to enforce consistency?

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
```diff
@@ -26,11 +25,14 @@
...
-namespace intel {
-namespace micro {
+
+GEMMSTONE_NAMESPACE_START
+namespace microkernel {
+
+template <typename T, typename U>
+static inline T divideUp(T num, U denom) {
+    return (num + denom - 1) / denom;
+}
```
Simonsays095, line 35: We already have a version of this in `include/internal/utils.hpp:div_up` we could use, now that it's part of the gemmstone code.

src/gpu/intel/microkernels/gemmstone_interop.cpp
rjoursler: No, the issue is that the specific names used for these arguments is the input as these depend on the implementation we are injecting code into. Likely we need to just rework this interface in the future to break this dependency.

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
rjoursler: True, I had thought about this, but there is enough discussion on exactly how this should be integrated with Gemmstone at this time, JIC we end up moving the implementation again in the near future.

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/microkernels/gemmstone_interop.cpp
Simonsays095: These names have to match those in `gemm_microkernel.cxx` - they're provided by gemmstone, not the user.

rjoursler: To my understanding, this is not correct. The nGEN interface is setup [here](https://github.com/uxlfoundation/oneDNN/blob/9b98b64fd79b157f71e3f5ceea3406afd7353934/src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp#L208-L228)  and does not rely on these names. Instead, these names are used by [generateShim](https://github.com/uxlfoundation/oneDNN/blob/9b98b64fd79b157f71e3f5ceea3406afd7353934/src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp#L209) when generating the type names for inclusion into the OpenCL source code. These need to be configurable in the OpenCL code to ensure there aren't any naming conflicts (like when multiple microkernels are used).

Simonsays095: Currently, the `Protocol` defines the name, order, and type of arguments. This is then used in two ways:
1. `gemm_microkernel.cxx:gemmMicrokernelPackage`: Protocol names are translated into internal names.
    - These internal names are the ones that have to match the ones given to the nGEN interface in `microkernel_provider.cpp`.
    - Internal names are used to map argument types to problem types.
    - This mapping starts with protocol-defined argument names, not user-defined names.
2. `shim.cpp:generateShim`: Verifies that all of the protocol-defined argument names match the ones supplied in the package. The type names are distinguished (in the event of multiple microkernels being used) by the `shimOptions.decorator`.

So kernel names are defined and consumed in internal code, exposed to the user via type definitions, and disambiguated via shim options selected by the user.

rjoursler: Yeah, looks like a mess. I don't think we should move this for now, but there definitely should be more refinement of the API in the future. Likely we just need to extend `GEMMOptions` to avoid the entaglement.

rjoursler:
make test
disable test_device_cpu

PR #4325, opened 2025-11-13
[rls-v3.10] change gemm impl list order

dyoussif:
addresses [MFDNN-14353](https://jira.devtools.intel.com/browse/MFDNN-14353). Per @echeresh's  comment we can change the order of implementations to reduce pd creation time.

PR #4555, opened 2026-01-16
xe: gemm: move gemm_with_po lower in impl list

dyoussif:
backport https://github.com/uxlfoundation/oneDNN/pull/4325

### DIFF
--- a/src/gpu/gpu_gemm_list.cpp
+++ b/src/gpu/gpu_gemm_list.cpp
@@ -34,12 +34,14 @@ namespace gpu {
 
 namespace {
 
+// gen_t is used most of the time, keep ahead of gemm_with_po
+// to reduce pd creation time
 // clang-format off
 constexpr impl_list_item_t impl_list[] = {
         GPU_INSTANCE_INTEL_DEVMODE(intel::gemm::conv_t)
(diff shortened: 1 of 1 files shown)


src/gpu/gpu_gemm_list.cpp
```diff
@@ -34,12 +34,14 @@ namespace gpu {
...
 
+// gen_t is used most of the time, keep ahead of gemm_with_po
+// to reduce pd creation time
 // clang-format off
 constexpr impl_list_item_t impl_list[] = {
         GPU_INSTANCE_INTEL_DEVMODE(intel::gemm::conv_t)
         GPU_INSTANCE_INTEL(intel::gemm::xe_hp_systolic_t)
-        GPU_INSTANCE_INTEL(intel::gemm::with_post_ops_t)
         GPU_INSTANCE_INTEL(intel::gemm::gen_t)
+        GPU_INSTANCE_INTEL(intel::gemm::with_post_ops_t)
```
rjoursler, line 44: We should remove the check to skip `gen_t` in main (this doesn't seem particularly important for backporting though): https://github.com/uxlfoundation/oneDNN/blob/7e4e9df49c7650e7d593ba0166c5a48afd9b2527/src/gpu/intel/gemm/with_post_ops.cpp#L113

PR #4571, opened 2026-01-21
[GPU] GEMM: fix issues with host scalars and scales

echeresh:
Spotted a few issues:

- Missing matmul reference kernel coverage for FP8 scales
- Incorrect weights groups check in `matmul_pd.hpp`
- Incorrect `dnnl_memory_create_host_scalar()` call from benchdnn due to type mismatch
- Naming: use `swap_ab` and `host_scalar` in JIT GEMM

### DIFF
--- a/src/gpu/intel/gemm/jit.cpp
+++ b/src/gpu/intel/gemm/jit.cpp
@@ -46,8 +46,8 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
         intel::stream_t *compute_stream, zero_pool_t *zero_pool,
         const memory_storage_t &a, const memory_storage_t &b,
         const memory_storage_t &c, const memory_storage_t *ao,
-        const memory_storage_t *bo, int16_t ao_hostscalar,
-        int16_t bo_hostscalar, const memory_storage_t *a_scales,
+        const memory_storage_t *bo, int16_t ao_host_scalar,
+        int16_t bo_host_scalar, const memory_storage_t *a_scales,
         const memory_storage_t *b_scales, const memory_storage_t *c_scales,
         const memory_storage_t *ag, const memory_storage_t *bg,
         const memory_storage_t &co, const memory_storage_t *c_temp,
@@ -56,7 +56,7 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
         int64_t offset_c, int64_t offset_aq, int64_t offset_bq,
         int64_t offset_co, int64_t *offset_po_src, int32_t lda, int32_t ldb,
         int32_t ldc, int32_t m, int32_t n, int32_t k, int32_t k0, float alpha,
-        float beta, int32_t cmask, bool last_k_block, bool swapab,
+        float beta, int32_t cmask, bool last_k_block, bool swap_ab,
         bool disable_hilbert) const {
     if (pd()->desc()->batch() == 0) return status::success;
 
@@ -94,8 +94,8 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
         arg_list.set(argn++, *ao);
     if (pd()->with_b_zero_points() && !problem->bOffsetHostScalar())
         arg_list.set(argn++, *bo);
-    if (problem->aOffsetHostScalar()) arg_list.set(argn++, ao_hostscalar);
-    if (problem->bOffsetHostScalar()) arg_list.set(argn++, bo_hostscalar);
+    if (problem->aOffsetHostScalar()) arg_list.set(argn++, ao_host_scalar);
+    if (problem->bOffsetHostScalar()) arg_list.set(argn++, bo_host_scalar);
     if (problem->aScale2D()) arg_list.set(argn++, *a_scales);
     if (problem->bScale2D()) arg_list.set(argn++, *b_scales);
     if (pd()->with_mx_scale()) arg_list.set(argn++, *c_scales);
@@ -345,7 +345,7 @@ status_t gen_t::execute(const exec_ctx_t &ctx) const {
     const auto d = pd()->desc();
     const auto &problem = *pd()->kernel_desc()->problem();
 
-    const bool swapab = pd()->swap_ab();
+    const bool swap_ab = pd()->swap_ab();
 
     auto a_type = pd()->eff_a_type();
     auto b_type = pd()->eff_b_type();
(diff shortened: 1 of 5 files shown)


tests/benchdnn/dnnl_memory.hpp
```diff
@@ -27,11 +27,27 @@
...
 #include "utils/dims.hpp"
+#include "utils/dnnl_query.hpp"
 #include "utils/wrapper.hpp"
 
 #define dnnl_mem_default_value 0xFF
 #define dnnl_mem_default_perf_test_value 0x3F
 
+template <typename T_>
+dnnl_data_type_t get_dt() {
+    using T = typename std::decay<T_>::type;
```
dzarukin: include <type_traits> for `std::decay`?

tests/benchdnn/dnnl_memory.hpp
```diff
@@ -27,11 +27,27 @@
...
 #include "common.hpp"
 #include "utils/dims.hpp"
+#include "utils/dnnl_query.hpp"
 #include "utils/wrapper.hpp"
 
 #define dnnl_mem_default_value 0xFF
 #define dnnl_mem_default_perf_test_value 0x3F
 
+template <typename T_>
+dnnl_data_type_t get_dt() {
```
dzarukin: Minor: I'm tempting to ask to move this to numeric.hpp and better expand with all supported data types to avoid surprises....

tests/benchdnn/dnnl_memory.hpp
```diff
@@ -27,11 +27,27 @@
...
+#define CASE(dt, type) \
+    if (std::is_same<T, type>()) return dt;
+    CASE(dnnl_f32, float);
+    CASE(dnnl_f64, double);
+    CASE(dnnl_s32, int32_t);
+    CASE(dnnl_s64, int64_t);
+    CASE(dnnl_s8, int8_t);
+    CASE(dnnl_u8, uint8_t);
+#undef CASE
+    return dnnl_data_type_undef;
```
dzarukin: Probably, assert wouldn't hurt to avoid (slightly longer) debugging.

tests/benchdnn/dnnl_memory.hpp
echeresh: I wanted to do exactly this but `numeric.hpp` doesn't include headers for float4/float8 so I thought it's not a good idea to bring these headers unconditionally just for the sake of covering these data types, especially since they are not used...

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

tests/benchdnn/dnnl_common.cpp
```diff
@@ -2048,19 +2048,15 @@ int init_ref_memory_args_default_case(int exec_arg, dnn_mem_t &mem,
         int local_exec_arg = exec_arg ^ DNNL_ARG_ATTR_ZERO_POINTS;
         TIME_FILL(SAFE(
                 fill_zero_points(attr, local_exec_arg, mem, ref_mem), WARN));
-    } else if (is_dropout_p && !attr.dropout.use_host_scalars) {
+    } else if (is_dropout_p) {
+        mem.set_f32_elem(0, attr.dropout.p);
```
dzarukin: I think `mem` calls should be `set_elem` in all three cases as the library data type might be different from ref data types.

tests/benchdnn/dnnl_memory.hpp
```diff
@@ -61,7 +79,20 @@ struct dnn_mem_t {
     dnn_mem_t(const dnn_mem_t &rhs, dnnl_data_type_t dt, const std::string &tag,
             dnnl_engine_t engine);
 
-    dnn_mem_t(const_dnnl_memory_desc_t md, void *value);
+    // Construct a host-scalar memory object, initialized with 0x3F pattern.
+    dnn_mem_t(const_dnnl_memory_desc_t md);
+
+    // Construct a host-scalar memory object, type `T` must match the memory descriptor.
+    template <typename T>
+    dnn_mem_t(const_dnnl_memory_desc_t md, T *value) {
```
dzarukin: Since the logic to get a value by pointer from the external interface was removed, is this interface still needed?

tests/benchdnn/dnnl_common.cpp
echeresh: It's actually fixed here, for probability it's always `f32` but I agree, to future-proof better to use `set_elem()`.

I have a question about `set_s64_elem()`:

- The dropout seed can be either `s32` or `s64` (from [documentation](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_dropout.html#api))
- `set_s64_elem()` is unconditional - no `s64` checks, just writes to memory
- `set_elem()` is safe and supports flexible conversion but it takes `f32` only
- AFAIU, for your suggestion to apply we need to do `s64` -> `f32` conversion to go through `set_elem()`. Maybe not a good idea?
- Do you have suggestions how to make it more robust?

dzarukin: Alright, it is doomed. Let's keep set_s64_elem. 
Benchdnn doesn't allow s32 seed it's going to utilize s64 always, but adding s32 data type case into set_s64_elem just to be safe wouldn't probably hurt.

echeresh: Since this is for dropout only, maybe just add type-check assertions and keep `s64` as you explained it's enough?

dzarukin: It should do, let's stick to that.

tests/benchdnn/dnnl_memory.hpp
echeresh: Makes sense, dropped.

echeresh:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

PR #4572, opened 2026-01-22
xe: gemm: fixup ukernel selection

Simonsays095:
Addresses 1st token regression reported in [MFDNN-14526](https://jira.devtools.intel.com/browse/MFDNN-14526). When the `kParallelLocal` option was disabled, the microkernel selector would still pick the first entry, even if it used k-parallel parameters. It would then disable those parameters, and potentially end up with a worse kernel than the catalog suggested.

We still do this strategy mutation if needed, but now we attempt to find a kernel that matches the protocol exactly first. Only if we can't find one do we go back and pick the otherwise "best" kernel to mutate.


Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

kealan-barbieri (approved):

PR #4576, opened 2026-01-22
fixup: fixup: xpu: ocl: link to OCL indirectly

dzarukin:
Despite Coverity reported a resource leak, `dlclose`-ing a handle leads to a problem and any potential solutions might lead to problems with applications that link with oneDNN.

### DIFF
--- a/src/xpu/utils.hpp
+++ b/src/xpu/utils.hpp
@@ -163,10 +163,19 @@ inline void *find_symbol(const char *library_name, const char *symbol) {
                 "error while searching for a %s symbol address in %s library: "
                 "%s",
                 symbol, library_name, dlerror());
-        dlclose(handle);
+        // See a comment below.
+        // dlclose(handle);
         return nullptr;
     }
-    dlclose(handle);
+    // Note: `dlclose` invalidates `symbol_address` if the application hadn't
+    // had a `handle` opened before. The solution to put a handle in some
+    // `static` object leads to other problems with unloading a library from
+    // applications linked with oneDNN when it comes to managing global
+    // resources passed between libraries such as contexts.
+    //
+    // Thus, the recommendation is not to `dlclose` the handle and let it slide.
+    //
+    // dlclose(handle);
     return symbol_address;
 }
 #endif


src/xpu/utils.hpp
```diff
@@ -163,10 +163,18 @@ inline void *find_symbol(const char *library_name, const char *symbol) {
...
     }
-    dlclose(handle);
+    // Note: `dlclose` invalidates `symbol_address` if the application hadn't
+    // had a `handle` opened before. It might be possible to put a handle in
+    // a `static` object, but it might lead to other problems with unloading the
+    // library from applications linked with oneDNN.
+    //
+    // Thus, the recommendation is not to `dlclose` the handle and let it slide.
+    //
+    // dlclose(handle);
```
rjoursler, line 178: In the past, we have gotten around this issue by associating the library handle with a oneDNN object (see [mdapi_helper_t](https://github.com/uxlfoundation/oneDNN/blob/8a610254546fbce1316b6cb34a51a1df9407080d/src/gpu/intel/ocl/stream.cpp#L40)). Could we do a similar thing here by having the interop API accessed through `engine` and `stream` objects?

dzarukin: The key difference here is global resources passed between library boundaries (I guess, the comment should reflect that).
Once the handle is wrapped into a static object, the order stops being specified and it leads to a trouble when the library gets unloaded earlier than some symbol from it frees global resources.

rjoursler: > Once the handle is wrapped into a static object, the order stops being specified and it leads to a trouble when the library gets unloaded earlier than some symbol from it frees global resources.

Agreed, but is this an issue in practice? I know we already ran into this issue with respect to the kernel cache and implemented a helper for when querying if destruction is safe: https://github.com/uxlfoundation/oneDNN/blob/8a610254546fbce1316b6cb34a51a1df9407080d/src/common/utils.cpp#L292-L344

This didn't actually require any changes outside of Windows.

dzarukin: > Agreed, but is this an issue in practice? 

I can't answer this question with full confidence. We can make a separate experiment to verify, but it leads to a race condition with, at least, kernel cache on our side where the handle must stay alive until all kernels are released to have `clReleaseKernel` a valid address space.

densamoilov: > Agreed, but is this an issue in practice?

Over the last 5 years every time when I thought it wouldn't be an issue I was wrong so I'd play it safe here too. As for `is_destroying_cache_safe` function I think it used to always return false for window.

PR #4578, opened 2026-01-23
cpu: x86: deconv fallback to non-brgemm kernel on unimplemented error

georgen117:
# Description
[MFDNN-14380](https://jira.devtools.intel.com/browse/MFDNN-14380)

The issue is at jit_brgemm_conv_bwd_utils.cpp:1460:

if (cd.use_inversion && has_uneven_spatial) return status::unimplemented;
This returns unimplemented when using deconvolution (use_inversion is true) with uneven spatial dimensions. The has_uneven_spatial condition checks if the spatial dimensions are not evenly divisible by the strides:

const bool has_uneven_spatial = jcp.id % jcp.stride_d != 0        || jcp.ih % jcp.stride_h != 0 || jcp.has_uneven_iw;
In your case, you have:

ih127 (input height = 127) with sh2 (stride = 2): 127 % 2 = 1 (not evenly divisible)
iw127 (input width = 127) with sw2 (stride = 2): 127 % 2 = 1 (not evenly divisible)
id127 (input depth = 127) with sd2 (stride = 2): 127 % 2 = 1 (not evenly divisible)
The problem: The brgemm deconvolution implementation doesn't support spatial dimensions that aren't evenly divisible by the strides when using the backwards-strided path.

This results in the deconvolution running the reference implementation of the post-ops. Causing a significant performance degradation.

This issue can be reproduced using the following benchdnn command:
**Any tag:**
```
ONEDNN_MAX_CPU_ISA=AVX512_CORE_BF16 ./benchdnn --deconv --reset --allow-enum-tags-only=0 --engine=cpu --dir=FWD_I --alg=direct --dt=bf16:bf16:bf16 --bia-dt=f32 --stag=any --wtag=any --dtag=any --attr-scratchpad=user mb1_ic16oc8_id128od256kd3sd2dd0pd0_ih127oh255kh3sh2dh0ph0_iw128ow256kw3sw2dw0pw0
```
**Channel last: ndhwc**
```
ONEDNN_MAX_CPU_ISA=AVX512_CORE_BF16 ./benchdnn --deconv --reset --allow-enum-tags-only=0 --engine=cpu --dir=FWD_I --alg=direct --dt=bf16:bf16:bf16 --bia-dt=f32 --stag=ndhwc --wtag=any --dtag=ndhwc --attr-scratchpad=user mb1_ic16oc8_id127od255kd3sd2dd0pd0_ih127oh255kh3sh2dh0ph0_iw127ow255kw3sw2dw0pw0
```

If a brgemm kernel is not found the code will try any other existing convolution kernels, before falling back to the reference implementation.

### DIFF
--- a/src/cpu/x64/jit_brgemm_deconv.cpp
+++ b/src/cpu/x64/jit_brgemm_deconv.cpp
@@ -199,7 +199,6 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
                 reinterpret_cast<const op_desc_t *>(&conv_d), attr(), nullptr);
         if (!it.is_initialized()) return status::out_of_memory;
 
-        bool found_impl = false;
         // First pass: try to find BRGEMM backward strided implementation
         while (++it != it.end()) {
             conv_pd_ = *it;
@@ -207,7 +206,6 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
                         typename brgemm_convolution_bwd_strided_t<isa>::pd_t>(
                         it)
                     == status::success) {
-                found_impl = true;
                 break;
             }
         }
@@ -215,24 +213,20 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
         // Second pass: fallback to any other backward data convolution implementation
         // This allows non-BRGEMM kernels (like jit_avx512_core_bf16) to handle
         // cases with uneven spatial dimensions
-        if (!found_impl) {
+        if (it == it.end()) {
             primitive_desc_iterator_t it2(engine,
                     reinterpret_cast<const op_desc_t *>(&conv_d), attr(),
                     nullptr);
             if (!it2.is_initialized()) return status::out_of_memory;
             while (++it2 != it2.end()) {
                 conv_pd_ = *it2;
-                if ((*it2)->kind() == primitive_kind::convolution) {
-                    found_impl = true;
-                    break;
-                }
+                if ((*it2)->kind() == primitive_kind::convolution) { break; }
             }
+            if (it2 == it2.end())
+                VDISPATCH_DECONVOLUTION_IC(false,
+                        "no suitable implementation found for strided "
+                        "deconvolution");
         }
-
-        if (!found_impl)
-            VDISPATCH_DECONVOLUTION_IC(false,
-                    "no suitable implementation found for strided "
-                    "deconvolution");
     } else {
         CHECK(fwd_conv_desc_create(fwd_deconv_d, &conv_d));
 


georgen117:
@tczeszun asked me to let him review this before committing.

src/cpu/x64/jit_brgemm_deconv.cpp
```diff
@@ -199,17 +199,28 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
                 reinterpret_cast<const op_desc_t *>(&conv_d), attr(), nullptr);
         if (!it.is_initialized()) return status::out_of_memory;
 
+        bool found_impl = false;
```
dzarukin: It seems this variable is not necessary. Relying on `it` should do.

src/cpu/x64/jit_brgemm_deconv.cpp
```diff
@@ -199,17 +199,29 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
...
         while (++it != it.end()) {
             conv_pd_ = *it;
+            // First, try BRGEMM backward strided implementation
```
dzarukin: It seems to me the logic doesn't coincide with the intention - if the first impl in the list is not strided, it checks if it was convolution (what else can it be), and picks it up. If strided is intended to be picked up first, two honest passes over it are needed.

tczeszun: Agree, @georgen117 we need to pick non-BRGEMM in a separate `while` loop

src/cpu/x64/jit_brgemm_deconv.cpp
georgen117: Agreed. I will remove the `found_impl` flag.

src/cpu/x64/jit_brgemm_deconv.cpp
georgen117: You are correct this doesn't properly prioritize checking the BRGEMM implementation.

I will break it up to two passes.

src/cpu/x64/jit_brgemm_deconv.cpp
tczeszun, line 178: (random place), please add a comment above jit_brgemm_conv_bwd_utils.cpp:1460
saying that deconvolution (use_inversion) is disabled, because for uneven spatial cases it's needed for binary post-ops to support bigger rhs tensor basing on enlarged output tensor

(we may partially enable it to run without post-ops, but it's not worth imo as the fallback code still will need to be there)

src/cpu/x64/jit_brgemm_deconv.cpp
georgen117: I ended up keeping the `found_impl` flag because the `primitive_desc_itterator_t` does not have any way to reset the iterator to the beginning.  There is no `reset()` method and no `begin()` method. Since the code disallows copy and assign operators there was no way that I could figure out to do the second pass other than creating a new iterator.

Now that there are two separate iterators `it == it.end()` no longer works. I decided the simplest to understand code was keeping the `found_impl` flag.

src/cpu/x64/jit_brgemm_deconv.cpp
georgen117: done.

src/cpu/x64/jit_brgemm_deconv.cpp
```diff
@@ -199,17 +199,40 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
...
+                    nullptr);
+            if (!it2.is_initialized()) return status::out_of_memory;
+            while (++it2 != it2.end()) {
+                conv_pd_ = *it2;
+                if ((*it2)->kind() == primitive_kind::convolution) {
+                    found_impl = true;
+                    break;
+                }
+            }
+        }
```
dzarukin, line 229: I think this might work:
```
it(...);
while(it) {
   ...
}
if (it == it.end()) {
    it2(...);
    while (it2) {
    }
    if (it2 == it2.end()) {
        report_error(...);
    }
}
return success;
```

src/cpu/x64/jit_brgemm_conv_bwd_utils.cpp
```diff
@@ -1457,6 +1457,9 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
     const bool has_uneven_spatial = jcp.id % jcp.stride_d != 0
             || jcp.ih % jcp.stride_h != 0 || jcp.has_uneven_iw;
 
+    // deconvolution (use_inversion) is disabled, for uneven spatial cases
+    // it's needed for binary post-ops to support bigger rhs tensor basing
+    // on enlarged output tensor
     if (cd.use_inversion && has_uneven_spatial) return status::unimplemented;
```
dzarukin: I wonder if it makes sense to add verbose message here instead of plain return.

georgen117: I will update the code to add a verbose message.

src/cpu/x64/jit_brgemm_deconv.cpp
georgen117: done

georgen117:
make test

georgen117:
I have checked the errors reported by the CI build. All but 1 of the errors existed in other builds. The 1 error that didn't exist in other builds was a failure to open a file. I don't think any of the CI errors should prevent this from being merged.

georgen117:
@dzarukin and @tczeszun I am still new enough that I am not sure what the expected procedure to continue is here. I have no merge conflicts and the change has been approved by two reviewers.

I have done my due diligence to make sure none of the CI errors were due to this change.

Am I as the developer expected to merge this PR or are maintainers going to do the merge. The `squash and merge` button is active for me. All the other PRs I have done were merged by maintainers. I am just trying to check if this is being held up by me or for some other reason, I am unaware of.

PR #4587, opened 2026-01-27
Xbyak Aarch64 Register Manager

henry-gar:
### Summary of Proposal

This document serves as an overview and proposal for a user-level register manager to allow developers working with Xbyak_aarch64 to  more easily track, manage and (de)allocate registers.

### Desired Outcomes:

1. Provide tracking and standardised, automatic allocate and free methods for registers.
2. Guarantee unique ownership of registers for each of the three register families: General Purpose (GP); SIMD/Floating Point (FP); Predicate (Pred).
3. Implement scoped registers following the RAII technique.
4. Ensure compatibility & minimum friction with current register manager techniques in oneDNN.

### Context & Motivation

The motivation for this project is to reduce the manual user-level register management for projects using Xbyak_aarch64 and to uphold the ABI (https://github.com/ARM-software/abi-aa/releases/download/2025Q1/aapcs64.pdf) to increase and enforce consistency across oneDNN.

### Assumptions

- Initially, the majority of kernels will continue to use the old method of register handling.
- ABI compliance is key - call-clobbered registers are allocated first, despite the jit_generator.hpp kernel conservatively saving general registers 16, 17 & 19-28, irrespective of if they are touched by a kernel.
- There is no specific order that registers should be allocated in their respective sets.

### Implementation

The following implementation notes is derived from the RFC that I developed alongside the project.

**Overview**

Below I've included a diagram describing the basic structure of the sets in the manager, and the methods that move the register indices between them. The * represents one of the three register families (GP, FP, Pred). Each family has these three sets that hold indices which represent the registers within the family. Further, each family has a used set, which simply tracks any register that has been touched by the manager up to this point.

**Data structures, data types and wider design choices**

image

Used std::sets to track register states as shown above. The manager aims to use STL based data structures and functions to ensure the helper is as understandable and compatible as possible.

**Register priority and management** 

For information on design choices for registers discussed below, see the ABI here: https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst. The project includes handling for all 25 available register types for Aarch64.

 - Call Clobbered: caller saved/volatile.

 - Call Preserved: callee saved/non-volatile (stored/restored during pre/postamble).

**Scoped Registers (RAII)**

```
  // scoped register handling with RAII
  // usage: XReg reg11 = rm.alloc<XReg>(11);
  //        auto scoped = rm.makeScoped(reg11);
  // or
  //        auto scoped_reg = rm.makeScoped(rm.alloc<XReg>());
  // reg11 & scoped_reg will free at end of scope when guards' dtors called.
  template<class Reg>
  class Scoped {
    public:
      explicit Scoped(RegPoolManager& rm, Reg r)
        // pointer to allocator, allocate scoped reg at construction & track, unowned = nullptr
        : rm_(&rm), reg_(r) {
          validate_scoped_reg(rm_, reg_);
        }
          
      // if object is owner of scoped reg and goes out of scope, deallocate
      ~Scoped() {
        if (!rm_) return;
        rm_->free(reg_);
      }      

      // disable copy - scoped regs are move only to avoid ownership/double free issues as per RAII
      Scoped(const Scoped&) = delete;
      Scoped& operator=(const Scoped&) = delete;

      // move constructor - used when scoped regs initialised from rvalue (incl. std::move)
      Scoped(Scoped&& other) noexcept
        : rm_(other.rm_), reg_(other.reg_) {
        other.rm_ = nullptr; // set previous owner to no longer own
      } 

      // expose underlying register for implicit use in JIT helpers
      operator const Reg &() const noexcept { return reg_; }
      const Reg &get() const noexcept { return reg_; }

    private:
      RegPoolManager* rm_ = nullptr; // pointer to allocator, initialised as nullptr
      Reg reg_{}; 
  };

  // helper factory - calls Scoped ctor
  template<class Reg>
  inline Scoped<Reg> makeScoped(Reg r) & {
    return Scoped<Reg>(*this, r);
  } 
```

Note: scoped registers should not be manually freed, as this will lead to double free errors and crashes. Users should use std::move to change the scoped guard of a register, and therefore its lifetime.

**Special XRegs**

```
  // helper methods to return special registers as per AArch64 calling convention
  // NOTE: indirect results register (x8) has been added to the call clobbered general pool
  // intra-procedure-call register 0 : x16
  inline XReg _ipc_reg_0() { used_gp.insert(16); return XReg(16); }
  // intra-procedure-call register 1 : x17
  inline XReg _ipc_reg_1() { used_gp.insert(17); return XReg(17); }
  // platform register : x18
  inline XReg _platform_reg() { used_gp.insert(18); return XReg(18); }
  // frame pointer : x29
  inline XReg _frame_pointer() { used_gp.insert(29); return XReg(29); }
  // link register : x30
  inline XReg _link_reg() { used_gp.insert(30); return XReg(30); }
```

**Call Preserved Registers**

In my register manager, for each of the three families of registers (GP/FP/Pred), I use call clobbered registers before using any of the call preserved ones, unless the user specifies a particular register index.

**Error handling & strictness**

The manager assumes that the user knows what they are doing, in so far that it will throw runtime errors if:

- The user attempts to allocate a specific register when it is in use.
- The user attempts to allocate a specific register that is not in the free or preserved sets.
- The user frees a register that is not in the in use set.
- The user enters an a register index outside of 0-31 (or 0-15 for predicate registers).
- The user attempts to add a register to the free GP pool (with add_to_gp_pool) when the register is already in one of the in-use/preserved/free sets.
- There are no registers in free_*_regs or preserved_* when alloc() is called for that a register type in that family.
- The user calls makeScoped on a register that is not in use.
- The user tries to handle a register type whose reg_family is not GP/FP/Pred.

Further, it is important to note that the following should be avoided by users:

Manual freeing of scoped registers can cause issues (double frees etc)
Avoid alias issues when handling registers - avoid: 

            XReg reg1 = ra.gp_reg(R)

            XReg reg2 = reg1

This can cause alias issues. Register objects should be freed if they are no longer used, or made scoped, as discussed earlier.

**Using the manager**

```
  // Usage:
  // XReg x_eg = rm.alloc<XReg>(); // Allocate next available X register (freed by user)
  // rm.free(x_eg);                // Free x_eg
```

**Helpers**

I wanted to keep the manage as lightweight as possible, so I included only the following helpers:

- reg_in_use: this takes a register object and returns if the index of that register type's family is in the in use set.
- *_idx_in_use: this takes an integer representing a register index for one of gp/fp/pred and returns if that index is in the in use set for the given register family.
- makeScoped: takes a register that is being tracked in the relevant in use set and creates an RAII guard as discussed earlier.
- add_to_gp_pool: method that takes an XReg and adds it to the free pool for tracking by the manager - implemented in case a kernel needs to use one of the special registers discussed earlier in its tracking.
- Special register helpers: these helpers return x16, x17, x18, x29 and x30 as discussed above.
- Getters for the in use, free, preserved and used sets for each register family.

### DIFF
--- a/src/cpu/aarch64/jit_generator.hpp
+++ b/src/cpu/aarch64/jit_generator.hpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2016 Intel Corporation
 * Copyright 2020-2025 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@
 #include "cpu/aarch64/cpu_isa_traits.hpp"
 
 #include "cpu/jit_utils/jit_utils.hpp"
+#include "xbyak_aarch64_reg_manager.hpp"
 
 #if defined(_WIN32) && !defined(__GNUC__)
 #define STRUCT_ALIGN(al, ...) __declspec(align(al)) __VA_ARGS__
@@ -64,12 +65,14 @@
             mn(op, (mask), ptr(addr_off(addr, off, X_DEFAULT_ADDR, X_TMP_0))); \
     } while (0)
 
-static const size_t CSIZE = sizeof(uint32_t);
+static const size_t CSIZE_generator = sizeof(uint32_t);
 
 namespace dnnl {
 namespace impl {
 namespace cpu {
 namespace aarch64 {
+
+using Xbyak_aarch64::RegPoolManager;
 
 // TODO: move this to jit_generator class?
 namespace {
@@ -762,7 +765,7 @@ private:
         if (!is_initialized()) return nullptr;
         const uint8_t *code
                 = reinterpret_cast<const uint8_t *>(CodeGenerator::getCode());
-        register_jit_code(code, getSize() * CSIZE);
+        register_jit_code(code, getSize() * CSIZE_generator);
         return code;
     }
 
@@ -779,6 +782,7 @@ private:
 protected:
     virtual void generate() = 0;
     const uint8_t *jit_ker_ = nullptr;
+    RegPoolManager rm;
 };
 
 } // namespace aarch64
--- a/src/cpu/aarch64/xbyak_aarch64_reg_manager.hpp
+++ b/src/cpu/aarch64/xbyak_aarch64_reg_manager.hpp
@@ -1,0 +1,548 @@
+/*******************************************************************************
+* Copyright 2026 Arm Ltd. and affiliates
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+#ifndef CPU_AARCH64_XBYAK_AARCH64_REG_MANAGER_HPP
+#define CPU_AARCH64_XBYAK_AARCH64_REG_MANAGER_HPP
+
+#include <cstdint>
+#include <set>
+#include <stdexcept>
+#include <vector>
+#include "xbyak_aarch64/xbyak_aarch64/xbyak_aarch64.h"
+#include <type_traits>
+
+namespace Xbyak_aarch64 {
+
+// Static definitions for different types of registers in relation to which family they belong to.
+enum class RegFamily { GP, FP, Pred };
+
+template <class RegT>
+struct reg_family;
+
+template <>
+struct reg_family<WReg> {
+    static constexpr RegFamily value = RegFamily::GP;
+};
+template <>
+struct reg_family<XReg> {
+    static constexpr RegFamily value = RegFamily::GP;
+};
+
+template <>
+struct reg_family<BReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<HReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<SReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<DReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<QReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg8B> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg16B> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg4H> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg8H> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg2S> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg4S> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg1D> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<VReg2D> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZReg> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZRegB> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZRegH> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZRegS> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZRegD> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+template <>
+struct reg_family<ZRegQ> {
+    static constexpr RegFamily value = RegFamily::FP;
+};
+
+template <>
+struct reg_family<PReg> {
+    static constexpr RegFamily value = RegFamily::Pred;
+};
+template <>
+struct reg_family<PRegB> {
+    static constexpr RegFamily value = RegFamily::Pred;
+};
+template <>
+struct reg_family<PRegH> {
+    static constexpr RegFamily value = RegFamily::Pred;
+};
+template <>
+struct reg_family<PRegS> {
+    static constexpr RegFamily value = RegFamily::Pred;
+};
+template <>
+struct reg_family<PRegD> {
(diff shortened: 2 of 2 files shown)


vpirogov:
Tagging @uxlfoundation/onednn-cpu-x64, who have quite similar problem.

vpirogov:
@henry-gar, have you considered adding this functionality into upstream [xbyak_aarch64](https://github.com/fujitsu/xbyak_aarch64)?

+@herumi

vpirogov:
+@kawakami-k

kawakami-k:
@henry-gar 
Thank you for the interesting RFC! I haven't had read all the details yet, but I think it's a great idea and a very useful feature. I'd like to merge this function into the original Xbyak_aarch64.

@herumi 
How do you think?

georgen117:
@vpirogov  I created a pull request of what an x64 implementation would look like. 

I used the design submitted by @henry-gar with minimal changes.

https://github.com/uxlfoundation/oneDNN/pull/4606

Ability to compare the implementation for another system could help judge if this is something we want.

vpirogov:
> Ability to compare the implementation for another system could help judge if this is something we want.

Great. Having a unified design would definitely help to get this feature into Xbyak.

georgen117:
A few thoughts after doing the x86 implementation. 

On the x86 implementation I used different name for the register families I used `{gp, vec, opmask}`.

`gp` for general purpose registers is well understood.
`fp` for SIMD/floating point registers does not really work. The registers are just as likely to hold integer values as floating point values. using `vec` for vector I think is more applicable.
`pred` does not makes sense x64 and `opmask` does not make sense for aarch64.  Maybe use `mask`?

This is based on the idea we want a unified design for x64 and aarch64.

In my experimentation I found this really useful for getting registers as needed but I didn't find it helped much for handling registers when calling other functions. It seems the main reason for having base_freed_gp and base_preserved_gp sets is for building a list of registers that needs to be preserved when making function calls / or when you want to preserve the contents of a registers as a callee. I didn't find it intuitive to use. Maybe you could provide a short example.

georgen117:
**note** I originally developed the samples using x86 code then manually translated it to aarch64 so it may have unintended issues. I hope the overall concept is understood.

I would propose adding some additional query methods:

```cpp
// Returns only in-use registers that are volatile (caller-saved)
// These MUST be saved by the caller before making a function call
std::vector<int> get_in_use_volatile_gps() const;

// Returns only in-use registers that are preserved (callee-saved)  
// These will be saved by the callee if it uses them
std::vector<int> get_in_use_preserved_gps() const;
```

Currently, `get_in_use_gps()` returns **all** in-use registers regardless of calling convention. When generating code that makes function calls, developers need to know:

1. **Which in-use registers are volatile (caller-saved)?**  --> Must be saved before the call if it needs the values after the call.
2. **Which in-use registers are preserved (callee-saved)?** --> Callee must save them if it uses them

Without these methods, developers must manually filter `get_in_use_gps()` by checking against `base_free_gp()` and `base_preserved_gp()`, which is error-prone and duplicates logic.

Current Approach (Inefficient)
```cpp

RegPoolManager rm;

// Allocate some registers
auto r1 = rm.alloc<XReg>();
auto r2 = rm.alloc<XReg>();
auto r3 = rm.alloc<XReg>(); // Might be a preserved register
auto r4 = rm.alloc<XReg>(); // Might be a preserved register

// Before making a function call, save ALL in-use registers
auto in_use = rm.get_in_use_gps();

for (int idx : in_use ) {
    push(XReg(idx)); // INEFFICIENT: Saves preserved registers unnecessarily!
}

call_external_function();

//Restore all registers
for (auto it = in_use.rbegin(); != in_use.rend(); ++it) {
    pop(XReg(*it));
}

// safe to use the popped registers.
rm.free(r1);
rm.free(r2);
rm.free(r3);
rm.free(r4);
```

**Problem:** We're saving/restoring preserved registers that the callee will handle anyway. This wastes stack space and instructions.

Proposed Approach (Optimal)

```cpp

RegPoolManager rm;

// Allocate some registers
auto r1 = rm.alloc<XReg>();
auto r2 = rm.alloc<XReg>();
auto r3 = rm.alloc<XReg>();
auto r4 = rm.alloc<XReg>();

// Save only volatile (caller-saved) registers
auto volatile_regs = rm.get_in_use_volatile_gps();

for (int idx : volatile_regs ) {
    push(XReg(idx); // Only save what the caller is responsible for
}

call_external_function();

//Restore registers
for (auto it = volatile_regs.rbegin(); != volatile_regs.rend(); ++it) {
    pop(XReg(*it)
}

// Preserved registers still have their values (callee handled them)

rm.free(r1);
rm.free(r2);
rm.free(r3);
rm.free(r4);
```
 
**Benefit:** Only saves registers that are actually at risk, avoiding redundant work.

Use Case: Callee Side
`get_in_use_preserved_gps()` helps when implementing a function (callee):

```cpp
void generate_optimized_function() {
    RegPoolManager rm;
    
    // Try to use only volatile registers for scratch work
    auto temp1 = rm.alloc<XReg>();  // volatile
    auto temp2 = rm.alloc<XReg>();  // volatile
    
    // ... do some work ...
    
    // Need more registers? Check if we'll use preserved ones
    auto preserved_in_use = rm.get_in_use_preserved_gps();
    
    if (preserved_in_use.empty()) {
        std::cout << "Optimized path: no prologue/epilogue needed!" << std::endl;
        // No need for function prologue/epilogue!
    } else {
        // Must save preserved registers we're using
        std::cout << "Saving " << preserved_in_use.size() << " preserved regs" << std::endl;
        
        for (int idx : preserved_in_use) {
            push(XReg(idx));
        }
        
        // ... function body ...
        
        for (auto it = preserved_in_use.rbegin(); it != preserved_in_use.rend(); ++it) {
            pop(XReg(*it));
        }
    }
}
```

kawakami-k:
@henry-gar @vpirogov 
Files within oneDNN whose names begin with Xbyak_aarch64 originate from [Xbyak_aarch64](https://github.com/fujitsu/xbyak_aarch64). Therefore, I believe it would be best to first incorporate xbyak_aarch64_reg_manager.hpp into Xbyak_aarch64, and then reflect it in oneDNN. Since xbyak_aarch64_reg_manager.hpp provides a generically useful feature, I strongly want to merge it into Xbyak_aarch64.

henry-gar:
Hi all - thanks for the interest and feedback! I'm working with @jondea currently to get this approved and upstreamed to Xbyak_aarch64, as was suggested, and then hopefully get it here as well.

@georgen117 - thank you for your thoughts! The x86 implementation looks great. As for you more specific thoughts:

- I'll discuss naming conventions with Jonathan before I upstream to Xbyak_aarch64.

> It seems the main reason for having base_freed_gp and base_preserved_gp sets is for building a list of registers that needs to be preserved when making function calls / or when you want to preserve the contents of a registers as a callee. I didn't find it intuitive to use. Maybe you could provide a short example.

My reasoning for this design choice was for future development of and with Xbyak_aarch64 - in oneDNN (specifically src/cpu/aarch64/jit_generator.hpp) the convention is to conservatively save/restore all call-preserved registers as well as x16 & x17 at every call of pre/postamble. My goal was to move away from jit_generator.hpp governing the calling convention, and instead rely on the ABI. 

I also looked into a look ahead/dynamic way of preserving only call-preserved registers if they are touched by the kernel, similar to you - this was outside my scope for this project however but it was my intention to lay the framework for future development in this area. That being said, your generate_optimized_function() seems like a sensible approach, so I will discuss that with @jondea as well.

herumi:
@kawakami-k 
Sorry for the slow response because I’ve been on a conference trip. I’ll take a look when I have some time.

vpirogov:
Upstream PR: https://github.com/fujitsu/xbyak_aarch64/pull/114.

openmodelzooci:

kawakami-k:
@openmodelzooci 
I'll revie this patch at https://github.com/fujitsu/xbyak_aarch64/pull/114

vpirogov:
> I'll revie this patch at [fujitsu/xbyak_aarch64#114](https://github.com/fujitsu/xbyak_aarch64/pull/114)

This bot is actually asking an administrator of Jenkins test automation to confirm that this PR is safe to test. I do not know who this Jenkins automation belongs to, so I banned the bot for nuisance. Also there's no reason to explain yourself to bots :)

jondea:
This [went into Xbyak_aarch64](https://github.com/fujitsu/xbyak_aarch64/pull/114) and into oneDNN here #5206

PR #4589, opened 2026-01-27
Level Zero: align style and shuffling code around

dzarukin:
This is a preparational PR with discrepancies identified along the way while developing Level Zero runtime support.


dzarukin:
make test

rjoursler (approved):

PR #4590, opened 2026-01-28
Fix perf regression in some brgemm convolutions

ankalinin:
This PR revert part of changes from commit 719786c1dc54f841ec413eb060ab099d8a3e9a6c to fix [MFDNN-14568](https://jira.devtools.intel.com/browse/MFDNN-14568) - performance regression for some AMX convolutions.
Adding a TODO comment to clarify how the `estimate_brgemm_ur` function should handle kernel initialization for the `use_uker` setting. The comment outlines the need to update the logic to use unrolled kernel initialization when appropriate and to improve efficiency calculations.

PR #4594, opened 2026-01-28
Fix perf regression in some brgemm convolutions on AMX

ankalinin:
This is follow-up of #4590 (fix [MFDNN-14568](https://jira.devtools.intel.com/browse/MFDNN-14568) - performance regression for some AMX convolutions):
This PR refines the efficiency estimation logic in the `brg_blocking_t::est_eff()` function, adapting it to AMX.

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_utils.cpp
@@ -870,8 +870,11 @@ dim_t brg_blocking_t::grid_coverage(
 }
 
 float brg_blocking_t::est_eff() {
-    const auto brgemm_microkernel_eff = (static_cast<float>(adj_ocblock) * ur)
-            / ((ur + adj_ocblock) * max_regs);
+    const auto N_regs = static_cast<float>(adj_ocblock);
+    const auto M_regs = is_amx(isa) ? div_up(ur, amx_h) : ur;
+    const auto tot_regs = is_amx(isa) ? brgemm_desc_t::AMX_TILES_NUM : max_regs;
+    const auto brgemm_microkernel_eff
+            = (N_regs * M_regs) / ((N_regs + M_regs) * tot_regs);
 
     const auto ur_eff = static_cast<float>(sp_block) / rnd_up(sp_block, ur);
     const auto brgemm_eff = squeeze_val(ur


ankalinin:
make test
set test_scope=NIGHTLY

disable os_win

disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022

disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

disable benchdnn_all
enable benchdnn_conv

tprimak (approved):

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -870,8 +870,11 @@ dim_t brg_blocking_t::grid_coverage(
 }
 
 float brg_blocking_t::est_eff() {
-    const auto brgemm_microkernel_eff = (static_cast<float>(adj_ocblock) * ur)
-            / ((ur + adj_ocblock) * max_regs);
+    const auto N_regs = static_cast<float>(adj_ocblock);
+    const auto M_regs = is_amx(isa) ? div_up(ur, 16) : ur;
+    const auto tot_regs = is_amx(isa) ? 8 : max_regs;
```
georgen117: Should we replace the magic number `8` with a constant `TILE_REG_COUNT` or `AMX_TILE_COUNT` for readability?

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -870,8 +870,11 @@ dim_t brg_blocking_t::grid_coverage(
 }
 
 float brg_blocking_t::est_eff() {
-    const auto brgemm_microkernel_eff = (static_cast<float>(adj_ocblock) * ur)
-            / ((ur + adj_ocblock) * max_regs);
+    const auto N_regs = static_cast<float>(adj_ocblock);
+    const auto M_regs = is_amx(isa) ? div_up(ur, 16) : ur;
```
georgen117: Should the magic number `16` be replaced by a constant?  something like `ROWS_PER_TILE` or `AMX_ROWS_PER_TILE` for readability and maintenance?

src/cpu/x64/jit_brgemm_conv_utils.cpp
georgen117: Instead of using const auto `tot_regs = is_amx(isa) ? 8 : max_regs;` Would it make sense to update the value assigned to `max_regs` instead?

src/cpu/x64/jit_brgemm_conv_utils.cpp
ankalinin: I replaced the magic numbers 8 and 16 in this patch with appropriate constants. I would prefer do not change `max_regs` now because it’s used elsewhere. 
But you’re right overall—we should replace all these numbers scattered throughout the code with constants. That should be a separate pull request.

georgen117 (approved):

PR #4595, opened 2026-01-28
[GPU] Simplify quantization attribute handling

Simonsays095:
The `quant_params` struct duplicates many of the variables stored in `jit::pd_t`. By moving this struct up to this level, we can skip a lot of duplication and potential inconsistencies - particularly related to `swap_ab`. This gets us a good chunk of the way toward handling `swap_ab` more intentionally and consistently.

### DIFF
--- a/src/gpu/intel/gemm/jit/gen_kernel.cpp
+++ b/src/gpu/intel/gemm/jit/gen_kernel.cpp
@@ -430,12 +430,12 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
         int eu_count, bool has_systolic, bool is_integrated, compute_mode mode,
         int batch_dims, bool trans_a, bool trans_b, bool trans_co, bool swap_ab,
         const quant_params &a_quant, const quant_params &b_quant,
-        const quant_params &c_quant, bool dst_sround, bool c_offset, bool bias,
-        sum_ab_t reduce_ab, float alpha, float beta, data_type_t a_type,
-        data_type_t b_type, data_type_t c_type, data_type_t co_type,
-        data_type_t acc_type, int align_a, int align_b, int align_c, dim_t m,
-        dim_t n, dim_t k, dim_t lda, dim_t ldb, dim_t ldc, dim_t batch,
-        gpu_post_ops_t &&post_ops) {
+        const quant_params &c_quant, bool mx_scales, bool dst_sround,
+        bool c_offset, bool bias, sum_ab_t reduce_ab, float alpha, float beta,
+        data_type_t a_type, data_type_t b_type, data_type_t c_type,
+        data_type_t co_type, data_type_t acc_type, int align_a, int align_b,
+        int align_c, dim_t m, dim_t n, dim_t k, dim_t lda, dim_t ldb, dim_t ldc,
+        dim_t batch, gpu_post_ops_t &&post_ops) {
     using namespace ngen;
     using namespace kcatalog;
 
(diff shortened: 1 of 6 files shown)


src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -571,14 +571,20 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
         problem_.autoTypeConversions(hw_, has_systolic);
 
     if (problem_.needsAGroupSums()) {
-        problem_.Tag = convert_dnnl_to_kernel_type(a_quant.gs_type);
+        data_type_t gs_dt = a_quant.gs_type == data_type::undef
```
kealan-barbieri, line 574: It seems like it should be an error for a problem to require group sums but not have `gs_type` intitialized, do we have some valid case where its happening?

kealan-barbieri: I see now this is reproducing the existing behavior - it doesnt need to be addressed for this PR.

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -241,45 +242,60 @@ status_t pd_t::init_attrs() {
...
+    a_quant.scales_type = a_scales.get_data_type();
+    a_quant.zp_type = a_zps.get_data_type();
+    a_quant.gs_type = a_gs.get_data_type();
+    a_quant.force_gs = !a_gs.has_default_values();
+    a_quant.zp_host_scalar = a_zp_host_scalar();
+    const auto &set_a_groups
+            = [](quant_params &quant, const quant_entry_t &entry) -> status_t {
+        int k_grp = entry.get_group(0);
+        int m_grp = entry.get_group(1);
+        if (quant.group_k > 0 && quant.group_k != k_grp)
```
kealan-barbieri, line 264: nit: it would be good to have a comment somewhere making explicit the requirement that multiple quant attributes with groups must have the same grouping.

kealan-barbieri (approved):

src/gpu/intel/gemm/jit/pd.cpp
Simonsays095: Done

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph

src/gpu/intel/gemm/jit/pd.hpp
```diff
@@ -165,31 +169,31 @@ struct pd_t : public gemm::pd_t {
...
-    bool with_b_scales() const { return (bsc_dims_ >= 0); }
+    bool with_a_scales() const { return (a_quant.scale_ndims >= 0); }
+    bool with_b_scales() const { return (b_quant.scale_ndims >= 0); }
     bool with_c_scales() const {
         return !attr()->scales_.has_default_values(DNNL_ARG_DST);
     }
 
     bool with_a_zero_points() const {
-        return (swap_ab_ ? bo_dims_ >= 0 : ao_dims_ >= 0);
+        return (swap_ab_ ? b_quant.zp_ndims >= 0 : a_quant.zp_ndims >= 0);
```
echeresh, line 179: I noticed that this one is different from `with_a_scales()` as this swaps A/B, what is the long term plan here?
If it's intentional, we could name it with `eff_` prefix to emphasize this difference, similar to `eff_a_type()`.

Simonsays095: I left a comment in this PR where I had to copy `a_quant` and `b_quant` specifically because of this swapping. I'm going to make a PR soon to address this issue - ultimately we should be swapping in two places:
1. Right before we make the `gemmstone::problem` during kernel selection
2. Swap a few pointers in `execute`.

In this way, all of our `eff_*` functions can go away and this confusion will be gone. But since `swap_ab` has invaded code in several places at several different times, this is a bit larger of an issue so I want to address it separately.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4518, opened 2026-01-09
xe: gemm: revert gemmstone API changes

Simonsays095:
Recent changes alter a user-facing API in gemmstone (the kernel selector's select function) which other teams use. Since the change in API is unnecessary, this PR reverts that change and instead adds a new API:
- `select`: Earlier API, unchanged. Returns the lowest-scoring (best) matching kernel from the catalog.
- `collect_kernels`: Returns a vector of entries, sorted by score. Also includes additional information in the form of `EvaluateAuxOutput` and score, for later use.

This also is the first step to simplifying `gen_desc_t`, as the `set_entry` function sets both `aux_params_` and `entry`, guaranteeing consistency in the state.

PR #4599, opened 2026-01-28
xe: gemm: jit: introduce pd_t::init_GEMMProblem

rjoursler:
Separate `GEMMProblem` and GEMMStrategy creation into their own functions.

### DIFF
--- a/src/gpu/intel/gemm/jit.hpp
+++ b/src/gpu/intel/gemm/jit.hpp
@@ -70,21 +70,13 @@ struct gen_t : public primitive_t {
             dev_info_ = intel_engine->device_info();
             arch_ = dev_info_->gpu_arch();
             int stepping = dev_info_->stepping_id();
-            VDISPATCH_GEMM_SC(init_attrs(), VERBOSE_UNSUPPORTED_TAG);
 
             const auto d = desc();
 
             CHECK(set_default_formats(false));
-
-            with_sround_ = attr()->rounding_mode_.get(DNNL_ARG_DST)
-                    == rounding_mode::stochastic;
+            CHECK(jit::pd_t::init(engine));
 
             // If m = 1, swap A/B to use more efficient n = 1 kernels if possible.
-            eff_lda_ = d->lda();
-            eff_ldb_ = d->ldb();
-            eff_transa_ = d->transa() == dnnl_trans;
-            eff_transb_ = d->transb() == dnnl_trans;
-
             bool check_lda = ((d->transa() == dnnl_notrans && d->lda() == 1)
                     || (d->transa() == dnnl_trans));
             swap_ab_ = (d->m() == 1 && d->ldc() == 1 && check_lda)
@@ -194,12 +186,6 @@ struct gen_t : public primitive_t {
                                        c_stride == 1 || c_stride % 2 == 0),
                         VERBOSE_SHAPE_RESTRICTION);
             }
-
-            VDISPATCH_GEMM(scales_ok(), VERBOSE_UNSUPPORTED_SCALES_CFG);
-            VDISPATCH_GEMM(zp_ok(), VERBOSE_UNSUPPORTED_ZP_CFG);
-            VDISPATCH_GEMM(gs_ok(), VERBOSE_UNSUPPORTED_PR_CFG);
-
-            VDISPATCH_GEMM_SC(init_post_ops(), VERBOSE_UNSUPPORTED_POSTOP);
 
             bool with_binary = (post_ops_.find(binary) != -1)
                     || (post_ops_.find(prelu) != -1);
--- a/src/gpu/intel/gemm/jit/pd.hpp
+++ b/src/gpu/intel/gemm/jit/pd.hpp
@@ -54,6 +54,25 @@ status_t transfer_post_ops(gemmstone::GEMMProblem &problem,
 struct pd_t : public gemm::pd_t {
     using gemm::pd_t::pd_t;
 
+    // Assumes desc() was already initialized with default formats
+    status_t init(impl::engine_t *engine) {
+
+        with_sround_ = attr()->rounding_mode_.get(DNNL_ARG_DST)
+                == rounding_mode::stochastic;
+
+        eff_lda_ = desc()->lda();
+        eff_ldb_ = desc()->ldb();
+        eff_transa_ = desc()->transa() == dnnl_trans;
+        eff_transb_ = desc()->transb() == dnnl_trans;
+
+        VDISPATCH_GEMM_SC(init_attrs(), VERBOSE_UNSUPPORTED_TAG);
+        VDISPATCH_GEMM(scales_ok(), VERBOSE_UNSUPPORTED_SCALES_CFG);
+        VDISPATCH_GEMM(zp_ok(), VERBOSE_UNSUPPORTED_ZP_CFG);
+        VDISPATCH_GEMM(gs_ok(), VERBOSE_UNSUPPORTED_PR_CFG);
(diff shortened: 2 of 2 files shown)


src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -721,9 +516,8 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
     // but only if there are no grouped scales (in these cases,
     // we apply scales before dpas, and we must use fp dpas)
     bool allow = gpu_utils::dev_getenv("ALLOW_IACC", true);
-    bool is_int
-            = types::is_integral_dt(a_type) && types::is_integral_dt(b_type);
-    if (a_quant.scale_ndims < 1 && b_quant.scale_ndims < 1 && is_int && allow) {
+    bool is_int = problem.Ta_ext.isInteger() && problem.Tb_ext.isInteger();
+    if (problem_.asPtrDims < 1 && problem_.bsPtrDims < 1 && is_int && allow) {
```
h-sadia, line 520: are the scales being stored in "asPtrDims"? Is a variable change or is the var type also changing?

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -435,6 +437,282 @@ bool pd_t::valid_2d_mask(int mask, int ndims, bool per_tensor_ok) {
...
+    if (swap_ab()) {
+        std::swap(a_quant, b_quant);
+        std::swap(a_quant.group_m, a_quant.group_n);
+        std::swap(b_quant.group_m, b_quant.group_n);
+    }
+
+    problem.Ta = problem.Ta_ext = convert_dnnl_to_kernel_type(a_type);
+    problem.Tb = problem.Tb_ext = convert_dnnl_to_kernel_type(b_type);
+    problem.Tc = convert_dnnl_to_kernel_type(acc_type);
+    problem.Tc_ext = convert_dnnl_to_kernel_type(c_type);
```
h-sadia, line 586: Could you explain why the selection of _ext for a data type var?

src/gpu/intel/gemm/jit/gen_kernel.cpp
rjoursler: The scales are being split into a few places in `GEMMProblem`. For the `A` buffer these are:
```
Ta_scale -> scale type
A_scale -> Layout information for the scale buffer
asPtrDims -> number of dimensions in the scale buffer
aqGroupK, aqGroupM -> group sizes for the scales
```
This section of code is just replacing the oneDNN version of scales with what is already stored in the GEMMProblem to maintain one source of truth and avoid unnecessary inputs to select_kernel.

src/gpu/intel/gemm/jit/pd.cpp
rjoursler: This is just the gemmstone API. The `_ext` values represent [the in memory](https://github.com/uxlfoundation/oneDNN/blob/9e4010530b5803cbd651ba549ecbcee3df4e41c5/src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp#L168) data type, which `{a,b,c}_type` refer to.

rjoursler:
make test
disable test_device_cpu

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -435,6 +437,282 @@ bool pd_t::valid_2d_mask(int mask, int ndims, bool per_tensor_ok) {
...
+    auto trans_co = eff_trans_bias();
+    auto dst_sround = with_sround_;
+    bool c_offset = with_c_zero_points();
+    bool bias = with_bias();
+    auto reduce_ab = eff_sum_ab();
+
+    jit::quant_params a_quant = this->a_quant;
+    jit::quant_params b_quant = this->b_quant;
+
+    if (swap_ab()) {
```
Simonsays095, line 577: Kind of off-topic, but just a comment on `swap_ab`: eventually, we want to initialize the problem without `swap_ab`, and instead decide to swap or not during kernel selection. The idea being that we would use `swap_ab` only when a matching kernel exists in the catalog and has a better score. Then it would be stored in `kernel_desc_` where we'd have access to it during `execute` for swapping of runtime parameters/pointers.

Your changes are equivalent to the existing behaviour and fine for now, I just wanted to share what I'm thinking for next steps.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -449,147 +383,8 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
...
-        data_type_t gs_dt = b_quant.gs_type == data_type::undef
-                ? data_type::s32
-                : b_quant.gs_type;
-        problem_.Tbg = convert_dnnl_to_kernel_type(gs_dt);
-        problem_.Bg.layout = MatrixLayout::N;
-        problem_.Bg.setAlignment(problem_.Tbg.paddedSize());
-        if (problem_.aqGroupK == 0) problem_.aqGroupK = problem_.bqGroupK;
-        if (problem_.bqGroupK == 0) problem_.bqGroupK = problem_.aqGroupK;
-    }
+    problem_ = problem;
```
Simonsays095, line 387: I don't like that we still change state in `select_kernel` by setting `problem_`. Alternatively we could pass in the problem to other `kernel_desc_t` functions like we do with these changes for `select_kernel`, and avoid storing it locally. But that has other drawbacks.

Simonsays095 (approved):

src/gpu/intel/gemm/jit/pd.cpp
rjoursler: > The idea being that we would use swap_ab only when a matching kernel exists in the catalog and has a better score. Then it would be stored in kernel_desc_ where we'd have access to it during execute for swapping of runtime parameters/pointers.

Makes sense. If doing this, it sounds like the kind of information should be passed aroudn via gemmstone::DriverInfo, although I am not clear on the exact details required get this information registered.

src/gpu/intel/gemm/jit/gen_kernel.cpp
rjoursler: Agreed. I mostly left this here to keep the current work flow in an abundance of caution as my main focus is elsewhere. Ideally, we would just initialize `problem_` as part of `pd_t::init`, but we cannot trivially do that due to the heuristics we use in problem selection..

src/gpu/intel/gemm/jit/pd.cpp
h-sadia: Is this swap_ab what you made a fix for recently @Simonsays095 ?

h-sadia (approved):

src/gpu/intel/gemm/jit/pd.cpp
Simonsays095: It's the same feature, but this is a different set of variables getting swapped. One major issue with the `swap_ab` feature right now is that it's broken into pieces, scattered across the code like this. It will be cleaned up in my next PR on the matter :)

Simonsays095: > it sounds like the kind of information should be passed aroudn via gemmstone::DriverInfo

I think it makes more sense to add it to `MatchParams` and the return value of `select`. That way we tell the selector "look for this pattern, but also the pattern with swapped A/B is OK". Then the selector could return "this kernel matches your pattern, but it does require swapping A/B". Right now `select` just returns an entry pointer, but soon (once I get the segfault in #4518 figured out) we'll have a struct that we can put this information into.

PR #4604, opened 2026-01-29
cpu: fixes potential integer handling issues

xuxinzen:
Partially fixes Coverity issues


xuxinzen:
make test

xuxinzen:
> To be straightforward, none of the changes in this PR are needed, as they either prevent from non-existent situations or create disputes which would trigger a chain of other unnecessary changes. I suggest rather close Coverity hits with "Intentional" mark as these are not false-positives.

Ok, closing it.

PR #4605, opened 2026-01-29
Fix segfaults in brgemm kernels in case of large shapes

asimonov1:
The PR fixes a number of crashes found during analysis of [MFDNN-14477 GPU Large buffer tests segfaults in CPU primitive](https://jira.devtools.intel.com/browse/MFDNN-14477).

Some problems that crashed are (tested with SPR or GNR CPU if "--engine=cpu"):
```
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb64ic9oc176379826
./tests/benchdnn/benchdnn --mode=R --ip --skip-impl=avx10 --engine=cpu --dt=u8:s8:u8 mb64ic9oc176379826
./tests/benchdnn/benchdnn --mode=R --ip --skip-impl=avx10,avx512 --engine=cpu --dt=u8:s8:u8 mb64ic9oc176379826
./tests/benchdnn/benchdnn --mode=R --ip --skip-impl=avx10,avx512,avx2 --engine=cpu --dt=u8:s8:u8 mb64ic9oc176379826

./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb8ic1oc536870912
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb4ic1oc1073741824

./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --stag=ab --wtag=ab --dtag=ab mb2270183416ic1oc2
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --stag=ab --wtag=ab --dtag=ab mb2147483648ic1oc2

./tests/benchdnn/benchdnn --mode=R --ip --skip-impl=avx10,avx512 --engine=cpu --dt=u8:s8:u8 mb64ic1oc176379817
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 --attr-post-ops=sum:0.5 mb64ic9oc176379826
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=f8_e4m3:f8_e4m3:f32 --attr-post-ops=sum:0.5 mb64ic9oc176379826
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 --attr-scales=src:common:0.25+wei:per_oc+dst:common:2.25 --attr-post-ops=sum:0.5+mul:u8 mb4ic1oc1073741824
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 --attr-scales=src:common:0.25+wei:per_oc+dst:common:2.25 --attr-post-ops=add:u8:per_oc mb4ic1oc1073741824
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 --attr-scales=src:common:0.25+wei:per_oc+dst:common:2.25 --attr-post-ops=add:u8:per_oc mb4ic1oc2147483648

./tests/benchdnn/benchdnn --mode=C --ip --engine=gpu --dir=BWD_W --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab mb1ic15oc149971097
```

Now, the following tests for GPU passed:
```./tests/benchdnn/benchdnn --ip --engine=gpu --batch=test_ip_large_gpu```
Reproducers from [MFDNN-13437](https://jira.devtools.intel.com/browse/MFDNN-13437) are processed correctly too.

#### Comments to changes:
- src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp: a value in register r11 (=reg_A, reg_long_offt) was overwritten in some cases when that was not allowed. This was reproducible with
```
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb64ic9oc176379826
```
- src/cpu/x64/brgemm/brgemm.cpp: most brgemm kernels do not support large (>INT_MAX) strides; a-,b-,c- strides were disabled earlier; this PR disables large d-strides for non-uker kernels (uker kernel should work correctly as it has been fixed in jit_brgemm_amx_uker.cpp). This was reproducible with
```
./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb4ic1oc1073741824
```
- src/cpu/x64/brgemm/brgemm_utils.cpp, src/cpu/x64/jit_brgemm_inner_product_utils.cpp: added checks for dimension downconversions from dim_t to int. Example of a fixed crash (reproduced on SPR):
```./tests/benchdnn/benchdnn --mode=R --ip --engine=cpu --dt=u8:s8:u8 mb4ic1oc2147483648```
If any dimension of an input tensor is larger than INT_MAX then brgemm ip implementation is skipped. Intermediate calculations could also overflow. One issue was related to div_up function; this has been fixed in another PR earlier. Several changes were added to src/cpu/x64/jit_brgemm_inner_product_utils.cpp to avoid overflows in calculations (mostly in cases when expressions like oc*ic, oc+os are used). 
Changes in src/cpu/x64/brgemm/brgemm_utils.cpp are not significant after fixing overflows on caller side but they looks  useful.
-  src/cpu/x64/matmul/brgemm_matmul.cpp, src/cpu/x64/matmul/brgemm_matmul_utils.cpp: 'int' is replaced with 'dim_t' to avoid overflows in offset calculations and, e.g., in this case
```
        brgemm_matmul_conf_t bgmmc; // and assign bgmmc.M>INT_MAX
        // ...
        const matmul_avx512_blocking_params_t::matmul_params_t matmul(
                bgmmc.M, bgmmc.N, bgmmc.K, bgmmc.batch);
```
This was reproducible using:
```
./tests/benchdnn/benchdnn -v5 --mode=R --ip --engine=cpu --stag=ab --wtag=ab --dtag=ab mb2270183416ic1oc2
```
- src/cpu/x64/jit_brgemm_transpose_utils.cpp: support of large offsets is added because jit_trans_to_vnni_t is used by uker brgemm kernel (which is supposed to work with large offsets). This was reproduced with this test on GPU:
```
./tests/benchdnn/benchdnn --mode=C --ip --engine=gpu --dir=BWD_W --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab mb1ic15oc149971097
```

### DIFF
--- a/src/cpu/x64/jit_brgemm_inner_product_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_inner_product_utils.cpp
@@ -18,6 +18,7 @@
 #include "common/math_utils.hpp"
 
 #include "cpu/x64/brgemm/brgemm.hpp"
+#include "cpu/x64/brgemm/brgemm_utils.hpp"
 
 namespace dnnl {
 namespace impl {
@@ -33,6 +34,8 @@ using namespace prop_kind;
 using namespace data_type;
 
 namespace brgemm_inner_product_utils {
+
+using brgemm_utils::safe_dim_to_int;
 
 // Returns amount of work on a thread when using parallel reduction.
 static int comp_work(
@@ -121,8 +124,10 @@ int jit_brgemm_ip_conf_t::get_os_block(
             const bool small_work_amt_per_thread
                     = min_nb_os * jbgp.nb_oc < 1.8f * jbgp.nthr;
             if (small_work_amt_per_thread)
-                max_os_block = saturate(16, max_os_block,
-                        div_up(jbgp.os * jbgp.nb_oc, 2 * jbgp.nthr));
+                max_os_block
+                        = static_cast<int>(saturate<dim_t>(16, max_os_block,
+                                div_up(static_cast<dim_t>(jbgp.os) * jbgp.nb_oc,
+                                        2 * jbgp.nthr)));
         }
     } else if (is_bwd_d) {
         int plat_max_os_block = 0;
@@ -136,7 +141,9 @@ int jit_brgemm_ip_conf_t::get_os_block(
         }
         max_os_block = nstl::min(plat_max_os_block, jbgp.os);
         min_os_block = is_amx_xf16 ? 16 : is_avx512 ? 6 : 4;
-        if (jbgp.isa == avx2 && jbgp.os * jbgp.oc > 512 * 1024) return jbgp.os;
+        if (jbgp.isa == avx2
+                && static_cast<dim_t>(jbgp.os) * jbgp.oc > 512 * 1024)
+            return jbgp.os;
 
     } else if (is_bwd_w) {
         constexpr int amx_xf16_row = 64;
@@ -722,8 +729,10 @@ status_t jit_brgemm_ip_fwd_conf_t::init_conf(cpu_isa_t isa,
     jbgp.N_tail = jbgp.oc % jbgp.oc_block;
     jbgp.K_tail = jbgp.use_buffer_a ? 0 : jbgp.ic % jbgp.K;
 
-    jbgp.LDA = jbgp.use_buffer_a ? jbgp.K * jbgp.gemm_batch_size
-                                 : jbgp.ic_without_padding * jbgp.ks();
+    const auto LDA = jbgp.use_buffer_a
+            ? static_cast<dim_t>(jbgp.K) * jbgp.gemm_batch_size
+            : static_cast<dim_t>(jbgp.ic_without_padding) * jbgp.ks();
+    CHECK(safe_dim_to_int(jbgp.LDA, LDA));
     jbgp.LDB = jbgp.N;
     jbgp.LDD = jbgp.oc_without_padding;
     jbgp.LDC = jbgp.LDD;
@@ -742,8 +751,8 @@ status_t jit_brgemm_ip_fwd_conf_t::init_conf(cpu_isa_t isa,
     if (jbgp.is_bf32) {
         const float M = static_cast<float>(jbgp.M);
         const float N = nstl::min<float>(jbgp.N, jbgp.oc);
-        const float K
-                = nstl::min<float>(jbgp.K * jbgp.gemm_batch_size, jbgp.ic);
+        const float K = nstl::min<float>(
+                static_cast<dim_t>(jbgp.K) * jbgp.gemm_batch_size, jbgp.ic);
         const float tmul_efficiency = (M / 16) * (N / 16) * (K / 32);
         // TODO: Adjust blocking such that bigger M, N, K are generated.
         if (one_of(true, M <= 8, K <= 8, N < 16, tmul_efficiency <= 2.25))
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by asimonov1** [comment]
make test
set test_scope=NIGHTLY

File: src/cpu/x64/matmul/brgemm_matmul_utils.cpp [hunk 1]
```diff
@@ -1100,29 +1101,29 @@ float compute_blocking_heuristic_avx2_f32(brgemm_matmul_conf_t &bgmmc,
...
+    dim_t min_m_blk = nstl::min((dim_t)32, matmul.M);
 
     int n_blk = bgmmc.N_blk;
     const int n_chunks = div_up(matmul.N, n_blk);
     const int max_n_chunks = bgmmc.use_buffer_a ? 16 : 1;
     const int n_chunks_start = nstl::min(max_n_chunks, n_chunks);
 
     int default_k_blk = 1024;
-    int k_blk = nstl::min(matmul.K, default_k_blk);
+    int k_blk = nstl::min(matmul.K, (dim_t)default_k_blk);
```
georgen117: nitpik:
the `default_k_blk` is not used anywhere else why not change the type of `default_k_blk` instead.

also it can be a `constexpr`

`constcepr dim_t default_k_blk = 1024;`

do we want to rely on compiler coercion or be explicit?

`int k_blk = static_cast<int>(nstl::min(matmul.K, (dim_t)default_k_blk));`

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1034,24 +1035,24 @@ float compute_blocking_heuristic_avx2(brgemm_matmul_conf_t &bgmmc,
         matmul_avx512_blocking_params_t &best_blocking) {
     const int nthr = bgmmc.nthr;
 
-    const int max_m_blk = nstl::min(/*64*/ 256, matmul.M);
-    int min_m_blk = nstl::min(32, matmul.M); // max_m_blk
+    const int max_m_blk = nstl::min(/*64*/ (dim_t)256, matmul.M);
+    int min_m_blk = nstl::min((dim_t)32, matmul.M); // max_m_blk
```
georgen117: nitpik: do you want to explicitly cast the results to `int`? or just rely on compiler type coercion?

i.e.

```cpp
const int max_m_blk = static_cast<int>(nstl::min(/*64*/ (dim_t)256, matmul.M));
int min_m_blk = static_cast<int>(nstl::min((dim_t)32, matmul.M)); // max_m_blk
```

asimonov1:
make test

src/cpu/x64/jit_brgemm_transpose_utils.cpp
```diff
@@ -1365,6 +1365,7 @@ struct jit_trans_to_vnni_t : public jit_brgemm_trans_to_vnni_t,
     zmm vidx1 = zmm31;
 
     reg32_t regw_tmp = r15d;
+    reg64_t reg_long_offt = r15;
```
georgen117, line 1368: Just asking because this is stretching my understanding of the code. 

Doesn't this risk the same register corruption that was solved by your commit [fix reg_A corruption](https://github.com/uxlfoundation/oneDNN/pull/4605/changes/4549205e3ea51dd690a9c041184c37c08da3ad90)

I don't see any `reg64_savable_guard_t`. Is it possible this could cause register corruption by mixing 32-bit and 64-bit uses of the same register?

src/cpu/x64/jit_brgemm_inner_product_utils.cpp
```diff
@@ -921,7 +930,8 @@ status_t jit_brgemm_ip_bwd_d_conf_t::init_conf(cpu_isa_t isa,
     jbgp.LDA = jbgp.use_buffer_a ? jbgp.K * jbgp.nb_oc_blocking
                                  : jbgp.oc_without_padding;
```
georgen117: At lines 732-734 you cast `jbgp.k` and `jbgp.oc_without_padding` to `dim_t` to widen the data type for the forward path but did not change it here for the backward path.

Whas this an oversite? Should the calculation be widened here as well?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
asimonov1: Let it be explicit (updated).

src/cpu/x64/jit_brgemm_inner_product_utils.cpp
asimonov1: K is equal to oc_block here, and oc_block is calculated in get_adjusted_oc_block(). This function returns something like 64, 32, 4*simd_w, etc.
So it was intentionally. I changed it; let it be done the same way everywhere.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
asimonov1: Ok. By the way, this default_k_blk was reassigned several lines below, but the new value was not used, I removed it now.

asimonov1:
make test

src/cpu/x64/jit_brgemm_transpose_utils.cpp
asimonov1: In jit_trans_to_vnni_t, the register regw_tmp (i.e. r15d) is used in 2 lambdas kmovw, kmovd, which are used at almost the beginning of generate() to prepare masks kFFFF and mask_tail. I do not see other usages of r15d in jit_trans_to_vnni_t. The register r15 is also not referenced by other register members of jit_trans_to_vnni_t. So I have not found any conflicts.
Another fixed case with r11 and rbx in jit_brgemm_amx_uker_base_t is very different. The register rbx is referenced as reg_tmp_gpr, reg_BS, reg_bias, reg_bias, reg_scales, reg_dst_scales, and many others. The registry r11 is referenced as reg_A and reg_long_offt. So we have to carefully use them.

And by the way, it is also should be mentioned that 2 these lines together
```
    auto ptr_C = EVEX_compress_addr_safe(reg_C, c_offset, reg_long_offt);
    auto ptr_D = EVEX_compress_addr_safe(reg_D, d_offset, reg_long_offt);
```
is always a bug if we use ptr_C, reg_long_offt>INT_MAX, and c_offset!=d_offset because both EVEX_compress_addr_safe add
```
    mov(tmp_reg, offt);
```
and the 2nd one overwrites what is done by the 1st one:
```
    mov(reg_long_offt, c_offset);
    mov(reg_long_offt, d_offset);
```

asimonov1:
make test

asimonov1:
make test

asimonov1:
make test
set test_scope=NIGHTLY

PR #4596, opened 2026-01-28
[draft for testing purposes, do not review] grouped gemm support

mzhukova:
nothing to see here for now :)

PR #4615, opened 2026-01-30
introduce grouped gemm support as experimental feature

mzhukova:
This PR follows recommended proposal from [RFC](https://github.com/uxlfoundation/oneDNN/pull/4412):
- Introduce grouped gemm support via existing matmul primitive and new memory type (grouped memory that is used to resolve the src, dst with 0 dynamic dimension)
- Feature is intended as experimented for now, so changes are guarded with `DNNL_EXPERIMENTAL_GROUPED_MEMORY`
- Items from RFC that could still be changed (appreciate feedback on that):
  - `offsets` for `dst` could be either input parameter (this allows re-using the same `offsets` values/memory ptr as for `src`) or could be computed by the library (therefore requiring user to allocate and provide two different buffers: one for `src` and one for `dst`) -- currently 1st option is used, see example
  - using scales for grouped memory (`src`, `dst`) currently implies different thinking (e.g., setting certain mask means "apply this behavior to each group" vs "apply to the whole memory"), could be resolved by changing grouped tensor to be perceived as "3D" (i.e. `{num_groups, total_M, K}`) -- currently 1st option is used, see example
- benchdnn update is WIP, will be introduced in a separate PR, this PR was tested using [this branch](https://github.com/uxlfoundation/oneDNN/pull/4596)

Supported in ref code for CPU/GPU:
- f32/bf16/f16 `src`/`wei`/`dst`/per-expert bias (shape is `[num_experts, N]`)
- row-wise (or per-token) `src` scales 
- column-wise or grouped (grouped along `K` dim) `wei` scales 
- transposed `wei` 
- (CPU only) s8/u8 `src`/`wei` w  row-wise src and grouped wei scales

### DIFF
--- a/.github/automation/x64/build_linters.sh
+++ b/.github/automation/x64/build_linters.sh
@@ -41,6 +41,7 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
           -DDNNL_EXPERIMENTAL=ON \
           -DDNNL_EXPERIMENTAL_PROFILING=ON \
           -DDNNL_EXPERIMENTAL_UKERNEL=ON \
+          -DDNNL_EXPERIMENTAL_GROUPED_MEMORY=ON \
           -DONEDNN_EXPERIMENTAL_LOGGING=ON \
           -DDNNL_CPU_RUNTIME=$CPU_RUNTIME \
           -DDNNL_GPU_RUNTIME=$GPU_RUNTIME \
@@ -49,7 +50,7 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
       set +x
     elif [[ "$GITHUB_JOB" == "pr-format-tags" ]]; then
       set -x
-      cmake -B../build -S. -DONEDNN_BUILD_GRAPH=OFF
+      cmake -B../build -S. -DONEDNN_BUILD_GRAPH=OFF -DDNNL_EXPERIMENTAL_GROUPED_MEMORY=ON
       set +x
     else
       echo "Unknown linter job: $GITHUB_JOB"
--- a/scripts/generate_dnnl_debug.py
+++ b/scripts/generate_dnnl_debug.py
@@ -188,6 +188,16 @@ def enum_abbrev(enum):
     }.get(enum, def_enum)
 
 
+# Map (enum_name, enum_value) to macro_name for experimental features
+experimental_features = {
+    # ~~~~~~~enum name~~~~~~    ~enum value~     ~~~~~~~~~~macro name~~~~~~~~~~
+    (
+        "dnnl_sparse_encoding_t",
+        "dnnl_grouped",
+    ): "DNNL_EXPERIMENTAL_GROUPED_MEMORY",
+}
+
+
 def sanitize_value(v):
     if "undef" in v:
         return "undef"
@@ -235,7 +245,13 @@ def func_to_str(enum, values):
     signature = func_to_str_decl(enum)
     func_blocks = []
     for v in values:
-        func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
+        # Add conditional compilation for experimental features
+        if (enum, v) in experimental_features:
+            func_blocks.append(f"#if {experimental_features[enum, v]}")
+            func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
+            func_blocks.append("#endif")
+        else:
+            func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
     if enum == "dnnl_primitive_kind_t":
         func_blocks.append(
             'if (v == dnnl::impl::primitive_kind::sdpa) return "sdpa";'
@@ -282,7 +298,13 @@ def str_to_func(enum, values, is_dnnl=True):
         if "any" in v:
             special_values.append(v)
             continue
-        func_blocks.append(f"CASE({sanitize_value(v)});")
+        # Add conditional compilation for experimental features
+        if (enum, v) in experimental_features:
+            func_blocks.append(f"#if {experimental_features[enum, v]}")
+            func_blocks.append(f"CASE({sanitize_value(v)});")
+            func_blocks.append("#endif")
+        else:
+            func_blocks.append(f"CASE({sanitize_value(v)});")
     func_blocks.append("#undef CASE")
     for v in special_values:
         match = re.search(r"(any|undef)", v)
--- a/src/common/memory_desc.cpp
+++ b/src/common/memory_desc.cpp
@@ -239,27 +239,26 @@ status_t memory_desc_init_with_grouped_encoding(memory_desc_t &memory_desc,
             ndims, dims, data_type, format_kind::undef);
     VCHECK_MEMORY(args_ok, invalid_arguments, VERBOSE_MEM_DESC_CHECK_FAIL);
 
-    VCHECK_MEMORY(
-            group_count > 0, invalid_arguments, "group_count must be positive");
+    VCHECK_MEMORY(group_count > 0, invalid_arguments, VERBOSE_BAD_PARAM,
+            "group_count");
 
     // Validate offsets data type
     VCHECK_MEMORY(offsets_dt == data_type::s32, invalid_arguments,
-            "offsets_dt must be an integer s32 type");
+            VERBOSE_INVALID_DATATYPE, "offsets");
 
     // Validate variable dimension index
     VCHECK_MEMORY(variable_dim_idx >= 0 && variable_dim_idx < ndims,
-            invalid_arguments, "variable_dim_idx must be in range [0, ndims)");
+            invalid_arguments, VERBOSE_BAD_PARAM, "variable_dim_idx");
 
     // Currently only first dimension can be variable
-    VCHECK_MEMORY(variable_dim_idx == 0, unimplemented,
-            "grouped encoding currently only supports variable first dimension "
-            "(M dimension).");
+    VCHECK_MEMORY(variable_dim_idx == 0, unimplemented, VERBOSE_BAD_PARAM,
+            "variable_dim_idx");
 
     for (int d = 0; d < ndims; ++d) {
         VCHECK_MEMORY(dims[d] != DNNL_RUNTIME_DIM_VAL, invalid_arguments,
-                "runtime dimensions not supported for grouped encoding");
-        VCHECK_MEMORY(dims[d] > 0, invalid_arguments,
-                "zero dimensions not supported for grouped encoding");
+                VERBOSE_RUNTIMEDIM_UNSUPPORTED);
+        VCHECK_MEMORY(
+                dims[d] > 0, invalid_arguments, VERBOSE_BAD_DIM, "dims", d);
     }
 
     dim_t K = dims[1]; // Uniform dimension
--- a/src/cpu/matmul/ref_grouped_gemm.cpp
+++ b/src/cpu/matmul/ref_grouped_gemm.cpp
@@ -18,6 +18,8 @@
 
 #if DNNL_EXPERIMENTAL_GROUPED_MEMORY
 
+#include <atomic>
+
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
 #include "common/memory_desc_wrapper.hpp"
@@ -42,10 +44,7 @@ status_t ref_grouped_t::execute(const exec_ctx_t &ctx) const {
     const dim_t group_count = src_grouped.group_count;
     const dim_t K = wei_d.dims()[1];
     const dim_t N = wei_d.dims()[2];
-
-#ifndef NDEBUG
     const dim_t total_M = src_d.dims()[0];
-#endif
 
     const void *src_data = CTX_IN_MEM(const void *, DNNL_ARG_SRC, 0);
     const int32_t *src_offsets = CTX_IN_MEM(const int32_t *, DNNL_ARG_SRC, 1);
@@ -98,6 +97,7 @@ status_t ref_grouped_t::execute(const exec_ctx_t &ctx) const {
     // Parallelize over groups (experts in MoE)
     // Expectation is to see 128-256+ groups, with varying M per group
     // and possibly some empty groups (M == 0)
+    std::atomic<status_t> st(status::success);
     parallel_nd(group_count, [&](dim_t group_id) {
         const dim_t src_offset_start
                 = (group_id == 0) ? 0 : src_offsets[group_id - 1];
@@ -106,17 +106,16 @@ status_t ref_grouped_t::execute(const exec_ctx_t &ctx) const {
                 = (group_id == 0) ? 0 : dst_offsets[group_id - 1];
         const dim_t dst_offset_end = dst_offsets[group_id];
 
-        assert(src_offset_start >= 0);
-        assert(src_offset_end <= total_M);
-        assert(src_offset_end >= src_offset_start);
-        assert(dst_offset_start >= 0);
-        assert(dst_offset_end <= total_M);
-        assert(dst_offset_end >= dst_offset_start);
+        // Validate offsets
+        if (src_offset_start < 0 || src_offset_end > total_M
+                || src_offset_end < src_offset_start || dst_offset_start < 0
+                || dst_offset_end > total_M
+                || dst_offset_end < dst_offset_start) {
+            st = status::invalid_arguments;
+            return;
+        }
 
         const dim_t M = src_offset_end - src_offset_start;
-        const dim_t M_dst = dst_offset_end - dst_offset_start;
-
-        assert(M == M_dst);
         if (M == 0) return; // skip if no rows in this group
 
         const dim_t src_base_idx = src_offset_start * K;
@@ -214,7 +213,7 @@ status_t ref_grouped_t::execute(const exec_ctx_t &ctx) const {
         }
     });
 
-    return status::success;
+    return st;
 }
 
 } // namespace matmul
--- a/tests/benchdnn/dnnl_debug_autogenerated.cpp
+++ b/tests/benchdnn/dnnl_debug_autogenerated.cpp
@@ -71,6 +71,9 @@ dnnl_sparse_encoding_t str2sparse_encoding(const char *str) {
     CASE(csr);
     CASE(packed);
     CASE(coo);
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+    CASE(grouped);
+#endif
 #undef CASE
     if (!strcmp("undef", str) || !strcmp("dnnl_sparse_encoding_undef", str))
         return dnnl_sparse_encoding_undef;


scripts/generate_dnnl_debug.py
atkassen, line 25: For ease of extension, consider:
```suggestion

experimental_features = {
    # ~~~~~~~enum name~~~~~~    ~enum value~     ~~~~~~~~~~macro name~~~~~~~~~~
    ("dnnl_sparse_encoding_t", "dnnl_grouped"): "DNNL_EXPERIMENTAL_GROUPED_GEMM",
}

```

scripts/generate_dnnl_debug.py
```diff
@@ -235,7 +235,13 @@ def func_to_str(enum, values):
     signature = func_to_str_decl(enum)
     func_blocks = []
     for v in values:
-        func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
+        # Add conditional compilation for experimental features
+        if v == "dnnl_grouped":
+            func_blocks.append("#if DNNL_EXPERIMENTAL_GROUPED_GEMM")
+            func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
+            func_blocks.append("#endif")
```
atkassen, line 252:
```suggestion
        if (enum, v) in experimental_features:
            func_blocks.append(f"#if {experimental_features[enum, v]}")
            func_blocks.append(f'if (v == {v}) return "{sanitize_value(v)}";')
            func_blocks.append("#endif")
```

scripts/generate_dnnl_debug.py
```diff
@@ -282,7 +288,13 @@ def str_to_func(enum, values, is_dnnl=True):
         if "any" in v:
             special_values.append(v)
             continue
-        func_blocks.append(f"CASE({sanitize_value(v)});")
+        # Add conditional compilation for experimental features
+        if v == "dnnl_grouped":
+            func_blocks.append("#if DNNL_EXPERIMENTAL_GROUPED_GEMM")
+            func_blocks.append(f"CASE({sanitize_value(v)});")
+            func_blocks.append("#endif")
```
atkassen:
```suggestion
        if (enum, v) in experimental_features:
            func_blocks.append(f"#if {experimental_features[enum, v]}")
            func_blocks.append(f"CASE({sanitize_value(v)});")
            func_blocks.append("#endif")
```

src/gpu/intel/matmul/ref_grouped_gemm.cl
```diff
@@ -0,0 +1,130 @@
...
+        __global const float *wei_scales
+#endif
+) {
+
+    const int group_id = get_global_id(0);
+    const int m = get_global_id(1);
+    const int n = get_global_id(2);
+
+    const int src_start = (group_id == 0) ? 0 : src_offsets[group_id - 1];
+    const int src_end = src_offsets[group_id];
```
Simonsays095, line 70: From the RFC:
> Offsets array is for tracking where each expert starts:
offsets = [0, M0, M0+M1, M0+M1+M2, M0+M1+M2+M3]

This behaviour (documented elsewhere in this PR consistently to the implementation here) is slightly different:
> // offsets is an array of size ngroups, with cumulative values [M_0, M_0 + M_1, M_0 + M_1 + M_2, ...]

The RFC should be updated to reflect this difference.

mzhukova: Good catch! There is a commit missing in RFC, that should have been pushed, I will update shortly.

mzhukova: Fixed now! Thanks for pointing this out.

scripts/generate_dnnl_debug.py
mzhukova: Thank you for the suggestion, this is cleaner! Applied.

scripts/generate_dnnl_debug.py
mzhukova: Applied

scripts/generate_dnnl_debug.py
mzhukova: Applied

.github/automation/x64/build_linters.sh
```diff
@@ -41,6 +41,7 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
           -DDNNL_EXPERIMENTAL=ON \
           -DDNNL_EXPERIMENTAL_PROFILING=ON \
           -DDNNL_EXPERIMENTAL_UKERNEL=ON \
+          -DDNNL_EXPERIMENTAL_GROUPED_GEMM=ON \
```
vpirogov: This should probably be 'grouped matmul', as we don't use GEMM name in oneDNN API or documentation.

mzhukova: The idea for this was that the requested / commonly known operation is called grouped GEMM, and the build options enables grouped GEMM via combination of new memory type/new codepath for matmul primitive. 
I'm worried that saying "grouped matmul" could be confusing, since there is no new grouped_matmul primitive, but rather new approach.

doc/primitives/matmul.md
```diff
@@ -306,3 +306,127 @@ information on sparse encoding.
...
+./benchdnn --matmul --grouped=M:8:800,600,700,500,650,450,550,750 8x512x256
+~~~
+
+`--grouped=M:8:800,600,700,500,650,450,550,750` indicates that dimension 0
+(`M`) varies across 8 groups with the specified sizes.
+
+### Implementation Notes
+
+- Currently, only single dimension `0` can vary.
+- Supported are scales and bias attributes.
```
vpirogov: Does this mean "supported scales attribute values"?

.github/automation/x64/build_linters.sh
vpirogov: Ok, I get the idea. Introducing term GEMM in the documentation and implementation names in addition to matmul feels a bit confusing.
Maybe stick to 'EXPERIMENTAL_GROUPED_MEMORY' for the experimental feature name like we did for sparse functionality?

doc/primitives/matmul.md
mzhukova: Yep, the wording is messed up, will fix. Thanks for catching this!

.github/automation/x64/build_linters.sh
mzhukova: Yeah, I think we could do `...GROUPED_MEMORY` instead, but somewhere in the documentation there still be mentions of grouped gemm in a sense, that one needs to use this and that to have it implemented via oneDNN :) will fix

mzhukova: Updated

doc/primitives/matmul.md
mzhukova: Fixed

mzhukova:
ping @uxlfoundation/onednn-arch, @uxlfoundation/onednn-doc and @uxlfoundation/onednn-gpu-intel to review

.github/automation/x64/build_linters.sh
```diff
@@ -49,7 +50,7 @@ if [[ "$ONEDNN_ACTION" == "configure" ]]; then
       set +x
     elif [[ "$GITHUB_JOB" == "pr-format-tags" ]]; then
       set -x
-      cmake -B../build -S. -DONEDNN_BUILD_GRAPH=OFF
+      cmake -B../build -S. -DONEDNN_BUILD_GRAPH=OFF -DDNNL_EXPERIMENTAL_GROUPED_GEMM=ON
```
vpirogov: (nit) This is not needed as format tags are not affected by this change

doc/primitives/matmul.md
```diff
@@ -308,3 +308,128 @@ information on sparse encoding.
 
 See @ref dev_guide_examples page for a complete list. MatMul examples are listed in the
 [Matrix Multiplication](@ref examples_matmul) section.
+
+@anchor dev_guide_matmul_grouped_gemm
+## Grouped GEMM Support
```
vpirogov, line 309: This section is added at the end of the document after implementation limitatoins and examples. Would it be better to stick at the same level as Sparsity? Examples sections could be merged together.

src/gpu/gpu_matmul_list.cpp
```diff
@@ -45,6 +48,7 @@ namespace {
 constexpr impl_list_item_t impl_list[] = REG_MATMUL_P({
         GPU_INSTANCE_INTEL(intel::matmul::gemm_t)
         GPU_INSTANCE_INTEL(intel::matmul::ref_sparse_t)
+        GPU_INSTANCE_GROUPED(intel::matmul::ref_grouped_gemm_t)
```
vpirogov: Maybe `ref_grouped`?

src/gpu/intel/matmul/ref_grouped_gemm.cl
```diff
@@ -0,0 +1,130 @@
...
+    const float wei_scale = wei_scales[wei_scale_idx];
+    acc *= (ACC_DATA_T)wei_scale;
+#endif
+
+#if WITH_BIAS
+    const long bias_idx = (long)group_id * N + n;
+    acc += BIA_TO_REF(bias[bias_idx]);
+#endif
+
+    // todo: prevent overflow
```
echeresh: Is this comment still relevant? Cast to `long` should be enough, isn't it?

src/common/memory_desc_wrapper.hpp
```diff
@@ -388,7 +396,25 @@ struct memory_desc_wrapper : public c_compatible {
...
+            }
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+            else if (sparse_desc().encoding == sparse_encoding::grouped) {
+                // Grouped encoding has values buffer and offsets buffer
+                switch (index) {
+                    case 0:
+                        // Return size for values.
+                        return nnz() * data_type_size();
+                    case 1: {
+                        // Return size for offsets (group_count + 1 offsets).
```
echeresh: Is it `group_count` or `group_count + 1`. Looks like according to the public header should be `group_count`.

tests/gtests/test_iface_grouped.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+    memory::desc md;
+    ASSERT_NO_THROW(md = memory::desc::grouped(
+                            {total_tokens, K}, dt::f32, 0, ngroups, dt::s32));
+
+    // Size of values buffer (buffer 0): total_tokens * K elements
+    size_t ref_values_size = total_tokens * K * memory::data_type_size(dt::f32);
+    ASSERT_EQ(md.get_size(0), ref_values_size);
+
+    // Size of offsets buffer (buffer 1): ngroups + 1
```
echeresh: `ngroups`?

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -3004,6 +3008,54 @@ struct memory : public handle<dnnl_memory_t> {
...
+        /// - Group count: 3 (number of experts)
+        /// - Values buffer: 9 x 256 elements
+        /// - Offsets buffer: [1, 4, 9] (cumulative token counts)
+        ///
+        /// @param adims Tensor dimensions. For the variable dimension,
+        ///     specify the total size (sum across all groups).
+        /// @param adata_type Data type of tensor elements.
+        /// @param variable_dim_idx Index of dimension that varies between groups.
+        /// @param group_count Number of groups (e.g., number of experts in MoE).
+        /// @param offsets_dt Data type of offsets buffer. Default: s32.
```
echeresh, line 3039: Can you please add that only `s32` is supported?

I see this is checked via ` VCHECK_MEMORY(offsets_dt == data_type::s32, ...)` in `memory_desc.cpp`

src/cpu/matmul/ref_grouped_gemm.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // dst: [total_tokens, N] grouped
+    const auto &src_grouped = src_d.sparse_desc().grouped_desc;
+    const dim_t ngroups = src_grouped.ngroups;
+    const dim_t K = wei_d.dims()[1];
+    const dim_t N = wei_d.dims()[2];
+
+    const void *src_data = CTX_IN_MEM(const void *, DNNL_ARG_SRC, 0);
+    const int32_t *offsets = CTX_IN_MEM(const int32_t *, DNNL_ARG_SRC, 1);
+    const void *wei_data = CTX_IN_MEM(const void *, DNNL_ARG_WEIGHTS);
+    void *dst_data = CTX_OUT_MEM(void *, DNNL_ARG_DST, 0);
```
echeresh, line 52: Since we have dst offsets as well here, does it make sense to compare them against `offsets`? Otherwise they are never used/validated but still required by the API.

cmake/options.cmake
```diff
@@ -213,6 +213,11 @@ option(DNNL_EXPERIMENTAL_UKERNEL
     independently from DNNL_EXPERIMENTAL."
     OFF) # disabled by default
 
+option(DNNL_EXPERIMENTAL_GROUPED_MEMORY
+    "Enable experimental support for grouped gemm. This option works
```
echeresh:
```suggestion
    "Enable experimental support for grouped memory format and grouped GEMM. This option works
```

doc/advanced/grouped.md
```diff
@@ -0,0 +1,101 @@
+Grouped Memory Format for Variable-Size Batching {#dev_guide_grouped_mem}
+=========================================================================
+
+@note This is an [experimental feature](@ref dev_guide_experimental). Build oneDNN
+with `ONEDNN_EXPERIMENTAL_GROUPED_MEMORY=ON` to enable grouped memory support.
+
+Grouped memory format handles data with variable-sized groups, commonly used in
+Mixture-of-Experts (MoE). Unlike regular batching where all in the batch share the same size,
```
echeresh:
```suggestion
Mixture-of-Experts (MoE). Unlike regular batching where all tensors in the batch share the same size,
```

doc/advanced/grouped.md
```diff
@@ -0,0 +1,101 @@
...
+    data_type dtype,               // Data type (e.g., f32, s8)
+    int variable_dim_idx,          // Index of dimension that varies per group
+    int num_groups);               // Number of groups
+~~~
+
+## Creating and Using Grouped Memory
+
+For instance, 2D grouped tensor `[total_M, K]` with `variable_dim_idx = 0`:
+- Dimension `0` (`M`) is variable per group with `total_M`
+    being a sum of all group sizes, number of groups is `N`
```
echeresh:
```suggestion
    being the sum of all group sizes, number of groups is `N`
```

include/oneapi/dnnl/dnnl.hpp
```diff
@@ -983,6 +983,10 @@ struct memory : public handle<dnnl_memory_t> {
         packed = dnnl_packed,
         /// Coordinate Sparse (COO) encoding.
         coo = dnnl_coo,
+    /// Grouped Encoding.
```
echeresh:
```suggestion
        /// Grouped encoding.
```

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -77,6 +77,14 @@ typedef enum {
     dnnl_packed,
     /// Coordinate Sparse Encoding (COO).
     dnnl_coo,
+/// Grouped Encoding represents a tensor where one dimension has variable
+/// size per group.
+/// Stored as concatenated blocks with an offsets specifying the size
+/// of each group along the variable dimension.
+/// Some of the blocks could be empty.
```
echeresh:
```suggestion
    /// Grouped Encoding represents a tensor where one dimension has variable
    /// size per group.
    /// Stored as concatenated blocks with an offsets specifying the size
    /// of each group along the variable dimension.
    /// Some of the blocks could be empty.
```

src/common/memory_desc.cpp
```diff
@@ -209,6 +209,84 @@ status_t memory_desc_init_by_packed_encoding(memory_desc_t &memory_desc,
...
+
+    // Grouped sparse encoding specifics
+    // Uses format_kind::sparse because grouped layout is a form of multi-buffer
+    // representation where values are stored together with metadata
+    // describing the variable structure (similar to CSR with rowptr/colind)
+    md.format_kind = format_kind::sparse;
+    md.format_desc.sparse_desc.encoding = sparse_encoding::grouped;
+    md.format_desc.sparse_desc.nnz = dims[0] /* total_M */ * K;
+    md.format_desc.sparse_desc.metadata_types[0] = offsets_dt;
+    md.format_desc.sparse_desc.grouped_desc.ngroups = group_count;
```
echeresh: Nit: `ngroups` vs `group_count` naming could be aligned for clarity.

src/gpu/gpu_matmul_list.cpp
mzhukova: sure, updated

src/gpu/intel/matmul/ref_grouped_gemm.cl
mzhukova: not relevant anymore, this got cleaned up now :)

src/common/memory_desc_wrapper.hpp
mzhukova: remnants of previous POC, updated everywhere now (hopefully)

tests/gtests/test_iface_grouped.cpp
mzhukova: fixed, thanks for catching this

src/common/memory_desc.cpp
mzhukova: sure, I'll update to `group_count` everywhere

.github/automation/x64/build_linters.sh
mzhukova: It looks like this check does more that just format tags, after oneDNN is configured (previously w/a `-DDNNL_EXPERIMENTAL_GROUPED_GEMM=ON`), the `types.xml` is created that has `dnnl_format_kind_t` -> `dnnl_sparse_encoding_t` -> `dnnl_grouped` value and therefore seems like I need both to configure it with new option and special case in scripts/generate_dnnl_debug.py

I dropped this particular change in [this job](https://github.com/uxlfoundation/oneDNN/actions/runs/21770696317/job/62817394186?pr=4658#logs) to double check

vpirogov: Got it.

doc/primitives/matmul.md
mzhukova: The idea was to separate it for now, as this functionality has its own limitations/supported configs/etc., but I'm not attached to it. 

However, I think "Sparsity" currently is kind of weirdly placed too, for instance, "Implementation Limitations" and "Performance Tips" have nothing to do with sparse matmul (it has its own limitations and performance tips probably not applicable also?), so maybe it would make more sense to place both Sparse Matmul and Grouped Matmul on the same level as those and before "Examples" (since this is just a pointer to the table which is common for regular/sparse/grouped matmul)?

vpirogov: This makes sense.

cmake/options.cmake
mzhukova: updated

doc/advanced/grouped.md
mzhukova: updated

doc/advanced/grouped.md
mzhukova: fixed :)

include/oneapi/dnnl/dnnl.hpp
mzhukova: sure, I put a note into C and C++ APIs

doc/primitives/matmul.md
mzhukova: re-arranged sparse and grouped sections

include/oneapi/dnnl/dnnl.hpp
mzhukova: ah yes, this was clang-format doing, I figured that the reasons was in comment being before `ifdef`, fixed now

include/oneapi/dnnl/dnnl_types.h
mzhukova: fixed

src/common/memory_desc.cpp
mzhukova: should be good now :)

src/common/matmul.cpp
```diff
@@ -37,6 +37,147 @@ using namespace dnnl::impl::types;
...
+
+    // Validate grouped encoding on src and dst
+    VCHECK_MATMUL_UNIMPL(src_d.is_grouped_desc() && dst_d.is_grouped_desc(),
+            VERBOSE_UNSUPPORTED_SPARSE_CFG);
+
+    // Weights should be dense, abc or acb format
+    VCHECK_MATMUL_UNIMPL(!wei_d.is_sparse_desc() && !wei_d.is_grouped_desc(),
+            VERBOSE_UNSUPPORTED_SPARSE_CFG);
+    VCHECK_MATMUL_UNIMPL(
+            wei_d.matches_one_of_tag(format_tag::abc, format_tag::acb),
```
dzarukin, line 75: Matmul supports `format_kind::any` programming model. Is it restricted with this functionality?

src/common/matmul.cpp
```diff
@@ -37,6 +37,147 @@ using namespace dnnl::impl::types;
...
+    const auto &dst_grouped = dst_d.sparse_desc().grouped_desc;
+    const dim_t group_count = src_grouped.group_count;
+
+    VCHECK_MATMUL_UNIMPL(src_grouped.group_count == dst_grouped.group_count,
+            VERBOSE_INCONSISTENT_DIM, "src_group_count",
+            (int)src_grouped.group_count, "dst_group_count",
+            (int)dst_grouped.group_count);
+
+    VCHECK_MATMUL_UNIMPL(
+            dst_grouped.variable_dim_idx == src_grouped.variable_dim_idx,
```
dzarukin, line 89: Would the functionality still work as expected if they both have `variable_dim_idx == 1`?

src/common/matmul.cpp
```diff
@@ -546,6 +687,20 @@ status_t dnnl_matmul_primitive_desc_create(
         const memory_desc_t *bias_desc, const memory_desc_t *dst_desc,
         const primitive_attr_t *attr) {
     auto matmul_desc = matmul_desc_t();
+
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+    const memory_desc_wrapper src_d(src_desc);
+    if (src_d.is_grouped_desc()) {
```
dzarukin, line 693: What would happen for src = any/blocked, wei = any/blocked, dst = grouped?

src/common/memory_desc.cpp
```diff
@@ -209,6 +209,84 @@ status_t memory_desc_init_by_packed_encoding(memory_desc_t &memory_desc,
...
+
+    bool args_ok = memory_desc_sanity_check(
+            ndims, dims, data_type, format_kind::undef);
+    VCHECK_MEMORY(args_ok, invalid_arguments, VERBOSE_MEM_DESC_CHECK_FAIL);
+
+    VCHECK_MEMORY(
+            group_count > 0, invalid_arguments, "group_count must be positive");
+
+    // Validate offsets data type
+    VCHECK_MEMORY(offsets_dt == data_type::s32, invalid_arguments,
```
avmanerikar, line 246: (nit): can be replaced by `VERBOSE_OFFSET_DT_MISMATCH`

src/common/memory_desc.cpp
```diff
@@ -209,6 +209,84 @@ status_t memory_desc_init_by_packed_encoding(memory_desc_t &memory_desc,
...
+    VCHECK_MEMORY(variable_dim_idx >= 0 && variable_dim_idx < ndims,
+            invalid_arguments, "variable_dim_idx must be in range [0, ndims)");
+
+    // Currently only first dimension can be variable
+    VCHECK_MEMORY(variable_dim_idx == 0, unimplemented,
+            "grouped encoding currently only supports variable first dimension "
+            "(M dimension).");
+
+    for (int d = 0; d < ndims; ++d) {
+        VCHECK_MEMORY(dims[d] != DNNL_RUNTIME_DIM_VAL, invalid_arguments,
```
avmanerikar, line 258: (nit): can be replaced by `VERBOSE_RUNTIMEDIM_UNSUPPORTED`

src/common/memory_desc.cpp
```diff
@@ -209,6 +209,84 @@ status_t memory_desc_init_by_packed_encoding(memory_desc_t &memory_desc,
...
+    VCHECK_MEMORY(
+            group_count > 0, invalid_arguments, "group_count must be positive");
+
+    // Validate offsets data type
+    VCHECK_MEMORY(offsets_dt == data_type::s32, invalid_arguments,
+            "offsets_dt must be an integer s32 type");
+
+    // Validate variable dimension index
+    VCHECK_MEMORY(variable_dim_idx >= 0 && variable_dim_idx < ndims,
+            invalid_arguments, "variable_dim_idx must be in range [0, ndims)");
```
avmanerikar: (nit): can be replaced by `VERBOSE_OUT_OF_RANGE_DIMS`

include/oneapi/dnnl/dnnl.h
```diff
@@ -1145,6 +1145,43 @@ dnnl_status_t DNNL_API dnnl_memory_desc_create_with_packed_encoding(
         dnnl_memory_desc_t *memory_desc, int ndims, const dnnl_dims_t dims,
         dnnl_data_type_t data_type, dnnl_dim_t nnz);
 
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+/// Creates a memory descriptor for grouped encoding, that
```
avmanerikar: (nit): The term "grouped encoding" can be mistaken for grouped position encoding in LLMs - perhaps a small note could be added for clarification?

mzhukova: will this work? 

> /// Creates a memory descriptor for grouped memory encoding, that stores multiple independent sub-tensors.

src/common/memory_desc.cpp
mzhukova: I feel like it could be a bit confusing though..? this is about a "variable index" being `[0,... number_of_dimensions)`, but it is not exactly a part of dims (dims for this tensor are `[total_M, K]`

include/oneapi/dnnl/dnnl.h
avmanerikar: Sure. that works great.

src/common/memory_desc.cpp
mzhukova: I guess I could do like `VERBOSE_BAD_PARAM`?

src/common/memory_desc.cpp
mzhukova: `VERBOSE_OFFSET_DT_MISMATCH "%s offsets do not fit into %s datatype"` -- not sure this works..

maybe `VERBOSE_INVALID_DATATYPE "invalid datatype for %s"` instead?

src/common/memory_desc.cpp
avmanerikar: That can be mentioned with the specifier: `(VERBOSE_OUT_OF_RANGE_DIMS, "variable_dim_idx")` so that the message reads as `"out-of-range dimensions for variable_dim_idx"`. But either way is fine.

src/common/memory_desc.cpp
avmanerikar: Yes - that works too.

src/common/memory_desc.cpp
mzhukova: updated this and a few other places to use more standard messages:
```
onednn_verbose,v1,common,create:check,memory, has a bad number of dimensions 3,src/common/memory_desc.cpp:236
onednn_verbose,v1,common,create:check,memory,bad param: group_count,src/common/memory_desc.cpp:242
onednn_verbose,v1,common,create:check,memory,bad param: group_count,src/common/memory_desc.cpp:242
onednn_verbose,v1,common,create:check,memory,bad param: variable_dim_idx,src/common/memory_desc.cpp:250
onednn_verbose,v1,common,create:check,memory,bad param: variable_dim_idx,src/common/memory_desc.cpp:254
onednn_verbose,v1,common,create:check,memory,runtime dimension is not supported,src/common/memory_desc.cpp:258
onednn_verbose,v1,common,create:check,memory,runtime dimension is not supported,src/common/memory_desc.cpp:258
onednn_verbose,v1,common,create:check,memory,bad dimension dims:0,src/common/memory_desc.cpp:260
onednn_verbose,v1,common,create:check,memory,bad dimension dims:1,src/common/memory_desc.cpp:260
onednn_verbose,v1,common,create:check,memory,invalid datatype for offsets,src/common/memory_desc.cpp:246
```

all covered by gtests :)

src/common/memory_desc.cpp
mzhukova: updated, see https://github.com/uxlfoundation/oneDNN/pull/4615#discussion_r2785370331

src/common/matmul.cpp
mzhukova: Currently, the only allow configuration is src = grouped * wei = 3D abc/acb equals dst = grouped, it is checked in `grouped_matmul_desc_init`

For the mentioned configuration the error would be `onednn_verbose,v1,primitive,create:check,matmul,unsupported sparse md configuration,src/common/matmul.cpp:68`

I've extended gtests to cover various patterns (see `iface_grouped_test_t.TestGroupedMatmulPatterns`)

src/common/matmul.cpp
mzhukova: Currently, grouped memory only supports `variable_dim_idx == 0`, checked in the `memory_desc_init_with_grouped_encoding`. I could also add a check into matmul (grouped codepath) if needed.

src/common/matmul.cpp
mzhukova: No, it is not supported for now for wei (I think we could add this later if needed), and for grouped memory it probably doesn't make sense.

src/cpu/matmul/ref_grouped_gemm.cpp
mzhukova: updated to use dst offsets and added validation

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul
run_documentation

src/common/matmul.cpp
mzhukova: OK to resolve this and other threads? :)

PR #4620, opened 2026-02-01
cpu: rv64: gemm: switch f32 GEMM to mx7 RVV JIT

xiazhuozhao:
# Summary
This PR switches the RV64 f32 GEMM path to an **mx7 RVV JIT kernel family**, replacing the previous mixed intrinsic/JIT microkernel structure with a unified JIT-based implementation for the main compute path.

The new kernel family blocks the N dimension by up to 7 columns and keeps the M tile size as a runtime parameter. It also supports all transpose combinations of A and B through the same JIT framework. The main motivation is to improve vector register utilization, increase the number of independent accumulators in flight, and provide a more uniform kernel structure across regular tiles and tail cases. 

The choice of `n_cols = 7` follows directly from RVV register allocation constraints for `f32` with `LMUL = 4`.

RVV provides 32 architectural vector registers. With `LMUL = 4`, one logical vector consumes 4 physical vector registers, so the kernel effectively has:

* 8 vector register groups total
* 7 groups available for C accumulators
* 1 remaining group used for the loaded A vector / temporary reuse during C update

That makes 7 the largest spill-free accumulator blocking factor for this configuration. A smaller N tile underutilizes the vector register file and leaves instruction-level parallelism on the table, while a larger tile would require spilling and would likely lose the intended compute-side benefit.

The existing A-packing / workspace logic is preserved. The main structural change is that the old mixed intrinsic/JIT microkernel organization in `rvv_gemm_f32.cpp` is replaced by a unified JIT-based kernel family, with the kernel interface extended to include runtime `m`, `n_cols`, and transpose-mode-aware dispatch. 

# Performance

The current result shows an average speedup of 1.20x on `sg2044` and 2.37x on `SpacemiT M1`.

In general, the new kernel provides clear gains on many larger and more compute-favorable GEMM shapes, where wider N blocking and higher accumulator count improve utilization of the RVV register file. At the same time, some shape-sensitive cases still show limited benefit or regressions, especially around tail-heavy, boundary-sensitive, or otherwise less favorable configurations.

## speedtest log files:
device: sg2044 VLEN=128 bit
batch: `perf_matmul_training`
[perf_matmul_training_opt.csv](https://github.com/user-attachments/files/26340022/perf_matmul_training_opt.csv)
[perf_matmul_training_raw.csv](https://github.com/user-attachments/files/26340024/perf_matmul_training_raw.csv)
[perf_matmul_training_speedup.csv](https://github.com/user-attachments/files/26340025/perf_matmul_training_speedup.csv)

device: muse pi pro (SpacemiT M1) VLEN=256 bit
batch: `perf_matmul_training`
[spacemit_m1_perf_matmul_training_opt.csv](https://github.com/user-attachments/files/26363604/spacemit_m1_perf_matmul_training_opt.csv)
[spacemit_m1_perf_matmul_training_raw.csv](https://github.com/user-attachments/files/26363605/spacemit_m1_perf_matmul_training_raw.csv)
[spacemit_m1_perf_matmul_training_speedup.csv](https://github.com/user-attachments/files/26363603/spacemit_m1_perf_matmul_training_speedup.csv)

### DIFF
--- a/src/cpu/rv64/gemm/jit_rvv_gemm_kernel.cpp
+++ b/src/cpu/rv64/gemm/jit_rvv_gemm_kernel.cpp
@@ -25,8 +25,12 @@ namespace gemm_utils {
 
 using namespace Xbyak_riscv;
 
-jit_rvv_gemm_kernel_t::jit_rvv_gemm_kernel_t(dim_t m)
-    : jit_generator_t("rv64_gemm_kernel_f32_jit"), m_(m) {
+jit_rvv_gemm_kernel_t::jit_rvv_gemm_kernel_t(
+        dim_t n_cols, bool isTransA, bool isTransB)
+    : jit_generator_t("rv64_gemm_kernel_f32_jit")
+    , n_cols_(n_cols)
+    , isTransA_(isTransA)
+    , isTransB_(isTransB) {
     create_kernel();
 }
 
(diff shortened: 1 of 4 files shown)


openmodelzooci:

zhangfeiv0 (approved):

zhangjian29:
(nit) There is a simple mistake. The speedups should be 1.12x and 1.83x respectively, according to the csv files you provided. Taking the average value directly from speedups is unfair.

xiazhuozhao:
> (nit) There is a simple mistake. The speedups should be 1.12x and 1.83x respectively, according to the csv files you provided. Taking the average value directly from speedups is unfair.

Thank you very much for your advice.

PR #4621, opened 2026-02-02
graph: centralize op kinds, schema, and shape infer functions

TaoLv:
- Currently, we have operator definitions scattered in different abstraction layers in graph component. This is mostly a legacy due to the original multiple backends design. We wanted isolate internal ops which were used by one backend but not another.
- This caused several problems in:
  - Op kinds and schema management, eg. we have op kinds in both dnnl::impl::graph:op_kind:: and dnnl::impl::graph::dnnl_backend::op_kind:: which causes ambiguity.
  - Implementation of functions like op_t::strinfy() or op_t::serialize() as internal ops/attributes are not visible in interface/op.hpp.
  - Code management. Adding a new op almost changes every layer.
- Though these may be resolved by adding another abstraction layer, given that we now have only one backend implementation, we decide to centralize op kinds, attributes, schema, shape inference functions inside the interface layer (src/graph/interface/*) to keep the design simple. This can also raise the bar of defining new internal ops and attributes (compared with hiding them in each backend).
- This PR:
  - is mostly a code refactor and movement, plus some styling fixes. 
  - Moves DNNL backend op kinds and attributes to interface/c_types_map.hpp
  - Moves DNNL backend op schemas to interface/op_def.cpp
  - Moves DNNL backend shape inference functions to interface/shape_infer.hpp/cpp.
  - Renames internal ops with a `_dnnl` prefix to avoid ambiguity.
  - strips some runtime specifics from op schema and keeps them stay in DNNL backend as they are not general op abstractions, but tightly related to primitive API behaviors.
- Next step:
  - Implements general op_t::strinfy() or op_t::serialize() functions which are useful for verbose and debug even for internal ops.
  - Potentially merges the schemas of a public op and internal ops, and reduce duplications. 
-


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/executables/base.cpp
```diff
@@ -31,11 +31,12 @@ void get_arg_indices_for_post_ops(
         if (pops[i]->is_post_sum()) {
             indices.insert({DNNL_GRAPH_ARG_POST_SRC,
                     {indices_t::type_t::input, base_index++}});
-        } else if (pops[i]->get_op()->get_kind() == op_kind::dnnl_binary) {
+        } else if (pops[i]->get_op()->get_kind() == op_kind::_dnnl_binary) {
```
ElaineBao: random place: maybe not in this PR, do you want to remove the dnnl backend namespace?

TaoLv: Do you mean to remove the whole dnnl backend namespace? I think we can discuss that as part of the next step. I personally tend to keep it as an isolation of implementation from interfaces. We may want to give it a better name.

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/executables/base.cpp
dzarukin: Discussed offline, proposing to drop `_dnnl_` altogether from all `op_kind`s.

dzarukin (approved):

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4622, opened 2026-02-02
cpu: rv64: add RVV JIT kernel for layer normalization

xiazhuozhao:
### Summary
Add an RVV JIT kernel for the layer‑norm normalization pass on RISC‑V, keep existing RVV non‑JIT implementation as fallback. Gate JIT usage to f32 forward with scale/shift flags (C or CH) and C<=512 to avoid regressions.  

JIT is gated to C/CH only to avoid regressions in H-only and no-flag paths.

### Performance (benchdnn, sg2044, clang)

  - Command (raw + opt):

```
./benchdnn --lnorm --mode=p --dir=FWD_I --dt=f32:f32 --ss_dt=f32 --flags=,C,H,CH --repeats-per-prb=1 --perf-template=csv --batch=inputs/lnorm/shapes_ci
```

  - Opt vs raw:
      - Average speedup: 1.55×

  - log:
[lnorm_opt_shapes_ci.csv](https://github.com/user-attachments/files/25004191/lnorm_opt_shapes_ci.csv)
[lnorm_raw_shapes_ci.csv](https://github.com/user-attachments/files/25004193/lnorm_raw_shapes_ci.csv)

### DIFF
--- a/src/cpu/rv64/rvv_layer_normalization.hpp
+++ b/src/cpu/rv64/rvv_layer_normalization.hpp
@@ -96,10 +96,21 @@ struct rvv_layer_normalization_fwd_t : public primitive_t {
 
         std::shared_ptr<primitive_desc_t> reorder_pd_;
         memory_desc_t reordered_stat_md_;
+        // The current RVV JIT kernel targets f32 forward paths with scale enabled.
+        // For shift-only or no-flag paths, the existing RVV non-JIT implementation
+        // is more stable and is kept as the default fallback.
+        // We also gate JIT usage by norm_axis() (C dimension) to avoid performance
+        // regressions on large-C cases.
+        // The current threshold is based on CI-validated performance results and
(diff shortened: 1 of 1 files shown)


src/cpu/rv64/rvv_layer_normalization.hpp
```diff
@@ -94,8 +96,20 @@ struct rvv_layer_normalization_fwd_t : public primitive_t {
 
         std::shared_ptr<primitive_desc_t> reorder_pd_;
         memory_desc_t reordered_stat_md_;
+        bool use_jit_ = false;
 
     private:
+        void init_jit_conf() {
```
zhangfeiv0, line 109: We can add a comment here to explain the reasoning, making it easier to optimize later.

zhangfeiv0 (approved):

PR #4626, opened 2026-02-02
Make DNNL_RISCV_USE_{RVV,ZVFH}_INTRINSICS overridable

kxxt:
# Description

Currently oneDNN enables RVV and ZVFH when the compiler supports compiling them.
However, that is not sufficient as the riscv64 board itself may not support those extensions.
Currently there's no way to disable RVV and ZVFH extension if the compiler supports it.

I don't think the feature detection should be based on whether the compiler supports that feature but it should be based on user supplied `-march` flags. But since it has already been implemented this way, this PR at least makes it possible to turn off RVV and ZVFH and removes the inaccurate information from the docs.

To reproduce it: compile oneDNN on a riscv64 board without RVV, and the tests would crash with SIGILL.

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -191,11 +191,6 @@ set(DNNL_ARCH_OPT_FLAGS "HostOpts" CACHE STRING
     - For GNU* Compiler Collection and Clang, the default option is `-msse4.1` which
       behaves similarly to the description above.
 
-    - For Clang and GCC compilers on RISC-V architecture this option accepts `-march=<ISA-string>` flag
-      to control wthether or not oneDNN should be compiled with RVV Intrinsics. Use this option with
-      `-march=rv64gc` or `-march=rv64gcv` value to compile oneDNN with and without RVV Intrinsics respectively.
-      If the option is not provided, CMake will decide based on the active toolchain and compiler flags.
-
     - For all other cases there are no special optimizations flags.
 
     If the library is to be built for generic architecture (e.g. built by a


dzarukin:
Hi @kxxt, please check out [this](https://github.com/uxlfoundation/oneDNN/pull/4466) PR. It looks to me that RV64 is moving away from CMake detection to runtime detection and once it's completely removed, it should address the mentioned problem.

It might be useful to have an exact reproducer of the crash so that when the move happens it would be easy to verify it the problem was solved. Thank you.

kxxt:
> Hi @kxxt, please check out [this](https://github.com/uxlfoundation/oneDNN/pull/4466) PR. It looks to me that RV64 is moving away from CMake detection to runtime detection and once it's completely removed, it should address the mentioned problem.

Thanks! I agree that runtime detection would be better but currently the cmake files would unconditional enable RVV and Zvfh if the compiler supports it.

> 
> It might be useful to have an exact reproducer of the crash so that when the move happens it would be easy to verify it the problem was solved. Thank you.

To reproduce it, compile oneDNN with a compiler that supports RVV and Zvfh on a riscv64 board that does not support RVV and run the tests. The tests would crash as RVV is used.  

> Hi @kxxt ,
> 
> I agree with what @dzarukin said. Currently we have: _compiling with CMake detections to check compiler's abilities and running with `mayiuse()` to check platform's abilities_.
> 
> What do you think on this topic?

But the CMake files does not only perform detections.
It checks whether the compiler supports RVV and Zvfh and sets RV64_MARCH_FLAG to enable them,
regardless of whether the target platform supports  RVV and Zvfh or not.

```
-- Using RV64 march flag: -march=rv64gcv_zvfh
```

kxxt:
Oh, Never mind, it is just that the CMake output is a little misleading.

```
-- Using RV64 march flag: -march=rv64gcv_zvfh
```

I thought this means that the march flag would apply to all compilation units. But that only applies when DNNL_ARCH_OPT_FLAGS is the default value "HostOpts"

kxxt:
But still, setting DNNL_ARCH_OPT_FLAGS to `-march=rv64gc` would fail the compilation:

```
FAILED: [code=1] src/cpu/rv64/CMakeFiles/dnnl_cpu_riscv.dir/rvv_matmul.cpp.o                                                                                                                                                                                
/usr/bin/c++ -DDNNL_DLL -DDNNL_DLL_EXPORTS -DDNNL_ENABLE_CPU_ISA_HINTS -DDNNL_ENABLE_MAX_CPU_ISA -DDNNL_RISCV_USE_RVV_INTRINSICS -DDNNL_RISCV_USE_ZVFH_INTRINSICS -DDNNL_RV64=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/build/onednn/src/oneDNN-3.
10.2/build/include -I/build/onednn/src/oneDNN-3.10.2/include -I/build/onednn/src/oneDNN-3.10.2/third_party -I/build/onednn/src/oneDNN-3.10.2/src -march=rv64gc -mabi=lp64d -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=
format-security         -fstack-clash-protection         -fno-omit-frame-pointer -Wp,-D_GLIBCXX_ASSERTIONS -g -ffile-prefix-map=/build/onednn/src=/usr/src/debug/onednn -flto=auto -fopenmp -fvisibility-inlines-hidden  -Wall -Wno-unknown-pragmas -Wundef 
-fvisibility=internal -march=rv64gc  -fPIC -Wformat -Wformat-security -D_FORTIFY_SOURCE=2 -fstack-protector-strong  -Wmissing-field-initializers  -Wno-strict-overflow -Wno-maybe-uninitialized -Wno-stringop-overflow -Wno-array-bounds  -O3 -DNDEBUG -std=
c++11 -MD -MT src/cpu/rv64/CMakeFiles/dnnl_cpu_riscv.dir/rvv_matmul.cpp.o -MF src/cpu/rv64/CMakeFiles/dnnl_cpu_riscv.dir/rvv_matmul.cpp.o.d -o src/cpu/rv64/CMakeFiles/dnnl_cpu_riscv.dir/rvv_matmul.cpp.o -c /build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/r
vv_matmul.cpp                                                                                                                                                                                                                                               
<command-line>: warning: ‘_FORTIFY_SOURCE’ redefined                                                                                                                                                                                                        
<command-line>: note: this is the location of the previous definition                                                        
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_matmul.cpp: In lambda function:                                             
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_matmul.cpp:94:71: error: argument type ‘vfloat32m1_t’ requires the V ISA extension
   94 |                     float partial = __riscv_vfmv_f_s_f32m1_f32(reduced);                                                                                                                                                                            
      |                                                                       ^                                                                                                                                                                             
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_matmul.cpp:94:71: error: argument type ‘vfloat32m1_t’ requires the V ISA extension                                                                                                                         
In file included from /build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_matmul.hpp:21,                                                                                                                                                                       
                 from /build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_matmul.cpp:16:                                                                                                                                                                       
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_postops.hpp: In member function ‘vfloat32m1_t dnnl::impl::cpu::rv64::rvv_postops_t::apply(vfloat32m1_t, size_t) const’:
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_postops.hpp:45:25: error: return type ‘vfloat32m1_t’ requires the V ISA extension
   45 |     inline vfloat32m1_t apply(vfloat32m1_t v, size_t vl) const {                                                     
      |                         ^~~~~                                                                                                                                                                                                                       
/build/onednn/src/oneDNN-3.10.2/src/cpu/rv64/rvv_postops.hpp:45:25: error: argument type ‘vfloat32m1_t’ requires the V ISA extension            
```

Would it be possible to guard such code with `__attribute__((target("arch=+v")))` if the code is used only after runtime detection?

zhangjian29:
> But still, setting DNNL_ARCH_OPT_FLAGS to -march=rv64gc would fail the compilation:

Yep. That is exactly what we have discussed in [this PR](https://github.com/uxlfoundation/oneDNN/pull/4466). We suppose that we do not need `DNNL_ARCH_OPT_FLAGS` for rv64 right now. Compile as more as toolchains can, and runtime checks can make sure the correct dispatches according to the platform's abilities.

kxxt:
> > But still, setting DNNL_ARCH_OPT_FLAGS to -march=rv64gc would fail the compilation:
> 
> Yep. That is exactly what we have discussed in [this PR](https://github.com/uxlfoundation/oneDNN/pull/4466). We suppose that we do not need `DNNL_ARCH_OPT_FLAGS` for rv64 right now. Compile as more as toolchains can, and runtime checks can make sure the correct dispatches according to the platform's abilities.

But in that case(leaving  `DNNL_ARCH_OPT_FLAGS` to `HostOpts`), then all compilation units will be compiled with `-march=rv64gcv_zvfh` and the compiled binary itself would contain RVV code inserted by the compiler optimization and not controlled by the runtime detection, which explains the SIGILL crashes I saw in the tests when not setting   `DNNL_ARCH_OPT_FLAGS`.

zhangjian29:
Could you please provide the detalied log on this SIGILL crashes?

kxxt:
> Could you please provide the detalied log on this SIGILL crashes?

Sure. I have opened https://github.com/uxlfoundation/oneDNN/issues/4638.

PR #4627, opened 2026-02-02
cpu: aarch64: simplify uni reorder stack

Sqvid:
# Description

The previous design had a pure interface that was only implemented by a monolithic class that had every method copied in. This means that as new assembly kernels (and their helper functions) are added, the size of this class would grow indefinitely. There's also the overhead for a new reader of figuring out which methods are shared helpers vs. which ones are kernel-specific.

This refactor groups all the common functionality (creating loops, data conversion, call operators, etc.) into `jit_uni_reorder_kernel_t`. Developers can then add their specialised kernels by inheriting and overriding the `compute()` method.

I have deliberately not made any changes that would affect the dispatch logic or the generated assembly. This patch is just to provide a framework for those future improvements to be made.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/reorder/jit_uni_reorder_kernel.hpp
+++ b/src/cpu/aarch64/reorder/jit_uni_reorder_kernel.hpp
@@ -31,54 +31,59 @@ namespace impl {
 namespace cpu {
 namespace aarch64 {
 namespace tr {
-struct call_param_t {
-    const void *in = nullptr;
-    void *out = nullptr;
-    const float *src_scales = nullptr;
-    const float *dst_scales = nullptr;
-    int32_t src_zp = 0;
-    int32_t dst_zp = 0;
-    int32_t *compensation_scratch = nullptr;
-};
-
-// The additional structure is needed because
-// using a data structure with tail processing
-// data for non-tail cases reduces kernel
-// performance. This is because there is too
-// much data that has to be transferred to the kernel.
-struct tail_call_param_t {
-    call_param_t base_params;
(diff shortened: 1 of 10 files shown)


src/cpu/aarch64/reorder/jit_uni_reorder_kernel.hpp
```diff
@@ -287,6 +241,113 @@ struct jit_uni_reorder_kernel_f32_t : public kernel_t, public jit_generator_t {
...
+    }
+
+private:
+    void compute(int ndims, int len, bool tail_processing) override;
+    void process_unroll_generic_step(int reg_unroll, const int *i_off,
+            const int *o_off, const int *s_off, const int *c_off,
+            const int *zero_padding, const bool tail_processing);
+};
+
+class tr8x8_kernel_t : public jit_uni_reorder_kernel_t {
```
jondea, line 288: Should each kernel go into its own file?

jondea:
I think it might be worth writing some words somewhere about the structure and what child class can/should override. Also, what is necessary to get it to be called (e.g. adding it to a list somewhere).

I'm sure you've thought this through, but are there any issues with jit_uni_reorder_kernel_t being both the driver and the parent class? E.g. what happens if a child class calls ()? You could end up using a different reorder than the child.

Also, where does jit_blk fit into this? Is doesn't appear to inherit from jit_uni_reorder_kernel_t, but should it?

openmodelzooci:

Sqvid:
Closing this for now. May revisit in the future.

PR #4628, opened 2026-02-02
src: xpu: introduce Level Zero vendor agnostic implementation

dzarukin:
Continuation of #4589.

PR introduces a `ze` folder inside `xpu` that implements vendor agnostic Level Zero-based essential abstractions and infrastructure to function.

This code will be always compiled for SYCL runtime (as a verification part). It's workability will be verified only once all pieces of implementation land to main and benchdnn validation is enabled.

### DIFF
--- a/src/gpu/intel/sycl/l0/utils.cpp
+++ b/src/gpu/intel/sycl/l0/utils.cpp
@@ -14,14 +14,14 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include "xpu/ze/utils.hpp"
+
 #include "gpu/intel/sycl/l0/utils.hpp"
-#include "oneapi/dnnl/dnnl_config.h"
 
 #include "gpu/intel/jit/binary_format.hpp"
 #include "gpu/intel/jit/utils/type_bridge.hpp"
 #include "ngen_level_zero.hpp"
 
-#include "level_zero/ze_api.h"
 #include "level_zero/ze_intel_gpu.h"
 
 #if !defined(__SYCL_COMPILER_VERSION)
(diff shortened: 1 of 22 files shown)


dzarukin:
make test

src/gpu/intel/sycl/l0/utils.cpp
atkassen, line 137: We've got a mix of `l0` and `ze` in function names in this file (also this directory). Can we change them all to `ze` for consistency?

atkassen (approved):

src/gpu/intel/sycl/utils.cpp
```diff
@@ -32,6 +33,15 @@ namespace gpu {
 namespace intel {
 namespace sycl {
 
+// FIXME: Currently SYCL doesn't provide any API to get device UUID so
+// we query it directly from Level0 with the zeDeviceGetProperties function.
+// The `get_device_uuid` function packs 128 bits of the device UUID, which are
+// represented as an uint8_t array of size 16, to 2 uint64_t values.
+xpu::device_uuid_t get_device_uuid(const ::sycl::device &dev) {
+    return xpu::ze::get_device_uuid(
+            xpu::sycl::compat::get_native<ze_device_handle_t>(dev));
```
echeresh, line 42: Can we have the OpenCL backend here? or are we eventually dropping SYCL/OpenCL? (if yes, I believe we still have many places assuming both OpenCL/L0 backends).

src/xpu/ze/engine_factory.cpp
```diff
@@ -0,0 +1,90 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "xpu/ze/engine_factory.hpp"
+
+// #if DNNL_GPU_VENDOR == DNNL_VENDOR_INTEL
+// #include "gpu/intel/ze/engine.hpp"
+// #endif
```
echeresh, line 21: Should be removed?

src/xpu/ze/engine_factory.cpp
```diff
@@ -0,0 +1,90 @@
...
+    CHECK(ze::zeContextCreate(driver, &context_desc, &context));
+
+    return engine_create(engine, driver, device, context, index);
+}
+
+status_t engine_factory_t::engine_create(impl::engine_t **engine,
+        ze_driver_handle_t driver, ze_device_handle_t device,
+        ze_context_handle_t context, size_t index) const {
+    // return gpu::intel::ze::engine_create(
+    //         engine, driver, device, context, index);
```
echeresh, line 90: Should be restored?

src/xpu/ze/engine_factory.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+size_t engine_factory_t::count() const {
+    uint32_t driver_count = 0;
+    CHECK(ze::zeDriverGet(&driver_count, nullptr));
+
+    std::vector<ze_driver_handle_t> drivers(driver_count);
+    CHECK(ze::zeDriverGet(&driver_count, drivers.data()));
+
+    uint32_t device_count = 0;
+    CHECK(ze::zeDeviceGet(drivers[0], &device_count, nullptr));
```
echeresh: Should we check `driver_count == 0` before accessing?

src/xpu/ze/engine_impl.cpp
```diff
@@ -0,0 +1,94 @@
...
+                    nullptr, 1, &ocl_device_.unwrap(), nullptr, nullptr, &err));
+            OCL_CHECK(err);
+        }
+    }
+
+    ze_device_properties_t device_properties = {};
+    device_properties.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
+    device_properties.pNext = nullptr;
+    CHECK(ze::zeDeviceGetProperties(device_, &device_properties));
+    name_ = device_properties.name;
```
echeresh, line 56: Looks like `runtime_version_` should be also initialized somewhere here.

src/xpu/ze/stream_impl.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    ze_event_desc_t event_desc = {};
+    event_desc.stype = ZE_STRUCTURE_TYPE_EVENT_DESC;
+    event_desc.pNext = nullptr;
+    event_desc.index = static_cast<uint32_t>(events_.size());
+    event_desc.signal = ZE_EVENT_SCOPE_FLAG_HOST;
+    event_desc.wait = ZE_EVENT_SCOPE_FLAG_HOST;
+
+    ze_event_handle_t event;
+    ze::zeEventCreate(event_pool_, &event_desc, &event);
```
echeresh:
```suggestion
    auto st = ze::zeEventCreate(event_pool_, &event_desc, &event);
    if (st != status::success) return nullptr;
```

src/gpu/intel/sycl/l0/utils.cpp
dzarukin: It's true. Names will be updated with the next portion of changes that introduces `gpu/intel/ze` side of ze backend. All parts combined of this support are 6.5k of lines total, so I split them in pieces to provide an ability to review it accordingly.

src/xpu/ze/engine_factory.cpp
dzarukin: I kept it as the next portion of changes with bring it back.

src/xpu/ze/engine_factory.cpp
dzarukin: Same here.

src/xpu/ze/engine_factory.cpp
dzarukin: This API actually extracts proper `device_count` if it's passed as `0` to this API. I can leave a comment but it's generally a programming model when it comes to query stuff from Level Zero APIs.

src/xpu/ze/engine_impl.cpp
dzarukin: So far it's empty as you noticed. I had a hard time to identify the API that would return me this information in the same way as SYCL does but directly through Level Zero.

src/gpu/intel/sycl/utils.cpp
dzarukin: It looks like this call was always related to Level Zero backend exclusively. If one wants to refer to OCL one, the proper namespace should be used like [here](https://github.com/uxlfoundation/oneDNN/blob/dzarukin/ze_api/src/gpu/intel/sycl/utils.cpp#L166). So this one doesn't assume any OCL usage.

It looks like we can drop SYCL/OpenCL only when there's a tool to compile kernels directly from SYCL.

src/xpu/ze/engine_impl.cpp
dzarukin: Alright, I peeked the implementation from oneAPI DPC++ Compiler and re-used it here though it looks like it's an undocumented method. Looks like it works:
```
onednn_verbose,v1,info,gpu,runtime:Level Zero
onednn_verbose,v1,info,gpu,engine,level zero device count:2
onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:1.14.36931,binary_kernels:enabled
```

src/xpu/ze/stream_impl.cpp
dzarukin: Makes sense, thanks, applied.

dzarukin:
make test

src/xpu/ze/utils.hpp
```diff
@@ -0,0 +1,251 @@
...
+}
+#undef ZE_LIB_NAME
+
+#define INDIRECT_ZE_CALL(f) \
+    template <typename... Args> \
+    status_t f(Args &&...args) { \
+        const ze_init_flags_t default_ze_flags = 0; \
+        static auto init_ = find_ze_symbol<decltype(&::zeInit)>("zeInit"); \
+        if (!init_) return status::runtime_error; \
+        ZE_CHECK(init_(default_ze_flags)); \
```
echeresh: Looks like `zeInit` is called on every L0 call, is it required?
I would think about wrapping `zeInit` by `std::call_once()` to reduce overhead.

src/xpu/ze/engine_factory.cpp
echeresh: Let me clarify a bit: I see this access: `drivers[0]`. But what if `drivers` is empty here? `drivers[0]` in this case is invalid access, right?

dzarukin: Totally valid concern. I've applied the logic to exit on zero driver count. Thanks for spotting!

src/xpu/ze/utils.hpp
dzarukin: I think I've managed to do just that. Two notes: given that it's unknown which ze call might be the first one, the common place still should be used for zeInit as none of the calls are valid until it's initialized (however, it must be true only for local testing as for interop case it should be already initialized by the user). Secondly, I figured that zeInit is deprecated, and replaced it with advised API. Please check it out and thanks for a suggestion.

src/xpu/ze/engine_impl.cpp
dzarukin: Suddenly, there was a piece of logic that updated runtime_version_ for device info (which will be a part of the next PR) but for some unknown to me reason, it returns different values, I'll need to follow up on this and see what's going on.

dzarukin:
make test

src/xpu/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ endif()
 
 if(DNNL_GPU_VENDOR STREQUAL "INTEL")
     add_subdirectory(ocl)
+    add_subdirectory(ze)
```
rjoursler:
```Suggestion
    if(DNNL_WITH_SYCL)
        add_subdirectory(ze)
    endif()
```

src/xpu/ze/utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+namespace impl {
+namespace xpu {
+namespace ze {
+
+status_t ze_initialize() {
+    static std::once_flag flag;
+    std::call_once(flag, [&] {
+        auto _init_drivers
+                = find_ze_symbol<decltype(&::zeInitDrivers)>("zeInitDrivers");
+        if (!_init_drivers) return status::runtime_error;
```
echeresh: Does this returned status go unused? It seems `call_once()` can't propagate this error so maybe just drop it - I think `find_ze_symbol()` already includes checks and messages

src/xpu/ze/utils.hpp
echeresh: Thanks, looks good to me (left one minor question).

src/xpu/CMakeLists.txt
dzarukin: Updated the logic, thanks.

src/xpu/ze/utils.cpp
dzarukin: Indeed, my bad, for some reason I thought it the call would return. Introduced a ret value which will be updated inside `call_once` and returned to the user. I decided it's probably better have it than not. Thanks.

src/xpu/ze/stream_impl.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+void stream_impl_t::create_event_pool() {
+    ze_event_pool_desc_t event_pool_desc = {};
+    event_pool_desc.stype = ZE_STRUCTURE_TYPE_EVENT_POOL_DESC;
+    event_pool_desc.pNext = nullptr;
+    event_pool_desc.flags = ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
+    if (is_profiling_enabled())
+        event_pool_desc.flags |= ZE_EVENT_POOL_FLAG_KERNEL_TIMESTAMP;
+    event_pool_desc.count = 16 * 1024;
```
mzhukova, line 61: (minor) wondering, what is this based on? (maybe some sort of comment would help)
and also what would happen if this is saturated?

src/xpu/ze/utils.hpp
```diff
@@ -0,0 +1,250 @@
...
+template <typename T>
+struct destroy_traits;
+// {
+//     static void destroy(T t) {}
+// };
+
+template <>
+struct destroy_traits<ze_command_list_handle_t> {
+    static void destroy(ze_command_list_handle_t t) {
+        (void)xpu::ze::zeCommandListHostSynchronize(t, UINT64_MAX);
```
mzhukova, line 173: is this always safe to ignore any possible errors after the first call?

dzarukin: Errors affecting the successful execution flow are not anticipated at this point. Synchronization here is to ensure that if the user who didn't pass their list didn't wait on a stream (for any reason) the list is still finished before destroying it. It is an API requirement the list is not executing at the destruction moment.

src/xpu/ze/stream_impl.cpp
dzarukin: There's no justification for these 16K. Just some number that worked. Specification doesn't say about implications for a larger number though it feels like it should have memory price for more event allocations.

If more than 16K events are issued, an error from Level Zero will pop up leading to execution failure.

Event handling will likely require some touch up anyway because events creation is not thread-safe, they are error-prone if a lot of event needed (as I mentioned above) but we can handle it once there are more requirements come. So far this is sufficient.

src/xpu/ze/utils.hpp
mzhukova: >Errors affecting the successful execution flow are not anticipated at this point

"not anticipated at this point" means during destroy call or at this stage of implementing the support and this will be improved in some future PR?

src/xpu/ze/stream_impl.cpp
mzhukova: Could you please add some comment, so that we know it is random number? :) is touch up planned in a foreseeable future PR?

src/xpu/ze/utils.hpp
dzarukin: The list is created under stream_impl if it's not supplied by the user (which is a more expected scenario, in which case the library doesn't own the list, thus, doesn't destroy it). If it wasn't provided, the list will be destroyed when the stream is destroyed. If user plans to get meaningful results out of the library, the stream must be synchronized (through `stream.wait()`), and then destroyed. In this case the call to synchronize the list does nothing as list must finish all the tasks submitted to it.

The only scenario this synchronize call covers is if something was submitted, stream wasn't waited and destroyed. In such case, whether a list errored on sync during that destruction or not doesn't mean anything. In that sense "at this point" refers to a point of stream end-of-life moment and the list inside of it.

src/xpu/ze/stream_impl.cpp
dzarukin: > Could you please add some comment...

Some comment added, thanks.

> is touch up planned in a foreseeable future PR?

Nope, not until the code hits customer production, and evidence of errors appear with their specific use cases to have a better vision of how to improve it and for what scenarios. It's not a part of leftover code promotion.

src/xpu/ze/utils.hpp
mzhukova: Thanks for the explanation!

PR #4636, opened 2026-02-03
doc: graph: operations: fix dropout formula

TaoLv:


TaoLv:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

PR #4479, opened 2025-12-23
cpu: rv64: add weekly ci test

zhangfeiv0:
# Description

Currently, the CI tests for the RISC-V architecture only run smoke-scale tests and skip a large number of time-consuming test cases. To address this, plan to introduce a weekly test job that runs at CI scale, split across 10 parallel jobs for improved efficiency.

PR #4637, opened 2026-02-03
cpu: rv64: rvv: fix im2col vector path with width padding and re-enable CI test

xiazhuozhao:
# Problem
The RVV GEMM convolution im2col vectorized path (stride_w == 1) may produce incorrect results and can read out-of-bounds when width padding is present (e.g. padR > 0). This was observed in the RISC-V CI with test_convolution_eltwise_forward_f32 showing large mismatches against reference.
```
        Start  67: test_convolution_eltwise_forward_f32
 67/230 Test  #67: test_convolution_eltwise_forward_f32 ............................***Failed    5.26 sec
```

# Root Cause
This issue appears to be introduced by the RVV im2col vectorization added in PR #4548 (commit b6282ee214aa642330014f9caa3e6facfe009e17):
The vectorized path pre-computed the base pointer with duplicated offsets (ih / kw / padding), leading to incorrect addressing. Vector loads were performed without guarding for left/right padding, which can cause out-of-bounds reads on the input row.

# Fix
- Correct the base pointer computation in the im2col vectorized path to avoid duplicated offsets.
- Disable the vectorized copy when width padding is present (lp != 0 || r_pad != 0) and fall back to the existing scalar path, which correctly handles boundaries. This keeps semantics correct while limiting the change to padding cases.
- Re-enable test_convolution_eltwise_forward_f32 in the RISC-V CI. The test is lightweight, keeping it enabled improves regression coverage.

# Performance
Potential minor performance regression only for width-padding cases where the vectorized path is disabled; non-padding paths and other kernels are unchanged.

### DIFF
--- a/src/cpu/rv64/rvv_gemm_convolution_utils.cpp
+++ b/src/cpu/rv64/rvv_gemm_convolution_utils.cpp
@@ -427,6 +427,7 @@ void im2col(const conv_gemm_conf_t &jcp, const data_type_t *__restrict im,
     const dim_t oh_end = last_oh + 1;
     const dim_t first_ow = ss % jcp.ow;
     const dim_t last_ow = (ss + sb - 1) % jcp.ow;
+    const bool no_w_padding = (lp == 0 && jcp.r_pad == 0);
 
     const data_t zero_val = 0;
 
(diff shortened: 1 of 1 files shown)


xiazhuozhao:
I have reverted [7eace69](https://github.com/xiazhuozhao/oneDNN/commit/7eace69e07aa6db08bf16c8f8d32c893f65954f6), which re-enabled test_convolution_eltwise_forward_f32 in CI.

Although this test completes in ~5 seconds locally, it runs significantly slower than expected in the QEMU-based RISC-V GitHub CI environment, taking several minutes.

zhangfeiv0:
A subset of tests has been excluded from the CI environment due to extended runtimes under QEMU. CI strengthening is pending the merger of #4479.

zhangjian29 (approved):

PR #4646, opened 2026-02-04
xe: gemm: revert unintentional dispatch modification

rjoursler:
Fixes [MFDNN-14611](https://jira.devtools.intel.com/browse/MFDNN-14611). 

In 6d45f42c, moving post-op initialization resulted in a change to the output of `with_bias()`. In particular, post-op initialization may move the bias calculation into a post-op, which results with `with_bias()` returning false.

This patch modifies dispatching logic to restore the original behavior due to some observed performance regressions.


rjoursler:
make test
disable test_device_cpu

PR #4648, opened 2026-02-04
[GPU] Refactor A/B swapping Part 1

Simonsays095:
Part 1 of the `swap_ab` refactor (out of 2 or 3) in jit GEMM. This PR focuses on unraveling the roots of `swap_ab` in functions inside `gemm::jit::pd_t` functions. This means that `pd()` returns the unswapped quantity in every case, and swapping is handled externally.

Every function removed is split into its own commit for easier bisecting, in case I missed an edge case. I checked these changes on `test_matmul_gpu` and it passes with no errors, but this logic is historically prone to bugs.

The next PR (part 2) is to simplify the logic on the `execute` side, moving all swapping logic to one place. And then part 3 of the refactor is to defer `swap_ab` logic until kernel selection and store this information in `kernel_desc_`.

### DIFF
--- a/src/gpu/intel/gemm/jit.cpp
+++ b/src/gpu/intel/gemm/jit.cpp
@@ -359,27 +359,27 @@ status_t gen_t::execute(const exec_ctx_t &ctx) const {
     auto a_type = pd()->get_type(DNNL_ARG_A);
     auto b_type = pd()->get_type(DNNL_ARG_B);
     auto c_type = d->c_type();
-    if (swap_ab) std::swap(a_type, b_type);
 
     auto m = into<int32_t>(pd()->desc()->m());
     auto n = into<int32_t>(pd()->desc()->n());
     auto k = into<int32_t>(d->k());
 
-    if (swap_ab) std::swap(m, n);
-
     bool trans_a = pd()->trans_a();
     bool trans_b = pd()->trans_b();
+
+    auto lda = into<int32_t>(pd()->ld(DNNL_ARG_A));
+    auto ldb = into<int32_t>(pd()->ld(DNNL_ARG_B));
+    auto ldc = into<int32_t>(d->ldc());
+    auto ldco = into<int32_t>(pd()->with_bias() ? d->ld_bias() : 0);
+
     if (swap_ab) {
+        std::swap(a_type, b_type);
+        std::swap(m, n);
+        std::swap(lda, ldb);
         std::swap(trans_a, trans_b);
         trans_a = !trans_a;
         trans_b = !trans_b;
     }
-
-    auto lda = into<int32_t>(pd()->ld(DNNL_ARG_A));
-    auto ldb = into<int32_t>(pd()->ld(DNNL_ARG_B));
-    if (swap_ab) std::swap(lda, ldb);
-    auto ldc = into<int32_t>(d->ldc());
-    auto ldco = into<int32_t>(pd()->with_bias() ? d->ld_bias() : 0);
 
     auto alpha = pd()->alpha();
     auto beta = pd()->beta();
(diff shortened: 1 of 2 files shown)


src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -669,7 +664,8 @@ status_t pd_t::init_GEMMProblem(
     CHECK(gpu_post_ops_t::make(
             gpu_post_ops, post_ops_, dst_md(), get_post_op_specializations()));
 
-    CHECK(transfer_post_ops(problem, std::move(gpu_post_ops), swap_ab()));
+    CHECK(transfer_post_ops(problem, std::move(gpu_post_ops)));
+    if (swap_ab()) problem.postOps.transpose();
```
rjoursler, line 668: Can this just be handled as part of the call to `problem.transpose()`? In general, it seems like we should rely on a `GEMMProblem` member function to handle this process, rather than have the logic within oneDNN.

Simonsays095: Using `problem.transpose()` is the eventual goal - but the problem initialization is so intertwined with `swap_ab` logic that I haven't done that yet. We don't call `problem.transpose()` in any case currently, we handle all of the swapping on the oneDNN side.

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -512,28 +502,43 @@ status_t pd_t::init_GEMMProblem(
...
+    auto align_a = align(DNNL_ARG_A);
+    auto align_b = align(DNNL_ARG_B);
+
+    auto lda = ld(DNNL_ARG_A);
+    auto ldb = ld(DNNL_ARG_B);
+
+    auto trans_a = this->trans_a();
+    auto trans_b = this->trans_b();
+
+    if (swap_ab_) {
```
kealan-barbieri, line 521: would it be practical to make  this function a member of `GEMMProblem` in order to totally get rid of swapping within pd?

Simonsays095: This does effectively exist in `GEMMProblem` as the `transpose()` function, and the goal is to use it eventually. I just haven't done that yet, since it has to be done after everything is initialized, which means this block needs to be moved through all the code to the bottom of the function - and then converted to `problem.transpose()`. It's part of the plan for the next PR.

src/gpu/intel/gemm/jit.cpp
```diff
@@ -169,8 +169,9 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
 
     if (pd()->batch_dims() >= 1) {
         for (int i = pd()->batch_dims() - 1; i >= 0; i--) {
-            auto stride_a = int32_t(pd()->eff_stride_a(i));
-            auto stride_b = int32_t(pd()->eff_stride_b(i));
+            auto stride_a = int32_t(pd()->stride(DNNL_ARG_A, i));
```
kealan-barbieri, line 172: We could avoid some swaps if we used an `eff_arg` here like you have for `scale_stride` .  I think it would be worthwhile to have one last remaining `eff_*`  function for this purpose (`eff_a_arg()`/`eff_b_arg()`).

kealan-barbieri:
@rjoursler I think perf testing would be sufficient to verify the same (or better) strategies are being selected and also covers more testcases.

src/gpu/intel/gemm/jit.cpp
Simonsays095: I might be convinced to keep `eff_a_arg`/`eff_b_arg`, although I'd like to remove `swap_ab` from the pd entirely. In my local changes for the next PR, I've already converted this to use the `eff_X_arg`, although I'm thinking of extracting strides into a vector and then swapping A/B strides in `execute` instead of here.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4652, opened 2026-02-05
cpu: x64: matmul: fix initialization of use_buffer_c in amx blocking

xuxinzen:
Fixes MFDNN-14444

Incorrect initialization of use_buffer_c causes failure for the case with sum.


xuxinzen:
make test

xuxinzen:
make test
set test_scope=NIGHTLY

PR #4654, opened 2026-02-05
Fixes for clang-cl driver

vpirogov:
`clang-cl` is a driver shipped with Windows version of Clang (and Intel oneAPI DPC++ Compiler) that mimics MSVC behavior. We have couple of compatibility issues:
* find_package(OpenMP) returns `-Xclang -fopenmp` flag instead of `-openmp`. The first one messes up with SYCL builds because it gets applied to device code.
* Options `-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer` are not supported resulting in warnings.

Fixes MFDNN-14612

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -197,7 +197,8 @@ if(MSVC)
 
     # Generating frame pointers for easier performance profiling
     if(DNNL_TARGET_ARCH STREQUAL "X64")
-        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        # clang-cl driver only supports MSVC flags
+        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT CMAKE_BASE_NAME STREQUAL "clang-cl")
             append(CMAKE_CCXX_FLAGS "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer")
         else()
             append(CMAKE_CCXX_FLAGS "/Oy-")


vpirogov:
make test
disable test_device_cpu
disable test_device_gpu
test_scope=NIGHTLY

cmake/OpenMP.cmake
```diff
@@ -38,26 +38,16 @@ if(DPCPP_HOST_COMPILER_KIND STREQUAL "DEFAULT")
     # the -fsycl option by default so it has to be explicitly disabled.
     set(_omp_original_cmake_cxx_flags "${CMAKE_CXX_FLAGS}")
     string(REGEX REPLACE "-fsycl" "-fno-sycl" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    # FindOpenMP module is not aware of "-openmp" flag supported by clang-cl driver
+    # Instead it finds -Xclang -fopenmp option that causes build errors for SYCL code
+    if(CMAKE_BASE_NAME STREQUAL "clang-cl")
+        set(OpenMP_C_FLAG "-openmp")
+        set(OpenMP_CXX_FLAG "-openmp")
+    endif()
```
dzarukin, line 49: Should `find_package()` come first and then workaround update OpenMP-specific variables? Otherwise, it feels they'll become incorrect again...

cmake/OpenMP.cmake
```diff
@@ -38,26 +38,16 @@ if(DPCPP_HOST_COMPILER_KIND STREQUAL "DEFAULT")
     # the -fsycl option by default so it has to be explicitly disabled.
     set(_omp_original_cmake_cxx_flags "${CMAKE_CXX_FLAGS}")
     string(REGEX REPLACE "-fsycl" "-fno-sycl" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    # FindOpenMP module is not aware of "-openmp" flag supported by clang-cl driver
+    # Instead it finds -Xclang -fopenmp option that causes build errors for SYCL code
+    if(CMAKE_BASE_NAME STREQUAL "clang-cl")
```
dzarukin, line 46: `CMAKE_BASE_NAME` seems to be undocumented variable.  Does regular driver identification work?

vpirogov: `CMAKE_CXX_COMPILER_ID` is Clang for both `clang++` and `clang-cl`. CMake 3.14 introduced [`CMAKE_<LANG>_COMPILER_FRONTEND_VARIANT`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_FRONTEND_VARIANT.html) to distinguish between these. We'll have to bump minimal CMake version to use it.

This is not the first place where we use `CMAKE_BASE_NAME` either. I'll check for alternatives.

cmake/OpenMP.cmake
vpirogov: `OpenMP_CXX_FLAG` gives FindOpenMP.cmake module a hint on what options to test. It will still execute a compile test and indicate that OpenMP is not supported if the test fails.

cmake/OpenMP.cmake
vpirogov: We are using `CMAKE_BASE_NAME` [extensively](https://github.com/search?q=repo%3Auxlfoundation%2FoneDNN%20CMAKE_BASE_NAME&type=code) to specialize for non-standard compiler drivers like `icx` and `dpcpp`. An alternative would be implementing logic to indentify driver by stripping down contents of `CMAKE_CXX_COMPILER`/`CMAKE_C_COMPILER`. This does not seem very useful though.

cmake/OpenMP.cmake
vpirogov: Extended the comment to explain what `OpenMP_*_FLAG` does.

cmake/OpenMP.cmake
dzarukin: Thanks, resolving.

cmake/OpenMP.cmake
dzarukin: Thanks, resolving!

cmake/OpenMP.cmake
```diff
@@ -38,26 +38,19 @@ if(DPCPP_HOST_COMPILER_KIND STREQUAL "DEFAULT")
...
+    endif()
     find_package(OpenMP)
     set(CMAKE_CXX_FLAGS "${_omp_original_cmake_cxx_flags}")
 endif()
 
-# special case for clang-cl (not recognized by cmake up to 3.17)
-if(NOT OpenMP_CXX_FOUND AND MSVC AND CMAKE_CXX_COMPILER_ID MATCHES "(Clang|IntelLLVM)")
-    # clang-cl and icx will fall under this condition
-    # CAVEAT: undocumented variable, may be inappropriate
-    if(CMAKE_BASE_NAME STREQUAL "icx")
```
mdtoguchi, line 49: I'm not familiar with the build requirements, but is use of `icx` no longer part of the build?  When using `icx`, the use of `/Qopenmp` instead of `-openmp`.

vpirogov: @mdtoguchi, this code was needed for old CMake versions. Currently CMake handles `icx` correctly out of the box:
```
>cmake .. -DONEDNN_CPU_RUNTIME=SYCL -DONEDNN_GPU_RUNTIME=SYCL -G Ninja
-- CMAKE_BUILD_TYPE is unset, defaulting to Release
-- The C compiler identification is IntelLLVM 2025.2.0 with MSVC-like command-line
-- The CXX compiler identification is IntelLLVM 2025.2.0 with MSVC-like command-line
<snip>
-- Found OpenMP_C: -Qiopenmp (found version "5.1")
-- Found OpenMP_CXX: -Qiopenmp (found version "5.1")
-- Found OpenMP: TRUE (found version "5.1")
```

The only situation where it needs help to find the right option is `clang-cl` driver.

PR #4655, opened 2026-02-05
xe: ukernel: Enable support for iterating through multiple entries

umar456:
# Description

This PR allows the microkernel selector to iterate through multiple strategies. The current implementation fails if the first entry returned by the select function fails. This change keeps testing additional entries until one passes so that there is a smaller likelihood of failing.

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
+++ b/src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
@@ -254,6 +254,7 @@ std::string generateShim(const Package &package, HostLanguage language,
 
             auto sname = typeName(stypes[i], language, &sizes);
             auto ename = typeName(stypes[i].type, language);
+            auto typedef_name = "_e_" + sname;
             int vlen = divideUp(sizes.blockElements(), options.subgroupSize);
 
             shim << "#ifndef MICRO_DECL_" << sname
@@ -261,10 +262,12 @@ std::string generateShim(const Package &package, HostLanguage language,
                     "#define MICRO_DECL_"
                  << sname
                  << "\n"
+                    "typedef " << ename <<
+                    " __attribute__((ext_vector_type(" << vlen << "))) "
+                 << typedef_name << ";\n"
                     "typedef struct {\n"
                     "    "
-                 << ename;
-            if (vlen > 1) shim << vlen;
+                 << typedef_name << ' ';
             shim << " x[" << sizes.blocks()
                  << "];\n"
                     "} "


src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -205,153 +239,129 @@ Package selectGEMM(const GEMMOptions &options, HWInformation hwInfo, SizeParams
...
-    auto entries = select(catalog, 1, &matchParams, evalParams, auxParams, observer);
-    // First, attempt to find a strategy that matches the protocol's options
-    for(const auto* this_entry: entries) {
+    std::vector<const kcatalog::Entry*> entries = select(catalog, 1, &matchParams, evalParams, auxParams);
+    auto last_entry = std::remove_if(begin(entries), end(entries), [&](const kcatalog::Entry* e) {
+        GEMMStrategy strategy(hw, stepping);
+        strategy.unroll[LoopM] = e->driverInfo.unroll[LoopM];
+        strategy.unroll[LoopN] = e->driverInfo.unroll[LoopN];
+        parseStrategy(e->strategy, hw, problem, strategy);
+        return strategy.kParallelLocal != kParallelLocal;
```
Simonsays095: If `options.kParallelLocal == true`, we should select strategies regardless of whether they use local k-parallelism.
```suggestion
        return !kParallelLocal && strategy.kParallelLocal;
```

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
```diff
@@ -263,8 +263,7 @@ std::string generateShim(const Package &package, HostLanguage language,
                  << "\n"
                     "typedef struct {\n"
                     "    "
-                 << ename;
-            if (vlen > 1) shim << vlen;
+                 << ename << vlen;
```
Simonsays095: I commented this on the other PR, but I don't think size=1 vectors are valid types according to [the spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-vector-data-types).
> Supported values of n are 2, 3, 4, 8, and 16 for all vector data types.

Two questions:
1. Why do we have an Nx1 tile, is this from strategies with unroll[MN]=1?
2. How do we handle this in SDPA?

If we really need this, we can possibly define it as float name[N] instead of floatN name - i.e. arrays instead of vectors.

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -205,153 +239,129 @@ Package selectGEMM(const GEMMOptions &options, HWInformation hwInfo, SizeParams
...
+    });
+    entries.erase(last_entry, end(entries));
+    entries.push_back(nullptr); // Try heuristics if no kernel found
```
Simonsays095: The current algorithm has 3 tries to get a valid strategy:
1. A catalog entry that matches protocol (for now, just the kParallelLocal option)
2. A catalog entry that doesn't match the protocol, but we adjust it to match (by removing kParallelLocal options)
3. Heuristic selection.

This change effectively removes that 2nd attempt. I'm not necessarily opposed to that, but it may cause regressions unless we add non-kparallel versions of the `kr` kernels currently in the catalog. I just want to make sure that this change is intentional. If that's the case, then we don't need the block below that disables k-parallelism.

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -205,153 +239,129 @@ Package selectGEMM(const GEMMOptions &options, HWInformation hwInfo, SizeParams
...
+                REG_XEHP_ISA(ARCH_DISPATCH(XeHP))
+                REG_XEHPG_ISA(ARCH_DISPATCH(XeHPG))
+                REG_XEHPC_ISA(ARCH_DISPATCH(XeHPC))
+                REG_XE2_ISA(ARCH_DISPATCH(Xe2))
+                REG_XE3_ISA(ARCH_DISPATCH(Xe3))
+                default: throw std::runtime_error("Unsupported architecture");
+            }
+            #undef ARCH_DISPATCH
+        } catch (const std::runtime_error &ex) {
+            /* Try next strategy */
```
Simonsays095, line 363: Please log these cases behind `debuginfo=1` (ideally we'd log as a oneDNN warning line, but anything works for now), so we can identify and fix them.

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
umar456: We run into this issue when we are searching for configs for SDPA, especially for smaller sizes. This allows us to use a wider variety of configurations when performing these searches.

We define float1 in [generic_vector_ops.h](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/include/generic_vector_ops.h#L22).

The tiles use both the vector dimension as well as the array indexing to define a tile. See: https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/include/tile_ops.h#L647

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
umar456: Wouldn't it be redundant with the verbose log here: https://github.com/uxlfoundation/oneDNN/pull/4655/changes/BASE..699daf2b37dd78d0230cae7dc2278f317650b165#diff-9c16a2dcef46c8753c4b07d9c5c37dd133e55f8e08a501d3d72568fa450d7637R326

We can add something on line [364 ](https://github.com/uxlfoundation/oneDNN/pull/4655/changes/BASE..699daf2b37dd78d0230cae7dc2278f317650b165#diff-9c16a2dcef46c8753c4b07d9c5c37dd133e55f8e08a501d3d72568fa450d7637R364)below

Simonsays095: I'd like these to show up as warnings even without any verbose options set - but for now I guess the logs you added at the top of this function will work.

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -205,153 +240,134 @@ Package selectGEMM(const GEMMOptions &options, HWInformation hwInfo, SizeParams
...
-    for(const auto* this_entry: entries) {
-        GEMMStrategy strategy(hw, stepping);
-        strategy.unroll[LoopM] = this_entry->driverInfo.unroll[LoopM];
-        strategy.unroll[LoopN] = this_entry->driverInfo.unroll[LoopN];
-        parseStrategy(this_entry->strategy, hw, problem, strategy);
-        if (!kParallelLocal && strategy.kParallelLocal) continue;
-        entry = this_entry;
-        break;
+    std::vector<const kcatalog::Entry*> entries = select(catalog, 1, &matchParams, evalParams, auxParams);
+    if(!kParallelLocal) {
```
kealan-barbieri, line 244: For future work it might be worthwhile to enable fallback between strategy architectures as well.

Simonsays095: The `entries` vector should contain kernels from all architectures, with newer architectures at the start of the vector. This should handle all levels of fallbacks appropriately.

umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

PR #4657, opened 2026-02-06
gpu: ze: prepare the ground: shuffling code and styling

dzarukin:
Continuation of #4628 

This PR doesn't introduce new ze code, just moves stuff here and there with better splitting. List of changes by commits from earliest to latest:
* Split xpu/ocl/utils.cpp into to translation units -> utils_shared.cpp that will be re-used in ze part and original utils.cpp. 
  * The reason for that is utils.cpp depend on ocl::engine_impl_t which is not available for ze backend. It's impossible to untie it or move it from utils.cpp.
  * ze backend requires some OpenCL functionality due to extensions extraction (though I haven't fully verified what extensions are actually essential, this is a future work after backend is entirely supported). Once these extensions are completely available in ZE API, this dependency can be removed and original code placement can be restored. For now ze backend will include ze folder and that one ocl cpp file to have necessary symbols defined.
  * I decided not to move needed functions bodies into a header file to keep compilation times a little bit smaller.
* Introduce a new function in shared ocl utils for future re-use.
* Align SYCL kernel generation with OpenCL kernel generation in terms of debug information.
* The main one - shred former L0 utils into sycl utils and ze utils that will be re-used for engine creation (next PR). It also updates some headers and moves some local functions into unnamed namespaces.
* Styling for names.

### DIFF
--- a/src/gpu/intel/sycl/utils.cpp
+++ b/src/gpu/intel/sycl/utils.cpp
@@ -55,7 +55,7 @@ status_t create_ocl_engine(
             ocl_ctx = xpu::ocl::make_wrapper(
                     xpu::sycl::compat::get_native<cl_context>(*sycl_ctx));
             break;
-        case xpu::sycl::backend_t::level0: {
+        case xpu::sycl::backend_t::ze: {
             cl_device_id d {nullptr};
             CHECK(sycl_dev2ocl_dev(&d, sycl_dev));
             ocl_dev = xpu::ocl::make_wrapper(d, true);
@@ -120,7 +120,7 @@ status_t get_ze_kernel_binary(
 } // namespace
 
 // FIXME: Currently SYCL doesn't provide any API to get device UUID so
-// we query it directly from Level0 with the zeDeviceGetProperties function.
+// we query it directly from Level Zero with the zeDeviceGetProperties function.
 // The `get_device_uuid` function packs 128 bits of the device UUID, which are
 // represented as an uint8_t array of size 16, to 2 uint64_t values.
 xpu::device_uuid_t get_device_uuid(const ::sycl::device &dev) {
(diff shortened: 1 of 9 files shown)


dzarukin:
make test

src/xpu/ocl/utils_shared.cpp
```diff
@@ -0,0 +1,307 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+// Note: this file shares function bodies with Level Zero runtime. The reason
+// it's a dedicated file is an engine_t dependency in utils.cpp that doesn't
+// let it link if compiled for Level Zero backend. Thus, this file contains
+// pure CL code bodeis only.
```
echeresh: Nit:

```suggestion
// pure CL code bodies only.
```

dzarukin: Thanks for the catch, updated.

src/gpu/intel/sycl/utils.cpp
```diff
@@ -33,15 +33,150 @@ namespace gpu {
...
+            ocl_ctx = xpu::ocl::make_wrapper(xpu::ocl::clCreateContext(
+                    nullptr, 1, &d, nullptr, nullptr, &err));
+            OCL_CHECK(err);
+            break;
+        }
+        default: assert(!"not expected"); return status::invalid_arguments;
+    }
+    impl::engine_t *ocl_engine_ptr;
+    size_t index;
+    CHECK(xpu::ocl::get_device_index(&index, ocl_dev));
```
avmanerikar, line 73: (nit): [VERROR_ENGINE](https://github.com/uxlfoundation/oneDNN/blob/93961ce39cb0d546defa138c6f5d561b4ffcfe3c/src/common/engine.hpp#L35) can be utilized here if needed.

avmanerikar (approved):

src/gpu/intel/sycl/utils.cpp
dzarukin: This is an internal, non-user-visible, function used for kernel compilation. If engine creation would have failed, it would do that long earlier it reaches this place.

PR #4661, opened 2026-02-08
cpu: x64: matmul: No prefetches when B is transformed - bug fix

yair-obodovsky:
This change avoids issuing prefetches when B transformation is used, as they access out-of-bounds addresses, leading to performance degradation without causing a segfault.
Related to jira: https://jira.devtools.intel.com/browse/MFDNN-14571


yair-obodovsky:
make test
set test_scope=NIGHTLY

PR #4677, opened 2026-02-09
src: gpu: introduce Level Zero implementation

dzarukin:
Continuation of https://github.com/uxlfoundation/oneDNN/pull/4657

This portion introduces gpu/intel/ze machinery changes that is responsible for kernel/binary generation and updates SYCL path with re-usage of newly introduced functions.


dzarukin:
make test

dzarukin:
make test

h-sadia:
Isn't it better for everyone to make smaller commits rather than one big one?

dzarukin:
> Isn't it better for everyone to make smaller commits rather than one big one?

It is. You can take my word for granted it the smallest meaningful granularity. You may follow the chain of PRs preceding this one. Those are mentioned in the description.

src/gpu/intel/gemm/jit/include/gemmstone/dsl/runtime.hpp
```diff
@@ -48,6 +48,11 @@ ::sycl::kernel make_kernel(
 #ifdef GEMMSTONE_WITH_OPENCL_RUNTIME
 cl_kernel make_kernel(const kernel_t &kernel, cl_context ctx, cl_device_id dev);
 #endif
+#ifdef GEMMSTONE_WITH_LEVEL_ZERO_RUNTIME
+std::pair<ze_module_handle_t, ze_kernel_handle_t> make_kernel(
+        const kernel_t &kernel, ze_context_handle_t ctx,
+        ze_device_handle_t dev);
+#endif
```
rjoursler, line 42: The Gemmstone changes look like a reimplementation of functionality already merged in the upstream Gemmstone repository. The API is a little different though, so please align the API's to avoid issues when Gemmstone is downstreamed. The things that standout to me is the use of `GEMMSTONE_WITH_LEVEL_ZERO_RUNTIME` in place of `GEMMSTONE_WITH_L0_RUNTIME` and using `std::pair<ze_module_handle_t, ze_kernel_handle_t>` in place of `LevelZeroKernelAndModule`.

src/gpu/intel/jit/config/gemmstone_config.hpp
```diff
@@ -35,6 +35,7 @@
 #define ZEBIN_OUTPUT
 #define GEMMSTONE_WITH_SYCL_RUNTIME
 #define GEMMSTONE_WITH_OPENCL_RUNTIME
+#define GEMMSTONE_WITH_LEVEL_ZERO_RUNTIME
```
rjoursler: Could you explain why this is necessary? If oneDNN is being built for SYCL support, we shouldn't need to build gemmstone for level zero as we can rely on the SYCL API instead.

dzarukin: It seems to me it's practically the same reason as for `GEMMSTONE_WITH_OPENCL_RUNTIME` under sycl backend - this allows us to include the whole `gpu/intel/ocl` backend folder when compiling sycl backend without looking back what's needed and what's not since we are trying to re-use some methods from there. Without this `define` the engine.cpp file wouldn't compile (which, obviously, shouldn't bother sycl backend, but then we get into code separation or code re-usage trouble).

It seems to me it can be done in a different manner with a single utils_backends.{h,c}pp files at `gpu/intel` level and carefully marking which functions should be declared/defined for which backend. This would avoid including whole backend directories for sycl backend and both macros could be removed in this case.

Let me put a TODO to investigate this path and update this piece later once the Level Zero interop support lands and there'll be more time for cleaning stuff up, and keep current implementation for now. Will it be fine with you?

src/gpu/intel/gemm/jit/include/gemmstone/dsl/runtime.hpp
dzarukin: Thanks for this catch, I was not aware of the process of working with two external repositories. I think I've aligned both ngen and gemmstone parts with upstream version. There are couple fixes that are required, I'll submit them shortly. Would appreciate if you could take a second look.

dzarukin:
make test

src/gpu/intel/jit/config/gemmstone_config.hpp
rjoursler: Sounds good. My guess without looking in-depth is that the implementations of [this interface](https://github.com/uxlfoundation/oneDNN/blob/297b7f993ba101592fc1a48adcffae0b5cf0bda5/src/gpu/intel/engine.hpp#L72-L75) just need to be modified so that they are only (functionally) generated for the enabled DNNL GPU runtime.

src/gpu/intel/ze/engine.cpp
```diff
@@ -0,0 +1,235 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "gpu/intel/ze/engine.hpp"
+#include "gpu/intel/compute/ukernels.hpp"
+#include "gpu/intel/gemm/jit/include/gemmstone/microkernel/fuser.hpp"
```
rjoursler:
```Suggestion
#include "gemmstone/microkernel/fuser.hpp"
```

This is necessary to enable (correct) cross-compilation against upstream by compiling with [ONEDNN_GEMMSTONE_DIR](https://github.com/uxlfoundation/oneDNN/blob/297b7f993ba101592fc1a48adcffae0b5cf0bda5/src/gpu/CMakeLists.txt#L36-L38). Really, most of the `gemm/jit` belong under `third_party` similar to `ngen`, we just haven't done the work yet

dzarukin: Thanks, I verified both sycl and ze builds minus failed compilation from recent host scalar changes.

dzarukin:
make test

third_party/ngen/ngen_level_zero.hpp
```diff
@@ -156,13 +158,30 @@ ze_module_handle_t LevelZeroCodeGenerator<hw>::getModule(ze_context_handle_t con
...
+
+    ze_kernel_desc_t kernelDesc = {
+        ZE_STRUCTURE_TYPE_KERNEL_DESC,
+        nullptr,
+        0,
+        kernelName
+    };
+
+    ze_kernel_handle_t kernelHandle;
+    detail::handleL0(dynamic::zeKernelCreate(moduleHandle, &kernelDesc, &kernelHandle));
```
mzhukova, line 177: it seems like handleL0 just checks an error and throws an exception, this should interfere with the cleanup logic below, right? (e.g., previous moduleHandle create succeeded, but this one didn't)

src/gpu/intel/ze/kernel.cpp
```diff
@@ -0,0 +1,212 @@
...
+        VERROR(primitive, gpu,
+                "parameter bytes requirements greater than device supports");
+        return status::invalid_arguments;
+    }
+
+    if (range.is_zero()) { return status::success; }
+
+    std::vector<uint32_t> global_size(3, 1);
+    switch (range.global_range().ndims()) {
+        case 3: global_size[2] = static_cast<uint32_t>(range.global_range()[2]);
```
mzhukova, line 133: (minor) would it be useful to have some sort of debug print/assert in case when range value > UINT32_T?

third_party/ngen/ngen_level_zero.hpp
dzarukin: This call, if successful, returns both objects to the user and transfers the ownership. The user is responsible for further handling of those objects.
Modules are independent between calls to `getModuleAndKernel` so as kernels.
In summary, the call cleans after itself, if something failed, or transfers the ownership, there should be no problem here.

src/gpu/intel/ze/kernel.cpp
dzarukin: The initialization of `global_range()` happens before parallel call, all the checks are done there. It's safe to assume it's a valid number and conversion at this point of the flow.

third_party/ngen/ngen_level_zero.hpp
mzhukova: sorry I didn't quite get it, what I was trying to ask -- 
`moduleHandle` creation succeeded
then `zeKernelCreate` fails and therefore `handleL0` checks for error and throws an exception
how is `moduleHandle` cleaned up if we shouldn't get to the `zeModuleDestroy` below?

dzarukin: Noted, but it seems to be an "issue" of ngen as it follows the same protocol for dummy module, too.
I can address it separately through a unique_ptr with a custom deleter later as this scenario is rather uncommon and leads to termination anyway.

mzhukova: Sure, later is fine! Could you create a jira, so that I could resolve a thread?

dzarukin: I've updated [MFDNN-14318](https://jira.devtools.intel.com/browse/MFDNN-14318).

mzhukova: Thanks!

PR #4678, opened 2026-02-10
gtests: graph: unit: fix allocator tests when CPU_RT=OMP and GPU_RT=SYCL

TaoLv:
The case will fail if the library is built with CPU_RT=OMP and GPU_RT=SYCL and tested with CPU engine kind. This PR fixes it by checking CPU runtime and calling different allocate and deallocate functions accordingly.


TaoLv:
make test
disable benchdnn_all

ElaineBao (approved):

TaoLv:
make test
disable benchdnn_all

PR #4679, opened 2026-02-10
cpu: aarch64: avoid overhead from spawning different num_threads from layer to layer

fadara01:
cpu: aarch64: avoid overhead from spawning different num_threads from layer to layer

# Description

At the moment, in jit int8 matmul, when we do: `parallel(std::min(pd()->mm_parallel_work, num_threads), [&](const int ithr, const int nthr)` libgomp will shrink the team size if mm_parallel_work < num_threads - i.e. extra threads will exit. However, these threads might need to be re-created for subsequent layers where mm_parallel_work >= num_threads. 

This addresses the current overhead from spawning different number of OMP threads from layer to layer and aligns the threading behavior of jit int8 with [jit bf16 matmul](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/aarch64/matmul/jit_bf16_matmul.cpp#L416) and [brgemm matmul](https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/aarch64/matmul/brgemm_matmul.cpp#L936).

Fixes https://github.com/vllm-project/vllm/issues/34229

Here are some performance numbers with a vLLM int8 dyn quant matmul benchmark which directly calls jit int8.
The reason I chose 1x1280 as a src tensor is because its max parallel work = 54 < max num threads = 64.
The dramatic 100x speedups for the thread binding case are explained by  https://github.com/vllm-project/vllm/issues/34229
The dramatic 10x speedups when no thread binding is applied are due to getting rid of overhead from spawning different num threads from layer to layer (the vLLM int8 op has 3 parallel regions: one for quantizing src tensor, one in the oneDNN jit int8 gemm and one for applying the offset contribution)

speedup_int8

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes

### DIFF
--- a/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
+++ b/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2025 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


fadara01:
cc: @renato-arantes @jondea @Ryo-not-rio

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -1121,12 +1121,11 @@ status_t jit_int8_matmul_t<isa>::execute(const exec_ctx_t &ctx) const {
         int k_blks = div_up(K, b.k_blk);
         int parallel_work = B * m_blks * k_blks;
         int blk_per_bt = m_blks * k_blks;
-        int nt = std::min(pd()->mm_parallel_work, num_threads);
```
jondea, line 1124: What about when `parallel_work=1`? Do we not want to recreate the behaviour of https://github.com/uxlfoundation/oneDNN/blob/3a4c793b289d9a0c19553c57933b3b13a8843497/src/cpu/aarch64/matmul/brgemm_matmul.cpp#L933

Also, I think a similarly explanatory comment would be useful, because this new behaviour is counter-intuitive and it's completely possible someone will come along and revert this

fadara01: good idea, jit bf16 does not have a special case for parallel_work=1, I'll add one there too.
I skipped this case because parallel_work does not tend to be 1 in practice, but never say never ig :)

fadara01: Done, could you please take another look?

fadara01:
>I'd just say it may be worth putting some perf numbers in the description.

Done

PR #4680, opened 2026-02-10
benchdnn: graph: fix non-dense-strided cases of displaced filling

dzarukin:
[MFDNN-14395](https://jira.devtools.intel.com/browse/MFDNN-14395)

Comments in commit messages.

Thanks to @umar456 for debugging the cases and providing an input for the fix.

### DIFF
--- a/tests/benchdnn/matmul/matmul.cpp
+++ b/tests/benchdnn/matmul/matmul.cpp
@@ -701,7 +701,7 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
 }
 
 void skip_invalid_prb(const prb_t *prb, res_t *res) {
-    if (!prb->attr.zero_points.is_def()
+    if (!prb->attr.zero_points.get(DNNL_ARG_WEIGHTS).is_def()
             && (prb->wei_dt() != dnnl_s8 && prb->wei_dt() != dnnl_u8
                     && prb->wei_dt() != dnnl_s4 && prb->wei_dt() != dnnl_u4)) {
         BENCHDNN_PRINT(2,


dzarukin:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -604,34 +604,82 @@ int partition_data_displacer_t::displace_input_data(size_t lt_id,
...
+            DNN_SAFE_V(dnnl_memory_desc_create_with_strides(&new_replace_md,
+                    mem.ndims(), mem.dims(), mem_replace.dt(), mem.strides()));
+            dnnl_memory_desc_destroy(mem_replace.md_);
+            dnnl_memory_desc_clone(&mem_replace.md_, new_replace_md);
+            SAFE(mem.reorder(mem_replace), WARN);
+            dnnl_memory_desc_destroy(new_replace_md);
+            break;
+        }
+
+        // Case when there're non-dense strides in `mem.md_` leading to "holes"
```
h-sadia, line 659: Does "holes" here mean sparsity in a dim?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -604,34 +604,82 @@ int partition_data_displacer_t::displace_input_data(size_t lt_id,
...
-    SAFE(mds_ok ? OK : FAIL, WARN);
+        // Case w/ grouped convolutions when number of dimensions would be +1.
+        bool is_grouped_conv = false;
+        if (main_op.kind_ == "Convolution"
+                || main_op.kind_ == "ConvTranspose") {
+            int64_t groups;
+            main_op.get_attr_s64(groups, "groups");
+            is_grouped_conv = groups > 1;
+        }
+        if (is_grouped_conv) {
```
h-sadia: Isn't it better to move this if inside the main_op.kind_ == "Convolution"? Since it is going to false otherwise?

tests/benchdnn/graph/input_displacer.cpp
```diff
@@ -604,34 +604,82 @@ int partition_data_displacer_t::displace_input_data(size_t lt_id,
...
-    }
-    dnnl_memory_desc_destroy(mem_replace.md_);
-    dnnl_memory_desc_clone(&mem_replace.md_, md);
-    SAFE(mem.reorder(mem_replace), WARN);
+        // Case when there're extra unit dims in replaced memory. Memory buffers
+        // are identical but different ndims are restricted in reorder API.
+        // `mem_replace.md_` requires manual adjustment before reordering.
+        const bool is_reshaped_dims = mem_replace.nelems() == mem.nelems()
+                && mem_replace.ndims() != mem.ndims();
+        if (is_reshaped_dims) {
```
h-sadia, line 648: in case of reshape dims, are we skipping any dim?

tests/benchdnn/graph/input_displacer.cpp
dzarukin: A "hole" means a physical memory space that is not supposed to be used by the operation and mustn't be read from/written to. I'm not quite sure what the "sparsity in a dim" term mean.

tests/benchdnn/dnnl_memory.cpp
```diff
@@ -239,6 +239,35 @@ size_t dnn_mem_t::size(int index) const {
...
+    const auto &_dims = with_padded_dims ? padded_dims() : dims();
+    if (ndims() == 0) return 0;
+
+    int64_t n = 1;
+    for (int i = 0; i < ndims(); ++i)
+        n *= _dims[i];
+    return n;
+}
+
+bool dnn_mem_t::is_dense() const {
```
h-sadia, line 252: could you please add a comment about when it returns true? by no holes do you mean the least we have is a unit stride? or is unit stride is am important part of a being a dense matrix?

tests/benchdnn/graph/input_displacer.cpp
dzarukin: Sure thing, thanks for a suggestion.

tests/benchdnn/graph/input_displacer.cpp
dzarukin: If you mean whether a reshaped dim can by any of original, then yes.

tests/benchdnn/dnnl_memory.cpp
dzarukin: > is unit stride an important part of a being a dense matrix?

It is! Otherwise, there'll be "holes" in the innermost dim which would stop being dense in that case.

The definition of "dense" is practically the same as inside the library, it just uses slightly different approach to verify it.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -701,7 +701,7 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
 }
 
 void skip_invalid_prb(const prb_t *prb, res_t *res) {
-    if (!prb->attr.zero_points.is_def()
+    if (!prb->attr.zero_points.get(DNNL_ARG_WEIGHTS).is_def()
```
avmanerikar, line 704: (nit): a comment for the updated check would be helpful

tests/benchdnn/dnnl_memory.cpp
avmanerikar: (nit): Would it be simpler to rename the method to `has_non_dense_strides()` or equivalent? That directly specifies what the function does.

avmanerikar (approved):

tests/benchdnn/matmul/matmul.cpp
dzarukin: I'm not sure what to put there. The following code checks weights data types, thus, **any** zero-point wouldn't work, it must be weights one. It's just a bug.

tests/benchdnn/dnnl_memory.cpp
dzarukin: This is the name used inside the library with the same meaning and purpose, it's easier to have them aligned.

tests/benchdnn/graph/input_displacer.cpp
h-sadia: I see, I was confusing hole with sparse matrix

tests/benchdnn/graph/input_displacer.cpp
h-sadia: I guess what I am trying to ask is; if the ndims in one of the memory desc is more than what is expected then what are we doing with the extra dim?

tests/benchdnn/matmul/matmul.cpp
h-sadia: Maybe just that could be put in the comment.. will help later on if someone happens on it during debugging.

tests/benchdnn/dnnl_memory.cpp
h-sadia: maybe we need to align them both to something more meaningful; I am confused by what this is doing. Is it checking if a stride is dense? or is it checking if a memory desc is dense?

dzarukin: It checks if the layout is dense based on strides and dims. memory_desc doesn't provide public queries for dense property.

tests/benchdnn/graph/input_displacer.cpp
dzarukin: Updating the memory descriptor and doing the reorder. Those actions are inside the `if` section.

PR #4681, opened 2026-02-11
cpu: x64: gemm: remove dead code

Stars-22:
cpu: x64: gemm: remove dead code

### DIFF
--- a/src/cpu/x64/gemm/f32/jit_avx512_common_gemm_f32.cpp
+++ b/src/cpu/x64/gemm/f32/jit_avx512_common_gemm_f32.cpp
@@ -1623,8 +1623,7 @@ dnnl_status_t sgemm_nocopy_driver(const char *transa, const char *transb,
     bool hasBias = (bias != nullptr);
     auto ker_bn = get_xbyak_gemm(isTransA, isTransB, *beta, hasBias);
     auto ker_b1 = get_xbyak_gemm(isTransA, isTransB, 1.0, false);
-    auto ker_b0 = get_xbyak_gemm(isTransA, isTransB, 0.0, false);
-    if (utils::any_null(ker_bn, ker_b1, ker_b0)) return dnnl_runtime_error;
+    if (utils::any_null(ker_bn, ker_b1)) return dnnl_runtime_error;
 
     dim_t BM = 4032, BN, BK;
     if (mayiuse(avx512_core) && __BUILD_GEMM_AVX512) {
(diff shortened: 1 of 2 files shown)


vpirogov:
@Stars-22, could you please elaborate on why you believe removed code is "dead code"?

Stars-22:
The modification is called **"Dead Code Elimination"** because the removed code was logically redundant and could never produce a result different from the remaining code.
Here is the detailed analysis:
### 1. Variable Definitions Analysis
In the original code, three kernel pointers were defined:
*   **`ker_bn`**: Initialized with the actual runtime parameters.
    ```cpp
    auto ker_bn = get_xbyak_gemm(isTransA, isTransB, *beta, hasBias);
    ```
*   **`ker_b0`**: Initialized specifically for the case where beta is 0 and there is no bias.
    ```cpp
    auto ker_b0 = get_xbyak_gemm(isTransA, isTransB, 0.0, false);
    ```
### 2. Usage Context Analysis
In the original code, `ker_b0` was only invoked under a very specific `if` condition:
```cpp
if (Bk == 0) {
    if (*beta == 0.0 && bias == nullptr)
        (*ker_b0)(...); // Uses ker_b0
    else
        (*ker_bn)(...); // Uses ker_bn
}
```
### 3. Proof of Redundancy
Let's examine the state of the variables when the condition `*beta == 0.0 && bias == nullptr` is true:
1.  **`*beta` is `0.0`**: Consequently, the `*beta` argument passed to `ker_bn` becomes `0.0`.
2.  **`bias` is `nullptr`**: Consequently, the `hasBias` variable (calculated as `bias != nullptr`) becomes `false`.
Therefore, inside this specific `if` block, the initialization of `ker_bn` becomes equivalent to:
```cpp
// Inside the condition (*beta == 0.0 && bias == nullptr)
ker_bn = get_xbyak_gemm(isTransA, isTransB, 0.0, false);
```
This is **exactly identical** to the definition of `ker_b0`.
### Conclusion
Since `ker_b0` and `ker_bn` point to the same kernel object (or functionally equivalent kernels) when the condition is met, the branch using `ker_b0` is redundant. The code can be simplified by removing `ker_b0` and the specific `if` check, relying solely on `ker_bn` to handle the logic correctly.
Removing code that does not change the program's behavior is standard **Dead Code Elimination**.

vpirogov:
While `ker_bn` and `ker_b0` are functionally equivalent the latter produces an implementation specialized for the case `beta = 0.0` for performance reasons. So this is definitely not "Dead Code". I'm closing this PR.

In the future please refrain from using AI slop to respond to questions about proposed code changes.

Stars-22:
Screenshot
You are correct that the specialized kernel is necessary for performance. However, looking at the `get_xbyak_gemm` implementation, it uses `beta_idx` to map `beta == 0.0` to index `0`.
When `*beta == 0.0` is true, `ker_bn` retrieves the kernel from `kernel_table[...][0]`, which **is** the specialized beta=0 kernel. `ker_b0` retrieves the exact same entry.
Since both pointers resolve to the same specialized implementation, the extra variable and branch are redundant.

PR #4682, opened 2026-02-11
cpu: aarch64: re-enable wtag ba for brgemm matmul sve_256 and gemv

michalowski-arm:
#4315 disabled wtag transposed entirely for aarch64 f32 brgemm matmul, but it should be supported on sve_256 (since it has a copy_b_transposed implementation) and for gemv cases (since they don't use copy_b_transposed). This change re-enables it in those cases.

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -21,6 +21,7 @@
 #include "common/dnnl_thread.hpp"
 #include "common/type_helpers.hpp"
 #include "common/utils.hpp"
+#include "cpu/aarch64/cpu_isa_traits.hpp"
 #include "cpu/aarch64/injectors/jit_uni_postops_injector.hpp"
 
 #include "cpu/binary_injector_utils.hpp"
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2024 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
```
jondea: I don't think we need this

src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -232,13 +233,35 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(
...
                 }
                 if (blocked_B_layouts_allowed) {
-                    bgmmc.wei_tag = memory_desc_matches_one_of_tag(B_md,
-                            plain_tensor_layout_tag, blocked_64n_B_layout_tag,
-                            blocked_48n_B_layout_tag, blocked_32n_B_layout_tag,
-                            blocked_16n_B_layout_tag);
+                    // Allows tranposed only for sve_256, since it has a copy_b_transpose
+                    // implemented, and for gemv cases since they do not use copy_b.
+                    // TODO: a working copy_b_transpose for sve_128
+                    if (bgmmc.isa == sve_256 || bgmmc.M == 1 || bgmmc.N == 1) {
```
jondea: You might want to check the later logic which sets the `is_gemv` flag, it's a bit more restrictive.

openmodelzooci:

PR #4683, opened 2026-02-12
xe: sdpa: restrict block2d_stores to rows(unroll_m_vs) < 64

umar456:
# Description

Block 2D loads/stores can only read one cache line worth of rows, the current restriction on 2d blocks assumed we would only transfer halfs but this caused issues when the return type if float. The new restrictions prevent the use of 2d block stores in these cases.

Fixes # [MFDNN-14397](https://jira.devtools.intel.com/browse/MFDNN-14397)

### DIFF
--- a/src/gpu/intel/sdpa/micro.cpp
+++ b/src/gpu/intel/sdpa/micro.cpp
@@ -502,7 +502,7 @@ status_t micro_t::pd_t::init_conf(impl::engine_t *engine) {
         conf.block_q = (ldq % 4 == 0);
         conf.block_a = (lda % 16 == 0 && v_full);
     } else if (pd->arch() >= compute::gpu_arch_t::xe_hpc
-            && config.unroll_m_vs < 64) {
+            && (config.unroll_m_vs * dst_mdw.data_type_size()) <= 64) {
         auto vbytes = d->values() * val_mdw.data_type_size();
         if (lda % 16 == 0 && vbytes % 4 == 0) conf.block_2d_a = true;
     }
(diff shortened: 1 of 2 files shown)


umar456:
make test
disable benchdnn_all
set test_scope=NIGHTLY
enable benchdnn_graph
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

umar456:
make test perf-gpu
set primitive=sdpa

src/gpu/intel/sdpa/micro.cpp
```diff
@@ -502,7 +502,7 @@ status_t micro_t::pd_t::init_conf(impl::engine_t *engine) {
         conf.block_q = (ldq % 4 == 0);
         conf.block_a = (lda % 16 == 0 && v_full);
     } else if (pd->arch() >= compute::gpu_arch_t::xe_hpc
-            && config.unroll_m_vs < 64) {
+            && (config.unroll_m_vs * dst_mdw.data_type_size()) <= 64) {
```
Simonsays095, line 505: If this size exceeds a cache line, can we resolve to multiple block 2D loads? e.g. 128 bytes -> 2 block2D loads?

umar456: Yeah, we can. It would require a larger refactor of the tile_ops header. Not sure if this would yield performance improvement over the current approach. The regular block stores are pretty fast and block 2d stores are typically only used when full blocks cannot be written to memory. Promoting block2d writes over regular block writes hurt performance(i.e. including it in the upper if block.)

Simonsays095: It looks like there's no impact on performance, according to CI. I guess the microkernel can still use the more performant loads, and this just doesn't affect it that much - so we just need it to be correct.

Simonsays095 (approved):

PR #4686, opened 2026-02-12
xe: gemm: jit: fixup gemmstone include path

rjoursler:
Prevents issues when compiling against upstream


rjoursler:
make test
disable test_device_cpu

PR #4687, opened 2026-02-13
[GPU] Enable interleaved k-parallel microkernels

Simonsays095:
Addresses [MFDNN-14586](https://jira.devtools.intel.com/browse/MFDNN-14586). Also adds some code for enhanced debugging of microkernels (ability to override ukernel selection for SDPA) and fixes an issue in a recent change to the microkernel interface.

These changed kernels are not in use currently, but will be once the ukernel version of grouped gemm is merged in PR #4505.


umar456 (approved):

src/gpu/intel/sdpa/micro.cpp
```diff
@@ -637,28 +639,65 @@ status_t micro_params_t::get_kernel_ctx(
...
                 ex.what());
     }
 
     /* Ask microkernel provider for microkernel */
+    auto vs_strat_override = [&](gemmstone::GEMMStrategy &strat) {
+        std::string newStrat;
+        newStrat = gpu_utils::dev_getenv("SDPA_VS_USTRATEGY", newStrat);
+        if (!newStrat.empty()) {
+            // Example: 16 16 aT32 aM32 aB wg 2x4 sys
+            auto product = ngen::npack::decodeHWIPVersion(hw_info.gmdid);
```
kealan-barbieri, line 676: nit: can we de-duplicate this chunk for parsing strategy from string?

h-sadia:
what are some kernels that will take benefit from k-parallelism? Isn't enabling this a double-edged sword.

src/gpu/intel/sdpa/micro.cpp
Simonsays095: I consider this as a working template for now - one thing to consider is that we override both the strategy and the problem in regular gemm kernel overriding. Eventually I think we should make a gemmstone-side function for setting the problem types and strategy from a string, but I'll leave that for later.

Simonsays095:
> what are some kernels that will take benefit from k-parallelism? Isn't enabling this a double-edged sword.

True, not all kernels will benefit from k-parallelism. In general, it should be used for smaller problems and kernels that need extra parallelism. This change just enables us to use (interleaved) k-parallelism, but we still rely on the kernel catalog and the performance model to select the best kernel for the given problem.

Simonsays095:
Manually tested performance using grouped gemm, no significant performance differences to report.

PR #4689, opened 2026-02-14
api: add Level Zero runtime and interoperability interfaces

dzarukin:
Continuation of #4677

This is the last piece of functionality needed for OpenVINO on ww8.
There's also validation (and infra enablement) and documentation coming after + potential bug fixes and couple leftovers.

Shout-out and part credit to @spalicki for all the code put together to make this runtime working and promoted till this point.


dzarukin:
make test

src/gpu/intel/gemm/jit/include/gemmstone/generator.hpp
```diff
@@ -53,6 +53,9 @@ GEMMSTONE_NAMESPACE_START
 #if DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
 #define FORWARD(hw) NGEN_FORWARD_OPENCL(hw)
 #endif
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_ZE
+#define FORWARD(hw) NGEN_FORWARD_LEVEL_ZERO(hw)
+#endif
```
dzarukin, line 58: I figured this part is different from what's in gemmstone repo, but I decided to follow the existing pattern and keep the improvement for maintainers.

src/gpu/intel/ze/kernel.cpp
echeresh, line 180: Is reference intended here?
```suggestion
    std::vector<ze_event_handle_t> &ze_out_deps
            = utils::downcast<xpu::ze::event_t *>(&out_dep)->ze_events_;
```

src/gpu/intel/jit/binary_format.cpp
```diff
@@ -45,6 +45,9 @@
 #if DNNL_GPU_RUNTIME == DNNL_RUNTIME_OCL
 #define FORWARD(hw) NGEN_FORWARD_OPENCL(hw)
 #endif
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_ZE
+#define FORWARD(hw) NGEN_FORWARD_LEVEL_ZERO(hw);
```
echeresh: For consistency:
```suggestion
#define FORWARD(hw) NGEN_FORWARD_LEVEL_ZERO(hw)
```

src/gpu/intel/ze/stream.cpp
echeresh, line 50: Left-shift of a negative value is UB in C++, I think it's equivalent to this:

```suggestion
                (1ULL << device_properties.kernelTimestampValidBits) - 1
```

src/gpu/intel/ze/device_info.cpp
echeresh, line 136:
```suggestion
    CHECK(xpu::ze::zeDeviceGetMemoryAccessProperties(
            device, &device_memory_access_properties));
```

src/gpu/intel/ze/utils.cpp
echeresh, line 259: Looks like this is a potential module leak - if kernel creation fails. I see you mentioned a similar place in nGEN here: [MFDNN-14318](https://jira.devtools.intel.com/browse/MFDNN-14318). Leaving a comment to combine/address this place in the future as well.

src/gpu/intel/ze/device_info.cpp
echeresh, line 129: I see in `dsl/runtime.cpp` this is queried by also checking `flags`. I think we need the same check here.

```c++
        for (uint32_t i = 0; i < numProp; i++) {
            if (deviceCacheProps[i].flags == 0) {
                l3_cache_size = deviceCacheProps[i].cacheSize;
                break;
            }
        }
```

include/oneapi/dnnl/dnnl_ze.hpp
```diff
@@ -0,0 +1,231 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
```
mzhukova: (minor) -> 2026 maybe

include/oneapi/dnnl/dnnl_ze.h
```diff
@@ -0,0 +1,157 @@
...
+/// command list.
+///
+/// @param stream Output execution stream.
+/// @param engine Engine to create the execution stream on.
+/// @param list Level Zero command list to use.
+/// @param profiling Flag enabling GPU kernels profiling.
+/// @returns #dnnl_success on success and a status describing the error
+///     otherwise.
+dnnl_status_t DNNL_API dnnl_ze_interop_stream_create(dnnl_stream_t *stream,
+        dnnl_engine_t engine, ze_command_list_handle_t list, bool profiling);
```
mzhukova: I thought we're mainly using `int` for boolean type of things in C headers, might we need extra header included here for this case?

include/oneapi/dnnl/dnnl_ze.h
```diff
@@ -0,0 +1,157 @@
...
+///     otherwise.
+dnnl_status_t DNNL_API dnnl_ze_interop_stream_get_list(
+        dnnl_stream_t stream, ze_command_list_handle_t *list);
+
+/// Creates a memory object.
+///
+/// Unless @p handle is equal to DNNL_MEMORY_NONE or DNNL_MEMORY_ALLOCATE, the
+///     constructed memory object will have the underlying buffer set.
+///     In this case, the buffer will be initialized as if
+///     dnnl_ze_interop_memory_set_mem_object() had been called.
```
mzhukova: `dnnl_ze_interop_memory_set_mem_object()` does this thing exist? I've only found `_ocl_` one (same for C++ header)

include/oneapi/dnnl/dnnl_ze.hpp
```diff
@@ -0,0 +1,231 @@
...
+            dnnl_ze_interop_engine_create(&aengine, adriver, adevice, acontext),
+            "could not create an engine");
+    return engine(aengine);
+}
+
+/// Returns the Level Zero context associated with an engine.
+///
+/// @param aengine Engine to query.
+///
+/// @returns The underlying Level Zero device of the engine.
```
mzhukova:
```suggestion
/// @returns The underlying Level Zero context of the engine.
```

include/oneapi/dnnl/dnnl_ze.hpp
```diff
@@ -0,0 +1,231 @@
...
+    error::wrap_c_api(dnnl_ze_interop_engine_get_context(aengine.get(), &ctx),
+            "could not get a context handle");
+    return ctx;
+}
+
+/// Returns the Level Zero device associated with an engine.
+///
+/// @param aengine Engine to query.
+///
+/// @returns The underlying Level Zero context of the engine.
```
mzhukova:
```suggestion
/// @returns The underlying Level Zero device of the engine.
```

src/xpu/ze/capi/primitive.cpp
```diff
@@ -0,0 +1,72 @@
...
+        ze_stream_impl->ze_ctx().set_deps(events);
+    }
+
+    // run primitive
+    exec_args_t exec_args;
+    CHECK(cvt_primitive_args(
+            primitive_iface->pd()->impl().get(), nargs, args, exec_args));
+
+    exec_ctx_t ctx(stream, std::move(exec_args));
+    CHECK(primitive_execute(primitive_iface, ctx));
```
mzhukova, line 58: if the primitive_execute fails and therefore CHECK exits with non-success status is it safe to not do `stream->after_exec_hook()` (seems like this does some cleaning of the dependencies..?) ?

src/gpu/intel/ze/kernel.cpp
dzarukin: I guess it is, I'll update the whole context piece (interfaces and force to use them) with testing PR. Thanks for the catch.

src/gpu/intel/jit/binary_format.cpp
dzarukin: Removed, thanks.

src/gpu/intel/ze/stream.cpp
dzarukin: Makes sense, will post a fix with testing PR, thanks for the catch.

src/gpu/intel/ze/device_info.cpp
dzarukin: Good catch, will update, thanks.

src/gpu/intel/ze/utils.cpp
dzarukin: This one turned out to be easier than anticipated, I'll be addressed with a testing PR. Thanks for spotting!

src/gpu/intel/ze/device_info.cpp
dzarukin: Indeed, will update, too. Thanks!

include/oneapi/dnnl/dnnl_ze.hpp
dzarukin: Updated, thanks.

include/oneapi/dnnl/dnnl_ze.h
dzarukin: You are right, need to use `int`. Updated, thanks for the catch.

include/oneapi/dnnl/dnnl_ze.h
dzarukin: It did until it didn't, updated to the proper `dnnl_memory_set_data_handle()` call. Thanks for the catch.

include/oneapi/dnnl/dnnl_ze.hpp
dzarukin: Thanks, updated.

include/oneapi/dnnl/dnnl_ze.hpp
dzarukin: Thanks, updated.

src/xpu/ze/capi/primitive.cpp
dzarukin: It's a good question. It should be safe as stream doesn't manage stuff in after_hook, it just clears events in its internal state which shouldn't matter on a failed stream. At least, none of backends assume it would happen that not cleared dependencies would affect any potential stuff with that failed stream.

dzarukin:
make test

src/xpu/ze/capi/primitive.cpp
mzhukova: if someone re-tries after unsuccessful submission, could there be any issues with using the stream with "previous" deps? or this wouldn't be the case?

dzarukin: I have a feeling it's possible to find a combination where it might hand or start depending on a dangling event or something, but that would be unconventional flow from what I can see from the code, like using interop execute and then use a regular one.

I'll let @echeresh to find a weak spot of this potential lack of cleaning up dependencies.

dzarukin: Discussed offline, agreed to drop CHECK, perform after_hook unconditionally and only then report status. Thanks for a question. I'll post an update for all backends in the next PR.

mzhukova: thanks for the update!

src/gpu/intel/gemm/jit/include/gemmstone/generator.hpp
Fozzibearr: Nice

PR #4692, opened 2026-02-16
Fix overflows in div_up for sub-maximal arguments

asimonov1:
This PR is extracted from another one related to overflows because the function div_up is used everywhere and needs high attention.

This function overflows if e.g. the sum of its arguments is larger than the maximal value of the type of that sum, but the function should not overflow because e.g. we always can divide an int by another int without overflow (and rounding up does not matter). At least, it should not overflow for any positive arguments.

Another question is how this function can be used.
We used it with the 2nd parameter of the type float (find "thr_eff_threshold = 0.9f" in the code; an example is [jit_brgemm_conv_utils.cpp)](https://github.com/uxlfoundation/oneDNN/blob/4be91e40c51e053c3de6cd40cf31ccbee8d4e0f0/src/cpu/x64/jit_brgemm_conv_utils.cpp#L1251-L1252). The div_up function has been replaced with ceil function in these cases. The parameter types of the div_up function are limited to integral types.
We also used this function with negative values of the argument a. The PR changes the behavior of div_up in this case: it returns 0 for negative a, and an assert(a>=0) has been added. The code has been updated to call div_up with non-negative values of a.
We used runtime value placeholder in calculations in several places accidentally. This was found due to the new assert. These cases have been fixed.

Similar div_up functions in GPU/SYCL code have been updated also.

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul.cpp
@@ -836,6 +836,8 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
                     = weights_d.size() - weights_d.additional_buffer_size();
             const size_t b_batch
                     = get_bb_idx(bgmmc.batch - 1, bgmmc_.bcast_B_desc) + 1;
+            assert(IMPLICATION(bgmmc.s8s8_compensation_required,
+                    !is_runtime_value(bgmmc.s8s8_comp_b_str)));
             const size_t s8s8_buffer_sz = bgmmc.s8s8_compensation_required
                     ? sizeof(int32_t) * b_batch * bgmmc.s8s8_comp_b_str
                     : 0;
@@ -1224,6 +1226,7 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
         const int n_blk_local = bgmmc_.use_buffer_b
                 ? n_blk_idx % bgmmc_.N_chunk_size
                 : n_blk_idx;
+        assert(!is_runtime_value(bgmmc_.s8s8_comp_b_str));
         return s8s8_compensation_ptr_ + ithr * bgmmc_.s8s8_comp_ithr_str
                 + get_bb_idx(b, bgmmc_.bcast_B_desc) * bgmmc_.s8s8_comp_b_str
                 + n_blk_local * bgmmc_.s8s8_comp_n_str;
--- a/src/cpu/aarch64/matmul/brgemm_matmul_reorders.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_reorders.cpp
@@ -189,6 +189,7 @@ status_t brgemm_matmul_copy_reorder_t::execute_body(
         ker_exec_ctx.current_N_blk
                 = is_N_tail ? kernel_conf.N_tail : kernel_conf.N_blk;
 
+        assert(!is_runtime_value(kernel_conf.s8s8_comp_b_str));
         const auto comp_offset = batch * kernel_conf.s8s8_comp_b_str
                 + n_blk_idx * kernel_conf.s8s8_comp_n_str;
 
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -1167,11 +1167,18 @@ void init_aux_values(brgemm_matmul_conf_t &bgmmc,
     bgmmc.M_chunk_elems = bgmmc.M_blk * bgmmc.M_chunk_size;
     bgmmc.N_chunk_elems = bgmmc.N_blk * bgmmc.N_chunk_size;
     bgmmc.K_chunk_elems = bgmmc.K_blk * bgmmc.brgemm_batch_size;
-    bgmmc.M_chunks = div_up(bgmmc.M, bgmmc.M_chunk_elems);
-    bgmmc.N_chunks = div_up(bgmmc.N, bgmmc.N_chunk_elems);
-    bgmmc.K_chunks = div_up(bgmmc.K, bgmmc.K_chunk_elems);
-    bgmmc.num_M_blocks = div_up(bgmmc.M, bgmmc.M_blk);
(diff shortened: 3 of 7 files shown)


asimonov1:
make test
set test_scope=NIGHTLY

dzarukin:
@asimonov1, thanks for drawing an attention to this function...

Here are my thoughts:
* I agree with overflow statement and proposed general definition looks good to me.
* I think it's safe to demand `a >= 0` under assert and also have a separate `if (a <= 0) return 0;` to remove conditional return for the general statement (keep only a new one).
* `float` doesn't seem to behave as expected for any body definition (e.g. `int a = 2, float b = 0.3f` suppose to give identical to `int a = 20, float b = 3.f` the answer `7` but it won't (it gives `4`) as the function is defined by `a` type which is integer. So, it'd be for better if we limit `T` and `U` to integral types only. I have no idea if it's going to be a disaster to restrict `T` and `U` being of the same type. If it's not, I'd probably align those, too, to have identical definitions to GPU functions.

> Also, we have several more implementations with the same overflow (e.g. find "(a + b - 1) / b" in the code).

Would be good to replace that.

> The assert condition is changed to b>0 to test if we use a negative b. It looks that we do not. Should it be restored?

I'd keep both arguments positive. I can't imagine why would we want to `div_up` negative numbers at all.

src/common/utils.hpp
```diff
@@ -356,8 +356,9 @@ constexpr const T &saturate(const T &low, const T &upper, const T &a) {
 
 template <typename T, typename U>
 inline typename remove_reference<T>::type div_up(const T a, const U b) {
-    assert(b);
-    return static_cast<typename remove_reference<T>::type>((a + b - 1) / b);
+    assert(b > 0);
+    return static_cast<typename remove_reference<T>::type>(
+            a >= 1 ? 1 + (a - 1) / b : (a + b - 1) / b);
```
rjoursler: The typing here can still lead to weird edge case. For example something like the following would not work as intended:
```
div_up(int(2), dim_t(INT_MAX) + 2)
```

rjoursler (commented):
This generally looks good. Please update the OpenCL headers you mentioned to use the:
```
1 + (a - 1) / b
```
definition as well, that looks like it will be more computationally efficient (especially as most Intel GPU's emulate integer division/modulus).

asimonov1:
> * I think it's safe to demand `a >= 0` under assert and also have a separate `if (a <= 0) return 0;` to remove conditional return for the general statement (keep only a new one).

@dzarukin, there are many places where div_up is called with negative `a` (I added logging of such cases locally). The code in such places looks like
```
        int kd_start = nstl::max(
                0, div_up(jcp.f_pad - jcp.stride_d * i_od, kd_step));
```
I did not verify that exactly this sample leads to negative `a`, but using `max` is a sign that negative values could be returned. I used Copilot to quickly find similar cases, and Copilot found suspicious cases in more than 30 files (verification is required) mostly in (de)convolution, but some cases are not fixed yet.

src/common/utils.hpp
asimonov1: What is expected here? The result is `(int)1`. The type of `a` is `int`, the type of `b` is `dim_t`, `b=(dim_t)2147483649`, the type of `(a-1)` is `int`, `a-1=(int)1`, the type of `(a-1)/b` is `dim_t` (`(a-1)` is converted to `dim_t`), the type of `1+(a-1)/b` is `dim_t` ("`int`+`dim_t`", 1 is converted to `(dim_t)1`), the result before the final cast is `(dim_t)1`. The type of `(a + b - 1) / b` is also `dim_t`.

rjoursler: I see, I wasn't thinking through the order of operations quite right.

In general, it may still be desirable to modify this function to return `decltype(a/b)` so that its type behavior matches C++ promotion semantics.

asimonov1:
make test
set test_scope=NIGHTLY

asimonov1:
make test
set test_scope=NIGHTLY

asimonov1:
make test
set test_scope=NIGHTLY

asimonov1:
3 commits are added. The 1st one adds an assert to div_up. The 2nd updates many places related to strides/padding. The 3rd fixes handling runtime dimensions in Matmul.

asimonov1:
make test
set test_scope=NIGHTLY
enable arch_cpu_adl
enable arch_cpu_clx
enable arch_cpu_dmr
enable arch_cpu_gnr
enable arch_cpu_hsw
enable arch_cpu_nhm
enable arch_cpu_nvl
enable arch_cpu_skx
enable arch_cpu_snb
enable arch_cpu_spr
enable arch_cpu_srf
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-fs1

zhangjian29 (approved):
Look good to RV64 part. Thanks.

tprimak (approved):

asimonov1:
> * I have no idea if it's going to be a disaster to restrict `T` and `U` being of the same type. If it's not,

This would require changing about 50 files just for the x64 code, so it's better to do it in a separate PR.

dzarukin (approved):

asimonov1:
make test
set test_scope=NIGHTLY
enable arch_cpu_adl
enable arch_cpu_clx
enable arch_cpu_dmr
enable arch_cpu_gnr
enable arch_cpu_hsw
enable arch_cpu_nhm
enable arch_cpu_nvl
enable arch_cpu_skx
enable arch_cpu_snb
enable arch_cpu_spr
enable arch_cpu_srf
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-fs1

PR #4693, opened 2026-02-17
cpu: x64: pool: add post-ops support for int8 max pooling

jstachowintel:
[MFDNN-8916](https://jira.devtools.intel.com/browse/MFDNN-8916) int8 maxpool with binary mul post-ops run into ref path

Added eltwise and binary post-ops support for the int8 max pooling JIT kernel. Previously, max pooling with post-ops was unconditionally rejected and fell back to the ref implementation.

The main challenge is that max pooling operates on packed i8 data, while post-ops injectors expect f32. This is solved with a 3-phase approach:
1. **Unpack:** Extract packed i8 results from vreg_dst into s32 chunks vreg_dst_s32, iterating in reverse order to avoid clobbering aliased registers.
2. **Apply post-ops:** Convert s32→f32, apply eltwise/binary post-ops via the injector, convert back f32→s32 (with u8 clamping if needed).
3. **Pack & store:** Reuse store_dst_avg_op to pack s32 back to i8 and write to memory.

ISA-specific max_postops_unpacker_t helpers handle the unpack step for AVX-512, AVX2, and SSE4.1.

### DIFF
--- a/src/cpu/x64/jit_uni_i8i8_pooling.cpp
+++ b/src/cpu/x64/jit_uni_i8i8_pooling.cpp
@@ -167,20 +167,8 @@ struct jit_uni_i8i8_pooling_fwd_ker_t : public jit_generator_t {
     };
 
     inline size_t get_offset_dst(int jj, int ll) const {
-        size_t offset = 0;
-        switch (jpp.alg) {
-            case pooling_max: {
-                offset = jj * jpp.c_block * sizeof_dst_dt();
-                break;
-            }
-            case pooling_avg_include_padding:
-            case pooling_avg_exclude_padding: {
-                offset = (ll * (jpp.c_block / max_num_ll) + jj * jpp.c_block)
-                        * sizeof_dst_dt();
-                break;
-            }
-            default: assert(!"unsupported pooling algorithm");
-        }
+        size_t offset = (ll * (jpp.c_block / max_num_ll) + jj * jpp.c_block)
+                * sizeof_dst_dt();
         return offset;
     }
 
--- a/src/cpu/x64/jit_uni_i8i8_pooling.cpp
+++ b/src/cpu/x64/jit_uni_i8i8_pooling.cpp
@@ -878,6 +878,59 @@ void jit_uni_i8i8_pooling_fwd_ker_t<avx512_core>::compute_max_op(const int jj) {
 }
 
 template <cpu_isa_t isa>
+struct max_postops_unpacker_t;
+
+template <>
+struct max_postops_unpacker_t<avx512_core> {
+    static void load_dst_chunk(
+            jit_uni_i8i8_pooling_fwd_ker_t<avx512_core> *self, int jj, int ll,
+            const Xbyak::Zmm &dst_s32) {
+        // Use vreg_tail as temp register since tail is not supported for max+postops
+        Xbyak::Xmm xreg_tmp(self->vreg_tail.getIdx());
+        Xbyak::Zmm zreg(self->vreg_dst(jj).getIdx());
+        self->vextracti32x4(xreg_tmp, zreg, ll);
+
+        if (self->jpp.src_dt == data_type::s8)
+            self->vpmovsxbd(dst_s32, xreg_tmp);
+        else
+            self->vpmovzxbd(dst_s32, xreg_tmp);
+    }
+};
+
+template <>
+struct max_postops_unpacker_t<avx2> {
+    static void load_dst_chunk(jit_uni_i8i8_pooling_fwd_ker_t<avx2> *self,
+            int jj, int ll, const Xbyak::Ymm &dst_s32) {
+        // Use vreg_tail as temp register since tail is not supported for max+postops
+        Xbyak::Xmm xreg_tmp(self->vreg_tail.getIdx());
+        Xbyak::Ymm yreg(self->vreg_dst(jj).getIdx());
+        self->vextracti128(xreg_tmp, yreg, ll / 2);
+        if (ll % 2 == 1) self->vpsrldq(xreg_tmp, xreg_tmp, 8);
+
+        if (self->jpp.src_dt == data_type::s8)
+            self->vpmovsxbd(dst_s32, xreg_tmp);
+        else
+            self->vpmovzxbd(dst_s32, xreg_tmp);
+    }
+};
+
+template <>
+struct max_postops_unpacker_t<sse41> {
+    static void load_dst_chunk(jit_uni_i8i8_pooling_fwd_ker_t<sse41> *self,
+            int jj, int ll, const Xbyak::Xmm &dst_s32) {
+        // Use vreg_tail as temp register since tail is not supported for max+postops
+        Xbyak::Xmm xreg_tmp(self->vreg_tail.getIdx());
+        self->movups(xreg_tmp, Xbyak::Xmm(self->vreg_dst(jj).getIdx()));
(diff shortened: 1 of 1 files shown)


jstachowintel:
make test
set test_scope=NIGHTLY

jstachowintel:
make test
set test_scope=NIGHTLY

jstachowintel:
make test
set test_scope=NIGHTLY

tczeszun (approved):

jstachowintel:
make test
set test_scope=NIGHTLY

jstachowintel:
Logs from `OMP_NUM_THREADS=8 ./tests/benchdnn/benchdnn --pool --batch=test_pool_ci`:

`main`
| Implementation        | Count | Share |
|----------------------|------:|------:|
| jit:avx512_core      | 1038  | 40%   |
| ref:any              | 804   | 31%   |
| jit_int:avx512_core  | 546   | 21%   |
| simple_nhwc:any      | 120   | 5%    |
| simple_nchw:any      | 102   | 4%    |

`jstachow/jit_int_post_ops`
| Implementation        | Count | Share |
|----------------------|------:|------:|
| jit:avx512_core      | 1038  | 40%   |
| jit_int:avx512_core  | 702   | 27%   |
| ref:any              | 648   | 25%   |
| simple_nhwc:any      | 120   | 5%    |
| simple_nchw:any      | 102   | 4%    |

As we can see, 156 test cases shifted from `ref:any` to `jit_int:avx512_core`

src/cpu/x64/jit_uni_i8i8_pooling.cpp
```diff
@@ -1389,10 +1572,25 @@ status_t jit_uni_i8i8_pooling_fwd_ker_t<isa>::init_post_ops_conf(
...
+        using namespace injector;
+        const bool po_ok = post_ops_ok(post_ops_ok_args_t(isa,
+                {binary, eltwise}, post_ops, &dst_d,
+                false /*sum_at_pos_0_only*/, false /*sum_requires_scale_one*/,
+                false /*sum_requires_zp_zero*/,
+                false /*sum_requires_same_params*/,
+                get_supported_bcast_strategies()));
+        if (!po_ok) return status::unimplemented;
+        return status::success;
+    }
```
dzarukin: My understanding is this piece of code is duplicating the one below and can be completely dropped.

src/cpu/x64/jit_uni_i8i8_pooling.cpp
```diff
@@ -921,6 +981,123 @@ void jit_uni_i8i8_pooling_fwd_ker_t<isa>::compute_max_step(
...
+                        reg_dst_f32.getIdx(), rhs_arg_params);
+
+                uni_vcvtps2dq(reg_dst_s32, reg_dst_f32);
+                if (is_u8) uni_vpmaxsd(reg_dst_s32, reg_dst_s32, vreg_zeros);
+
+                store_dst_max_op(jj, 0, jj * c_block * sizeof_dst_dt(), masked,
+                        jpp.tail[0]);
+            }
+            return;
+        }
```
dzarukin: As I can see, this piece of code doesn't have to separated from the rest of the code below as it simply repeats phase 2 stuff. Could you, please, incorporate both?

src/cpu/x64/jit_uni_i8i8_pooling.cpp
```diff
@@ -921,6 +981,123 @@ void jit_uni_i8i8_pooling_fwd_ker_t<isa>::compute_max_step(
...
+            for (int ll = 0; ll < max_num_ll; ll++) {
+                tail_per_ll[ll] = m & msk_msk;
+                m >>= msk_gran;
+            }
+        }
+
+        // For AVX-512: reload opmask registers with per-ll f32-level
+        // masks needed by the binary injector and Phase 3 store.
+        // The byte-level masks loaded by init_mask() are no longer
+        // needed since the max pooling loop has already completed.
```
dzarukin: Is it actually no longer needed? I can see this is a part of `compute_step()` which may be called several `c_iter` times but mask gets initialized only once at the beginning of computations.

src/cpu/x64/jit_uni_i8i8_pooling.cpp
```diff
@@ -921,6 +981,123 @@ void jit_uni_i8i8_pooling_fwd_ker_t<isa>::compute_max_step(
...
+                        rhs_arg_params.vmm_tail_idx_.emplace(
+                                reg_dst_f32.getIdx());
+                }
+                postops_injector_->compute_vector(
+                        reg_dst_f32.getIdx(), rhs_arg_params);
+
+                uni_vcvtps2dq(reg_dst_s32, reg_dst_f32);
+                if (is_u8) uni_vpmaxsd(reg_dst_s32, reg_dst_s32, vreg_zeros);
+            }
+        }
```
dzarukin: Could you, please, explain how it works, I'm not sure I have clear understanding.
int8 output is packed with 64 elements per Zmm (let's stick to AVX512 for now).
One would require 4 Zmm to convert that into s32/f32. The code does it completely in separated loops - conversion first, then applying, then converting back.

Based on that, you'd need `ur_c` Zmms to save original int8 dst with conversions on registers needing `5 * ur_c` registers in total. Are those available for use? If it is somehow true for AVX512 with 32 vregs, what about AVX2's 16 vregs available only? 

I was thinking in rather conservative conversions - pick 16 int8 values, convert them into a different register, apply post-ops, convert them back. In this case it would require only one extra per `ur_c` vreg, thus, with total `2 * ur_c`.

src/cpu/x64/jit_uni_i8i8_pooling.cpp
```diff
@@ -877,6 +881,62 @@ void jit_uni_i8i8_pooling_fwd_ker_t<avx512_core>::compute_max_op(const int jj) {
...
+        // which holds the min init value needed across c_block iterations.
+        self->movups(dst_s32, Xbyak::Xmm(self->vreg_dst(jj).getIdx()));
+        if (ll > 0) self->psrldq(dst_s32, 4 * ll);
+
+        if (self->jpp.src_dt == data_type::s8)
+            self->pmovsxbd(dst_s32, dst_s32);
+        else
+            self->pmovzxbd(dst_s32, dst_s32);
+    }
+};
```
dzarukin, line 930: I think you can enable post-ops for AVX2+ dropping SSE41 in init and from here. We don't test it anyway.

src/cpu/x64/jit_uni_i8i8_pooling.cpp
jstachowintel: Yes, the reload is needed. `init_mask()` loads a byte-level mask into `mask(0)` (e.g. a 48-bit mask for c_tail=48 with 64 u8 elements per Zmm). However, the post-ops path processes data in s32 chunks and needs per-ll masks at s32 granularity (e.g. 16-bit masks for 16 s32 elements per Zmm). These are fundamentally different mask widths, so the reload converts the single byte-level `tail[0]` into per-ll s32-level masks.

The reload is guarded by `c_tail != 0`, so it only runs during the tail iteration (the last `compute_step` call). Main-loop iterations pass `c_tail = 0` and skip it entirely — the byte-level masks from `init_mask()` survive intact until the tail call's max pooling loop finishes using them.

jstachowintel:
make test
set test_scope=NIGHTLY

vpirogov:
Closing as stale.

PR #4611, opened 2026-01-30
xe: gemm: add host scalar support for dst zero points

skazakov1:
The last part of host scalar support for jit gemm - for DST zero points.
Addresses https://jira.devtools.intel.com/browse/MFDNN-14149.

key points:
- new gemm kernel argument "co_host_scalar"
- generator<hw>::gemmScalarBinaryOpC() w/ argument from subregister
- max consistently with A/B support

PR #4695, opened 2026-02-17
[GPU] Fix unexpected tf32 type error

Simonsays095:
Addresses [MFDNN-14638](https://jira.devtools.intel.com/browse/MFDNN-14638).

The main issue is in the kernel retry process, where something like the following happens:
1. Try using a tf32 kernel
2. Set the external problem type to tf32 (interpret f32 directly as tf32)
3. Kernel generation fails (this was exposed by a bug in #4611, also fixed here in the 2nd commit)
4. Try another tf32 kernel
5. Try setting external type to tf32 again - but it's already tf32, not f32. This is unexpected and throws an assertion error.

The fix is to reset the problem with every kernel retry, that way the 1st kernel's details don't leak into the 2nd attempted kernel.

### DIFF
--- a/src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
+++ b/src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
@@ -270,7 +270,7 @@ struct GEMMProblem : public CommonProblem {
     bool needsAGroupSums() const { return (bOffset == ABOffset::Calc && quantized2DB() && !earlyDequantizableOffset(Tb_ext, Tbo, Tb)); }
     bool needsBGroupSums() const { return (aOffset == ABOffset::Calc && quantized2DA() && !earlyDequantizableOffset(Ta_ext, Tao, Ta)); }
 
-    bool usesCOPtr() const { return (hasCOffsetPtr() && ((cOffset != COffset::None) || sumA || sumB)); }
+    bool usesCOPtr() const { return (hasCOffsetPtr() && cOffset != COffset::None) || sumA || sumB; }
     bool allowMatrixOffset() const { return (cOffset == COffset::Pre); }
 
     Type Tc_compute() const {


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
```diff
@@ -270,7 +270,7 @@ struct GEMMProblem : public CommonProblem {
     bool needsAGroupSums() const { return (bOffset == ABOffset::Calc && quantized2DB() && !earlyDequantizableOffset(Tb_ext, Tbo, Tb)); }
     bool needsBGroupSums() const { return (aOffset == ABOffset::Calc && quantized2DA() && !earlyDequantizableOffset(Ta_ext, Tao, Ta)); }
 
-    bool usesCOPtr() const { return (hasCOffsetPtr() && ((cOffset != COffset::None) || sumA || sumB)); }
+    bool usesCOPtr() const { return (hasCOffsetPtr() && cOffset != COffset::None) || sumA || sumB; }
```
kealan-barbieri, line 273: is the `hasCOffsetPtr` check just redundant if `sumA`/`sumB` is true?

dyoussif (approved):

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -383,12 +383,9 @@ gen_nocopy_desc_t::select_kernel(compute::gpu_arch_t arch, int stepping,
     disable_systolic_ = !has_systolic;
     relaxed_acc_ = mode & mode_relaxed_acc;
 
-    // Set up problem structure.
-    problem_ = problem;
```
h-sadia, line 387: I think the reason behind this is that internal variables/objs  in the library always have a _ in their name.

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
Simonsays095: Essentially, yes. This function controls whether we claim registers for `offsetCO`, which is needed for `sumA` and `sumB` cases regardless of the `cOffset`/bias settings.

src/gpu/intel/gemm/jit/gen_kernel.cpp
Simonsays095: I'm not removing the `problem_` variable from the struct, I'm just not setting it in this function. Instead, it gets set (reinitialized) right after we select a kernel to try to build.

atkassen:
make test perf-gpu
set primitive=gpu:gemm

PR #4696, opened 2026-02-17
common: vtune: fix perf regressions due to ITT metadata representation

avmanerikar:
# Description

The PR introduces a fix for the perf regressions observed due to metadata-related updates made for VTune representation in PR #3714. 

The regressions arise when constructing metadata strings during ITT task setup for cached primitive creation. The strings are constructed using calls to `pd->info()` which utilize the cached `info_` string whenever there is a cache hit. But this is not the case for nested primitives and the string is repeatedly constructed for each call adding to the overhead.

The fix includes adding an option for `pd->info()` to disable `info_` string construction if it is not already initialized and return an empty string instead. This allows us to skip adding metadata for ITT tasks when the info() string is not present.

Fixes [MFDNN-14521](https://jira.devtools.intel.com/browse/MFDNN-14521)

### DIFF
--- a/src/common/primitive_iface.cpp
+++ b/src/common/primitive_iface.cpp
@@ -68,10 +68,8 @@ status_t primitive_create(primitive_iface_t **primitive_iface,
 #if defined(DNNL_ENABLE_ITT_TASKS)
     const bool enable_itt = itt::get_itt(itt::__itt_task_level_low);
     if (enable_itt) {
-        const char *pd_info = primitive_desc_iface->info();
-        auto task_id = itt::make_itt_id(pd_info, get_msec());
-        itt::primitive_task_start(primitive_desc_iface->impl()->kind(), pd_info,
-                VERBOSE_create, task_id);
+        itt::primitive_task_start(
+                primitive_desc_iface->impl()->kind(), VERBOSE_create);
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


src/common/primitive_desc.hpp
```diff
@@ -70,8 +70,20 @@ struct primitive_desc_t : public c_compatible {
...
+    //     tensor info, attributes and problem description.
+    // Note: If the info string is not already constructed, it will be
+    //     automatically initialized and built before returning.
+    //     This can be optionally disabled by clearing `do_init` - in that
+    //     case, the call to `info()` returns an empty string if the info_
+    //     is not yet constructed. This feature avioids overheads associated
+    //     with building very large info_ strings.
+    // Note: The `do_init` workaround is utilized during ITT task setup for
+    //     primitive operations, since the info_ string used as metadata for
+    //     ITT tasks and can slow down execution times.
```
dzarukin:
```suggestion
    // Returns an `info` string for the primitive configuration including
    //     details like name, implementation, prop_kind, tensor info,
    //     attributes and problem description.
    // Note: If the info string hasn't already been constructed, it will be
    //     automatically initialized and built before returning.
    //     This can be optionally disabled by setting `do_init` argument to
    //     `false` - in that case, the call to `info(...)` returns an empty string
    //     if the `info_` is empty. This helps avoiding overheads associated
    //     with constructing `info_` strings when it isn't desired.
    // Note: The `do_init` workaround is utilized during ITT task setup for
    //     primitive operations, since the `info_` string used as metadata
    //     for ITT tasks and can significantly slow down execution times.
```

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/primitive_iface.cpp
```diff
@@ -111,12 +127,22 @@ status_t primitive_execute(
...
+        itt::primitive_task_start(pd->impl()->kind(), VERBOSE_exec);
+
+        // Before metadata is added to the ITT task, the `info()` string is
+        // checked for initialization.
+        // An empty info_ string implies kernel or nested primitive operations,
+        // in which case the metadata addition is skipped to avoid additional
+        // overheads.
+        const auto *pd_info
+                = pd->impl()->info(primitive_iface->engine(), false);
+        if ((pd_info && pd_info[0] != '\0')) {
```
dzarukin: I'd think that checking for empty pd_info should be enough? What's the scenario when it's not empty but has a single escape symbol in it?

dzarukin (approved):

src/common/primitive_iface.cpp
avmanerikar: I observed that because `info()` returns `info_.c_str()`, it will always return a non-null pointer even when `info_` is not initialized. (`std::string::c_str()` will return a valid pointer to a NUL-terminated buffer for the empty string).

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/primitive_iface.cpp
dzarukin: I think this should work.
```suggestion
        if ((pd_info && *pd_info)) {
```

avmanerikar: Applied. Thanks!

avmanerikar:
make test
set test_scope=NIGHTLY

mzhukova (approved):

PR #4697, opened 2026-02-17
CPU: x64: deconv added checks in fallback code

georgen117:
[MFDNN-14627](https://jira.devtools.intel.com/browse/MFDNN-14627)

This adds checks:
 - the destination data type and bias data type match.
 - the data types for the fallback are floating data types.

The convolution failure was a result of the fallback code choosing an older jit kernel when the bias data type and destination data type did not match.

Additionally added a check to not use the fallback code except for floating data types.

This is purposely reducing the possible cases the fallback loop is used.

Note: Adding the bias data type check to the init code of the jit kernels was investigated, but and was rejected due to the jit convolution kernels inability to handle mixed data types. Other deconvolution code did handle the mixing of data types correctly.

### DIFF
--- a/src/cpu/x64/jit_brgemm_deconv.cpp
+++ b/src/cpu/x64/jit_brgemm_deconv.cpp
@@ -211,15 +211,32 @@ status_t brgemm_deconvolution_fwd_t<isa>::pd_t::init(engine_t *engine) {
 
         // Second pass: fallback to any other backward data convolution implementation
         // This allows non-BRGEMM kernels (like jit_avx512_core_bf16) to handle
-        // cases with uneven spatial dimensions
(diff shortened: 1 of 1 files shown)


georgen117:
make test
set test_scope=NIGHTLY
disable os_win
disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv

georgen117:
make test

georgen117:
Closing this. Although it fixes the breaks from MFDNN-14627 it reintroduces the issue that was MFDNN-14380 was trying to solve. I need to continue investigating.

PR #4700, opened 2026-02-18
cpu: x64: conv: remove unnecessary work around

yair-obodovsky:
Remove unnecessary flow that touches every page in the input buffer in conv layer.
Related to jira: https://jira.devtools.intel.com/browse/MFDNN-14582


yair-obodovsky:
make test
set test_scope=NIGHTLY

vpirogov:
There's one more spot with similar workaround:
https://github.com/uxlfoundation/oneDNN/blob/3e50d903dbba0e958da92353b930523f5db8d677/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp#L783-L789

I do not see anything in matmul, but it's worth double checking.

yair-obodovsky:
> There's one more spot with similar workaround:
> 
> https://github.com/uxlfoundation/oneDNN/blob/3e50d903dbba0e958da92353b930523f5db8d677/src/cpu/x64/jit_brgemm_conv_bwd_strided.cpp#L783-L789
> 
> I do not see anything in matmul, but it's worth double checking.

Removed from conv_bwd. We didn’t find it anywhere else.

PR #4701, opened 2026-02-18
cpu: x64: matmul: added checks for invalid blocking options

yair-obodovsky:
Added validation for invalid blocking configurations. The issue was reproducible when heuristics received incorrect cache size values.


yair-obodovsky:
make test
set test_scope=NIGHTLY

PR #4688, opened 2026-02-13
xe: conv: jit: fix incorrect get_fma_friendly_layout

rjoursler:
For grouped BWD_W, batch dimension blocks were not included in the B buffer dimension mapping. This may result in get_fma_friendly_layout failing when vectorization is performed over the batch dimension.

Fixes [MFDNN-14630](https://jira.devtools.intel.com/browse/MFDNN-14630).

PR #4702, opened 2026-02-18
[rls-v3.11] xe: conv: jit: fix incorrect get_fma_friendly_layout

rjoursler:
For grouped BWD_W, b_blocks_ did not include the group dimension for the dst buffer, causing get_fma_friendly_layout to fail as that is the vectorized dimension.

Backport #4688.


rjoursler:
make test
disable test_device_cpu

PR #4656, opened 2026-02-06
cpu: x86: deconv fallback to older deconv on brgemm failure

georgen117:
Fix for [MFDNN-14380](https://jira.devtools.intel.com/browse/MFDNN-14380)

This is an update to the reverted changes for https://github.com/uxlfoundation/oneDNN/pull/4578

PR#4578 was caused failures in the deconvolution that were caught by the nightly build and was reverted in responce. See [MFDNN-14600](https://jira.devtools.intel.com/browse/MFDNN-14600)

Investigating the failures from MFDNN-14600 showed it was due to the fallback kernels only using bias if the kernel was forward inference.

This contains the same fallback code that was commited for PR#4578 with additional guard preventing fallback if bias exists for anything other than forward inference.

The initial issue is at `jit_brgemm_conv_bwd_utils.cpp:1460:`

`if (cd.use_inversion && has_uneven_spatial) return status::unimplemented;`

This returns unimplemented when using deconvolution (use_inversion is true) with uneven spatial dimensions. The has_uneven_spatial condition checks if the spatial dimensions are not evenly divisible by the strides:

```cpp
const bool has_uneven_spatial = jcp.id % jcp.stride_d != 0 
                              || jcp.ih % jcp.stride_h != 0
							  || jcp.has_uneven_iw;
```

In the case reported by MFDNN-14380, we have have:

ih127 (input height = 127) with sh2 (stride = 2): 127 % 2 = 1 (not evenly divisible)
iw127 (input width = 127) with sw2 (stride = 2): 127 % 2 = 1 (not evenly divisible)
id127 (input depth = 127) with sd2 (stride = 2): 127 % 2 = 1 (not evenly divisible)

**The problem:** The brgemm deconvolution implementation doesn't support spatial dimensions that aren't evenly divisible by the strides when using the backwards-strided path.

This results in the deconvolution running the reference implementation of the post-ops. Causing a significant performance degradation.

**The solution:** Fallback on `jit_avx512_*` or `convolution_gemm` in the cases that brgemm deconv is not supported.

This issue can be reproduced using the following benchdnn commands:

Any tag:

[...]

PR #4709, opened 2026-02-19
revert: cpu: x86: deconv fallback to older deconv on brgemm failure

georgen117:
[MFDNN-14627](https://jira.devtools.intel.com/browse/MFDNN-14627)

This reverts commit fcb57d030d3793e9d298af4b59ddb98ef8d87f01.

https://github.com/uxlfoundation/oneDNN/pull/4656

The fallback code is failing in two situations.
When the bias data type does not match the destination data type.
It also fails when trying to fall back for s8:s8:s32 data types on a system without vnni.

I created a PR that resolved both those issues https://github.com/uxlfoundation/oneDNN/pull/4697 only to discover it would prevent the code catching the original shapes from [MFDNN-14380](https://jira.devtools.intel.com/browse/MFDNN-14380).

I will need more time to investigate an additional fix for MFDNN-14380 in the meantime to prevent nightly build failures I am reverting this change.


georgen117:
make test

PR #4710, opened 2026-02-20
Level Zero: add benchdnn, gtests and address feedback comments

dzarukin:
Continuation of #4689

Addresses all the feedback found during previous two PRs review.
Adds benchdnn and gtest validation.
So far, the plan is to validate Level Zero backend on Xe2+ platforms since there's no functional support for OpenCL kernels compilation in Level Zero for pre-Xe2 systems and there's no plan to backport such support there.

Enabling validation in regular processes would start post this PR promotion.


dzarukin:
make test

src/gpu/intel/ze/engine.cpp
```diff
@@ -75,25 +75,28 @@ status_t engine_t::create_kernel(
...
-    ze_module_handle_t ze_module = nullptr;
+    std::shared_ptr<xpu::ze::wrapper_t<ze_module_handle_t>> ze_module_ptr;
     std::vector<ze_kernel_handle_t> ze_kernels;
-    CHECK(ze::create_kernels(
-            device(), context(), kernel_names, binary, &ze_module, ze_kernels));
-    auto ze_module_ptr
-            = std::make_shared<xpu::ze::wrapper_t<ze_module_handle_t>>(
-                    ze_module);
+    CHECK(ze::create_kernels(device(), context(), kernel_names, binary,
+            &ze_module_ptr->unwrap(), ze_kernels));
```
echeresh: Isn't `ze_module_ptr` a null pointer here which we first need to construct?

src/gpu/intel/ze/stream.cpp
```diff
@@ -27,28 +27,37 @@ namespace ze {
...
         device_properties.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES_1_2;
+        CHECK(xpu::ze::zeDeviceGetProperties(
+                utils::downcast<engine_t *>(engine())->device(),
+                &device_properties));
 
-        xpu::ze::zeDeviceGetProperties(
-                utils::downcast<engine_t *>(engine)->device(),
-                &device_properties);
+        uint64_t max_timestamp_value
+                = (1 << device_properties.kernelTimestampValidBits) - 1;
```
echeresh: Can `device_properties.kernelTimestampValidBits` be >= 32 bits?

```suggestion
                = (1ULL << device_properties.kernelTimestampValidBits) - 1;
```

tests/gtests/ze/api/test_memory.cpp
```diff
@@ -0,0 +1,326 @@
...
+    memory::dim n = 100;
+    memory::desc mem_d({n}, memory::data_type::f32, memory::format_tag::x);
+
+    std::vector<float> system_buf(n);
+    bool system_memory_supported = false;
+
+    ze_driver_handle_t ze_driver = nullptr;
+    ze_device_handle_t ze_dev = nullptr;
+    ze_context_handle_t ze_ctx = nullptr;
+    find_ze_device(&ze_driver, &ze_dev, &ze_ctx);
```
echeresh, line 200: Looks like `ze_ctx` is never destroyed

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1143,6 +1154,18 @@ int get_gpu_ram_sizes(size_t &ram_size, size_t &max_alloc_size) {
...
+#elif DNNL_GPU_RUNTIME == DNNL_RUNTIME_ZE
+    auto eng = dnnl::engine(get_test_engine(), true);
+    auto ze_dev = dnnl::ze_interop::get_device(eng);
+    ze_device_properties_t ze_dev_props;
+    ze_dev_props.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
+    ze_dev_props.pNext = nullptr;
+    auto status = zeDeviceGetProperties(ze_dev, &ze_dev_props);
+    if (status != ZE_RESULT_SUCCESS) return FAIL;
+
+    ram_size = (size_t)ze_dev_props.maxMemAllocSize;
```
echeresh: Is it a typo? or doesn't L0 have a dedicated query like SYCL `global_mem_size`?

src/gpu/intel/ze/device_info.cpp
```diff
@@ -121,19 +121,20 @@ status_t device_info_t::init_attributes(impl::engine_t *engine) {
             device_cache_properties_count);
     for (ze_device_cache_properties_t &p : device_cache_properties) {
         p.stype = ZE_STRUCTURE_TYPE_DEVICE_CACHE_PROPERTIES;
-        p.pNext = nullptr;
     }
 
     CHECK(xpu::ze::zeDeviceGetCacheProperties(device,
             &device_cache_properties_count, device_cache_properties.data()));
-    l3_cache_size_ = device_cache_properties[0].cacheSize;
+    if (device_cache_properties[0].flags == 0) {
```
echeresh: Nit: maybe better to use a loop like in `src/gpu/intel/gemm/jit/dsl/runtime.cpp`? For consistency and just in case, if `device_cache_properties` is empty.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1192,6 +1215,20 @@ int get_gpu_cache_size(size_t &cache_size) {
...
+    ze_cache_props.stype = ZE_STRUCTURE_TYPE_DEVICE_CACHE_PROPERTIES;
+    ze_cache_props.pNext = nullptr;
+    uint32_t count = 0;
+    auto status = zeDeviceGetCacheProperties(ze_dev, &count, nullptr);
+    if (status != ZE_RESULT_SUCCESS) return FAIL;
+
+    status = zeDeviceGetCacheProperties(ze_dev, &count, &ze_cache_props);
+    if (status != ZE_RESULT_SUCCESS) return FAIL;
+
+    _cache_size = ze_cache_props.cacheSize;
```
echeresh: Please align with the other place when `cacheSize` is accessed, need `flags == 0` check and loop.

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1192,6 +1215,20 @@ int get_gpu_cache_size(size_t &cache_size) {
...
+    auto eng = dnnl::engine(get_test_engine(), true);
+    auto ze_dev = dnnl::ze_interop::get_device(eng);
+    ze_device_cache_properties_t ze_cache_props;
+    ze_cache_props.stype = ZE_STRUCTURE_TYPE_DEVICE_CACHE_PROPERTIES;
+    ze_cache_props.pNext = nullptr;
+    uint32_t count = 0;
+    auto status = zeDeviceGetCacheProperties(ze_dev, &count, nullptr);
+    if (status != ZE_RESULT_SUCCESS) return FAIL;
+
+    status = zeDeviceGetCacheProperties(ze_dev, &count, &ze_cache_props);
```
echeresh: I think we need to define `ze_cache_props` as `std::vector` with `count` elements.

tests/gtests/ze/api/test_engine.cpp
```diff
@@ -0,0 +1,144 @@
...
+*******************************************************************************/
+
+#include "dnnl_test_common.hpp"
+#include "gtest/gtest.h"
+
+#include "oneapi/dnnl/dnnl_ze.h"
+#include "oneapi/dnnl/dnnl_ze.hpp"
+
+#include <string>
+#include <CL/cl.h>
```
echeresh: Is it unused?

src/gpu/intel/ze/engine.cpp
dzarukin: True, it will happen [here](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/ze/utils.cpp#L244-L248). We pass an underlying pointer which will be initialized to the caller.

src/gpu/intel/ze/stream.cpp
dzarukin: It's `uint32_t` but I'll apply the change anyway. Thanks.

tests/gtests/ze/api/test_memory.cpp
dzarukin: Indeed, thanks for spotting.

src/gpu/intel/ze/engine.cpp
echeresh: Maybe I'm missing something but what I meant is that as is this code should produce nullptr dereference. Please check this [snippet](https://godbolt.org/z/zqqofhW5h). Is it the same idea?

tests/benchdnn/dnnl_common.cpp
dzarukin: It wasn't. I failed to identify the right marker for the first time. I gave it a second chance and found it, thanks for noticing.

src/gpu/intel/ze/device_info.cpp
dzarukin: Absolutely, I must close my eyes when I was coding. Thanks.

tests/benchdnn/dnnl_common.cpp
dzarukin: Makes sense, there are just two places - device_info and benchdnn. Now they code stuff identically. Thanks.

tests/gtests/ze/api/test_engine.cpp
dzarukin: It was, a product of copy-paste, thanks for spotting, cleaned all gtest files from unneeded headers.

src/gpu/intel/ze/engine.cpp
dzarukin: You got it right, it was indeed crashing, testing didn't get to it on BMG.
I think I've fixed it now by using `shared_ptr` on the interface level and fill it right after the module was constructed. Thank you for the catch!

rjoursler (approved):

src/xpu/ze/stream_impl.cpp
```diff
@@ -121,20 +121,36 @@ status_t stream_impl_t::barrier() {
     return status::success;
 }
 
+status_t stream_impl_t::init_flags(
+        unsigned *flags, ze_command_list_handle_t list, bool profiling) {
+    *flags = 0;
+    // Note: determininig if the passed list is in-order/out-of-order is
```
mzhukova:
```suggestion
    // Note: determining if the passed list is in-order/out-of-order is
```

src/gpu/intel/ze/engine.cpp
```diff
@@ -116,6 +119,8 @@ status_t engine_t::create_kernels(std::vector<compute::kernel_t> *kernels,
     const char *source = nullptr;
     for (size_t i = 0; source == nullptr && i < kernel_names.size(); i++)
         source = intel::get_kernel_source(kernel_names[i]);
+    VERROR_ENGINE(source, status::runtime_error,
+            "No OpenCL source was found for kernel");
```
mzhukova, line 123: OpenCL -- intentional error msg?

src/gpu/intel/ze/utils.cpp
```diff
@@ -243,9 +243,14 @@ status_t compile_ocl_module_to_binary(ze_device_handle_t device,
...
         const std::vector<const char *> &kernel_names,
-        const xpu::binary_t &binary, ze_module_handle_t *module_ptr,
+        const xpu::binary_t &binary,
+        std::shared_ptr<xpu::ze::wrapper_t<ze_module_handle_t>> &module_ptr,
         std::vector<ze_kernel_handle_t> &kernels) {
-    CHECK(compile_native_module(module_ptr, device, context, binary));
+    ze_module_handle_t ze_module = nullptr;
+    CHECK(compile_native_module(&ze_module, device, context, binary));
+
+    module_ptr = std::make_shared<xpu::ze::wrapper_t<ze_module_handle_t>>(
```
mzhukova, line 255: wouldn't there be a problem still (leak) if this throws bad_alloc or something?

src/gpu/intel/ze/engine.cpp
dzarukin: It is intentional, this piece works with OpenCL sources.

src/gpu/intel/ze/utils.cpp
dzarukin: What's your suggestion for such scenario?

mzhukova: could swapping the order of `compile_native_module` and `make_shared` help? (e.g., "pre-allocate", so if it fires early - no issue, and then seems like there is `unwrap` that helps writing back)

dzarukin: I might have a change but I'm not comfortable to introduce it without validation which is not there yet. I'll postpone this change once infra is set up.

src/xpu/ze/stream_impl.cpp
dzarukin: Thanks, updated.

PR #4714, opened 2026-02-20
add benchdnn support of grouped gemm

mzhukova:
Initial grouped gemm support in benchdnn (to allow testing features targeting 3.12-pc):
- `--grouped=DIM_IDX:NUM_GROUPS:size0,size1,...,sizeN` option to specify variable dimension, group count and sizes per group.
  - example: `--grouped=0:4:32,0,32,96 160x512:4x512x256` (for `dim_idx=0`, the `--grouped` is followed by `total_MxK:GxKxN`, where G is number of groups, to explicitly set WEI as 3D tensor)
- `5` and `7` policies are used to express mask required for WEI in order to set per-expert-per-column scales/zps and per-expert-blocked scales/zps, picked up changes from https://github.com/uxlfoundation/oneDNN/pull/4726 to [test](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f1133e37-f949-f137-9c57-a4bf010d0e2d).

Next steps:
[MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704)

### DIFF
--- a/tests/benchdnn/matmul/matmul.cpp
+++ b/tests/benchdnn/matmul/matmul.cpp
@@ -49,12 +49,13 @@ namespace matmul {
 // - Input validation is done in verify_grouped_input()
 static benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_grouped_md(
         const prb_t *prb, data_kind_t kind, dnnl_data_type_t dt) {
+    dnnl_memory_desc_t md {};
     int arg = (kind == SRC) ? DNNL_ARG_SRC
             : (kind == DST) ? DNNL_ARG_DST
                             : DNNL_ARG_UNDEF;
-    if (arg == DNNL_ARG_UNDEF) return nullptr;
-    if (prb->sparse_options.get_encoding(arg) != dnnl_grouped) return nullptr;
-    if (prb->sparse_options.get_variable_dim_idx(arg) != 0) return nullptr;
+    if (arg == DNNL_ARG_UNDEF) return md;
+    if (prb->sparse_options.get_encoding(arg) != dnnl_grouped) return md;
+    if (prb->sparse_options.get_variable_dim_idx(arg) != 0) return md;
 
     const int64_t group_count = prb->sparse_options.get_group_count();
 
@@ -85,6 +86,7 @@ dims_t get_runtime_dims(const dims_t &dims, const dims_mask_t &mask) {
 // start supporting it.
 benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_md(const prb_t *prb,
         data_kind_t kind, dnnl_data_type_t dt = dnnl_data_type_undef) {
+    dnnl_memory_desc_t md {};
     if (kind == SRC) {
         if (dt == dnnl_data_type_undef) dt = prb->src_dt();
         const auto &src_rt_dims = get_runtime_dims(
@@ -108,7 +110,7 @@ benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_md(const prb_t *prb,
                     return dnn_mem_t::init_coo_md(
                             prb->ndims, src_rt_dims.data(), dt, nnz, dnnl_s32);
                     break;
-                default: assert(!"unsupported encoding"); return nullptr;
+                default: assert(!"unsupported encoding"); return md;
             }
         } else
             return dnn_mem_t::init_md(prb->ndims, src_rt_dims.data(), dt,
@@ -137,7 +139,7 @@ benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_md(const prb_t *prb,
                     return dnn_mem_t::init_sparse_packed_md(
                             prb->ndims, weights_rt_dims.data(), dt, nnz);
                     break;
-                default: assert(!"unsupported encoding"); return nullptr;
+                default: assert(!"unsupported encoding"); return md;
             }
         } else {
             // for grouped matmul, prb->ndims is not equal to the actual number
@@ -162,7 +164,7 @@ benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_md(const prb_t *prb,
         return dnn_mem_t::init_md(prb->ndims, dst_rt_dims.data(), dt, prb->dtag,
                 prb->strides[STRIDES_DST]);
     }
-    return nullptr;
+    return md;
 }
 
 dnnl_status_t init_pd(init_pd_args_t<prb_t> &init_pd_args) {


tests/benchdnn/doc/driver_matmul.md
```diff
@@ -26,15 +26,28 @@ where *matmul-knobs* are:
...
             `weights` that indicates whether a dimension is
             `DNNL_RUNTIME_DIM_VAL` (indicated as 1-bit in the corresponding
             dimension position). The default is `0` for all dimensions, meaning
             all tensor dimensions are fully defined at primitive creation. For
             tensors with option values other than `0`, a correspondent memory
             format tag must be specified.
-- `--encoding=STRING` - sparse encodings and sparsity. No encodings are set by
+ - `--encoding=STRING` - sparse encodings and sparsity. No encodings are set by
             default. Refer to [encodings](knobs_encoding.md) for details.
+ - `--grouped=DIM_IDX:NUM_GROUPS:size0,size1,...,sizeN` -- specifies grouped
```
dzarukin, line 41: Minor: once `size` arguments are gone, I would think that the option would look like:
`--grouped=NUM_GROUPS[:DIM_IDX]` as dim_idx would be 0 95-99% of all times and there's not much sense to type it every time, thus, keep an optional argument at the end.

Saying that, you might want to swap them in the initial version to avoid disruption once the option gets modified, if you have time for that.

mzhukova:
ping @uxlfoundation/onednn-arch to review :)

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
```diff
@@ -0,0 +1,127 @@
+# Grouped GEMM tests
```
dzarukin, line 1: Please check commit `b85ca83fa` on main. You'll need to re-create it for GROUPED functionality to avoid failures for builds that don't have GROUPED enabled.

tests/benchdnn/matmul/bench_matmul.cpp
```diff
@@ -127,10 +127,137 @@ int verify_input(const settings_t &s, const settings_t &def) {
...
+                = s.sparse_options[0].get_variable_dim_idx(DNNL_ARG_SRC);
+
+        // Validate variable_dim_idx (only M dimension supported)
+        if (variable_dim_idx != 0) {
+            BENCHDNN_PRINT(0,
+                    "ERROR: grouped encoding only supports M dimension "
+                    "(dim 0), got dim %d\n",
+                    variable_dim_idx);
+            SAFE_V(FAIL);
+        }
```
dzarukin: This check seems to be a job for the library. API allows it, any integer is a valid index from test perspective.

tests/benchdnn/matmul/bench_matmul.cpp
```diff
@@ -127,10 +127,137 @@ int verify_input(const settings_t &s, const settings_t &def) {
...
+        for (const auto &rt_dims_mask_config : s.rt_dims_masks) {
+            for (const auto &mask : rt_dims_mask_config) {
+                if (mask.any()) {
+                    BENCHDNN_PRINT(0, "%s\n",
+                            "ERROR: grouped encoding does not support runtime "
+                            "dimensions");
+                    SAFE_V(FAIL);
+                }
+            }
+        }
```
dzarukin: Ditto.

tests/benchdnn/matmul/bench_matmul.cpp
```diff
@@ -127,10 +127,137 @@ int verify_input(const settings_t &s, const settings_t &def) {
...
+        // Note: currently not supported, support will require updated filling path for
+        // grouped DST values
+        for (const auto &i_attr : s.attributes) {
+            if (i_attr.post_ops.find(attr_t::post_ops_t::SUM) >= 0) {
+                BENCHDNN_PRINT(0, "%s\n",
+                        "ERROR: grouped matmul does not support sum "
+                        "post-op");
+                SAFE_V(FAIL);
+            }
+        }
```
dzarukin: Ditto. Should return unimplemented by the library (and not used in input lists).

tests/benchdnn/matmul/bench_matmul.cpp
```diff
@@ -127,10 +127,137 @@ int verify_input(const settings_t &s, const settings_t &def) {
...
+                if (i_bia_mask != 2) {
+                    BENCHDNN_PRINT(0,
+                            "ERROR: grouped matmul only supports "
+                            "bia_mask=2 (per-column per-expert bias), "
+                            "got bia_mask=%d\n",
+                            i_bia_mask);
+                    SAFE_V(FAIL);
+                }
+            }
+        }
```
dzarukin: Ditto.

tests/benchdnn/doc/driver_matmul.md
```diff
@@ -33,8 +33,18 @@ where *matmul-knobs* are:
...
+            encoding to execute grouped GEMM. `DIM_IDX` is the dimension index
```
avmanerikar:
```suggestion
            encoding for a given dimension to execute grouped GEMM. `DIM_IDX` is the dimension index
```

tests/benchdnn/utils/parser.cpp
```diff
@@ -710,6 +710,144 @@ bool parse_encoding(std::vector<sparse_options_t> &sparse_options,
...
+                sizes.push_back(std::stoll(
+                        sizes_str.substr(size_pos, comma_pos - size_pos)));
+                size_pos = comma_pos + 1;
+            }
+        } catch (...) {
+            BENCHDNN_PRINT(
+                    0, "%s\n", "Error: invalid size value in group sizes");
+            SAFE_V(FAIL);
+        }
+    }
```
avmanerikar, line 785: Might be useful to add a check for non-negative integers:
```
.>./tests/benchdnn/benchdnn --matmul --dt=f16:f16:f16 --grouped=0:4:8,8,24,-8 --wtag=abc 32x64:4x64x32
Error: Function 'doit' at (/nfs/site/proj/mkl/hal9000/amanerik/oneDNN/tests/benchdnn/matmul/matmul.cpp:1374) returned '1'
0:FAILED (116 ms) __REPRO: --matmul --dt=f16:f16:f16 --grouped=0:4:8,8,24,-8 --wtag=abc 32x64:4x64x32
| ref_grouped:any : 1 (100%)                               |
0:FAILED (116 ms) __REPRO: --matmul --dt=f16:f16:f16 --grouped=0:4:8,8,24,-8 --wtag=abc 32x64:4x64x32
============================
total: 0.14s; create_pd: 0.00s (0%); create_prim: 0.00s (0%); fill: 0.10s (71%); execute: 0.02s (13%); compute_ref: 0.00s (0%); compare: 0.00s (0%);
```

tests/benchdnn/utils/parser.cpp
```diff
@@ -710,6 +710,144 @@ bool parse_encoding(std::vector<sparse_options_t> &sparse_options,
...
 }
 
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+// Format: DIM_IDX:NUM_GROUPS:size0,size1,...,sizeN
+// - DIM_IDX is the dimension index, where src MxK * weights KxN = dst MxN
+//   0 = M,  1 = K, 2 = N
+// - NUM_GROUPS is the number of tensors in the group (number of experts)
+// - size0, size1,..., sizeN are the sizes for each in the group, with sum
+//   equal to the total size along DIM_IDX
+bool parse_grouped(std::vector<sparse_options_t> &sparse_options,
```
dzarukin, line 722: Nit: please add a comment:
`TODO: move to 'parse_single_value_option' once experimental is removed`

tests/benchdnn/dnn_types.cpp
```diff
@@ -1091,21 +1091,44 @@ std::ostream &operator<<(std::ostream &s, dnnl_sparse_encoding_t se) {
...
+                if (!sparse_options.is_encoding_def(arg)) {
+                    s << sparse_options.get_encoding(arg);
+                    if (!sparse_options.is_sparsity_def(arg))
+                        s << "+" << sparse_options.get_sparsity(arg);
+                }
+                if (i != (int)args.size() - 1)
+                    s << ":";
+                else
+                    s << " ";
+            }
```
avmanerikar, line 1131: Random spot: This will also require updating `scripts/verbose_converter` to parse for `--grouped=`

tests/benchdnn/utils/parser.cpp
```diff
@@ -710,6 +710,144 @@ bool parse_encoding(std::vector<sparse_options_t> &sparse_options,
...
+    sparse_options_t v;
+
+    // Parse format: DIM_IDX:NUM_GROUPS:size0,size1,...
+    size_t pos = 0;
+    auto first_colon = s.find(':');
+    if (first_colon == std::string::npos) {
+        BENCHDNN_PRINT(0, "%s\n",
+                "Error: grouped format requires DIM_IDX:NUM_GROUPS:sizes");
+        SAFE_V(FAIL);
+    }
```
dzarukin, line 756: Minor: typically, get_substr function tracks the delimiter position and is `pos` would be equal to std::string::npos, it means it wasn't found and return a fail status. You may check `parse_attr_post_ops_func` for this principle.

tests/benchdnn/utils/parser.cpp
```diff
@@ -710,6 +710,144 @@ bool parse_encoding(std::vector<sparse_options_t> &sparse_options,
...
+
+    std::string dim_idx_str = s.substr(0, first_colon);
+
+    // Parse dimension index
+    int variable_dim_idx = -1;
+    try {
+        variable_dim_idx = std::stoi(dim_idx_str);
```
dzarukin: Nit: you may re-use `parser_utils::stoll_safe`.

tests/benchdnn/utils/parser.cpp
```diff
@@ -710,6 +710,144 @@ bool parse_encoding(std::vector<sparse_options_t> &sparse_options,
...
+    }
+
+    // Validate number of sizes
+    if (sizes.size() != (size_t)group_count) {
+        BENCHDNN_PRINT(0,
+                "Error: number of sizes (%zu) doesn't match "
+                "group_count (%lld)\n",
+                sizes.size(), (long long)group_count);
+        SAFE_V(FAIL);
+    }
```
dzarukin, line 794: I've seen this check in matmul, probably should be removed there and kept here.

tests/benchdnn/utils/compare.cpp
```diff
@@ -265,7 +265,33 @@ int compare_t::compare_p2p(const dnn_mem_t &exp_mem, const dnn_mem_t &got_mem,
...
+        const int64_t nelems_to_copy = got_f32.nelems();
+        benchdnn_parallel_nd(nelems_to_copy, [&](int64_t i) {
+            got_f32.set_f32_elem(i,
+                    got_mem.get_elem(i, sparse_options_t::grouped_values_idx));
+        });
+    } else
+#endif
+    {
+        got_f32 = dnn_mem_t(got_mem, dnnl_f32, tag::abx, get_cpu_engine());
+    }
```
dzarukin: Discussed offline. The agreement is to rely on `execute_reorder` (with adjustments needed to make it work).

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -30,6 +30,131 @@ int64_t wei_ba_off_f(const prb_t *prb, int64_t mb, int64_t k, int64_t n) {
     return (mb * prb->n + n) * prb->k + k;
 }
 
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+// Reference implementation for grouped gemm
+// Computes per-expert matmuls: for each expert e, computes dst[e] = src[e] * wei[e]
+//
+// Note, that all tensors are concatenated
+// src and dst in grouped format and wei in 3D [group_count, K, N] format
+void compute_ref_grouped_matmul(const prb_t *prb, const args_t &args) {
```
dzarukin, line 41: Nit: TODO in the comment from the PR description would be nice.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -354,6 +422,99 @@ int fill_sparse_data(data_kind_t kind, const prb_t *prb, dnn_mem_t &mem_dt,
...
+    const int nhandles = query_md_num_handles(mem_dt.md_);
+    if (nhandles != 2) {
+        BENCHDNN_PRINT(0, "Error: grouped memory requires 2 handles, got %d\n",
+                nhandles);
+        return FAIL;
+    }
+
+    // Fill offsets buffer
+    int64_t total_M = 0;
+    SAFE(fill_grouped_offsets(mem_dt, prb, total_M), WARN);
```
dzarukin: I'm not sure I'm following why `total_M` should be obtained from `fill_grouped_offsets` if you can query it directly from the `prb`?

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -34,6 +34,44 @@
 
 namespace matmul {
 
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+// Helper to create grouped memory descriptor
+//
+// Current input format for grouped matmul is:
+//     --grouped=indx:group_count:size1,size2,...,sizeN total_MxK:group_countxKxN
+// group_count is the number of groups and size1,...,sizeN are the sizes of the variable dimension for each group,
```
dzarukin: Nit: please fit in 80. Feels like IDE could use some setting to highlight/adjust such things.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -34,6 +34,44 @@
...
+//
+// Notes:
+// - Currently supports only M dimension,
+//   therefore only SRC and DST can be created with grouped encoding
+// - Input validation is done in verify_grouped_input()
+static benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_grouped_md(
+        const prb_t *prb, data_kind_t kind, dnnl_data_type_t dt) {
+    int arg = (kind == SRC) ? DNNL_ARG_SRC
+            : (kind == DST) ? DNNL_ARG_DST
+                            : DNNL_ARG_UNDEF;
```
dzarukin, line 55: `data_kind2exec_arg(...)` should handle this.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -34,6 +34,44 @@
...
+    assert(arg != DNNL_ARG_UNDEF && "grouped encoding only supports SRC/DST");
+    assert(prb->sparse_options.get_encoding(arg) == dnnl_grouped
+            && "create_grouped_md should be called only for grouped encoding");
+    assert(prb->sparse_options.get_variable_dim_idx(arg) == 0
+            && "currently only M dimension grouping is supported");
```
dzarukin: Seems like can be converted to straight FAILs:
`SAFE(arg != DNNL_ARG_UNDEF ? OK : FAIL, WARN);`

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -34,6 +34,44 @@
...
+    // [total_M, K] for SRC
+    // [total_M, N] for DST
+    dnnl_dims_t dims_2d;
+    dims_2d[0]
+            = prb->m; // total_M, at this point we've already validated that sum of group sizes equals M dimension
+    dims_2d[1] = (arg == DNNL_ARG_SRC) ? prb->k : prb->n;
+
+    // Create memory descriptor with grouped encoding with multiple handles
+    return dnn_mem_t::init_grouped_md(
+            2, dims_2d, dt, /* variable_dim_idx = */ 0, group_count, dnnl_s32);
```
dzarukin, line 71: `variable_dim_idx ` seems to be queryable from `sparse_options::get_variable_dim_idx(...)`.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -97,15 +140,24 @@ benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_md(const prb_t *prb,
                     break;
                 default: assert(!"unsupported encoding"); return nullptr;
             }
-        } else
-            return dnn_mem_t::init_md(prb->ndims, weights_rt_dims.data(), dt,
-                    prb->wtag, prb->strides[STRIDES_WEI]);
+        } else {
+            return dnn_mem_t::init_md((int)weights_rt_dims.size(),
```
dzarukin, line 147:
```suggestion
            // `prb->ndims` might not coincide with `weights_rt_dims.size()` for grouped_gemm.
            return dnn_mem_t::init_md((int)weights_rt_dims.size(),
```

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -354,6 +422,99 @@ int fill_sparse_data(data_kind_t kind, const prb_t *prb, dnn_mem_t &mem_dt,
...
+
+        for (int64_t i = idx_start; i < idx_end; i++) {
+            float val = gen(int_seed);
+            mem_dt.set_elem(i,
+                    round_to_nearest_representable(cfg.get_dt(kind), val),
+                    sparse_options_t::grouped_values_idx);
+        }
+    });
+
+    // Copy values to fp memory for reference computation
```
dzarukin:
```suggestion
    // Copy values to fp memory for reference computation, since there's no library reorder for this type of md.
```

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -354,6 +422,99 @@ int fill_sparse_data(data_kind_t kind, const prb_t *prb, dnn_mem_t &mem_dt,
...
+            idx_start = 1;
+        }
+
+        for (int64_t i = idx_start; i < idx_end; i++) {
+            float val = gen(int_seed);
+            mem_dt.set_elem(i,
+                    round_to_nearest_representable(cfg.get_dt(kind), val),
+                    sparse_options_t::grouped_values_idx);
+        }
+    });
```
dzarukin: I can see that for some reason you dropped the logic that adjusted src feeling based on density, which will likely lead to overflow output values for large shapes or low precision data types. Have you inspected output values they will provide a full specter of values when filling src always densely?

I think that the part that fills buffers can be moved to a dedicated function and used in both places to have a common ground.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -368,13 +529,27 @@ int fill_data(data_kind_t kind, int exec_arg, const prb_t *prb,
     const bool is_sparse_csr_coo
             = sparse_encoding == dnnl_csr || sparse_encoding == dnnl_coo;
     is_sparse_packed = sparse_encoding == dnnl_packed;
+    const bool is_grouped_dt = (query_md_num_handles(mem_dt.md_) > 1)
```
dzarukin: Nit: suggest to set it to `false` and define only when `sparse_encoding == dnnl_grouped` as a safer approach.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -927,23 +1174,41 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
...
                 const auto &dims = mem.dims();
-                dnnl_dims_t strides {};
-                dnnl_dim_t stride = 1;
-                for (int d = ndims - 2; d >= 0; d--) {
-                    strides[d] = stride * dims[d + 1];
-                    stride = strides[d];
+
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+                if (ndims == 3 && is_grouped) {
+                    // For 3D grouped weights, use plain abc format (no transpose)
```
dzarukin, line 1161:
```suggestion
                    // For 3D grouped weights, use plain abc format (no transpose)
                    // TODO: replace with `acb` once the transition to common ref will happen.
```

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
mzhukova: are you talking about https://github.com/uxlfoundation/oneDNN/pull/4714/changes/b4631d81353fb8dd9b81088e00f37d79eea92d99 ?

dzarukin: Yes, resolving.

tests/benchdnn/matmul/bench_matmul.cpp
mzhukova: I agree that we could drop most of the checks, but this particular one is also because benchdnn currently can't support `variable_dim_idx != 0`, so probably leave it be since it is testing limitation too?

tests/benchdnn/utils/compare.cpp
mzhukova: yep, that's been updated already!

tests/benchdnn/matmul/matmul.cpp
mzhukova: shouldn't clang-format handle this for comments too?

tests/benchdnn/doc/driver_matmul.md
mzhukova: updated wording

tests/benchdnn/utils/parser.cpp
mzhukova: sure, I've added a check into verify_input

tests/benchdnn/matmul/bench_matmul.cpp
dzarukin: If it fails or asserts before getting to the library, it's better to have it, sure.

tests/benchdnn/dnn_types.cpp
mzhukova: yep, verbose will be a separate story, I'll put a jira ticket here for the next steps

tests/benchdnn/matmul/bench_matmul.cpp
mzhukova: I've removed a bunch of checks that should be verified by the library, thanks for the suggestion!

tests/benchdnn/utils/parser.cpp
mzhukova: done! I'll add it to the list for the future activities :)

tests/benchdnn/utils/parser.cpp
mzhukova: updated, thanks for pointing this out, I was not aware of this functionality

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: done

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/matmul.cpp
dzarukin: We have that setting disabled. I can look into that :)

tests/benchdnn/utils/parser.cpp
mzhukova: sure, I've updated the code

tests/benchdnn/matmul/matmul.cpp
mzhukova: yes, that's a good catch, thanks! (this is an artifact that should have been updated)

tests/benchdnn/matmul/matmul.cpp
mzhukova: fixed

tests/benchdnn/matmul/matmul.cpp
mzhukova: It doesn't seem that way, I'm getting `DNNL_ARG_DIFF_SRC` for `SRC`

tests/benchdnn/matmul/matmul.cpp
mzhukova: so since this is create and supposed to return pointer, I think I will change assert to regular check and returning nullptr instead, `SAFE` won't work for it seems like (no status returning here)

tests/benchdnn/matmul/matmul.cpp
mzhukova: the idea here was to keep it simple (explicit `2` for 2D and `0` for variable index, since this is the only configuration supported, but I'm not attached, so could change to query instead

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/matmul.cpp
mzhukova: I understand the issue, I will extract the common part and re-use it, thanks!

tests/benchdnn/utils/parser.cpp
mzhukova: I feel like this make sense here as well, as it validates that the number of provided grouped matches number of provided sizes, and it is more about input validity vs operation (matmul). OK if I leave it?

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/doc/driver_matmul.md
mzhukova: I've added into [MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704)

tests/benchdnn/dnn_types.cpp
mzhukova: I've added into [MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704)

tests/benchdnn/utils/parser.cpp
mzhukova: I've added into [MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704)

tests/benchdnn/matmul/matmul.cpp
mzhukova: this was done too :)

tests/benchdnn/utils/parser.cpp
dzarukin: I meant to keep this one here as it is indeed an input validation and drop [this one](https://github.com/uxlfoundation/oneDNN/pull/4714/changes#diff-64a19b9dc29173212b979a6d7efbd45addbfe062ef9117ae0a6922e7337476e3R154-R163) as it repeats this one. It's fine to drop it later.

mzhukova:
@dzarukin and @avmanerikar please take another look (FYI linters/CI failures are not cause by the changes in this PR)

vpirogov:
> @dzarukin and @avmanerikar please take another look (FYI linters/CI failures are not cause by the changes in this PR)

I took care of linter false positive.

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
```diff
@@ -0,0 +1,127 @@
...
+# WOQ: f16 activation + u8 weights with per-expert per-column scales and ZPs
+--reset
+--dt=f16:u8:f16
+--wtag=abc,acb
+--grouped=0:4:8,8,8,8
+--attr-fpmath=f16:true
+--attr-scales=wei:5:f16
+--attr-zero-points=wei:5:u8
+32x64:4x64x32
+
```
avmanerikar, line 127: (nit): minor suggestion to add a case for `dim_idx=1` to cover all grounds of usage.

tests/benchdnn/matmul/matmul.cpp
dzarukin: Alright, need to update the interface, thanks for bringing this up.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -34,6 +34,43 @@
...
+// Notes:
+// - Currently supports only M dimension,
+//   therefore only SRC and DST can be created with grouped encoding
+// - Input validation is done in verify_grouped_input()
+static benchdnn_dnnl_wrapper_t<dnnl_memory_desc_t> create_grouped_md(
+        const prb_t *prb, data_kind_t kind, dnnl_data_type_t dt) {
+    int arg = (kind == SRC) ? DNNL_ARG_SRC
+            : (kind == DST) ? DNNL_ARG_DST
+                            : DNNL_ARG_UNDEF;
+    if (arg == DNNL_ARG_UNDEF) return nullptr;
```
dzarukin: I'd recommend to replace `nullptr` with `dnnl_memory_desc_t md {};` and returning this `md` throughout this function. I'm not sure benchdnn is ready to get `nullptr` instead of zero_md...

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -354,53 +423,9 @@ int fill_sparse_data(data_kind_t kind, const prb_t *prb, dnn_mem_t &mem_dt,
...
-        return fill_random_real(mem_dt, mem_fp, res);
-    }
-    if (has_bench_mode_bit(mode_bit_t::perf) && !is_any_sparse) {
-        return fill_random_real(
-                mem_dt, mem_fp, res, get_perf_fill_cfg(mem_dt.dt()));
-    }
-
+// Filling for mem_fp that takes into account density and zero points
+static void fill_dense_fp_values(data_kind_t kind, const prb_t *prb,
+        const cfg_t &cfg, int64_t nelems, dnn_mem_t &mem_fp) {
```
dzarukin: Minor: I'd drop `nelems` and query it: `mem_fp.nelems()` to reduce the number of arguments.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -453,7 +478,158 @@ int fill_data(data_kind_t kind, int exec_arg, const prb_t *prb,
...
+    if (nhandles != 2) {
+        BENCHDNN_PRINT(0, "Error: grouped memory requires 2 handles, got %d\n",
+                nhandles);
+        return FAIL;
+    }
+
+    // Fill offsets buffer
+    SAFE(fill_grouped_offsets(mem_dt, prb), WARN);
+
+    const int64_t nelems = prb->m * prb->k;
```
dzarukin: Wouldn't `mem_dt.nelems()` interface work here?

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -453,7 +478,158 @@ int fill_data(data_kind_t kind, int exec_arg, const prb_t *prb,
...
+    fill_dense_fp_values(kind, prb, cfg, nelems, mem_fp);
+
+    // Copy values from fp memory into grouped values handle,
+    // as grouped memory doesn't support reorder
+    benchdnn_parallel_nd(nelems, [&](int64_t i) {
+        mem_dt.set_elem(i,
+                round_to_nearest_representable(
+                        cfg.get_dt(kind), mem_fp.get_f32_elem(i)),
+                sparse_options_t::grouped_values_idx);
+    });
```
dzarukin: My impression was `mem_dt.reorder(mem_fp)` should exactly handle this situation.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -899,23 +1107,53 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
...
+            dnnl_dims_t dims = {prb->m, prb->k};
+            ref_mem_map.emplace(exec_arg,
+                    dnn_mem_t(2, dims, dnnl_f32, tag::abx, ref_engine,
+                            /* prefill = */ false));
+        } else if (is_grouped && exec_arg == DNNL_ARG_DST) {
+            // Plain 2D f32 reference buffer
+            dnnl_dims_t dims = {prb->m, prb->n};
+            ref_mem_map.emplace(exec_arg,
+                    dnn_mem_t(2, dims, dnnl_f32, tag::abx, ref_engine,
+                            /* prefill = */ false));
```
dzarukin: It seems this handling is not needed as it's taken care of by `create_md` and existing logic for sparse should just work.

tests/benchdnn/matmul/matmul.cpp
mzhukova: sure, I could do that
I only used `nullptr` because that's what `create_md` returns with assert or if no cases hit

tests/benchdnn/matmul/matmul.cpp
mzhukova: sure, fixed

tests/benchdnn/matmul/matmul.cpp
mzhukova: it will :) will address

tests/benchdnn/matmul/matmul.cpp
mzhukova: not right now, but we could make it work, sure :)

tests/benchdnn/matmul/matmul.cpp
dzarukin: If you could update that as well, it would be nice. Thanks.

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
mzhukova: This would be added once supported via benchdnn and the library, for now this parameter is basically not allowed :)

tests/benchdnn/matmul/matmul.cpp
mzhukova: fixed now

tests/benchdnn/matmul/matmul.cpp
mzhukova: fixed now

tests/benchdnn/matmul/matmul.cpp
mzhukova: So this is to create reference plain 2D buffers, if I let it fall through sparse logic (and `create_md`) we would end up in `create_grouped_md` I think, and therefore get grouped memory vs plain 2D

tests/benchdnn/matmul/matmul.cpp
mzhukova: updated https://github.com/uxlfoundation/oneDNN/pull/4714/changes/fea9b46e41b4449ec9dbe89941add3fb29dc87ec

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/matmul.cpp
mzhukova: Ok, so re-routing to generic branch (not sparse), seem to produce the same result, so I've removed those two :) Thanks for the suggestion (and explaining)! I'll wait for broader testing to ensure everything is still as expected

PR #4632, opened 2026-02-03
x64: matmul: Fix matrix A segfault issue

tczeszun:
The code incorrectly merged batch dimensions into M for 4D tensors and applied 3D-specific stride adjustments. The `A_strides` array only stores the last 3 dimensions, causing incorrect memory offset calculations when the stride adjustment logic `A_strides[1] = A_strides[2]` was applied to 4D cases.

Fixes [MFDNN-14493](https://jira.devtools.intel.com/browse/MFDNN-14493)

PR #4716, opened 2026-02-23
cpu: aarch64: matmul: fix segfault when merging batch dims for acbd

Sqvid:
# Description

Port of 326678af (PR #4632) to cpu/aarch64
Resolves: Issue #4396

# Checklist

## General

### Bug fixes

### DIFF
--- a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2023-2024 FUJITSU LIMITED
-* Copyright 2024-2025 Arm Ltd. and affiliates
+* Copyright 2024-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1035,8 +1039,10 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
 
     // We need to correct A_strides if batched dimensions are merged in M and
     // A layout is formally transposed but could be treated as plain
+    // For 4D tensors, only apply adjustment for treat_A_as_plain, not for acbd tag
```
jondea, line 1042: What's special about 4? If the crux is that we have >1 batch dim in this case, you might want to say that explicitly, I realise that they may be logically equivalent, but only because we only support 4 dims (I think, correct me if I'm wrong)

Sqvid: I think this really is `acbd` related (and therefore 4D-specific) for this implementation. We do support `ndims > 4` however, only 4D tensors can have tags that are not simple transposes of the non-batch dimensions. That is, for all other dimensions you may only transpose the innermost dims (i.e, `abc -> acb`, `abcde -> abced`), but for 4D you get an additional two supported cases (`acbd` and `adbc`). I admit I have not looked into exactly _why_ these two tags are singled out but the consequence is we need tag-specific code for these two cases everywhere.
See:
https://github.com/uxlfoundation/oneDNN/blob/e6b70c82acee931b42f087de3fca7eebcd8a053a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp#L172-L177

And the 4D special-cases here:
https://github.com/uxlfoundation/oneDNN/blob/e6b70c82acee931b42f087de3fca7eebcd8a053a/src/cpu/aarch64/matmul/brgemm_matmul_utils.cpp#L283-L286

@tczeszun or @densamoilov may be able to provide a deeper explanation

jondea: Okay, brill, thanks

jondea (approved):

openmodelzooci:

PR #4720, opened 2026-02-24
xe: gemm: add Xe2 HHS TNN strategy for gemv (n=1)

skazakov1:
PR fixes regression reported by https://jira.devtools.intel.com/browse/MFDNN-14613.


skazakov1:
make test perf-gpu
set primitive=gpu:gemm

skazakov1:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/selector/db/kernel.db
```diff
@@ -1206,6 +1206,7 @@ auto _CATALOG_ = kcatalog::toArray({
...
 {{'G', "gemm", {"H", "H", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {-1, 128, -1}, {16, 16, 1}, "ABI"}, "at16x2+m32@48 am32+m16@64 aB wg 4x2x4 kr xaf st vav hi pt sr br sb64 bk0 sm sn grf256 sys kv afb", {16, (LoopType) 255, 256, {(LoopType) 208, (LoopType) 255, (LoopType) 2}, {262144, 262144, 16777216}, {262144, 262144, 32}, {16, 16, 32}, {4, 2, 4}, 1, (WGType) 1, 445, 0, 8192, {16, 16, 4}, {true, true, true}}, {'E', 17, {241719, 28050.5, 9262.5, 2137.5, 552672, 876888, 1.91538, 1.82334, 3.84174, 8.3282, 0.015781, 0.00988075, 0.00545694, 0.959871, 1.33252, 0.965141, 5.50087e-12}}},
+{{'G', "gemm", {"H", "H", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, 1, 3968}, {-1, 1, 4096}, {-1, 1, 3968}, {-1, 1, 4096}, {16, 16, 1}, "ABIH"}, "am32+C32@64 at32 aS wg 1x1x16 ikr af vav sr sb256 bk0 bm0 sys rr", {16, (LoopType) 255, 128, {(LoopType) 0, (LoopType) 1, (LoopType) 2}, {16777216, 8192, 16777216}, {8192, 8192, 16777216}, {4, 16, 32}, {1, 1, 16}, 1, (WGType) 1, 268439813, 0, 256, {16, 16, 4}, {true, true, true}}, {'W', 1, {64}}},
```
Simonsays095, line 1209: Have you tried this kernel on N=1-16? Since the C tile is `4x16`, it should have the same performance on N=16 as it does on N=1 (maybe even better), so the "allow" condition might be able to be loosened a bit. If so, you'd probably have to refit with `--model E` so it can use the performance model to select against other kernels (outside of the "accept" region).

Simonsays095 (approved):

src/gpu/intel/gemm/jit/selector/db/kernel.db
skazakov1: Thanks for the suggestion. That is just fix of regression, that may be overridden if I find better performance.

PR #4726, opened 2026-02-25
benchdnn: introduce an integer input for scales/zero-points/precompute-reductions/prelu

dzarukin:
https://jira.devtools.intel.com/browse/MFDNN-14115

There are a bunch of minor fixes on top:
* CPU brgemm_matmul was optimistic about what it supported. Reduced the supported scope.
* verbose_converter always printed quant options even with empty content. I decided to stop it.

Rest modifications/unifications/cleanups are done to gracefully work with either input mask or policy.

### DIFF
--- a/tests/benchdnn/dnn_types.cpp
+++ b/tests/benchdnn/dnn_types.cpp
@@ -185,24 +185,24 @@ dnnl_rounding_mode_t str2rounding_mode(const std::string &str) {
     return dnnl_rounding_mode_environment;
 }
 
-int attr_t::get_default_mask(policy_t policy, int ndims) {
+static int get_default_mask(policy_t policy, int ndims) {
     switch (policy) {
-        case PER_DIM_0: return (1 << 0);
-        case PER_OC:
-        case PER_DIM_1: return (1 << 1);
-        case PER_OCIC:
-        case PER_DIM_01: return (1 << 0) + (1 << 1);
-        case PER_DIM_2: return (1 << 2);
-        case PER_DIM_3: return (1 << 3);
-        case MX:
-        case DYNAMIC_FP:
-        case PER_TENSOR:
+        case attr_t::policy_t::PER_DIM_0: return (1 << 0);
+        case attr_t::policy_t::PER_OC:
+        case attr_t::policy_t::PER_DIM_1: return (1 << 1);
+        case attr_t::policy_t::PER_OCIC:
+        case attr_t::policy_t::PER_DIM_01: return (1 << 0) + (1 << 1);
+        case attr_t::policy_t::PER_DIM_2: return (1 << 2);
+        case attr_t::policy_t::PER_DIM_3: return (1 << 3);
+        case attr_t::policy_t::MX:
+        case attr_t::policy_t::DYNAMIC_FP:
+        case attr_t::policy_t::PER_TENSOR:
             assert(ndims > 0 && ndims <= DNNL_MAX_NDIMS);
             return (1 << ndims) - 1;
-        case COMMON: return 0;
-        case HOST_SCALAR:
-            return 0; // mask=0 is required for compatibility with preprocessing logic
-        default: SAFE(FAIL, CRIT); return 0;
+        case attr_t::policy_t::COMMON: return 0;
+        // Note: use mask=0 for compatibility with reference compute paths.
+        case attr_t::policy_t::HOST_SCALAR: return 0;
+        default: SAFE(FAIL, CRIT); return INT_MIN;
     }
 }
 
@@ -218,28 +218,28 @@ int attr_t::policy2mask(int arg, policy_t policy, int ndims,
         dnnl_primitive_kind_t prim_kind, bool has_groups) {
 
     if (policy == policy_t::HOST_SCALAR) { // shortcut
-        return attr_t::get_default_mask(policy, ndims);
+        return get_default_mask(policy, ndims);
     }
 
     // Handle of weights mask for various primitives.
     if (prim_kind == dnnl_convolution || prim_kind == dnnl_deconvolution
             || prim_kind == dnnl_inner_product) {
         if (arg != DNNL_ARG_WEIGHTS || policy == policy_t::COMMON)
-            return attr_t::get_default_mask(policy, ndims);
+            return get_default_mask(policy, ndims);
 
         switch (policy) {
             case PER_OC:
                 if (has_groups)
-                    return attr_t::get_default_mask(PER_DIM_01, ndims);
+                    return get_default_mask(PER_DIM_01, ndims);
                 else
-                    return attr_t::get_default_mask(PER_DIM_0, ndims);
+                    return get_default_mask(PER_DIM_0, ndims);
             default: SAFE(FAIL, CRIT); return -1;
         }
     } else if (prim_kind == dnnl_matmul) {
         if ((arg != DNNL_ARG_SRC && arg != DNNL_ARG_WEIGHTS
                     && arg != DNNL_ARG_DST)
                 || policy == policy_t::COMMON)
-            return attr_t::get_default_mask(policy, ndims);
+            return get_default_mask(policy, ndims);
 
         if (ndims < 2) SAFE_V(FAIL);
         switch (policy) {
@@ -248,12 +248,12 @@ int attr_t::policy2mask(int arg, policy_t policy, int ndims,
             case PER_OCIC: return (1 << (ndims - 1)) + (1 << (ndims - 2));
             case MX:
             case DYNAMIC_FP:
-            case PER_TENSOR: return attr_t::get_default_mask(policy, ndims);
+            case PER_TENSOR: return get_default_mask(policy, ndims);
             default: SAFE_V(FAIL); return -1;
         }
     } else if (prim_kind == dnnl_layer_normalization) {
         if (arg != DNNL_ARG_SRC_1 || policy != policy_t::PER_OC)
-            return attr_t::get_default_mask(policy, ndims);
+            return get_default_mask(policy, ndims);
 
         // PER_OC
         assert(policy == policy_t::PER_OC);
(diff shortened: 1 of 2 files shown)


dzarukin:
make test

tests/benchdnn/utils/parser.cpp
```diff
@@ -69,6 +69,13 @@ void add_option_to_help(const std::string &option,
     help_added.push_back(option);
 }
 
+bool has_only_digits(const std::string &s) {
```
mzhukova, line 72: maybe also add check for an empty string? otherwise it seems `has_only_digits()` will be true

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -90,9 +90,13 @@ as host_scalars (`true`) or as device memory objects (`false`, default).
   - `dst` corresponds to `DNNL_ARG_DST`.
   - `msrci` corresponds to `DNNL_ARG_MULTIPLE_SRC + i`.
 
-`POLICY` specifies the way scale values will be applied to an `ARG` tensor. By
-default no scales are applied which corresponds to `common` policy and `1.f`
-scale value. Supported values are:
+`MASK_INPUT` specifies the way scale values will be applied to an `ARG` tensor.
+`MASK_INPUT` may be provided in two ways:
+* A plain integer value (which is provided by the verbose functionality) which
+  will be processed directly.
```
vpirogov: BTW do we have any documentation on how masks encoding works?

```suggestion
* An integer value encoding the mask. This is the way masks are reported in oneDNN verbose logs. 
```

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -90,9 +90,13 @@ as host_scalars (`true`) or as device memory objects (`false`, default).
...
+* A `POLICY` value (see below), that encodes an integer value in a lexical form.
+
+`POLICY` supported values are:
```
vpirogov: We probably don't need to introduce `POLICY` term at all

```suggestion
* One of pre-defined mask names:
```

scripts/verbose_converter/src/benchdnn_generator.py
```diff
@@ -65,18 +65,6 @@ def _get_alg(self):
     def _get_policies() -> Tuple[str, ...]:
         return "common", "per_oc"
```
atkassen: Not needed?

scripts/verbose_converter/src/benchdnn_generator.py
```diff
@@ -193,14 +181,15 @@ def _get_quantization(
...
+            mask = param.mask
+            policy = ""
             # Set policy to "host_scalar" if is_host_scalar is True
             if param.is_host_scalar:
                 policy = "host_scalar"
             if param.quantization_mode == "dynamic_mx":
                 policy = "mx"
-            result = f"{arg}:{policy}"
-            if policy == "common" or policy == "host_scalar":
+            result = f"{arg}:{policy}" if policy else f"{arg}:{mask}"
```
atkassen: nit:
```suggestion
            result = f"{arg}:{policy or mask}"
```

scripts/verbose_converter/src/benchdnn_generator.py
```diff
@@ -214,19 +203,19 @@ def _get_quantization(
     @property
     def scales(self):
         params = self._get_quantization(self.entry.exts.scales, 0.5, "f32")
-        return f"--attr-scales={params}"
+        return f"--attr-scales={params}" if params else f""
```
atkassen: `f""` -> `""`

tests/benchdnn/dnn_types.hpp
mzhukova: do we need to update those ctors too?
I think rn if something is constructed using one of those, the mask would still be `none` and therefore `is_def()` would also return true since the mask is the only thing being checked (vs previously checking for default policy/value/etc.)..?

mzhukova: (I was trying to understand more on this and noticed this sort of constructing should probably be used in self tests, which seem to be broken rn, idk for sure whether this is for the reason above or some other one, but just thought I'd mention)

scripts/verbose_converter/src/benchdnn_generator.py
atkassen: also random spot above this: do we need to handle `dynamic_fp`?

tests/benchdnn/doc/knobs_attr.md
dzarukin: The best description is probably provided at `doc/programming_model/attributes_quantization.md`.

Thanks for a suggestion. Applied.

tests/benchdnn/doc/knobs_attr.md
dzarukin: I removed any notices of policy across the document. Please check it out.

scripts/verbose_converter/src/benchdnn_generator.py
dzarukin: Shouldn't be, thanks for spotting, removed.

scripts/verbose_converter/src/benchdnn_generator.py
dzarukin: Applied, thanks for a suggestion.

`dynamic_fp` is indeed missing. Added, thanks for the catch.

scripts/verbose_converter/src/benchdnn_generator.py
dzarukin: Applied, thanks.

tests/benchdnn/utils/parser.cpp
dzarukin: Sure thing, thanks for a suggestion, applied.

tests/benchdnn/dnn_types.hpp
dzarukin: Your observation for ctors is correct, failures in self are indeed connected to those. I've updated ctors and also updated tests to set new expectations.

dzarukin:
make test

dzarukin:
make test

tests/benchdnn/dnn_types.cpp
```diff
@@ -340,24 +347,33 @@ int attr_t::arg_scales_t::entry_t::from_str(const std::string &s) {
...
-                if (this->groups.size() != 2) {
-                    BENCHDNN_PRINT(0, "%s\n",
-                            "Error: number of groups should be equal to number "
-                            "of dimension bits set in the mask.");
-                    SAFE_V(FAIL);
-                }
-                break;
-            default:
+        if (this->mask_input == attr_t::mask_input_t::mask) {
+            if (mask < 1) {
```
mzhukova: do we also need to check for consistency like group.size() against number of 1s in mask?

tests/benchdnn/dnn_types.hpp
```diff
@@ -204,40 +220,45 @@ struct attr_t {
...
-            policy_t policy = COMMON;
+            policy_t policy = POLICY_TOTAL;
             float scale = 1.f;
             dnnl_data_type_t dt = dnnl_f32;
             std::vector<dnnl_dim_t> groups;
+            int64_t mask = INT_MIN;
+            mask_input_t mask_input = mask_input_t::none;
         };
 
         void set(int arg, const entry_t &scale) { scales[arg] = scale; }
```
mzhukova, line 264: btw does this need is_def guard?

tests/benchdnn/dnn_types.hpp
```diff
@@ -78,34 +79,48 @@ struct attr_t {
...
-            bool is_def() const {
-                return policy == COMMON && value == 0 && dt == dnnl_s32
-                        && groups.empty();
-            }
+            bool is_def() const { return mask_input == mask_input_t::none; }
 
             bool is_host_scalar() const { return policy == HOST_SCALAR; }
+            bool has_single_element() const {
+                if (mask_input == mask_input_t::mask) return mask == 0;
+                assert(mask_input == mask_input_t::policy);
```
mzhukova, line 114: seems like this would assert for things constructed with default (since the mask will be `none`), is this intentional?

mzhukova (approved):

tests/benchdnn/dnn_types.cpp
dzarukin: This kind of verification would be passed to the library at the attribute creation moment or primitive_desc create moment and would be verified with verbose.

This particular mask branch of this check might actually fit the same category and can be probably dropped. I'll check it if it's safe to drop it.

tests/benchdnn/dnn_types.hpp
dzarukin: Since `scales` is a `std::map`, `operator[]` would insert a new element if there's none with such `arg` key.
It should be functionally working.

tests/benchdnn/dnn_types.hpp
dzarukin: I see your point, this could be an exit for `mask_input == none`, yes, but for faster debugging of incorrect usage I prefer assert given that this is a testing tool and not the library with the contract with the user, meaning we can afford that assert over returning a "correct" answer and then triaging the whole stack what went wrong searching for a spot where incorrect scale was called in case of a bug.

There's pending refactor of scales/zps/prds to avoid boilerplate code for parsing, processing and implementing identical interfaces, I haven't decided if I want to keep those ctors (as they are solely used in self testing). If they will stay, all interfaces will be improved to deal with all their states properly. So far I'd like to leave them as is just for the sake of time as long as it doesn't crash.

tests/benchdnn/dnn_types.cpp
mzhukova: I'm fine either way, it just seemed like we were checking for group.size previously

tests/benchdnn/dnn_types.hpp
mzhukova: I see, thanks for explanation :)

tests/benchdnn/dnn_types.cpp
dzarukin: This is correct, I adjusted the check to continue verifying it regardless of mask input. Thanks.

Issue #4638, opened 2026-02-03
RV64: Runtime CPU feature detection does not cover all RVV code in the built binary


kxxt:
# Summary

Runtime CPU feature detection does not cover all RVV code in the built binary, causing crashes on rv64gc platform.

# Version

3.10.2 (git tag v3.10.2)

# Environment
oneDNN includes hardware-specific optimizations and may behave
differently on depending on the compiler and build environment. Include
the following information to help reproduce the issue:
* CPU make and model: rv64gc (SG2042)
* OS version (`uname -a`): `Linux 6.6.46-pisces #2024.08.18.16.00+26c645731 SMP  riscv64 GNU/Linux`
* Compiler version (`gcc --version`): 15.2.1
* CMake version (`cmake --version`): 4.2.3
* CMake output log:

```
-- CMAKE_BUILD_TYPE is unset, defaulting to Release
-- The C compiler identification is GNU 15.2.1
-- The CXX compiler identification is GNU 15.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- DNNL_TARGET_ARCH: RV64
-- DNNL_LIBRARY_NAME: dnnl
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found OpenMP_C: -fopenmp (found version "4.5")
-- Found OpenMP_CXX: -fopenmp (found version "4.5")
-- Found OpenMP: TRUE (found version "4.5")
-- Performing Test CAN_COMPILE_RVV_INTRINSICS
-- Performing Test CAN_COMPILE_RVV_INTRINSICS - Success
-- Performing Test CAN_COMPILE_ZVFH_INTRINSICS
-- Performing Test CAN_COMPILE_ZVFH_INTRINSICS - Success
-- Can compile RVV Intrinsics: TRUE
-- Can compile Zvfh Intrinsics: TRUE
-- DNNL_RISCV_USE_RVV_INTRINSICS: TRUE
[...]

PR #4734, opened 2026-02-27
cpu: rv64: add JIT affine kernel optimization for softmax f32 with runtime ISA check

xiazhuozhao:
# Summary

According to #4638, this PR converts the f32 softmax (and logsoftmax) post-processing stage on RV64 to a pure JIT-based RVV affine kernel and introduces an explicit runtime ISA check.

The key goals are:

- Eliminate dependency on compile-time RVV intrinsics in the softmax affine stage
- Prevent potential crashes caused by mismatches between build-time and runtime ISA
- Align with the RISC-V team's direction toward full runtime dispatch
- Maintain overall performance parity

# Motivation

Previously, parts of the RV64 softmax implementation relied on RVV intrinsics compiled into the binary.

This can cause instability in scenarios where:

- The library is built on a system with RVV enabled
- But deployed on a system where the V extension is unavailable

In such cases, illegal instruction faults or segmentation faults may occur.

To address this, this PR:

1. Replaces intrinsic-based affine post-processing with a JIT-generated RVV kernel.
2. Adds a runtime ISA check:

```cpp
VDISPATCH_SOFTMAX(mayiuse(v), VERBOSE_UNSUPPORTED_ISA);
```

This ensures that RVV code paths are only used when the runtime environment actually supports the V extension.

# Changed

For f32 softmax on RV64:

- Introduced `jit_rvv_softmax_affine_kernel_t`
- Replaced intrinsic-based affine loops with:
  - `y = x - sub` (logsoftmax)
  - `y = x * mul` (softmax normalization)
- Kernel instantiated once per primitive when JIT is enabled
- Added explicit runtime ISA validation
- Scalar fallback retained for non-JIT builds

No functional behavior changes are intended.

# Performance Evaluation
Benchmark Batch:  
```
--reset

--inplace=true,false
--alg=SOFTMAX,LOGSOFTMAX
--dir=FWD_D
--sdt=f32
--ddt=f32

--stag=abx
--axis=0,1
--batch=set_0d
--axis=1,3
--batch=shapes_2d
--axis=1,4
--batch=shapes_3d

--stag=axb
--axis=1
--batch=shapes_2d
--axis=1
--batch=shapes_3d
```
(This is the test_softmax_all batch with only f32 cases retained.)

Benchmark result file: 
[bench_softmax.csv](https://github.com/user-attachments/files/25597590/bench_softmax.csv)  
Total cases: 368

Two implementations compared:

- Baseline: intrinsic-based implementation
- New: full JIT affine implementation

## Overall Throughput (Total Time Metric)

Total raw time:
    329.67

Total JIT time:
    291.61

Metric | Value | Meaning
-- | -- | --
Arithmetic Mean Speedup | 1.0084 | Appears almost unchanged
Geometric Mean Speedup | 0.9243 | A typical case is about 7.5% slower
Total Runtime Speedup | 1.1305 | Overall 13% faster
Median Speedup | 0.95 | Half of the cases regress
Percentage of Cases with Speedup < 1 | 58.4% | About half of the cases regress

# f16

An initial attempt was made to apply a similar JIT-based approach to the f16 path. However, preliminary benchmarking shows noticeable regression compared to the current implementation. A likely contributing factor is the additional data type conversion overhead in the f16 pipeline, which may offset the benefits of the JIT affine stage.
Further optimization for f16 remains future work.

### DIFF
--- a/src/cpu/rv64/jit_rvv_softmax_affine_kernel.cpp
+++ b/src/cpu/rv64/jit_rvv_softmax_affine_kernel.cpp
@@ -42,8 +42,6 @@ void jit_rvv_softmax_affine_kernel_t::generate() {
     const FReg f_mul = fa1;
 
     const VReg v_src(0);
-    const VReg v_sub(1);
-    const VReg v_mul(2);
 
     // call_params_t layout:
     //  0: src, 8: dst, 16: len, 24: sub, 28: mul
@@ -60,11 +58,9 @@ void jit_rvv_softmax_affine_kernel_t::generate() {
     L(loop);
     beqz(reg_len, done);
     vsetvli(reg_vl, reg_len, SEW::e32, LMUL::m1);
-    vfmv_v_f(v_sub, f_sub);
-    vfmv_v_f(v_mul, f_mul);
     vle32_v(v_src, reg_src);
-    vfsub_vv(v_src, v_src, v_sub);
-    vfmul_vv(v_src, v_src, v_mul);
+    vfsub_vf(v_src, v_src, f_sub);
+    vfmul_vf(v_src, v_src, f_mul);
     vse32_v(v_src, reg_dst);
     slli(reg_bytes, reg_vl, 2);
     add(reg_src, reg_src, reg_bytes);


src/cpu/rv64/jit_rvv_softmax_affine_kernel.cpp
```diff
@@ -0,0 +1,85 @@
...
+    fmv_w_x(f_sub, reg_tmp);
+    lw(reg_tmp, reg_param, 28);
+    fmv_w_x(f_mul, reg_tmp);
+
+    Label loop, done;
+    L(loop);
+    beqz(reg_len, done);
+    vsetvli(reg_vl, reg_len, SEW::e32, LMUL::m1);
+    vfmv_v_f(v_sub, f_sub);
+    vfmv_v_f(v_mul, f_mul);
```
zhangjian29: Suggesting move `vfmv` outside the loop, which may help solve performance regressions.

zhangfeiv0: Can vector-scalar (.vf) instructions be used to directly operate on floating-point registers, thereby eliminating the splat overhead?

```
vle32_v(v_src, reg_src);
vfsub_vf(v_src, v_src, f_sub);  
vfmul_vf(v_src, v_src, f_mul);  
```

openmodelzooci:

xiazhuozhao:
Thank you for the helpful suggestions!

Following the review comments, we have updated the implementation to use the vector–scalar (.vf) instructions (vfsub_vf / vfmul_vf) instead of the previous vv + splat approach. This removes the need for broadcasting the scalar values into vector registers and simplifies the loop body, which also reduces register pressure.

We have also re-run the performance benchmarks:
[bench_softmax_vf.csv](https://github.com/user-attachments/files/25791689/bench_softmax_vf.csv)

For the Benchmark Batch described in the PR (the same configuration previously listed), the updated implementation achieves an average speedup of 1.10× compared to the non-JIT version.

Please note that the “baseline” performance numbers in this new log differ slightly from those uploaded when the PR was first opened. This is due to variations in the runtime state of the test server. To provide the most representative comparison, we performed back-to-back runs of the two code versions on the same machine under the same conditions, and computed the speedup from those consecutive measurements.

PR #4738, opened 2026-03-02
graph: backend: dnnl: memory_planning: fix reorder inplace

TaoLv:
DNNL reorder primitive does not support inplace.


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4745, opened 2026-03-02
fix: common: memory_desc: add upper bounds check for md tensor dimensions

avmanerikar:
# Description

The PR adds out-of-bounds checks for tensor dimensions within `memory_desc_sanity_check()` and `size()` calculations to prevent potential overflow situations during memory creation.
`memory_desc_sanity_check()` has a lower bound check for tensor dimensions `(dims<0)` but not for the case when the product of tensor dimensions exceeds numeric bounds. 

Addresses [MFDNN-14665](https://jira.devtools.intel.com/browse/MFDNN-14665).

#### Test: 
```
DNNL_VERBOSE=all ./tests/benchdnn/benchdnn --matmul 4x4611686018427387905:4611686018427387905x4
```

#### Before fix:
```
onednn_verbose,v1,info,oneDNN v3.12.0 (commit 946487bb0f9b2c5a6f425b4d5c35b1d8d3fd5298)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
onednn_verbose,v1,info,cpu,isa:Intel AVX-512 with float16, Intel DL Boost and bfloat16 support and Intel AMX with bfloat16 and 8-bit integer support
onednn_verbose,v1,info,gpu,runtime:none
onednn_verbose,v1,info,graph,backend,0:dnnl_backend
onednn_verbose,v1,graph,info,template:operation,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,implementation,backend,exec_time
onednn_verbose,v1,primitive,create:dispatch,brgemm_matmul,datatype configuration not supported on this isa,src/cpu/x64/matmul/brgemm_matmul_utils.cpp:1369
onednn_verbose,v1,primitive,create:dispatch,brgemm_matmul,datatype configuration not supported on this isa,src/cpu/x64/matmul/brgemm_matmul_utils.cpp:1369
onednn_verbose,v1,primitive,create:dispatch,brgemm_matmul,datatype configuration not supported on this isa,src/cpu/x64/matmul/brgemm_matmul_utils.cpp:1369
onednn_verbose,v1,primitive,create:dispatch,brgemm_matmul,datatype configuration not supported on this isa,src/cpu/x64/matmul/brgemm_matmul_utils.cpp:1369
onednn_verbose,v1,primitive,create:cache_miss,cpu,matmul,brg_matmul:avx512_core,undef,src:f32:a:blocked:ab:4611686018427387905x1:f0 wei:f32:ap:blocked:AB16a64b::f0 dst:f32:a:blocked:ab::f0,,,4x4611686018427387905:4611686018427387905x4,109.795
onednn_verbose,v1,primitive,create:cache_hit,cpu,matmul,brg_matmul:avx512_core,undef,src:f32:a:blocked:ab:4611686018427387905x1:f0 wei:f32:ap:blocked:AB16a64b::f0 dst:f32:a:blocked:ab::f0,,,4x4611686018427387905:4611686018427387905x4,0.126953
AddressSanitizer:DEADLYSIGNAL
=================================================================
==613136==ERROR: AddressSanitizer: SEGV on unknown address 0x52700000c000 (pc 0x634320c247cf bp 0x7ffd91188b30 sp 0x7ffd91188b10 T0)
==613136==The signal is caused by a WRITE memory access.
```

#### After fix:
```
onednn_verbose,v1,info,oneDNN v3.12.0 (commit 946487bb0f9b2c5a6f425b4d5c35b1d8d3fd5298)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
onednn_verbose,v1,info,cpu,isa:Intel AVX-512 with float16, Intel DL Boost and bfloat16 support and Intel AMX with bfloat16 and 8-bit integer support
onednn_verbose,v1,info,gpu,runtime:none
onednn_verbose,v1,info,graph,backend,0:dnnl_backend
onednn_verbose,v1,graph,info,template:operation,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,implementation,backend,exec_time
Error: Function 'init_md' at (/nfs/site/proj/mkl/hal9000/amanerik/oneDNN/tests/benchdnn/dnnl_memory.cpp:678) returned 'invalid_arguments'
```

### DIFF
--- a/src/common/type_helpers.hpp
+++ b/src/common/type_helpers.hpp
@@ -1135,6 +1135,10 @@ inline bool memory_desc_strides_check(
     };
     std::sort(perm, perm + md.ndims, idx_sorter);
 
+    // tracks max stride for integral overflow checks
+    dim_t max_stride = 1;
+    int max_stride_d = 0;
+
     dim_t min_stride = block_size;
     for (int idx = 0; idx < md.ndims; ++idx) {
         const int d = perm[idx];
@@ -1154,6 +1158,22 @@ inline bool memory_desc_strides_check(
         // update min_stride for next iteration
         const auto padded_dim = md.padded_dims[d];
         min_stride = block_size * strides[d] * (padded_dim / blocks[d]);
+        if (max_stride <= strides[d]) {
+            max_stride = strides[d];
+            max_stride_d = d;
+        }
+    }
+
+    const size_t dt_size = types::data_type_size(md.data_type);
+
+    // guard against integral overflow due to strides exceeding numeric limits
+    if (!is_runtime_value(md.padded_dims[max_stride_d])) {
+        size_t dim_val = static_cast<size_t>(
+                md.padded_dims[max_stride_d] / blocks[max_stride_d]);
+        dim_val = dim_val == (size_t)max_stride ? 1 : dim_val;
+        if (dim_val > SIZE_MAX / max_stride) return false;
+        if (dt_size && ((dim_val * max_stride) > SIZE_MAX / dt_size))
+            return false;
     }
     return true;
 }
@@ -1331,9 +1351,22 @@ inline bool memory_desc_sanity_check(int ndims, const dims_t dims,
                     f8_e4m3, f16, bf16, f32, f64, s64, s32, s8, u8, s4, u4);
     if (!ok) return false;
 
+    // A bounds check on the dimensions ensures that the tensor size
+    // computation does not trigger a overflow during memory creation.
+    dim_t prod = 1;
+    for (int d = 0; d < ndims; ++d) {
+        if (dims[d] != DNNL_RUNTIME_DIM_VAL) {
+            if (dims[d] < 0) return false;
+            if (dims[d] > 0) {
+                if (prod > std::numeric_limits<dim_t>::max() / dims[d])
+                    return false;
+                prod *= dims[d];
+            }
+        }
+    }
+
     bool has_runtime_dims = false;
     for (int d = 0; d < ndims; ++d) {
-        if (!is_runtime_value(dims[d]) && dims[d] < 0) return false;
         if (is_runtime_value(dims[d])) has_runtime_dims = true;
     }
 
--- a/src/common/utils.hpp
+++ b/src/common/utils.hpp
@@ -25,6 +25,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
+#include <limits>
 #include <locale>
 #include <sstream>
 #include <string>
@@ -296,8 +297,11 @@ constexpr T array_product(const T *arr) {
 template <typename T, typename R = T>
 inline R array_product(const T *arr, size_t size) {
     R prod = 1;
-    for (size_t i = 0; i < size; ++i)
+    for (size_t i = 0; i < size; ++i) {
+        assert(IMPLICATION(arr[i] > 0 && prod > 0,
+                prod <= std::numeric_limits<R>::max() / arr[i]));
         prod *= arr[i];
+    }
     return prod;
 }
 
--- a/src/common/memory_desc_wrapper.cpp
+++ b/src/common/memory_desc_wrapper.cpp
@@ -63,6 +63,10 @@ status_t fill_blocked(memory_desc_t &md, std::initializer_list<int> perm,
                 ? runtime_value_for(md.padded_dims[d])
                 : utils::rnd_up(md.dims[d], blocks[d]);
 
+    // tracks max stride for integral overflow checks
+    dim_t max_stride = 1;
+    int max_stride_d = 0;
+
     // setting the strides
     {
         dim_t stride = block_size;
@@ -77,7 +81,24 @@ status_t fill_blocked(memory_desc_t &md, std::initializer_list<int> perm,
             else if (pdim != 0)
                 stride *= pdim / blocks[d];
 
+            if (max_stride <= stride) {
+                max_stride = stride;
+                max_stride_d = d;
+            }
+
         } while (iter_d != perm.begin());
+    }
+
+    const size_t dt_size = types::data_type_size(md.data_type);
+
+    // guard against integral overflow due to strides exceeding numeric limits
+    if (!is_runtime_value(md.padded_dims[max_stride_d])) {
+        size_t dim_val = static_cast<size_t>(
+                md.padded_dims[max_stride_d] / blocks[max_stride_d]);
+        dim_val = dim_val == (size_t)max_stride ? 1 : dim_val;
+        if (dim_val > SIZE_MAX / max_stride) return status::invalid_arguments;
+        if (dt_size && ((dim_val * max_stride) > SIZE_MAX / dt_size))
+            return status::invalid_arguments;
     }
 
     return status::success;


avmanerikar:
make test
set test_scope=NIGHTLY

src/common/utils.hpp
```diff
@@ -296,8 +296,10 @@ constexpr T array_product(const T *arr) {
 template <typename T, typename R = T>
 inline R array_product(const T *arr, size_t size) {
     R prod = 1;
-    for (size_t i = 0; i < size; ++i)
+    for (size_t i = 0; i < size; ++i) {
+        if (arr[i] != 0) assert(prod <= std::numeric_limits<R>::max() / arr[i]);
```
dzarukin: Otherwise, this function wouldn't return an expected value for `arr[i] = 0`.
```suggestion
        if (arr[i] != 0) { assert(prod <= std::numeric_limits<R>::max() / arr[i]); }
```

src/common/type_helpers.hpp
```diff
@@ -1344,6 +1344,16 @@ inline bool memory_desc_sanity_check(int ndims, const dims_t dims,
...
 
+    // A bounds check on the dimensions ensures that the tensor size
+    // computation does not trigger a overflow during memory creation.
+    dim_t prod = 1;
+    for (int d = 0; d < ndims; ++d) {
+        if (dims[d] != DNNL_RUNTIME_DIM_VAL && dims[d] > 0) {
+            if (prod > INT64_MAX / dims[d]) return false;
+            prod *= dims[d];
+        }
+    }
```
dzarukin, line 1366: Minor: maybe fuse into a loop below?
```
for (int d = 0; d < ndims; ++d) {
    if (dims[d] != DNNL_RUNTIME_DIM_VAL) {
        if (dims[d] < 0) return false;
        if (dims[d] > 0) {
            ...
        }
    }
}
```

src/common/memory_desc_wrapper.hpp
```diff
@@ -337,8 +337,13 @@ struct memory_desc_wrapper : public c_compatible {
 
             // `div_up` guarantees a spot in memory for odd number of half-byte
             // elements. Crucial case is `1` when simple division returns 0.
-            size_t data_size = utils::div_up(max_size * data_type_size(),
-                    sub_byte_data_type_multiplier());
+            size_t prod = max_size * data_type_size();
+            // guard against integer overflow due to size() exceeding numeric bounds.
+            if (max_size != 0 && prod / max_size != data_type_size())
+                return SIZE_MAX;
```
dzarukin: I'm actually not sure whether this is needed. Because whatever number we report here, it would be just wrong.
Validating dims should be good enough.

avmanerikar: Even when the tensor dims are validated, the returned `size()` value can overflow depending on `data_type_size()` because of: 
```
size_t prod = max_size * data_type_size();
```
A reproducer for this scenario is provided in the ticket.

src/common/type_helpers.hpp
avmanerikar: Applied. Thanks!

src/common/memory_desc_wrapper.hpp
dzarukin: Maybe then verify if this could happen in sanity check as well?

avmanerikar: This check is specific to a blocking descriptor. Computation of `max_size` requires knowledge of the number of blocks, inner blocks and block parameters that must be passed to `memory_desc_sanity_check` arguments for the operation.

dzarukin: That's fair.

It's definitely possible to check prod * data_type_size in existing sanity_check. That would provide some guidance for format_kind::any or undef.

Additionally, a new sanity check for strides is needed which should also check the product of `biggest stride * dim_val (correspondent to this stride) * data_type_size`. And call this function in `memory_desc_init_by_strides` and `memory_desc_init_by_tag` that are associated with public API to create memory descriptors.

You might also want to check if everything is good with rest public ctors of memory desriptors (such as sparse, host_scalar, etc.)

My point is `size()` shouldn't do any checks, it should provide the size. If it provides the size that is "incorrect" in any meaning, that memory descriptor should have been rejected at earlier point.

avmanerikar: > You might also want to check if everything is good with rest public ctors of memory desriptors (such as sparse, host_scalar, etc.)

All the public ctors for the memory descriptors are safeguarded by `dnnl::impl::memory_desc_sanity_check()` so including a bounds check within it will work for all.

> My point is `size()` shouldn't do any checks, it should provide the size. If it provides the size that is "incorrect" in any meaning, that memory descriptor should have been rejected at earlier point.
> Additionally, a new sanity check for strides is needed which should also check the product of biggest stride * dim_val (correspondent to this stride) * data_type_size. And call this function in memory_desc_init_by_strides and memory_desc_init_by_tag that are associated with public API to create memory descriptors.

That makes sense.  It looks like the `dnnl::impl::memory_desc_sanity_check()` will need to be expanded to account for each scenario.

dzarukin: sanity_check operates with basic inputs of md while there are format_kind specific ones which may not make sense for all scenarios and all calls, that's why I referred to a dedicated check. But feel free to implement it the way it seems best, I was assuming shaking existing one might not be the best choice considering the number of sptos calling this routine.

src/common/type_helpers.hpp
```diff
@@ -1344,6 +1344,19 @@ inline bool memory_desc_sanity_check(int ndims, const dims_t dims,
...
+            if (dims[d] > 0) {
+                if (prod > INT64_MAX / dims[d]) return false;
+                prod *= dims[d];
+            }
+        }
+    }
+
     bool has_runtime_dims = false;
     for (int d = 0; d < ndims; ++d) {
         if (dims[d] != DNNL_RUNTIME_DIM_VAL && dims[d] < 0) return false;
```
mzhukova: it seems like `dims[d] < 0` is no longer needed here as it is checked above?

src/common/utils.hpp
```diff
@@ -296,8 +296,12 @@ constexpr T array_product(const T *arr) {
 template <typename T, typename R = T>
 inline R array_product(const T *arr, size_t size) {
     R prod = 1;
-    for (size_t i = 0; i < size; ++i)
+    for (size_t i = 0; i < size; ++i) {
+        if (arr[i] != 0) {
+            assert(prod <= std::numeric_limits<R>::max() / arr[i]);
```
mzhukova: need to include `<limits>` probably? (didn't notice this one in the list)

src/common/memory_desc_wrapper.hpp
```diff
@@ -337,8 +337,14 @@ struct memory_desc_wrapper : public c_compatible {
...
+
+            size_t data_size
+                    = utils::div_up(prod, sub_byte_data_type_multiplier());
+
             if (is_additional_buffer()) {
```
mzhukova: it seems like below there are a few more places where the size is increased, should we introduce some sort of guarding mechanism there too?

src/common/memory_desc_wrapper.hpp
avmanerikar: Moved sanity check from `size()` to `memory_desc_init_by_tag()`. Thanks!

src/common/type_helpers.hpp
```diff
@@ -1344,6 +1344,19 @@ inline bool memory_desc_sanity_check(int ndims, const dims_t dims,
...
     if (!ok) return false;
 
+    // A bounds check on the dimensions ensures that the tensor size
+    // computation does not trigger a overflow during memory creation.
+    dim_t prod = 1;
+    for (int d = 0; d < ndims; ++d) {
+        if (dims[d] != DNNL_RUNTIME_DIM_VAL) {
+            if (dims[d] < 0) return false;
+            if (dims[d] > 0) {
+                if (prod > INT64_MAX / dims[d]) return false;
```
mzhukova: just wondering why `INT64_MAX` here and not `std::numeric_limits<dim_t>::max`?

src/common/memory_desc.cpp
mzhukova, line 84: do we need similar check for init_by_strides ..?

src/common/memory_desc.cpp
```diff
@@ -71,6 +71,29 @@ status_t memory_desc_init_by_tag(memory_desc_t &memory_desc, int ndims,
         // nop
     } else if (format_kind == format_kind::blocked) {
         status = memory_desc_wrapper::compute_blocking(md, tag);
+        if (status != success) return status;
+
+        // guard against integer overflow due to size() exceeding numeric bounds.
+        const memory_desc_wrapper mdw(md);
```
mzhukova: could this be moved to some common function and re-used for `size()`?

dzarukin: `size()` shouldn't do any checks, as it supposed to operate over a valid md.

But I second the consolidation to re-use it for strides.

src/common/memory_desc.cpp
```diff
@@ -71,6 +71,29 @@ status_t memory_desc_init_by_tag(memory_desc_t &memory_desc, int ndims,
...
+        const auto &bd = mdw.blocking_desc();
+        const size_t dt = mdw.data_type_size();
+
+        dims_t blocks = {0};
+        mdw.compute_blocks(blocks);
```
dzarukin: Does this one have anything in common with `memory_desc_wrapper::compute_blocking`? If yes, why double calling it? The problem with extra compute block is that it will pop up as an extra overhead for all operations and the more light-weighted checks we can end up with, the better.

src/common/utils.hpp
```diff
@@ -296,8 +296,15 @@ constexpr T array_product(const T *arr) {
...
 inline R array_product(const T *arr, size_t size) {
     R prod = 1;
-    for (size_t i = 0; i < size; ++i)
+    for (size_t i = 0; i < size; ++i) {
+        if (arr[i] != 0) {
+            auto max_lim = std::numeric_limits<R>::max() / arr[i];
+            auto min_lim = std::numeric_limits<R>::min() / arr[i];
+            assert(prod <= std::max(max_lim, min_lim)
+                    && prod >= std::min(max_lim, min_lim));
+        }
```
dzarukin: Please move all the conditions under assert, otherwise, it would hurt performance.

src/common/memory_desc.cpp
```diff
@@ -71,6 +71,29 @@ status_t memory_desc_init_by_tag(memory_desc_t &memory_desc, int ndims,
...
+            dim_t spd = mdw.padded_dims()[d] / blocks[d];
+            dim_t estr = spd == 1 ? 1 : bd.strides[d];
+            max_size = nstl::max<size_t>(max_size, spd * estr);
+        }
+
+        if (max_size == 1 && bd.inner_nblks != 0) {
+            max_size = static_cast<size_t>(mdw.blk_size());
+        }
+
+        if (dt && max_size > SIZE_MAX / dt) return invalid_arguments;
```
dzarukin: This sanity check can be done with less overhead - check the max_stride * dim_val * dt_size for overflow, it should be enough.

src/common/memory_desc.cpp
mzhukova: Oh, I didn't mean check, but the computations for `max_size` seem to be the same as in `size()` and here

src/common/type_helpers.hpp
avmanerikar: The two are equivalent from `dim_t` definition: https://github.com/uxlfoundation/oneDNN/blob/75a92fb3847fc4da950aa3a81cc3c21fdede61fa/include/oneapi/dnnl/dnnl_common_types.h#L126

src/common/type_helpers.hpp
avmanerikar: Makes sense. Applied!

src/common/utils.hpp
avmanerikar: Applied. Thanks!

src/common/memory_desc.cpp
avmanerikar: Instead of adding a new function, the safeguard has been integrated into existing checks for blocking and strides during memory_desc initialization - this avoids repeated computations which are shared by these checks.
- For `memory_desc_init_by_tag()`, the check is now a part of `memory_desc_wrapper::fill_blocked()` called within `memory_desc_wrapper::compute_blocking()`.
- For `memory_desc_init_by_strides()`, it is added to `memory_desc_strides_check()`.

src/common/memory_desc.cpp
avmanerikar: The check is now moved to `memory_desc_wrapper::fill_blocked` which shares the block computations - this executes the check without repeated overheads.

src/common/memory_desc.cpp
avmanerikar: Done. Thanks!

src/common/memory_desc.cpp
avmanerikar: Yes - this has been added too in the latest update

src/common/utils.hpp
avmanerikar: Artifact of rebasing - done.

mzhukova: umm sorry I'm not following, what changed..?

src/common/type_helpers.hpp
mzhukova: right, I guess I was mostly referring to consistency, you have `prod` as `dim_t`, so maybe either `int64_t` everywhere or `dim_t` with `std::numeric_limits`..?

avmanerikar: Done - Thanks!

src/common/utils.hpp
avmanerikar: Resolving - discussed offline and added the header.

avmanerikar:
make test
set test_scope=NIGHTLY

src/common/utils.hpp
```diff
@@ -296,8 +297,16 @@ constexpr T array_product(const T *arr) {
...
     R prod = 1;
-    for (size_t i = 0; i < size; ++i)
+    for (size_t i = 0; i < size; ++i) {
+        if (arr[i] > 0) {
+            assert(prod <= std::numeric_limits<R>::max() / arr[i]
+                    && prod >= std::numeric_limits<R>::min() / arr[i]);
+        } else if (arr[i] < 0) {
+            assert(prod <= std::numeric_limits<R>::min() / arr[i]
+                    && prod >= std::numeric_limits<R>::max() / arr[i]);
+        }
```
dzarukin: Could you please, share, when it is expected to see negative values in the array?

src/common/type_helpers.hpp
```diff
@@ -1154,6 +1157,19 @@ inline bool memory_desc_strides_check(
         // update min_stride for next iteration
         const auto padded_dim = md.padded_dims[d];
         min_stride = block_size * strides[d] * (padded_dim / blocks[d]);
+        max_stride = nstl::max(max_stride, strides[d]);
```
dzarukin: You may track for `d`, then just checking for `max_stride * dims[d] * dt_size` for overflow instead of the whole for loop below.

avmanerikar: That will be less conservative but still works - applied.

src/common/utils.hpp
avmanerikar: That is a safeguard against assuming only positive values for the templated arguments. This happens when `array_product()` acts on uninitialized `dims()` values and breaking code. I can skip the assert entirely for negative values if preferred for efficiency.

dzarukin: My concert is extra `if` for release mode for nothing. I'd like it to be inside assert. With that, current version becomes very obscure, thus, the question, why we want to legitimize negative values while the product doesn't assume operating over them.

I'd suggest to modify it like this:
```
    assert(arr[i] >= 0 && IMPLICATION(arr[i] > 0, prod_checks_from_PR_for_positive_arr_i));
```

This way debug mode would be the only suffering spot. And then the most intriguing part to make sure it won't explode.

avmanerikar: Done. The check is limited to only positive values for arrays - even if negative values are encountered by `array_product()` for an uninitialized `dims()`, it is safe to assume that it will be subsequently caught by `memory_desc_sanity_check` during md initialization.

avmanerikar:
make test
set test_scope=NIGHTLY

avmanerikar:
> Is it possible to add a small test case for gtests where we could hit all of the new checks?

Discussed offline - this will be added as part of a separate PR addressing a broader set of `dims()` checks.

dzarukin (approved):

PR #2351, opened 2025-01-07
xe: ocl: gemm: fix post-op typing for gemm_with_po

rjoursler:
Fixes [MFDNN-12893](https://jira.devtools.intel.com/browse/MFDNN-12893).

PR #4746, opened 2026-03-02
WIP: xe: use double with post-ops with f64 to preserve accuracy

echeresh:
[MFDNN-14668](https://jira.devtools.intel.com/browse/MFDNN-14668)

### DIFF
--- a/src/gpu/intel/matmul/ref.cl
+++ b/src/gpu/intel/matmul/ref.cl
@@ -273,12 +273,12 @@ __kernel void ref_matmul(__global SRC_DATA_T *A, __global WEI_DATA_T *B,
         temp += BIA_TO_REF(bia[bia_off]);
 #endif // WITH_BIAS
 
-        float dst_data;
+        POST_OP_DATA_T dst_data;
 #if WITH_SUM
-        dst_data = convert_float(DATA_TO_REF(C[dst_off]));
+        dst_data = (POST_OP_DATA_T)DATA_TO_REF(C[dst_off]);
 #endif // WITH_SUM
 
-        float po_acc = convert_float(temp);
+        POST_OP_DATA_T po_acc = temp;
 
 #if WITH_DROPOUT
 #if WITH_SEED_S64 && USE_OFFSET
--- a/src/gpu/intel/include/eltwise.h
+++ b/src/gpu/intel/include/eltwise.h
@@ -240,8 +240,8 @@ POST_OP_DATA_T gelu_erf_bwd(POST_OP_DATA_T dd, POST_OP_DATA_T s) {
                     + v * two_over_sqrt_pi * exp(-v * v));
 }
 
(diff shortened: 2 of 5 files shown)


echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/matmul/ref.cl
```diff
@@ -273,12 +273,12 @@ __kernel void ref_matmul(__global SRC_DATA_T *A, __global WEI_DATA_T *B,
         temp += BIA_TO_REF(bia[bia_off]);
 #endif // WITH_BIAS
 
-        float dst_data;
+        POST_OP_DATA_T dst_data;
 #if WITH_SUM
-        dst_data = convert_float(DATA_TO_REF(C[dst_off]));
+        dst_data = (POST_OP_DATA_T)DATA_TO_REF(C[dst_off]);
```
Simonsays095, line 278: `DATA_TO_REF` here will convert to float. We either need to change `DATA_TO_REF` for f64 to be `convert_double`, or carve out an exception via macros here.

Simonsays095 (approved):
There are still many other OpenCL kernels that use `float` instead of `POST_OP_DATA_T`. I guess those will be handled separately?

dzarukin (approved):

src/gpu/intel/matmul/ref.cl
rjoursler: Likely `DATA_TO_REF` should be replaced with `FLT_ACC_DATA_T` which is already used in this file. Alternatively, we can just refactor this to use the load/store API so that we don't need to go over this with a fine-grained comb looking for invalid casts.

As an aside, currently we have 3 different floating point accumulator types `DATA_TO_REF`, `POST_OP_DATA_T`, and `FLT_ACC_DATA_T`. I am not convinced we really need all of these.

echeresh: Thanks for the catch, let me move the PR to WIP for now. I want to try to migrate `REF` family of macros to load/store. Eliminating extra macros would be nice.

rjoursler: Just in case this is helpful, [here as a sample commit](https://github.com/uxlfoundation/oneDNN/pull/2351/changes/ea3297843438283b885416d8a81eb11c11b34b08) porting gemm to use the load store interface.

PR #4747, opened 2026-03-02
build: x64: workaround for GCC 8.5 ICE

vpirogov:
GCC 8.5 started to throw ICE on autogenerated GEMM kernels after unrelated changes in the codebase. These files already have special treatment in the build system to avoid slowing down compilation. Additionally disabling `-ftree-dominator-opt` prevents compiler from exploding.


vpirogov:
make test linters

PR #4752, opened 2026-03-03
cpu: x64: update APX feature detection

xuxinzen:
Implements MFDNN-14488

### DIFF
--- a/src/cpu/x64/cpu_isa_traits.hpp
+++ b/src/cpu/x64/cpu_isa_traits.hpp
@@ -370,6 +370,18 @@ inline const Xbyak::util::Cpu &cpu() {
     return cpu_;
 }
 
+namespace apx {
+
+inline bool is_available() {
+    const bool xsave_supported = cpu().has(Xbyak::util::Cpu::tOSXSAVE);
+    if (!xsave_supported) return false;
+
+    const uint64_t xcr0_features = Xbyak::util::Cpu::getXfeature();
+    return ((xcr0_features >> 19) & 1) == 1;
+}
+
+} // namespace apx
+
 namespace amx {
 
 // Return the target palette for AMX instructions. Currently this is `0` if AMX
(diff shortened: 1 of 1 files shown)


xuxinzen:
make test

src/cpu/x64/cpu_isa_traits.cpp
```diff
@@ -263,6 +263,16 @@ status_t set_cpu_isa_hints(dnnl_cpu_isa_hints_t isa_hints) {
...
 
+bool is_apx_available() {
+    // XSAVE feature must be supported in order to check APX feature
+    const bool xsave_supported = cpu().has(Xbyak::util::Cpu::tOSXSAVE);
+    if (!xsave_supported) return false;
+
+    // Check APX feature in XCR0[19]
+    uint64_t xcr0_features = Xbyak::util::Cpu::getXfeature();
+    return xcr0_features >> 19 == 1;
+}
```
dzarukin: It feels like this body should be moved into a header to comply with oneDNN gtests.
It also seems it's safe enough to do as it doesn't seem to use any cpp local objects.

georgen117:
I ended up investigating the exact same problem and came up with a more targeted solution.

https://github.com/uxlfoundation/oneDNN/pull/4808

src/cpu/x64/cpu_isa_traits.cpp
```diff
@@ -263,6 +263,16 @@ status_t set_cpu_isa_hints(dnnl_cpu_isa_hints_t isa_hints) {
...
 }
 
+bool is_apx_available() {
+    // XSAVE feature must be supported in order to check APX feature
+    const bool xsave_supported = cpu().has(Xbyak::util::Cpu::tOSXSAVE);
+    if (!xsave_supported) return false;
+
+    // Check APX feature in XCR0[19]
+    uint64_t xcr0_features = Xbyak::util::Cpu::getXfeature();
+    return xcr0_features >> 19 == 1;
```
georgen117: **Suggestion:** Replace the magic number 19 with a constexpr.
**Required:** change `xcr0_features >> 19 == 1` to `(xcr0_features >> 19) & 1` or `(xcr0_features >> 19) & 1 == 1` xcr0 is a bit field we need to isolate that bit

```
constexpr int XSAVE_APX_EGPR = 19; // or just use XSAVE_APX = 19
return (xcr0_features >> XSAVE_APX_EGPR) & 1;
```

I also agree with @dzarukin the body can be moved to the header. This is also could be an inline function.

georgen117:
I would suggest also adding the is_apx_available to the registry_scratchpad_t::ExtendedRegisters function found in injector_utils.hpp like I did in PR#4808.

I closed https://github.com/uxlfoundation/oneDNN/pull/4808/changes in favor of this PR.

My PR only added a check to the ExtendedRegisters there are a lot of potential instructions that can have #UD if the xcr0[APX] EGPR bit is not set. Since all AVX10.2 systems are expected to have APX it is probably best to group them together for developer simplicity. This will lock out some AVX10.2 instructions if the xcr0[APX] is not set (i.e. `vdpphps`) that don't rely on the APX feature.

ankalinin (approved):

PR #4761, opened 2026-03-04
rfc: graph: revisit sdpa training backward

ElaineBao:
# Description

as title


vpirogov (approved):

vpirogov:
LGTM as the approach was validated with PyTorch and fused GPU kernel.

TaoLv (approved):
@ElaineBao Could you please update this RFC with the recommended/approved solution?  With that, I think we get this document merged. Thanks.

PR #4766, opened 2026-03-05
xe: gemm: reintroduce kernels to spotfix regressions

Simonsays095:
Addresses [MFDNN-14483](https://jira.devtools.intel.com/browse/MFDNN-14483). Reintroduces one kernel that was removed. Since the performance model sometimes dispatches to this kernel when the other kernels would be better, this kernel is strictly limited to a few cases present in currently-used models. This is why one kernel is duplicated 3x - each entry covers another shape.


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4770, opened 2026-03-05
cpu: rv64: conv: add brgemm conv kernel for rv64

zhangjian29:
# Description

This PR introduces an intrinsically optimized, natively integrated `brgemm:rvv` primitive for RISC-V architectures, providing a streamlined *im2col-free* implicitly batched GEMM implementation. 

It directly invokes heavily-optimized underlying `extended_sgemm` hardware mechanisms while mapping convolution geometries directly to memory layout. 

We achieve exceptional acceleration over `gemm:rvv` for structurally complex convolution networks and dense layers (VGG-11, GoogleNet V3).

## Key Features

- **Im2col Overhead Elimination**: Instead of allocating and performing costly large-patch spatial transitions to memory arrays, memory limits dynamically adjust `extended_sgemm("N", "N", OC, valid_ow, IC, ...)` execution logic sequentially per spatial depth. This allows us to fully leverage the native vectorization of `rvv_gemm_f32` without the typical memory bandwidth bottlenecks of im2col.

### Implementation Details

- **Post-ops Support**: Currently only supports no post-ops or a single `sum` operation at the first position.
- **Data Types & Quantization**: Hardcoded to accept only `f32` inputs/outputs/bias. Non-f32 types, dynamic quantization, and zero-point compensation are currently excluded.
- **Memory Formats**: Strictly limited to `NSPC` layout (Channel-Last, e.g., `nhwc` / `nwc` / `ndhwc`). Other memory layout tags are rejected directly.
- **Shape Restrictions**: Explicitly rejects workloads where the output width is small (`ow < 20`) or the per-group input channels are small (`ic < 16`), as batched GEMM per-kernel-position loses efficiency on such dimensions.

---

# Checklist

## General

## Performance improvements

All experiments are performed on the remote test platform (SG2044 with VLEN=128), utilizing 8 execution cores. We draw comparisons among:

1. **main / ref** (upstream reference architecture containing `gemm:rvv` baseline im2col implementations and `jit_1x1:rvv`).
2. **brgemm:rvv** (this PR - optimized im2col-free implicitly batched convolution).

### Overall Topology Performance

| Dataset | Dispatch Precentage | Baseline (main) Time | This PR Time | Speedup |
|---------|---------------------|----------------------|--------------|-------------|
| **shapes_resnet_50** | 20% | 2796.72 ms | 2672.43 ms | **1.05×** |
| **shapes_googlenet_v3** | 14% | 2256.07 ms | 1888.99 ms | **1.19×** |
| **shapes_vgg_11** | 50% | 1651.25 ms | 1202.99 ms | **1.37×** |
| **shapes_gemm** | 5% | 1348.77 ms | 974.266 ms | **1.38×** |

**Analysis**:
- **Huge speedups** (up to >2× in some layers) obtained in spatial-heavy memory-bound models with large convolutions, such as VGG-11 and GoogleNet V3. The im2col-free approach eliminates massive memory bandwidth and processing overheads present in these large spatial kernels natively, translating to high performance bounds.
- **ResNet-50 Performance**: ResNet relies heavily on pure small convolutions which demand little to no `im2col` data expansion. For these blocks, we use *Shape Restrictions* to dispatch them to other implementations, which is more efficient for such small spatial dimensions.

### DIFF
--- a/src/cpu/rv64/rvv_brgemm_conv.cpp
+++ b/src/cpu/rv64/rvv_brgemm_conv.cpp
@@ -184,8 +184,10 @@ status_t rvv_brgemm_convolution_fwd_t::execute(const exec_ctx_t &ctx) const {
                             float *C = dst_base + oh_s * dst_h_str
                                     + ow_s * OC_all;
                             dim_t M = OC, K_dim = IC;
-                            extended_sgemm("N", "N", &M, &N, &K_dim, &one, A,
-                                    &LDA, B, &LDB, &one, C, &LDC);
+                            status_t st = extended_sgemm("N", "N", &M, &N,
+                                    &K_dim, &one, A, &LDA, B, &LDB, &one, C,
+                                    &LDC);
+                            if (st != status::success) return;
                         } else {
                             // Edge (partial width): per-row GEMM.
                             for (int oh = oh_s; oh < oh_e; oh++) {
@@ -196,8 +198,10 @@ status_t rvv_brgemm_convolution_fwd_t::execute(const exec_ctx_t &ctx) const {
                                 float *C = dst_base + oh * dst_h_str
                                         + ow_s * OC_all;
                                 dim_t M = OC, N_dim = valid_ow, K_dim = IC;
-                                extended_sgemm("N", "N", &M, &N_dim, &K_dim,
-                                        &one, A, &LDA, B, &LDB, &one, C, &LDC);
+                                status_t st = extended_sgemm("N", "N", &M,
+                                        &N_dim, &K_dim, &one, A, &LDA, B, &LDB,
+                                        &one, C, &LDC);
+                                if (st != status::success) return;
                             }
                         }
                     }
@@ -261,8 +265,10 @@ status_t rvv_brgemm_convolution_fwd_t::execute(const exec_ctx_t &ctx) const {
                             float *C = dst_row + ow_s * OC_all;
 
                             dim_t M = OC, N = valid_ow, K_dim = IC;
-                            extended_sgemm("N", "N", &M, &N, &K_dim, &one, A,
-                                    &LDA, B, &LDB, &one, C, &LDC);
+                            status_t st = extended_sgemm("N", "N", &M, &N,
+                                    &K_dim, &one, A, &LDA, B, &LDB, &one, C,
+                                    &LDC);
+                            if (st != status::success) return;
                         }
                     }
                 }


src/cpu/rv64/rvv_brgemm_conv.cpp
```diff
@@ -0,0 +1,290 @@
...
+                            const float *A = wei
+                                    + ((kd * KH + kh) * KW + kw) * wei_kpos_str
+                                    + g * OC;
+                            const float *B = src + n * src_mb_str
+                                    + id * src_d_str + ih * src_h_str
+                                    + iw_start * src_w_str + g * IC;
+                            float *C = dst_row + ow_s * OC_all;
+
+                            dim_t M = OC, N = valid_ow, K_dim = IC;
+                            extended_sgemm("N", "N", &M, &N, &K_dim, &one, A,
```
zhangfeiv0: It is recommended to add a return status check here.

zhangjian29: Sure. Thanks.

src/cpu/rv64/rvv_brgemm_conv.cpp
```diff
@@ -0,0 +1,296 @@
...
+                            // Full-width: source/dst contiguous across rows.
+                            dim_t num_rows = oh_e - oh_s;
+                            dim_t N = num_rows * OW;
+                            const float *B = src + n * src_mb_str
+                                    + id * src_d_str + ih_start * src_h_str
+                                    + iw_start * src_w_str + g * IC;
+                            float *C = dst_base + oh_s * dst_h_str
+                                    + ow_s * OC_all;
+                            dim_t M = OC, K_dim = IC;
+                            status_t st = extended_sgemm("N", "N", &M, &N,
```
dzarukin, line 187: This would mean that the proper impl name is `gemm_convolution`. BRGeMM is a different type of kernel: [this](https://github.com/uxlfoundation/oneDNN/tree/main/src/cpu/x64/brgemm) or [this](https://github.com/uxlfoundation/oneDNN/tree/main/src/cpu/aarch64/brgemm).

Names don't mean much but it might be misleading to other developers/users, though I can see another `riscv_gemm_convolution_fwd_t` which might be the same as this one... So, what would be the difference of a new one over the old one?

zhangjian29:
> Names don't mean much but it might be misleading to other developers/users, though I can see another riscv_gemm_convolution_fwd_t which might be the same as this one... So, what would be the difference of a new one over the old one?

Thank you for the detailed review and insightful comments @dzarukin .

## The Architectural Motivation

The primary goal of this PR is to learn from the `jit_brgemm_conv` implementations from `x64`and `aarch64`. By using block/batching iterators to compute over sparse/discrete data implicitly, it completely eliminates the massive `im2col` memory and bandwidth overheads, showing speedups up to 1.37x on VGG-11 compared to the standard `gemm_convolution`.

## The Current Fallback State
Currently, the RISC-V backend lacks a true JIT brgemm micro-kernel engine. Therefore, as a transitional approach, this PR hooks the batched inner computation loops into the existing `extended_sgemm` macro-kernel. While this inherently introduces some GEMM launch and repacking overheads, the gained benefits of avoiding `im2col` still exist.

## Roadmap and Naming Proposal
Once a true RVV BRGeMM micro-kernel is ready in the future, we can seamlessly switch this primitive to use it with very minimal code changes. Because this primitive shares the exact same "macro-loop shell" paradigm as `x64`/`aarch64`, simply renaming it to gemm_convolution might be misleading of its intents.

I suggest keeping the `rvv_brgemm_conv` name for now. I am actively planning to develop the actual RVV BRGeMM micro-kernel engine in subsequent PRs. Retaining this name aligns perfectly with the `x64`/`aarch64` naming conventions.

Does this sound like a reasonable solution to you? Thank you again for your time!

dzarukin (approved):
Fine with me, just wanted to make sure it's a deliberate choice.

PR #4773, opened 2026-03-05
xe: gemm: fix binary post-op layout when swap_ab

karturov:
Fixes a correctness bug in GPU matmul binary post-ops when `swap_ab` is in effect (transposed destination layout). A recent refactor moved the swap logic out of `transfer_post_ops` and added `postOps.transpose()` afterwards, but that call only updated the `binaryRow`/`binaryCol` bitsets and did not transpose the `MatrixAddressing` entries in `problem.binary[]`, causing the kernel to use wrong memory access patterns for the binary source. The fix also calls `b.transpose()` on each `problem.binary` entry to keep it consistent.

Jira: [MFDNN-14726](https://jira.devtools.intel.com/browse/MFDNN-14726)

### DIFF
--- a/tests/benchdnn/inputs/matmul/test_matmul_gpu
+++ b/tests/benchdnn/inputs/matmul/test_matmul_gpu
@@ -144,3 +144,6 @@
 # precomputed reductions for zero points
 --reset
 --batch=harness_matmul_precomp_reductions
+
+# float16 regressions
+--batch=harness_matmul_regression_float16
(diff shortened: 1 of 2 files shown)


karturov:
make test
disable test_device_cpu

tests/benchdnn/inputs/matmul/test_matmul_gpu
```diff
@@ -144,3 +144,12 @@
...
 --batch=harness_matmul_precomp_reductions
+
+# regression: binary post-op with explicit 2D format tag and transposed dst (swap_ab path)
+# MFDNN-14726: MatrixAddressing layout in problem.binary was not updated on swap_ab
+--reset
+--allow-enum-tags-only=0
+--dt=f16:f16:f16
+--stag=ab,ba --wtag=ab,ba --dtag=ba
+--attr-post-ops=linear:1.2+add:f16:3:ab+linear:1,linear:1.2+add:f16:3:ba+linear:1
+10x50:50x25
```
dzarukin: Nit: please put into `harness_matmul_regression_float16` and include this file into `test_matmul_gpu`.

PR #4774, opened 2026-03-05
update xbyak library to  v7.35.3

georgen117:
This updates the Xbyak library to be in sync with the upstream version of Xybak.

This addvances the version from Xybak v7.28 to Xbyak v7.35.3

This PR is broken down into 3 commits.

**1st commit:** Updates Xbyak to v7.35.3:
Changes to oneDNN to build with Xbyak v7.35.3:
- New feature XBYAK_STRICT_CHECK_MEM_REG_SIZE disabled in the
      cpu_isa_traits header. Several oneDNN JIT kernels use
      size-mismatched memory operands that are architecturally correct
      but will cause Xbyak to throw an ERR_BADD_MEM_SIZE error.
- Removed all instances of Xbyak::Address(0) the Address(size_t)
      constructor is removed. The added Xbyak::Address() constructor
      is used instead.
 
**2nd commit:**
- enables `XBYAK_STRICT_CHECK_MEM_REG_SIZE` macro and fixes places in the code that caused the `ERR_BAD_MEM_SIZE` failure.
- This is the highest risk change in this PR since the failures happen at runtime during JIT assembly, not at compile time so have to be caught by unit tests and benchdnn.

--------------------------------

Changes to the Xbyak library:

**Overall themes (v7.29-7.35):** Label and integer-based addressing improvements (RegExp/Address robustness and clarity), the new CpuTopology/CpuMask API for hybrid CPU topology, instruction additions (tcvtrowd2ps, prefetchrst2, extended NOPs), removal of deprecated ISA entries, and portability fixes (UWP, macOS, Clang C++14).

v7.29
- Added **Labels support in addressing**: `ptr[label]` now works. `ptr[integer]` format merged into `RegExp`
v7.30
- Added `tcvtrowd2ps` instruction support
v7.31
- **Major addition:** introduced `util::CpuTopology`, `LogicalCpu`, and `CpuMask` classes. for handling CPU cache and core topology on hybrid systems.
- `util::CpuTopology` uses OS specific methods to populate.
v7.32
- **Removed depricated instrucitons** referenced in Intel manual 319433-059
- Added comparison operators to `CpuMask` class
- added `put()` helper functions to `LogicalCpu` and `Cache` class for debug output
v7.33
- `CpuMask` string format added
- `CpuMask` storage switched from `std::vector<size_t` to `CpuMask`
- Fixed `inline` linkage issues
- `cmpxchg16b` now accepts `xword` operand (e.g., `cmpxchg16b xword[rax]`
v3.34
- Added `prefetchrst2 instruction
- added muti-byte NOP support for 10-15 byte encodings (previously caped at 9 bytes)
- Improved backward compatability for `Address` and `RegExp`
- made `RegExp` constructor explicit to resolve integer literal ambiguity
- Added default `Address` constructor
- Added `CpuMask::appendRange()`
v7.35
- **`RegExp` Addressing fixes**: `RegExp(0)` now treated as normal displacement.
- Removed `constespr` from `RegExp` to fix comilation error with `clang++ -std=c++14`

### DIFF
--- a/third_party/xbyak/xbyak.h
+++ b/third_party/xbyak/xbyak.h
@@ -82,8 +82,19 @@
 	#define XBYAK_GNUC_PREREQ(major, minor) 0
 #endif
 
+// User defined (must define all 3)
+#if defined(XBYAK_STD_UNORDERED_SET) || defined(XBYAK_STD_UNORDERED_MAP) || defined(XBYAK_STD_UNORDERED_MULTIMAP)
+	#ifndef XBYAK_STD_UNORDERED_SET
+		#error "Define XBYAK_STD_UNORDERED_SET"
+	#endif
+	#ifndef XBYAK_STD_UNORDERED_MAP
+		#error "Define XBYAK_STD_UNORDERED_MAP"
+	#endif
+	#ifndef XBYAK_STD_UNORDERED_MULTIMAP
+		#error "Define XBYAK_STD_UNORDERED_MULTIMAP"
+	#endif
 // This covers -std=(gnu|c)++(0x|11|1y), -stdlib=libc++, and modern Microsoft.
-#if ((defined(_MSC_VER) && (_MSC_VER >= 1600)) || defined(_LIBCPP_VERSION) ||\
+#elif ((defined(_MSC_VER) && (_MSC_VER >= 1600)) || defined(_LIBCPP_VERSION) ||\
 	 			 ((__cplusplus >= 201103) || defined(__GXX_EXPERIMENTAL_CXX0X__)))
 	#include <unordered_set>
 	#define XBYAK_STD_UNORDERED_SET std::unordered_set
@@ -168,8 +179,10 @@
 	#define XBYAK_TLS thread_local
 	#define XBYAK_VARIADIC_TEMPLATE
 	#define XBYAK_NOEXCEPT noexcept
+	#define XBYAK_OVERRIDE override
 #else
 	#define XBYAK_NOEXCEPT throw()
+	#define XBYAK_OVERRIDE
 #endif
 
 // require c++14 or later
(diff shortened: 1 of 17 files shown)


third_party/xbyak/xbyak.h
```diff
@@ -211,13 +211,8 @@
...
 // Define this macro as 0 to disable strict checking of memory operand and register size matching.
 // This macro may be removed in future versions.
-// oneDNN-specific note: several JIT kernels intentionally use size-mismatched memory operands
-// (e.g. zword[]-annotated addresses with 64-bit GPR instructions, or dword[] fields loaded into
-// Reg64 registers). These are architecturally correct on x86-64 but trigger this check. Until all
-// call sites are updated to use properly-sized operands or stripped annotations, set this to 0 to
-// suppress the false-positive errors.
 #ifndef XBYAK_STRICT_CHECK_MEM_REG_SIZE
-	#define XBYAK_STRICT_CHECK_MEM_REG_SIZE 0
+	#define XBYAK_STRICT_CHECK_MEM_REG_SIZE 1
```
dzarukin, line 215: Enabling such macros are done through `src/cpu/x64/cpu_isa_traits.hpp` file, not by modifying XByak sources.

georgen117: The macro was added to the upstream xbyak source code. It is defined as `1` by default. When updating the library, I changed it to `0` due to the errors.  Then later tried to modify it to match the upstream with the 3rd commit.

The upstream code has a comment that the macro may be removed in the future. To me this indicates xbyak intends the strict memory check to be the default behavior.

OneDNN is not obligated to use this check. Not using it does mean we are different than the default xbyak. We already maintain a few oneDNN specific modifications. (Mostly new ISA that have not been upstreamed yet.)


If we find an instruction that this check causes many issues with (`mov` may be a candidate) I believe the `T_ALLOW_DIFF_SIZE` flag seen a few places in the `xybak_mnemonic.h` bypasses the strict mem reg size check. I however didn't actually play around with changing xybak to see it that would work.

georgen117:
> Changes LGTM, one comment to address.
> 
> You might want to swap 1st and 2nd commit to preserve buildability of oneDNN for each commit (if default Address ctor is already available). If it's not available, it seems those two commits better be merged together.

The default Address ctor was not avalible.
To preserve buildability I added the Address(nullptr) ctor that was removed by the 2nd commit.

I will merge the two commits together.

georgen117:
@dzarukin I merged the original 1st and 2nd commits into a single commit. I also moved enabling/disabling the `XBYAK_STRICT_CHECK_MEM_REG_SIZE` to the `src/cpu/x64/cpu_isa_traits.hpp` as suggested.

Thanks for both your suggestions.

tprimak (approved):

georgen117:
make test

georgen117:
I removed the second commit that enabled the new `XBYAK_STRICT_CHECK_MEM_REG_SIZE` feature. That change resulted in changes to the code that I would like to have focused attention by reviewers.

I will commit it as its own PR. This will help if any new errors occur as a result of enabling the feature it can more easily be traced to a single commit and PR.

densamoilov (approved):

PR #4775, opened 2026-03-05
benchdnn: matmul: resolving a few follow-ups from grouped matmul enabling

mzhukova:
Addresses portion of the [MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704) to allow for future grouped gemm extensions:
- migrate ref grouped matmul to the computation kernel shared with regular matmul
- get rid of special handling for ref WEI creation
- add more parallelism for compute

Some moving around of the code happened, so easier to review by commits :)

### DIFF
--- a/tests/benchdnn/matmul/ref_matmul.cpp
+++ b/tests/benchdnn/matmul/ref_matmul.cpp
@@ -14,7 +14,6 @@
 * limitations under the License.
 *******************************************************************************/
 
-#include <algorithm>
 #include <float.h>
 
 #include "utils/parallel.hpp"
@@ -339,8 +338,10 @@ void compute_ref_grouped_matmul(const prb_t *prb, const args_t &args) {
 
     std::vector<int64_t> group_offsets(group_count + 1);
     group_offsets[0] = 0;
+    int64_t max_M_g = 0;
     for (int64_t g = 0; g < group_count; g++) {
         group_offsets[g + 1] = group_offsets[g] + M_dims[g];
+        max_M_g = MAX2(max_M_g, M_dims[g]);
     }
 
     // Precompute common parameters for the different chunks computations
@@ -350,11 +351,15 @@ void compute_ref_grouped_matmul(const prb_t *prb, const args_t &args) {
     const int64_t wei_k_stride = params.wei_m->strides()[wei_ndims - 2];
     const int64_t wei_n_stride = params.wei_m->strides()[wei_ndims - 1];
 
-    // Parallelize over groups
-    // Each group is a separate matmul of size M_dims[g] x K by K x N
-    benchdnn_parallel_nd(group_count, [&](int64_t g) {
+    const int64_t N_chunks = div_up(prb->n, params.dst_N_group);
+    const int64_t max_M_chunks = div_up(max_M_g, params.dst_M_group);
+
+    // Parallelize over groups and (mc, nc) chunks within each group
+    benchdnn_parallel_nd(group_count, max_M_chunks, N_chunks,
+            [&](int64_t g, int64_t mc, int64_t nc) {
         const int64_t M_g = M_dims[g];
         if (M_g == 0) return;
+        if (mc * params.dst_M_group >= M_g) return;
 
         // Precompute offsets for this group
         //   src(m, k)    = (src_row_base + m) * K + k
@@ -373,16 +378,10 @@ void compute_ref_grouped_matmul(const prb_t *prb, const args_t &args) {
             bia_n_stride = 1;
         }
 
-        // Computation: iterate over (mc, nc) dst-scale blocks.
-        const int64_t M_blocks = div_up(M_g, params.dst_M_group);
-        const int64_t N_blocks = div_up(prb->n, params.dst_N_group);
-
-        for (int64_t mc = 0; mc < M_blocks; mc++)
-            for (int64_t nc = 0; nc < N_blocks; nc++)
-                compute_ref_matmul_chunk(params, M_g, prb->n, prb->k, mc, nc,
-                        src_row_base, wei_base, wei_k_stride, wei_n_stride,
-                        dst_row_base, bia_base, bia_m_stride, bia_n_stride,
-                        prb->attr, args);
+        compute_ref_matmul_chunk(params, M_g, prb->n, prb->k, mc, nc,
+                src_row_base, wei_base, wei_k_stride, wei_n_stride,
+                dst_row_base, bia_base, bia_m_stride, bia_n_stride, prb->attr,
+                args);
     });
 }
 #endif // DNNL_EXPERIMENTAL_GROUPED_MEMORY


mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -571,140 +488,27 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
...
+                = dst_m.get_idx(dst_off_f(prb, mb, 1, 0), bias_broadcast_mask)
+                - bia_base;
+        const int64_t bia_n_stride
+                = dst_m.get_idx(dst_off_f(prb, mb, 0, 1), bias_broadcast_mask)
+                - bia_base;
+
+        // Precompute offsets for this chunk
+        const int64_t src_row_base = src_mb * M;
+        const int64_t wei_ab_base = wei_mb * K * N;
+        const int64_t wei_ba_base = wei_mb * N * K;
```
dzarukin: This look like the same base, duplication might be eliminated. Having a different base point for weight depending on the format might confuse, too.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -571,140 +488,27 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
...
-            dst_m.set_f32_elem(dst_off, dst_val);
-        }
+        const int64_t bia_base
+                = dst_m.get_idx(dst_off_f(prb, mb, 0, 0), bias_broadcast_mask);
+        const int64_t bia_m_stride
+                = dst_m.get_idx(dst_off_f(prb, mb, 1, 0), bias_broadcast_mask)
+                - bia_base;
+        const int64_t bia_n_stride
+                = dst_m.get_idx(dst_off_f(prb, mb, 0, 1), bias_broadcast_mask)
+                - bia_base;
```
dzarukin: You might want to consider avoiding unnecessary overhead when bias was not requested.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -571,140 +488,27 @@ void compute_ref_matmul(const prb_t *prb, const args_t &args) {
...
+
+        // Precompute offsets for this chunk
+        const int64_t src_row_base = src_mb * M;
+        const int64_t wei_ab_base = wei_mb * K * N;
+        const int64_t wei_ba_base = wei_mb * N * K;
+        const int64_t wei_k_stride = 1;
+        const int64_t wei_n_stride = K;
```
dzarukin: Maybe pick up strides from wei_m directly so that changing strides on the upper level would be reflected here?

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -324,128 +324,64 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
...
-    // Use finest granularity k-group (GCD) to handle mixed scale/ZP groups
-    const int64_t smallest_k_group
-            = gcd<int64_t>({wei_scale_group_k, wei_zp_group_k});
-    const int64_t n_k_groups = K / smallest_k_group;
 
     std::vector<int64_t> group_offsets(group_count + 1);
     group_offsets[0] = 0;
     for (int64_t g = 0; g < group_count; g++) {
         group_offsets[g + 1] = group_offsets[g] + M_dims[g];
     }
```
dzarukin, line 345: Maybe take those values directly from the src memory object instead?

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -324,128 +324,64 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
...
+        //   src(m, k)    = (src_row_base + m) * K + k
+        //   wei_ab(k, n) = wei_ab_base + k * N + n (for scales and zps)
+        //   wei_ba(k, n) = wei_ba_base + k * wei_k_stride + n * wei_n_stride
+        //   dst(m, n)    = (dst_row_base + m) * N + n
+        //   bia(m, n)    = bia_base + m * bia_m_stride + n * bia_n_stride
+        const int64_t src_row_base = group_offsets[g];
+        const int64_t wei_ab_base = g * prb->k * prb->n;
+        const int64_t wei_ba_base = g * prb->k * prb->n; // abc: k-major
+        const int64_t wei_k_stride = prb->n;
+        const int64_t wei_n_stride = 1;
```
dzarukin: Ditto about taking strides directly.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -324,128 +324,64 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
...
     for (int64_t g = 0; g < group_count; g++) {
         group_offsets[g + 1] = group_offsets[g] + M_dims[g];
     }
 
+    // Precompute common parameters for the different chunks computations
+    const chunk_params_t params = make_chunk_params(prb, args);
+
+    // Parallelize over groups
+    // Each group is a separate matmul of size M_dims[g] x K by K x N
     benchdnn_parallel_nd(group_count, [&](int64_t g) {
```
dzarukin: Maybe try to squeeze some more parallelization over chunks? With 4 groups and big matmuls underneath it feels like a lot of resources are wasted...

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: good point, updated

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: done

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: sure, updated

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: this should have been updated now too

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: so `src` here is just plain 2D f32 buffer (built based on values of `src` that goes into the library), therefore it doesn't have an offset buffer; I think I don't have access here to the grouped `src` (I believe the `compute_ref` only has access to `ref_args`, right..?), so the only option is to compute offsets based on whatever is stored in `prb->sparse_options` seems like

dzarukin: I was thinking about ref_src following 2-buffered memory, data is 2D f32 buffer, offsets is 1D int32 buffer, both packed inside at filling stage.

And group_offsets values to be obtained as `params.src->get_elem(idx, 1)`.
Once you move from command-line input to filling, this transition must have happened anyway...

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: updated in [benchdnn: matmul: better parallelization for grouped ref](https://github.com/uxlfoundation/oneDNN/pull/4775/commits/4286d2681a9481d070b765d72365c26124a43a60)

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: I would like to get current changes in and switch gears for a bit, so I've added this to the list in [MFDNN-14704](https://jira.devtools.intel.com/browse/MFDNN-14704); thanks for the suggestion! I'll investigate

PR #4777, opened 2026-03-06
graph: updates for f16/bf16 gated mlp patterns

TaoLv:
Clarify the intermediate f32 data type used in f16/bf16 gated mlp patterns.

- Up and Gate matmuls will have f16/bf16 inputs and generate f32 outputs.
- Activation and Multiply are performed in f32, hence can be fused into Up or Gate.
- An additional TypeCast to downconvert the intermediate results from f32 to f16 or bf16.
- Down matmul has f16/bf16 inputs, the output can be f16/bf16 or f32 when needed.

### DIFF
--- a/doc/graph/fusion_patterns/gated_mlp.md
+++ b/doc/graph/fusion_patterns/gated_mlp.md
@@ -80,10 +80,13 @@ optional.
 
 ## Data Types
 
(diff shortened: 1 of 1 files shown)


TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4779, opened 2026-03-06
common: deduce runtime value marker from types

asimonov1:
The PR removes explicit use of constants the `DNNL_RUNTIME_DIM_VAL`, `DNNL_RUNTIME_SIZE_VAL`, `DNNL_RUNTIME_S32_VAL`, `DNNL_RUNTIME_F32_VAL` in CPU-related part of the library.
We already had overloads of the `is_runtime_value` functions, but they weren't used in many places.
New variadic function templates have also been added to check whether at least one or all values in a set are equal to the corresponding runtime placeholder value, taking into account the type of values.
E.g., the call
```
all_runtime_values(
    DNNL_RUNTIME_DIM_VAL,
    DNNL_RUNTIME_SIZE_VAL,
    DNNL_RUNTIME_S32_VAL,
    DNNL_RUNTIME_F32_VAL,
    runtime_value_for<int>(),
    runtime_value_for<dim_t>(),
    runtime_value_for<float>(),
    runtime_value_for<size_t>())
```
returns `true`.
Functions to deduce runtime placeholder values from a variable's type have also been added. This can help avoid bugs when, for example, the type of a structure member has changed. If the type has changed, we need to find all occurrences of that member and update all assignments and conditional uses of runtime value placeholders.

Functions are placed in utils.hpp (not type_helpers.hpp) to avoid circular dependencies with files included in type_helpers.hpp. One another option is to place them in a separate file.

### DIFF
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -644,11 +644,11 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
         const dim_t b_dim = with_bias ? op_d.bias_desc.dims[d] : 0;
         const dim_t r_dim = with_reduce ? op_d.reduce_desc.dims[d] : 0;
 
-        if (one_of(DNNL_RUNTIME_DIM_VAL, s_dim, w_dim, d_dim, b_dim)) {
+        if (any_runtime_value(s_dim, w_dim, d_dim, b_dim)) {
 
-            VCHECK_MATMUL(everyone_is(DNNL_RUNTIME_DIM_VAL, s_dim, w_dim, d_dim)
+            VCHECK_MATMUL(all_runtime_values(s_dim, w_dim, d_dim)
                             && IMPLICATION((bia_mask & (1 << d)) && with_bias,
-                                    b_dim == DNNL_RUNTIME_DIM_VAL),
+                                    is_runtime_value(b_dim)),
                     VERBOSE_RUNTIMEDIM_INCONSISTENT, d);
         } else {
             // This follows numpy semantics of broadcasting when 0 is involved.
--- a/src/common/utils.hpp
+++ b/src/common/utils.hpp
@@ -970,6 +970,76 @@ public:
     }
 };
 
+/** returns true if fp32 value denotes DNNL_RUNTIME_F32_VAL */
+inline bool is_runtime_value(float val) {
+    return utils::bit_cast<unsigned>(val) == DNNL_RUNTIME_F32_VAL_REP.u;
+}
+
+/** returns true if s32 value denotes DNNL_RUNTIME_S32_VAL */
+inline bool is_runtime_value(int val) {
+    return val == DNNL_RUNTIME_S32_VAL;
+}
+
+/** returns true if dim_t value denotes DNNL_RUNTIME_DIM_VAL */
+inline bool is_runtime_value(dim_t val) {
+    return val == DNNL_RUNTIME_DIM_VAL;
+}
+
+/** returns true if size_t value denotes DNNL_RUNTIME_SIZE_VAL */
+inline bool is_runtime_value(size_t val) {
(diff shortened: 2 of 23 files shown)


asimonov1:
make test
enable build_mode_cpp20
enable compiler_icx
enable compiler_icx-previous
enable compiler_icx-oss
enable compiler_gnu8
enable compiler_gnu9
enable compiler_gnu10
enable compiler_gnu11
enable compiler_gnu12
enable compiler_gnu13
enable compiler_clang11
enable compiler_clang14
enable compiler_clang16
enable compiler_clang19
enable compiler_vs2022

asimonov1:
make test
enable build_mode_cpp20
enable compiler_icx
enable compiler_icx-previous
enable compiler_icx-oss
enable compiler_gnu8
enable compiler_gnu9
enable compiler_gnu10
enable compiler_gnu11
enable compiler_gnu12
enable compiler_gnu13
enable compiler_clang11
enable compiler_clang14
enable compiler_clang16
enable compiler_clang19
enable compiler_vs2022

src/common/matmul.cpp
```diff
@@ -644,11 +644,11 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
         const dim_t b_dim = with_bias ? op_d.bias_desc.dims[d] : 0;
         const dim_t r_dim = with_reduce ? op_d.reduce_desc.dims[d] : 0;
 
-        if (one_of(DNNL_RUNTIME_DIM_VAL, s_dim, w_dim, d_dim, b_dim)) {
+        if (any_runtime_value(s_dim, w_dim, d_dim, b_dim)) {
```
dzarukin, line 647: Maybe `one_of_is_runtime_value`? Same with `everyone_is_runtime_value`?
It seems longer but combines two existing concepts.

src/common/utils.hpp
```diff
@@ -970,6 +970,82 @@ class mask_iterator { // NOLINT(readability-identifier-naming)
...
+    return DNNL_RUNTIME_DIM_VAL;
+}
+
+/** @copydoc runtime_value_for() */
+template <>
+constexpr size_t runtime_value_for<size_t>() {
+    return DNNL_RUNTIME_SIZE_VAL;
+}
+
+/** returns the runtime placeholder constant for the type of @p v */
```
dzarukin:
```suggestion
/** returns the runtime placeholder constant for the type of @p T */
```

src/common/utils.hpp
```diff
@@ -970,6 +970,82 @@ class mask_iterator { // NOLINT(readability-identifier-naming)
...
+constexpr bool all_runtime_values(T item) {
+    return is_runtime_value(item);
+}
+template <typename T, typename... Args>
+constexpr bool all_runtime_values(T item, Args... item_others) {
+    return is_runtime_value(item) && all_runtime_values(item_others...);
+}
+
+/** returns the placeholder constant used to indicate that a value of type @p T
+ * will be provided at execution time rather than at primitive creation time */
```
dzarukin: Maybe remove this? Doesn't seem to be connected.

src/common/utils.hpp
asimonov1: Fixed.

src/common/matmul.cpp
asimonov1: Maybe. I tried to find names that are not too long. And we have `any_null` and `any_memory_desc_host_scalar`  functions which names are similar. I considered also `any_of_is_rt_val`/`any_is_rt_val`, `every_is_rt_val`, and I think that "every" is better than "all" because "all" and "any" look similar.

asimonov1: one_* sounds misleading to me. One is one, not two or three.

asimonov1:
make test
set test_scope=NIGHTLY

asimonov1:
make test
set test_scope=NIGHTLY

PR #4782, opened 2026-03-06
gpu: fix xe softmax large-shape offset overflow

h-sadia:
# Description

Softmax xe kernel can't handle large shapes since we are doing integer arithmatic for offset calculation which exceeds 2^31 int max capacity. This fixes data_off integer overflow. 

Fixes # (MFDNN-14479)

# Checklist

## General

### Bug fixes

### DIFF
--- a/tests/benchdnn/inputs/softmax/harness_softmax_regression
+++ b/tests/benchdnn/inputs/softmax/harness_softmax_regression
@@ -5,3 +5,5 @@
 --ddt=s8
 --attr-scales=dst:common:0.125,src:common:2
 --axis=3 1x4x192x256 1x4x16x32
+
+--reset --allow-enum-tags-only=0 --dir=FWD_I --stag=abcd --dtag=abcd --alg=SOFTMAX --attr-scratchpad=user --axis=3 1x1x65536x65536


h-sadia:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_softmax

src/gpu/intel/softmax/xe.cl
```diff
@@ -489,7 +492,8 @@ xe_softmax_bwd(__global DST_DATA_T *dst, __global SRC_DATA_T *diff_src,
     }
 
 #else
-    const int data_off = (get_global_id(0) / GROUP_SIZE) * SOFTMAX_AXIS_SIZE;
+    const dim_t data_off
+            = (dim_t)(get_global_id(0) / GROUP_SIZE) * SOFTMAX_AXIS_SIZE;
```
dzarukin: Random: is it the only implementation that can't handle large strides?

tests/benchdnn/inputs/softmax/test_softmax_gpu
```diff
@@ -47,6 +47,15 @@
...
 --alg=LOGSOFTMAX --axis=3 64x3x224x224
 
+# regression: large-axis inf-as-zero
+--reset
+--allow-enum-tags-only=0
+--dir=FWD_I
+--stag=abcd --dtag=abcd
+--alg=SOFTMAX_INF_AS_ZERO
+--attr-scratchpad=user
+--axis=3 1x1x65536x65536
```
dzarukin: Please use this line:
```
--reset --dir=FWD_I --stag=abx --alg=SOFTMAX --axis=3 1x1x65536x65536_n"large_offsets"
```

And please move to `harness_softmax_regression`.

src/gpu/intel/softmax/xe.cl
h-sadia: This is the one that this test calls in particular. But I can check the other ones, and see if offset type has to be changed there as well.

atkassen (approved):

src/gpu/intel/softmax/xe.cl
```diff
@@ -146,8 +146,8 @@ xe_softmax_fwd(__global SRC_DATA_T *src, __global DST_DATA_T *dst,
 
         __global SRC_DATA_T *src_copy = src;
         int axis_offset = find_axis_offset(i, subgroup_local_id, subgroup_id);
```
kealan-barbieri, line 148: is there any chance of values like `axis_offset` exceeding the int limits as well?

h-sadia: Not with this current regression-case. In this case the axis size is the maximum size of the tensor which is 65536, less than UINT32_MAX.

h-sadia:
make test
disable device_cpu
disable benchdnn_all
enable benchdnn_softmax

h-sadia:
> Can you use `kernel_ctx.register_buffer_size()` to register all of the relevant buffers during kernel creation, then use `off_t` instead of `dim_t`? Then we'll use 64-bit integers only when required by one of the buffers.

Added `register_buffer_size()`

tests/benchdnn/inputs/softmax/harness_softmax_regression
```diff
@@ -5,3 +5,5 @@
 --ddt=s8
 --attr-scales=dst:common:0.125,src:common:2
 --axis=3 1x4x192x256 1x4x16x32
+
+--reset --allow-enum-tags-only=0 --dir=FWD_I --stag=abcd --dtag=abcd --alg=SOFTMAX_INF_AS_ZERO --attr-scratchpad=user --axis=3 1x1x65536x65536
```
dzarukin: `SOFTMAX_INF_AS_ZERO` is an internal kind. Better use public one:
```suggestion
--reset --allow-enum-tags-only=0 --dir=FWD_I --stag=abcd --dtag=abcd --alg=SOFTMAX --attr-scratchpad=user --axis=3 1x1x65536x65536
```

dzarukin:
Just in case, I verified if the reported json would pass, and it wouldn't because of binary, too:
```
create: --binary --engine=gpu --mem-check=false --stag=abcd:abcd --dtag=abcd --alg=mul 1x1x65536x65536:1x1x1x1
...
oneDNN implementation: ocl:xe
...
Segmentation fault from GPU at 0xff00000200024000, ctx_id: 1 (CCS) type: 0 (NotPresent), level: 2 (PDP), access: 0 (Read), banned: 1, aborting.
Segmentation fault from GPU at 0xff00000200024000, ctx_id: 1 (CCS) type: 0 (NotPresent), level: 2 (PDP), access: 0 (Read), banned: 1, aborting.
Abort was called at 288 line in file:
./shared/source/os_interface/linux/drm_neo.cpp
Aborted (core dumped)
```

Please let me know if you want to handle it in a different PR, or in this one.

src/gpu/intel/softmax/xe.cl
```diff
@@ -146,8 +146,8 @@ xe_softmax_fwd(__global SRC_DATA_T *src, __global DST_DATA_T *dst,
...
-        int data_off = mb * CHANNELS_PADDED * SOFTMAX_AXIS_SIZE + axis_offset
-                + channel_id;
+        off_t data_off = (off_t)mb * CHANNELS_PADDED * SOFTMAX_AXIS_SIZE
```
rjoursler, line 149: Just change the data type of `mb` to `off_t` rather than casting.

src/gpu/intel/softmax/xe.hpp
```diff
@@ -224,6 +224,8 @@ struct xe_fwd_t : public primitive_t {
         def_memory_desc_info(kernel_ctx, src_md_info, "SRC");
         kernel_ctx.set_data_type(dst_mdw.data_type());
         kernel_ctx.require_stateless_addressing(pd()->has_large_buffers());
```
rjoursler, line 226: The stateless addressing check is automatically performed by [register_buffer_size](https://github.com/uxlfoundation/oneDNN/blob/a47809054a56203ef4e2763be99f7d8a277b5067/src/gpu/intel/compute/kernel_ctx.hpp#L169-L172), we can go ahead and drop these lines.

src/gpu/intel/softmax/simple.cl
```diff
@@ -99,7 +99,7 @@ simple_softmax_fwd_generic(__global SRC_DATA_T *src, __global DATA_T *dst,
     // finding max value for each sub_group
     if (!(NEEDS_PADDING(dim[0], dim[1], dim[2], dim[3], dim[4], begin))) {
         for (int i = begin; i < end && i < DD(SOFTMAX_AXIS_IDX); ++i) {
-            dim_t data_off
+            off_t data_off
                     = DATA_OFF(dim[0], dim[1], dim[2], dim[3], dim[4], i);
```
rjoursler, line 103: `dim` needs updated to type `off_t` as well, otherwise we can have an overflow before assigning to `data_off`.

src/gpu/intel/softmax/simple.hpp
```diff
@@ -183,6 +183,8 @@ struct simple_fwd_t : public primitive_t {
         def_memory_desc_info(kernel_ctx, src_md_info, "SRC");
         kernel_ctx.set_data_type(dst_mdw.data_type());
         kernel_ctx.require_stateless_addressing(pd()->has_large_buffers());
+        kernel_ctx.register_buffer_size(*pd()->src_md());
+        kernel_ctx.register_buffer_size(*pd()->dst_md());
```
rjoursler: We also need to register any post-op binary buffers.  Maybe we can add it these registrations to `def_post_ops_cfg`?

src/gpu/intel/softmax/xe.hpp
```diff
@@ -354,6 +356,9 @@ struct xe_bwd_t : public primitive_t {
         def_memory_desc_info(kernel_ctx, diff_dst_md_info, "DST");
         kernel_ctx.set_data_type(pd()->diff_src_md()->data_type);
         kernel_ctx.require_stateless_addressing(pd()->has_large_buffers());
+        kernel_ctx.register_buffer_size(*pd()->dst_md());
+        kernel_ctx.register_buffer_size(*pd()->diff_src_md());
+        kernel_ctx.register_buffer_size(*pd()->diff_dst_md());
```
rjoursler: Many of these register_buffer_size calls are unnecessary as they are performed by [def_memory_desc_info](https://github.com/uxlfoundation/oneDNN/blob/a47809054a56203ef4e2763be99f7d8a277b5067/src/gpu/intel/primitive_conf.cpp#L375)

src/gpu/intel/softmax/xe.cl
h-sadia: Its not really casting mb in particular, rather the result of this multiplication. The product of `mb * CHANNELS_PADDED * SOFTMAX_AXIS_SIZE` can overflow, I can put the calculation in parenthesis to make it clear.

src/gpu/intel/softmax/xe.hpp
Simonsays095: `pd()->has_large_buffers()` checks all input/output buffers (e.g. scales buffers as well) which we aren't necessarily registering here. We can drop here, since I think scales should always be no larger than the src buffer, but it's something to be aware of.

src/gpu/intel/softmax/xe.hpp
h-sadia: Would removing this `require_stateless_addressing(pd()->has_large_buffers());` also make sense here?

src/gpu/intel/softmax/simple.hpp
h-sadia: I see that we are only sending dst_md() in def_post_ops_cfg which is already registered in def_memory_desc_info.

src/gpu/intel/softmax/xe.hpp
rjoursler: > pd()->has_large_buffers() checks all input/output buffers (e.g. scales buffers as well) which we aren't necessarily registering here.

Then this check should be handle as part `def_attr_info`, similar to how `def_post_ops_cfg` works in my suggestion below.

src/gpu/intel/softmax/simple.hpp
rjoursler: This isn't correct, we need to register [e.binary.src1_desc](https://github.com/uxlfoundation/oneDNN/blob/9897c116b52f2d47a7503a7876ceb34166c328c6/src/gpu/intel/primitive_conf.cpp#L522) and [weight_mem_desc](https://github.com/uxlfoundation/oneDNN/blob/9897c116b52f2d47a7503a7876ceb34166c328c6/src/gpu/intel/primitive_conf.cpp#L526-L528).

src/gpu/intel/softmax/xe.cl
rjoursler: While true for this particular case, there is nothing in the implementation preventing `mb` from being larger than `INT_MAX` in the implementaition. For simplicity, we may as well just modify the `mb` type, unless there is some major performance issue.

src/gpu/intel/softmax/xe.hpp
rjoursler: Yes, the `require_stateless_addressing` comment applies to all the implementations.

src/gpu/intel/softmax/simple.hpp
h-sadia: Oh, so you want me to register the memory desc which are being initialized within the function? How will that be relevant? Like `post_op::relative_md_t src_rmd;` for `e.binary.src1_desc`? Or, e.binary.src1_desc ?

rjoursler: Correct, these memory descriptors define the binary post-op buffer. It is possible that this buffer could require large buffer handling, so they need checked. For example, if we were operating on a large int8 destination buffer, but binary buffer was `f32`, the binary buffer may trigger large buffer handling when the int8 destination doesn't.

src/gpu/intel/softmax/xe.hpp
h-sadia: I am keeping the stateless check as is, since this is a longer refactor related to other primitives as well.

h-sadia: Created a jira for this issue here: https://jira.devtools.intel.com/browse/MFDNN-14750 @Simonsays095 @rjoursler

PR #4783, opened 2026-03-06
cpu: matmul: Enable int8 matmul weight decompression

inteldimitrius:
Implements int4->int8 weight decompression for matmul.
The case when the computations are in int8. This flow is differs from normal weight decompression.
This PR includes:
- Weight upconversion `u8:s4:u8`, `s8:s4:s8`, `u8:u4:u8` , `s8:u4:s8` implemented in `brgemm_matmul_copy_utils.cpp`. 
- Common/per_oc/per_ocic (grouped) zero point shift in `brgemm_matmul_copy_utils.cpp` **only for s8 zero point dt**
- Scales per_ocic  implemented in `jit_brgemm_kernel.cpp`. Applying scales between dot-product and the accumulation. For AMX case it should be K_group % 64 == 0, in order to process the whole tile as one group element. Other cases will fallback to AVX512 brgemm implementation. Scales supports data types: f16, bf16, f32. Scales applying inetersects with s8s8-compensation funcionality, so it requires a custom processing for it.
- Test coverage

Supported isa: AVX512+, AVX2-VNNI, AMX

Performance testing on SPR:
[performance_report.pdf](https://github.com/user-attachments/files/26304794/performance_report.pdf)
[performance_data.xlsx](https://github.com/user-attachments/files/26304793/performance_data.xlsx)

### DIFF
--- a/src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
@@ -2682,12 +2682,16 @@ void jit_brgemm_kernel_t<Wmm>::gemm_microkernel(dim_t bd_block2,
         }
 
         if (brg.req_s8s8_compensation) {
-            if (brg.is_ic_wei_scales) {
-                // vmm_inp_shift() corrupted in IC scales + compensation.
-                mov(reg_s8_input_shift, 128);
-                uni_vpbroadcastb(vmm_inp_shift(), reg_s8_input_shift.cvt8());
-            }
             uni_vpaddb(vmm_bcast, vmm_bcast, vmm_inp_shift());
+        }
+    };
+
+    // vmm_inp_shift() is corrupted in dot-product with IC scales
+    // and this lambda restores it when needed for compensation.
+    auto maybe_restore_vmm_inp_shift = [&]() {
+        if (brg.req_s8s8_compensation && brg.is_ic_wei_scales) {
+            mov(reg_s8_input_shift, 128);
+            uni_vpbroadcastb(vmm_inp_shift(), reg_s8_input_shift.cvt8());
         }
     };
 
@@ -2777,6 +2781,7 @@ void jit_brgemm_kernel_t<Wmm>::gemm_microkernel(dim_t bd_block2,
 
     for (dim_t rd = 0; rd < rd_loop; rd += brg.rd_step) {
         if (brg.n_bcast_1_load) {
+            if (rd > 0) maybe_restore_vmm_inp_shift();
             for (dim_t bd = bd_b; bd < bd_e && !is_emdbd; bd++)
                 broadcast_A(bcst(bd), bd, rd);
             for (dim_t ld = 0; ld < ld_block2; ld++) {
@@ -2811,6 +2816,7 @@ void jit_brgemm_kernel_t<Wmm>::gemm_microkernel(dim_t bd_block2,
             }
 
             for (dim_t bd = bd_b; bd < bd_e; bd++) {
+                if (rd > 0 || bd > bd_b) maybe_restore_vmm_inp_shift();
                 if (!is_emdbd) broadcast_A(bcst(), bd, rd);
                 if (brg.is_fp8_via_convert_non_amx())
                     maybe_pre_process_data(brg.dt_a, bcst(), vmm_fp8_bcst());
--- a/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
@@ -3159,9 +3159,11 @@ protected:
         const auto addr = ptr[rdx];
         if (is_tail && !mask_supported) {
             load_bytes(vmm_zp, addr, ncolumns);
-        } else {
+        } else if (mask_supported) {
             const auto masked_vmm = maybe_mask(vmm_zp, is_tail);
             vmovdqu8(masked_vmm, addr);
+        } else {
+            uni_vmovdqu(vmm_zp, addr);
         }
         // Apply zero point shift
         uni_vpsubb(vmm_src, vmm_src, vmm_zp);
@@ -3191,9 +3193,11 @@ protected:
         const auto addr = ptr[reg_tmp];
         if (is_tail && !mask_supported) {
             load_bytes(vmm_zp, addr, ncolumns);
-        } else {
+        } else if (mask_supported) {
             const auto masked_vmm = maybe_mask(vmm_zp, is_tail);
             vmovdqu8(masked_vmm, addr);
+        } else {
+            uni_vmovdqu(vmm_zp, addr);
         }
         // Apply zero point shift
         uni_vpsubb(vmm_src, vmm_src, vmm_zp);
@@ -3710,10 +3714,12 @@ private:
                         auto offset = (is_dynamic_stride_ ? 0 : i * src_stride_)
                                 + pass * simd_w_;
                         offset /= src_elems_per_byte_;
+                        const auto tail_elements = ncolumns - pass * simd_w_;
                         const auto tail_size
-                                = div_up((ncolumns - pass * simd_w_),
-                                        src_elems_per_byte_);
+                                = div_up(tail_elements, src_elems_per_byte_);
                         load_ymm(i % 4, offset, do_tail, tail_size);
+                        maybe_apply_zero_points(get_ymm(i % 4), i, do_tail,
+                                tail_elements, pass * simd_w_);
                         if (is_dynamic_stride_) add(reg_src, reg_src_stride);
                     } else {
                         const auto src_ymm_1 = get_ymm(i % 4);


inteldimitrius:
make test

inteldimitrius:
make test

inteldimitrius:
make test

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2284,6 +2351,124 @@ void jit_brgemm_kernel_t<Wmm>::dot_product(Vmm v1, Vmm v2, Vmm v3) {
...
+        auto prepare_and_apply_mask = [&]() {
+            reg_grouped_k.restore();
+            reg_wei_scales_group_sz.restore();
+
+            // rax = k group idx, rdx = idx within group
+            xor_(rdx, rdx);
+            mov(rax, reg_grouped_k);
+            add(rax, rd);
+            add(rax, loc_idx);
+            div(reg_wei_scales_group_sz);
```
tczeszun, line 2395: consider replacing with bit shifts, at least for power of 2 dividend

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2475,11 +2670,19 @@ void jit_brgemm_kernel_t<Wmm>::gemm_microkernel(dim_t bd_block2,
             }
         }
 
-        if (brg.req_s8s8_compensation)
+        if (brg.req_s8s8_compensation) {
+            if (brg.is_ic_wei_scales) {
+                // vmm_inp_shift() corrupted in IC scales + compensation.
+                mov(reg_s8_input_shift, 128);
+                uni_vpbroadcastb(vmm_inp_shift(), reg_s8_input_shift.cvt8());
```
tczeszun: probably can do it outside of lambda/loop?

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -357,3 +357,29 @@
 --attr-zero-points=,wei:common:2,wei:per_oc:s4,wei:per_ocic:s4:192x1
 --attr-fpmath=strict:true
 12x4x576:12x576x192
+
+## Int8 with int4 weights
+### Upconversion int4->int8
+--reset
+--dt=u8:s4:u8,s8:s4:s8,u8:u4:u8
```
tczeszun: what about s8:u4:s8?

inteldimitrius: Added, thanks!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Yes, that's true, restoration logic is suboptimal. Moved out to separate lambda and it's applying under conditions now. Thanks!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: There're two motivations to use `div` instead of shifting bits:
1. Group size must be divisible by 32, but it's not guaranteed being power of 2. E.g. group size = 96 is valid.
2. The code below is using both `quant` and `rem` (`rax`, `rdx` correspondingly), so changing bit-shift-division will require changing the whole logic in this function.

tczeszun: I see, let's leave it as it is then

inteldimitrius:
make test

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -889,6 +889,27 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
+                && dnnl::impl::utils::one_of(
+                        prb->dst_dt(), dnnl_u8, dnnl_s8, dnnl_s32);
+        const auto &wei_scales = prb->attr.scales.get(DNNL_ARG_WEIGHTS);
+        const bool has_grouped_scales
+                = !wei_scales.is_def() && !wei_scales.groups.empty();
+        if (is_x8s4x8 && has_grouped_scales) {
+            BENCHDNN_PRINT(2,
+                    "[SKIP][%s:%d]: GPU doesn't support int8 source with "
+                    "int4 weights, integer destination and grouped "
+                    "scales.\n",
```
vpirogov, line 905: This does not sound right. Can you share benchdnn problem that fail?

inteldimitrius: I've got these cases failed: 
```
507: 10090:FAILED (90 ms) __REPRO: --mode-modifier=P --matmul --engine=gpu --dt=u8:s4:u8 --attr-scales=wei:per_ocic:f32:128x1 --attr-zero-points=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 128x256:256x16
507: 10102:FAILED (81 ms) __REPRO: --mode-modifier=P --matmul --engine=gpu --dt=s8:s4:s8 --attr-scales=wei:per_ocic:f32:128x1 --attr-zero-points=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 128x256:256x16
507: 10114:FAILED (85 ms) __REPRO: --mode-modifier=P --matmul --engine=gpu --dt=u8:u4:u8 --attr-scales=wei:per_ocic:f32:128x1 --attr-zero-points=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 128x256:256x16
507: 10126:FAILED (88 ms) __REPRO: --mode-modifier=P --matmul --engine=gpu --dt=s8:u4:s8 --attr-scales=wei:per_ocic:f32:128x1 --attr-zero-points=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 128x256:256x16
```
Testcases added in this PR, but fails when engine=gpu.
Failures are not reproducible on CPU.

densamoilov: Looks like a bug. Unsupported cases are supposed to return unimplemented.

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -807,7 +825,9 @@ int brgemm_cmp(const brgemm_desc_t &lhs, const brgemm_desc_t &rhs) {
     CMP_BRGEMM_FIELD(zp_type_c);
 
     CMP_BRGEMM_FIELD(skip_scales);
-    CMP_BRGEMM_FIELD(is_oc_scale);
+    CMP_BRGEMM_FIELD(is_single_wei_scale);
+    CMP_BRGEMM_FIELD(is_oc_wei_scales);
+    CMP_BRGEMM_FIELD(is_ic_wei_scales);
```
georgen117, line 830: Should this include wei_scale_k_group_size as well?

```cpp
CMP_BRGEMM_FIELD(wei_scale_k_group_size);
```

tests/benchdnn/matmul/matmul.cpp
dzarukin: This piece should be dropped, there are no configs supported by CPU that are not supported by GPU.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1458,11 +1487,19 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
         bgmmc.wei_scales_dt = wei_scales.get_data_type();
         bgmmc.wei_scales_dt_sz = types::data_type_size(bgmmc.wei_scales_dt);
         bgmmc.wei_scales_k_gsize = wei_scales.get_group(0);
 
+        // Due to hardware restrictions of AMX the effective group size should
+        // be divisible by tile-size 64
+        // Otherwise fallback to AVX512 kernels.
+        VCONDCHECK_BG(IMPLICATION(bgmmc.is_wei_scale_per_k && bgmmc.is_amx,
+                              bgmmc.wei_scales_k_gsize % 64 == 0),
+                VERBOSE_UNSUPPORTED_SCALES_CFG);
```
georgen117, line 1501: Should this validation check be moved after the check below that downgrades to per_oc?

Move this after line 1587.

```cpp
// this will change is_wei_scale_per_k to false when K_group >= K
if (bgmmc.is_wei_scale_per_k && !bgmmc.is_runtime_K
        && bgmmc.wei_scales_k_gsize >= bgmmc.K) {
    bgmmc.is_wei_scale_per_k = false;
}

// is_wei_scale_per_k could be changed to false leaving those shapes to still be run on AMX hw
// Any remaining bgmmc.is_wei_scale_per_k
VCONDCHECK_BG(IMPLICATION(bgmmc.is_wei_scale_per_k && bgmmc.is_amx,
                          bgmmc.wei_scales_k_gsize % 64 == 0),
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2937,6 +2947,304 @@ struct jit_brgemm_matmul_copy_b_int8_t : public jit_brgemm_matmul_copy_b_t,
...
+
+        mov(reg_tmp, 0xF0);
+        uni_vpbroadcastb(vmm_lut, reg_tmp.cvt8());
+
+        uni_vmovups(vmm_sign_extend, vmm_src);
+        vpslld(vmm_sign_extend, vmm_sign_extend, 4);
+
+        uni_vpxor(vmm_sign_extend, vmm_sign_extend, vmm_lut);
+        vpshufb(vmm_sign_extend, vmm_lut, vmm_sign_extend);
+        vpord(vmm_src, vmm_src, vmm_sign_extend);
```
georgen117, line 2979: The shared `vmm_zero` register is being used as a lookup table without restoring it back to zero.

Other code may expect the vmm_zero to be zero.

suggest adding code to explicity restore `vmm_zero` to zero. Similar to what was done for `cvt_int4_to_int8`.
```cpp
// Clean vmm_zero
uni_vpxor(vmm_zero, vmm_zero, vmm_zero);
```

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2295,6 +2362,124 @@ void jit_brgemm_kernel_t<Wmm>::dot_product(Vmm v1, Vmm v2, Vmm v3) {
...
+            uni_vaddps(acc, acc, tmp_acc);
+        }
+        jmp(dot_product_loop);
+        L_aligned(dot_product_loop_end, 64);
+
+        return true;
+    }
+    return false;
+}
+
```
georgen117, line 2482: `maybe_dot_product_with_ic_scales` emits an x86 `div` instruction inside `prepare_and_apply_mask` (line 2395) to compute the current scale group index at runtime. This lambda is called on every `(rd, ld, bd)` iteration of the innermost microkernel, which is significant inside a compute-saturated int8 GEMM kernel.

**I believe the division always yields quotient 0, making all this runtime bookkeeping unnecessary**. Here's why:

The blocking heuristics enforce `K_blk == wei_scales_k_gsize` (see `compute_blocking_heuristic` around line 1273 in `brgemm_matmul_utils.cpp` and `update_configuration` in `amx_blocking_heuristics.cpp` around line 69). Because of this, `maybe_set_ic_scales` in `brgemm_matmul.cpp` (line 854) always computes `k_start = k % wei_scales_k_gsize == 0`. So `reg_grouped_k` ( which is initialized from `k_start`), is always `0` at the start of each brgemm call. Since `rd < rd_block == K_blk == group_size`, the division `(0 + rd + loc_idx) / group_size` always yields quotient `0`.

Consequences:

The scale group index (`rax` after the `div`) is always `0`, the scales pointer row offset is always 0.
All `rd_step` bytes are always in the same group, so the byte-activity mask is always `0xFFFFFFFF`; the `AND` with `vmm_A_bcst` is a no-op.
The outer `dot_product_loop` (lines 2448–2475), which exists to handle mid-step group boundaries, always exits after exactly one iteration.
If this analysis is correct, the `div`, the outer JIT loop, and the `reg_global_k`/`reg_grouped_k`/`reg_wei_scales_group_sz` GPR save/restore chain (declared at lines 236–240, used across `read_params` at lines 958–961 and `bs_loop` at lines 2884–2920) could all be removed, leaving a much simpler straight-line sequence.

**Could you confirm whether the `k_start == 0` invariant always holds?** If so, simplifying this path would both improve performance and make the logic considerably easier to follow.

If the `k_start == 0` always holds true then the code can be significantly simplified.

[PR4783_remove_div_ic_scales.patch](https://github.com/user-attachments/files/26450094/PR4783_remove_div_ic_scales.patch)

I played around with the attached change which removes the `prepare_and_apply_mask` and `load_cales` lambda entirely.

I almost didn't post this because it is such a big difference I feel like there is something critical, I am missing.

If `k_start` is not always `0` then we could add a fast path for situations when it is zero. Which as best I can tell `k_start` will always be `0` for amx.

[PR4783_ic_scales_fast_path.patch](https://github.com/user-attachments/files/26450543/PR4783_ic_scales_fast_path.patch)

With both attached patchs I was able to run: 
```batch
benchdnn --mode=C --matmul --reset --skip-impl=ref --dt=u8:s4:u8,s8:s4:s8,u8:u4:u8,s8:u4:s8 --attr-scales="wei:common:2,wei:per_oc:f32,wei:per_ocic:f32:32x1" --attr-zero-points=",wei:common:2,wei:per_oc:s8,wei:per_ocic:s8:32x1" --attr-fpmath=strict:true 2x32:32x32

benchdnn --mode=C --matmul --batch=tests/benchdnn/inputs/matmul/harness_matmul_decompression
```
without issues.  I didn't get a chance to fully test the patches for all possible ISA.

I really hope this is helpful.

georgen117:
I ran the benchdnn with `DNNL_MAX_CPU_ISA=AVX512_CORE` and got a lot of failed cases. Indicating that somewhere is incorrect handling the max_isa.

```batch
DNNL_MAX_CPU_ISA=AVX512_CORE ./benchdnn --mode=C --matmul --batch=./inputs/matmul/harness_matmul_decompression
```

Summary:
```
2843:FAILED (errors:6 total:8) (16 ms) __REPRO: --matmul --dt=s8:s4:s8 1x1:1x8
2845:FAILED (errors:7 total:8) (14 ms) __REPRO: --matmul --dt=s8:u4:s8 1x1:1x8
2847:FAILED (errors:8 total:8) (25 ms) __REPRO: --matmul --dt=s8:s4:s8 1x16:16x8
2849:FAILED (errors:8 total:8) (24 ms) __REPRO: --matmul --dt=s8:u4:s8 1x16:16x8
2851:FAILED (errors:24 total:24) (60 ms) __REPRO: --matmul --dt=s8:s4:s8 3x65:65x8
2853:FAILED (errors:24 total:24) (58 ms) __REPRO: --matmul --dt=s8:u4:s8 3x65:65x8
2866:FAILED (errors:64 total:64) (30 ms) __REPRO: --matmul --dt=s8:s4:s8 --attr-scales=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 2x32:32x32
2867:FAILED (errors:64 total:64) (32 ms) __REPRO: --matmul --dt=s8:s4:s8 --attr-scales=wei:common:2 --attr-zero-points=wei:common:2 --attr-fpmath=strict:true --skip-impl=ref 2x32:32x32
2868:FAILED (errors:64 total:64) (28 ms) __REPRO: --matmul --dt=s8:s4:s8 --attr-scales=wei:common:2 --attr-zero-points=wei:per_oc:s8 --attr-fpmath=strict:true --skip-impl=ref 2x32:32x32
2869:FAILED (errors:64 total:64) (29 ms) __REPRO: --matmul --dt=s8:s4:s8 --attr-scales=wei:common:2 --attr-zero-points=wei:per_ocic:s8:32x1 --attr-fpmath=strict:true --skip-impl=ref 2x32:32x32
(... 60 more cases ...)
============================
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: In `cvt_int4_to_int8()` there're literally two last lines: 
```
signed_mask_int4(vmm_src);
// Clean vmm_zero
uni_vpxor(vmm_zero, vmm_zero, vmm_zero);
```
So it will generate double VPXOR instruction on vmm_zero.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Yes, you are right, performance-wise AMX fallback should be the last, I will move it, thanks!

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Actually, I don't mind to duplicate vpxor(vmm_zero) in `signed_mask_int4()` to make it more consistent. But it's quite a `hot` block of the kernel and could affect the performance (I didn't measure how bad it would be, tbh).

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Hi George, thanks for reviewing! 
Let me carefully analyze your suggestion and I will come back in replies.

src/cpu/x64/brgemm/brgemm.cpp
inteldimitrius: Added, thanks!

tests/benchdnn/matmul/matmul.cpp
inteldimitrius: I will remove this skip block and create a jira ticket for it.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Yes, it was really helpful. You are right, my solution is working with arbitrary K group size, however the group size must be divisible by 32 by design. So yeah, no masks needed since rd-step is entirely in one group. However, quant can be equal 1 for the cases e.g. Mx64:64:N, with the group size 32. This register is updating the offset between microkernel calls, so I think it definitely can be > 0. Maybe the problem is in variable naming...

This code branch doesn't depend on AMX, because it has it's own `maybe_apply_scales_amx()`, so K_blk here could be arbitrary. AMX can handle only "full-tile" group sizes % 64 == 0. 

I think it's really helpful and I will try to adapt you patches, but slightly in different way.
Thanks!

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -773,6 +784,81 @@ void brgemm_matmul_t<isa>::compute_kernel(
     brgmm_ctx.maybe_backup_dst_values_to_buffer(
             ithr, b_idx, m_blk_idx, n_blk_idx);
 
+    auto is_amx = [&](const int brg_idx) {
+        return is_superset(
+                pd()->get_brg_desc(brg_idx).isa_impl, avx512_core_amx);
+    };
```
georgen117, line 790: This `is_amx(int)` lambda is being shadowed by the `bool is_amx` at line 863 and again as line 896.

These would cause a compiler warning if the code was built with -Wshadow.

I would suggest renaming the lambda or rename the inner bools at lines 863 and 896.

The inner bools are using the same logic as the lambda so they could actually be updated to use the lambda.

i.e. change lines 863 and 896 to `bool brg_ker_is_amk = is_amx(brg_ker_idx); `

src/cpu/x64/brgemm/brgemm.hpp
```diff
@@ -246,6 +246,13 @@ void brgemm_kernel_execute(const brgemm_kernel_t *brg_kernel, int bs,
         void *scratch = nullptr,
         const brgemm_dynamic_values_t *dynamic_values = nullptr);
 
+/// Execute BRGEMM kernel with pre-set params
+/// @param brg_kernel BRGEMM kernel
+/// @param params Pre-set parameters for the BRGEMM kernel
+/// Useful for setting addtitional parameters, like IC scales or post-ops.
```
georgen117, line 252: nit: `addtitional` -> `additional`

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -322,7 +324,7 @@ brgemm_matmul_conf_utils_t::brgemm_matmul_conf_utils_t(
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::bf16,
                       bgmmc.src_dt == bf16)
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict,
-                      bgmmc.src_dt == f32)
+                      one_of(bgmmc.src_dt, f32, u8, s8))
               && attr.fpmath_.apply_to_int_)
```
georgen117, line 328: weights_decompression_support may eclude the new int8 path

`weights_decompression_support` gates on the `attr.fpmath_.apply_to_int_` which is an explicit user opt-in.  The new `u8/s8 x s4/u4` path does not require the user to set this flag. It is possible the weights_decompression_support could be `false` for int8 weights thanks to the `apply_to_int_` flag.

This affects bgmmc.with_wei_dcompression (line 1422), the `use_buffer_a` guard (line 1787), and `check_datatype_cfg` (line 289)

Suggestion:
Since `int8_with_quant_wei_dt` has its own initializer later in the list, I feel the cleanest fix is to OR the int8 condition directly into `weights_decompression_support`:
```cpp
weights_decompression_support(one_of(bgmmc.wei_dt, u8, s8, u4, s4)
              && one_of(attr.fpmath_.mode_, fpmath_mode::bf16, fpmath_mode::f16,
                      fpmath_mode::strict, fpmath_mode::any)
                      bgmmc.src_dt == f16)
                      bgmmc.src_dt == bf16)
              && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict,
                      one_of(bgmmc.src_dt, f32, u8, s8))
              && attr.fpmath_.apply_to_int_)
             // int8 x int 4 does not require fpmath opt-in
             || (one_of(bgmmc.src_dt, u8, s8) && one_of(bgmc.wei_dt, s4, u4)
```


The other existing `bf16_with_int_wei_dt`/`f16_with_int_wei_dt`/`f32_with_int_wei_dt` are unaffected their `src_dt == bf16/f16/f32` already eclude the u8/s8 case.

inteldimitrius:
Closed due to being part of bigger feature #4999.

PR #4786, opened 2026-03-08
benchdnn: parser: print similar options for unrecognized input option

dzarukin:
```
$ benchdnn --matmul --memory-check=0 32x64:64x32
Error: unknown option for 'matmul' driver: '--memory-check=0'.
The most similar option is:
        mem-check
        memory-kind
```

```
$ benchdnn --matmul --attr-postops=relu 32x64:64x32
Error: unknown option for 'matmul' driver: '--attr-postops=relu'.
The most similar option is:
        attr-post-ops
```

Additionally, included a fix for [MFDNN-14741](https://jira.devtools.intel.com/browse/MFDNN-14741) as it's a "one-liner".

### DIFF
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -651,6 +651,14 @@ void notify_gpu_profiling_complete(dnnl_stream_t stream) {
 
 void finalize() {
     finalize_tbb();
+
+#if DNNL_CPU_THREADING_RUNTIME == DNNL_RUNTIME_THREADPOOL
+    // See: DUMMY_PARALLEL.
+    if (is_cpu()) {
+        auto tp = dnnl::testing::get_threadpool();
+        tp->parallel_for(1, [](int, int) {});
+    }
+#endif
 }
 
 inline int measure_perf_individual(timer::timer_t &t, dnnl_stream_t stream,


tests/benchdnn/utils/parser.cpp
```diff
@@ -1776,9 +1827,35 @@ void catch_unknown_options(const char *str) {
...
+
+        BENCHDNN_PRINT(0, "Error: unknown option for \'%s\' driver: \'%s\'.\n",
+                driver_name.c_str(), str);
+
+        // Note: print options when there's some similarity, but limit it by 6
+        // (taken experimentally) for long given options.
+        size_t good_dist = std::min(s.size() - 1, size_t(6));
+        if (opt_distances.begin()->first >= good_dist) std::abort();
+
+        BENCHDNN_PRINT(0, "%s\n", "The most similar option is:");
```
avmanerikar: (nit): Could be replaced by the standard:
```
Did you mean?
     mem-check
     memory-kind
```

tests/benchdnn/utils/parser.cpp
```diff
@@ -1776,9 +1827,35 @@ void catch_unknown_options(const char *str) {
...
-                "driver: ERROR: unknown option:", driver_name.c_str(), str);
-        exit(2);
+        std::string s(str);
+        auto equal_pos = s.find_first_of('=');
+        s = s.substr(2, equal_pos - 2);
+
+        // Try to provide similar options in case of minor typos.
+        // The search is based on Levenshtein distance between two strings.
+        std::multimap<size_t, std::string> opt_distances;
+        for (const auto &opt : parser_utils::get_option_names()) {
```
avmanerikar, line 1844: IIUC that would be printing the similar options for the first incorrect option and not for multiple typos in a row? (for example, `--egnine=.. --attr-psot-ops=.. --bathc=.. ` will only print the options for `engine`)

tests/benchdnn/utils/parser.cpp
```diff
@@ -1776,9 +1827,35 @@ void catch_unknown_options(const char *str) {
...
+            opt_distances.emplace(dist, opt);
+        }
+
+        BENCHDNN_PRINT(0, "Error: unknown option for \'%s\' driver: \'%s\'.\n",
+                driver_name.c_str(), str);
+
+        // Note: print options when there's some similarity, but limit it by 6
+        // (taken experimentally) for long given options.
+        size_t good_dist = std::min(s.size() - 1, size_t(6));
+        if (opt_distances.begin()->first >= good_dist) std::abort();
```
mzhukova: could we end up with an empty map here? (so that doing `->first` would be an issue)

tests/benchdnn/utils/parser.cpp
```diff
@@ -1776,9 +1827,35 @@ void catch_unknown_options(const char *str) {
...
+
+        BENCHDNN_PRINT(0, "%s\n", "The most similar option is:");
+
+        for (const auto &e : opt_distances) {
+            if (e.first >= good_dist) continue;
+            BENCHDNN_PRINT(0, "        %s\n", e.second.c_str());
+        }
+
+        std::abort();
```
mzhukova: just wondering why `std::abort` and not regular `exit`?

tests/benchdnn/utils/parser.cpp
```diff
@@ -117,6 +121,53 @@ float stof_safe(const std::string &s) {
     return value;
 }
 
+// Computes Levenshtein distance between two strings, which indicated how many
+// symbols should be appended, replaced, or removed to obtain the input string.
+// The implementation follows this article:
+// https://en.wikipedia.org/wiki/Levenshtein_distance
+//
+// The function is designed to be called recursively, decresing symbols in `in`
```
mzhukova:
```suggestion
// The function is designed to be called recursively, decreasing symbols in `in`
```

tests/benchdnn/utils/parser.cpp
```diff
@@ -117,6 +121,53 @@ float stof_safe(const std::string &s) {
...
+size_t compute_distance(const std::string &in, const std::string &opt,
+        bool reset_cache = false) {
+    if (in.empty()) return opt.size();
+    if (opt.empty()) return in.size();
+    if (in.front() == opt.front())
+        return compute_distance(in.substr(1), opt.substr(1));
+
+    static std::map<std::pair<std::string, std::string>, size_t> cache;
+    if (reset_cache) cache.clear();
+    std::map<std::pair<std::string, std::string>, size_t>::iterator it {};
```
mzhukova, line 146: add `<map>` to includes?

tests/benchdnn/utils/parser.cpp
dzarukin: I took the message from `git`. Where's the proposed standard coming from (for my education purpose)?

tests/benchdnn/utils/parser.cpp
dzarukin: The parser would spit an error on the first incorrect option by design (as it's a state machine), so this is an expected behavior.

tests/benchdnn/utils/parser.cpp
dzarukin: It's not anticipated as the can't be no options in the list. I can add a check just for the sake of safe code practice.

tests/benchdnn/utils/parser.cpp
dzarukin: I want to switch from exit to abort all over the places as it doesn't trigger global objects check such as clearing memory or not emptied task_executor list. Additionally, it provides a convenient ability to debug the exact spot of the problem instantly over finding it with a break and triggering it.

tests/benchdnn/utils/parser.cpp
avmanerikar: Not a standard but a common convention for CLI, git command suggestions: https://github.com/git//git/blob/master/help.c - L880.

dzarukin:
```
$ git stotus
git: 'stotus' is not a git command. See 'git --help'.

The most similar command is
        status
$ git --version
git version 2.40.1
```

echeresh: Nit:
```suggestion
        BENCHDNN_PRINT(0, "%s\n", "The most similar options are:");
```

tests/benchdnn/utils/parser.cpp
echeresh: Do you know if it's hard to fix the issues with global objects? `std::abort` on incorrect user input is not a common practice plus it may produce undesired crash dumps.

Looking at benchdnn, we have `finalize()` call - can it serve as an entry point for all global-state validation?

dzarukin: I didn't consider core dumps as a negative side effect of abort.
I guess, I'll have to go through a conventional way of checking the failure state instead, removed `std::abort()`, and thanks for the discussion.

tests/benchdnn/utils/parser.cpp
dzarukin: Removed this map entry dereference line due to other comments. Thanks.

tests/benchdnn/utils/parser.cpp
dzarukin: Updated, thanks.

tests/benchdnn/utils/parser.cpp
dzarukin: Guess it comes from other dependent headers. Added it and multimap, too. Thanks.

tests/benchdnn/utils/parser.cpp
dzarukin: Added a fork to report a single version or plural option depending on the number of candidates. Thanks.

dzarukin:
make test
set test_scope=SMOKE

dzarukin:
make test

mzhukova (approved):

PR #4498, opened 2026-01-06
fused sdpa training

syurkevi:
# Description

This PR implements fused sdpa training. Supported features include:
* supported forward stats output (logsumexp)
* supported backward training for fp16, bf16, f32
* supported GQA forward/backwards
* support for all attention masks (1d, 2d, causal_tl,br)
* optional output gradient for attention mask
* host side scalar support
* transpose_k inputs

A follow up PR will add dropout support forwards and backwards.

Current performance is very much problem size dependent but ranges from ~30% speedups (non-masked, non-grouped) to 4x speedups (grouped, causal). Head sizes are currently limited to 128. Slightly more tuning to follow shortly.

PR #4787, opened 2026-03-08
tests: gtests: internals: overhaul gated_mlp random value generation

hidefromkgb:
1. Gated MLP primitive now supports BF16 and FP32
1. Random values in internal GMLP tests are now moved around in a less inelegant fashion
1. Additionally resolves [MFDNN-14677](https://jira.devtools.intel.com/browse/MFDNN-14677)

### DIFF
--- a/tests/gtests/internals/CMakeLists.txt
+++ b/tests/gtests/internals/CMakeLists.txt
@@ -50,4 +50,11 @@ register_exe(${TEST_EXE}_env_vars_onednn
         "test" "dnnl_gtest")
 list(REMOVE_ITEM TEST_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/test_env_vars_onednn.cpp)
 
+# Register GMLP tests as a separate executable
+register_exe(${TEST_EXE}_gmlp
+    "${MAIN_SRC_GTEST};${CMAKE_CURRENT_SOURCE_DIR}/test_gated_mlp.cpp;${CMAKE_CURRENT_SOURCE_DIR}/test_utils.cpp"
+        "test" "dnnl_gtest")
+
+list(REMOVE_ITEM TEST_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/test_gated_mlp.cpp)
+
 register_exe(${TEST_EXE} "${TEST_SOURCES}" "test" "dnnl_gtest")


tests/gtests/internals/test_gated_mlp.cpp
```diff
@@ -133,32 +136,34 @@ std::ostream &operator<<(std::ostream &ss, const dnnl_alg_kind_t &act) {
...
+        ss << "_gu_group_size_" << p.gateup_group_size;
+        ss << "_gd_group_size_" << p.down_group_size;
+    }
+
+    ss << "_src_dt_" << dnnl_dt2str(memory::convert_to_c(p.src_dt));
+    ss << "_dst_dt_" << dnnl_dt2str(memory::convert_to_c(p.dst_dt));
 
     ss << "_wgu_wt_" << dnnl_dt2str(memory::convert_to_c(p.wgu_wt));
-    if (p.wgu_wt != mdt::f16) {
+    if (has_quant) {
```
syurkevi, line 156: 👍

tests/gtests/internals/test_gated_mlp.cpp
```diff
@@ -169,152 +174,30 @@ std::string PrintToString(const ::testing::TestParamInfo<mlp_dims_t> &info) {
...
-            break;
-        case memory::data_type::f32:
-            write_to_dnnl_memory_common<float>(data, mem);
-            break;
-        case memory::data_type::f16:
-            write_to_dnnl_memory_common<float16_t>(data, mem);
-            break;
-        default: assert(false);
+template <typename T>
+void move_data(std::vector<T> &v, memory &mem, bool to_mem = true) {
```
syurkevi: How does this differ from write_to_dnnl_memory in test_utils.hpp (not the local one)? Can the two be combined?

.github/CODEOWNERS
```diff
@@ -39,7 +39,11 @@
 /tests/benchdnn/graph/ @uxlfoundation/onednn-graph @uxlfoundation/onednn-arch
 /tests/benchdnn/inputs/graph/ @uxlfoundation/onednn-graph @uxlfoundation/onednn-arch
 /tests/gtests/graph/ @uxlfoundation/onednn-graph
+## Special case of internal tests
 /tests/gtests/internals/test_sdpa.cpp @uxlfoundation/onednn-gpu-intel
+/tests/gtests/internals/test_utils.cpp @uxlfoundation/onednn-gpu-intel
+/tests/gtests/internals/test_utils.hpp @uxlfoundation/onednn-gpu-intel
+/tests/gtests/internals/test_gated_mlp.cpp @uxlfoundation/onednn-gpu-intel
```
syurkevi, line 46: Due to potential timeouts with (planned) increased testing, can we also make a separate executable for test_internals_gmlp?

tests/gtests/internals/test_gated_mlp.cpp
syurkevi: Can we also remove these? If they are useful for debug let's add them to test_utils.

tests/gtests/internals/test_gated_mlp.cpp
```diff
@@ -1138,7 +922,7 @@ TEST_P(mlp_test_t, compare) {
     for (int i = 0; i < int(resih.size()); ++i) {
         float abs_diff = std::abs(resih[i] - resph[i]);
         float rel_diff = std::abs((resih[i] - resph[i]) / resih[i]);
-        if (abs_diff > 1e-4 && rel_diff > 5e-3) {
+        if (abs_diff > 1e-4 && rel_diff > 5e-2) {
```
syurkevi, line 860: Do bounds differ much for f32/f16/quantized?

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: Using `fill_hceye()` and `fill_const()` right now.
Will move them to test_utils.

src/gpu/intel/gated_mlp/ref.hpp
```diff
@@ -94,17 +96,40 @@ struct ref_t : public primitive_t {
...
+                case data_type::u8:
+                case data_type::s8:
+                    dt = arg_md(DNNL_ARG_SRC)->data_type;
+                    if (!utils::one_of(dt, data_type::bf16, data_type::f16))
+                        dt = data_type::f16;
+                    break;
+                default: return status::unimplemented;
+            }
+            return get_gate_up_dst_md(retn, dt);
+        }
```
Simonsays095, line 116: Perhaps a `utils::one_of` guard check would be easier to understand here, than a switch with a default failure? Also, what's the logic of converting to f16/bf16 always? Is that an API requirement?

tests/gtests/internals/test_gated_mlp.cpp
```diff
@@ -536,77 +436,59 @@ gmlp_tensors_t get_descriptors(engine &eng, mlp_dims_t p) {
...
-        if (wgu_zp_dt == mdt::s4 || wgu_zp_dt == mdt::s8) {
-            printf("s4/s8 quant init\n");
-            w_gate_data = dequantize(w_gate_quantized_data, w_gate_md,
-                    w_gate_scales_md, w_gate_zp_data_signed, w_gate_scales_data,
-                    wgu_group_size, p.qtype, out.wgu_groups, 0);
+        bool wgu_zp_unsigned = (wgu_zp_dt == mdt::u4 || wgu_zp_dt == mdt::u8);
+        if (wgu_zp_unsigned)
+            printf("unsigned gateup quant init\n");
+        else
+            printf("signed gateup quant init\n");
```
Simonsays095: Should we be directly printing out messages like this?

tests/gtests/internals/test_gated_mlp.cpp
Simonsays095: This function no longer handles int4 data types, is that a problem?

hidefromkgb: It does, just not as an std::vector type.
There's no point in supporting `std::vector<int4_t>` since the source data can be passed as `int` or `float`, and the destination data doesn't need to be packed.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: Right now we don't need to.
It used to be convenient at first, when I ran the CI on multiple machines and watched it fail, but now it can safely be hidden.

src/gpu/intel/gated_mlp/ref.hpp
hidefromkgb: This isn't an API requirement, it just reflects the need for the Down GEMM to run fast, via DPAS preferably, but avoid losing too much precision.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: It very much depends on the RNG seed — but I've seen crazy deltas, both absolute and relative, that couldn't be anything but a result of the different computation order.

.github/CODEOWNERS
hidefromkgb: I believe the better option is to add both SDPA and GMLP to benchdnn.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: I believe `move_data` can supersede `write_to_dnnl_memory`: it should be faster because it doesn't duplicate the CPU buffer, and it's (almost) type-agnostic — the only exception being `std::vector<unsigned>`, as oneDNN doesn't have a corresponding `memory::desc::data_type` for it.

src/gpu/intel/gated_mlp/ref.hpp
Simonsays095: Surely this is selected by the user, right? I'm not familiar with the gated mlp API, but I'd guess that the user submits an md that defines the data type - especially since they're in charge of filling the data.

hidefromkgb: The data types for intermediate matrices are chosen implicitly, without asking the user.
With Up it's not that much of a problem, since its destination type is hardcoded as FP32, but with GateUp it's more complex: the resulting intermediate matrix needs to be either A or B for the Down GEMM.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: Moved to test_utils.hpp, added a TODO.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: Done.

tests/gtests/internals/test_gated_mlp.cpp
hidefromkgb: Moved all `printf()`'s and `std::cout`'s under `if (verbose)`.

.github/CODEOWNERS
syurkevi: I'm seeing timeouts on xe_lp as-is even when [separating ](https://github.com/uxlfoundation/oneDNN/pull/4498/changes#diff-18cf1e096f9e04e034c041b6e8f6d51b27b21879771eb955dad417c4cdebec47)out all sdpa tests from [test_internals](https://ecmd.jf.intel.com/commander/link/workspaceFile/workspaces/JF-COMMON-CI?jobStepId=f11e0b55-cedd-f1b5-b870-a4bf010d0e2d&fileName=gtest.f11e0b55-cedd-f1b5-b870-a4bf010d0e2d.log&jobName=CI_oneDNN_2026-03-12-07%3A11%3A08-syurkevi_fused_sdpa_training-328bef9&jobId=f11de2a4-8722-f188-9abb-a4bf010d0e2d&diagCharEncoding=&resourceName=sdpdnn712446&completed=1). Looking forward to benchdnn support as well!

hidefromkgb: It's in a separate executable now.
Granted, this is a temporary stopgap, but until benchdnn gains GMLP support it'll do.

dzarukin:
Just a general late comment - I see that filling is sequential while it can be done in parallel speeding up stuff a lot.
You may check https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/dnnl_test_common.hpp#L452 as an example how it can be done.

PR #4793, opened 2026-03-09
xe: lnorm: use uniform pointer with block read

echeresh:
`intel_sub_group_block_read()` must be called with a uniform pointer across the subgroup:
https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups.html#_add_a_new_section_6_13_x_sub_group_read_and_write_functions

LNorm kernel has this adjustment for `dst_vect` but it's missing for shift/scale.

### DIFF
--- a/src/gpu/intel/lnorm/reusable_vectorized.cl
+++ b/src/gpu/intel/lnorm/reusable_vectorized.cl
@@ -75,10 +75,10 @@ lnorm_reusable_vectorized(__global SRC_DATA_T *src, __global float *mean,
 
     if (USE_SCALE)
         scale = GWS_GET_BUFFER_POS(SS, gws_params, scale)
-                + ((greads - 1) * SG_STRIDE);
+                - get_sub_group_local_id() + ((greads - 1) * SG_STRIDE);
     if (USE_SHIFT)
         shift = GWS_GET_BUFFER_POS(SS, gws_params, shift)
-                + ((greads - 1) * SG_STRIDE);
+                - get_sub_group_local_id() + ((greads - 1) * SG_STRIDE);
 
     /// Normalize layer
     FLT_ACC_DATA_T sqrt_variance = rsqrt(local_variance + eps);


echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_lnorm
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

kealan-barbieri:
@echeresh is the issue that some of the work-items wont hit this?

echeresh:
> @echeresh is the issue that some of the work-items wont hit this?

No, the issue is in the pointers, see this OpenCL dump:

```c++
   // greads is fixed, passed to the kernel as an argument.
   if (1) scale = &scale[(get_global_id(0))] + ((greads - 1) * 64);
   // ...
        if (1)
            res *= cvt_bf16_to_f32(as_ushort2(intel_sub_group_block_read_us2(
                    (const __global ushort *)(scale))));
```

What happens is:

- `get_global_id(0)` varies between items of a subgroup
- `scale` pointer is different between subgroup items
- `intel_sub_group_block_read_us2()` is called with non-uniform pointer values

This contradicts the requirement (see the link in the description):

> Furthermore, since these are block operations, the pointer, image, and coordinate arguments to these built-in functions must be the same for all work items in the sub-group

src/gpu/intel/lnorm/reusable_vectorized.cl
```diff
@@ -75,10 +75,10 @@ lnorm_reusable_vectorized(__global SRC_DATA_T *src, __global float *mean,
 
     if (USE_SCALE)
         scale = GWS_GET_BUFFER_POS(SS, gws_params, scale)
-                + ((greads - 1) * SG_STRIDE);
+                - get_sub_group_local_id() + ((greads - 1) * SG_STRIDE);
     if (USE_SHIFT)
         shift = GWS_GET_BUFFER_POS(SS, gws_params, shift)
-                + ((greads - 1) * SG_STRIDE);
+                - get_sub_group_local_id() + ((greads - 1) * SG_STRIDE);
```
rjoursler, line 81: Would it make sense to add a new interface `GWS_GET_SUBGROUP_BUFFER_POS` to avoid implementing fixup logic whenever blocked load/store is required? Also, does the spec actually guarantee that `global_id - sub_group_local_id` is scalar? To really guarantee this, I would think we need to use `get_group_id()` directly.

echeresh: > Would it make sense to add a new interface GWS_GET_SUBGROUP_BUFFER_POS to avoid implementing fixup logic whenever blocked load/store is required?

After a quick check, this file is the only place that uses `GWS_GET_BUFFER_POS` together with block reads. This fix is also aligned with the adjustment done for `dst_vect`.

Considering that layer normalization is rarely updated, I suggest to keep the fix as is, more local, without adding new macros.

> Also, does the spec actually guarantee that global_id - sub_group_local_id is scalar?

In general no, but this is the assumption of this kernel. ND range is created so that LWS is equal to the subgroup size so that subtraction fixes the divergent pointer to make it subgroup-uniform.

PR #4795, opened 2026-03-09
xe: grp_gemm: Fix xe2 failures by adding new grp_gemm configurations. Fixed named barrier condition

umar456:
# Description

This PR fixes failures in grouped gemm on Xe2 that were caused by the reuse of bf16 kernels from PVC. This caused several tests cases to freeze during testing on Xe2HPG GPUs. This PR also addresses an incorrect condition to exclude named barriers from being candidates because they are not supported by the microkernel shim function.

PR #4796, opened 2026-03-09
[backport v3.12-pc]xe: grp_gemm: Fix xe2 failures by adding new grp_gemm configurations. Fixed named barrier condition

umar456:
# Description

Backport of #4795

This PR fixes failures in grouped gemm on Xe2 that were caused by the reuse of bf16 kernels from PVC. This caused several tests cases to freeze during testing on Xe2HPG GPUs. This PR also addresses an incorrect condition to exclude named barriers from being candidates because they are not supported by the microkernel shim function.


umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

PR #4797, opened 2026-03-09
cpu: x64: mm_ip: add wtag=ba support

dzarukin:
[MFDNN-14744](https://jira.devtools.intel.com/browse/MFDNN-14744)

When user passes wtag=ba, the impl rejected it though doesn't seem to have issues to handle that properly on forward.

Extended validation with wtag=ba cases.

### DIFF
--- a/src/cpu/x64/matmul_inner_product.cpp
+++ b/src/cpu/x64/matmul_inner_product.cpp
@@ -83,7 +83,7 @@ status_t init_matmul_md(memory_desc_t &mm_md, const memory_desc_t &ip_md,
 
 static bool check_training_formats(const memory_desc_wrapper &src_d,
         const memory_desc_wrapper &wei_d, const memory_desc_wrapper &bias_d,
-        const memory_desc_wrapper &dst_d) {
+        const memory_desc_wrapper &dst_d, prop_kind_t prop_kind) {
     using namespace format_tag;
     using namespace utils;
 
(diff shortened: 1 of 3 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_ip

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -206,6 +212,11 @@ status_t matmul_inner_product_fwd_t::pd_t::init_matmul_params(
                 break;
             }
         }
+        // Special case: `ba` input format.
+        if (mm_wei_tag == format_tag::undef
```
densamoilov: Is there any issue with extending the table with supported formats? I think that would be the proper way to support it.

dzarukin: The problem is to set a key for a new tag. Duplicated key is not functional in `std::map`, extending the vector with another 2D tag is not possible without modification rest of the code.
Feel free to guide with a better solution, I just did it to address it faster.

densamoilov: I'm not sure I'm following you.
We could add `{ ab, {ba, bca, bcda, bcdea}}` to the table and extend checks. This should do.

dzarukin: But we don't want `ab` when the input is `ba`, we want to keep `ba` still for better performance...

densamoilov: Why do you think performance should be better for `ba`? Those are plain layouts so we're supposed to use copy routines in both cases.

And a side question, are you sure the case is tested? It seems weird that we don't care about matmul weights layout for dlrm shapes. Could it be that `--fast-ref=true` option causes false negatives?

dzarukin: Discussed offline, figured out the problem, addressed with a new change.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_ip

src/cpu/x64/matmul_inner_product.cpp
```diff
@@ -92,17 +92,25 @@ static bool check_training_formats(const memory_desc_wrapper &src_d,
...
     if (!bias_d.is_zero()) ok = ok && bias_d.matches_tag(x);
 
-    ok = ok && IMPLICATION(src_d.matches_tag(ab), wei_d.matches_tag(ab))
-            && IMPLICATION(src_d.matches_tag(acb), wei_d.matches_tag(acb))
-            && IMPLICATION(src_d.matches_tag(acdb), wei_d.matches_tag(acdb))
-            && IMPLICATION(src_d.matches_tag(acdeb), wei_d.matches_tag(acdeb));
+    auto tag = utils::pick(src_d.ndims() - 2, ab, acb, acdb, acdeb);
+    bool tag_matches
+            = IMPLICATION(src_d.matches_tag(tag), wei_d.matches_tag(tag));
+    // Note: allow forward_training weights' OC to be the densiest.
```
densamoilov:
```suggestion
    // Note: allow forward_training weights' OC to be the innermost.
```

src/cpu/x64/matmul_inner_product.hpp
```diff
@@ -72,12 +72,10 @@ struct matmul_inner_product_fwd_t : public primitive_t {
...
                     VERBOSE_UNSUPPORTED_ATTR);
 
-            if (get_prop_kind() == prop_kind::forward_training) {
-                VDISPATCH_INNER_PRODUCT_SC(
-                        set_training_formats(
-                                &src_md_, &weights_md_, &bias_md_, &dst_md_),
-                        VERBOSE_UNSUPPORTED_TAG);
-            }
+            VDISPATCH_INNER_PRODUCT_SC(
+                    set_training_formats(&src_md_, &weights_md_, &bias_md_,
```
densamoilov: (nit) i think checking prop kind before calling this function would be cleaner but it's up to you

src/cpu/x64/matmul_inner_product.cpp
dzarukin: Updated, thanks.

src/cpu/x64/matmul_inner_product.hpp
dzarukin: Restored the natural order of things, NP ;)

PR #4801, opened 2026-03-10
xe: gemm: jit: use 2d for grouped M/N scale

kealan-barbieri:
# Description

Force GEMM scales with M/N group to be treated as 2D. 2D handling is required to handle group offsets internally even if scales only have one non-trivial dim. Also prevent such scales from being converted to post-ops as post-op handling does not properly broadcast values for grouped scales. 

Fixes # [MFDNN-14740](https://jira.devtools.intel.com/browse/MFDNN-14740)
# Checklist

## General

### Bug fixes


src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -127,12 +132,13 @@ status_t pd_t::init_post_ops() {
...
 
     auto maybe_convert_scales_to_postop
-            = [this](const memory_desc_t &scale_md, int arg, data_type_t dt,
-                      bool mx, bool &converted) -> status_t {
+            = [this](const memory_desc_t &scale_md, int arg,
+                     int scale_ndims, bool mx, bool &converted) -> status_t {
         auto ndims = desc()->c_desc.ndims;
         // Scales on A/B can be converted to postops if
-        // the scales md has K=1
+        // the scales md has K=1 and M/N is not bcast.
```
Simonsays095, line 145: Have you identified the root cause of this bug in gemmstone? Could we just fix the bug instead of restricting cases to a less optimized path?

kealan-barbieri: Its not clear to me that dispatching these cases to post-ops is generally better for perf. For example the case this issue was reported against is a weights decompression, where I expect scaling during up-conversion to be better than post-ops. I'll run perf to check if this will change dispatching for any existing cases. 

I think the more in-depth fix for this would be to separate the application of non-trivial groups from quant dimensions. This would be a larger and more risky refactor for not much gain.

kealan-barbieri:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/pd.cpp
Simonsays095: When possible, it's usually better to convert to post-ops. Scaling during upconversion can be close-to-zero overhead in some cases (where we have optimized upconversion routines) but in general requires injecting a bunch of vector operations before dpas - inside the k-loop - which can be slow. Applying as a post-op does the scaling once on the accumulation registers, outside of the k-loop. That's why we have so far unconditionally converted to post-ops when doing so is theoretically possible.

Edit: even in cases where we have optimized upconversion+dequantization, the additional load instruction each iteration of the k-loop can have unintended consequences for performance (in some cases, acting effectively as a barrier while the threads load the data), which makes kernel performance hard to predict.

kealan-barbieri: That makes sense, I can see how that could improve performance, however looking at `post_ops.cxx` I dont think its right to characterize it as an internal bug, we just dont have support for handling the grouped offset/broadcast on this path. It sounds like this would be a pretty achievable feature to implement if we encounter KPI cases that use this path.  I'm reaching out to the submitter to check if they can share KPIs that will hit this case.

Simonsays095: That's kind of what I was trying to get at - is this something that we could just as easily fix in gemmstone (and allow the optimized path), or is it something that would take a bit of work to integrate. In theory, it should just essentially require something like `c -> c / xqGroupMN` to pick out the correct layout element, but if there's something more fundamental that would require effort to get right, I'm ok with this as a band-aid.

kealan-barbieri: I see, it doesnt look like a trivial implementation at the moment there are multiple paths for post-ops scales and they dont perform broadcast currently except for common scales, we'll need to refactor transforming scales to post-ops in order to incorporate the group into the post-op stride.

Simonsays095: Thanks Kealan, looks good.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

kealan-barbieri:
[ORPHAN PERF](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f11cc6de-813d-f154-8d41-a4bf010d0e2d?objectId=job-f11cc6de-813d-f154-8d41-a4bf010d0e2d&filterName0=jobsPageSearch&filterDepth=1)

syurkevi (approved):

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -45,11 +45,18 @@ int quant_entry_ndims(
     for (int i = qmd.ndims - 2; i < qmd.ndims; ++i) {
         if (qmd.dims[i] > 1) { count++; }
     }
+    int gcount = 0;
```
dzarukin, line 58: Looks like if `count == 0`, a fast exit right here would simplify the following code:
```
if (count == 0) return 0;
// for gemmstone, 1D quantization implies a full column vector
// (i.e. not on the K dimension). If quantization varies over K,
if (k_idx >= 0 && qmd.dims[k_idx] > 1) return 2;

<get_gcount>
if (gcount == 2) return 2;

return 1;
```

kealan-barbieri: done, thanks!

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

kealan-barbieri:
[Orphan CI](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f11e2f46-bfff-f12c-a0f3-a4bf010d0e2d?objectId=job-f11e2f46-bfff-f12c-a0f3-a4bf010d0e2d&filterName0=jobsPageSearch&filterDepth=1)

PR #4806, opened 2026-03-10
K-grouped f16 scales support for int SRC for grouped gemm in ref, benchdnn

mzhukova:
This PR should cover the grouped scales requested in [MFDNN-14729](https://jira.devtools.intel.com/browse/MFDNN-14729) for ref and benchdnn (inputs addition).

### DIFF
--- a/doc/primitives/matmul.md
+++ b/doc/primitives/matmul.md
@@ -378,11 +378,19 @@ auto dst_mem = memory(dst_md, engine, {dst_data, offsets.data()});
 
 Setting attributes for grouped GEMM follows the regular matmul attribute API.
 Below are some examples of common use cases for MoE workloads.
+For more details on how to set attributes, refer to the @ref dev_guide_attributes page.
 
 Per-token source scales:
 ~~~cpp
 attr.set_scales_mask(DNNL_ARG_SRC, (1 << 0));  // Varies along M dimension
 // Scale tensor: [total_tokens] - one scale per token
+// Layout: concatenated like source data, uses same offsets
+~~~
+
+K-grouped source scales with group size of 128:
+~~~cpp
+attr.set_scales(DNNL_ARG_SRC, (1 << 0) | (1 << 1), {1, 128}, memory::data_type::f16);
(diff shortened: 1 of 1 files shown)


mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

doc/primitives/matmul.md
```diff
@@ -378,6 +378,7 @@ auto dst_mem = memory(dst_md, engine, {dst_data, offsets.data()});
 
 Setting attributes for grouped GEMM follows the regular matmul attribute API.
 Below are some examples of common use cases for MoE workloads.
+For more details on how to set attributes, please refer to the @ref dev_guide_attributes page.
```
dzarukin: (Minor) We don't seem to use "please" much:
```
$ grep --color=always -ri "please" doc/ | wc -l
13
```
thus,
```suggestion
For more details on how to set attributes, refer to the @ref dev_guide_attributes page.
```

doc/primitives/matmul.md
```diff
@@ -407,7 +415,10 @@ The following are supported:
 - Currently, only single dimension `0` can vary.
 - Source and destination must use identical grouping.
 - Scales attribute for source and weights tensors:
-  - Source Scales: row-wise (per-token, `mask = (1 << 0)`) are applied to all experts equally.
+  - Source Scales: row-wise (per-token, `mask = (1 << 0)`) and K-grouped
+    (per-token-per-K-block, `mask = (1 << 0) | (1 << 1)`) with group specification
```
dzarukin: I don't recall `per-token-per-K-block` term used officially. Is it used in other documentary sources?
If not, maybe just drop it (to avoid new terms generation)?

mzhukova: I just wanted to make it clear what it means in the context of MoE/grouped gemm, but I could drop if that is confusing

doc/primitives/matmul.md
mzhukova: updated

doc/primitives/matmul.md
mzhukova: dropped here and for wei

PR #4807, opened 2026-03-10
cpu: aarch64: optimize n_bcast_1_load path in jit_brgemm_kernel_t

jondea:
# Description

Enable the `n_bcast_1_load` path for 128 bit SVE, and optimize by loading quadwords (`ld1rqw`) and using indexed FMLA/DOTs, e.g. for a `1024x1024:1024x1024` matmul

```
ld1rqw  {z0.s}, p0/z, [x27]
add     x28, x27, x17
ld1rqw  {z1.s}, p0/z, [x28]
add     x28, x27, x17, lsl #1
ld1rqw  {z2.s}, p0/z, [x28]
mov     x26, #0x3000                    // #12288
add     x28, x27, x26
ld1rqw  {z3.s}, p0/z, [x28]
add     x28, x27, x17, lsl #2
ld1rqw  {z4.s}, p0/z, [x28]
mov     x26, #0x5000                    // #20480
add     x28, x27, x26
ld1rqw  {z5.s}, p0/z, [x28]
ld1w    {z6.s}, p0/z, [x10]
fmla    z31.s, z6.s, z0.s[0]
fmla    z27.s, z6.s, z1.s[0]
fmla    z23.s, z6.s, z2.s[0]
fmla    z19.s, z6.s, z3.s[0]
fmla    z15.s, z6.s, z4.s[0]
fmla    z11.s, z6.s, z5.s[0]
ld1w    {z6.s}, p0/z, [x10, #1, mul vl]
fmla    z30.s, z6.s, z0.s[0]
fmla    z26.s, z6.s, z1.s[0]
...
fmla    z12.s, z6.s, z4.s[0]
fmla    z8.s, z6.s, z5.s[0]
addvl   x16, x10, #16
ld1w    {z6.s}, p0/z, [x16]
fmla    z31.s, z6.s, z0.s[1]
fmla    z27.s, z6.s, z1.s[1]
...
```

This reduces the number of left hand side loads by a quarter. Using this path leads to 10-20% speedups on Neoverse V2 for BRGEMM based implementations(see tables below), but performance seemed to be similar to the old path on Neoverse V1, so I left it as-is. We could revisit this and add a heurisitc, but for now this keeps things simple.

Plus some other small changes to the iteration, e.g. storing the stride to remove movs

Before
```
fmla    z28.s, p0/m, z0.s, z4.s
mov     x23, #0x1000                    // #4096
add     x25, x25, x23
ld1rw   {z0.s}, p0/z, [x25]
fmla    z27.s, p0/m, z0.s, z7.s
fmla    z26.s, p0/m, z0.s, z6.s
fmla    z25.s, p0/m, z0.s, z5.s
fmla    z24.s, p0/m, z0.s, z4.s
mov     x23, #0x1000                    // #4096
add     x25, x25, x23
ld1rw   {z0.s}, p0/z, [x25]
fmla    z23.s, p0/m, z0.s, z7.s
```

After
```
fmla    z28.s, p0/m, z4.s, z0.s
add     x27, x27, x17
ld1rw   {z0.s}, p0/z, [x27]
fmla    z27.s, p0/m, z7.s, z0.s
fmla    z26.s, p0/m, z6.s, z0.s
fmla    z25.s, p0/m, z5.s, z0.s
fmla    z24.s, p0/m, z4.s, z0.s
add     x27, x27, x17
ld1rw   {z0.s}, p0/z, [x27]
fmla    z23.s, p0/m, z7.s, z0.s
```

n_bcast_1_load matmu speedupsl

n_bcast_1_load conv speedups

## Detailed list of changes

- Enable n_bcast_1_load by loading quadwords and laned DOT/FMLAs to reduce loads per flop, this leads to 10-20% speedup on Neoverse V2 (little effect on Neoverse V1)
- We do not use quadwords+indexed DOT for n_bcast_1_load == false path due to A64FX using 2 uops for indexed DOT (it would also be tricky to save any loads in this case anyway). Too many different code paths for different hardware makes testing much harder.
- Optimizations and simplifications to broadcast load (LHS)
  - Remove broadcast lambda and replace with stateless member functions which are more descriptive and specific
  - In general use register + immediate/MUL_VL if we can, a strided shift/add or otherwise we use the general add_imm
  - Save stride in a register to avoid lots of movs in the hot loop
  - Tried several other approaches to incrementing, which are noted inline
  - Long term, I think we would want to have separate pointers for each row of the LHS which are post incremented by stride, but this will need us to find a range of free x regs
- Move GEMV microkernel into separate function, it doesn't make sense to keep with GEMM. For example, GEMV was using the broadcast function, but confusingly not actually broadcasting (because GEMV loads two whole vectors instead of broadcasting the LHS)
- Remove ld_full_mask, the masking functionality is currently broken, and I'm not convinced it is useful. We can restrict the vector vl at the system level, we don't need to do this error-prone predication approach (especially when it interacts with MUL_VL). This also moves jit_brgemm_kernel_t closer to VLA
- Also included #4682 so that I could test the GEMV path, this will need to go in first
- Use movi instead of eor for clearing, because it sometimes zero latency on AArch64

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
+++ b/src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
@@ -187,6 +187,8 @@ private:
 
     const XReg reg_tmp_ = x16;
 
+    const XReg reg_stride_bytes_A = x17;
+
     constexpr static int origin_offs_batch_offs_ = 0;
     constexpr static int origin_strd_batch_offs_ = 0;
     constexpr static int reg_bias_offs_ = 8;
@@ -220,28 +222,30 @@ private:
     constexpr static int max_vregs = 32;
     const int max_effective_vregs;
 
-    PReg ld_full_mask = PReg(2);
+    PReg rd_tail_mask = PReg(2);
     PReg ld_tail_mask = PReg(3);
-    PReg gemv_tail_mask = PReg(4);
 
     ZReg accm(int ld_block, int bd, int ld) const {
+        // Starts at the highest (e.g. 31), descending and using ld_block * bd_block registers as accumulators
         return ZReg(max_effective_vregs - 1 - (bd * ld_block + ld));
     }
 
-    ZReg bcst(int bd = 0) {
+    ZReg bcst(int bd = 0) const {
         if (n_bcast_1_load) {
-            int idx = max_effective_vregs - 1 - (brg.ld_block2 * brg.bd_block)
-                    - bd;
-            assert(idx > 0);
-            return ZReg(idx);
-        } else
-            return this->z0;
+            // Indexed FMLA/DOT instructions are only defined in the architecture for z0-z7
+            // https://developer.arm.com/documentation/111108/2025-12/SVE-Instructions/FMLA--indexed---Floating-point-fused-multiply-add-by-indexed-element-
+            assert(bd <= 7);
+            return ZReg(bd);
+        } else {
+            return ZReg(0);
+        }
     }
 
-    ZReg load(int ld = 0) {
+    ZReg load(int ld = 0) const {
         if (n_bcast_1_load) {
-            return this->z0;
+            return ZReg(brg.bd_block);
         } else {
+            // Starts off from lowest accm register, and continues descending
             int idx = max_effective_vregs - 1 - (brg.ld_block2 * brg.bd_block)
                     - ld;
             assert(idx > 0);
@@ -252,9 +256,27 @@ private:
     const ZReg &z_tmp_2() const noexcept { return this->z1; }
     const ZReg &z_tmp_3() const noexcept { return this->z2; }
     const ZReg &z_tail_mask() const noexcept { return this->z1; }
-    const ZReg &z_one_bytes() const noexcept { return this->z3; }
-    const ZReg &z_zp_a_shift() const noexcept { return this->z2; }
-    const ZReg &z_inp_shift() const noexcept { return this->z1; }
+    ZReg z_one_bytes() const noexcept {
+        if (n_bcast_1_load) {
+            return ZReg(brg.bd_block + 3);
+        } else {
+            return this->z3;
+        }
+    }
+    ZReg z_zp_a_shift() const noexcept {
+        if (n_bcast_1_load) {
+            return ZReg(brg.bd_block + 2);
+        } else {
+            return this->z2;
+        }
+    }
+    ZReg z_inp_shift() const noexcept {
+        if (n_bcast_1_load) {
+            return ZReg(brg.bd_block + 1);
+        } else {
+            return this->z1;
+        }
+    }
 
     ZReg int8_ones_words() const noexcept { return ZReg(max_vregs - 1); }
     ZReg int8_dot_product_temp() const noexcept { return ZReg(max_vregs - 2); }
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -220,28 +222,31 @@ struct jit_brgemm_kernel_t : public jit_generator_t {
...
+    ZReg bcst(int bd = 0) const {
         if (n_bcast_1_load) {
-            int idx = max_effective_vregs - 1 - (brg.ld_block2 * brg.bd_block)
-                    - bd;
-            assert(idx > 0);
-            return ZReg(idx);
-        } else
-            return this->z0;
+            // For indexed FMLA/DOTs we must use the first 8 vector registers
+            assert(bd <= 7);
```
renato-arantes, line 237: Why bd must be <= 7?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1452,6 +1527,80 @@ void jit_brgemm_kernel_t::compute_int8_compensation(int rd_loop, int bd_b,
...
+                && mul_vl <= 7) {
+            auto addr = ptr(reg_A_ptr, mul_vl, MUL_VL);
+            if (dt_bytes == 1) ld1b(dst.b, rd_tail_mask / T_z, addr);
+            if (dt_bytes == 2) ld1h(dst.h, rd_tail_mask / T_z, addr);
+        } else {
+            add_imm(X_DEFAULT_ADDR, reg_A_ptr, offset_bytes, tmp);
+            auto addr = ptr(X_DEFAULT_ADDR);
+            if (dt_bytes == 1) ld1b(dst.b, rd_tail_mask / T_z, addr);
+            if (dt_bytes == 2) ld1h(dst.h, rd_tail_mask / T_z, addr);
+        }
```
renato-arantes, line 1567: Can be replaced with LD_MUL_VL?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1452,6 +1527,80 @@ void jit_brgemm_kernel_t::compute_int8_compensation(int rd_loop, int bd_b,
...
+        const int32_t rd, const XReg &tmp) {
+    auto dt_bytes = dnnl_data_type_size(brg.dt_a);
+
+    int64_t offset_bytes = A_offset(bd, rd) - base_offset;
+    // Stride between bd
+    const auto A_stride_bytes = brg.typesize_A * brg.LDA;
+
+    // Tail here means fewer elements than needed to make up a word
+    if (!is_tail || brg.typesize_A == 4) {
+        if (offset_bytes > (63 * brg.typesize_A) || offset_bytes < 0
```
renato-arantes: Why 63?

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -252,9 +257,27 @@ struct jit_brgemm_kernel_t : public jit_generator_t {
     const ZReg &z_tmp_2() const noexcept { return this->z1; }
     const ZReg &z_tmp_3() const noexcept { return this->z2; }
     const ZReg &z_tail_mask() const noexcept { return this->z1; }
-    const ZReg &z_one_bytes() const noexcept { return this->z3; }
-    const ZReg &z_zp_a_shift() const noexcept { return this->z2; }
-    const ZReg &z_inp_shift() const noexcept { return this->z1; }
+    ZReg z_one_bytes() const noexcept {
+        if (n_bcast_1_load) {
+            return ZReg(brg.bd_block + 3);
```
Sqvid, line 261: is `brg.bd_block + 3` always guaranteed to yield a valid `ZReg`? Seems like a modulus might be needed? Ditto for the lines below.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Good suggestion, but we can't in this case. I need to use `add_imm` in the case where the `MUL_VL` fails. I also think that because this is a helper function, I can afford to be a bit more explicit. It makes it clear why we would take different paths for offsetting.

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: Indexed FMLA/DOTs are only defined for z0-z7, see https://developer.arm.com/documentation/111108/2025-12/SVE-Instructions/FMLA--indexed---Floating-point-fused-multiply-add-by-indexed-element-?lang=en#iclass_single_precision

I've hopefully made the comment a bit clearer

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: LD1RW can offset by up to 63 words.

I've added a comment with a link to the [docs](https://developer.arm.com/documentation/ddi0596/2021-03/SVE-Instructions/LD1RW--Load-and-broadcast-unsigned-word-to-vector-) and I've also changed the 63 * 4 to the more explicit 252 to match the docs.

renato-arantes (approved):
LGTM

src/cpu/aarch64/brgemm/jit_brgemm_kernel.cpp
jondea: A modulus would lead to potential silent failures when we change the regsiter picking logic. The parameters (e.g. bd_block) need to be chosen so that that vector registers chosen for A, B and accumulators are distinct and don't number more than 32. It's not bullet proof, but there is logic when these parameters are chosen that account for the total number of free vregs. Realistically, `bd_block` will always be well below 28, however you pick your registers.

Sqvid: Makes sense, we want to move towards a register manager eventually anyway.

PR #4809, opened 2026-03-11
cpu: rv64: switch layernorm fwd to fully JIT-based RVV kernels

xiazhuozhao:
# Description

This PR reworks the RV64 layernorm forward implementation around dedicated JIT kernels.

Instead of relying on the previous mixed path, it introduces separate fused and data JIT kernels for layernorm fwd, then incrementally tightens the implementation with a more stable variance formulation, RVV reduction cleanup, and a wider data path. The end result is that the layernorm fwd RVV path is handled entirely inside JIT-generated kernels. This also avoids the specific RVV-dependent execution path involved in the runtime crash seen on systems without RVV support, although it does not claim to solve all such issues globally. 

The 4 commits in this PR do the following:

* `272ee65`: introduce fused and data JIT kernels for layernorm fwd, establishing the basic RVV JIT structure for statistics and data processing. 
* `0416980`: switch the fused kernel to centered variance. This follows the x86 JIT approach, but keeps the computation in fp32 rather than introducing precision promotion to fp64. 
* `865a9fb`: use `vfredusum` for reductions and move the data path to 2x unrolling. 
* `1f2c0d3`: restore the 4x-unrolled RVV data path and rework tail handling to stay on the vector path with dynamic VL, removing the scalar tail fallback used in the previous revision. 

Numerically, the main change is the move from the `E[x^2] - E[x]^2` style variance computation to centered variance. This mirrors the x86 JIT idea and avoids precision upcasting, which keeps the RV64 implementation simpler while improving robustness of the variance computation. 

Performance was measured with:

```bash
./benchdnn --mode=P --lnorm --batch=inputs/lnorm/test_lnorm_f32
```

where `test_lnorm_f32` is a reduced batch derived from `test_lnorm_all`, keeping only f32 forward cases:

* `--dt=f32`
* `--dir=FWD_D,FWD_I`
* `--inplace=true,false`
* `--flags=,CH,G,GCH`
* `--batch=option_set_all`

The uploaded result files compare the optimized JIT implementation (`jit_lnorm:rvv`) against the raw baseline (`RISCV64GCV`). Across 688 cases, the `0time` column improves by about 3.0% geometric mean overall, with medians also around 2.6%. The gain is fairly consistent between inplace and non-inplace cases. By shape, the more visible improvements are on cases such as `256x768`, `257x768`, and `31x301`, while very large shapes like `40x1024` and `5120x1024` are roughly flat. The raw benchmark outputs are in the uploaded perf files.  

[sg2044_lnorm_opt.csv](https://github.com/user-attachments/files/26115970/sg2044_lnorm_opt.csv)
[sg2044_lnorm_raw.csv](https://github.com/user-attachments/files/26115972/sg2044_lnorm_raw.csv)

A concise summary of the observed perf trend is:

* average speedup: 1.030048308
* median speedup 1.025831992
* medium-sized irregular shapes benefit more than very large shapes
* no broad steady-state regression is visible from the uploaded data set

### DIFF
--- a/src/cpu/rv64/jit_rvv_layernorm_kernel.cpp
+++ b/src/cpu/rv64/jit_rvv_layernorm_kernel.cpp
@@ -214,8 +214,7 @@ void jit_rvv_layernorm_fused_kernel_t::generate() {
     fsw(f_var, reg_var_ptr, 0);
 
     L(start_data);
-    lw(reg_tmp, reg_param, GET_FUSED_OFF(eps));
-    fmv_w_x(f_aux, reg_tmp);
+    flw(f_aux, reg_param, GET_FUSED_OFF(eps));
     fadd_s(f_aux, f_var, f_aux);
     fsqrt_s(f_inv, f_aux);
     li(reg_tmp, 0x3f800000);
@@ -411,10 +410,8 @@ void jit_rvv_layernorm_data_kernel_t::generate() {
     ld(reg_scale, reg_param, GET_DATA_OFF(scale));
     ld(reg_shift, reg_param, GET_DATA_OFF(shift));
     ld(reg_len, reg_param, GET_DATA_OFF(len));
-    lw(reg_tmp, reg_param, GET_DATA_OFF(mean));
-    fmv_w_x(f_mean, reg_tmp);
-    lw(reg_tmp, reg_param, GET_DATA_OFF(inv_std));
-    fmv_w_x(f_inv, reg_tmp);
+    flw(f_mean, reg_param, GET_DATA_OFF(mean));
+    flw(f_inv, reg_param, GET_DATA_OFF(inv_std));
 
     vsetvli(reg_vlmax, x0, SEW::e32, LMUL::m1);
     slli(reg_bytes, reg_vlmax, 2);


src/cpu/rv64/jit_rvv_layernorm_kernel.cpp
```diff
@@ -64,24 +406,20 @@ void jit_rvv_layernorm_kernel_t::generate() {
...
+    ld(reg_src, reg_param, GET_DATA_OFF(src));
+    ld(reg_dst, reg_param, GET_DATA_OFF(dst));
+    ld(reg_scale, reg_param, GET_DATA_OFF(scale));
+    ld(reg_shift, reg_param, GET_DATA_OFF(shift));
+    ld(reg_len, reg_param, GET_DATA_OFF(len));
+    lw(reg_tmp, reg_param, GET_DATA_OFF(mean));
     fmv_w_x(f_mean, reg_tmp);
-    lw(reg_tmp, reg_param, 44);
+    lw(reg_tmp, reg_param, GET_DATA_OFF(inv_std));
     fmv_w_x(f_inv, reg_tmp);
```
zhangjian29: Any reason to use `lw` + `fmv_w_x` instead of `flw`?

src/cpu/rv64/jit_rvv_layernorm_kernel.cpp
```diff
@@ -23,16 +25,356 @@ namespace rv64 {
...
+    fsw(f_mean, reg_mean_ptr, 0);
+
+    L(store_var);
+    ld(reg_var_ptr, reg_param, GET_FUSED_OFF(variance));
+    beqz(reg_var_ptr, start_data);
+    fsw(f_var, reg_var_ptr, 0);
+
+    L(start_data);
+    lw(reg_tmp, reg_param, GET_FUSED_OFF(eps));
+    fmv_w_x(f_aux, reg_tmp);
```
zhangjian29: Same question here

src/cpu/rv64/jit_rvv_layernorm_kernel.cpp
xiazhuozhao: > Any reason to use `lw` + `fmv_w_x` instead of `flw`?

Thank you. The current use of `lw` + `fmv_w_x` is a leftover from earlier design, with no specific reason, and switching to `flw` is clearer and ensures type safety

PR #4817, opened 2026-03-12
cpu: x64: binary injector: per_mb_spatial non-trivial stride support

asztukow:
Addresses MFDNN-14499.
This PR adds non-trivial stride support in binary injector for per_mb_spatial.

### DIFF
--- a/src/cpu/x64/injectors/jit_uni_binary_injector.cpp
+++ b/src/cpu/x64/injectors/jit_uni_binary_injector.cpp
@@ -117,9 +117,58 @@ bool is_supported(cpu_isa_t isa, const dnnl::impl::memory_desc_t &src1_desc,
         const bcast_set_t &supported_strategy_set) {
     VCHECK_BIN_INJ_BOOL(is_data_supported(isa, src1_desc.data_type),
             VERBOSE_ISA_DT_MISMATCH);
-    VCHECK_BIN_INJ_BOOL(memory_desc_wrapper(src1_desc).is_dense(true),
-            VERBOSE_NONTRIVIAL_STRIDE);
+    const memory_desc_wrapper rhs_d(src1_desc);
+    const bool rhs_nontrivial_strides = !rhs_d.is_dense(true);
+    if (rhs_nontrivial_strides) {
+        VCHECK_BIN_INJ_BOOL(rhs_d.is_plain(), VERBOSE_NONTRIVIAL_STRIDE);
+        const auto bcast_type = get_rhs_arg_broadcasting_strategy(
+                src1_desc, dst_d, supported_strategy_set);
+        VCHECK_BIN_INJ_BOOL(
+                bcast_type == broadcasting_strategy_t::per_mb_spatial,
+                VERBOSE_NONTRIVIAL_STRIDE);
+        VCHECK_BIN_INJ_BOOL(dst_d.is_plain()
+                        && injector_utils::get_layout_type(dst_d)
+                                == injector_utils::layout_t::ncsp,
+                VERBOSE_NONTRIVIAL_STRIDE);
+    }
     return is_bcast_supported(src1_desc, dst_d, supported_strategy_set);
+}
+
+static dim_t rhs_offset_per_mb_spatial(const memory_desc_wrapper &dst_d,
+        const memory_desc_wrapper &rhs_d, dim_t dst_offset_bytes) {
+    // dst_off = n*dst_stride_n + c*dst_stride_c + d*dst_stride_d + h*dst_stride_h + w*dst_stride_w
+    // rhs_off = n*rhs_stride_n + d*rhs_stride_d + h*rhs_stride_h + w*rhs_stride_w
+    // per_mb_spatial: channel c is broadcast, so rem = (dst_off % dst_stride_n) % dst_stride_c.
+    const auto &dst_strides = dst_d.blocking_desc().strides;
+    const auto &rhs_strides = rhs_d.blocking_desc().strides;
+    const int ndims = dst_d.ndims();
+    const int dst_dt_shift
+            = math::ilog2q(types::data_type_size(dst_d.data_type()));
+    const dim_t dst_offset_elems = dst_offset_bytes >> dst_dt_shift;
+
+    const dim_t n = dst_offset_elems / dst_strides[0];
+    dim_t rem = dst_offset_elems % dst_strides[0];
+    if (dst_strides[1] != 0) rem %= dst_strides[1];
+
+    dim_t d = 0;
+    dim_t h = 0;
+    dim_t w = 0;
+    if (ndims >= 5) {
+        d = rem / dst_strides[ndims - 3];
+        rem %= dst_strides[ndims - 3];
+    }
+    if (ndims >= 4) {
+        h = rem / dst_strides[ndims - 2];
+        rem %= dst_strides[ndims - 2];
+    }
+    if (ndims >= 3) w = rem / dst_strides[ndims - 1];
+
+    // output = rhs_off (in elements)
+    dim_t rhs_offset = n * rhs_strides[0];
+    if (ndims >= 5) rhs_offset += d * rhs_strides[ndims - 3];
+    if (ndims >= 4) rhs_offset += h * rhs_strides[ndims - 2];
+    if (ndims >= 3) rhs_offset += w * rhs_strides[ndims - 1];
+    return rhs_offset;
 }
 
 bool binary_args_broadcast_supported(const post_ops_t &post_ops,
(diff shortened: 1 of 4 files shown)


src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -117,11 +117,59 @@ bool is_supported(cpu_isa_t isa, const dnnl::impl::memory_desc_t &src1_desc,
...
+                src1_desc, dst_d, supported_strategy_set);
+        VCHECK_BIN_INJ_BOOL(rhs_d.is_plain(), VERBOSE_NONTRIVIAL_STRIDE);
+        if (bcast_type == broadcasting_strategy_t::per_mb_spatial) {
+            VCHECK_BIN_INJ_BOOL(dst_d.is_plain()
+                            && injector_utils::get_layout_type(dst_d)
+                                    == injector_utils::layout_t::ncsp,
+                    VERBOSE_NONTRIVIAL_STRIDE);
+        } else {
+            VCHECK_BIN_INJ_BOOL(false, VERBOSE_NONTRIVIAL_STRIDE);
+        }
```
asimonov1:
```suggestion
        VCHECK_BIN_INJ_BOOL(rhs_d.is_plain(), VERBOSE_NONTRIVIAL_STRIDE);
        const auto bcast_type = get_rhs_arg_broadcasting_strategy(
                src1_desc, dst_d, supported_strategy_set);
        VCHECK_BIN_INJ_BOOL(
                bcast_type == broadcasting_strategy_t::per_mb_spatial,
                VERBOSE_NONTRIVIAL_STRIDE);
        VCHECK_BIN_INJ_BOOL(dst_d.is_plain()
                        && injector_utils::get_layout_type(dst_d)
                                == injector_utils::layout_t::ncsp,
                VERBOSE_NONTRIVIAL_STRIDE);
```

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
+
+    host_->xor_(r8, r8);
+    host_->mov(rax, tmp_reg);
+    host_->mov(tmp_reg, dst_strides[0]);
+    host_->xor_(rdx, rdx);
+    host_->div(tmp_reg);
+    if (rhs_strides[0] != 0) {
+        host_->mov(tmp_reg, rhs_strides[0]);
+        host_->imul(rax, tmp_reg);
+        host_->add(r8, rax);
```
inteldimitrius, line 1394: (random place) Is it possible to add test coverage for these changes in benchdnn?

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
-                    break;
-                case injector_utils::layout_t::cspn:
-                    calculate_mb_sp_cspn_base(strides, tmp_reg);
-                    break;
-                default: assert(!"Unknown layout");
-            }
-
-            if (elem_size_bytes == 1) {
-                host_->add(addr_reg, rax);
+            if (layout == injector_utils::layout_t::ncsp && rhs_d.is_plain()) {
```
georgen117: possible performance issue: the `is_plain()` is true for both dense and non-dense plain tensors, so this condition also captures the existing dense ncsp case and routes it through `calculate_mb_sp_ncsp_base_rhs_strided`.

- `calculate_mb_sp_ncsp_base_rhs_strided` has 3-5 `div` + `imul` instructions
- `calculate_mb_sp_ncsp_base` has 2 `div` + `mul` instructions.

both calculate_mb_sp_ncsp_base_rhs_strided and calculate_mb_sp_ncsp_base produce the same results but the calculate_mb_sp_ncsp_base should be faster if the rhs_d tensor is dense.
if you changed this from:
```cpp
if (layout == injector_utils::layout_t::ncsp && rhs_d.is_plain()) {
```
to:
```cpp
if (layout == injector_utils::layout_t::ncsp && rhs_d.is_plain()
        && !rhs_d.is_dense(true)) {
```
then it should continue to call `calculate_mb_sp_ncsp_base` for the dense case. Which based on the instructions should be faster.

This is all based on the idea that the `is_plain()` is true for both dense and non-dense tensors.  Which I might be wrong about.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
-                case injector_utils::layout_t::nspc:
-                    calculate_mb_sp_nspc_partial(strides, it_off_val->second,
-                            tmp_reg, elem_size_bytes);
-                    break;
-                case injector_utils::layout_t::cspn:
-                    calculate_mb_sp_cspn_partial(strides, it_off_val->second,
-                            tmp_reg, elem_size_bytes);
-                    break;
-                default: assert(!"Unknown layout");
+            if (layout == injector_utils::layout_t::ncsp && rhs_d.is_plain()) {
```
georgen117: Possible same issue as line 1301 above:

suggest changing to:
```cpp
if (layout == injector_utils::layout_t::ncsp && rhs_d.is_plain()
        && !rhs_d.is_dense(true)) {
```

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
         }
     }
 }
 
+template <cpu_isa_t isa, typename Vmm>
+void jit_uni_binary_injector_t<isa, Vmm>::calculate_mb_sp_ncsp_base_rhs_strided(
+        const memory_desc_wrapper &dst_d, const memory_desc_wrapper &rhs_d,
+        const dim_t *dst_strides, const Xbyak::Reg64 &addr_reg,
+        const Xbyak::Reg64 &tmp_reg, std::size_t elem_size_bytes) const {
+    const auto ndims = dst_d.ndims();
```
georgen117, line 1378: Every other `calculate_mb_sp_*` function has a clear comment explaining the mathmatical derivation of the block.

For example `calculate_mb_sp_ncsp_base` has:
```
void jit_uni_binary_injector_t<isa, Vmm>::calculate_mb_sp_ncsp_base(
        const dim_t *strides, const Xbyak::Reg64 &tmp_reg) const {
    // offset = (n * stride_n) + (c * stride_c) + (d * stride_d) + (h * stride_h) + (w * stride_w)
    // mb_sp_off = (n * (stride_n/C)) + (d * stride_d) + (h * stride_h) + (w * stride_w)
    // mb_sp_off = offset - (c * stride_c) - (n * (C - 1)DHW)
    // output = rax
```

adding something similar here could make this much easier to review and maintain.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
+        host_->div(tmp_reg);
+        if (rhs_strides[ndims - 3] != 0) {
+            host_->mov(tmp_reg, rhs_strides[ndims - 3]);
+            host_->imul(rax, tmp_reg);
+            host_->add(r8, rax);
+        }
+    }
+
+    if (ndims >= 4) {
+        host_->mov(rax, rdx);
```
georgen117, line 1417: isn't `rax` already set to `rdx` from line 1392?

not a bug but a repeated op.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
...
+        host_->div(tmp_reg);
+        if (rhs_strides[ndims - 2] != 0) {
+            host_->mov(tmp_reg, rhs_strides[ndims - 2]);
+            host_->imul(rax, tmp_reg);
+            host_->add(r8, rax);
+        }
+    }
+
+    if (ndims >= 3) {
+        host_->mov(rax, rdx);
```
georgen117, line 1430: isn't `rax` already set to `rdx` from line 1392?

once again not a bug but a repeated op.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1249,60 +1298,168 @@ void jit_uni_binary_injector_t<isa, Vmm>::append_mb_sp_offset(
                             host_,
                             {is_out_reg ? it_out_reg->second : Xbyak::Reg64()}};
```
georgen117, line 1301: minor: the `calculate_mb_sp_ncsp_partial_rhs_strided` only uses `rax`, `rdx`, `r8` and scratch register `temp_reg`. It does not use `r9` so the `register_guard` will push/pop `r9` unnecessarily when taking the new `calculate_mb_sp_ncsp_partial_rhs_strided`. This is likely negligible. For code simplicity I think I would leave this unchanged. I am pointing it out just in case you have a different opinion about the extra push/pop.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
asztukow: Yes, you are right, `is_plain()` is true for both. Fixed the if statement, thanks!

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
asztukow: My mistake - that outer `mov` was redundant and I removed it (also added one in `ndims >= 5` case). `mov rax, rdx` must remain inside each conditional block because after each `div` the remainder is in `rdx`, and that remainder must be moved into `rax` before `xor` (which clears `rdx`) for the next division. With a single `mov` outside, subsequent `xor`/`div` may operate on wrong values and spatial index reconstruction will be incorrect.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
asztukow: Thanks, done.

src/cpu/x64/injectors/jit_uni_binary_injector.cpp
asztukow: Sure, is there a preferred place in benchdnn where this test should be added?

inteldimitrius: The basic path is `tests/benchdnn/inputs/binary` and it creating `harness_binary_strides` should be ok.

PR #4757, opened 2026-03-03
revert: gpu: intel: ze: engine: resolve potential module leaks

dzarukin:
This reverts commit 2dd6b8a7e7b36ecf32aa1dfb0a35f4e3a8c4a836.

Improving the code from potential leaks didn't do good. Will take another attempt later.

PR #4818, opened 2026-03-12
[backport][rls-v3.12-pc] ze api fixes

dzarukin:
Backport of #4710 and #4757 (removing bad commit from this set).


dzarukin:
make test
disable test_device_cpu

PR #4724, opened 2026-02-24
backport xe3p

dyoussif:

PR #4348, opened 2025-11-19
xe: jit: gemm: clear non-dpas tokens at end of k loop

petercad:
In some cases, nGEN is unable to infer that A/B load message tokens are inactive after the k loop due to the complicated control flow. Occasionally this can lead to poor token usage during the C update, causing low performance. An example for Xe2/Xe3:

```
--dt=s8:u4:f16 --stag=ab --wtag=ba --dtag=ab --attr-post-ops=add:f16:3:ab 4096x4096:4096x4096
```

This PR manually adds syncs at the end of the k loop to help inform nGEN these tokens are no longer active.

PR #4819, opened 2026-03-12
[GPU] Fix SWSB dependencies after 512 GRF update

Simonsays095:
Addresses [MFDNN=14713](https://jira.devtools.intel.com/browse/MFDNN-14713).

#4348 Added dependencies on all registers at the end of the k-loop, to clear tokens and improve post-op performance. In #4724 we increased the number of registers from 256 to 512. When these features combine, we end up trying to add dependencies on all 512 registers, even when hardware doesn't support it. This translates in auto-SWSB and encoding to a stricter `A@1` dependency than the required `@1`.

This change fixes this by switching to using the number of registers available in the allocator directly (in the dependency insertion loop), and then further limiting the strategy during preflight according to the hardware capabilities.

I also added some exceptions when encoding instructions that use 8-bit register numbers, so we fail to compile these kernels if the register number doesn't fit in 8 bits (i.e. >255).

### DIFF
--- a/third_party/ngen/ngen_core.hpp
+++ b/third_party/ngen/ngen_core.hpp
@@ -900,10 +900,14 @@ inline RegData abs(const RegData &r)
     return result.setMods(1);
 }
 
+// Forward declare
+static constexpr int maxRegs(HW hw);
+
 inline void RegData::fixup(HW hw, int execSize, int execWidth, DataType defaultType, int srcN, int arity)
 {
 #ifdef NGEN_SAFE
     if (isInvalid()) throw invalid_object_exception();
+    if (!isARF() && !isIndirect() && static_cast<int>(base) >= maxRegs(hw)) throw grf_out_of_bounds_exception();
 #endif
 
     if (getType() == DataType::invalid) {
@@ -1299,6 +1303,10 @@ public:
             : 256;
     }
 };
+
+static constexpr int maxRegs(HW hw) {
+    return GRF::maxRegs(hw);
+}
 
 class ARF : public Register
 {


third_party/ngen/ngen_gen12.hpp
```diff
@@ -679,6 +679,8 @@ static inline constexpr14 BinaryOperand12 encodeBinaryOperand12(const RegData &r
 
 #ifdef NGEN_SAFE
     if (rd.isInvalid()) throw invalid_object_exception();
+    if (!rd.isIndirect() && rd.getBase() > 255)
```
kealan-barbieri: is `isIndirect` just referenced here because in such cases base is uninitialized and potentially some arbitrary value?

kealan-barbieri:
nit: theres a handful of places it looks like we're using maxregs where we could switch to the actual value if it doesnt seem too risky, eg:

- https://github.com/intel-innersource/libraries.performance.math.onednn/blob/85b3b18319b28612f927322279aaae3a23d2266a/src/gpu/intel/gemm/jit/generator/pieces/copy.cxx#L268
- https://github.com/intel-innersource/libraries.performance.math.onednn/blob/main/src/gpu/intel/gemm/jit/generator/pieces/register_layout.cpp#L1853
- a whole bunch related to post ops where its not clear if the ra variable is in scope.

third_party/ngen/ngen_gen12.hpp
Simonsays095: I'm not sure what the expected state of `rd` is for indirect operands, but the idea is that we only encode the register index from `rd` when `!rd.isIndirect()`. In indirect cases, the register index isn't even encoded in the instruction.

Simonsays095:
> nit: theres a handful of places it looks like we're using maxregs where we could switch to the actual value if it doesnt seem too risky, eg:
> 
> * https://github.com/intel-innersource/libraries.performance.math.onednn/blob/85b3b18319b28612f927322279aaae3a23d2266a/src/gpu/intel/gemm/jit/generator/pieces/copy.cxx#L268
> * https://github.com/intel-innersource/libraries.performance.math.onednn/blob/main/src/gpu/intel/gemm/jit/generator/pieces/register_layout.cpp#L1853
> * a whole bunch related to post ops where its not clear if the ra variable is in scope.

`GRF::maxRegs()` is constexpr, whereas the new `GRF::maxRegs(HW hw)` is not. So making that change would mean converting arrays -> vectors (which I did here, but want to avoid in general). I'd only want to do this if there's a confirmed bug in that code. Going forward, we should be using the HW-aware version pretty much always.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

third_party/ngen/ngen_gen12.hpp
kealan-barbieri: Ideally it seems we should have a reliable default value for register index in  `rd` so this check isnt a concern.

Simonsays095: I don't think the default is the issue (at least, it's not in this case). Having an invalid `rd.getBase()` just won't break anything if using an indirect region (since the register number isn't encoded), so we shouldn't throw exceptions in this case.

atkassen:
> `GRF::maxRegs()` is constexpr, whereas the new `GRF::maxRegs(HW hw)` is not.

Why not? It satisfies the requirements for C++11 constexpr functions. Then you can keep the arrays.

third_party/ngen/ngen_core.hpp
```diff
@@ -1293,6 +1293,7 @@ class GRF : public Register
     static constexpr int bytesToGRFs(HW hw, unsigned x)    { return (x + bytes(hw) - 1) >> log2Bytes(hw); }
 
     static constexpr int maxRegs()                         { return 512; }
+    static int maxRegs(HW hw)                              { return (hw == HW::XE3P_35_11) ? 512 : 256; }
```
echeresh: What about `XE3P_35_10`? Looks like it's different in register allocator:


https://github.com/uxlfoundation/oneDNN/blob/4362af6b8a5ac422ccab7fa2bf42d3f2c785a8ae/third_party/ngen/ngen_register_allocator.hpp#L381-L384

third_party/ngen/ngen_gen12.hpp
kealan-barbieri: Understood, my thought was that if an indirect rd will always have a sub 256 value theres no harm in performing this check and the addidional indirectness check is redundant, however looking at the code it appears the base is used by indirect `rd` s so theres some chance this value is over 256 and will result in throwing an erroneous exception as you describe: 
https://github.com/intel-innersource/libraries.performance.math.onednn/blob/85b3b18319b28612f927322279aaae3a23d2266a/third_party/ngen/ngen_core.hpp#L834

Simonsays095:
> Why not? It satisfies the requirements for C++11 constexpr functions. Then you can keep the arrays.

I forgot that:
1. functions with scalar arguments, and only a single (return) statement can be constexpr
2. These functions can be used as array bounds when the argument is defined at compile-time, such as template parameters.

I'll make this change.

third_party/ngen/ngen_core.hpp
Simonsays095: This looks like another existing bug, but we unconditionally call `setRegisterCount(strategy.GRFs);` so it never showed up. I've updated it to default to `maxRegs(hw)` with only the `hw < HW::XeHP` exception. I'm not sure how many registers the hardware has in those cases, so I left it (possibly `maxRegs(hw)` should return 128 in those cases).

Simonsays095: I went ahead and pulled this into `maxRegs(HW)`. I think this is an actual hardware limitation.

third_party/ngen/ngen_gen12.hpp
atkassen: Is the issue with bases >= 256 that it overflows the bitfield? If so, why not just mask off the high bit below?

```
        op.direct.regNum = rd.getBase() & 0xFF;
```

Simonsays095: That wouldn't have fixed this bug, either, since we're still calling `wrdep` on all those extra registers (so we'd just be adding every register as a dependency, not just non-C registers). The idea is that if we're using a register that doesn't exist on HW, then we're doing something wrong.

atkassen: But doesn't this forbid using any register with base > 255 in a unary/binary instruction?

Simonsays095: These exceptions have been added to overloads of `encodeBinaryOperand12` and `encodeTernaryOperand12`. If we ever get to this point we should be encoding an instruction that only has 8 bits reserved for the register base. So this really is only valid for base < 256, enforced by encoding. Do you know of an exception to this, where we need base > 255 (in a binary/ternary operand as part of an `Instruction12`)?

atkassen: The high bits are encoded [elsewhere](https://github.com/uxlfoundation/oneDNN/blob/4362af6b8a5ac422ccab7fa2bf42d3f2c785a8ae/third_party/ngen/ngen_gen12.hpp#L493-L498) but using those registers should still be valid.

The copy planner, for example, will happily allocate and use these registers.

Simonsays095: Thanks for the explanation @atkassen. I understand that this would break Xe3p instructions using >=r256, which is completely valid. I've updated the code to remove the exceptions, replacing them with one at `RegData::fixup` that checks specifically if the base is out-of-bounds using knowledge of the hardware.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph

third_party/ngen/ngen_core.hpp
```diff
@@ -900,10 +900,14 @@ inline RegData abs(const RegData &r)
...
 }
 
+// Forward declare
+static constexpr int maxRegs(HW hw);
+
 inline void RegData::fixup(HW hw, int execSize, int execWidth, DataType defaultType, int srcN, int arity)
 {
 #ifdef NGEN_SAFE
     if (isInvalid()) throw invalid_object_exception();
+    if (base > maxRegs(hw)) throw grf_out_of_bounds_exception();
```
atkassen: `>=`

Simonsays095: Thanks, fixed.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph

PR #4827, opened 2026-03-13
cpu: aarch64: Support for Bf16 LUT Eltwise Operations

nikhil-arm:
# Description

Please include a summary of the change. Please also include relevant motivation and context. See [contribution guidelines](https://github.com/uxlfoundation/oneDNN/blob/main/CONTRIBUTING.md) for more details. If the change fixes an issue not documented in the project's Github issue tracker, please document all steps necessary to reproduce it.

Fixes # (github issue)

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes


nikhil-arm:
Performance Improvements on 128 bit machine 

BenchDNN commands
- OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=gelu_erf 6432x5120
 - OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=swish --alpha=1 6432x5120
 - OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=gelu_tanh 6432x5120
 - OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=exp 6432x5120
 - OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=log 6432x5120
 - OMP_NUM_THREADS=1 ./build/tests/benchdnn/benchdnn --eltwise --mode=P --dt=bf16
    --alg=sqrt 6432x5120

  | threads | op        | lut avg (ms) | base avg (ms) | speedup |
  |--------:|-----------|-------------:|--------------:|--------:|
  | 1       | gelu_erf   | 12.7663  | 105.556   | 8.268331 |
  | 1       | swish      | 12.9223  | 108.851   | 8.423500 |
  | 1       | gelu_tanh  | 12.8577  | 136.556   | 10.620562 |
  | 1       | exp        | 12.7922  | 59.2863   | 4.634566 |
  | 1       | log        | 12.8866  | 80.5249   | 6.248731 |
  | 1       | sqrt       | 12.8670  | 20.4943   | 1.592780 |
  | 8       | gelu_erf   | 1.62322  | 13.2287   | 8.149665 |
  | 8       | swish      | 1.62900  | 13.6354   | 8.370411 |
  | 8       | gelu_tanh  | 1.63109  | 17.0969   | 10.481886 |
  | 8       | exp        | 1.62035  | 7.43281   | 4.587163 |
  | 8       | log        | 1.62276  | 10.0804   | 6.211886 |
  | 8       | sqrt       | 1.62573  | 2.56726   | 1.579143 |
  | 64      | gelu_erf   | 0.208485 | 1.66726   | 7.997026 |
  | 64      | swish      | 0.208498 | 1.71613   | 8.230918 |
  | 64      | gelu_tanh  | 0.208785 | 2.15408   | 10.317216 |
  | 64      | exp        | 0.207944 | 0.940454  | 4.522631 |
  | 64      | log        | 0.208260 | 1.26797   | 6.088399 |
  | 64      | sqrt       | 0.207303 | 0.328519  | 1.584729 |

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+            VDISPATCH_ELTWISE(src_d.is_dense(true), VERBOSE_NONTRIVIAL_STRIDE);
+            VDISPATCH_ELTWISE(
+                    src_d == dst_d, VERBOSE_INCONSISTENT_MDS, "src", "dst");
+            VDISPATCH_ELTWISE(
+                    attr()->has_default_values(), VERBOSE_UNSUPPORTED_ATTR);
+
+            VDISPATCH_ELTWISE(data_type == ::dnnl::impl::data_type::bf16,
+                    VERBOSE_UNSUPPORTED_DT);
+
+            const auto *spec = get_bf16_fwd_lut_spec_(desc()->alg_kind);
```
jondea: Do we need all this spec machinery? Can we replace it all with
```
if (!swish){
    alpha = 0;
}
beta = 0;
```

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+            VDISPATCH_ELTWISE(spec != nullptr, VERBOSE_BAD_ALGORITHM);
+
+            const float alpha = spec->ignore_alpha_beta ? 0.f : desc()->alpha;
+            const float beta = spec->ignore_alpha_beta ? 0.f : desc()->beta;
+            bf16_lut_.resize(1u << 16);
+            for (uint32_t raw = 0; raw < (1u << 16); ++raw) {
```
jondea: raw is a bit vague, consider raw -> x_u16

src/cpu/cpu_eltwise_list.cpp
```diff
@@ -62,6 +63,7 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<avx512_core>)
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<avx2>)
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<sse41>)
+            CPU_INSTANCE_AARCH64(ref_eltwise_lut_fwd_t<bf16>)
```
jondea: I'm not sure I'd call it a ref, when we see ref in a verbose output, it usually implies it is unoptimized. I'd just call it `eltwise_lut_fwd_t`, I don't think we even need the template, it's not really doing anything. If we come to extend this to other data types, I think we can reconsider how.

src/cpu/aarch64/ref_eltwise_lut.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    const auto offset_bytes = types::elements_to_bytes(
+            pd()->src_md()->data_type, data_d.offset0());
+    src_u8 += offset_bytes;
+    dst_u8 += offset_bytes;
+
+    const auto *src = reinterpret_cast<const data_t *>(src_u8);
+    auto *dst = reinterpret_cast<data_t *>(dst_u8);
+
+    static_assert(sizeof(data_t) == sizeof(bfloat16_t),
```
jondea: I don't think this is needed, validation should be kept to the `pd_t::init()` if possible

src/cpu/aarch64/ref_eltwise_lut.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (data_d.has_zero_dim()) return status::success;
+
+    const dim_t n = data_d.nelems(true);
+
+    const auto *src_u8 = CTX_IN_MEM(const uint8_t *, DNNL_ARG_SRC);
+    auto *dst_u8 = CTX_OUT_MEM(uint8_t *, DNNL_ARG_DST);
+
+    const auto offset_bytes = types::elements_to_bytes(
+            pd()->src_md()->data_type, data_d.offset0());
+    src_u8 += offset_bytes;
```
jondea: I don't think we need to cast back and forward to u8. We can just use bf16 and do
```
src += data_d.offset0();
```
etc

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+            using namespace ::dnnl::impl::utils;
+
+            const memory_desc_wrapper src_d(src_md());
+            const memory_desc_wrapper dst_d(dst_md());
+
+            VDISPATCH_ELTWISE(is_fwd(), VERBOSE_BAD_PROPKIND);
+            VDISPATCH_ELTWISE(everyone_is(data_type, src_md()->data_type,
+                                      dst_md()->data_type),
+                    VERBOSE_UNSUPPORTED_DT);
+            VDISPATCH_ELTWISE(platform::has_data_type_support(data_type),
```
jondea: I don't think we need to enforce this. I think this check is for bfdot, but your code doesn't need to use any bf16 instructions. The calculation of the LUT should automatically fallback, and the LUT access doesn't use bf16.

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+                    VERBOSE_UNSUPPORTED_DT);
+            VDISPATCH_ELTWISE(
+                    set_default_formats_common(), VERBOSE_UNSUPPORTED_TAG);
+            VDISPATCH_ELTWISE(src_d.is_dense(true), VERBOSE_NONTRIVIAL_STRIDE);
+            VDISPATCH_ELTWISE(
```
jondea: Redundant due to line 54

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+                    src_d == dst_d, VERBOSE_INCONSISTENT_MDS, "src", "dst");
+            VDISPATCH_ELTWISE(
+                    attr()->has_default_values(), VERBOSE_UNSUPPORTED_ATTR);
+
+            VDISPATCH_ELTWISE(data_type == ::dnnl::impl::data_type::bf16,
```
jondea: Redundant due to line 54

src/cpu/aarch64/ref_eltwise_lut.hpp
```diff
@@ -0,0 +1,134 @@
...
+                const float x = static_cast<float>(x_bf16);
+                const float y = compute_eltwise_scalar_fwd(
+                        desc()->alg_kind, x, alpha, beta);
+                bf16_lut_[raw] = bfloat16_t(y);
+            }
+
+            return status::success;
+        }
+
+        std::vector<bfloat16_t> bf16_lut_;
```
jondea: I think that the LUT should be stored in the primitive rather than primitive::pd_t, similar to how the JIT primitives do codegen.

jondea (changes requested):
Lovely speedups and such an elegant idea, thank you!

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: done

src/cpu/cpu_eltwise_list.cpp
nikhil-arm: done

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: I have removed this for now. 
We are using ```compute_eltwise_scalar_fwd()``` to populate the LUT table. We might need to bring this back if in future we chose to populate table using accelerated ops?

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: ```src_d.is_dense(true)``` check is still needed because ```set_default_formats_common()``` does not prove the source memory is contiguous; the linear LUT loop depends on dense storage?

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: done

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: done

src/cpu/aarch64/ref_eltwise_lut.cpp
nikhil-arm: done

src/cpu/aarch64/ref_eltwise_lut.cpp
nikhil-arm: done

src/cpu/aarch64/ref_eltwise_lut.hpp
nikhil-arm: I have addressed this in a slightly diff. way. ```compute_eltwise_scalar_fwd()``` is responsible to generate lut values that takes care of an algos alpha beta. 
As long as ```compute_eltwise_scalar_fwd()``` is accurate, descriptor and its lut will be specialized for alpha/beta

src/cpu/aarch64/ref_eltwise_lut.hpp
jondea: I don't think we would need to because we could just use a nested primitive like we do with deconv and matmul_inner_product. Then we can make use of the oneDNN's built in feature selection and fallback.

.gitignore
```diff
@@ -33,3 +33,4 @@ __pycache__
 .cache
 .gdb_history
 .gdbinit
+
```
jondea: optional: getting rid of this would mean that we wouldn't need onednn-arch approval

nikhil-arm:
```suggestion
```

nikhil-arm: :(

nikhil-arm: done

src/cpu/aarch64/eltwise_lut.hpp
```diff
@@ -0,0 +1,91 @@
...
+namespace impl {
+namespace cpu {
+
+template <::dnnl::impl::data_type_t data_type>
+struct eltwise_lut_fwd_t : public primitive_t {
+    using data_t = typename ::dnnl::impl::prec_traits_t<data_type>::type;
+
+    struct pd_t : public cpu_eltwise_fwd_pd_t {
+        using cpu_eltwise_fwd_pd_t::cpu_eltwise_fwd_pd_t;
+        DECLARE_COMMON_PD_T("eltwise_lut", eltwise_lut_fwd_t);
```
jondea: nit: this could just be `lut`, we don't usually include the primitive type in the implementation verbose name

jondea (approved):
Great stuff! Thank you

src/cpu/aarch64/eltwise_lut.hpp
nikhil-arm: Should I rename just the message ("eltwise_lut") in DECLARE_COMMON_PD_T  or the eltwise_lut_* to lut_* in the overall implementation?

nikhil-arm:
```suggestion
        DECLARE_COMMON_PD_T("lut", eltwise_lut_fwd_t);
```

nikhil-arm: I have changed the log message only. Let me know if you are suggesting to rename class and functions

jondea: Great, that's what I meant

.gitignore
Sqvid: `cpu_eltwise_list.cpp` :laughing:

src/cpu/aarch64/eltwise_lut.hpp
```diff
@@ -0,0 +1,91 @@
...
+    eltwise_lut_fwd_t(const pd_t *apd) : primitive_t(apd) {}
+    status_t init(engine_t * /*engine*/) override { return status::success; }
+
+    status_t execute(const exec_ctx_t & /*ctx*/) const override {
+        return status::unimplemented;
+    }
+
+private:
+    const pd_t *pd() const { return (const pd_t *)primitive_t::pd().get(); }
+    std::vector<data_t> lut_;
```
Sqvid, line 84: Given that this is going to be the majority of the data you hold, have you considered having it passed in via `scratchpad`? Otherwise the framework-level caller has no control over allocation.

Sqvid (dismissed):

Sqvid:
@dzarukin could you have a look at this for onednn-arch?

src/cpu/aarch64/eltwise_lut.hpp
nikhil-arm: hmm, good observation. Yeah I think we should allow the allocator to control this

src/cpu/cpu_eltwise_list.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include "cpu/cpu_engine.hpp"
 
+#include "cpu/aarch64/eltwise_lut.hpp"
```
dzarukin: This is the wrong spot. Should be guarded with aarch64 macro.

src/cpu/cpu_eltwise_list.cpp
```diff
@@ -62,6 +63,7 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<avx512_core>)
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<avx2>)
             CPU_INSTANCE_X64(jit_uni_eltwise_int_fwd_t<sse41>)
+            CPU_INSTANCE_AARCH64(eltwise_lut_fwd_t<bf16>)
```
dzarukin: I highly encourage to drop data type from a templated argument to avoid unnecessary bloating once data types support is increased.

dzarukin (commented):
What's the point of moving computations from execute to create part?

Secondly, why focusing on a separate implementation when eltwise post-op would always have an advantage over a separate implementation?

src/cpu/aarch64/eltwise_lut.hpp
dzarukin: This won't work as scratchpad is expected to be passed at execute() time and all the stashing part is done at init() time.

Sqvid: Oh true, ignore my comment then.

jondea: I don't think this is possible. Scratchpad granting is only done at execute time, but this LUT needs to have the same lifetime as the primitive.

jondea:
> What's the point of moving computations from execute to create part?
> 
> Secondly, why focusing on a separate implementation when eltwise post-op would always have an advantage over a separate implementation?

We've seen pretty big speedups at the framework level for the transcendental BF16 functions, especially as they need upcasting and downcasting to fp32. It turned out to be a very simple impl, so it made sense.

Post-ops would be a good next step, but the oneDNN numerical contract still requires post ops to be done in F32 (even when `fpmath=bf16` or `acc_mode=bf16`). For the transcendental functions, providing a fast BF16->BF16 impl essentially allows frameworks to opt into doing post ops with intermediate BF16 casts if they are happy with it numerically. With more aggressive quantization methods becoming popular, we are seeing this as something that frameworks are happy to accept, but not possible with oneDNN post-ops. It's on my todo list to write an RFC on this, but I don't yet have a clear picture of how it would look.

Also, we don't have good support for JIT fused post-ops yet. It's something we want to work on, but it is technically much more difficult than this, so this made sense as low hanging fruit.

src/cpu/cpu_eltwise_list.cpp
nikhil-arm: Fixed

src/cpu/cpu_eltwise_list.cpp
nikhil-arm: Fixed

src/cpu/aarch64/eltwise_lut.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "common/utils.hpp"
+
+#include "cpu/cpu_eltwise_pd.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+struct eltwise_lut_fwd_t : public primitive_t {
```
dzarukin, line 37: It would be good to put a comment here explaining that all compute-intense part was moved to creation stage which increases initialization time for this implementation and also requires extra 128K buffer for each unique value in bf16 data type.

dzarukin (approved):

src/cpu/aarch64/eltwise_lut.hpp
nikhil-arm: Fixed

Sqvid (changes requested):
Can you please test your changes with the following:
```
$ ./build/tests/benchdnn/benchdnn --eltwise --global-impl='lut' --batch=test_eltwise_all
```

There are quite a few failed tests.

```
33792:FAILED (1 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=exp --alpha=0 --beta=0 --inplace=true --impl=lut 5x16x3
33793:FAILED (0 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=exp --alpha=0 --beta=0 --impl=lut 5x16x3
33800:FAILED (1 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=log --alpha=0 --beta=0 --inplace=true --impl=lut 5x16x3
33801:FAILED (2 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=log --alpha=0 --beta=0 --impl=lut 5x16x3
33802:FAILED (1 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=logistic --alpha=0 --beta=0 --inplace=true --impl=lut 5x16x3
33803:FAILED (2 ms) __REPRO: --eltwise --dt=bf16 --tag=ABx16a16b --alg=logistic --alpha=0 --beta=0 --impl=lut 5x16x3
34332:FAILED (1 ms) __REPRO: --eltwise --dt=bf16 --tag=aBx8b --alg=exp --alpha=0 --beta=0 --inplace=true --impl=lut 3x7x3x2
34333:FAILED (2 ms) __REPRO: --eltwise --dt=bf16 --tag=aBx8b --alg=exp --alpha=0 --beta=0 --impl=lut 3x7x3x2
34340:FAILED (1 ms) __REPRO: --eltwise --dt=bf16 --tag=aBx8b --alg=log --alpha=0 --beta=0 --inplace=true --impl=lut 3x7x3x2
34341:FAILED (2 ms) __REPRO: --eltwise --dt=bf16 --tag=aBx8b --alg=log --alpha=0 --beta=0 --impl=lut 3x7x3x2
(... 62 more cases ...)
```

Sqvid (approved):
Thanks for the patch!

PR #4830, opened 2026-03-13
benchdnn: a bunch of fixes and maintenance clean ups

dzarukin:
This PR contains several changes:
* Restores a warm-up run on CPU. Thanks for @ankalinin for reporting the problem.
* Makes the norm check complementary to point-to-point instead of exclusive.
* Attempts to solve [MFDNN-14550](https://jira.devtools.intel.com/browse/MFDNN-14550) by relying on point-to-point check with an additional custom rule for minor discrepancies.
* Other minor improvements.

### DIFF
--- a/tests/benchdnn/utils/compare.cpp
+++ b/tests/benchdnn/utils/compare.cpp
@@ -127,13 +127,6 @@ bool negative_converts_to_zero(const attr_t &attr, dnnl_data_type_t target_dt) {
     return false;
 }
 
-void reset_case_stats(res_t *res) {
-    // TODO: introduce res->stats and ctor to replace just the stats part.
-    res->state = EXECUTED;
-    res->total = 0;
-    res->errors = 0;
-}
-
 } // namespace
 
 bool compare_extreme_values(float a, float b) {
@@ -643,7 +636,7 @@ int compare_t::compare(const dnn_mem_t &exp_mem, const dnn_mem_t &got_mem,
         }
 
         if (call_norm_check) {
-            reset_case_stats(res);
+            res->reset_stats(EXECUTED);
             st = compare_norm(exp_mem, got_mem, attr, res);
         }
     }
--- a/tests/benchdnn/utils/res.hpp
+++ b/tests/benchdnn/utils/res.hpp
@@ -108,15 +108,39 @@ struct check_mem_size_args_t {
 };
 
 struct res_t {
-    res_state_t state;
-    size_t errors, total;
-    timer::timer_map_t timer_map;
-    std::string impl_name;
-    std::string prim_ref_repro;
+    // The state of the `res` object. Changes as the flow continues. The typical
+    // progression starts with UNTESTED and follows steps:
+    // Creation: -> INITIALIZED/INVALID_ARGUMENTS/UNIMPLEMENTED/SKIPPED;
+    // Execution: -> EXECUTED;
+    // Result: -> PASSED/FAILED/MISTRUSTED.
+    res_state_t state = UNTESTED;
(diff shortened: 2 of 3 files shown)


dzarukin:
make test
set test_scope=NIGHTLY

tests/benchdnn/rnn/rnn.cpp
```diff
@@ -1054,6 +1044,21 @@ void setup_cmp(compare::compare_t &cmp, const prb_t *prb, data_kind_t kind,
...
+        // be same is unknown. Anyway, reconstruct original DST_LAYER values and
+        // if they are off-by-1, let them through.
+        if (prb->cfg.is_int8(DST_LAYER)
+                && (args.dk == rnn_data_kind2data_kind(DST_ITER)
+                        || args.dk == rnn_data_kind2data_kind(DST_ITER_C))) {
+            const int exp_q = static_cast<int>(
+                    args.exp * prb->data_scale + prb->data_shift);
+            const int got_q = static_cast<int>(
+                    args.got * prb->data_scale + prb->data_shift);
+            return fabs(got_q - exp_q) <= 1;
```
mzhukova: just `abs` maybe since both int?

tests/benchdnn/utils/res.hpp
```diff
@@ -108,15 +108,38 @@ struct check_mem_size_args_t {
...
+    // Registered timers during the run.
     timer::timer_map_t timer_map;
+    // The implementation name of the validated primitive.
     std::string impl_name;
+    // The repro line for a primitive used as a baseline over benchdnn ref.
     std::string prim_ref_repro;
-    std::string reason;
+    // The amount of bytes of 'i'nput and 'o'utput.
     // TODO: fuse `ibytes` and `obytes` into `mem_size_args`.
     size_t ibytes, obytes;
```
mzhukova: (minor) seems like all others are initialized, so probably here `= 0` too?

tests/benchdnn/utils/compare.cpp
```diff
@@ -616,15 +593,54 @@ int compare_t::compare_p2p(const dnn_mem_t &exp_mem, const dnn_mem_t &got_mem,
...
             get_kind_str().c_str(), zero_trust_percent_, add_args.c_str());
-    if (use_norm_) return compare_norm(exp_mem, got_mem, attr, res);
-    return compare_p2p(exp_mem, got_mem, attr, res);
+    auto st = compare_p2p(exp_mem, got_mem, attr, res);
+    if (st != OK) {
+        bool call_norm_check = allow_norm_check_;
+        // Note: the following code specifies additional driver's individual
+        // desires when to enable norm check. This one purely depends on the
+        // result of p2p comparison. So far graph is the only driver needing
+        // such. When this becomes a trend, move it to a registred function
```
mzhukova:
```suggestion
        // such. When this becomes a trend, move it to a registered function
```

tests/benchdnn/rnn/rnn.cpp
dzarukin: Yes, thanks for the suggestion, updated.

tests/benchdnn/utils/res.hpp
dzarukin: Makes sense, updated, thanks.

tests/benchdnn/utils/compare.cpp
dzarukin: Thanks, fixed.

dzarukin:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_rnn

vpirogov (approved):

PR #4834, opened 2026-03-16
cpu: aarch64: add JIT ASIMD implementations for exp-based eltwise

Anndrey24:
# Description

This commit introduces JIT ASIMD implementations for 7 out of 8 exp-based eltwise ops (excluding `gelu_erf`), following the existing JIT SVE algorithms:
 - `tanh`
 - `logistic`
 - `elu`
 - `mish`
 - `swish`
 - `gelu_tanh`
 - `soft_relu`

Certain optimisations (e.g. removing unnecessary `ldr`/`mov` instructions and reducing the number of preserved auxiliary registers) were also applied to those SVE implementations, resulting in performance improvements.

## Performance improvements (f32)

eltwise_speedups

The table above summarises the average benchdnn speedup for each op, aggregated over the following shapes:

- `1536x384`
- `1539x387`
- `1024x4096`
- `1025x4099`
- `4096x4096`
- `4099x4099`

For each shape, measurements were collected using 1, 4, 16, and 64 threads, the reported value being the average across all tested configurations.

The comparison baseline depends on the target platform:

- **c6g**
  - `tanh`, `logistic`, and `elu` are compared against `acl`
  - `mish`, `swish`, `gelu_tanh`, and `soft_relu` are compared against `ref`
- **c7g** / **c8g**
  - all ops are compared against the previous `jit_sve` implementation

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -468,29 +468,28 @@ template <cpu_isa_t isa>
 void jit_uni_eltwise_injector_t<isa>::exp_compute_vector_fwd(
         const TRegS &vmm_src, float min_input, float max_input) {
 
-    const auto &t0 = ZRegS(IDX(vmm_src));
-    const auto &t1 = ZRegS(IDX(vmm_aux0));
-    const auto &t2 = ZRegS(IDX(vmm_aux1));
+    const auto &t0 = vmm_src;
+    const auto &t1 = vmm_aux0;
+    const auto &t2 = vmm_aux1;
     const float ln_flt_max = logf(FLT_MAX);
     const float ln_flt_min = logf(FLT_MIN);
     if (max_input > ln_flt_max)
-        h->fmin(t0, p_all, ZRegS(IDX(table_val(exp_ln_flt_max_f, z_tmp))));
+        h->fmin(t0, p_all, table_val(exp_ln_flt_max_f, z_tmp));
     if (min_input < ln_flt_min)
-        h->fmax(t0, p_all, ZRegS(IDX(table_val(exp_ln_flt_min_f, z_tmp))));
-    h->fmul(t0, t0, ZRegS(IDX(table_val(exp_log2ef, t1))));
+        h->fmax(t0, p_all, table_val(exp_ln_flt_min_f, z_tmp));
+    h->fmul(t0, t0, table_val(exp_log2ef, t1));
     h->frintm(t1, p_all, t0);
     h->fcvtzs(t2, p_all, t1);
     h->fsub(t1, t0, t1);
-    h->fadd(t0, t1, ZRegS(IDX(table_val(one, z_tmp))));
+    h->fadd(t0, t1, table_val(one, z_tmp));
     h->lsr(t1, t0, 17);
     h->fexpa(t1, t1);
     h->fscale(t1, p_all, t2);
     h->and_(ZRegD(t2.getIdx()), ZRegD(t0.getIdx()),
             ZRegD(IDX(table_val(exp_not_mask17, z_tmp))));
     h->fsub(t2, t0, t2);
-    h->fmad(ZRegS(IDX(table_val(exp_coeff2, t0))), p_all, t2,
-            ZRegS(IDX(table_val(exp_coeff1, z_tmp))));
-    h->fmad(t0, p_all, t2, ZRegS(IDX(table_val(one, z_tmp))));
+    h->fmad(table_val(exp_coeff2, t0), p_all, t2, table_val(exp_coeff1, z_tmp));
+    h->fmad(t0, p_all, t2, table_val(one, z_tmp));
     h->fmul(t0, t1, t0);
 }
 
@@ -556,8 +555,7 @@ void jit_uni_eltwise_injector_t<isa>::tanh_polynomial_approx_compute_vector_fwd(
         h->add_imm(h->X_TMP_1, x_table,
                 table_off(tanh_pol_table, coeff_idx * tanh_n_polynomials),
                 h->X_TMP_0);
-        h->ld1w(ZRegS(IDX(vmm_coeff)), p_all,
-                ptr(h->X_TMP_1, ZRegS(IDX(vmm_pol_idx)), SXTW));
+        h->ld1w(vmm_coeff, p_all, ptr(h->X_TMP_1, vmm_pol_idx, SXTW));
     };
 
     // because tanh(x) = -tanh(-x), we extract sign to make x postive
@@ -565,8 +563,7 @@ void jit_uni_eltwise_injector_t<isa>::tanh_polynomial_approx_compute_vector_fwd(
     h->fabs(vmm_src_pos, p_all / T_z, vmm_src);
 
     // Compute indices for the table lookup
-    h->sub(ZRegS(IDX(vmm_indices)), ZRegS(IDX(vmm_src_pos)),
-            ZRegS(IDX(table_val(tanh_idx_bias, z_tmp))));
+    h->sub(vmm_indices, vmm_src_pos, table_val(tanh_idx_bias, z_tmp));
     h->and_(ZRegD(IDX(vmm_indices)), ZRegD(IDX(vmm_indices)),
             ZRegD(IDX(table_val(tanh_idx_mask, z_tmp))));
     h->lsr(ZRegD(IDX(vmm_indices)), ZRegD(IDX(vmm_indices)), 20);
@@ -574,7 +571,7 @@ void jit_uni_eltwise_injector_t<isa>::tanh_polynomial_approx_compute_vector_fwd(
     // Argument reduction
     h->and_(ZRegD(IDX(vmm_src_shift)), ZRegD(IDX(vmm_src_pos)),
             ZRegD(IDX(table_val(tanh_idx_mask, z_tmp))));
-    h->fsub(vmm_src_pos, vmm_src_pos, ZRegS(IDX(vmm_src_shift)));
+    h->fsub(vmm_src_pos, vmm_src_pos, vmm_src_shift);
 
     gather_coefficient(vmm_pol, 6, vmm_indices);
     for (int deg = 5; deg >= 0; --deg) {
@@ -799,7 +796,7 @@ void jit_uni_eltwise_injector_t<isa>::soft_relu_compute_vector_fwd(
 
     // calculate exp(x)
     // fx = x * log2ef + 0.5
-    h->fmul(vmm_src, vmm_src, ZRegS(IDX(table_val(exp_log2ef, z_tmp))));
+    h->fmul(vmm_src, vmm_src, table_val(exp_log2ef, z_tmp));
     h->fadd(vmm_src, p_all / T_m, 0.5f);
 
     // tmp = floorf(fx)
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -753,13 +745,13 @@ void jit_uni_eltwise_injector_t<isa>::mish_compute_vector_fwd(
     h->fmul(vmm_src, vmm_src, vmm_src);
 
     // save (e^x+1)^2 as it appears in both the denominator and the numerator
-    h->mov(ZRegD(vmm_aux1.getIdx()), ZRegD(vmm_src.getIdx()));
+    h->mov(ZRegD(vmm_aux0.getIdx()), ZRegD(vmm_src.getIdx()));
```
jondea: It may not have any perf effect (although it will make the code clearer), but we can fuse this move into the next fadd e.g. something like

```
h->fadd(vmm_aux0, p_all / T_m, vmm_src, 1.);
```

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -753,13 +745,13 @@ void jit_uni_eltwise_injector_t<isa>::mish_compute_vector_fwd(
...
 
     // x * ((e^x + 1)^2 - 1) / ((e^x + 1)^2 + 1)
     h->fsub(vmm_src, p_all / T_m, 1.);
-    h->fadd(vmm_aux1, p_all / T_m, 1.);
-    h->fdiv(vmm_src, p_all / T_m, vmm_aux1);
-    h->fmul(vmm_src, vmm_src, vmm_aux3);
+    h->fadd(vmm_aux0, p_all / T_m, 1.);
+    h->fdiv(vmm_src, p_all / T_m, vmm_aux0);
+    h->fmul(vmm_src, vmm_src, vmm_aux2);
```
jondea, line 745: This may have non-trivial numerical effects, but we should be able to fuse this fmul into either the fadd or fsub because div and mul commute

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -908,10 +898,11 @@ void jit_uni_eltwise_injector_t<isa>::soft_relu_compute_vector_fwd(
     if (alpha_ == 1.f) { // standard soft_relu case
         // Skip an instruction.
     } else if (alpha_ == -1) { // logsigmoid case
-        /* Do not use -1.f, which is a float constant,
-       but -1., which is a double constant. */
-        h->fmov(z_tmp, -1.);
-        h->fmul(vmm_src, vmm_src, z_tmp);
+        h->fneg(vmm_src, p_all / T_m, vmm_src);
+    } else if (alpha_ == 0.5) {
```
jondea: I don't have strong feelings about which one is best, we should probably be consistent with the previous
```
if (alpha_ == 0.5 || alpha_ == 2.0) 
```

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -922,53 +913,42 @@ void jit_uni_eltwise_injector_t<isa>::logistic_compute_vector_fwd(
...
 
+    // Store the original sign and make x negative
+    constexpr unsigned sign_mask = 0x80000000;
+    h->fcmgt(p_mask.s, p_all / T_z, vmm_src, 0.);
+    h->orr(vmm_src, sign_mask);
+
+    // Compute exponent
+    const auto old_max_input = max_input_;
+    max_input_ = 0.f;
     exp_compute_vector_fwd(vmm_src);
```
jondea: Modifying state feels a little bit brittle. Given how commonly exp is used by other algorithms, it might make sense to make a stateless `exp_compute_vector_fwd` function with a default e.g. something like

```
void exp_compute_vector_fwd(TReg vmm_src, float  min = min_input_,  float max = max_input_);
```

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2695,6 +2678,96 @@ void jit_uni_eltwise_injector_t<asimd>::relu_compute_vector_fwd(
...
+    h->mov(VReg16B(vmm_aux5.getIdx()), VReg16B(vmm_src.getIdx()));
+
+    // Compute exponent
+    const auto old_max_input = max_input_;
+    max_input_ = 0.f;
+    exp_compute_vector_fwd(vmm_src);
+    max_input_ = old_max_input;
+
+    // alpha * (exp(x) - 1) = alpha * exp(x) - alpha
+    if (alpha_ != 1.f) { h->fmul(vmm_src, vmm_src, vmm_aux4); }
```
jondea: We can do
```
if (alpha_ != 1.f) {
FMLS
} else {
FSUB with immediate
}
```

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2695,6 +2678,96 @@ void jit_uni_eltwise_injector_t<asimd>::relu_compute_vector_fwd(
...
+            VReg16B(vmm_aux0.getIdx()));
+}
+
+template <>
+void jit_uni_eltwise_injector_t<asimd>::tanh_compute_vector_fwd(
+        const TRegS &vmm_src) {
+    // tanh(x) = x(1 + (-1/3)x^2) for |x| < tanh_range
+    // tanh(x) = 1 - 2/(1 + exp(2 x)) for otherwise
+
+    const auto &t0 = VReg4S(IDX(vmm_src));
```
jondea: Can we just do

```
const auto t0 = vmm_src;
```

These should already be VReg4S I think

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I might be wrong, but I think the "with immediate" variant of FADD only supports having the same src and dst register. However, it looks like a version that loads and uses the table constant `one` instead of the immediate is slightly faster than the previous approach, so I've changed it to that.

I've also realised that we have a `max_input_` for the exponential thanks to the prior FMINNM instruction, which reduces the execution time too.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I've changed it to:
```
    } else if (alpha_ == 0.5 || alpha_ == 2.0) {
        h->fmul(vmm_src, p_all / T_m, 1.f / alpha_);
```
I wasn't sure if I could rely on computing `1.f / alpha_`, but it seems to result in the correct f32 values.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I think using FMLS here requires adding an extra FNEG too, though the instruction should be pipelined fairly efficiently, as it is completely independent of the exp portion that comes before it. This approach is a bit faster than before.

Also, I believe there isn't a "with immediate" variant for ASIMD FSUB, though that shouldn't matter in this case, since the alpha vector register (vmm_aux4) would already contain 1.0f on that branch, so we can just keep using that.

I've also changed the SVE implementation to use the FNMSB instruction, since I missed that before. This finally brings some speedups to SVE ELU too, making the entire speedup table green!

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Yes, it's completely redundant. I've now removed it in a few more functions that this PR has touched, but there are still over 70 occurrences of `ZReg(IDX(...))` left, most of them unnecessary. 

I can also change those, but I'm wary of bloating this PR with too many unrelated changes. Do you think it's worth opening a separate PR for that?

Anndrey24:
> What were the single threaded comparisons between ACL and the new asimd impls? ACL threading typically has a bit of overhead, and eltwise is particularly hard to scale to 64 threads, so I'm worried that the high thread count/small problems may be tipping the scales when you take the average. I don't think we need to be faster than ACL single threaded to get this in, but it would tell us that there are further optimizations (which could come later).

@jondea As you said, JIT:ASIMD vs ACL c6g speedups tend to increase with higher thread counts, though the JIT implementation also seems to match the latter's single-threaded performance too. (`tanh` looks like a bit of an outlier, though)
asimd_vs_acl

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Agreed! I've overloaded the function with this new signature, and the original `exp_compute_vector_fwd(const TRegS &vmm_src)` now acts as a thin wrapper around it.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I've tried a version where the multiplication by x takes place before the division:

```
// (e^x+1)^2
h->fadd(vmm_src, p_all / T_m, 1.);
h->fmul(vmm_src, vmm_src, vmm_src);
// A = (e^x+1)^2
// x * (A - 1) / (A + 1) = (x * (A - 1)) / (A + 1)
h->fadd(vmm_aux0, vmm_src, table_val(one, z_tmp));
h->fsub(vmm_src, p_all / T_m, 1.);
h->fmul(vmm_src, vmm_src, vmm_aux2);
h->fdiv(vmm_src, p_all / T_m, vmm_aux0);
```

It seems to lead to the numerator overflowing to +inf:
```
[   8][DST][0:8] exp_f32:     88.5327 exp:     88.5327 got:         inf diff:     inf rdiff:     inf
[  10][DST][0:10] exp_f32:     44.6536 exp:     44.6536 got:         inf diff:     inf rdiff:     inf
[  17][DST][0:17] exp_f32:         100 exp:         100 got:         inf diff:     inf rdiff:     inf
[  21][DST][0:21] exp_f32:     88.2776 exp:     88.2776 got:         inf diff:     inf rdiff:     inf
[  23][DST][0:23] exp_f32:     44.6202 exp:     44.6202 got:         inf diff:     inf rdiff:     inf

```

jondea: Hmmm, interesting, thanks for trying!

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: Thanks. Entirely up to you, one thing you can do is just put it in a separate commit which is marked as no functional change. Happy for you to leave the other occurences too

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: I did forget it was `1/alpha`, but glad it worked!

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: I appreciate you looking into these.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -532,9 +543,9 @@ void jit_uni_eltwise_injector_t<isa>::tanh_polynomial_approx_compute_vector_fwd(
     const int tanh_n_polynomials = 32;
 
     // Register mapping
-    TRegS vmm_dst = vmm_aux1, vmm_src_shift = vmm_aux1, vmm_coeff = vmm_aux1,
-          vmm_pol = vmm_aux2, vmm_indices = vmm_aux3, vmm_tmp = vmm_aux3,
-          vmm_src_pos = vmm_aux4, vmm_sign = vmm_aux4;
+    TRegS vmm_dst = vmm_aux0, vmm_src_shift = vmm_aux0, vmm_coeff = vmm_aux0,
```
Sqvid, line 546: nit: would be nice to have these on separate lines for readability.

vpirogov:
@jondea, @Sqvid, just a reminder that merging PRs requires write permissions. So @Anndrey24 would need help when it's ready.

PR #4836, opened 2026-03-16
Prefill detection and Refactor grp_gemm strategy handling and GRF flag application

umar456:
# Description

* We now check the strategy to determine if the larger register file flag is set in grouped gemm kernels. This was causing failures when the BBS kernels were launching on Xe2 with strategies from PVC. The current register detection did not sufficiently detected if we needed the flag so for some problems the kernel would freeze on Xe2. 
* Added logic to detect "prefill" shapes based on the average number of entries per expert, influencing kernel configuration and parallelism. Prefill shapes now use different tile sizes and parallelism settings for optimal execution. [[1]](diffhunk://#diff-1978bd9fad370b8eed3ebf7df3cc0ef78a2fd1786cbca27fec39bfa7f5eefd45R275-R279) [[2]](diffhunk://#diff-1978bd9fad370b8eed3ebf7df3cc0ef78a2fd1786cbca27fec39bfa7f5eefd45L159-R159)
* The initialization of microkernels is moved to the primitive descriptor (`pd_t`). [[1]](diffhunk://#diff-1978bd9fad370b8eed3ebf7df3cc0ef78a2fd1786cbca27fec39bfa7f5eefd45L38-R38) [[2]](diffhunk://#diff-1978bd9fad370b8eed3ebf7df3cc0ef78a2fd1786cbca27fec39bfa7f5eefd45L58-R60)
* Removed BBS microkernels entries for Xe2 systems now that the GRF flag should address any freezing issues we were experiencing earlier

### DIFF
--- a/src/gpu/intel/matmul/grouped_micro_gemm.cpp
+++ b/src/gpu/intel/matmul/grouped_micro_gemm.cpp
@@ -89,7 +89,7 @@ status_t grouped_micro_gemm_t::pd_t::init_microkernels(impl::engine_t *engine) {
     opts.scaleB = src_quant_.with_scale() && src_group_sizes_[1] < K();
     opts.offsetB = src_quant_.with_zp();
     opts.slmPtr = true;
-    opts.kParallelLocal = true;
+    opts.kParallelLocal = is_gemv_;
 
     if (opts.scaleA) {
         data_type_t wei_scale_dt = wei_quant_.scale_dt();
@@ -155,7 +155,7 @@ status_t grouped_micro_gemm_t::pd_t::init_microkernels(impl::engine_t *engine) {
 
     SizeParams sizes;
     sizes.m = static_cast<uint16_t>(N());
-    sizes.n = static_cast<uint16_t>(M());
+    sizes.n = is_gemv_ ? 1 : 32;
     sizes.k = static_cast<uint16_t>(K());
 
     auto strat_override = [&](gemmstone::GEMMStrategy &strat) {
@@ -272,6 +272,8 @@ status_t grouped_micro_gemm_t::pd_t::init(impl::engine_t *engine) {
             (int)src_grouped.group_count, (int)dst_grouped.group_count);
 
     ngroups_ = src_grouped.group_count;
+    is_gemv_ = M() < ngroups_;
+
     // only supported dt for now
     VDISPATCH_MATMUL(utils::one_of(src_dt, f32, f16, bf16, u8, s8, s4, u4),
             VERBOSE_UNSUPPORTED_DT_CFG);
@@ -433,6 +435,8 @@ status_t grouped_micro_gemm_t::pd_t::init(impl::engine_t *engine) {
     auto bia_dt = weights_md(1)->data_type;
     def_data_type(kernel_ctx_, bia_dt, "BIA");
     kernel_ctx_.define_int("WITH_BIAS", with_bias());
+    kernel_ctx_.define_int("K_PARALLEL_LOCAL", is_gemv_);
+    kernel_ctx_.define_int("WITH_SLM", gemm_.getSetting("slm_size") > 0);
 
     return status::success;
 }
--- a/src/gpu/intel/matmul/grouped_micro_gemm.hpp
+++ b/src/gpu/intel/matmul/grouped_micro_gemm.hpp
@@ -70,6 +70,7 @@ struct grouped_micro_gemm_t : public primitive_t {
         status_t init(impl::engine_t *engine);
         status_t init_microkernels(impl::engine_t *engine);
 
+        bool is_gemv_ = false;
         int sg_size_ = 0;
         int strategyGRFs_ = 0;
         dim_t ngroups_ = 0;
--- a/src/gpu/intel/matmul/grouped_micro_gemm.cl
+++ b/src/gpu/intel/matmul/grouped_micro_gemm.cl
@@ -68,6 +68,7 @@ DECLARE_2D_TILE(c_tile_type_dst, DST_DATA_T, SUBGROUP_SIZE,
     OPTIONAL(AND(WITH_WEI_SCALES, WEI_SCALES_GROUPED), wei_attr_scales)
 #define WEI_ZP_ARGS OPTIONAL(WITH_WEI_ZP, wei_attr_zp)
 #define WEI_LD_ARGS OPTIONAL(OR(WITH_WEI_ZP, WEI_SCALES_GROUPED), ldweiq)
+#define K_PARALLEL_LOCAL_ARGS OPTIONAL(K_PARALLEL_LOCAL, sg_k)
 
 void store_results(ugemm_grouped_c_type *tile, global DST_DATA_T *ptr, int n,
         int m, int lddst, int sg_i0, int sg_j0) {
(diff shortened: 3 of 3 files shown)


umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -389,69 +392,61 @@ status_t grouped_micro_gemm_t::pd_t::init(impl::engine_t *engine) {
...
-
-    pd()->src_quant_.define_macros(kernel_ctx_, "SRC");
-    pd()->wei_quant_.define_macros(kernel_ctx_, "WEI");
+    src_quant_.define_macros(kernel_ctx_, "SRC");
+    wei_quant_.define_macros(kernel_ctx_, "WEI");
 
     kernel_ctx_.set_data_type(dst_dt);
 
-    if (gemm_.grfMin > 128)
+    if (gemm_.grfMin > 128 || strategyGRFs_ > 128)
```
Simonsays095, line 399: Do you know why these kernels were hanging before this fix? In those kernels, were we somehow getting `grfMin <= 128` while actually using `>=r128` in the microkernel?

src/gpu/intel/matmul/grouped_micro_gemm.hpp
```diff
@@ -68,23 +68,25 @@ struct grouped_micro_gemm_t : public primitive_t {
         DECLARE_COMMON_PD_T("grouped_gemm:micro", grouped_micro_gemm_t);
 
         status_t init(impl::engine_t *engine);
+        status_t init_microkernels(impl::engine_t *engine);
```
Simonsays095, line 71: Could you please discuss the reasoning behind moving this logic into `pd_t`? It's not clear to me if this refactor is required for some bugfix or if it's just supposed to simplify things.

src/gpu/intel/matmul/grouped_micro_gemm.hpp
```diff
@@ -68,23 +68,25 @@ struct grouped_micro_gemm_t : public primitive_t {
...
 
         int sg_size_ = 0;
+        int strategyGRFs_ = 0;
         dim_t ngroups_ = 0;
+        bool is_prefill_ = false;
```
Simonsays095: Can we use a variable name that's more reflective of the intent? Something like `is_more_tokens_than_experts`, or `allow_kparallel`? We don't use the term "prefill" anywhere else in oneDNN.

Also, could we add a comment clarifying the intent of this variable, indicating that it's a workaround until we get better local k-parallelism dispatching?

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -274,6 +272,11 @@ status_t grouped_micro_gemm_t::pd_t::init(impl::engine_t *engine) {
             (int)src_grouped.group_count, (int)dst_grouped.group_count);
 
     ngroups_ = src_grouped.group_count;
+
+    // Prefill shapes have many entries per expert. We are determining
+    // prefill shapes checking the average number of entries per expert
+    is_prefill_ = (M() / static_cast<float>(ngroups_)) > 1.f;
```
Simonsays095:
```suggestion
    is_prefill_ = (M() > ngroups_);
```

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: Not sure the exact reason. The main issue was the strategy was using grf256 but the calculation that the package performs to calculate the grf requirement was lower than 128 causing the issue. One of the things that the GRF strategy defines is the number of threads per EU in the gemm calculation. I am assuming this was causing the failures in Xe2 even though the same code worked in PVC.

src/gpu/intel/matmul/grouped_micro_gemm.hpp
umar456: The refactor was required becasue the pd was const when we are selecting the microkernel so the strategyGRFs_ variable required the removal of the constness of the pd. This seemed like the incorrect approach so I brought the microkenrel generation into the pd init.

src/gpu/intel/matmul/grouped_micro_gemm.hpp
umar456: I added this variable to match the way OpenVINO was configuring the kernel. I would like to remove this at some point so that we can select the right problem based on better heuristics than this approach. I can change it but these changes will be temporary either way.

umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu

src/gpu/intel/gemm/jit/selector/db/ukernel_mmr.db
```diff
@@ -629,10 +629,6 @@ auto _CATALOG_ = kcatalog::toArray({
...
 {{'G', "ugemm", {"F", "H", "S"}, {"T", "N", "N"}}, {-1, -1, {-1, -1, -1}, {-1, 8, -1}, {-1, -1, -1}, {-1, 8, -1}, {16, 16, 1}, "IAB"}, "at32 am128 rB wg 2x1x8 ikr xaf st vav sb128 nmk sys", {16, (LoopType) 255, 128, {(LoopType) 209, (LoopType) 255, (LoopType) 2}, {16777216, 131072, 16777216}, {262144, 131072, 16777216}, {16, 8, 128}, {2, 1, 8}, 1, (WGType) 0, 4357, 0, 1024, {16, 16, 4}, {true, true, true}}, {'W', 1, {128}}},
-{{'G', "ugemm", {"B", "B", "S"}, {"N", "N", "N"}}, {-1, -1, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {-1, -1, -1}, {16, 16, 1}, "ABI"}, "av64x2+m48@40 am64x2+m32@88 rB wg 4x1x8 kr sys af vav sb128 sm dm grf256", {16, (LoopType) 255, 256, {(LoopType) 144, (LoopType) 255, (LoopType) 2}, {524288, 65536, 16777216}, {524288, 65536, 16777216}, {32, 4, 64}, {4, 1, 8}, 1, (WGType) 1, 268435717, 0, 2048, {16, 16, 4}, {true, true, true}}, {'E', 17, {576908, -9616.73, 1812.82, -16.2442, 0, 0, 1.71218, 16.8178, 3.85905, 5.67627, 0.13492, 0.13492, 0, 0.0814212, 1.00074, 1.00076, -2.21732e-15}}},
```
kealan-barbieri, line 632: so is the idea these strategies are unnecessary now and the PVC strategy for this signature will be used with smaller GRF? is it worth testing perf?

umar456: Yes, correct. Additionally these strategies seem to be slower than their PVC counterparts for the sizes OpenVINO were interested in with their current implementation.

PR #4838, opened 2026-03-17
graph: enable quantized gated mlp dispatch

TaoLv:
1. The example and test files in benchdnn are updated with f32 intermediate data type. (Need Arch review.)
2. Graph backend gated mlp kernel to support quantized inputs.
3. Dispatch quantized gated mlp patterns to the quantized gated mlp primitive.

Still, the dispatching is disabled by default, to enable it, set `_ONEDNN_GRAPH_GATED_MLP_FORCE_PRIMITIVE=0`.

### DIFF
--- a/src/graph/backend/dnnl/patterns/mlp.cpp
+++ b/src/graph/backend/dnnl/patterns/mlp.cpp
@@ -15,7 +15,6 @@
 *******************************************************************************/
 
 #include "graph/backend/dnnl/kernels/gated_mlp.hpp"
-#include "graph/backend/dnnl/kernels/large_partition.hpp"
 
 #include "graph/backend/dnnl/patterns/fusions.hpp"
 #include "graph/backend/dnnl/patterns/pattern_matcher_pass.hpp"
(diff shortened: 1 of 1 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/passes/transform.cpp
```diff
@@ -4656,11 +4660,64 @@ status_t fuse_gated_mlp(std::shared_ptr<subgraph_t> &sg) {
         return status::unimplemented;
     }
 
+    fusion_info_t fusion_info;
+    if (gate->has_attr(op_attr::fusion_info)) {
```
mzhukova: (minor) seems like a repetitive code, that could be moved to some helper function with ({gate, up, down}, DNNL_ARG_WEIGHTS_*) arguments

TaoLv: Good point. It's changed now.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #4843, opened 2026-03-17
doc: specified primitive arguments mutability

vpirogov:
This PR updates primitives documentation Developer Guide to specify role of every execution argument as input or output.

We have a few cases where the answer is "It's complicated", in particularly with mean and variance arguments for convolution. This seems to be sufficiently covered in accompanying tables explaining impact of flags.

### DIFF
--- a/doc/primitives/batch_normalization.md
+++ b/doc/primitives/batch_normalization.md
@@ -111,21 +111,24 @@ requires different inputs and outputs.  For clarity, a summary is shown below.
 When executed, the inputs and outputs should be mapped to an execution
 argument index as specified by the following table.
 
-| Primitive Input/Output    | Execution Argument Index |
-|---------------------------|--------------------------|
-| \src                      | DNNL_ARG_SRC             |
-| \f$\src_1\f$              | DNNL_ARG_SRC_1           |
-| \f$\gamma\f$              | DNNL_ARG_SCALE           |
-| \f$\beta\f$               | DNNL_ARG_SHIFT           |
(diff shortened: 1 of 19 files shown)


doc/primitives/batch_normalization.md
```diff
@@ -111,21 +111,21 @@ requires different inputs and outputs.  For clarity, a summary is shown below.
...
-| mean (\f$\mu\f$)          | DNNL_ARG_MEAN            |
-| variance (\f$\sigma^2\f$) | DNNL_ARG_VARIANCE        |
-| \dst                      | DNNL_ARG_DST             |
-| workspace                 | DNNL_ARG_WORKSPACE       |
-| \diffdst                  | DNNL_ARG_DIFF_DST        |
-| \diffsrc                  | DNNL_ARG_DIFF_SRC        |
-| \f$\diffsrc_1\f$          | DNNL_ARG_DIFF_SRC_1      |
-| \f$\diffgamma\f$          | DNNL_ARG_DIFF_SCALE      |
-| \f$\diffbeta\f$           | DNNL_ARG_DIFF_SHIFT      |
+| Argument                  | Index                    | Type         |
```
densamoilov, line 114: (random spot)
nit: should we also add the same note for scratchpad https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_scratchpad.html#index-0?

vpirogov: Good point. Added a mention of scratchpad argument as well. I'm on the fence between classifying it as 'output' or introducing something like `temporary`

PR #4848, opened 2026-03-18
xe: reduce build time with Xe3p

echeresh:
This reduces ~5.5 minutes build time down to ~2.5 minutes (on my machine) by parallelizing Xe3p GEMM generator sub-targets.

In the long term we need to unify all these sub-targets: [MFDNN-14690](https://jira.devtools.intel.com/browse/MFDNN-14690)


echeresh:
make test

Issue #470, opened 2019-04-30
How to retreive the data format from a memory descriptor ?


zeno40:
How can I get the mkldnn_format_tag_t value from a memory descriptor?

thanks

emfomenk:
Hi @zeno40,

According to the [RFC: Intel(R) MKL-DNN API 1.0](https://github.com/intel/mkl-dnn/pull/384) there is no more possible to retrieve the tag from the memory descriptor. The tags are only used to specify the memory descriptor format, but the actual information is stored in `md.format_desc.{blocking,other..}`.
See the last but one paragraph in section [5. Memory descriptor rework](https://github.com/intel/mkl-dnn/tree/rfc-api-changes-v1.0/doc/rfc/api-v1.0#5-memory-descriptor-rework) of the RFC.

**The rational**: there might be different tags correspond to the same memory format. Also a user might want to initialize memory descriptor simply by specifying the strides (i.e. no tags are used at all).

dev2y:
I too run into same issue. mem::desc().data.format doesn't work any more. We are updating to MKLDNN 1.0 and we check formats and decide to reorder or not. format_kind is different than format_tag. 

auto temp = conv1_prim_desc.weights_desc().data.format_kind; 

@emfomenk looks like we should compare memory_desc instead of memory format.

I am wondering why format is hidden from memory desc in 1.0. I think we still want to know the memory formats. Especially for Convolution primitive. Any plans to make format_tag accessible from memory description?

    /* create a convolution */
    auto conv1_desc = convolution_forward::desc(prop_kind::forward_inference,
            algorithm::convolution_direct, conv1_src_md, conv1_weights_md, conv1_bias_md,
            conv1_dst_md, conv1_strides, conv1_padding, conv1_padding,
            padding_kind::zero);
    auto conv1_prim_desc = convolution_forward::primitive_desc(conv1_desc, eng);
    if (conv1_prim_desc.src_desc() != user_src_memory.get_desc()) {
        conv1_src_memory = memory(conv1_prim_desc.src_desc(), eng);
        net.push_back(reorder(user_src_memory, conv1_src_memory));
        net_args.push_back({ { MKLDNN_ARG_FROM, user_src_memory },
                { MKLDNN_ARG_TO, conv1_src_memory } });
    }

zeno40:
[...]

PR #4850, opened 2026-03-18
cpu: rv64: vector-length-agnostic RVV JIT 1x1 convolution kernel

alexandrelimassantana:
# Description

This patch includes utility functions, new memory layouts, and related adjustments to extend the RVV JIT 1x1 convolution kernel to 256- and 512-bit vector systems. Specifically, this patch:

- Adds the **get_platform_vlen**() function to the **cpu_isa_traits** as a way of querying the vlen at runtime;
- Adds the **Abc8a** and **aBcdef8b** layout tags and aliases for conv weight tensor used on 256-bit vector platforms;
- Updates the 128-bit **RVV JIT 1x1 convolution kernel** to also accept 256-bit and 512-bit systems.

While vector-length-specific implementations are a good starting point, maximizing the vector register state usage is one of the most impactful software-level optimization on RISC-V vector systems. This patch aims to start this transition by providing a reference on how to build a primitive that can work on any short-vector platforms (i.e. vlen <= cache line) while still using pre-packed weights.

We analyzed the speed-up obtained in Bananapi-f3, an 8-core 256-bit RVV system, running the set_conv_all convolution dataset.
The updated jit_1x1:rvv kernel is selected on 38% of the workloads where it obtains a **geometric mean speedup of 1.78x** by leveraging the full vector register state.
Below are two images showcasing the speedup scatterplot, where the x-axis enumerates the workloads by ascending number of flops, and the speedup histogram, rendered with 20 bins.

conv scatterplot

conv histogram

# Checklist

## General

## Performance improvements

### New features

- [ ] Was the RFC approved?

### Bug fixes

### DIFF
--- a/src/cpu/rv64/cpu_isa_traits.hpp
+++ b/src/cpu/rv64/cpu_isa_traits.hpp
@@ -106,7 +106,7 @@ inline uint32_t get_platform_vlen() {
 inline int get_vlen_implementation_id(int vlen) {
     static constexpr int VLEN_MIN = 128;
     if (math::is_pow2(vlen) && vlen >= VLEN_MIN) {
-        return math::ilog2q(vlen) - (math::ilog2q(VLEN_MIN) + 1);
+        return math::ilog2q(vlen) - math::ilog2q(VLEN_MIN);
     } else {
         return -1;
     }
--- a/src/cpu/rv64/jit_rvv_1x1_conv_kernel.cpp
+++ b/src/cpu/rv64/jit_rvv_1x1_conv_kernel.cpp
@@ -63,12 +63,13 @@ status_t jit_rvv_1x1_conv_kernel_t::init_conf(jit_1x1_conv_conf_t &jcp,
     jcp.oc = jcp.oc_without_padding;
     jcp.ic = jcp.ic_without_padding;
 
-    // Targeting SEW=32 (float), LMUL=1, VLEN=128 -> simd_w = 4
-    const int simd_w = 4;
+    // Targeting SEW=32 (float), LMUL=1, VLEN <= 1024
+    const int vlen = nstl::min(1024u, get_platform_vlen());
+    jcp.simd_w = vlen / (sizeof(float) * 8);
 
-    // OC is padded to match oc_block in weights format (Oihw4o)
+    // OC is padded to match oc_block in weights format (OihwXo)
     // IC is not padded; kernel handles IC tail processing
-    jcp.oc = rnd_up(jcp.oc, simd_w);
+    jcp.oc = rnd_up(jcp.oc, jcp.simd_w);
 
     // 3D convolution support
     jcp.id = (ndims == 5) ? src_d.dims()[2] : 1;
@@ -83,8 +84,8 @@ status_t jit_rvv_1x1_conv_kernel_t::init_conf(jit_1x1_conv_conf_t &jcp,
     jcp.os = jcp.od * jcp.oh * jcp.ow;
     jcp.is = jcp.id * jcp.ih * jcp.iw;
 
-    jcp.oc_block = simd_w;
-    jcp.ic_block = simd_w;
+    jcp.oc_block = jcp.simd_w;
+    jcp.ic_block = jcp.simd_w;
 
     // Dynamic parameter calculation
     // Register constraint: (ur * load_loop_blk) + (unroll * load_loop_blk) + 1 <= 32
--- a/src/cpu/rv64/jit_rvv_1x1_convolution.hpp
+++ b/src/cpu/rv64/jit_rvv_1x1_convolution.hpp
@@ -19,6 +19,7 @@
 
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
+#include "common/math_utils.hpp"
 #include "common/memory_tracking.hpp"
 #include "common/primitive.hpp"
 #include "common/utils.hpp"
@@ -61,33 +62,6 @@ struct jit_rvv_1x1_convolution_fwd_t : public primitive_t {
             VDISPATCH_CONV(post_ops_ok(), VERBOSE_UNSUPPORTED_POSTOP);
             VDISPATCH_CONV(!has_zero_dim_memory(), VERBOSE_EMPTY_TENSOR, "");
 
-            // Only support: data = nwc/nhwc/ndhwc, weights = blocked formats (Oiw4o/gOiw4o/etc)
-            const int n = ndims();
-            const bool g = with_groups();
-            const auto dat_tag_nxc = utils::pick(n - 3, nwc, nhwc, ndhwc);
-            const auto wei_tag_blocked = utils::pick(2 * n - 6 + (g ? 1 : 0),
-                    Oiw4o, gOiw4o, Oihw4o, gOihw4o, Oidhw4o, gOidhw4o);
-
-            // Check if src/dst match supported format (nxc)
-            // Only accept format_kind::any as a fallback, reject explicit
-            // unsupported formats
-            VDISPATCH_CONV(IMPLICATION(src_d.matches_one_of_tag(dat_tag_nxc)
-                                           != dat_tag_nxc,
-                                   src_d.format_kind() == format_kind::any),
-                    VERBOSE_UNSUPPORTED_TAG);
-            VDISPATCH_CONV(IMPLICATION(dst_d.matches_one_of_tag(dat_tag_nxc)
-                                           != dat_tag_nxc,
-                                   dst_d.format_kind() == format_kind::any),
-                    VERBOSE_UNSUPPORTED_TAG);
-            VDISPATCH_CONV(
-                    IMPLICATION(weights_d.matches_one_of_tag(wei_tag_blocked)
-                                    != wei_tag_blocked,
-                            weights_d.format_kind() == format_kind::any),
-                    VERBOSE_UNSUPPORTED_TAG);
-
-            // Set default formats if format_kind == any
-            VDISPATCH_CONV(set_default_formats(), VERBOSE_UNSUPPORTED_TAG);
-
             // ISA check
             VDISPATCH_CONV(mayiuse(v), VERBOSE_UNSUPPORTED_ISA);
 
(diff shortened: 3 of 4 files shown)


zhangfeiv0:
> - Updates the 128-bit **RVV JIT 1x1 convolution kernel** to also accept 256-bit and 512-bit systems.

[Spacemit](https://www.spacemit.com/products/keystone/k3) already has hardware supporting 1024-bit RVV; can the code be extended to support 1024 bits?

src/cpu/rv64/cpu_isa_traits.hpp
```diff
@@ -56,15 +56,18 @@ struct Riscv64Cpu {
 
     bool get_has_v() const { return has_v; }
     bool get_has_zvfh() const { return has_zvfh; }
+    int get_vlen() const { return vlen; }
 
 private:
     bool has_v = false;
     bool has_zvfh = false;
+    int vlen = 0;
```
zhangjian29: (Nit) `xbyak_cpu.getVlen()` returns `uint32_t`, it would be better to mark `vlen` as `uint32_t` here.

src/cpu/rv64/jit_rvv_1x1_convolution.hpp
```diff
@@ -19,9 +19,9 @@
 
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
+#include "common/math_utils.hpp"
 #include "common/memory_tracking.hpp"
 #include "common/primitive.hpp"
-#include "common/utils.hpp"
```
zhangjian29, line 24: (Nit) Suggest to keep this include since `utils::pick` etc. are still used via transitive includes.

zhangjian29:
Hi @alexandrelimassantana, thanks for the patch!

I suppose it would be better if you could address these different issues in separate PRs. E.g., one for new layout tags (common/API changes), and one for kernel VLA changes (rv64 changes).

alexandrelimassantana:
Hi @zhangfeiv0 and @zhangjian29,

I have updated the PR with all of your petitions, including the support for 1024-bit vector architectures.
As requested, I have moved the addition of the new weights tags into the new [PR 4861](https://github.com/uxlfoundation/oneDNN/pull/4861).

I am confused on how to proceed now, this PR now contains a duplicated commit, included in the other PR, that is also necessary for the other two commits. Could you clarify what actions should I take with this PR?

alexandrelimassantana:
Hi @zhangfeiv0 and @zhangjian29,

I would also like to comment something related to this PR.
The main aspect preventing this kernel from being usable on architectures with higher vector length is the need to add new memory tags to cater to longer vector systems.
We do have systems with longer vectors and I would love to hear your comments on [Issue 4860](https://github.com/uxlfoundation/oneDNN/issues/4860), which I created to discuss packed memory layouts for RISC-V.

vpirogov:
> I am confused on how to proceed now, this PR now contains a duplicated commit, included in the other PR, that is also necessary for the other two commits. Could you clarify what actions should I take with this PR?

I thought I posted the comment on this, but apparently it got eaten by /dev/null. The only reason for adding new tags is new kernel introduced in this PR and also changes in this PR are not functional without new tags. So #4861 should be closed.

> The main aspect preventing this kernel from being usable on architectures with higher vector length is the need to add new memory tags to cater to longer vector systems.

Format tags are merely a convenience feature and adding new tags should not be strictly necessary to support new data formats (see discussion in #470). Consider constructing memory descriptors required directly using [`memory_desc_init_by_blocking_desc`](https://github.com/uxlfoundation/oneDNN/blob/main/src/common/type_helpers.hpp#L1227).

src/cpu/rv64/jit_rvv_1x1_convolution.hpp
```diff
@@ -125,15 +123,23 @@ struct jit_rvv_1x1_convolution_fwd_t : public primitive_t {
...
-                    gOiw4o, Oihw4o, gOihw4o, Oidhw4o, gOidhw4o);
-
-            return set_default_formats_common(dat_tag, wei_tag, dat_tag);
+            const int vlen_min = 128 / (sizeof(float) * 8);
+            const int vlid = math::ilog2q(jcp_.oc_block / vlen_min);
+            return utils::pick(vlid * 6 + (2 * n - 6 + (g ? 1 : 0)), Oiw4o,
+                    gOiw4o, Oihw4o, gOihw4o, Oidhw4o, gOidhw4o, Oiw8o, gOiw8o,
+                    Oihw8o, gOihw8o, Oidhw8o, gOidhw8o, Oiw16o, gOiw16o,
+                    Oihw16o, gOihw16o, Oidhw16o, gOidhw16o, Oiw32o, gOiw32o,
+                    Oihw32o, gOihw32o, Oidhw32o, gOidhw32o, Oiw32o, gOiw32o);
```
zhangjian29: (nit) Duplicate entries of `Oiw32o` and `gOiw32o`. Harmless but should be cleaned up.

alexandrelimassantana: good catch, I just cleared this up on the following patch

src/cpu/rv64/jit_rvv_1x1_convolution.hpp
```diff
@@ -125,15 +123,23 @@ struct jit_rvv_1x1_convolution_fwd_t : public primitive_t {
...
+        inline format_tag_t get_wei_tag() const {
             using namespace format_tag;
             const int n = ndims();
             const bool g = with_groups();
-            const auto dat_tag = utils::pick(n - 3, nwc, nhwc, ndhwc);
-            const auto wei_tag = utils::pick(2 * n - 6 + (g ? 1 : 0), Oiw4o,
-                    gOiw4o, Oihw4o, gOihw4o, Oidhw4o, gOidhw4o);
-
-            return set_default_formats_common(dat_tag, wei_tag, dat_tag);
+            const int vlen_min = 128 / (sizeof(float) * 8);
```
dzarukin: If `128` will be extended to `1024`, does it mean we will receive a set of brand new tags to support every combination?

Besides, is it a big difference which O block to pick for a single case? Let's say why not just use the block size of 32 every time?

alexandrelimassantana: Yes, I added the new memory block formats, with block sizes of 32, to support vl=1024.

On 128-bit and 256-bit vector processors, the JIT code generator may leverage vector register grouping to compute the larger 1024-bit vectors of blocked data.
So, no, there is not much of a difference to pick a smaller or larger block conceptually.
However, this is something I wish to address on another PR because the current JIT code generation logic lacks the flexibility to support different vector register allocation schemes.

Using a custom convolution implementation I have, I found empirical evidence that grouping registers, using LMUL, to match vlen=1024 yields the best performance on Milk-V Pioneer (128-bit OoO core) and Bananapi-F3 (256-bit in-order) systems.
The advantages are many, vector grouping reduces the number of dynamic instructions (larger vlen), better instruction cache locality (fewer registers for unrolling), L1D accesses are coallesced (vlen >= L1D line), and others.
Since these advantages did not carry over to vlen=2048 on Bananapi using lmul=8, I believe the best way to handle grouping, and blocks > vlen in general, is to equip JIT code generators with a flexible logic for vector register allocation.
At the moment, equipping the current 1x1 conv primitive with this feature requires a deeper refactor on the code generation logic.
Adding this feature, to 1x1 conv and other primitives, is something I would rather do on its own PR.

dzarukin: @alexandrelimassantana, it feels like you covered a slightly different aspect of the answer I was trying to understand. Reworded question sounds like this: should oneDNN anticipate more format_tags in the next several months that would be a product of a different vector length?

If yes, then the ask would be to figure out if this can be avoided as the current mindset is to keep it at some level of scaling.

If no, and this set introduced in this PR is sufficient for future needs, then feel free to address the coming feedback.

include/oneapi/dnnl/dnnl_types.h
```diff
@@ -242,13 +242,15 @@ typedef enum {
     // Opaque blocked formats
 
     dnnl_Abc16a,
+    dnnl_Abc32a,
```
dzarukin: All new tags (not aliases) must be put right before `tag_last` to preserve ABI compatibility.

src/cpu/rv64/jit_rvv_1x1_conv_kernel.cpp
```diff
@@ -63,12 +63,12 @@ status_t jit_rvv_1x1_conv_kernel_t::init_conf(jit_1x1_conv_conf_t &jcp,
     jcp.oc = jcp.oc_without_padding;
     jcp.ic = jcp.ic_without_padding;
 
-    // Targeting SEW=32 (float), LMUL=1, VLEN=128 -> simd_w = 4
-    const int simd_w = 4;
+    // Targeting SEW=32 (float), LMUL=1
+    jcp.simd_w = nstl::min(1024u, get_platform_vlen()) / (sizeof(float) * 8);
```
dzarukin: Minor: it would benefit to name `1024u` in a common infrastructure as it seems to be some fundamental value.

src/cpu/rv64/jit_rvv_1x1_convolution.hpp
```diff
@@ -125,15 +123,23 @@ struct jit_rvv_1x1_convolution_fwd_t : public primitive_t {
...
+            const int vlid = math::ilog2q(jcp_.oc_block / vlen_min);
+            return utils::pick(vlid * 6 + (2 * n - 6 + (g ? 1 : 0)), Oiw4o,
+                    gOiw4o, Oihw4o, gOihw4o, Oidhw4o, gOidhw4o, Oiw8o, gOiw8o,
+                    Oihw8o, gOihw8o, Oidhw8o, gOidhw8o, Oiw16o, gOiw16o,
+                    Oihw16o, gOihw16o, Oidhw16o, gOidhw16o, Oiw32o, gOiw32o,
+                    Oihw32o, gOihw32o, Oidhw32o, gOidhw32o);
```
dzarukin: Would you mind to document the algorithm of picking a tag and document constants for better understanding the logic behind?

alexandrelimassantana:
@dzarukin this update adds what was requested.

> Should oneDNN anticipate more format_tags in the next several months that would be a product of a different vector length?

No, this is impractical and the current set of memory formats are enough to support implementations with short vectors. Nonetheless, this issue has manifested on our local fork where we support very long vectors (up to 16384-bits). Our solution was to make the [fill_blocked method from memory_desc_wrapper](https://github.com/uxlfoundation/oneDNN/blob/db17ac9b72049677cc837224a5daac14517bbcdc/src/common/memory_desc_wrapper.cpp#L31) accessible to our primitives so we could easily populate blocking descriptors without creating new tags. On the next months, we expect to propose a similar solution to avoid new tags when support vlen > 1024.

> it would benefit to name 1024u in a common infrastructure as it seems to be some fundamental value.

We have labeled this as the implementation limit for the 1x1 primitive for now, since our observations I shared previously are not yet reflected anywhere on the upstream code.

> Would you mind to document the algorithm of picking a tag and document constants for better understanding the logic behind?

The logic was commented and broken down into two steps.
The code handling selection based on the vector length has been moved to a new method [get_vlen_implementation_id](https://github.com/alexandrelimassantana/oneDNN/blob/bsc/vla/src/cpu/rv64/cpu_isa_traits.hpp#L106), which acts similarly to the enumerations that differentiate SVE among sve_128, sve_256, sve_512, but for RISC-V systems.

dzarukin (approved):

PR #4853, opened 2026-03-18
cmake: disable LTO for debug builds

Simonsays095:
Fixes the following warning for builds on Windows:
> LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification

Debug builds will use `/INCREMENTAL` for faster build times, while other builds will use `/LTCG` for more optimization.


cmake/SDL.cmake
```diff
@@ -85,7 +85,10 @@ if(UNIX)
 elseif(WIN32)
     if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
         append(ONEDNN_SDL_COMPILER_FLAGS "/EHsr /GS /Gy /guard:cf /DYNAMICBASE /sdl")
-        append(ONEDNN_SDL_LINKER_FLAGS "/NXCOMPAT /LTCG")
+        append(ONEDNN_SDL_LINKER_FLAGS "/NXCOMPAT")
+        if(NOT UPPERCASE_CMAKE_BUILD_TYPE STREQUAL "DEBUG")
+            append(ONEDNN_SDL_LINKER_FLAGS "/LTCG /INCREMENTAL:NO")
+        endif()
```
vpirogov, line 91: I wonder why `/LTCG` ended up in security flags...
This logic seems to belong in platform.cmake.

avmanerikar: This was bundled as a part of compiler flag requirements when implementing SDL guidelines:
https://readthedocs.intel.com/SecureCodingStandards/latest/compiler/c-cpp/#__tabbed_5_4

Simonsays095: @avmanerikar From my read of that doc, we need to have `/LTCG` enabled for all builds, so I should just set `/INCREMENTAL:NO` for all builds to comply with SDL and silence the warning. Do you agree?

vpirogov: Sounds right.

Note, that we have flexibility with these flags if there are significant adverse effects of using them. Specifically with `/LTCG` I don't see any explanation on what it does to improve security posture in MSVS documentation or Secure Coding Standards.

Simonsays095: Personally, I think it makes more sense to use incremental linking for debug builds to speed up build time in development, and leave `/LTCG` for release builds (to get whatever security benefits it may provide, in production). I'm not sure if "longer build time for debug builds" is significant enough to warrant bending the rules, though.

vpirogov: These are more of a guideline than a hard rule. I'm fine with this approach.

Simonsays095: Thanks, Vadim. I'll leave the logic as-is, and keep it in SDL.cmake since it's a related flag.

PR #4864, opened 2026-03-20
Work around a jit post-op issue with layout not matching dst

karturov:
Fixes MFDNN-14794. Backport to v3.11 (to allow ref path in Graph SDPA training work).

### DIFF
--- a/src/gpu/intel/gemm/jit/gen_kernel.cpp
+++ b/src/gpu/intel/gemm/jit/gen_kernel.cpp
@@ -401,6 +401,20 @@ status_t gen_desc_t::transfer_post_ops(
 
             bool trans = is_multi_row && !src_rmd.inner_dim.is_innermost();
 
+            // Reject row-only or col-only binary post-ops where the
+            // physically innermost dimension is not the varying one
+            // (e.g. batch dim is innermost). Block access requires
+            // contiguous elements along the varying dimension.
+            {
+                int rmd_ndims = src_rmd.ndims();
+                bool row_only = is_multi_row && !is_multi_col;
+                bool col_only = is_multi_col && !is_multi_row;
+                if (row_only && !src_rmd.inner_dim.is_innermost())
+                    return status::unimplemented;
+                if (col_only && !src_rmd.is_inner_dim(rmd_ndims - 2, rmd_ndims))
+                    return status::unimplemented;
+            }
+
             if (swap_ab) {
                 trans = !trans;
                 std::swap(is_multi_row, is_multi_col);


src/gpu/intel/gemm/jit/generator/pieces/gemm_setup.cxx
```diff
@@ -491,8 +491,10 @@ void Generator<hw>::gemmOffsetABC(bool initial, Subregister i0, Subregister j0,
             eadd(1, offset, offset, tempQ0.reinterpret(0, offset.getType()), strategy, state);
         } else if (row)
             eaddScaled(1, offset, offset, i0, T, strategy, state);
-        else if (col)
-            eaddScaled(1, offset, offset, j0, T, strategy, state);
+        else if (col) {
+            emul(1, tempQ0, j0, ld, strategy, state);
+            eadd(1, offset, offset, tempQ0.reinterpret(0, offset.getType()), strategy, state);
+        }
```
Simonsays095: These changes don't look right. The intent is that we pass `state.inputs.binaryLDs` only when we have 2D post-ops. For either row or column (1D) post-ops, we assume packed data, and here we scale by the data type size to get the stride.

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -412,6 +412,17 @@ status_t gen_desc_t::transfer_post_ops(
...
+            // For col-only binary, check whether the col direction has unit stride.
+            // Col direction corresponds to dim[ndims-2] (no swap) or dim[ndims-1] (swap_ab).
+            bool col_only = is_multi_col && !is_multi_row;
+            int rmd_ndims = src_rmd.ndims();
+            bool col_unit = !col_only
+                    || (swap_ab ? src_rmd.is_inner_dim(rmd_ndims - 1, rmd_ndims)
+                                : src_rmd.is_inner_dim(rmd_ndims - 2, rmd_ndims));
+            // Non-unit stride requires Scattered access with correct multi-block offsets,
+            // which is not yet supported. Reject at PD time to avoid silent wrong results.
+            if (!col_unit) return status::unimplemented;
```
Simonsays095: This fix may work, but only this block of code is needed. Everything else looks wrong. Please also move this to before the `if (swap_ab)` block above, and assume `swap_ab = false`, then inside of the `if (swap_ab)` block do any necessary "fixing up". Eventually this logic will be moved somewhere else, so we need to keep it separate from the base logic.

atkassen: ~~Could we force 2d in these cases? E.g.~~
```c++
        if (!is_multi_row && is_multi_col && !src_rmd.is_inner_dim(rmd_ndims - 2, rmd_ndims))
            is_multi_row = true;
```

edit: nevermind, I thought the problem shape had N=1.

Simonsays095:
@rjoursler Please comment on the logic related to `relative_md_t`. I'm not that familiar with the semantics of `is_innermost` and relative mds, so I want to make sure you're on board with the changes.

karturov:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -401,6 +401,20 @@ status_t gen_desc_t::transfer_post_ops(
...
+            // (e.g. batch dim is innermost). Block access requires
+            // contiguous elements along the varying dimension.
+            {
+                int rmd_ndims = src_rmd.ndims();
+                bool row_only = is_multi_row && !is_multi_col;
+                bool col_only = is_multi_col && !is_multi_row;
+                if (row_only && !src_rmd.inner_dim.is_innermost())
+                    return status::unimplemented;
+                if (col_only && !src_rmd.is_inner_dim(rmd_ndims - 2, rmd_ndims))
+                    return status::unimplemented;
```
rjoursler, line 415: I don't see how this addresses the linked issue. The logic here is effectively
```
if(is_1d &&  ...) return status::unimplemented
```
but the post-op mentioned in the ticket is not a 1D post-op.

vpirogov:
Closed as stale.

PR #4865, opened 2026-03-20
common: verbose: drop extra CSV header column

jondea:
# Description

Fix verbose CSV headers for primitive and graph output by removing an extra header-only column.

The verbose header printed `template:operation` / `template:timestamp,operation`, but the corresponding data rows never had a column which corresponded to `operation`. This made the header wider than the data and broke CSV tools such as `xsv table`. This change updates the primitive and graph headers to match the emitted data rows, and refreshes the verbose documentation examples.

## Details 

```sh
ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn --mode=r --matmul 1x1:1x1 \
    | grep 'onednn_verbose,v1,primitive' \
    | xsv table
```

Before the fix, this failed because the verbose header had one more CSV field than the emitted primitive rows, e.g.

```text
onednn_verbose,v1,primitive,exec,cpu,matmul,gemm:jit:f32,undef,src:f32:a:blocked:ab::f0 wei:f32:a:blocked:ab::f0 dst:f32:a:blocked:ab::f0,,,1x1:1x1,0.0170898
```

After the fix, the header is:
```text
onednn_verbose,v1,primitive,info,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
```
and the `xsv table` command above succeeds.

The same issue existed in the graph verbose header and can be checked with:
```sh
ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn --mode=P --graph --case=op/f32/abs.json \
    | grep 'onednn_verbose,v1,graph' \
    | xsv table
```

# Checklist

## General

### Bug fixes

- [ ] Have you added relevant regression tests? I don't think there's a simple way to add a C++ `ctest` here, especially without adding dependencies, but ideas are welcome!

### DIFF
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
 * Copyright 2018 Intel Corporation
-* Copyright 2023 Arm Ltd. and affiliates
+* Copyright 2023, 2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -149,14 +149,14 @@ void print_header() noexcept {
         verbose_printf("info,experimental SYCL kernel compiler is enabled\n");
 #endif
         verbose_printf(
-                "primitive,info,template:%soperation,engine,primitive,"
+                "primitive,info,%sengine,primitive,"
                 "implementation,prop_kind,memory_descriptors,attributes,"
                 "auxiliary,problem_desc,exec_time\n",
                 get_verbose_timestamp() ? "timestamp," : "");
 
 #ifdef ONEDNN_BUILD_GRAPH
         verbose_printf(
-                "graph,info,template:%soperation,engine,partition_id,"
+                "graph,info,%sengine,partition_id,"
                 "partition_kind,op_names,data_formats,logical_tensors,fpmath_"
                 "mode,implementation,backend,exec_time\n",
                 get_verbose_timestamp() ? "timestamp," : "");
--- a/doc/performance_considerations/verbose.md
+++ b/doc/performance_considerations/verbose.md
@@ -98,8 +98,8 @@ onednn_verbose,v0,info,cpu,runtime:OpenMP,nthr:128
 onednn_verbose,v0,info,cpu,isa:Intel AVX-512 with Intel DL Boost
 onednn_verbose,v0,info,gpu,runtime:none
 onednn_verbose,v0,info,graph,backend,0:dnnl_backend
-onednn_verbose,v0,primitive,info,template:operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
-onednn_verbose,v0,graph,info,template:operation,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,backend,exec_time
+onednn_verbose,v0,primitive,info,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
+onednn_verbose,v0,graph,info,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,backend,exec_time
 onednn_verbose,v0,primitive,create:check,matmul,dimension src:1 is inconsistent with weights:0,src/common/matmul.cpp:144
 ~~~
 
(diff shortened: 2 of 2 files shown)


vpirogov:
The intent here was to provide the names of column headers. `onednn_verbose,v1,primitive,info,template:` prefix indicates that it will be followed by column names. The names themselves are correct with `operation` corresponding to `exec`, `create`, and likes:
```
operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
exec,cpu,matmul,gemm:jit:f32,undef,src:f32:a:blocked:ab::f0 wei:f32:a:blocked:ab::f0 dst:f32:a:blocked:ab::f0,,,1x1:1x1,0.0170898
```

I get that this does not work CVS though. Maybe we just need to drop `template:` thing?

And please check that [verbose converter](https://github.com/uxlfoundation/oneDNN/tree/main/scripts/verbose_converter) does not choke on the new format.

avmanerikar:
The `template` descriptor is required for parsing the header info in both benchdnn and the [verbose converter](https://github.com/uxlfoundation/oneDNN/tree/main/scripts/verbose_converter). 
One workaround is to strip it before use by CSV workflows:
```
ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn --mode=P --graph --case=op/f32/abs.json \
    | grep 'onednn_verbose,v1,graph' \
    | sed '1s/info,template://'\
    | xsv table
```

atkassen:
The verbose converter generally doesn't _need_ the template to function correctly. It will simply ignore these lines after the change.

There are a number of ways we could change the template, but there are many more cases where the verbose line doesn't line up with the header (try `ONEDNN_VERBOSE=dispatch`). Do we have to fix all of those too?

src/common/verbose.cpp
```diff
@@ -149,14 +149,14 @@ void print_header() noexcept {
         verbose_printf("info,experimental SYCL kernel compiler is enabled\n");
 #endif
         verbose_printf(
-                "primitive,info,template:%soperation,engine,primitive,"
+                "primitive,info,%sengine,primitive,"
                 "implementation,prop_kind,memory_descriptors,attributes,"
```
vpirogov, line 153: Output format actually used is "[timestamp,]operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes"

So if the goal is making columns align in CSV format we'll need something like this:
```suggestion
                "primitive,%soperation,engine,primitive,"
"implementation,prop_kind,memory_descriptors,attributes,"
```

This breaks a convention that informational lines use `operation = info`. Alternatively we could do something like `primitive:template`.

jondea: Ah yes, good point, I missed the operation column, I will fix that.

About the convention of `operation = info`, it's worth nothing that this convention is broken in the initial info lines, which put info in the same column as the API name (`primitive`, `graph` etc). So I think it would still be coherant to have any of:

- `onednn_verbose,v1,primitive,operation,...`
- `onednn_verbose,v1,primitive:template,operation,...`
- `onednn_verbose,v1,primitive:info,operation,...`

jondea:
> The intent here was to provide the names of column headers. `onednn_verbose,v1,primitive,info,template:` prefix indicates that it will be followed by column names. The names themselves are correct with `operation` corresponding to `exec`, `create`, and likes:

That makes sense, although a bit programmatically unwieldy. So the canonical way of grabbing the verbose output as a CSV would be to strip `onednn_verbose,v1,primitive,info,template:` or `onednn_verbose,v1,primitive,` e.g.
```
| grep 'onednn_verbose,v1,primitive' | sed 's/onednn_verbose,v1,primitive,//g' | sed '1s/info,template://'
```

> There are a number of ways we could change the template, but there are many more cases where the verbose line doesn't line up with the header (try ONEDNN_VERBOSE=dispatch). Do we have to fix all of those too?

Fair point, although the dispatch verbose appears to be even more broken. It appears to repeat the primitive name twice, and it doesn't align with the last column being `exec_time`. Instead it has a message and a source location. I think to fix this, we would need another header which applies only to dispatch, with it made clear which one applied to which.

```
> ONEDNN_VERBOSE=all ./tests/benchdnn/benchdnn --mode=r --matmul 1x1:1x1 | grep primitive
onednn_verbose,v1,primitive,create:dispatch,matmul,cpu,matmul,brg:sve_512,undef,src:f32:a:any:any::f0 wei:f32:a:any:any::f0 dst:f32:a:any:any::f0,,,1x1:1x1,unsupported isa,src/cpu/aarch64/matmul/brgemm_matmul.cpp:152
```

Although I don't want perfect to be the enemy of good. I would like to make profile more easily parseable even if there isn't an easy fix for dispatch. I usually just read dispatch verbose with my eyes, but I often want to do data analysis on profile output.

atkassen:
> Although I don't want perfect to be the enemy of good. I would like to make profile more easily parseable even if there isn't an easy fix for dispatch. I usually just read dispatch verbose with my eyes, but I often want to do data analysis on profile output.

Another option is to add a csv generator to the verbose converter:
```
$ DNNL_VERBOSE=1 build/tests/benchdnn/benchdnn --matmul --engine=gpu 1x1:1x1 | \
      scripts/verbose_converter/verbose_converter.py -g csv | \
      xsv table
operation  engine  prim_kind  impl             prop_kind  mds                                                                         exts  aux  shapes   exec_time
exec       cpu     matmul     brg_matmul:avx2  undef      src:f32:a:blocked:ab::f0 wei:f32:a:blocked:ab::f0 dst:f32:a:blocked:ab::f0             1x1:1x1  0.0458984
```

PR #4866, opened 2026-03-20
tests: benchdnn: print csv friendly perf header

jondea:
# Description

Most of the `benchdnn` `perf-template`s already emit CSV-shaped data, but there are no headers, and the template doesn't match the columns. Specifically, `%DESC%` can expand to multiple columns for some drivers, which makes the output hard to parse. This change adds a CSV header row that matches each driver's concrete `%DESC%` expansion while preserving the existing `Output template: ...` banner. This makes the CSV output easy to consume with standard tools (e.g. `xsv`).

E.g. after this change,

```sh
./tests/benchdnn/benchdnn --conv --mode=P --perf-template=csv  mb1_ic1oc1_id1ih5iw5_od1oh3ow3_kd1kh3kw3_sd1sh1sw1_pd0ph0pw0_dd0dh0dw0 \
    | grep '^perf,' \
    | xsv table

perf  engine  impl         name  dir    sdt          stag  wtag  dtag  alg     attr  g   mb  ic  id  ih  iw  oc  od  oh  ow  kd  kh  kw  sd  sh  sw  pd  ph  pw  dd  dh  dw  Gops      max_ctime  min_time  min_Gflops  avg_time  avg_Gflops
perf  cpu     jit:sve_128        FWD_B  f32:f32:f32  any   any   any   direct        1   1   1   1   5   5   1   1   3   3   1   3   3   1   1   1   0   0   0   0   0   0   1.62e-07  0.994385   0.137695  0.00117651  0.32537   0.000497895
```

Not sure how to test this using ctest, I'm open to ideas...

# Checklist

## General

### DIFF
--- a/tests/benchdnn/doc/knobs_perf_report.md
+++ b/tests/benchdnn/doc/knobs_perf_report.md
@@ -136,7 +136,7 @@ CSV-style:
                --batch=inputs/ip/test_ip_all
 ```
 ```
-Output template: perf,%engine%,%name%,%dir%,%cfg%,%attr%,%DESC%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
+Output template: perf,%engine%,%name%,%dir%,%cfg%,%attr%,%desc%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
 perf,cpu,"resnet:ip1",FWD_B,f32,,112,1000,2048,1,1,0.458752,0,0.520264,881.768,0.564043,813.328
 ```
 
--- a/tests/benchdnn/doc/knobs_perf_report.md
+++ b/tests/benchdnn/doc/knobs_perf_report.md
@@ -124,8 +124,10 @@ dumping results with a standard performance template:
                --batch=inputs/ip/test_ip_all
 ```
 ```
-Output template: perf,%engine%,%name%,%prb%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
-perf,cpu,"resnet:ip1",mb112oc1000ic2048n"resnet:ip1",0.458752,0,0.521729,879.293,0.576451,795.822
+Template entries: perf,%engine%,%impl%,%name%,%prb%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
+perf,engine,impl,name,prb,Gops,max_ctime,min_time,min_Gflops,avg_time,avg_Gflops
+perf,cpu,brg_matmul:avx512_core,"resnet:ip1",--mode=P --max-ms-per-prb=6000 --ip ic2048oc1000n"resnet:ip1",0.008192,0.935547,0.0234375,349.525,0.0269902,303.518
+...
 ```
 
 Runs a set of inner products measuring performance and dumping results in
@@ -135,8 +137,10 @@ CSV-style:
                --batch=inputs/ip/test_ip_all
 ```
 ```
-Output template: perf,%engine%,%name%,%dir%,%cfg%,%attr%,%desc%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
-perf,cpu,"resnet:ip1",FWD_B,f32,,112,1000,2048,1,1,0.458752,0,0.520264,881.768,0.564043,813.328
+Template entries: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%attr%,%desc%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
+perf,engine,impl,name,dir,sdt,stag,wtag,dtag,attr,desc,Gops,max_ctime,min_time,min_Gflops,avg_time,avg_Gflops
+perf,cpu,brg_matmul:avx512_core,"resnet:ip1",FWD_B,f32:f32:f32,any,any,any,,ic2048oc1000n"resnet:ip1",0.008192,0.9375,0.0234375,349.525,0.027302,300.051
+...
 ```
 
 Runs a set of inner products measuring performance and dumping custom template -
@@ -147,6 +151,8 @@ not a special symbol here; any other delimiter can be used:
                --batch=inputs/ip/test_ip_all
 ```
 ```
-Output template: %prb%,%-time%,%-Gflops%
-mb112oc1000ic2048n"resnet:ip1",0.521973,878.881
+Template entries: %prb%,%-time%,%-Gflops%
+prb,min_time,min_Gflops
+--mode=P --ip ic2048oc1000n"resnet:ip1",0.0234375,349.525
+...
 ```
--- a/tests/benchdnn/doc/knobs_perf_report.md
+++ b/tests/benchdnn/doc/knobs_perf_report.md
@@ -54,7 +54,6 @@ Other problem specific options supported:
 | %attr%       | All                                                                   | Primitive attributes
 | %axis%       | Concat, Shuffle, Softmax                                              | Primitive axis
 | %desc%       | All                                                                   | String style problem descriptor
-| %DESC%       | All                                                                   | CSV-style problem descriptor values only
 | %dir%        | All, except Concat, RNN, Reorder, Sum                                 | Primitive prop kind
 | %direction%  | RNN                                                                   | RNN direction execution
 | %driver%     | All                                                                   | Name of the current driver (e.g. conv, reorder)
--- a/tests/benchdnn/utils/perf_report.cpp
+++ b/tests/benchdnn/utils/perf_report.cpp
@@ -21,7 +21,7 @@
 #include "utils/perf_report.hpp"
 
 void base_perf_report_t::report(res_t *res, const char *prb_str) const {
-    dump_perf_footer();
+    dump_perf_header();
 
     dnnl::impl::stringstream_t ss;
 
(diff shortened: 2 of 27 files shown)


vpirogov:
I wonder if we actually need information included into problem descriptor (`%DESC%`) split into multiple columns. If the only goal is getting 1-1 correspondence between column headers and data I would argue for printing %DESC% in the same format as used in benchdnn input. Something like this in your example:
```
Output template: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%alg%,%attr%,%DESC%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,jit:avx512_core,,FWD_B,f32:f32:f32,any,any,any,direct,,mb1ic1id1ih5iw5oc1od1oh3ow3kd1kh3kw3pd0ph0pw0,1.62e-07,0.689209,0.00146484,0.110592,0.00220769,0.07338
```

This approach has a benefit of keeping the same number of columns between different primitives in CSV format.

jondea:
That's `%desc%`, %DESC% is supposed to be the CSV version, but you make a good point.

Maybe this is an X-Y problem, I want to be able to use `--perf-template=csv` on all the drivers and be able to easily parse the output as a CSV. Currently, that fails because `csv` uses `%DESC%`, e.g.

```
> ./tests/benchdnn/benchdnn --conv  --mode=f --perf-template=csv --dt=bf16 mb1_ic1oc1_id1ih5iw5_od1oh3ow3_kd1kh3kw3_sd1sh1sw1_pd0ph0pw0_dd0dh0dw0
Output template: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%alg%,%attr%,%DESC%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,ref:any,,FWD_B,bf16:bf16:bf16,any,any,any,direct,--attr-scratchpad=library ,1,1,1,1,5,5,1,1,3,3,1,3,3,1,1,1,0,0,0,0,0,0,1.62e-07,0.741699,1.30591,0.000124052,2.08535,7.76847e-05
```

I actually prefer the `%desc%` style, so I'd be happy for `--perf-template=csv` to use it. Would this be acceptable, or would it class as a breaking change? If so, we would perhaps add something like `csv_desc`?

Although, if we leave %DESC% as an option (used in `csv`) or not, it should probably still output the right header.

Also, are you happy to add the printing of the template as a CSV header as well as the template? Currently I use a hacky sed line to strip out the `Output template:` `%` and `-/0/+`.

vpirogov:
> I actually prefer the `%desc%` style, so I'd be happy for `--perf-template=csv` to use it. Would this be acceptable, or would it class as a breaking change? If so, we would perhaps add something like `csv_desc`?

We don't have infrastructure relying on predefined `csv` template. So I don't see any issues in changing it to use %desc% instead of %DESC%.

> Also, are you happy to add the printing of the template as a CSV header as well as the template? Currently I use a hacky sed line to strip out the Output template: % and -/0/+.

Fine as long as it does not break verbose converter. Couple ideas:
* Drop 'Output template: ' piece for csv template only
* or print `Output template: ` thing on a new line instead

jondea:
> Couple ideas:
> 
> * Drop 'Output template: ' piece for csv template only
> * or print `Output template: ` thing on a new line instead

I think it makes sense to print the output template verbatim _and_ a CSV header in all cases (on different lines). I probably didn't make that clear in my output because I trimmed it, sorry!
- The output template is useful for reproducing the output. You have to `| grep perf,` anyway because `--mode=p` prints lots of lines which aren't csv, so the extra line doesn't mess up parsing CSV
- If we only print the header when `--perf-template=csv` then it limits users to just one output template if they want the header. I often want to customize the output but still have a header.


> This approach has a benefit of keeping the same number of columns between different primitives in CSV format.

It would make them more similar, but I think there would still be some differences. E.g eltwise has no `%wtag%`.

If I output the CSV header in all cases, then I will still need all the machinery to print the `%DESC%` columns, even if I switch `--perf-template=csv` to use `%desc%`.

vpirogov:
> If I output the CSV header in all cases, then I will still need all the machinery to print the `%DESC%` columns, even if I switch `--perf-template=csv` to use `%desc%`.

Not sure I follow. `%DESC%` basically splits `%desc%` into columns, the information is the same. Compare:
```
$ ./benchdnn --conv --mode=P --perf-template=csv ic1oc1ih1iw1kh1kw1
Output template: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%alg%,%attr%,%DESC%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,jit_1x1:avx512_core,,FWD_B,f32:f32:f32,any,any,any,direct,,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,0,4e-09,44.2292,0.000488281,0.008192,0.000761455,0.0052531
$ ./benchdnn --conv --mode=P --perf-template="perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%alg%,%attr%,%desc%,%Gops%,%+ctime%,%-time%,%-Gflops%
,%0time%,%0Gflops%" ic1oc1ih1iw1kh1kw1
Output template: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%alg%,%attr%,%desc%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,jit_1x1:avx512_core,,FWD_B,f32:f32:f32,any,any,any,direct,,ic1ih1oc1oh1kh1ph0,4e-09,0.611084,0.000488281,0.008192,0.000759286,0.00526811
```

I agree that adding proper column names for extra columns that `%DESC%` expands into makes sense. The mode itself is not particularly useful though. You just get 20 columns describing convolution shape instead of just one.

jondea:
> I agree that adding proper column names for extra columns that `%DESC%` expands into makes sense. The mode itself is not particularly useful though. You just get 20 columns describing convolution shape instead of just one.

I think we agree. Are you suggesting I remove it completely?

If not, I'll add a commit where I switch `%DESC%`-> `%desc%` for `--perf-template=csv`.

With respect to the verbose converter, it appears not to like `--mode=p` before and after, emitting the same output: 
```
WARNING: verbose_converter: parse error: too many fields to parse
```
I'll have a look if I can fix this in a separate PR, because it should probably just silently filter them out.

vpirogov:
> I think we agree. Are you suggesting I remove it completely?

Yeah, that's one thing I'd like to get @dzarukin's perspective on once he's back later this week.

dzarukin:
I'm cool with removing %DESC% and replace it with %desc% in `csv` output.

tests/benchdnn/utils/perf_report.hpp
```diff
@@ -83,6 +84,8 @@ struct base_perf_report_t {
         }
     }
 
+    void dump_perf_csv_header() const;
```
dzarukin, line 87: There a virtual call to remove with a second commit:
```
    virtual void dump_desc_csv(std::ostream &) const { SAFE_V(FAIL); }
```

tests/benchdnn/utils/perf_report.cpp
```diff
@@ -156,3 +157,35 @@ void base_perf_report_t::handle_option(std::ostream &s, const char *&option,
             opt_name.c_str());
     SAFE_V(FAIL);
 }
+
+void base_perf_report_t::dump_perf_csv_header() const {
```
dzarukin, line 161: Suggest to fuse `dump_perf_footer` with this new one, keep the former name, a single static guard, and convert
`Output template: ` text into `Template entries: ` since that's what it is.

dzarukin (approved):
Couple minor suggestions, rest looks good to me.

tests/benchdnn/utils/perf_report.cpp
jondea: Makes sense!

Although, why is this one called footer? It seems to become before any other perf output.

dzarukin: Well, 5 years ago I thought it is a footer. Little did I know...

tests/benchdnn/doc/knobs_perf_report.md
```diff
@@ -136,7 +135,7 @@ CSV-style:
                --batch=inputs/ip/test_ip_all
 ```
 ```
-Output template: perf,%engine%,%name%,%dir%,%cfg%,%attr%,%DESC%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
+Output template: perf,%engine%,%name%,%dir%,%cfg%,%attr%,%desc%,%Gops%,%Gfreq%,%-time%,%-Gflops%,%0time%,%0Gflops%
 perf,cpu,"resnet:ip1",FWD_B,f32,,112,1000,2048,1,1,0.458752,0,0.520264,881.768,0.564043,813.328
```
vpirogov: (nit) Actually the output looks like this now:

```suggestion
Output template: perf,%engine%,%impl%,%name%,%dir%,%sdt%,%stag%,%wtag%,%dtag%,%attr%,%desc%,%Gops%,%+ctime%,%-time%,%-Gflops%,%0time%,%0Gflops%
perf,cpu,brg_matmul:avx512_core,"resnet:ip1",FWD_B,f32:f32:f32,any,any,any,,ic2048oc1000n"resnet:ip1",0.008192,1.42505,0.781006,10.489,0.802466,10.2085
```

vpirogov (approved):

tests/benchdnn/utils/perf_report.cpp
jondea: Done

tests/benchdnn/doc/knobs_perf_report.md
jondea: Good point! I've updated all of them

PR #4867, opened 2026-03-20
xe: gemm: disable swapab for weights decompression

Simonsays095:
Addresses [MFDNN-14797](https://jira.devtools.intel.com/browse/MFDNN-14797). We generally don't have type-transposed versions of weights-only quantization layers in the kernel catalog. When we attempt to use the `swap_ab` mechanic on these shapes, we fail to find a kernel and dispatch to the reference kernel instead.

Fixes the issue by disabling `swap_ab` in all weights-only quantization cases, so we look for the familiar A/B types in the catalog. The long-term solution will eventually be to make the catalog aware of when swapping is possible, and return a kernel that's tagged with `swap_ab` so we can search for both patterns and use the better one.

### DIFF
--- a/src/common/gemm_types.hpp
+++ b/src/common/gemm_types.hpp
@@ -77,10 +77,18 @@ struct gemm_desc_t : public op_desc_t {
     // Simplified accessors that comply to GEMM API
     static transpose_t get_trans(const memory_desc_t &md) {
         if (!md.ndims) return transpose::notrans; // arbitrary
-        return md.dims[md.ndims - 1] != 1
-                        && md.format_desc.blocking.strides[md.ndims - 1] != 1
-                ? transpose::trans
-                : transpose::notrans;
+
+        // Leading dimension must be byte-aligned
+        using namespace data_type;
+        bool is_4bit = utils::one_of(md.data_type, f4_e2m1, f4_e3m0, s4, u4);
+        dim_t last_dim = md.dims[md.ndims - 1];
+        auto strides = md.format_desc.blocking.strides;
+        dim_t notranspose_ld
+                = md.dims[md.ndims - 2] > 1 ? strides[md.ndims - 2] : last_dim;
+        if (is_4bit && notranspose_ld % 2 != 0) return transpose::trans;
+
+        return last_dim != 1 && strides[md.ndims - 1] != 1 ? transpose::trans
+                                                           : transpose::notrans;
     }
     transpose_t transa() const { return get_trans(b_desc); }
     transpose_t transb() const { return get_trans(a_desc); }


src/gpu/intel/gemm/jit.hpp
```diff
@@ -86,9 +86,9 @@ struct gen_t : public primitive_t {
                     || d->transc() == dnnl_trans;
 
             // We cannot swap A/B if we don't have kernels to support the
-            // swapped data type/alignment requirements
-            swap_ab_ &= !(utils::one_of(d->a_type(), f8_e5m2, f8_e4m3)
-                    && d->b_type() == bf16);
+            // swapped data type/alignment requirements. Currently mostly affects
+            // weights-only compression cases, since A/B have different data types
+            swap_ab_ &= !wei_decomp_;
```
kealan-barbieri, line 91: Do we lose anything by dropping transpose? Would it be at all worthwhile to handle this by selecting strat in untransposed format and then applying the transpose following select?

Simonsays095: We shouldn't lose anything with this change, since any `wei_decomp_` case that's swapped will be looking for kernels like `u8:f16:f16` that don't exist, and fall back to reference currently.

I agree in principle, though, that swapping should happen later in the decision tree. Swapping after kernel selection is risky though, as it would significantly change the performance and register pressure for the kernel, and may result in failures to compile. We also don't have any logic to do this currently (for example, would we swap the tile dimension, too?). That's why I think the best solution is to let the catalog search for transposed cases if we allow it to, then we're not modifying any kernels.

kealan-barbieri: That makes sense, I was imagining a scenario in which we would permit a transposed case to select strategies with signature oHH rather than HoH and then implicitly handle it as transposed, it would definitely require some potentially fragile logic to handle and I can see how it might result in sub-optimal selection or even fail to compile.

vpirogov:
Why `gemm_types.hpp` lives in `common`, not in `gpu/intel` where it's actually used?

Simonsays095:
> Why `gemm_types.hpp` lives in `common`, not in `gpu/intel` where it's actually used?

I'm not sure, it's lived here for years. There's also `gemm_pd.hpp` (if we're ok with gemm being a GPU-only internal primitive, as it is currently) and `gemm_utils.hpp` that could be moved. @vpirogov I can open a follow-up PR to move these and rename them appropriately if you like.

vpirogov:
> I can open a follow-up PR to move these and rename them appropriately if you like.

Thanks, @Simonsays095. This would be nice.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4874, opened 2026-03-21
x64: matmul: Add 3D scales and zero points support

inteldimitrius:
[MFDNN-14588](https://jira.devtools.intel.com/browse/MFDNN-14588)
This PR enables 3D-scales and zero points for matmul.

Library changes:
- `brgemm_matmul_utils.hpp`: Add `wei_scales_batch_gsize`, `wei_scales_batch_offset`, `wei_zp_batch_gsize`, `wei_zp_batch_offset` attributes to `brgemm_matmul_conf_t`.
- `brgemm_matmul_utils.cpp`: Read batch group size from `get_group(2)` and compute batch offsets for scales and zero points.
- `brgemm_matmul.cpp`: Pass batch index to `get_wei_scales_ptr()` and `get_wei_zp_ptr()`, add batch offset calculation in those functions.
- `primitive_attr_quant.hpp`: Update `get_md()` to divide the batch dimension by corresponding group element.

Benchdnn changes:
- `dnn_types.cpp`: Enable 3D scales and zero points.
- `dnnl_memory.cpp`: Add one more dimension `get_idx()` .
- `matmul/matmul.cpp`: Enable 3D in  `overload_quant_mask` .
- `harness_matmul_decompression`: Add test cases for batch dimension weight decompression.

### DIFF
--- a/doc/primitives/matmul.md
+++ b/doc/primitives/matmul.md
@@ -168,6 +168,12 @@ The following masks are supported by the primitive:
 - 2, which applies a scale / zero point values per column along the
   `n`-dimension for #DNNL_ARG_WEIGHTS.
 
+For batched matmul (3D and higher), scales and zero points can additionally
+be grouped over the batch dimension. When batch grouping is specified (via
+a third group element), the corresponding batch dimension bit is added to
+the mask. This allows different quantization parameters per batch group,
+enabling per-batch-group weight decompression.
+
 When scales and/or zero-points masks are specified, the user must
 provide the corresponding scales and/or zero-points as additional
 input memory objects with argument `DNNL_ARG_ATTR_SCALES |
--- a/src/common/primitive_attr_quant.hpp
+++ b/src/common/primitive_attr_quant.hpp
@@ -130,8 +130,14 @@ struct quant_entry_t : public c_compatible {
         utils::copy_dims_with_mask(quant_dims, in_dims, ndims, mask_,
                 /* fill_with_ones = */ true);
         if (!has_default_groups()) {
-            quant_dims[ndims - 2] /= get_group(0);
-            quant_dims[ndims - 1] /= get_group(1);
+            // Note: nstl::max guards are needed because get_group()
+            // returns 0 for out-of-bound access and MSVC raises C4723
+            // warning (potential divide by 0) treated as error.
+            quant_dims[ndims - 2] /= nstl::max(get_group(0), (dim_t)1);
+            quant_dims[ndims - 1] /= nstl::max(get_group(1), (dim_t)1);
+            if (group_ndims_ > 2 && ndims >= 3) {
+                quant_dims[ndims - 3] /= nstl::max(get_group(2), (dim_t)1);
+            }
         }
 
         CHECK(memory_desc_init_by_tag(
--- a/src/cpu/matmul/ref_matmul.cpp
+++ b/src/cpu/matmul/ref_matmul.cpp
@@ -115,6 +115,7 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
     const auto &wei_zp_dt = attr_zps.get_data_type(DNNL_ARG_WEIGHTS);
     const auto wei_zp_group_k = attr_zps.get_group(DNNL_ARG_WEIGHTS, -2);
     const auto wei_zp_group_n = attr_zps.get_group(DNNL_ARG_WEIGHTS, -1);
+    const auto wei_zp_group_b = attr_zps.get_group(DNNL_ARG_WEIGHTS, 2);
     // Initialize a memory desc for quant entries for easier offset calculation.
     memory_desc_t wei_zp_md {};
     CHECK(attr_zps.get(DNNL_ARG_WEIGHTS).get_md(wei_zp_md, *weights_d.md_));
@@ -146,6 +147,7 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
     const auto wei_scale_group_k = attr_scales.get_group(DNNL_ARG_WEIGHTS, -2);
     const auto wei_scale_ngroups_k = K / wei_scale_group_k;
     const auto wei_scale_group_n = attr_scales.get_group(DNNL_ARG_WEIGHTS, -1);
+    const auto wei_scale_group_b = attr_scales.get_group(DNNL_ARG_WEIGHTS, 2);
     // Initialize a memory desc for quant entries for easier offset calculation.
     memory_desc_t wei_scale_md {};
     CHECK(attr_scales.get(DNNL_ARG_WEIGHTS)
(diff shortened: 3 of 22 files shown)


inteldimitrius:
make test

doc/primitives/matmul.md
```diff
@@ -168,6 +168,12 @@ The following masks are supported by the primitive:
 - 2, which applies a scale / zero point values per column along the
   `n`-dimension for #DNNL_ARG_WEIGHTS.
 
+For batched matmul (3D and higher), scales and zero points can additionally
+be grouped over the batch dimension. When batch grouping is specified (via
+a third group element), the corresponding batch dimension bit is added to
+the mask. This allows different quantization parameters per batch group,
+enabling per-batch-group weight decompression.
```
vpirogov, line 175: This looks like an extension of group semantics affecting global library behavior. As such it requires an RFC.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -357,3 +357,14 @@
...
 --attr-fpmath=strict:true
 12x4x576:12x576x192
+
+## Grouping over batch dimension (3D scales/ZP groups)
+--reset
+--skip-impl=ref
+--dt=bf16:s8:bf16
+--wtag=any,abc
+--attr-scales=wei:per_ocic:bf16:128x1x1,wei:per_ocic:bf16:128x1x3
+--attr-zero-points=,wei:per_ocic:s8:128x1x1,wei:per_ocic:s8:128x1x3
```
mzhukova, line 366: could we simply use mask `7` (per N/per K/per B)? wouldn't it be enough to cover the request? and also I think that it might remove the need for the overload in `tests/benchdnn/matmul/matmul.cpp`

src/common/primitive_attr_quant.hpp
```diff
@@ -132,6 +132,12 @@ struct quant_entry_t : public c_compatible {
         if (!has_default_groups()) {
             quant_dims[ndims - 2] /= get_group(0);
             quant_dims[ndims - 1] /= get_group(1);
+            if (group_ndims_ > 2 && ndims >= 3) {
```
mzhukova, line 138: (semi-random place) I think verbose would be broken (i.e., will not print the ~3rd group info)

src/cpu/matmul/ref_matmul.cpp
```diff
@@ -113,8 +113,9 @@ status_t ref_matmul_t::execute_ref(const exec_ctx_t &ctx) const {
             = !attr_zps.has_default_values(DNNL_ARG_WEIGHTS);
     int wei_zp_mask = attr_zps.get_mask(DNNL_ARG_WEIGHTS);
     const auto &wei_zp_dt = attr_zps.get_data_type(DNNL_ARG_WEIGHTS);
-    const auto wei_zp_group_k = attr_zps.get_group(DNNL_ARG_WEIGHTS, -2);
```
mzhukova, line 116: I think we should continue w negative indices, `-3` should work probably?

src/cpu/matmul/ref_matmul_int8.cpp
```diff
@@ -97,6 +97,7 @@ status_t ref_matmul_int8_t::execute_ref(const exec_ctx_t &ctx) const {
     const auto &wei_zp_dt = attr_zps.get_data_type(DNNL_ARG_WEIGHTS);
     const auto wei_zp_group_k = attr_zps.get_group(DNNL_ARG_WEIGHTS, 0);
     const auto wei_zp_group_n = attr_zps.get_group(DNNL_ARG_WEIGHTS, 1);
+    const auto wei_zp_group_b = attr_zps.get_group(DNNL_ARG_WEIGHTS, 2);
```
mzhukova, line 100: (random place) there should be a validation for `get_group(2)` value somewhere, to ensure ~`batch_dim % batch_group == 0` or something of this sort (as it is done for `K` and `N`)

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -357,3 +357,14 @@
 --attr-zero-points=,wei:common:2,wei:per_oc:s4,wei:per_ocic:s4:192x1
 --attr-fpmath=strict:true
 12x4x576:12x576x192
+
+## Grouping over batch dimension (3D scales/ZP groups)
+--reset
+--skip-impl=ref
```
mzhukova: this seem to be the only added test, and it skips reference (although it was modified), so maybe add another lightweight test case to ensure `ref_matmul.cpp` is correct?

tests/benchdnn/doc/knobs_attr.md
```diff
@@ -394,6 +398,14 @@ Run a matmul problem with weight-only quantization where weights `bf16` scales a
              --attr-zero-points=wei:per_ocic:s8:2x1 6x4:4x5
 ```
 
+Run a batched matmul problem with weight-only quantization where scales and
+zero points are grouped over both `ic` and batch dimensions:
+``` sh
+  ./benchdnn --matmul --dt=bf16:s8:bf16 --attr-fpmath=bf16:true \
+             --attr-scales=wei:per_ocic:bf16:32x1x3 \
```
mzhukova, line 405: (I would change to integer mask here as well, as it seem a bit confusing)

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -937,15 +966,36 @@ void skip_invalid_prb(const prb_t *prb, res_t *res) {
...
                 BENCHDNN_PRINT(2,
                         "[INVALID][%s:%d]: Weight-only quantization "
                         "zero-points "
-                        "groups support only two dimensions\n",
+                        "groups support up to three dimensions\n",
                         __FILE__, __LINE__);
                 res->state = SKIPPED;
                 res->reason = skip_reason::invalid_case;
                 return;
+            } else if (groups.size() == 3 && prb->ndims < 3) {
```
mzhukova, line 978: I think this should be problem check w early exit vs skipped test case

doc/primitives/matmul.md
```diff
@@ -168,6 +168,12 @@ The following masks are supported by the primitive:
 - 2, which applies a scale / zero point values per column along the
   `n`-dimension for #DNNL_ARG_WEIGHTS.
```
mzhukova, line 169: would be good to have newly supported masks explicitly mentioned here (like `7` for per-K-per-N-per-B if that's an intent)
you could in addition move the new sentence from below here

mzhukova (changes requested):
Added a bunch of comments, I think this new feature needs a bit more attention.
Also, it would be better to not have everything as a single commit, as changes are quite diverse.

tests/benchdnn/matmul/matmul.cpp
mzhukova: also library should return an error here, right?

azhai219:
Hi @inteldimitrius
I have read your patch already. Please consider the below things.
1. I remember some i4/u4 with `zero_point` case accuracy may loss. refer to https://github.com/openvinotoolkit/oneDNN/commit/fd1955dabbc7780c8e81085964608ff76e339b49
2. I am not familar with onednn test case. But test case should `i4/u4` & `i8/u8` & with `zero_point` & without `zero_point`.

doc/primitives/matmul.md
inteldimitrius: Hi, okay! I'm getting familiar with RFC procedures and will create the RFC for it.

vpirogov: You may also want to clarify the need of groups in batch dimension with OpenVINO. This looks really odd from the math perspective.

inteldimitrius: Yes, @azhai219  can  confirm the request. They are already use the patch without benchdnn-related part.

PR #4875, opened 2026-03-23
benchdnn: graph: test rmsnorm with different epsilon values

TaoLv:
Fixes MFDNN-14781

1. In Graph API, if epsilon is not specified, it's default to 1e-5 per the op definition. Epsilon can be specified through the JSON file or --op-attrs rewriting.
2. In lnorm driver, epsilon = 1.f/16 (0.0625f) is always used for lnorm primitive testing.
3. This change passes the epsilon value from RMS Norm test case to the lnorm driver in benchdnn for correctness validation.

### DIFF
--- a/tests/benchdnn/inputs/graph/op/harness_f32_all
+++ b/tests/benchdnn/inputs/graph/op/harness_f32_all
@@ -873,3 +873,4 @@
 --reset --in-shapes=1:1 --case=op/f32/greaterequal.json
 --reset --op-attrs=0:mode:gelu_erf,0:mode:gelu_tanh --case=op/f32/gelu.json
 --reset --case=op/f32/rmsnorm.json
+--reset --op-attrs=0:epsilon:0.00001 --case=op/f32/rmsnorm.json
--- a/tests/benchdnn/lnorm/lnorm.hpp
+++ b/tests/benchdnn/lnorm/lnorm.hpp
@@ -58,8 +58,11 @@ struct settings_t : public base_settings_t {
     std::vector<flags_t> flags {NONE};
     check_alg_t check_alg = check_alg_t::ALG_AUTO;
 
+    std::vector<float> eps {1.f / 16};
+
     const char *perf_template_csv() const {
-        static const std::string args = "%dir%,%dt%,%tag%,%stat_tag%,%flags%";
+        static const std::string args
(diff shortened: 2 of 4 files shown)


tests/benchdnn/lnorm/lnorm.hpp
mzhukova, line 71: does `has_single_setup()` need an update too?

tests/benchdnn/inputs/graph/op/harness_f32_all
```diff
@@ -873,3 +873,4 @@
 --reset --in-shapes=1:1 --case=op/f32/greaterequal.json
 --reset --op-attrs=0:mode:gelu_erf,0:mode:gelu_tanh --case=op/f32/gelu.json
 --reset --case=op/f32/rmsnorm.json
+--reset --op-attrs=0:epsilon:0.00001 --case=op/f32/rmsnorm.json
```
mzhukova, line 876: it seems like fallback in `get_lnorm_eps` sets the same `eps` value by default, is it intentional? (essentially wouldn't this and previous lines be the same in behavior..?)

TaoLv: The previous line tests epsilon=0.0625 which is specified inside the JSON file: 

https://github.com/uxlfoundation/oneDNN/blob/61f315e1532d7e45020c5332265dbc288a534b3f/tests/benchdnn/inputs/graph/op/f32/rmsnorm.json#L22-L25

The new test line here tests epsilon=1e-5 for which benchdnn previously reported correctness issue as the reference path used epsilon=0.0625.

We can remove the epsilon attribution section from the JSON file to test the default epsilon logic. But that will need us to create and add another JSON file. I feel this new test line can trigger the same computational behavior as the case without epsilon specified.

tests/benchdnn/lnorm/lnorm.hpp
TaoLv: Good catch. Updated now.

tests/benchdnn/inputs/graph/op/harness_f32_all
mzhukova: I see now, thanks for clarification!

TaoLv:
make test
set test_scope=NIGHTLY
disable benchdnn_all
enable benchdnn_graph

PR #4879, opened 2026-03-23
cpu: x64: matmul: add f8 and int8 support for transposed src

densamoilov:
This PR implements MFDNN-14728. It adds support for f8 and int8 for the case when src is transposed.
A new 8-bit transposition kernel was introduced that works with the raw bytes directly and doesn't interpret the values numerically.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -571,13 +571,12 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_tags(memory_desc_t &A_md,
         } else {
             const bool is_int8_avx512_core
                     = this->is_int8() && is_superset(bgmmc.isa, avx512_core);
-            const bool is_adbc_allowed = this->is_f8()
-                    || (this->is_bf16() || this->is_f32() || this->is_bf32()
-                            || this->is_f16() || this->is_f32_f16()
-                            || this->is_f32_bf16()
-                            || this->is_bf16_with_int_wei()
-                            || this->is_f16_with_int_wei() || this->is_tf32()
-                            || this->is_f32_with_int_wei());
+            const bool is_adbc_allowed = this->is_f8() || is_int8_avx512_core
+                    || this->is_bf16() || this->is_f32() || this->is_bf32()
+                    || this->is_f16() || this->is_f32_f16()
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
georgen117, line 1408: Minor: init_masks() is unconditionally called line 1446 of `jit_brgemm_matmul_copy_a_tronsposed_impl_t`

Which ends up initializing vidx1-vidx5.  Which, as best I can tell, is only needed for `transpose_bf16`.

These are using zmm27-zmm31 which ends up being written over by the `vpunpck*` instructions in the `jit_brgemm_matmul_copy_a_transposed_impl_t<Xbyak::Zmm>::transpose_8bit`

this is minor do you want to add `if ( is_f8 || is_int8 ) return;` to the top of the `init_masks()`?

georgen117:
Was the intention to use the new 'transpose_8bit` for int8 data types?

Lines 554, 585, 1367 and 1418-1419 suggest the intent was to have int8 use the new `transpose_8bit` 

However, I don't think `transpose_8bit` is ever called for int8 kernels.
The `create_brgemm_copy_a_transposed` at line 6047 has an unconditional `one_of(src_dt, s8, u8)` check that always redirects `int8` to the pre-existing `jit_brgemm_matmul_copy_a_transposed_int8_impl_t`, before the `is_superset(avx512_core)` branch (line 6050) that instantiates the `Zmm` transpose kernel. As a result all the is_int8 checks (Lines 554, 585, 1367 and 1418-1419) will never be hit. Effectively, making them dead code.

Note the `int8_impl_t` has zero-point compensation support that `transpose_8bit` lacks. 

Is the goal to eventually migrate to using `transpose_8bit` for int8 types?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
densamoilov: I added a check to call `init_masks()` only for fp8 to avoid adding unnecessary instructions.

As for int8, you’re right. After I removed the limitation on the format tag we started going to the int8 copy kernel you mentioned, which turned out to work fine. I renamed the new 8‑bit kernel to f8 to emphasize data type it's used for.

> Is the goal to eventually migrate to using transpose_8bit for int8 types?

No, that was a mistake. Thank you for pointing this out!

PR #4880, opened 2026-03-23
[GPU} xe: jit: gemm: add Xe3 OOI TNN strategy

skazakov1:
Minor practical result of investigation/tuning for https://jira.devtools.intel.com/browse/MFDNN-14666. It does not solve this JIRA, just king of side effect.
PRs adds a strategy that seems to be ~5-10% better on Xe3 for the cases w/ full tensor postops and no regressions for other ones.


skazakov1:
make test perf-gpu
set primitive=gpu:gemm

Issue #2228, opened 2024-11-26
Test test_bfloat16_plus_float fails on s390x


skriesch:
# Summary
Our build fails for s390x because the numbers are not rounded enough:
tests/gtests/internals/test_bfloat16.cpp:79: Failure
[18667s] Expected: (float {expect_bf16}) > (float {initial_value_bf16}), actual: 0 vs 4.59163e-41

# Version
3.6

# Environment
oneDNN includes hardware-specific optimizations and may behave
differently on depending on the compiler and build environment. Include
the following information to help reproduce the issue:
* CPU make and model (try `lscpu`; if your `lscpu` does not list CPU flags,
  try running `cat /proc/cpuinfo | grep flags | sort -u`): s390x
* OS version (`uname -a`): openSUSE Tumbleweed
* Compiler version (`gcc --version`): gcc-14-2.1
* CMake version (`cmake --version`): cmake-3.31.0-1.1
* CMake output log:
```
[...]

PR #4883, opened 2026-03-24
s390x: Fix bfloat16 conversion on big-endian architectures

MarkVeerasingam:
This PR fixes bfloat16 conversions on big-endian architectures (s390x).

`src/common/bfloat16.cpp` relied on accessing the high 16 bits of a float via `iraw[1]` after bit_cast, which assumes a little-endian memory layout. On big-endian systems, this caused incorrect conversions and test failures in:
- `test_bfloat16_plus_float.TestDenormF32`
- `test_bfloat16_converting_constructor_and_assignment.BitsFromIntegerSameAsFromFloat`

This change removes the memory-layout dependency by:
- Converting the float to a uint32_t representation with bit_cast<uint32_t>.
- Taking the top 16 bits using bit shifts instead of array indexing.
- Apply rounding on the lower 16 bits
- Reconstruct the float values by left-shifting 16 bits from bfloat16 when converting back

The change ensures endian-safe conversions across architectures.

Fixes: #2228

### DIFF
--- a/tests/gtests/internals/test_bfloat16.cpp
+++ b/tests/gtests/internals/test_bfloat16.cpp
@@ -117,4 +117,21 @@ TEST(test_bfloat16_converting_constructor_and_assignment,
     assert_same_bits_from_integer_as_from_float<long long>();
 }
 
+TEST(test_bfloat16_rounding, RoundToNearestEven) {
+    // Test round-to-nearest-even (RNE) behavior during float->bfloat16
+    // conversion. A bfloat16 keeps the upper 16 bits of a float. The lower
+    // 16 bits are the "tail" that gets rounded off. When the tail is exactly
+    // at the halfway point, RNE rounds to make the result LSB even.
+
+    // Halfway with odd result LSB: should round up to even.
+    // float 1.01171875f is 0x3F818000 (bf16 candidate 0x3F81, tail 0x8000).
+    constexpr float halfway_odd_lsb {1.01171875f};
+    ASSERT_EQ(bfloat16_t {halfway_odd_lsb}.raw_bits_, 0x3F82u);
+
+    // Halfway with even result LSB: should stay even (no round up).
+    // float 1.00390625f is 0x3F808000 (bf16 candidate 0x3F80, tail 0x8000).
+    constexpr float halfway_even_lsb {1.00390625f};
+    ASSERT_EQ(bfloat16_t {halfway_even_lsb}.raw_bits_, 0x3F80u);
+}
+
 } // namespace dnnl


vpirogov:
make test
disable test_device_gpu

skriesch:
Interesting that @IBM is ignoring such critical bugs for such a long time.

vpirogov:
These changes cause bfloat16 reorder fails on AArch64:
```
95/223 Test #100: test_reorder ....................................................***Failed    6.82 sec
Note: Google Test filter = *:-*_GPU*
[==========] Running 237 tests from 13 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from ACLCases/reorder_simple_test_t_f32_bf16
[ RUN      ] ACLCases/reorder_simple_test_t_f32_bf16.TestsReorder/0
/home/runner/work/oneDNN/oneDNN/oneDNN_build/tests/gtests/test_reorder_common.hpp:48: Failure
Expected equality of these values:
  s
    Which is: 258
  d
    Which is: 256
mismatch at position 257
[  FAILED  ] ACLCases/reorder_simple_test_t_f32_bf16.TestsReorder/0, where GetParam() = 40-byte object <03-00 00-00 37-03 00-00 E0-17 2E-04 6C-AB 00-00 F0-17 2E-04 6C-AB 00-00 F0-17 2E-04 6C-AB 00-00 00-00 00-00 00-00 00-00> (48 ms)
[ RUN      ] ACLCases/reorder_simple_test_t_f32_bf16.TestsReorder/1
/home/runner/work/oneDNN/oneDNN/oneDNN_build/tests/gtests/test_reorder_common.hpp:48: Failure
Expected equality of these values:
  s
    Which is: 258
  d
    Which is: 256
mismatch at position 257
[  FAILED  ] ACLCases/reorder_simple_test_t_f32_bf16.TestsReorder/1, where GetParam() = 40-byte object <03-00 00-00 38-03 00-00 E0-17 2E-04 6C-AB 00-00 F0-17 2E-04 6C-AB 00-00 F0-17 2E-04 6C-AB 00-00 00-00 00-00 00-00 00-00> (51 ms)
```

@uxlfoundation/onednn-cpu-aarch64, any insights?

vpirogov:
Looks like AArch64 tests caught incorrect rounding. Added explicit test for this behavior to the branch.

MarkVeerasingam:
Thanks for looking into this and creating the test case!
The new test failed on my s390x machine too. I had a look into this.

The original code used `(iraw[1] & 0x1)`  which becomes the LSB of the bf16 result. The change made replaced it with `(low16 >> 15)` which is always setting 1 when `low16 >= 0x8000`, losing the "even" check.
I replaced `(low16 >> 15)` with `(f_raw >> 16) & 0x1u` in the FP_NORMAL case - making it match what the original masking did with `(iraw[1] & 0x1)`

I tested the new `RoundToNearestEven`  test and it  passes locally.

Sqvid:
@MarkVeerasingam thank you for the patch! I've tested this locally on our AArch64 machines and it seems to be correct.

When we squash this we might as well set the commit scope to `cpu: common: ...` since it seems to be a shared bug.

vpirogov:
> When we squash this we might as well set the commit scope to `cpu: common: ...` since it seems to be a shared bug.

@Sqvid, it's not a common bug. It's specific to big endian systems because the implementation was dependent on little endian ordering. You may want to check if these changes make any difference in performance on AArch64 just in case though. This codepath is not normally used on x64.

Sqvid:
@vpirogov Oops you're right I skimmed the failure too lightly

vpirogov (approved):

PR #4884, opened 2026-03-24
benchdnn: add SDPA driver

kwieloch-intel:
### This PR adds a new benchdnn driver for the **Scaled Dot-Product Attention (SDPA)** primitive, enabling correctness validation and performance benchmarking of the SDPA GPU kernel directly from the benchdnn application.

**JIRA:** [MFDNN-14715](https://jira.devtools.intel.com/browse/MFDNN-14715)

---

### Problem description

Currently oneDNN includes a fused GPU SDPA kernel (`ocl:micro:reusable`) but has **no dedicated benchdnn driver** to test it. All other comparable primitives (matmul, softmax, etc.) have benchdnn drivers. This means:
- No **performance benchmarking** through benchdnn's `--mode=F/P` infrastructure.
- No **correctness validation** against an independent reference (only gtests).
- No **CI coverage** for SDPA across data types, mask types, scale modes, GQA/MQA, and propagation directions.

---

### Proposed Solution

Implement a new benchdnn driver `--sdpa` that calls the internal `sdpa_primitive_desc_create()` API and validates GPU output against a CPU reference. The driver covers forward inference, forward training, and backward propagation.

$$\text{DST} = \text{softmax}\left(\frac{QK^T}{\sqrt{d}} + \text{mask}\right) V$$

**Feature coverage:**
- **Propagation:** `FWD_I`, `FWD_D` (training, produces workspace), `BWD_D` (gradients for Q, K, V).
- **Data types:** `f32`, `f16`, `bf16` with per-tensor broadcast (`--dt=f16` or `--dt=f16:f16:f16:f32`).
- **Attention masks (`--mask`):** `none`, `buffer` (explicit `[B,H,S_q,S_kv]` tensor), `buffer_1d` (broadcast `[1,1,1,S_kv]`), `buffer_2d` (broadcast `[1,1,S_q,S_kv]`), `causal_top_left`, `causal_bottom_right`.
- **Scale modes (`--scale`):** `library` (default $1/\sqrt{H}$), `mul`, `div`.
- **GQA/MQA:** KV head count derived from the K tensor shape; supports grouped-query attention.
- **Dropout:** via standard attribute (`--attr-dropout`), with Philox-based mask generation.
- **Memory tags:** independent `--qtag`, `--ktag`, `--vtag`, `--dtag`.

> [!IMPORTANT]  
> This driver tests **internal SDPA primitive functionality** that is not part of the public oneDNN API. It uses `sdpa_primitive_desc_create()` from `src/common/sdpa_test_iface.hpp`.

**Reference implementation** (`ref_sdpa.cpp`) generates gold data by composing oneDNN primitives on the CPU:
```
Forward:
  score   = matmul(Q, K)                           - oneDNN matmul primitive
  score  *= scale                                  - 1/sqrt(head_size) or user-provided
  score  += mask                                   - buffer (with broadcast) or generated causal
  prob    = softmax(score)                         - softmax_accurate_inf_as_zero
  prob_dp = dropout(prob)                          - optional, Philox-based
  DST     = matmul(prob_dp, V)                     - oneDNN matmul primitive

Backward:
  dS2 = matmul(dO, V^T)                            - gradient through BMM2
  dV  = matmul(S2_dp^T, dO)                        - uses post-dropout probs
  dS2 = dropout_bwd(dS2)                           - same mask, same scaling
  dS  = softmax_bwd(S2, dS2)                       - uses pre-dropout probs
  dS *= scale                                      - same scale as forward
  dQ  = matmul(dS, K^T)                            - gradient for queries
  dK  = matmul(Q^T, dS)                            - gradient for keys
  + GQA reduce for dK, dV when kv_heads < q_heads
```
---

### Files changed

**Scope limited to `tests/benchdnn/`** (no changes in `src/`).

Core driver:
| File | Description |
|------|-------------|
| `sdpa/sdpa.hpp` | Problem descriptor (`prb_t`), settings, enums (`mask_type_t`, `scale_type_t`), semantic `DNNL_ARG_*` macros |
| `sdpa/sdpa.cpp` | `init_pd`, `fill_data`, `fill_mask` (0/−inf), `init_ref_memory_args`, `skip_*`, `setup_cmp`, `doit` |
| `sdpa/ref_sdpa.cpp` | Reference via oneDNN matmul + softmax primitives; forward + backward + GQA + dropout |
| `sdpa/sdpa_aux.cpp` | String conversions, `get_md` (semantic ARGs), `set_repro_line` |
| `sdpa/bench_sdpa.cpp` | CLI parser (`--mask`, `--scale`, `--mdt`) |
| `sdpa/cfg.cpp` | Fill ranges for Q (`SRC`), K/V (`WEI`), DST |

Test inputs:
| File | Description |
|------|-------------|
| `inputs/sdpa/shapes_basic` | 2D/3D/4D smoke shapes |
| `inputs/sdpa/shapes_transformer_gpu` | LLM-derived shapes with provenance (LLaMA-2/3, Qwen2, Phi3) |
| `inputs/sdpa/shapes_transformer` | Extended shapes including decode (`seq_q=1`) |
| `inputs/sdpa/shapes_bwd` | Backward shapes respecting GPU kernel constraints |
| `inputs/sdpa/test_sdpa_smoke` | Smoke tests |
| `inputs/sdpa/test_sdpa_ci` | CI: all mask/scale variants + fwd/bwd + transformer shapes |
| `inputs/sdpa/test_sdpa_gpu` | Nightly GPU: comprehensive coverage |

Infrastructure:
| File | Description |
|------|-------------|
| `utils/wrapper.hpp` | `dnnl_api_traits<dnnl_stream_t>` for RAII stream management |
| `benchdnn.cpp` | `--sdpa` dispatcher |
| `utils/parser.cpp` | Help text |
| `README.md` | Driver list entry |
| `doc/driver_sdpa.md` | Full driver documentation |

---

###  Results (`test_sdpa_gpu`)

| Metric | DG2 (xe_hpg) | BMG (xe2) |
|--------|--------------|-----------|
| Total | 124 | 124 |
| Passed | 124 | 124 |
| Skipped | 0 | 0 |
| Unimplemented | 0 | 0 |
| Failed | 0 | 0 |

> [!WARNING]  
> **Note:** All 2D/3D shapes removed from tests. All FP32 tests removed from tests. All backward cases removed from tests. These tests have been removed as they do not currently support all platforms.

<details>
<summary>Example repro commands</summary>

```sh
# Smoke test
benchdnn --sdpa --engine=gpu --batch=inputs/sdpa/test_sdpa_smoke

# CI suite
benchdnn --sdpa --engine=gpu --batch=inputs/sdpa/test_sdpa_ci

# Full GPU suite
benchdnn --sdpa --engine=gpu --batch=inputs/sdpa/test_sdpa_gpu

# Single shape with causal mask
benchdnn --sdpa --engine=gpu --dt=f16 --mask=causal_bottom_right 1x32x2048x128:1x32x128x2048:1x32x2048x128

# Backward
benchdnn --sdpa --engine=gpu --dir=BWD_D --dt=f16 1x4x256x64:1x4x64x256:1x4x256x64

# GQA (LLaMA-3 32:8)
benchdnn --sdpa --engine=gpu --dt=f16 1x32x384x128:1x8x128x384:1x8x384x128

# Performance
benchdnn --mode=F --sdpa --engine=gpu --dt=f16 1x32x2048x128:1x32x128x2048:1x32x2048x128
```

</details>

---

#### General

#### New features

- [ ] ~~Have you published an RFC for the new feature?~~ (N/A - internal API, benchdnn only)
- [ ] ~~Was the RFC approved?~~ (N/A)
- [x] Have you added relevant tests?
- [x] Have you added relevant documentation? (`doc/driver_sdpa.md`)

### DIFF
--- a/tests/benchdnn/doc/driver_sdpa.md
+++ b/tests/benchdnn/doc/driver_sdpa.md
@@ -1,0 +1,114 @@
+# SDPA Driver
+
+**Note:** This driver tests internal SDPA primitive functionality that is not
+part of the public oneDNN API.
+
+## Usage
+``` sh
+    ./benchdnn --sdpa [benchdnn-knobs] [sdpa-knobs] [sdpa-desc] ...
+```
+
+where *sdpa-knobs* are:
+
+ - `--dir={FWD_I [default], FWD_D, BWD_D}` -- propagation direction.
+            `FWD_I` is forward inference, `FWD_D` is forward training (needed
+            for workspace output used by backward), `BWD_D` is backward data
+            which computes gradients for Q, K, and V.
+            Refer to [direction](knobs_dir.md) for details.
+ - `--dt={f32:f32:f32:f32 [default], ...}` -- source (Q, K, V) and destination
+            data types. Interface supports broadcasting, when a single input is
+            provided, e.g., `--dt=f32`, the value is applied for all tensors.
+            Refer to [data types](knobs_dt.md) for details.
+ - `--qtag={abx [default], ...}` -- memory format of the queries tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--ktag={abx [default], ...}` -- memory format of the keys tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--vtag={abx [default], ...}` -- memory format of the values tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--mask={none [default], buffer, buffer_1d, buffer_2d, causal_top_left, causal_bottom_right}`
+            -- specifies the attention mask type.
+            `none` uses no mask, `buffer` provides an explicit mask tensor
+            of shape `[B, H, S_q, S_kv]` (added element-wise to attention
+            scores before softmax), `buffer_1d` provides a broadcast mask of
+            shape `[1, 1, 1, S_kv]` (same mask for all batches, heads, and
+            queries), `buffer_2d` provides a broadcast mask of shape
+            `[1, 1, S_q, S_kv]` (same mask for all batches and heads),
+            `causal_top_left` and `causal_bottom_right` apply a causal mask
+            aligned to the respective corner.
+ - `--mdt={f32 [default], f16, bf16}` -- data type of the attention mask
+            buffer. Only used when `--mask` is one of `buffer`,
+            `buffer_1d`, or `buffer_2d`.
+ - `--scale={library [default], mul, div}` -- specifies how the scale
+            value is passed to the primitive. Attention scores are always
+            scaled; the knob controls the API path used. `library` passes
+            `1/sqrt(head_size)` as a multiplicative scale, `mul` does the same
+            explicitly via the `invert_scale=false` API path, `div` passes
+            `sqrt(head_size)` via the `invert_scale=true` API path. All three
+            produce the same mathematical result `scores / sqrt(head_size)`.
+ - `--match=REGEX` -- skip problems not matching the regular expression in
+            `REGEX`. By default no pattern is applied (run everything).
+            Note: Windows may interpret only string arguments surrounded by
+            double quotation marks.
+ - Any attributes options. Refer to [attributes](knobs_attr.md) for details.
+
+and *sdpa-desc* is a problem descriptor. The canonical form is:
+```
+    Q_DIMS:K_DIMS:V_DIMS
+```
+Here `x` is the delimiter for dimensions within a tensor and `:` is the
+delimiter for tensors, provided in the order queries (Q), keys (K), and
+values (V). The destination dimensions are derived automatically.
+A typical 4D tensor layout is `BxHxSxD` where `B` is batch, `H` is heads,
+`S` is sequence length, and `D` is head size.
+
+## Essence of Testing
+
+The SDPA operation computes
+`DST = softmax((Q * K^T) * scale [+ mask]) * V`.
+The reference executes each step independently using f32 matmul and softmax
+primitives on the CPU. The driver compares the fused primitive output against
+this stepwise reference.
+
+For backward (`--dir=BWD_D`), the driver creates a forward-training primitive
+(to produce workspace) followed by a backward primitive. The backward reference
+recomputes the forward intermediates (softmax probabilities), then derives
+gradients via softmax backward and transposed matmuls:
+`dS = softmax_bwd(S2, dO * V^T)`, `dQ = scale(dS) * K^T`, `dK = Q^T * scale(dS)`,
+`dV = S2^T * dO`, with GQA reduction where applicable.
+
+## Examples
+
+Run the default validation set of SDPA using `inputs/sdpa/shapes_basic` file:
+``` sh
+    ./benchdnn --sdpa --batch=inputs/sdpa/shapes_basic
+```
+
+Run f16 SDPA with a causal mask on a transformer-like shape:
+``` sh
+    ./benchdnn --sdpa --dt=f16 --mask=causal_top_left \
+               1x12x128x64:1x12x64x128:1x12x128x64
+```
+
+Run SDPA with explicit scale (division mode) and mixed data types:
+``` sh
+    ./benchdnn --sdpa --dt=f16:f16:f16:f32 --scale=div \
+               2x8x32x64:2x8x64x32:2x8x32x64
+```
+
+Run SDPA performance benchmark on GPU:
+``` sh
+    ./benchdnn --mode=f --sdpa --engine=gpu --dt=f16 \
+               --mask=causal_bottom_right \
+               1x32x2048x128:1x32x128x2048:1x32x2048x128
+```
+
+Run SDPA backward (training) on GPU:
+``` sh
+    ./benchdnn --sdpa --engine=gpu --dir=BWD_D --dt=f16 \
+               1x12x128x64:1x12x64x128:1x12x128x64
+```
+
+More examples with different driver options can be found at
+inputs/sdpa/test_\*.
--- a/tests/benchdnn/inputs/sdpa/test_sdpa_ci
+++ b/tests/benchdnn/inputs/sdpa/test_sdpa_ci
@@ -1,0 +1,53 @@
+--reset
+# No mask
+--dt=f16,bf16
+--batch=shapes_basic
+
+# Buffer mask (bf16 excluded: GPU kernel limitation)
+--reset
+--mask=buffer
+--dt=f16
+--batch=shapes_basic
+
+# Broadcast buffer masks (1D and 2D, matching gtest oneD/twoD)
+--reset
+--mask=buffer_1d
+--dt=f16
+--batch=shapes_basic
+
+--reset
+--mask=buffer_2d
+--dt=f16
+--batch=shapes_basic
+
+# Causal mask
+--reset
+--mask=causal_top_left
+--dt=f16
+--batch=shapes_basic
+
+# Scale
+--reset
+--scale=mul
+--dt=f16
+--batch=shapes_basic
+
+# Transformer-derived shapes (cf. tests/gtests/internals/test_sdpa.cpp)
+--reset
+--dt=f16,bf16
+--batch=shapes_transformer_gpu
+
+# Forward training (needed for backward; verifies DST with training prop)
+--reset
+--dir=FWD_D
+--dt=f16,bf16
+--batch=shapes_transformer_gpu
+
+# Backward
+# --reset
+# --dir=BWD_D
+# --dt=f16,bf16
+# --batch=shapes_bwd
+# --mask=causal_top_left
+# --dt=f16
+# --batch=shapes_bwd
--- a/tests/benchdnn/sdpa/bench_sdpa.cpp
+++ b/tests/benchdnn/sdpa/bench_sdpa.cpp
@@ -1,0 +1,130 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "dnnl_common.hpp"
+#include "utils/parser.hpp"
+#include "utils/task_executor.hpp"
+
+#include "sdpa/sdpa.hpp"
+
+namespace sdpa {
+
+TASK_EXECUTOR_DECL_TYPES;
+
+void check_correctness(
+        const settings_t &s, driver_task_executor_t &task_executor) {
+    for_(const auto &i_dir : s.dir)
+    for_(const auto &i_dt : s.dt)
+    for_(const auto &i_qtag : s.qtag)
+    for_(const auto &i_ktag : s.ktag)
+    for_(const auto &i_vtag : s.vtag)
+    for_(const auto &i_dtag : s.dtag)
+    for_(const auto &i_mdt : s.mdt)
+    for_(const auto &i_mask_type : s.mask_type)
+    for_(const auto &i_scale_type : s.scale_type)
+    for_(const auto &i_attr : s.attributes)
+    for_(const auto &i_ctx_init : s.ctx_init)
+    for (const auto &i_ctx_exe : s.ctx_exe) {
+        const prb_t prb(s.prb_vdims, i_dir, i_dt, i_qtag, i_ktag, i_vtag,
+                i_dtag, i_mdt, i_mask_type, i_scale_type, i_attr, i_ctx_init,
+                i_ctx_exe, s.impl_filter);
+        if (s.pattern && !match_regex(prb.str(), s.pattern)) return;
+
+        task_executor.submit(prb, s.perf_template, createit, checkit, doit);
+    }
+}
+
+int verify_input(const settings_t &s) {
+    static constexpr int n_inputs = 4; // Q, K, V, DST dt
+
+    for (const auto &i_dt : s.dt) {
+        if (i_dt.size() != 1 && i_dt.size() != n_inputs) {
+            BENCHDNN_PRINT(0,
+                    "ERROR: sdpa driver: `dt` option expects either a single "
+                    "input or four inputs in Q, K, V, DST order. Current "
+                    "size is: \"%ld\"\n",
+                    (long)i_dt.size());
+            SAFE_V(FAIL);
+        }
+    }
+
+    static constexpr int n_vdims_inputs = 3; // Q, K, V dims
+    if (s.prb_vdims.n_inputs() != n_vdims_inputs) {
+        BENCHDNN_PRINT(0,
+                "ERROR: Expected number of dims arguments is `%d`, provided "
+                "`%d`.\n",
+                n_vdims_inputs, s.prb_vdims.n_inputs());
+        SAFE_V(FAIL);
+    }
+    return OK;
+}
+
+static const char *help_mask
+        = "MASK    (Default: `none`)\n    Specifies the attention mask "
+          "type for SDPA.\n    `none` - no mask, `buffer` - explicit mask "
+          "buffer, `buffer_1d` - broadcast [1,1,1,S_kv],\n"
+          "    `buffer_2d` - broadcast [1,1,S_q,S_kv],\n"
+          "    `causal_top_left` - causal mask from top-left,\n"
+          "    `causal_bottom_right` - causal mask from bottom-right.\n";
+
+static const char *help_scale
+        = "SCALE    (Default: `library`)\n    Specifies the attention "
+          "scale type for SDPA.\n    `library` - library computes "
+          "1/sqrt(head_size),\n    `mul` - multiply by scale, `div` - divide "
+          "by scale.\n";
+
+int bench(int argc, char **argv) {
+    driver_name = "sdpa";
+    using namespace parser;
+    static settings_t s;
+    static const settings_t def {};
+    static driver_task_executor_t task_executor;
+    for (; argc > 0; --argc, ++argv) {
+        const bool parsed_options = parse_bench_settings(argv[0])
+                || parse_batch(bench, argv[0])
+                || parse_dir(s.dir, def.dir, argv[0])
+                || parse_multi_dt(s.dt, def.dt, argv[0], "dt")
+                || parse_tag(s.qtag, def.qtag, argv[0], "qtag")
+                || parse_tag(s.ktag, def.ktag, argv[0], "ktag")
+                || parse_tag(s.vtag, def.vtag, argv[0], "vtag")
+                || parse_tag(s.dtag, def.dtag, argv[0], "dtag")
+                || parse_dt(s.mdt, def.mdt, argv[0], "mdt")
+                || parse_vector_option(s.mask_type, def.mask_type,
+                        str2mask_type, argv[0], "mask", help_mask)
+                || parse_vector_option(s.scale_type, def.scale_type,
+                        str2scale_type, argv[0], "scale", help_scale)
+                || parse_driver_shared_settings(s, def, argv[0]);
+        if (!parsed_options) {
+            catch_unknown_options(argv[0]);
+
+            parse_prb_vdims(s.prb_vdims, argv[0], /* min_inputs = */ 3);
+
+            SAFE(verify_input(s), WARN);
+
+            s.finalize();
+            check_correctness(s, task_executor);
+        }
+    }
+
+    task_executor.flush();
+
+    return parse_last_argument();
+}
+
+} // namespace sdpa
--- a/tests/benchdnn/sdpa/ref_sdpa.cpp
+++ b/tests/benchdnn/sdpa/ref_sdpa.cpp
@@ -1,0 +1,412 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cmath>
+#include <cstring>
+#include <limits>
+
+#include "oneapi/dnnl/dnnl.h"
+
+// Internal alg_kind used by the GPU SDPA kernel. Must be removed once
+// softmax_accurate_inf_as_zero is promoted to a public value.
+#include "src/common/c_types_map.hpp"
+
+#include "dnnl_common.hpp"
+#include "dnnl_memory.hpp"
+
+#include "sdpa/sdpa.hpp"
+
+namespace sdpa {
+
+// Reference SDPA: generates gold data by composing existing oneDNN primitives
+// (matmul, softmax) instead of reimplementing the algorithm from scratch.
+//
+// Pipeline: score = matmul(Q, K)  ->  scale  ->  [mask/causal]
+//           ->  softmax(score)  ->  matmul(prob, V)  ->  DST
+//
+// All intermediate computation is done in f32 on the CPU engine.
+
+namespace {
+
+// Execute a matmul primitive on CPU: dst = src x wei.
+void exec_matmul(dnnl_engine_t eng, dnnl_stream_t strm, const dnn_mem_t &src,
+        const dnn_mem_t &wei, dnn_mem_t &dst) {
+    dnnl_primitive_desc_t pd {};
+    DNN_SAFE_V(dnnl_matmul_primitive_desc_create(
+            &pd, eng, src.md_, wei.md_, nullptr, dst.md_, nullptr));
+    auto pd_w = make_benchdnn_dnnl_wrapper(pd);
+
+    dnnl_primitive_t prim {};
+    DNN_SAFE_V(dnnl_primitive_create(&prim, pd));
+    auto prim_w = make_benchdnn_dnnl_wrapper(prim);
+
+    dnnl_exec_arg_t args[] = {
+            {DNNL_ARG_SRC, src.m_},
+            {DNNL_ARG_WEIGHTS, wei.m_},
+            {DNNL_ARG_DST, dst.m_},
+    };
+    DNN_SAFE_V(dnnl_primitive_execute(prim, strm, 3, args));
+    DNN_SAFE_V(dnnl_stream_wait(strm));
+}
+
+// Execute in-place softmax on CPU over the given axis.
+void exec_softmax(
+        dnnl_engine_t eng, dnnl_stream_t strm, dnn_mem_t &mem, int axis) {
+    // Use softmax_accurate_inf_as_zero to match the GPU SDPA kernel.
+    const auto alg = static_cast<dnnl_alg_kind_t>(
+            dnnl::impl::alg_kind::softmax_accurate_inf_as_zero);
+    dnnl_primitive_desc_t pd {};
+    DNN_SAFE_V(dnnl_softmax_forward_primitive_desc_create(&pd, eng,
+            dnnl_forward_inference, alg, mem.md_, mem.md_, axis, nullptr));
+    auto pd_w = make_benchdnn_dnnl_wrapper(pd);
+
+    dnnl_primitive_t prim {};
+    DNN_SAFE_V(dnnl_primitive_create(&prim, pd));
+    auto prim_w = make_benchdnn_dnnl_wrapper(prim);
+
+    dnnl_exec_arg_t args[] = {
+            {DNNL_ARG_SRC, mem.m_},
+            {DNNL_ARG_DST, mem.m_},
+    };
+    DNN_SAFE_V(dnnl_primitive_execute(prim, strm, 2, args));
+    DNN_SAFE_V(dnnl_stream_wait(strm));
+}
+
+// Execute softmax backward on CPU: diff_src = softmax_bwd(fwd_dst, diff_dst).
+void exec_softmax_bwd(dnnl_engine_t eng, dnnl_stream_t strm,
+        const dnn_mem_t &fwd_dst, const dnn_mem_t &diff_dst,
+        dnn_mem_t &diff_src, int axis) {
+    const auto alg = static_cast<dnnl_alg_kind_t>(
+            dnnl::impl::alg_kind::softmax_accurate_inf_as_zero);
+    dnnl_primitive_desc_t fwd_pd {};
+    DNN_SAFE_V(dnnl_softmax_forward_primitive_desc_create(&fwd_pd, eng,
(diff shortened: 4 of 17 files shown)


tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+            Refer to [data types](knobs_dt.md) for details.
+ - `--qtag={abx [default], ...}` -- memory format of the queries tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--ktag={abx [default], ...}` -- memory format of the keys tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--vtag={abx [default], ...}` -- memory format of the values tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--mask_type={none [default], buffer, causal_top_left, causal_bottom_right}`
```
umar456: `buffer` should have a type and shape associated with it. We can pass in 1D buffers and 2D buffers with different data types(They may be related to one of the other tensor data types).

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
+# SDPA Driver
+
+## Usage
+``` sh
+    ./benchdnn --sdpa [benchdnn-knobs] [sdpa-knobs] [sdpa-desc] ...
+```
+
+where *sdpa-knobs* are:
+
+ - `--dt={f32:f32:f32:f32 [default], ...}` -- source (Q, K, V) and destination
```
umar456, line 18: Maybe f16 should be the default because we have wider support for it?

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+ - `--vtag={abx [default], ...}` -- memory format of the values tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--mask_type={none [default], buffer, causal_top_left, causal_bottom_right}`
+            -- specifies the attention mask type.
+            `none` uses no mask, `buffer` provides an explicit mask tensor,
+            `causal_top_left` and `causal_bottom_right` apply a causal mask
+            aligned to the respective corner.
+ - `--scale_type={none [default], mul, div}` -- specifies how the attention
```
umar456: I think `none` is actually not performing the scale at all and `mul` is treating the division as a multiplication (i.e. value=(1/sqrt(head_size)) and div is passing the denominator (i.e. value = sqrt(head_size))

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+            scores are scaled. `none` uses the default `1/sqrt(head_size)`,
+            `mul` multiplies scores by the scale value, `div` divides by it.
+ - `--kv_head_number={0 [default], INT}` -- number of KV heads for grouped
```
umar456: I think this is redundant because we pass the kv_head_number as part of the shape K V and Q tensors.

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
+# SDPA Driver
+
+## Usage
+``` sh
+    ./benchdnn --sdpa [benchdnn-knobs] [sdpa-knobs] [sdpa-desc] ...
+```
+
+where *sdpa-knobs* are:
```
karturov, line 11: SDPA training without and with dropout are now missing features after they've been added to v3.12.
I'm fine with amending it via a separate PR though.

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+            `none` uses no mask, `buffer` provides an explicit mask tensor,
+            `causal_top_left` and `causal_bottom_right` apply a causal mask
+            aligned to the respective corner.
+ - `--scale_type={none [default], mul, div}` -- specifies how the attention
+            scores are scaled. `none` uses the default `1/sqrt(head_size)`,
+            `mul` multiplies scores by the scale value, `div` divides by it.
+ - `--kv_head_number={0 [default], INT}` -- number of KV heads for grouped
+            query attention (GQA) or multi-query attention (MQA). `0` means
+            standard multi-head attention where KV heads equal Q heads.
+ - `--match=REGEX` -- skip problems not matching the regular expression in
```
umar456, line 50: I am not sure what this does. Do you have an example?

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+            `REGEX`. By default no pattern is applied (run everything).
+            Note: Windows may interpret only string arguments surrounded by
+            double quotation marks.
```
umar456, line 53: We should add --dir= for forward, forward inference and backwards. This is enabled in the main branch now.

Additionally we should add support for dropout which was added recently.

tests/benchdnn/sdpa/sdpa.hpp
```diff
@@ -0,0 +1,259 @@
...
+#include "oneapi/dnnl/dnnl.h"
+
+#include "common.hpp"
+#include "dnnl_common.hpp"
+#include "utils/cfg.hpp"
+#include "utils/perf_report.hpp"
+#include "utils/settings.hpp"
+
+// Internal test interface for SDPA primitive.
+#include "tests/gtests/internals/sdpa_internal.hpp"
```
karturov: Not sure if it makes sense to include "gtests/internals" to define iface to SDPA from "src/common/sdpa_test_iface.hpp", unless we really need something from there. SDPA testing support in gtests will likely be a candidate for removal after we have it supported by benchdnn.

tests/benchdnn/inputs/sdpa/test_sdpa_ci
```diff
@@ -0,0 +1,21 @@
+--reset
+# No mask
+--dt=f32,f16,bf16
+--batch=shapes_basic
+# Buffer mask (bf16 excluded: GPU kernel bug)
+--mask_type=buffer
```
umar456: Can you add spaces between groups of tests with appropriate --reset flags?

tests/benchdnn/sdpa/bench_sdpa.cpp
```diff
@@ -0,0 +1,132 @@
...
+        BENCHDNN_PRINT(0,
+                "ERROR: Expected number of dims arguments is `%d`, provided "
+                "`%d`.\n",
+                n_vdims_inputs, s.prb_vdims.n_inputs());
+        SAFE_V(FAIL);
+    }
+    return OK;
+}
+
+static const std::string help_mask_type
```
umar456: These are global variables so we should just reference them as const char* instead of string. String requires and allocation before we hit main where as const char* does not. Its fine to use strings in functions.

tests/benchdnn/sdpa/ref_sdpa.cpp
```diff
@@ -0,0 +1,214 @@
...
+    const auto &eng = get_cpu_engine();
+    dnnl_stream_t strm {};
+    DNN_SAFE_V(dnnl_stream_create(&strm, eng, dnnl_stream_default_flags));
+
+    const dnn_mem_t &q_m = args.find(DNNL_ARG_SRC_0);
+    const dnn_mem_t &k_m = args.find(DNNL_ARG_SRC_1);
+    const dnn_mem_t &v_m = args.find(DNNL_ARG_SRC_2);
+    const dnn_mem_t &dst_m = args.find(DNNL_ARG_DST);
+
+    const int64_t MB = prb->mb; // product of all batch dims (incl. heads)
```
umar456, line 296: Why do we combine heads and mb together? We provide this information as inputs and we can use the num heads dim instead of --kv-head-num to enable GQA and MHA use cases.

tests/benchdnn/sdpa/ref_sdpa.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // Step 3: Add attention mask buffer.
+    if (prb->with_mask()) {
+        float *sp = static_cast<float *>(score);
+        const float *mp = static_cast<float *>(args.find(DNNL_ARG_SHIFT));
+        for (int64_t i = 0, n = MB * SQ * SK; i < n; i++)
+            sp[i] += mp[i];
+    }
+
+    // Step 4: Apply causal mask (set future positions to -inf).
+    if (prb->with_causal_mask()) {
```
umar456: Its probably more consistent to create a causal mask buffer and take the buffer path. It will simplify the execution path and probably faster and you can add it to the matmul primitive as a binary postop. Should be faster this way. You can reference the test_sdpa file on how to do that.

tests/benchdnn/sdpa/sdpa.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+    auto dnnl_attr = make_benchdnn_dnnl_wrapper(
+            create_dnnl_attr(prb->attr, attr_args_t(), prb->ndims));
+
+    TIME_C_PD(DNN_SAFE_STATUS(sdpa_primitive_desc_create(&init_pd_args.pd,
+            init_pd_args.engine, q_d, k_d, v_d, dst_d,
+            prb->with_mask() ? (const_dnnl_memory_desc_t)mask_d : nullptr,
+            scale_d, invert, kv_hn, attn_mask_type_val, softmax_alg,
+            dnnl_forward_inference, dnnl_attr,
+            /* kq_attr = */ nullptr, /* vs_attr = */ nullptr)));
```
umar456: Are you planning on adding quantized input support in this PR?

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
...
+ - Any attributes options. Refer to [attributes](knobs_attr.md) for details.
```
umar456, line 54: It would also be cool to have a flag that calls the sdpa function using the graph API so it can be tested through here as well.

tests/benchdnn/doc/driver_sdpa.md
```diff
@@ -0,0 +1,84 @@
+# SDPA Driver
+
+## Usage
+``` sh
+    ./benchdnn --sdpa [benchdnn-knobs] [sdpa-knobs] [sdpa-desc] ...
```
umar456, line 8: Since this is an internal api driver, perhaps we call it --internal-sdpa? This is a question for the arch team.

tests/benchdnn/sdpa/sdpa.cpp
```diff
@@ -0,0 +1,342 @@
...
+}
+
+std::vector<int> supported_exec_args(dir_t dir) {
+    // SRC_0=Q, SRC_1=K, SRC_2=V, SHIFT=attn_mask. SCALE is added in doit().
+    static const std::vector<int> exec_args = {
+            DNNL_ARG_SRC_0,
+            DNNL_ARG_SRC_1,
+            DNNL_ARG_SRC_2,
+            DNNL_ARG_DST,
+            DNNL_ARG_SHIFT,
```
umar456: Use or define these macros so its more readable:

https://github.com/uxlfoundation/oneDNN/blob/main/src/common/sdpa_types.hpp#L31-L39

tests/benchdnn/sdpa/sdpa.cpp
```diff
@@ -0,0 +1,342 @@
...
+                        WARN);
+                break;
+            case DNNL_ARG_SRC_1:
+            case DNNL_ARG_SRC_2:
+                SAFE(fill_data(exec_arg, WEI, prb, cfg, mem, ref_mem, res),
+                        WARN);
+                break;
+            case DNNL_ARG_DST: break;
+            case DNNL_ARG_SHIFT:
+                SAFE(fill_data(exec_arg, SRC, prb, cfg, mem, ref_mem, res),
```
umar456: Masks are supposed to be 0 and -inf values.

If we want to use other values it should be a different mode and we need to be carful about the range of these values since internally this operation is implemented as an addition to avoid precision issues.

dzarukin (dismissed):

vpirogov:
@kwieloch-intel, how did you verify that this test is adequate to identify functional issues for sizes used in LLMs?

tests/benchdnn/doc/driver_sdpa.md
vpirogov: This doesn't seem necessary. We can just add a note in benchdnn documentation that this test covers internal functionality.

tests/benchdnn/sdpa/ref_sdpa.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+} // anonymous namespace
+
+void compute_ref(
+        const prb_t *prb, dir_t dir, const args_t &args, dnnl_primitive_t) {
+    const auto &eng = get_cpu_engine();
+    dnnl_stream_t strm {};
+    DNN_SAFE_V(dnnl_stream_create(&strm, eng, dnnl_stream_default_flags));
```
echeresh: Can we use `make_benchdnn_dnnl_wrapper()` for the stream handle as well? This can help to avoid a memory leak if exiting from this function due to an error.

I think `dnnl_api_traits` should be extended for that.

tests/benchdnn/sdpa/sdpa_aux.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    if (canonical || !has_default_dts) s << "--dt=" << dt << " ";
+    if (canonical || qtag != def.qtag[0]) s << "--qtag=" << qtag << " ";
+    if (canonical || ktag != def.ktag[0]) s << "--ktag=" << ktag << " ";
+    if (canonical || vtag != def.vtag[0]) s << "--vtag=" << vtag << " ";
+    if (canonical || dtag != def.dtag[0]) s << "--dtag=" << dtag << " ";
+    if (canonical || mask_type != def.mask_type[0])
+        s << "--mask_type=" << mask_type2str(mask_type) << " ";
+    if (canonical || scale_type != def.scale_type[0])
+        s << "--scale_type=" << scale_type2str(scale_type) << " ";
```
echeresh: Suggest to rename: `--mask-type` to `--mask` and `--scale-type` to `--scale`.

This will be better aligned with other existing benchdnn knobs such as `--alg`, `bia_mask`, `--scaling`, etc.

tests/benchdnn/sdpa/sdpa_aux.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (canonical || kv_head_number != def.kv_head_number[0])
+        s << "--kv_head_number=" << kv_head_number << " ";
```
echeresh: Suggestion: `--kv-heads` - shorter and uses a hyphen for consistency with other arguments.

Update: after reading the related comment, maybe it should be just dropped and deduced from K/V/Q sizes.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: The mask shape is fixed to `4D` `[B, H, S_q, S_kv]`, which is the only layout supported by the GPU kernel (there is an explicit `ndims == 4` enforcement in `micro.hp`p). If the kernel adds support for `1D` or `2D` masks in the future, the shape handling can be extended accordingly. Do you see a need to already enable such use case?

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: `f32` is the conventional default across all benchdnn drivers (matmul, conv, bnorm, softmax, etc.). The benchdnn driver uses `dt=f32` by default, this is a design convention. CI exercises `f16`/`bf16` extensively. SDPA with `f32` is currently functional and passes the tests. I think keeping `f32` as default maintains consistency with the rest of the framework.

tests/benchdnn/doc/driver_sdpa.md
umar456: This is for broadcasting a [1,1,1,S_kv] across S_q dimension. You can also pass in a mask which sets B and H to be 1 and the same mask is applied in all dims. You can find the supported shapes for masks here:
https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/internals/test_sdpa.cpp#L625

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Scaling is always applied. I’ll clarify this in the documentation.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: `kv_head_number` is a required parameter of the SDPA primitive desc API (`sdpa_primitive_desc_create`). The knob defaults to $0$, which auto-derives the value from `K.dims[1]`, **so users don't need to specify it manually for standard cases** but it's exposed as a knob for explicit control when testing the API. Should I remove it anyway?

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: `--match` is a standard benchdnn knob shared by all drivers (matmul, conv, bnorm, etc.). It filters test cases by regex on the repro line, e.g., `--match=bf16` runs only `bf16` cases.

#### Example:
```py
benchdnn.exe --engine=gpu --sdpa  --dt=f16 1x4x8x16:1x4x16x8:1x4x8x16 --dt=bf16 1x4x8x16:1x4x16x8:1x4x8x16
# tests:2 passed:2 skipped:0 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
# total: 0.22s; create_pd: 0.00s (0%); create_prim: 0.06s (26%); fill: 0.01s (3%); execute: 0.00s (2%); compute_ref: 0.00s (1%); compare: 0.00s (1%);

benchdnn.exe --engine=gpu --sdpa --match=bf16 --dt=f16 1x4x8x16:1x4x16x8:1x4x8x16 --dt=bf16 1x4x8x16:1x4x16x8:1x4x8x16
# tests:1 passed:1 skipped:0 mistrusted:0 unimplemented:0 invalid_arguments:0 failed:0 listed:0
# total: 0.18s; create_pd: 0.00s (0%); create_prim: 0.03s (14%); fill: 0.01s (3%); execute: 0.00s (2%); compute_ref: 0.00s (1%); compare: 0.00s (0%);
```

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Same issue mentioned by @karturov: https://github.com/uxlfoundation/oneDNN/pull/4884#discussion_r3047643390.
This change is a very large‑scale one, but it will be addressed in this PR, `--dir={FWD_I,FWD_D,BWD_D}` will be enabled implemented with forward training and backward support. Dropout will be supported via `--attr-dropout` attribute.

tests/benchdnn/doc/driver_sdpa.md
umar456: Yeah, its a vestigial parameter. You can do it in this PR or you can remove the parameter now and we can address it internally in a different PR.

tests/benchdnn/doc/driver_sdpa.md
umar456: Okay, I looked though the code and you are right according to the API it is a required parameter.
Internally we check the data type to determine if we generate the code for scaling in the kernel. If the data type is `undef` then the scaling is not performed. We can ignore this functionality but it is useful for debugging purposes.

If you want to proceed with the approach you are currently taking then you should rename it to `library` instead of `none `to differentiate that the scale is still applied by the library determines what is passed in.

tests/benchdnn/doc/driver_sdpa.md
umar456: Ah interesting. I would have thought something like that would be part of the knob_common documentation but it seems it is defined everywhere.

tests/benchdnn/doc/driver_sdpa.md
echeresh: @umar456 What scenarios do you have mind? I would expect the graph API to be tested via `--graph` driver while `--sdpa` is reserved for the primitive-based API.

tests/benchdnn/sdpa/ref_sdpa.cpp
kwieloch-intel: With `--kv_head_number` removed, the head count is now always derived from K's shape ([k_dims[nd-3]]). The 3D flattening ([MB = batch × heads]) is intentional. I can try to refactor to `4D` if you feel strongly about it.

tests/benchdnn/sdpa/sdpa.cpp
kwieloch-intel: Quantized input support is currently out of scope for this PR. The focus here is on addressing the core functionality first. Quantization can be considered in a follow‑up PR once the fundamental issues are resolved and tested.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: I’ll add the appropriate information to `driver_sdpa.md`.

kwieloch-intel:
> @kwieloch-intel, how did you verify that this test is adequate to identify functional issues for sizes used in LLMs?

@vpirogov Test shapes are derived from configurations in `test_sdpa.cpp`, which cover models such as LLaMA‑2/3, Qwen2, Phi and others. I'll modify each shape to cite its source. The test doesn't claim to discover new kernel issues, but confirms the benchdnn driver works correctly on production-relevant sizes.

dzarukin:
> > @kwieloch-intel, how did you verify that this test is adequate to identify functional issues for sizes used in LLMs?
> 
> @vpirogov Test shapes are derived from configurations in `test_sdpa.cpp`, which cover models such as LLaMA‑2/3, Qwen2, Phi and others. I'll modify each shape to cite its source. The test doesn't claim to discover new kernel issues, but confirms the benchdnn driver works correctly on production-relevant sizes.

The question can be re-phrased this way: if you change even a little piece inside the micro_sdpa implementation to make the internal logic incorrect, will the new driver catch it and report it, which is basically its main purpose?

kwieloch-intel:
> > > @kwieloch-intel, how did you verify that this test is adequate to identify functional issues for sizes used in LLMs?
> > 
> > 
> > @vpirogov Test shapes are derived from configurations in `test_sdpa.cpp`, which cover models such as LLaMA‑2/3, Qwen2, Phi and others. I'll modify each shape to cite its source. The test doesn't claim to discover new kernel issues, but confirms the benchdnn driver works correctly on production-relevant sizes.
> 
> The question can be re-phrased this way: if you change even a little piece inside the micro_sdpa implementation to make the internal logic incorrect, will the new driver catch it and report it, which is basically its main purpose?

Yes. The benchdnn reference is architecturally independent from the GPU kernel. It decomposes SDPA into separate CPU matmul + softmax primitives and compares element-by-element against the fused micro_sdpa output. Any logic error in the kernel produces large diffs that fail the threshold. Already observed this in practice when a mask broadcasting mismatch produced rdiff > 800, caught immediately by compare.

tests/benchdnn/sdpa/ref_sdpa.cpp
umar456: I don't mind it if it works for you if you remove the `--kv-head-number`.

tests/benchdnn/doc/driver_sdpa.md
umar456: Yeah, that seems reasonable. I was thinking it would be a quick test for the graph api with a simple flag. You could use it to check if the same set of parameters trigger the SDPA microkernel through the graph API and check if the same set of parameters can be processed using the direct approach. It would be difficult to implement considering the additional nodes you need for quantization.

tests/benchdnn/sdpa/sdpa.hpp
kwieloch-intel: Fixed in 9a73897b95d91cd798a23454b6e6bf022d81f1bc. The SDPA test interface no longer relies on gtests/internals.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Addressed in d96da47b4b4327e4c1d2f39c1739b31c8a62c394 and 365c859366f871b06b45763461567a045fdbd03c.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Docs updated in a701f18d4a126a53bf8f1cc91f8b47efde42208d and parameter renamed in 7a3105d3f32a85f77c571deca0218d621207471a.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Parameter removed as suggested in 8684ab6248df021afdf64a0bc1dd589e3318cb4c.

tests/benchdnn/inputs/sdpa/test_sdpa_ci
kwieloch-intel: Thanks for the suggestion, updated in 9048643fad68125069382fcf5468dbb829ae29c1.

tests/benchdnn/sdpa/bench_sdpa.cpp
kwieloch-intel: Updated in d266117da846483642cb58c9c4df15939cedae52 - replaced global `std::string` with `const char*`.

tests/benchdnn/sdpa/ref_sdpa.cpp
kwieloch-intel: Addressed by removing `--kv_head_number` in 8684ab6248df021afdf64a0bc1dd589e3318cb4c.

tests/benchdnn/sdpa/ref_sdpa.cpp
kwieloch-intel: Resolved in a45bab10359c5413c1075a6ca104c97fb42fdd30.

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Added documentation note clarifying the internal nature of the SDPA driver in 947df3731f7355f75d104e31849cd5751da20a7a.

tests/benchdnn/sdpa/sdpa.cpp
kwieloch-intel: Addressed in d02d47564a64136835b48fc3442bd966ca6488a9.

tests/benchdnn/sdpa/sdpa.cpp
kwieloch-intel: Updated in 39f1f56ab87ee0c98ce28f2a9279af5010680c11 - mask values restricted to 0 and -inf.

tests/benchdnn/sdpa/ref_sdpa.cpp
kwieloch-intel: Thanks for the suggestion, updated in 9279273bd48ed2ef77b9138ae5af3d536d4b603b.

tests/benchdnn/sdpa/sdpa_aux.cpp
kwieloch-intel: Done: c21f64078bedb986463e4acbec9e600e2e0126ef.

tests/benchdnn/sdpa/sdpa_aux.cpp
kwieloch-intel: This parameter has been removed entirely: 8684ab6248df021afdf64a0bc1dd589e3318cb4c (https://github.com/uxlfoundation/oneDNN/pull/4884#discussion_r3047642194).

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Thanks. SDPA training support, both with and without dropout, has now been implemented as part of this PR (https://github.com/uxlfoundation/oneDNN/pull/4884/commits/967749a9595bce36e69f872d7f48832d76a9d079).

tests/benchdnn/doc/driver_sdpa.md
kwieloch-intel: Done 967749a9595bce36e69f872d7f48832d76a9d079.

tests/benchdnn/utils/wrapper.hpp
```diff
@@ -68,6 +68,13 @@ struct dnnl_api_traits<dnnl_memory_desc_t> {
     }
 };
 
+template <>
+struct dnnl_api_traits<dnnl_stream_t> {
+    static void destroy(dnnl_stream_t t) {
+        CHECK_DESTROY(dnnl_stream_destroy(t));
+    }
+};
```
dzarukin: Why is this needed?

tests/benchdnn/sdpa/sdpa.hpp
```diff
@@ -0,0 +1,293 @@
...
+            msk_dims[ndims - 2] = (mask_type == MASK_BUFFER_1D) ? 1 : n_queries;
+            // S_kv: always full.
+            msk_dims[ndims - 1] = n_keys;
+        }
+
+        mb = 1;
+        for (int i = 0; i < ndims - 2; i++)
+            mb *= qdims[i];
+
+        ops = 2.0 * mb * n_queries * n_keys * head_size
```
umar456: The ops calculations probably needs to be updated. There are couple of different ways of counting flops with SDPA depending on how its called.
For causal mask you will need to divide the ops in half.
Forward and backward passes will have their own calculations.
You can refer to the internal SDPA code calculations here: https://github.com/uxlfoundation/oneDNN/blob/main/tests/gtests/internals/test_sdpa.cpp#L2704

We can also address this at a later time.

umar456 (approved):

tests/benchdnn/utils/wrapper.hpp
kwieloch-intel: It was added as part of this request: https://github.com/uxlfoundation/oneDNN/pull/4884#discussion_r3054508317.  I've simplified this by using `stream_t` in: https://github.com/uxlfoundation/oneDNN/pull/4884/commits/22ef1967d927833949a1f657c80ca01504970133

tests/benchdnn/sdpa/sdpa.hpp
kwieloch-intel: Implemented: https://github.com/uxlfoundation/oneDNN/pull/4884/commits/c40c8b33136ede969328633f10541d5bb543860e.

dzarukin (approved):

tests/benchdnn/sdpa/sdpa.cpp
```diff
@@ -0,0 +1,479 @@
...
+            case MASK_CAUSAL_TOP_LEFT: return 2; // dnnl_attn_mask_top_left
+            case MASK_CAUSAL_BOTTOM_RIGHT:
+                return 3; // dnnl_attn_mask_bottom_right
+            default: return 0;
+        }
+    }(prb->mask_type);
+    dnnl_alg_kind_t softmax_alg = dnnl_softmax_accurate;
+
+    // KV head count is always derived from the K tensor's head dimension.
+    dnnl_dim_t kv_hn = prb->k_dims()[prb->ndims - 3];
```
atkassen, line 83: The first few cases in `shapes_basic` are 2d and are causing segfaults here.

echeresh (approved):

tests/benchdnn/sdpa/sdpa.cpp
kwieloch-intel: Great catch, thanks! This was a regression from removing the `--kv_head_number knob` (commit 7c3835d, https://github.com/uxlfoundation/oneDNN/pull/4884#discussion_r3047642194), the replacement `k_dims()[ndims - 3]` is OOB for 2D/3D shapes where ndims < 4.

On Windows/MSVC Release builds the UB silently read adjacent memory, the API returned `dnnl_unimplemented`, and `skip_unimplemented_prb` masked it as SKIPPED, so the test results looked correct. On Linux I guess this manifests as a segfault.

Fixed in 94493b5: moved the `ndims < 4` guard from `skip_unimplemented_prb` to `skip_invalid_prb`, which runs before `init_pd` and prevents the OOB access entirely. 2D/3D shapes now cleanly skip as "Invalid case".

atkassen:
make test
disable test_device_cpu

PR #4892, opened 2026-03-25
[GPU] conv xe3p fixes

dyoussif:
Addresses [MFDNN-14683](https://jira.devtools.intel.com/browse/MFDNN-14683) and [MFDNN-14535](https://jira.devtools.intel.com/browse/MFDNN-14535).
1) Detect src operands in dpas with unaligned offsets and return `invalid_fma_layout`.
2) nregs should not be 0
3) stride should default to 1
4) dnscl stride was not in relation to actual data type size but to assumed size ngen was using. This led to incorrect processing in passes in the copy planner like legalizeSIMD. Make stride logic consistent with actual data types of operands.
5) Make sround routine compliant with xe3p register regioning restrictions.
6) mach/l instructions not supported on xe3p; we can use mul more directly.
7) Do not use FWD modifier on unsupported platform (per observed testing results).

### DIFF
--- a/src/gpu/intel/conv/jit/ir_builder.cpp
+++ b/src/gpu/intel/conv/jit/ir_builder.cpp
@@ -252,7 +252,7 @@ public:
         }
 
         // Assign {Fwd} for dpas when applicable.
-        if (cfg_.hw() >= ngen::HW::XE3P_35_10)
+        if (cfg_.hw() > ngen::HW::XE3P_35_10)
             x2r_mul_stmt_ = inject_dpas_fwd(x2r_mul_stmt_);
         // Assign {Atomic} for dpas(w) when applicable.
         x2r_mul_stmt_ = inject_dpas_atomic(x2r_mul_stmt_);


dyoussif:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

third_party/ngen/ngen_emulation.hpp
```diff
@@ -599,12 +599,22 @@ struct EmulationImplementation {
...
-                g.add(mod, dstHi, dstHi, dstLo, loc);
-                g.mov(mod, dstLo, accLo, loc);
+                if (emulateDWxDW){
+                    g.mul(mod, accLo, s0Lo, s1Lo, loc);
+                    g.mach(mod, dstLo, s0Lo, src1, loc);
+                    g.mul(mod, accHi, s0Hi, s1Lo, loc);
+                    g.macl(mod, dstHi, s0Hi, src1, loc);
+                    g.add(mod, dstHi, dstHi, dstLo, loc);
+                    g.mov(mod, dstLo, accLo, loc);
+		}else{
```
echeresh: Please review formatting in nGEN files.

src/gpu/intel/jit/eltwise_injector.cpp
```diff
@@ -457,18 +457,18 @@ void eltwise_injector_f32_t<ngen_generator_t>::philox_4x32(
...
+            4, offs.ud(0)(1), key.ud(0)(4, 4, 1), offs.uw(0)(2, 1, 0));
+    h->template mov<uint16_t>(8, offs.uw(16)(1), Immediate::uv(0x77665544));
+    h->template mov<uint16_t>(8, key.ud(8)(8, 8, 1), key.ud(0)(8, 8, 1));
     h->template mul<uint32_t>(
-            8, key.ud(8)(1), key.ud(0)(8, 8, 1), offs.uw(8)(8, 8, 1));
+            8, key.ud(8)(1), key.ud(8)(8, 8, 1), offs.uw(16)(2, 1, 0));
+    h->template mov<uint16_t>(8, offs.uw(0)(1), Immediate::uv(0x33221100));
     h->template mul<uint32_t>(
-            8, key.ud(0)(1), key.ud(0)(8, 8, 1), offs.uw(0)(8, 8, 1));
+            8, key.ud(0)(1), key.ud(0)(8, 8, 1), offs.uw(0)(2, 1, 0));
```
echeresh: Is it supposed to match `(8, 8, 1)` region? It seems `(2, 1, 0)` would produce `0, 1, 2, 3, ...` sequence instead of previous `0, 0, 1, 1, 2, 2, ... ` as we have one element per row so only `vs` (vertical stride) is relevant here. Is my interpretation correct?

Do we have benchdnn-side validation for stochastic rounding?

src/gpu/intel/jit/eltwise_injector.cpp
```diff
@@ -509,15 +509,20 @@ void eltwise_injector_f32_t<ngen_generator_t>::philox_4x32(
...
-
-        auto acc = h->acc0.retype(DataType::ud);
-        for (int i = 0, off = 0; i < steps; ++i, off += 2 * esize)
-            h->mul(esize, acc[i](2), ctr.ud(off)(8, 4, 2),
-                    ctr_mul.uw(off)(8, 2, 4));
-        h->mach(8, ctrLo, ctr.ud(0)(8, 4, 2), ctr_mul.ud(0)(8, 4, 2));
-        h->mov(8, ctrHi, ctrLo);
-        for (int i = 0, off = 0; i < steps; ++i, off += 2 * esize)
-            h->mov(esize, ctr.ud(off)(2), acc[i](2));
+        if (hw() < gpu_xe3p_35_10) {
```
kealan-barbieri, line 527: are the emulation routines accessible here? it seems like it might be possible to replace some of this with `emul`.

src/gpu/intel/jit/eltwise_injector.cpp
atkassen: Stochastic rounding is broken and jit SR is hopelessly so. I can't recall exactly, but it might be guaranteed to always result in RTNE, so as long as this thing runs, it won't break SR.

echeresh: @atkassen Hmm, does SR need to be removed from JIT then? or what would you suggest?

atkassen: There is no demand for SR currently. If it ever is needed, I could revive #2903 and remove the JIT implementation there (and probably add the OCL version to the GEMM post-ops kernel).

dyoussif: I changed the regioning per Eugene's suggestion and kept the old code in place as a reference for older platforms.

src/gpu/intel/jit/eltwise_injector.cpp
dyoussif: Given the above discussion, I don't think it makes sense to spend much time on this.

src/gpu/intel/conv/jit/ir_builder.cpp
```diff
@@ -252,7 +252,7 @@ class compute_builder_t {
         }
 
         // Assign {Fwd} for dpas when applicable.
-        if (cfg_.hw() >= ngen::HW::XE3P_35_10)
+        if (cfg_.hw() > ngen::HW::XE3P_35_10)
```
kealan-barbieri, line 255: it seems in conflict with spec we should report it somehow if it doesnt behave as expected.

third_party/ngen/ngen_gen12.hpp
```diff
@@ -1103,6 +1100,9 @@ static inline uint8_t encodeDnsclCtrl(uint8_t mode, RoundingType rnd, RegData &d
     dst.setOffset(dst.getByteOffset() >> 2);
     src0.setOffset(src0.getByteOffset() >> 2);
     src1.setOffset(src1.getByteOffset() >> 2);
+    dst.setRegion(dst.getVS()/ 8, dst.getWidth(), dst.getHS()/ 8);
```
kealan-barbieri, line 1103: it seems we are applying a transform to region in `copy_plan.cpp` and removing it here, can we fix it in the copyplan transforms rather than doing this? Perhaps there can be an instruction specific exception.

src/gpu/intel/conv/jit/ir_builder.cpp
dyoussif: This limitation is documented in the spec. Real life behavior seems consistent with it.

third_party/ngen/ngen_gen12.hpp
dyoussif: In the few preceding lines, this function expects regioning of `dst` to be with respect to a 4bit data type and `src `to be with respect to a 16 bit data type. Then it converts those offsets to be with respect to a` ud `data type. However the logic is missing for strides. Copy planner should be specify regioning with respect to the actual data types of the regions to 1) avoid problems in subsequent passes if the strides are inconsistent with the specified data type 2) match the inputs `encodeDNsclCtrl `expects.

src/gpu/intel/conv/jit/ir_builder.cpp
kealan-barbieri: I see it, thanks.

third_party/ngen/ngen_gen12.hpp
kealan-barbieri: I see, so dnscl already expects to modify offsets to handle the actual arg data type, but within copy planner we will reason about the arguments in terms of their underlying type.

dyoussif:
make test
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_reorder
enable benchdnn_concat

skazakov1:
this PR also fixes https://jira.devtools.intel.com/browse/MFDNN-14807.

kealan-barbieri:
Will this get a backport to rls-v3.12?

PR #4893, opened 2026-03-25
xe: gemm: reintroduce bf16 kernel for Xe2

Simonsays095:
Addresses [MFDNN-14792](https://jira.devtools.intel.com/browse/MFDNN-14792). A PVC kernel is more optimal on the given shape than existing Xe2 kernels, so it is reintroduced for Xe2. A few other kernels have been refit for the performance model, since they were selected instead of the newer, more optimized kernel.


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #4894, opened 2026-03-25
fix for md.size() using sub byte types

mzhukova:
- f4 types were missing from the sub byte multiplier function (in common/ a graph/ analogue)
- grouped mem desc also didn't take this multiplier into account,

so fixed those two and added simple gtest cases to ensure correctness

### DIFF
--- a/tests/gtests/api/test_memory_desc_ops.cpp
+++ b/tests/gtests/api/test_memory_desc_ops.cpp
@@ -97,6 +97,34 @@ TEST(memory_desc_properties_test, TestMemoryDescSize) {
     ASSERT_EQ(md2_blocked.get_size(), 8 * sizeof(float));
 }
 
+TEST(memory_desc_properties_test, TestMemoryDescSizeSubByte) {
+    using dt = memory::data_type;
+
+    const memory::dim M = 10;
+    const memory::dim K = 256;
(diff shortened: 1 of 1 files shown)


mzhukova:
make test
enable run_documentation

src/graph/interface/logical_tensor.hpp
```diff
@@ -161,7 +161,10 @@ struct logical_tensor_wrapper_t {
     /** For sub-byte data types returns number of elements per byte.
      * For the rest data types returns 1. */
     size_t sub_byte_data_type_multiplier() const {
-        if (utils::one_of(data_type(), data_type::s4, data_type::u4)) return 2;
+        if (utils::one_of(data_type(), data_type::s4, data_type::u4,
+                    dnnl::impl::data_type::f4_e2m1,
+                    dnnl::impl::data_type::f4_e3m0))
+            return 2;
```
TaoLv, line 167: LGTM. Currently graph API's logical tensor does not support fp4 data types yet ([api link](https://github.com/uxlfoundation/oneDNN/blob/89b6222fe28e567bc1f9e0a0921e43eb1736a94c/include/oneapi/dnnl/dnnl_graph.hpp#L252)). But eventually we will support them and need to fix this logic.

PR #4899, opened 2026-03-26
xe: GGEMM: Expand support of grouped gemm to older GPUs

umar456:
# Description

This PR expands support of Grouped GEMM to older GPUs. The grouped gemm kernel was failing previously because the microkernel code was modifying some of the input parameters of the host kernel. This PR addresses this issue by modifying the input parameter to a different type than what the microkernel function is accepting. This should eventually needs to be addressed by IGC.

This PR also addresses some other bugs with the tile_op header and how the sg_size is queried. These changes are unrelated to the original issue.

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
+++ b/src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
@@ -459,23 +459,6 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
         s.kb_load = 32;
     }
 
-    if (hw > HW::XeHPG) {
-        if(block2DA) {
-            problem.A.alignment = std::min(problem.A.alignment,
-                                           static_cast<uint8_t>(block2DMinAlignment(hw, problem.A, strategy.A)));
-        } else {
-            problem.A.alignment = std::min<uint8_t>(16, problem.A.alignment);
-        }
-        if(block2DB) {
-            problem.B.alignment = std::min(problem.B.alignment,
-                                           static_cast<uint8_t>(block2DMinAlignment(hw, problem.B, strategy.B)));
-        } else {
-            problem.B.alignment = std::min<uint8_t>(16, problem.B.alignment);
-        }
-    } else {
-        problem.A.setAlignment(problem.Ta.paddedSize());
-        problem.B.setAlignment(problem.Tb.paddedSize());
-    }
     s.C.accessType = AccessType::Block;
 
     s.A.base = localA ? AddressBase::createSLM() : AddressBase::createA64(true);
@@ -517,6 +500,19 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
         case StrategyRequirement::WGN:         s.wg[LoopN] = req.value; break;
         case StrategyRequirement::WGK:         s.wg[LoopK] = req.value; break;
         default: break;
+    }
+
+    if(block2DA) {
+        problem.A.alignment = std::min(problem.A.alignment,
+                                        static_cast<uint8_t>(block2DMinAlignment(hw, problem.A, strategy.A)));
+    } else {
+        problem.A.setAlignment(std::min<uint8_t>(problem.A.alignment, s.unroll[LoopM] * problem.Ta));
+    }
+    if(block2DB) {
+        problem.B.alignment = std::min(problem.B.alignment,
+                                        static_cast<uint8_t>(block2DMinAlignment(hw, problem.B, strategy.B)));
+    } else {
+        problem.B.alignment = std::min<uint8_t>(16, problem.B.alignment);
     }
 
     if (s.wgTile(LoopM) * s.wgTile(LoopN) == 0)


src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -424,7 +424,7 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
     } else if (!block2DA) {
         s.A.accessType = AccessType::Block;
         if (systolic)
-            s.ka_load = (problem.A.layout == MatrixLayout::T) ? (64 / problem.Ta_ext) : 16;
+            s.ka_load = (problem.A.layout == MatrixLayout::T) ? (64 / problem.Ta.paddedSize()) : 16;
```
kealan-barbieri: is the switch from `Ta` to `Ta_ext` intentional here?

umar456: Good catch. I was experimenting with both and I think i forgot to set it back.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
```diff
@@ -180,19 +184,33 @@ status_t grouped_micro_gemm_t::pd_t::init_microkernels(impl::engine_t *engine) {
...
+        int m_unroll = 0;
+        int max_n_unroll = 0;
+
+        switch (dev_info->gpu_arch()) {
+            case compute::gpu_arch_t::xe_lp:
+            case compute::gpu_arch_t::xe_hp:
+            case compute::gpu_arch_t::xe_hpg:
+                m_unroll = problem.Ta.isInt4() ? sg_size_ / problem.Ta
+                                               : sg_size_;
+                max_n_unroll = (problem.Ta_ext.bits() <= 8)
```
kealan-barbieri: its not clear why this should be separate from the default case. It appears the only difference is initialization when `Ta_ext` is fp4, is this case unsupported for eg. Xe2 or just only with `max_n_unroll` 32?

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -424,7 +424,7 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
     } else if (!block2DA) {
         s.A.accessType = AccessType::Block;
         if (systolic)
-            s.ka_load = (problem.A.layout == MatrixLayout::T) ? (64 / problem.Ta_ext) : 16;
+            s.ka_load = (problem.A.layout == MatrixLayout::T) ? (64 / problem.Ta_ext.paddedSize()) : 16;
```
Simonsays095: I think this change means that we load 64 int4/fp4 elements at a time (paddedSize=1), which is only half a cache line. What is the intent of this change?

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
```diff
@@ -527,7 +534,9 @@ static inline bool getStrategyByHeuristics(HW hw, GEMMStrategy &strategy, bool l
     if (systolic && hw >= HW::XeHPC)
         s.extendedAtomicFMA = s.atomicFMA = true;
     s.registerScheme = GEMMStrategy::VAvoid;
-    if (s.wgTile(LoopM) * s.wgTile(LoopN) > 512)
+
+    int grf_limit = hw >= HW::XeHPC ? 512 : 256 * problem.Ta_ext;
+    if (s.wgTile(LoopM) * s.wgTile(LoopN) > grf_limit)
```
Simonsays095: Can you add a todo here to improve the accuracy of this logic? I'm certain there are lots of times (especially for smaller data types) where we're setting `GRF=256` when we only need 128. It should approximate A/B/C GRF requirements and add some head room.

src/gpu/intel/matmul/grouped_micro_gemm.cl
```diff
@@ -150,14 +150,15 @@ void load_wei_attr_scales(wei_attr_scales_tile_type *tile,
 __attribute__((intel_reqd_sub_group_size(SUBGROUP_SIZE)))
 __attribute__((reqd_work_group_size(ugemm_grouped_sg_per_wg_m * SUBGROUP_SIZE,
         ugemm_grouped_sg_per_wg_n, ugemm_grouped_sg_per_wg_k))) kernel void
-grouped_micro_gemm(const global SRC_DATA_T *src, int ldsrc,
+grouped_micro_gemm(const global SRC_DATA_T *src, long ldsrc,
```
Simonsays095, line 153: So I understand: the idea here is that you pass in `long` as arguments to the kernel but the microkernel interface requires `int`, forcing it to copy the data and downconvert before passing into the kernel? Trying to figure out how this change fixes the "overwriting host kernel arguments" issue.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
Simonsays095, line 215: Can you explain why we have these requirements? I think I remember something about adjusting heuristic strategies using these, but wouldn't this also disqualify many catalog entries?

src/gpu/intel/matmul/grouped_micro_gemm.cl
umar456: Yes exactly. I am not sure the mechanism with which the original data was modified as I haven't analyzed the assembly. I verified this was the main culprit for the failures in DG2.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: These lines specifically reduce the size of the tile for very small problems. The idea is to increase the number of resident WG/CU. 
This is in contrast the unroll calculation above that seem to be more sensitive to types and how they interact with the size of the sub-groups. there are several cases where the alignment and unroll values are intertwined. I often ran into this error: https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/gemm/jit/generator/strategy.cpp#L103

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: DG2 seems to be more sensitive to int8 and fp8 data types. I believe most of the errors were due to "insufficient registers in bundle".

Both DG2 and Xe2 have alignment issues with block loads if the sg_tile_m 2x sg_size for int4 data types only(Fails here: https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/gemm/jit/generator/strategy.cpp#L103).

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
umar456: Sure, I can. I am not sure how we can better calculate that information. Doesn't wgTile(LoopM/LoopN) capture that sufficiently? We also have to consider the requirements of the host kernel such as in SDPA or gated MLP.

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
umar456: Several of the kernels were failing with ka_load of 128. Perhaps the quantization requirements requires additional register pressure. I can limit this to DG2 because XeHPC+ hardware seemed to work with the previous version.

src/gpu/intel/matmul/grouped_micro_gemm.cpp
umar456: To answer your second question. This code path only executes if there are no catalog entries that match this problem or cannot be generated. Here we are defining a sane flexible strategy that can solve all problem but may not be super optimized like the ones in the catalog. Ideally we cover all cases with the kernel catalog and we never have to generate heuristic based kernels.

umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg
enable arch_gpu_xe-hpg
enable arch_gpu_xe-hpc
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-lpg
enable arch_gpu_xe-hpg
enable arch_gpu_xe-hpc
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

umar456:
make test perf-gpu
set primitive=sdpa

src/gpu/intel/gemm/jit/generator/microkernel_selector.cpp
Simonsays095: Almost all register space in the kernel is usually from A/B/C matrices in registers, with the following sizes (assuming no repacking):
- A: `problem.Ta_ext * strategy.unroll[LoopM] * strategy.ka_load`
- B: `problem.Tb_ext * strategy.unroll[LoopN] * strategy.kb_load`
- C: `problem.Tc * strategy.unroll[LoopM] * strategy.unroll[LoopN]`

If this is, for example, over 90 registers, then we can request 256 GRF mode to give some wiggle room for the host kernel.

Edit: just realized this was merged a while ago. Sorry for reviving it!

umar456: This is great. Thank you. I will try it out to see if this works.

PR #4901, opened 2026-03-26
[rls-v3.11] revert: graph: backend: dnnl: fix use-of-uninitialized-value

TaoLv:
This reverts commit 2cbf9a0d1ca89fd358112d1832dd1bdc0a0e5c8c.

`std::call_once` should not be called in header file.

I tried but moving the `get_global_cache()` function into a new cpp file is a bit complicated, involving templated class explicit instantiation in cpp file. Hence, I decided to just revert the previous change.

Fixes [MFDNN-14824](https://jira.devtools.intel.com/browse/MFDNN-14824)

### DIFF
--- a/src/graph/backend/dnnl/thread_local_cache.hpp
+++ b/src/graph/backend/dnnl/thread_local_cache.hpp
@@ -185,8 +185,8 @@ private:
                 static auto global_cache = std::shared_ptr<global_cache_type_t>(
                         new global_cache_type_t {},
                         [](global_cache_type_t *ptr) {
-                            return ptr->release();
-                        });
+                    return ptr->release();
+                });
                 return global_cache.get();
             } catch (...) { return nullptr; }
         }


src/graph/backend/dnnl/thread_local_cache.hpp
```diff
@@ -181,13 +181,14 @@ class thread_local_cache_t {
         static global_cache_type_t *get_global_cache() {
             // A global table to store cached values in ALL threads. This global
             // table takes the ownership of cached values
-            static std::once_flag flag;
-            static global_cache_type_t *global_cache = nullptr;
-
-            std::call_once(
-                    flag, []() { global_cache = new global_cache_type_t {}; });
-
-            return global_cache;
```
ElaineBao, line 190: What's the problem here?

TaoLv: There is a symbol visibility issue with `std::call_once` in a header file and including the header everywhere. It causes symbol conflict in PyTorch where two copies of oneDNN (cpu & xpu) are used together.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #4902, opened 2026-03-26
cpu: aarch64: Optimize RNN by introduce BRGEMM RNN utilities and common cell helpers (non-JIT first)

abhishek-iitmadras:
# Description

This PR introduces the core BRGEMM-based RNN infrastructure for AArch64 as a first “Phase A” step, focusing on shared utilities and common cell (non-JIT). JIT related broader tuning will be follow in subsequent PRs.

### What’s included
- BRGEMM RNN foundational helpers:
    -     rnn_brgemm_utils.hpp/.cpp
- BRGEMM forward cell:
    -     brgemm_cell_common_fwd.hpp/.cpp
- BRGEMM backward cell:
    -     brgemm_cell_common_bwd.hpp/.cpp
- BRGEMM reorders support:
    -      brgemm_cell_common_reorders.hpp/.cpp

> Plan to bring optimized RNN for aarch64

This is Phase A of a 7-phase effort to bring full optimize RNN support to AArch64. Structured whole RNN development as each phase is a self-contained, reviewable unit with a clear scope which make it easier on my end to develop/tested and also for community end to review easily. 

**Phase A (this PR)**: BRGEMM core — the non-JIT compute path, end-to-end fwd + bwd
**Phase B**: JIT helper kernels — bf16 emulation and fp32 add-to-bf16 convert for bf16 training path
**Phase C**: JIT helper kernels — transpose, gates reduction, peephole, BF16 conversion
**Phase D**: Post-GEMM infrastructure — shared abstractions and vanilla RNN cell post-GEMM
**Phase E**: LSTM optimization
**Phase F**: GRU optimization
**Phase G**: LBR-GRU optimization

# Checklist

## General

## Performance improvements

| **RNN Benchmarking Name** | **Before (total perf (ms))** | **This Patch (total perf (ms))** |
|----------|----------|----------|
| **_shapes_deepspeech_2_**  | 89.7729  | 6.9917  |
| **_shapes_inference_**  | 42.4207  | 15.1135  |
| **_shapes_training_**  | 289.159 | 113.703 |
| **_perf_rnn_inference_sb_**  | 159.371 | 97.6782 |
| **_perf_rnn_inference_lb_**  | 1068.06 | 875.916 |
| **_perf_rnn_training_**  | 4803.65 | 3554.14 |

AWS instance used for benchamrking is [c7g.metal](https://instances.vantage.sh/aws/ec2/c7g.metal?currency=USD)

### DIFF
--- a/src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
+++ b/src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
@@ -1,0 +1,517 @@
+/*******************************************************************************
+* Copyright 2021 Intel Corporation
+* Copyright 2026 FUJITSU LIMITED
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "cpu/aarch64/rnn/brgemm_cell_common_bwd.hpp"
+
+#include "common/dnnl_thread.hpp"
+#include "common/utils.hpp"
+
+#include "cpu/aarch64/rnn/brgemm_cell_common_reorders.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+using namespace dnnl::impl::utils;
+
+template <typename weights_t, typename scratch_t, typename gemm_acc_t>
+brgemm_diff_src_layer_iter_t<weights_t, scratch_t,
+        gemm_acc_t>::brgemm_diff_src_layer_iter_t(const ref_rnn_brgemm_t
+                                                          &rnn_brgemm,
+        const rnn_utils::rnn_conf_t &rnn,
+        rnn_utils::cell_position_t cell_position, scratch_t *scratch_gates,
+        weights_t *w_iter, weights_t *w_layer, gemm_acc_t *diff_src_iter,
+        gemm_acc_t *diff_src_layer,
+        aarch64::brgemm_batch_element_t *addr_batch_global)
+    : rnn_brgemm_(rnn_brgemm)
+    , rnn_(rnn)
+    , A_(scratch_gates)
+    , B_wei_iter_(w_iter)
+    , B_wei_layer_(w_layer)
+    , C_diff_iter_(diff_src_iter)
+    , C_diff_layer_(diff_src_layer)
+    , k_blocks_n_gates_(rnn.diff_src_brgemm.K_blocks)
+    , k_blocks_(rnn.diff_src_brgemm.K_blocks / rnn.n_gates)
+    , k_tail_(rnn.diff_src_brgemm.k_tail)
+    , k_block_(rnn.diff_src_brgemm.k_block)
+    , A_k_tail_offset_(k_blocks_ * k_block_)
+    , B_k_tail_offset_(A_k_tail_offset_ * rnn.diff_src_brgemm.n_block)
+    , B_nb_offset_(rnn.diff_src_brgemm.Kpadded * rnn.diff_src_brgemm.n_block)
+    , B_kb_offset_(k_block_ * rnn.diff_src_brgemm.n_block)
+    , B_gb_iter_offset_(rnn.diff_src_brgemm.Kpadded
+              * rnn.diff_src_brgemm.n_block * rnn.diff_src_brgemm.N_iter_blocks)
+    , B_gb_layer_offset_(rnn.diff_src_brgemm.Kpadded
+              * rnn.diff_src_brgemm.n_block
+              * rnn.diff_src_brgemm.N_layer_blocks)
+    , LDA_(rnn.diff_src_brgemm.LDA)
+    , LDC_(rnn.diff_src_brgemm.LDC)
+    , max_nthr_(nstl::min(dnnl_get_current_num_threads(), rnn.nthr))
+    , n_blocking_(rnn.diff_src_brgemm.N_blocks)
+    , m_blocking_(rnn.diff_src_brgemm.M_blocks)
+    , work_amount_(n_blocking_ * m_blocking_)
+    , max_n_layer_blocks_(rnn.diff_src_brgemm.N_layer_blocks)
+    , max_n_iter_blocks_(rnn.diff_src_brgemm.N_iter_blocks)
+    , gemm_layer_needed_(rnn.need_gemm_layer(cell_position))
+    , kernel_iter_full_blocks_b0_(
+              rnn_brgemm_.diff_src_.kernel_iter_layer_beta0_.get())
+    , kernel_iter_n_tail_b0_(
+              rnn_brgemm_.diff_src_.kernel_iter_N_tail_beta0_.get())
+    , kernel_iter_k_tail_(
+              rnn_brgemm_.diff_src_.kernel_iter_layer_K_tail_beta1_.get())
+    , kernel_iter_nk_tail_(
+              rnn_brgemm_.diff_src_.kernel_iter_NK_tail_beta1_.get())
+    , kernel_layer_full_blocks_b0_(
+              rnn_brgemm_.diff_src_.kernel_iter_layer_beta0_.get())
+    , kernel_layer_n_tail_b0_(
+              rnn_brgemm_.diff_src_.kernel_layer_N_tail_beta0_.get())
+    , kernel_layer_k_tail_(
+              rnn_brgemm_.diff_src_.kernel_iter_layer_K_tail_beta1_.get())
+    , kernel_layer_nk_tail_(
+              rnn_brgemm_.diff_src_.kernel_layer_NK_tail_beta1_.get())
+    , addr_batch_global_(addr_batch_global) {}
+
+template <typename weights_t, typename scratch_t, typename gemm_acc_t>
+void brgemm_diff_src_layer_iter_t<weights_t, scratch_t, gemm_acc_t>::execute()
+        const {
+    parallel(max_nthr_, [this](const int ithr, const int nthr) {
+        this->kernel(ithr, nthr);
+    });
+}
+
+template <typename weights_t, typename scratch_t, typename gemm_acc_t>
+void brgemm_diff_src_layer_iter_t<weights_t, scratch_t, gemm_acc_t>::kernel(
+        const int ithr, const int nthr) const {
+
+    int start = 0, end = 0;
+    balance211(work_amount_, nthr, ithr, start, end);
+
+    int n_block_id = 0, m_block_id = 0;
+    nd_iterator_init(start, n_block_id, n_blocking_, m_block_id, m_blocking_);
+
+    aarch64::brgemm_batch_element_t *const addr_batch
+            = addr_batch_global_ + ithr * (k_blocks_n_gates_ + 1);
+
+    const auto n_gates = rnn_.n_gates;
+
+    while (start < end) {
+        const int m = m_block_id * rnn_.diff_src_brgemm.m_block;
+        const int n = n_block_id * rnn_.diff_src_brgemm.n_block;
+
+        const scratch_t *const A_m = A_ + m * LDA_;
+        const auto B_n_offset = n_block_id * B_nb_offset_;
+        const weights_t *const B_wei_iter_n = B_wei_iter_ + B_n_offset;
+        const weights_t *const B_wei_layer_n = B_wei_layer_ + B_n_offset;
+
+        const auto C_offset = m * LDC_ + n;
+        gemm_acc_t *const C_diff_iter_n = C_diff_iter_ + C_offset;
+        gemm_acc_t *const C_diff_layer_n = C_diff_layer_ + C_offset;
+
+        const brgemm_kernel_t *kernel_iter = kernel_iter_full_blocks_b0_;
+        const brgemm_kernel_t *kernel_iter_k_tail = kernel_iter_k_tail_;
+        const brgemm_kernel_t *kernel_layer = kernel_layer_full_blocks_b0_;
+        const brgemm_kernel_t *kernel_layer_k_tail = kernel_layer_k_tail_;
+
+        const bool should_calc_diff_src_layer
+                = gemm_layer_needed_ && n_block_id < max_n_layer_blocks_;
+        const bool should_calc_diff_src_iter = n_block_id < max_n_iter_blocks_;
+
+        if (should_calc_diff_src_iter) {
+            const bool do_n_iter_tail = (n + rnn_.diff_src_brgemm.n_block)
+                    > rnn_.diff_src_brgemm.N_iter;
+            if (do_n_iter_tail) {
+                kernel_iter = kernel_iter_n_tail_b0_;
+                kernel_iter_k_tail = kernel_iter_nk_tail_;
+            }
+
+            for (int gate_id = 0; gate_id < n_gates; gate_id++) {
+                const auto g_block_id = gate_id * k_blocks_;
+                const auto A_gb_offset = gate_id * rnn_.diff_src_brgemm.K;
+                const auto B_gb_offset = gate_id * B_gb_iter_offset_;
+                const auto A_gm = A_m + A_gb_offset;
+                const auto B_wei_iter_gn = B_wei_iter_n + B_gb_offset;
+
+                for (int k_block_id = 0; k_block_id < k_blocks_; k_block_id++) {
+                    addr_batch[g_block_id + k_block_id].ptr.A
+                            = A_gm + k_block_id * k_block_;
+                    addr_batch[g_block_id + k_block_id].ptr.B
+                            = B_wei_iter_gn + k_block_id * B_kb_offset_;
+                }
+            }
+            brgemm_kernel_execute(kernel_iter, k_blocks_n_gates_, addr_batch,
+                    reinterpret_cast<void *>(C_diff_iter_n), nullptr);
+        }
+
+        if (should_calc_diff_src_layer) {
+            const bool do_n_layer_tail = (n + rnn_.diff_src_brgemm.n_block)
+                    > rnn_.diff_src_brgemm.N_layer;
+            if (do_n_layer_tail) {
+                kernel_layer = kernel_layer_n_tail_b0_;
+                kernel_layer_k_tail = kernel_layer_nk_tail_;
+            }
+
+            for (int gate_id = 0; gate_id < n_gates; gate_id++) {
+                const auto g_block_id = gate_id * k_blocks_;
+                const auto A_gb_offset = gate_id * rnn_.diff_src_brgemm.K;
+                const auto B_gb_offset = gate_id * B_gb_layer_offset_;
+                const auto A_gm = A_m + A_gb_offset;
+                const auto B_wei_layer_gn = B_wei_layer_n + B_gb_offset;
+
+                for (int k_block_id = 0; k_block_id < k_blocks_; k_block_id++) {
+                    addr_batch[g_block_id + k_block_id].ptr.A
+                            = A_gm + k_block_id * k_block_;
+                    addr_batch[g_block_id + k_block_id].ptr.B
+                            = B_wei_layer_gn + k_block_id * B_kb_offset_;
+                }
+            }
+            brgemm_kernel_execute(kernel_layer, k_blocks_n_gates_, addr_batch,
+                    reinterpret_cast<void *>(C_diff_layer_n), nullptr);
+        }
+
+        if (should_calc_diff_src_iter && k_tail_) {
+            for (int gate_id = 0; gate_id < n_gates; gate_id++) {
+                const auto A_gb_offset = gate_id * rnn_.diff_src_brgemm.K;
+                const auto B_gb_offset = gate_id * B_gb_iter_offset_;
+                addr_batch[gate_id].ptr.A
+                        = A_m + A_gb_offset + A_k_tail_offset_;
+                addr_batch[gate_id].ptr.B
+                        = B_wei_iter_n + B_gb_offset + B_k_tail_offset_;
+            }
+            brgemm_kernel_execute(kernel_iter_k_tail, n_gates, addr_batch,
+                    reinterpret_cast<void *>(C_diff_iter_n), nullptr);
+        }
+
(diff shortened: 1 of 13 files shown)


src/cpu/rnn/postgemm_dispatcher.hpp
```diff
@@ -105,9 +106,12 @@ struct rnn_postgemm_dispatcher_t {
     virtual ~rnn_postgemm_dispatcher_t() = default;
 
     status_t init(const rnn_utils::rnn_conf_t &rnn) {
-        DNNL_X64_ONLY(CHECK(initialize_jit(rnn)));
+        #if DNNL_X64
```
jondea: This change appears to be a no-op. Do we need it?

Although on the other hand, this looks suspicious. I would expect that is `x64` has to check the status of init, `aarch64` would too

src/cpu/rnn/ref_rnn.hpp
```diff
@@ -114,6 +117,8 @@ struct ref_rnn_common_t : public primitive_t {
             = ref_rnn_common_t<aprop, src_type, weights_type, acc_type>;
 #if DNNL_X64
     using ref_rnn_brgemm_t = x64::rnn_brgemm_utils::rnn_brgemm_t<aprop>;
+#elif DNNL_AARCH64
```
jondea: Given that `aarch64` rnn seems to have the same API as `x64` can we not avoid all of these ifdefs and have a single `using namespace`?

src/cpu/rnn/ref_rnn.cpp
```diff
@@ -533,6 +534,193 @@ ref_rnn_common_t<aprop, src_type, weights_type, acc_type>::pd_t::init_brgemm(
     }
 
     return status::success;
+
+#elif DNNL_AARCH64
```
jondea: This is probably the biggest case. Why do we need these large additions in common code if the APIs are the same?

jondea (changes requested):
Thank you for this patch, the speedups look great!

It is a lot of code though, and I assume very similar to the `x64` version. Could you explain your decision to duplicate rather than use architecture agnostic common utilities? And further to also detail what (if anything) you had to change to get it working on AArch64. That would make this patch much more reviewable and maintainable going forward.

abhishek-iitmadras:
> Thank you for this patch, the speedups look great!
> 
> It is a lot of code though, and I assume very similar to the `x64` version. Could you explain your decision to duplicate rather than use architecture agnostic common utilities? 


Hi @jondea 
Now most of the common utilities code is under ifdef of `#if DNNL_X64 || DNNL_AARCH64`. Without any issue occur on x64 as i test on x64 machine also and all build and test passed sucessfully.

> Plan going forward

This is Phase A of a 7-phase effort to bring full optimize RNN support to AArch64. Structured whole RNN development as each phase is a self-contained, reviewable unit with a clear scope which make it easier on my end to develop/tested and also for community end to review easily. 

**Phase A (this PR)**: BRGEMM core — the non-JIT compute path, end-to-end fwd + bwd
**Phase B**: JIT helper kernels — bf16 emulation and fp32 add-to-bf16 convert for bf16 training path
**Phase C**: JIT helper kernels — transpose, gates reduction, peephole, BF16 conversion
**Phase D**: Post-GEMM infrastructure — shared abstractions and vanilla RNN cell post-GEMM
**Phase E**: LSTM optimization
**Phase F**: GRU optimization
**Phase G**: LBR-GRU optimization

abhishek-iitmadras:
Hi @jondea 

Please approve other workflows in the CI

src/cpu/rnn/postgemm_dispatcher.hpp
abhishek-iitmadras: > This change appears to be a no-op. Do we need it?
> 
> Although on the other hand, this looks suspicious. I would expect that is `x64` has to check the status of init, `aarch64` would too

We don't need this for now for aarch64 as its jit post-gemm init coming in next PRs for aarch64.

vpirogov:
> Please approve other workflows in the CI

Here you go.

abhishek-iitmadras:
Related CI's are green.

Please have a review.
cc: @jondea

src/cpu/rnn/ref_rnn.cpp
```diff
@@ -2052,9 +2280,18 @@ status_t ref_rnn_common_t<aprop, src_type, weights_type, acc_type>::execute(
...
+    const auto scratch_gates_blocked
+            = scratchpad.template get<scratch_t>(key_rnn_gates_blocked);
+    const auto scratch_src_layer
+            = scratchpad.template get<scratch_t>(key_rnn_src_layer_trans);
+    const auto scratch_src_iter
+            = scratchpad.template get<scratch_t>(key_rnn_src_iter_trans);
 #endif
 
     gemm_acc_t *amx_scratchpad = nullptr;
```
jondea, line 2292: AMX is x64 specific, it would be safer to put this in an ifdef rather than using MAYBE_UNSED

src/cpu/rnn/ref_rnn.cpp
```diff
@@ -2052,9 +2280,18 @@ status_t ref_rnn_common_t<aprop, src_type, weights_type, acc_type>::execute(
             = scratchpad.template get<scratch_t>(key_rnn_src_layer_trans);
     const auto scratch_src_iter
             = scratchpad.template get<scratch_t>(key_rnn_src_iter_trans);
+#elif DNNL_AARCH64
```
jondea, line 2283: We don't need this ifdef, the two blocks are identical

jondea (changes requested):
I'm not sure that so many phases make sense. It is likely that by the last phase we will have forgotten important context for any refactoring. The purpose of a review is to assess the whole design atomically. Doing this in phases is going to mean adding a lot of duplicated code, and then removing it again. It would be easier and much quicker to just do this all in one review.

If a PR has to be large to get the full picture, then that's how it should be.

With this aside, I think even for this PR I don't think we have exhausted all the ways to make the code common. E.g. the amx scratchpad can be made generic (maybe a nullptr for AArch64?) so that the interface can be common.

src/cpu/rnn/ref_rnn.hpp
abhishek-iitmadras: Ok Done,
Now under `#if DNNL_X64 || DNNL_AARCH64`

src/cpu/rnn/ref_rnn.cpp
abhishek-iitmadras: DOne and now under umbrella of `#if DNNL_X64 || DNNL_AARCH64`

src/cpu/rnn/ref_rnn.cpp
abhishek-iitmadras: Ok DOne , amx here is put in ifdef of `#if DNNL_X64`

src/cpu/rnn/ref_rnn.cpp
abhishek-iitmadras: Now both x64 and aarch64 is under ifdef like` #if DNNL_X64 || DNNL_AARCH64`

src/cpu/rnn/ref_rnn.hpp
```diff
@@ -32,12 +33,27 @@
...
+#define RNN_AMX_SCRATCHPAD_ACTUAL amx_scratchpad,
+#else
+#define RNN_AMX_SCRATCHPAD_ACTUAL
+#endif
+
 namespace dnnl {
 namespace impl {
 namespace cpu {
 
+#if DNNL_X64
```
jondea: The typical way to deal with this in oneDNN is with a `using namespace` in the include ifdef. You can see this pattern in the impl lists.

src/cpu/rnn/rnn_utils.hpp
```diff
@@ -31,6 +32,14 @@
 
 #if DNNL_X64
 #include "cpu/x64/cpu_isa_traits.hpp"
+#elif DNNL_AARCH64
+#include "cpu/aarch64/cpu_isa_traits.hpp"
+#endif
+
+#if DNNL_X64
+#define RNN_AMX_SCRATCHPAD_ARG gemm_acc_t *amx_scratchpad,
```
jondea: This macro magic does not feel right, I'm not sure I can explain why, but I don't think we need to do this. AArch64 could just accept an accumulation scratchpad pointer too (although it can be a nullptr for now). I wouldn't be surprised if the AArch64 implementation will _need_ a scratchpad for some implementations (e.g. for those using MMLA instructions).

abhishek-iitmadras: Agree, Now removed macros and make x64/AArch64 signatures identical

src/cpu/rnn/ref_rnn.hpp
abhishek-iitmadras: Agreed , DOne

abhishek-iitmadras:
Please help with running all CI's here

Sqvid:
@abhishek-iitmadras 
> Please help with running all CI's here

I believe the CI has run all checks. @jondea do you have another round of comments?

src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
```diff
@@ -0,0 +1,520 @@
...
+#include "cpu/aarch64/rnn/brgemm_cell_common_reorders.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+using namespace dnnl::impl::utils;
+
+template <typename weights_t, typename scratch_t, typename gemm_acc_t>
```
jondea, line 32: In the x64 version it looks like there are some comments here which explain the maths. Is there a reason they haven't been copied across?

src/cpu/aarch64/rnn/brgemm_cell_common_fwd.cpp
```diff
@@ -0,0 +1,907 @@
...
+#if DNNL_CPU_THREADING_RUNTIME == DNNL_RUNTIME_TBB
+    constexpr dim_t k2_i8_asimd = 192;
+    constexpr dim_t k2_f32_asimd = 112;
+#elif DNNL_CPU_THREADING_RUNTIME == DNNL_RUNTIME_OMP
+    constexpr dim_t k2_i8_asimd = 176;
+    constexpr dim_t k2_f32_asimd = 96;
+#endif
+#if (DNNL_CPU_THREADING_RUNTIME == DNNL_RUNTIME_TBB) \
+        || (DNNL_CPU_THREADING_RUNTIME == DNNL_RUNTIME_OMP)
+    if (rnn_.brgemm_isa == aarch64::asimd && !need_gemm_layer_ && rnn_.M == 1
```
jondea: We don't have asimd brgemm kernels, so this line seems odd. What is it trying to do on AArch64? And why does that directly translate from `x64::avx2`?

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
```diff
@@ -0,0 +1,945 @@
...
+
+aarch64::cpu_isa_t brgemm_calc_isa(
+        const cpu::rnn_utils::rnn_conf_t &rnn, dim_t K1, dim_t K2) {
+
+    // Feature gating: reject early if HW lacks required dtype support
+    if (rnn.is_cell_dt_bf16() && !mayiuse_bf16()) return isa_undef;
+    // TODO: Add int8 dotprod / f16 gating if brgemm_desc_init
+    //       doesn't handle it robustly enough downstream.
+
+    if (mayiuse(sve_512)) return sve_512;
```
jondea, line 73: Can we just use `sve` instead of being vector length specific at this point. As far as I know, `jit_brgemm_kernel_t` doesn't require us to be.

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
```diff
@@ -0,0 +1,945 @@
...
+
+    return std::make_pair(K1, K2);
+}
+
+std::pair<dim_t, dim_t> brgemm_calc_k_block_vanilla_rnn(dim_t K1, dim_t K2,
+        dim_t M, dim_t n_block, dim_t src_layer_type_size, dim_t As, dim_t Bs,
+        dim_t Cs, dim_t l2_cache_size, bool is_xf16, aarch64::cpu_isa_t isa) {
+
+    const bool is_sve = is_superset(isa, aarch64::sve_128);
+    const float l2_occupancy = is_sve ? 0.35f : 0.70f;
```
jondea: This has been copied verbatim from x64 applied from `x64::avx512_core`, but it doesn't make sense when applied to `sve`. The L2 occupancy is orthogonal to the ISA. It might be related to the vector length, but you have applied it here to _all_ vector lengths. I think it would be worth rationalise this to some function of vector length/L2 cache size, or just remove it completely

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
```diff
@@ -0,0 +1,945 @@
...
+    if (work_by_N > 2.0f || (work_by_N > 1.0f && adj_by_l2))
+        return M;
+    else
+        return adjust_m_block_lstm(nthr, M, N_blocks);
+}
+
+dim_t adjust_m_block_lstm(dim_t nthr, dim_t M, dim_t N_blocks) {
+
+    const dim_t max_m_blocks = 4 * utils::div_up(nthr, N_blocks);
+    const dim_t max_m_value = 32;
```
jondea: It would be good to explain what these variables mean. Also, are any of them dependent on vector length? Looking at the x64 code would imply that they could be, but these have not been adjusted for such.

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
```diff
@@ -0,0 +1,945 @@
...
+        }
+
+    if (m_block == 1) m_block = M;
+    return m_block;
+}
+
+dim_t brgemm_calc_n_block(
+        const cpu::rnn_utils::rnn_conf_t &rnn, alg_kind_t cell_kind) {
+
+    const int simd_w = isa_max_vlen(rnn.brgemm_isa) / (int)sizeof(float);
```
jondea: nit: use `simd_elems`

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
```diff
@@ -0,0 +1,945 @@
...
+        cpu::rnn_utils::rnn_conf_t &rnn, alg_kind_t cell_kind,
+        dim_t src_layer_type_size, dim_t scratch_type_size) {
+
+    using namespace cpu::rnn_utils;
+
+    // no int8 brgemm RNN support for now
+    if (rnn.is_int8_conf() || rnn.is_cell_dt_int8())
+        return status::unimplemented;
+
+    // FOr now only support vanilla RNN/LSTM/GRU/AUGRU
```
jondea: nit: typo

jondea (changes requested):
Given the enormity of this PR I have focused on reviewing just the diff between this and the `x64/rnn` directory. I have left some minor comments, but otherwise it looks good.

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
abhishek-iitmadras: Corrected

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
abhishek-iitmadras: Done

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
abhishek-iitmadras: Done ,provide comment with detail explanations

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
abhishek-iitmadras: Done , 
Values of L2 occupancy was takes by performing "L2 occupancy sweep" — VANILLA_RNN, Graviton3 (c7g.metal, sve_256, 1 MiB L2), 15 shapes, min-of-5. Metric is worst-case per-shape regret vs each shape's best 
(lower = better).

| L2 Occupancy | Cold (Worst-Case Regret) | Warm (Worst-Case Regret) |
|-------------:|-------------------------:|-------------------------:|
| 0.20 | 13.9% | 14.7% |
| 0.25 | 14.3% | 15.4% |
| 0.30 | 13.8% | 14.6% |
| 0.35 | 13.5% | 13.5% |
| 0.40 | 11.2% | 9.8% |
| **0.50** | **4.3%** | **4.9%** |
| 0.65 | 6.6% | 16.6% |
| 0.75 | 16.8% | 9.6% |
| 0.85 | 8.9% | 6.5% |

make SVE = 0.50 — clear winner in both warm and cold

You were right that this is ISA-independent: it multiplies the runtime-queried L2 size, so I kept it a single constant rather than function of vector length.

src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
abhishek-iitmadras: Now provide Comment about maths

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
abhishek-iitmadras: Tried this but had to revert as i got LSTM test failures.
maybe the brgemm kernel infrastructure (brgemm_desc_init / brgemm_kernel_create) internally dispatches via is_superset(isa, sve_128)-style checks, which return false on plain sve (only sve_bit | asimd_bit set).
Keeping concrete VL here for now

abhishek-iitmadras:
> Given the enormity of this PR I have focused on reviewing just the diff between this and the `x64/rnn` directory. I have left some minor comments, but otherwise it looks good.


thanks @jondea for your positive review and Sorry for the long delay in replying to your latest comments. I had serious health emergency

src/cpu/aarch64/rnn/brgemm_cell_common_fwd.cpp
abhishek-iitmadras: Correct ,
Rather than port a dead branch or invent an SVE threshold with no data behind it, I removed the heuristic and return the full thread count now and also put detail comment

src/cpu/aarch64/rnn/brgemm_cell_common_fwd.cpp
```diff
@@ -0,0 +1,889 @@
...
+        const auto nb = (rnn_.unfused_post_gemm) ? nb_i / rnn_.n_gates : nb_i;
+        const auto n = nb * rnn_.n_block;
+        const auto g_unfused
+                = (rnn_.unfused_post_gemm) ? nb_i % rnn_.n_gates : 0;
+
+        const auto *const Al_m = Al_ + m * LDAl_;
+        const auto *const Ai_m = Ai_ + m * LDAi_;
+        const auto *const Bl_n = Bl_ + nb * Bl_n_offset_;
+        const auto *const Bi_n = Bi_ + nb * Bi_n_offset_;
+        auto *const C_n = C_gates_ + m * rnn_.LDC + n;
```
Sqvid: I might be missing some domain-specific context here, but I think very terse variable names make it really hard for a future developer to easily understand the code. I think more descriptive variable names are worth the extra characters in this case.

I had to look at the initialisation to see what `Al` vs `Ai` was. Why not just
`s/Al/A_layer/g` and `s/Ai/A_iter/g` for example? `LDA/LDB/LDC` are fine to me since at least they are very standard in numerical code at this point.

Sqvid:
> thanks @jondea for your positive review and Sorry for the long delay in replying to your latest comments. I had serious health emergency

I left a non-blocking comment until @jondea re-reviews. Sorry to hear about your health emergency, hope you are better now :slightly_smiling_face:

src/cpu/aarch64/rnn/brgemm_cell_common_fwd.cpp
abhishek-iitmadras: Agreed and Done. 
thanks for flagging it

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
jondea: No problem, thanks for trying!

src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
jondea: It looks like nothing has changed here?

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
jondea: Sorry, I should have explained a bit more. There is a function called `simd_elems` which returns how many SIMD elems there are of a given datatype type for a given ISA.

src/cpu/aarch64/rnn/brgemm_cell_common_bwd.cpp
abhishek-iitmadras: its is in src/cpu/aarch64/rnn/brgemm_cell_common_bwd.hpp , not in this .cpp file

jondea: So it is, sorry!

src/cpu/aarch64/rnn/brgemm_cell_common_fwd.cpp
jondea: Makes sense, thanks

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
jondea: Thanks for investigating

src/cpu/aarch64/rnn/rnn_brgemm_utils.cpp
jondea: This was a nit though, so happy to approve without

jondea:
> > Given the enormity of this PR I have focused on reviewing just the diff between this and the `x64/rnn` directory. I have left some minor comments, but otherwise it looks good.
> 
> thanks @jondea for your positive review and Sorry for the long delay in replying to your latest comments. I had serious health emergency

No problem, thank you for addressing the comments. Sorry to hear about this, I hope you are better now!

jondea (approved):

abhishek-iitmadras:
> I left a non-blocking comment until @jondea re-reviews. Sorry to hear about your health emergency, hope you are better now 🙂

Hi @Sqvid , please help with review and merge
Regarding my health, feeling better now 😊
Thanks

Sqvid:
You need a review from the x64 side. @densamoilov would you be able to take a look? Thank you!

abhishek-iitmadras:
Hi @densamoilov , 
Could you please take a look at this when you are free ?

densamoilov (approved):

PR #4903, opened 2026-03-26
cpu: aarch64: introduce assembly loop generation utils

Sqvid:
# Description
This PR introduces two new assembly utilities:
- `asm_for`
- `asm_do_while`

The intent is to simplify the creation of complex loop structures and improve readability by abstracting away the need to manually place labels and jumps. This can be especially helpful for nested loops.

I have also used these utilities to simplify the code in `jit_int8_matmul*`, and `jit_brgemm_post_ops` to demonstrate the usage model and the improvement to readability.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/jit_brgemm_post_ops.hpp
+++ b/src/cpu/aarch64/jit_brgemm_post_ops.hpp
@@ -141,6 +141,7 @@ private:
     const XReg reg_out = x14;
     const XReg aux_reg_in = x13;
     const XReg aux_reg_out = x12;
+    const XReg reg_mb_loop = x16;
 
     const XReg reg_bias = x11;
     const XReg aux_reg_bias = x10;
@@ -773,23 +774,16 @@ private:
 
         // Unrolling over mb can result in very large post-op assembly being
         // generated (see Issue #4089), so branch and loop instead.
-        if (mb > 0) {
-            Label mb_loop;
-            mov_imm(x16, mb);
-
-            L(mb_loop);
-
+        asm_for(reg_mb_loop, mb, [&]() {
             loop_by_N(m_block, nb2, nb2_tail, nb_tail);
 
             if (brg.alpha != 0)
                 add_imm(reg_in, reg_in, inp_typesize_ * (m_block * brg.LDC),
                         X_TMP_0);
+
             add_imm(reg_out, reg_out, out_typesize_ * (m_block * LDD_),
                     X_TMP_0);
-
-            subs(x16, x16, 1);
-            bgt(mb_loop);
-        }
+        });
 
         if (mb_tail > 0) loop_by_N(mb_tail, nb2, nb2_tail, nb_tail);
 


src/cpu/aarch64/jit_generator.hpp
```diff
@@ -674,6 +674,45 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
             dup(v_tmp, w_tmp);
     }
 
+    // Roughly equivalent to:
+    // for(index = max - 1; index >= 0; --index) {
+    //   body();
+    // }
+    template <typename TReg, typename Func>
```
jondea, line 680: Func takes no args and returns nothing, can we not concretely type it?

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -674,6 +674,45 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
...
+    void asm_for(const TReg &index_reg, const TReg &max_reg, const Func &body) {
```
jondea, line 681: Optional: is it worth providing one with an immediate max?

src/cpu/aarch64/jit_generator.hpp
Sqvid: I did want to. But if you use `std::function<void(void)>` then you can't pass in a lambda. And doing [something like `decltype([&](){} -> void)` isn't allowed either](https://stackoverflow.com/questions/36030589/i-cannot-pass-lambda-as-stdfunction).

jondea: Ah okay, no problem.

src/cpu/aarch64/jit_generator.hpp
Sqvid: That would let us pass in the immediate as a C++ variable but the concern I have is the need to throw a runtime error if the value passed in cannot be moved into a register in a single instruction.

We could check the immediate at compile time if we take it in as a template parameter but would that be useful in our typical use-cases?

jondea: We have a `mov_imm` function which handles all cases.

Sqvid: After looking into it, I still think we can run into [cases where we get runtime exceptions](https://github.com/fujitsu/xbyak_aarch64/blob/dd44ddedf9dc6398c063e24c4d21f8c17aa5b12f/xbyak_aarch64/xbyak_aarch64_meta_mnemonic.h#L456-L458). For instance if `TReg` is deduced to `WReg` but the user passes in a value greater than the 32-bit signed/unsigned max.

Sqvid: I have added a version that accepts an immediate value but only with a 64-bit index register. This should allow us to sidestep overflow exceptions.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -674,6 +673,67 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
...
+        L(loop_end);
+    }
+
+    // Roughly equivalent to:
+    // do {
+    //   body();
+    // } while (--n > 0)
+    template <typename TReg, typename Func>
+    void asm_do_while(const TReg &index_reg, const Func &body) {
+        Xbyak_aarch64::Label loop_begin, loop_end;
```
renato-arantes: ``loop_end`` is declared but never used.

Sqvid: Done.

PR #4907, opened 2026-03-26
cpu: x64: xbyak: enable XBYAK_STRICT_CHECK_MEM_REG_SIZE Xbyak v7.25 feature

georgen117:
## Enable the `XBYAK_STRICT_CKECK_MEM_REG_SIZE` feature.

The `XBYAK_STRICT_CKECK_MEM_REG_SIZE` feature has been part of Xbyak since v7.25 but was not imported into OneDNN till the recent PR to update Xbyak to v7.35.3.

## What the `XBYAK_STRICT_CHECK_MEM_REG_SIZE` feature does

If a memory operand has an explicit size annotation (e.g. `dword[]` = 32, `zword[]` = 512) and that size does not match the register's bit-width, it will throw an error immediately during JIT kernel construction. The check is intentionally non-intrusive for unannotated `ptr[]` addresses.

# Why this check is valuable

## Catches subtle bugs at the point of emission, not at runtime

Without this check, a mismatch operand pair like `mov(Reg64, dword[...])` assembles silently into well-formed but wrong machine code (a 32-bit `mov`) that the programmer may have intended to be a 64-bit load. The bug would manifest as a subtle data corruption or wrong data result at inference time. Often with no crash, and no assertion. The check converts this class of bug from nondeterministic runtime anomaly into an immediate error when the JIT kernel is generated.

## Helps enforce x86 operand-size discipline in JIT code

Unlike C++ code the JIT code does not protect against unit32_t/uint64_t mismatches. This check helps fill that gap, acting like a type-system layer for the assembly-level operand size.

## Zero runtime overhead
The check runs only during JIT kernel construction, never during the execution of the generated code. In testing the extra check was negligible.

## Enabling the check Aligns oneDNN with the upsteam intent of the Xbyak maintainer.

The `XBYAK_STRICT_CHECK_MEM_REG_SIZE` check defaults to `1` in the upstream project. With a comment explicitly calling out that the macro "may be removed in future version". Indicating that this will eventually become the unconditional default.

### DIFF
--- a/src/cpu/x64/cpu_isa_traits.hpp
+++ b/src/cpu/x64/cpu_isa_traits.hpp
@@ -46,13 +46,8 @@
 #define XBYAK_NO_EXCEPTION
 #endif
 
-// Several JIT kernels intentionally or unintentionally use size-mismatched memory operands
-// (e.g. zword[]-annotated addresses with 64-bit GPR instructions, or dword[] fields loaded into
-// Reg64 registers). These are architecturally correct on x86-64 but trigger this check. Until all
-// call sites are updated to use properly-sized operands or stripped annotations, set this to 0 to
-// suppress the false-positive errors.
 #if !defined(XBYAK_STRICT_CHECK_MEM_REG_SIZE)
-#define XBYAK_STRICT_CHECK_MEM_REG_SIZE 0
+#define XBYAK_STRICT_CHECK_MEM_REG_SIZE 1
 #endif
 
 #if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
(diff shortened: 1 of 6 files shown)


georgen117:
make test
set test_scope=NIGHTLY

georgen117:
make test
set test_scope=NIGHTLY

georgen117:
make test
set test_scope=NIGHTLY
disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

amakarev (approved):
lgtm

inteldimitrius (approved):

PR #4909, opened 2026-03-27
graph: backend: dnnl: fix conditions of f32 sdpa dispatching

TaoLv:
Fix the conditions of f32 sdpa dispatching.
Previously the checks in the code actually did not align with the error message.
Now with this PR, f32 sdpa with implicit causal masks will be dispatched to gpu ukernel sdpa. Other cases will dispatch to primitive based kernels.


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

rongzha1 (approved):

PR #4910, opened 2026-03-27
graph: backend: dnnl: fix a symbol leak issue caused by std::call_once

TaoLv:
- Forward port PR: #4901 
- Fix MFDNN-14824


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #4911, opened 2026-03-27
cpu: aarch64: add LUT-based `gelu_erf` JIT implementations

Anndrey24:
# Description

This commit adds ASIMD and SVE JIT implementations for `gelu_erf` with LUT-based ERF computation, adapting the `verfq_f32` implementation from Arm Compute Library (ACL).

The ASIMD implementation is a direct adaptation of the ACL kernel, which uses a single 513-pair LUT with interleaved `erf(r)` and `scale(r)` values. The SVE implementation uses the same approach, but stores `erf(r)` and `scale(r)` in separate LUTs so it can use `ld1w` gather loads directly. The existing 512-bit SVE minimax path remains unchanged.

Also, a few existing `vlen` checks were fixed to gate the 512-bit-only paths correctly.

### c6g

| Shape     | Threads | jit:asimd (ms)   | acl (ms)        | Speedup |
|-----------|---------|-------------|-------------|---------|
| 1539x387  | 1       | 1.27431     | 1.434686667 | 1.13    |
| 1539x387  | 4       | 0.319794333 | 0.380329    | 1.19    |
| 1539x387  | 16      | 0.0825511   | 0.100743    | 1.22    |
| 1539x387  | 64      | 0.030120033 | 0.0449317   | 1.49    |
| 1024x4096 | 1       | 8.93697     | 10.13786667 | 1.13    |
| 1024x4096 | 4       | 2.239206667 | 2.5386      | 1.13    |
| 1024x4096 | 16      | 0.563989667 | 0.64518     | 1.14    |
| 1024x4096 | 64      | 0.150102    | 0.180736667 | 1.20    |
| 4096x4096 | 1       | 35.73536667 | 40.12503333 | 1.12    |
| 4096x4096 | 4       | 8.940213333 | 10.1733     | 1.14    |
| 4096x4096 | 16      | 2.240803333 | 2.55782     | 1.14    |
| 4096x4096 | 64      | 0.588767    | 0.664909333 | 1.13    |

### c7g

| Shape     | Threads | jit:sve_256 (after) | jit:sve_256 (before) | Speedup |
|-----------|---------|---------------------|----------------------|---------|
| 1539x387  | 1       | 0.51054             | 0.959155             | 1.88    |
| 1539x387  | 4       | 0.130314            | 0.24537              | 1.88    |
| 1539x387  | 16      | 0.0350539           | 0.0635002            | 1.81    |
| 1539x387  | 64      | 0.0178918           | 0.025823             | 1.44    |
| 1024x4096 | 1       | 3.59149             | 6.74169              | 1.88    |
| 1024x4096 | 4       | 0.903596            | 1.71063              | 1.89    |
| 1024x4096 | 16      | 0.228847            | 0.433806             | 1.90    |
| 1024x4096 | 64      | 0.0647632           | 0.123271             | 1.90    |
| 4096x4096 | 1       | 14.3601             | 27.1085              | 1.89    |
| 4096x4096 | 4       | 3.60209             | 6.80219              | 1.89    |
| 4096x4096 | 16      | 0.909855            | 1.72027              | 1.89    |
| 4096x4096 | 64      | 0.283271            | 0.446649             | 1.58    |

### c8g
| Shape     | Threads | jit:sve_128 (after) | jit:sve_128 (before) | Speedup |
|-----------|---------|---------------------|----------------------|---------|
| 1539x387  | 1       | 0.520478667         | 1.70666              | 3.28    |
| 1539x387  | 4       | 0.1316              | 0.428746             | 3.26    |
| 1539x387  | 16      | 0.034794133         | 0.109844             | 3.16    |
| 1539x387  | 64      | 0.015572533         | 0.0326526            | 2.10    |
| 1024x4096 | 1       | 3.669626667         | 12.0174              | 3.27    |
| 1024x4096 | 4       | 0.918106            | 3.00667              | 3.27    |
| 1024x4096 | 16      | 0.232221333         | 0.755418             | 3.25    |
| 1024x4096 | 64      | 0.078077167         | 0.195619             | 2.51    |
| 4096x4096 | 1       | 14.93903333         | 48.0708              | 3.22    |
| 4096x4096 | 4       | 3.69309             | 12.0244              | 3.26    |
| 4096x4096 | 16      | 0.936244333         | 3.0116               | 3.22    |
| 4096x4096 | 64      | 0.283405            | 0.761588             | 2.69    |

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -1551,7 +1551,9 @@ size_t jit_uni_eltwise_injector_t<isa>::aux_vecs_count() {
     if (is_fwd_) {
         switch (alg_) {
             case eltwise_relu_use_dst_for_bwd:
-            case eltwise_relu: return (alpha_ == 0.f) ? 2 : 3;
+            case eltwise_relu:
+                return (isa == asimd) ? ((alpha_ > 1.f) ? 3 : 2)
+                                      : ((alpha_ == 0.f) ? 0 : 2);
             case eltwise_elu_use_dst_for_bwd:
             case eltwise_elu: return (isa == asimd) ? 7 : 5; /* = exp + 2 */
             case eltwise_tanh_use_dst_for_bwd:
@@ -2662,10 +2664,20 @@ void jit_uni_eltwise_injector_t<asimd>::relu_zero_ns_compute_vector_fwd(
 template <>
 void jit_uni_eltwise_injector_t<asimd>::relu_compute_vector_fwd(
         const TRegS &vmm_src) {
-    // vmm_tmp = alpha * vmm_tmp
-    h->fmul(vmm_aux0, vmm_src, z_tmp);
-    // vmm_src = max(vmm_src, vmm_tmp)
-    h->fmaxnm(vmm_src, vmm_src, vmm_aux0);
+    // Compute x > 0 ? x : alpha * x.
+    // For alpha <= 1, this is equivalent to max(x, alpha * x).
+    if (alpha_ <= 1.f) {
+        h->fmul(vmm_aux0, vmm_src, z_tmp);
+        h->fmaxnm(vmm_src, vmm_src, vmm_aux0);
+    } else {
+        // For alpha > 1, keep positive lanes unchanged and scale only
+        // x <= 0 lanes.
+        h->mov(VReg16B(vmm_aux1.getIdx()), VReg16B(vmm_src.getIdx()));
+        h->fcmgt(vmm_aux0, vmm_aux1, 0.);
+        h->fmul(vmm_src, vmm_src, z_tmp);
+        h->bit(VReg16B(vmm_src.getIdx()), VReg16B(vmm_aux1.getIdx()),
+                VReg16B(vmm_aux0.getIdx()));
+    }
 }
 
 template <>
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -52,22 +52,10 @@ bool is_alg_supported(alg_kind_t alg) {
 }
 
 bool is_supported(cpu_isa_t isa, alg_kind_t alg) {
-    if (is_isa_supported(isa)) {
-        if (isa == asimd) {
-            using namespace alg_kind;
-            return utils::one_of(alg, eltwise_relu, eltwise_tanh, eltwise_elu,
-                    eltwise_square, eltwise_abs, eltwise_sqrt, eltwise_linear,
-                    eltwise_soft_relu, eltwise_logistic, eltwise_mish,
-                    eltwise_exp, eltwise_gelu_tanh, eltwise_hardsigmoid,
-                    eltwise_hardswish, eltwise_swish, eltwise_clip,
-                    eltwise_clip_v2, eltwise_gelu_erf, eltwise_round,
-                    eltwise_clip_v2_use_dst_for_bwd);
-        } else {
-            return is_alg_supported(alg);
-        }
-    } else {
-        return false;
-    }
+    using namespace alg_kind;
+    // ASIMD does not currently support eltwise_log.
+    if (isa == asimd && alg == eltwise_log) return false;
+    return is_isa_supported(isa) && is_alg_supported(alg);
 }
 
 } // namespace eltwise_injector
@@ -1130,7 +1118,7 @@ void jit_uni_eltwise_injector_t<isa>::gelu_erf_compute_vector_fwd(
     //   erf(x) = erf(r) + scale(r) * d * (1 - r * d - 1/3 * d^2)
     //
     // where:
-    //   r = floor(x * 128) / 128
+    //   r = round(x * 128) / 128
     //   d = x - r
     //
     // erf(r) and scale(r) are stored in a 513-entry lookup table.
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: Are there now any ASIMD algorithms we don't support? We should probably simplify this statement either way.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2976,6 +3011,88 @@ void jit_uni_eltwise_injector_t<asimd>::swish_compute_vector_fwd(
...
+    // erf(r) and scale(r) are stored in a 513-entry lookup table.
+    // The LUT covers the range from 0 to 4 with the step of 1/128.
+    //
+    // Special cases:
+    //   erf(x) =  1 for x >  3.9375
+    //   erf(x) = -1 for x < -3.9375
+
+    // Find LUT indices by rounding to the step of 1/128.
+    // `gelu_erf_lut_shift` pushes out the 16 LSBs. Only 7 fraction bits remain.
+    h->fadd(vmm_aux2, vmm_aux1, table_val(gelu_erf_lut_shift, z_tmp)); // z
```
jondea: Could this be a shift by immediate?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -1111,64 +1111,74 @@ template <cpu_isa_t isa>
...
+    //   erf(x) = -1 for x < -3.9375
+
+    // Track saturation mask from |x| and clamp x used for LUT indexing.
+    table_val(gelu_erf_lut_max, vmm_aux0);
+    h->facgt(p_tmp0.s, p_all / T_z, vmm_aux5, vmm_aux0);
+    h->fmin(vmm_aux1, p_all / T_m, vmm_aux0);
+
+    // Find LUT indices by rounding to the step of 1/128.
+    // `gelu_erf_lut_shift` pushes out the 16 LSBs. Only 7 fraction bits remain.
+    h->fadd(vmm_aux2, vmm_aux1, table_val(gelu_erf_lut_shift, z_tmp));
```
jondea: I think this can be shift by immediate

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -2443,6 +2459,25 @@ void jit_uni_eltwise_injector_t<isa>::register_table_entries() {
     if (need.gelu_erf()) push_entries_of(gelu_erf_polynomial);
     if (need.gelu_erf()) push_entries_of(gelu_erf_minimax_consts);
     if (need.gelu_erf()) push_entries_of(gelu_erf_minimax_polynomial);
+    if (need.gelu_erf()) {
+        push_entries_of(gelu_erf_lut_consts);
+        const int erf_lut_size = 513;
+        const float erf_lut_step = 1.f / 128.f;
+        const float two_over_sqrt_pi = 1.12837922573089599609375f;
+        for (int i = 0; i < erf_lut_size; i++) {
```
jondea, line 2460: Is this 513 vector lengths or are we just storing words?

jondea (commented):
This looks great, thank you. I just had some minor comments

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Great point! It seems that `eltwise_log` is the only algorithm left that doesn't have an ASIMD forward pass, so we could simplify the check to something like:
```
bool is_supported(cpu_isa_t isa, alg_kind_t alg) {
    using namespace alg_kind;
    // ASIMD does not currently support eltwise_log.
    if (isa == asimd && alg == eltwise_log) return false;
    return is_isa_supported(isa) && is_alg_supported(alg);
}
```
I've also noticed that the check here is direction-agnostic, and has been mainly treated as a "is the forward pass of this algorithm supported" check, leaving the responsibility of gating the unsupported backward passes to the caller of the eltwise injector. 

There is no ASIMD support for any of the eltwise backward passes yet, but they are currently gated by the `cpu_eltwise_list.cpp` list for the standalone eltwise primitive, while eltwise postops only exist for SVE, as far as I can tell. I'm not sure if this will cause some problems down the line, but I thought it was worth mentioning.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: It is 2 x 513 words (erf and scale), not vector lengths, which is defined by the `broadcast` argument being set to `false` in `push_arg_entry_of(gelu_erf_lut, float2int(erf_val), false)`. The LUT would therefore be 4,104 bytes long (≈4kB).

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: I believe this combination of `fadd -> fsub` is not quite equivalent to a simple shift, as it relies on floating-point rounding behaviour. Adding the `65536.0f` bias (`2^16`) forces the result into a range where the float ULP is `2^(16-23) = 1/128`, so the add rounds x to that grid (i.e. `round(x * 128) / 128`), while subtracting the bias recovers that quantised value. A shift-immediate would not match that exact behaviour, truncating instead of rounding.

I'll update the comments around it to better reflect this.

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Anndrey24: Same as other comment ^

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: Interesting, okay, thanks for giving it some thought anyway!

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: Much simpler, great. 

WRT backward, we could reject it here because the injector has an `is_fwd` member variable, but it's up to you. You didn't introduce the bug

jondea (approved):

Anndrey24:
While changing the ASIMD algorithm list from an inclusion list to an exclusion list, some `*_dst` forward paths with pre-existing implementation support were also enabled. 

Enabling ASIMD `relu_dst` exposed a pre-existing ASIMD `eltwise_relu` bug for `alpha > 1`, where the shared ReLU kernel produced incorrect outputs. This only showed up through `relu_dst` because the benchdnn eltwise CI/smoke coverage tests `relu_dst` with `alpha=2`, while plain `relu`, which uses the same implementation, has no equivalent `alpha > 1` coverage .

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
renato-arantes, line 2467: Why can't we precompute it and store the precomputed values, like the other tables such as `gelu_erf_minimax_polynomial`?

Anndrey24: We could precompute the table to save some time during the primitive creation / JIT generation stage, where this additional computation takes place, although it would require adding some much larger tables in `register_table_entries()` than `gelu_erf_minimax_polynomial` (i.e. 1026-entry table for ASIMD and 2x513-entry tables for SVE, so a possible ~2k extra lines).

As it currently stands, there should be no performance difference during the primitive execution stage between the two approaches (i.e. after the table is emitted), but I can try precomputing the values if we want to eliminate the additional primitive generation time too.

renato-arantes: No need to do this. I believe adding around 2k extra lines to this file is too much. Let's leave it as it is for now.

renato-arantes (approved):
LGTM

PR #4913, opened 2026-03-27
doc: Fix misspelling

andrey-khropov:
subj


andrey-khropov:
> You would need to update commit message as per https://github.com/uxlfoundation/oneDNN/blob/main/CONTRIBUTING.md#code-contribution-guidelines (probably something like "common: fix spelling", but otherwise good catch :)

done

PR #4914, opened 2026-03-27
cpu: x64: pool: enable different dt on src and dst in jit

asztukow:
Addresses MFDNN-14412.

This PR adds support for different data types on source and destination in jit pooling.
Depending on the direction:
- int8 -> fp and int8 -> int8 are handled by `jit_int`
- fp -> int8 is handled by `jit_uni`

This branch includes squashed commit of changes from [#3440](https://github.com/uxlfoundation/oneDNN/pull/3440) These changes enable correct saturation logic for int8, needed in `jit_uni` while destination is of int8 type.


vpirogov:
Closing as stale.

PR #4922, opened 2026-03-30
cpu: x64: matmul: add MXFP4 weights support

amakarev:
Description:

Add brgemm matmul support for MXFP4 (f4_e2m1) weights with e8m0 scales across multiple source data types:

f32:f4_e2m1:f32 — f4-to-f32 conversion in brgemm microkernel during B load, e8m0 scales applied via post-ops
bf16:f4_e2m1:{bf16,f32} — dequantization in copy_b kernel (f4→f32→bf16) for AMX/VNNI compute
f16:f4_e2m1:{f16,f32} — dequantization in copy_b kernel (f4→f32→f16) for AMX fp16, or f4→f32 on avx512_core_fp16
Each source type uses a dedicated is_{src}_with_f4_wei flag, separate from the existing *_with_int_wei paths, since f4 and integer weights use different load mechanisms in copy_b kernels.

Supported ISAs:

avx512_core (f32 source)
avx512_core_bf16 (bf16 source)
avx512_core_fp16 / avx512_core_amx_fp16 (f16 source)


src/common/matmul.cpp
```diff
@@ -665,6 +665,16 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
         }
     }
 
+    // FP4 types store 2 elements per byte, so K must be even
+    const bool is_f4_weights = one_of(
+            weights_desc->data_type, data_type::f4_e2m1, data_type::f4_e3m0);
+    if (is_f4_weights) {
+        const auto &w_dims = weights_desc->dims;
+        const int K = w_dims[weights_desc->ndims - 2];
```
mzhukova: int -> dim_t?

src/common/matmul_pd.hpp
```diff
@@ -218,7 +218,9 @@ struct matmul_pd_t : public primitive_desc_t {
                 ok = ok && wei_k_group_ok && wei_n_group_ok;
 
                 // Mask over K dim is allowed for fp types or weights decompression only.
-                if (types::is_integral_dt(weights_md(0)->data_type)) {
+                const bool is_wei_fp
```
mzhukova: I feel like double negation is actually making it harder to read the code, any particular reason to make this change?

src/common/matmul.cpp
```diff
@@ -665,6 +665,16 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
...
+        VCHECK_MATMUL(
+                !is_runtime_value(K) && K % 2 == 0, VERBOSE_BAD_PARAM, "K");
```
mzhukova: probably VERBOSE_BAD_DIM instead of VERBOSE_BAD_PARAM might be better

mzhukova (commented):
Do we have benchdnn input file that covers f32:f4:f32? 
Does it make sense to add gtest case to ensure K % 2 == 0 condition is tested for matmul?

src/common/matmul.cpp
amakarev: fixed

amakarev:
> Do we have benchdnn input file that covers f32:f4:f32? Does it make sense to add gtest case to ensure K % 2 == 0 condition is tested for matmul?

Thank you. Updated the input file. I think the tests make sense, added a few.

src/common/matmul.cpp
amakarev: You're right, I made the changes

src/common/matmul_pd.hpp
amakarev: Definitely agree. Refactored

amakarev:
found issue with amx path.fixing

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3973,12 +3984,20 @@ void jit_brgemm_matmul_copy_b_bf16_t<Vmm>::init_masks() {
...
             vmovdqa32(vmm_permd, ptr[reg_tmp]);
         }
+
+        if (is_src_f4) {
+            alignas(64) static constexpr const float f4_e2m1_table[16]
+                    = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f,
+                            -.5f, -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
+            mov(reg_tmp, reinterpret_cast<size_t>(f4_e2m1_table));
+            vmovdqa32(vmm_f4_lut, ptr[reg_tmp]);
+        }
```
georgen117, line 3999: This will interpret all f4 data types as f4_e2m1 what about f4_e3m0.  Shouldn't you handle that case as well. Similar to line 4451 of this document.

Should the if block be:
```cpp
        if (is_src_f4) {
            alignas(64) static constexpr const float f4_e2m1_table[16]
                    = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f,
                            -.5f, -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
            alignas(64) static constexpr const float f4_e3m0_table[16]
                    = {0.0f, .25f, .5f, 1.0f, 2.0f, 4.0f, 8.0f, 16.0f, -0.0f,
                            -.25f, -.5f, -1.0f, -2.0f, -4.0f, -8.0f, -16.0f};
            switch (conf_->orig_wei_dt) {
                case data_type::f4_e2m1:
                    mov(reg_tmp, reinterpret_cast<size_t>(f4_e2m1_table));
                    break;
                case data_type::f4_e3m0:
                    mov(reg_tmp, reinterpret_cast<size_t>(f4_e3m0_table));
                    break;
                default: assert(!"unsupported f4 data type");
            }
            vmovdqa32(vmm_f4_lut, ptr[reg_tmp]);
        }
``` 

I took the f4_e3m0_table from line 4454 I assume that is correct.

I may be wrong about needing f4_e3m0. My investigation indicated that both `is_bf16_with_f4_wei` and `is_f16_with_f4_wei` include `f4_e3m0` and will make this code reachable.

If this is reachable with `f4_e3m0` type then the if block as is will silently produce the wrong results for `f4_e3m0` weights.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -5696,6 +5741,14 @@ void jit_brgemm_matmul_copy_b_cvt_bf16_t<Vmm>::init_masks() {
...
         mov(reg_tmp, 0x5555555555555555);
         kmovq(k5555, reg_tmp);
+
+        if (is_src_f4_) {
+            alignas(64) static constexpr const float f4_e2m1_table[16]
+                    = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f,
+                            -.5f, -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
+            mov(reg_tmp, reinterpret_cast<size_t>(f4_e2m1_table));
+            vmovdqa32(vmm_f4_lut, ptr[reg_tmp]);
+        }
```
georgen117, line 5781: Same issue with f4_e3m0 here.

The logic will silently output values from the `f4_e2m1` even for `f4_e3m0` inputs.

suggested change:
```cpp
        if (is_src_f4_) {
            alignas(64) static constexpr const float f4_e2m1_table[16]
                    = {0.0f, .5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f, -0.0f,
                            -.5f, -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f};
            alignas(64) static constexpr const float f4_e3m0_table[16]
                    = {0.0f, .25f, .5f, 1.0f, 2.0f, 4.0f, 8.0f, 16.0f, -0.0f,
                            -.25f, -.5f, -1.0f, -2.0f, -4.0f, -8.0f, -16.0f};
            switch (conf_->orig_wei_dt) {
                case data_type::f4_e2m1:
                    mov(reg_tmp, reinterpret_cast<size_t>(f4_e2m1_table));
                    break;
                case data_type::f4_e3m0:
                    mov(reg_tmp, reinterpret_cast<size_t>(f4_e3m0_table));
                    break;
                default: assert(!"unsupported f4 data type");
            }
            vmovdqa32(vmm_f4_lut, ptr[reg_tmp]);
        }
```

Again this is based on my understanding of the code. And the suggested table for f4_e3m0 came from line 4454 of the code I didn't come up with the numbers.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
georgen117: Random location for this comment:

the `f4_e2m1_table` shows up 6 times in the code. If you use my suggestions the `f4_e3m0_table` will also show up 6 times.  That is 6 possible places the table could be different. We should have one instance of each table that is used in all 6 locations.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -4896,7 +4932,8 @@ void jit_brgemm_matmul_copy_b_transposed_t<Vmm>::copy_row_x_col(
             const auto next_addr = EVEX_compress_addr(reg_src, next_src_offset);
             if (is_bf32_)
                 vmovups(src_next_masked, next_addr);
-            else if (is_bf16_with_int_wei_ || conf_->is_f16_with_int_wei) {
+            else if (is_bf16_with_int_wei_ || is_bf16_with_f4_wei_
+                    || conf_->is_f16_with_int_wei) {
```
georgen117: Should this add `is_f16_with_f4_wei_ like the branch at line 4905?

suggested change
```cpp
else if (is_bf16_with_int_wei_ || is_bf16_with_f4_wei_
            || is_f16_with_f4_wei_ || conf_->is_f16_with_int_wei) {
```

this would impact weights with transposed layout when `ncolumns > req_cvt_bf16_k_blk_step_`

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2629,6 +2621,10 @@ struct jit_brgemm_matmul_copy_b_common_t : public jit_brgemm_matmul_copy_b_t,
                 uni_vpslld(scale_vmm, scale_vmm, 16);
                 break;
             case data_type::f16: vcvtph2psx(scale_vmm, addr); break;
+            case data_type::e8m0:
+                vpbroadcastb(scale_vmm, addr);
+                uni_vpslld(scale_vmm, scale_vmm, 23);
```
georgen117, line 2613: If I understand this block correctly you are converting a f8_e8_m0 to f32.

I think there is a subtle bug here:  f8_e8_m0 does not have a sign bit. (if it has a sign bit then i have evaluated this wrong.)

lets say addr cotains the f8_e8m0 value of 1.0 which is 0x7F

if the `vpbraodcastb` fills every byte with 0x7F so now the f32 bits are `0x7F7F7F7F`
we then shift left by 23 we end up with `0xBF800000` The exponent bits[30:23] are correct but the sign bit is `1` not `0` which is incorrect.

The upper 3 bytes bits[31:8] before the shift need to be zeroed at least bit[8] needs to be zero.

suggestion:
option 1 (recommended)
```cpp
case data_type::e8m0:
    vpbroadcastb(scale_vmm, addr);
    uni_vpmovzxbd(scale_vmm, Xmm(scale_vmm.getIdx())); // zero-extend bytes
    uni_vpslld(scale_vmm, scale_vmm, 23);
    break;
```

or 
option 2
```
case data_type::e8m0:
    vpbroadcastb(scale_vmm, addr);
    vpandd(scale_vmm, scale_vmm, vmm_mask_0xff); // needs pre-loaded mask register (not shown)
    uni_vpslld(scale_vmm, scale_vmm, 23);
    break;
```

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1822,15 +1837,21 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
...
 
-    dim_t get_data_B_kn_off(dim_t k, dim_t n) const {
+    int wei_packed_elems_per_byte() const {
+        return utils::one_of(bgmmc_.orig_wei_dt, data_type::s4, data_type::u4,
+                       data_type::f4_e2m1, data_type::f4_e3m0)
+                ? 2
+                : 1;
+    }
+
+    dim_t get_data_B_kn_off(int k, int n) const {
```
georgen117: Shouldn't this be 
`dim_t get_data_B_kn_off(dim_t k, dim_t n) const {`

silently narrowind the data type from `dim_t` (`int64_t`) to `int`

The narrowing is only a risk for really large models but it is inconsistent with the rest of the API.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1865,7 +1886,8 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
         } else {
             b_off = wei_d_.off_l(b * bgmmc_.K * bgmmc_.N) * bgmmc_.b_dt_sz;
         }
-        if (bgmmc_.is_int4_weights) b_off = b_off / 2;
+        if (wei_packed_elems_per_byte() > 1)
+            b_off /= wei_packed_elems_per_byte();
```
georgen117: really minor: since `wei_packed_elms_per_byte() is really cheep but I had to look it up to see if it was trivial or computationally expensive.

`wei_packed_elems_per_byte() called twice;

suggestion:
```cpp
const int pack = wei_packed_elems_per_byte();
if (pack > 1) b_off /= pack;
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
tprimak: We don't need to implement e3m0. It's support is [deprecated](https://uxlfoundation.github.io/oneDNN/dev_guide_data_types.html#data-types) in the library and will be removed in the next major release. We just need to make sure cpu implementation return unimplemented rather than failing or segfaulting when you pass e3m0 type

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -220,8 +229,9 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             if (is_runtime_value(N())) ok = false;
         }
         // Impl suppports f32 scales only for non-weight decompression
```
georgen117: nit: typo in comment `suppports` -> `supports`

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -170,12 +170,19 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             = src_dt == f32 && wei_dt == bf16 && one_of(dst_dt, bf16, f32);
     const bool is_bf16_with_int_wei = src_dt == bf16
             && one_of(wei_dt, s8, u8, s4, u4) && one_of(dst_dt, bf16, f32);
+    const bool is_bf16_with_f4_wei = src_dt == bf16
+            && one_of(wei_dt, data_type::f4_e2m1, data_type::f4_e3m0)
```
georgen117: Discussion:

The `one_of(wei_dt, data_type::f4_e2m1, data_type::f4_e3m0)` check is duplicated in 8-10 locations that I could find.

brgemm_matmul.cpp lines 174, 179, 182
brgemm_matmul_utils.cpp lines 346, 360, 1574, 2004, 2009
brgemm_matmul_reorders.cpp lines 324, 329

Should a single helper function be added? 

A natural location for a helper function could be placed in `common/type_helpers.hpp` right beside the existing `is_integral_dt` function.

```cpp
inline bool is_f4_dt(data_type_t dt) {
    using namespace data_type;
    return utils::one_of(dt, f4_e2m1, f4_e3m0);
}
```

Thing is I don't see a similar helper for any of the fp8 code which would have also benefited from this type of helper function.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
georgen117: > We don't need to implement e3m0. It's support is [deprecated](https://uxlfoundation.github.io/oneDNN/dev_guide_data_types.html#data-types) in the library and will be removed in the next major release. We just need to make sure cpu implementation return unimplemented rather than failing or segfaulting when you pass e3m0 type

Thanks for letting me know. I saw it in some places in the code but not in others and it stood out as an oversight.

src/cpu/x64/matmul/brgemm_matmul.cpp
georgen117: based on @tprimak comment above about the f5_e3m0 format support is being dropped this idea is likely also best left unimplemented.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Removed f4_e3m0 from all brgemm matmul code.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: fixed

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Fixed with option 1

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: refactored

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: f4_e3m0 removed from brgemm matmul code

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: fixed

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: fixed

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: Given f4_e3m0 deprecation, all one_of(dt, f4_e2m1, f4_e3m0) checks are now just dt == f4_e2m1, so the helper is no longer needed.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Considered extracting the duplicated f4_e2m1_table to file scope. However, there are only 3 occurrences, and keeping them inline is consistent with how int4_permute tables are already defined next to their usage, making the context immediately visible. If we decide to refactor all LUT/permute tables to file scope, I'd prefer to do that as a separate cleanup PR covering all similar cases at once.

src/common/matmul.cpp
```diff
@@ -665,6 +665,16 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
         }
     }
 
+    // FP4 types store 2 elements per byte, so K must be even
+    const bool is_f4_weights = one_of(
+            weights_desc->data_type, data_type::f4_e2m1, data_type::f4_e3m0);
+    if (is_f4_weights) {
+        const int k_idx_wei = weights_desc->ndims - 2;
+        const dim_t K = weights_desc->dims[k_idx_wei];
+        VCHECK_MATMUL(!is_runtime_value(K) && K % 2 == 0, VERBOSE_BAD_DIM,
```
dzarukin: Why K and not N?

src/cpu/reorder/simple_reorder.hpp
```diff
@@ -2180,9 +2180,8 @@ struct simple_reorder_impl_t<SIMPLE_REORDER_TEMPL_CALL,
                 simple_attr_check(attr, false, true), VERBOSE_UNSUPPORTED_ATTR);
         VDISPATCH_REORDER_IC(
                 input_d.is_dense(), VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "src");
-        VDISPATCH_REORDER_IC(
-                IMPLICATION(!output_d.is_dense(), output_d.is_plain()),
-                VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "dst");
+        // Note: non-dense output (e.g. blocked with padding) is supported
+        // through the scratchpad-based need_transform path in execute().
```
dzarukin: Could you, please, expand what this comment suppose to mean? What is the "scratchpad-based need_transform path"? How does it relate to simple_reorder functionality?

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -14,14 +14,59 @@
...
+--bia_mask=4
+--dt=f32:f4_e2m1:f32,bf16:f4_e2m1:bf16,bf16:f4_e2m1:f32,f16:f4_e2m1:f16,f16:f4_e2m1:f32
+--attr-scales=wei:per_ocic:e8m0:32x1
+--batch=shapes_mx_3d
+
+# 4D shapes
+--bia_mask=8
+--dt=f32:f4_e2m1:f32,bf16:f4_e2m1:bf16,bf16:f4_e2m1:f32,f16:f4_e2m1:f16,f16:f4_e2m1:f32
+--attr-scales=wei:per_ocic:e8m0:32x1
+--batch=shapes_mx_4d
```
dzarukin: Just add one 3D shape into `shapes_mx` and revert rest changes, that should be enough.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1538,6 +1583,11 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     bgmmc.is_runtime_N = is_runtime_value(bgmmc.N);
     bgmmc.is_runtime_K = is_runtime_value(bgmmc.K);
 
+    // FP4 types store 2 elements per byte, so K must be even
+    const bool is_f4_weights = bgmmc.orig_wei_dt == data_type::f4_e2m1;
+    VCONDCHECK_BG(!is_f4_weights || bgmmc.is_runtime_K || bgmmc.K % 2 == 0,
+            VERBOSE_BAD_PARAM, "K");
```
dzarukin: Ditto.

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -36,6 +81,6 @@
 
 --reset
 --stag=abx --wtag=abx,any --dtag=abx
---dt=f4_e2m1:f4_e2m1:f32,f4_e2m1:f4_e2m1:bf16,f8_e5m2:f4_e2m1:bf16
+--dt=f4_e2m1:f4_e2m1:f32,f4_e2m1:f4_e2m1:bf16
 --attr-scales=src:per_tensor:f8_e5m2:1x128+wei:per_tensor:f8_e5m2:128x128,src:per_tensor:f16:32x32+wei:per_tensor:f8_e5m2:32x32
 --batch=shapes_2d_attr
```
georgen117, line 41: Just asking question not expected a code change.

When running the test suite all the tests run by the `## 2D Blocked Scales` section return unimplemented?

I know this was not part of this PR. Should these tests be part of the tests? Should they be commented out with an added TODO.  

I don't think this commit contains per_tensor scaling for f8_e5m2 src?

Not saying these need to be removed/changed just trying to figure out why tests that return UNIMPLEMENTED are part of the test suit.

georgen117 (approved):

src/common/matmul.cpp
amakarev: Good point. The check here is wrong — for plain layout (ab), packing is along N, so N must be even, not K. The existing INT4 handling in brgemm_matmul_reorders.cpp correctly checks N % 2 == 0. I'll fix this to check the innermost dimension instead, consistent with how INT4 validates it.

src/cpu/reorder/simple_reorder.hpp
amakarev: The need_transform path is in execute() of this same simple_reorder (line 2211). When output is non-dense (e.g. blocked with padding), it allocates a dense f32 scratchpad, zeroes it, reorders input into it using logical offsets, then compresses to 4-bit. This handles any output layout correctly.

The old check IMPLICATION(!output_d.is_dense(), output_d.is_plain()) was overly restrictive - it rejected non-dense blocked outputs that execute() already handles. With MXFP4 weights, brgemm matmul creates blocked formats (e.g. BA16a64b2a) that can be non-dense when dimensions don't align with block sizes, so the check had to go.

The comment itself probably is unnecessary - it explains why code was removed, which is a commit message concern, not a code comment. I'll remove it.

tests/benchdnn/inputs/matmul/test_matmul_fp4
amakarev: The shapes are split because bia_mask is positional: per-N bias requires bia_mask=2 for 2D, 4 for 3D, 8 for 4D. I can merge shapes back into one shapes_mx and keep the per-dimension sections in the test file — coverage stays the same, but each section would also run shapes of other dimensions that return UNIMPLEMENTED, adding noise to the output and wasting CI time. Is that acceptable? If so, I'll revert.

tests/benchdnn/inputs/matmul/test_matmul_fp4
amakarev: This section was already on main before this PR. These tests use f8_e5m2 scales which aren't supported yet. They serve as a placeholder: once f8_e5m2 scale support lands, these tests will start passing automatically without needing a separate test PR. I think it's better to leave them as is.

tests/benchdnn/inputs/matmul/test_matmul_fp4
georgen117: My personal opinion is to keep the shapes split so we don't have a lot of `UNIMPLEMENTED` tests making noise. I would prefer to have multiple test files over the noise.

dzarukin: I'm pretty sure all the bugs wouldn't be coming from the bias applicability given that its data type doesn't even change from f32 to anything else. Saying that, I'd drop bias and fuse stuff according to my initial proposal.

src/cpu/platform.cpp
```diff
@@ -143,7 +143,12 @@ bool has_data_type_support(data_type_t data_type) {
             return false;
 #endif
         case data_type::f4_e3m0:
```
dzarukin: This should return `false` to be correct.

src/common/matmul.cpp
```diff
@@ -665,6 +665,16 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
...
 
+    // FP4 types store 2 elements per byte, so N must be even
+    const bool is_f4_weights = one_of(
+            weights_desc->data_type, data_type::f4_e2m1, data_type::f4_e3m0);
+    if (is_f4_weights) {
+        const int n_idx_wei = weights_desc->ndims - 1;
+        const dim_t N = weights_desc->dims[n_idx_wei];
+        VCHECK_MATMUL(!is_runtime_value(N) && N % 2 == 0, VERBOSE_BAD_DIM,
+                "weights", n_idx_wei);
+    }
```
dzarukin: This looks incorrect to me either.

For `ab` format, N must be divisible by 2 while for `ba` format, K must be. The requirement is that the most dense dimension is divisible by 2, not one or the other. This includes batch dimensions if they are the innermost and present.

There must be something for int4 as it follows the same logic, I'd check it out and apply copy-exact approach.

src/cpu/reorder/simple_reorder.hpp
```diff
@@ -2180,9 +2180,6 @@ struct simple_reorder_impl_t<SIMPLE_REORDER_TEMPL_CALL,
                 simple_attr_check(attr, false, true), VERBOSE_UNSUPPORTED_ATTR);
         VDISPATCH_REORDER_IC(
                 input_d.is_dense(), VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "src");
-        VDISPATCH_REORDER_IC(
-                IMPLICATION(!output_d.is_dense(), output_d.is_plain()),
-                VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "dst");
```
dzarukin, line 2185: Would you mind to point where these cases are covered in benchdnn indicating this condition can be dropped?

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -211,17 +215,20 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
     auto check_attr_scales = [&]() -> bool {
         const std::vector<int> supported_args
                 = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST};
-        bool ok = attr_scales_ok(supported_args);
+        const std::vector<int> supported_qmodes = {
+                quantization_mode::static_sazp, quantization_mode::dynamic_mx};
```
dzarukin: `quantization_mode::dynamic_mx` means computing dst scales on-the-fly, which this PR doesn't provide. This one should be dropped until the support is actually added.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -1822,15 +1833,21 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
         return batch_ptr + A_strides_[1] * m + A_strides_[0] * k;
     }
 
+    int wei_packed_elems_per_byte() const {
+        return utils::one_of(bgmmc_.orig_wei_dt, data_type::s4, data_type::u4,
```
dzarukin: Nit: would look more compact if `using namespace data_type;` is used.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -205,8 +205,11 @@ struct brgemm_matmul_conf_t {
...
     bool is_bf32 = false;
     bool is_bf16_with_int_wei = false;
+    bool is_bf16_with_f4_wei = false;
     bool is_f16_with_int_wei = false;
+    bool is_f16_with_f4_wei = false;
     bool is_f32_with_int_wei = false;
+    bool is_f32_with_f4_wei = false;
     bool is_f32_f16 = false;
     bool is_f32_bf16 = false;
     bool is_int4_weights = false;
```
dzarukin, line 215: If this one was used only to indicate a subbyte state, please drop since a new method for such technology is introduced. If it indicates something more, would be nice to have it documented here.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2358,29 +2358,25 @@ struct jit_brgemm_matmul_copy_b_common_t : public jit_brgemm_matmul_copy_b_t,
...
+    Vmm maybe_mask_4bit(Vmm vmm, bool is_tail) {
         // Transposed kernel uses the same kTail mask for both cases
-        const auto tail_mask
-                = is_int4 && !conf_->transposed_B ? kTail_int4 : kTail;
+        const auto tail_mask = !conf_->transposed_B ? kTail_4bit : kTail;
+        return maybe_mask_impl(vmm, is_tail, tail_mask);
+    }
+
+    template <typename Vmm>
+    Vmm maybe_mask_impl(Vmm vmm, bool is_tail, Xbyak::Opmask tail_mask) {
```
dzarukin: Minor: make it private since it's `impl`?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3685,6 +3689,8 @@ struct jit_brgemm_matmul_copy_b_bf16_t
     const int typesize, tr_typesize, wei_scales_typesize;
     const dim_t src_stride, tr_src_stride;
     const bool is_src_int4;
+    const bool is_src_f4;
+    const bool is_src_4bit;
```
dzarukin, line 3691: It feels like indication if it's 4bit + data_type (which is two members) should be enough to reach the goal over 3 current members (and growing with other data types adding). Going further, just a data type member + a method that would answer the question if it's 4 bit or not would be even more compact.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -3729,6 +3735,7 @@ struct jit_brgemm_matmul_copy_b_bf16_t
     Vmm vmm_zp_b_shift = Vmm(2);
     Vmm vmm_permd = Vmm(3);
     Vmm vmm_wei_scales = Vmm(4);
+    Vmm vmm_f4_lut = Vmm(5);
```
dzarukin: Nit: what is `lut`?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -5566,17 +5587,21 @@ struct jit_brgemm_matmul_copy_b_cvt_bf16_t
...
         , req_zp_b_shift_(
                   conf_->has_zero_point_b && conf_->with_wei_decompression)
         , req_apply_wei_scales_(conf_->apply_scales_in_buffer_b)
-        , reserved_regs_(req_apply_wei_scales_ ? 6
-                          : req_zp_b_shift_    ? 4
-                          : is_src_int4_       ? 1
-                                               : 0)
+        , reserved_regs_(is_src_f4_ && req_apply_wei_scales_ ? 7
+                          : req_apply_wei_scales_            ? 6
+                          : is_src_f4_                       ? 1
```
dzarukin: I might be wrong but setting `1` before `4` would definitely lead to a bug for some combination which is hard to debug.
This spot is clumsy and already requires better handling such as dynamic identification of features and reserving the number of register, static approach doesn't work any longer.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -313,18 +320,27 @@ brgemm_matmul_conf_utils_t::brgemm_matmul_conf_utils_t(
...
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::f16,
                       bgmmc.src_dt == f16)
               && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::bf16,
                       bgmmc.src_dt == bf16)
-              && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict,
+              // f4_e2m1 can be losslessly upconverted to bf16,
+              // so strict mode should not restrict src to f32.
+              && IMPLICATION(attr.fpmath_.mode_ == fpmath_mode::strict
+                              && bgmmc.wei_dt != data_type::f4_e2m1,
                       bgmmc.src_dt == f32)
```
dzarukin: Let's do one more iteration here:
Updated version states that if fpmath_mode == strict and wei_dt == f4_e2m1, then src_dt doesn't have any requirements.

Let's say, src_dt is bf16, wei_dt is f4_e2m1 and fpmath-mode is strict, what would be weights upconverted to - f32 or bf16, and why?

And even if you find an answer on this question, the proposal is to do more explicit coding (extrapolate to f16 dt if needed):
```
, bf16_with_f4_wei_dt(bgmmc.src_dt == bf16 && one_of(bgmmc.dst_dt, bf16, f32) && bgmmc.wei_dt == data_type::f4_e2m1)
, weights_decompression_support(bf16_with_f4_wei_dt || former_condition_for_integers)
```

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1404,19 +1441,27 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
     // happens during copy-buffer computations
-    if (bgmmc.is_bf32 || bgmmc.is_bf16_with_int_wei) {
+    if (bgmmc.is_bf32 || bgmmc.is_bf16_with_int_wei
+            || bgmmc.is_bf16_with_f4_wei) {
         bgmmc.src_dt = bf16;
         bgmmc.wei_dt = bf16;
         bgmmc.tr_a_dt_sz = types::data_type_size(bf16);
         bgmmc.tr_b_dt_sz = types::data_type_size(bf16);
-    } else if ((bm_conf_utils.is_f16() || bgmmc.is_f16_with_int_wei)
+    } else if (bgmmc.is_f16_with_f4_wei && bgmmc.is_amx) {
```
dzarukin: How is this different from line 1480?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1538,6 +1583,11 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     bgmmc.is_runtime_N = is_runtime_value(bgmmc.N);
     bgmmc.is_runtime_K = is_runtime_value(bgmmc.K);
 
+    // FP4 types store 2 elements per byte, so N must be even
+    const bool is_f4_weights = bgmmc.orig_wei_dt == data_type::f4_e2m1;
+    VCONDCHECK_BG(!is_f4_weights || bgmmc.is_runtime_N || bgmmc.N % 2 == 0,
+            VERBOSE_BAD_PARAM, "N");
```
dzarukin: Ditto. BTW, this shouldn't repeat API constraints but actually put unimplemented parts of allowed API.

dzarukin (commented):
Additional question - have you checked if ukernel API works with mxfp4 changes introduced? Is there coverage for it?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: LUT = lookup table. It holds the 16 f4_e2m1→f32 mapping values used by vpermps to convert 4-bit indices to floats.

dzarukin: Just `table` feels more natural and descriptive, but I'll leave it up to you.

tests/benchdnn/inputs/matmul/test_matmul_fp4
amakarev: done

amakarev:
> Additional question - have you checked if ukernel API works with mxfp4 changes introduced? Is there coverage for it?

No ukernel API coverage for mxfp4. FP4 dequantization happens in copy_B kernels at the matmul primitive level, before data reaches brgemm.

src/cpu/platform.cpp
amakarev: done

dzarukin:
> > Additional question - have you checked if ukernel API works with mxfp4 changes introduced? Is there coverage for it?
> 
> No ukernel API coverage for mxfp4. FP4 dequantization happens in copy_B kernels at the matmul primitive level, before data reaches brgemm.

ukernel provides transform routine which takes input and output data types. I'd expect fp4 in and xf16/f32 out should work or explained why not.

src/common/matmul.cpp
amakarev: done [Commit](https://github.com/uxlfoundation/oneDNN/pull/4922/changes/7f20b560b312c12a06fcae8a9868d98103f68f1f)

src/cpu/reorder/simple_reorder.hpp
amakarev: Restored the check. Blocked layouts are now disabled for f4 weights in matmul (same as int4), so there's no use case for non-dense blocked f4 output in this PR.

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: Dropped dynamic_mx - reverted to the default attr_scales_ok(supported_args) which only allows static_sazp, same as before this PR.

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: Done. using namespace data_type; is already at file scope, removed the redundant qualifiers.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
amakarev: is_int4_weights was not introduced by this PR - it existed before. Could you clarify if you're referring to is_int4_weights itself, or to one of the new is_*_with_f4_wei fields? Also, what is the "new method" you're referring to?

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: done

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Renamed vmm_f4_lut --> vmm_f4_lookup_table as a compromise - keeps the distinction from permute tables while being more descriptive.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Dropped is_src_int4 and is_src_f4 members, kept is_src_4bit. Specific type checks now use conf_->orig_wei_dt directly.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
amakarev: Refactored: bf16_with_f4_wei_dt, f16_with_f4_wei_dt, f32_with_f4_wei_dt are now standalone (no fpmath_mode dependency). weights_decompression_support is defined as their union with the int weights condition.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: Replaced the ternary chain with IIFE + nstl::max - each feature independently declares its highest register index, order doesn't matter

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
amakarev: Both branches set the same values. Merged is_f16_with_f4_wei && is_amx into the existing is_f16_with_int_wei branch at line 1482.

amakarev:
> > > Additional question - have you checked if ukernel API works with mxfp4 changes introduced? Is there coverage for it?
> > 
> > 
> > No ukernel API coverage for mxfp4. FP4 dequantization happens in copy_B kernels at the matmul primitive level, before data reaches brgemm.
> 
> ukernel provides transform routine which takes input and output data types. I'd expect fp4 in and xf16/f32 out should work or explained why not.

should work with [this commit](https://github.com/uxlfoundation/oneDNN/pull/4922/changes/bdb9184fb84c124d9b6eb163d2f6917f84cabe37)

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -219,9 +223,10 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             // This case requires scratchpad
             if (is_runtime_value(N())) ok = false;
         }
-        // Impl suppports f32 scales only for non-weight decompression
-        if (!(is_bf16_with_int_wei || is_f16_with_int_wei
-                    || is_f32_with_int_wei)) {
+        // Impl supports f32 scales only for non-weight decompression
+        if (!(is_bf16_with_int_wei || is_bf16_with_f4_wei || is_f16_with_int_wei
+                    || is_f16_with_f4_wei || is_f32_with_int_wei
+                    || is_f32_with_f4_wei)) {
```
dzarukin, line 229: Minor: I've checked through the code and it looks like the only external var can be `is_xf_with_f4_wei` which would combine all three cases. Additional clarity can be added through wei_dt in copy kernels, but that's just couple spots.

This is not a blocking feedback to address but should definitely help for future development and readability.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -317,8 +323,10 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             VERBOSE_UNSUPPORTED_POSTOP);
 
     VDISPATCH_MATMUL(check_attr_scales(), VERBOSE_UNSUPPORTED_SCALES_CFG);
-    VDISPATCH_MATMUL(check_attr_zero_points(is_bf16_with_int_wei
-                             || is_f16_with_int_wei || is_f32_with_int_wei),
+    VDISPATCH_MATMUL(
+            check_attr_zero_points(is_bf16_with_int_wei || is_bf16_with_f4_wei
+                    || is_f16_with_int_wei || is_f16_with_f4_wei
+                    || is_f32_with_int_wei || is_f32_with_f4_wei),
```
dzarukin: There are no zero-points for f4 "quantization", please revert.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -5566,17 +5606,23 @@ struct jit_brgemm_matmul_copy_b_cvt_bf16_t
...
-                          : is_src_int4_       ? 1
-                                               : 0)
+        , reserved_regs_([&]() {
+            int n = 0;
+            if (is_src_4bit_) n = nstl::max(n, 1); // vmm_permd(0)
+            if (req_zp_b_shift_) n = nstl::max(n, 4); // + zp/tmp(1..3)
+            if (req_apply_wei_scales_) n = nstl::max(n, 6); // + scales(4..5)
+            if (is_src_f4_) n = nstl::max(n, 7); // + f4_lookup_table(6)
+            return n;
+        }())
```
dzarukin, line 5639: This one looks like a more complex version of previous assignment where `is_src_f4_` would ask for 7 reserved registers. This is not quite what I meant.
These feature may or may not intersect, e.g. f4 src would have scales but no zp, thus could use 1-3 indices for their needs, instead, those be empty and won't be used. I'd think about an algorithm that would check for a feature and it's present, assign a reserved register index(-ices) for it and the code to get those indices through a function.
Since it's a bigger rework, this can be documented in a separate JIRA as a technical debt and addressed later.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
dzarukin: I'm referring to `wei_packed_elems_per_byte()` which basically substituted the logic that `is_int4_weights` was covering - defining if division by 2 should happen or not.

Though it existed before, it seems it's redundant with `wei_packed_elems_per_byte()` introduced, so I don't see a compelling reason to keep it alive, unless I'm missing it's additional usage which can't substituted in any other way.

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -23,6 +24,12 @@
 --attr-scales=src:per_tensor:e8m0:1x32+wei:per_tensor:e8m0:32x1+dst:mx:e8m0:1x32
 --batch=shapes_mx
 
+## MXFP4 weights
+
+--dt=f32:f4_e2m1:f32,bf16:f4_e2m1:bf16,bf16:f4_e2m1:f32,f16:f4_e2m1:f16,f16:f4_e2m1:f32
+--attr-scales=wei:per_ocic:e8m0:32x1
+--batch=shapes_mx
```
dzarukin: Please move this to `harness_matmul_decompression`, it has a dedicated section for fp4 weights-only-quantization.

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: I looked into this but the individual flags are still needed in backup_isa selection (line 360) where f16 and bf16 require different ISA fallbacks, so I'd prefer to keep as-is.

src/cpu/x64/matmul/brgemm_matmul.cpp
amakarev: done

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
amakarev: The previous ternary chain was order-dependent - a developer adding a new feature had to carefully determine the correct insertion point, otherwise register requirements for some combinations could be missed. The max-based approach makes the order irrelevant: each feature independently declares its requirement, so it's not possible to accidentally shadow another feature's registers. From my perspective this doesn't add complexity - it prevents easy-to-make mistakes.

I agree that a fully dynamic index allocation (where each feature claims only the slots it actually needs rather than reserving a fixed range) would be a better long-term solution. I'll estimate the effort and see if it can be done within this PR or create a separate JIRA for it.

amakarev: @dzarukin  Does this address your concern? [commit](https://github.com/uxlfoundation/oneDNN/pull/4922/commits/8fd8833f11fbdf5c84ce2dcb2e32686e4c84c61e)

tests/benchdnn/inputs/matmul/test_matmul_fp4
amakarev: done

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
amakarev: wei_packed_elems_per_byte() is a private method of brgemm_matmul_t and can't be accessed from brgemm_matmul_utils.cpp where the config initialization happens. To make it available in both places, I moved it to a field in brgemm_matmul_conf_t, replacing both is_int4_weights and is_f4_weights.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
dzarukin: This looks better, thank you.

tests/benchdnn/inputs/matmul/test_matmul_fp4
```diff
@@ -14,6 +14,7 @@
 
 
 ## MXFP4
+
```
dzarukin, line 17: Please drop all the changes from this file as it was supposed to work without them.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -262,6 +262,14 @@
 --attr-fpmath=f16:true
 --attr-scales=wei:common:2,wei:per_oc:f16,wei:per_ocic:f16:128x1
 1x4096:4096x4096
+
+# mxfp4 weights decompression
+--reset
+--skip-impl=ref
+--dt=f32:f4_e2m1:f32,bf16:f4_e2m1:bf16,bf16:f4_e2m1:f32,f16:f4_e2m1:f16,f16:f4_e2m1:f32
+--attr-scales=wei:per_ocic:e8m0:32x1
```
dzarukin, line 270: I doubt this scale data type is correct. e8m0 is designed for pure MXFP8 configuration which is not it.
Looking on what's available above, I'd say you need to leave f32:f4_e2m1:f32 data type only with f32 scale, existing cases should provide the rest of the coverage.

src/cpu/platform.cpp
```diff
@@ -142,8 +142,13 @@ bool has_data_type_support(data_type_t data_type) {
...
 #endif
-        case data_type::f4_e3m0:
-        case data_type::f4_e2m1: return false;
+        case data_type::f4_e3m0: return false;
+        case data_type::f4_e2m1:
+#if DNNL_X64
+            return x64::mayiuse(x64::avx512_core);
+#else
+            return false;
+#endif
```
dzarukin, line 151: I anticipate this change is needed only for ref_matmul implementation but I don't see any changes to it. At least I'd expect f4_e2m1 data type as allowed for weights. Did you verify that ref_matmul can dispatch to new configurations? If not, please add correspondent changes. If yes, please provide a verbose line with a benchdnn case dispatching into it.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1390,12 +1420,18 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    bgmmc.is_bf16_with_f4_wei = bm_conf_utils.is_bf16_with_f4_wei();
     bgmmc.is_f16_with_int_wei = bm_conf_utils.is_f16_with_int_wei();
+    bgmmc.is_f16_with_f4_wei = bm_conf_utils.is_f16_with_f4_wei();
     bgmmc.is_f32_with_int_wei = bm_conf_utils.is_f32_with_int_wei();
+    bgmmc.is_f32_with_f4_wei = bm_conf_utils.is_f32_with_f4_wei();
     bgmmc.is_f32_f16 = bm_conf_utils.is_f32_f16();
     bgmmc.is_f32_bf16 = bm_conf_utils.is_f32_bf16();
     bgmmc.with_wei_decompression = bm_conf_utils.with_weights_decompression();
-    bgmmc.is_int4_weights = one_of(bgmmc.wei_dt, data_type::s4, data_type::u4);
+    bgmmc.wei_packed_elems_per_byte = one_of(bgmmc.orig_wei_dt, data_type::s4,
```
dzarukin, line 1431: Please leave a comment (aligned on 80):
```
    // The usage of this variable hardcodes the assumption there's only supported value > 1, which is 2.
    // If this assumption changes, the update to all other places is required, too.
```

tests/benchdnn/inputs/matmul/harness_matmul_decompression
tprimak: > I doubt this scale data type is correct. e8m0 is designed for pure MXFP8 configuration which is not it. Looking on what's available above, I'd say you need to leave f32:f4_e2m1:f32 data type only with f32 scale, existing cases should provide the rest of the coverage.

@dzarukin all MX datatypes use e8m0 scale format as per OCP specification. https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf

e8m0 scale support for MXFP4 even explicitly mentioned in our documentation:
Scale data type includes: f32, bf16, f16, e8m0 for MXFP8 and MXFP4, f8_e4m3 for NVFP4 block scales.
https://uxlfoundation.github.io/oneDNN/dev_guide_matmul.html#implementation-notes

tests/benchdnn/inputs/matmul/test_matmul_fp4
amakarev: done

src/cpu/platform.cpp
amakarev: Added f4_e2m1 to the allowed weight types in ref_matmul. Verified that ref_matmul dispatches and passes for f32:f4_e2m1:f32 configuration.

Repro:
`DNNL_VERBOSE=dispatch,all benchdnn --matmul --dt=f32:f4_e2m1:f32 --stag=abx --wtag=abx --dtag=abx --skip-impl=brg 2x64:64x2
`

`onednn_verbose,v1,primitive,create:cache_miss,cpu,matmul,ref:any,undef,src:f32::blocked:ab::f0 wei:f4_e2m1::blocked:ab::f0 dst:f32::blocked:ab::f0,,,2x64:64x2
`

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
amakarev: done

amakarev:
Closing this PR — it grew beyond a reviewable scope while iterating on all three data types (f32, bf16, f16) at once, and the comment thread became hard to navigate.

Splitting the work into a series of smaller PRs, one data type at a time:

f32:mxfp4 — opened as #5645 .
bf16:mxfp4 — will follow once f32 lands.
f16:mxfp4 — after bf16.

PR #4924, opened 2026-03-30
cpu: x64: fix vcvt_f8_to_f16 for is_fp8_native path

georgen117:
[MFDNN-14831](https://jira.devtools.intel.com/browse/MFDNN-14831)

### Problem

After updating Xbyak from v7.28 to v7.35.3, the `f8_e4m3` pooling test fails with a `runtime_error` during `create_primitive` when running under SDE with `DNNL_MAX_CPU_ISA=AVX10_2_512_AMX_2`: `--pool --dt=f8_e4m3:f8_e4m3 ic64iw32ow16kw3sw2pw0`

Xbyak v7.35.3 introduced `checkCvt1`, a new operand-size validation for conversion instructions. `vcvthf82ph` requires the source register to be half the width of the destination (e.g. `(xmm/ymm dst, xmm src)` or `(zmm dst, ymm src)`).

| dst | src | valid |
|----|-----|------|
| XMM/YMM | XMM | yes |
| ZMM | YMM | yes |
| XMM/YMM | YMM | no |
| ZMM | ZMM | no |
| any | MEM | yes |

 Callers of `fp8_conversionion_e4m3::vcvt_f8_to_f16` and `vcv_f8_to_f32` were passing same-width registers (e.g. YMM src to YMM dst),  which was silently accepted before v7.35.3 but now throws `ERR_BAD_COMBINATION`, propagating as a `runtime_error`.

### Fix

Updated `fp8_conversion_e4m3::vcvt_f8_to_f16` to adjust register sizes to conform with the `checkCvt1` requirements. `cvct_f8_to_f16` will have the same behavior as it had before the Xbyak library update. It will silently accept registers of the same width and different width. 

`vcvt_f8_to_f32` is simplified to delegate to `vcvt_f8_to_f16`, removing a previously duplicated native-path inline that was the original source of the bug.

Note: `f8_e5m2` conversions use `vpmovzxbw` internally, which does not have the `checkCvt1` size constraint and required no changes.

### Testing

```bash
DNNL_MAX_CPU_ISA=AVX10_2_512_AMX_2 \
  sde -chip_check_exe_only -dmr -- \
  benchdnn --pool --dt=f8_e4m3:f8_e4m3 ic64iw32ow16kw3sw2pw0 #from nightly build

DNNL_MAX_CPU_ISA=AVX10_2_512_AMX_2 \
  sde -chip_check_exe_only -dmr -- \
  benchdnn --reorder --sdt=f8_e4m3 --ddt=f32,bf16,f16,f8_e5m2 8x8 64x64 128x128 #based on code changed
```
**PASSES**

### DIFF
--- a/src/cpu/x64/jit_avx512_core_fp8cvt.cpp
+++ b/src/cpu/x64/jit_avx512_core_fp8cvt.cpp
@@ -260,7 +260,15 @@ void fp8_conversion_e4m3_t::vcvt_f8_to_f16(
             true, op_in.isXMM(), op_in.isYMM(), op_in.isZMM(), op_in.isMEM()));
 
     if (is_fp8_native()) {
-        host_->vcvthf82ph(xmm_out, op_in);
+        if (!op_in.isMEM()) {
+            // src register must be at least half the size of dst register (xbyak checkCvt1)
+            if (xmm_out.isZMM())
+                host_->vcvthf82ph(xmm_out, Xbyak::Ymm(op_in.getIdx()));
+            else
+                host_->vcvthf82ph(xmm_out, Xbyak::Xmm(op_in.getIdx()));
+        } else {
+            host_->vcvthf82ph(xmm_out, op_in);
+        }
         return;
     }
 
@@ -355,11 +363,7 @@ void fp8_conversion_e4m3_t::vcvt_f8_to_f32(
     const Xbyak::Zmm zmm_out(xmm_out.getIdx());
 
     // f16 <- f8_e4m3
-    if (is_fp8_native())
-        host_->vcvthf82ph(ymm_mask(xmm_out), op_in);
-    else
-        vcvt_f8_to_f16(ymm_mask(xmm_out), op_in);
-
+    vcvt_f8_to_f16(ymm_mask(xmm_out), op_in);
     // f32 <- f16
     host_->vcvtph2psx(zmm_out, ymm_out);
 }


georgen117:
make test
set test_scope=NIGHTLY
disable os_win
disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

amakarev:
Suggest adding assert after line 260 to catch ZMM→XMM truncation:

assert(IMPLICATION(!op_in.isMEM() && op_in.isZMM(), xmm_out.isZMM())
        && "ZMM src requires ZMM dst to avoid data loss");

georgen117:
> Suggest adding assert after line 260 to catch ZMM→XMM truncation:
> 
> assert(IMPLICATION(!op_in.isMEM() && op_in.isZMM(), xmm_out.isZMM()) && "ZMM src requires ZMM dst to avoid data loss");

@amakarev  Did you mean `assert(IMPLICATION(!op_in.isMEM() && op_in.isZMM(), xmm_out.isXMM()) && "ZMM src requires ZMM dst to avoid data loss");`

Although I think anyone calling `vcvt_f8_to_f16(xmm_reg_out, zmm_reg_in)` is likely calling it with incorrect inputs. it can still be valid. As long as they are only trying to convert 8 numbers (the lower 64 bits).

I feel like getting the assert correct for all the combinations of registers would be difficult to get right and the right combination is the existing `checkCvt1` from Xbyak.
```
if (!op.isMEM() && !(x.is(Operand::XMM | Operand::YMM) && op.isXMM()) && !(x.isZMM() && op.isYMM())) XBYAK_THROW(ERR_BAD_COMBINATION)
```
If we want to go that route the best solution would be to make sure everyone calling `vcvt_f8_to_f16` adheres to the checkCvt1.

If that is what we want is to catch register size mismatches there are only about 8-20 places in the code that I would need to update. Would you like me to submit a patch showing the change?

georgen117:
@amakarev this change basically restores the behavior before the xbyak v7.35.3 update. Calling `vcvt_f8_to_f16` will silently accept any combination of registers. Silently ignoring the upper bits of the source register if the conversion does not fit in destination register. This was the behavior before xbyak added the register size check.

If we want to remove the silent conversion, then I propose we use the checkCvt1 logic. This would involve updating every location that is calling `vcvt_f8_to_f16`. Around 8-20 locations in the code base.

Two solutions:
1. silently accept different register sizes for f8 -> f16 casting (this PR)
2. Be explicit about registers don't silently accept register sizes that can't be converted without ignoring the upper bits. (this is what upstream xbyak checks) changing all places that call `vcvt_f8_to_f16` so they explicitly convert the registers.

I am happy with using either solution. I chose the solution in this PR (solution 1) as it was the smallest code change.

amakarev:
@georgen117 I agree my assert was naive and didn't cover all valid partial conversion cases. Between the two solutions, I'd prefer solution 2 (explicit checkCvt1 everywhere) 
It gives us **type safety** by catching incorrect usage at compile/runtime instead of silent truncation. The code becomes more **self-documenting** - you immediately see whether it's a partial or full conversion. And it follows the upstream xbyak design philosophy- I think they added checkCvt1 for a reason. And 8-20 fixes is a reasonable one-time cost for long-term correctness.

georgen117:
make test
set test_scope=NIGHTLY
disable os_win
disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

georgen117:
make test
set test_scope=NIGHTLY
disable compiler_gnu9
disable compiler_icx-previous
disable compiler_icx-oss
disable compiler_vs2022
disable build_vendor_amd
disable build_vendor_nvidia
disable build_graph
disable build_cpu_runtime_tbb
disable build_gpu_runtime_ocl
disable build_mode_no_cpu
disable test_device_gpu

georgen117:
> @georgen117 I agree my assert was naive and didn't cover all valid partial conversion cases. Between the two solutions, I'd prefer solution 2 (explicit checkCvt1 everywhere) It gives us **type safety** by catching incorrect usage at compile/runtime instead of silent truncation. The code becomes more **self-documenting** - you immediately see whether it's a partial or full conversion. And it follows the upstream xbyak design philosophy- I think they added checkCvt1 for a reason. And 8-20 fixes is a reasonable one-time cost for long-term correctness.

@amakarev Done. Thanks for the feedback.

georgen117:
@ankalinin  and @tprimak  based on review feedback the code has changed considerably. It no longer silently converts the inputs registers used by vcvt_f8_tof16 to match the required register size. Instead it forces the user to adhere to the checkCvt1 that was added to Xbyak.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -701,7 +701,12 @@ void jit_brgemm_kernel_t<Wmm>::cvt2ps(data_type_t type_in, const Vmm vmm_in,
         case data_type::s8: uni_vpmovsxbd(vmm, op); break;
         case data_type::u8: uni_vpmovzxbd(vmm, op); break;
         case data_type::f8_e5m2: f8_e5m2_cvt_->vcvt_f8_to_f32(vmm, op); break;
-        case data_type::f8_e4m3: f8_e4m3_cvt_->vcvt_f8_to_f32(vmm, op); break;
+        case data_type::f8_e4m3:
+            if (!op.isMEM())
+                f8_e4m3_cvt_->vcvt_f8_to_f32(vmm, Xmm(op.getIdx()));
+            else
+                f8_e4m3_cvt_->vcvt_f8_to_f32(vmm, op);
```
dzarukin: Have you considered to move this logic inside the routine?
It feels not wrong to pass an `Operand` inside this call and let it do the right thing instead of making devs doing it.

amakarev: I think the explicit approach here is better, even though it creates some 
duplication. The encapsulation would be cleaner and the assert would guide 
usage, but in low-level code like this, functions are rarely written and 
frequently read.

Having the explicit `if/else` with `Xmm()` makes the constraint immediately 
visible:

```cpp
case f8_e5m2: 
    f8_e5m2_cvt_->vcvt_f8_to_f32(vmm, op); 
    break;
case f8_e4m3:
    if (!op.isMEM())
        f8_e4m3_cvt_->vcvt_f8_to_f32(vmm, Xmm(op.getIdx()));
    else
        f8_e4m3_cvt_->vcvt_f8_to_f32(vmm, op);
    break;
```
When reviewing or debugging, you can see right away that f8_e4m3 has special
register handling

Yes, it's ~15 lines of duplication across call sites, but for low-level JIT
code where hardware constraints are critical, I think the readability trade-off
is worth it. The code documents the vcvthf82ph limitation in place.

Just my opinion though - both approaches are valid.

georgen117: @dzarukin 

> Have you considered to move this logic inside the routine? It feels not wrong to pass an `Operand` inside this call and let it do the right thing instead of making devs doing it.

I had this conversation with @amakarev 

In my opinion there were two ways to handle the Xbyak::error that is a result of the checkCvt1 check added to Xybak

1. silently accept different register sizes for f8 -> f16 handling the register conversion inside the `vcvt_f8_to_f16` function [how I originally handled it](https://github.com/uxlfoundation/oneDNN/commit/f7b3f3dcf5b12bb348255a4932cd308d4c7f27ab)
2. Be explicit about registers.  Don't silently accept register sizes that can't be converted without ignoring the upper bits. (this is what upstream xbyak checks) (the current state of the PR)

option 1 is a much smaller change and behaves like the code before the `checkCvt1` was added. However, it silently accepts register combinations that could indicate the function is not being used correctly. 

For example calling: `cvct_f8_to_f16(ymm, zmm);` option 1 would silently change the zmm -> xmm when likely the actual intent was `cvct_f8_to_f16(zmm, ymm);`.  

Option1 makes the common pattern of expanding into the same register much simpler: `cvct_f8_to_f16(zmm, zmm);`

while option 2 requires `cvct_f8_to_f16(zmm, YMM(zmm.getIdx());` the second option is explicit about register size.

Personally, I am happy with both options.

option 2, Being explicit about register/mem inputs can catch some tricky errors at JIT time.  **Huge advantage.**
option 1, is easer to use but if you get the registers wrong it will silently change them and the error would have to caught at runtime through correctness testing instead.

@amakarev seemed to think catching the error at jit is worth the repeated code. Catching errors early also seems to be the philosophy Xbyak maintainer is using.

I feel like a little extra work to protect us from using the isa incorrectly is worth repeating the pattern in a few places.

dzarukin: I see a bunch of branches and not safe coding approach here.
If someone not familiar with this practice is going to call the current conversion routine, they'll get an error an will have to spend more time to figure out what's going on, study existing approaches (as there's no clear function description what's expected and what's not) and the very same question would pop up - why did the author handled that through explicit code over adding support inside the routine (I also failed to find any explanatory notes in the code)?

I don't see a compelling answer why it should be checked externally over making things cleaner internally and setting asserts if needed. After all, it's about using a proper instruction with proper arguments. If it's safe to make a down conversion of an `op` into a narrow register every time, I don't see a problem with doing it. If it's not safe, then the current routine must be updated to accept safe arguments only and not a generic Operand.

ankalinin: IMHO: I vote to have this logic inside the routine.

PR #4925, opened 2026-03-31
[wip, do not review]

mzhukova:
- New optional parameter for grouped memory: `max_variable_dim`, that allows to provide an upper bound on max number of tokens per expert
  - In general, tokens are routed randomly and this knowledge is kept on GPU, therefore in the current GPU kernel we end up with number of work groups launched `== total tokens assigned (M dim) * N * number of groups`, which is significantly larger than actual number of work groups needed to compute grouped gemm operation
  - The proposed change aims to reduce the number of work groups launched, by using upper bound on individual M instead of `total tokens assigned`:
    - Since MoE uses topK strategy (each token is routed to K experts), we could always ask framework to provide `total assigned tokens / topK` as upper bound, since this is the maximum what an expert could get
    - Additionally, if framework is doing any sort of balancing / limiting tokens that an expert could receive / any other sort of extra work, where they could spend time finding M max, this could also serve as an upper bound

[Jira with performance data points](https://jira.devtools.intel.com/browse/MFDNN-14467)

### DIFF
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -558,7 +558,16 @@ std::string md2fmt_str(
         case format_kind::wino:
         case format_kind::rnn_packed:
         case format_kind::opaque: ss << "::"; break;
-        case format_kind::sparse: ss << ":" << mdw.encoding() << ":"; break;
+        case format_kind::sparse: ss << ":" << mdw.encoding();
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+            if (mdw.is_grouped_desc()) {
+                const auto &gd = mdw.sparse_desc().grouped_desc;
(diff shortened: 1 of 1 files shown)


src/common/matmul.cpp
```diff
@@ -90,6 +90,12 @@ status_t grouped_matmul_desc_init(matmul_desc_t *matmul_desc,
             VERBOSE_INCONSISTENT_DIM, "dst_variable_dim_idx", 0,
             "src_variable_dim_idx", 0);
 
+    VCHECK_MATMUL_UNIMPL(
```
umar456, line 93: We can probably remove this check. I don't think this is necessary to be enforced by the primitive.

PR #4929, opened 2026-03-31
xe: ggemm: Add support for grouped quantization for src. Add f4 and f8 support

umar456:
# Description

This PR adds support for grouped quantization for the src tensor in Grouped GEMM. It also adds the f4 and f8 type support for microkernels and Grouped GEMM weights and src inputs. Scale quantization attribuites can also accept these new floating point data types. 

Partially addresses: [MFDNN-14187](https://jira.devtools.intel.com/browse/MFDNN-14187)

### DIFF
--- a/src/gpu/intel/include/tile_ops.h
+++ b/src/gpu/intel/include/tile_ops.h
@@ -676,7 +676,7 @@ DEF_BLOCK2D_LOAD_STORE(float, uint, 8, 16, u32_m8k16v1, 16, 8)
         _Pragma("unroll") for (int jj = 0; jj < nbc; jj++, ptr += ld * bc) { \
             _Pragma("unroll") for (int ii = 0; ii < nbr; ii++)(t) \
                     ->x[ii + nbr * jj] \
-                    = block_load(ptr + ii * br, br / SUBGROUP_SIZE); \
+                    = block_load(ptr + ii * br, br / sg); \
         } \
     } \
     __attribute__((overloadable)) void tile_store_block(tile_type t, \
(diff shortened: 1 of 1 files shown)


umar456:
make test
disable benchdnn_all
enable benchdnn_graph
enable benchdnn_matmul
disable test_device_cpu
enable test_device_gpu
enable arch_gpu_xe-hpc
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/gemm/jit/generator/microkernel/shim.cpp
```diff
@@ -86,7 +86,12 @@ int typeSize(StructuredType::Type type) {
         case StructuredType::f16:
         case StructuredType::bf16: return 2;
         case StructuredType::s8:
-        case StructuredType::u8: return 1;
+        case StructuredType::u8:
+        case StructuredType::bf8:
+        case StructuredType::hf8:
+        case StructuredType::f8_e8m0:
+        case StructuredType::f4_e2m1:
+        case StructuredType::f4_e3m0: return 1;
```
kealan-barbieri, line 94: how are you handling half-byte offsets here? is it assumed that fp4s are strided by 1 so they can be treated as bytes or is it being handled internally by Gemmstone?

umar456: This function only defines the sizes of the individual arguments that are passed into the shim function by value and not the size of memory. If we were to pass any f8 or f4 values they would be as char types cast to those values. It is unlikely we will use it. I can remove it if you prefer.

kealan-barbieri: I see it makes sense in the context of OCL kernel arguments. For that purpose it seems correct, I know we have handling in OCL for half-byte types with byte pointers.

umar456:
make test perf-gpu
set primitive=sdpa matmul

PR #4931, opened 2026-03-31
rfcs: proposal for optional upper bound arg for grouped matmul

mzhukova:
[design document](https://github.com/uxlfoundation/oneDNN/tree/rfcs/rfcs/20260331-grouped-matmul-max-variable-dim#readme)

[POC for recommended option](https://github.com/uxlfoundation/oneDNN/pull/4946)
[POC for option B](https://github.com/uxlfoundation/oneDNN/pull/4925), you could see perf data attached there via Jira link (even though the approach is different, the data is valid as it mainly evaluates performance with tighter upper bound)

### DIFF
--- a/rfcs/20260331-grouped-matmul-max-variable-dim/README.md
+++ b/rfcs/20260331-grouped-matmul-max-variable-dim/README.md
@@ -1,0 +1,120 @@
+# Add DNNL_ARG_HINT_MAX_GROUP_M for Grouped GEMM
+
+## Introduction
+
+> [!NOTE]
+> Grouped GEMM support is currently marked as experimental
+> (available under `ONEDNN_EXPERIMENTAL_GROUPED_MEMORY`), so API/ABI stability is not guaranteed yet,
+> and the API may evolve based on feedback.
+
+The grouped GEMM feature (see [previous RFC](https://github.com/uxlfoundation/oneDNN/tree/rfcs/rfcs/20251203-grouped-gemm-support))
+enables Mixture-of-Experts (MoE) support by representing variable-length groups of tokens
+as a single memory descriptor, where tokens are concatenated contiguously and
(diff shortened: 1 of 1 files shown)


mzhukova:
FYI @umar456

rfcs/20260331-grouped-matmul-max-variable-dim/README.md
```diff
@@ -0,0 +1,120 @@
...
+> and the API may evolve based on feedback.
+
+The grouped GEMM feature (see [previous RFC](https://github.com/uxlfoundation/oneDNN/tree/rfcs/rfcs/20251203-grouped-gemm-support))
+enables Mixture-of-Experts (MoE) support by representing variable-length groups of tokens
+as a single memory descriptor, where tokens are concatenated contiguously and
+per-group boundaries are communicated via a **device-side** offsets buffer.
+
+For a grouped matmul with `E` experts and `total_M` tokens (the sum of all per-expert M values),
+the GPU kernel grid is currently: `gws ~ N * total_M * E`.
+This generates a much higher number of workers than necessary, since each expert need to process
```
densamoilov, line 17: Is there a way to manage it without relying on a hint or are we likely to continue using the hint going forward?

mzhukova: AFAIK eventually there is a plan to try out different strategies for launching workers, however all of it would come with a price for traversing the offsets array in one way or another, so we would need to carefully analyze where this would work
this one seem like a natural and quite cheap way to get the improvement

mzhukova:
FYI @riverlijunjie

dzarukin:
Left a minor comment in PoC branch. If it wouldn't affect API, I'm fine with the proposal.

PR #4936, opened 2026-04-01
xe: fix kernel cache key validation

rjoursler:
@echeresh Noticed that kernel cache key debug validation was not behaving as expected. This patch resolves this issue and addresses some missing functionality needed for the validation.

### DIFF
--- a/src/gpu/intel/conv/jit/zero_out.hpp
+++ b/src/gpu/intel/conv/jit/zero_out.hpp
@@ -51,7 +51,9 @@ public:
             const intel::engine_t &engine, compute::kernel_t &kernel) const;
     serialization_stream_t serialize() const override;
     static zero_out_kernel_desc_t deserialize(const serialization_stream_t &s);
-
+#if __cplusplus >= 202002L
+    bool operator==(const zero_out_kernel_desc_t &) const = default;
+#endif
     static compute::nd_range_t nd_range(int simd, size_t size);
 
 private:
--- a/src/gpu/intel/jit/ir/kernel_desc.hpp
+++ b/src/gpu/intel/jit/ir/kernel_desc.hpp
@@ -56,6 +56,9 @@ public:
             primitive_t *primitive, impl::engine_t *engine) const
             = 0;
     virtual serialization_stream_t serialize() const = 0;
+#if __cplusplus >= 202002L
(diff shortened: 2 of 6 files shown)


rjoursler:
make test

src/gpu/intel/jit/ir/kernel_desc.hpp
```diff
@@ -56,6 +56,9 @@ class kernel_desc_base_t {
             primitive_t *primitive, impl::engine_t *engine) const
             = 0;
     virtual serialization_stream_t serialize() const = 0;
+#if __cplusplus >= 202002L
+    bool operator==(const kernel_desc_base_t &) const = default;
```
kealan-barbieri, line 60: Are these operators required due to the new validator implementation?

rjoursler: Yeah, previously no validation was being performed on `zero_out_desc_t`. After fixing the check to enable validation, we now run the following which requires `operator==` to be defined.

```
desc == zero_out_desc_t::deseriailze(desc.serialize()); 
```

src/gpu/intel/conv/jit/zero_out.hpp
```diff
@@ -51,7 +51,9 @@ class zero_out_kernel_desc_t : public kernel_desc_base_t {
             const intel::engine_t &engine, compute::kernel_t &kernel) const;
     serialization_stream_t serialize() const override;
     static zero_out_kernel_desc_t deserialize(const serialization_stream_t &s);
-
+#if __cplusplus >= 202002L
+    bool operator==(const zero_out_kernel_desc_t &) const = default;
+#endif
```
dzarukin, line 56: Minor: would it hurt to drop macro (not compile) around it?

rjoursler: This is part of the requirements for the validation method. Effectively, the GPU kernel cache keys are hashed via `hash(key.serialize())`, and when we add members to a key structure, we can introduce caching bug which are difficult to catch and debug if we do not propagate those changes into `serialize()`. By relying on default `operator==` (which is a C++20 feature), this class of bug gets caught by checking: `key == deserialize(serialize(key))`. Implementing a custom operator== would be largely counterproductive, as it would be susceptible to the same class of bugs.

PR #4939, opened 2026-04-01
common: TLS freelist for memory descriptors to speed up PyTorch CPU inference

lslusarczyk:
### Summary

Adds a **thread-local freelist** for C API memory descriptors in `memory_desc.cpp`, so `dnnl_memory_desc_create_*` / `dnnl_memory_desc_destroy` can reuse allocations on hot paths instead of always hitting the heap. Pool size is configurable via **ONEDNN_MEMORY_DESC_FREELIST_CAPACITY** / **DNNL_MEMORY_DESC_FREELIST_CAPACITY** (default 256; 0 disables pooling). Developer documentation is added under _Performance profiling and inspection_.

### Motivation

In CPU-bound batch inference for a sample Vision Transformer workload, a noticeable share of time sits under `dnnl::memory::desc::desc`, which wraps C API creation and `shared_ptr` setup. Profiling shows substantial cost in the allocation path inside `dnnl_memory_desc_create_with_tag` (e.g. `dnnl::impl::malloc` / `posix_memalign`) plus a separate operator new for the `shared_ptr` control block in the C++ wrapper.

This change targets the **C API descriptor allocation** side by reusing freed descriptors per OS thread, which removes or shrinks the repeated **malloc/aligned allocation** stack seen under `create_with_tag` in the “before” profiles.

### Benchmark (PyTorch + oneDNN)

- Scenario: CPU-bound ViT-style inference, batch mode (same setup before/after except oneDNN build).
- Overall wall time: **~4%** faster end-to-end on the measured run.
- **oneDNN-attributed overhead** (share of total time attributed to oneDNN in this scenario): **15.6%** → **12.6%**, i.e. a **~19% relative reduction** of that overhead slice (not 19% of full application time).
- GPU: B580, CPU: i9 Core 14900K

#### Reproducer / measurement code
[real_world_app.py](https://github.com/lslusarczyk/pytorch/blob/main/benchmarks/dynamo/microbenchmarks/xpu/real_world_app.py)
Command: 
```
python3 pytorch/benchmarks/dynamo/microbenchmarks/xpu/real_world_app.py \
--device xpu --iter 100000 --model vit --batch 50 --compile
```

### Flame graphs 
Illustrate both: the matmul hot path and the descriptor constructor search hit.

#### 1) Matmul stack — before vs after

**Before** Under `at::native::onednn::matmul` → `dnnl::memory::desc::desc` → `dnnl_memory_desc_create_with_tag`, a wide slice is `dnnl::impl::malloc` → `__posix_memalign` (aligned allocation for the descriptor body).

**After** The same frame is reduced: reused descriptors avoid that **malloc/posix_memalign** path when the thread-local pool has entries; remaining cost is mostly init logic and (unchanged) C++ wrapper overhead where applicable.

dnn_before_desc
dnn_after_desc

#### 2) Search / aggregate: desc::desc — before vs after

**Before** Search for `desc::desc` (or matching symbol in your viewer) shows a large matched fraction of total time (e.g. **~4.1%** of the profile in one capture), reflecting frequent descriptor construction on the path to primitive creation.

**After** The same search shows a smaller matched share (e.g. **~2.5%** in the corresponding capture), consistent with fewer heap allocations on the C API create/destroy path.

dnn_before
dnn_after

**Note:** Absolute percentages depend on model, batch size, CPU, and how “oneDNN overhead” is attributed in the profiler; the numbers above match the labeled runs used for this PR.

### Technical notes

- **ABI**: Unchanged public C signatures; only implementation of allocation/release in `memory_desc.cpp`.
- **Threading**: Destroy does not need to run on the same thread as create; pooling is per thread that calls `destroy`.
- **Docs**: `@ref dev_guide_memory_desc_freelist` (new page in the dev guide).

### DIFF
--- a/src/common/nstl.hpp
+++ b/src/common/nstl.hpp
@@ -64,6 +64,12 @@ struct c_compatible { // NOLINT(readability-identifier-naming)
         UNUSED(sz);
         return p;
     }
+    // MSVC C4291: placement new requires a matching placement delete (no-op here;
+    // storage is caller-owned). Invoked only if the constructor throws.
+    static void operator delete(void *p, void *place) noexcept {
+        UNUSED(p);
+        UNUSED(place);
+    }
     static void *operator new[](size_t sz) {
         return MALLOC(sz, default_alignment);
     }


vpirogov:
PyTorch [recommends](https://docs.pytorch.org/tutorials/recipes/recipes/tuning_guide.html#switch-memory-allocator) jemalloc as a solution to memory allocation problem. Does it help?

lslusarczyk:
> PyTorch [recommends](https://docs.pytorch.org/tutorials/recipes/recipes/tuning_guide.html#switch-memory-allocator) jemalloc as a solution to memory allocation problem. Does it help?

Thanks for the pointer. **Yes — a faster general-purpose allocator helps** in this scenario (less time in malloc / heap paths). My measurements used tcmalloc as the alternative allocator; I’d expect jemalloc to move results in a similar direction (less allocator overhead), though the exact numbers will differ.

Even with that, **the memory-descriptor freelist still improves latency on top of the allocator**, because hot paths can **reuse descriptors** and avoid calling into the allocator for every create/destroy pair. Below are **three-run means ± sample standard deviation** (mseconds per one `model(x)` call), plus **speedup** of the optimized build vs the non-optimized build within the same allocator setting.

| Allocator setup |Without MD freelist (mean ± σ) |With MD freelist (mean ± σ) |Speedup (freelist vs baseline) |
| :--- | :--- | :--- | :--- |
| With tcmalloc | 1.437 ± 0.005 | 1.400 ± 0.009 | 2.6% ± 0.8% |
| Without tcmalloc (default) |1.803 ± 0.026 |1.687 ± 0.015 | 6.4% ± 2.1% |

**Takeaway**: tcmalloc closes a large part of the gap from allocator cost, but **the freelist still adds a clear extra gain** — and it does so by avoiding those allocations in the tight loop, which remains useful even when the global allocator is already “fast.”

lslusarczyk:
@vpirogov , oneDNN Team,
CI passed, could you please review and merge if all is OK?

dzarukin:
> @vpirogov , oneDNN Team, CI passed, could you please review and merge if all is OK?

Hi @lslusarczyk, thank you for the contribution and providing proof points. However, some details are still missed to make an educated choice towards accepting this contribution.

Could you, please, answer this:
* What is exactly measured in a table with tcmalloc comparison? What are those times?
* From what I can understand, freelist and tcmalloc solve exactly same problem - retrieve the memory piece from the pool instead of doing allocation. I'm not sure I understand how they work together and where the speed up with using freelist is coming from?

Thank you.

lslusarczyk:
> > @vpirogov , oneDNN Team, CI passed, could you please review and merge if all is OK?
> 
> Hi @lslusarczyk, thank you for the contribution and providing proof points. However, some details are still missed to make an educated choice towards accepting this contribution.
> 
> Could you, please, answer this:
> 
>     * What is exactly measured in a table with tcmalloc comparison? What are those times?

The table uses the an average Latency from [real_world_app.py](https://github.com/lslusarczyk/pytorch/blob/main/benchmarks/dynamo/microbenchmarks/xpu/real_world_app.py), i.e. `1000 * (end_time - start) / N` ms, where `start / end_time` are `time.time()` around the whole timed loop and `N` being number of iterations.

Run configuration: `--device xpu --iter 100000 --model vit --batch 50 --compile`.

Per “iteration” in the divisor: one forward of the compiled ViT (`vit_tiny_patch16_224` path for `--model vit`), that is time of `model(x)` after `torch.compile`.
`--batch 50`: the loop runs 50 forwards, then `torch.xpu.synchronize()`, then the next 50, and so on, for 100000 iterations total. So the measured wall time includes periodic device sync every 50 steps, not a single sync only at the end.

What the number is: mean wall‑clock milliseconds per forward, over the full timed region — Python + host/driver/SYCL + allocator effects (tcmalloc vs default malloc shows up here if the hot path allocates/frees on the host) + XPU work, as seen by `time.time()`, with the above batching/sync pattern.

So the tcmalloc comparison is that reported average ms/iter for identical flags, only the process allocator changed.

> 
>     * From what I can understand, freelist and tcmalloc solve exactly same problem - retrieve the memory piece from the pool instead of doing allocation. I'm not sure I understand how they work together and where the speed up with using freelist is coming from?

What tcmalloc optimizes is the generic allocator path: fast thread caches, size classes, and returning freed blocks to a central pool so that a subsequent malloc/free of similar size is cheap. You still pay for going through the allocator API, cache lookup, bookkeeping, and any locking or metadata updates on that path.

What the freelist does here is avoid that path entirely for memory_desc_t: after the first few creations, dnnl_memory_desc_destroy / acquire typically reuse a fixed object size (one memory_desc_t), so we only run the destructor, optionally push_back/pop_back on a per-thread vector, and placement-new into storage we already own. There is no call into the heap allocator for the descriptor body on the hot reuse path.

So they work together in the sense that tcmalloc still backs the first allocation (and the vector’s internal buffer growth, any non-pooled paths, etc.), while the freelist removes repeated malloc/free for this specific, uniform type from the hot path.

Even a very good allocator is not free - this is why we still see ~2% with tcmalloc. Skipping it for a tight loop of create/destroy of the same POD-like object saves predictable work (allocator entry, possible cache refill, metadata). The gain is smaller than without tcmalloc because tcmalloc already makes “small repeated alloc” cheap, but the freelist removes a layer tcmalloc cannot remove by design.

~4% case without tcmalloc still matters because not all oneDNN users link against tcmalloc (default glibc malloc, other runtimes, embedded/static builds, different platforms). In those configurations the same freelist avoids a slower baseline allocator more often, so the relative win is larger. Optimizing both cases is useful: we get a modest win when tcmalloc is present and a larger win when it is not.

> 
> 
> Thank you.

vpirogov:
> @vpirogov , oneDNN Team, CI passed, could you please review and merge if all is OK?

Upon discussion we decided not to accept these changes, as these effectively represent a custom memory manager implementation for a single specific oneDNN object and produce negligible performance gain in a very narrow case.

vpirogov:
ViT benchmark you are using runs a static model with tiny sizes. In fact it executes only 10 unique oneDNN calls out of ~5000 calls executed for one iteration.

| Function | Shape |
| :--- | :--- |
| convolution | mb1_ic3oc192_ih16oh1kh16sh16dh0ph0_iw16ow1kw16sw16dw0pw0 |
| matmul | 2x192:192x576 |
| matmul | 1x3x2x64:1x3x64x2 |
| softmax	| 1x3x2x2 |
| matmul | 1x3x2x2:1x3x2x64 |
| graph:sdpa | in0_f32:0:strided:undef:1x3x2x64:192s64s576s1 in1_f32:1:strided:undef:1x3x2x64:192s64s576s1 in2_f32:2:strided:host_scalar:0:2 in3_f32:5:strided:undef:1x3x2x64:192s64s576s1 out0_f32:6:strided:undef:1x3x2x64:192s64s192s1
| matmul | 2x192:192x192 |
| matmul | 2x192:192x768 |
| matmul | 2x768:768x192 |
| matmul | 1x192:192x1000 |

Here're some ideas on improving performance for this case:
1. Look at why PyTorch is making hundreds thousands allocations in this case. oneDNN programming model does not require that. So can we improve PyTorch here?
2. The shapes executed by PyTorch are relatively small so there's definitely low GPU utilization and overhead from kernel submission. Are there model level optimizations that can make computation more efficient?

PR #4946, opened 2026-04-02
all: support for optional upper bound arg for grouped matmul

mzhukova:
Implements recommended proposal from [RFC 4931](https://github.com/uxlfoundation/oneDNN/pull/4931) with a minor renaming to generalize (in case we would have SRC grouped by K dim).

Latest discussion and perf data could be found at [MFDNN-14695](https://jira.devtools.intel.com/browse/MFDNN-14695).

### DIFF
--- a/doc/primitives/matmul.md
+++ b/doc/primitives/matmul.md
@@ -412,6 +412,31 @@ auto bias_md = memory::desc({num_groups, N},
 // Layout: standard ab layout
 ~~~
 
+### Execution Hints
+
+An optional execution-time hint `DNNL_ARG_HINT_MAX_GROUP_SIZE` can be provided to
+communicate the maximum size of the group across the variable dimension for the
+current execution call. Implementations may choose to use this input to tune
+dispatch, and therefore using this hint may provide performance benefits.
+
+If chosen, the hint is passed as a host scalar `s32` memory at execution time:
+~~~cpp
+int32_t max_size = 950; // upper bound on variable dimension across all groups for this call
+auto hint_md = memory::desc::host_scalar(memory::data_type::s32);
+auto hint_mem = memory(hint_md, engine, &max_size);
+
+matmul_prim.execute(stream, {
+    {DNNL_ARG_SRC, src_mem},
+    {DNNL_ARG_WEIGHTS, weights_mem},
+    {DNNL_ARG_DST, dst_mem},
+    {DNNL_ARG_HINT_MAX_GROUP_SIZE, hint_mem}  // optional
+});
+~~~
+
+@warning Providing a value smaller than the actual maximum variable dimension across
+groups for the current call will produce incorrect results. It is the caller's
+responsibility to ensure the hint is a valid upper bound.
+
 ### Implementation Notes
 
 The following are supported:


src/common/primitive_desc.hpp
```diff
@@ -216,6 +216,11 @@ struct primitive_desc_t : public c_compatible {
                     ? arg_usage_t::input
                     : arg_usage_t::unused;
 
+        // Hint is an optional execution-time values with no related pd state
+        // TODO: For now make any primitive accept it,
+        // and it is safe for kernel to ignore it if not supported
+        if (arg == DNNL_ARG_HINT_MAX_GROUP_M) return arg_usage_t::input;
```
dzarukin: Should it pop up in matmul_pd only under `is_grouped()` condition?

The concern is without accounting for a hint through `n_inputs`, the check for the number of input arguments would be broken. One may pass a hint but not pass a mandatory buffer and the existing check would be tweaked and let such scenario roll. At least, it's negative side-effect.

A memory descriptor could have a field whether to expect the hint or not, then it could be accounted properly.

mzhukova: Yes, that's a valid concern! I was going to redo that after getting some traction from POC.
It is not moved to matmul_pd and I've added a few gtest cases to test scenario you described, I could see it correctly rejected and also waning messages in verbose mode

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul
enable run_documentation

mzhukova:
FYI @umar456

mzhukova:
ping @uxlfoundation/onednn-arch for review

// note that the current CI failure is a known issue and unrelated to the changes in this PR

tests/gtests/test_iface_grouped.cpp
```diff
@@ -414,6 +414,78 @@ TEST(iface_grouped_test_t, TestGroupedMatmulPatterns) {
             dnnl::error);
 }
 
+HANDLE_EXCEPTIONS_FOR_TEST(iface_grouped_test_t, TestMaxGroupMHint) {
+    engine eng = get_test_engine();
+    const bool is_unimplemented = (eng.get_kind() == engine::kind::gpu
+            || DNNL_CPU_RUNTIME == DNNL_RUNTIME_SYCL);
+    if (is_unimplemented) return;
```
dzarukin: There's `SKIP_IF` interface for such kind of functionality - it provides verbosity to the output.

tests/gtests/test_iface_grouped.cpp
```diff
@@ -414,6 +414,78 @@ TEST(iface_grouped_test_t, TestGroupedMatmulPatterns) {
...
+
+    auto hint_md = memory::desc::host_scalar(dt::s32);
+    memory hint_mem(hint_md, int32_t(3));
+
+    // Grouped matmul: hint is accepted
+    {
+        const int ngroups = 2;
+        const int M = 6, K = 4, N = 4;
```
dzarukin, line 430: Should sizes be bigger to be accepted by the optimized implementation?

tests/benchdnn/dnnl_common.hpp
```diff
@@ -936,6 +936,18 @@ void init_memory_args(dnn_mem_map_t &mem_map, const prb_t *prb,
...
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+    // Grouped max variable dim hint
+    // Optional host scalar, created when src is a grouped descriptor
+    {
+        const auto *src_md = query_md(const_pd, DNNL_ARG_SRC);
+        if (query_md_sparse_encoding(src_md) == dnnl_grouped) {
+            auto hint_md = dnn_mem_t::init_host_scalar_md(dnnl_s32);
+            mem_map.emplace(DNNL_ARG_HINT_MAX_GROUP_SIZE, dnn_mem_t(hint_md));
+        }
+    }
```
dzarukin:
```suggestion
    if (query_md_sparse_encoding(query_md(const_pd, DNNL_ARG_SRC)) == dnnl_grouped) {
        auto hint_md = dnn_mem_t::init_host_scalar_md(dnnl_s32);
        mem_map.emplace(DNNL_ARG_HINT_MAX_GROUP_SIZE, dnn_mem_t(hint_md));
    }
```

tests/benchdnn/utils/parser.cpp
```diff
@@ -830,11 +833,26 @@ bool parse_grouped(std::vector<sparse_options_t> &sparse_options,
...
+    dnnl_dim_t max_variable_dim = 0;
+    std::string sizes_str;
+    const size_t colon_pos = remainder.find(':');
+    if (colon_pos != std::string::npos) {
+        sizes_str = remainder.substr(0, colon_pos);
+        max_variable_dim
+                = parser_utils::stoll_safe(remainder.substr(colon_pos + 1));
+    } else {
+        sizes_str = remainder;
+    }
```
dzarukin: The logic might get straightened by getting a substr till `:` for `sizes`. If start_pos == string::npos after that, it means there's no hint. If it's not, then it was specified.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -1219,6 +1219,13 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
                     }
                 }
             } break;
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+            case DNNL_ARG_HINT_MAX_GROUP_SIZE:
+                mem.set_elem(0,
```
dzarukin, line 1226:
```suggestion
                // The hint is for the library only, reference is not affected.
                mem.set_elem(0,
```

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
```diff
@@ -172,3 +172,16 @@
...
+--dt=f32:f32:f32,bf16:bf16:bf16
+--wtag=abc
+--grouped=0:4:8,8,8,8:8
+32x64:4x64x32
+
+--reset
+--dt=f16:f16:f16
+--wtag=abc
+--grouped=0:4:10,20,30,40:40
+100x64:4x64x32
```
dzarukin, line 187: Maybe incorporate the hint into some of existing cases?

As a side note - it feels like using `,` for sizes separation hurts the ability to run several grouped cases as it can be done with other parameters. If you plan to replace them with a runtime generation sooner, probably, not a big deal. If not, might be an option to consider to replace `,` with other delimiter to allow running several cases. E.g. a single case with and without the hint:
```
--grouped=0:4:10+20+30+40,0:4:10+20+30+40:40
```

doc/primitives/matmul.md
```diff
@@ -412,6 +412,31 @@ auto bias_md = memory::desc({num_groups, N},
...
+    {DNNL_ARG_SRC, src_mem},
+    {DNNL_ARG_WEIGHTS, weights_mem},
+    {DNNL_ARG_DST, dst_mem},
+    {DNNL_ARG_HINT_MAX_GROUP_SIZE, hint_mem}  // optional
+});
+~~~
+
+@warning Providing a value smaller than the actual maximum variable dimension across
+groups for the current call will produce incorrect results. It is the caller's
+responsibility to ensure the hint is a valid upper bound.
```
dzarukin, line 438: Would it hurt performance badly to pass this hint inside the kernel and verify its value against current M obtained from the memory?

If it does in the release mode, maybe check it under debug at least. It would be sad to waste debugging time on figuring out that the hint passed was wrong.

tests/benchdnn/inputs/matmul/test_matmul_grouped_ci
mzhukova: Good point, I've added to https://jira.devtools.intel.com/browse/MFDNN-14704, will address in a separate PR

doc/primitives/matmul.md
mzhukova: tag @umar456 for input

tests/benchdnn/dnnl_common.hpp
mzhukova: sure, I've merged lines

tests/benchdnn/utils/parser.cpp
mzhukova: good suggestion, simplified the code

tests/benchdnn/matmul/matmul.cpp
mzhukova: added the comment

tests/gtests/test_iface_grouped.cpp
mzhukova: thanks for the tip, I've updated the whole file to rely on it instead

tests/gtests/test_iface_grouped.cpp
mzhukova: I'm not sure, I think there is no constraint currently for sizes so that opt kernel could be picked up
also, this test is more of an API/interface one to ensure passing hint through API doesn't produce errors, real testing is done via benchdnn

doc/primitives/matmul.md
mzhukova: I think we could leave it to @umar456 to evaluate/implement and have it in another PR, I could create a task for it

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

doc/primitives/matmul.md
umar456: Its not too difficult to check this but I am not sure what we can do after we find that the value is incorrect. We would need to perform a synchronization to read that value before exiting the primitive which is not ideal. We can also abort but I don't think that is a great solution either.

What do you suggest we do if we find out if its incorrect?

mzhukova: @umar456 I think one of the suggestions was to provide more info in debug build, so that it would be easier to catch such issues, I agree that synchronization / abort is not ideal.

umar456: Okay, yeah, we can add that info as a warning line. Is it okay to optionally synchronize if the warning logs are enabled via ONEDNN_VERBOSE? I am not sure if we want to have a different behavior based the log levels.

dzarukin: I'd think the kernel can simply abort the execution if the hint would lead to a correctness problem (has a smaller value than the one provided at runtime). If kernels return statuses, it can be utilized. If not, maybe just a print would be sufficient (I don't recall verbose works under cl).

tests/benchdnn/dnnl_common.hpp
dzarukin: Nit: outer parenthesis can be dropped, too as there're no local variables.

mzhukova: oh yes, haven't noticed it, good catch

doc/primitives/matmul.md
mzhukova: Added into [MFDNN-14886](https://jira.devtools.intel.com/browse/MFDNN-14886)

PR #4951, opened 2026-04-03
benchdnn: add GatedMLP driver

kwieloch-intel:
### This PR adds a new benchdnn driver for the **Gated MLP** primitive, enabling correctness validation and performance benchmarking of the fused GatedMLP GPU kernel directly from the benchdnn application.

**JIRA:** [MFDNN-14716](https://jira.devtools.intel.com/browse/MFDNN-14716)

---

Adds a `--gated_mlp` benchdnn driver for correctness validation and performance benchmarking of the fused GatedMLP GPU kernel.

**Scope:** `tests/benchdnn/` and `scripts/verbose_converter/` only. Zero changes to `src/`.

#### Features

- Data types: f32, bf16, f16, s8, u8, s4, u4 (SRC/weights/DST independently)
- Activations: swish (default), gelu_erf, gelu_tanh
- Quantization: `--attr-scales` / `--attr-zero-points` for `wei`, `wei_up`, `wei_down`, `src` with `per_dim_1` and `per_dim_01` policies (grouped quantization)
- 2D (`--stag=ab --wtag=ba`) and fake-3D (`--stag=abc --wtag=cab`) tensor layouts
- Problem descriptor: `MBxICxOC`
- Verbose converter support (`GatedMLPConverter`)

#### Reference implementation

Composes existing oneDNN primitives on CPU (matmul + eltwise + binary) with explicit weight/src dequantization when quantization attributes are set.

#### Test suites

```
test_gated_mlp_smoke  - f32/f16/bf16, basic shapes, all activations
test_gated_mlp_ci     - f32/f16/bf16, basic shapes
test_gated_mlp_gpu    - f16/bf16, LLM shapes + u4 quantized (scales, scales+ZPs)
```

---

#### General

#### New features

- [ ] ~~Have you published an RFC for the new feature?~~ (N/A — benchdnn driver only)
- [ ] ~~Was the RFC approved?~~ (N/A)
- [x] Have you added relevant tests?
- [x] Have you added relevant documentation? (`doc/driver_gated_mlp.md`)

### DIFF
--- a/tests/benchdnn/dnn_types.cpp
+++ b/tests/benchdnn/dnn_types.cpp
@@ -96,6 +96,8 @@ static const std::map<int, std::vector<const char *>> supported_args {
         {DNNL_ARG_SRC, {"src", "src0"}},
         {DNNL_ARG_SRC_1, {"src1"}},
         {DNNL_ARG_WEIGHTS, {"wei"}},
+        {DNNL_ARG_WEIGHTS_1, {"wei_up"}},
+        {DNNL_ARG_WEIGHTS_2, {"wei_down"}},
         {DNNL_ARG_DST, {"dst"}},
         {DNNL_ARG_ATTR_POST_OP_DW | DNNL_ARG_DST, {"attr_post_op_dw_dst"}},
         {DNNL_ARG_ATTR_POST_OP_DW | DNNL_ARG_WEIGHTS, {"attr_post_op_dw_wei"}},
--- a/tests/benchdnn/doc/driver_gated_mlp.md
+++ b/tests/benchdnn/doc/driver_gated_mlp.md
@@ -1,0 +1,128 @@
+# GatedMLP Driver
+
+## Usage
+``` sh
+    ./benchdnn --gated_mlp [benchdnn-knobs] [gated_mlp-knobs] [gated_mlp-desc] ...
+```
+
+where *gated_mlp-knobs* are:
+
+ - `--dt={f32 [default], ...}` -- source, weight, and destination data types.
+            Interface supports broadcasting, when a single input is provided,
+            e.g., `--dt=f32`, the value is applied for all tensors. Five
+            individual values can be specified in SRC, W_GATE, W_UP, W_DOWN,
+            DST order.
+            Refer to [data types](knobs_dt.md) for details.
+ - `--stag={abx [default], ...}` -- memory format of the source tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--wtag={abx [default], ...}` -- memory format of the weight tensors
+            (applied to all three: gate, up, down).
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--activation={swish [default], gelu_erf, gelu_tanh}` -- specifies the
+            gated activation function applied after the gate matmul.
+ - `--match=REGEX` -- skip problems not matching the regular expression in
+            `REGEX`. By default no pattern is applied (run everything).
+            Note: Windows may interpret only string arguments surrounded by
+            double quotation marks.
+
+and *gated_mlp-desc* is a problem descriptor. The canonical form is:
+```
+    MBxICxOC
+```
+Here `x` is the delimiter for the three dimensions: `MB` (batch size), `IC`
+(input channels / model dimension), and `OC` (intermediate dimension). All
+tensor shapes are derived from these three values:
+- Source: `[MB, IC]`
+- Gate weights: `[IC, OC]`
+- Up weights: `[IC, OC]`
+- Down weights: `[OC, IC]`
+- Destination: `[MB, IC]`
+
+When a 3D tag is specified (e.g. `--stag=abc --wtag=cab --dtag=abc`), the
+driver creates "fake 3D" tensors with a unit dimension matching the layout used
+by OpenVINO networks:
+- Source: `[MB, 1, IC]`
+- Gate/Up weights: `[1, IC, OC]`
+- Down weights: `[1, OC, IC]`
+- Destination: `[MB, 1, IC]`
+
+## Quantization Attributes
+
+The driver supports `--attr-scales` and `--attr-zero-points` for input and
+weight dequantization. Supported argument names are `src` (for quantized
+activations), `wei`, `wei_up`, and `wei_down` (corresponding to the gate, up,
+and down projection weights).
+
+Supported policies:
+- `common` -- single scale/zero-point shared across the entire tensor.
+- `per_dim_1` -- one value per output channel (dimension 1 of the weight tensor).
+- `per_dim_01` -- grouped quantization along both dimensions (group size
+            specified inline, e.g., `per_dim_01:f16:32x1`).
+
+Example: run u4 quantized weights with per-channel f32 scales:
+``` sh
+    ./benchdnn --gated_mlp --dt=f32:u4:u4:u4:f32 \
+               --attr-scales=wei:per_dim_1:f32+wei_up:per_dim_1:f32+wei_down:per_dim_1:f32 \
+               64x128x256
+```
+
+Example: grouped quantization with groups along the K-dimension:
+``` sh
+    ./benchdnn --gated_mlp --dt=f32:u4:u4:u4:f32 \
+               --attr-scales=wei:per_dim_01:f32:32x1+wei_up:per_dim_01:f32:32x1+wei_down:per_dim_01:f32:32x1 \
+               --attr-zero-points=wei:per_dim_01:s4:32x1+wei_up:per_dim_01:s4:32x1+wei_down:per_dim_01:s4:32x1 \
+               64x128x256
+```
+
+Example: INT8 SRC with per-tensor SRC scale and quantized weights:
+``` sh
+    ./benchdnn --gated_mlp --dt=s8:u4:u4:u4:f16 \
+               --attr-scales=src:common:f16+wei:per_dim_01:f16:128x1+wei_up:per_dim_01:f16:128x1+wei_down:per_dim_01:f16:128x1 \
+               64x256x4864
+```
+
+## Post-ops
+
+The driver supports post-ops applied to the final (down) matmul output.
+Supported kinds include `sum`, `eltwise`, and `binary`.
+
+Example: binary add post-op:
+``` sh
+    ./benchdnn --gated_mlp --dt=f16:u4:u4:u4:f16 \
+               --attr-scales=wei:per_dim_01:f16:128x1+wei_up:per_dim_01:f16:128x1+wei_down:per_dim_01:f16:128x1 \
+               --attr-post-ops=binary_add:f16:2:ab \
+               64x128x4864
+```
+
+## Essence of Testing
+
+The GatedMLP operation computes
+`DST = (activation(SRC • W_gate) ⊙ (SRC • W_up)) • W_down`,
+where `•` denotes matrix multiplication and `⊙` denotes element-wise
+(Hadamard) product. The reference executes each step independently using f32
+matmul and element-wise primitives on the CPU. The driver compares the fused
+primitive output against this stepwise reference.
+
+## Examples
+
+Run the default validation set of GatedMLP using `inputs/gated_mlp/shapes_basic`
+file:
+``` sh
+    ./benchdnn --gated_mlp --batch=inputs/gated_mlp/shapes_basic
+```
+
+Run f16 GatedMLP with gelu_erf activation on a small shape:
+``` sh
+    ./benchdnn --gated_mlp --dt=f16 --activation=gelu_erf 64x128x256
+```
+
+Run GatedMLP performance benchmark on GPU with an LLM shape:
+``` sh
+    ./benchdnn --mode=f --gated_mlp --engine=gpu --dt=f16 \
+               --activation=swish 1024x896x4864
+```
+
+More examples with different driver options can be found at
+inputs/gated_mlp/test_\*.
--- a/tests/benchdnn/gated_mlp/cfg.cpp
+++ b/tests/benchdnn/gated_mlp/cfg.cpp
@@ -1,0 +1,95 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "gated_mlp/gated_mlp.hpp"
+
+namespace gated_mlp {
+
+cfg_t::cfg_t(const prb_t *prb, const std::vector<data_kind_t> &kinds)
+    : base_cfg_t {prb->attr.acc_mode} {
+    for (const auto kind : kinds) {
+        auto orig_data_type = prb->get_dt(kind);
+        auto data_type = deduce_cfg_data_type(orig_data_type, prb->attr, kind);
+        cfg_entry_.emplace(kind,
+                cfg_entry_t {
+                        kind, orig_data_type, data_type, get_cfg_map(kind)});
+    }
+
+    adjust_ranges();
+    print_fill_cfg_verbose();
+}
+
+// Adjust density based on accumulation chain length.
+float cfg_t::get_density(const cfg_t::density_args_t &density_args) const {
+    float density = 1.f;
+    if (density_args.data_kind != SRC) return density;
+
+    const int64_t safe_n_acc = get_safe_n_acc();
+    float safe_density = (float)safe_n_acc / density_args.n_acc;
+    density = MIN2(density, safe_density);
+
+    BENCHDNN_PRINT(6, "%s safe_n_acc=%d density=%f\n", "[FILL_CFG]",
+            (int)safe_n_acc, density);
+
+    return density;
+}
+
+cfg_t::cfg_entry_t::cfg_map_t cfg_t::get_cfg_map(data_kind_t kind) const {
+    // Tight ranges for chained pipeline. SRC is further scaled in fill.
+    static const cfg_t::cfg_entry_t::cfg_map_t src_cfg_map = {
+            {{dnnl_f32}, {-1, 1}},
+            {{dnnl_bf16}, {-1, 1}},
+            {{dnnl_f16}, {-1, 1}},
+            {{dnnl_s8}, {-1, 1}},
+            {{dnnl_u8}, {0, 2}},
+            {{dnnl_s4}, {-1, 1}},
+            {{dnnl_u4}, {0, 2}},
+    };
+
+    // Weight tensor ranges, small to control accumulation magnitude.
+    static const cfg_t::cfg_entry_t::cfg_map_t wei_cfg_map = {
+            {{dnnl_f32}, {-1, 1}},
+            {{dnnl_bf16}, {-1, 1}},
+            {{dnnl_f16}, {-1, 1}},
+            {{dnnl_s8}, {-1, 1}},
+            {{dnnl_u8}, {0, 2}},
+            {{dnnl_s4}, {-1, 1}},
+            {{dnnl_u4}, {0, 2}},
+    };
+
+    // Destination tensor ranges.
+    static const cfg_t::cfg_entry_t::cfg_map_t dst_cfg_map = {
+            {{dnnl_f32}, {-8, 8}},
+            {{dnnl_bf16}, {-4, 4}},
+            {{dnnl_f16}, {-4, 4}},
+            {{dnnl_s32}, {-8, 8}},
+            {{dnnl_s8}, {-4, 4}},
+            {{dnnl_u8}, {0, 8}},
+            {{dnnl_s4}, {-2, 2}},
+            {{dnnl_u4}, {0, 4}},
+    };
+
+    switch (kind) {
+        case SRC: return src_cfg_map;
+        case WEI: return wei_cfg_map;
+        case DST: return dst_cfg_map;
+        default: assert(!"unsupported data kind"); break;
+    }
+    static cfg_t::cfg_entry_t::cfg_map_t dummy;
+    return dummy;
+}
+
+} // namespace gated_mlp
--- a/tests/benchdnn/gated_mlp/gated_mlp.cpp
+++ b/tests/benchdnn/gated_mlp/gated_mlp.cpp
@@ -1,0 +1,349 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <math.h>
+#include <random>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "oneapi/dnnl/dnnl.h"
+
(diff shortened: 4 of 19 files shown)


tests/benchdnn/doc/driver_gated_mlp.md
```diff
@@ -0,0 +1,72 @@
...
+- Source: `[MB, IC]`
+- Gate weights: `[IC, OC]`
+- Up weights: `[IC, OC]`
+- Down weights: `[OC, IC]`
+- Destination: `[MB, IC]`
+
+## Essence of Testing
+
+The GatedMLP operation computes
+`DST = (activation(SRC * W_gate) * (SRC * W_up)) * W_down`.
```
hidefromkgb: I believe there has to be a distinction here between the matmul '\*' and the elementwise '\*'.

tests/benchdnn/doc/driver_gated_mlp.md
```diff
@@ -0,0 +1,72 @@
...
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--activation={swish [default], gelu_erf, gelu_tanh}` -- specifies the
+            gated activation function applied after the gate matmul.
+ - `--match=REGEX` -- skip problems not matching the regular expression in
+            `REGEX`. By default no pattern is applied (run everything).
+            Note: Windows may interpret only string arguments surrounded by
+            double quotation marks.
+ - Any attributes options. Refer to [attributes](knobs_attr.md) for details.
```
hidefromkgb: The existing implementations only support a narrow subset of attrs — namely, grouped scales and grouped zero points for 2 kinds of weights (Gate/Up should be of the same shape and type, Down can differ).

tests/benchdnn/doc/driver_gated_mlp.md
```diff
@@ -0,0 +1,72 @@
...
+ - `--stag={abx [default], ...}` -- memory format of the source tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--wtag={abx [default], ...}` -- memory format of the weight tensors
+            (applied to all three: gate, up, down).
+            Refer to [tags](knobs_tag.md) for details.
+ - `--dtag={abx [default], ...}` -- memory format of the destination tensor.
+            Refer to [tags](knobs_tag.md) for details.
+ - `--activation={swish [default], gelu_erf, gelu_tanh}` -- specifies the
+            gated activation function applied after the gate matmul.
+ - `--match=REGEX` -- skip problems not matching the regular expression in
```
hidefromkgb, line 25: Should we really do that instead of splitting the test set into chunks that can be included or excluded?

tests/benchdnn/gated_mlp/ref_gated_mlp.cpp
```diff
@@ -0,0 +1,159 @@
...
+    exec_matmul(eng, strm, src_ref, w_up_ref, up_result);
+
+    // Step 2: gate_result = matmul(src, W_gate).
+    exec_matmul(eng, strm, src_ref, w_gate_ref, gate_result);
+
+    // Step 3: gate_result = activation(gate_result).
+    exec_eltwise(eng, strm, gate_result, prb->activation);
+
+    // Step 4: gate_result = gate_result * up_result (element-wise).
+    {
```
hidefromkgb: why not exec_binary()?

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
```diff
@@ -0,0 +1,15 @@
...
+64x896x4864
+
+# --reset
+# --dt=f16,bf16
+# --activation=swish,gelu_erf,gelu_tanh
+# --batch=shapes_basic
+#
+# --reset
+# --dt=f16
+# --batch=shapes_llm
```
hidefromkgb: There has to be a second set with these shapes where all weights are u4 and have scales/ZPs applied to them.

tests/benchdnn/doc/driver_gated_mlp.md
kwieloch-intel: Fixed.

tests/benchdnn/doc/driver_gated_mlp.md
kwieloch-intel: Done.

tests/benchdnn/doc/driver_gated_mlp.md
kwieloch-intel: @hidefromkgb This is a direct copy of the convention used in other drivers. Knob `--match` is a standard option, documented the same way across all drivers, so removing it would break consistency. Should I remove it anyway?

tests/benchdnn/gated_mlp/ref_gated_mlp.cpp
kwieloch-intel: Done.

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
kwieloch-intel: I think quantized variant (u4 + grouped scales/ZPs) support was planned as a follow-up once the base driver infrastructure lands and is validated with the current f16/bf16/f32 coverage but because of dequantization questions I've added a preliminary version on a experimental basis here.

tests/benchdnn/doc/driver_gated_mlp.md
hidefromkgb: Okay. Mea culpa.
Resolved.

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
```diff
@@ -0,0 +1,28 @@
...
+
+# Quantized weights: u4 with per_oc scales
+--reset
+--dt=f16:u4:u4:u4:f16
+--attr-scales=wei_gate:per_oc:f16+wei_up:per_oc:f16+wei_down:per_oc:f16
```
hidefromkgb: 1. These tests need `--attr-fpmath=any:true` because all of the implementations involve GEMMs that operate on FP×INT inputs; otherwise the GEMMs would simply fail.
1. Also of importance is `--wtag=ba`, since this is the only option that would let `ocl:micro_horz` be selected instead of `ocl:ref`.
1. And, finally, `per_oc` here is incredibly misleading, since OC is not the same dimension in Down vs Gate/Up; see my suggestion below.

```suggestion
--stag=ab
--wtag=ba
--dtag=ab
--attr-fpmath=any:true
--attr-scales=wei_gate:per_dim_01:f16:896x1+wei_up:per_dim_01:f16:896x1+wei_down:per_dim_01:f16:896x1
```

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
```diff
@@ -0,0 +1,28 @@
...
+64x896x4864
+
+# Quantized weights: u4 with per_oc scales and zero-points
+--reset
+--dt=f16:u4:u4:u4:f16
+--attr-scales=wei_gate:per_oc:f16+wei_up:per_oc:f16+wei_down:per_oc:f16
+--attr-zero-points=wei_gate:per_oc:s4+wei_up:per_oc:s4+wei_down:per_oc:s4
```
hidefromkgb: The comment above also applies to zero points; also, we need some grouped quantization cases — I'm suggesting `128x1` instead of `896x1`.

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_ci
```diff
@@ -0,0 +1,11 @@
+# GPU gated_mlp bug: executing multiple primitives with different shapes
+# in one process produces incorrect results.
+# TODO: uncomment full test set when the GPU implementation is fixed.
```
hidefromkgb: Debugging complete.
Turns out I forgot to add `CASE(gated_mlp)` to `serialize_desc()` when adding GMLP to oneDNN.
The fix is [ready](5192).

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
kwieloch-intel: Thanks for the suggestion! All three points applied: `[--attr-fpmath=any:true]`, `[--wtag=ba]`, and `per_dim_01:...:128x1` for grouped quantization (both scales and zero-points).

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
kwieloch-intel: Resolved 0fda9ac022c851556f8fc87cd6cd7ab3a7eb7fe5.

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_ci
kwieloch-intel: Great! Thanks for tracking this down! Once #5192 lands, I'll rebase and uncomment the full multi-shape test suites.

hidefromkgb: Merged!

kwieloch-intel: Rebased on top of #5192. Full multi-shape test suites are now uncommented and **passing**.

tests/benchdnn/gated_mlp/gated_mlp.cpp
```diff
@@ -0,0 +1,286 @@
...
+    }
+}
+
+void setup_cmp(compare::compare_t &cmp, const prb_t *prb, data_kind_t kind,
+        const args_t &ref_args) {
+    // Thresholds are empirical; 3 chained matmuls + activation compound errors.
+    // Factor of 384 = 64 * 6 accounts for error propagation through 3 matmuls
+    // and element-wise ops with non-deterministic GPU accumulation order.
+    const int64_t max_acc = std::max(prb->ic, prb->oc);
+    const float trh = 384.f * (1 + max_acc) * epsilon_dt(prb->dst_dt());
```
hidefromkgb: This threshold is so high that it doesn't, in fact, check the output for correctness.
I just fixed an error in the microkernel implementation of GMLP that benchdnn didn't even register as an error because of this threshold.
I believe that, with GMLP, relative diff (±1%) and a threshold on the number of mismatches (0.01% of DST size) would be a much better metric.

kwieloch-intel: Reduced threshold to 5% relative. My GPU measurements (non-deterministic accumulation order) show up to ~2% rel_diff for f32, so 1% would cause false failures. The 5% threshold + absolute driver_check for f16/bf16 should still catch real implementation bugs while avoiding GPU false positives. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/551a0c4ecbd58c8d76b1406e125286e674931e91

hidefromkgb:
Here are some more [fixes](https://github.com/user-attachments/files/28367378/benchdnn_fixes.patch).
Firstly, the extent of CPU memory duplication is reduced.
Secondly, `up_result` is now the Gate/Up destination buffer instead of `gate_result`, which helps with more involved debugging when only the first GEMM is performed.

hidefromkgb:
False negative:
```
--dt=f16:u4:u4:u4:f16 --wtag=ba --attr-fpmath=any:true --attr-scales=wei_gate:per_dim_01:f16:896x1+wei_up:per_dim_01:f16:896x1+wei_down:per_dim_01:f16:128x1 --impl=ref 64x896x4864
```
Result:
```
[46735][DST][52:143] exp_f32:     65520.3 exp:         inf got:       65504 diff:     inf rdiff:    -nan
[COMPARE_STATS][DST]: trh=0.05 err_max_diff:     inf err_max_rdiff:    -nan all_max_diff:     inf all_max_rdiff:0.000969932
0:FAILED (errors:1 total:57344) (3021 ms) __REPRO: --gated_mlp --engine=gpu --dt=f16:u4:u4:u4:f16 --wtag=ba --attr-scales=wei:per_dim_01:f16:896x1+wei_up:per_dim_01:f16:896x1+wei_down:per_dim_01:f16:128x1 --attr-fpmath=any:true --impl=ref 64x896x4864
```

hidefromkgb:
Please also add a parser for `gated_mlp` to `verbose_converter.py`.

hidefromkgb:
OK, here's my approval.

Please note though that the false negatives have to be addressed before the PR gets merged.

One of the error patterns, the only one I could reliably identify, is as follows: one of the buffers contains a value that is several ULP short of ±INF and the corresponding value from the other buffer has crossed that boundary and became ±INF — which leads to both abs_delta and rel_delta diverging to infinity.

There are others, though — e.g. this case yields 835 errors, some with 20%+ relative delta:
```
--dt=f16:u4:u4:u4:f16 --wtag=ba --attr-fpmath=any:true --attr-scales=wei_gate:per_dim_01:f16:128x1+wei_up:per_dim_01:f16:128x1+wei_down:per_dim_01:f16:128x1 --attr-zero-points=wei_gate:per_dim_01:s4:128x1+wei_up:per_dim_01:s4:128x1+wei_down:per_dim_01:s4:128x1 1024x2048x4096
```

hidefromkgb (approved):

tests/benchdnn/dnn_types.cpp
```diff
@@ -94,7 +94,9 @@ const char *prop2str(dnnl_prop_kind_t prop) {
 static const std::map<int, std::vector<const char *>> supported_args {
         {DNNL_ARG_SRC, {"src", "src0"}},
         {DNNL_ARG_SRC_1, {"src1"}},
-        {DNNL_ARG_WEIGHTS, {"wei"}},
+        {DNNL_ARG_WEIGHTS, {"wei", "wei_gate"}},
```
hidefromkgb:
```suggestion
        {DNNL_ARG_WEIGHTS, {"wei_gate"}},
```
— just to avoid confusion when comparing tests with their verbose_converter counterparts.

tests/benchdnn/gated_mlp/cfg.cpp
```diff
@@ -0,0 +1,86 @@
...
+    return density;
+}
+
+cfg_t::cfg_entry_t::cfg_map_t cfg_t::get_cfg_map(data_kind_t kind) const {
+    // Source tensor ranges, kept small to prevent overflow in the chained
+    // matmul pipeline (3 matmuls with an element-wise multiply in between).
+    static const cfg_t::cfg_entry_t::cfg_map_t src_cfg_map = {
+            {{dnnl_f32}, {-2, 2}},
+            {{dnnl_bf16}, {-1, 1}},
+            {{dnnl_f16}, {-1, 1}},
```
hidefromkgb, line 55: According to OpenVINO we need INT4/INT8 SRC as well.
Please also add INT32/INT4/INT8 DST.

tests/benchdnn/gated_mlp/ref_gated_mlp.cpp
```diff
@@ -0,0 +1,221 @@
...
+        const auto &zp_groups = has_zp ? attr.zero_points.get(wei_arg).groups
+                                       : std::vector<dnnl_dim_t> {};
+        auto retn = make_2d(eng, K, N);
+        std::memcpy((float *)retn, (float *)w, K * N * sizeof(float));
+        dequantize_2d((float *)retn, K, N, sc, zp, has_scale, has_zp, sc_mask,
+                zp_mask, sc_groups, zp_groups);
+        return retn;
+    };
+
+    auto w_gate_ref = dequant_wei(w_gate_m, IC, OC, DNNL_ARG_WEIGHTS_GATE);
```
hidefromkgb: According to the new data from OV, scales/ZPs can also be applied to SRC, but the cases they shared with us only include SRC scales, not ZPs — I hope it stays that way.
Please leave the possibility of SRC ZPs open, just in case, but don't add SRC ZP tests.

hidefromkgb:
> Known GPU implementation bug
> 
> The GPU ocl:ref:any gated_mlp implementation has a bug: executing multiple primitives with different shapes in the same process produces incorrect results (NaN, garbage) or CL_INVALID_KERNEL_ARGS (errcode -52). Each shape works correctly in isolation. This is a bug in src/gpu/intel/gated_mlp/ref.hpp, not in this driver.
> 
> The CI test files are limited to one shape per file as a workaround. Full multi-shape test configurations are included as comments with a TODO to uncomment once the GPU implementation is fixed.

For the record: as of 2026-05-29 this bug has been [fixed](5192).

tests/benchdnn/dnn_types.cpp
hidefromkgb: This isn't the place to add these highly specialized names; there are other oneDNN primitives that use custom wei types, but none of these types are listed here.

tests/benchdnn/doc/driver_gated_mlp.md
```diff
@@ -0,0 +1,99 @@
...
+## Quantization Attributes
+
+The driver supports `--attr-scales` and `--attr-zero-points` for weight
+dequantization. Supported argument names are `wei_gate`, `wei_up`, and
+`wei_down`, corresponding to the gate, up, and down projection weights.
+
+Supported policies:
+- `common` -- single scale/zero-point shared across the entire tensor.
+- `per_oc` -- one value per output channel (dimension 1 of the weight tensor).
+- `per_ocic` -- grouped quantization along both dimensions (group size specified
```
hidefromkgb: `per_dim_*`, as discussed earlier.

tests/benchdnn/doc/driver_gated_mlp.md
```diff
@@ -0,0 +1,99 @@
...
+            double quotation marks.
+
+and *gated_mlp-desc* is a problem descriptor. The canonical form is:
+```
+    MBxICxOC
+```
+Here `x` is the delimiter for the three dimensions: `MB` (batch size), `IC`
+(input channels / model dimension), and `OC` (intermediate dimension). All
+tensor shapes are derived from these three values:
+- Source: `[MB, IC]`
```
hidefromkgb, line 37: OV networks also use 'fake 3D' shapes: MB×1×IC for SRC/DST, 1×IC×OC for W_GATE/W_UP, 1×OC×IC for W_DOWN.
The user needs to be able to request fake 3D shapes — e.g. by passing `abc` and `cab` in tags instead of `ab` and `ba` respectively.

tests/benchdnn/dnn_types.cpp
kwieloch-intel: Fixed. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/08e50d7bbdd432c19e26d3c537b69116effaaf7e

tests/benchdnn/gated_mlp/cfg.cpp
kwieloch-intel: Done. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/b17ca1bcba943368d10c9cdc815a0a6295837adb

tests/benchdnn/gated_mlp/ref_gated_mlp.cpp
kwieloch-intel: Addressed. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/05fb93f9bba9afa71ddf7955c27ddd14c09f913d

tests/benchdnn/doc/driver_gated_mlp.md
kwieloch-intel: Done, updated to `per_dim_1`/`per_dim_01`. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/ada781ae536f2903a67ce18ee1792e6f07cc947c

tests/benchdnn/doc/driver_gated_mlp.md
kwieloch-intel: Done. The driver now supports fake 3D shapes via explicit 3D tags (`--stag=abc --wtag=cab --dtag=abc`). The primitive currently returns `UNIMPLEMENTED` for ndims≠2, once the PD is extended, these cases will pass through. https://github.com/uxlfoundation/oneDNN/pull/4951/commits/e3645a342113d4ff616adb1f691363afacc4ace5

hidefromkgb:
Another [batch of fixes](https://github.com/user-attachments/files/28487004/benchdnn_fixes.patch), this time mostly for the dequantization routine.
Coupled with #5221 the GMLP benchdnn driver can be called mostly working — if we forget for a minute that on large tests the results still diverge.

hidefromkgb:
Third [batch of fixes](https://github.com/user-attachments/files/28661375/benchdnn_fixes.patch): post-ops.
To properly execute ref GMLP with Binary post-ops (unlike Sum and Eltwise), the fix below should be applied — which is out of scope for this PR and will be added separately:
```diff
diff --git a/src/common/gated_mlp_pd.hpp b/src/common/gated_mlp_pd.hpp
index 5f020fa8ad..71cb3fd242 100644
--- a/src/common/gated_mlp_pd.hpp
+++ b/src/common/gated_mlp_pd.hpp
@@ -76,7 +76,7 @@ struct gated_mlp_pd_t : public primitive_desc_t {

     int n_outputs() const override { return 1; }
     int n_inputs() const override {
-        return int(all_idxs().size()) - n_outputs();
+        return int(all_idxs().size()) - n_outputs() + n_binary_po_inputs();
     }
     const op_desc_t *op_desc() const override {
         return reinterpret_cast<const op_desc_t *>(this->desc());
```

The issue of false-negative and false-positive results also has to be addressed — e.g., removing these 3 lines from the patch still yields a pass with the case below, despite the values of the post-op buffer being complete garbage, read as FP16 instead of FP32:
```diff
+    for (auto &po : down_attr.post_ops.entry)
+        if (po.is_binary_kind())
+            po.binary.src1_dt = dnnl_f32;
```
```
--engine=gpu --gated_mlp --dt=s8:u4:u4:u4:f16 --stag=ab --wtag=ba --dtag=ab --attr-scales=wei_down:common:0.00125 --attr-post-ops=binary_add:f16:3:ab --impl=ref 64x256x4864
```
Please don't ignore this issue, as it is extremely serious — in fact, in the eyes of the Arch team it's disqualifying for the whole driver and will never get this PR a positive review from Arch.

dzarukin:
I'll approve it as long as it doesn't update the common infrastructure at the final stage.

It just proves Arch point that building a separate driver is as much difficult as it was building the graph driver validation, just doubles the effort and doesn't give much back. ROI is very low. Those two or three weeks of delay from Graph API to promote a new feature side will be constantly invested into this driver development and debugging which would simply kill the purpose that this driver was assumed on. It may not happen always but I anticipate it happens much more often that it seemed initially.

tests/benchdnn/inputs/gated_mlp/test_gated_mlp_gpu
```diff
@@ -28,4 +28,13 @@
 --attr-scales=wei:per_dim_01:f16:128x1+wei_up:per_dim_01:f16:128x1+wei_down:per_dim_01:f16:128x1
 --attr-zero-points=wei:per_dim_01:s4:128x1+wei_up:per_dim_01:s4:128x1+wei_down:per_dim_01:s4:128x1
 64x128x4864
-64x896x4864
```
hidefromkgb: The ability to test arbitrary user-supplied cases, with arbitrary shapes, for correctness is kinda the entire point of a benchdnn driver.

tests/benchdnn/gated_mlp/gated_mlp.cpp
```diff
@@ -252,6 +253,20 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
                 }
                 break;
             default:
+                // For int SRC types, multiply SRC scales by coeff to prevent
+                // overflow (buffer multiply would round to zero for integers).
+                if ((exec_arg & DNNL_ARG_ATTR_SCALES)
+                        && (exec_arg ^ DNNL_ARG_ATTR_SCALES) == DNNL_ARG_SRC) {
```
hidefromkgb: What if SRC has no scales on it but W_GATE/W_UP do?

kwieloch-intel: Good catch. If SRC is integral without its own scale, the coeff is now applied to WEI gate/up scales instead. Added a test case for `s8:u4:u4:u4:f16` with only WEI scales (no SRC scale). https://github.com/uxlfoundation/oneDNN/pull/4951/commits/db45226cb842ca8ccef5ae4255e237f7b7df9843

hidefromkgb:
Please modify the driver according to the latest upstream changes (e.g. the introduction of `base_prb_t`).

vpirogov:
I think this one can be put to rest considering the change of direction in GatedMLP block optimization. As we are not introducing internal primitive there's no need for benchdnn driver.

kwieloch-intel:
> I think this one can be put to rest considering the change of direction in GatedMLP block optimization. As we are not introducing internal primitive there's no need for benchdnn driver.

Closing this PR due to the **updated direction for GatedMLP optimization**. As we are no longer pursuing the internal primitive approach, the functionality introduced here is no longer required.

PR #4974, opened 2026-04-08
cpu: aarch64: implement jitted lnorm in SVE and ASIMD

Sqvid:
# Description

Implement forward `lnorm` in SVE and ASIMD.

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/jit_generator.hpp
+++ b/src/cpu/aarch64/jit_generator.hpp
@@ -416,6 +416,20 @@ public:
         fadd(dst, src, src2);
     }
 
+    // Float addition across all vector elements
+    void uni_fadd_reduce(
+            const Xbyak_aarch64::SReg &dst, const Xbyak_aarch64::ZRegS &src) {
+        faddv(dst, P_ALL_ONE / Xbyak_aarch64::T_z, src);
+    }
+
+    void uni_fadd_reduce(
+            const Xbyak_aarch64::SReg &dst, const Xbyak_aarch64::VReg4S &src) {
+        const auto &vreg_dst = Xbyak_aarch64::VReg(dst.getIdx());
+
+        faddp(vreg_dst.s, src, src);
+        faddp(vreg_dst.s2, vreg_dst.s2, vreg_dst.s2);
+    }
+
     void uni_fcvtzs(
             const Xbyak_aarch64::VReg4S &d, const Xbyak_aarch64::VReg4S &s) {
         fcvtzs(d, s);
@@ -471,6 +485,18 @@ public:
             const Xbyak_aarch64::VReg4S &src2) {
         fmul(dst, dst, src);
         fadd(dst, dst, src2);
+    }
+
+    void float_point_fused_multiply_add(const Xbyak_aarch64::ZReg &dst_vmm,
+            const Xbyak_aarch64::ZReg &src1_vmm,
+            const Xbyak_aarch64::ZReg &src2_vmm) {
+        fmla(dst_vmm.s, P_ALL_ONE / Xbyak_aarch64::T_m, src1_vmm.s, src2_vmm.s);
+    }
+
+    void float_point_fused_multiply_add(const Xbyak_aarch64::VReg &dst_vmm,
+            const Xbyak_aarch64::VReg &src1_vmm,
+            const Xbyak_aarch64::VReg &src2_vmm) {
+        fmla(dst_vmm.s, src1_vmm.s, src2_vmm.s);
     }
 
     template <typename T>
@@ -529,18 +555,6 @@ public:
     template <typename T>
     void uni_fmul(const T &dst, const T &src, const T &src2) {
         fmul(dst, src, src2);
-    }
-
-    void float_point_fused_multiply_add(const Xbyak_aarch64::ZReg &dst_vmm,
-            const Xbyak_aarch64::ZReg &src1_vmm,
-            const Xbyak_aarch64::ZReg &src2_vmm) {
-        fmla(dst_vmm.s, P_ALL_ONE / Xbyak_aarch64::T_m, src1_vmm.s, src2_vmm.s);
-    }
-
-    void float_point_fused_multiply_add(const Xbyak_aarch64::VReg &dst_vmm,
-            const Xbyak_aarch64::VReg &src1_vmm,
-            const Xbyak_aarch64::VReg &src2_vmm) {
-        fmla(dst_vmm.s, src1_vmm.s, src2_vmm.s);
     }
 
     void contiguous_load_unsigned_words(const Xbyak_aarch64::ZReg &src_vmm,
--- a/src/cpu/aarch64/jit_uni_layer_normalization.cpp
+++ b/src/cpu/aarch64/jit_uni_layer_normalization.cpp
@@ -1,0 +1,648 @@
+/*******************************************************************************
+* Copyright 2019 Intel Corporation
+* Copyright 2026 Arm Ltd. and affiliates
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cassert>
+#include <cmath>
+
+#include "common/c_types_map.hpp"
+#include "common/dnnl_thread.hpp"
+#include "common/reorder.hpp"
+#include "common/stream.hpp"
+#include "common/type_helpers.hpp"
+#include "cpu/platform.hpp"
+
+#include "cpu/aarch64/injectors/jit_uni_postops_injector.hpp"
+#include "cpu/aarch64/jit_generator.hpp"
+#include "cpu/aarch64/jit_uni_layer_normalization.hpp"
+#include "cpu/aarch64/utils/jit_io_helper.hpp"
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace aarch64 {
+
+using namespace data_type;
+using namespace memory_tracking::names;
+using namespace Xbyak_aarch64;
+
+namespace {
+
+bcast_set_t get_supported_bcast_strategies(int ndims) {
+    assert(ndims > 1 && ndims <= 5);
+    bcast_set_t set {broadcasting_strategy_t::scalar};
+    switch (ndims) {
+        case 2: set.insert(broadcasting_strategy_t::per_oc); break;
+        case 3:
+        case 4:
+        case 5: set.insert(broadcasting_strategy_t::per_w); break;
+        default: assert(!"Unsupported ndims");
+    }
+    return set;
+}
+
+template <cpu_isa_t isa>
+class jit_lnorm_stat_and_data_kernel_t : public stat_and_data_kernel_iface_t,
+                                         public jit_generator_t {
+public:
+    DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_lnorm_stat_and_data_kernel_t)
+
+    using TReg = typename cpu_isa_traits<isa>::TReg;
+
+    jit_lnorm_stat_and_data_kernel_t(const layer_normalization_pd_t *pd)
+        : jit_generator_t(nullptr, MAX_CODE_SIZE, true)
+        , pd_(pd)
+        , src_d_(pd_->src_md())
+        , dst_d_(pd_->dst_md())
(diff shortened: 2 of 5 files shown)


src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+        ldr(reg_src_scales_, ptr(reg_param_, PARAM_OFF(src_scales)));
+        ldr(reg_dst_scales_, ptr(reg_param_, PARAM_OFF(dst_scales)));
+        ldr(reg_block_end_, ptr(reg_param_, PARAM_OFF(block_size)));
+        add_imm(X_DEFAULT_ADDR, reg_param_, PARAM_OFF(eps), X_TMP_0);
+        ldr(W_TMP_0, ptr(X_DEFAULT_ADDR));
+#undef PARAM_OFF
+
+        dup(z_eps_.s, W_TMP_0);
+        fmov(z_ones_.s, 1.0);
+        mov(W_TMP_0, float2int(static_cast<float>(C_)));
```
jondea, line 388: Given that we are going to be using `1/C`, it would be better to calculate `1/C` at generate time so that we can replace the fdivs with fmul

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+        if (axis_simd_tail_) {
+            io_->at(src_d_.data_type())
+                    ->load(addr_with_offt(reg_src_,
+                                   axis_simd_full_ * simd_w_
+                                           * src_d_.data_type_size()),
+                            0, z_tmp_, true);
+            fadd(z_acc_.s, z_acc_.s, z_tmp_.s);
+        }
+        hreduce_sum(z_acc_);
+        uni_fdiv(z_mean_.s, z_acc_.s, z_c_.s, z_tmp_.s, P_ALL_ONE);
```
jondea, line 333: This may not make much difference in the grand scheme of things, but complex iterative instructions like fdiv and sqrt are faster when done as scalars. I think it would also be clearer for these to be scalars at this point (even if you do dup them later becaue you want to apply them across lanes)

It would also be clearer because you could use faddv directly

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+                    addr_with_offt(reg_shift_, offt_elems * sizeof(float)), 0,
+                    z_shift_, tail);
+        }
+
+        io_->at(src_d_.data_type())
+                ->load(addr_with_offt(
+                               reg_src_, offt_elems * src_d_.data_type_size()),
+                        0, z_dst_, tail);
+
+        if (!skip_mean_) fsub(z_dst_.s, z_dst_.s, z_mean_.s);
```
jondea, line 290: Can any of the sub/mul/adds be fused into fmla/fmls?

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+        args.src = src;
+        args.dst = dst;
+        args.scale = scale;
+        args.shift = shift;
+        args.mean = mean;
+        args.var = var;
+        args.src_scales = src_scales;
+        args.dst_scales = dst_scales;
+        args.post_ops_binary_rhs_arg_vec = post_ops_binary_rhs_arg_vec;
+        args.block_size
```
jondea, line 158: This is confusing, this variable means diferent things depending on where it is. Can we disambiguate the two kinds of blocks?

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+                    get_supported_bcast_strategies(dst_d_.ndims()),
+                    binary_rhs_params};
+
+            postops_injector_ = utils::make_unique<
+                    injector::jit_uni_postops_injector_t<isa>>(this,
+                    pd_->attr()->post_ops_, binary_params, eltwise_params);
+        }
+    }
+
+    void operator()(const void *src, void *dst, const float *scale,
```
jondea, line 143: It doesn't feel like this wrapper does much, do we need it?

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+                    == status::success,
+            VERBOSE_INCONSISTENT_MDS, "src", "stat");
+
+    if (reordered_stat_md_ != *stat_md() && !stats_are_tmp()) {
+        CHECK(reorder_primitive_desc_create(reorder_pd_, engine,
+                stats_are_src() ? stat_md() : &reordered_stat_md_,
+                stats_are_src() ? &reordered_stat_md_ : stat_md()));
+    }
+
+    nthr_ = dnnl_get_max_threads();
```
jondea, line 489: If I remember correctly, this can change between init and execute time. I think it's a good hint, but we do need to re-check when we execute

src/cpu/aarch64/jit_uni_layer_normalization.cpp
```diff
@@ -0,0 +1,639 @@
...
+
+        (*stat_and_data_kernel_)(src_ptr, dst_ptr, scale, shift, mean_ptr,
+                &variance[N_start], src_scales, dst_scales_inv_ptr,
+                post_ops_binary_rhs_arg_vec.data(), block_size);
+    });
+
+    return status::success;
+}
+
+template struct jit_uni_layer_normalization_fwd_t<sve_512>;
```
jondea, line 632: I would be eager to try to use the VLA `isa` class if we can

Sqvid: Done.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: I am not sure if `fsub+fmul` can be fused? I know `fmul+fsub/fadd` can and I think I have now fused these where possible.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: Done. Thanks for pointing this out!

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: I tried this but it seems to not be as numerically stable.

Sqvid:
I will need to redo the benchmarks as both this implementation and the ACL primitive have changed since the initial round. Will post those soon but it is **expected** that this version be significantly slower than ACL since that kernel is actually not numerically stable due to it [only doing a single pass over the data](https://github.com/ARM-software/ComputeLibrary/blob/3add408310176222330f6eb7c36688b305241964/src/cpu/kernels/meanstddevnorm/generic/neon/impl.cpp#L60-L65). A discussion on the numerical consequences of doing this can be found in [_Algorithms for computing the sample variance: analysis and recommendations_ (Chan, Golub, LeVeque)](https://engineering.yale.edu/application/files/3817/3714/7395/tr222.pdf)

src/cpu/aarch64/jit_uni_layer_normalization.cpp
jondea: Your notation is ambiguous, but yeah I think I agree `fmul(fsub(a,b),c)` is not fusable

jondea:
Good find. I think for now we can accept the regression if the new algorithm is stable. We can then work to improve the performance if needed. I'm surprised this has not bit us already, the current algorithm can be arbitrarily bad.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
jondea: I'm surprised that this was noticable, but given that it is applied once, I think we can err on the side of numerical accuracy.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: Done. De-duplicated.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: Done. Disambiguated.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
Sqvid: I don't think we have to re-check here. The reason we store `nthr_` at configure time is because we decide scratchpad memory based on that value:
```cpp
            if (!attr()->scales_.has_default_values(DNNL_ARG_DST)) {
                scratchpad.book(key_lnorm_dst_scales,
                        static_cast<size_t>(nthr_) * sizeof(float), 64);
            }
```
So at runtime we want to run with the configure-time value regardless of whether it was changed in the meantime. Of course there will be performance penalty if the user tweaks the max thread count between primitive creation and execution but we assume they know what they are doing if they are calling `omp_set_num_threads()` directly.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -416,6 +416,20 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
         fadd(dst, src, src2);
     }
 
+    // Float addition across all vector elements
+    void uni_hreduce(
```
jondea: `hreduce` is an odd name. I'm not sure what the h means but if it means horizontal then vectors only have one dimension so it feels redundant. Also, is it doing floating point or integer maths?

I think `uni_fadd_reduce` would probably be clearest.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -473,6 +487,18 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
         fadd(dst, dst, src2);
     }
 
+    void float_point_fused_multiply_add(const Xbyak_aarch64::ZReg &dst_vmm,
```
jondea, line 490: Can this just be called `fmla`? I think it would be more recognisable and specific

jondea: Oh, this has just been moved. Feel free to ignore

src/cpu/cpu_layer_normalization_list.cpp
```diff
@@ -48,6 +49,8 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
     static const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> the_map = REG_LNORM_P({
         {{forward}, {
             CPU_INSTANCE_X64(jit_uni_layer_normalization_fwd_t)
+            CPU_INSTANCE_AARCH64(jit_uni_layer_normalization_fwd_t<sve>)
```
jondea, line 52: If we are putting it above ACL and it implements everything ACL does, can we remove the ACL impl? If not in this PR, it needs doing at some point anyway, and I'd prefer to do it while we are confident we have matched it in functionality and perf.

dzarukin (approved):

src/cpu/cpu_layer_normalization_list.cpp
Sqvid: This primitive does not yet support `f16` (It will once I have time to cleanup https://github.com/uxlfoundation/oneDNN/pull/5092), while `acl_layer_normalization_t` does, which is why I've kept it around.

src/cpu/aarch64/jit_generator.hpp
Sqvid: I got bit by the fact that `asimd fadd != sve faddv`. In fact there is no 1-2-1 equivalent instruction, which is why I went for a neutral third name. Would `uni_fadd_reduce` suggest we are using `fadd`? This is in floating-point math.

jondea: Your functions _is_ a series of floating point adds, with no promises of the order of operations. Both faddv and faddp have fadd in them. Also, hreduce doesn't tell you what binary operation you are reducing with. You could reduce using e.g. max or mul.

Sqvid: Point taken, I'll rename it as you suggested. Final note though, should I keep it as a general utility in `jit_generator` or do you think it is niche enough to be scoped to `lnorm`?

jondea: Definitely a general utility, there are a lot of places it could be used. E.g. the JIT GEMV microkernel or there's a whole set of reduction primitives too.

src/cpu/aarch64/jit_uni_layer_normalization.cpp
jondea: Okay, so it would be safe, but it may be slower (e.g. running 9 tasks on 8 CPUs will cause overhang).

src/cpu/aarch64/jit_generator.hpp
Sqvid: Done!

jondea (approved):

PR #4977, opened 2026-04-08
cpu, benchdnn: add reorder to/from grouped with different dts and use grouped in matmul ref

mzhukova:
TBD


mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_reorder

PR #4987, opened 2026-04-09
aarch64: support for per_dim_0 scales and bf16 dst_dt in jit int8 matmul

michalowski-arm:
# Description

This change adds AArch64 jit:int8 matmul support for row-wise source scales (src:per_dim_0) and bf16 destination output.

The original motivation for this change is to improve W8A8 serving performance for workloads such as Llama and Whisper in vLLM. In these models, activations are symmetrically quantized with per-token scales, which map to matmul src:per_dim_0 scales (the matmul M dimension).

Before this change, the AArch64 jit:int8 matmul path did not support this case fully, so vLLM had to run an additional epilogue to apply the activation scales and convert the result to bf16. With this PR, the scale application and bf16 destination handling can stay inside the oneDNN matmul path, removing that extra epilogue and reducing extra memory traffic and output-side work.

In vLLM testing, this improved output-token throughput by roughly 5–10%, depending on the model.

## Checklist

### DIFF
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2019 Intel Corporation
+* Copyright 2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -299,7 +300,7 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
             const int mask_src = sc.get_mask(DNNL_ARG_SRC);
 
             VCHECK_MATMUL_UNIMPL(
-                    utils::one_of(mask_src, 0, src_qmask_K,
+                    utils::one_of(mask_src, 0, src_qmask_M, src_qmask_K,
                             src_qmask_M + src_qmask_K, full_tensor_mask),
                     VERBOSE_UNSUPPORTED_SCALES_CFG);
 
--- a/src/common/matmul_pd.hpp
+++ b/src/common/matmul_pd.hpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
 * Copyright 2019 Intel Corporation
+* Copyright 2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -18,6 +19,7 @@
 #define COMMON_MATMUL_PD_HPP
 
 #include <assert.h>
+#include <map>
 
 #include "oneapi/dnnl/dnnl.h"
 
@@ -199,9 +201,19 @@ struct matmul_pd_t : public primitive_desc_t {
     virtual bool attr_scales_ok(const std::vector<int> &supported_args
             = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST},
             const std::vector<int> &supported_qmodes
-            = {quantization_mode::static_sazp}) const {
+            = {quantization_mode::static_sazp},
+            const std::map<int, std::vector<int>> &extra_masks = {}) const {
         const auto &scales = attr()->scales_;
         if (scales.has_default_values()) return true;
+
+        const auto extra_mask_ok = [&](int arg, int mask) {
+            const auto it = extra_masks.find(arg);
+            if (it != extra_masks.end()) {
+                for (const auto &extra_mask : it->second)
+                    if (mask == extra_mask) return true;
+            }
+            return false;
+        };
 
         bool ok = scales.has_default_values(supported_args);
         for (int arg : supported_args) {
@@ -238,10 +250,13 @@ struct matmul_pd_t : public primitive_desc_t {
                                     (mask & wei_qmask_K()), is_decompression);
                 }
             } else if (arg == DNNL_ARG_SRC) {
+                // Masks supported across all implementations. Implementation
+                // specific masks can be passed through `extra_masks`.
                 ok = ok
-                        && utils::one_of(mask, 0, src_qmask_K(),
-                                src_qmask_M() + src_qmask_K(),
-                                full_tensor_mask());
+                        && (utils::one_of(mask, 0, src_qmask_K(),
+                                    src_qmask_M() + src_qmask_K(),
+                                    full_tensor_mask())
+                                || extra_mask_ok(arg, mask));
                 ok = ok
                         && IMPLICATION((mask & src_qmask_K()),
                                 !scales.get(arg).has_default_groups());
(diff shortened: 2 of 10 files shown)


src/common/matmul.cpp
```diff
@@ -268,7 +269,7 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
             const int mask_src = sc.get_mask(DNNL_ARG_SRC);
 
             VCHECK_MATMUL_UNIMPL(
-                    utils::one_of(mask_src, 0, src_qmask_K,
+                    utils::one_of(mask_src, 0, src_qmask_M, src_qmask_K,
```
dzarukin, line 303: Could you, please, share the use case for such quantization?

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -719,6 +737,30 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
+            if (!supported) {
+                BENCHDNN_PRINT(2,
+                        "[SKIP][%s:%d]: src:per_dim_0 is only supported "
+                        "for AArch64 SVE128/SVE256 jit:int8 matmul.\n",
+                        __FILE__, __LINE__);
+                res->state = SKIPPED;
+                res->reason = skip_reason::case_not_supported;
+                return;
+            }
+        }
```
dzarukin: The preferred way of doing things is to enable reference (shared implementations between backends) to avoid multiple parties handling the case that is going to be supported in a single point of a single backend.

dzarukin (commented):
Please share the motivation of extending functional capabilities and confirm the readiness to enable it uniformly across the stack. The further guidance will follow based on those answers. Thank you.

src/common/matmul.cpp
michalowski-arm: I added motivation to the PR description, but in essence this is needed to support per-token scales in the dynamic quantization path used by vLLM.

tests/benchdnn/matmul/matmul.cpp
michalowski-arm: That makes sense. I can rework it to use ref_matmul_int8 as the generic fallback instead. If I’m not wrong, the support is already there and the case is only rejected during pd creation. Would that work?

src/common/matmul.cpp
dzarukin: Thanks, it seems like a grouped effort to enable new flavors of vllm. resolving.

tests/benchdnn/matmul/matmul.cpp
dzarukin: Yes, it will.
Shared parts besides touched also include `matmul_pd::attr_scales_ok()` function and ref_matmul.{c,h}pp files.

michalowski-arm: Let me know if the approach I took with the latest commit works.

src/common/matmul_pd.hpp
```diff
@@ -199,7 +200,8 @@ struct matmul_pd_t : public primitive_desc_t {
     virtual bool attr_scales_ok(const std::vector<int> &supported_args
             = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST},
             const std::vector<int> &supported_qmodes
-            = {quantization_mode::static_sazp}) const {
+            = {quantization_mode::static_sazp},
+            bool allow_src_per_m = false) const {
```
dzarukin: It seems to me the signature of this function is not very suitable for mask support extension in one or two specific matmul implementations and should be further extended.

If you need to wrap up this activity earlier, I suggest to expand implementation local conditions and let new mask in.
If you are up to take an extra mile, I think the idea below should suit for the future:
```
// This function covers a common ground across ALL implementations. Masks that are supported
// by exclusive implementations can be passed through `extra_masks`. It takes an elements as pairs of
// `{arg, {1, 5, ...}}` and will be additionally checked.
virtual bool attr_scales_ok(const std::vector<int> &supported_args
            = {DNNL_ARG_SRC, DNNL_ARG_WEIGHTS, DNNL_ARG_DST},
            const std::vector<int> &supported_qmodes
            = {quantization_mode::static_sazp},
            const std::map<int, std::vector<int>> &extra_masks = {}) {
            ...
            // Masks supported in all implementations.
            bool mask_ok = utils::one_of(mask, 0, src_qmask_K(),
                                src_qmask_M() + src_qmask_K(),
                                full_tensor_mask());
            // If mask passed wasn't found, let check extra masks coming from the impl.
            if (!mask_ok) {
                if (extra_masks.find(arg) != extra_masks.end()) {
                    for (auto &em : extra_masks.at(arg).second)
                        if (mask == em) mask_ok = true;
                }
            }
            ...
```

Having it this way should prevent from updating other backends without affecting their testing results and support capabilities.

tests/benchdnn/dnn_types.cpp
```diff
@@ -243,6 +244,7 @@ int attr_t::policy2mask(int arg, policy_t policy, int ndims,
 
         if (ndims < 2) SAFE_V(FAIL);
         switch (policy) {
+            case PER_DIM_0: return (1 << (ndims - 2));
```
dzarukin: The name would be misleading for cases when `ndims > 2`. `per_dim_0` implies batch dimension in that case, and scaling wouldn't be doing that...

tests/benchdnn/inputs/matmul/test_matmul_ci
```diff
@@ -72,6 +72,16 @@
 --batch=shapes_2d
 --batch=shapes_3d
 
+# Int8 src per_dim_0 scales
+--reset
+--dt=s8:s8:f32,s8:s8:bf16
+--attr-scales=src:per_dim_0,src:per_dim_0+wei:common:2,\
+              src:per_dim_0+wei:per_oc
```
dzarukin: Suggest to make separate cases with integer masks, e.g.,
```
--reset --dt=s8:s8:f32 --attr-scales=src:1+wei:2 32x64:64x48
...
```

5 or 6 should be enough with different data types and shapes.
I'd like to avoid expanding usage of named masks when their names are not reflecting reality properly.

tests/benchdnn/dnn_types.cpp
michalowski-arm: removed

tests/benchdnn/inputs/matmul/test_matmul_ci
michalowski-arm: done

src/cpu/matmul/ref_matmul.hpp
```diff
@@ -108,7 +109,8 @@ struct ref_matmul_t : public primitive_t {
                                                     DNNL_ARG_DST},
                                      {quantization_mode::static_sazp,
                                              quantization_mode::dynamic_mx,
-                                             quantization_mode::dynamic_fp}),
+                                             quantization_mode::dynamic_fp},
+                                     {{DNNL_ARG_SRC, {src_qmask_M()}}}),
```
dzarukin: This impl shouldn't get boosted (at least, I can't see it got covered) as it's responsible for `fp : int/fp : fp` combinations.

src/common/matmul_pd.hpp
```diff
@@ -238,22 +250,46 @@ struct matmul_pd_t : public primitive_desc_t {
...
+                } else {
+                    const auto g0 = scales.get_group(arg, 0);
+                    const auto g1 = scales.get_group(arg, 1);
+                    ok = ok
+                            && IMPLICATION(
+                                    (mask & src_qmask_K()), has_src_groups);
+                    ok = ok
+                            && IMPLICATION(has_src_groups,
+                                    g0 > 0 && g1 > 0 && K() % g1 == 0);
+                }
```
dzarukin: I think the previous structure worked fine. Each mask value had its dependencies on allowed groups, let's keep it as it was before (pseudo-code):
```
ok = masks_ok || extra_mask_ok;
if (mask_1) {
    ok = ...
} else if (mask_2) {
    ok = ...
} ...
```

I also think you shouldn't check for allowed extra (`is_src_extra_per_m`), just check for the mask directly. If mask is allowed, it will get through, if not, it won't pass the initial barrier anyway.

If you are adding support for groups for M, please add benchdnn coverage for that (though my understanding of per-token scaling is grouping is not needed there). In case it's added by mistake, please remove it to keep it cleaner.

src/cpu/matmul/ref_matmul.hpp
michalowski-arm: I wasn't sure on that one, reverted it now.

src/common/matmul_pd.hpp
michalowski-arm: Added it by mistake.

dzarukin (approved):

aditew01 (approved):

PR #4998, opened 2026-04-10
cpu: aarch64: make post-op injectors vector length agnostic

Sqvid:
# Description

Makes the post-op injectors, and related utilities, vector-length agnostic.

# Checklist

## General

cc: @jondea

### DIFF
--- a/src/cpu/aarch64/injectors/injector_utils.cpp
+++ b/src/cpu/aarch64/injectors/injector_utils.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2020 Intel Corporation
 * Copyright 2022 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@
 * limitations under the License.
 *******************************************************************************/
 #include <numeric>
-#include "common/broadcast_strategy.hpp"
+#include <vector>
 #include "cpu/aarch64/injectors/injector_utils.hpp"
 
 namespace dnnl {
@@ -30,26 +30,38 @@ register_preserve_guard_t<isa>::register_preserve_guard_t(jit_generator_t *host,
         std::initializer_list<Xbyak_aarch64::XReg> reg64_to_preserve,
         std::initializer_list<Xbyak_aarch64::VReg> vmm_to_preserve)
     : host_(host)
-    , reg64_stack_(reg64_to_preserve)
-    , vmm_stack_(vmm_to_preserve)
+    , gpr_regs_(reg64_to_preserve)
+    , vec_regs_(vmm_to_preserve)
     , vmm_to_preserve_size_bytes_(
               calc_vmm_to_preserve_size_bytes(vmm_to_preserve)) {
 
-    for (const auto &reg : reg64_to_preserve)
-        host_->str(reg, pre_ptr(host_->X_SP, -8));
+    if (reg64_to_preserve.size() != 0) {
+        const auto store_pairs = reg64_to_preserve.size() / 2;
+        const bool has_lone_store = reg64_to_preserve.size() % 2;
+        uint32_t i = 0;
+        for (; i < store_pairs; ++i)
+            host_->stp(gpr_regs_[2 * i], gpr_regs_[2 * i + 1],
+                    pre_ptr(host_->X_SP, -16));
 
-    if (!vmm_stack_.empty()) {
+        if (has_lone_store) {
+            // The hardware requires the stack pointer to be quad-word aligned
+            // https://github.com/ARM-software/abi-aa/blob/e2534ac15f02fa2e03b7f336f069f7cf392257d9/aapcs64/aapcs64.rst?#645the-stack
+            host_->str(gpr_regs_[2 * i], pre_ptr(host_->X_SP, -16));
+        }
+    }
+
+    if (!vec_regs_.empty()) {
         host_->sub(host_->X_SP, host_->X_SP, vmm_to_preserve_size_bytes_);
 
         uint32_t stack_offset = vmm_to_preserve_size_bytes_;
         for (const auto &vmm : vmm_to_preserve) {
-            const uint32_t bytes = cpu_isa_traits<isa>::vlen;
-            stack_offset -= bytes;
+            const uint32_t vlen_bytes = simd_bytes(isa);
+            stack_offset -= vlen_bytes;
             const auto idx = vmm.getIdx();
-            if (is_superset(isa, sve_256)) {
-                if (stack_offset % cpu_sveLen_ == 0) {
+            if (vlen_bytes > 16) {
+                if (stack_offset % vlen_bytes == 0) {
                     host_->st1w(Xbyak_aarch64::ZRegS(idx), host_->P_ALL_ONE,
-                            ptr(host_->X_SP, stack_offset / bytes,
+                            ptr(host_->X_SP, stack_offset / vlen_bytes,
                                     Xbyak_aarch64::MUL_VL));
                 } else {
                     host_->add_imm(host_->X_DEFAULT_ADDR, host_->X_SP,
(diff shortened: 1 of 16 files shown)


src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -862,8 +869,8 @@ void jit_uni_binary_injector_t<isa>::calculate_oc_blocked(
     // c = ((offset % strides[0]) / strides[1]) * strides[ndims - 1] + offset % blk_size
     // output = X_TMP_0
     const auto dst_d = rhs_arg_static_params_.dst_d;
-    const int simd_w = cpu_isa_traits<isa>::vlen
-            / types::data_type_size(dst_d.data_type());
+    const int simd_w
+            = isa_max_vlen(isa) / types::data_type_size(dst_d.data_type());
```
jondea: There's a `simd_elems` function for this

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -496,17 +496,24 @@ void jit_uni_binary_injector_t<isa>::compute_vector_range(
...
-                            ? std::initializer_list<Xbyak_aarch64::XReg>(
-                                      {rhs_arg_static_params_.rhs_addr_reg,
-                                              rhs_arg_static_params_
-                                                      .rhs_helper_reg})
-                            : std::initializer_list<Xbyak_aarch64::XReg>()),
-            (rhs_arg_static_params_.preserve_vmm_helper && dt_helper_vmm_needed
-                            ? std::initializer_list<Xbyak_aarch64::VReg>(
-                                      {Xbyak_aarch64::VReg(vmm_hint)})
-                            : std::initializer_list<Xbyak_aarch64::VReg>())};
+    const injector_utils::register_preserve_guard_t<to_vla_sve(isa)>
```
jondea, line 506: Can we pull out some of these variables to make it more readable?

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -112,7 +112,7 @@ size_t register_preserve_guard_t<isa>::calc_vmm_to_preserve_size_bytes(
     return std::accumulate(vmm_to_preserve.begin(), vmm_to_preserve.end(),
             std::size_t(0u),
             [](std::size_t accum, const Xbyak_aarch64::VReg &vmm) {
-        return accum + cpu_isa_traits<isa>::vlen;
+        return accum + isa_max_vlen(isa);
```
jondea: I think `simd_bytes(isa)` would be more explicit

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -43,7 +43,7 @@ register_preserve_guard_t<isa>::register_preserve_guard_t(jit_generator_t *host,
 
         uint32_t stack_offset = vmm_to_preserve_size_bytes_;
         for (const auto &vmm : vmm_to_preserve) {
-            const uint32_t bytes = cpu_isa_traits<isa>::vlen;
+            const uint32_t bytes = isa_max_vlen(isa);
             stack_offset -= bytes;
             const auto idx = vmm.getIdx();
             if (is_superset(isa, sve_256)) {
```
jondea: I think this logic is faulty and will fail when we use a `register_preserve_guard_t<sve>` on a machine with vlen >= 256 bits. Specifically, on this line it would decide to store a quadword even though the full vector to be preserved would be bigger.

It may work to swap this check (and the one in the destructor) with `simd_bytes(isa) > 16`

jondea (changes requested):
Thank you for taking a stab at this! I think there's a couple of issues with it though. I'm also slightly concerned that the CI passed even though there is a logical error which should break the register guard

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
Sqvid: It is now only as unreadable as it was before.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
Sqvid: Done.

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: Done.

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: Thanks for spotting this. I think this whole thing is a massive mess. It would also break on hardware because it was not keeping the stack-pointer 16-aligned. That should hopefully be fixed now.

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -30,26 +30,38 @@ register_preserve_guard_t<isa>::register_preserve_guard_t(jit_generator_t *host,
...
         for (const auto &vmm : vmm_to_preserve) {
-            const uint32_t bytes = cpu_isa_traits<isa>::vlen;
-            stack_offset -= bytes;
+            const uint32_t vlen_bytes = simd_bytes(isa);
+            stack_offset -= vlen_bytes;
             const auto idx = vmm.getIdx();
-            if (is_superset(isa, sve_256)) {
-                if (stack_offset % cpu_sveLen_ == 0) {
+            if (vlen_bytes > 16) {
+                if (stack_offset % vec_reg_len == 0) {
```
jondea: Can `vec_reg_len` be `vlen_bytes`? Why do we have two variables?

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -90,17 +102,26 @@ register_preserve_guard_t<isa>::~register_preserve_guard_t() {
...
+    if (!gpr_regs_.empty()) {
+        const bool has_lone_store = gpr_regs_.size() % 2;
+
+        if (has_lone_store) {
+            host_->ldr(gpr_regs_.back(), post_ptr(host_->X_SP, 16));
+            gpr_regs_.pop_back();
+        }
+
+        for (int i = gpr_regs_.size() - 1; i > 0; --i) {
+            host_->ldp(
```
jondea, line 122: Do we need to `pop` like we did before?

src/cpu/aarch64/injectors/injector_utils.hpp
```diff
@@ -88,15 +69,15 @@ class register_preserve_guard_t {
 
 private:
     jit_generator_t *host_;
-    std::stack<Xbyak_aarch64::XReg> reg64_stack_;
-    std::stack<Xbyak_aarch64::VReg> vmm_stack_;
-    const uint64_t cpu_sveLen_ = get_sve_length();
+    std::vector<Xbyak_aarch64::XReg> gpr_regs_;
+    std::vector<Xbyak_aarch64::VReg> vec_regs_;
+    const uint64_t vec_reg_len = isa_max_vlen(isa);
```
jondea: `len` is ambiguous, it would be good to give units here (bytes) and use the `simd_bytes` function. Although tbh, I don't think this member variable is necessary, and violates the single source of truth.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
jondea: I'd say it's slightly more readable, mission accomplished!

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: This is all in the destructor, so the vector is going to be deallocated at the end of the function anyway. I popped in the lone store just to make the logic easier.

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: You're right, I will remove the private member variable.

Sqvid: Done.

src/cpu/aarch64/injectors/injector_utils.hpp
Sqvid: Removed.

jondea (approved):
Great work, thank you!

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -30,26 +30,38 @@ register_preserve_guard_t<isa>::register_preserve_guard_t(jit_generator_t *host,
...
         host_->sub(host_->X_SP, host_->X_SP, vmm_to_preserve_size_bytes_);
 
         uint32_t stack_offset = vmm_to_preserve_size_bytes_;
         for (const auto &vmm : vmm_to_preserve) {
-            const uint32_t bytes = cpu_isa_traits<isa>::vlen;
-            stack_offset -= bytes;
+            const uint32_t vlen_bytes = simd_bytes(isa);
+            stack_offset -= vlen_bytes;
             const auto idx = vmm.getIdx();
-            if (is_superset(isa, sve_256)) {
```
renato-arantes, line 49: nit: for me `is_superset(isa, sve_256)` is better than `vlen_bytes > 16`.

renato-arantes (approved):

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: `is_superset` won't work as expected when the template is instantiated with `sve`.

PR #5000, opened 2026-04-10
common: move gemm utils to gpu/intel

Simonsays095:
In #4867 @vpirogov Asked me to move some gemm-related code into the GPU folder. Here's the summary:
- `common/gemm_utils.h`: Moved in this PR. This is the code I touched in the previous PR, and it's the only gemm code in `common/` used exclusively by the GPU team.
- `common/gemm_types.h`: Cannot be moved up due to common-level primitive hashing dependencies.
- `common/gemm_pd.h`: Cannot be moved up due to common-level verbose logging dependencies.
- `common/gemm.cpp`: Should probably be moved to the `cpu/` folder, but I'll leave someone on CPU/Arch teams to do that one.

### DIFF
--- a/src/common/gemm_pd.hpp
+++ b/src/common/gemm_pd.hpp
@@ -20,7 +20,6 @@
 #include "oneapi/dnnl/dnnl.h"
 
 #include "common/c_types_map.hpp"
-#include "common/gemm_utils.hpp"
 #include "common/primitive_desc.hpp"
 #include "common/utils.hpp"
 
--- a/src/gpu/intel/matmul/gemm.hpp
+++ b/src/gpu/intel/matmul/gemm.hpp
@@ -18,13 +18,12 @@
 #define GPU_INTEL_MATMUL_GEMM_HPP
 
 #include "common/c_types_map.hpp"
-#include "common/gemm_utils.hpp"
 #include "common/matmul_pd.hpp"
 #include "common/memory_desc.hpp"
-#include "common/memory_desc_wrapper.hpp"
 #include "common/primitive.hpp"
 #include "common/primitive_attr_quant.hpp"
 #include "common/primitive_desc_iterator.hpp"
+#include "gpu/intel/gemm/utils.hpp"
 #include "gpu/intel/matmul/config.hpp"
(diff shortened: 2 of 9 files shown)


vpirogov:
`common/gemm.cpp` is on it's way to be moved to `/dev/null` :)

src/gpu/intel/matmul/gemm.hpp
```diff
@@ -18,13 +18,12 @@
 #define GPU_INTEL_MATMUL_GEMM_HPP
 
 #include "common/c_types_map.hpp"
-#include "common/gemm_utils.hpp"
 #include "common/matmul_pd.hpp"
 #include "common/memory_desc.hpp"
-#include "common/memory_desc_wrapper.hpp"
```
kealan-barbieri, line 24: are this and `sdpa/config.hpp` just unused?

kealan-barbieri (approved):

src/gpu/intel/matmul/gemm.hpp
Simonsays095: Yeah, it's giving an unused header warning. I figured I'd remove it while I'm messing with the inclusions.

dzarukin (approved):

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

src/common/gemm_pd.hpp
```diff
@@ -20,7 +20,6 @@
 #include "oneapi/dnnl/dnnl.h"
 
 #include "common/c_types_map.hpp"
-#include "common/gemm_utils.hpp"
```
rjoursler, line 23: Longer term question: could we just remove the use of primitive descriptors for GEMM? I am not convinced this abstraction is actually buying us anything except headaches (especially with the implicit a/b swap).

Simonsays095: Maybe eventually, but at that point we should probably talk about removing gemm primitives entirely, using just matmul for everything. I remember there was some quirk regarding IP/matmul that required gemm pd and primitives, but I don't remember the details. Something about bwd propagation, potentially?

PR #5008, opened 2026-04-13
cpu: aarch64: enable binary op and binary post-ops on ASIMD

renato-arantes:
# Description

This PR enables JIT `binary` and binary `post-ops` for ASIMD. Previously, they were routed to `ref:any`, and now they are routed to `jit:uni`.

Fixes # (github issue)

# Checklist

## General

All the NIGHTLY tests are passing when the command below is executed: 
`ctest -j$(nproc) -E $(../.github/automation/aarch64/skipped-tests.sh)`

## Performance improvements

```
Number of tests                    : 3372
Number of threads                  : 16
Average JIT speedup (mean time)    : 5.7059
Average JIT speedup (min time)     : 6.0998

Top 15 JIT speedups by mean time
| mean_time_speedup | min_time_speedup | mean_time_ms_jit | mean_time_ms_ref | prb                                                                                                                                                   
-----------+-------------------+------------------+------------------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------
| 131.6289          | 139.5362         | 0.0811           | 10.6792          | --mode=P --binary --stag=abcde:abcde --dtag=abcde --alg=ge 1x2x6x407x407:1x2x6x407x407                                                                
| 119.2072          | 127.7289         | 0.0127           | 1.5120           | --mode=P --binary --stag=axb:any --attr-post-ops=mul:s8+add:f32:common+sum:0.5+abs 32x65x2x3x2x3:32x65x1x1x1x1                                        
| 100.5980          | 107.3243         | 0.0097           | 0.9782           | --mode=P --binary --stag=axb:any --inplace=true --attr-post-ops=ge:f32 32x65x2x3x2x3:32x65x1x1x1x1                                                    
| 99.9313           | 107.0304         | 0.0088           | 0.8829           | --mode=P --binary --attr-post-ops=mul:s8+add:f32:common+sum:0.5+abs 1000x60:1000x60                                                                   
| 97.1838           | 103.2106         | 0.0100           | 0.9675           | --mode=P --binary --stag=axb:any --attr-post-ops=ge:f32 32x65x2x3x2x3:32x65x1x1x1x1                                                                   
| 91.1309           | 95.7048          | 0.0114           | 1.0344           | --mode=P --binary --stag=axb:any --attr-post-ops=sum:0.25+relu:-0.01+gt:f32 32x65x2x3x2x3:32x65x1x1x1x1                                               
| 83.4453           | 84.1845          | 0.7605           | 63.4622          | --mode=P --binary --sdt=s8:f32 --ddt=u8 --stag=axb:axb --alg=sub --attr-post-ops=div:f32:per_oc+add:f32:per_oc+mul:f32:per_oc 8x1024x19x19:1x1024x1x1 
| 82.9788           | 90.9999          | 0.0068           | 0.5605           | --mode=P --binary --inplace=true --attr-post-ops=ge:f32 1000x60:1000x60                                                                               
| 82.3568           | 83.1222          | 0.7628           | 62.8225          | --mode=P --binary --sdt=u8:f32 --ddt=u8 --stag=axb:axb --alg=sub --attr-post-ops=div:f32:per_oc+add:f32:per_oc+mul:f32:per_oc 8x1024x19x19:1x1024x1x1 
| 81.3736           | 81.6808          | 0.8397           | 68.3260          | --mode=P --binary --sdt=u8:f32 --ddt=u8 --stag=axb:axb --alg=sub --attr-post-ops=div:f32:per_oc+add:f32:per_oc+mul:f32:per_oc 16x2048x10x10:1x2048x1x1
| 81.2526           | 85.9788          | 0.0122           | 0.9953           | --mode=P --binary --stag=axb:any --attr-post-ops=add:f32:per_oc+linear:2:1 32x65x2x3x2x3:32x65x1x1x1x1                                                
| 81.1151           | 84.2705          | 0.0123           | 1.0003           | --mode=P --binary --stag=axb:any --inplace=true --attr-post-ops=add:f32:per_oc+linear:2:1 32x65x2x3x2x3:32x65x1x1x1x1                                 
| 81.0118           | 87.5384          | 0.0069           | 0.5609           | --mode=P --binary --attr-post-ops=ge:f32 1000x60:1000x60                                                                                              
| 80.9378           | 81.5603          | 0.3825           | 30.9554          | --mode=P --binary --sdt=u8:f32 --ddt=u8 --stag=axb:axb --alg=sub --attr-post-ops=div:f32:per_oc+add:f32:per_oc+mul:f32:per_oc 2x512x38x38:1x512x1x1   
| 80.6038           | 80.6693          | 0.3786           | 30.5133          | --mode=P --binary --sdt=u8:f32 --ddt=u8 --stag=axb:axb --alg=sub --attr-post-ops=div:f32:per_oc+add:f32:per_oc+mul:f32:per_oc 1x256x75x75:1x256x1x1   

```

### DIFF
--- a/src/cpu/aarch64/injectors/injector_utils.cpp
+++ b/src/cpu/aarch64/injectors/injector_utils.cpp
@@ -157,8 +157,10 @@ conditional_register_preserve_guard_t<
                                 vmm_to_preserve}
                       : register_preserve_guard_t<isa> {nullptr, {}, {}}} {};
 
+template class register_preserve_guard_t<asimd>;
 template class register_preserve_guard_t<sve>;
 template class conditional_register_preserve_guard_t<sve>;
+template class conditional_register_preserve_guard_t<asimd>;
 
 } // namespace injector_utils
 } // namespace aarch64
--- a/src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
@@ -1364,6 +1364,12 @@ void jit_uni_binary_injector_t<isa>::calculate_w_cspn(
     calculate_w_nspc(strides, tmp_reg);
 }
 
+template <>
+void jit_uni_binary_injector_t<cpu_isa_t::asimd>::cvt_to_f32(
+        const Xbyak_aarch64::VReg &tmp_vmm) const {
+    host_->scvtf(tmp_vmm.s, tmp_vmm.s);
+}
+
 template <cpu_isa_t isa>
 void jit_uni_binary_injector_t<isa>::inject_binary(
         const dnnl_post_ops::entry_t &post_op, Vmm dst,
@@ -1408,6 +1414,38 @@ void jit_uni_binary_injector_t<isa>::inject_binary(
         }
 
         execute_binary(alg, dst, mask, lhs, rhs_addr);
+    }
+}
+
+template <>
+void jit_uni_binary_injector_t<asimd>::inject_binary(
+        const dnnl_post_ops::entry_t &post_op, Xbyak_aarch64::VReg dst,
+        const rhs_address_t &rhs_addr, bool with_tail,
+        const tail_lode_mode_t tail_load_mode) const {
+
+    const auto &alg = post_op.binary.alg;
+    const bool div_op = alg == alg_kind::binary_div;
+    const auto &rhs_arg_data_type = post_op.binary.src1_desc.data_type;
+    const bool process_rhs_arg_using_tmp = with_tail
+            || rhs_arg_data_type != data_type::f32
+            || !binary_op_with_unaligned_mem_operand_allowed_ || div_op;
+
+    if (process_rhs_arg_using_tmp) {
+        const Xbyak_aarch64::VReg tmp = Xbyak_aarch64::VReg(
+                rhs_arg_static_params_.rhs_dt_helper_vmm_idx);
+
+        if (rhs_addr.isBroadcast())
+            execute_broadcast(rhs_arg_data_type, tmp,
+                    remove_bcast_bit(rhs_addr), tail_load_mode, with_tail);
+        else
+            load_rhs(rhs_arg_data_type, tmp, rhs_addr, tail_load_mode,
+                    with_tail);
+
+        if (rhs_arg_data_type != data_type::f32) cvt_to_f32(tmp);
+
+        execute_binary(alg, dst, host_->P_ALL_ONE, dst, tmp);
+    } else {
+        execute_binary(alg, dst, host_->P_ALL_ONE, dst, rhs_addr);
     }
 }
 
@@ -1476,6 +1514,27 @@ void jit_uni_binary_injector_t<isa>::execute_broadcast_no_tail(
     }
 }
 
+template <>
+void jit_uni_binary_injector_t<asimd>::execute_broadcast_s8u8_no_tail(
+        const data_type_t &data_type, const Xbyak_aarch64::VReg &tmp_vmm,
+        const rhs_address_t &rhs_addr) const {
+    assert(utils::one_of(data_type, data_type::s8, data_type::u8)
+            && "unsupported data type");
+
+    const Xbyak_aarch64::XReg x_addr = host_->addr_off(rhs_addr.base_,
+            rhs_addr.offt_, host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+
+    if (data_type == data_type::s8) {
+        // load byte from memory into wtmp and sign-extend
+        host_->ldrsb(host_->W_TMP_0, Xbyak_aarch64::ptr(x_addr));
+    } else if (data_type == data_type::u8) {
+        // unsigned load
+        host_->ldrb(host_->W_TMP_0, Xbyak_aarch64::ptr(x_addr));
+    }
+    // broadcast into 4x32-bit
+    host_->dup(tmp_vmm.s4, host_->W_TMP_0);
+}
+
 template <cpu_isa_t isa>
 void jit_uni_binary_injector_t<isa>::execute_broadcast_s8u8_no_tail(
         const data_type_t &data_type, const Vmm &tmp_vmm,
@@ -1529,6 +1588,41 @@ void jit_uni_binary_injector_t<isa>::execute_broadcast_tail_with_opmask(
     }
 }
 
+template <>
+void jit_uni_binary_injector_t<asimd>::execute_broadcast_tail_with_opmask(
+        const data_type_t &data_type, const Xbyak_aarch64::VReg &tmp_vmm,
+        const rhs_address_t &rhs_addr) const {
+
+    const Xbyak_aarch64::XReg x_addr = host_->addr_off(
+            rhs_addr.base_, rhs_addr.offt_, host_->X_TMP_1, host_->X_TMP_0);
+
+    host_->uni_clear(tmp_vmm);
+    switch (data_type) {
+        case data_type::f32:
+        case data_type::s32:
+            host_->ld1(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[0],
+                    Xbyak_aarch64::ptr(x_addr));
+            for (size_t i = 1; i < rhs_arg_static_params_.tail_size; ++i) {
+                host_->ins(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i],
+                        Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[0]);
+            }
+            break;
+        case data_type::s8:
+        case data_type::u8:;
+            if (data_type == data_type::s8) {
+                host_->ldrsb(host_->W_TMP_0, Xbyak_aarch64::ptr(x_addr));
+            } else {
+                host_->ldrb(host_->W_TMP_0, Xbyak_aarch64::ptr(x_addr));
+            }
+            for (size_t i = 0; i < rhs_arg_static_params_.tail_size; ++i) {
+                host_->ins(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i],
+                        host_->W_TMP_0);
+            }
+            break;
+        default: assert(!"unsupported data type");
+    }
+}
+
 template <cpu_isa_t isa>
 void jit_uni_binary_injector_t<isa>::load_rhs_no_tail(
         const data_type_t &data_type, const Vmm &tmp_vmm,
@@ -1548,6 +1642,28 @@ void jit_uni_binary_injector_t<isa>::load_rhs_no_tail(
     }
 }
 
+template <>
+void jit_uni_binary_injector_t<asimd>::load_rhs_i8_no_tail(
+        const data_type_t &data_type, const Xbyak_aarch64::VReg &tmp_vmm,
+        const rhs_address_t &rhs_addr) const {
+    assert(utils::one_of(data_type, data_type::s8, data_type::u8)
+            && "unsupported data type");
+
+    const Xbyak_aarch64::XReg x_addr = host_->addr_off(rhs_addr.base_,
+            rhs_addr.offt_, host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+
+    for (size_t i = 0; i < 4; ++i) {
+        const Xbyak_aarch64::XReg lane_addr = host_->addr_off(x_addr,
+                i * sizeof(int8_t), host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+        if (data_type == data_type::s8) {
+            host_->ldrsb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+        } else {
+            host_->ldrb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+        }
+        host_->ins(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i], host_->W_TMP_0);
+    }
+}
+
 template <cpu_isa_t isa>
 void jit_uni_binary_injector_t<isa>::load_rhs_i8_no_tail(
         const data_type_t &data_type, const Vmm &tmp_vmm,
@@ -1564,6 +1680,45 @@ void jit_uni_binary_injector_t<isa>::load_rhs_i8_no_tail(
                 Xbyak_aarch64::ptr(host_->X_DEFAULT_ADDR));
     } else
         assert(!"unsupported data type");
+}
+
+template <>
+void jit_uni_binary_injector_t<asimd>::load_rhs_tail_dynamically_with_opmask(
+        const data_type_t &data_type, const Xbyak_aarch64::VReg &tmp_vmm,
+        const rhs_address_t &rhs_addr) const {
+
+    const Xbyak_aarch64::XReg x_addr = host_->addr_off(
+            rhs_addr.base_, rhs_addr.offt_, host_->X_TMP_1, host_->X_TMP_0);
+
+    host_->uni_clear(tmp_vmm);
+    switch (data_type) {
+        case data_type::f32:
+        case data_type::s32:
+            for (size_t i = 0; i < rhs_arg_static_params_.tail_size; ++i) {
+                const Xbyak_aarch64::XReg lane_addr
+                        = host_->addr_off(x_addr, i * sizeof(float),
+                                host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+                host_->ld1(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i],
+                        Xbyak_aarch64::ptr(lane_addr));
+            }
+            break;
+        case data_type::s8:
+        case data_type::u8:
+            for (size_t i = 0; i < rhs_arg_static_params_.tail_size; ++i) {
+                const Xbyak_aarch64::XReg lane_addr
+                        = host_->addr_off(x_addr, i * sizeof(int8_t),
+                                host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+                if (data_type == data_type::s8) {
+                    host_->ldrsb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+                } else {
+                    host_->ldrb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+                }
+                host_->ins(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i],
+                        host_->W_TMP_0);
+            }
+            break;
+        default: assert(!"unsupported data type");
+    }
 }
 
 template <cpu_isa_t isa>
@@ -1797,6 +1952,42 @@ void get_v_rhs_value(
     }
 }
 
+template <>
+void jit_uni_binary_injector_t<asimd>::execute_cmp_binary(
+        const Xbyak_aarch64::VReg &dst, const Xbyak_aarch64::PReg &mask,
+        const Xbyak_aarch64::VReg &lhs, const Xbyak_aarch64::VReg &rhs,
+        const unsigned int cmp_predicate) const {
+    UNUSED(mask);
+
+    switch (cmp_predicate) {
+        case jit_generator_t::_cmp_nlt_us:
+            host_->fcmge(dst.s, lhs.s, rhs.s);
+            break;
+        case jit_generator_t::_cmp_nle_us:
+            host_->fcmgt(dst.s, lhs.s, rhs.s);
+            break;
+        case jit_generator_t::_cmp_le_os:
+            host_->fcmge(dst.s, rhs.s, lhs.s);
+            break;
+        case jit_generator_t::_cmp_lt_os:
+            host_->fcmgt(dst.s, rhs.s, lhs.s);
+            break;
+        case jit_generator_t::_cmp_eq_oq:
+            host_->fcmeq(dst.s, lhs.s, rhs.s);
+            break;
+        case jit_generator_t::_cmp_neq_uq:
+            host_->fcmeq(dst.s, lhs.s, rhs.s);
+            host_->not_(dst.b, dst.b);
+            break;
+        default: assert(!"unsupported compare mode"); break;
+    }
+
+    // ASIMD compare instructions produce all-ones/all-zeros integer masks.
+    // Convert them to the expected 1.0f / 0.0f compare result in place.
+    host_->ushr(dst.s, dst.s, 31);
+    host_->scvtf(dst.s, dst.s);
+}
+
 template <cpu_isa_t isa>
 void jit_uni_binary_injector_t<isa>::execute_cmp_binary(const Vmm &dst,
         const Xbyak_aarch64::PReg &mask, const Vmm &lhs, const Vmm &rhs,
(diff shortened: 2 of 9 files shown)


jondea:
It's surprising how similar the performance is to ref. Also, as far as I know, how is it relative to ACL? I'm pretty sure there are some cases that should go to ACL on a c6g

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -136,9 +136,11 @@ conditional_register_preserve_guard_t<
                                 vmm_to_preserve}
                       : register_preserve_guard_t<isa> {nullptr, {}, {}}} {};
 
+template class register_preserve_guard_t<asimd>;
```
Sqvid, line 160: This is not sufficient. You need to look at the `.hpp` as well:
https://github.com/uxlfoundation/oneDNN/blob/31c97ba3b4d59e459892b7ad72a3111736663b60/src/cpu/aarch64/injectors/injector_utils.hpp#L93

src/cpu/aarch64/jit_uni_binary_kernel.cpp
```diff
@@ -251,45 +252,108 @@ unsigned int jit_uni_binary_kernel_t<isa>::cmp_predicate(alg_kind_t alg) {
...
-        NGE_UQ = 25,
-        NGT_UQ = 26,
-        FALSE_OS = 27,
-        NEQ_OS = 28,
-        GE_OQ = 29,
-        GT_OQ = 30,
-        TRUE_US = 31,
-    };
+namespace {
+enum {
```
Sqvid: Do we ever actually use the integer values? If not let's convert this to an `enum class`.

src/cpu/aarch64/jit_uni_binary.cpp
Sqvid: This function looks like it is reinventing stuff that is in `cpu_isa_traits.cpp/hpp`. Won't `get_max_cpu_isa()` do this?

src/cpu/aarch64/jit_uni_binary.cpp
```diff
@@ -468,7 +470,7 @@ bool jit_uni_binary_t::post_ops_ok(const primitive_attr_t *attr,
 
     const auto is_binary = [&](int idx) { return p.entry_[idx].is_binary(); };
 
-    if (!mayiuse(sve_128)) return false;
+    if (!mayiuse(sve_128) && !mayiuse(asimd)) return false;
```
Sqvid: `!mayiuse(asimd)` already implies `!mayiuse(sve_128)` so you can use it on it's own.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1808,6 +1997,100 @@ void jit_uni_binary_injector_t<isa>::execute_cmp_binary(const Vmm &dst,
     pop_opmask(host_, cmp_mask);
 }
 
+template <>
+template <typename T>
+void jit_uni_binary_injector_t<asimd>::execute_binary(alg_kind_t binary_alg,
+        const Xbyak_aarch64::VReg &dst, const Xbyak_aarch64::PReg &mask,
+        const Xbyak_aarch64::VReg &lhs, const T &rhs) const {
+    const int SIMD_SZ = 16; // SIMD register length in bytes
```
Sqvid: Just use `simd_bytes()` and leave out the comment.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1808,6 +1997,100 @@ void jit_uni_binary_injector_t<isa>::execute_cmp_binary(const Vmm &dst,
...
+    const bool isAddr = std::is_same<T, rhs_address_t>::value;
+    Vmm v_rhs(0);
```
Sqvid: Does `Vmm` make sense in a `isa` specialization.

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -87,6 +87,18 @@ jit_io_helper_t<Vmm>::jit_io_helper_t(jit_generator_t *host,
...
+// That is ok for NEON paths because we handle tail manually with tail_conf_->tail_size_.
+template <>
+void jit_io_helper_t<Xbyak_aarch64::VReg>::prepare_opmask(
+        const std::size_t how_many_bits_to_set,
+        const Xbyak_aarch64::XReg &reg_tmp0,
+        const Xbyak_aarch64::XReg &reg_tmp1, const Xbyak_aarch64::PReg &mask) {
+    UNUSED(how_many_bits_to_set);
+    UNUSED(reg_tmp0);
+    UNUSED(reg_tmp1);
+    UNUSED(mask);
```
Sqvid, line 99: Would it be a programming error to call this for `asimd`? In that case we should `assert`.

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -114,7 +126,7 @@ template <typename Vmm>
 void jit_io_helper_t<Vmm>::prepare_full_mask() {
     assert(gather_conf_.has_value() && "Config for loading with the use of gather instruction is not set.");
 
-    assert(is_superset(isa_, sve_128));
+    assert(is_superset(isa_, sve_128) || isa_ == asimd);
```
Sqvid: Why not just combine this to `is_superset(isa_, asimd)`?

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -145,6 +157,58 @@ void jit_io_helper_t<Vmm>::init_saturate_f32() const {
...
+            && "Config for tail processing is not set.");
+
+    const unsigned number_of_values_to_load = tail ? tail_conf_->tail_size_ : 4;
+
+    host_->movi(dst_vmm.b16, 0); // zeroing dst_vmm for tail processing
+
+    switch (data_type_) {
+        case data_type::f32:
+        case data_type::s32:
+            for (unsigned j = 0; j < number_of_values_to_load; j++) {
```
Sqvid: I understand that we are simulating gather-loads by manually reading indices and loading them. But this is going to be incredibly inefficient right? To the point that I think a caller might want to re-think the algorithm for `asimd` if they find themselves needing to do this. Happy to be corrected. @jondea what do you think?

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -145,6 +157,58 @@ void jit_io_helper_t<Vmm>::init_saturate_f32() const {
...
+                host_->mov(host_->W_TMP_0,
+                        Xbyak_aarch64::VReg4S(indices_vmm.getIdx())[j]);
+                host_->add(host_->X_DEFAULT_ADDR, src_reg, host_->X_TMP_0);
+                host_->ld1(Xbyak_aarch64::VReg4S(dst_vmm.getIdx())[j],
+                        Xbyak_aarch64::ptr(host_->X_DEFAULT_ADDR));
+            }
+            break;
+        case data_type::s8:
+        case data_type::u8:
+            for (unsigned j = 0; j < number_of_values_to_load; j++) {
```
Sqvid: nit: style: `j` is a little confusing as I instinctively assumed I must be inside an inner (`i`) loop.

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -662,6 +1054,16 @@ void uni_vpmovsxbd(jit_generator_t *host_, const Vmm &dst, const Vmm &src) {
     host_->sxtb(z_dst.s, host_->P_ALL_ONE / Xbyak_aarch64::T_m, z_dst.s);
 }
 
+void uni_vpmovzxbd(jit_generator_t *host_, const Xbyak_aarch64::VReg &dst,
```
Sqvid: Can we use this patch as an opportunity to change names that do not apply to this architecture?

Sqvid (changes requested):

src/cpu/aarch64/utils/jit_io_helper.cpp
renato-arantes: As agreed with @jondea, this is beyond the scope of this PR and will be addressed later.

renato-arantes:
> It's surprising how similar the performance is to ref. Also, as far as I know, how is it relative to ACL? I'm pretty sure there are some cases that should go to ACL on a c6g

I was wrongly benchmarking the test `test_benchdnn_modeC_binary_all_cpu`, where about 2/3 of the tests were going to `ref`. Now I have properly benchmarked with 1/3 of the test that goes to JIT.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1364,6 +1364,12 @@ void jit_uni_binary_injector_t<isa>::calculate_w_cspn(
     calculate_w_nspc(strides, tmp_reg);
 }
 
+template <>
+void jit_uni_binary_injector_t<cpu_isa_t::asimd>::cvt_to_f32(
```
Sqvid: I think this (and the `sve` version) are [already implemented as `uni_scvt`](https://github.com/uxlfoundation/oneDNN/blob/7f0a3e02c55c1ad505c51db1897286b2053a21dd/src/cpu/aarch64/jit_generator.hpp#L633-L641) right?

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -482,6 +482,16 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
...
+    void uni_fmax(const Xbyak_aarch64::VReg4S &dst,
+            const Xbyak_aarch64::VReg4S &src,
+            const Xbyak_aarch64::VReg4S &src2) {
+        uint32_t dstIdx = dst.getIdx();
+        uint32_t srcIdx = src.getIdx();
+        if (dstIdx != srcIdx)
+            mov(Xbyak_aarch64::ZRegD(dstIdx), Xbyak_aarch64::ZRegD(srcIdx));
+        fmax(dst, src, src2);
+    }
+
```
Sqvid, line 494: This function seems extremely strange to me. It accepts only `asimd` registers but then uses `ZRegD` which is an `SVE` register that may not be available. Can you please clarify what you are trying to do?

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -506,6 +516,16 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
         fmin(dst, P_ALL_ONE / Xbyak_aarch64::T_m, src2);
     }
 
+    void uni_fmin(const Xbyak_aarch64::VReg4S &dst,
+            const Xbyak_aarch64::VReg4S &src,
+            const Xbyak_aarch64::VReg4S &src2) {
+        uint32_t dstIdx = dst.getIdx();
+        uint32_t srcIdx = src.getIdx();
+        if (dstIdx != srcIdx)
+            mov(Xbyak_aarch64::ZRegD(dstIdx), Xbyak_aarch64::ZRegD(srcIdx));
```
Sqvid: Ditto. Using `ZReg` in a function that seems to be for `asimd` support.

src/cpu/aarch64/jit_uni_binary_kernel.cpp
```diff
@@ -237,140 +238,241 @@ XReg jit_uni_binary_kernel_t<isa>::dst_ptr(size_t offt) {
     return X_DEFAULT_ADDR;
 }
 
+namespace {
```
Sqvid: Can we leave a link to a an official public resource where a developer can understand what these comparison types mean. Also is this a general enough concept that it should be in `jit_generator_t`? (cc: @jondea)

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1797,6 +1952,42 @@ void get_v_rhs_value(
...
 }
 
+template <>
+void jit_uni_binary_injector_t<asimd>::execute_cmp_binary(
+        const Xbyak_aarch64::VReg &dst, const Xbyak_aarch64::PReg &mask,
+        const Xbyak_aarch64::VReg &lhs, const Xbyak_aarch64::VReg &rhs,
+        const unsigned int cmp_predicate) const {
+    UNUSED(mask);
+
+    switch (cmp_predicate) {
```
Sqvid, line 1930: Don't these have a slightly confusing overlap with the `cmp_mode_t` class that you introduce later?

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -198,11 +262,54 @@ void jit_io_helper_t<Vmm>::load(const Xbyak_aarch64::XReg &src_addr,
...
         case data_type::s8:
-        case data_type::u8: load_i8(src_addr, offt, dst_raw_vmm, mask); break;
+        case data_type::u8:
+            load_i8(src_addr, offt, dst_raw_vmm, tail, mask);
+            break;
         default: assert(!"Unsupported data type.");
     }
 }
 
+template <>
```
Sqvid: Can these all be removed in favour of the `uni_*` functions in `jit_generator_t`?
(E.g. `uni_ld1rw`)

Sqvid (changes requested):

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
renato-arantes: This is the binary op. The other one is binary post-op.

src/cpu/aarch64/jit_uni_binary_kernel.cpp
renato-arantes: I agree with that, this code is messy and need refactoring, but is beyond the scope of this PR.

src/cpu/aarch64/jit_generator.hpp
Sqvid: @renato-arantes you marked this as resolved but it looks like the code is still there. Did I misunderstand or is there another patchset you intended to push?

renato-arantes: Sorry, I shouldn't have marked this as solved before putting the patch. Doing this right now.

src/cpu/aarch64/utils/jit_io_helper.cpp
renato-arantes: Yes, but kept `float_point_fused_multiply_add` and `contiguous_load_unsigned_words` as they only used inside `jit_uni_binary_kernel.cpp`.

Sqvid (approved):
Approving this patch, but I am uneasy with the pre-existing code-quality in these files. The design of `jit_io_helper_t` forces `asimd` and `sve` to define implementaions that really ought to apply only to one or the other.

I will let @jondea decide on the final approval as he has a better picture of the urgency of these changes vs. the tech-debt.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1548,6 +1636,28 @@ void jit_uni_binary_injector_t<isa>::load_rhs_no_tail(
...
+
+    for (size_t i = 0; i < 4; ++i) {
+        const Xbyak_aarch64::XReg lane_addr = host_->addr_off(x_addr,
+                i * sizeof(int8_t), host_->X_DEFAULT_ADDR, host_->X_TMP_0);
+        if (data_type == data_type::s8) {
+            host_->ldrsb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+        } else {
+            host_->ldrb(host_->W_TMP_0, Xbyak_aarch64::ptr(lane_addr));
+        }
+        host_->ins(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[i], host_->W_TMP_0);
```
jondea: Really not sure 4x loads and `ins` are necessary. This will be _really_ slow too. I think the idiomatic Neon way to implement `ld1sb(v., ...)` is a word load and then two `uxtl`, but I think it's worth investigating. If you implement uni functions, this can be merged into the sve version

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1529,6 +1582,41 @@ void jit_uni_binary_injector_t<isa>::execute_broadcast_tail_with_opmask(
...
+        const rhs_address_t &rhs_addr) const {
+
+    const Xbyak_aarch64::XReg x_addr = host_->addr_off(
+            rhs_addr.base_, rhs_addr.offt_, host_->X_TMP_1, host_->X_TMP_0);
+
+    host_->uni_clear(tmp_vmm);
+    switch (data_type) {
+        case data_type::f32:
+        case data_type::s32:
+            host_->ld1(Xbyak_aarch64::VReg4S(tmp_vmm.getIdx())[0],
```
jondea, line 1569: For the tail, are we better doing `tail_size` number of laned loads or the same number of `ins`s?

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1411,6 +1411,38 @@ void jit_uni_binary_injector_t<isa>::inject_binary(
...
+        if (rhs_addr.isBroadcast())
+            execute_broadcast(rhs_arg_data_type, tmp,
+                    remove_bcast_bit(rhs_addr), tail_load_mode, with_tail);
+        else
+            load_rhs(rhs_arg_data_type, tmp, rhs_addr, tail_load_mode,
+                    with_tail);
+
+        if (rhs_arg_data_type != data_type::f32) cvt_to_f32(tmp);
+
+        execute_binary(alg, dst, host_->P_ALL_ONE, dst, tmp);
```
jondea: If we are going to pass in a `PReg` to the `asimd` `execute_binary`, why can't we have a single method for `inject_binary`?

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -145,6 +157,58 @@ void jit_io_helper_t<Vmm>::init_saturate_f32() const {
...
+        const Xbyak_aarch64::XReg &src_reg,
+        const Xbyak_aarch64::VReg &indices_vmm,
+        const Xbyak_aarch64::VReg &dst_vmm, const bool tail) {
+    assert(gather_conf_.has_value() && "Config for loading with the use of gather instruction is not set.");
+    assert(IMPLICATION(tail, tail_conf_.has_value())
+            && "Config for tail processing is not set.");
+
+    const unsigned number_of_values_to_load = tail ? tail_conf_->tail_size_ : 4;
+
+    host_->movi(dst_vmm.b16, 0); // zeroing dst_vmm for tail processing
```
jondea: uni_clear

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -145,6 +157,58 @@ void jit_io_helper_t<Vmm>::init_saturate_f32() const {
...
+
+    switch (data_type_) {
+        case data_type::f32:
+        case data_type::s32:
+            for (unsigned i = 0; i < number_of_values_to_load; i++) {
+                host_->mov(host_->W_TMP_0,
```
jondea, line 178: Here we are using laned loads, and before we were using `ins`, I think we should have a consistent approach to handling tails. I think we could move a lot of this logic out of here too

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -198,11 +262,41 @@ void jit_io_helper_t<Vmm>::load(const Xbyak_aarch64::XReg &src_addr,
...
-        case data_type::u8: load_i8(src_addr, offt, dst_raw_vmm, mask); break;
+        case data_type::u8:
+            load_i8(src_addr, offt, dst_raw_vmm, tail, mask);
+            break;
         default: assert(!"Unsupported data type.");
     }
 }
 
+template <>
+void jit_io_helper_t<Xbyak_aarch64::ZReg>::float_point_fused_multiply_add(
```
jondea: All these simple helpers which only require `P_ALL_ONE` should be moved to the jit_generator

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -497,12 +591,69 @@ void jit_io_helper_t<Xbyak_aarch64::VReg>::load_f32(
...
+    if (tail && tail_conf_->tail_size_ > 0) {
+        // tail_size_ = nelems % simd_w_, so it cannot be greater than (simd_w_ - 1), which is 4 for VReg.
+        // Refer to binary_kernel_t::get_tail_size().
+        const int SZ = sizeof(float);
+        switch (tail_conf_->tail_size_) {
+            case 1:
+                host_->ld1(dst_vmm.s[0], Xbyak_aarch64::ptr(src_addr));
+                break;
+            case 2:
+                host_->ld1(dst_vmm.s[0], Xbyak_aarch64::ptr(src_addr));
```
jondea: In the case of 2 I think we can just do load a double with: `ld1(dst_vmm.d[0])`. Equally, for the case of 3, we can load a double and then a single. This pattern can be used repeatedly for other kinds of tails too. E.g. 7 bytes is 1 word, 1 half and 1 byte rather than loading 7 bytes

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -514,9 +665,39 @@ void jit_io_helper_t<Vmm>::load_s32(const Xbyak_aarch64::XReg &src_addr,
     host_->scvtf(dst_vmm.s, host_->P_TMP / Xbyak_aarch64::T_m, dst_vmm.s);
 }
 
+template <>
+void jit_io_helper_t<Xbyak_aarch64::VReg>::load_i8(
```
jondea, line 612: This could be generic I think

src/cpu/aarch64/jit_uni_binary_kernel.cpp
jondea: Given that you are pulling them out of a function scope and putting them into the cpp scope, I think it's worth thinking about what we are doing here. I'm a bit confused as to why we need them at all, why can't we use the `binary_eq` etc enums directly?

src/cpu/aarch64/jit_uni_binary_kernel.cpp
```diff
@@ -237,140 +238,241 @@ XReg jit_uni_binary_kernel_t<isa>::dst_ptr(size_t offt) {
...
+        uni_fdiv(v0.s, v0.s, v1.s);
+    else if (alg == binary_sub)
+        uni_fsub(v0.s, v0.s, v1.s);
+    else if (cmp_op) {
+        const unsigned int alg_enum = get_cmp_alg_enum(alg);
+        compute_cmp_alg(v0, v0, v1, alg_enum);
+        // ASIMD compare instructions produce all-ones/all-zeros integer masks.
+        // Convert them to the expected 1.0f / 0.0f compare result in place.
+        ushr(v0.s, v0.s, 31);
+        scvtf(v0.s, v0.s);
```
jondea: Rather than an expensive `scvtf` we can just `and` the result with a `fmov`ed `1.0f`

jondea (changes requested):
I think there's still quite a few points of this that need some more thought. The approach feels a little inconsistent

renato-arantes:
> Approving this patch, but I am uneasy with the pre-existing code-quality in these files. The design of `jit_io_helper_t` forces `asimd` and `sve` to define implementaions that really ought to apply only to one or the other.
> 
> I will let @jondea decide on the final approval as he has a better picture of the urgency of these changes vs. the tech-debt.

Thanks, @Sqvid, for your review. I agree with your concern!
From reviewing the AArch64 JIT code, this appears to be pre-existing technical debt rather than something specific to this patch.
I can see the same in other AArch64 JIT areas, such as the [eltwise injector](https://github.com/renato-arantes/oneDNN/blob/9ec04ceed14607f6aefc1d70bcfac26d0fbbf9eb/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp#L3132), [shuffle kernel](https://github.com/renato-arantes/oneDNN/blob/9ec04ceed14607f6aefc1d70bcfac26d0fbbf9eb/src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp#L71), and [batch normalisation](https://github.com/renato-arantes/oneDNN/blob/9ec04ceed14607f6aefc1d70bcfac26d0fbbf9eb/src/cpu/aarch64/jit_uni_batch_normalization.cpp#L1029), where the shared template requires empty ASIMD methods, ISA branches, or ISA-specific load/store helpers.
So I agree the long-term fix should probably be to split the helper contract by ISA capability, or keep only a small common layer and move ASIMD/SVE-specific IO behaviour into separate helper classes.
I would prefer not to include that refactoring in this PR, as it is broader than the binary-op enablement and would make the functional change harder to review. I am happy to track/follow up on it separately in another PR.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
renato-arantes: For `execute_broadcast_tail_with_opmask`, one load plus ins is the better choice. This path broadcasts a single scalar value into the active tail lanes, so doing `tail_size` lane loads would just reload the same address repeatedly.

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
renato-arantes: Yes, we can, but we still have to address the `with_tail_fusable_to_binary_op` branch, which will require a helper function to handle the mask in the `mov`. Splitting `inject_binary` makes the code clearer and easier to read.

src/cpu/aarch64/utils/jit_io_helper.cpp
renato-arantes: It's different here. It's not loading contiguous elements from memory; instead, it's accessing indexed elements from `indices_vmm`.

src/cpu/aarch64/jit_uni_binary_kernel.cpp
jondea: If you are stuck for temp registers, you can also do a logical right and left shift to truncate all the 1s to
```
00111111100000000000000000000000
```

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
jondea: We can skip the `with_tail_fusable_to_binary_op` branch by testing `isa != asimd`. This make sense too, if we don't have the ability to mask, then we can't fuse it.

jondea (approved):

PR #5011, opened 2026-04-13
xe: gemm: unrestrict hf8 downconversion pathway

Simonsays095:
Unrestrict `f32`->`f8_e4m3` downconversion pathway using the copy planner. Allows the use of the `ar` strategy parameter (alternate C-remainder) for `fp8` downversion GEMM problems such as `--dt=f8_e4m3`. Intended to increase support for testing as `ar` generates relatively few lines of assembly compared to `sr br` but is slightly slower.

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/pieces/c_update.cxx
+++ b/src/gpu/intel/gemm/jit/generator/pieces/c_update.cxx
@@ -2101,7 +2101,7 @@ void Generator<hw>::convert(const GRFMultirange &range, Type Told, Type Tnew, co
     if (Told == Tnew)
         return;
     if (Told.isInt4() || Tnew.isInt4()) stub();
-    if (Told == Type::hf8 || Tnew == Type::hf8) stub();
+    if (Told == Type::hf8) stub();
 
     // Special path: x32->FP.
     if (one_of(Told, {Type::s32, Type::u32}) && Tnew.isFP()) {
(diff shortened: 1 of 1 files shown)


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

src/gpu/intel/gemm/jit/generator/pieces/c_update.cxx
```diff
@@ -2115,24 +2115,14 @@ void Generator<hw>::convert(const GRFMultirange &range, Type Told, Type Tnew, co
     // Special path: f32->bf8.
     if (hw >= HW::XeHPC && Told == Type::f32 && Tnew == Type::bf8) {
         int ne = elementsPerGRF<uint32_t>(hw);
-        for (int i = 0; i < range.getLen(); i++)
```
kealan-barbieri, line 2118: No platforms support direct float->fp8 per spec, I think we need to preserve some of these routines.

Simonsays095: The copy planner will handle these conversions. With current platforms/support it'll downconvert f32->f16, then emulated or native f16->fp8 depending on platform. It even shaves off a few instructions by combining registers when possible compared to this existing code. I checked on Xe3 and it uses the same downconversion routine as in this code that I'm removing.

kealan-barbieri: I see, missed that fundamental aspect of this, we should do a perf check hopefully the `CopyPlan` routines are faster, I think theyre newer.

kealan-barbieri (approved):

src/gpu/intel/gemm/jit/generator/pieces/c_update.cxx
kealan-barbieri: actually pursuant to the use of `CopyPlan` do we even need to check arch/src/dst type?

Simonsays095: Good catch - I removed the check for `hw >= XeHPC`. I thought about making this entire function a thin wrapper on the copy planner like this new code, but the comment before the function declares that it does the conversion "in-place", which isn't guaranteed with the copy planner currently. We likely rely on this when using this function inside the k-loop, so I only wanted to make use of the copy planner during post-k-loop downconversion for now.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

h-sadia (approved):

PR #5014, opened 2026-04-14
cpu: x64: matmul: Enable int8 grouped quantization

inteldimitrius:
### What was done:
1. int4->int8 upconversion for the quantized weights. Implemented in `brgemm_matmul_copy_utils.cpp`
2. Weight zero points shift grouped, common, per_oc, per_ocic. Implemented in `brgemm_matmul_copy_utils.cpp`
3. Weight scales application: grouped, per_ocic. Implemented in `jit_brgemm_kernel.cpp`, `jit_brgemm_amx_uker.cpp`
4. Source zero points shift grouped, per_ocic. Implemented in  `brgemm_matmul_copy_utils.cpp`
5. Source scales application grouped, per_ocic. Implemented in `jit_brgemm_kernel.cpp`, `jit_brgemm_amx_uker.cpp`

### Details of implementation: 
Grouped scales for both src and wei are applied in brgemm kernel, AMX requires the group size being divisible by 64. Other cases are handled via non-AMX brgemm kernel.  

Grouped zero points for both src and wei are applied in copy routines and require buffer B (buffer A for src) usage. 
### Test coverage
is already in the main branch, was using reference implementation before or skipped:
Implementation statistics in comparison to main: 
 

#### Implementation dispatch (`harness_matmul_decompression`)

| Implementation                 |  main | branch | delta |
| ------------------------------ | ----: | -----: | ----: |
| `brg_matmul:avx10_1_512`       |  1659 |   1676 |   +17 |
| `brg_matmul:avx10_1_512_amx`   |   489 |    540 |   +51 |
| `brg_matmul:avx512_core`       |   184 |    184 |     0 |
| `brg_matmul:avx512_core_bf16`  |    48 |     48 |     0 |
| `ref:any`                      |    18 |     16 |    −2 |
| **TOTAL (dispatched)**         |  **2398** | **2464** | **+66** |

#### Test outcomes

|            | tests | passed | skipped | mistrusted | unimplemented | invalid_arguments | failed |
| ---------- | ----: | -----: | ------: | ---------: | ------------: | ----------------: | -----: |
| **main**   |  2842 |   2398 |     444 |          0 |             0 |                 0 |      0 |
| **branch** |  2842 |   2464 |     378 |          0 |             0 |                 0 |      0 |
| **delta**  |     0 |    +66 |     −66 |          0 |             0 |                 0 |      0 |

The cases with dynamic quantization in `harness_matmul_weight_decompression` are passing now. 

Supported ISAs: AVX512+, AVX2-VNNI, AMX.

### Performance testing on SPR:
[grouped_quantization_report.pdf](https://github.com/user-attachments/files/28267679/grouped_quantization_report.pdf)

### DIFF
--- a/src/cpu/x64/brgemm/brgemm.cpp
+++ b/src/cpu/x64/brgemm/brgemm.cpp
@@ -166,6 +166,7 @@ void brgemm_kernel_execute_postops(const brgemm_kernel_t *brg_kernel, int bs,
     brgemm_p.a_zp_compensations = post_ops_data.a_zp_compensations;
     brgemm_p.b_zp_compensations = post_ops_data.b_zp_compensations;
     brgemm_p.c_zp_values = post_ops_data.c_zp_values;
+    brgemm_p.ptr_per_mn_compensation = post_ops_data.per_mn_compensation;
     if (dynamic_values) {
         brgemm_p.dynamic_LDA = dynamic_values->dynamic_LDA;
         brgemm_p.dynamic_LDB = dynamic_values->dynamic_LDB;
@@ -210,6 +211,7 @@ void brgemm_kernel_execute_postops(const brgemm_kernel_t *brg_kernel, int bs,
     brgemm_p.b_zp_compensations = post_ops_data.b_zp_compensations;
     brgemm_p.a_zp_values = post_ops_data.a_zp_values;
     brgemm_p.c_zp_values = post_ops_data.c_zp_values;
+    brgemm_p.ptr_per_mn_compensation = post_ops_data.per_mn_compensation;
     if (dynamic_values) {
         brgemm_p.dynamic_LDA = dynamic_values->dynamic_LDA;
         brgemm_p.dynamic_LDB = dynamic_values->dynamic_LDB;
@@ -529,7 +531,8 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
         zp_type = brgemm_broadcast_t::none;
 
         const bool skip_zero_point
-                = mem_arg == DNNL_ARG_WEIGHTS && brg->skip_zp_b_compensation;
+                = (mem_arg == DNNL_ARG_WEIGHTS && brg->skip_zp_b_compensation)
+                || (mem_arg == DNNL_ARG_SRC && brg->skip_zp_a_compensation);
         if (skip_zero_point) return status::success;
 
         if (!zp.has_default_values(mem_arg)) {
--- a/src/cpu/x64/brgemm/brgemm.cpp
+++ b/src/cpu/x64/brgemm/brgemm.cpp
@@ -486,6 +486,7 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     const auto &src_scales = attr->scales_.get(DNNL_ARG_SRC);
     const auto &wei_scales = attr->scales_.get(DNNL_ARG_WEIGHTS);
     brg->with_src_scales = !src_scales.has_default_values();
+    if (brg->with_src_scales) brg->dt_src_scales = src_scales.get_data_type();
     brg->with_wei_scales
             = !brg->skip_wei_scales && !wei_scales.has_default_values();
 
@@ -516,7 +517,7 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     const bool scales_ok = attr->scales_.has_default_values({DNNL_ARG_SRC,
                                    DNNL_ARG_WEIGHTS, DNNL_ARG_DST})
             && IMPLICATION(!src_scales.has_default_values(),
-                    src_scales.get_mask() == 0)
+                    src_scales.get_mask() == 0 || brg->is_per_k_src_scales)
             && IMPLICATION(!dst_scales.has_default_values(),
                     dst_scales.get_mask() == 0);
     if (!scales_ok) return status::unimplemented;
@@ -839,6 +840,9 @@ int brgemm_cmp(const brgemm_desc_t &lhs, const brgemm_desc_t &rhs) {
     CMP_BRGEMM_FIELD(is_per_n_wei_scales);
     CMP_BRGEMM_FIELD(is_per_k_wei_scales);
     CMP_BRGEMM_FIELD(with_src_scales);
+    CMP_BRGEMM_FIELD(is_per_k_src_scales);
+    CMP_BRGEMM_FIELD(src_scale_m_stride);
+    CMP_BRGEMM_FIELD(dt_src_scales);
     CMP_BRGEMM_FIELD(with_wei_scales);
     CMP_BRGEMM_FIELD(with_dst_scales);
     CMP_BRGEMM_FIELD(dt_wei_scales);
--- a/src/cpu/x64/brgemm/brgemm.cpp
+++ b/src/cpu/x64/brgemm/brgemm.cpp
@@ -488,7 +488,10 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     brg->with_src_scales = !src_scales.has_default_values();
     brg->with_wei_scales
             = !brg->skip_wei_scales && !wei_scales.has_default_values();
-    if (brg->with_wei_scales) {
+
+    bool wei_scales_are_set = brg->is_single_wei_scale
+            || brg->is_per_n_wei_scales || brg->is_per_k_wei_scales;
+    if (brg->with_wei_scales && !wei_scales_are_set) {
         // Note. the current version supports only two different wei scales
         // types:
         //     1) common (mask = 0)
@@ -500,7 +503,11 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
         // So if wei_scales.get_mask() > 0 (not common) it's assumed here that
         // scale type is per_n_dim_scale and driver which calls brgemm kernel
         // checked that mask has correct value for this case
-        brg->is_oc_scale = wei_scales.get_mask() > 0;
+
+        const auto &wei_scales_mask = wei_scales.get_mask();
+        brg->is_single_wei_scale = wei_scales_mask == 0;
+        brg->is_per_n_wei_scales = wei_scales_mask > 0;
+        brg->is_per_k_wei_scales = false;
         brg->dt_wei_scales = wei_scales.get_data_type();
     }
 
@@ -828,7 +835,9 @@ int brgemm_cmp(const brgemm_desc_t &lhs, const brgemm_desc_t &rhs) {
     CMP_BRGEMM_FIELD(zp_type_c);
 
     CMP_BRGEMM_FIELD(skip_wei_scales);
-    CMP_BRGEMM_FIELD(is_oc_scale);
+    CMP_BRGEMM_FIELD(is_single_wei_scale);
+    CMP_BRGEMM_FIELD(is_per_n_wei_scales);
+    CMP_BRGEMM_FIELD(is_per_k_wei_scales);
     CMP_BRGEMM_FIELD(with_src_scales);
     CMP_BRGEMM_FIELD(with_wei_scales);
     CMP_BRGEMM_FIELD(with_dst_scales);
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -297,6 +297,7 @@ struct brgemm_desc_t {
     bool with_eltwise = false;
     bool with_binary = false;
     bool skip_zp_b_compensation = false;
+    bool skip_zp_a_compensation = false;
     bool n_bcast_1_load = false;
 
     brgemm_broadcast_t zp_type_a = brgemm_broadcast_t::none;
@@ -311,6 +312,8 @@ struct brgemm_desc_t {
     bool is_per_k_src_scales = false;
     dim_t src_scale_m_stride = 0;
     data_type_t dt_src_scales = data_type::undef;
+
+    bool with_per_mn_compensation = false;
 
     bool has_per_k_scales() const {
         return is_per_k_wei_scales || is_per_k_src_scales;
@@ -488,9 +491,11 @@ struct brgemm_desc_t {
 
     // Determines if the accumulator dt is integer.
     //
-    // When the per-k scales are present, need to accumulate
-    // in f32 due to precision loss.
-    bool is_integer_acc() const { return is_int8 && !has_per_k_scales(); }
+    // When the per-k scales or per-MN compensation are present, need to
+    // accumulate in f32 due to precision loss.
+    bool is_integer_acc() const {
+        return is_int8 && !has_per_k_scales() && !with_per_mn_compensation;
+    }
 
     // Determines if the DQ2PS conversion is needed for the accumulator.
     bool do_dq2ps_cvt() const {
@@ -756,6 +761,7 @@ struct brgemm_kernel_params_t {
     const void *b_zp_compensations = nullptr;
     const void *a_zp_values = nullptr;
     const void *c_zp_values = nullptr;
+    const void *ptr_per_mn_compensation = nullptr;
     size_t skip_accm = 0;
     int32_t zp_a_val = 1;
     dim_t dynamic_LDA = 0;
@@ -881,7 +887,8 @@ struct brgemm_post_ops_data_t {
             int32_t zp_a_val = 1, bool do_only_comp = false,
             bool do_only_zp_a_val = false, const void *src_scales = nullptr,
             const void *wei_scales = nullptr, const void *dst_scales = nullptr,
-            const void *a_zp_values = nullptr)
+            const void *a_zp_values = nullptr,
+            const void *per_mn_compensation = nullptr)
         : bias(bias)
         , binary_post_ops_rhs(binary_post_ops_rhs)
         , oc_logical_off(oc_logical_off)
@@ -898,7 +905,8 @@ struct brgemm_post_ops_data_t {
         , src_scales(src_scales)
         , wei_scales(wei_scales)
         , dst_scales(dst_scales)
-        , a_zp_values(a_zp_values) {}
+        , a_zp_values(a_zp_values)
+        , per_mn_compensation(per_mn_compensation) {}
 
     const void *bias = nullptr;
     const void *binary_post_ops_rhs = nullptr;
@@ -917,6 +925,7 @@ struct brgemm_post_ops_data_t {
     const void *wei_scales = nullptr;
     const void *dst_scales = nullptr;
     const void *a_zp_values = nullptr;
+    const void *per_mn_compensation = nullptr;
 };
 
 } // namespace x64
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -308,7 +308,14 @@ struct brgemm_desc_t {
     bool is_per_n_wei_scales = false;
     bool is_per_k_wei_scales = false;
     bool with_src_scales = false;
-    bool has_per_k_scales() const { return is_per_k_wei_scales; }
+    bool is_per_k_src_scales = false;
+    dim_t src_scale_m_stride = 0;
+    data_type_t dt_src_scales = data_type::undef;
+
+    bool has_per_k_scales() const {
+        return is_per_k_wei_scales || is_per_k_src_scales;
+    }
+
     bool with_wei_scales = false;
     // `dst_scales` passed as a bare pointer making kernel change multiplication
     // to division was proved to be significantly slower, both for pure divps
--- a/src/cpu/x64/brgemm/brgemm_types.hpp
+++ b/src/cpu/x64/brgemm/brgemm_types.hpp
@@ -303,11 +303,12 @@ struct brgemm_desc_t {
     brgemm_broadcast_t zp_type_b = brgemm_broadcast_t::none;
     brgemm_broadcast_t zp_type_c = brgemm_broadcast_t::none;
 
-    // `skip_wei_scales` is controlled by the implementation and not by kernel
-    // API.
     bool skip_wei_scales = false;
-    int is_oc_scale = 0;
+    bool is_single_wei_scale = false;
+    bool is_per_n_wei_scales = false;
+    bool is_per_k_wei_scales = false;
     bool with_src_scales = false;
+    bool has_per_k_scales() const { return is_per_k_wei_scales; }
     bool with_wei_scales = false;
     // `dst_scales` passed as a bare pointer making kernel change multiplication
     // to division was proved to be significantly slower, both for pure divps
@@ -475,7 +476,18 @@ struct brgemm_desc_t {
         assert(is_gemv);
         // Per-N scales collapse to a single scalar unless `y` is treated as a
         // row.
-        return !is_oc_scale || !treat_y_as_row;
+        return !is_per_n_wei_scales || !treat_y_as_row;
+    }
+
+    // Determines if the accumulator dt is integer.
+    //
+    // When the per-k scales are present, need to accumulate
+    // in f32 due to precision loss.
+    bool is_integer_acc() const { return is_int8 && !has_per_k_scales(); }
+
+    // Determines if the DQ2PS conversion is needed for the accumulator.
+    bool do_dq2ps_cvt() const {
+        return is_integer_acc() && (alpha != 1.f || beta != 1.f);
     }
 
     // return 'true' when FP8 MAC is not natively supported by the CPU ISA
@@ -705,33 +717,33 @@ struct brgemm_dynamic_values_t {
 };
 
 struct brgemm_kernel_params_t {
-    const void *ptr_A;
-    const void *ptr_B;
-    const brgemm_batch_element_t *batch;
-    void *ptr_C;
+    const void *ptr_A = nullptr;
+    const void *ptr_B = nullptr;
+    const brgemm_batch_element_t *batch = nullptr;
+    void *ptr_C = nullptr;
 
-    const void *ptr_bias;
-    void *ptr_D;
+    const void *ptr_bias = nullptr;
+    void *ptr_D = nullptr;
 
     const void *ptr_src_scales = nullptr;
     const void *ptr_wei_scales = nullptr;
     const void *ptr_dst_scales = nullptr;
-    void *ptr_buf;
+    void *ptr_buf = nullptr;
 
-    size_t do_post_ops;
-    size_t do_apply_comp;
-    size_t BS;
+    size_t do_post_ops = 0;
+    size_t do_apply_comp = 0;
+    size_t BS = 0;
 
     /*
      * ptr to table of void * elements that are pointers to post_op binary
      * src1 tensors
      */
-    const void *post_ops_binary_rhs_arg_vec;
-    size_t oc_logical_off;
-    size_t first_mb_matrix_addr_off;
-    size_t dst_row_logical_off;
+    const void *post_ops_binary_rhs_arg_vec = nullptr;
+    size_t oc_logical_off = 0;
+    size_t first_mb_matrix_addr_off = 0;
+    size_t dst_row_logical_off = 0;
 
-    const char *data_C_ptr_;
+    const char *data_C_ptr_ = nullptr;
 
     const void *a_zp_compensations = nullptr;
     const void *b_zp_compensations = nullptr;
@@ -822,7 +834,7 @@ private:
 
 /// @param bias Vector of bias (vector length is N)
 /// @param scales - Vector of scale factor values which represents combination
-///     scale factors for matrixes A and B. If brgemm_desc_t::is_oc_scale = true
+///     scale factors for matrixes A and B. If brgemm_desc_t::is_per_n_scale = true
 ///     vector length is N otherwise it must be broadcasted to vector of simd
 ///     width length
 /// @param binary_post_ops_rhs - Ptr to table of pointers to tensors used as rhs
--- a/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
+++ b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
@@ -398,8 +398,7 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
 
     if (brg.with_wei_scales) {
         reg_aux_wei_scales.restore();
-        const bool is_single_scale = !brg.is_oc_scale;
-        if (!is_single_scale) {
+        if (!brg.is_single_wei_scale) {
             assert(brg.dt_wei_scales == data_type::f32);
             lea(reg_aux_wei_scales,
                     ptr[reg_aux_wei_scales + reg_aux_N * sizeof(float)]);
@@ -420,7 +419,7 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
             const Vmm vmm_wei_scales = vmm_tmp(0);
             const auto addr
                     = ptr[reg_aux_wei_scales + wei_scales_offset(n, v_i)];
-            if (is_single_scale) {
+            if (brg.is_single_wei_scale) {
                 if (has_ptr_b_support) {
                     // No need to check for mask support separately, as masks
                     // are supported with the same isa that introduced address
--- a/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
@@ -577,7 +577,8 @@ private:
         const bool need_to_apply_post_ops
                 = are_post_ops_applicable_ && apply_post_ops;
         const auto store_by_vectors = need_to_apply_alpha_beta_
-                || need_to_apply_post_ops || brg.brgattr.bd_mask_level;
+                || need_to_apply_post_ops || brg.brgattr.bd_mask_level
+                || brg.has_per_k_scales();
         return store_by_vectors;
     }
     bool actual_ils(bool apply_post_ops, bool skip_accumulation = false) const {
@@ -829,7 +830,7 @@ dim_t jit_brgemm_amx_uker_base_t::bias_offset(int ldb) const noexcept {
 }
 
 dim_t jit_brgemm_amx_uker_base_t::scales_offset(int ldb) const noexcept {
-    return brg.is_oc_scale * ldb * ld_block_scales_size_;
+    return brg.is_per_n_wei_scales * ldb * ld_block_scales_size_;
 }
 
 dim_t jit_brgemm_amx_uker_base_t::zp_comp_a_offset(int ldb) const noexcept {
@@ -975,8 +976,9 @@ void jit_brgemm_amx_uker_base_t::apply_alpha_beta_to_vector(
     const bool apply_beta = brg.beta != 0.f;
     if (!apply_alpha && !apply_beta) return;
 
-    const bool dq2ps_required = brg.is_int8 && (apply_alpha || brg.beta != 1.f);
-    const bool use_vadd_for_beta = brg.beta == 1.f && !dq2ps_required;
+    // When K-scales (wei or src) are applied, accumulator is already
+    // converted to float (dq2ps + scales before alpha_beta), C stores floats.
+    const bool use_vadd_for_beta = brg.beta == 1.f && !brg.do_dq2ps_cvt();
 
     if (apply_beta && !use_vadd_for_beta) {
         mov(reg_tmp_gpr, float2int(static_cast<float>(brg.beta)));
@@ -988,12 +990,12 @@ void jit_brgemm_amx_uker_base_t::apply_alpha_beta_to_vector(
         vmovq(Xmm(zmm_alpha.getIdx()), reg_tmp_gpr);
         vbroadcastss(zmm_alpha, Xmm(zmm_alpha.getIdx()));
     }
-    if (dq2ps_required) vcvtdq2ps(zmm, zmm);
+    if (brg.do_dq2ps_cvt()) vcvtdq2ps(zmm, zmm);
     if (apply_alpha) vmulps(zmm, zmm, zmm_alpha);
     if (apply_beta) {
         if (use_vadd_for_beta) {
             auto zmm_masked = zmm | k_mask | T_z;
-            if (brg.is_int8)
+            if (brg.is_integer_acc())
                 vpaddd(zmm_masked, zmm, addr);
             else
                 vaddps(zmm_masked, zmm, addr);
@@ -1131,8 +1133,42 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers_ldb(
 
 void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
         brgemm_iteration_t &bi) {
-    if (!bi.apply_postops) return;
     const auto ldi = bi.ldi;
+
+    // Load wei_scales for per K-block application (is_per_k_wei_scales).
+    // This must happen for both apply_postops and non-apply_postops paths.
+    if (brg.with_wei_scales && brg.is_per_k_wei_scales) {
+        mov(reg_scales, ptr[param1 + GET_OFF(ptr_wei_scales)]);
+        for (int ldb = 0; ldb < ldi->block2(); ldb++) {
+            auto scales_ptr = EVEX_compress_addr(
+                    reg_scales, scales_offset(ldi->pos(ldb)));
+            auto k_mask = ldi->is_tail(ldb) ? ld_tail_mask : ld_full_mask;
+            if (brg.is_single_wei_scale) {
+                // Broadcast a single scale value — handle non-f32 types.
+                switch (brg.dt_wei_scales) {
+                    case data_type::f32:
+                        vbroadcastss(zmm_scales(ldb), scales_ptr);
+                        break;
+                    case data_type::bf16:
+                        vpbroadcastw(zmm_scales(ldb), scales_ptr);
+                        vpslld(zmm_scales(ldb), zmm_scales(ldb), 16);
+                        break;
+                    case data_type::f16:
+                        vpbroadcastw(zmm_scales(ldb), scales_ptr);
+                        vcvtph2ps(zmm_scales(ldb),
+                                Xbyak::Ymm(zmm_scales(ldb).getIdx()));
+                        break;
+                    default: vbroadcastss(zmm_scales(ldb), scales_ptr); break;
+                }
+            } else {
+                // Load a vector of scale values with proper type conversion.
+                cvt2ps(brg.dt_wei_scales, zmm_scales(ldb), scales_ptr, true,
+                        false, k_mask);
+            }
+        }
+    }
+
+    if (!bi.apply_postops) return;
 
     if (brg.with_bias) {
         mov(reg_bias, ptr[param1 + GET_OFF(ptr_bias)]);
@@ -1156,18 +1192,17 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
         }
     }
 
-    if (brg.with_wei_scales) {
+    if (brg.with_wei_scales && !brg.is_per_k_wei_scales) {
         mov(reg_scales, ptr[param1 + GET_OFF(ptr_wei_scales)]);
         for (int ldb = 0; ldb < ldi->block2(); ldb++) {
             auto scales_ptr = EVEX_compress_addr(
                     reg_scales, scales_offset(ldi->pos(ldb)));
             auto k_mask = ldi->is_tail(ldb) ? ld_tail_mask : ld_full_mask;
-            const bool is_single_scale = !brg.is_oc_scale;
 
             const auto zmm_scale = zmm_scales(ldb);
             const auto zmm_scale_masked = zmm_scales(ldb) | k_mask | T_z;
 
-            if (is_single_scale) {
+            if (brg.is_single_wei_scale) {
                 if (brg.with_src_scales) {
                     // Single value is not anticipated to be of any other type
                     // when both scales are defined.
@@ -1457,6 +1492,18 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
             vmovups(vreg_acc, ptr[reg_buf + buf_offset + wsp_offset]);
         }
 
+        // For K-scales (wei and/or src), convert int32->float and apply
+        // the current K-group's scales BEFORE alpha_beta accumulation.
+        if (brg.has_per_k_scales() && !bi.skip_accumulation) {
+            vcvtdq2ps(zmm, zmm);
+
+            if (brg.is_per_k_wei_scales) {
+                const Xbyak::Zmm scaled_zmm
+                        = vmm_mask(zmm, true, false, k_mask);
+                vmulps(scaled_zmm, scaled_zmm, zmm_scales(ldb));
+            }
+        }
+
         if (need_to_apply_alpha_beta_ || bi.skip_accumulation) {
             const auto c_offset = C_offset(bi, bdb, bd, bi.ldi->pos(ldb));
             const auto ptr_C
@@ -1467,7 +1514,8 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
 
         if (!bi.apply_postops) continue;
 
-        if (dq2ps_required) vcvtdq2ps(zmm, zmm);
+        // When K-scales (wei or src) are used, dq2ps already applied above.
+        if (dq2ps_required && !brg.has_per_k_scales()) vcvtdq2ps(zmm, zmm);
 
         if (brg.req_comp_pads_with_bcast)
             apply_comp_pad_to_vector(bi, bdb, bd, ldb, zmm.getIdx());
@@ -1503,7 +1551,11 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
         }
     }
 
-    if (brg.with_src_scales || brg.with_wei_scales) {
+    // When K-scales (wei) are used, they were already applied
+    // per K-block above. Only apply the remaining scales (non-K) in postops.
+    const bool apply_scales_in_postops = brg.with_src_scales
+            || (brg.with_wei_scales && !brg.is_per_k_wei_scales);
+    if (apply_scales_in_postops) {
         for (auto bd = bd_start; bd < bd_finish; bd++) {
             if (!is_out_bd(bi.bdi, bdb, bd)) continue;
 
@@ -2944,7 +2996,7 @@ void jit_brgemm_amx_uker_base_t::init(brgemm_iteration_t &bi) {
                 = brg.alpha != 1.0f || brg.beta != 0.f;
         const bool beta_uses_vadd = brg.beta == 1.f
                 && IMPLICATION(brg.is_int8, brg.alpha == 1.0f);
-        dt_requires_saturation_ = brg.is_int8
+        dt_requires_saturation_ = brg.is_int8 && !brg.has_per_k_scales()
                 && !IMPLICATION(alpha_or_beta_applicable, beta_uses_vadd);
     }
     use_sat_cvt_ = dt_requires_saturation_
--- a/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
+++ b/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
@@ -191,6 +191,7 @@ private:
     const reg64_savable_t reg_zp_comp_b {regscratchpad_, rbx, r19};
     const reg64_savable_t reg_zp_c_values {regscratchpad_, rbx, r20};
     const reg64_savable_t reg_src_scales_per_k {regscratchpad_, rbx, r21};
+    const reg64_savable_t reg_per_mn_comp {regscratchpad_, rbx, r22};
     const reg64_t reg_ptr_sum_zp = rbx;
     const reg64_t reg_converted_stride = rsi;
     const reg64_t reg_zp_comp_pad_a = rsi;
@@ -579,7 +580,7 @@ private:
                 = are_post_ops_applicable_ && apply_post_ops;
         const auto store_by_vectors = need_to_apply_alpha_beta_
                 || need_to_apply_post_ops || brg.brgattr.bd_mask_level
-                || brg.has_per_k_scales();
+                || brg.has_per_k_scales() || brg.with_per_mn_compensation;
         return store_by_vectors;
     }
     bool actual_ils(bool apply_post_ops, bool skip_accumulation = false) const {
@@ -619,6 +620,8 @@ private:
             int inp_bd, int ldb) const noexcept;
     dim_t zp_comp_b_offset(int bd) const noexcept;
     dim_t zp_c_values_offset(brgemm_iteration_t &bi, int ldb) const noexcept;
+    dim_t per_mn_comp_offset(const brgemm_iteration_t &bi, int bdb, int inp_bd,
+            int ldb) const noexcept;
     bool is_out_bd(const bd_iteration_t *bdi, int bdb, int inp_bd) const;
     int get_out_bd(const bd_iteration_t *bdi, int bdb, int inp_bd) const;
 
@@ -862,6 +865,20 @@ dim_t jit_brgemm_amx_uker_base_t::zp_c_values_offset(
     return 0;
 }
 
+dim_t jit_brgemm_amx_uker_base_t::per_mn_comp_offset(
+        const brgemm_iteration_t &bi, int bdb, int inp_bd,
+        int ldb) const noexcept {
+    const auto bi_bd_start = get_out_bd(bi.bdi, 0, 0);
+    const auto bd = get_out_bd(bi.bdi, bdb, inp_bd);
+    const auto bd_shift = bd - (ununroll_bd_loop ? bi_bd_start : 0);
+    const dim_t ldc_elem = (dim_t)ldb * brg.ld_block;
+    const dim_t bloc_idx = ldc_elem / brg.LDC;
+    const dim_t in_block = ldc_elem % brg.LDC;
+    return (dim_t)sizeof(float)
+            * ((dim_t)bd_shift * brg.LDC2_M + (dim_t)bloc_idx * brg.LDC2_N
+                    + in_block);
+}
+
 bool jit_brgemm_amx_uker_base_t::is_out_bd(
         const bd_iteration_t *bdi, int bdb, int inp_bd) const {
     const auto bd = bdi->pos(bdb) + inp_bd;
@@ -940,6 +957,11 @@ void jit_brgemm_amx_uker_base_t::read_params() {
     if (brg.is_per_k_src_scales) {
         mov(reg_src_scales_per_k, ptr[param1 + GET_OFF(ptr_src_scales)]);
         reg_src_scales_per_k.save();
+    }
+
+    if (brg.with_per_mn_compensation) {
+        mov(reg_per_mn_comp, ptr[param1 + GET_OFF(ptr_per_mn_compensation)]);
+        reg_per_mn_comp.save();
     }
 }
 
@@ -1172,6 +1194,26 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
                         false, k_mask);
             }
         }
+
+        if (brg.with_src_scales && !brg.is_per_k_src_scales) {
+            mov(reg_scales, ptr[param1 + GET_OFF(ptr_src_scales)]);
+            auto zmm_src_sc = zmm_tmp_1();
+            auto src_sc_addr = EVEX_compress_addr(reg_scales, 0);
+            switch (brg.dt_src_scales) {
+                case data_type::bf16:
+                    vpbroadcastw(zmm_src_sc, src_sc_addr);
+                    vpslld(zmm_src_sc, zmm_src_sc, 16);
+                    break;
+                case data_type::f16:
+                    vpbroadcastw(zmm_src_sc, src_sc_addr);
+                    vcvtph2ps(zmm_src_sc, Xbyak::Ymm(zmm_src_sc.getIdx()));
+                    break;
+                case data_type::f32:
+                default: vbroadcastss(zmm_src_sc, src_sc_addr); break;
+            }
+            for (int ldb = 0; ldb < ldi->block2(); ldb++)
+                vmulps(zmm_scales(ldb), zmm_scales(ldb), zmm_src_sc);
+        }
     }
 
     if (!bi.apply_postops) return;
@@ -1187,7 +1229,8 @@ void jit_brgemm_amx_uker_base_t::prepare_post_ops_registers(
         }
     }
 
-    if (brg.with_src_scales && !brg.is_per_k_src_scales) {
+    if (brg.with_src_scales && !brg.is_per_k_src_scales
+            && !brg.is_per_k_wei_scales) {
         mov(reg_scales, ptr[param1 + GET_OFF(ptr_src_scales)]);
         for (int ldb = 0; ldb < ldi->block2(); ldb++) {
             // Hard-coded assumption for a single src scale value being
@@ -1515,10 +1558,28 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
             vmovups(vreg_acc, ptr[reg_buf + buf_offset + wsp_offset]);
         }
 
-        // For K-scales (wei and/or src), convert int32->float and apply
-        // the current K-group's scales BEFORE alpha_beta accumulation.
-        if (brg.has_per_k_scales() && !bi.skip_accumulation) {
+        // Per-(M,N) compensation: convert int32->float and subtract the
+        // unscaled delta BEFORE per-K scale multiply.
+        if (brg.with_per_mn_compensation && !bi.skip_accumulation) {
             vcvtdq2ps(zmm, zmm);
+
+            reg_per_mn_comp.restore();
+            const auto global_ldb = bi.ldi->pos(ldb);
+            const auto delta_off = per_mn_comp_offset(bi, bdb, bd, global_ldb);
+            const auto delta_ptr = EVEX_compress_addr_safe(
+                    reg_per_mn_comp, delta_off, reg_long_offt);
+            const auto zmm_delta = zmm_tmp_1();
+            const Xbyak::Zmm zmm_delta_masked
+                    = vmm_mask(zmm_delta, true, false, k_mask);
+            vmovups(zmm_delta_masked, delta_ptr);
+            vsubps(zmm, zmm, zmm_delta);
+        }
+
+        // For K-scales (wei and/or src), convert int32->float (if not
+        // already done by per-MN compensation) and apply the current
+        // K-group's scales BEFORE alpha_beta accumulation.
+        if (brg.has_per_k_scales() && !bi.skip_accumulation) {
+            if (!brg.with_per_mn_compensation) vcvtdq2ps(zmm, zmm);
 
             const Xbyak::Zmm scaled_zmm = vmm_mask(zmm, true, false, k_mask);
             // Apply K-wei_scales if present (pre-loaded per-ldb vector).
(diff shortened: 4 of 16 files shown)


inteldimitrius:
make test

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1437,6 +1444,10 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
         bgmmc.wei_dt = f16;
         bgmmc.tr_a_dt_sz = types::data_type_size(f16);
         bgmmc.tr_b_dt_sz = types::data_type_size(f16);
+    } else if (bgmmc.with_int8_grouped_quantization) {
+        bgmmc.wei_dt = s8;
+        bgmmc.tr_a_dt_sz = types::data_type_size(bgmmc.src_dt);
+        bgmmc.tr_b_dt_sz = types::data_type_size(s8);
```
dzarukin, line 1450: This better preserve original data type. If u4 was requested, it should be converted to u8 instead, not s8.

src/cpu/x64/brgemm/brgemm.hpp
```diff
@@ -246,6 +246,13 @@ void brgemm_kernel_execute(const brgemm_kernel_t *brg_kernel, int bs,
         void *scratch = nullptr,
         const brgemm_dynamic_values_t *dynamic_values = nullptr);
 
+/// Execute BRGEMM kernel with pre-set params
+/// @param brg_kernel BRGEMM kernel
+/// @param params Pre-set parameters for the BRGEMM kernel
+/// Useful for setting additional parameters, like IC scales or post-ops.
+void brgemm_kernel_execute_params(
+        const brgemm_kernel_t *brg_kernel, brgemm_kernel_params_t &params);
```
dzarukin, line 254: Introducing methods like this doesn't lead to better state of API. It doesn't announce benefits over existing methods, the code is not updated to get benefit from benefits (if there are) in other places, and removing obsolete APIs with better versions.

How should user know which API to use, execute or execute_params?

I recommend to stick to the existing ways to avoid disruption and propose these changes separately for the team to collect the feedback and discuss the further plans.

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -305,8 +305,12 @@ struct brgemm_desc_t {
 
     // `skip_scales` is controlled by the implementation and not by kernel API.
     bool skip_scales = false;
-    int is_oc_scale = 0;
+    int is_single_wei_scale = 0;
+    int is_oc_wei_scales = 0;
+    int is_ic_wei_scales = 0;
+    int wei_scale_k_group_size = 0;
     bool with_src_scales = false;
+    bool has_ic_scales() const { return is_ic_wei_scales; }
```
dzarukin, line 313: Please stick to M, N, K terms instead of OC and IC.

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -305,8 +305,12 @@ struct brgemm_desc_t {
 
     // `skip_scales` is controlled by the implementation and not by kernel API.
     bool skip_scales = false;
-    int is_oc_scale = 0;
+    int is_single_wei_scale = 0;
+    int is_oc_wei_scales = 0;
+    int is_ic_wei_scales = 0;
```
dzarukin, line 310: Commit name that enables partial accumulator in brgemm should reflect brgemm in it, not matmul to avoid misleading information.

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1467,7 +1519,8 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
 
         if (!bi.apply_postops) continue;
 
-        if (dq2ps_required) vcvtdq2ps(zmm, zmm);
+        // When IC scales (wei or src) are used, dq2ps already applied above.
+        if (dq2ps_required && !brg.has_ic_scales()) vcvtdq2ps(zmm, zmm);
```
dzarukin, line 1523: `dq2ps_required` should already handle this. There are other feature that rely on this conversion that are not attached to scaling anyhow.

src/cpu/x64/brgemm/brgemm.hpp
inteldimitrius: Defining `brgemm_kernel_params_t` is more sufficiently be filled on driver side it gives more flexibility, existing methods are of 2 kind - execute and execute_postops. Introducing grouped scales requires passing pointers and other data each call, not only postops. In anyway it requires to introduce the new method.
Introducing new methos like `execute_grouped_scales ` seems to me too specific and will cause code duplication, for my estimation it will generate 30+ duplicated lines. So:
I've created universal method which could arbitrary fill params for brgemm kernel and can be completely reusable.

Updating other (except matmul) primitive seems to me outside this PRs scope. However it can be easily updated in separate one.

inteldimitrius:
make test
disable test_device_gpu

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1457,6 +1503,41 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
...
+
+            const Xbyak::Zmm scaled_zmm = vmm_mask(zmm, true, false, k_mask);
+            // Apply IC wei_scales if present (pre-loaded per-ldb vector).
+            if (brg.is_per_k_wei_scales) {
+                vmulps(scaled_zmm, scaled_zmm, zmm_scales(ldb));
+            }
+            // Apply IC src_scales per-bd: each M-row has its own scalar.
+            if (brg.is_per_k_src_scales) {
+                reg_src_scales_ic.restore();
+                const auto src_sc_offset = bd * brg.src_scale_m_stride;
```
densamoilov: Is this offset correct? `reg_src_scales_ic` seems to contain the base address of the scales buffer. And `src_sc_offset` is calculated using the stride and `bd`, which is an M index within a tile so it's not an absolute offset.

The case below fails:
```
--matmul --dt=s8:u8:f16 --attr-scales=src:per_ocic:f16:1x256+wei:per_ocic:f16:128x1 32x256:256x64
```

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -690,33 +700,35 @@ struct brgemm_dynamic_values_t {
...
+    void *ptr_buf = nullptr;
 
-    size_t do_post_ops;
-    size_t do_apply_comp;
-    size_t BS;
+    size_t do_post_ops = 0;
+    size_t do_apply_comp = 0;
+    size_t BS = 0;
+    // Used for K-grouped scales application
+    dim_t k_start = 0;
```
densamoilov: Is it really used? I couldn't find the place it's used after initialization.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2834,9 +2832,14 @@ struct jit_brgemm_matmul_copy_b_int8_t : public jit_brgemm_matmul_copy_b_t,
                   do_compute_compensation_ && !isa_has_int8_vnni(conf->isa))
         , is_dynamic_stride_(is_runtime_value(src_stride_))
         , is_dynamic_N_(conf->is_runtime_N)
-        , comp_acc_idx_(is_ymm_                      ? 13
+        , is_src_int4_(one_of(conf->orig_wei_dt, data_type::s4, data_type::u4))
+        , has_vpermb_(cpu().has(Xbyak::util::Cpu::tAVX512_VBMI))
```
dzarukin, line 2836: Minor: it feels like this could be added to jit_generator in a more readable way (across the whole CPU stack).

src/cpu/x64/brgemm/brgemm.hpp
densamoilov: The motivation behind introducing this API is understandable. But I think this API has a serious issue because it somewhat exposes an implementation detail of the brgemm kernel, namely it leaks the internal `brgemm_kernel_params_t` structure, which is the brgemm kernel ABI.

1. The caller is expected to set only some fields of `brgemm_kernel_params_t` (e.g. scales, post-ops pointers, etc).
2. For all the remaining fields the caller silently relies on the struct's default member initializers being correct for the kernel.

The correctness of every call therefore hinges on the assumption that the defaults stay compatible with whatever subset of fields the caller happens to set. If we decide to change the default value of some of these fields it may break matmul in many different ways.

This is why we need a clear contract and to have it we need a well defined API for what we are trying to achieve here without leaking brgemm ABI.

For reference, I sketched out one possible direction on [dmitriyo/matmul_int8_grouped_quantization_experiment](https://github.com/uxlfoundation/oneDNN/tree/dmitriyo/matmul_int8_grouped_quantization_experiment) just to make the alternative concrete so it's easier to compare tradeoffs. I reused the existing API but reorganized the way it's called to accommodate the feature you introduced.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -2937,6 +2944,179 @@ struct jit_brgemm_matmul_copy_b_int8_t : public jit_brgemm_matmul_copy_b_t,
             vpaddd(v1, v1, vmm_dot_product_temp);
         }
     }
+
+    /**
+    * @brief Applies sign extension for 4-bit signed integers stored in 8-bit lanes using AVX-512.
+    * Due to lack of byte-wise right arithmetic shift(vpsrad) this method is used.
+    * This method checks the sign bit (bit 3) of each byte and fills the upper nibble with 0xF0
+    * if the sign bit is set, preserving the signed semantics after conversion from int4 to int8.
```
dzarukin, line 2952: Let's add that this function converts 1xxxb s4 value into 11111xxxb s8 value and 0xxxb value into 00000xxxb as this are the way they are represented in binary way.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -463,7 +472,8 @@ struct brgemm_matmul_conf_utils_t {
     const bool f32_dt, bf16_dt, f16_dt, f4_via_convert_dt, f8_dt, bf8_dt,
             int8_dt, bf32_dt, tf32_dt;
     const bool weights_decompression_support, bf16_with_int_wei_dt, f32_f16_dt,
-            f32_bf16_dt, f16_with_int_wei_dt, f32_with_int_wei_dt;
+            f32_bf16_dt, f16_with_int_wei_dt, f32_with_int_wei_dt,
+            int8_grouped_quantization_dt;
```
dzarukin, line 493: Random spot in first commit: it says it enables int4 to int8 conversion but the commit seems to do more than that (allowing src scales with groups, but doesn't add this support anywhere in this commit), it might require splitting to avoid that confusion.
Additionally, could you, please, clarify which kind of int4->int8 support is added if int4 seems to be in copy kernels before?

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -470,8 +477,13 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
     const auto &wei_scales = attr->scales_.get(DNNL_ARG_WEIGHTS);
     brg->with_src_scales = !src_scales.has_default_values();
     brg->with_wei_scales
-            = !brg->skip_wei_scales && !wei_scales.has_default_values();
-    if (brg->with_wei_scales) {
+            = !brg->skip_scales && !wei_scales.has_default_values();
```
dzarukin: Looks like a product of bad rebase. Please use `skip_wei_scales` instead everywhere.

src/cpu/x64/brgemm/brgemm.cpp
```diff
@@ -470,8 +477,13 @@ status_t brgemm_desc_set_postops(brgemm_desc_t *brg,
...
+    // In order to parse the mask correctly the `ndims` value is required,
```
dzarukin: I can't see any `ndims` value around. Is this comment still relevant?

src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
```diff
@@ -398,7 +398,7 @@ void jit_brdgmm_kernel_base_t<Wmm>::store_accumulators_apply_post_ops(
 
     if (brg.with_wei_scales) {
         reg_aux_wei_scales.restore();
-        const bool is_single_scale = !brg.is_oc_scale;
+        const bool is_single_scale = brg.is_single_wei_scale;
```
dzarukin: Nit: drop a local variable in a favor of a new one?

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -975,7 +976,11 @@ void jit_brgemm_amx_uker_base_t::apply_alpha_beta_to_vector(
     const bool apply_beta = brg.beta != 0.f;
     if (!apply_alpha && !apply_beta) return;
 
-    const bool dq2ps_required = brg.is_int8 && (apply_alpha || brg.beta != 1.f);
+    // When IC scales (wei or src) are applied, accumulator is already
```
dzarukin: `When K scales...`
There are several references in comments across this commit.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -998,6 +1001,7 @@ void jit_brgemm_kernel_t<Wmm>::read_params() {
 
     mov(reg_C, ptr[param1 + GET_OFF(ptr_C)]);
     mov(reg_D, ptr[param1 + GET_OFF(ptr_D)]);
+    reg_D.save();
```
dzarukin: Do you want to document why `reg_D.save()` is needed?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1370,71 +1374,14 @@ void jit_brgemm_kernel_t<Wmm>::gemv_apply_wei_scales(
...
-    }
-}
-
 template <typename Wmm>
 void jit_brgemm_kernel_t<Wmm>::apply_alpha_beta(
         dim_t bd_block, dim_t ld_block2, bool is_ld_tail, bool is_bdb_tail) {
     const bool apply_alpha = brg.alpha != 1.f;
-    const bool dq2ps_required = brg.is_int8 && (apply_alpha || brg.beta != 1.f);
+    // apply_scales() already converted the accumulator to f32.
+    const bool is_integer_acc = brg.is_int8 && !brg.has_per_k_scales();
```
dzarukin: Minor: maybe introduce `is_integer_acc()` and `do_dq2ps_cvt()` based on `brg` state as it seems they are reused across different kernels?

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
```diff
@@ -60,6 +60,25 @@ void matmul_amx_blocking_params_t::update_configuration(
     bgmmc.is_macro_heuristics
             = dynamic_cast<const matmul_amx_blocking_params_macro_t *>(this)
             != nullptr;
+
+    // Force K_blk alignment to scales group size for AMX per K scales.
+    // Each brgemm call must cover exactly one scale group so that
+    // post-gemm-ops can apply a single scale row per call.
+    if (bgmmc.with_int8_grouped_quantization) {
+        dim_t ic_group_sz = 0;
```
dzarukin: `k_group_size` instead?

tests/benchdnn/matmul/cfg.cpp
```diff
@@ -29,6 +29,19 @@ cfg_t::cfg_t(const prb_t *prb, const std::vector<data_kind_t> &kinds)
...
+    // that byte-wise (wei - zp) in the copy_b kernel does not underflow the
+    // unsigned byte domain. Filler for non-single-element ZPs uses [.., 2].
+    const auto &wei_zp = prb->attr.zero_points.get(DNNL_ARG_WEIGHTS);
+    const bool wei_is_unsigned
+            = prb->get_dt(WEI) == dnnl_u8 || prb->get_dt(WEI) == dnnl_u4;
+    if (!wei_zp.is_def() && wei_is_unsigned && cfg_entry_.count(WEI)) {
+        const int zp_max = wei_zp.has_single_element() ? wei_zp.value : 2;
+        if (zp_max > 0 && get_range_min(WEI) < zp_max)
+            set_range_min(WEI, zp_max);
+    }
```
dzarukin: Copy kernel can't apply zero points for int8xint8 scenario exactly for the reason of potential overflow, and it's the reason why compensation exists.

Copy kernel can apply zero-points directly in two scenarios:
* Apply int4 zp after converting src/wei data point and zp from int4 to int8.
* Apply int8 zp after converting wei data point and zp from int8 to xf16 (weights-only-quantization).

Other scenarios require compensation support. Please drop benchdnn change and adjust the kernel logic instead.

dzarukin (commented):
Though this version looks better than the previous one, I think this PR must be broken up into independent features since it mixes up many things together. I envision the following split:
* batch wei grouped scales support for WoQ (PR open).
* (batch) wei grouped zp support for WoQ. It would apply zp in copy_b kernels directly. Format support for wtag=abx and wtag=ba,acb,abdc can be split if needed.
* brgemm new kernel scheme with per_k + brgemm driver validation
* introducing grouped_quant support with SRC SCALES ONLY - all new members to support int8 x int4/int8 flow. Might require limiting zp support for weights since for int8 weights they can't be applied directly.
* Introduce wei zp support for group quant. Split wtag support if needed.
* Introduce src_zp support for group quant.

Each of feature would enable a portion of existing validation and can be validated properly reducing the risk of introducing functional bugs.

It gets hard to provide meaningful feedback starting from 4th commit.

tests/benchdnn/matmul/cfg.cpp
inteldimitrius: There's a failing cases with unsigned quantized integer, when the ZP is generated bigger than unsigned WEI, and it causes underflow. This config adjusts range min value when ZP is set for unsigned datatype. It's not about compensation and datatype matching.

ZP should be at most min(WEI), otherwise correctness will cost additional runtime checks.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: Before was `int4 -> {f16, bf16, f32}`.
Now: `int4->int8`.
The main difference is there's no special instruction supported. Especially related to signed case, since there's no signed left shift operation available.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -151,6 +163,95 @@ struct jit_brgemm_matmul_copy_a_impl_t : public jit_brgemm_matmul_copy_a_t,
...
+        mov(ptr[rsp + reg_src_save_offs_], reg_src);
+
+        // Compute absolute K position = current_K_start + k_idx * k_step_
+        mov(rax, ptr[rsp + reg_src_zp_cur_k_offs_]);
+        add(rax, k_idx * k_step_);
+
+        // Compute K-group index: k_group_idx = abs_k / gsize
+        xor_(rdx, rdx);
+        mov(regq_tmp, gsize);
+        div(regq_tmp);
```
densamoilov:
```suggestion
       if (math::is_pow2(gsize)) {
           shr(rax, math::ilog2q(gsize));
        } else {
            xor_(rdx, rdx);
            mov(regq_tmp, gsize);
            div(regq_tmp);
         }
```

I don't know how much it could save in general, but it's basically a zero risk optimization and the group size is a power of 2 anyway.

Same optimization could be applied to `maybe_apply_int8_per_k_zp`.

tests/benchdnn/matmul/cfg.cpp
dzarukin: Could you, please, provide a reproducer?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
dzarukin: Is it a two-stage conversion after this change for weights-only-quantization?
int4->int8 isn't helping for it (at least, I don't see why issuing more instructions would be more beneficial) and only relevant when src is int8 too, but this commit doesn't provide such capabilities.

tests/benchdnn/matmul/cfg.cpp
densamoilov: If I understand it right, the current problem is that wei(int8) - zp(int8) can cause underflow (I actually think that that applies to both signed and unsigned cases). If so, then the problem here is that we can control our test data but can't control deployment data, this is why we can't have a kernel that does incorrect math.
Looks like there are two options:
1. Support it via compensation
2. Fallback to the reference for that particular case if it wasn't requested

Maybe I'm missing something but I don't see those problematic cases in harness_matmul_weight_decompression.

densamoilov: We discussed this problem offline. The issue arises when wei (uint8) > zp (uint8), causing an underflow because we use `vpsubb` to subtract the zero points. Since `vpsubb` is a non-saturating instruction, we either need to use saturating subtraction or adopt a different approach.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2155,6 +2257,22 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
...
+        const auto dt_sz = types::data_type_size(bgmmc_.src_zp_dt);
+        const auto is_int4
+                = one_of(bgmmc_.src_zp_dt, data_type::s4, data_type::u4);
+
+        if (is_int4) { return src_zp_ptr_; }
+
+        const auto m_stride = num_k_groups * dt_sz;
+        auto offset = m * m_stride;
+        return ((const char *)src_zp_ptr_ + offset);
+    }
```
georgen117: for `get_src_zp_ptr` the 'k' parameter is declared but never used and likely should be removed.

The **int8** and **int4** paths handle indexing inconsistently:
- **int4**: returns the base pointer `zrc_zp_ptr_` unchanged, ignoring both `m` and `k`.  The kernel (`mabye_apply_src_zp`) reconstructs the full flat index itself using `current_M_start` and `current_K_start` from the stack.
- **int8**: advances the pointer by `m * (num_k_groups * dt_sz)`, applying the M-row offset. The k-group offst within the row is then computed inside the kernel.

The use of `get_src_zp_ptr` is hard to reason about: for **int8** the caller must supply the correct `m` or the pointer is wrong; for **int4** `m` is irrelevant.  The `k` parameter suggests the caller also controls k-group selection, but never does for either type.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -452,6 +570,8 @@ template <typename Vmm>
 void jit_brgemm_matmul_copy_a_impl_t<Vmm>::generate() {
     preamble();
```
georgen117, line 454: The `vmm_src_zp` register aliases the same register as `vmm_comp_mul` (see lines ~133-148 of this file, brgemm_matmul_copy_utils.cpp )

This is safe today because `do_compute_compensation_` is forced to false when `with_int8_groupd_quantization`  is true.  However, the aliasing is a maintenance trap. 

Suggestion: add an assert to maintain the exclusive use of the register

```cpp
 // vmm_src_zp and vmm_comp_mul share the same physical register.
 // They are safe only because the two paths are mutually exclusive:
 // s8s8 compensation (do_compute_compensation_) uses vmm_comp_mul,
 // per-K src ZP (do_apply_src_zp_) uses vmm_src_zp.
 assert(IMPLICATION(do_compute_compensation_, !do_apply_src_zp_)
         && "vmm_comp_mul and vmm_src_zp alias the same register");
```

**Reviewer note*: perhapse you can come up with a better warning.  The one I wrote likely mentions too much internal details.

The assertion belongs here rather than at the class member declaration site because `do_computer_compenation_` and `do_apply_src_zp` are runtime booleans set in the constructor. `generate()` is called once per kenel creation not per-inference, and it is the earliest point where both flags are live and accessible.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Fixed!

src/cpu/x64/brgemm/brgemm.hpp
inteldimitrius: Okay, will take your version for now! Thanks!

tests/benchdnn/matmul/cfg.cpp
inteldimitrius: Yes, @densamoilov  is right. The only note here is it's only for **unsigned** int8. The question, that I have rn, is do we need to saturate it or add this configuration in benchdnn to modify ranges.

inteldimitrius: After another discussion it was concluded, that the original range is valid and `ZP_i > WEI_i` possible for all data types, so I will revert this change and use saturated `vpsubusb` in kernels.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: That's a valid point, `vmm_src_zp` does corrupt `vmm_comp_mul`, but these features are planned to work both together, so I suggest to add restoring for this case:
```cpp
    auto maybe_compute_compensation = [this, num_acc](int k_idx, Vmm vmm_copy) {
        if (do_compute_compensation_) {
           // *** Added lines here   
           if (do_apply_src_zp_) {
                mov(imm_addr64, 1);
                uni_vpbroadcastb(vmm_comp_mul, imm_addr64.cvt8());
            }
            // end of edit ***
            const auto vmm_comp_acc = get_vmm_comp_acc(k_idx % num_acc);
            if (conf_->src_dt == data_type::s8)
                dot_product(vmm_comp_acc, vmm_comp_mul, vmm_copy);
            else
                dot_product(vmm_comp_acc, vmm_copy, vmm_comp_mul);
        }
    };
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Thanks, added.

tests/benchdnn/matmul/cfg.cpp
dzarukin: I would clarify the behavior and the target configuration with requestors first, then apply changes once there's clarification.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: `int8_grouped_quantization_dt` is required here to use it kernel's logic. However I moved "enabling" in matmul_utils.cpp to the last commit, so the code is unreachable until the last commit.
Is this solution good enough?

src/cpu/x64/brgemm/brgemm_types.hpp
inteldimitrius: Removed, thanks!

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
inteldimitrius: Thanks! Fixed!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: It's redundant line, moved here from previous version. Will be removed on the next edit.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Done, thanks!

src/cpu/x64/matmul/amx_blocking_heuristics.cpp
inteldimitrius: Renamed, thanks!

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Yes, thanks! Removed unused `k` argument.

tests/benchdnn/matmul/cfg.cpp
inteldimitrius: I have some takes here, which we partially discussed with @densamoilov :
- there're cases when `ZP_u4 > WEI_u4` cause underflow in `(WEI_u8 - ZP_u8)` **unsigned int8** datatype.
- GPU implementation converts int4->f16 and no underflow happened.
- Ref implementation uses f32, no underflow.
- I tried saturated `vpsubusb` subtraction, but it fails because ref-impl doesn't do it. 

I found the less painful solution for it - to convert `u4->s8`, we could reuse all AMX/avx512 possibilities, but with signed dot-product. What do you think guys? I pushed into the branch, so it's the last commit. 

Saturation also truncates the result and it could be critical.

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
inteldimitrius: Done, thanks!

src/cpu/x64/brgemm/brgemm.cpp
inteldimitrius: Updated, thanks. It's just renaming, not sure if it will affect on anything,

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: Added to jit_generator.hpp.

src/cpu/x64/brgemm/brgemm.cpp
inteldimitrius: Removed outdated commit. Thanks!

src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp
```diff
@@ -1457,6 +1499,42 @@ void jit_brgemm_amx_uker_base_t::process_output_range(
...
+            vcvtdq2ps(zmm, zmm);
+
+            const Xbyak::Zmm scaled_zmm = vmm_mask(zmm, true, false, k_mask);
+            // Apply K-wei_scales if present (pre-loaded per-ldb vector).
+            if (brg.is_per_k_wei_scales) {
+                vmulps(scaled_zmm, scaled_zmm, zmm_scales(ldb));
+            }
+            // Apply K-src_scales per-bd: each M-row has its own scalar.
+            if (brg.is_per_k_src_scales) {
+                reg_src_scales_ic.restore();
```
densamoilov: Please use brgemm notation (bs, m, n, k, bd, rd, ld, etc) for all variable names

src/cpu/x64/brgemm/brgemm_types.hpp
inteldimitrius: Updated commit messages, now: "cpu: matmul+brgemm: int8 Add XXX scales support".

inteldimitrius:
make test perf-cpu
disable arch_cpu_adl
disable arch_cpu_icx
disable arch_cpu_spr
disable arch_cpu_srf
set primitive=matmul

inteldimitrius:
make test
disable benchdnn_all
enable benchdnn_matmul

inteldimitrius:
make test perf-cpu
disable arch_cpu_adl
disable arch_cpu_icx
disable arch_cpu_spr
disable arch_cpu_srf
set primitive=matmul

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+            const auto addr = ptr[reg_a + c * n_lanes];
+            if (isa_has_masks(isa_)) {
+                build_chunk_mask(reg_iter);
+                if (wei_is_signed_)
+                    vpmovsxbd(vmm_data_sr() | k_mask | T_z, addr);
+                else
+                    vpmovzxbd(vmm_data_sr() | k_mask | T_z, addr);
+            } else {
+                if (wei_is_signed_)
+                    vpmovsxbd(vmm_data_sr(), addr);
```
densamoilov: We use masked versions only for avx512, is it expected?

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2110,7 +2233,8 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
         return bias_ptr_ + n * bgmmc_.bias_dt_sz;
     }
 
-    int32_t *get_s8s8_comp_ptr(int ithr, int b, int n_blk_idx) const {
+    int32_t *get_s8s8_comp_ptr(
+            int ithr, int b, int n_blk_idx, int k_blk_idx = 0) const {
```
densamoilov: `k_blk_idx` seems to be unused. Is it a bug or do we not care about the k index?

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -297,17 +297,28 @@ struct brgemm_desc_t {
...
     brgemm_broadcast_t zp_type_b = brgemm_broadcast_t::none;
     brgemm_broadcast_t zp_type_c = brgemm_broadcast_t::none;
 
-    // `skip_wei_scales` is controlled by the implementation and not by kernel
-    // API.
     bool skip_wei_scales = false;
-    int is_oc_scale = 0;
+    int is_single_wei_scale = 0;
+    int is_per_n_wei_scales = 0;
+    int is_per_k_wei_scales = 0;
```
densamoilov: nit: `bool`?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1760,31 +1768,179 @@ void jit_brgemm_kernel_t<Wmm>::reduce_gemv_accumulators(dim_t bd_block) {
...
+                    uni_vpslld(vmm_scales, vmm_scales, 16);
+                    break;
+                case data_type::f16: vcvtph2ps(vmm_scales_masked, addr); break;
+                default: assert(!"unsupported scales data type");
+            }
+        } else {
+            assert(one_of(
+                    type_in, data_type::f32, data_type::bf16, data_type::f16));
+            const dim_t tail_size = brg.is_gemv
+                    ? (brg.gemv_acc_is_vector()
```
densamoilov: We have a dedicated function that's responsible for applying scales in the GEMV code path. It seems that this new `load_scales_to_vmm` is never used in the GEMV code path.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1800,11 +1956,10 @@ void jit_brgemm_kernel_t<Wmm>::store_accumulators_apply_post_ops(dim_t bd_block,
         dq2ps_cvt_done = true;
     }
 
-    if (brg.with_wei_scales) {
+    if (brg.with_wei_scales && !brg.is_per_k_wei_scales) {
         if (!brg.is_gemv) {
-            apply_wei_scales(bd_block, ld_block2, dq2ps_required,
-                    dq2ps_cvt_done, is_ld_tail);
-            dq2ps_cvt_done = true;
+            apply_wei_scales(bd_block, ld_block2, is_ld_tail, dq2ps_required,
```
densamoilov, line 1941: It seems that this function and all the functions it calls have GEMV specific code but it's never used for GEMV because we handle it in `gemv_apply_wei_scales`.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -1760,31 +1768,179 @@ void jit_brgemm_kernel_t<Wmm>::reduce_gemv_accumulators(dim_t bd_block) {
...
     const bool alpha_or_beta_applicable = brg.alpha != 1.0f || brg.beta != 0.f;
     const bool beta_uses_vadd
             = brg.beta == 1.f && IMPLICATION(brg.is_int8, brg.alpha == 1.0f);
     const bool dq2ps_required = brg.is_int8
-            && IMPLICATION(alpha_or_beta_applicable, beta_uses_vadd);
-    const bool has_ptr_b_support = is_superset(brg.isa_impl, avx512_core);
+            && IMPLICATION(alpha_or_beta_applicable, beta_uses_vadd)
+            && !brg.has_per_k_scales();
+    const bool has_ptr_b_support = is_superset(brg.isa_impl, avx512_core)
+            && brg.dt_src_scales == data_type::f32;
```
densamoilov, line 1908: Can you please clarify this line with a comment?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2116,6 +2271,40 @@ void jit_brgemm_kernel_t<Wmm>::apply_compensation(
     }
 }
 
+template <typename Wmm>
+void jit_brgemm_kernel_t<Wmm>::apply_per_mn_compensation(
+        dim_t bd_block, dim_t ld_block2, bool is_ld_tail) {
+    assert(brg.with_per_mn_compensation);
+    assert(!brg.is_gemv && "per-(M,N) compensation is not supported for gemv");
```
densamoilov, line 2258: should we add `assert(!brg.integer_acc())` just in case?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2283,7 +2472,8 @@ void jit_brgemm_kernel_t<Wmm>::store_accumulators(dim_t bd_block2,
                         }
 
                         if (apply_zp_a_compensation)
-                            apply_compensation(adj_bd_block, 1, is_ld_tail);
+                            apply_zp_int_compensation(
```
densamoilov: Can you please clarify this renaming?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2560,7 +2755,7 @@ void jit_brgemm_kernel_t<Wmm>::maybe_tileloadd_nt(matrix_kind_t matrix_kind,
                                : brg.get_B_tensor(idx, is_tail);
     auto t1 = Tmm(tmm_idx);
 
-    auto reg_base = is_A ? reg_aux_A : reg_aux_B;
+    auto &reg_base = is_A ? reg_aux_A : reg_aux_B;
```
densamoilov: nit: a register is a tiny object, is there any specific reason we create a reference? Should we make it const too?

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -3450,6 +3645,15 @@ void jit_brgemm_kernel_t<Wmm>::bdb_loop() {
         }
         add(reg_a_offset, bdb_A_offset(bd_block2));
 
+        if (brg.is_per_k_src_scales) {
+            const auto adj_bd_block = is_bdb_tail ? brg.bdb_tail : brg.bd_block;
```
densamoilov, line 3631: nit: i think we generally stick with  the `XX_block` name when we select a tail vs full block.

```suggestion
            const auto bd_block = is_bdb_tail ? brg.bdb_tail : brg.bd_block;
```

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2187,6 +2343,67 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
...
+        // Per-thread tile mirrors buffer_c's stripe layout for AMX:
+        // ceil(N_blk / LDC) stripes of M_blk × LDC floats.
+        const size_t per_thr_elems = static_cast<size_t>(bgmmc_.M_blk)
+                * static_cast<size_t>(rnd_up(bgmmc_.N_blk, bgmmc_.LDC));
+        return per_mn_comp_ptr_ + ithr * per_thr_elems;
+    }
+
+    // Fills the per-thread per-(M, N) compensation tile with the symmetric
+    // src+wei zero-point correction delta[m,n]
+    void fill_per_mn_compensation(int ithr, int b_idx, int m_blk_idx,
```
densamoilov: I don't think placing a `per_mn_comp_kernel_` call in `brg_matmul_exec_ctx_t` is the right thing to do. The abstraction implies that it should contain execution time parameters and interfaces to access them, meaning that it should not be responsible for any actions with the input/output data directly as well as storing pointers to kernel objects that are created at the primitive construction time.

Is there any issue with having it outside of `brg_matmul_exec_ctx_t`?

note: we have some exceptions like `maybe_restore_dst_values_(to|from)_buffer` and maybe something else but they shouldn't have been added there.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -5312,14 +5880,23 @@ void jit_brgemm_matmul_copy_b_transposed_t<Vmm>::compute_K_loop(bool is_N_tail,
...
+    // a complete, finalized comp value (raw sum * 128, sign-flipped, +1)
+    // independent of other chunks.
+    const bool comp_per_chunk = conf_->with_int8_grouped_quantization;
+    const bool eff_first = comp_per_chunk ? true : is_first_K_iter;
+    const bool eff_last = comp_per_chunk ? true : is_last_K_iter;
+
     if (req_s8s8_comp_) {
         const auto addr = ptr[reg_comp_ptr];
-        if (!is_first_K_iter)
+        if (!eff_first)
```
densamoilov: What is `eff` here? :)

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1730,9 +1754,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
         bgmmc.wei_dt = f16;
         bgmmc.tr_a_dt_sz = types::data_type_size(f16);
         bgmmc.tr_b_dt_sz = types::data_type_size(f16);
+    } else if (bgmmc.with_int8_grouped_quantization) {
```
densamoilov, line 1757: Are `bgmmc.is_f4_via_convert` and `bgmmc.with_int8_grouped_quantization` not mutually exclusive? Can they both be true?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1730,9 +1754,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+                = bgmmc.src_dt == s8 && !isa_has_s8s8(isa);
     }
 
-    bgmmc.acc_dt = bm_conf_utils.is_int8() ? s32 : f32;
+    // Grouped int8 quantization accumulates in f32: per-K scale/ZP
+    // application happens between brgemm calls and requires float
+    // accumulation (per-MN comp subtracts a float delta each K-group).
+    const bool grouped_quant_needs_f32_acc
+            = bgmmc.with_int8_grouped_quantization;
+    bgmmc.acc_dt = (bm_conf_utils.is_int8() && !grouped_quant_needs_f32_acc)
```
densamoilov: maybe just use `! bgmmc.with_int8_grouped_quantization`? The comment above already makes it clear.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -2536,6 +2662,8 @@ void init_aux_values(brgemm_matmul_conf_t &bgmmc,
     bgmmc.has_zero_point_a = bgmmc.src_zp_type != brgemm_broadcast_t::none;
     bgmmc.has_zero_point_b = bgmmc.wei_zp_type != brgemm_broadcast_t::none;
     bgmmc.has_zero_point_c = bgmmc.dst_zp_type != brgemm_broadcast_t::none;
+    bgmmc.with_per_mn_compensation = bgmmc.with_int8_grouped_quantization
```
densamoilov, line 2664: Is this compensation always needed for `with_int8_grouped_quantization` regardless of the data types configuration?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -274,14 +286,27 @@ struct brgemm_matmul_conf_t {
 
     data_type_t src_zp_dt = data_type::undef;
 
+    // Per-K src zero-points: stride between K-group slots (in
+    // elements). Used by the per-mn compensation src ZP gather.
+    bool is_src_zp_per_k = false;
+    bool is_src_zp_per_m = false;
+    dim_t src_zp_k_gsize = 0;
```
densamoilov, line 293: Is this `stride between K-group slots`?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -274,14 +286,27 @@ struct brgemm_matmul_conf_t {
...
 
+    // Batched (4D) per-tensor grouped scales/ZP plane stride.
+    dim_t wei_zp_batch_stride = 0;
+    dim_t wei_scales_batch_stride = 0;
+
     dim_t zp_a_comp_shift_n;
     dim_t zp_a_comp_elems_per_thr;
+    // Stride between per-K-group zp_a compensation slots (in elements).
+    // Used by the int8 grouped quantization src ZP post-brgemm path.
+    dim_t zp_a_comp_k_str = 0;
```
densamoilov: Initialized but never used?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -274,14 +286,27 @@ struct brgemm_matmul_conf_t {
...
+
     dim_t wei_zp_k_gsize = 0;
     bool is_wei_zp_per_k = false;
     bool is_wei_zp_per_n = false;
     bool is_wei_zp_common = false;
     data_type_t wei_zp_dt = data_type::undef;
 
+    // Batched (4D) per-tensor grouped scales/ZP plane stride.
+    dim_t wei_zp_batch_stride = 0;
+    dim_t wei_scales_batch_stride = 0;
```
densamoilov: Unused?

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+    const dim_t src_m_stride_bytes_;
+    const dim_t src_k_stride_bytes_;
+    const dim_t wei_k_stride_bytes_;
+    const dim_t LDC_;
+    const dim_t M_blk_static_;
+    const int chunks_;
+
+    static constexpr int f32_sz = sizeof(float);
+    static constexpr int i32_sz = sizeof(int32_t);
+    static constexpr int n_lanes = vreg_traits_t<Vmm>::vlen / i32_sz;
```
densamoilov: nit: we don't really use the term "lanes" in the library, we use `simd_w` instead.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+        float *delta_ptr;
+        dim_t m_base;
+        dim_t n_base;
+        dim_t M_blk;
+        dim_t stripe_w;
+        dim_t k_len;
+        // Scalar inputs already cast to f32 by the cpp-side.
+        float src_zp_f32; // unused when !has_src_zp_
+        float wei_zp_f32; // unused when wei_zp_per_n_ or !has_wei_zp_
+        float Gzs_f32; // = K_len * src_zp_f32; unused unless both zps present
```
densamoilov: what does `Gzs` stand for?

inteldimitrius:
make test
disable benchdnn_all
enable benchdnn_matmul

inteldimitrius:
make test perf-cpu
disable arch_cpu_adl
disable arch_cpu_icx
disable arch_cpu_spr
disable arch_cpu_srf
set primitive=matmul

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+    const bool wei_zp_is_signed_;
+    const bool src_is_signed_;
+    const bool wei_is_signed_;
+    const bool wei_is_subbyte_;
+    const bool has_vnni_;
+    const dim_t src_m_stride_bytes_;
+    const dim_t src_k_stride_bytes_;
+    const dim_t wei_k_stride_bytes_;
+    const dim_t LDC_;
+    const dim_t M_blk_static_;
```
densamoilov: dead member

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+
+#define GET_OFF(x) offsetof(call_params_t, x)
+
+    const cpu_isa_t isa_;
+    const bool has_src_zp_;
+    const bool has_wei_zp_;
+    const bool wei_zp_per_n_;
+    const data_type_t wei_zp_dt_;
+    const bool wei_zp_is_int4_;
+    const bool wei_zp_is_signed_;
```
densamoilov: dead member

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+// Per-axis (src zp/sc, wei zp/sc) supply:
+//   * scalar f32 (per-tensor or per-K-only)  -> cpp-side pre-converts
+//   * vector load from typed user buffer    -> JIT loads, casts to f32
+//
+// The delta tile is the only output buffer.
+template <typename Vmm>
+struct jit_per_mn_comp_kernel_t : public jit_generator_t {
+    DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_per_mn_comp_kernel_t)
+
+    struct call_params_t {
```
densamoilov: The kernel structure is over-layered. We have 3 layers for a single compensation kernel. I can't see why we need all of them unless I'm missing something.

* We should use inheritance and not composition to get rid of the redundant layer with the kernel: `per_mn_comp_kernel_impl_t::gen_`;
* And then we won't need to reconstruct `call_params_t` object from `per_mn_comp_kernel_t::ctx_t` and re-derive other things

This will leave us with a two-layer design, similar to how the copy kernels are currently designed.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+        // Persistent f32 broadcasts loaded once for the whole call.
+        if (has_src_zp_) bcast_f32_param(vmm_zs_bc(), GET_OFF(src_zp_f32));
+        if (has_src_zp_ && has_wei_zp_)
+            bcast_f32_param(vmm_Gzs_bc(), GET_OFF(Gzs_f32));
+        if (has_wei_zp_ && !wei_zp_per_n_)
+            bcast_f32_param(vmm_zw(), GET_OFF(wei_zp_f32));
+
+        // int4 wei_zp helpers: load perm table & 0x0F broadcast.
+        if (wei_zp_per_n_ && wei_zp_is_int4_) {
+            alignas(64) static constexpr const uint32_t int4_interleave_dw[16]
```
densamoilov: we could probably make it a class member as it's re-used in two places

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+                    = {0, 16, 1, 17, 2, 18, 3, 19, 4, 20, 5, 21, 6, 22, 7, 23};
+            mov(reg_tmp,
+                    reinterpret_cast<uintptr_t>(
+                            static_cast<const void *>(int4_interleave_dw)));
+            vmovdqu32(vmm_int4_perm(), ptr[reg_tmp]);
+            mov(reg_tmp.cvt32(), 0x0F);
+            vpbroadcastd(vmm_int4_0F(), reg_tmp.cvt32());
```
densamoilov: We should probably also move it to a function and use it to initialize `vmm_int4_perm()` and `vmm_int4_0F()` because these are the same as `vmm_perm_sr()` and `vmm_0f_sr()`.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,868 @@
...
+                    cmp(reg_iter, n_lanes);
+                    Label cap_ok;
+                    jle(cap_ok, T_NEAR);
+                    mov(reg_iter, n_lanes);
+                    L(cap_ok);
+                    inc(reg_iter);
+                    shr(reg_iter, 1);
+                    mov(reg_n, -1);
+                    bzhi(reg_n, reg_n, reg_iter);
+                    kmovw(k_byte_mask, reg_n.cvt32());
```
densamoilov: maybe move this block to `build_subbyte_byte_mask(reg)`? There is the same block at 443-452.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Do you mean better renaming? Or adding comments? "int" means here integer dt, I can suggest "apply_common_zp_compensation" as the  opposite of "apply_per_mn_compensation".

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Specifically here is the product of bad rebasing, I will remove this change, thanks! 
The motivation was proper using of `reg64_savable_t`.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Yes, regardless of data types, only zero point presence is matter: 
```cpp
bgmmc.with_per_mn_compensation = bgmmc.with_int8_grouped_quantization
            && (bgmmc.has_zero_point_a || bgmmc.has_zero_point_b);
```

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius:
```cpp
    , f4_via_convert_dt(utils::one_of(bgmmc.wei_dt, data_type::f4_e2m1,
                                data_type::f4_e3m0)
              && isa == avx10_1_512)
   ...
    , int8_grouped_quantization_dt(one_of(bgmmc.src_dt, u8, s8)
              && one_of(bgmmc.wei_dt, s4, u4, s8)
              && one_of(bgmmc.dst_dt, f16, bf16, f32, s8, u8)
              && (!attr.zero_points_.get(DNNL_ARG_SRC).has_default_values()
                      || !attr.zero_points_.get(DNNL_ARG_WEIGHTS)
                                  .has_default_values())
              && (utils::one_of(bgmmc.wei_dt, s4, u4)
                      || !attr.scales_.get(DNNL_ARG_SRC).has_default_groups()
                      || !attr.scales_.get(DNNL_ARG_WEIGHTS)
                                  .has_default_groups()
                      || !attr.zero_points_.get(DNNL_ARG_SRC)
                                  .has_default_groups()
                      || !attr.zero_points_.get(DNNL_ARG_WEIGHTS)
                                  .has_default_groups())
              && bgmmc.ndims
                      == 2) // 3D/4D cases will be enabled in separate PRs.
```

Yes, they do not intersect on data types. so they are mutually exclusive.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
inteldimitrius: There was `is_first_K_iter` and `is_last_K_iter`, requiring special processing, however after enabling compensation per chunk, these flags are adjusted:
```cpp
const bool comp_per_chunk = conf_->with_int8_grouped_quantization;
const bool do_first = comp_per_chunk || is_first_K_iter; // This was previously named eff_first
const bool do_last = comp_per_chunk || is_last_K_iter; // This was previously named eff_last
```
Yes, the naming was bad, so I renamed it `do_last` & `do_first` instead of previous naming.

src/cpu/x64/brgemm/brgemm.hpp
inteldimitrius: This is issue was fixed by moving to per-mn-compensation, so this thread is could be resolved.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Yes, AVX2 doesn't support masked loading, using another approach for avx2.

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Yes, product of the previous versions. Removed. Thanks!

src/cpu/x64/brgemm/brgemm_types.hpp
inteldimitrius: Updated, thanks.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Removed all gemv-related branches.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Removed!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Cleaned!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Flattened the structure, removed call_params_t, also removed `_impl` wrapper. Thanks!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Made it static on the class level + added re-usable `init_int4_helpers()`. Thanks!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Updated!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Moved out into a separate method, invoked in 2 places. Thanks!

inteldimitrius:
@dzarukin Since the feature is "turned on" via a single flag `with_int8_grouped_quantization` splitting into several PRs leads to dead code merging and disability to test.
On top of that several PRs would require strong order, because of reusing same flags as `per_mn_compensation`. 
The best solution is to merge it as is, keeping the order of commits and changes consistent among all code changes in the lib.

inteldimitrius:
make test perf-cpu
disable arch_cpu_adl
disable arch_cpu_icx
disable arch_cpu_spr
disable arch_cpu_srf
set primitive=matmul

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: It's a group size per K for src ZP, it's used for striding and offset computation in `get_src_zp_ptr()`,

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: It's `group_size * src_zp` term, which is later used in delta. I don't know a specific name for it. I also can suggest `grouped_src_zp_factor`. Or smth similar.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius:
```bash
(base) bash-5.1$ grep -r -n adj_bd_block  src/
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:421:    void advance_bdb_post_op_regs(dim_t adj_bd_block);
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:912:void jit_brgemm_kernel_t<Wmm>::advance_bdb_post_op_regs(dim_t adj_bd_block) {
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2412:            const dim_t adj_bd_block = (brg.is_M_tail && is_bdb_tail)
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2438:                            for (dim_t bd = 0; bd < adj_bd_block; bd++) {
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2445:                            for (dim_t bd = 0; bd < adj_bd_block; bd++) {
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2458:                                    adj_bd_block, 1, is_ld_tail);
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2462:                                    adj_bd_block, 1, is_ld_tail, is_bdb_tail);
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2467:                            store_accumulators_apply_post_ops(adj_bd_block, 1,
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2476:                                    adj_bd_block, 1, is_ld_tail, is_bdb_tail);
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:2516:                        advance_bdb_post_op_regs(adj_bd_block);
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:3631:            const auto adj_bd_block = is_bdb_tail ? brg.bdb_tail : brg.bd_block;
src/cpu/x64/brgemm/jit_brgemm_kernel.cpp:3635:                    = adj_bd_block * bd_block2 * brg.src_scale_m_stride;
(base) bash-5.1$ 
```
The naming was introduced earlier, Do I need to change all entries or just mine?

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Moved out into `brgemm_matmul_t` method. Thanks!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Added!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Added a comment, thanks! `has_ptr_b_support` is active only when the `vmulps` support memory operand, so it possible to multiply right out of memory.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: Okay, it's up to you if you want to make this change at all. It seems that `adj_bd_block` is used only in the AMX code path.
I was thinking about this line where we define `bd_block` and pass it down to a bunch of function where it's considered just a `bd_block`. It's a preference, not an argument so it's your call.
https://github.com/uxlfoundation/oneDNN/blob/ce50f37eb9817e9405f5ed075a36f6e8b1c9befb/src/cpu/x64/brgemm/jit_brgemm_kernel.cpp#L2394

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
densamoilov: I guess what confused me is the comment above.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
densamoilov: Okay, then we don't have to put it under `else` because if we do it may look like these branches are not mutually exclusive.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
densamoilov: Oh, okay. Should it be Gsz_f32 instead of Gzs_f32 then?

`g_sz_f32` looks a bit more aligned with the surrounding code style though

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
densamoilov: My question was mostly about correctness of the code. I assumed that if we need using mask in the avx512 branch then we'd also need to use mask in the avx2 branch. But currently we're not using it in the avx2 branch.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: I just wanted to understand what was the driver for the renaming. But I think you answered it now.

If I understand it right it works this way now:
* `apply_zp_int_compensation` works with integer accumulators and it can do one or two things:
  1. Apply compensation in the case of s8s8
  2. Apply compensation in the case of s8s8 + apply zp compensation
* `apply_per_mn_compensation` takes integer accumulators and converts them to float accumulators and applies the pre-computed zp compensation

So if we need s8s8 compensation and per_mn_compensation then both of the functions will be called.

So having `zp` in the name of `apply_zp_int_compensation` is misleading because it's only one of the two actions the function can take.
And having `int` is also misleading because technically both functions take integer accumulators.

Maybe `apply_zp_s8s8_compensation` could be better here.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
densamoilov: Thanks!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
```diff
@@ -0,0 +1,834 @@
...
+        L(idx_table_);
+        for (int i = 0; i < simd_w; ++i)
+            dd(i);
+    }
+
+#undef GET_OFF
+};
+
+template <typename Vmm>
+void jit_per_mn_comp_kernel_t<Vmm>::operator()(ctx_t *ctx) const {
```
densamoilov: The `ctx` should be taken by const pointer here. Currently, you modify some parts of `ctx` during this call, which is not correct because `ctx` contains `input` parameters. This `operator` is simply supposed to forward `ctx` to the jit kernel (smallest building block) and the driver (harness) of the jit kernel should be implemented in `fill_per_mn_compensation`.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
densamoilov: This case fails:
`DNNL_MAX_CPU_ISA=AVX2_VNNI ./benchdnn --matmul --stag=ab --wtag=ab --dt=s8:s8:f32 --attr-zero-points=wei:per_ocic:s8:32x1 16x256:256x17`

Do we have cases like this in the the input files so that we can catch such issues in validation?

It looks like this kernel precomputes compensation using user-provided weights, which may be in either blocked or plain format. When the format is blocked, I think we will simply read from the padded area, which should be fine. However, if the weights are in plain format, we may read past the boundary.

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Done, thanks!

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Fixed, thanks!

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
inteldimitrius: Renamed to `apply_zp_s8s8_compensation`.

densamoilov (approved):

src/cpu/x64/matmul/jit_brgemm_matmul_per_mn_comp.cpp
inteldimitrius: Renamed.

inteldimitrius:
Failed cases are related to aarch64 implementation and not reproducible on x64. I think it should be fixed separately.

However, it fails on NEWLY added test cases, which are testing tail processing per N. 

UPD: reported an issue related to it: [5377](https://github.com/uxlfoundation/oneDNN/issues/5377)

ankalinin (approved):

inteldimitrius:
make test perf-cpu
disable arch_cpu_adl
disable arch_cpu_icx
disable arch_cpu_spr
disable arch_cpu_srf
set primitive=matmul

inteldimitrius:
make test
disable benchdnn_all
enable benchdnn_matmul

PR #5020, opened 2026-04-15
cpu: x64: matmul: fix VDISPATCH_REORDER_IC use with memory_desc_reshape

asztukow:
Addresses MFDNN-14804

`VDISPATCH_REORDER_IC` expects a boolean but `memory_desc_reshape()` returns a `status_t`. Since `status::success == 0`, a success was evaluated as false, causing the macro to return `status::unimplemented` and reject the `brgemm_matmul_copy_reorder_t` implementation.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
@@ -80,10 +80,12 @@ status_t calculate_plain_transpose_blocks(dim_t &batch, dim_t &M, dim_t &K,
 
     memory_desc_t src_md_reduced, dst_md_reduced;
     VDISPATCH_REORDER_IC(memory_desc_reshape(src_md_reduced, src_md,
-                                 non_unit_dim, non_unit_dims),
+                                 non_unit_dim, non_unit_dims)
+                    == status::success,
             VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "src");
     VDISPATCH_REORDER_IC(memory_desc_reshape(dst_md_reduced, dst_md,
-                                 non_unit_dim, non_unit_dims),
+                                 non_unit_dim, non_unit_dims)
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
```diff
@@ -80,10 +80,12 @@ status_t calculate_plain_transpose_blocks(dim_t &batch, dim_t &M, dim_t &K,
 
     memory_desc_t src_md_reduced, dst_md_reduced;
     VDISPATCH_REORDER_IC(memory_desc_reshape(src_md_reduced, src_md,
-                                 non_unit_dim, non_unit_dims),
+                                 non_unit_dim, non_unit_dims)
```
avmanerikar, line 83: Suggestion to use [`VDISPATCH_REORDER_SC`](https://github.com/uxlfoundation/oneDNN/blob/15476365497a5dd4944e3eeaf4fb2a2662aa3902/src/common/reorder_pd.hpp#L41) that checks for the status directly.

avmanerikar (approved):

src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
asztukow: Thanks, unfortunately it’s not straightforward here: `VDISPATCH_REORDER_SC` expands to `this->info(engine)` and requires a primitive‑PD member context (`this`/`engine`), while `calculate_plain_transpose_blocks` is a free function with no such context, so it can’t be used directly.

asimonov1 (approved):

src/cpu/x64/matmul/brgemm_matmul_reorders.cpp
avmanerikar: Makes sense - resolving.

PR #5021, opened 2026-04-15
cpu: aarch64: shuffle: make SVE instantiation vector-length agnostic

Sqvid:
# Description
Makes jit shuffle vector-length agnostic for SVE.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
+++ b/src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2022-2024 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -18,11 +18,12 @@
 
 #include <cassert>
 
-#include "common/bfloat16.hpp"
 #include "common/c_types_map.hpp"
 
 #include "cpu/aarch64/jit_generator.hpp"
 #include "cpu/aarch64/shuffle/jit_uni_shuffle_kernel.hpp"
+
+#include "xbyak_aarch64/xbyak_aarch64/xbyak_aarch64_util.h"
 
 namespace dnnl {
 namespace impl {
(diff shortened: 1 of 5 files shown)


src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.hpp
```diff
@@ -98,12 +94,11 @@ struct jit_uni_shuffle_kernel_t : public jit_generator_t {
...
     const XReg &reg_tmp5_ = x12;
     const XReg &reg_tmp6_ = x13;
-    const XReg &reg_padded_block = x14; //WReg(x14.getIdx());
+    const XReg &reg_padded_block = x14;
 
     const jit_shuffle_conf_t conf_;
     const size_t padding_size_;
-    const uint64_t isa_sveLen
-            = is_superset(isa, sve_128) ? cpu_isa_traits<isa>::vlen : 0;
+    const uint64_t sve_length_ = get_sve_length();
```
jondea: I think this should be `simd_bytes(isa)` or similar. Otherwise you could make this crash by running asimd on a machine with SVE length > 128

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
```diff
@@ -310,7 +311,7 @@ void jit_uni_shuffle_kernel_t<isa>::append_zero_padding(
         for (; off + simd_w_byte < padding_to_add; off += simd_w_byte) {
             add_imm(X_DEFAULT_ADDR, reg_dst_addr, off_start + off, X_TMP_0);
 
-            if (is_superset(isa, sve_128))
+            if (isa == sve)
```
jondea: I think 
```
is_superset(isa, sve)
```
would be more idiomatic here, so that it would work for vector length specific cases too. Although it does beg the question of why we can't just use `uni_str` in both cases...

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
```diff
@@ -337,9 +338,9 @@ void jit_uni_shuffle_kernel_t<isa>::generate() {
     preamble();
 
     /* Overwrite P_ALL_ONE, if required sve size < 512. */
-    if (isa_sveLen == util::SVE_128)
+    if (sve_length_ == util::SVE_128)
```
jondea, line 341: Will this not SIGILL on a C6g. This is very odd...

jondea (changes requested):
Thanks for taking a stab at simplifying things, although the existing code quality of this one is making me nervous...

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.hpp
Sqvid: The original code relies on the fact that `isa_sveLen` would be zero on `asimd` which is why I thought this was the simplest way to achieve the goal of this PR.

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
Sqvid: `sve_length_` is zero on `asimd`

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.hpp
Sqvid: Oh I see what you're saying. You mean if we force it via `benchdnn` for example. That's a good point, thanks!

dzarukin (approved):

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.hpp
Sqvid: Done.

src/cpu/aarch64/shuffle/jit_uni_shuffle_kernel.cpp
Sqvid: Done.

jondea (approved):
After in person chat I'm happy that the code quality issues with this implementation are orthogonal to this change, and it is a step in the right direction at least

PR #5023, opened 2026-04-15
cpu: aarch64: make softmax SVE instantiation vector length agnostic

Sqvid:
# Description

Makes jit softmax vector-length agnostic for SVE.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/jit_uni_softmax.cpp
+++ b/src/cpu/aarch64/jit_uni_softmax.cpp
@@ -16,23 +16,21 @@
 * limitations under the License.
 *******************************************************************************/
 
-#include <assert.h>
+#include <cassert>
 #include <memory>
 
 #include "common/c_types_map.hpp"
 #include "common/dnnl_thread.hpp"
-#include "common/math_utils.hpp"
 #include "common/memory_tracking.hpp"
 #include "common/nstl.hpp"
-#include "common/type_helpers.hpp"
 #include "common/utils.hpp"
 
+#include "cpu/aarch64/injectors/jit_uni_eltwise_injector.hpp"
+#include "cpu/aarch64/jit_generator.hpp"
+#include "cpu/aarch64/jit_uni_softmax.hpp"
 #include "cpu/cpu_primitive.hpp"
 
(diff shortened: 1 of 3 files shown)


src/cpu/aarch64/jit_uni_softmax.cpp
```diff
@@ -167,7 +165,7 @@ size_t jit_softmax_base_t::compute_axis_stride(
 
 void jit_softmax_base_t::load_common_params() {
     mov(W_TMP_0, float2int(-FLT_MAX));
-    if (vlen_ > cpu_isa_traits<sve_128>::vlen) {
+    if (vlen_ > util::SVE_128) {
```
jondea, line 168: Not sure we need to use this `util` thing. This is actuallya purely numerical question, and throwing the concept of architecture in is unnecessary

src/cpu/aarch64/jit_uni_softmax.hpp
```diff
@@ -169,8 +167,6 @@ struct jit_uni_softmax_bwd_t : public primitive_t {
                                            diff_dst_md()->data_type,
                                            diff_src_md()->data_type),
                             mayiuse_bf16())
-                    && (mayiuse(sve_512) || mayiuse(sve_256)
```
jondea, line 172: If you remove this, it will enable it for asimd. Is this intentional?

jondea (commented):
Two minor comments, generally looks fine

dzarukin (approved):

src/cpu/aarch64/jit_uni_softmax.cpp
Sqvid: Hmm I just think it's more readable and grepable than a raw number like `16`.

src/cpu/aarch64/jit_uni_softmax.hpp
Sqvid: Will it? There is a check at the start of this chain for `mayiuse(isa)` and we only instantiate the `bwd` template for `<sve>`

jondea (approved):

PR #3027, opened 2025-04-04
cpu: x64: matmul: update set_or_check_B_tag

xuxinzen:
Fixes MFDNN-13446

PR #5038, opened 2026-04-16
cpu: x64: matmul: enable treat_as_plain for weights format

xuxinzen:
Fixes MFDNN-14900

Using plain format when N == 1 for transpose format, which works the same before guilty commit. And the TF confirms that the patch works fine on their end.

[CI tests](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f139abf2-8a32-f19c-b124-a4bf010d0e2d)

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -513,8 +513,10 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(memory_desc_t &B_md,
                     = memory_desc_matches_tag(B_md, plain_tensor_layout_tag)
                     && memory_desc_matches_tag(
                             B_md, transposed_tensor_layout_tag);
+            const bool treat_as_plain = plain_transposed_matched && bgmmc.N == 1
+                    && !bgmmc.is_int4_weights;
             if (bgmmc.wei_tag == format_tag::undef
-                    || plain_transposed_matched) {
+                    || (plain_transposed_matched && !treat_as_plain)) {
                 if (gemm_based::check_gemm_input_format(B_md)) {
                     // Note: Here we batch layout may not be accurately represented
                     // by the wei_tag string, due to all the permutations of the


src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -513,8 +513,10 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(memory_desc_t &B_md,
                     = memory_desc_matches_tag(B_md, plain_tensor_layout_tag)
                     && memory_desc_matches_tag(
                             B_md, transposed_tensor_layout_tag);
+            const bool treat_as_plain = plain_transposed_matched && bgmmc.N == 1
+                    && !bgmmc.is_int4_weights;
```
dzarukin, line 517: I anticipate to see a comment explain the necessity of this variable and its conditions. It took several people for several days to acknowledge the issue and narrow down it to matmul specific setting behavior.

Additionally, I wonder why no recently introduced canonical call is not used.

Thirdly, it's unclear why parallel creation lead to the crash, which shouldn't happen by primitive cache design, insights are desired.

dzarukin (changes requested):

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
dzarukin: And please convert the reproducer from the tracker into a test_concurrency gtest expansion as a regression test.

densamoilov: I'm also wondering whether the issue has been there for long time or did we recently introduce it?

xuxinzen: I'll add some comments and this patch is a temporary fixup. I need more time to investigate the cause and get a real solution

As for why not using is_canonical(), I tried to use it before and the issue was resolve with c++ reproducer. But the results from TF team said the segfault issue was still there.

xuxinzen: I can reproduce this issue with rls-v3.7 when using correct weights format for transposed. brgemm matmul uses plain for transpose format when N == 1 before the guilty commit.

vpirogov: > I'm also wondering whether the issue has been there for long time or did we recently introduce it?

It was introduced a year ago by PR #3027. Specifically [this stride check](https://github.com/uxlfoundation/oneDNN/pull/3027/changes#diff-27337a02b11079d84fa9408f56adeb19a9fee3cbe8b6287eabefb63ca5b34156R362-R364) creates undefined behavior when corresponding dimension is trivial.

xuxinzen: If the `stride_check` is case, then the reason of not using is_canonial() that mentioned above can be explained as I still keep the code to reach the stride check. I'll continue investigating that. Thanks

densamoilov: I don’t fully understand this fix since we don’t know the root cause yet. You said you needed more time to find it and come up with a real solution so it’s not clear whether this is actually fixing the issue or just happens to make it disappear. I don’t think we should promote it as a fix until we understand that

tprimak:
https://github.com/uxlfoundation/oneDNN/pull/5158

vpirogov:
> #5158

I see that the fail does not reproduce after changes in #5158, but it's just a coincidence. The buggy code is still there on main:

https://github.com/uxlfoundation/oneDNN/blob/main/src/cpu/x64/matmul/brgemm_matmul_utils.cpp#L590-L613

At a minimum the patch that landed to rls-v3.12 should be merged to main. Ideally we need to understand why plain copy-B does not work for odd size (i.e. properly fix MFDNN-13446).

densamoilov:
@vpirogov, https://github.com/uxlfoundation/oneDNN/pull/5171 to port the fix to main.

PR #5042, opened 2026-04-16
x64: conv: relax large size check for 3d f32/int8 shapes

tczeszun:
Fixes MFDNN-14642. Large 3D convs (nnUNet) were falling to `ref:any` because `has_large_size()` rejected shapes with per-image elements > `INT_MAX`, even when no `jcp` field actually overflowed. 

- Added `has_large_size_relaxed()` in `jit_primitive_conf.hpp` that checks only per-tensor spatial product instead of `channels * spatial` and used it in the brgemm conv path. 
- Also added a `size_t` cast on `jcp.LDA * jcp.os` in the brgemm rtus buffer size calculation, which can now exceed `INT_MAX` under the relaxed guard. 

Repro shape now picks `brg_conv_fwd:avx512_core` (f32) / `brg_conv_fwd:avx10_1_512_amx` (int8) instead of `ref:any`.

Potential TODO: May need to extend it to other kernels as well.

### DIFF
--- a/src/cpu/x64/jit_brgemm_conv_utils.cpp
+++ b/src/cpu/x64/jit_brgemm_conv_utils.cpp
@@ -1679,9 +1679,10 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
     const memory_desc_wrapper dst_d(&dst_md);
     const memory_desc_wrapper bias_d(&bias_md);
 
-    // Big int (> INT_MAX) values are unsupported and jcp fields may overflow
+    // Per-tensor spatial product, weights nelems, and per-dim
+    // strides/paddings/dilations must fit in int.
     // TODO: change data type of jcp fields to size_t
-    VDISPATCH_CONV_IC(!has_large_size(cd, src_d, weights_d, dst_d),
+    VDISPATCH_CONV_IC(!has_large_size_relaxed(cd, src_d, weights_d, dst_d),
             VERBOSE_BAD_PARAM, "large size is not supported");
 
     const bool with_groups = weights_d.ndims() == src_d.ndims() + 1;
@@ -2658,7 +2659,7 @@ status_t init_1x1_conf(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
 
     const size_t rtus_buffer_size = jcp.is_reduced_rtus
             ? jcp.rtus_padded_ic_size * jcp.os_block
-            : jcp.LDA * jcp.os;
+            : static_cast<size_t>(jcp.LDA) * jcp.os;
     jcp.inp_buffer_size
             = jcp.is_rtus ? rnd_up(rtus_buffer_size, align_size) : 0;
 
--- a/src/cpu/x64/jit_primitive_conf.hpp
+++ b/src/cpu/x64/jit_primitive_conf.hpp
@@ -305,6 +305,36 @@ inline bool has_large_size(const convolution_desc_t &cd,
     return false;
 }
 
+// Relaxed version of has_large_size(). The only difference from
+// has_large_size() is that this variant checks the per-tensor spatial product
+// instead of the count of all consecutive elements excluding the minibatch
+// (i.e. spatial * channels). Weights nelems and per-dim strides/paddings/dilations
+// are checked the same way. Use this when element offsets into memory are computed
+// as dim_t and only per-dim / spatial values need to fit in int.
+inline bool has_large_size_relaxed(const convolution_desc_t &cd,
+        const memory_desc_wrapper &src_d, const memory_desc_wrapper &weights_d,
+        const memory_desc_wrapper &dst_d) {
+    auto is_large = [](const dim_t val) { return val > INT_MAX; };
+
+    const int ndims = src_d.ndims();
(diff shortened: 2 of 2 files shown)


src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -1681,7 +1681,7 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
 
     // Big int (> INT_MAX) values are unsupported and jcp fields may overflow
```
densamoilov: (nit) should we adjust this comment to mention that it's a per-tensor requirement to the spatial?

src/cpu/x64/jit_brgemm_conv_utils.cpp
```diff
@@ -1681,7 +1681,7 @@ status_t init_jcp(jit_brgemm_conv_conf_t &jcp, cpu_isa_t isa,
...
     // TODO: change data type of jcp fields to size_t
-    VDISPATCH_CONV_IC(!has_large_size(cd, src_d, weights_d, dst_d),
+    VDISPATCH_CONV_IC(!has_large_size_relaxed(cd, src_d, weights_d, dst_d),
```
densamoilov, line 1685: What criteria are we using to decide that checking only spatial per-tensor is enough here?

tczeszun: No `jcp` field stores `channels * spatial` as a single `int`. Channels (`jcp.ic`, `jcp.oc`) and spatial (`jcp.os`, `jcp.is`) are separate `int` fields, so we only need each to fit in `int` independently. The original `has_large_size()` checks `nelems / mb` (i.e. `channels * spatial`), which rejects shapes like `ic=64, id*ih*iw=320*320*576` even though `ic` and `id*ih*iw` each fit in `int` on their own.

src/cpu/x64/jit_brgemm_conv_utils.cpp
tczeszun: done

src/cpu/x64/jit_brgemm_conv_utils.cpp
asimonov1: But what about intermediate results in expressions like `jcp.mb * jcp.ngroups * jcp.ic * jcp.oc` or similar? The 1st multiplier is casted to dim_t or size_t in some places, but not always, and it is not obvious sometimes if overflow is not possible.

tczeszun: @asimonov1 This is a pre-existing issue - `has_large_size()` doesn't guard against `ic * oc` overflow either (e.g. ic=50000, oc=50000 with 1x1 spatial passes the old check but overflows `mb * ngroups * ic * oc`). A full audit of `int` overflow in multi-field expressions would require a separate effort. Though, I believe our sanitizer CI testing (UBsan) catches these kinds of errors.

densamoilov: I’m not sure I fully understand the intent of this function. Is its purpose mainly to ensure that the int fields in `jit_conv_conf_t` do not overflow? If so, would it make more sense to address this at the type level instead, for example by changing `int` to `dim_t` (for dimensions/strides) and `size_t` (for sizes in bytes)? That approach could eliminate the need for this new function and the check.

Another point I’m wondering about is whether the core issue here is actually overflow within `jit_conv_conf_t` or rather the use of memory offsets that exceed `INT_MAX`, which can trigger Xbyak exceptions. In that context could you clarify whether the brgemm-based convolution path handles such large offsets correctly (e.g. via offset compression)?

tczeszun: @densamoilov Changing `int` to `dim_t` is the right long-term fix but a large refactor. This is a minimal targeted fix.

The brgemm path already handles large memory offsets correctly (`dim_t` pointer math, `safe_add()` in JIT). The issue is purely `int` overflow in `jcp` config fields like `jcp.os`.

densamoilov: Okay, I might be missing something here. Based on my understanding we should be able to simply replace int with dim_t in the structure and be done with it. If that’s all that’s required that approach seems much better than introducing a new function, especially since we’d then need to explain the difference between the relaxed and non‑relaxed versions and why the relaxed version is only used in this particular implementation.

tczeszun: @densamoilov It's not just a struct change - `jcp` fields are used in JIT codegen (registers/immediates), in `brgemm_desc_t` (LDA/LDB are `int`), and across all conv kernels (avx2, avx512, amx, brgemm). Changing to `dim_t` requires updating every JIT usage + fixing hundreds of narrowing warnings.

tczeszun: FYI, #2827 introduced this function

densamoilov: I see. Thank you for the explanation.

densamoilov (approved):

asimonov1 (approved):

PR #5043, opened 2026-04-17
[rls-v3.11] cpu: x64: matmul: fixed undefined behavior for trivial dimensions case

vpirogov:
This PR removes explicit stride check in `brgemm_matmul_conf_utils_t::set_or_check_B_tag` method. This check resulted in undefined behavior for memory descriptors with a trivial dimension (e.g. `N=1`). In particular, a tensor may be interpreted as plain or transposed based on a stride value for trivial dimension despite having the same physical layout. This undefined behavior resulted in primitive cache hash collision and eventual segfault due to different requirements to scratchpad size between implementations for plain and transposed cases.

The fix is confirmed by a Tensorflow test.

Orphan CI: https://intel-ci.intel.com/f13a7e9e-6ba1-f1c3-8889-a4bf010d0e2d

Fixes MFDNN-14900

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -501,25 +501,14 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(memory_desc_t &B_md,
                     : memory_desc_matches_one_of_tag(B_md,
                               plain_tensor_layout_tag,
                               transposed_tensor_layout_tag, acbd, adbc);
-            const bool plain_transposed_matched
-                    = memory_desc_matches_tag(B_md, plain_tensor_layout_tag)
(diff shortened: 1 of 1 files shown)


vpirogov:
This one actually breaks int4 WoQ cases. Here's an example:
```
./benchdnn --matmul --dt=bf16:u4:bf16 --stag=acb --wtag=acb --dtag=abc --attr-fpmath=bf16:true --skip-impl=ref 8x599x1388:8x1388x1
```

vpirogov:
Figured out that the logic around `is_plain_transposed` was introduced as a workaround for lack of odd sizes support for int4 in plain copy-B routines. Updated the patch to reflect that.

Now this PR is effectively a new version of #3027.

densamoilov (approved):

tprimak (approved):

PR #5044, opened 2026-04-17
git: ignore .gdbinit and .gdb_history

Sqvid:
# Description

Ignore the local `gdb` config and `gdb` history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 #===============================================================================
 # Copyright 2019-2021 Intel Corporation
-# Copyright 2024-2025 Arm Limited and affiliates.
+# Copyright 2024-2026 Arm Limited and affiliates.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -31,3 +31,5 @@ compile_commands.json
 **/.DS_Store
 __pycache__
 .cache
+.gdb_history
+.gdbinit


.gitignore
```diff
@@ -1,6 +1,6 @@
 #===============================================================================
 # Copyright 2019-2021 Intel Corporation
-# Copyright 2024-2025 Arm Limited and affiliates.
+# Copyright 2024-2026 Arm Limited and affiliates.
```
rjoursler, line 3: It looks like this file didn't get update to the new copyright format. @vpirogov In general, does really even make sense to maintain a License  header on .gitignore?

rjoursler (approved):

.gitignore
vpirogov: In general no. It's mainly there to avoid making various checkers sad.

.gitignore
```diff
@@ -31,3 +31,5 @@ compile_commands.json
 **/.DS_Store
 __pycache__
 .cache
+.gdb_history
+.gdbinit
```
dzarukin, line 35: I always thought these files belong to user home directory (which is usually not DNNLROOT directory).
Is it a common case for your team development environment?
There's also a way to exclude specific files from tracking locally (through .git/info/exclude), would it work for you?

.gitignore
vpirogov: I'm fine with dropping license header from this file unless @Sqvid has reasons to keep it.

.gitignore
Sqvid: I do indeed have `~/.config/gdbinit` for generally-applicable settings. But it is useful to have a project-local config for oneDNN-specific `gdb` utilities (typically for things like JIT debugging). The same applies to the history file ([which by default is created in the current directory](https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_182.html)).

.gitignore
Sqvid: We typically don't remove existing headers so I'd prefer to avoid that headache in this PR if possible.

dzarukin (approved):

PR #5017, opened 2026-04-14
[GPU] GEMM Acc fixup

kealan-barbieri:
# Description

- Modify primitive init to always use int acc when both inputs are ints
- Add strategy coverage for transpose int inputs. 
- Move runtime dim check to avoid assertions in `ld` initialization. 

Fixes [MFDNN-14871](https://jira.devtools.intel.com/browse/MFDNN-14871)

# Checklist

## General

- [x] Do all unit and benchdnn tests (`make test` and `make test_benchdnn_*`) pass locally for each commit?
- [x] Have you formatted the code using clang-format?

### Bug fixes

- [x] Have you included information on how to reproduce the issue (either in a github issue or in this PR)?
- [x] Have you added relevant regression tests?

PR #5057, opened 2026-04-20
[GPU][NVL-P] Add acc mode restriction on atomic acc

kealan-barbieri:
# Description

Since NVL-P supports f16/bf16 atomic acc we need to restrict atomic accumulation to avoid precision issues.

Fixes # [MFDNN-14692](https://jira.devtools.intel.com/browse/MFDNN-14692)

# Checklist

## General

### Bug fixes

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/strategy.cpp
+++ b/src/gpu/intel/gemm/jit/generator/strategy.cpp
@@ -86,6 +86,7 @@ bool useAutoAtomic(HW hw, const GEMMProblem &problem, const GEMMStrategy &strate
     return (hw >= HW::XeHPG)
             && (ignoreBeta || problem.beta1())
             && hasNativeAtomicAdd(hw, problem.Tc_ext.real(), problem.C, strategy.C)
+            && (problem.Tacc_min == Type::invalid || problem.Tacc_min.isSubsetOf(problem.Tc_ext))
             && !strategy.cLoadAhead
             && (problem.postOps.len() == 0 || problem.hasSum1PostOpAtEnd())
             && (problem.cOffset != COffset::Post)
--- a/src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
+++ b/src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
@@ -169,6 +169,7 @@ struct GEMMProblem : public CommonProblem {
     Type Tao, Tbo, Tco;                             // Types for A/B/C offsets.
     Type Ta_scale, Tb_scale, Tc_scale;              // Types for A/B/C scales.
     Type Tag, Tbg;                                  // Types for A/B group sums.
+    Type Tacc_min;                                  // Type for minimum C accumulation precision.
 
     Scalar alpha, beta;                             // Scaling factors for A*B and C, respectively.
     MatrixAddressing A, B, C;                       // Addressing information for A/B/C matrices.
@@ -301,6 +302,7 @@ struct GEMMProblem : public CommonProblem {
     bool usesCOPtr() const { return (hasCOffsetPtr() && cOffset != COffset::None) || sumA || sumB; }
     bool allowMatrixOffset() const { return (cOffset == COffset::Pre); }
 
+    // Data type of dst registers for dpas/mad/dp4a.
     Type Tc_compute() const {
         if (Ta.isInteger() && Tb.isInteger() && Tc == Type::f32)
             return Type::s32;
--- a/src/gpu/intel/gemm/jit/pd.cpp
+++ b/src/gpu/intel/gemm/jit/pd.cpp
@@ -630,16 +630,28 @@ status_t pd_t::init_GEMMProblem(
             || (post_ops_.find(primitive_kind::prelu) != -1);
 
     bool need_x32_acc = with_binary || !IMPLICATION(with_sum_, sum_at_begin_);
+    auto acc_mode = attr()->acc_mode_;
 
-    switch (attr()->acc_mode_) {
+    // Strict acc mode default.
+    auto Tacc_mode = problem.Tc_ext.isFP() ? data_type::f32 : data_type::s32;
+
+    // Initialize non-strict acc mode.
+    switch (acc_mode) {
         case accumulation_mode::any:
-            if (!need_x32_acc) acc_type = data_type::undef;
+            if (!need_x32_acc) Tacc_mode = data_type::undef;
             break;
-        case accumulation_mode::f16: acc_type = data_type::f16; break;
-        case accumulation_mode::f32: acc_type = data_type::f32; break;
-        case accumulation_mode::s32: acc_type = data_type::s32; break;
+        case accumulation_mode::f16: Tacc_mode = data_type::f16; break;
+        case accumulation_mode::f32: Tacc_mode = data_type::f32; break;
+        case accumulation_mode::s32: Tacc_mode = data_type::s32; break;
         default: break;
     }
+
+    // Minimum precision type for applying post-ops based on acc mode.
+    // Limits use of atomic add.
+    problem.Tacc_min = convert_dnnl_to_kernel_type(Tacc_mode);
+
+    if (acc_mode != accumulation_mode::strict) acc_type = Tacc_mode;
+
     if (wei_decomp_) { acc_type = data_type::f32; }
 
     auto trans_co = trans_bias();


dyoussif (approved):

src/gpu/intel/gemm/jit/generator/strategy.cpp
```diff
@@ -86,6 +86,7 @@ bool useAutoAtomic(HW hw, const GEMMProblem &problem, const GEMMStrategy &strate
     return (hw >= HW::XeHPG)
             && (ignoreBeta || problem.beta1())
             && hasNativeAtomicAdd(hw, problem.Tc_ext.real(), problem.C, strategy.C)
+            && (problem.Tacc.ngen() == DataType::invalid || problem.Tacc.isSubsetOf(problem.Tc_ext))
```
echeresh, line 89: I'm looking at `Type::ngen()` implementation. Invalid `Type` has `val = 0` so `Type::ngen()` returns `ngen::DataType::hf`:

https://github.com/uxlfoundation/oneDNN/blob/f2a4710ba4625ba11277880286dcd0310bd857fd/src/gpu/intel/gemm/jit/include/gemmstone/type.hpp#L109-L122

This is not directly related to the PR but the check works by accident - better to fix to make it behaving as intended.

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
```diff
@@ -169,6 +169,7 @@ struct GEMMProblem : public CommonProblem {
     Type Tao, Tbo, Tco;                             // Types for A/B/C offsets.
     Type Ta_scale, Tb_scale, Tc_scale;              // Types for A/B/C scales.
     Type Tag, Tbg;                                  // Types for A/B group sums.
+    Type Tacc;                                      // Type for minimum C Accumulation Precision.
```
echeresh, line 172: Nit: follow capitalization rules consistently:

```suggestion
    Type Tacc;                                      // Type for minimum C accumulation precision.
```

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -594,6 +586,14 @@ status_t pd_t::init_GEMMProblem(
...
     }
+
+    // Minimum precision type for applying post-ops based on acc mode.
+    // Limits use of atomic add.
+    auto strict_acc = problem.Tc_ext.isFP() ? data_type::f32 : data_type::s32;
+    problem.Tacc = (acc_mode == accumulation_mode::strict)
+            ? convert_dnnl_to_kernel_type(strict_acc)
+            : convert_dnnl_to_kernel_type(acc_type);
+
     if (wei_decomp_) { acc_type = data_type::f32; }
```
echeresh, line 597: `acc_type` is used for `Tacc` initialization but updated afterwards when `wei_decomp_` is `true`. Is this intentional? This doesn't look right.

echeresh (approved):

src/gpu/intel/gemm/jit/pd.cpp
kealan-barbieri: It is intentional, but maybe not ideal for clarity. There are other attributes that impact the `acc_type` vs the `Tacc` which is a standin for `acc_mode`, so the later modifications are still necessary, but I'm catching `acc_type` when it has been potentially set based on `acc_mode` but before its potentially overriden. 

I'll refactor this so `acc_mode` is used to initialize something separate for clarity.

src/gpu/intel/gemm/jit/generator/strategy.cpp
kealan-barbieri: thanks for catching, droped the `ngen()` call and compared `Tacc` directly to avoid this.

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
kealan-barbieri: fixed, thanks.

src/gpu/intel/gemm/jit/pd.cpp
kealan-barbieri: Refactored to track `acc_mode` separately, apply separately to `Tacc`, `acc_mode`.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph
enable benchdnn_lrn
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-lpg

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
```diff
@@ -169,6 +169,7 @@ struct GEMMProblem : public CommonProblem {
     Type Tao, Tbo, Tco;                             // Types for A/B/C offsets.
     Type Ta_scale, Tb_scale, Tc_scale;              // Types for A/B/C scales.
     Type Tag, Tbg;                                  // Types for A/B group sums.
+    Type Tacc;                                      // Type for minimum C accumulation precision.
```
Simonsays095: Isn't `Tacc === Tc`? I think we should be able to just use that directly.

src/gpu/intel/gemm/jit/pd.cpp
```diff
@@ -584,16 +575,28 @@ status_t pd_t::init_GEMMProblem(
             || (post_ops_.find(primitive_kind::prelu) != -1);
 
     bool need_x32_acc = with_binary || !IMPLICATION(with_sum_, sum_at_begin_);
+    auto acc_mode = attr()->acc_mode_;
+
+    // Strict acc mode default.
+    auto Tacc_mode = problem.Tc_ext.isFP() ? data_type::f32 : data_type::s32;
```
Simonsays095, line 636: I think this logic should match `acc_type`, at least with f64 cases.

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
kealan-barbieri: Unfortunately this isnt true, eg in cases like the target of this fix with type signature `s8:s8:f16` , `Tc` will be `s32` so we need a separate value to track the required accumulation precision separately from the actual FMA accumulation type (`Tc`) and the actual destination type (`Tc_ext`).

src/gpu/intel/gemm/jit/pd.cpp
kealan-barbieri: this is a separate value from `acc_type` this line is meant to set the default `strict` acc_mode type, separate from the acc type that will be used for FMA internal to the kernel.

Simonsays095: Right, my point is that f64 cases should accumulate with f64 when using `strict` accumulation, right?

kealan-barbieri: I see, my understanding is we only enforce accumulation precision up to f32 as in its not guaranteed to use f64 accumulation even in a case with f64 dst. https://uxlfoundation.github.io/oneDNN/enum_dnnl_accumulation_mode.html#index-0 the docs reflect this in that strict acc is either f32 or s32. In this case the `Tacc` is only enforced as a minimum precision so it shouldnt stop f64 acc from being used where its supported.

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
Simonsays095: Alright... Just to make sure I understand fully: `s8:s8:f16` will use an `s8:s8:s32` dpas/fma followed by downconversion to f16 at the end of the kernel. If the kernel uses atomic accumulation (kv/kb strategies), then it'll do partial s32 accumulation in-thread followed by partial f16 (atomic) accumulation between threads, right?

If that's all correct... We already support `Tc` being different than the compute type. Take `s8:s8:f32` kernels for example, in gemmstone type notation these can either use `OO[IS]` (implicit s32 compute, accumulate in s32, convert to f32 on write) or `OOIS` (s32 compute, accumulate in f32, implicit no conversion on write). These show up in the catalog as `OOS` and `OOI` kernels. I think these are the only cases that make use of this currently, but it is something we support. We also have `problem.Ts` which is related but I don't know the specifics.

Instead of adding `Tacc` to the problem, is there an issue with just using `&& !problem.Tc_ext.isSubsetOf(problem.Tc))` in `useAutoAtomic`?

kealan-barbieri: I think such cases are being eliminated by https://github.com/uxlfoundation/oneDNN/pull/5017 and should be avoided. To me it is clearly better to add a new parameter with a specific purpose than to rely on variable internal behavior that makes the purpose of the `Tc` parameter less clear.

In the affected case (`--ip --engine=gpu --allow-enum-tags-only=false --dir=FWD_I --dt=s8:s8:f16 --bia-dt=f32 --stag=abcd --dtag=ab --attr-post-ops=mul:f32:2+sum mb784ic768ih1oc192n"c78e401edef547035e7bc7e2631b3e67*5"`) the issue is the application of sum using atomic add, so its fully separated from the "compute" type. 

Ideally `Tc` should refer to the accumulation type during fma (dpas/mad/dp4a) and should always be s32 for cases where both inputs are ints. Meanwhile `Tc_ext` is the "real" dst type, potentially a float in cases with int `Tc` and `Tacc` is the user parameter specifying the minimum precision for accumulation and application of post-ops. In practice we dont have to check that `Tc` meets `Tacc` threshold, but if we add an fp64 accumulation mode in the future we will have to potentially apply it there.

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph
enable benchdnn_lrn
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-lpg

kealan-barbieri:
make test perf-gpu
set primitive=gpu:gemm

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph
enable benchdnn_lrn
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-lpg

kealan-barbieri:
make test perf-gpu
set primitive=gpu:gemm

kealan-barbieri:
make test
set test_scope=NIGHTLY
disable test_device_cpu
enable test_device_gpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn
enable benchdnn_graph
enable benchdnn_lrn
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg
enable arch_gpu_xe3p-lpg

kealan-barbieri:
make test perf-gpu
set primitive=gpu:gemm

PR #4839, opened 2026-03-17
graph: backend: dnnl: executables: call internal sdpa iface

TaoLv:
- Call internal sdpa iface in graph backend.
- It simplifies the integration code a bit and importantly, it enables sdpa verbose log under graph API.
- Add sdpa_test_iface.hpp in common so it can be included in both graph backend and internal gtests.

See sdpa verbose below after the change:

```
$ ONEDNN_VERBOSE=1 ./tests/benchdnn/benchdnn --graph --engine=gpu --case=complex_fusion/mha/sdpa-plain-simplified-f16-f32.json
onednn_verbose,v1,info,oneDNN v3.12.0 (commit d23923d92bcc93968b195be2cb8f1d22d3b7e835)
onednn_verbose,v1,info,cpu,runtime:OpenMP,nthr:224
onednn_verbose,v1,info,cpu,isa:Intel AVX-512 with float16, Intel DL Boost and bfloat16 support and Intel AMX with bfloat16 and 8-bit integer support
onednn_verbose,v1,info,gpu,runtime:OpenCL
onednn_verbose,v1,info,gpu,engine,opencl device count:4 
onednn_verbose,v1,info,gpu,engine,0,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,gpu,engine,1,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,gpu,engine,2,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,gpu,engine,3,name:Intel(R) Data Center GPU Max 1100,driver_version:25.18.33578,binary_kernels:enabled
onednn_verbose,v1,info,graph,backend,0:dnnl_backend
onednn_verbose,v1,primitive,info,template:operation,engine,primitive,implementation,prop_kind,memory_descriptors,attributes,auxiliary,problem_desc,exec_time
onednn_verbose,v1,graph,info,template:operation,engine,partition_id,partition_kind,op_names,data_formats,logical_tensors,fpmath_mode,implementation,backend,exec_time
onednn_verbose,v1,primitive,exec,cpu,reorder,jit_direct_copy:uni,undef,src:f32::blocked:abdc::f0 dst:f32::blocked:abdc::f0,,,1x16x64x384,7.03687
[...]

PR #5060, opened 2026-04-21
graph: backend: dnnl: exec: gated_mlp uses unique_ptr instead of raw ptr

TaoLv:
1. Similar to what was discussed at https://github.com/uxlfoundation/oneDNN/pull/4839#discussion_r3029309642, here we also apply the `unique_ptr` changes to gated_mlp code.
2. Add `is_initialized()` function to all executables as we have more and more subclasses need to check the initialization status. It simplifies the code logic in `compiles_ops` by avoiding per op kind checks.

### DIFF
--- a/src/graph/backend/dnnl/executables/batch_norm.hpp
+++ b/src/graph/backend/dnnl/executables/batch_norm.hpp
@@ -80,6 +80,15 @@ struct bn_folding_t : public op_executable_t {
             const std::vector<cl_event> &deps) const override;
 #endif
 
+    bool is_initialized() const override {
+        return add_prim_ && mul_prim_ && sub_prim_
+#if DNNL_GPU_RUNTIME != DNNL_RUNTIME_NONE \
+        && DNNL_GPU_VENDOR == DNNL_VENDOR_NVIDIA
+                && sqrt_prim_
+#endif
+                ;
+    }
+
 private:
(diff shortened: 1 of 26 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/executables/batch_norm.hpp
```diff
@@ -80,6 +80,15 @@ struct bn_folding_t : public op_executable_t {
             const std::vector<cl_event> &deps) const override;
 #endif
 
+    bool is_initialized() const override {
+        return add_prim_ && mul_prim_ && sub_prim_
```
ElaineBao, line 84:
```suggestion
        return bool(add_prim_) && bool(mul_prim_) && bool(sub_prim_)
```

ElaineBao (approved):

rongzha1 (approved):

src/graph/backend/dnnl/executables/batch_norm.hpp
TaoLv: Thanks. They are not required as `&&` operator will convert them to `bool` implicitly.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #5068, opened 2026-04-22
xe: gemm: fix stride interface

rjoursler:
Avoids incorrect offset calculations for batched GEMM. Fixes the following issue reported in [MFDNN-14479](https://jira.devtools.intel.com/browse/MFDNN-14479)

```
$ ./tests/benchdnn/benchdnn --matmul --engine=gpu --dt=f32:f32:f32 --stag=acbd --wtag=adbc --dtag=abcd --attr-post-ops=binary_mul:f32:0 --attr-scratchpad=user 4x3x16413x16:4x3x16x16413
Segmentation fault from GPU at 0xff0000000000f000, ctx_id: 1 (CCS) type: 0 (NotPresent), level: 1 (PDE), access: 1 (Write), banned: 1, aborting.
Segmentation fault from GPU at 0xff0000000000f000, ctx_id: 1 (CCS) type: 0 (NotPresent), level: 1 (PDE), access: 1 (Write), banned: 1, aborting.
Abort was called at 306 line in file:
./shared/source/os_interface/linux/drm_neo.cpp
Aborted (core dumped) 

```

### DIFF
--- a/src/gpu/intel/gemm/jit.cpp
+++ b/src/gpu/intel/gemm/jit.cpp
@@ -173,10 +173,10 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
 
     if (pd()->batch_dims() >= 1) {
         for (int i = pd()->batch_dims() - 1; i >= 0; i--) {
-            auto stride_a = int32_t(pd()->stride(DNNL_ARG_A, i));
-            auto stride_b = int32_t(pd()->stride(DNNL_ARG_B, i));
+            auto stride_a = int64_t(pd()->stride(DNNL_ARG_A, i));
+            auto stride_b = int64_t(pd()->stride(DNNL_ARG_B, i));
             if (swap_ab) std::swap(stride_a, stride_b);
-            auto stride_c = int32_t(pd()->desc()->stride_c(i));
+            auto stride_c = int64_t(pd()->desc()->stride_c(i));
             if (jit::enable_generator_dsl()) {
                 auto hw = ngen::getCore(
                         ((ngen::Product *)&utils::downcast<intel::engine_t *>(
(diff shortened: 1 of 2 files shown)


dyoussif (approved):

atkassen (approved):

src/gpu/intel/gemm/jit.cpp
echeresh, line 220: This is now 64-bit but the kernel interface uses 32-bit. I guess we don't have any type checks but theoretically that could lead to a similar overflow issue.

Simonsays095:
Could we do something similar to what we do with the opencl dispatcher, and use 64-bit strides/dimensions only if required by the problem?

src/gpu/intel/gemm/jit.cpp
h-sadia: Could this be true for stride_binary? https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/gemm/jit.cpp#L238

rjoursler:
> Could we do something similar to what we do with the opencl dispatcher, and use 64-bit strides/dimensions only if required by the problem?

We definitely can, but I only intend to do this if we encounter a performance regression. If the offset calculations is not important for performance (which is generally the case for GEMM), then there is no benefit to adding this control, it only creates an extra point of failure.

rjoursler:
make test_ov

echeresh (approved):

src/gpu/intel/gemm/jit.cpp
rjoursler: Good points, I will try updating these as well and check if there are any issues.

rjoursler:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

PR #5071, opened 2026-04-22
rfc: improve functional coverage processes

dzarukin:
[Rendered document](https://github.com/uxlfoundation/oneDNN/tree/dzarukin/rfcs/improve_functional_coverage/rfcs/20260421-improve-functional-coverage)

The document proposes how to move functional level quality and processes associated with it to the next level.

### DIFF
--- a/rfcs/20260421-improve-functional-coverage/README.md
+++ b/rfcs/20260421-improve-functional-coverage/README.md
@@ -1,0 +1,86 @@
+# Introduction
+The purpose of this RFC is to re-establish the content of oneDNN functional validation and its procedural update. The request comes from external parties that rely on oneDNN quality.
+The challenge is to provide sufficient coverage to avoid bugs escapes in a rapidly changing environment of new functionality added to the library.
+
+## Vision
+Validation exists to prove a change, or a patch set is safe to promote into the repo.
+(Though validation should cover both functional and performance aspects, this RFC will focus on the functional part of it).
+The more complete functional coverage tests provide the higher chances are the change is safe.
+This proof must come in reasonable time. If it takes a long time (e.g., a day) to verify any kind of change, such process gives less value for the development compared to one that responds to results much quicker.
+However, expanding functional coverage with the bigger number of test cases takes more testing time, thus, the resulting vision statement takes the following form:
+The balance between execution time and coverage must be kept wisely; the team must achieve the most possible coverage under specific time budget.
+
+# Goals
+Here are the ultimate goals the team wants to achieve:
+* Time resources spent per single validation instance, or testing job (under assumption of infinite hardware resources):
+    - CI: 2 hours.
+    - Nightly: 8 hours.
+* Functional coverage:
+    - Most useful API combinations to ensure customer satisfaction with day-to-day functionality used in their environment.
+      This coverage should be expressed in a generalized way, not tied to specific shapes or a fixed set of API parameters.
+    - Corner cases to ensure getting off the streamline path doesn't lead to undesired behavior.
+    - Hit optimized library implementations with 80+% rate as slow implementations (reference codes) don't bring value to customers.
+
+# Proposal
+To fit more combinations with less test cases is possible when a single test case has various flags with non-default values.
+Given the infinite space of possibilities of various combinations, the best shot is to create a tool that would generate distributed points in that combination space. Coarse distribution can cover more various, usually independent implementation-wise, features of the library versus the targeted specific feature validation through iteration over some of options (as it's done currently). A tool (over the manual update) is preferred as more reliable and faster way to extend the coverage points in shorter time.
+
+A smart and scalable test case generator that would provide a single (no iterations over any arguments) benchdnn test case as its output is envisioned as such a tool.
+Its smartness comes in two aspects:
+    - It's adapting to the library API combinations, returning only test cases that make sense from the library perspective, not a sporadic combination (a.k.a. black box testing).
+        E.g., quantization is applied to integer data types; test case masks follow dimensions used and reflect real-world applications; etc.
+    - Uses probability mechanism to balance the amount of most useful combinations versus corner cases (test most useful functionality more often).
+    - Control ranges of random values to reach the desired goal of 80% hit rate.
+Scalability comes with an architecture that relies on regex strings to express more supported patterns with less implementation logic.
+E.g., data types are represented as x4 for s4 or u4, xf16 for bf16 and f16, etc.; use cases are combined in some generic policies like MXFP8 quantization; not fixed values represented with '*', etc.
+
+Once such generator exists and covers a benchdnn driver, here is the proposed model of its usage on the example of CI set:
+- Determine the list of features for a given driver that would form the core functionality.
+    * Core functionality must satisfy the following requirements:
+        - It must be shared between CPU and GPU backends.
+        - It should have optimized support in the library. It doesn't necessarily mean the optimized scope must be identical between backends.
+    * Core functionality should occupy the most space of coverage for a driver.
+    * Test cases targeting core would reside in `harness_driver_core_ci` file.
+- Determine the list of exclusive features for a backend or limited support scope.
+    * Such features should occupy a small portion of the total number of all test cases.
+    * Test case targeting such features would reside in `harness_driver_featureName_ci`.
+        E.g., ‘featureName’ can be 'f64' which is exclusive to GPUs and exists in various primitives.
+- A set of harnesses form `test_driver_ci` file which will replace existing file and its scope.
+- Verify the coverage fits the time limits across all supported hardware and satisfies proposed optimized implementation rates.
+    - In case it doesn't, analyze rates and/or times, make decisions based on the assessment.
+    - Refine the coverage.
+- Rough estimation is 5,000 cases.
+
+Same process will apply to Nightly set. The difference between Nightly and CI is the number of cases covered, which in case of Nightly would be around 20,00 - 25,000 and allowed problems might exceed 50 MBs up to (500 MB).
+Larger shapes will be validated in Nightly as it has bigger time budget.
+
+The process of appending coverage to existing files will be manual and will consist of several steps:
+1. Update the generator with a new scope (extending existing feature capabilities / introducing new feature through benchdnn flags) to the desired percent-wise state of core functionality as if it becomes a part of the core. If it takes a dedicated file, there is no need to take this step.
+2. Once the desired state is achieved, generate test cases containing only the new scope and append these cases to a correspondent file manually.
+
+The renewal process of core files (or re-generation) may be triggered by several factors:
+* Periodically, somewhere in the middle of development process for a coming release.
+* Once the time budget is exhausted, leading to longer jobs times stop fitting the time goal.
+* Once the scope of functionality changes, e.g. proportions of cases were adjusted in a major way to comply with the trend industry path.
+
+This tool is expected to be posted in oneDNN repository for everyone to use it.
+Every developer will be responsible for extending the tool to cover functionality they are adding whether it's a new feature or a specific important set of settings.
+Additional value of the tool is when it's mastered, a developer will be able to tailor local testing coverage for features/bugs they are working on.
+
+The final state of input files includes a various harness_driver_xxx_ci files and three test_driver_{smoke/ci/nightly} files.
+Any kind of filtering problems (don't run these on that HW or run those on this HW) must be regulated through other approaches but not through “test” files.
+Same approach shouldn't be also used for better balancing.
+Multiple "test" files per driver lead to repetitive validation, higher pressure on Intel internal automation system and lack of transparency what's included and validated.
+Having a single file should address these problems.
+Balancing can be done by CMake, if or when needed, by splitting a list into chunks and keeping shorter test files and targets associated with them in a build folder.
+
+## Synthdnn
+Currently, there's already a script that can perform case generation, located at src/scripts/synthdnn.
+So far it supports various data types, fpmath-mode, tags and shapes.
+The problem is it lacks a probability architecture piece which makes it less controlled on the output it provides.
+It acts more like a black box testing without adjusting to actually supported scenarios which wouldn't help with the announced goals once more features are added to it.
+
+## Fuzzdnn
+The showcase work is done on the base of another internal generator, "fuzzdnn".
+Its latest output can be checked out [here](harness_matmul_core_ci). Notice the variability of knobs used per case, sizes and other settings.
+This is still work-in-progress, some complete test cases may not make total sense, and they will be adjusted once it becomes easier to analyze what the limitations are.
--- a/rfcs/20260421-improve-functional-coverage/harness_matmul_core_ci
+++ b/rfcs/20260421-improve-functional-coverage/harness_matmul_core_ci
@@ -1,0 +1,100 @@
+--reset --dt=f16:f32:bf16 --stag=abx --wtag=abx --dtag=abx --attr-scales=src:host_scalar:4:f32+wei:host_scalar:4:bf16 --attr-post-ops=exp 872x1312:1312x148
+--reset --dt=f16:f4_e2m1:f16 --stag=abx --wtag=abx --dtag=abx --attr-scales=wei:per_tensor:f8_e4m3:16x1 2x6x1x1x896:1x6x1x896x932
+--reset --dt=f16:s4:f16 --stag=abx --wtag=ba --dtag=abx --attr-scales=wei:per_tensor:f16:64x1 --attr-zero-points=wei:per_oc:s4 --attr-fpmath=strict:true 1x1088:1088x2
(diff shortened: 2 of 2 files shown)


rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+The more complete functional coverage tests provide the higher chances are the change is safe.
+This proof must come in reasonable time. If it takes a long time (e.g., a day) to verify any kind of change, such process gives less value for the development compared to one that responds to results much quicker.
+However, expanding functional coverage with the bigger number of test cases takes more testing time, thus, the resulting vision statement takes the following form:
+The balance between execution time and coverage must be kept wisely; the team must achieve the most possible coverage under specific time budget.
+
+# Goals
+Here are the ultimate goals the team wants to achieve:
+* Time resources spent per single validation instance, or testing job (under assumption of infinite hardware resources):
+    - CI: 2 hours.
+    - Nightly: 8 hours.
```
rjoursler, line 17: What are your thoughts on extending tests based on hardware? There is a massive performance difference between something like a tiger lake iGPU test system and a Sapphire Rapids Server with a PVC GPU. Additionally, there is an opportunity for transfer validation  possible by using a faster machine to run (as close as possible) the primitive implementation used on a slower machine.

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+- Determine the list of features for a given driver that would form the core functionality.
+    * Core functionality must satisfy the following requirements:
+        - It must be shared between CPU and GPU backends.
+        - It should have optimized support in the library. It doesn't necessarily mean the optimized scope must be identical between backends.
+    * Core functionality should occupy the most space of coverage for a driver.
+    * Test cases targeting core would reside in `harness_driver_core_ci` file.
+- Determine the list of exclusive features for a backend or limited support scope.
+    * Such features should occupy a small portion of the total number of all test cases.
+    * Test case targeting such features would reside in `harness_driver_featureName_ci`.
+        E.g., ‘featureName’ can be 'f64' which is exclusive to GPUs and exists in various primitives.
```
rjoursler, line 47: I am not convinced we should be adding files specific to a feature. At it's core, the purpose of test sampling is to turn bug identification into a [coupon collector](https://en.wikipedia.org/wiki/Coupon_collector) problem over the sample space. As a consequence the number of tests required to find almost all bugs should roughly grow as `O(num_features*log(num_features))`.  When we start introducing feature tests, we lose logarthimic scaling property of this method, and end up with testing taking `O(num_features * log(num_features) * num_extra_features)` to get equivalent validation. What if we try to group addition features by different backends, i.e, have a harness_driver_gpu_ci instead?

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+A smart and scalable test case generator that would provide a single (no iterations over any arguments) benchdnn test case as its output is envisioned as such a tool.
+Its smartness comes in two aspects:
+    - It's adapting to the library API combinations, returning only test cases that make sense from the library perspective, not a sporadic combination (a.k.a. black box testing).
+        E.g., quantization is applied to integer data types; test case masks follow dimensions used and reflect real-world applications; etc.
+    - Uses probability mechanism to balance the amount of most useful combinations versus corner cases (test most useful functionality more often).
+    - Control ranges of random values to reach the desired goal of 80% hit rate.
```
rjoursler, line 33: Do we have evidence that 80% an achievable target? From what I have seen, I am not convinced this can be practically managed once we are discussing different backends. The core issue is that when you combine different features, the probability of hitting an optimized implementation is `O((1-probability_feature_unsupported)^num_features)`, i.e, we have exponential decay over the number of features. Because of this, I suspect that it will be much simpler to handle this by just skipping testing unoptimized implementations in benchdnn if they greedily exceed some fixed rate.

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+Given the infinite space of possibilities of various combinations, the best shot is to create a tool that would generate distributed points in that combination space. Coarse distribution can cover more various, usually independent implementation-wise, features of the library versus the targeted specific feature validation through iteration over some of options (as it's done currently). A tool (over the manual update) is preferred as more reliable and faster way to extend the coverage points in shorter time.
+
+A smart and scalable test case generator that would provide a single (no iterations over any arguments) benchdnn test case as its output is envisioned as such a tool.
+Its smartness comes in two aspects:
+    - It's adapting to the library API combinations, returning only test cases that make sense from the library perspective, not a sporadic combination (a.k.a. black box testing).
+        E.g., quantization is applied to integer data types; test case masks follow dimensions used and reflect real-world applications; etc.
+    - Uses probability mechanism to balance the amount of most useful combinations versus corner cases (test most useful functionality more often).
```
rjoursler, line 32: Could you add some more information on how such a mechanism would work? When I have looked at this, the necessary controls generally seem like they would be excessively complex from a usability and maintenance perspective.

rfcs/20260421-improve-functional-coverage/README.md
echeresh: Is there clear evidence that this hardware-specific handling will be worth the investment?

My impression is that most of our timeout issues in testing could be prevented by smarter test case engineering. As a data point for matmul: LNL or PTL take an hour or so to run 20k CI matmul cases. I believe we can reduce this time budget even further with the random sampling approach while actually improving our coverage.

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Synthdnn
+Currently, there's already a script that can perform case generation, located at src/scripts/synthdnn.
+So far it supports various data types, fpmath-mode, tags and shapes.
+The problem is it lacks a probability architecture piece which makes it less controlled on the output it provides.
+It acts more like a black box testing without adjusting to actually supported scenarios which wouldn't help with the announced goals once more features are added to it.
+
+## Fuzzdnn
+The showcase work is done on the base of another internal generator, "fuzzdnn".
+Its latest output can be checked out [here](harness_matmul_core_ci). Notice the variability of knobs used per case, sizes and other settings.
```
echeresh, line 85: I noticed that MxNxK distribution is somewhat skewed in this file - e.g., too many M = 1 or N = 1 cases (but no K = 1 cases). Need to check this carefully.

Also a point worth checking - synthdnn relies on a log-uniform distribution. We may need to use it as well. However, I would first check if it gives better coverage vs uniform in practice. For example, one argument is that log-uniform can oversample smaller-sized cases which may prefer smaller-tile strategies leaving less representation for larger ones. Also a log-uniform distribution can use different bases to tweak this more.

We can compare the number of unique strategies vs number of cases in uniform vs log-uniform distribution. @rjoursler can probably give more background about the log-uniform approach.

rfcs/20260421-improve-functional-coverage/README.md
echeresh: I agree - it makes sense to limit such files. However, I'm not sure if we can avoid this at all. As an example, `--runtime_dims_masks` may need dedicated coverage (does it?) or large shapes (testing 32-bit overflow) or longer chains of post-ops. Also features like sparse GEMM or group GEMM may need special handling.

rfcs/20260421-improve-functional-coverage/README.md
rjoursler: My main motivations for the log-uniform distribution were:
* Smaller tests run faster, so we can (generally) get more testing for a given time budget
* Most significant strategy transition points happen when a dimension is relatively small as large dimensions saturate the device leading to semi-uniform behavior.
* To better scale to multidimensional workloads. For example a 5d workload with each dimension selected uniformly between [1, 4096] is 4,000,000x larger on average than a 3d workload with the same sampling method.
  * To some extent, my goal was that for each problem case we had a fixed compute budget and we are sampling within that budget.
* We do not need large problem sizes to (mostly) validate a given primitive, for example on GPU, once the problem is large enough to generate a few workgroups and run a few internal loop iterations, we will have hit all control flow that could result in errors. 
  * There is actually a more fundamental testing algorithm issue here, as we currently require large test sizes due to strategy selection heuristics. This results in lots of "wasted" computation to ensure we test all strategies adequately.

So long as we are getting sufficient coverage and testing performance, I don't think the specific sampling method is too important though.

rfcs/20260421-improve-functional-coverage/README.md
rjoursler: We can definitely have some extra files where it makes sense, but to me this sounds like we will hit difficulties with coverage. For example, lets say we have testing for large_buffer, runtime dimes, and longer post-op chains as separate extended features. When do we validate the combination `large_buffer + runtime_dims`? The more special feature tests we have, the harder it is to manage this complexity.

rfcs/20260421-improve-functional-coverage/README.md
dzarukin: My considerations towards hardware differentiation is this: though more powerful HW can fit more cases, true, it doesn't mean we have to do it. Uniformity would pay off better as it comes with predictability for every regular and devs jobs, and straight comparability between hardware results, which often helps with the question "did it fail just on XXX, or everywhere?"
Lacking uniformity would give harder times to identify this kind of information.

I orient on time as an essential and feelable metric. Coverage is not the only thing that affects the final time, the are definitely other factors including hardware to runs that coverage on, as you mentioned. Technically, this time target translates into a statement "the weakest hardware can't run testing coverage for a single driver for more than 8 hours minus time for build minus time for setting up, under assumption that we have enough silicon to parallelize all driver steps". This is exaggerated statement but we can use it as our guideline and solve problems along the way.

Re-phrasing it would sound like "let's not try to fit 300k cases to find all the bugs, let's use the more reasonable amount".

rfcs/20260421-improve-functional-coverage/README.md
dzarukin: It's a valid concern about additional complexity for separated files.
I had in mind separated features that are not developed any longer or very specific, but, maybe, it wouldn't provide the value of reducing number of test cases (really small % of total number of cases) for such features.
I'll drop them and will check if that affects skip rates.

rfcs/20260421-improve-functional-coverage/README.md
dzarukin: 80% hit rate number should be taken with a grain of salt. First, it's the percent of supported cases for a given configuration of hardware/backend.

The number expresses the desire to cover API capabilities that are optimized 4 out of 5 cases in general. It is put as a number to quantify that desire which can actually be measured. It doesn't mean if we hit 75-78% consistently, we'd refine coverage until 80% are hit, but if it is around 60%, it would require to analyze what's going on (dispatching or generation). I had mostly flagman products in mind to reach that high. Flagman minus two generation can have lower rates, at least, it is acceptable in my picture, but I'm happy to hear other opinions on this.

Skipping wouldn't help to reach the target, it would just make to run problems with optimized implementations. I specifically want to generate problems that use settings that hit optimized implementations. It's a combination of what's supported on all platforms and backends (which is basically a full API space of allowed values in it) crossed-out some "wild" values that API allows to set.

rfcs/20260421-improve-functional-coverage/README.md
dzarukin: Sure. Probability mechanism is something that is controlled by devs in terms of chances. Here's the syntax example:
```
self.ngroups = int_sampler("1-16*14|60-64:4**1|128*1")
```
It means "setup 16 buckets of values to pick an integer value from, where 14 buckets have a random value from 1 to 16, 1 bucket of values from 60 to 64 with a step of 4, and one bucket with a value of 128".
First it picks the bucket randomly, then it picks a value out of the bucket randomly. These 14, 1 and 1 is something to control the probability of cases to support.
Same thing applies to strings, one can put as many of one string and one or several of other strings that should be picked less often.

The syntax and idea is the courtesy of @echeresh.

echeresh (approved):

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+The renewal process of core files (or re-generation) may be triggered by several factors:
+* Periodically, somewhere in the middle of development process for a coming release.
+* Once the time budget is exhausted, leading to longer jobs times stop fitting the time goal.
+* Once the scope of functionality changes, e.g. proportions of cases were adjusted in a major way to comply with the trend industry path.
+
+This tool is expected to be posted in oneDNN repository for everyone to use it.
+Every developer will be responsible for extending the tool to cover functionality they are adding whether it's a new feature or a specific important set of settings.
+Additional value of the tool is when it's mastered, a developer will be able to tailor local testing coverage for features/bugs they are working on.
+
+The final state of input files includes a various harness_driver_xxx_ci files and three test_driver_{smoke/ci/nightly} files.
```
jondea, line 70: Do we need the word `harness`? If we are getting rid of the existing input files, and these will be the default, why not just `driver_xxx_ci`?

rfcs/20260421-improve-functional-coverage/README.md
```diff
@@ -0,0 +1,86 @@
...
+Validation exists to prove a change, or a patch set is safe to promote into the repo.
+(Though validation should cover both functional and performance aspects, this RFC will focus on the functional part of it).
+The more complete functional coverage tests provide the higher chances are the change is safe.
+This proof must come in reasonable time. If it takes a long time (e.g., a day) to verify any kind of change, such process gives less value for the development compared to one that responds to results much quicker.
+However, expanding functional coverage with the bigger number of test cases takes more testing time, thus, the resulting vision statement takes the following form:
+The balance between execution time and coverage must be kept wisely; the team must achieve the most possible coverage under specific time budget.
+
+# Goals
+Here are the ultimate goals the team wants to achieve:
+* Time resources spent per single validation instance, or testing job (under assumption of infinite hardware resources):
```
jondea, line 15: Is this total runtime (i.e. all the runner times added up) or the elapsed wallclock time?

rfcs/20260421-improve-functional-coverage/README.md
jondea: I'm inclined to agree with @rjoursler that log-uniform is a better choice. It is simple and robust. Unless I have good reason, I always sample dimensions from a log-uniform (or log equidistant).

A couple of other points:

- All numbers in the range are possible (nowhere to hide bugs!)
- 1s appear a lot, but they also appear a lot in real problems. So many problems use singleton dimensions which typically have special handling in the kernels. The interesting behaviour happens on the boundaries.

I think this tool should lean into the randomness, there's a exploratory power to not letting developers tweak the distribution too much.

I do think it would be interesting to see how well the different distributions explore the space though.

rfcs/20260421-improve-functional-coverage/README.md
jondea: I guess the key question is: how many resoucres do we want to spend testing ref? We need to test it a _bit_, but we don't want to spend most of our time on it.

Also, from my experience bringing up the AArch64 backend, it was tough early on because our test cases were dominated by ref and so our CI was slow.

Could we balance these two requirements by having a small and a big set (as well as CI/nightly) where we skip ref on the big set and don't skip ref on the small set. That way we get an unbiased sample of test cases on ref, but we don't spend too many computational resources checking a conceptually simpler implementation.

jondea (commented):
I think this is a great idea, I think a sparse random sample from the space of parameters is much more efficient and robust than the cartesian product.

rfcs/20260421-improve-functional-coverage/README.md
rjoursler: > Could we balance these two requirements by having a small and a big set (as well as CI/nightly) where we skip ref on the big set and don't skip ref on the small set.

Alternatively, it may also make sense to have a small set that only tests ref instead. When enabling a new architecture on GPU, I know that we have hit the following scenario on oneDNN releases: 

unrecognized architecture  -> fallback to ref -> ref bug causes issues

Due to our testing process, these reference bugs were never found as we would always dispatch to optimized implementations during development. Another advantage of this method is that it will be easier to model test performance.  For example, if the reference implementation is running at `1/10000`  the performance of an optimized implementation on a problem (I have seen this occur when a reference fails to parallelize on a GPU), it becomes very easy for a single problem to exceed the testing budget. This kind of performance disparity is hard to account for with the current setup. This reference test set can then be constrained to the Nightly test suite as reference development is relatively uncommon.

echeresh: > to have a small set that only tests ref instead

As an option, we can reuse smoke scope input files for that. If they are small/medium sizes then GPU reference is not terribly slow. Even with 5 minutes time budget we can get decent coverage out of that.

jondea: If we skip ref on the big set, we may want to enforce a certain number of passes on each platform so that we don't silently lose coverage

rjoursler (approved):

vpirogov (approved):

PR #5077, opened 2026-04-24
Add a new GEMV kernel to BRGEMM and enable it in MatMul

densamoilov:
This PR adds a new GEMV kernel to BRGEMM to support the remaining cases and complete GEMV coverage.

The existing and the new GEMV kernels enable all four GEMV cases required for full support across layout and parameter combinations.

**GEMV coverage in MatMul**
| Vector dimension | A layout | B layout | Corresponding BRGEMV Operation | BRGEMV `transA` parameter | BRGEMV `treat_y_as_row` parameter |
|------------------|----------|----------|--------------------------------|---------------------------|------------------------------------------|
| N = 1            | ab       | ab,ba    | `y = A * x`                    | false                     | n/a                                      |
| M = 1            | ab,ba    | ba       | `yᵀ = xᵀ * Aᵀ`                 | false                     | if true output is `yᵀ`                   |
| N = 1            | ba       | ab,ba    | `y = Aᵀ * x`                   | true                      | n/a                                      |
| M = 1            | ab,ba    | ab       | `yᵀ = xᵀ * A`                  | true                      | if true output is `yᵀ`                   |

**Note**
* `transA`: selects whether the BRGEMV uses `A` or `Aᵀ`
* `treat_y_as_row`: for M=1, interprets `y` as a row vector
* Batch dimensions are supported
* Bias, post-ops and scales are supported

At the matmul level, these GEMV configurations are represented via `gemv_strategy_t`.
At the BRGEMM level, they are implemented using `transA` and `treat_y_as_row`.

**Performance**
Performance was evaluated on ADL and SRF showing parity with auto-generated GEMM kernels.
As a result, GEMM implementations are no longer used in performance validation and have been fully replaced by BRGEMM matmul.
image
image


densamoilov:
make test

ankalinin (approved):

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -2372,37 +2680,47 @@ template <typename Wmm>
...
-    for (dim_t ld = 0; ld < ld_block2; ld++) {
-        load_B(load(), rd, ld);
         for (dim_t bd = 0; bd < bd_block; bd++) {
-            load_A(bcst(), bd, rd);
-            auto acc = accm(ld_block2, bd, ld);
-            uni_vfmadd231ps(acc, bcst(), load());
+            const bool is_tail_acc
+                    = gemv_is_tail_acc(bd, bd_block, is_bdb_tail);
+            // TODO: adjust A_offset(bd, 0)
+            const auto a_addr = ptr[reg_aux_A + bd * 8 * sizeof(float)];
```
georgen117: Suggestions:
1. remove the TODO unless there is something you still need to do here.
2.replace
```
const auto a_addr = ptr[reg_aux_A + bd * 8 * sizeof(float)];
```

with

```
const auto a_addr = ptr[reg_aux_A + bd * vreg_traits_t<Vmm>::vlen];
```

This removes the hardcoded `8` and make is clear that the bd-accumulator is a full register width.

I did a little checking to make sure the `8` is always valid. It looks like this always takes a Ymm register so the `8` is valid. However, changing the code to remove the hardcoded `8` would improve maintenance.

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -381,6 +392,76 @@ struct brgemm_desc_t {
...
+    //
+    // non-transA:
+    //   - same as `bd_block`
+    //
+    // transA:
+    //  - number of full vector accumulators updated per reduction step
+    //  - equal to `gemv_transa_bd_unroll`
+    dim_t gemv_bd_block() const {
+        assert(is_gemv);
+        if (!is_gemv) return 0;
```
georgen117, line 405: if callers guarantee `is_gemv` as implied by the `assert` then the `if (!is_gemv) return 0;` is not needed.

- if callers guarantee `is_gemv`, keep only the assert (same as `gemv_acc_is_vector(), and `gemv_num_acc_blocks(...)`)
- If the method must be safe to call from non-GEMV paths, drop the `assert` and keep only the `if`.

src/cpu/x64/brgemm/brgemm_types.hpp
```diff
@@ -381,6 +392,76 @@ struct brgemm_desc_t {
...
+    // non-transA:
+    //   - same as `bdb_tail`
+    //
+    // transA:
+    //   - number of full vector accumulators in the tail block
+    //   - the remaining first-level register tail is stored separately in
+    //     `gemv_tail`
+    dim_t gemv_bdb_tail() const {
+        assert(is_gemv);
+        if (!is_gemv) return 0;
```
georgen117, line 420: Same as above, if callers guarantee `is_gemv` as implied by the `assert` then the `if (!is_gemv) return 0;` is not needed.

- if callers guarantee `is_gemv`, keep only the assert (same as `gemv_acc_is_vector(), and `gemv_num_acc_blocks(...)`)
- If the method must be safe to call from non-GEMV paths, drop the `assert` and keep only the `if`.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -575,9 +620,13 @@ dim_t jit_brgemm_kernel_t<Wmm>::bdb_po_offset(dim_t bd_block2) const noexcept {
     return bd_block2 * brg.bd_block * brg.LDD;
 }
 
+// TODO: check that these offsets are correct or 4 different kinds of offsets are needed
```
georgen117: As best I can tell the offsets are correct.  Does this TODO still need to be here? What remains to be checked?

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -454,9 +479,29 @@ bool is_gemv_applicable(const brgemm_matmul_conf_t &bgmmc,
...
+            = bm_conf_utils.get_gemv_strategy(gemv_A_tag, gemv_B_tag);
+
+    const bool is_transA_strategy = utils::one_of(gemv_strategy,
+            gemv_strategy_t::n1_A_trans, gemv_strategy_t::m1_B_plain);
+    const bool is_swap_strategy = utils::one_of(gemv_strategy,
+            gemv_strategy_t::m1_B_plain, gemv_strategy_t::m1_B_trans);
+
+    // transA GEMV requires contiguous output (INCY == 1) for tensors with
+    // batch dimensions. Some batched layouts, e.g. `bac`, place the patch
+    // dimension between consecutive GEMV output elements and make INCY > 1.
```
georgen117, line 495: possible typo in comment: `patch dimensions` -> `batch dimensions`?

src/cpu/x64/brgemm/brgemm_utils.cpp
```diff
@@ -722,31 +725,109 @@ status_t brgemm_blocking_tmm(brgemm_desc_t *brg) {
...
+
+        brg->gemv_tail = brg->rdb_tail;
+
+        return status::success;
+    }
+
+    // Blocking parameters for the transposed case.
     brg->ld_block = 1;
     brg->ldb = brg->load_dim / brg->ld_block;
     brg->ldb_tail = brg->load_dim % brg->ld_block;
```
georgen117, line 813: Low: Should this have an `asset(brg->ldb_tail ==0)`  like line 790 of the not transA path above?

both calculations are the same why does one have an assert and not the other?

src/cpu/x64/brgemm/brgemm_types.hpp
densamoilov: This function is supposed to be called only for the GEMV code paths to enforce better GEMV and non-GEMV code separation at the brgemm level. However, just because it's not supposed to be called for non-GEMV code paths doesn't mean that it won't happen and it may not always be possible to catch such misuse in debug mode. So given that we don't use exceptions and we don't want to overcomplicate simple functions by making the return a status we use asserts plus a deterministic fallback to avoid UB in non-debug modes. Returning 0 here is relatively safe as it means that 0 blocks will be handled, which will just skip most of the kernel code.

I think it would make more sense to make `gemv_num_acc_blocks()` always return 0 (which is already true but implicit) for non-GEMV cases and `gemv_acc_is_vector()` can be left as is because whether it's true or false doesn't make much difference and transA is not GEMV-specific.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: Oh, I left this TODO as part of the development process and simply overlooked it. It has now been completed. Thanks for pointing it out.

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: This one had been completed. I just forgot to remove the comment. Removed it now, thanks.

src/cpu/x64/brgemm/brgemm_utils.cpp
densamoilov: Yeah, let's add it.

densamoilov:
make test

densamoilov:
make test

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
```diff
@@ -481,9 +501,49 @@ struct jit_brgemm_kernel_t : public jit_base_brgemm_kernel_t {
     palette_config_t palette_;
 };
 
+/**
+ * Checks whether the current accumulator is the final partial veector
```
inteldimitrius: nit: `veector` => `vector`

inteldimitrius (approved):

densamoilov:
make test

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
densamoilov: fixed

src/cpu/x64/brgemm/jit_brgemm_kernel.cpp
densamoilov: fixed

PR #4275, opened 2025-11-03
rfcs: gpu: add Level Zero runtime

spalicki:
This is a proposal for adding a Level Zero runtime using immediate command lists. Rendered version available [here](https://github.com/uxlfoundation/oneDNN/tree/spalicki/rfcs/l0_runtime/rfcs/20251103-level-zero-runtime).

PR #5091, opened 2026-04-29
doc: add Level Zero runtime materials

dzarukin:
This PR adds documentation for Level Zero GPU runtime.


dzarukin:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,87 @@
...
+
+1. dnnl::ze_interop::execute() interface
+
+    This interface enables the user to pass dependencies between primitives
+    using Level Zero events. In this case, the user is responsible for passing
+    proper dependencies for every primitive execution.
+
+2. In-order oneDNN stream
+
+    oneDNN enables the user to create in-order streams when submitted primitives
```
densamoilov:
```suggestion
    oneDNN enables the user to create in-order streams in which submitted primitives
```

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,87 @@
...
+    are executed in the order they were submitted. Using in-order streams
+    prevents possible read-before-write or concurrent read/write issues.
+
+@note oneDNN doesn't own user-provided Level Zero objects. This ensures that
+they will not be destroyed while oneDNN objects store a reference to it.
```
densamoilov:
```suggestion
they will not be destroyed while oneDNN objects store a reference to them.
```

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,87 @@
...
+@note The access interfaces do not retain the Level Zero object. It is the
+user's responsibility to retain the returned Level Zero object if necessary.
+
+@note Current version of API manages Level Zero events lifetime, and it's
+attached to the stream lifetime. All returned events after stream destruction
+become invalidated.
+
+@note USM memory doesn't support retain/release Level Zero semantics. When
+constructing a oneDNN memory object using a user-provided USM pointer oneDNN
+doesn't own the provided memory. It's user's responsibility to manage its
```
densamoilov:
```suggestion
doesn't own the provided memory. It's the user's responsibility to manage its
```

densamoilov (approved):

doc/advanced/level_zero_interoperability.md
dzarukin: Thanks, applied all edits.

dzarukin:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

mzhukova (approved):
Should we mention that it is not supported for graph API? Otherwise LGTM!

doc/rst/programming_model.rst
```diff
@@ -10,4 +10,4 @@ oneDNN Concepts
    dev_guide_data_types
    page_cross_engine_reorder_cpp
    dev_guide_c_and_cpp_apis
-   interop_with_dpcpp_and_opencl
\ No newline at end of file
+   runtimes_interoperability
```
vpirogov: I think this will result in 404 for this link: https://uxlfoundation.github.io/oneDNN/interop_with_dpcpp_and_opencl.html

Maybe remove the old page from index and keep it in orphans?

doc/build/build_options.md
```diff
@@ -7,7 +7,7 @@ oneDNN supports the following build-time options.
 |:--------------------------------|:----------------------------------------------------|:-------------------------------------------------------------------------------------------------------------------|
 | ONEDNN_LIBRARY_TYPE             | **SHARED**, STATIC                                  | Defines the resulting library type                                                                                 |
 | ONEDNN_CPU_RUNTIME              | NONE, **OMP**, TBB, SEQ, THREADPOOL, SYCL           | Defines the threading runtime for CPU engines                                                                      |
-| ONEDNN_GPU_RUNTIME              | **NONE**, OCL, SYCL                                 | Defines the offload runtime for GPU engines                                                                        |
+| ONEDNN_GPU_RUNTIME              | **NONE**, OCL, SYCL, ZE                             | Defines the offload runtime for GPU engines                                                                        |
```
vpirogov, line 10: There must be a story behind `ZE` vs `L0`. Since `ZE` is not self-explanatory some kind of explanation in description will be helpful.

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,87 @@
...
+
+- dnnl::ze_interop::make_memory(const memory::desc &, const engine &, void \*)
+
+    Constructs a USM-based memory object. The `handle` could be one of special
+    values #DNNL_MEMORY_ALLOCATE or #DNNL_MEMORY_NONE, or it could be a
+    user-provided USM pointer.
+
+## Handling Dependencies
+
+Only Level Zero in-order immediate command lists are supported.
```
echeresh, line 56: Maybe put it as a `@warning` or `@note`?

echeresh (approved):

vpirogov (approved):

dzarukin:
> Should we mention that it is not supported for graph API? Otherwise LGTM!

It doesn't look like that we provide any documentation on interoperability with Graph API nor specify any limitations. I guess, it's only a build time error for us.

doc/build/build_options.md
dzarukin: [This](https://github.com/uxlfoundation/oneDNN/pull/4275#discussion_r2662366561) is the thread that lead to this naming. It was picked for compatibility with level zero APIs. If users are about to use Level Zero runtime, they must be familiar with their API naming.

doc/advanced/level_zero_interoperability.md
dzarukin: Added a note, thanks.

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,88 @@
...
+    Constructs a USM-based memory object. The `handle` could be one of special
+    values #DNNL_MEMORY_ALLOCATE or #DNNL_MEMORY_NONE, or it could be a
+    user-provided USM pointer.
+
+## Handling Dependencies
+
+@note
+Only Level Zero in-order immediate command lists are supported.
+
+Unlike OpenCL API, Level Zero API doesn't have a notion of retain/release
```
rjoursler: "Unlike the OpenCL API, the Level Zero API" ...

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,88 @@
...
+    proper dependencies for every primitive execution.
+
+2. In-order oneDNN stream
+
+    oneDNN enables the user to create in-order streams in which submitted
+    primitives are executed in the order they were submitted. Using in-order
+    streams prevents possible read-before-write or concurrent read/write issues.
+
+@note oneDNN doesn't own user-provided Level Zero objects. This ensures that
+they will not be destroyed while oneDNN objects store a reference to them.
```
rjoursler: > This ensures that they will not be destroyed while oneDNN objects store a reference to them.

This doesn't seem to be saying the right thing. Shouldn't this be another "It is the user responsibility to ensure ...".

rjoursler (approved):

doc/advanced/level_zero_interoperability.md
dzarukin: Updated, thanks.

doc/advanced/level_zero_interoperability.md
dzarukin: I've removed this sentence and put a new one in the paragraph above:

> Thus, objects created by oneDNN are managed by oneDNN, and objects created by the user and passed to oneDNN are not managed by oneDNN, the library only stores references to objects.

> It means it's the user responsibility to manage those objects lifetime while the library operates with them.

Would you find it fitting better?

doc/rst/programming_model.rst
dzarukin: I've kept the old name for a file but updated the content. Left an additional note on the reasoning in commit message. Please review.

dzarukin:
make test
disable build_vendor_intel
disable run_source_scans
enable run_documentation

doc/advanced/level_zero_interoperability.md
rjoursler: That is better. One more note in the final sentence: "those objects" is a bit unclear as we are discussing two different groups of objects. I think we should explicitly call out that we are discussing user created objects.

A side question: is there a mechanism for users to know whether oneDNN is still retaining a given object. In particular, I am wondering how this interacts with objects stashed in the kernel/primitive cache.

doc/advanced/level_zero_interoperability.md
```diff
@@ -0,0 +1,88 @@
...
+| Stream             | `ze_command_list_handle_t`                                            |
+| Memory (USM-based) | Unified Shared Memory (USM) pointer                                   |
+
+The table below summarizes how to construct oneDNN objects based on Level Zero
+objects and how to query underlying Level Zero objects for existing oneDNN
+objects.
+
+| oneDNN object      | API to construct oneDNN object                                                             | API to access Level Zero object(s)                                                                                                                |
+|:-------------------|:-------------------------------------------------------------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------------------------|
+| Engine             | dnnl::ze_interop::make_engine(ze_driver_handle_t, ze_device_handle_t, ze_context_handle_t) | dnnl::ze_interop::get_driver(const engine &) <br> dnnl::ze_interop::get_device(const engine &) <br> dnnl::ze_interop::get_context(const engine &) |
```
vpirogov: `<br>` does not render right in HTML. Consider something like this:

```suggestion
| Engine             | dnnl::ze_interop::make_engine(ze_driver_handle_t, ze_device_handle_t, ze_context_handle_t) | dnnl::ze_interop::get_driver(const engine &)  |
| \                  | \                                                                                          | dnnl::ze_interop::get_device(const engine &)  |
| \                  | \                                                                                          | dnnl::ze_interop::get_context(const engine &) |

```

doc/rst/programming_model.rst
vpirogov: Looks good now.

doc/advanced/level_zero_interoperability.md
dzarukin: Updated in all related documents, thanks.

doc/advanced/level_zero_interoperability.md
dzarukin: > I think we should explicitly call out that we are discussing user created objects.

Updated, thanks.

> A side question: is there a mechanism for users to know whether oneDNN is still retaining a given object. In particular, I am wondering how this interacts with objects stashed in the kernel/primitive cache.

The kernel depends on the module and the module depends on the context. We keep both the kernel and the module alive on our side without exposing them to the user, and I think Level Zero tracks module-context relations internally to keep track dependencies to ensure module destruction would be safe. At least, till now there were no crashes at global destruction with the cache enabled for Level Zero kernels.

PR #5092, opened 2026-04-29
cpu: aarch64: add f16 and bf16 support to jit_io_helper and jit_uni_binary

Sqvid:
# Description

This PR adds `f16` and `bf16` support to `jit_io_helper_t`. I have also demonstrated the impact of these changes by adding support for these data-types to `jit_uni_binary_t`.

```
# Dispatch before:
$ ./build/tests/benchdnn/benchdnn --binary --batch=test_binary_{float16,bfloat16} --batch=harness_binary_{f16,bf16}
...
| ref:any : 8900 (100%)                                    |
tests:20000 passed:8872 skipped:11100 mistrusted:28 unimplemented:0 invalid_arguments:0 failed:0 listed:0

# Dispatch after:
$ ./build/tests/benchdnn/benchdnn --binary --batch=test_binary_{float16,bfloat16} --batch=harness_binary_{f16,bf16}
...
| ref:any : 6240 (54%)                                     |
| jit:uni : 5320 (46%)                                     |
tests:20000 passed:11520 skipped:8440 mistrusted:40 unimplemented:0 invalid_arguments:0 failed:0 listed:0
```

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/utils/jit_io_helper.cpp
+++ b/src/cpu/aarch64/utils/jit_io_helper.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2021 Intel Corporation
 * Copyright 2022 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -73,14 +73,14 @@ jit_io_helper_t<Vmm>::jit_io_helper_t(jit_generator_t *host,
     , saturation_conf_(saturation_conf)
     , gather_conf_(gather_conf) {
 
-    assert(utils::one_of(data_type_, data_type::f32, data_type::s8,
-                   data_type::u8, data_type::s32)
-            && "Supported data types f32, s8, u8, s32");
+    assert(utils::one_of(data_type_, data_type::f16, data_type::bf16,
+                   data_type::f32, data_type::s8, data_type::u8, data_type::s32)
+            && "Supported data types f16, bf16, f32, s8, u8, s32");
 
     static constexpr bool is_zmm
             = std::is_same<Vmm, Xbyak_aarch64::ZReg>::value;
     MAYBE_UNUSED(is_zmm);
-    assert(IMPLICATION(!is_superset(isa_, sve_128), !is_zmm)
+    assert(IMPLICATION(!is_superset(isa_, sve), !is_zmm)
             && "This architecture does not support z registers.");
 }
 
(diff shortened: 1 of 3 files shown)


src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -514,6 +527,26 @@ void jit_io_helper_t<Vmm>::load_s32(const Xbyak_aarch64::XReg &src_addr,
     host_->scvtf(dst_vmm.s, host_->P_TMP / Xbyak_aarch64::T_m, dst_vmm.s);
 }
 
+template <typename Vmm>
+void jit_io_helper_t<Vmm>::load_bf16(const Xbyak_aarch64::XReg &src_addr,
```
michalowski-arm, line 531: I guess I wonder if we need these single use helpers like `load_bf16` if they only seem to be consumed by `load`? I feel it may be clearer to simply write the per-dt load logic directly in `load`, unless we expect to use them elsewhere as well?

Sqvid: I think it's simply nicer to have separate functions for maintainability. `load_bf16/s8/etc..` are all private methods anyway, so the user only sees `load()` anyway.

michalowski-arm: It's up to personal preferences I suppose, not a problem then.

michalowski-arm (approved):
LGTM

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -602,6 +641,51 @@ void jit_io_helper_t<Vmm>::store_f32(const Vmm &src_vmm,
...
+    if (mayiuse_bf16()) {
+        host_->bfcvt(Xbyak_aarch64::ZRegH(src_vmm.getIdx()),
+                mask / Xbyak_aarch64::T_m,
+                Xbyak_aarch64::ZRegS(src_vmm.getIdx()));
+        host_->st1h(Xbyak_aarch64::ZRegS(src_vmm.getIdx()), mask,
+                Xbyak_aarch64::ptr(dst_addr));
+        return;
+    }
+
+    host_->str(host_->z31,
```
jondea, line 659: How do we know we can always use `z31`? Also, I think a comment explaining that this is a fallback would be useful.

src/cpu/aarch64/utils/jit_io_helper.cpp
```diff
@@ -682,6 +766,17 @@ void jit_io_helper_t<Vmm>::convert_to_f32(const Vmm &dst_vmm,
             host_->uni_scvtf(dst_vmm.s, dst_vmm.s);
             break;
         }
+        case data_type::bf16: {
+            if (dst_vmm.getIdx() != src_vmm.getIdx())
+                host_->mov(dst_vmm.s, host_->P_ALL_ONE / Xbyak_aarch64::T_m,
```
jondea, line 771: Can we get rid of this `mov` with

```
host_->lsl(dst_vmm.s, src_vmm.s, 16);
```

If not, a comment explaining why would be useful

jondea (changes requested):

Sqvid:
Abandoning this in favour of reworking the io helper completely.

PR #5097, opened 2026-04-29
benchdnn: preparation for better skipped statuses accounting

dzarukin:
Proposal #5071 suggests better functional coverage and sets some preliminary goals in optimized implementations hit rates.

When the case hits skipped status, it's important to understand the reason what functionality is not supported and why. Current implementation doesn't help to understand the precise statistics of certain events happening, such as when it's specific configs are not supported, or something general is not working on a particular system.

The plan is to build a better tracking system for skipped cases on benchdnn side through personalized skip tags translated into clearer messages of reasons and also providing a summary table with all collected skipped statuses per run same as benchdnn provides for impl name hits.

This PR reworks reasons on benchdnn side, and add a new printer that handles filename and line addition automatically and also checks the formatting (same as verbose).
Additionally, e3m0 support is dropped from benchdnn as data type is deprecated.


dzarukin:
make test

dzarukin:
make test

tests/benchdnn/common.cpp
```diff
@@ -54,14 +54,21 @@ const char *state2str(res_state_t state) {
...
+        case reason_t::invalid: return "Invalid case";
+        case reason_t::failed_ref_not_expected: return "Ref Impl Not Expected";
+        case reason_t::skip_not_enough_ram: return "Not enough RAM";
+        case reason_t::skip_impl_hit: return "Skip-impl option hit";
+        case reason_t::skip_start: return "Skip-start option hit";
+        case reason_t::skip_not_supported: return "Case not supported";
+        case reason_t::skip_data_type: return "Data type not supported";
+        default: assert(!"unknown reason"); return std::string();
+    }
+    return std::string();
```
mzhukova: unreachable code seems like (there is a return even in default case)

tests/benchdnn/common.hpp
```diff
@@ -99,6 +99,52 @@ extern bool check_ref_impl;
...
+    printf("%s (%s:%zu)\n", msg.c_str(), filename, line);
+}
+
+// This is a new implementation to move benchdnn print calls to.
+// Note: using former one can't be done as is since the 'format' is mixed with
+// other string literals.
+#define BENCHDNN_PRINTF(v, fmt, ...) \
+    do { \
+        if (verbose >= v) { \
+            format_type_checker(fmt, __VA_ARGS__); \
```
mzhukova, line 145: is it safe to have `__VA_ARGS__` twice here? e.g., if someone does `BENCHDNN_PRINTF(..., do_smth)` then `do_smth` would be done twice..?

mzhukova (commented):
what would be happening after these changes if we pass `f4_e3m0` in any form? is it just up to library now to return correct status?

tests/benchdnn/common.hpp
dzarukin: It would call the function twice, that is correct. I didn't think of calling a function from this macro, and checker is not working with c++ template stuff (unless it's c++23, there's std::format + std::print combo), so either we are not checking formatting, or we don't call functions that are critical to be called once.

Given the intention, I'd probably would like to go with formatting as the only use case in mind is strings and plain values dumping, not function calls. But I've put a fat warning about the behavior of functions.

Thanks for spotting this behavior.

tests/benchdnn/common.cpp
dzarukin: True, removed, thanks.

mzhukova (approved):

dzarukin:
> what would be happening after these changes if we pass `f4_e3m0` in any form? is it just up to library now to return correct status?

Those are interesting questions that you ask...
After giving it a thought, I updated the debug_generation script to drop the value from dt2str function so that users can't pass this value to benchdnn (it will return unknown data type) and also drop it from supported data types in memory_desc_create, so users can't create them. Thus, it should be safe, the only action would be to cleanup all internal from this data type, but from this moment it's basically unfunctional. Please check it out.

dzarukin:
make test

TaoLv (approved):

dzarukin:
make test

PR #5101, opened 2026-04-30
cpu: aarch64: generalize post-op fallback execution

leoken01:
# Description

  This PR fixes ACL post-op eltwise execution for AArch64 primitives so nested
  eltwise post-ops run on the current intermediate buffer instead of always using
  the original destination argument.

  The issue affects ACL post-op chains where an eltwise post-op is executed after
  another post-op and needs to consume the current post-op buffer. The fix executes
  nested eltwise post-ops through the primitive interface with runtime memory bound
  to the active source buffer.

  This PR also registers nested scratchpads for ACL post-op primitives in the
  affected AArch64 primitives.

  Fixes # N/A

  # Validation

  ```sh
  ONEDNN_VERBOSE=profile_exec ./build/tests/benchdnn/benchdnn --graph --mode=R --in-shapes=0:16x256+1:256x1+2:1x1
  --case=pattern/int8/int8_matmul_logistic_fusion.json

  Result:
```

  # Checklist

  ## General

  - [x] Do all unit and benchdnn tests (make test and make test_benchdnn_*) pass locally for each commit?

  ### Bug fixes

  Fixes segfault here: https://github.com/uxlfoundation/oneDNN/pull/3950
  Enables feature here: https://github.com/uxlfoundation/oneDNN/pull/3767

### CODE REVIEW

File: src/cpu/aarch64/acl_post_ops.cpp [hunk 1]
```diff
@@ -53,45 +90,13 @@ status_t acl_post_ops_t::execute(
         } else if (post_op->kind() == primitive_kind::eltwise) {
             // The post op at the sum index must be binary
             if (post_op_index == sum_index) return status::runtime_error;
-
-            auto eltwise_post_op
-                    = dynamic_cast<acl_eltwise_fwd_t *>(post_op.get());
-            if (eltwise_post_op == nullptr) return status::runtime_error;
-
-            if (dst_data_type == data_type::f16) {
```
michalowski-arm, line 61: Looks like we dropped our special handling for f16 eltwise in this patch?

michalowski-arm (dismissed):

src/cpu/aarch64/acl_post_ops.cpp
leoken01: Do f16 post ops need to be included specifically? This was changed to allow oneDNN selected primitives to define the data types, @puneetmatharu may know more on this?

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
```diff
@@ -231,6 +231,7 @@ status_t acl_lowp_matmul_t::pd_t::init(engine_t *engine) {
     // quantization/dequantization layer has empty workspace.
     auto scratchpad = scratchpad_registry().registrar();
     CHECK(init_scratchpad(scratchpad, aux_mem_req));
+    acl_post_ops.init_scratchpad(scratchpad);
```
jondea: Do we need to `CHECK` this too?

src/cpu/aarch64/acl_post_ops.cpp
jondea: This was previouslty numerically incorrect, the way to do f16 post ops is to do a f16*f16->f32 matmul, do post ops in f32 and then cast down to f16 after. It is worth checking that we haven't dropped support for anything though @leoken01. I would build up a benchdnn input set which previously ran with ACL, and just check they all still go to ACL after

jondea (changes requested):

src/cpu/aarch64/acl_post_ops.cpp
leoken01: ACK, checking and will report back

leoken01: What kind of criteria or checks should the benchdnn input set look for when checking if they still go after acl? Not sure what to 'base' the input set on, to make sure it's correct and we're not missing anything.

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
leoken01: This depends on the value of `acl_post_ops.init_scratchpad(scratchpad);` if it returns void, then there's nothing to check which is why CHECK is not used right now, but I think if this returns something like `status_t` which it currently doesn't, we would need to add CHECK

src/cpu/aarch64/acl_post_ops.cpp
michalowski-arm: I would go with `tests/benchdnn/inputs/matmul/test_matmul_ci` as a good starting point, and then maybe try to stress test different eltwise post-ops combinations from there. Look for the number of problems dispatched to ACL matmul before and after the patch and any correctness issues that might show.

src/cpu/aarch64/acl_post_ops.hpp
jondea: Now you have generalised these post ops to use not just acl, I think it would make sense to rename this as something more descriptive. E.g.

```
post_ops_fallback_t
```

leoken01: do we want this just for acl_post_ops.hpp, or (I;m assuming) all files within this workspace? (acl_lowp_matmul.cpp etc...)

jondea: Just this post ops class. The class is no longer specific to ACL, but all the other classes still are

src/cpu/aarch64/acl_binary.hpp
```diff
@@ -50,7 +50,7 @@ struct acl_binary_t : public primitive_t {
     private:
         arm_compute::Status validate(const acl_binary_conf_t &asp);
 
-        friend struct acl_post_ops_t;
+        friend struct post_ops_fallback_t;
```
Sqvid: I know this wasn't added by your patch but it does stand out to me. Using `friend` here strikes me as a code-smell. What private data does `post_ops_fallback_t` need here, and can it be shared without sharing all private information via `friend`?

jondea: It used to need it because it called `execute` directly, but it shouldn't need it anymore

leoken01: will remove it if it's ok to do so then.

leoken01: I tried to run a build, the code fails because of the friend removal, but I think it's to do with `init()` and more specifically `acl_binary_t::init()`.

leoken01: Not sure what the best move is in terms of making init public or adding some helper which constructs and initialises without needing private access.

Sqvid: I am fine with you leaving it as-is in this patch. We have some planned refactors to this code anyway.

jondea: Had a discussion with @leoken01 and I think we want this gone. Essentially, if we have truly decoupled the post ops class from ACL specific primitives, then we won't need these

leoken01: Going to resolve as latest patch decouples these from the post ops class.

src/cpu/aarch64/acl_post_ops.hpp
leoken01: ACK and done

src/cpu/aarch64/acl_post_ops.cpp
leoken01: Different combinations have been tested and stress tests done, it looks like this doesn't raise any correctness issues so far. If it's ok to do so, will resolve comment.

src/cpu/aarch64/cpu_isa_traits.hpp
```diff
@@ -107,6 +107,10 @@ constexpr cpu_isa_t to_vla_sve(const cpu_isa_t isa) {
     return (cpu_isa_t)(isa & sve);
 }
 
+constexpr cpu_isa_t to_post_ops_eltwise_isa(const cpu_isa_t isa) {
+    return isa == asimd ? asimd : to_vla_sve(isa);
+}
+
```
Sqvid: This does not look right to me. `to_vla_sve(asimd)` should already be equivalent to `asimd`.
Look at the `cpu_isa_t` definitions:
https://github.com/uxlfoundation/oneDNN/blob/563c5fcfcf8ad591b43d3dc7e0bc9292102eeba0/src/cpu/aarch64/cpu_isa_traits.hpp#L69-L72

`to_vla_sve()` just uses a logical `AND`:
https://github.com/uxlfoundation/oneDNN/blob/563c5fcfcf8ad591b43d3dc7e0bc9292102eeba0/src/cpu/aarch64/cpu_isa_traits.hpp#L105-L108

`sve` is defined as `sve_bit | asimd`. So I would expect that doing the logical `AND` between `sve` and `asimd` would result in `asimd` anyway. Could you clarify why this function is needed?

Sqvid (changes requested):

src/cpu/aarch64/cpu_isa_traits.hpp
leoken01: Good point, I wanted to make the asimd ISA selection explicit as a helper for c6g ASIMD ISA selection, but it doesn't change behaviour, if needed I can remove it.

Sqvid: Yes I think it is confusing to have two functions that do the same thing

jondea: To be fair, the name isn't particularly clear WRT to what it does to `asimd` but I thought
```
to_vla_sve_or_asimd
```
also wasn't clear. I'm open to suggestions!

leoken01: Done

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
```diff
@@ -2441,7 +2442,10 @@ TEST(test_convolution_execute, ConvBiasEltwise) {
         strm->wait();
         dst = eltwise_dst_ts.as_vec_type<float>();
         for (size_t i = 0; i < dst.size(); ++i) {
-            ASSERT_FLOAT_EQ(dst[i], param.ref_dst[i]);
+            if (param.op_kind == graph::op_kind::Mish)
```
Sqvid: How did you decide on `1e-5` as the tolerance? Why does only `Mish` require this?

Sqvid (changes requested):

src/cpu/aarch64/jit_uni_binary.cpp
```diff
@@ -83,7 +83,9 @@ static bool data_type_supported(const data_type_t dtype) {
 
 static bool data_format_supported(
         const memory_desc_wrapper &mdw, const cpu_isa_t isa) {
+    if (!is_superset(isa, sve_128)) return false;
```
Sqvid: Why not `!is_superset(isa, sve)`?

Sqvid (changes requested):

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
leoken01: I think 1e-5 is the best one for error allowance whilst not being an outright bug or failure, Mish only required this because it was the only one that failed even on the slightest rounding difference, essentially Mish assertion is very strictly set to fail upstream on c6g unless the c6g unit test itself is changed (which I presume we cannot do or don't want to do).

src/cpu/aarch64/jit_uni_binary.cpp
leoken01: I think for the isa guard we need to be specific with sve values, otherwise asimd can be excluded and also it might let other sve values through, even if it should only expect a concrete sve_128 or sve_256. However, if this is allowed, I can remove it.

jondea: `is_superset(asimd, sve_128) == is_superset(asimd, sve) == false`

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
Sqvid: We are modifying a common test here. It is important to know _why_ a given tolerance is needed. We cannot just adjust it to address the failures as they come up.

> I think 1e-5 is the best one for error allowance whilst not being an outright bug or failure

For example, why do you think that?

leoken01: I'm open to suggestion of changing the allowance or code, but how or in what efficient way could we determine the accuracy needed whilst accounting for functionality of code (i.e. the test not failing). Unless we can change the parameters of the test to accommodate the original code, or develop a dedicated benchmarking and test suite to determine the best tolerance, I'm not sure what else to do in this case. 1e-5 is an assertion tolerance rather than an actual value that is represented or calculated as a result. Maybe some insight and expertise from @jondea on this? I'm not sure what else could be done.

src/cpu/aarch64/jit_uni_binary.cpp
leoken01: Done, changed to standard `sve` marker.

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
jondea: I agree with Sid, I think this needs exploring a bit before we change this in common code. It's a bit suspicious that it is only c6g and mish. You could try exploring: What is the tolerance for a standalone eltwise mish? Why is it different for `asimd` and `sve` implementations? What implementation is it calling underneath, is it ACL or JIT?

Also, how does `1e-5` compare to the default value?

leoken01: Do we know where I can find these specific runs or tests within oneDNN? Does it have parameters for tolerance within cli builds or is it change value within code and rebuild?

jondea: I'd use `benchdnn --eltwise --alg=mish` or something similar. You can add `-v99` and it will show you absolute and relative error even when it doesn't fail. To change the actual tolerance you would have to recompile benchdnn though, the tolerance is set in the eltwise benchdnn driver I think.

Sqvid: https://github.com/uxlfoundation/oneDNN/commit/70273f74c52c5ebb0cda4d4e9882d6f3d2d51948

Looks like there is precedent for problems with `Mish`. Given that, I'd be okay with replicating this code in the current test and moving on. We can then tackle it separately without blocking the rest of this. What do you think @jondea?

@leoken01 Could you see if replicating this tolerance check is sufficient for your case? (you can omit the `gpu` condition)
https://github.com/uxlfoundation/oneDNN/blob/4f915a57ce2fd600b9ca04bf5a312c365a4b1830/tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp#L2563-L2573

leoken01: This is sufficient, gpu has been omitted but in this case as you said it is fine. I tested this on the c6g instance and for 24 tests and skipped 2 tests that needed avx512 cores. Otherwise, the 24 tests pass.

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
```diff
@@ -2441,7 +2442,12 @@ TEST(test_convolution_execute, ConvBiasEltwise) {
         strm->wait();
         dst = eltwise_dst_ts.as_vec_type<float>();
         for (size_t i = 0; i < dst.size(); ++i) {
-            ASSERT_FLOAT_EQ(dst[i], param.ref_dst[i]);
+            if (param.op_kind == graph::op_kind::Mish
```
Sqvid, line 2435: Can you please add a comment above this, similar to the one below:
```cpp
// We noticed mish test has slight accuracy issue on some AArch64 CPUs
```

Sqvid:
Have kicked off a nightly run on this patch since it touches quite a bit of tests indirectly:
https://github.com/uxlfoundation/oneDNN/actions/runs/27145533389

tests/gtests/graph/unit/backend/dnnl/test_convolution.cpp
leoken01: done, comment added.

leoken01: Done, comment added for this line.

src/cpu/aarch64/acl_post_ops.hpp
```diff
@@ -17,19 +17,21 @@
 #ifndef CPU_AARCH64_ACL_POST_OPS_HPP
```
jondea: The filename and ifdef guard should be updated to match the new class name

src/cpu/aarch64/acl_post_ops.hpp
```diff
@@ -131,50 +112,56 @@ struct acl_post_ops_t {
...
+
+    status_t create_eltwise_primitive(engine_t *engine,
+            const eltwise_desc_t &eltwise_desc,
+            std::shared_ptr<primitive_t> &primitive) const {
+        auto empty_attr = dnnl_primitive_attr();
+
+        primitive_desc_iterator_t it(engine,
+                reinterpret_cast<const op_desc_t *>(&eltwise_desc), &empty_attr,
+                nullptr);
+        if (++it == it.end()) return status::unimplemented;
```
jondea: We need to iterate over `primitive_desc_iterator_t` before checking whether we got to the end. Take a look at other examples of this elsewhere in oneDNN for inspiration. But broadly speaking, we create the iterator, iterate over it one implementation at a time, taking the first one that returns success. If we get tot he end without success, then we declare that it is unimplemented.

(same for the binary version of this)

src/cpu/aarch64/acl_post_ops.hpp
leoken01: Should this be applied to other files using the new class name? Or is this just for this post ops file?

jondea: I'm not sure I understand the question, but this may help.

The causation is:
```
class name -determines-> source file name where class lives -determines-> ifdef guard name in file
```
Take a look at some other files to see the pattern.

Other files which include this source file will need to update their `#include` but they will only change their `ifdef` guard if _their_ filename changes too.

leoken01: done, guard has been updated to reflect this as well

src/cpu/aarch64/acl_post_ops.hpp
leoken01: done

TaoLv (approved):

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
```diff
@@ -411,30 +415,36 @@ status_t acl_lowp_matmul_t::execute(const exec_ctx_t &ctx) const {
...
         dst_s8_tensor.allocator()->free();
+        dst_s8_tensor_imported = false;
     }
 
+    if (dst_s8_tensor_imported) { dst_s8_tensor.allocator()->free(); }
+    if (dst_cast) { dst_cast_tensor.allocator()->free(); }
     dst_tensor.allocator()->free();
 
-    return status::success;
+    return post_ops_status;
```
Sqvid: If we get to the end we should be explicitly return `status::success` in my opinion. Errors should be handled early.

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
```diff
@@ -411,30 +415,36 @@ status_t acl_lowp_matmul_t::execute(const exec_ctx_t &ctx) const {
...
+    status_t post_ops_status
+            = pd()->fallback_post_ops.execute(ctx, src_post_ops, dst_post_ops);
 
     // free() here tells ACL it can no longer use it, it does not deallocate
     src_tensor.allocator()->free();
     wei_tensor.allocator()->free();
     if (with_bias) { bia_tensor.allocator()->free(); }
 
-    if (pd()->almc_.dst_is_s8) {
+    if (post_ops_status == status::success && pd()->almc_.dst_is_s8) {
```
Sqvid, line 426: Doing the check of `post_ops_status` here is confusing. Can we just use `CHECK()` on the `fallback_post_ops.execute()` call?

src/cpu/aarch64/acl_inner_product.cpp
```diff
@@ -86,7 +86,7 @@ status_t acl_inner_product_fwd_t::execute_forward(const exec_ctx_t &ctx) const {
     inner_product_op_->run(run_pack);
 
     void *dst = dst_tensor.buffer();
-    pd()->post_ops.execute(ctx, dst);
+    CHECK(pd()->post_ops.execute(ctx, dst));
```
Sqvid, line 89: This is a good example of the `CHECK()` pattern I was alluding to in my previous comment.

src/cpu/aarch64/acl_deconvolution.cpp
```diff
@@ -54,14 +54,14 @@ status_t acl_deconvolution_fwd_t::execute_forward(const exec_ctx_t &ctx) const {
     acl_obj.deconv.run();
 
     void *dst = acl_obj.dst_tensor.buffer();
-    pd()->post_ops.execute(ctx, dst);
+    status_t status = pd()->post_ops.execute(ctx, dst);
```
Sqvid, line 57: `CHECK()` instead, and leave the return as `status::success`

src/cpu/aarch64/acl_utils.cpp
```diff
@@ -139,6 +139,35 @@ status_t convert_to_acl_act(const post_ops_t::entry_t::eltwise_t &elt,
     return convert_to_acl_act(elt.alg, elt.alpha, elt.beta, act_info);
 }
 
+status_t try_fuse_first_acl_post_op(const post_ops_t &post_ops,
+        data_type_t dst_data_type, int post_op_start_index,
+        arm_compute::ActivationLayerInfo &act_info, int &next_post_op_index) {
+    act_info = arm_compute::ActivationLayerInfo();
+    next_post_op_index = post_op_start_index;
+
+    if (post_op_start_index >= post_ops.len()) return status::success;
```
Sqvid, line 148: Why do we return `status::success` in all paths of this function?

Sqvid (commented):
Nightly is passing. I have a few notes on the error handling.

src/cpu/aarch64/acl_post_ops.cpp
```diff
@@ -15,17 +15,98 @@
 *******************************************************************************/
 
 #include "common/float16.hpp"
-#include "cpu/aarch64/acl_gemm_convolution.hpp"
+#include "common/memory.hpp"
+#include "common/memory_desc_wrapper.hpp"
+#include "common/stream.hpp"
+#include "cpu/aarch64/post_ops_fallback.hpp"
```
jondea: This cpp file should be called `post_ops_fallback.cpp` to match the class name and the header

src/cpu/aarch64/acl_utils.cpp
leoken01: I think it's treating it so the whole ACL primitive isn't rejected, so it's treating cannot fuse as a normal outcome, not as failure. Is this something we should return individually then?

src/cpu/aarch64/acl_post_ops.cpp
leoken01: Acknowledged, will rename to match class name and header

leoken01: done

src/cpu/aarch64/acl_deconvolution.cpp
leoken01: done

src/cpu/aarch64/acl_inner_product.cpp
leoken01: acknowledged, implemented into request

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
leoken01: done, `CHECK() `used instead of `status`

src/cpu/aarch64/matmul/acl_lowp_matmul.cpp
leoken01: done, return status::success implemented.

jondea (approved):
LGTM, thanks for taking this on!

Sqvid (approved):

PR #5105, opened 2026-04-30
[rls-v3.12] build: exclude Xe3p from default DNNL_ENABLE_PRIMITIVE_GPU_ISA

echeresh:
For background see [MFDNN-15003](https://jira.devtools.intel.com/browse/MFDNN-15003).

Opening a PR to discuss and to map out potential side effects of making Xe3p support opt-in.


src/common/impl_registration.hpp
```diff
@@ -246,7 +246,8 @@
 #define REG_XE3_ISA(...)
 #endif
 
-#if BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XE3P
+// XE3P is excluded from BUILD_PRIMITIVE_GPU_ISA_ALL, opt in explicitly.
+#if BUILD_XE3P
```
rjoursler, line 250: Do we need any gating on OpenCL implementations? For the most part, I would expect those to  just work, but there are [some workarounds](https://github.com/uxlfoundation/oneDNN/blob/4f7899af5bb44a2965ee6bff18adb1ae75c41755/src/gpu/intel/pool/xe.cpp#L79-L84) for driver issues.

src/gpu/intel/gemm/jit/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ endif()
 
 # Use oneDNN names for ALL to ensure string replacement functions correctly
 set(GPUS ${DNNL_ENABLE_PRIMITIVE_GPU_ISA})
-string(REPLACE "ALL" "XELP;XEHP;XEHPG;XEHPC;XE2;XE3;XE3P" GPUS "${GPUS}")
+string(REPLACE "ALL" "XELP;XEHP;XEHPG;XEHPC;XE2;XE3" GPUS "${GPUS}")
```
rjoursler, line 24: Would it be reasonable to use a runtime guard (via environment variable) rather than compile time? This way we can still use the same build for testing purpose.

vpirogov: From my perspective runtime opt-in is the best choice. It would be challenging to design and implement it in time for v3.12 RTM.

src/common/impl_registration.hpp
vpirogov: We don't have any control on driver version at the user end, so I'd say no.

src/gpu/intel/gemm/jit/CMakeLists.txt
echeresh: What we can do before the RTM is opt-in via `ONEDNN_ENABLE_MAX_GPU_ISA` environment variable (similar to [CPU](https://uxlfoundation.github.io/oneDNN/dev_guide_cpu_dispatcher_control.html)). But to limit the scope:

1. No API, just the environment variable
2. Opt-in is only via the environment variable - no changes in `DNNL_ENABLE_PRIMITIVE_GPU_ISA` behavior

@vpirogov @rjoursler Do you see issues with any of the bullets? Are frameworks fine with environment variable control for opt-in (if they still want to enable Xe3p platforms)?

vpirogov: I would rather focus on developing proper guard consistent with the way we handle environment variables for v3.13. Build time opt-in looks adequate for v3.12.

echeresh:
make test

echeresh:
make test

dzarukin (commented):
LGTM.

echeresh:
Excluding a platform via `DNNL_ENABLE_PRIMITIVE_GPU_ISA` is not tested currently so we have issues coming from different places: missing coverage for reference kernels, testing issues, missing checks, etc. These issues have to be addressed first and this mode is covered in regular testing before using it for "preview" platforms.

Closing the PR for now, will proceed with a proper RFC to implement this support in main.

PR #5109, opened 2026-05-02
add limited postops support (eltwise_swish, binary_mul) for grouped gemm in CPU ref and benchdnn

mzhukova:
Implements [MFDNN-15013](https://jira.devtools.intel.com/browse/MFDNN-15013):
- eltwise_swish, binary_mul for `[total_M, 1]` dense tensor and `[total_M, N]` tensor 

Biggest change is to allow grouped tensor for binary_mul: for CPU ref we're making sure to validate offsets against `dst` offsets, and for benchdnn it is required to create a grouped memory instead of dense and be able to recognize `:grouped` as a new "tag"

### DIFF
--- a/doc/primitives/matmul.md
+++ b/doc/primitives/matmul.md
@@ -383,6 +383,13 @@ Setting attributes for grouped GEMM follows the regular matmul attribute API.
 Below are some examples of common use cases for MoE workloads.
 For more details on how to set attributes, refer to the @ref dev_guide_attributes page.
 
+#### Scales and Zero Points
+
+Scales and zero points for grouped GEMM follow the same API as regular matmul.
+Tensors must use dense memory descriptors.
+The data in the scales/zero points tensor must follow the same flat concatenated
+order as the grouped source/weights.
+
 Per-token source scales:
 ~~~cpp
 attr.set_scales_mask(DNNL_ARG_SRC, (1 << 0));  // Varies along M dimension
@@ -410,6 +417,42 @@ Bias per expert:
 auto bias_md = memory::desc({num_groups, N},
     memory::data_type::f32, memory::format_tag::ab);
 // Layout: standard ab layout
+~~~
+
+#### Post-ops Support
+
+Post-ops for grouped GEMM follow the same API as regular matmul.
+Eltwise and binary multiplication post-ops are supported.
+
+Binary post-op tensors accept both dense and grouped memory descriptors.
+Note that when a grouped descriptor is provided, we use the binary tensor's
+own per-group offsets for addressing, but they must describe the same grouped
+partition as the grouped destination tensor.
+
+SiLU and element-wise matrix multiplication as post-op:
+~~~cpp
+// [total_tokens, N] tensor, that could be either grouped or dense with
+// memory::format_tag::ab
+auto binary_md = /*...*/;
+
+post_ops po;
+po.append_eltwise(algorithm::eltwise_swish, 1.0f, 0.f);  // SiLU first
+po.append_binary(algorithm::binary_mul, binary_md);      // then mul
+attr.set_post_ops(po);
+~~~
+
+Note that dense tensor must follow the same flat concatenated order as the grouped dst.
+
+Binary multiply with per-row broadcast:
+~~~cpp
+// [total_tokens, 1] tensor, that is one scalar per row
+// in the same flat concatenated order as the grouped destination
+auto routing_md = memory::desc({total_tokens, 1},
+    memory::data_type::f32, memory::format_tag::ab);
+
+post_ops po;
+po.append_binary(algorithm::binary_mul, routing_md);
+attr.set_post_ops(po);
 ~~~
 
 ### Execution Hints
@@ -455,10 +498,15 @@ The following are supported:
 - Zero points attribute for source and weights tensors:
   - The masks must match the scales mask
   - Source zero points data types include `u8`, `s8`
-  - Weights zero points data types include `u8`, `s8`, `u4`, `s4‘
-- Post-ops: binary post-ops are supported (e.g., binary `mul` with a scalar
-  `f32` tensor can be used to apply a global scale factor, as needed for NVFP4
-  two-level scaling).
+  - Weights zero points data types include `u8`, `s8`, `u4`, `s4`
+- Post-ops: eltwise and binary multiplication post-ops are supported. Binary post-op
+  tensors accept both dense and grouped memory descriptors.
+  Common patterns include `eltwise_swish` for SiLU activation,
+  `binary_mul` with a `[total_tokens, N]` grouped or dense tensor,
+  `binary_mul` with a `[total_tokens, 1]` dense tensor,
+  and `[1, 1]` that could be used to apply a global scale for NVFP4.
+  For grouped binary tensors, the per-group offsets must match the grouped dst
+  partition.
 - Bias supports per-expert shape.
 - Supported on CPU and GPU engines.
 
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -174,7 +174,38 @@ status_t grouped_matmul_attr_check(
             attr->has_default_values(allowed_mask, desc.dst_desc.data_type),
             VERBOSE_UNSUPPORTED_ATTR);
 
-    // Specific checks are happening in impls for now
+    // Post-ops: only eltwise and binary_mul are supported
+    const memory_desc_wrapper dst_d(&desc.dst_desc);
+    const auto &po = attr->post_ops_;
+    for (int i = 0; i < po.len(); ++i) {
+        const auto &e = po.entry_[i];
+        VCHECK_MATMUL_UNIMPL(
+                e.is_eltwise() || e.is_binary(), VERBOSE_UNSUPPORTED_POSTOP);
+
+        if (e.is_binary()) {
+            VCHECK_MATMUL_UNIMPL(e.binary.alg == alg_kind::binary_mul,
+                    VERBOSE_UNSUPPORTED_POSTOP);
+            const memory_desc_wrapper src1_d(e.binary.src1_desc);
+
+            if (src1_d.is_grouped_desc()) {
+                // Validate grouped binary post-op descriptor:
+                // group_count and variable_dim_idx must match dst
+                const auto &bin_g = src1_d.sparse_desc().grouped_desc;
+                const auto &dst_g = dst_d.sparse_desc().grouped_desc;
+                VCHECK_MATMUL_UNIMPL(bin_g.group_count == dst_g.group_count
+                                && bin_g.variable_dim_idx
+                                        == dst_g.variable_dim_idx,
+                        VERBOSE_INCONSISTENT_MDS, "binary post-op", "dst");
+            } else if (!src1_d.format_any()) {
+                // Dense binary post-op:
+                // only support N == 1 (scalar/per-row) or ab layout
+                const dim_t bin_N = src1_d.dims()[src1_d.ndims() - 1];
+                VCHECK_MATMUL_UNIMPL(
+                        bin_N == 1 || src1_d.matches_one_of_tag(format_tag::ab),
+                        VERBOSE_UNSUPPORTED_TENSOR_LAYOUT, "binary post-op");
+            }
+        }
+    }
 
     return status::success;
 }
--- a/src/common/verbose.cpp
+++ b/src/common/verbose.cpp
@@ -809,6 +809,13 @@ std::ostream &operator<<(std::ostream &ss, const primitive_attr_t *attr) {
                                     ss << ":" << strides_str;
                             }
                             break;
+                        case format_kind::sparse:
+                            if (mdw.is_grouped_desc()) {
+                                ss << ":grouped";
+                                break;
+                            }
+                            assert(!"unsupported sparse encoding");
+                            break;
                         case format_kind::any: ss << ":any"; break;
                         default: assert(!"unsupported format_kind");
                     }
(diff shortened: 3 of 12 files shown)


mzhukova:
FYI @h-sadia

src/cpu/matmul/ref_grouped_gemm.hpp
```diff
@@ -216,18 +216,40 @@ struct ref_grouped_t : public primitive_t {
...
+                    // Only scalar [1, 1] is allowed to have format_any
+                    // (i.e. NVFP4 global scale support)
+                    // All other binary post-ops are expected to have
+                    // explicit format
+                    if (src1_d.format_any()) {
+                        VDISPATCH_MATMUL(src1_d.count_non_unit_dims(0),
+                                VERBOSE_UNSUPPORTED_POSTOP);
+                        CHECK(memory_desc_init_by_strides(
+                                e.binary.src1_desc, nullptr));
+                    }
```
dzarukin, line 231: All conditions seem to be an API restrictions, maybe move to a common matmul_desc_create section to let GPU impl save on some code?

src/common/verbose.cpp
```diff
@@ -809,6 +809,15 @@ std::ostream &operator<<(std::ostream &ss, const primitive_attr_t *attr) {
                                     ss << ":" << strides_str;
                             }
                             break;
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+                        case format_kind::sparse:
+                            if (mdw.encoding() == sparse_encoding::grouped) {
```
dzarukin: Minor: I think it would really help to have mdw.is_grouped() written like this:
```
bool is_grouped() const {
#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
    return ...
#else
    return false;
#endif
}
```
to have less macro leakages into the codebase.

doc/primitives/matmul.md
```diff
@@ -412,6 +419,41 @@ auto bias_md = memory::desc({num_groups, N},
...
+Eltwise and binary post-ops are supported.
+
+Binary post-op tensors accept both dense and grouped memory descriptors.
+Note that when a grouped descriptor is provided, we use the binary tensor's
+own per-group offsets for addressing, but they must describe the same grouped
+partition as the grouped destination tensor.
+
+SiLU and element-wise matrix multiplication as post-op:
+~~~cpp
+// [total_tokens, N] tensor, that could be either grouped or dense with ab
```
dzarukin:
```suggestion
// [total_tokens, N] tensor, that could be either grouped or dense with memory::format_tag::ab
```

src/cpu/matmul/ref_grouped_gemm.cpp
```diff
@@ -286,14 +287,58 @@ status_t ref_grouped_t::execute(const exec_ctx_t &ctx) const {
...
+                            }
+                            row_base = bin_offset_start;
+                        }
+                        const dim_t bin_M = bin_d.dims()[0];
+                        const dim_t bin_N = bin_d.dims()[1];
+                        const dim_t eff_m = bin_M > 1 ? (row_base + m) : 0;
+                        const dim_t eff_n = bin_N > 1 ? n : 0;
+                        const float val
+                                = io::load_float_value(bin_d.data_type(),
+                                        bin_data, eff_m * bin_N + eff_n);
```
dzarukin, line 337: It seems this assumes `format_tag::ab` for dense which wouldn't be true if `N != 1`. You might want to add a check for that N == 1 condition for dense case to be safe (seems like to be one more API check).

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -656,6 +669,21 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
+        for (int idx = 0; idx < po.len(); ++idx) {
+            const auto &e = po.entry[idx];
+            if (e.is_binary_kind() && e.binary.alg != dnnl_binary_mul) {
+                res->state = SKIPPED;
+                res->reason = skip_reason::case_not_supported;
+                return;
+            }
+        }
+    }
+#endif
```
dzarukin: If you move unimpl conditions to API level, I suggest to drop this as API conditions shouldn't be triggered by testing cases.

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -497,10 +498,22 @@ static void fill_dense_fp_values(data_kind_t kind, const prb_t *prb,
...
+
+// Check if exec_arg corresponds to a grouped binary post-op tensor
+// DNNL_ARG_ATTR_MULTIPLE_POST_OP(idx) = BASE * (idx + 1),
+// so checking idx that equals exec_arg / BASE - 1
+static bool is_grouped_binary_po_arg(
+        int exec_arg, const attr_t::post_ops_t &po) {
+    const int po_idx = exec_arg / DNNL_ARG_ATTR_MULTIPLE_POST_OP_BASE - 1;
+    return po_idx >= 0 && po_idx < po.len() && po.entry[po_idx].is_binary_kind()
+            && po.entry[po_idx].binary.grouped;
+}
```
dzarukin: Minor: maybe keep it as a lambda closer the calling spot?

tests/benchdnn/dnn_types.cpp
```diff
@@ -1351,6 +1366,14 @@ int attr_args_t::prepare_post_ops_mds(const attr_t &attr, int ndims,
             auto rhs_tensor_desc = dnn_mem_t::init_md(ndims, rhs_tensor_dims,
                     po_rhs_tensor_entry.dt, po_rhs_tensor_entry.tag,
                     po_rhs_tensor_entry.strides);
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
+            if (e.binary.grouped && grouped_var_dim_idx >= 0
```
dzarukin: It feels like `e.binary.grouped` condition should be enough to get inside `if`, then query values from sparse_options and pass them to `init_grouped_md`. If they are somewhat incorrect, that call should handle that. Right now if something goes wrong, we either don't get in and have no idea what happened when we crash later at the stack, or we returned an error somewhere earlier and then these checks do nothing...

src/cpu/matmul/ref_grouped_gemm.hpp
mzhukova: make sense, done in https://github.com/uxlfoundation/oneDNN/pull/5109/changes/f4ec7430d951d8a450782548129d7d88b188b050 !

src/common/verbose.cpp
mzhukova: yes, I should have used this as we have this already, fixed now; thanks!

src/cpu/matmul/ref_grouped_gemm.cpp
mzhukova: yes, I've added check; thanks for catching this!

tests/benchdnn/matmul/matmul.cpp
mzhukova: only used once, so moved to where it is being checked

tests/benchdnn/dnn_types.cpp
mzhukova: make sense, further library call would ensure dim_idx and such are correct, so remove those

src/common/verbose.cpp
```diff
@@ -809,6 +809,13 @@ std::ostream &operator<<(std::ostream &ss, const primitive_attr_t *attr) {
                                     ss << ":" << strides_str;
                             }
                             break;
+                        case format_kind::sparse:
+                            if (mdw.is_grouped_desc()) {
+                                ss << ":grouped";
```
dzarukin, line 814: It may happen verbose_convertor needs a touch up to generate benchdnn lines with this new entry... Could you, please, check?

tests/benchdnn/utils/parser.cpp
```diff
@@ -390,7 +390,15 @@ attr_t::post_ops_t str2attr_post_ops(const std::string &s) {
...
 
                 // parse tag input - processed for both src1/2 tensors.
+                //
+                // WARNING: "grouped" in the tag position means post-op tensor
+                // follows dst layout (e.g., offsets, layout, strides),
+                // in this case tag is left as default (any)
                 const auto tag_str = get_substr(s, src_subpos, delim);
+                if (tag_str == "grouped") {
+                    e.binary.grouped = true;
+                    return;
```
dzarukin: I think it would be a little bit more safer to keep the rest code in `else` branch over `return`ing here for actual tag processing. Any further extension would end up with unpleasant surprises if not checked carefully for this `return`.

tests/benchdnn/dnn_types.cpp
```diff
@@ -965,6 +965,9 @@ std::ostream &operator<<(std::ostream &s, const attr_t::post_ops_t &post_ops) {
             if (e.binary.tag != tag::any) s << src_delim << e.binary.tag;
             if (!e.binary.strides.empty())
                 s << src_delim << dims2str(e.binary.strides);
+#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
```
dzarukin: Minor: I think it's safe to drop this macro guard.

dzarukin (approved):

src/common/verbose.cpp
mzhukova: thanks for the reminder! updated now

tests/benchdnn/utils/parser.cpp
mzhukova: sure, updated

tests/benchdnn/dnn_types.cpp
mzhukova: that makes sense, updated

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

doc/primitives/matmul.md
```diff
@@ -412,6 +419,42 @@ auto bias_md = memory::desc({num_groups, N},
 // Layout: standard ab layout
 ~~~
 
+#### Post-ops Support
+
+Post-ops for grouped GEMM follow the same API as regular matmul.
+Eltwise and binary post-ops are supported.
```
avmanerikar:
```suggestion
Eltwise and binary post-ops (for binary multiplication only) are supported.
```
It will be helpful to specify the supported binary operation

doc/primitives/matmul.md
```diff
@@ -455,10 +498,15 @@ The following are supported:
 - Zero points attribute for source and weights tensors:
   - The masks must match the scales mask
   - Source zero points data types include `u8`, `s8`
-  - Weights zero points data types include `u8`, `s8`, `u4`, `s4‘
-- Post-ops: binary post-ops are supported (e.g., binary `mul` with a scalar
-  `f32` tensor can be used to apply a global scale factor, as needed for NVFP4
-  two-level scaling).
+  - Weights zero points data types include `u8`, `s8`, `u4`, `s4`
+- Post-ops: eltwise and binary post-ops are supported. Binary post-op
```
avmanerikar: Ditto

src/common/matmul.cpp
```diff
@@ -174,7 +174,38 @@ status_t grouped_matmul_attr_check(
...
+            VCHECK_MATMUL_UNIMPL(e.binary.alg == alg_kind::binary_mul,
+                    VERBOSE_UNSUPPORTED_POSTOP);
+            const memory_desc_wrapper src1_d(e.binary.src1_desc);
+
+            if (src1_d.is_grouped_desc()) {
+                // Validate grouped binary post-op descriptor:
+                // group_count and variable_dim_idx must match dst
+                const auto &bin_g = src1_d.sparse_desc().grouped_desc;
+                const auto &dst_g = dst_d.sparse_desc().grouped_desc;
+                VCHECK_MATMUL_UNIMPL(bin_g.group_count == dst_g.group_count
```
avmanerikar, line 195: (nit): perhaps use ` VERBOSE_INCONSISTENT_DIM`?

src/common/matmul.cpp
```diff
@@ -174,7 +174,38 @@ status_t grouped_matmul_attr_check(
...
+                                && bin_g.variable_dim_idx
+                                        == dst_g.variable_dim_idx,
+                        VERBOSE_UNSUPPORTED_POSTOP);
+            } else if (!src1_d.format_any()) {
+                // Dense binary post-op:
+                // only support N == 1 (scalar/per-row) or ab layout
+                const dim_t bin_N = src1_d.dims()[src1_d.ndims() - 1];
+                VCHECK_MATMUL_UNIMPL(
+                        bin_N == 1 || src1_d.matches_one_of_tag(format_tag::ab),
```
avmanerikar, line 204: This can be split into separate messages for ` bin_N == 1` and `src1_d.matches_one_of_tag(format_tag::ab)`

avmanerikar (approved):

doc/primitives/matmul.md
mzhukova: good catch, updated!

src/common/matmul.cpp
mzhukova: that's not really dims, but I've updated to `VERBOSE_INCONSISTENT_MDS`

src/common/matmul.cpp
mzhukova: both checks are for the same purpose, so I've updated to `VERBOSE_UNSUPPORTED_TENSOR_LAYOUT` to make it clearer

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

src/common/matmul.cpp
avmanerikar: That works too. Thanks!

PR #5110, opened 2026-05-02
build: graph: enable msvc c4244 and fix warnings

TaoLv:
- Follow-up work of #5069 
- Enable the MSVC C4244 flag in graph component. See src/graph/CMakeLists.txt.
- Fix all the type conversion warnings by C4244 in graph component.

### DIFF
--- a/src/graph/backend/dnnl/common.hpp
+++ b/src/graph/backend/dnnl/common.hpp
@@ -99,7 +99,7 @@ std::vector<int64_t> get_permutation(int ndims, const std::string &from_format,
 
 std::vector<int64_t> get_last_two_dims_permutation(int ndims);
 
-memory::desc transpose(const memory::desc &adesc, dim dim0, dim dim1);
+memory::desc transpose(const memory::desc &adesc, int dim0, int dim1);
 
 memory::desc to_grouped(const memory::desc &adesc, dim groups);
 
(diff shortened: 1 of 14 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

src/graph/backend/dnnl/common.hpp
```diff
@@ -99,7 +99,7 @@ std::vector<int64_t> get_permutation(int ndims, const std::string &from_format,
 
 std::vector<int64_t> get_last_two_dims_permutation(int ndims);
 
-memory::desc transpose(const memory::desc &adesc, dim dim0, dim dim1);
+memory::desc transpose(const memory::desc &adesc, int dim0, int dim1);
```
ElaineBao, line 102: why changing from dim to int?

TaoLv: The dim0 and dim1 will be passed to memory::desc::permute_axes() API which accepts std::vector<int>. Firstly, this change aligns with the data type requested by that API. Otherwise we will need cast from dim to int before passing them to the API.
Secondly, I searched in our code base, this transpose() function is called with immediate values (eg. transpose(md, 0, 1) or transpose(md, 1, 0)). So this change will not cause change to the caller sides.

rongzha1 (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

ElaineBao (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #5115, opened 2026-05-04
x64: conv: fixed reduced lowering copy routine for VNNI cases

### LINKED ISSUE #5080: brg_conv_fwd: incorrect output at width-padded boundary columns for s8 NHWC convolution on AVX-512 VNNI
## Summary

`brg_conv_fwd:avx512_core_vnni` produces incorrect output for `s8` `NHWC` forward convolution at output columns whose receptive field reads into the explicit width-padding region. The errors appear only at the boundary columns; all interior columns match a hand-computed zero-padded reference exactly. Selecting any non-`brg_conv_fwd` implementation on the same machine, with the same primitive descriptor and inputs, produces 0 mismatches.

## Environment

- oneDNN: **v3.8.1** (commit `be6dfdebf25976211c98deb08dd42fa7186ac7b4`)
- CPU: Intel Xeon Platinum 8368 (Ice Lake-SP), AVX-512 + VNNI, no AMX
- ISA reported by oneDNN: `Intel AVX-512 with Intel DL Boost`
- Runtime: `sequential`, `nthr:1`
- Compiler: g++ 11.4.0, `-O2 -std=c++17`

## Reproducer

Self-contained, ~90 lines, no third-party deps beyond oneDNN. Runs in <1s.

Build:

```bash
g++ -O2 -std=c++17 -I$ONEDNN/include repro.cpp \
    -L$ONEDNN/lib -Wl,-rpath,$ONEDNN/lib -ldnnl -o repro
```

Two modes:

| Run | Dispatched impl | Mismatches | Exit |
|---|---|---|---|
| `./repro` | `brg_conv_fwd:avx512_core_vnni` | **48 / 1024** | 1 |
| `./repro --skip-brg` | `jit_int8:avx512_core_vnni` | **0 / 1024** | 0 |

The only difference between the two runs is a sequence of `pd.next_impl()` calls advancing past `brg_conv_fwd` on the same primitive descriptor. Same problem definition, same data, same reference math.

<details>
<summary><code>repro.cpp</code></summary>

```cpp
// repro.cpp — build: g++ -O2 -std=c++17 -I$ONEDNN/include repro.cpp \
//                   -L$ONEDNN/lib -Wl,-rpath,$ONEDNN/lib -ldnnl -o repro
// Default run: triggers brg_conv_fwd:avx512_core_vnni and shows wrong s32
// outputs at the leftmost / rightmost output columns when explicit width
// padding is used on a small-IC NHWC s8 forward convolution.
// Pass --skip-brg to advance past brg_conv_fwd via next_impl(); the same
// problem then disappears (control: confirms the reference is correct and
// isolates the bug to the brgemm convolution path).
#include <cstdint>
#include <cstdio>
#include <cstring>
#include <vector>
#include "oneapi/dnnl/dnnl.hpp"
int main(int argc, char** argv) {
    using namespace dnnl;
    bool skip_brg = (argc > 1 && std::strcmp(argv[1], "--skip-brg") == 0);
    using tag = memory::format_tag;
    using dt  = memory::data_type;
    // mb=1 ic=1 oc=1 ih=32 oh=16 kh=2 sh=2 pad_h=0
    //              iw=64 ow=64 kw=4 sw=1 pad_w_l=1 pad_w_r=2
    const int N = 1, IC = 1, OC = 1;
    const int IH = 32, IW = 64;
    const int KH = 2,  KW = 4;
    const int SH = 2,  SW = 1;
    const int padT = 0, padB = 0, padL = 1, padR = 2;
    const int OH = (IH + padT + padB - KH) / SH + 1; // 16
    const int OW = (IW + padL + padR - KW) / SW + 1; // 64
    engine eng(engine::kind::cpu, 0);
    stream s(eng);
    std::vector<int8_t> src(N * IC * IH * IW);
    for (size_t i = 0; i < src.size(); ++i)
        src[i] = static_cast<int8_t>(int(i % 31) - 15); // -15..15, never 0
    std::vector<int8_t> wei(OC * IC * KH * KW, 1);
    auto user_src_md = memory::desc({N, IC, IH, IW}, dt::s8,  tag::nhwc);
    auto user_wei_md = memory::desc({OC, IC, KH, KW}, dt::s8, tag::oihw);
    auto any_src_md  = memory::desc({N, IC, IH, IW}, dt::s8,  tag::any);
    auto any_wei_md  = memory::desc({OC, IC, KH, KW}, dt::s8, tag::any);
    auto dst_md      = memory::desc({N, OC, OH, OW}, dt::s32, tag::nhwc);
    primitive_attr attr;
    attr.set_scratchpad_mode(scratchpad_mode::user);
    convolution_forward::primitive_desc pd(
        eng, prop_kind::forward_inference, algorithm::convolution_direct,
        any_src_md, any_wei_md, dst_md,
        {SH, SW}, {padT, padL}, {padB, padR}, attr);
    if (skip_brg) {
        for (;;) {
            const char* s = pd.impl_info_str();
            if (s == nullptr || std::strstr(s, "brg_conv_fwd") == nullptr) break;
            if (!pd.next_impl()) { std::printf("no non-brg impl available\n"); break; }
        }
    }
    std::printf("dispatched impl: %s\n", pd.impl_info_str());
    auto user_src_mem = memory(user_src_md, eng, src.data());
    auto user_wei_mem = memory(user_wei_md, eng, wei.data());
    auto src_mem      = memory(pd.src_desc(),         eng);
    auto wei_mem      = memory(pd.weights_desc(),     eng);
    auto dst_mem      = memory(pd.dst_desc(),         eng);
    auto sp_mem       = memory(pd.scratchpad_desc(),  eng);
    reorder(user_src_mem, src_mem).execute(s, user_src_mem, src_mem);
    reorder(user_wei_mem, wei_mem).execute(s, user_wei_mem, wei_mem);
    s.wait();
    convolution_forward(pd).execute(s, {
        {DNNL_ARG_SRC, src_mem}, {DNNL_ARG_WEIGHTS, wei_mem},
        {DNNL_ARG_DST, dst_mem}, {DNNL_ARG_SCRATCHPAD, sp_mem},
    });
    s.wait();
    auto src_at = [&](int n, int c, int h, int w) -> int32_t {
        if (h < 0 || h >= IH || w < 0 || w >= IW) return 0;
        return src[((n * IH + h) * IW + w) * IC + c];
    };
    std::vector<int32_t> ref(size_t(N) * OC * OH * OW, 0);
    for (int n = 0; n < N; ++n)
      for (int oc = 0; oc < OC; ++oc)
        for (int oh = 0; oh < OH; ++oh)
          for (int ow = 0; ow < OW; ++ow) {
              int32_t acc = 0;
              for (int ic = 0; ic < IC; ++ic)
                for (int kh = 0; kh < KH; ++kh)
                  for (int kw = 0; kw < KW; ++kw) {
                      int ih = oh * SH + kh - padT;
                      int iw = ow * SW + kw - padL;
                      acc += int32_t(src_at(n, ic, ih, iw)) *
                             int32_t(wei[((oc * IC + ic) * KH + kh) * KW + kw]);
                  }
              ref[((n * OH + oh) * OW + ow) * OC + oc] = acc;
          }
    auto* dnn_dst = static_cast<int32_t*>(dst_mem.get_data_handle());
    int diffs = 0;
    for (size_t i = 0; i < ref.size(); ++i)
        if (dnn_dst[i] != ref[i]) {
            if (diffs < 16) {
                int n  =  int(i) / (OH * OW * OC);
                int oh = (int(i) / (OW * OC)) % OH;
                int ow = (int(i) / OC) % OW;
                int oc =  int(i) % OC;
                std::printf("mismatch n=%d oc=%d oh=%d ow=%d  dnn=%d  ref=%d\n",
                            n, oc, oh, ow, dnn_dst[i], ref[i]);
            }
            ++diffs;
        }
    std::printf("total diffs: %d / %zu\n", diffs, ref.size());
    return diffs == 0 ? 0 : 1;
}
```

</details>

## Observed

Problem geometry: `mb1_ic1oc1_ih32oh16kh2sh2dh0ph0_iw64ow64kw4sw1dw0pw1`, source `s8:nhwc`, weights `s8:oihw`, dst `s32:nhwc`, no post-ops, scratchpad mode `user`.

oneDNN verbose for the failing run:

```
onednn_verbose,v1,info,oneDNN v3.8.1 (commit be6dfdebf25976211c98deb08dd42fa7186ac7b4)
onednn_verbose,v1,info,cpu,runtime:sequential,nthr:1
onednn_verbose,v1,info,cpu,isa:Intel AVX-512 with Intel DL Boost
onednn_verbose,v1,primitive,exec,cpu,convolution,brg_conv_fwd:avx512_core_vnni,
forward_inference,src:s8:a:blocked:acdb::f0 wei:s8:ap:blocked:Adcb16a::f1:s8m1
bia:undef::undef::: dst:s32::blocked:acdb::f0,attr-scratchpad:user,
alg:convolution_direct,
mb1_ic1oc1_ih32oh16kh2sh2dh0ph0_iw64ow64kw4sw1dw0pw1
```

All 48 mismatches lie at exactly three output columns — `ow ∈ {0, 62, 63}` — across all 16 output rows:

```
48 = 16 OH rows × 3 boundary columns
ow=0       <- receptive field reaches into left  pad (pad_w_l=1)
ow=62, 63  <- receptive field reaches into right pad (pad_w_r=2)
```

Every interior column (`ow=1..61`) is bit-exact. The brg path appears to read non-zero stale data where the input should be zero-padded; for example, with the first row of `src` and unit weights, output `[oh=0, ow=0]` should compute `0 + src[0] + src[1] + src[2] = -15 + (-14) + (-13) + (-12) = -54` (well, with the deterministic schedule above, the reference printed `-78`). The brg run prints `178` instead — a difference consistent with three padding lanes being sourced from an adjacent NHWC region rather than zero.

**Killer line:** Calling `pd.next_impl()` once on the same primitive descriptor moves dispatch from `brg_conv_fwd:avx512_core_vnni` to `jit_int8:avx512_core_vnni`, and the same reproducer prints `total diffs: 0`. No other change. See `--skip-brg` in `repro.cpp`.

## Workarounds (both validated end-to-end on real workloads)

1. **Skip `brg_conv_fwd` via `next_impl()`** — at primitive-descriptor build time, advance past any implementation whose `impl_info_str()` contains `brg_conv_fwd` whenever the primitive has any non-zero spatial padding. Cleanest, no API misuse, no extra memory.
2. **Pre-zero-pad the NHWC source** — allocate a wider `(IH+padT+padB) × (IW+padL+padR)` zero buffer, copy `src` into the interior, build the primitive with zero `pad_l`/`pad_r`. Works but wastes memory and bandwidth for small-IC convs.

We currently ship workaround #1.

## Hypothesis (best-guess from the user side, please correct me)

The symptom strongly points to the `brg_conv_fwd` "exec_trans" / `inp_buffer` path not zero-initializing the slack region of the per-tile transformed input buffer when reading boundary tiles whose source rows extend into spatial padding. The buggy values appear to come from positions adjacent to the pad lanes in the original source layout, not from random/uninitialized memory. Possibly related history:

- [#1354](https://github.com/oneapi-src/oneDNN/issues/1354) — earlier `brg_conv` boundary issue with explicit padding.
- [#3435](https://github.com/oneapi-src/oneDNN/issues/3435) — discusses removal of an `inp_buffer` zeroing step in `exec_trans` for performance reasons; the symptom we're seeing is what one would expect if that zeroing is now skipped on a code path it shouldn't be.

Happy to provide additional traces (`ONEDNN_VERBOSE=dispatch`, larger reproducers, alternate ISA caps) if useful.

vpirogov:
Fix for correctness issue in s8s8 VNNI convolution implementation. Previously the copy routine did not include -128 shift in the padded area when compensation is required producing incorrect results.

Fixes #5080.

### DIFF
--- a/src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
+++ b/src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
@@ -757,7 +757,14 @@ void jit_avx512_core_amx_copy_to_pbuffer_t::copy_row_reduced_lowering() {
     // number of zero-padded columns before src buffer to copy
     mov(reg_rov, ptr[param1 + GET_OFF(back_overflow)]);
 
-    vpxord(zmm_zero, zmm_zero, zmm_zero);
+    if (jcp.signed_input) {
+        // Apply -128 shift to zero padded area when using VPDPBUSD instruction
+        // to work with s8 inputs.
+        assert(jcp.src_dt == data_type::s8);
+        mov(reg_tmp, -128);
+        vpbroadcastb(zmm_zero, reg_tmp.cvt8());
+    } else
+        vpxord(zmm_zero, zmm_zero, zmm_zero);
 
     { // Handle Left Overflow
         Label label_lov, label_lov_skip;
--- a/src/cpu/x64/jit_brgemm_conv.cpp
+++ b/src/cpu/x64/jit_brgemm_conv.cpp
@@ -959,6 +959,9 @@ status_t brgemm_convolution_fwd_t<isa>::init(engine_t *engine) {
         ajcp.oc_block = jcp.oc_block;
         ajcp.ic_block = 16;
         ajcp.nb_oc = jcp.nb_oc;
+        // let copy routine know that zero padding needs special handling
+        // for legacy ISA without native s8s8 instructions
+        ajcp.signed_input = jcp.s8s8_compensation_required;
 
         CHECK(safe_ptr_assign(copy_to_relo_pbuffer_,
                 new jit_avx512_core_amx_copy_to_pbuffer_t(ajcp)));
--- a/tests/benchdnn/inputs/conv/harness_conv_regression_general
+++ b/tests/benchdnn/inputs/conv/harness_conv_regression_general
@@ -437,3 +437,8 @@ mb64_ic3oc192_ih224oh14kh16sh16dh0ph0_iw224ow14kw16sw16dw0pw0n"rpad0_bf16_accura
 # Case with vpad != 0 in brgconv:sve_128 which triggered tmp register reuse issue
 --reset
 --dt=f32 --dir=FWD_B --impl=brgconv mb1ic342iw7oc10000ow7kw7pw3
+
+# accuracy issue in reduced lowering approach for s8s8 case on VNNI code path
+--reset
+--dt=s8:s8:s32 --dir=FWD_I
+mb1_ic1oc1_ih32oh16kh2sh2dh0ph0_iw64ow64kw4sw1dw0pw1


dzarukin (approved):

vpirogov:
make test
set test_scope=NIGHTLY
disable test_device_gpu
enable benchdnn_conv

src/cpu/x64/jit_brgemm_conv.cpp
```diff
@@ -959,6 +959,7 @@ status_t brgemm_convolution_fwd_t<isa>::init(engine_t *engine) {
         ajcp.oc_block = jcp.oc_block;
         ajcp.ic_block = 16;
         ajcp.nb_oc = jcp.nb_oc;
+        ajcp.signed_input = jcp.s8s8_compensation_required;
```
densamoilov, line 964: (nit) maybe add a clarification that it's a legacy code path for the platforms that don't support s8s8?

densamoilov (approved):

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
georgen117, line 715: Minor: may want to consider adding an assert here:
```
// signed_input implies s8 input dtype; the -128 fill of zmm_zero must not
// reach the float-type padding path
assert(!jcp.signed_input || (!is_f32 && !isxf16));
```
This is already true, but the assert makes it clear (documents the fact via code) that the signed_input is exclusive to the int8 type.

georgen117 (approved):

src/cpu/x64/jit_avx512_core_amx_conv_kernel.cpp
vpirogov: Thanks! We can do even better considering that shift is specific to `s8`.

PR #2988, opened 2025-03-31
gpu: intel: sycl: add support for kernel compilation

spalicki:
Enabling experimental support for SYCL online kernel compilation. Feature is guarded by `ONEDNN_EXPERIMENTAL_SYCL_KERNEL_COMPILER`.

PR #5116, opened 2026-05-04
build, xpu: fix SYCL kernel compiler issues

echeresh:
PR fixes build issues:

- To query feature support we need to use `ext_oneapi_can_build()` instead of `ext_oneapi_can_compile()`
    - `can_compile()` returns `false` in my tests + we use `syclex::build()` so the can-build check looks like the correct approach
- `get_device_uuid()` is used by recently added L0 runtime support - removed the guard
- `create_ocl_engine()` is unused with `DNNL_EXPERIMENTAL_SYCL_KERNEL_COMPILER` - added the guard

### DIFF
--- a/cmake/SYCL.cmake
+++ b/cmake/SYCL.cmake
@@ -128,20 +128,18 @@ endif()
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsycl")
 
 if(DNNL_EXPERIMENTAL_SYCL_KERNEL_COMPILER)
-    include(CheckCXXSourceRuns)
+    include(CheckCXXSourceCompiles)
     set(CHECK_SYCL_KERNEL_COMPILER_SOURCE
     "
         #include <sycl/sycl.hpp>
         namespace syclex = sycl::ext::oneapi::experimental;
         int main() {
-            for (auto platform : sycl::platform::get_platforms())
-                for (auto d : platform.get_devices())
-                    if (!d.ext_oneapi_can_build(syclex::source_language::opencl))
-                        return 1;
+            sycl::device d;
+            (void)d.ext_oneapi_can_build(syclex::source_language::opencl);
             return 0;
         }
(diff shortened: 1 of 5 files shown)


echeresh:
make test

atkassen (approved):

kealan-barbieri (approved):

cmake/SYCL.cmake
vpirogov, line 149: This build time check actually checks runtime behavior. I would suggest dropping it altogether.

vpirogov:
Thanks for taking care of these issues!

dzarukin (approved):

cmake/SYCL.cmake
echeresh: I agree, this check is not reliable as oneDNN can be compiled on a host without SYCL devices.

It was added as a safeguard: https://github.com/uxlfoundation/oneDNN/pull/2988#discussion_r2116299839

I suggest to change it to a compilation check - it can still be useful to catch outdated or misconfigured SYCL compilers.

echeresh:
make test

cmake/SYCL.cmake
vpirogov: Even as a compile check it should not be a fatal error. Build configuration does not require `ocloc` or whatever other dependencies SYCL online compiler may have. Maybe leave it as a warning for the sake of the first adopters?

echeresh: Hmm, `ocloc` shouldn't be a compile time dependency. This is a preflight check - we compile a simple test to confirm that the compiler is able to compile `d.ext_oneapi_can_build(syclex::source_language::opencl)`:

https://github.com/uxlfoundation/oneDNN/blob/a980383066d466fd4ec039d2e38ff153bd1eeab7/cmake/SYCL.cmake#L130-L144

vpirogov: Oh, you are right.

PR #5118, opened 2026-05-05
benchdnn: improve perf time reporting for mode=F on GPU

dzarukin:
[MFDNN-13816](https://jira.devtools.intel.com/browse/MFDNN-13816)

This PR seems to settle a part of the problem when it comes to reporting BW perf numbers more than the GPU HW can achieve. It comes from several factors:
* GPU driver tends to compress the similar data that it meets. Thus, a single output value will be stored times faster that it should be. This effect is referred as "data compression". It affects both inputs and outputs. Output suffers more as during reduction in Conv or MM, infinities, NaNs, zeros can be obtained, esp. due to randomness of data filling.
* Cache is flushed after queue synchronization. After cache is flushed, the next run doesn't need to clean before writing in it, saving a lot of bandwidth and leading to faster times. Additionally, there's some GPU driver heuristic when it decided to flush the cache without clear external reason, in that case such submission would be way faster than others.

This PR rather targets the second scenario issues with two main changes:
* Change the way submission happens for measure_perf_aggregate - instead of a batch of 6 times (one first loop and 5 consecutive), it does a single warm-up loop which doesn't go towards time collection, and then a single submission loop with roughly same number of runs which was before. It relies on the warm-up time (for time criterion) and time budget (max_ms_per_prb) to estimate the number of runs.
* Filters the times it gets. The first half of collection will be dropped. This is motivated by that empty cache effect described earlier, plus by the time of second half the frequency should be stabilized. Secondly, the rest of collection will be inspected for spikes between times (4% diff) and number of times for large collections (usually, small problems are affected) which expects to find at least 1% of all collection to have the same time value (granularity of profiling allows to rely on this).

There are some other minor fixes - increase the amount of memory for cold-cache on GPU for better stability; adjusted iobytes to report more realistic amount of bytes processed by the kernel.

The part about data compression will be handled separately as it requires some benchdnn infrastructure change as random filling must become semi-random - certain tensors can use random values, otherwise, they got hit by compression for the reasons explained above.

### DIFF
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -1486,7 +1486,12 @@ int check_total_size(res_t *res, dnnl_primitive_t prim_ref) {
 
 void add_md_size(const_dnnl_memory_desc_t md,
         check_mem_size_args_t &check_mem_size_args) {
-    const auto mem_size = dnnl_memory_desc_get_size(md);
+    size_t mem_size = SIZE_MAX;
+    if (check_mem_size_args.use_logical_size) {
+        mem_size = get_logical_size(md);
+    } else {
+        mem_size = dnnl_memory_desc_get_size(md);
+    }
     // Runtime mem size is not defined.
     if (mem_size == 0 || mem_size == DNNL_RUNTIME_SIZE_VAL) return;
 
@@ -1671,15 +1676,18 @@ int collect_mem_size(check_mem_size_args_t &mem_size_args,
 }
 
 int get_memory_footprint(const_dnnl_primitive_desc_t const_pd, res_t *res) {
-    check_mem_size_args_t check_mem_in_size_args(
-            const_pd, /* want_input = */ true, DIR_UNDEF);
+    check_mem_size_args_t check_mem_in_size_args(const_pd,
+            /* want_input = */ true, DIR_UNDEF,
+            /* use_logical_size */ true);
     get_memory_bytes(check_mem_in_size_args); // Get input bytes.
-    check_mem_size_args_t check_mem_out_size_args(
-            const_pd, /* want_input = */ false, DIR_UNDEF);
+    check_mem_size_args_t check_mem_out_size_args(const_pd,
+            /* want_input = */ false, DIR_UNDEF,
+            /* use_logical_size */ true);
     get_memory_bytes(check_mem_out_size_args); // Get output bytes.
 
-    // Sum post-ops include dst bytes as an input. Not included in get_memory_bytes
-    // since it would cause `collect_mem_size` to double-count dst bytes.
+    // Sum post-ops include dst bytes as an input. Not included in
+    // `get_memory_bytes` since it would cause `collect_mem_size` to
+    // double-count dst bytes.
     auto const_attr_po = query_post_ops(const_pd);
     auto po_len = dnnl_post_ops_len(const_attr_po);
     for (int idx = 0; idx < po_len; ++idx) {
--- a/tests/benchdnn/dnnl_common.cpp
+++ b/tests/benchdnn/dnnl_common.cpp
@@ -667,6 +667,37 @@ void finalize() {
 #endif
 }
 
+int update_timer_with_profiling_info(timer::timer_t &t, bool use_profiling,
+        const std::vector<stream_t> &v_stream, int execute_count) {
+    if (!use_profiling) {
+        t.stamp(execute_count * num_streams);
+        return OK;
+    }
+
+    std::vector<std::vector<uint64_t>> v_nsecs(num_streams);
+    std::vector<std::vector<uint64_t>> v_cycles(num_streams);
+    for (size_t j = 0; j < v_stream.size(); j++) {
+        SAFE(get_gpu_profiling_info(
+                     v_stream[j], v_nsecs[j], v_cycles[j], execute_count),
+                CRIT);
+        reset_gpu_profiling(v_stream[j]);
+
+        // Profiling should have information to report, otherwise, stop.
+        if (v_nsecs[j].empty()) {
+            BENCHDNN_PRINT(0, "%s\n",
+                    "WARNING: no counters were found during profiling.");
+            return FAIL;
+        }
+    }
+
+    for_(size_t j = 0; j < v_stream.size(); j++)
+    for (size_t i = 0; i < v_nsecs[j].size(); i++) {
+        t.stop(1, (int64_t)v_cycles[j][i], v_nsecs[j][i] / 1e6);
+    }
+
+    return OK;
+}
+
 inline int measure_perf_individual(timer::timer_t &t, dnnl_stream_t stream,
         perf_function_t &perf_func, std::vector<dnnl_exec_arg_t> &dnnl_args) {
     // Warm-up run.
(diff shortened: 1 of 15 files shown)


dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test perf-gpu
set primitive=gpu:gemm

dzarukin:
make test

tests/benchdnn/utils/timer.cpp
```diff
@@ -74,6 +76,8 @@ void timer_t::stop(int add_times, int64_t add_ticks, double add_ms) {
     ticks_[mode_t::avg] += d_ticks;
     ticks_[mode_t::sum] += d_ticks;
 
+    ms_vec_.push_back(d_ms);
```
echeresh: This implicitly assumes `add_times == 1`, maybe add an assertion?

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1717,6 +1728,49 @@ int get_memory_footprint(const_dnnl_primitive_desc_t const_pd, res_t *res) {
...
+        auto wei_ndims = query_md_ndims(wei_md);
+        auto with_groups = src_ndims < wei_ndims;
+        auto wei_dims = query_md_dims(wei_md);
+
+        auto strides = query_strides(const_pd);
+        double kernel_ratio = 1;
+        for (int d = 0; d < src_ndims - 2; d++) {
+            auto stride = strides[d];
+            auto kernel = wei_dims[with_groups + 2 + d];
+            if (stride > kernel) { kernel_ratio *= stride / kernel; }
```
echeresh: This is an integer division, so need a cast:
```suggestion
            if (stride > kernel) { kernel_ratio *= stride / (double)kernel; }
```

tests/benchdnn/utils/timer.cpp
```diff
@@ -92,6 +96,107 @@ void timer_t::stamp(int add_times) {
...
+
+    // After all undesired values discarded, re-compute stats.
+    ms_[mode_t::sum] = 0;
+    ms_[mode_t::min] = *ms_vec_.begin();
+    ms_[mode_t::max] = *ms_vec_.rbegin();
+    for (size_t i = 0; i < ms_vec_.size(); i++) {
+        ms_[mode_t::sum] += ms_vec_[i];
+    }
+    ms_[mode_t::avg] = ms_[mode_t::sum];
+    times_ = ms_vec_.size();
```
echeresh, line 194: Nit: `ticks_` are not updated here so maybe better to zero them out - to avoid their misuse as `ms_vec_` now is the up-to-date version and `ticks_` are outdated.

tests/benchdnn/utils/dnnl_query.cpp
```diff
@@ -36,6 +36,18 @@ dnnl_alg_kind_t query_alg_kind(const_dnnl_primitive_desc_t pd) {
     return alg_kind;
 }
 
+const dnnl_dims_t &query_strides(const_dnnl_primitive_desc_t pd) {
+    dnnl_dims_t *strides;
+    dnnl_primitive_desc_query(pd, dnnl_query_strides, 0, &strides);
```
echeresh, line 41: `DNN_SAFE_V()` just in case? and the same in `query_kernels`

tests/benchdnn/dnnl_common.cpp
```diff
@@ -689,96 +720,68 @@ inline int measure_perf_individual(timer::timer_t &t, dnnl_stream_t stream,
...
-            = fix_times_per_prb ? fix_times_per_prb : min_times_per_prb;
+    int num_submissions = 0;
+    if (fix_times_per_prb) {
+        // If user specified the number of runs, just use it.
+        num_submissions = fix_times_per_prb;
+    } else {
+        // Otherwise, use an estimate run, which is not a part of final time
+        // collection. It's used to calculate the number of submissions to run
+        // before synchronization. Done on a single stream.
+        t.reset();
```
echeresh, line 745: Is `min_times_per_prb` ignored now?

tests/benchdnn/dnnl_common.cpp
```diff
@@ -1717,6 +1728,49 @@ int get_memory_footprint(const_dnnl_primitive_desc_t const_pd, res_t *res) {
...
+        }
+        const auto src_size = dnnl_memory_desc_get_size(src_md);
+        const auto fixed_src_size
+                = static_cast<size_t>(src_size / kernel_ratio);
+        check_mem_in_size_args.total_size_device += fixed_src_size - src_size;
+    }
+
+    // When a dimensional stride is bigger than kernel window, it means there
+    // are less reads from source by stride/kernel times.
+    if (is_fwd_prop_kind(prop) && kind == dnnl_pooling) {
```
echeresh, line 1767: Can we introduce a helper to reuse for convolution and pooling?

```c++
  static void adjust_src_bytes_for_stride(const_dnnl_memory_desc_t src_md,
          const dnnl_dim_t *strides, const dnnl_dim_t *kernels,
          int spatial_ndims, check_mem_size_args_t &args) {
      double kernel_ratio = 1;
      for (int d = 0; d < spatial_ndims; d++) {
          if (strides[d] > kernels[d]) {
              kernel_ratio *= strides[d] / (double)kernels[d];
          }
      }
      const auto src_size = dnnl_memory_desc_get_size(src_md);
      const auto fixed_src_size = static_cast<size_t>(src_size / kernel_ratio);
      args.total_size_device += fixed_src_size - src_size;
  }
```

echeresh (approved):

tests/benchdnn/utils/timer.cpp
dzarukin: I replaced this with `ms_vec_.insert(ms_vec_.end(), add_times, d_ms);` after `d_ms` is divided on add_times, now it should be OK for any `add_times`. Thanks for pointing this out.

tests/benchdnn/utils/timer.cpp
dzarukin: Makes sense, I need to drop them altogether, they don't seem to be used.

tests/benchdnn/utils/dnnl_query.cpp
dzarukin: We don't check statuses for queries, hoping for best. :)

tests/benchdnn/dnnl_common.cpp
dzarukin: Indeed, thanks for spotting. Updated in both places.

tests/benchdnn/dnnl_common.cpp
dzarukin: Sure, added it as lambda, thanks.

tests/benchdnn/dnnl_common.cpp
dzarukin: Restored, thanks.

dzarukin:
make test

mzhukova (approved):

PR #5119, opened 2026-05-05
x64: matmul: add acdb/adbc dst support to brgemm matmul

tczeszun:
Issue converted to draft, as JAX team confirmed their MHA lowering approach has changed and they no longer encounter the original case. Ticket closed by requester. MFDNN-11647. Keeping this version here in case ticket reopens.

## Description

brgemm matmul cannot write to acdb or adbc dst layouts directly.
Before this patch such cases fell back to ref matmul.

This patch adds support for both layouts by computing into a scratch
buffer and transposing out with a JIT AVX-512 16×16 kernel.

Not supported yet: sum post-op, parallel K reduction.

## Performance

56-core Intel AVX10.1/AMX, f32.

**Ref** = ref matmul (only path before this change).
**abcd+reorder** = brgemm to abcd dst + separate reorder (best a framework could do manually).

### adbc dst

| M | Shape | vs Ref | vs abcd+reorder |
|--:|:------|-------:|----------------:|
| 64  | $\color{goldenrod}{\textrm{1×16×64×384 : 1×16×384×384}}$ | $\color{green}{\textrm{1609x}}$ | $\color{green}{\textrm{1.24x}}$ |
| 64  | 1×16×64×384 : 1×16×384×64   | $\color{green}{\textrm{741x}}$ | $\color{green}{\textrm{1.64x}}$ |
| 64  | 1×12×64×64  : 1×12×64×64    | $\color{green}{\textrm{145x}}$ | $\color{green}{\textrm{1.93x}}$ |
| 64  | 1×32×64×64  : 1×32×64×64    | $\color{green}{\textrm{304x}}$ | $\color{green}{\textrm{1.71x}}$ |
| 128 | 1×16×128×384 : 1×16×384×128 | $\color{green}{\textrm{1120x}}$ | $\color{gray}{\textrm{0.97x}}$ |
| 256 | 1×8×256×256  : 1×8×256×256  | $\color{green}{\textrm{963x}}$ | $\color{red}{\textrm{0.75x}}$ |

### acdb dst

| M | Shape | vs Ref | vs abcd+reorder |
|--:|:------|-------:|----------------:|
| 64  | $\color{goldenrod}{\textrm{1×16×64×384 : 1×16×384×384}}$ | $\color{green}{\textrm{857x}}$ | $\color{red}{\textrm{0.72x}}$ |
| 64  | 1×16×64×384 : 1×16×384×64   | $\color{green}{\textrm{362x}}$ | $\color{red}{\textrm{0.81x}}$ |
| 64  | 1×12×64×64  : 1×12×64×64    | $\color{green}{\textrm{67x}}$  | $\color{red}{\textrm{0.90x}}$ |
| 64  | 1×32×64×64  : 1×32×64×64    | $\color{green}{\textrm{132x}}$ | $\color{red}{\textrm{0.74x}}$ |
| 128 | 1×16×128×384 : 1×16×384×128 | $\color{green}{\textrm{803x}}$ | $\color{red}{\textrm{0.66x}}$ |
| 256 | 1×8×256×256  : 1×8×256×256  | $\color{green}{\textrm{749x}}$ | $\color{red}{\textrm{0.56x}}$ |

$\color{goldenrod}{\textrm{Highlighted}}$ = JAX MHA model shape.

### Summary

- **67–1609×** faster than ref for both layouts.
- **adbc** beats abcd+reorder by **1.24–1.93×** at M≤64 (typical MHA shapes).
- **acdb** is 0.56–0.90× vs abcd+reorder but removes the need for the framework to manage a separate reorder.

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul.cpp
@@ -514,6 +514,9 @@ status_t brgemm_matmul_t<isa>::init(engine_t *engine) {
         CHECK(sparse_decompress_kernel_->create_kernel());
     }
 
+    if (bgmmc.dst_is_acdb || bgmmc.dst_is_adbc)
+        CHECK(create_brgemm_matmul_copy_c(copy_C_kernel_, &bgmmc));
+
     return status::success;
 }
 
@@ -601,7 +604,12 @@ status_t brgemm_matmul_t<isa>::execute_body(const exec_ctx_t &ctx) const {
         int m_chunks_per_thread = div_up(M_chunks, bgmmc.nthr_m);
         int n_chunks_per_thread = div_up(N_chunks, bgmmc.nthr_n);
         int batch_per_thread = div_up(bgmmc.batch, bgmmc.nthr_b);
-        if (brgmm_ctx.is_chunks_horizontal_process_order())
+
+        if (bgmmc.dst_is_acdb)
+            nd_iterator_init(start, bt, bgmmc.nthr_b, mt, bgmmc.nthr_m, nt,
+                    bgmmc.nthr_n, mc_per_t, m_chunks_per_thread, nc_per_t,
+                    n_chunks_per_thread, b_per_t, batch_per_thread);
+        else if (brgmm_ctx.is_chunks_horizontal_process_order())
             nd_iterator_init(start, bt, bgmmc.nthr_b, mt, bgmmc.nthr_m, nt,
                     bgmmc.nthr_n, b_per_t, batch_per_thread, mc_per_t,
                     m_chunks_per_thread, nc_per_t, n_chunks_per_thread);
@@ -615,7 +623,12 @@ status_t brgemm_matmul_t<isa>::execute_body(const exec_ctx_t &ctx) const {
 
         auto advance_func = [&]() {
             ++start;
-            if (brgmm_ctx.is_chunks_horizontal_process_order())
+            if (bgmmc.dst_is_acdb)
+                nd_iterator_step(bt, bgmmc.nthr_b, mt, bgmmc.nthr_m, nt,
+                        bgmmc.nthr_n, mc_per_t, m_chunks_per_thread,
+                        nc_per_t, n_chunks_per_thread, b_per_t,
+                        batch_per_thread);
+            else if (brgmm_ctx.is_chunks_horizontal_process_order())
                 nd_iterator_step(bt, bgmmc.nthr_b, mt, bgmmc.nthr_m, nt,
                         bgmmc.nthr_n, b_per_t, batch_per_thread, mc_per_t,
                         m_chunks_per_thread, nc_per_t, n_chunks_per_thread);
@@ -708,6 +721,40 @@ status_t brgemm_matmul_t<isa>::execute_body(const exec_ctx_t &ctx) const {
     maybe_reduce_and_convert_partial_results_A(brgmm_ctx_ptr);
     maybe_reduce_partial_results_and_apply_postops(brgmm_ctx_ptr);
 
+    // Phase 2: copy scratch buffer to acdb dst.
+    if (pd()->get_brgemm_matmul_conf().dst_is_acdb) {
+        const auto &bgmmc = pd()->get_brgemm_matmul_conf();
+        const auto &brgmm_ctx = *brgmm_ctx_ptr;
+        const int M = bgmmc.M;
+        const int N = bgmmc.N;
+        const int inner_batch = bgmmc.acdb_inner_batch;
+        const int outer_batch = bgmmc.batch / inner_batch;
+        const dim_t buf_row_stride = N * bgmmc.c_dt_sz;
+        const dim_t dst_m_stride = bgmmc.C_strides[1];
+        const dim_t dst_n_stride = bgmmc.C_strides[0];
+        const int dt_sz = bgmmc.c_dt_sz;
+
+        const int N_tile = 64;
+        const int n_tiles = utils::div_up(N, N_tile);
+
+        parallel_nd(outer_batch, n_tiles, M, [&](int ob, int nt, int m) {
+            const int n_start = nt * N_tile;
+            const int n_len = nstl::min(N_tile, N - n_start);
+            const int b_start = ob * inner_batch;
+            const char *buf_base
+                    = brgmm_ctx.get_buf_D_acdb_ptr(b_start);
+            char *dst_base = brgmm_ctx.get_data_C_ptr(b_start, 0, 0);
+            jit_brgemm_matmul_copy_c_t::ctx_t copy_c_ctx;
+            copy_c_ctx.src = buf_base + m * buf_row_stride
+                    + n_start * dt_sz;
+            copy_c_ctx.dst = dst_base + m * dst_m_stride
+                    + n_start * dst_n_stride;
+            copy_c_ctx.current_M_blk = 1;
+            copy_c_ctx.current_N_blk = n_len;
+            (*copy_C_kernel_)(&copy_c_ctx);
+        });
+    }
+
     return status::success;
 }
 
@@ -742,12 +789,6 @@ void brgemm_matmul_t<isa>::compute_kernel(
     const int brg_ker_idx = pd()->get_brg_kernel_idx(
             is_bs_tail, do_init, m_ker_idx, n_ker_idx, false, prefetch);
     const auto ptr_bias = brgmm_ctx.get_bias_ptr(n);
-    auto ptr_D = brgmm_ctx.get_data_C_ptr(
-            b_idx, brgmm_ctx.get_M_idx(m_blk_idx, true), n);
-    auto ptr_C = (bgmmc.use_buffer_c)
-            ? brgmm_ctx.get_buf_C_ptr(ithr, m_blk_idx, n_blk_idx)
-            : ptr_D;
-
     const auto zp_comp_a
             = brgmm_ctx.get_zp_a_compensation_ptr(ithr, b_idx, n_blk_idx);
     const auto zp_comp_b
@@ -756,6 +797,17 @@ void brgemm_matmul_t<isa>::compute_kernel(
             = brgmm_ctx.get_post_ops_binary_rhs_arg_vec();
     const bool post_ops_applicable = bgmmc.post_ops_applicable
             && (brgmm_ctx.get_num_threads_for_k() <= 1 || bgmmc.K_chunks == 1);
+
+    // For acdb dst: ptr_D points into shared buf_D at the correct (m,n) offset.
+    auto ptr_D = bgmmc.dst_is_acdb
+            ? (brgmm_ctx.get_buf_D_acdb_ptr(b_idx)
+                    + (brgmm_ctx.get_M_idx(m_blk_idx, true) * bgmmc.N + n)
+                            * bgmmc.c_dt_sz)
+            : brgmm_ctx.get_data_C_ptr(
+                      b_idx, brgmm_ctx.get_M_idx(m_blk_idx, true), n);
+    auto ptr_C = (bgmmc.use_buffer_c)
+            ? brgmm_ctx.get_buf_C_ptr(ithr, m_blk_idx, n_blk_idx)
+            : ptr_D;
 
     brgemm_dynamic_values_t leading_dimensions(
             bgmmc.LDA, bgmmc.LDB, brgmm_ctx.get_LDC(), brgmm_ctx.get_LDD());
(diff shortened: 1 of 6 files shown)


src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1631,6 +1631,11 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
 
     VCHECK_BG(bm_conf_utils.set_or_check_tags(src_md, dst_md, bias_md, helper),
             VERBOSE_UNSUPPORTED_TAG);
+    bgmmc.dst_is_acdb = (bgmmc.dst_tag == format_tag::acdb);
```
densamoilov: Is it possible for `bgmmc.dst_tag == format_tag::acdb` to be true if the batch dimension `b` is unit? 

In general, I think it'd be better to come up with a more general check that relies on dimensions and strides values to find out if we need the copy C kernel.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1631,6 +1631,11 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    // acdb dst with sum post-op is not yet supported (sum reads from D buffer
+    // which would not contain prior dst values).
+    VCONDCHECK_BG(
+            !(bgmmc.dst_is_acdb && bgmmc.with_sum), VERBOSE_UNSUPPORTED_POSTOP);
```
densamoilov: I also think we may need to disable this copy C kernel for gemv cases because brgemv can handle strided writes except for only one gemv strategy that might need this copy C kernel but such cases are explicitly banned for now via enforcing INCY to always be 1 for that strategy.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
tczeszun: Done, switched to stride-based detection: 
`dst_strides[ndims-1] != 1` and `dst_strides[ndims-2] == 1`. 
Unit-batch cases now skip the copy C path automatically.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
tczeszun: Done, added a guard rejecting acdb/adbc dst for gemv cases.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1631,6 +1632,38 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
             VERBOSE_UNSUPPORTED_TAG);
+    // Detect dst layouts where N (ndims-1) is not innermost, requiring a
+    // copy C kernel. Use stride-based checks for generality: layouts with
+    // unit batch may degenerate to plain strides and skip the copy path.
+    bgmmc.dst_is_acdb = false;
+    bgmmc.dst_is_adbc = false;
+    if (bgmmc.ndims >= 4) {
+        const auto &dst_strides = dst_d.blocking_desc().strides;
+        const bool dst_N_not_innermost
+                = dst_strides[bgmmc.ndims - 1] != 1;
```
densamoilov, line 1643: Accessing strides for unit dimensions is undefined behavior.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1631,6 +1632,38 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+        const bool dst_M_innermost
+                = dst_strides[bgmmc.ndims - 2] == 1;
+        bgmmc.dst_is_adbc = dst_N_not_innermost && dst_M_innermost;
+        bgmmc.dst_is_acdb = dst_N_not_innermost && !dst_M_innermost;
```
densamoilov, line 1647: Actually, my ask to use a general check was about moving away from using format tags entirely for the layout decisions in this PR. A `blocking_desc` is meant to be consumed directly via its dims and strides. Format tags exist only as a convenience for initializing a descriptor, not as a property to query afterward. What matmul does in general and this PR in particular is reverse-engineer a format tag from the blocking descriptor and then branch on it. That action is fundamentally ill-posed for degenerate shapes: a unit dimension doesn't have a defined stride so one physical layout maps to several tags. Inferring a single tag from such a layout is therefore impossible.

So every layout-based decision must be made from the layout's own properties (dims + strides), never from an inferred tag.

I propose introducing a new function to check whether a dimension is the innermost. We could then use this function to identify what actions we should take instead of relying on a tag name.

For example:
```cpp
//  if N != 1 then it should be innermost to avoid the copy
bgmmc.is_copy_c_required = !dims[N]==1 && !is_dim_innermost(dst, bgmmc.ndims - 1)); 
// if the innermost dst dim lives inside a single brgemm dst
bgmmc.fuse_copy_c_allowed =
        bgmmc.is_copy_c_required && is_dim_innermost(dst, bgmmc.ndims - 2);
```

```cpp
// Returns: among dims > 1, is `d` the unit-stride (innermost) dim?
// (unit dims are excluded)
bool is_dim_innermost(const memory_desc_wrapper &md, int d){
   const auto &bd = md.blocking_desc();
   assert(bd.inner_nblks == 0);
   const dim_t *dims = md.dims();

   if (!is_runtime_value(dims[d]) && dims[d] == 1) return false;

   const dim_t sd = bd.strides[d];
   for (int i = 0; i < md.ndims(); ++i) {
       if (i == d) continue;
       if (!is_runtime_value(dims[i]) && dims[i] == 1) continue;
       if (bd.strides[i] < sd) return false;
   }
   return true;
} 
```

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -6069,6 +6069,440 @@ status_t create_brgemm_matmul_copy_a(
     return copy_ker->create_kernel();
 }
 
+// Copy-C kernel for acdb dst: 16×16 transpose from scratch buffer to dst.
+struct jit_brgemm_matmul_copy_c_avx512_t : public jit_brgemm_matmul_copy_c_t,
```
densamoilov, line 6073: * The kernel does transposition so it should have `transpose` in the name
* The current convention is to have isa as a template parameter and keep the name generic

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -6069,6 +6069,440 @@ status_t create_brgemm_matmul_copy_a(
...
+    add(reg_dst, dst_m_stride_);
+    mov(reg_dst_row, reg_dst);
+    jmp(m_loop_label, T_NEAR);
+
+    L(done_label);
+    postamble();
+}
+
+// Copy-C kernel for adbc dst: 16×16 transpose from scratch buffer to dst.
+struct jit_brgemm_matmul_copy_c_adbc_avx512_t
```
densamoilov, line 6274: The kernel should not have a specific format tag in the name as it should rely in dims an strides.

src/cpu/x64/matmul/brgemm_matmul_copy_utils.cpp
```diff
@@ -6069,6 +6069,440 @@ status_t create_brgemm_matmul_copy_a(
...
+            jl(n_tail_label, T_NEAR);
+
+            // Load 16 rows of 16 N elements each
+            for (int i = 0; i < 16; i++) {
+                vmovups(zmm_src(i),
+                        ptr[reg_src_row + i * src_row_stride_
+                                + reg_n_iter * dt_sz_]);
+            }
+
+            emit_16x16_transpose();
```
densamoilov, line 6401: Both `jit_brgemm_matmul_copy_c_avx512_t` and `jit_brgemm_matmul_copy_c_adbc_avx512_t` seem to have the same core code that emits a 16x16 transposition block. The difference is the stride and tail handling (one has a static tail and the other one has a dynamic tail). 

We shouldn't have dedicated copy kernels for different format tags. They should be merged into a single, generic copy c kernel.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2464,6 +2548,10 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
 
     bool packed_sparse_weights() const { return bgmmc_.packed_sparse_weights; }
 
+    char *get_buf_D_acdb_ptr(int b_idx) const {
```
densamoilov, line 2551: The tag agnostic `get_buf_D_ptr` should be extended to support the new case (e.g. `b_idx` and `ithr` parameters could be generalized into a single one. The branching logic inside could use the new action variables in bgmmc.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -873,6 +925,38 @@ void brgemm_matmul_t<isa>::compute_kernel(
 
     brgmm_ctx.maybe_restore_dst_values_from_buffer(
             ithr, b_idx, m_blk_idx, n_blk_idx);
+
+    // For acdb dst: copy from buf_C to shared buf_D.
+    // The actual transpose to dst happens in Phase 2 after the parallel section.
+    if (bgmmc.dst_is_acdb && is_last_K_blk
+            && !post_ops_applicable && bgmmc.use_buffer_c) {
```
densamoilov, line 932: When `post_ops_applicable = true` we write to D directly. So for the case when we reorder data separately at the end we don't seem to have this issue. But for the fused case the output contains nans.

`--matmul --dtag=adbc --attr-post-ops=relu 1x16x64x384:1x16x384x384`

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -708,6 +721,40 @@ status_t brgemm_matmul_t<isa>::execute_body(const exec_ctx_t &ctx) const {
     maybe_reduce_and_convert_partial_results_A(brgmm_ctx_ptr);
     maybe_reduce_partial_results_and_apply_postops(brgmm_ctx_ptr);
 
+    // Phase 2: copy scratch buffer to acdb dst.
+    if (pd()->get_brgemm_matmul_conf().dst_is_acdb) {
```
densamoilov, line 725: This code should be moved to a separate function to follow the current maybe_* convetion.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1631,6 +1632,38 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
 
     VCHECK_BG(bm_conf_utils.set_or_check_tags(src_md, dst_md, bias_md, helper),
             VERBOSE_UNSUPPORTED_TAG);
+    // Detect dst layouts where N (ndims-1) is not innermost, requiring a
+    // copy C kernel. Use stride-based checks for generality: layouts with
+    // unit batch may degenerate to plain strides and skip the copy path.
+    bgmmc.dst_is_acdb = false;
```
densamoilov, line 1638: The shapes below map to both of the tags. 
* The primitive cache relies on the invariant that implementation behaves identically for identical physical layouts. In our case, depending on which format tag we used to create the first primitive we may see different results (one code path has a bug, the other works fine). This points to the false branching inside the implementation
* It can cause performance issues as we may dispatch the slower code path (the one with a separate copy c)

`--matmul --dtag=acdb,adbc 2x1x1x16:2x1x16x16 - passed`
`--matmul --dtag=adbc,acdb 2x1x1x16:2x1x16x16 - failed`

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -514,6 +514,9 @@ status_t brgemm_matmul_t<isa>::init(engine_t *engine) {
         CHECK(sparse_decompress_kernel_->create_kernel());
     }
 
+    if (bgmmc.dst_is_acdb || bgmmc.dst_is_adbc)
+        CHECK(create_brgemm_matmul_copy_c(copy_C_kernel_, &bgmmc));
```
densamoilov, line 518: We don't seem to check for ISA here so if we have one of those tags but it's an avx2 instantiation we may end up creating the avx512 specific copy c kernel on avx2.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -601,7 +604,12 @@ status_t brgemm_matmul_t<isa>::execute_body(const exec_ctx_t &ctx) const {
         int m_chunks_per_thread = div_up(M_chunks, bgmmc.nthr_m);
         int n_chunks_per_thread = div_up(N_chunks, bgmmc.nthr_n);
         int batch_per_thread = div_up(bgmmc.batch, bgmmc.nthr_b);
-        if (brgmm_ctx.is_chunks_horizontal_process_order())
+
+        if (bgmmc.dst_is_acdb)
+            nd_iterator_init(start, bt, bgmmc.nthr_b, mt, bgmmc.nthr_m, nt,
```
densamoilov, line 609: Have you explored an option to move the transpose to the compute body? It seems like we could do a 16x16 block transposition since we already adjusted the iterator.

densamoilov (changes requested):

PR #5120, opened 2026-05-05
benchdnn: matmul: remove obsolete skip unimpl and invalid checks

dzarukin:
Dropped nvidia and amd checks from matmul driver.
Also dropped checks related to invalid API calls such as:
* Calling ZP with non-integer data types;
* Using odd dim size for dense subbyte tensors;
Updated input lists correspondently once incorrect usage popped up.

### DIFF
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -640,6 +640,9 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
 
         int n_unit_strides = 0;
         for (int d = 0; d < ndims; d++) {
+            // This check restricts N=1 cases for subbyte types:
+            // `--wtag=abc 6x975x2080:6x2080x1` will return invalid_args.
+            // TODO: remove the check and verify the library.
             if (wei_strides[d] == 1) {
                 n_unit_strides++;
                 VCHECK_MATMUL(
--- a/tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
+++ b/tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
@@ -36,20 +36,22 @@
 --runtime_dims_masks=1:0,3:3
 --batch=shapes_2d_ci
 
-# Decompression
+# Weights-Only-Quantization
 --reset
---impl=ref # Intentionally test reference impl coverage
+--impl=ref # Test ref impl for supported coverage
 --wtag=any,abc,acb
---dt=s8:u4:f32,f16:s4:f16,s8:s8:f16,f16:s4:f32
+--dt=f16:s4:f16,f16:s4:f32
 --attr-scales=wei:per_ocic:f16:192x1
 --attr-zero-points=wei:per_ocic:s4:192x1
 --attr-fpmath=f16:true
 12x4x576:12x576x192
 12x6x192:12x192x100
 
+# Grouped Quantization
+--reset
+--impl=ref # Test ref impl for supported coverage
 --wtag=any,abcd,abdc
---dt=s8:s4:f16
---attr-fpmath=f16:true
+--dt=s8:s4:f16,s8:u4:f32,s8:s8:f16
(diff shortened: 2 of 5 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_graph

mzhukova (approved):

tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
```diff
@@ -36,20 +36,22 @@
 --runtime_dims_masks=1:0,3:3
 --batch=shapes_2d_ci
 
-# Decompression
+# Weights-only-Quantization
```
densamoilov: (nit) 

```suggestion
# Weights-Only Quantization
```

tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
```diff
@@ -36,20 +36,22 @@
...
+--dt=f16:s4:f16,f16:s4:f32
 --attr-scales=wei:per_ocic:f16:192x1
 --attr-zero-points=wei:per_ocic:s4:192x1
 --attr-fpmath=f16:true
 12x4x576:12x576x192
 12x6x192:12x192x100
 
+# Grouped Quantization
+--reset
+--impl=ref # Intentionally test reference impl coverage
```
densamoilov: I think stating a reason would be better because having this option here kind of implies what the comment says. Something like "only ref impl exists" or some other reason

tests/benchdnn/matmul/matmul.cpp
```diff
@@ -862,136 +852,24 @@ void skip_unimplemented_prb(const prb_t *prb, res_t *res) {
...
-                res->state = SKIPPED;
-                res->reason = reason_t::skip_not_supported;
-                return;
-            }
-        }
     }
 }
 
 void skip_invalid_prb(const prb_t *prb, res_t *res) {
-    if (!prb->attr.zero_points.get(DNNL_ARG_WEIGHTS).is_def()
```
densamoilov, line 893: I'm wondering why was this added in the first place.

In general, what are the rules now for keeping some checks here? Maybe it'd be worth documenting 
how we should use `skip_invalid_prb` and `skip_unimplemented_prb`?

src/common/matmul.cpp
```diff
@@ -609,6 +609,9 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
 
         int n_unit_strides = 0;
         for (int d = 0; d < ndims; d++) {
+            // This check restricts GeMV with N=1 cases for subbyte types:
```
densamoilov: (nit) maybe let's drop "GeMV" here since this is just the N=1 case of matmul and the more specific term feels a bit confusing

densamoilov (approved):

tests/benchdnn/inputs/matmul/harness_matmul_smoke_ref
dzarukin: I slightly updated the text but what's written is the reason - we want to verify ref impl supports these cases because this is the design of the whole file.

src/common/matmul.cpp
dzarukin: Sure, dropped, thanks.

tests/benchdnn/matmul/matmul.cpp
dzarukin: It was added as a preventive measure from debugging why something is not working. I guess it was done long before verbose dispatch developed into its current state and it was really hard to understand what exactly went wrong. Right now it's not a problem anymore.

I've updated documentation at the spot of calling, I don't think I can do any better unless starting a new page with technical details which we are not doing neither for benchdnn nor for the library.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_graph

PR #5067, opened 2026-04-22
xe: remove with_punning

rjoursler:
In [MFDNN-14925](https://jira.devtools.intel.com/browse/MFDNN-14925), we found multiple primitive failure due to inconsistent use of WITH_PUNNING across buffers. This PR completely removes this knob so that this failure mode can no longer occur. This fixes  [MFDNN-14925](https://jira.devtools.intel.com/browse/MFDNN-14925) along with a few other existing issues I found during testing.

PR #5123, opened 2026-05-07
[GPU] Fix issues found when testing without JIT kernels

echeresh:
While working on https://github.com/uxlfoundation/oneDNN/pull/5105, I found and fixed several issues:

- Correctness issues in reference implementations
- Missing `mayiuse_ngen_kernels` check in JIT pooling
- Invalid or incorrect tests, e.g. `mayiuse_ngen_kernels` check doesn't work when JIT kernels for the current architecture are disabled via `ONEDNN_ENABLE_PRIMITIVE_GPU_ISA`

### DIFF
--- a/src/common/verbose.hpp
+++ b/src/common/verbose.hpp
@@ -113,6 +113,12 @@ static inline int format_type_checker(const char *, ...) {
                     __LINE__); \
     } while (0)
 
+// Per-primitive VDISPATCH_<PRIM>{,_SC,_IC} wrappers (in <primitive>_pd.hpp) are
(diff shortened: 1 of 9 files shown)


echeresh:
make test
disable device_cpu

rjoursler:
FYI: I believe #5067 fixed some hidden OpenCL kernel issues as well. I mostly tasked copilot with performing the refactoring, and it definitely went off on some tangents to fix some existing issues that were found during testing (although as a result, these tangents are not cleanly tracked).

atkassen (approved):

dzarukin (approved):

PR #5124, opened 2026-05-07
cpu: aarch64: rationalise bnorm vector-length logic

Sqvid:
# Description
This PR simplifies the vector-length related logic in our batch normalisation implementations.

The commits do roughly the following:
  1. Remove unnecessarily complex special-cases for `asimd`
  2. Introduce `sve_128` support to `jit_uni_batch_normalization_{fwd,bwd}_t`
  3. Make the `jit_uni_batch_normalization_{fwd,bwd}_t` template instantiations vector-length agnostic
  4. Add support for plain formats on 16-byte SIMD-length implementations
  5. Add verbose dispatch info

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/jit_uni_batch_normalization.cpp
+++ b/src/cpu/aarch64/jit_uni_batch_normalization.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
 * Copyright 2020 Intel Corporation
 * Copyright 2020-2024 FUJITSU LIMITED
-* Copyright 2025 Arm Ltd. and affiliates
+* Copyright 2025-2026 Arm Ltd. and affiliates
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@
 * limitations under the License.
 *******************************************************************************/
 
-#include <assert.h>
+#include <cassert>
 #include <functional>
 
 #include "common/c_types_map.hpp"
@@ -26,6 +26,7 @@
 #include "common/nstl.hpp"
 #include "common/type_helpers.hpp"
 #include "common/utils.hpp"
+#include "common/verbose_msg.hpp"
 
 #include "cpu/cpu_batch_normalization_utils.hpp"
 #include "cpu/platform.hpp"
@@ -131,15 +132,14 @@ struct jit_bnorm_conf_t {
             iters_ = 1;
         }
 
-        is_spatial_thr_
-                = this->thread_partition(/* is_spatial_thr_ = */ true, nthr,
-                        /* dimensions */
-                        N, C_blks_per_iter_, SP,
-                        /* outputs */
-                        C_nthr_, N_nthr_, S_nthr_);
+        is_spatial_thr_ = thread_partition(/* is_spatial_thr_ = */ true, nthr,
+                /* dimensions */
+                N, C_blks_per_iter_, SP,
+                /* outputs */
+                C_nthr_, N_nthr_, S_nthr_);
 
         if (iters_ > 1)
-            this->thread_partition(is_spatial_thr_, nthr,
+            thread_partition(is_spatial_thr_, nthr,
                     /* dimensions */
                     N, C_blks_last_iter_, SP,
                     /* outputs */
@@ -158,9 +158,17 @@ struct jit_bnorm_conf_t {
             S_nthr = 1;
         } else {
             if (is_nspc_) {
+                const int min_c_blks_per_thr = nstl::max(1, 16 / simd_w_);
                 if (C_blks <= 8)
                     C_nthr = 1;
-                else if (nthr >= 8 && C_blks <= 32)
+                else if (simd_w_ < 16 && C_blks % min_c_blks_per_thr == 0) {
+                    // Keep each channel thread on a 16-channel boundary.
+                    const dim_t C_chunks = C_blks / min_c_blks_per_thr;
+                    C_nthr = (int)nstl::min<dim_t>(
+                            nstl::min<dim_t>(8, nthr), C_chunks);
+                    while (C_nthr > 1 && C_chunks % C_nthr != 0)
+                        --C_nthr;
+                } else if (nthr >= 8 && C_blks <= 32)
                     C_nthr = 8;
                 else {
                     C_nthr = (int)math::gcd((dim_t)nthr, C_blks);
@@ -236,7 +244,7 @@ struct jit_bnorm_t : public jit_generator_t {
     bool is_bf16_ = false;
     bool is_f16_ = false;
 
-    const int vlen = isa == asimd ? 32 : cpu_isa_traits<isa>::vlen;
+    const int vlen = simd_bytes(isa);
     int vlen_spat_data_; // set by ctor depending on data type (BF16 or FP32);
 
     XReg reg_param = abi_param1;
@@ -288,6 +296,9 @@ struct jit_bnorm_t : public jit_generator_t {
 
     size_t unroll_blocks;
     size_t unroll_regs;
+    // Index 0 is temporal value. is_fwd() ? vdiff_beta : vbeta
+    TReg vzero = TReg(0);
+    TReg vacc = TReg(1);
     TReg vdiff_beta = TReg(21);
     TReg vdiff_gamma = TReg(22);
     TReg vsqrtvar = TReg(23);
@@ -300,8 +311,6 @@ struct jit_bnorm_t : public jit_generator_t {
     TReg vtail_mask = TReg(30);
     TReg t_tmp0 = TReg(31);
     TReg t_tmp1 = TReg(20);
-    TReg vzero = TReg(
-            0); // Index 0 is temporal value. is_fwd() ? vdiff_beta : vbeta
 
     size_t spat_size;
     size_t chan_data_offt;
@@ -460,7 +469,7 @@ struct jit_bnorm_t : public jit_generator_t {
         XReg r = jbp_->is_nspc_ ? reg_soff_nspc : reg_soff;
         ZRegS zzero = ZRegS(vzero.getIdx());
 
-        assert(isa == sve_256 || isa == sve_512);
+        assert(is_superset(isa, sve));
 
         assert(bits < 64);
         lsr(r, r, bits);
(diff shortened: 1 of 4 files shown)


tests/benchdnn/inputs/bnorm/test_bnorm_all_blocked
```diff
@@ -7,7 +7,7 @@
 --dir=FWD_D,BWD_DW
 --dt=f32
 
---tag=aBx8b
+--tag=aBx8b,aBx4b
```
dzarukin: Same feedback as for deconvolution - we aim to move in a backward direction from blocked activations to `acdb` (or `nhwc`), this shouldn't be extended.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -2436,7 +2370,7 @@ status_t jit_uni_batch_normalization_bwd_t<isa>::pd_t::init(engine_t *engine) {
     }
 
     if (fuse_norm_relu()) {
-        if (isa != sve_256 && isa != sve_512) return status::unimplemented;
+        if (!is_superset(isa, sve)) return status::unimplemented;
```
dzarukin, line 2386: Minor, random spot: you might want to combine newly introduced unimplemented returns with adding verbose support for them, but it's up to you, feel free to resolve any time.

tests/benchdnn/inputs/bnorm/test_bnorm_all_blocked
Sqvid: To clarify my understanding, you are saying that `bwd` primitives should not accept blocked formats for their `src` (like `aBcd4b`) and should only accept a plain format (i.e, `acdb`)? I assume this is true for both `src` and `diff_src`?

The problem here is, [like the `cpu/x64` counterpart](https://github.com/uxlfoundation/oneDNN/blob/fe59f72bc909a67d0a0af7d6f018983dec83a7dc/src/cpu/x64/jit_uni_batch_normalization.cpp#L2530-L2531), in some cases the backward primitive _only_ supports blocked formats.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -1088,16 +1059,16 @@ struct jit_bnorm_t : public jit_generator_t {
                     subs_imm(reg_ctr, reg_ctr, 1, X_TMP_0);
                     b(NE, mean_reduction_thrs);
                 }
-                if (isa == sve_256 || isa == sve_512)
+                if (is_superset(isa, sve))
                     fdiv(ZRegS(1), P_ALL_ONE / T_m, ZRegS(vchan_size.getIdx()));
```
jondea: Can we use `uni_fdiv(TReg(1))`? (Probably want to use a variable too rather than hard coding an index of 1.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -1088,16 +1059,16 @@ struct jit_bnorm_t : public jit_generator_t {
...
                 else
                     fdiv(VReg4S(1), VReg4S(1), VReg4S(vchan_size.getIdx()));
                 uni_store_maybe_tail(mean_ptr(), TReg(1));
 
-                if (isa == sve_256 || isa == sve_512)
+                if (is_superset(isa, sve))
                     add_imm(reg_coff, reg_coff, vlen, X_TMP_0);
```
jondea: true and false follow the same path right?

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -1171,14 +1130,14 @@ struct jit_bnorm_t : public jit_generator_t {
                     subs(reg_ctr, reg_ctr, 1);
                     b(NE, var_reduction_thrs);
                 }
-                if (isa == sve_256 || isa == sve_512)
+                if (is_superset(isa, sve))
                     fdiv(ZRegS(1), P_ALL_ONE / T_m, ZRegS(vchan_size.getIdx()));
                 else { fdiv(VReg4S(1), VReg4S(1), VReg4S(IDX(vchan_size))); }
                 uni_store_maybe_tail(var_ptr(), TReg(1));
-                if (isa == sve_256 || isa == sve_512)
+                if (is_superset(isa, sve))
```
jondea: Similarly, can we get rid of this if?

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -2012,22 +1938,17 @@ struct jit_bnorm_t : public jit_generator_t {
         , is_f16_(pd_->src_md()->data_type == data_type::f16)
         , vlen_spat_data_(vlen / (1 + is_xf16())) // 32B of xF16 -> 64B of FP32
         , unroll_blocks(
-                  (isa == sve_256 || isa == sve_512) && !jbp_->is_spatial_thr_
-                          ? 4
-                          : 1)
+                  (is_superset(isa, sve)) && !jbp_->is_spatial_thr_ ? 4 : 1)
```
jondea: What's the rationale for this logic? If it's about unrolling, then `sve_128` may be more like `asimd` right?

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -2293,32 +2211,38 @@ status_t jit_uni_batch_normalization_fwd_t<isa>::pd_t::init(engine_t *engine) {
...
+
+        default: return status::unimplemented;
+    }
+
+    if (src_tag == format_tag::undef) return status::unimplemented;
 
     if (is_fwd() ? with_relu_post_op(is_training()) || fuse_norm_relu()
                  : fuse_norm_relu())
-        if (isa != sve_512) return status::unimplemented; // TODO
+        if (!(is_superset(isa, sve) && simd_bytes(isa) == 64))
```
jondea, line 2245: Do we need both of these checks?

jondea (changes requested):
Excellent, thanks for yet another code quality win! Minor comments

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Just thought the most equivalent replacement was to check that it was both `sve` and `simd_bytes == 64`. It's not like ISAs being added everyday but I though I'd err on the side of precision.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: That's a good point, I have made it a function of SIMD length now. I believe @keanucz has some work which goes deeper into the performance heuristics.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Done.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Fixed. This is what I get for using `vim` find and replace instead of a trillion-parameter transformer.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Done.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -1074,30 +1046,27 @@ struct jit_bnorm_t : public jit_generator_t {
             {
                 mov(reg_roff, reg_coff);
                 uni_eor(TReg(0), TReg(0), TReg(0));
```
jondea, line 1056: nit: `uni_clear` is clearer

jondea (approved):

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Done!

tests/benchdnn/inputs/bnorm/test_bnorm_all_blocked
Sqvid: @dzarukin If I have understood you correctly, could you check if the latest commit addresses your point?

dzarukin: It would be completely addressed if the coverage wasn't extended. :)
I'm not so sure it would work on x64 or, if would, would pass (in adequate time).

> To clarify my understanding, you are saying that bwd primitives should not accept blocked formats for their src (like aBcd4b) and should only accept a plain format (i.e, acdb)?

Yes, this is direction. If you want to have these blocked formats for batch norm, your ResNet-50 convolutions must support blocked output format for dst, too, otherwise, a pointless reorder in-between conv and batch norm appears, which kills the performance (which this blocked format aims to find). Having that blocked layout support in convolution is extra maintenance cost out of nothing.

For x64 in that case bwd training will just go to less performant impl which is fine as we are not anticipating any training happening on avx2 (14-year old hardware). avx512+ has acdb support several lines above what you've pointed to.

Sqvid: @dzarukin thank you for the explanation, that makes sense. The test suite now only adds `aBx4b` for the forward inference case, and I have removed blocked-format support in the `bwd_t` for 16-byte SIMD widths. I have left the pre-existing blocked formats in-place though, as I feel we're better off removing support for those as part of a separate change.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: Thanks for the reminder! Done.

tests/benchdnn/inputs/bnorm/test_bnorm_all_blocked
dzarukin: Separate change for sure. Thank you.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_bnorm
set test_scope=NIGHTLY

tests/benchdnn/inputs/bnorm/test_bnorm_all_blocked
dzarukin: It seems x64 backend just skips aBx4b format as it hits ref impl. All good to go!

dzarukin (approved):

PR #4944, opened 2026-04-02
cpu: aarch64: enable BWD_D for `jit:sve_128` and `jit_1x1:sve_128` conv

Anndrey24:
# Description

Based on #3022, this commit enables the BWD_D direction for f32 `jit:sve_128` and `jit_1x1:sve_128` convolutions. It also makes certain deconvolution shapes (width-only, unit height) use the `conv:any+jit:sve_128` implementation over `acl`, leading to significant speedups.

Support for the BWD_W direction was also added for both JIT paths, but they remain unselected because of a few observed regressions for small shapes with 1x1 kernel at 64/96 threads.

## Performance improvements (f32)
The following conv shape batches were benchmarked on a c8g instance, at various thread counts (1/4/16/64/96) in BWD_D direction:

- `shapes_resnet_50`
- `shapes_yolo_v2`
- `shapes_mobilenet`
- `shapes_efficientdet`

The shapes where implementation changed after applying the patch are:
| desc | impl change | 1 thread | 4 threads | 16 threads | 64 threads | 96 threads |
|---|---|---|---|---|---|---|
| `g1mb50ic3ih224iw224oc64oh112ow112kh7kw7sh2sw2ph3pw3n"resnet_50:conv1"` | `gemm:ref -> jit:sve_128` | 0.87x (397.934 ms -> 457.032 ms) | 1.01x (115.479 ms -> 114.314 ms) | 1.67x (47.9829 ms -> 28.6858 ms) | 3.33x (32.7874 ms -> 9.85014 ms) | 4.83x (33.7085 ms -> 6.98194 ms) |
| `g1mb50ic256ih56iw56oc512oh28ow28kh1kw1sh2sw2ph0pw0n"resnet_50:res3a_branch1"` | `gemm:ref -> jit_1x1:sve_128` | 1.73x (290.12 ms -> 168.059 ms) | 1.85x (77.9032 ms -> 42.084 ms) | 2.66x (33.4627 ms -> 12.5755 ms) | 10.23x (46.9699 ms -> 4.59243 ms) | 7.93x (24.3662 ms -> 3.07351 ms) |
[...]

PR #5125, opened 2026-05-07
fixup: cpu: aarch64: enable plain BWD_D for jit:sve_128 conv

Anndrey24:
# Description

This commit fixes BWD_D tail-width `jit:sve_128` conv kernels where local `ur_w` can be smaller than `jcp.ur_w` (e.g. when coming from `ur_w_tail`). The old register allocation, introduced in #4944,  reserved a full-width output block and could push broadcast/weight temporaries past the 32 SVE registers.

The new formula bases the temporary register offset on the actual tail width, so the final partial width block does not over-reserve accumulator registers.

### DIFF
--- a/src/cpu/aarch64/jit_sve_conv_kernel.cpp
+++ b/src/cpu/aarch64/jit_sve_conv_kernel.cpp
@@ -1651,7 +1651,7 @@ void jit_sve_conv_bwd_data_kernel_f32_t<isa>::compute_loop_fma_core(
     };
 
     auto zreg_inp_s = [=](int i_ic, int nb_x_blocking) {
-        int idx = (i_ic + nb_x_blocking * jcp.ur_w);
+        int idx = (i_ic + (nb_x_blocking - 1) * jcp.ur_w + ur_w);
         assert(idx < 31);
         return ZRegS(idx);
     };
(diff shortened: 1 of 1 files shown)


Sqvid:
https://github.com/uxlfoundation/oneDNN/actions/runs/25510445398
Nightly run here

Sqvid (approved):
LGTM, thanks for the fix!

jondea:
Is this a fix for something that was enabled but broken? If so, are there failing tests in CI?

Or to say it another way, is this reachable and is it tested in CI?

Sqvid:
> Is this a fix for something that was enabled but broken? If so, are there failing tests in CI?

@jondea Failing nightly here: https://github.com/uxlfoundation/oneDNN/actions/runs/25539074642/job/74962528774

jondea (approved):

Anndrey24:
Right, the CI set doesn't seem to have a shape that reaches this bug, but the nightly set caught the assertion failures it caused: 
```
benchdnn: /home/runner/work/oneDNN/oneDNN/oneDNN_build/src/cpu/aarch64/jit_sve_conv_kernel.cpp:1794: void dnnl::impl::cpu::aarch64::jit_sve_conv_bwd_data_kernel_f32_t<isa>::compute_loop_fma_core(int, int, int, int) [with dnnl::impl::cpu::aarch64::cpu_isa_t isa = dnnl::impl::cpu::aarch64::sve_128]: Assertion `wei_reg_ofs + num_wei_reg <= 32' failed.
```

Interestingly enough, without the asserts enabled (like how I mistakenly ran it during development), the code would still "accidentally" work even with a seemingly over-reserved register window, because the logic conservatively accounts for 2 weight registers (i.e. it reserves space as if `ZReg(31)` and `ZReg(32)` could be needed), even if these specific shapes only ever used the first one (`ZReg(31)`). The formula should be more in line with what actually happens now.

This is also a great reminder that I should always build with `RelWithAssert` instead of `Release` :)

PR #5137, opened 2026-05-11
benchdnn: graph: enable sycl graph execution mode

TaoLv:
Add SYCL graph recording (--execution-mode=graph) support to benchdnn graph driver.

### DIFF
--- a/tests/benchdnn/graph/graph.cpp
+++ b/tests/benchdnn/graph/graph.cpp
@@ -688,38 +688,48 @@ int doit(const prb_t *prb, res_t *res) {
 
         ref_partition_t ref_partition(dg, partitions[i], inputs, outputs);
 
-        // Construct memory for both perf & corr modes
-        SAFE(ref_partition.init_ref(graph_in_ports, res), WARN);
-        if (res->state == SKIPPED) return OK;
+        // Reference execution must use direct execution mode. Use RAII guard to
+        // ensure the original execution mode is restored.
+        // TODO: introduce exec_args_t for `execute_and_wait` to explicitly
+        // specify `execution_mode` value instead of updating the global value.
+        {
+            execution_mode_guard_t exec_mode_guard(execution_mode_t::direct);
 
-        if (has_bench_mode_bit(mode_bit_t::corr)) {
-            // correctness mode, run ref partition
-            if (res->state == UNTESTED || res->state == EXECUTED
-                    || res->state == DEFERRED) {
-                ref_partition.exec_ops(res);
-                if (res->state == FAILED) return FAIL;
-                if (res->state == SKIPPED || res->state == UNIMPLEMENTED)
-                    return OK;
-            } else {
-                // once a partition failed on init_ref, terminate whole graph execution
-                return FAIL;
+            // Construct memory for both perf & corr modes
+            SAFE(ref_partition.init_ref(graph_in_ports, res), WARN);
+            if (res->state == SKIPPED) return OK;
+
+            if (has_bench_mode_bit(mode_bit_t::corr)) {
+                // correctness mode, run ref partition
+                if (res->state == UNTESTED || res->state == EXECUTED
+                        || res->state == DEFERRED) {
+                    ref_partition.exec_ops(res);
+                    if (res->state == FAILED) return FAIL;
+                    if (res->state == SKIPPED || res->state == UNIMPLEMENTED)
+                        return OK;
+                } else {
+                    // once a partition failed on init_ref, terminate whole graph execution
+                    return FAIL;
+                }
             }
+
+            // To ensure data consistency between graph and reference paths,
+            // memory initialization and data displacement for reference
+            // primitives (performed in ref_partition.init_ref and
+            // ref_partition.exec_ops) must be completed before initializing
+            // graph memory.
+            SAFE(ref_partition.init_graph_mem(partition_mem_map_v[i], res),
+                    WARN);
+            if (res->state == SKIPPED) return OK;
+
+            // unmap memory from host to device
+            SAFE(map_unmap_partition_mem(
+                         partition_mem_map_v[i], inputs, UNMAP, res),
+                    WARN);
+            SAFE(map_unmap_partition_mem(
+                         partition_mem_map_v[i], outputs, UNMAP, res),
+                    WARN);
         }
-
-        // To ensure data consistency between graph and reference paths,
-        // memory initialization and data displacement for reference primitives
-        // (performed in ref_partition.init_ref and ref_partition.exec_ops)
-        // must be completed before initializing graph memory.
-        SAFE(ref_partition.init_graph_mem(partition_mem_map_v[i], res), WARN);
-        if (res->state == SKIPPED) return OK;
-
-        // unmap memory from host to device
-        SAFE(map_unmap_partition_mem(
-                     partition_mem_map_v[i], inputs, UNMAP, res),
-                WARN);
-        SAFE(map_unmap_partition_mem(
-                     partition_mem_map_v[i], outputs, UNMAP, res),
-                WARN);
 
         const op_ref_list_t &op_list = ref_partition.get_partition_ops();
         const auto &inplace_ports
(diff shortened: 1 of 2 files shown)


tests/benchdnn/graph/utils.hpp
```diff
@@ -282,5 +282,85 @@ struct stream_staller_t {
...
+// Returns nullptr on failure (with res->state set to FAILED).
+template <typename F>
+inline std::unique_ptr<sycl_exec_graph_t> record_and_finalize(
+        cpp_stream_t &stream, ::sycl::queue &queue, F &&record_func,
+        res_t *res) {
+    BENCHDNN_PRINT(
+            2, "%s\n", "[INFO] Using experimental SYCL graph execution.");
+
+    // Drain the queue to clear any pending events before recording.
+    static_cast<dnnl::stream &>(stream).wait();
```
dzarukin: Is this needed?
I thought `cpp_stream_t::operator dnnl::stream &()` should cover this cast.

dzarukin (commented):
Thanks @TaoLv for adding this support and I really like some of the introduced abstractions/functions. I can see that it mostly duplicates the existing code when it comes to managing sycl_graph, and, I guess, the primary reason is the difference in C++ and C engines and streams used in graph driver and rest drivers.
I'd really love to merge streams and engines and consolidate the sycl_graph logic under a single umbrella re-used everywhere. Could you, please, post a TODO in utils.hpp with a message to merge them? I'll open a JIRA to address a technical debt of different engines and streams. Having two versions of the code becomes quite expensive and doesn't seem to be necessary.

Secondly, I don't see drivers have performance mode support with sycl_graph. I also see graph driver uses the copy-exact approach a lot from aggregated perf mode support which is under redo process: https://github.com/uxlfoundation/oneDNN/pull/5118. I think it would be beneficial to consolidate this logic first before introducing sycl_graph support for perf mode for both graph and rest drivers. ~~Would you mind to postpone this support for graph driver and drop it from this PR until we have less friction in two paths?~~

I think the most important is to have sycl_graph enabled in weekly testing for graph driver which is covered in graph.cpp part and it should be enough so far to satisfy coming requests and validate them from functional point of view.

What would be your thoughts on this?

Edit: Vadim pointed out we didn't add perf mode support for primitive drivers on purpose, so the point about dropping should fade. I still feel it's good to have aligned and reusable code but I won't insist on it.

TaoLv:
Thank you for the review, @dzarukin . I just reverted the changes in utils.cpp (execute_and_wait(), measure_perf_aggregate(), and measure_perf_individual()) and added the TODO in utils.hpp.

Now running a graph case with --execution-mode=graph looks like below. We still need some fixes on the library side to pass the test. Regarding the repeated "[INFO] Using experimental SYCL graph execution.", it's because the reference path calls other drivers which also detect the execution mode setting.

```
$ ./tests/benchdnn/benchdnn --graph --engine=gpu -v2 --execution-mode=graph --case=complex_fusion/mha/sdpa-plain-simplified-f32.json 
run: --graph --engine=gpu --execution-mode=graph --case=complex_fusion/mha/sdpa-plain-simplified-f32.json
[INFO] Using experimental SYCL graph execution.
[INFO] Using experimental SYCL graph execution.
[INFO] Using experimental SYCL graph execution.
[INFO] Using experimental SYCL graph execution.
[INFO] Using experimental SYCL graph execution.
[INFO] Using experimental SYCL graph execution.
[ERROR] SYCL graph execution exception: Graph nodes cannot depend on events from outside the graph.
0:FAILED (348 ms) __REPRO: --graph --engine=gpu --execution-mode=graph --case=complex_fusion/mha/sdpa-plain-simplified-f32.json
0:FAILED (348 ms) __REPRO: --graph --engine=gpu --execution-mode=graph --case=complex_fusion/mha/sdpa-plain-simplified-f32.json
============================
total: 0.35s; create_pd: 0.06s (16%); create_prim: 0.01s (2%); fill: 0.00s (0%); execute: 0.01s (2%); compute_ref: 0.00s (0%); compare: 0.00s (0%);
```

ElaineBao (approved):

rongzha1 (approved):

tests/benchdnn/graph/utils.hpp
```diff
@@ -282,5 +282,107 @@ struct stream_staller_t {
...
+    execution_mode_guard_t &operator=(const execution_mode_guard_t &) = delete;
+
+private:
+    execution_mode_t saved_mode_;
+};
+
+// Returns true if SYCL command graph execution mode is active.
+inline bool use_sycl_graph_exec() {
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_SYCL
+    return is_gpu() && execution_mode == execution_mode_t::graph;
```
dzarukin: Using `is_sycl()` instead should simplify this function a bit.

tests/benchdnn/graph/graph.cpp
```diff
@@ -688,38 +688,46 @@ int doit(const prb_t *prb, res_t *res) {
...
-                    || res->state == DEFERRED) {
-                ref_partition.exec_ops(res);
-                if (res->state == FAILED) return FAIL;
-                if (res->state == SKIPPED || res->state == UNIMPLEMENTED)
-                    return OK;
-            } else {
-                // once a partition failed on init_ref, terminate whole graph execution
-                return FAIL;
+        // Reference execution must use direct execution mode. Use RAII guard to
+        // ensure the original execution mode is restored.
```
dzarukin, line 692:
```suggestion
        // ensure the original execution mode is restored.
        // TODO: introduce exec_args_t for `execute_and_wait` to explicitly
        // specify `execution_mode` value instead of updating the global value.
```

tests/benchdnn/graph/graph.cpp
```diff
@@ -751,15 +759,30 @@ int doit(const prb_t *prb, res_t *res) {
...
-        stream_staller_t staller(strm);
-        // Need following clean-up steps as the memories have been mappped to
-        // device. Otherwise the deconstruction will fail.
-        DNN_GRAPH_SAFE(
-                c_partitions[i - idx_offset].execute(strm, input_ts, output_ts),
-                (WARN | NEED_CLEANUP), res);
-        staller.release();
-
-        DNN_GRAPH_SAFE(strm.wait(), WARN, res);
+#if DNNL_GPU_RUNTIME == DNNL_RUNTIME_SYCL
```
dzarukin: Nit: please move macro inside the `if`.
A TODO to consolidate with primitives would help too.

tests/benchdnn/graph/utils.hpp
```diff
@@ -282,5 +282,107 @@ struct stream_staller_t {
...
+    }
+    return OK;
+}
+
+// Record operations into a SYCL command graph and return the finalized
+// executable. `record_func` is called while the queue is in recording state.
+// Returns nullptr on failure (with res->state set to FAILED).
+template <typename F>
+inline std::unique_ptr<sycl_exec_graph_t> record_and_finalize(
+        cpp_stream_t &stream, ::sycl::queue &queue, F &&record_func,
```
dzarukin: Could you, please, check if this would work:
```
std::unique_ptr<sycl_exec_graph_t> record_and_finalize(
        cpp_stream_t &stream, ::sycl::queue &queue, std::function<void()> &record_func,
        res_t *res) { ... }
```
And call it like this:
```
    std::function<void()> record_func = std::bind(compiled_partition_executor, c_partitions[i - idx_offset], strm,
                        input_ts, output_ts);
    record_and_finalize(stream, queue, record_func, res);
```

If yes, let's stick to this. It allows to move all bodies into cpp and also helps to consolidate with primitives.

TaoLv:
make test
disable benchdnn_all
enable beenchdnn_graph

tests/benchdnn/graph/utils.hpp
TaoLv: I couldn't find `is_sycl()`. Called `is_sycl_engine()` instead.

tests/benchdnn/graph/utils.hpp
TaoLv: Sure. Changed it to `std::function`.

dzarukin (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #5142, opened 2026-05-11
xe: skip JIT kernels at runtime when excluded from build

echeresh:
Partial port of #5105.

### DIFF
--- a/src/gpu/intel/jit/binary_format.cpp
+++ b/src/gpu/intel/jit/binary_format.cpp
@@ -21,7 +21,9 @@
 
 #include "gpu/intel/jit/binary_format.hpp"
 
+#include "common/impl_registration.hpp"
 #include "common/utils.hpp"
+#include "common/verbose.hpp"
 #include "gpu/intel/compute/utils.hpp"
 #include "gpu/intel/engine.hpp"
 #include "gpu/intel/jit/generator.hpp"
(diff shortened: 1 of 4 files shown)


echeresh:
make test
disable test_device_cpu

src/gpu/intel/ocl/hw_info.cpp
```diff
@@ -56,6 +58,53 @@ xpu::runtime_version_t get_driver_version(cl_device_id device) {
...
+
+    // Bail out if the detected architecture is disabled at build time.
+    bool arch_enabled = false;
+    switch (hw) {
+        case ngen::HW::XeLP:
+            arch_enabled = BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XELP;
+            break;
+        case ngen::HW::XeHP:
+            arch_enabled = BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XEHP;
+            break;
```
rjoursler: Comment: We should likely just remove XeHP from nGEN support. To my knowledge, there was never any released hardware and it is never tested in CI.

echeresh: Yes, you are right - we should do that at some point. But I don't think it's actionable in this PR - this has to be done separately.

rjoursler: >  But I don't think it's actionable in this PR - this has to be done separately.

Just to make sure we are on the same page, I am proposing we just go ahead and disable support via this check, not perform a "proper" removal throughout the oneDNN codebase at this point.

echeresh: Thanks for the clarification. I interpreted your suggestion as full-scale work to drop it from nGEN, etc.

I'm not sure it's the right approach since it's part of the public documentation and build options.

@vpirogov please advise - Roy is suggesting to disable XeHP support via this check since there is no publicly released hardware and no coverage in testing.

vpirogov: Sounds reasonable.

echeresh: Thanks, dropped from the switch, also added a comment.

rjoursler (approved):

echeresh:
make test
disable test_device_cpu

kealan-barbieri (approved):

src/gpu/intel/jit/binary_format.cpp
```diff
@@ -365,6 +368,50 @@ status_t gpu_supports_binary_format(bool *ok, impl::engine_t *engine) {
...
+        case compute::gpu_arch_t::xe2:
+            arch_enabled = BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XE2;
+            break;
+        case compute::gpu_arch_t::xe3:
+            arch_enabled = BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XE3;
+            break;
+        case compute::gpu_arch_t::xe3p:
+            arch_enabled = BUILD_PRIMITIVE_GPU_ISA_ALL || BUILD_XE3P;
+            break;
+        default: break;
```
Simonsays095: nit: we'll need to remember to add future architectures to this switch statement. I prefer removing the default and explicitly adding cases for all existing cases, so we can catch a missing enum value via static analysis.

Also, could be simplified by a helper macro + `REG_<arch>_ISA` wrappers.

src/gpu/intel/jit/binary_format.cpp
```diff
@@ -365,6 +368,50 @@ status_t gpu_supports_binary_format(bool *ok, impl::engine_t *engine) {
...
+    }
+    if (!arch_enabled) return status::success;
+
+    if (arch <= compute::gpu_arch_t::xe3) {
```
Simonsays095: Why do Xe3p (and all future platforms) default to true? This is the existing behaviour for `ocl/hw_info.cpp` but this changes behaviour for `ze/utils.cpp`.

echeresh:
make test

src/gpu/intel/jit/binary_format.cpp
echeresh: Good catch, changed to `<= xe3p`. After checking the code, this binary format check is usually bypassed (on all platforms) due to this skip-flag:

https://github.com/uxlfoundation/oneDNN/blob/63b994eca746de4bb0668444a0b9335c0e01de25/src/gpu/intel/jit/binary_format.cpp#L182

On modern drivers we always skip it to avoid startup overhead. I wouldn't mind if someone removed this check completely.

PR #5145, opened 2026-05-12
rfcs: Simplify the AArch64 Stack by Replacing Compute Library with Arm KleidiAI

jondea:
[Rendered document](https://github.com/uxlfoundation/oneDNN/blob/jondea/rfcs/replace-acl-with-kleidiai/rfcs/20260507-replace-acl-with-kleidiai-ops/README.md)

[PoC branch](https://github.com/uxlfoundation/oneDNN/tree/jondea/replace-acl-with-kleidiai)


Sqvid:
@vpirogov @dzarukin

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+The biggest open question is: how exactly do we integrate the library into oneDNN?
+
+**note**: the PoC branch currently uses a submodule, this is temporary and we will replace it with whatever approach is chosen.
+
+### Solution 1: add KleidiAI to third party
+
+Place the KleidiAI source (just the parts we use, which is a subset of the `experimental/ops` directory) in `third_party`, commit it to the oneDNN repo.
+We could then build it by default on AArch64.
+This is in line with existing third party libraries (e.g. `Xbyak_aarch64`).
```
dzarukin, line 134: This option is OK with me.

It might be good to explicitly spell several things:
* Is it a completely header-based library or not?
* What is the licensing? Using code in third-party implies compliance with current distribution model.
* Promises about API and how two versions of KAI would co-exist in a single framework (e.g., one version in oneDNN, another in PT used for direct kernel development, any potential symbol conflicts, etc.?)

Additionally, will it be allowed to specify a path to own version of KAI without the necessity of updating it in third_party?

If yes, there should be an explanation how old oneDNN with new KAI and new oneDNN with old KAI would work - any guarantees or promises? This might circle back to the co-existence of two versions in a single binary.

dzarukin (commented):
@jondea , thanks for the RFC! I posted some questions/suggestions for further clarification.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Pros
+- Simple and reliable
+- Fully accelerated AArch64 build without the need for extra build flags. The extra build step and flags cause considerable downstream pain (e.g. when building in PyTorch).
+
+#### Cons 
+- Increased repo size for all platforms, ~18% increase
+  - oneDNN is 1,080,353 lines and 84MB of source code
+  - After stripping out what we don't need, kleidiai/experimental/ops is 331,597 lines (mostly inline assembly) and 15MB of source code
+- Updates to KleidiAI may be large diffs of inline assembly (although they would be limited to `third_party/kleidiai`)
```
vpirogov, line 145: How much churn you would realistically expect in this  code? I see that `ukernel` part in KAI is relatively stable and updates are very manageable.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Open Question: How to Integrate into oneDNN?
+
+The biggest open question is: how exactly do we integrate the library into oneDNN?
+
+**note**: the PoC branch currently uses a submodule, this is temporary and we will replace it with whatever approach is chosen.
+
+### Solution 1: add KleidiAI to third party
+
+Place the KleidiAI source (just the parts we use, which is a subset of the `experimental/ops` directory) in `third_party`, commit it to the oneDNN repo.
```
vpirogov, line 132: Would you see any impact on support matrix from building KAI from source with oneDNN? For example, does it add requirements related to build system or compiler vendor/versions?

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
```diff
@@ -0,0 +1,215 @@
...
+Matmul makes up a large proportion of model runtime, but we also need to replace other ACL implementations. Inner product can be implemented as a wrapper around matmul. We discuss the remaining primitives in the table below.
+
+
+| Primitive  | Data Type    | Adv SIMD impl | SVE impl |
+| ---------- | ------------ | ---------- | -------- |
+| matmul/ip  | f32/bf16/f16 | KAI ✅     | KAI ✅   |
+| matmul/ip  | u8/s8        | KAI ✅     | jit ✅^1 |
+| conv       | f32          | KAI ✅ ^2a  | jit ✅ ^2b   |
+| conv       | f16          | KAI ✅ ^2a  | KAI/JIT ✅ ^2b  |
+| conv       | bf16         | KAI ✅ ^2a  | KAI/JIT ✅ ^2b |
```
vpirogov: What are your view on the long term picture for use of KAI? Will it remain a legacy path for tricky cases (like autogenerated GEMM in x64) or reliance on KAI will grow?

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Cons
+- All downstream projects have extra build steps and flags (from experience, this is a big con)
+- No canonical version of KleidiAI
+
+## Appendix
+
+### Overheads
+
+The overheads are not large, but they are measurable by considering at a small problem. For a single thread we see (~0.001ms = 1us) and for more threads, it becomes larger (~0.007ms = 7us)
```
renato-arantes: This part feels a bit lost to me, as if you accidentally cut a paragraph earlier. What is this overhead related to? The primitive creation?

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: - The `ops` library is not header only (unlike the `ukernel` KleidiAI library). It is a separate CMake project which compiles to a library called `kleidiai_ops`
- All the KleidiAI source we will use is licensed under the same license as oneDNN, Apache-2.0 (some of the testing infrastructure is under BSD-3 but this is not necessary for the oneDNN integration).
- Currently PT only uses the `ukernel` part of the library and oneDNN will only use the `ops` part. However, if they do end up with crossover, I think we have options
  - prefix the symbols with `onednn_`
  - I'm not sure on the details but we can link privately with `-Wl,--exclude-libs,libkleidiai.a` and/or using `PRIVATE` CMake options

I think we should be able to provide a `KAI_OPS_INSTALL_PATH` flag to override the linking in oneDNN.

Initially, while the `ops` library is experimental, we won't make any guarantees about API stability or interoperability between versions. Once it is stable we plan it to be versioned using the root KleidiAI semantic version number.

I will add this information to the RFC too.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: We don't expect much churn, but it is likely that the `ops` part of the library will have more churn than the `ukernel` part, because the `ukernel` is a lower level API, so there are fewer degrees of freedom.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: Good question, I need to investigate this further. Some initial notes:
- Currently [`KleidiAIOps`](https://gitlab.arm.com/kleidi/kleidiai/-/blob/feature/add-kleidiai-ops/experimental/ops/CMakeLists.txt) requires CMake 3.25, but I know this can come down to 3.22. We may be able to bring it down to 3.13
- We can't build the `KleidiAIOps` library with MSVC yet, but we don't claim this yet for AArch64, instead preferring Clang for Windows
- As far as I can tell, it builds with `CMAKE_CXX_STANDARD=11` like the root oneDNN project

Broadly speaking though, because the `ops` library is also in ACL (with some minor changes), it won't require anything more than oneDNN+ACL did.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: We don't have a concrete roadmap, the plan is to consider kernels on a case-by-case basis. It's a balance between JIT allowing us to have broad coverage and specialize without binary bloat (if we can justify the time) and KleidiAI allowing us to reuse well-tested code. Broadly speaking, F32 conv/matmul and 128bit vector length favours KleidiAI, but quantized and longer vector lengths favour JIT. So we don't envision a dramatic expansion of KleidiAI usage from this PoC, or reduction to zero. Some growth/shrinkage is inevitable depending on what delivers best perf on Arm as new hardware arrives and our JIT implementations mature.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: Runtime and creation overhead. But good point, I have added some context to clarify.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
puneetmatharu: @vpirogov are there any plans to bump the minimum CMake version from 3.13? (The current [build from source instructions](https://github.com/uxlfoundation/oneDNN#requirements-for-building-from-source) currently state 3.13 as the lower bound.) CMake 3.13 was released in 2018, which is a very long time ago. oneDNN also only seems to test [with CMake 3.31](https://github.com/search?q=repo%3Auxlfoundation%2FoneDNN%20cmakeVersion%3A&type=code). How do you guarantee that the build will work for 3.13?

KleidiAI is meant to support all current LTS versions of Ubuntu, making the lowest version of CMake 3.22  (released 2021) currently, which is part of the reason why we were hoping to leave the `KleidiAIOps` CMake project at that version. (N.B. the top-level `KleidiAI` project is currently at 3.16 but that might be updated to 3.22 in the somewhat near future.)

Also, for reference, PyTorch currently use [CMake 3.27 as a minimum](https://github.com/pytorch/pytorch/blob/090b49d99bb9c26144cedea3b9c566fbe6ede275/CMakeLists.txt#L1).

vpirogov: There's no specific plans to update minimal supported CMake version. These changes are made on case by case basis when a need for an update arises. Does KAI use any specific CMake features not supported by 3.13?

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
vpirogov: Shrinkage is definitely welcome :) Assuming that there's no plans for growth this looks very much like a stopgap on the path to JIT, which is great.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
puneetmatharu: The most beneficial feature is `CMakePresets.json` support which allows us to clearly separate the CI from the project configuration (i.e. all CI flags can be explicitly specified locally in a CMake preset inside the project). CMake 3.13 doesn't provide that. As mentioned above, KleidiAI supports all LTS versions of Ubuntu so we'd like to set that to CMake 3.22 to align. Of course, updating the requirements needs to take into consideration how problematic the change will be for people that rely on oneDNN. Is there a downstream user of oneDNN that guides the minimum CMake version? Or would it be reasonable to also bring the minimum version up in oneDNN?

vpirogov: oneDNN is primarily consumed in source code form using CMake build system by downstream users, so bumping minimal CMake version may be disruptive. I see [OpenVINO](https://github.com/openvinotoolkit/openvino/blob/master/CMakeLists.txt#L10) and [PaddlePaddle](https://github.com/PaddlePaddle/Paddle/blob/develop/CMakeLists.txt#L20) are using 3.18. So bumping to anything above 3.18 is not safe.

We also [had a discussion](https://github.com/uxlfoundation/oneDNN/pull/2225) about bumping minimal version based on OS configs, but decided that it should be driven by CMake features use instead. A proposal to bump CMake minimal version to 3.19 in order to use `cmake-presets` or version X to use feature Y would be the right place to collect feedback.

vpirogov (approved):
Thanks for the detailed proposal, @jondea! Option 1 looks reasonable to me.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: One potential large change to the KleidiAI ops source (probably landing late this year) will be moving from inline asm to separate assembly files so we can support MSVC.

renato-arantes (approved):

Sqvid (approved):
LGTM! On the note of library integration I lean slightly towards copying the directory into `third_party` like we do for `Xbyak_aarch64` but I don't have strong feelings on it.

theComputeKid:
Does the original ACL RFC for oneDNN need updating to communicate it now no longer applies: https://github.com/uxlfoundation/oneDNN/tree/jondea/rfcs/replace-acl-with-kleidiai/rfcs/20200731-Using-Arm-Compute-Library-kernels-on-AArch64

vpirogov:
> Does the original ACL RFC for oneDNN need updating to communicate it now no longer applies

I'd say it's unnecessary.

rfcs/20260507-replace-acl-with-kleidiai-ops/README.md
jondea: As far as I know, after some tweaks, I think we can keep the oneDNN CMake version unchanged. Is that right @puneetmatharu ?

jondea:
Can we merge this? I know that I can click the button, but I have never done this for RFCs, so I wanted to double check.

dzarukin (approved):

PR #5146, opened 2026-05-12
cpu: aarch64: bnorm: improve SVE128 threading

keanucz:
# Description

This PR builds on @Sqvid's patch #5124 to make bnorm VLA-agnostic. By forcing spatial threading for SVE128 NSPC inference (simd_w<8, N==1, SP>=225) and disabling channel threading by setting C_nthr=1 for simd_w<8 (instead of just narrowing that to MB=1, SP>=225), we observe a 1.332x geometric mean speedup across the entire bnorm benchdnn test suite compared with upstream. This holds true while plain-layout cases still use 16-input-channel grouping.

abba_4var_5x_pr_vs_upstream_chart

Overall, mine and Sid's changes have basically delivered a 2x geomean speedup on the benchdnn bnorm cases compared to upstream previously.
abba_4var_5x_summary_chart

This patch also delivers some other minor improvements:
1. Gates ZReg non-temporal stores on the is_nt_store flag instead of unconditionally emitting stnt1w (because SVE128 NSPC stores are immediately reused and benefit from cacheable st1w) - should deliver a small speedup ~5-10% by taking advantage of cache

~~2. Always initialises P_ALL_ONE in the kernel prologue for SVE ISAs instead of skipping when ISA VL matches HW VL - not related to performance specifically, but this at least avoids the use of an uninitialized predicate register.~~

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/jit_uni_batch_normalization.cpp
+++ b/src/cpu/aarch64/jit_uni_batch_normalization.cpp
@@ -151,33 +151,46 @@ struct jit_bnorm_conf_t {
     bool thread_partition(bool spatial_thr_allowed, int nthr, dim_t N,
             dim_t C_blks, dim_t SP, int &C_nthr, int &N_nthr,
             int &S_nthr) const {
-        if (((nthr <= C_blks) && IMPLICATION(is_nspc_, N == 1))
-                || !dnnl_thr_syncable()) {
+        // For SVE128 NSPC with MB=1 and SP >= 225, force spatial threading
+        // instead of channel threading (currently yields better performance
+        // on certain bnorm shapes)
+        const bool force_sve128_spatial
+                = is_nspc_ && simd_w_ < 8 && N == 1 && SP >= 225;
+
+        if ((((nthr <= C_blks) && IMPLICATION(is_nspc_, N == 1)
+                     && !force_sve128_spatial)
+                    || !dnnl_thr_syncable())) {
             C_nthr = nthr;
             N_nthr = 1;
             S_nthr = 1;
         } else {
             if (is_nspc_) {
                 const int min_c_blks_per_thr = nstl::max(1, 16 / simd_w_);
-                if (C_blks <= 8)
+                if (C_blks <= 8) {
                     C_nthr = 1;
-                else if (simd_w_ < 16 && C_blks % min_c_blks_per_thr == 0) {
+                } else if (simd_w_ < 16 && C_blks % min_c_blks_per_thr == 0) {
                     // Keep each channel thread on a 16-channel boundary.
                     const dim_t C_chunks = C_blks / min_c_blks_per_thr;
                     C_nthr = (int)nstl::min<dim_t>(
                             nstl::min<dim_t>(8, nthr), C_chunks);
                     while (C_nthr > 1 && C_chunks % C_nthr != 0)
                         --C_nthr;
-                } else if (nthr >= 8 && C_blks <= 32)
+                } else if (nthr >= 8 && C_blks <= 32) {
                     C_nthr = 8;
-                else {
+                } else if (simd_w_ < 8) {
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -1949,9 +1958,10 @@ struct jit_bnorm_t : public jit_generator_t {
 
         size_t simd_w = simd_elems(data_type::f32, isa);
         if (is_superset(isa, sve)) {
-            if (simd_w != cpu_sveLen / sizeof(float)) {
-                set_preg(P_ALL_ONE.s, simd_w, X_TMP_0, X_TMP_1);
-            }
+            // Always initialize the predicate used as "all-active lanes".
```
Sqvid: I think the logic behind this comment made sense when we had separate instantiations for `sve_128`, `sve_256`, and `sve_512`. Could you review it in light of the unified `sve` template?

I think your new `set_preg` line might just be repeating work done in the preamble.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
```diff
@@ -151,33 +151,38 @@ struct jit_bnorm_conf_t {
...
-                    // Keep each channel thread on a 16-channel boundary.
-                    const dim_t C_chunks = C_blks / min_c_blks_per_thr;
-                    C_nthr = (int)nstl::min<dim_t>(
-                            nstl::min<dim_t>(8, nthr), C_chunks);
-                    while (C_nthr > 1 && C_chunks % C_nthr != 0)
-                        --C_nthr;
-                } else if (nthr >= 8 && C_blks <= 32)
+                else if (simd_w_ >= 8 && nthr >= 8 && C_blks <= 32)
+                    // Only use C_nthr=8 heuristic for larger simd widths
+                    // (sve_256+). For sve_128 (simd_w=4), this causes
```
Sqvid: > For sve_128 (simd_w=4), this causes correctness issues with reduction buffers

This isn't actually true in as long as you had the `min_c_blks_per_thr` code. I think you should explain _why_ you can run into issues here so that a future reader is well informed about what they need to do to modify your heuristic without running into bugs.

Sqvid (changes requested):
Great work! Just a couple of minor points.

src/cpu/aarch64/jit_uni_batch_normalization.cpp
keanucz: completely agree with you - in light of this I've reverted the change

src/cpu/aarch64/jit_uni_batch_normalization.cpp
keanucz: agreed - I've added back your min_c_blks_per_thr code for handling non-SVE128 cases & have added a comment that better explains why (for now) it's better if channel threading is disabled for SVE128 cases

src/cpu/aarch64/jit_uni_batch_normalization.cpp
Sqvid: I actually think we can just remove this block completely. You should verify, but I think `P_ALL_ONE` should be set in `preamble()` anyway.

Sqvid (changes requested):

src/cpu/aarch64/jit_uni_batch_normalization.cpp
keanucz: you're right, https://github.com/keanucz/oneDNN/blob/6c1cab0586407fd0f5c356066f90984be25a47ba/src/cpu/aarch64/jit_generator.hpp#L220 already sets P_ALL_ONE to true for any SVE type anyways - I'll remove this redundant code

Sqvid (approved):
Looks good to me, thank you for the patch!

jondea (approved):

PR #5148, opened 2026-05-12
Fixed extended APX registers clobbering in BRDGMM-based convolution

vpirogov:
This PR fixes clobbering of `reg_src_zero_point`, `reg_zp_compensation`, and `reg_s8s8_comp` registers on Intel AVX10.2 codepath with APX enabled. Previously the following lines shifted the value resulting in loss of original pointer when `restore()` turns into a `NOP`:
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp#L683-L687
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp#L698-L707

Fixes [MFDNN-15006](https://jira.devtools.intel.com/browse/MFDNN-15006)

~~The design choice to make `restore()` do nothing when scratchpad is not allocated looks bug prone. I wonder how bad it would be if scratchpad gets allocated unconditionally for all registers used by the kernel.~~

Update: Added a set of `aux` registers to preserve respective buffers base in `reg_src_zero_point`, `reg_zp_compensation`, and `reg_s8s8_comp` and use `aux` counterparts during the iteration. This does not change anything for non-APX path, but allows to avoid using scratchpad when APX is available.

### DIFF
--- a/src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
+++ b/src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
@@ -239,6 +239,10 @@ private:
     const injector_utils::reg64_savable_t reg_s8s8_comp {
             regscratchpad_, r14, r29};
 
+    const reg64_t reg_aux_src_zp = rax;
+    const reg64_t reg_aux_zp_comp = r12;
+    const reg64_t reg_aux_s8s8_comp = r14;
+
     Xbyak::Opmask k_mask = Xbyak::Opmask(2);
     Xbyak::Opmask k_tail_mask = Xbyak::Opmask(3);
     Xbyak::Opmask kblend_mask = Xbyak::Opmask(4);
--- a/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
+++ b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp
@@ -680,13 +680,13 @@ void jit_brdgmm_kernel_base_t<Wmm>::maybe_transpose_interleaved_vnni_to_plain(
 
 template <typename Wmm>
 void jit_brdgmm_kernel_base_t<Wmm>::load_src_zp() {
-    reg_src_zero_point.restore();
-    lea(reg_src_zero_point,
+    reg_src_zero_point.restoreTo(reg_aux_src_zp);
+    lea(reg_aux_src_zp,
             is_src_zp_bcast_
-                    ? ptr_b[reg_src_zero_point]
-                    : ptr[reg_src_zero_point + reg_aux_N * sizeof(int32_t)]);
+                    ? ptr_b[reg_aux_src_zp]
+                    : ptr[reg_aux_src_zp + reg_aux_N * sizeof(int32_t)]);
     if (!is_superset(brg.isa_impl, avx512_core) && is_src_zp_bcast_)
-        uni_vpbroadcastd(vmm_bcast(), ptr[reg_src_zero_point]);
+        uni_vpbroadcastd(vmm_bcast(), ptr[reg_aux_src_zp]);
 }
 
 template <typename Wmm>
@@ -696,14 +696,15 @@ void jit_brdgmm_kernel_base_t<Wmm>::compute_int8_compensation(
     const int v_substep = vnni_substep();
 
     if (brg.req_s8s8_compensation) {
-        reg_s8s8_comp.restore();
-        lea(reg_s8s8_comp, ptr[reg_s8s8_comp + reg_aux_N * sizeof(int32_t)]);
+        reg_s8s8_comp.restoreTo(reg_aux_s8s8_comp);
+        lea(reg_aux_s8s8_comp,
+                ptr[reg_aux_s8s8_comp + reg_aux_N * sizeof(int32_t)]);
     }
     if (compute_src_zp_) {
         load_src_zp();
-        reg_zp_compensation.restore();
-        lea(reg_zp_compensation,
-                ptr[reg_zp_compensation + reg_aux_N * sizeof(int32_t)]);
+        reg_zp_compensation.restoreTo(reg_aux_zp_comp);
+        lea(reg_aux_zp_comp,
+                ptr[reg_aux_zp_comp + reg_aux_N * sizeof(int32_t)]);
     }
 
     for_(int v_i = 0; v_i < v_substep; ++v_i)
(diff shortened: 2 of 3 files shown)


src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
```diff
@@ -224,10 +224,14 @@ struct jit_brdgmm_kernel_base_t : public jit_base_brgemm_kernel_t {
             regscratchpad_, rax, r22};
     const injector_utils::reg64_savable_t reg_dst_zero_point {
             regscratchpad_, rax, r23};
+    // reg_src_zero_point is restored after being modified in load_src_zp
```
densamoilov: (random spot)

> The design choice to make restore() do nothing when scratchpad is not allocated looks bug prone. I wonder how bad it would be if scratchpad gets allocated unconditionally for all registers used by the kernel.

I think the main goal behind that design choice was to preserve the current register usage model, which relies on manual register allocation. Without APX we would simply need to ensure that the registers are used correctly and that there’s no overlap, pretty much what we've always been doing although that needs to change.

I think allocating a scratchpad for all cases where APX is unavailable may lead to poor manual register management, which will hurt performance.

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
```diff
@@ -224,10 +224,14 @@ struct jit_brdgmm_kernel_base_t : public jit_base_brgemm_kernel_t {
...
     const injector_utils::reg64_savable_t reg_src_zero_point {
-            regscratchpad_, rax, r24};
+            regscratchpad_,
+            regscratchpad_.ExtendedRegisters() ? r24 : rax};
+    // reg_zp_compensation is restored after being modified in compute_int8_compensation
     const injector_utils::reg64_savable_t reg_zp_compensation {
-            regscratchpad_, r12, r25};
+            regscratchpad_,
+            regscratchpad_.ExtendedRegisters() ? r25 : r12};
```
densamoilov: If I read it right we just call this ctor here, which always books:
```cpp
reg64_savable_t(registry_scratchpad_t &regscratchpad,
        const Xbyak::Reg64 &reg, bool is_storable = true);
```
Previously, we had similar logic implemented through delegating constructors but we explicitly asked the constructor above not to perform booking for the cases when APX is available.

This change appears to force scratchpad booking even when APX registers are available and I’m not sure I understand why.

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
vpirogov: Maybe I'm missing something in save/restore semantics of `reg64_saveable_t`. The place that confuses me is that save/restore operations quietly do nothing when APX is available: 
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/injectors/injector_utils.hpp#L137-L139

This directly contradicts the stated purpose of this class:
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/injectors/injector_utils.hpp#L117-L120

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
vpirogov: Indeed, this change forces scratchpad booking for the case APX when registers are available. Otherwise `restore()` turns into a `NOP` and results into incorrect address for compensation loaded here:
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp#L702-L707

densamoilov: When APX registers are available we don’t have anything to restore because the register is available and is supposed to contain the correct data, _assuming we don’t overwrite it_ somewhere. When APX registers are not available we have to restore their contents from the stack (i.e. emulate their presence). This is why `restore()` is a NOP when APX registers are available.

I suspect your change might fix the issue because we now always force the use of the stack but the real problem is likely that we overwrite the APX registers somewhere.

vpirogov: Thanks, I get how save/restore is supposed to work now. Looks like the piece of code I quoted breaks this assumption and relies on `restore()` to revert `reg_zp_compensation` to the base of compensation buffer:
https://github.com/uxlfoundation/oneDNN/blob/49f6d238780130ae24774a8e01d8b5972be3db3b/src/cpu/x64/brgemm/jit_brdgmm_kernel.cpp#L159-L160

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
vpirogov: This one is rescinded.

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
densamoilov: I may be missing something but I don’t think I follow this part. Here we only read the address of the zp compensation buffer and call `save()`. Since `save()` becomes a NOP when APX registers are available `reg_zp_compensation` must preserve that address for the entire kernel execution. Otherwise, if APX registers are not available, we spill it to the stack and restore it later.

vpirogov: The register value is not preserved in this code. In no-APX case the code works by a coincidence because it calls restore without doing a save. Here's what `reg_zp_compensation` values look like before line 705:
| aux_N | no-APX             | APX                |
|:------|:-------------------|:-------------------|
| 0     | 0x000055ef90129100 | 0x0000563ad1aa3100 |
| 32    | 0x000055ef90129100 | 0x0000563ad1aa3100 |
| 64    | 0x000055ef90129100 | 0x0000563ad1aa3180 |
| 96    | 0x000055ef90129100 | 0x0000563ad1aa3280 |
| 128   | 0x000055ef90129100 | 0x0000563ad1aa3400 |
| 160   | 0x000055ef90129100 | 0x0000563ad1aa3600 |
| 192   | 0x000055ef90129100 | 0x0000563ad1aa3880 |
| 224   | 0x000055ef90129100 | 0x0000563ad1aa3b80 |
| 256   | 0x000055ef90129100 | 0x0000563ad1aa3f00 |
| 288   | 0x000055ef90129100 | KABOOM                     |

densamoilov (approved):

vpirogov:
make test
set test_scope=NIGHTLY
disable test_device_gpu
enable benchdnn_conv

src/cpu/x64/brgemm/jit_brdgmm_kernel.hpp
ankalinin: Ah, I only now realized...
For APX case we suppose keeping correct data in register and the restore operation is a NOP. But specifically for this registers  we overwrite the data by later operations `lea..` 
For non-APX case everything should work fine. "restore" actually reloads the value into the register from scratchpad. And the fact that we immediately overwrite it by the next instruction `lea..` isn’t a problem.

ankalinin (approved):

PR #5155, opened 2026-05-14
common, cpu: random fixes and updates

dzarukin:
Fixes and updates for random spots found during generated coverage testing.

* Restricting e8m0 and e4m3 scales data types to be used with multiple values only as they are part of scaling policies that assume grouping.
* Fixing brgemm kernel and brg_matmul for WoQ cases with a common src_scale.
* Fixing API and ref impl to work properly with precomputed reductions.

### DIFF
--- a/src/common/matmul.cpp
+++ b/src/common/matmul.cpp
@@ -508,20 +508,40 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     pr_dt == data_type::s32, VERBOSE_UNSUPPORTED_PR_CFG);
 
             if (!pr.get(DNNL_ARG_SRC).has_default_groups()) {
+                const auto &sc = attr->scales_;
                 const dim_t src_pr_group_k = pr.get_group(DNNL_ARG_SRC, 1);
 
+                // Pre-computed reductions can only be used when the PR group
+                // size is the minimal amongst all group sizes along the K
+                // dimension from a mathematical point of view.
+                dim_t minimal_group = INT64_MAX;
                 dim_t wei_zp_group_k = 1;
                 if (!zp.get(DNNL_ARG_WEIGHTS).has_default_groups()) {
-                    const int mask_wei = zp.get_mask(DNNL_ARG_WEIGHTS);
-                    if (mask_wei & wei_qmask_K)
-                        wei_zp_group_k = zp.get_group(DNNL_ARG_WEIGHTS, 0);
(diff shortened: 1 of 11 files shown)


dzarukin:
make test

src/common/matmul.cpp
```diff
@@ -510,21 +508,39 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     pr_dt == data_type::s32, VERBOSE_UNSUPPORTED_PR_CFG);
 
             if (!pr.get(DNNL_ARG_SRC).has_default_groups()) {
+                const auto &sc = attr->scales_;
                 const dim_t src_pr_group_k = pr.get_group(DNNL_ARG_SRC, 1);
 
-                dim_t wei_zero_point_group_k = 1;
+                // It's possible to use PRs only when its group is the minimal
+                // among all K groups from mathematics point of view.
```
avmanerikar: (nit):
```suggestion
                // Pre-computed reductions can only be used when the PR group size is 
                // minimal amongst all group sizes along the K dimension from a
                // mathematical point of view.
```

avmanerikar (approved):

src/common/matmul.cpp
```diff
@@ -421,7 +421,7 @@ status_t matmul_attr_check(const matmul_desc_t &desc, const engine_t *engine,
                     VERBOSE_UNSUPPORTED_ZP_CFG);
         }
 
-        dim_t wei_zero_point_group_k = 1;
+        dim_t wei_zp_group_k = 1;
```
densamoilov, line 424: Why did you rename it?

densamoilov (approved):

src/common/matmul.cpp
dzarukin: To save some space., nothing more.

densamoilov: Let's rename `wei_zero_point_group_n` too for consistency?

src/common/matmul.cpp
dzarukin: Applied, thanks.

src/common/matmul.cpp
dzarukin: Sure thing, updated all `_zero_point_` variables into `_zp_` for consistency, thanks for pointing on it.

densamoilov: Thank you!

PR #4297, opened 2025-11-07
all: asynchronous Eigen threadpool runtime support

dzarukin:
This is the final PR that covers all the current needs for asynchronous Eigen threadpool runtime support.

Important notes:
* Benchdnn treats async runtime same way as GPU when it comes for perf measurements. It also got an abstraction to ensure that for correctness validation submission happens before any thread starts execution.
* Auto-gen based implementation becomes disabled for **any** threadpool implementation as it's impossible to determine the threadpool flag at creation time.
* Verbose functionality doesn't properly return times as the call can't be blocking, otherwise, it'll lead to a deadlock. It's an area for a future improvement.
* The support itself (if not considering the amount of preliminary work promoted to main) is limited to ~two~ three major changes:
  - Parallel calls change their capture from the capture-by-reference to the capture-by-copy. This ensures that all stack variables and objects and also an execution context will be used in delayed parallel tasks with their actual values or addresses on the heap instead of referencing them on stack where they can be already destroyed by the time of execution.
  - Sequential calls to perform the last reduction (or any kind of post parallel work) are wrapped into a parallel call with a single task issued. This is necessary to ensure the dependencies are followed (guarantees of proper synchronization between consecutive parallel calls are provided by the runtime itself). Paired with a change on the main branch that forces a parallel call with a single task to be redirected into the threadpool runtime and not to shortcut an execution on-the-spot as for other runtimes, it works. The latter runtimes shouldn't be affected performance-wise (besides one extra call).
  - (New) Access to memory buffers can't be done in submitter thread as there's no guarantees the data in those memory buffers will be available at the time of submitting new tasks. Any access, either user provided memory or scratchpad memory must be done _inside_ parallel tasks.

These two key principles now must be followed for updated CPU implementations (in the future, probably, every primitive and every implementation).

In a nutshell, a developer should treat async Eigen runtime in the following way:
* Submitter thread comes and extracts all or most of the pointers from the execution context, creates local variables, and reaches:
  - the parallel section - in this case it saves all the stack variable till that moment and provides this snapshot to parallel tasks (lambda functions). All the objects will be saved either by value (POD types) or by ref_count increase, ensuring those objects won't be destroyed till all the tasks finished.
  - nested primitive - in this case it prepares a new execution context and scratchpad grantor ensuring the lifetime of all newly created objects, and steps into the next execute call, either repeating this step, or getting into the previous one.
[...]

PR #5156, opened 2026-05-14
cpu: aarch64: replace acl with kleidiai

jondea:
See related RFC: https://github.com/uxlfoundation/oneDNN/pull/5145


src/cpu/aarch64/kai_convolution.cpp
```diff
@@ -0,0 +1,746 @@
+/*******************************************************************************
+* Copyright 2020-2026 Arm Ltd. and affiliates
```
renato-arantes: Why 2020-2026? Shouldn't it be only 2026?

src/cpu/aarch64/kai_utils.hpp
```diff
@@ -0,0 +1,190 @@
+/*******************************************************************************
+* Copyright 2023-2026 Arm Ltd. and affiliates
```
renato-arantes, line 2: Same here: Shouldn't be only 2026?

src/cpu/aarch64/matmul_inner_product.hpp
```diff
@@ -0,0 +1,81 @@
+/*******************************************************************************
+* Copyright 2025 Intel Corporation
```
renato-arantes, line 2: Why do we have to include Intel's copyright in those new files?

src/cpu/cpu_deconvolution_list.cpp
```diff
@@ -69,7 +66,6 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
             CPU_INSTANCE_AARCH64(brgemm_deconvolution_fwd_t<sve_256>)
             CPU_INSTANCE_AARCH64(brgemm_deconvolution_fwd_t<sve_128>)
             CPU_INSTANCE_AARCH64(jit_sve_512_core_x8s8s32x_deconvolution_fwd_t)
-            CPU_INSTANCE_AARCH64_ACL(acl_deconvolution_fwd_t)
```
renato-arantes, line 72: We need an ASIMD implementation.

src/cpu/cpu_prelu_list.cpp
```diff
@@ -41,7 +38,6 @@ const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> &impl_list_map() {
     static const std::map<pk_impl_key_t, std::vector<impl_list_item_t>> the_map = REG_PRELU_P({
         {{forward}, {
             CPU_INSTANCE_X64(jit_prelu_fwd_t)
-            CPU_INSTANCE_AARCH64_ACL(acl_prelu_fwd_t)
```
renato-arantes: Also here: we need an AArch64 implementation.

src/cpu/aarch64/kai_convolution.cpp
jondea: Significant content was taken from acl convolution, hence the backdating

src/cpu/aarch64/kai_utils.hpp
jondea: Significant content was taken from acl_utils.*, hence the backdating

src/cpu/aarch64/matmul_inner_product.hpp
jondea: Significant content was taken from src/cpu/x64/matmul_inner_product.hpp, hence the copyright.

src/cpu/cpu_deconvolution_list.cpp
jondea: Agreed!

cfRod:
Does the coverage increase / decrease compared to before?

jondea:
This change should not cause anything to go from an optimized to an unoptimized implementation. If you spot any (particularly ones at the framework level!) then it is a bug, and needs reporting here. Known issues are:
- prelu
- asimd log/log softmax
- MMLA s8:s8:s8
- gemm/conv + any post ops apart from relu (alpha=0) or sum
- batch norm with channels not a multiple of 16 
I think I discuss most of these in [the RFC](https://github.com/uxlfoundation/oneDNN/pull/5145), and we intend to get these fixed before merging. No doubt I have missed some though, so bug reports are very welcome here!

There will also be some things that go from an unoptimized to an optimized implementation, but they are mostly
- edge cases which become easier to handle with `kai` ops because a layer of API has been removed
- most of the JIT impls are more flexible than their ACL counterparts, so expanding their use tends to unintentionally bring extra coverage

It's worth saying that this change gives us more flexibility and will let us increase coverage more easily afterwards, however, it isn't the primary purpose of this patch itself. But, if you have specific and straightforward asks e.g. s8s8s32 we may be able to easily enable them in this patch.

src/cpu/aarch64/kai_convolution.cpp
```diff
@@ -0,0 +1,746 @@
...
+}
+
+bool can_use_direct_1x1_src(const kai_convolution_fwd_t::pd_t &pd) {
+    return !pd.indirect_ && exact_src_dst_tag(*pd.src_md()) == format_tag::nhwc
+            && pd.KH() == 1 && pd.KW() == 1 && pd.KSH() == 1 && pd.KSW() == 1
+            && pd.padT() == 0 && pd.padL() == 0 && pd.OH() == pd.IH()
+            && pd.OW() == pd.IW();
+}
+
+int cap_threads_for_small_gemm(
```
fadara01: We should be cautious with capping threads. In micro-benchmarks or benchdnn, capping threads will likely show nice speedups. However, this can be problematic when you benchmark models end to end, due to scenarios like: 
- Max threads is 64
- We have a model with Layer A and Layer B
- Layer A has threads capped from 64 to 32
- As a result, OMP runtime might shrink the team size (from 64 to 32) - I say might because it's up to the OMP implementation: libomp doesn't tend to do this, but libgomp does this more aggressively.
- Layer B needs all 64 threads, as a result we have to pay the overhead of re-spawningg the 32 threads destroyed in the previous step.

See https://github.com/vllm-project/vllm/issues/34229 and https://github.com/uxlfoundation/oneDNN/pull/4679 for practical examples where this bit us before.

I assume we're now relying on `dnnl::prarallel_for` in place of ACL's `OMPScheduler`. When doing something like `parallel(std::min(pd()->mm_parallel_work, num_threads)`, with libgomp, the team size will shrink from `num_threads` to `mm_parallel_work` as this call becomes: `#pragma omp parallel num_threads(mm_parallel_work)`.

The same pattern wouldn't cause team size to shrink with ACL's `OMPScheduler` as it keeps the extra threads (`omp_num_threads - amount_of_work`) alive and just schedules 0 iterations on them:
```
#pragma omp parallel for firstprivate(info) num_threads(omp_num_threads) default(shared) proc_bind(close) \
    schedule(static, 1)
    for (unsigned int wid = 0; wid < amount_of_work; ++wid)
    {
        info.thread_id = wid;
        workloads[wid](info);
    }
}
```

src/cpu/aarch64/matmul/kai_matmul.cpp
```diff
@@ -0,0 +1,509 @@
...
+    if (get_verbose(verbose_t::profile_externals)) {
+        std::cout << "profile_externals: " << _kernel->get_config().filter
+                  << std::endl;
+    }
+
+    const auto scratchpad = ctx.get_scratchpad_grantor();
+
+    const kai::ops::ndrange_t window_size = _kernel->get_window_size();
+    const int num_windows = static_cast<int>(window_size.total_size());
+    int num_threads = std::min(num_windows, dnnl_get_current_num_threads());
```
fadara01: See other comment about this.

fadara01 (commented):
This is awesome!

I didn't have a very close look, just commented on some concerns about changes in threading behavior

src/cpu/aarch64/kai_convolution.cpp
jondea: Excellent point, I have fixed this now.

cfRod:
Can you exclude all the kai implementations from threadpool using " VDISPATCH_MATMUL(DNNL_CPU_THREADING_RUNTIME != DNNL_RUNTIME_THREADPOOL,
            VERBOSE_UNSUPPORTED_RUNTIME);" until we can make them async compatible
            
 See this PR https://github.com/uxlfoundation/oneDNN/pull/4297

src/cpu/aarch64/matmul/kai_matmul.cpp
```diff
@@ -0,0 +1,605 @@
...
+            kernel_execute_work(*pd()), max_threads);
+
+    std::unique_ptr<kai::ops::IGemmCommon> _kernel = nullptr;
+    if (pd()->is_dequant()) {
+        DEFINE_ARG_SCALES_BUFFER(src_scale, DNNL_ARG_SRC);
+        DEFINE_ARG_SCALES_BUFFER(wei_scale, DNNL_ARG_WEIGHTS);
+        kai::ops::DequantizeFloat dequant(src_scale[0] * wei_scale[0]);
+        _kernel = pd()->create_kai_gemm_dequant(dequant, num_threads);
+    } else {
+        _kernel = pd()->create_kai_gemm(num_threads);
```
renato-arantes, line 459: I understand why you need to recreate the kernel inside `execute()` when quantisation is needed, but I think the `else` is redundant. If `num_threads` is needed, you can also get it inside `init` [here](https://github.com/uxlfoundation/oneDNN/blob/c03026f088f8d53ad53ed375930efcbf55dffb42/src/cpu/aarch64/matmul/kai_matmul.cpp#L323).

PR #4685, opened 2026-02-12
cpu: rv64: add-build-flag-for-rv64gc

zhangjian29:
# Description

Fixes https://github.com/uxlfoundation/oneDNN/issues/4638

This PR propose a temporary method to compile correctly on `rv64gc` platform. See detailed discussions in issue: https://github.com/uxlfoundation/oneDNN/issues/4638, PR: https://github.com/uxlfoundation/oneDNN/pull/4466 and https://github.com/uxlfoundation/oneDNN/pull/4626.

These build flags will be removed once runtime checks are ready for RV64.

# Checklist

- We can compile specific code for `rv64gc`, `rv64gcv`, and `rv64gcv_zvfh` arch now.

<details>
  <summary>logs</summary>

```bash
zhangjian@localhost:~/oneDNN/build_rvv (main) $ rm -rf * && cmake .. -DDNNL_TARGET_ARCH="RV64" -DDNNL_ARCH_OPT_FLAGS="-march=rv64gcv"
-- CMAKE_BUILD_TYPE is unset, defaulting to Release
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/zhangjian/tools/tool-gcc14.2/bin/riscv64-unknown-linux-gnu-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/zhangjian/tools/tool-gcc14.2/bin/riscv64-unknown-linux-gnu-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- DNNL_TARGET_ARCH: RV64
-- DNNL_LIBRARY_NAME: dnnl
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- Performing Test CAN_COMPILE_RVV_INTRINSICS
-- Performing Test CAN_COMPILE_RVV_INTRINSICS - Success
[...]

PR #5157, opened 2026-05-15
cpu: rv64: matmul: add brgemm matmul for rv64 platform

zhangjian29:
# Description

This PR introduces a BRGEMM-based matmul primitive for RISC-V RV64 using the existing JIT BRGEMM micro-kernel with copy_A packing.

This version provides:

1. BRGEMM matmul for broadcast weight inference (batch > 1, weights shared across batch)
2. copy_A packing that reads A matrix sequentially (stride bd_block=32) instead of strided (stride N), achieving ~2x DRAM throughput when A exceeds L2 cache compared to gemm kernel
3. Strict three-guard dispatch: K >= 4096, A exceeds L2 cache, batch * M >= 128

## Key Features

- **copy_A packing**: Transforms the weight matrix A from col-major stride N to contiguous stride bd_block per M-tile, enabling sequential DRAM reads. Packed once per M-tile for all K-blocks, amortizing packing cost.
- **K-blocking**: Uses BRGEMM_BK=256 K-blocking tiles, calling the JIT kernel per K-block with offset into the packed buffer.
- **Bias + post-ops**: Applied after BRGEMM computation via the existing rvv_postops infrastructure, supporting broadcast bias with arbitrary dimension matching.
- **Safe tail handling**: The pack_a_tile function uses a separate valid_rows parameter to avoid out-of-bounds reads on the tail M-tile.

### Dimension Mapping

Broadcast matmul `Y = X @ W` with X=[batch,M,K], W=[1,K,N] is mapped to BRGEMM as:

```
C(N x batch*M) = W^T(N x K) * X^T(K x batch*M)
  M_brg = N, K_brg = K, N_brg = batch * M
```

### Dispatch Conditions

Three guards ensure BRGEMM only fires when beneficial:

1. **K >= 4096**: K-blocking amortizes the per-tile packing overhead. Below this threshold, GEMM's existing copy_A is sufficient.
2. **N * K * 4 > L2 cache size**: When the A matrix (weights) exceeds L2 cache, GEMM's copy_A reads with stride N cause every access to miss cache and reload from DRAM. BRGEMM's sequential packing reads (stride bd_block=32) achieve ~2x higher throughput from main memory.
3. **batch * M >= 128**: The BRGEMM kernel's N-loop needs sufficient iterations (>= 32 at n_step=4) to pipeline efficiently. Tested: batch*M=32 causes -4% regression; batch*M=128 provides consistent improvement.

### Why BRGEMM Wins

When A exceeds L2 cache, GEMM reads A with stride N (e.g., 4096 floats) per K column. Two such rows fill the entire 32KB L1D cache, causing constant eviction. GEMM throughput drops to ~3-5 Gflops.

BRGEMM packs A into contiguous blocks with stride bd_block=32 per K element. Sequential reads from DRAM achieve ~7-9 Gflops, yielding 1.4-2.0x speedup.

For broadcast matmul specifically, the advantage is even larger: GEMM processes each batch element separately, re-loading the entire weight matrix from DRAM for each element. BRGEMM loads weights once and shares across all batch elements via the strided B panel, achieving **2.45x speedup** on GPT-J prefill shapes.

---

# Checklist

## General

## Performance improvements

All experiments are performed on Spacemit X60 (VLEN=256, L1D=32KB, L2=1MB) with single core (`taskset -c 32`). We draw comparisons between:

1. **RISCV64GCV** (upstream main - GEMM-based matmul using rvv intrinsics)
2. **brgemm:rvv** (this PR - BRGEMM-based matmul with copy_A packing)

### GPT-J Prefill Phase (3D Broadcast - Real LLM Workload)

The benchdnn GPT-J harness (`harness_matmul_gpt-j_2016-32_inf_lb_f32`) expresses LLM inference as 2D matmul (M flattened to batch*seq_len). In real LLM frameworks (PyTorch, TensorFlow), the same computation uses **3D broadcast matmul** where weights are shared across the batch dimension.

We reshape the GPT-J prefill shapes as 3D broadcast to match how real frameworks call oneDNN:

| GPT-J Operation | 2D Shape (original) | 3D Broadcast Shape | brgemm:rvv (ms) | RISCV64GCV (ms) | Speedup |
|-----------------|---------------------|--------------------|-----------------|-----------------|---------|
| gemm0: hidden FC | 48384x4096:4096x4096 | 128x32x4096:1x4096x4096 | 36,869 | 90,206 | **2.45x** |

**Why GEMM is 2.45x slower for broadcast matmul:**
- GEMM processes each of 128 batch elements as a separate M=32 GEMM call
- Each call re-loads the entire 64MB weight matrix from DRAM
- Total weight reads: 128 × 64MB = 8GB of redundant DRAM traffic
- BRGEMM loads weights once and shares across all 128 × 32 = 4096 output columns

### Synthesized LLM Inference Shapes

Additional shapes representing typical BERT/GPT-2/LLaMA broadcast matmul workloads:

| Shape | K | N | batch*M | brgemm:rvv (ms) | RISCV64GCV (ms) | Speedup |
|-------|---|---|---------|-----------------|-----------------|---------|
| gpt2_attn (16x40x4096:1x4096x4096) | 4096 | 4096 | 640 | 3,009 | 6,069 | **2.02x** |
| bert_output (128x1x4096:1x4096x3072) | 4096 | 3072 | 128 | 357 | 683 | **1.91x** |
| bert_intermediate (128x1x4096:1x4096x768) | 4096 | 768 | 128 | 104 | 153 | **1.47x** |
| gpt2_fc_up (128x1x4096:1x4096x4096) | 4096 | 4096 | 128 | 661 | 946 | **1.43x** |

**Summary for BRGEMM-dispatched shapes:**
- Average speedup: **1.71x** (synthesized), **2.45x** (GPT-J prefill)
- All BRGEMM shapes show significant improvement (43-145%)

### DIFF
--- a/src/cpu/rv64/rvv_brgemm_matmul.cpp
+++ b/src/cpu/rv64/rvv_brgemm_matmul.cpp
@@ -150,13 +150,13 @@ protected:
 // ---------------------------------------------------------------------------
 struct jit_bias_postops_row_t : public jit_generator_t {
     // bias_type: 0=none, 1=scalar broadcast, 2=per-element vector
-    // postop: 0=none, 1=relu
     struct call_params_t {
         float *row_dst; // offset 0
         dim_t N; // offset 8
         const float *bias_ptr; // offset 16
         int32_t bias_type; // offset 24
         int32_t postop; // offset 28
+        float postop_alpha; // offset 32
     };
 
     DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_bias_postops_row_t)
@@ -185,9 +185,12 @@ protected:
         const Reg reg_postop = t4;
 
         const FReg f_bias_scalar = fa0;
-        const FReg f_zero = fa1;
-        const VReg v_acc(0);
-        const VReg v_bias(1);
+        const FReg f_alpha = fa1;
+        const FReg f_zero = fa2;
+        const VReg v_acc(1);
+        const VReg v_bias(2);
+        const VReg v_neg(3);
+        const VReg v_orig(4);
 
         // Load parameters
         ld(reg_dst, reg_param, 0);
@@ -196,7 +199,9 @@ protected:
         lw(reg_bias_type, reg_param, 24);
         lw(reg_postop, reg_param, 28);
 
-        // Prepare f_zero = 0.0 for ReLU
+        // Load alpha for Leaky ReLU
+        flw(f_alpha, reg_param, 32);
+
         xor_(reg_tmp, reg_tmp, reg_tmp);
         fmv_w_x(f_zero, reg_tmp);
 
@@ -235,10 +240,19 @@ protected:
 
         L(after_bias);
 
-        // Post-op: ReLU (max(acc, 0))
-        Label postop_done;
+        // Post-op: ReLU with negative slope alpha
+        Label postop_done, leaky_relu;
         beqz(reg_postop, postop_done);
+        vmv_v_v(v_orig, v_acc);
+        fmv_x_w(reg_tmp, f_alpha);
+        bnez(reg_tmp, leaky_relu);
         vfmax_vf(v_acc, v_acc, f_zero);
+        j_(postop_done);
+
+        L(leaky_relu);
+        vmfgt_vf(VReg(0), v_orig, f_zero);
+        vfmul_vf(v_neg, v_orig, f_alpha);
+        vmerge_vvm(v_acc, v_neg, v_orig);
         L(postop_done);
 
         // Store result
@@ -494,11 +508,13 @@ status_t rvv_brgemm_matmul_t::execute(const exec_ctx_t &ctx) const {
     const dim_t *src_dims_ptr = src_d.dims();
     const dim_t dst_batch_stride = M * N;
 
-    // Determine postop type for JIT kernel
     int32_t postop = 0;
+    float postop_alpha = 0.f;
     if (post_ops.len() > 0 && post_ops.entry_[0].is_eltwise()
-            && post_ops.entry_[0].eltwise.alg == alg_kind::eltwise_relu)
+            && post_ops.entry_[0].eltwise.alg == alg_kind::eltwise_relu) {
         postop = 1;
+        postop_alpha = post_ops.entry_[0].eltwise.alpha;
+    }
 
     parallel_nd(batch, [&](dim_t b) {
         float *dst_base = dst + b * dst_batch_stride;
@@ -553,6 +569,7 @@ status_t rvv_brgemm_matmul_t::execute(const exec_ctx_t &ctx) const {
             bp.bias_ptr = bias_ptr;
             bp.bias_type = bias_type;
             bp.postop = postop;
+            bp.postop_alpha = postop_alpha;
             (*bias_postops_kernel)(&bp);
         }
     });
--- a/src/cpu/rv64/rvv_brgemm_matmul.cpp
+++ b/src/cpu/rv64/rvv_brgemm_matmul.cpp
@@ -1,0 +1,348 @@
+/*******************************************************************************
+* Copyright 2026 ZTE Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "common/c_types_map.hpp"
+#include "common/dnnl_thread.hpp"
+#include "common/memory_desc_wrapper.hpp"
+#include "common/memory_tracking.hpp"
+#include "common/nstl.hpp"
+#include "common/type_helpers.hpp"
+#include "common/utils.hpp"
+#include "common/verbose.hpp"
+
+#include "cpu/platform.hpp"
+#include "cpu/rv64/rvv_brgemm_matmul.hpp"
+#include "cpu/rv64/rvv_postops.hpp"
+
+#include <riscv_vector.h>
+
+namespace dnnl {
+namespace impl {
+namespace cpu {
+namespace rv64 {
+namespace matmul {
+
+using namespace dnnl::impl::status;
+using namespace dnnl::impl::utils;
+using namespace data_type;
+
+// Pack bd rows × K_inner columns from col-major A (with LDA_orig) into
+// contiguous workspace with LDA=bd.
+static void pack_a_tile(float *ws, const float *A, dim_t LDA_orig, dim_t bd,
+        dim_t valid_rows, dim_t K_inner) {
+    for (dim_t k = 0; k < K_inner; k++) {
+        const float *A_col = A + k * LDA_orig;
+        float *ws_row = ws + k * bd;
+        dim_t i = 0;
+        while (i < valid_rows) {
+            size_t vl = __riscv_vsetvl_e32m1(valid_rows - i);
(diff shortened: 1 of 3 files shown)


dzarukin (approved):

src/cpu/rv64/rvv_brgemm_matmul.cpp
```diff
@@ -0,0 +1,348 @@
...
+#include "common/nstl.hpp"
+#include "common/type_helpers.hpp"
+#include "common/utils.hpp"
+#include "common/verbose.hpp"
+
+#include "cpu/platform.hpp"
+#include "cpu/rv64/rvv_brgemm_matmul.hpp"
+#include "cpu/rv64/rvv_postops.hpp"
+
+#include <riscv_vector.h>
```
xiazhuozhao: Could we make more of this path JIT-based where practical? 

I checked the current f32 coverage under src/cpu/rv64: binary, layernorm, matmul/gemm, several conv paths, GEMM/BRGEMM IP, and part of softmax are already JIT or JIT-backed, so we are quite close to finishing the JIT migration.

zhangfeiv0: > Could we make more of this path JIT-based where practical?
> 
> I checked the current f32 coverage under src/cpu/rv64: binary, layernorm, matmul/gemm, several conv paths, GEMM/BRGEMM IP, and part of softmax are already JIT or JIT-backed, so we are quite close to finishing the JIT migration.

I agree with this opinion.

zhangjian29:
Hi @zhangfeiv0 @xiazhuozhao ,

Thank you for your suggestion. Changes made.

By the way, are we going to abandon RVV intrinsics and replace them with JIT style completely?

xiazhuozhao:
> Hi @zhangfeiv0 @xiazhuozhao ,
> 
> Thank you for your suggestion. Changes made.
> 
> By the way, are we going to abandon RVV intrinsics and replace them with JIT style completely?

It looks like we may still have a small misunderstanding here, so I checked the previous discussions again.

My understanding is that #4685 fixed the issue where the build result can be made consistent with a target that does not support the V extension. However, it does not fully solve the broader problem that the same build artifact should both accelerate on platforms with V and still run correctly on platforms without V.

In the earlier discussions, we also mentioned JIT / runtime detection as the longer-term direction, with the CMake-based detection eventually being removed:

https://github.com/uxlfoundation/oneDNN/pull/4626#issuecomment-3836397073

https://github.com/uxlfoundation/oneDNN/issues/4638#issuecomment-3845512663

https://github.com/uxlfoundation/oneDNN/issues/4638#issuecomment-3862612172

So my current understanding is not that RVV intrinsics must be completely abandoned immediately. Rather, #4685 is an effective interim fix for the build-flag consistency problem, while the longer-term direction is still runtime dispatch / runtime CPU feature detection. In that direction, using JIT or JIT-backed implementations for new optimized RV64 hot paths, where practical, seems more aligned with the previous discussions.

Of course, this is just my interpretation of the previous discussions. If I missed something or misunderstood the direction, we can discuss it here.

zhangjian29:
Hi @xiazhuozhao ,

I agree with what we discussed before in #4685 #4626. To my understanding, we need `mayiuse(v)` runtime check everywhere we use vector extensions.

> I checked the current f32 coverage under src/cpu/rv64: binary, layernorm, matmul/gemm, several conv paths, GEMM/BRGEMM IP, and part of softmax are already JIT or JIT-backed, so we are quite close to finishing the JIT migration.

So RVV instrinsics are okay if they are under correct checking of  `mayiuse(v)` and other types of runtime check. Is that true to you?

xiazhuozhao:
> Hi @xiazhuozhao ,
> 
> I agree with what we discussed before in #4685 #4626. To my understanding, we need `mayiuse(v)` runtime check everywhere we use vector extensions.
> 
> > I checked the current f32 coverage under src/cpu/rv64: binary, layernorm, matmul/gemm, several conv paths, GEMM/BRGEMM IP, and part of softmax are already JIT or JIT-backed, so we are quite close to finishing the JIT migration.
> 
> So RVV instrinsics are okay if they are under correct checking of `mayiuse(v)` and other types of runtime check. Is that true to you?

I think this is true only if the RVV code is also isolated from the baseline compilation path.

mayiuse(v) is necessary, but it is not sufficient if a large part of the library is built with -march=rv64gcv_zvfh. In that case, the compiler may emit V instructions automatically for optimization, even in code that is not explicitly using RVV intrinsics or guarded by mayiuse(v).

That was my understanding of #4638 as well: the failures were broad, not limited to some RVV-optimized primitives, which suggests the generated binary itself was not safe for non-V platforms, not only the optimized path, even if it's protected.

On the other hand, without compiling the relevant code with a V-enabled -march, code that uses RVV intrinsics cannot be compiled successfully.

So I agree RVV intrinsics are acceptable when properly guarded, but the V-specific code also needs to be isolated, for example via JIT, function attributes, or separate compilation units. This is why I think JIT/JIT-backed implementations are preferable for new optimized f32 hot paths where practical.

zhangjian29:
Yes, I agree. Let's work on other primitives without guards of JIT. Thank you for your comments @xiazhuozhao .

src/cpu/rv64/rvv_brgemm_matmul.cpp
```diff
@@ -0,0 +1,567 @@
...
+        // Advance bias pointer for next iteration
+        slli(reg_bytes, reg_vl, 2);
+        add(reg_bias, reg_bias, reg_bytes);
+
+        L(after_bias);
+
+        // Post-op: ReLU (max(acc, 0))
+        Label postop_done;
+        beqz(reg_postop, postop_done);
+        vfmax_vf(v_acc, v_acc, f_zero);
```
xiazhuozhao: It accepts eltwise_relu:0.1, but actually executes it as eltwise_relu:0. This causes the following command to produce incorrect results:
```
OMP_NUM_THREADS=1 DNNL_VERBOSE=1 ./benchdnn \
  --matmul --mode=C \
  --dt=f32 --stag=abc --wtag=abc --dtag=abc \
  --attr-post-ops=relu:0.1 \
  128x1x8192:1x8192x64
```

Although this issue seems to have also existed in the previous matmul implementation.

xiazhuozhao: The old path in src/cpu/rv64/rvv_postops.hpp:139 also directly uses `return __riscv_vfmax_vv_f32m1(v, zero, vl);` and does not use alpha either.

src/cpu/rv64/rvv_brgemm_matmul.cpp
```diff
@@ -0,0 +1,574 @@
...
+        add(reg_bias, reg_bias, reg_bytes);
+
+        L(after_bias);
+
+        // Post-op: Leaky ReLU f(x) = max(x, x * alpha)
+        Label postop_done;
+        beqz(reg_postop, postop_done);
+        vmv_v_v(v_orig, v_acc);
+        vfmul_vf(v_acc, v_acc, f_alpha);
+        vfmax_vv(v_acc, v_acc, v_orig);
```
xiazhuozhao: oneDNN’s ReLU semantics are not `max(x, alpha * x)` in the general case. They are:

```cpp
x > 0 ? x : alpha * x
```

You can refer to [this code](https://github.com/uxlfoundation/oneDNN/blob/9a7e1ac215b0f571749dff8a5d8373cb3922a618/src/common/math_utils.hpp#L149).

With the current fix, the result is correct when `alpha <= 1`, but it is still incorrect when `alpha > 1`. 

Image

For example, the following command still fails:

```bash
OMP_NUM_THREADS=1 DNNL_VERBOSE=1 ./benchdnn \
  --matmul --mode=C \
  --dt=f32 --stag=abc --wtag=abc --dtag=abc \
  --attr-post-ops=relu:2 \
  128x1x8192:1x8192x64
```

This is the same reproducer as before, with the alpha value changed to `2`.

This issue also exists in the previous RVV post-op implementation, which may have been misleading here.

xiazhuozhao: Hi, i fixed this. Do you want to have a try?
[fix.patch](https://github.com/user-attachments/files/28109846/fix.patch)

xiazhuozhao (approved):

PR #5158, opened 2026-05-15
Use strides to pick GEMV strategy robustly

densamoilov:
Fixes [MFDNN-15036](https://jira.devtools.intel.com/browse/MFDNN-15036).

For some degenerate GEMV cases in MatMul such as `1x1:1x200` we couldn't reliably identify the tag for matrix B and always picked `ab`. This led to the wrong GEMV strategy and an incorrect LDA since A and B are swapped on this path and LDA is inferred from B's strides. This PR uses strides to inspect user-provided memory descriptors and pick the right GEMV strategy. We still have to derive the tag from the inspected strides because the current MatMul design requires it but the choice should now be more robust for GEMV cases.

[Nightly](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f151a05d-7127-f1ef-a760-a4bf010d0e2d)
[CI](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f1507f6c-a754-f1ce-9b72-a4bf010d0e2d?objectId=job-f1507f6c-a754-f1ce-9b72-a4bf010d0e2d&filterName0=jobsPageSearch&filterDepth=1)

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -396,7 +396,7 @@ gemv_strategy_t brgemm_matmul_conf_utils_t::get_gemv_strategy(
 /**
  * Select a compatible format for A in GEMV cases.
  *
- * If A has format is "any", the function selects plain layout.
+ * If A has format "any", the function selects plain layout.
  * Otherwise, it checks if the provided format is compatible.
  *
  * Returns a valid format tag on success, or format_tag::undef otherwise.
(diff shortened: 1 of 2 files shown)


dzarukin (approved):

xuxinzen (approved):

Sqvid:
Given the similarity between our paths could I request a `benchdnn` reproducer so I can investigate if the fix needs porting to `cpu/aarch64`? Thanks.

dzarukin:
> Given the similarity between our paths could I request a `benchdnn` reproducer so I can investigate if the fix needs porting to `cpu/aarch64`? Thanks.

Hi @Sqvid, there you go:

```
# test a degenerate gemv case when K=1
--reset --stag=ab --wtag=ba --dtag=ab 1x1:1x200
```

For the future, the usual practice is to add a case that led to the problem into a regression test file and they can be found in the PR.

Sqvid:
Absolutely no idea how I missed it with only two files :facepalm: My bad, thanks!

PR #5166, opened 2026-05-18
xe: sycl: use gpu_eu_count extension query for EU count

echeresh:
This is a workaround for [GSD-12286](https://jira.devtools.intel.com/browse/GSD-12286).

`max_compute_units` query may report an incorrect EU count with SYCL/L0 on SKUs with reduced EU count. This may cause regressions and potentially hangs (though we didn't observe hangs in practice).


echeresh:
make test

atkassen (approved):

Simonsays095 (approved):

PR #5171, opened 2026-05-19
cpu: x64: matmul: fixed undefined behavior for trivial dimensions case

densamoilov:
Forward port of #5047

Supporting odd sizes in the non-transpose copy routine is a feature that hasn't been implemented yet and not a bug. Unless we get a feature request to support it we don't need to add it.


vpirogov (approved):

densamoilov:
make test
disable test_device_gpu

ankalinin (approved):

PR #5173, opened 2026-05-20
rfcs: graph api scratchpad management

TaoLv:
This RFC describes the current scratchpad memory management design in oneDNN's Graph API execution, identifies its incompatibilities with SYCL command graph capture and asynchronous thread pool runtime, and proposes alternative designs.

The rendered version: [link](https://github.com/uxlfoundation/oneDNN/blob/lvtao/rfcs/graph-api-spad-mgmt/rfcs/20260520-graph-api-scratchpad/README.md) .

### DIFF
--- a/rfcs/20260520-graph-api-scratchpad/README.md
+++ b/rfcs/20260520-graph-api-scratchpad/README.md
@@ -1,0 +1,332 @@
+# RFC: Scratchpad Memory Management for Graph API
+
+## Summary
+
+This RFC describes the current scratchpad memory management design in oneDNN's
+Graph API execution, identifies its incompatibilities with SYCL command graph
+capture and asynchronous thread pool runtime, and proposes alternative designs.
+
+## 1. Current Design
+
+### 1.1 Allocation and Deallocation Timing
+
+Each graph partition allocates scratchpad memory per execution call. The
+scratchpad memory will be deallocated when the executioncall exits. Take the
+`larger_partition_kernel_t` as an example:
+
+```cpp
+status_t larger_partition_kernel_t::sycl_execute_impl(...) {
+    // RAII. Allocated at the start of each execute call
+    temporary_scratchpad_t scratchpad(
+            memory_planner_.total_internal_temporary_size(), p_engine_, *g_alloc_);
+
+    // Scratchpad is divided and reused as arguments of executables.
+    prepare_args_set(res, inputs, outputs, scratchpad);
+
+    // Loop over executables/primitives.
+    for (size_t i = 0; i < subgraph_->execs_.size(); i++) {
+        returned_event = subgraph_->execs_[i]->execute_sycl(
+                p_stream, res->get_exec_args()[i], deps);
+    }
+
+    scratchpad.set_deps(returned_event);
+    // Deallocated here when scratchpad goes out of scope.
+    // The event dependency ensures the GPU finishes before the
+    // user-provided deallocator frees the device memory.
+}
+```
+
+### 1.2 Key Characteristics
+
+- The user-provided allocator (via the `allocator` and `engine` API) manages the
+  actual device memory. oneDNN merely requests and releases buffers.
+- Scratchpad allocation happens at the beginning of each `execute()` /
+  `sycl_execute_impl()` call.
+- Scratchpad deallocation happens at the end of each `execute()` call. The
+  `temporary_scratchpad_t` destructor passes a dependency event (if applicable)
+  to the user's deallocator so that asynchronous device work completes before
+  memory is freed.
+- The scratchpad object is scoped to a single `execute()` invocation. The
+  underlying device memory lifetime can be prolonged by the dependency event.
+- Each thread allocates its own scratchpad (stack-local), so concurrent
+  `execute()` calls on the same partition are safe.
+- The memory planning stage at partition compilation computes a flat, contiguous
+  scratchpad buffer. Intermediate tensors inside the partition and scratchpad
+  memories required by underlying primitives are assigned offsets within that
+  buffer.
+- In most cases, the scratchpad size is determined at partition compilation
+  time. When needed, the size can also be adjusted at execution time before the
+  scratchpad object is created.
+
+### 1.3 Pros and Cons
+
+**Pros:**
+
+- Thread-safe: Stack-local allocation means concurrent `execute()` calls on the
+  same compiled partition are inherently safe without any synchronization.
+- Minimal memory footprint: Scratchpad memory is held only during execution.
+  Idle compiled partitions consume zero temporary memory.
+- Honors allocator contract: Memory is allocated through allocator API. Users
+  can implement pool-based or per-device allocators externally and count memory
+  consumption if they wish. Memory alignment is passed to the allocator as part
+  of the interface.
+- Simple implementation: No cache management, eviction policy, or lifetime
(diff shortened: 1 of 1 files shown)


LuFinch:
Allocating persist scratchpad memory may reduce memory reusing and reserve too much memory by OneDNN graph.

Will you consider to combine your main proposal with alternative B?

If user provide allocated device memory for scratchpad tensor, then user should manage the memory live and thread-safe. 

If user doesn't provide allocated memory for scratchpad, then use the the memory allocating function to allocate persist scratchpad memory during sycl graph capture.

TaoLv:
> If user provide allocated device memory for scratchpad tensor, then user should manage the memory live and thread-safe.
> 
> If user doesn't provide allocated memory for scratchpad, then use the the memory allocating function to allocate persist scratchpad memory during sycl graph capture.

Thank you for the review, @LuFinch .  It means we will need to provide two different ways to support SYCL graph capturing which unnecessarily complicate the library's design and implementation. Can you please elaborate why you want both and when each of them will be used? Thanks.

LuFinch:
@TaoLv I have discussed with Eikan. We prefer to use Alternative B because this is best way to manage memory. Allocating persist scratchpad memory and not releasing memory is a kind of workaround. 

If you could consider Alternative B, that's what we need. If you insist the main proposal, please also provide Alternative B implementation.

TaoLv:
> @TaoLv I have discussed with Eikan. We prefer to use Alternative B because this is best way to manage memory. Allocating persist scratchpad memory and not releasing memory is a kind of workaround.
> 
> If you could consider Alternative B, that's what we need. If you insist the main proposal, please also provide Alternative B implementation.

Hi @LuFinch, I also feel the proposal B is more future proof. I updated the recommendation section to reflect that. Please check.

LuFinch:
LGTM now

rfcs/20260520-graph-api-scratchpad/README.md
```diff
@@ -0,0 +1,310 @@
...
+sycl_interop::execute(cp, stream, inputs, outputs, spad, deps);
+```
+
+**Pros:**
+
+- Zero internal allocation overhead.
+- User has full ownership over scratchpad lifetime, sharing, and memory
+  placement.
+- User ensures the buffer is valid and stable for SYCL graph capture.
+- User ensures the lifetime for async CPU thread pool runtime.
```
dzarukin: Let's combine three-in-one: 
* The user gains total control over scratchpad memory including lifetime management and memory properties (alignment, placement, etc.)

rfcs/20260520-graph-api-scratchpad/README.md
```diff
@@ -0,0 +1,310 @@
...
+- Matches established patterns (cuDNN workspace, oneDNN primitive user mode
+  scratchpad).
+
+**Cons:**
+
+- API change: Requires new public API surface.
+- User burden: User must manage buffer lifetime and concurrency.
+- Requires pre-calculated scratchpad size: In some cases the library cannot
+  determine the scratchpad size at compilation time.
+- Allocator contract: for example, the memory alignment requirement.
```
dzarukin: Could you, please, expand this bullet? I can't see what you mean and why it's a con.

rfcs/20260520-graph-api-scratchpad/README.md
```diff
@@ -0,0 +1,310 @@
...
+- User ensures the buffer is valid and stable for SYCL graph capture.
+- User ensures the lifetime for async CPU thread pool runtime.
+- Matches established patterns (cuDNN workspace, oneDNN primitive user mode
+  scratchpad).
+
+**Cons:**
+
+- API change: Requires new public API surface.
+- User burden: User must manage buffer lifetime and concurrency.
+- Requires pre-calculated scratchpad size: In some cases the library cannot
```
dzarukin: It shouldn't be the case. If it happens for some reason, the error should be reported and escalated.

rfcs/20260520-graph-api-scratchpad/README.md
```diff
@@ -0,0 +1,310 @@
...
+  placement.
+- User ensures the buffer is valid and stable for SYCL graph capture.
+- User ensures the lifetime for async CPU thread pool runtime.
+- Matches established patterns (cuDNN workspace, oneDNN primitive user mode
+  scratchpad).
+
+**Cons:**
+
+- API change: Requires new public API surface.
+- User burden: User must manage buffer lifetime and concurrency.
```
dzarukin: I'd remove this from cons. Control always has a price, but it's not a con.

rfcs/20260520-graph-api-scratchpad/README.md
```diff
@@ -0,0 +1,310 @@
...
+  determine the scratchpad size at compilation time.
+- Allocator contract: for example, the memory alignment requirement.
+
+**Backward compatibility:**
+
+If scratchpad buffer is passed to `execute()`, it will be used for kernel
+executions. The library will skip the internal temporary sctachpad allocation.
+
+If no scratchpad buffer is passed to `execute()`, the library falls back to the
+current behavior: creating a temporary scratchpad object and allocating through
```
dzarukin, line 304: I suggest to allow such fallback till the next major release. We can drop it once API compatibility is not promised. I assume that allocator API can be dropped with this change as well. Additionally, I'd allow such behavior for synchronous runtimes only and return an error if for asynchronous runtime scratchpad is not specified.

rfcs/20260520-graph-api-scratchpad/README.md
TaoLv: I was thinking about the runtime dims.

rfcs/20260520-graph-api-scratchpad/README.md
TaoLv: I was thinking that we could enforce memory attributes through the allocator API. One example is the memory alignment requirement. Misaligned memory can cause performance penalty. Currently we pass the alignment to the user through the allocator API and the user need to fulfill this requirement when returning a buffer to the library. With passing a scratchpad buffer directly to `execute()` API, we will lose this kind of enforcement. Alternatively, we can document the memory alignment requirement or request a bit more memory in `get_scratchpad_size()`.

dzarukin: I think a better way of handling that would be checking alignment at `execute(...)` and issue a verbose warning message if it doesn't satisfy expected alignment. If the user controls the memory, the control is complete, not partial.

rfcs/20260520-graph-api-scratchpad/README.md
dzarukin: I'm ware of a single implementation that violates it, and this implementation is about to die soon. In other cases, the kernel should operate with blocks with definitive scratchpad size required for a block resulting in a predicted size at create level.

rfcs/20260520-graph-api-scratchpad/README.md
TaoLv: I updated the document to reflect this discussion.

rfcs/20260520-graph-api-scratchpad/README.md
TaoLv: I updated the document to reflect this discussion.

dzarukin (approved):

ElaineBao (approved):

PR #5176, opened 2026-05-20
[GEMM] "Unknown" platform fix

Simonsays095:
Addresses [MFDNN-15080](https://jira.devtools.intel.com/browse/MFDNN-15080). If we are unable to query the GPU platform with `CL_DEVICE_IP_VERSION_INTEL`, we fall back to querying a dummy kernel for HW versions injected by IGC. In this case, the platform type is left as `ngen::PlatformType::Unknown`, even when we know whether it's integrated or discrete from this information.

Then, when searching the catalog for strategies, the "Unknown" platform can select strategies that have been restricted to discrete-only due to the tag matching logic.

This PR fixed both of these bugs:
- Deduce platform type from binary scraping path
- Introduce a tag-matching feature to enforce integrated/discrete strategy restrictions on unknown platforms.


Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

third_party/ngen/ngen_elf.hpp
echeresh: Should we initialize `outProduct.type = ProductFamily::unknown` as well, just in case?

src/gpu/intel/gemm/jit/include/gemmstone/kernel_selector.hpp
```diff
@@ -57,7 +57,7 @@ struct MatchParamsBase
     const StrategyRequirement *extraReqs = nullptr;
 
     MatchParamsBase() = default;
-    MatchParamsBase(ngen::HW hw, bool systolicAvailable, bool isIntegrated, const GEMMProblem &problem);
+    MatchParamsBase(ngen::HW hw, bool systolicAvailable, bool isIntegrated, bool isDiscrete, const GEMMProblem &problem);
```
echeresh: Nit: could probably switch to `ngen::PlatformType` here (and add a getter in the device info). It can replace these two boolean flags.

echeresh (approved):

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
```diff
@@ -420,6 +424,10 @@ MatchParamsBase::MatchParamsBase(ngen::HW hw, bool systolicAvailable, bool isInt
         *tagPtr++ = ReqSystolic;
 
     if (isIntegrated) *tagPtr++ = ReqIntegrated;
+    if (!isIntegrated && !isDiscrete) {
+        *tagPtr++ = ReqIntegrated;
+        *tagPtr++ = ReqNoIntegrated;
```
kealan-barbieri: Why cant we just disable discrete strategies in this case?

src/gpu/intel/gemm/jit/gen_kernel.hpp
```diff
@@ -140,11 +140,11 @@ struct gen_nocopy_desc_t : public gen_desc_t {
     }
 
     std::vector<const gemmstone::kcatalog::Entry *> select_kernel(
-            compute::gpu_product_t product, int stepping, int eu_count,
-            bool has_systolic, bool is_integrated, compute_mode mode,
-            const gemmstone::GEMMProblem &problem, float alpha, float beta,
-            dim_t m, dim_t n, dim_t k, dim_t lda, dim_t ldb, dim_t ldc,
-            dim_t batch);
+            compute::gpu_product_t product, int stepping,
+            const compute::device_info_t &dev_info, bool has_systolic,
```
kealan-barbieri: can we also drop `has_systolic` in favor of `dev_info.mayiuse_systolic()`?

third_party/ngen/ngen_opencl.hpp
```diff
@@ -352,9 +352,11 @@ Product OpenCLCodeGenerator<hw>::detectHWInfo(cl_context context, cl_device_id d
         product = ELFCodeGenerator<hw>::getBinaryHWInfo(binary);
     }
 
-    cl_bool integrated;
-    if (dynamic::clGetDeviceInfo(device, CL_DEVICE_HOST_UNIFIED_MEMORY, sizeof(integrated), &integrated, nullptr) == CL_SUCCESS)
-        product.type = integrated ? PlatformType::Integrated : PlatformType::Discrete;
+    if (product.type == PlatformType::Unknown) {
```
kealan-barbieri, line 354: is this check preventing `product.type` from being overwritten with a different value initialized before?

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
Simonsays095: That's the goal with this change - disable discrete except where we know we're on a discrete platform, and same for integrated. The issue is that there is no tag matching the "unknown" platform, so this is the simplest pattern I came up with to detect that case.

third_party/ngen/ngen_opencl.hpp
Simonsays095: Yeah - this query is not a theoretically perfect indicator of platform type. If we can get the platform type directly from the IP version or the dummy binary, then we should use that instead.

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
Simonsays095: Just to clarify a bit... The way tag matching is confusing, since it's asymmetric - the entries can have upper and lowercase tags, but so far the patterns (from `MatchParams`) only have uppercase tags.
- To use integrated entries: Add `ReqIntegrated` to the pattern tags.
- To use discrete entries: Don't add `ReqIntegrated` to the pattern tags.

But this means that discrete entries are used for unknown platforms as well. The "contradiction pattern" I introduce in this PR is the first time we're using lowercase tags in the pattern, so it doesn't affect existing logic. Now the logic is:
- To use integrated entries: Add `ReqIntegrated` to the pattern tags.
- To use discrete entries: Don't add `ReqIntegrated` to the pattern tags.
- To disable both (unknown platform type): Add `ReqIntegrated` and `ReqNoIntegrated` to the pattern tags.

src/gpu/intel/gemm/jit/include/gemmstone/kernel_selector.hpp
Simonsays095: Thanks, I've updated the signature. I decided to just pass in the product.

src/gpu/intel/gemm/jit/gen_kernel.hpp
Simonsays095: Good idea, thanks. I've also removed the `product` and `stepping` arguments, since they also come from `dev_info`.

third_party/ngen/ngen_elf.hpp
Simonsays095: I just added default initializers to `ngen::Product` and removed it here, since the default values are well-defined.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
h-sadia: Can't you come up with a new tag for Discrete? Having `ReqIntegrated` and `ReqNoIntegrated` together is kind of confusing.

Simonsays095: We could add a `ReqDiscrete` tag, although that would require editing all existing entries with `ReqNoIntegrated` to have the new `ReqDiscrete` tag instead. I don't have a strong preference either way (other than having to rewrite the PR) - I can change it if you feel strongly.

rjoursler (approved):

h-sadia (approved):

src/gpu/intel/gemm/jit/selector/kernel_selector.cpp
Simonsays095: I went ahead and changed this, since it makes more sense - if discrete, use discrete kernels; if integrated, use integrated kernels; if neither (unknown), skip both.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

PR #5177, opened 2026-05-20
gpu: refactor gpu_primitive_attr_t to grf per thread

Simonsays095:
The `gpu_primitive_attr_t` was created a time when there were only two modes (both limited physically by 1024 GRF/EU):
- 4 threads/EU -- 256 GRF/thread
- 8 threads/EU -- 128 GRF/thread

At the time this feature was originally developed, it was not clear whether threads/EU or GRF/thread would be better. Nowadays it's clear that GRF/thread is more uniformly used in the library (likely because it separates software requirements from hardware capabilities), so it doesn't make sense to switch to threads/EU for this object then back to GRF/thread where it's used.

This PR simplifies the feature by sticking to GRF/thread everywhere. This is the first step in fixing a few features related to Variable Registers per Thread (VRT) on Xe3p.

### DIFF
--- a/src/gpu/intel/compute/kernel_ctx.hpp
+++ b/src/gpu/intel/compute/kernel_ctx.hpp
@@ -179,7 +179,7 @@ private:
         if (attr && attr->gpu_attr_) {
             auto *gpu_attr = utils::downcast<gpu_primitive_attr_t *>(
                     attr->gpu_attr_.get());
-            if (gpu_attr->threads_per_eu() == 4) {
+            if (gpu_attr->grf_per_thread() > 128) {
                 add_option("-cl-intel-256-GRF-per-thread");
             }
         }
--- a/src/gpu/intel/rnn/grid.cpp
+++ b/src/gpu/intel/rnn/grid.cpp
@@ -151,7 +151,7 @@ static status_t init_layouts_data(offsets_t &off,
 }
 
 static status_t init_ocl_conf(ocl_conf_t &ocl_conf, const pd_t *pd,
-        const conf_t &conf, int threads_per_eu,
+        const conf_t &conf, int grf_per_thread,
         const compute::device_info_t &device_info, offsets_t &off) {
 
     const memory_desc_wrapper &src_iter_c_d = pd->src_md(2);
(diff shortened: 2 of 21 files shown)


src/gpu/intel/compute/kernel_ctx.hpp
```diff
@@ -179,7 +179,7 @@ class kernel_ctx_t {
         if (attr && attr->gpu_attr_) {
             auto *gpu_attr = utils::downcast<gpu_primitive_attr_t *>(
                     attr->gpu_attr_.get());
-            if (gpu_attr->threads_per_eu() == 4) {
+            if (gpu_attr->grfs_per_thread() > 128) {
```
echeresh: `-cl-intel-256-GRF-per-thread` vs `grfs_per_thread` might be confusing to distinguish: GRF vs GRFs

I would rename it to either:

- `grf_per_thread` to align with the OpenCL option or
- `grf_count` - shorter, also self-descriptive, aligned with zeinfo

src/gpu/intel/rnn/grid.cpp
```diff
@@ -772,12 +772,13 @@ status_t simple_common_t<aprop>::pd_t::init(impl::engine_t *engine) {
...
+        if (grfs_per_thread == 0 || verbose) {
             auto t = 0;
-            auto s = gemm_pd->query(query::preferred_gpu_threads_per_eu, 0, &t);
-            if (threads_per_eu == 0)
-                threads_per_eu = (status::success != s) ? t : 128;
-            if (verbose && t != threads_per_eu)
+            auto s = gemm_pd->query(
+                    query::preferred_gpu_grfs_per_thread, 0, &t);
+            if (grfs_per_thread == 0)
+                grfs_per_thread = (status::success != s) ? t : 128;
```
echeresh: Looks like a preexisting bug:
```suggestion
                grfs_per_thread = (status::success == s) ? t : 128;
```

echeresh (approved):

src/gpu/intel/rnn/grid.cpp
Simonsays095: It was setting `threads_per_eu = 128` before as well, which was obviously wrong. Thanks for the catch.

src/gpu/intel/compute/kernel_ctx.hpp
Simonsays095: Changed to `grf_per_thread`. I like that the units are encoded in the name - I find that it makes the meaning much more obvious. e.g. `grf_count` could mean "per EU" or "per slice" as well as "per thread".

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

dzarukin (approved):

PR #5178, opened 2026-05-20
benchdnn: styling and minor changes

dzarukin:
This PR mostly updates stylistic part.

One notice is that continuation of #5118 which should improve data filling for mode=F requires careful handling of mapping as regular approach is to keep all memories mapped by default but no_ref_memory modifier changes that default. It means that to update the library memory objects, they must be either:
1. directly filled with a dedicated GPU-side kernel (as it's done now with `gpu_fill_random`), or
2. explicitly mapped if `dnn_mem_t::sel_elem(...)` interface is used, or
3. filled through reorder.

The global idea is to get rid of approach 2 in benchdnn (as it creates a lot of extra code to explicitly map and unmap an access spot), use approach 1 in cases when it doesn't lead to perf issues (read, data compression on GPU by its driver), which wouldn't do any mapping, and use approach 3 when the data must be controlled, with mapping happening in the common for benchdnn spot - execute_reorder function which is the essense of filling for all library memory objects.

There will be PRs that update approach 2 with approach 3. The first commit is presented here.

### DIFF
--- a/tests/benchdnn/matmul/bench_matmul.cpp
+++ b/tests/benchdnn/matmul/bench_matmul.cpp
@@ -127,7 +127,6 @@ int verify_input(const settings_t &s, const settings_t &def) {
     return OK;
 }
 
-#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
 // Validate input consistency for grouped encoding
 //
 // Notes:
@@ -135,7 +134,7 @@ int verify_input(const settings_t &s, const settings_t &def) {
 //   DST, so we only validate SRC parameters
 // - Currently only M dimension grouping is supported
 int verify_grouped_input(const settings_t &s) {
-    if (s.sparse_options[0].get_encoding(DNNL_ARG_SRC) == dnnl_grouped) {
+    if (s.sparse_options[0].is_grouped(DNNL_ARG_SRC)) {
 
         const int variable_dim_idx
                 = s.sparse_options[0].get_variable_dim_idx(DNNL_ARG_SRC);
@@ -179,7 +178,6 @@ int verify_grouped_input(const settings_t &s) {
     }
     return OK;
 }
-#endif
 
 static const std::string help_bia_mask
         = "UINT    (Default: `2`)\n    Specifies a bit-mask that indicates "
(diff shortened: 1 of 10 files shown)


tests/benchdnn/matmul/matmul.cpp
```diff
@@ -961,16 +924,11 @@ int init_ref_memory_args(dnn_mem_map_t &ref_mem_map, dnn_mem_map_t &mem_map,
...
     const auto wei_encoding
             = prb->sparse_options.get_encoding(DNNL_ARG_WEIGHTS);
-    const auto dst_encoding = prb->sparse_options.get_encoding(DNNL_ARG_DST);
 
-#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
-    // The only supported configuration of grouped matmul
-    const bool is_grouped
-            = src_encoding == dnnl_grouped && dst_encoding == dnnl_grouped;
-#endif
+    const bool is_grouped = prb->sparse_options.is_grouped(DNNL_ARG_SRC)
```
mzhukova, line 930: I somewhat doubt that current PR fully enables other grouped flavor (i.e. grouped by grouped into 3D), so could we for now stick to using src is grouped **AND** dst is grouped (and not **OR**)?

tests/benchdnn/matmul/bench_matmul.cpp
```diff
@@ -127,15 +127,18 @@ int verify_input(const settings_t &s, const settings_t &def) {
...
-#if DNNL_EXPERIMENTAL_GROUPED_MEMORY
 // Validate input consistency for grouped encoding
 //
 // Notes:
 // - `--grouped` parameter uses the same configuration for both SRC and
 //   DST, so we only validate SRC parameters
 // - Currently only M dimension grouping is supported
 int verify_grouped_input(const settings_t &s) {
-    if (s.sparse_options[0].get_encoding(DNNL_ARG_SRC) == dnnl_grouped) {
+#if !DNNL_EXPERIMENTAL_GROUPED_MEMORY
```
mzhukova: wouldn't it be better to swap it though? i.e. have `#if DNNL_EXPERIMENTAL_GROUPED_MEMORY` and validation code inside, and otherwise return OK
otherwise we have a dead code for the case when oneDNN is compiled with grouped memory off

mzhukova (approved):

tests/benchdnn/matmul/matmul.cpp
dzarukin: That was a copy-paste error from above. Fixed in both places, thanks for the catch.

tests/benchdnn/matmul/bench_matmul.cpp
dzarukin: I dropped this as it turned out not needed. Thanks.

dzarukin:
make test

densamoilov (approved):

dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul

dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul

PR #5179, opened 2026-05-20
xe: gemm: fix cti application when enabled

Simonsays095:
Addresses [MFDNN-14993](https://jira.devtools.intel.com/browse/MFDNN-14993). The C-interleaved tile (cti) feature added to GEMM encoded the size of the dst data type into the kernel, but we allow dst conversions when selecting a strategy from the catalog. Some downconversion cases require extra padding to work group and dispatch sizes, which is not accounted for in kernel evaluation and could result in improperly overestimating performance of this feature.

Fixes the disagreement between scores and execution time by simplifying the cti feature. If it's enabled, the kernel will always choose an interleave chunk size so that the worst case alignment will have perfect cache line alignment. This was the previous default option, but now the chunk size correctly accounts for dst conversion. The option to supply a custom chunk size with `ctiN` is removed since it's unneeded.


src/gpu/intel/gemm/jit/include/gemmstone/driver_info.hpp
```diff
@@ -137,10 +137,13 @@ struct CommonDriverInfo {
...
-        // Only the lowest 7 bits are set after the shift, so this cast is safe.
-        int chunk = static_cast<int>((flags & FlagMaskCInterleave) >> FlagShiftCInterleave);
-        return chunk ? chunk : 1;
+    bool cInterleaveEnabled() const {
+        return flags & FlagCInterleave;
+    }
+
+    // Returns the C interleave chunk size (columns of C per thread).
+    int cInterleaveChunk(Type Tc_ext = Type::invalid) const {
+        return (cInterleaveEnabled() && Tc_ext != Type::invalid) ? 64 / Tc_ext : 1;
```
kealan-barbieri: It might be better to fail on invalid type, to make sure the behavior is consistent throughout.

kealan-barbieri (approved):

src/gpu/intel/gemm/jit/include/gemmstone/driver_info.hpp
Simonsays095: I actually don't expect to use `Type::invalid` anywhere, that's left over from testing. Dropped.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
Since we're touching the performance model which is imperfect, some cases improve while others regress. Overall, performance improves slightly with no model level regressions. perf-gpu failure is an issue on the infra side

-- From PVC perf-gpu CI --
matmul improvements:

|                                               layer                                               | 9a7e1ac21 | 6d67faf44 | ratio |
| :------------------------------------------------------------------------------------------------ | --------: | --------: | ----: |
| --dt=f16:f16:f16 --stag=abc --wtag=acb --dtag=abc 3x3136x32:3x32x49                               |  0.021120 |  0.009440 |  2.24 |
| --dt=bf16:bf16:bf16 --stag=ab --wtag=ba --dtag=ab 11456x32:32x54                                  |  0.014880 |  0.009120 |  1.63 |
| --dt=bf16:bf16:bf16 --stag=ab --wtag=ba --dtag=ab 11456x32:32x51                                  |  0.015520 |  0.009760 |  1.59 |
| --dt=bf16:bf16:bf16 --stag=abc --wtag=acb --dtag=abc 80x4096x64:80x64x77                          |  0.248000 |  0.214880 |  1.15 |
| --dt=bf16:bf16:bf16 --stag=ab --wtag=ba --dtag=ab 96744x512:512x29                                |  0.203520 |  0.180320 |  1.13 |
| --dt=bf16:bf16:bf16 --stag=abc --wtag=acb --dtag=abc 40x4096x64:40x64x77                          |  0.121600 |  0.110080 |  1.10 |
| --dt=f16:f16:f16 --stag=abc --wtag=cab --dtag=abc --bia-dt=f16 --bia_mask=4 1x514x768:1x768x50265 |  0.363200 |  0.330400 |  1.10 |
| --dt=bf16:bf16:bf16 --stag=ab --wtag=ba --dtag=ab 11456x16:16x51                                  |  0.014880 |  0.013600 |  1.09 |

matmul regressions:

|                                                layer                                                 | 9a7e1ac21 | 6d67faf44 | ratio |
| :--------------------------------------------------------------------------------------------------- | --------: | --------: | ----: |
| --dt=f16:f16:f16 --stag=abc --wtag=abc --dtag=abc 16x256x160:16x160x77                               |  0.014240 |  0.015680 |  0.91 |
| --dt=f16:f16:f16 --stag=cabd --wtag=dabc --dtag=abcd --attr-post-ops=clip:-1:1 1x8x100x32:1x8x32x900 |  0.008640 |  0.009920 |  0.87 |
| --dt=bf16:bf16:bf16 --stag=acb --wtag=abc --dtag=abc 128x32x49:128x49x49                             |  0.010720 |  0.013600 |  0.79 |
| --dt=bf16:bf16:bf16 --stag=ab --wtag=ab --dtag=ab --bia-dt=bf16 11456x96:96x16                       |  0.008160 |  0.010880 |  0.75 |
| --dt=bf16:bf16:bf16 --stag=acb --wtag=abc --dtag=abc 64x160x64:64x64x77                              |  0.016640 |  0.031040 |  0.54 |

h-sadia:
How are we now *correctly* estimating these down-conversions that were being over estimated before?

src/gpu/intel/gemm/jit/include/gemmstone/strategy.hpp
```diff
@@ -335,6 +335,7 @@ struct GEMMStrategy : public GEMMStrategyPOD
     void trimKChain(ngen::HW hw, int k, const GEMMProblem &problem);
 
     int wgTile(LoopType l)                            const { return unroll[l] * wg[l]; }
+    int cInterleaveChunk(Type Tc_ext)                 const { return cInterleave ? 64 / Tc_ext : 1; }
```
h-sadia, line 338: What is the reason behind dividing *64* by the exterior data type? Is 64 the word size? I guess this is the logic which *correctly* estimates the score?

Simonsays095: 64 is the cache line size, so 64/Tc_ext is the number of Tc_ext elements that fit in a cache line. Although the reason for it being here is a bit more nuanced...

For any Tc_ext type, the "worst" leading dimension in terms of cache line efficiency is when the ld and 64/Tc_ext are coprime (e.g. ld=64/Tc_ext+1). In those cases, we only get back to a cache line at the start of the row (assuming row-major) after 64/Tc_ext columns. Picking this as the chunk size guarantees that each column assigned to a given thread has the same starting offset relative to the nearest cache line start.

Simonsays095:
> How are we now _correctly_ estimating these down-conversions that were being over estimated before?

Before, we were using the catalog entry's C type as Tc_ext directly. Now we pass in the actual Tc_ext to the eval_params and use it instead. In practice, that led to the following issues:
- Tc_ext is smaller than the catalog entry's C type
- The C matrix is no longer cache line aligned with the smaller type (in some cases)
- Being cache line unaligned requires extra work groups to cover "tail" cases in the cti implementation
- These extra work groups have an execution time cost that's not accounted for, which overestimates performance of the cti kernel.

src/gpu/intel/gemm/jit/include/gemmstone/strategy.hpp
h-sadia: So essentially, all threads will end up having the same offset because of this estimation? In case of worst case of starting after 64/Tx_ext columns; can we assume it will badly affect the kernel strategy score?

h-sadia (approved):

src/gpu/intel/gemm/jit/include/gemmstone/strategy.hpp
Simonsays095: Yep, each thread will be assigned columns that have the same relative offset to the nearest cache line. That's the backbone of the cti optimization (plus shifting so that each thread starts at the beginning of the cache line).

In the worst-case scenario, the kernel evaluator will score the regular kernel poorly, due to `cacheLineEffect` being large. But cti kernels always have `cacheLineEffect=1` (or close to it) so we will score it better in this regard (already working well) - at the cost of potentially extra work groups, which we now account for more accurately.

PR #5181, opened 2026-05-21
[RFC] Drop internal GEMM API

echeresh:
The rendered document: [link](https://github.com/uxlfoundation/oneDNN/tree/echeresh/gemm-to-matmul/rfcs/20260520-gemm-to-matmul).

Plan:

- [x] Extend `matmul/ref.cl` to use runtime sizes - to match `gemm/ref.cl`: https://github.com/uxlfoundation/oneDNN/pull/5339
- [x] Migrate RNN to use matmul: https://github.com/uxlfoundation/oneDNN/pull/5290
- [x] Migrate IP to use matmul: https://github.com/uxlfoundation/oneDNN/pull/5289
- [ ] Introduce kernel config in JIT GEMM to use it as a coherent problem description: https://github.com/uxlfoundation/oneDNN/pull/5245
- [ ] Introduce a wrapper matmul primitive to hide reshape
- [ ] Migrate JIT GEMM to use matmul PD
- [ ] Clean up the remaining GEMM API

### DIFF
--- a/rfcs/20260520-gemm-to-matmul/README.md
+++ b/rfcs/20260520-gemm-to-matmul/README.md
@@ -1,0 +1,99 @@
+# RFC: Drop the internal GEMM API
+
+## Background
+
+oneDNN exposes two distinct representations for matrix multiplication on GPU:
+
+- **Matmul** - row-major by default, user-facing, with `DNNL_ARG_SRC / WEIGHTS / DST` arguments
+- **GEMM** - column-major, internal, BLAS-shaped (`M, N, K, lda, ldb, ldc, transa,
+  transb`), with cross-wired mapping (`DNNL_ARG_SRC` -> `B` and `DNNL_ARG_WEIGHTS` -> `A`) for `B^T x A^T = C^T` reformulation from row- to column-major layouts
+
+These two descriptions express the same arithmetic. Today the GEMM internal API
+exists primarily as the bridge between matmul and the JIT GEMM kernels
+(gemmstone-based) and as an internal interface for RNN and inner product.
+
+The GEMM API is represented by:
+
+- Descriptor: `gemm_desc_t`
+- Primitive descriptor (PD): `gemm_pd_t`
+- Primitive kind: `primitive_kind::gemm`
+
+oneDNN GPU has four GEMM implementations:
+
+- `intel::gemm::xe_hp_systolic_t` - XeHP+ implementation with blocked layouts
+- `intel::gemm::gen_t` - JIT GEMM implementation
+- `intel::gemm::with_post_ops_t` - a wrapper calling JIT GEMM and applying post-ops in a separate OpenCL kernel
+- `intel::gemm::ref_t` - reference GEMM implementation
+
+## Motivation
+
+Maintaining a parallel description for the matmul operation is hard to justify:
+
+- JIT GEMM itself uses an internal `GEMMProblem` abstraction so we end up with
+  a `matmul_desc` -> `gemm_desc` -> `GEMMProblem` translation chain, where the middle layer can be removed.
+- The GEMM API handles row-to-column-major translation poorly: it
+  [auto-swaps](https://github.com/uxlfoundation/oneDNN/blob/9a7e1ac215b0f571749dff8a5d8373cb3922a618/src/common/gemm_types.hpp#L144)
+  some properties but never re-assigns indices, and attribute-related translation is done ad hoc inside the JIT GEMM implementations.
+- One of the reasons to have an internal GEMM API was to support custom A or B reduction (used for backward inner product).
+  Matmul has since gained the same [capability](https://github.com/uxlfoundation/oneDNN/blob/455ac84e8fd8f2450b13fc29f06f43f18caf842c/src/common/opdesc.hpp#L279),
+  so this reason no longer holds.
+
+## Design
+
+The proposal is to delete the internal GEMM descriptor and primitive descriptor
+(`impl::gemm_desc_t` and `impl::gemm_pd_t`) and let matmul be the single
+internal representation for the operation:
+
+1. **One internal descriptor.** All GPU consumers of GEMM - matmul
+   itself, inner-product and RNN are ported to matmul API (`create_gemm_pd` -> `create_matmul_pd`).
+   The internal GEMM descriptor, its PD base, its A/B/C argument keys, its construction
+   helper, and its registration list are removed.
+
+2. **All GEMM kernels become matmul implementations.** The existing JIT
+   GEMM implementations (`intel::gemm::xe_hp_systolic_t` and `intel::gemm::gen_t`) register under
+   matmul. They share a single base PD that owns the JIT GEMM common
+   logic - including post-op canonicalization, swap-ab translation,
+   quantization-related logic. `intel::gemm::with_post_ops_t` is reimplemented
+   on top of the matmul descriptor and PD, and `intel::gemm::ref_t` is merged with
+   `intel::matmul::ref_t`.
+
+3. **Column-major orientation is a private JIT GEMM implementation detail.** The matmul
+   descriptor is kept unchanged. The decision to swap A and B
+   in order to match a column-major kernel lives entirely inside the JIT GEMM PDs.
+
+## Out of scope
+
+- The public GEMM API (`dnnl_sgemm`, `dnnl_gemm_*`). It dispatches directly to
+  CPU and is independent of the internal PD.
+- `brgemm_desc_t` and the BRGEMM ecosystem - a separate path with a separate API.
+- CPU GEMM implementations. The proposal is a GPU-side cleanup. CPU matmul/GEMM paths are out of scope.
+
+## Update: handling transformations and swap-ab inside JIT GEMM
+
+Today, row-to-column-major mapping is split between `gemm_desc_t`, which
+implicitly rewires arguments, and JIT GEMM PDs, which handle `swap_ab` through
+scattered init/execute branches. Additionally, PDs apply post-op-related
+transformations (e.g. bias-to-post-op), which forces a scattered, intermediate
+representation that diverges from the external descriptor and PD.
+
+The proposal is to build one canonical "view" from the descriptor and
+attributes. The view is first populated directly from the descriptor and PD.
+Two transformations are then applied - (1) `swap_ab` and (2) post-op - to match
+the kernel's expectations. This design gives us a single source of truth that
+is fully coherent at any given time.
+
+A few comments for further discussion:
+
+- Can we use `GEMMProblem` as the "view" representation? Partially yes, but
+  with nuances:
+    - `GEMMProblem` doesn't include M/N/K/LDA/LDB, host scalars, or offsets so
+       we still need a complementary structure to represent these
+    - Moving to `GEMMProblem` may require operating based on `GEMMProblem`
+      derived parameters or adding another conversion back to the oneDNN world.
+      It's hard to fully separate the oneDNN and gemmstone spaces, so the
+      decision of how much to rely on `GEMMProblem` falls on a spectrum.
+      Additional input is appreciated.
+- The first step in implementing the RFC is to clean up the existing PD flow by
+  introducing a view abstraction. This will concentrate the mapping logic in one
+  place and make the transition from the GEMM API to the matmul descriptor/PD
+  more mechanical.


rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,69 @@
...
+   The internal GEMM descriptor, its PD base, its A/B/C argument keys, its construction
+   helper, and its registration list are removed.
+
+2. **All GEMM kernels become matmul implementations.** The existing JIT
+   GEMM implementations (`intel::gemm::xe_hp_systolic_t` and `intel::gemm::gen_t`) register under
+   matmul. They share a single base PD that owns the JIT GEMM common
+   logic - including post-op canonicalization, swap-ab translation,
+   quantization-related logic. `intel::gemm::with_post_ops_t` is reimplemented
+   on top of the matmul descriptor and PD, and `intel::gemm::ref_t` is merged with
+   `intel::matmul::ref_t`.
```
rjoursler, line 58: I have looked at this in the past, merging `gemm::ref_t` with `matmul::ref_t` is likely too much additional complexity for this process. The key issue is that `gemm::ref_t` is a reusable implementation and this is used to implement runtime dimensions support and avoid kernel compilation overhead in RNN (since RNN can instantiate up to 12 GEMM primitives). I would suggest that we just move `gemm:ref_t` to be `matmul::reusable_ref_t` to align with other primitives.

rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,69 @@
...
+  so this reason no longer holds.
+
+## Design
+
+The proposal is to delete the internal GEMM descriptor and primitive descriptor
+(`impl::gemm_desc_t` and `impl::gemm_pd_t`) and let matmul be the single
+internal representation for the operation:
+
+1. **One internal descriptor.** All GPU consumers of GEMM - matmul
+   itself, inner-product and RNN are ported to matmul API (`create_gemm_pd` -> `create_matmul_pd`).
```
rjoursler, line 48: I looked at doing this for RNN a couple of years ago. As I recall, there are definitely some support gaps that will need to be addressed. While I don't recall much as to what I saw, one example is that `matmul::ref` does not support buffer offsets which is a necessary feature for RNN's computation.

rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,69 @@
...
+   The internal GEMM descriptor, its PD base, its A/B/C argument keys, its construction
+   helper, and its registration list are removed.
+
+2. **All GEMM kernels become matmul implementations.** The existing JIT
+   GEMM implementations (`intel::gemm::xe_hp_systolic_t` and `intel::gemm::gen_t`) register under
+   matmul. They share a single base PD that owns the JIT GEMM common
+   logic - including post-op canonicalization, swap-ab translation,
+   quantization-related logic. `intel::gemm::with_post_ops_t` is reimplemented
```
rjoursler, line 56: Would it make sense to just make `with_post_ops_t` into a general post-op applying kernel separated from GEMM and let primitives opt in to instantiating/calling it when they do not support post-ops?

rfcs/20260520-gemm-to-matmul/README.md
rjoursler, line 48: In my experience, especially when targeting reusable kernels, the nested primitive abstraction is rather clunky. Take for example, [this initialization](https://github.com/uxlfoundation/oneDNN/blob/d90d2b202a34f879261edb60b3bc6a9901af79f5/src/gpu/intel/rnn/grid.cpp#L792-L937) and [dispatching sequence](https://github.com/uxlfoundation/oneDNN/blob/d90d2b202a34f879261edb60b3bc6a9901af79f5/src/gpu/intel/rnn/grid.cpp#L1106-L1167) in RNN. This has led me to wonder if it would be worthwhile to just strip out the internal descriptor indirection and operate more directly on some kind of kernel abstraction whose usage is more like [this](https://github.com/uxlfoundation/oneDNN/blob/d90d2b202a34f879261edb60b3bc6a9901af79f5/src/gpu/intel/rnn/cell_common.cpp#L131-L133).

rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,69 @@
...
+Maintaining a parallel description for the matmul operation is hard to justify:
+
+- JIT GEMM itself uses an internal `GEMMProblem` abstraction so we end up with
+  a `matmul_desc` -> `gemm_desc` -> `GEMMProblem` translation chain, where the middle layer can be removed.
+- The GEMM API handles row-to-column-major translation poorly: it
+  [auto-swaps](https://github.com/uxlfoundation/oneDNN/blob/9a7e1ac215b0f571749dff8a5d8373cb3922a618/src/common/gemm_types.hpp#L144)
+  some properties but never re-assigns indices, and attribute-related translation is done ad hoc inside the JIT GEMM implementations.
+- One of the reasons to have an internal GEMM API was to support custom A or B reduction (used for backward inner product).
+  Matmul has since gained the same [capability](https://github.com/uxlfoundation/oneDNN/blob/455ac84e8fd8f2450b13fc29f06f43f18caf842c/src/common/opdesc.hpp#L279),
+  so this reason no longer holds.
```
Simonsays095, line 39: Another motivation: benchdnn's `skip-impl` feature is implemented in a way that doesn't play well with internal implementation suites like GEMM. If you use `--skip-impl` to skip the GEMM implementation that's selected by default, all remaining GEMM implementations will be skipped and it'll go back to trying matmul implementations. This is because all GEMM implementations are disguised as one entry in the matmul implementation list.

rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `intel::gemm::gen_t` - JIT GEMM implementation
+- `intel::gemm::with_post_ops_t` - a wrapper calling JIT GEMM and applying post-ops in a separate OpenCL kernel
+- `intel::gemm::ref_t` - reference GEMM implementation
+
+## Motivation
+
+Maintaining a parallel description for the matmul operation is hard to justify:
+
+- JIT GEMM itself uses an internal `GEMMProblem` abstraction so we end up with
+  a `matmul_desc` -> `gemm_desc` -> `GEMMProblem` translation chain, where the middle layer can be removed.
```
Simonsays095, line 33: One thing to consider: the `matmul_desc` -> `gemm_desc` conversion process applies an important renormalization (dimensionality reduction) optimization, which is essentially required for performant OpenVINO models. Without it, most OV layers are treated as batched+broadcast (and memory-bound) which slows them down significantly.

The renormalization is already generalized and could be applied to all matmul problems before selecting an implementation. It doesn't remove the extra link in the translation chain, but it would at least become `matmul_desc_t` -> `matmul_desc_t` (renormalized) -> `GEMMProblem` and would come with performance improvements to other matmul implementations too. It would just take some thought and effort to put this at the correct point in primitive initialization.

rfcs/20260520-gemm-to-matmul/README.md
echeresh: Thanks for the pointer, I'll take a look. This sounds like something that would be nice to address before starting this consolidation.

echeresh: Thanks for the comment, I haven't look into that - but sounds like it's the right direction.

Can we mark this as a non-goal for this RFC? I would like to limit the scope solely to the GEMM -> matmul migration.

rfcs/20260520-gemm-to-matmul/README.md
echeresh: Similar comment here - sounds like it's a good thing to do but it's a larger scope.

rfcs/20260520-gemm-to-matmul/README.md
echeresh: Thanks, let me check this. Hypothetically, if merging is possible, would it be a good idea to make `matmul::ref_t` a reusable kernel?

rfcs/20260520-gemm-to-matmul/README.md
echeresh: Thanks, good point. I also see some translation logic around prelu and bias.

This looks like the crux of all this migration. This pattern of intermediate desc/PD translation exists in oneDNN (GPU GEMM or convolution) but it doesn't have proper architectural support. After thinking a bit about this, I think having an intermediate representation (or state or wrapper) is the right design choice here:

- `GEMMProblem` doesn't include M/N/K sizes so it can't fully replace the descriptor
- Using `matmul_pd_t` getters alongside kernel-side `GEMMProblem`-derived getters is error-prone (`pd_t::M()` can refer to either `M` or `N` depending on the kernel view)

I'll extend the RFC to describe the implementation design around translation to discuss it from the implementation side.

> would come with performance improvements to other matmul implementations too

Talking about the practical side - which implementations are you referring to here? My understanding is that it's important for GPU JIT GEMM (or in other words, all gemmstone-based implementations). Reference - not that important, with-post-ops - looks irrelevant as well.

rfcs/20260520-gemm-to-matmul/README.md
rjoursler: Yes I think this is a good idea. The main blocker to this is post-ops handling as we do not have the infrastructure for rescuable kernels with post-ops. In [#4744](https://github.com/uxlfoundation/oneDNN/pull/4474), I did implement the necessary infrastructure, but in its current state there are performance regressions on Bnorm. If we are going to make `matmul::ref_t` reusable, we likely should get that work completed.

rfcs/20260520-gemm-to-matmul/README.md
rjoursler: Of course, I just wanted to mention this as something to keep in mind during the design/implementation in case the refactoring is more complicated than a rote interface change.

rfcs/20260520-gemm-to-matmul/README.md
Simonsays095: with-post-ops should just be jit:gemm plus a separate post-op kernel. Dimensionality reduction could improve performance in the post-op kernel as well as reference matmul (by removing one or more batch dim loops), although it'll still be limited by poor implementation performance in general. In both cases the better situation for performance would be to enable jit:gemm for that case. My point was that dimensionality reduction -> simpler problem -> faster performance is free with this transformation. Although the impact on other implementations is very minor compared to jit:gemm.

echeresh: > My point was that dimensionality reduction -> simpler problem -> faster performance is free with this transformation. Although the impact on other implementations is very minor compared to jit:gemm.

Yeah, I see your point but I don't see a simple path forward to bring this transformation to a higher level:

- Where to put `matmul_desc_t` -> `matmul_desc_t` reshape? Into to-be-created `create_matmul_pd`? What about the main path via the PD iterator creation?
- What to do with the PD attributes? A proper reshape must transform the MDs in the attributes as well. It can't be done in-place as this breaks the contract: the caller initializes a PD with batched MDs so this has to be preserved.

The only option I see here is to keep an intermediate abstraction - say at the existing GEMM API level. This abstraction will hide transformation and do proper forwarding of user MDs to the underlying implementations.

Before discussing this direction (about which I have a few major concerns), do you see a better option?

rjoursler: > Yeah, I see your point but I don't see a simple path forward to bring this transformation to a higher level
> ...
> The only option I see here is to keep an intermediate abstraction - say at the existing GEMM API level.
 
Tangentially related: Another thing I have noticed after working with Gemmstone, the build/test cycle is much faster. Trying to implement this so that we can independently compile and tests a well defined matmul interface may be worth considering.

echeresh: Update: I added a section covering the JIT PD implementation design:

https://github.com/uxlfoundation/oneDNN/tree/echeresh/gemm-to-matmul/rfcs/20260520-gemm-to-matmul#update-handling-transformations-and-swap-ab-inside-jit-gemm

The simplest approach is to prototype it so I'm going to share a branch to discuss the directions.

rfcs/20260520-gemm-to-matmul/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Can we use `GEMMProblem` as the "view" representation? Partially yes, but
+  with nuances:
+    - `GEMMProblem` doesn't include M/N/K/LDA/LDB, host scalars, or offsets so
+       we still need a complementary structure to represent these
+    - Moving to `GEMMProblem` may require operating based on `GEMMProblem`
+      derived parameters or adding another conversion back to the oneDNN world.
+      It's hard to fully separate the oneDNN and gemmstone spaces, so the
+      decision of how much to rely on `GEMMProblem` falls on a spectrum.
+      Additional input is appreciated.
```
rjoursler, line 95: The key distinction here is the notion between compile time problem parameters and runtime problem parameters. I would expect we need something similar to [atomic_conf_t](https://github.com/uxlfoundation/oneDNN/blob/dd77e900f6c44a851f29ef6f2a7b8001fad43744/src/gpu/intel/reduction/atomic.hpp#L69-L79), where we employ a single structure wrapping `GEMMProblem` and another containing runtime parameters. In the current Gemmstone API, runtime parameters are largely embedded in [CommonDriverInfo](https://github.com/uxlfoundation/oneDNN/blob/dd77e900f6c44a851f29ef6f2a7b8001fad43744/src/gpu/intel/gemm/jit/include/gemmstone/driver_info.hpp#L82-L149). This abstraction mainly targets runtime parameters derived from the strategy though and is severely underpowered relative to our needs.

rjoursler (approved):

vpirogov (approved):

PR #5184, opened 2026-05-21
aarch64: fixes for jit_binary and post_op injectors

Sqvid:
# Description

This change de-duplicates the functions used to check support for post-op injection.

Also fixes an issue where the wrong instruction is used for broadcast loads on `asimd`, causing crashes in nightly: https://github.com/uxlfoundation/oneDNN/actions/runs/26145272841/job/76901306317

Failing case:
```
$ DNNL_MAX_CPU_ISA='advanced_simd' ./build/tests/benchdnn/benchdnn --binary --alg=add --attr-post-ops=binary_div:f32:common 1x4:1x4
zsh: segmentation fault (core dumped)
```

# Checklist

## General

### Bug fixes

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
@@ -22,6 +22,7 @@
 
 #include "common/primitive_attr.hpp"
 #include "common/utils.hpp"
+#include "common/verbose_msg.hpp"
 #include "cpu/aarch64/injectors/jit_uni_binary_injector.hpp"
 
 namespace dnnl {
@@ -30,10 +31,14 @@ namespace cpu {
 namespace aarch64 {
 namespace binary_injector {
 
+#define VCHECK_BIN_INJ_BOOL(cond, msg) \
+    VCONDCHECK(primitive, create, check, binary_injector, cond, false, msg);
+
 static bcast_set_t get_all_strategies_supported_by_injector() {
     return bcast_set_t {broadcasting_strategy_t::scalar,
             broadcasting_strategy_t::per_oc,
             broadcasting_strategy_t::per_oc_spatial,
+            broadcasting_strategy_t::per_mb_spatial,
             broadcasting_strategy_t::per_mb_w, broadcasting_strategy_t::per_w,
             broadcasting_strategy_t::no_broadcast};
 }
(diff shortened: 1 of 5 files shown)


dzarukin (approved):

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
```diff
@@ -1450,27 +1465,23 @@ rhs_address_t jit_uni_binary_injector_t<isa>::remove_bcast_bit(
...
-                    rhs_addr.offt_, host_->X_TMP_0);
-            host_->uni_ldr(tmp_vmm, host_->X_DEFAULT_ADDR);
-            break;
         case data_type::s32:
             host_->add_imm(host_->X_DEFAULT_ADDR, rhs_addr.base_,
                     rhs_addr.offt_, host_->X_TMP_0);
-            host_->ld1rw(Xbyak_aarch64::ZRegS(tmp_vmm.getIdx()),
-                    host_->P_ALL_ONE / Xbyak_aarch64::T_z,
-                    Xbyak_aarch64::ptr(host_->X_DEFAULT_ADDR));
+            if (is_superset(isa, sve)) {
```
jondea, line 1477: Do we not have a generic load replicate function in jit_generator for this? It would allow us to have some logic to fuse the the add_imm in some cases (I think I did this in jit_brgemm). If there isn't one already and it's too onerous, feel free to leave.

src/cpu/aarch64/injectors/jit_uni_postops_injector.cpp
```diff
@@ -195,54 +173,58 @@ post_ops_ok_args_t::post_ops_ok_args_t(const cpu_isa_t isa,
...
-    const bool with_sum = sum_idx != -1;
-    const auto &entry
-            = with_sum ? post_ops.entry_[sum_idx] : dnnl_post_ops::entry_t();
-    const auto sum_scale = with_sum ? entry.sum.scale : 0;
-    const auto sum_zero_point = with_sum ? entry.sum.zero_point : 0;
+    const auto first_sum_idx = post_ops.find(primitive_kind::sum);
+    const bool with_sum = first_sum_idx != -1;
+    const auto &maybe_sum_entry = with_sum ? post_ops.entry_[first_sum_idx]
+                                           : dnnl_post_ops::entry_t();
+    const auto first_sum_scale = with_sum ? maybe_sum_entry.sum.scale : 0;
```
jondea, line 187: I'm surprised the compiler doesn't complain here about the implicit int to float cast.

jondea (approved):

src/cpu/aarch64/injectors/jit_uni_binary_injector.cpp
Sqvid: I think `jit_io_helper_t` has a generic broadcast function (though it is terrible) but I'd rather not try to integrate it here.

PR #5188, opened 2026-05-21
cmake: modify compilation to improve ccache hit rate

rjoursler:
This change modifies the oneDNN build with ccache to be build directory independent. The goal is to increase ccache hits across build directories. As motivation, when debugging a regression I often create different build directories like:

```
build_good // Build with correct behavior
build_bad  // Build with incorrect behavior
build      // Build for testing fixes
```

and then compare outputs between the builds to locate the cause.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,37 @@ configure_file(
     "${PROJECT_SOURCE_DIR}/include/oneapi/dnnl/dnnl_config.h.in"
     "${PROJECT_BINARY_DIR}/include/oneapi/dnnl/dnnl_config.h"
 )
-include_directories_with_host_compiler_before(${PROJECT_BINARY_DIR}/include)
+
+if(CMAKE_CXX_COMPILER_LAUNCHER)
+    # Adapt the build process when using ccache to improve the hit rate
+    get_filename_component(LAUNCHER "${CMAKE_CXX_COMPILER_LAUNCHER}" NAME)
+    if(LAUNCHER STREQUAL "ccache")
+        file(GLOB_RECURSE GENERATED_HEADERS "${PROJECT_BINARY_DIR}/include/*.h")
+        list(SORT GENERATED_HEADERS)
+        set(GENERATED_HASH "")
+        foreach(HDR IN LISTS GENERATED_HEADERS)
+            file(SHA256 "${HDR}" HDR_HASH)
+            string(APPEND GENERATED_HASH "${HDR_HASH}")
+        endforeach()
+        string(SHA256 GENERATED_HASH "${GENERATED_HASH}")
+        string(SUBSTRING "${GENERATED_HASH}" 0 16 GENERATED_HASH)
+
+        # Use a source level generated directory so that the build is
+        # independent of the build directory name. Uses a hash on generated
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -137,7 +137,37 @@ configure_file(
     "${PROJECT_SOURCE_DIR}/include/oneapi/dnnl/dnnl_config.h.in"
     "${PROJECT_BINARY_DIR}/include/oneapi/dnnl/dnnl_config.h"
 )
-include_directories_with_host_compiler_before(${PROJECT_BINARY_DIR}/include)
+
+if(CMAKE_CXX_COMPILER_LAUNCHER)
+    # Adapt the build process when using ccache to improve the hit rate
+    get_filename_component(LAUNCHER "${CMAKE_CXX_COMPILER_LAUNCHER}" NAME)
+    if(LAUNCHER STREQUAL "ccache")
+        file(GLOB_RECURSE GENERATED_HEADERS "${PROJECT_BINARY_DIR}/include/*.h")
```
mzhukova, line 145: should this include `*.hpp` in case it appears in the future?

CMakeLists.txt
```diff
@@ -137,7 +137,37 @@ configure_file(
...
+            string(APPEND GENERATED_HASH "${HDR_HASH}")
+        endforeach()
+        string(SHA256 GENERATED_HASH "${GENERATED_HASH}")
+        string(SUBSTRING "${GENERATED_HASH}" 0 16 GENERATED_HASH)
+
+        # Use a source level generated directory so that the build is
+        # independent of the build directory name. Uses a hash on generated
+        # files to avoid conflicting build configurations.
+        set(DNNL_GENERATED_INCLUDE_DIR
+          "${PROJECT_SOURCE_DIR}/.dnnl_generated/${GENERATED_HASH}/include")
```
mzhukova, line 159: do we need a `.gitignore` etnry for this?

CMakeLists.txt
```diff
@@ -137,7 +137,37 @@ configure_file(
...
+        file(COPY "${PROJECT_BINARY_DIR}/include/"
+          DESTINATION "${DNNL_GENERATED_INCLUDE_DIR}")
+        message(STATUS "Generated headers: ${DNNL_GENERATED_INCLUDE_DIR}")
+    endif()
+endif()
```
mzhukova, line 164: also wondering how should we clean this directory?

rjoursler: These files would have to be user managed as there is no reliable way to inform the build system whether a set of headers is currently being used by another build directory. As these files are relatively small, such management should rarely be necessary.

This is actually related to your comment:

> do we need a .gitignore entry for this?

My main motivation for not adding  this to `.gitignore` is related this requirement. My thinking is that these files should be obvious to developers when running commands like `git status` as a method to inform them that manual management is occasionally required.

h-sadia:
Does this change the way we build the library if we have to use ccache?

rjoursler:
@h-sadia it mostly would have changed the file locations of generated headers.

Anyway, I think I found a way to avoid cmake changes by relying on `CCACHE_DIR`, so I am closing this.

PR #5190, opened 2026-05-22
grouped matmul: support per-group global NVFP4 scale 

mzhukova:
PR is to address [MFDNN-15180](https://jira.devtools.intel.com/browse/MFDNN-15180).
Essentially, for NVFP4 (that requires two-levels of scaling: grouped + global one), we need to change support to allow global value per-group and not per tensor.

### DIFF
--- a/tests/benchdnn/dnn_types.cpp
+++ b/tests/benchdnn/dnn_types.cpp
@@ -1355,8 +1355,19 @@ int attr_args_t::prepare_post_ops_mds(const attr_t &attr, int ndims,
 
             // deduce binary, prelu dims based on input policy
             dnnl_dims_t rhs_tensor_dims = {};
-            for (auto d = 0; d < ndims; ++d)
-                rhs_tensor_dims[d] = (!(mask & (1 << d))) ? 1 : dims[d];
+            // Overload mask 0 meaning for when binary post-op is applied
+            // to grouped memory, changing it to be per-group (i.e. one value
+            // per concatenated tensor, for instance for NVFP4 global scale).
+            // The grouped dst has no explicit num_groups in dims,
+            // so form a [group_count, 1, ...] descriptor manually
+            if (grouped_count > 0 && mask == 0) {
+                for (auto d = 0; d < ndims; ++d)
+                    rhs_tensor_dims[d]
+                            = (d == grouped_var_dim_idx) ? grouped_count : 1;
+            } else {
+                for (auto d = 0; d < ndims; ++d)
+                    rhs_tensor_dims[d] = (!(mask & (1 << d))) ? 1 : dims[d];
+            }
 
             auto rhs_tensor_desc = dnn_mem_t::init_md(ndims, rhs_tensor_dims,
                     po_rhs_tensor_entry.dt, po_rhs_tensor_entry.tag,
--- a/tests/benchdnn/matmul/ref_matmul.cpp
+++ b/tests/benchdnn/matmul/ref_matmul.cpp
@@ -186,7 +186,8 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
         int64_t N, int64_t K, int64_t mc, int64_t nc, int64_t src_row_base,
         int64_t wei_base, int64_t wei_k_stride, int64_t wei_n_stride,
         int64_t dst_row_base, int64_t bia_base, int64_t bia_m_stride,
-        int64_t bia_n_stride, const attr_t &attr, const args_t &args) {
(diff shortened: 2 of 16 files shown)


umar456 (approved):

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -247,8 +250,19 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
             dst += p.bia_m->get_f32_elem(bia_idx);
         }
 
-        const auto v_po_vals
-                = prepare_po_vals(*p.dst_m, args, p.v_po_masks, dst_off);
+        auto v_po_vals = prepare_po_vals(*p.dst_m, args, p.v_po_masks, dst_off);
+        // NOTE: mask == 0 means "apply per-group" for grouped memory,
+        // so changeing behavior here for that particular case
```
dzarukin:
```suggestion
        // so changing behavior here for that particular case.
```

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -247,8 +250,19 @@ static void compute_ref_matmul_chunk(const chunk_params_t &p, int64_t M,
...
+        for (size_t d = 0; d < v_po_vals.size(); d++) {
+            const int po_idx = p.v_po_masks[d].first
+                            / DNNL_ARG_ATTR_MULTIPLE_POST_OP_BASE
+                    - 1;
+            if (attr.post_ops.entry[po_idx].is_binary_kind()
+                    && p.v_po_masks[d].second == 0) {
+                v_po_vals[d] = args.find(p.v_po_masks[d].first)
+                                       .get_f32_elem(p.group_id);
```
dzarukin: I think it's reasonable to pass `group_id` into prepare_po_vals explaining it's 0 for conventional use and an order of the group for grouped_mm instead of re-updating the value here.

tests/benchdnn/dnn_types.cpp
```diff
@@ -1355,8 +1355,23 @@ int attr_args_t::prepare_post_ops_mds(const attr_t &attr, int ndims,
...
+            // Overload mask 0 meaning for when binary post-op is applied
+            // to grouped memory, changing it to be per-group (i.e. one value
+            // per concatenated tensor, for instance for NVFP4 global scale)
+            // WARN: HOST_SCALAR doesn't make sense for this case (single value),
+            // therefore keeping it as-is
+            //
+            // The grouped dst has no explicit num_groups in dims,
+            // so form a [group_count, 1, ...] descriptor manually
+            if (grouped_count > 0 && mask == 0
+                    && e.binary.policy != attr_t::policy_t::HOST_SCALAR) {
```
dzarukin: Do we actually allow `HOST_SCALAR` for post-ops? If not, I'd drop this condition...

If somehow yes, it's better to verify it's `binary.mask_input == mask_input_t::policy` to use `policy` member directly.

tests/benchdnn/matmul/ref_matmul.cpp
```diff
@@ -377,7 +391,10 @@ void compute_ref_grouped_matmul(const prb_t *prb, const args_t &args) {
             bia_n_stride = 1;
         }
 
-        compute_ref_matmul_chunk(params, M_g, prb->n, prb->k, mc, nc,
+        auto params_g = params;
+        params_g.group_id = g;
```
dzarukin: Suggestion to move `group_id` into `compute_ref_matmul_chunk` as a product of parallelization and keep params intact.

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: fixed, thanks for the catch! :)

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: sure, I've moved it

tests/benchdnn/dnn_types.cpp
mzhukova: I just wanted to be explicit / on the safe side; we do not use it for binary post ops (it is used for dropout/scales/zps/grouped matmul hint)
I'm not attached to this, so I've dropped it

tests/benchdnn/matmul/ref_matmul.cpp
mzhukova: done, thanks!

mzhukova:
make test
disable benchdnn_all
enable benchdnn_matmul

dzarukin (approved):

PR #5191, opened 2026-05-22
[GPU] Generalize GRF mode options

Simonsays095:
Addresses [MFDNN-15100](https://jira.devtools.intel.com/browse/MFDNN-15100). With the Variable Registers per Thread (VRT) feature on Xe3p, "large GRF mode" no longer holds a specific meaning. GRF/thread>128 now has several options, with various properties of each (mainly, different maximum work group sizes are possible).

This PR replaces every instance of `large_grf_mode` and `grfs > 128` proxies with GRF/thread explicitly, and then computes the threads/WG as `(GRF/EU) / (GRF/thread) * (EU/WG)` which is accurate across all platforms currently. Some duplication of `grf_per_eu()` queries is present, which may be refactored later to deduplicate (currently exists in base oneDNN, gemmstone, and jit conv code).

Edit: Most Xe3p kernels used `wg 8x4` or `wg 4x8` which are incompatible with `grf512`. In those cases, I've reduced the work group to `wg 4x4` which will change performance but recover correctness.


h-sadia:
Is there a performance impact due to this change: (GRF/EU) / (GRF/thread) * (EU/WG)

src/gpu/intel/compute/device_info.cpp
```diff
@@ -242,30 +242,33 @@ int device_info_t::max_subgroup_size(data_type_t type) const {
...
         case gpu::intel::compute::gpu_arch_t::xe_hp:
         case gpu::intel::compute::gpu_arch_t::xe_hpg:
         case gpu::intel::compute::gpu_arch_t::xe_hpc:
         case gpu::intel::compute::gpu_arch_t::xe2:
         case gpu::intel::compute::gpu_arch_t::xe3p:
-        case gpu::intel::compute::gpu_arch_t::xe3:
-            return large_grf_mode ? 4 : 8;
-        case gpu::intel::compute::gpu_arch_t::unknown: return 7;
+        case gpu::intel::compute::gpu_arch_t::xe3: return 1024;
+        case gpu::intel::compute::gpu_arch_t::unknown: return 896;
```
h-sadia, line 266: Nit: You could combine the switch in terms of returning the grf_per_eu (896, 1024)

Simonsays095: I plan on eventually returning 1024 as the default for unknown arch since it's more common - I just didn't want to mess with that in this PR.

Simonsays095:
> Is there a performance impact due to this change: (GRF/EU) / (GRF/thread) * (EU/WG)

I'll run perf CI - it's possible. The main thing this changes is how `grf192` and `grf512` kernels are evaluated in the performance model, since they saturate the GPU at different problem sizes. `grf512` was previously evaluated assuming 4 threads/EU when it's actually 2 threads/EU - the GPU is saturated 2x faster than expected, thus we expect worse performance for larger shapes.

Since we're evaluating this more accurately, I expect performance to increase if anything. But we don't have many of these strategies yet, so I wouldn't be surprised if we don't have any meaningful performance impact.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

src/gpu/intel/gemm/jit/generator/pieces/hw_utils.hpp
```diff
@@ -115,12 +115,26 @@ static inline size_t maxSLMPerWG(ngen::Product p, int grfCount)
...
+
 static inline int threadsPerEU(ngen::HW hw, const CommonStrategy &strategy)
 {
-    if (hw >= ngen::HW::XeHP)
-        return (strategy.GRFs > 128) ? 4 : 8;
-    else
+    if (hw <= ngen::HW::XeLP)
         return 7;
+
+    return GRFPerEU(hw) / strategy.GRFs;
```
h-sadia, line 137: Is there a chance of any sort of integer rounding here? If yes, are there cases where we could be getting lesser number of threads than what's supported and optimal?

rjoursler (approved):

h-sadia (approved):

src/gpu/intel/gemm/jit/generator/pieces/hw_utils.hpp
Simonsays095: There are cases where this will round down, and that's expected. For example, Xe3p with grf192. 1024/192=5.333 threads/EU. Since we can't split GRFs between EUs, the driver sets 5 threads/EU as the maximum, and the maximum work group size is 5 threads/EU * 8 EU/subslice = 40 threads/subslice (or subgroups/workgroup). This can be confirmed with clinfo and OpenCL kernel queries - that's how I came up with this function in the first place.

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

PR #5202, opened 2026-05-27
[GPU] src: gpu: intel: sdpa: enable f16 SLM path for dKV f16 cases

h-sadia:
# Description

This PR adds a down-conversion path to DV and DK matmuls in micro sdpa kernel for training to align GQA tests in f16 between framework team (PT) to oneDNN GPU implementation. Although less in precision but aligns with framework expectation.

Addresses CI failure here: https://jira.devtools.intel.com/browse/MFDNN-14955

### DIFF
--- a/src/gpu/intel/sdpa/micro_bwd.cl
+++ b/src/gpu/intel/sdpa/micro_bwd.cl
@@ -436,10 +436,18 @@ inline void tile_store_k_slm(
 
 #if KV_GROUP_SIZE > 1
 #define IS_GQA 1
+#if defined(DST_DT_F16)
+#define REDUCE_DKDV_F16 1
+#else
+#define REDUCE_DKDV_F16 0
+#endif
 #if DST_DATA_T != float
 #define NEEDS_INTERMEDIATE_DKV 1
+#else
+#define NEEDS_INTERMEDIATE_DKV 0
 #endif
 #endif
+
 #if QRY_DATA_T != float
 #define NEEDS_INTERMEDIATE_DQ 1
 #endif
@@ -450,20 +458,55 @@ inline void tile_store_k_slm(
 #define DST_DATA_T_DKDV DST_DATA_T
 #endif
 
-// round f32 intermediate values to DST_DATA_T precision before GQA atomic
-// accumulation. Although less accurate, it matches the unfused path
-// where each query group matmul output passes through DST_DATA_T
-// storage before the reduction
+#if REDUCE_DKDV_F16
+// Enable SLM acummulation in f16 for f16 dV and dK reduce path
+// The global scratch is f32 (DST_DATA_T_DKDV=float); tile_atomic_add_f32
+// converts f16->f32 at the global write step, avoiding __builtin_IB_atomic_add_global_f16.
+DECLARE_2D_TILE(dv_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_vs_c_type_block0, ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
+        ugemm_vs_c_type_nblock1)
+DECLARE_2D_TILE(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+DECLARE_2D_TILE_SLM_ADD(dv_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_vs_c_type_block0, ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
+        ugemm_vs_c_type_nblock1)
+DECLARE_2D_TILE_SLM_ADD(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+DECLARE_2D_TILE_SLM_ADD_T(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+#define SLM_DKDV_DATA_T DST_DATA_T
+typedef dv_reduce_tile_type dv_acc_tile_type;
+typedef a_reduce_tile_type dk_acc_tile_type;
+inline float round_to_dst(float v) {
+    return v;
+}
+#else
+#define SLM_DKDV_DATA_T float
+typedef dv_tile_type dv_acc_tile_type;
+typedef a_tile_type dk_acc_tile_type;
 inline float round_to_dst(float v) {
     return CONVERT_FLOAT_T(CONVERT_DATA_T(v));
 }
+#endif
 
-inline void tile_store_dV(dv_tile_type *dV_tile_slm, global DST_DATA_T_DKDV *dV,
-        int m, int n, int ld, int offset_r, int offset_c, int rem) {
+inline void tile_store_dV(dv_acc_tile_type *dV_tile_slm,
+        global DST_DATA_T_DKDV *dV, int m, int n, int ld, int offset_r,
+        int offset_c, int rem) {
 
 #if IS_GQA
-    tile_elementwise_s(*dV_tile_slm, round_to_dst);
-    tile_atomic_add(*dV_tile_slm, dV, m, n, ld, offset_r, offset_c);
+    // tile_copy: f16->f32 for REDUCE_DKDV_F16 (avoids __builtin_IB_atomic_add_global_f16
+    // which can produce NaN), identity for f32 SLM.
+    // round_to_dst: identity for f16 (precision already implicit from SLM);
+    // rounds to bf16 for bf16 dst to match the unfused path.
+    dv_acc_tile_type dV_tile_slm_copy = *dV_tile_slm;
+    dv_tile_type dV_tile_f32;
+    tile_copy(dV_tile_slm_copy, dV_tile_f32);
+    tile_elementwise_s(dV_tile_f32, round_to_dst);
+    tile_atomic_add(dV_tile_f32, dV, m, n, ld, offset_r, offset_c);
+
 #else // MHA update
 
     dv_tile_type_dst dV_tile_dst; // convert to half
@@ -475,18 +518,21 @@ inline void tile_store_dV(dv_tile_type *dV_tile_slm, global DST_DATA_T_DKDV *dV,
     tile_store(dV_tile_dst, (global DST_TILE_DATA_T *)dV, m, n, ld, offset_r,
             offset_c);
 #endif
-
 #endif
 }
 
 #if TRANSPOSE_K
 // uses transposed dv_tile_type (D*Bc) for dK update
-inline void tile_store_dK_t(dv_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
-        int m, int n, int ld, int offset_r, int offset_c, int rem) {
+inline void tile_store_dK_t(dv_acc_tile_type *dK_tile,
+        global DST_DATA_T_DKDV *dK, int m, int n, int ld, int offset_r,
+        int offset_c, int rem) {
 
 #if IS_GQA
-    tile_elementwise_s(*dK_tile, round_to_dst);
-    tile_atomic_add(*dK_tile, dK, m, n, ld, offset_r, offset_c);
+    dv_acc_tile_type dK_tile_slm_copy_t = *dK_tile;
+    dv_tile_type dK_tile_f32;
+    tile_copy(dK_tile_slm_copy_t, dK_tile_f32);
+    tile_elementwise_s(dK_tile_f32, round_to_dst);
+    tile_atomic_add(dK_tile_f32, dK, m, n, ld, offset_r, offset_c);
 #else // MHA update
     dv_tile_type_dst dK_tile_dst;
     tile_copy_reblock(*dK_tile, &dK_tile_dst);
@@ -503,12 +549,15 @@ inline void tile_store_dK_t(dv_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
 #else
 
 // uses qdSt tile (Bc*D) for dK update
-inline void tile_store_dK(a_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
+inline void tile_store_dK(dk_acc_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
         int m, int n, int ld, int offset_r, int offset_c) {
 
 #if IS_GQA
-    tile_elementwise_s(*dK_tile, round_to_dst);
-    tile_atomic_add(*dK_tile, dK, m, n, ld, offset_r, offset_c);
+    dk_acc_tile_type dK_tile_slm_copy = *dK_tile;
+    a_tile_type dK_tile_f32;
+    tile_copy(dK_tile_slm_copy, dK_tile_f32);
+    tile_elementwise_s(dK_tile_f32, round_to_dst);
+    tile_atomic_add(dK_tile_f32, dK, m, n, ld, offset_r, offset_c);
 #else // MHA update
 
     a_tile_type_dst dK_tile_dst;
@@ -651,8 +700,8 @@ micro_sdpa_bwd(const global KEY_DATA_T *K, const global QRY_DATA_T *Q,
 #define S2_f32_slm_size 0
 #endif
 
-#define dK_slm_size (ugemm_kq_wg_tile_m * D_MAX * sizeof(float))
-#define dV_slm_size (ugemm_kq_wg_tile_m * D_MAX * sizeof(float))
+#define dK_slm_size (ugemm_kq_wg_tile_m * D_MAX * sizeof(SLM_DKDV_DATA_T))
+#define dV_slm_size (ugemm_kq_wg_tile_m * D_MAX * sizeof(SLM_DKDV_DATA_T))
 
 #define ugemm_slm_size \
     MAX(MAX(MAX(MAX(ugemm_kq_slm_size, ugemm_vs_slm_size), \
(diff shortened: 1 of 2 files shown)


h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/sdpa/micro_bwd.cl
```diff
@@ -450,20 +458,59 @@ inline void tile_store_k_slm(
...
+#endif
+DECLARE_2D_TILE_SLM_ADD_T(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+#define SLM_DKDV_DATA_T DST_DATA_T
+typedef dv_reduce_tile_type dv_slm_tile_type;
+typedef a_reduce_tile_type dk_slm_tile_type;
+#else
+#define SLM_DKDV_DATA_T float
+typedef dv_tile_type dv_slm_tile_type;
```
umar456: I am not a fan of the slm in the name of these types. Its confusing because these are registers.

umar456: I don't think the typedefs are needed here anyway right? Can't we use the 'dv_tile_type` directly?

src/gpu/intel/sdpa/micro_bwd.cl
```diff
@@ -450,20 +458,59 @@ inline void tile_store_k_slm(
...
+        || (ugemm_qdSt_c_type_nblock0 != ugemm_vs_c_type_nblock0) \
+        || (ugemm_qdSt_c_type_nblock1 != ugemm_vs_c_type_nblock1)
+DECLARE_2D_TILE_SLM_ADD(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+#endif
+DECLARE_2D_TILE_SLM_ADD_T(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+#define SLM_DKDV_DATA_T DST_DATA_T
```
umar456, line 480: I don't think the internal types are related to the dst type. This is just an additional type we need to support from my understanding of the problem.

src/gpu/intel/sdpa/micro_bwd.cl
```diff
@@ -450,20 +458,59 @@ inline void tile_store_k_slm(
...
+DECLARE_2D_TILE(dv_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_vs_c_type_block0, ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
+        ugemm_vs_c_type_nblock1)
+DECLARE_2D_TILE(a_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_qdSt_c_type_block0, ugemm_qdSt_c_type_block1,
+        ugemm_qdSt_c_type_nblock0, ugemm_qdSt_c_type_nblock1)
+DECLARE_2D_TILE_SLM_ADD(dv_reduce_tile_type, DST_DATA_T, SUBGROUP_SIZE,
+        ugemm_vs_c_type_block0, ugemm_vs_c_type_block1, ugemm_vs_c_type_nblock0,
+        ugemm_vs_c_type_nblock1)
+#if (ugemm_qdSt_c_type_block0 != ugemm_vs_c_type_block0) \
```
umar456: This is a strange condition. If the values are not equal you should transpose the addition? Is this a case that can be better defined by another variable? Does this only happen when TRANSPOSE_K is set? I am asking because I don't know.

src/gpu/intel/sdpa/micro_bwd.cl
```diff
@@ -503,12 +557,19 @@ inline void tile_store_dK_t(dv_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
...
 // uses qdSt tile (Bc*D) for dK update
-inline void tile_store_dK(a_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
+inline void tile_store_dK(dk_slm_tile_type *dK_tile, global DST_DATA_T_DKDV *dK,
         int m, int n, int ld, int offset_r, int offset_c) {
 
 #if IS_GQA
+#if REDUCE_DKDV_F16
+    dk_slm_tile_type dK_tile_slm_src = *dK_tile;
+    a_tile_type dK_tile_slm_dst;
+    tile_copy(dK_tile_slm_src, dK_tile_slm_dst);
```
umar456: Can we replace both branches with this approach of REDUCE_DKDV_F16 with this approach? I see that the round_to_dst is kind of doing something similar. I think the idea was that it would be more register efficient, but if its not working then I am not sure if it makes sense to do it.

umar456 (commented):
I made a quick pass with couple of comments with some quesitons. I am not super familiar with the structure and will need to do a deeper dive into the problem. I will give it another go tomorrow.

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: This was just copied from the[ kernel itself](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/sdpa/micro_bwd.cl#L352) when its creating regular SLM tiles. I think the idea is that TRANSPOSE K is the same tile as V and regular K is different. Look at the comment here:  [ // transposed dK_slm (D*Bc) matches dV tile layout](https://github.com/uxlfoundation/oneDNN/blob/main/src/gpu/intel/sdpa/micro_bwd.cl#L1089C3-L1089C55)

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: This is the tile that is stored in slm, let me know if you have a better name? https://github.com/uxlfoundation/oneDNN/pull/5202/changes#diff-0e1031c89766fac9ec154d5e7ab544a90c3eb2450e4aa086c4b62fc9a1d5781dR1149 - if I don't have the typedef I will have create 3 different if/else branches. See the link for the use of dv_slm_tile_type usage. It is the same for dk_slm_tile_type.

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: SLM is usually float, but in this case it becomes DST_DATA_T; how else can we simplify it?

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: Sure, I can simplify it.

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: Had to remove them anyway because gtests (bwd_gqa) were failing to run because of this condition.

src/gpu/intel/sdpa/micro_bwd.cl
h-sadia: One approach can be that i can have float or half slm based on REDUCED_DKDV_F16 macro, hence creating two branches in the already existing plethora of branches in the kernel. Creating a datatype seemed simpler ..... we have to choose our poison.

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

umar456 (approved):

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

h-sadia:
@TaoLv - added the smaller dim test for stability as discussed, please take a look.

Edit: the noise in dV path was curtailed by adding an opencl path, thanks to @dzarukin for pointing it out.

Re-Edit: spoke too soon, added back the small test-case with a comment explaining why. BMG fails for this test with both compilers for a larger size of GQA test with max_diff being 2.5
Important caveat for atomics:

Atomic add order is not deterministic across threads/workgroups.
Different summation orders can shift the result up or down, especially with cancellation or large dynamic range.
Dropout can amplify this sensitivity (masking + 1/q scaling), so spread in results can grow, but that is still usually better than accumulating directly in f16.

h-sadia:
https://ecmd.jf.intel.com/commander/link/jobStepDetails/jobSteps/f15f7da2-48da-f105-ae31-d4f5ef20c6a0?stepName=gtest&jobId=f15f7bd3-8fab-f125-91ed-d4f5ef20c6a0&jobName=CI_oneDNN_2026-06-03-18%3A41%3A26-hsadix_fix_sdpa_accum_type-efae527 - f16 errors fixed in this CI for latest commit.

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe-hpc
enable arch_gpu_xe-hpg-atsm
enable arch_gpu_xe-hpg-dg2
enable arch_gpu_xe-lp
enable arch_gpu_xe-lpg
enable arch_gpu_xe-lpg+
enable arch_gpu_xe2-hpg-bmg
enable arch_gpu_xe2-lpg
enable arch_gpu_xe3-lpg

src/gpu/intel/sdpa/micro_bwd.cl
```diff
@@ -475,18 +518,21 @@ inline void tile_store_dV(dv_tile_type *dV_tile_slm, global DST_DATA_T_DKDV *dV,
...
+inline void tile_store_dK_t(dv_acc_tile_type *dK_tile,
+        global DST_DATA_T_DKDV *dK, int m, int n, int ld, int offset_r,
+        int offset_c, int rem) {
 
 #if IS_GQA
-    tile_elementwise_s(*dK_tile, round_to_dst);
-    tile_atomic_add(*dK_tile, dK, m, n, ld, offset_r, offset_c);
+    dv_acc_tile_type dK_tile_slm_copy_t = *dK_tile;
+    dv_tile_type dK_tile_f32;
+    tile_copy(dK_tile_slm_copy_t, dK_tile_f32);
```
kealan-barbieri, line 533: should these copies be conditional on the case where we want to accumulate to a different type?

h-sadia: Yes, it was like that originally. But Umar's review wanted to combine the two together since we have so many if/else macros everywhere in the code. Other than f16, everything else will be f32 here. So its an extra copy.

h-sadia:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_graph
enable arch_gpu_xe2-hpg-bmg

kealan-barbieri (approved):

dzarukin:
It may happen that bigger problem goes into different path which was not modified. Division supposed to address those 6k error points but it's still 6k there, something else is wrong. Besides, lpg+ and dg2 produce zeros starting from a second tile or something. 512th point coincide for both.

h-sadia:
@dzarukin - the errors you are mentioning are separate from this ticket. a) they exist on forward pass b) DG2 is disabled on backward sdpa. Moreover, there isn't another path, I have compared cldumps of both the kernels. The difference exists in tiling size for the microgemm kernels - now you must ask that the tiling for accumulation should also change accordingly, and it does.  FP16 path is just very unstable and less accurate - there is a reason why it wasn't enabled in the first place. We can't do fp16 global atomic add because threads produce NaNs. The atomic update order is nondeterministic and multiple threads race to update the same global element, and the hardware can commit those adds in different orders run-to-run. Its worse in MHA/GQA backward case. The dK/dV are reduced across multiple Q heads (group accumulation), so contention on the same output element is high. More contenders means more possible orderings, so more rounding variance. Dropout can amplify perceived noise because masked/sparse gradients increase variance of partial sums before accumulation. So “unstable” here usually means numerically noisy and non-bitwise-deterministic, not mathematically wrong. The common mitigation is accumulating in FP32 and converting to FP16 at the end. And thats what we are doing matmul returns in fp32 which gets converted to fp16 in SLM and then back to fp32 for scratch buffers. The postprocess kernel then takes those fp32 scratch buffers to finally copy out the dst in fp16.

Edit: If you are still not okay with the explanation, I can merge my changes without the new test-case. And issues can be investigated on the side. But as far as enabling the fp16 path is concerned, it is done correctly. As discussed and proved offline. If there is a specific way to implement the *precise division* that you mentioned offline - we can keep on looking into it. What I also see is that PVC and BMG use different drivers.

dzarukin:
> @dzarukin - the errors you are mentioning are separate from this ticket. a) they exist on forward pass b) DG2 is disabled on backward sdpa. Moreover, there isn't another path, I have compared cldumps of both the kernels. The difference exists in tiling size for the microgemm kernels - now you must ask that the tiling for accumulation should also change accordingly, and it does. FP16 path is just very unstable and less accurate - there is a reason why it wasn't enabled in the first place. We can't do fp16 global atomic add because threads produce NaNs. The atomic update order is nondeterministic and multiple threads race to update the same global element, and the hardware can commit those adds in different orders run-to-run. Its worse in MHA/GQA backward case. The dK/dV are reduced across multiple Q heads (group accumulation), so contention on the same output element is high. More contenders means more possible orderings, so more rounding variance. Dropout can amplify perceived noise because masked/sparse gradients increase variance of partial sums before accumulation. So “unstable” here usually means numerically noisy and non-bitwise-deterministic, not mathematically wrong. The common mitigation is accumulating in FP32 and converting to FP16 at the end. And thats what we are doing matmul returns in fp32 which gets converted to fp16 in SLM and then back to fp32 for scratch buffers. The postprocess kernel then takes those fp32 scratch buffers to finally copy out the dst in fp16.
> 
> Edit: If you are still not okay with the explanation, I can merge my changes without the new test-case. And issues can be investigated on the side. But as far as enabling the fp16 path is concerned, it is done correctly. As discussed and proved offline. If there is a specific way to implement the _precise division_ that you mentioned offline - we can keep on looking into it. What I also see is that PVC and BMG use different drivers.

Have you checked if there's a matching reorder on benchdnn side to round down dst output to f16 and convert it back to f32? It's a standard for benchdnn technique to match the numerical behavior of internal implementation and if the fused implementation does f32->f16->f32 and benchdnn ref path doesn't, that's your culprit.

h-sadia:
I can look into the benchdnn compatibility with the internal implementation, and file a ticket if the behaviour has to be matched/changed. Dropped the change in the test-case for BMG, until we see benchdnn compatibility. @dzarukin 

> > @dzarukin - the errors you are mentioning are separate from this ticket. a) they exist on forward pass b) DG2 is disabled on backward sdpa. Moreover, there isn't another path, I have compared cldumps of both the kernels. The difference exists in tiling size for the microgemm kernels - now you must ask that the tiling for accumulation should also change accordingly, and it does. FP16 path is just very unstable and less accurate - there is a reason why it wasn't enabled in the first place. We can't do fp16 global atomic add because threads produce NaNs. The atomic update order is nondeterministic and multiple threads race to update the same global element, and the hardware can commit those adds in different orders run-to-run. Its worse in MHA/GQA backward case. The dK/dV are reduced across multiple Q heads (group accumulation), so contention on the same output element is high. More contenders means more possible orderings, so more rounding variance. Dropout can amplify perceived noise because masked/sparse gradients increase variance of partial sums before accumulation. So “unstable” here usually means numerically noisy and non-bitwise-deterministic, not mathematically wrong. The common mitigation is accumulating in FP32 and converting to FP16 at the end. And thats what we are doing matmul returns in fp32 which gets converted to fp16 in SLM and then back to fp32 for scratch buffers. The postprocess kernel then takes those fp32 scratch buffers to finally copy out the dst in fp16.
> > Edit: If you are still not okay with the explanation, I can merge my changes without the new test-case. And issues can be investigated on the side. But as far as enabling the fp16 path is concerned, it is done correctly. As discussed and proved offline. If there is a specific way to implement the _precise division_ that you mentioned offline - we can keep on looking into it. What I also see is that PVC and BMG use different drivers.
> 
> Have you checked if there's a matching reorder on benchdnn side to round down dst output to f16 and convert it back to f32? It's a standard for benchdnn technique to match the numerical behavior of internal implementation and if the fused implementation does f32->f16->f32 and benchdnn ref path doesn't, that's your culprit.

PR #5203, opened 2026-05-27
cpu: x64: matmul: restore use of transposed copy-B for N == 1

vpirogov:
Besides fixing an issue with `int4` weights handling PR #3027 included (undocumented) changes that forced transposed copy-B kernels for cases when N == 1. Dropping these changes as part of #5171 resulted in a major performance regression. This PR restores and clarifies this logic.

[Orphan CI](https://intel-ci.intel.com/f15a0bc0-30ce-f13a-b505-d4f5ef20c6a0)

Fixes [MFDNN-15115](https://jira.devtools.intel.com/browse/MFDNN-15115)

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.cpp
@@ -554,9 +554,19 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(memory_desc_t &B_md,
             bgmmc.wei_tag = blocked_B_layouts_allowed && !bgmmc.is_runtime_N
                             && !bgmmc.is_int4_weights
                     ? this->pick_blocked_B_layout(default_n_block)
-                    : bgmmc.is_int4_weights && bgmmc.N % 2 != 0
-                    ? transposed_tensor_layout_tag
                     : plain_tensor_layout_tag;
+
+            // For N == 1 force transposed layout because copy-B kernel is
+            // significantly faster.
+            if (bgmmc.wei_tag == plain_tensor_layout_tag && bgmmc.N == 1) {
+                bgmmc.wei_tag = transposed_tensor_layout_tag;
+            }
+
+            // Plain copy-B kernel does not support odd sizes for subbyte types.
+            // Using transposed version for these cases.
+            if (bgmmc.is_int4_weights && bgmmc.N % 2 != 0) {
+                bgmmc.wei_tag = transposed_tensor_layout_tag;
+            }
         }
         VCONDCHECK_BG(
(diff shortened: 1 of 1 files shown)


src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -554,9 +554,19 @@ status_t brgemm_matmul_conf_utils_t::set_or_check_B_tag(memory_desc_t &B_md,
...
                             && !bgmmc.is_int4_weights
                     ? this->pick_blocked_B_layout(default_n_block)
-                    : bgmmc.is_int4_weights && bgmmc.N % 2 != 0
-                    ? transposed_tensor_layout_tag
                     : plain_tensor_layout_tag;
+
+            // For N == 1 force transposed layout because copy-B kernel is
+            // significantly faster.
+            if (bgmmc.wei_tag == plain_tensor_layout_tag && bgmmc.N == 1) {
+                bgmmc.wei_tag = transposed_tensor_layout_tag;
```
densamoilov, line 562: This change will force matmul to use a copy B kernel even though we might have picked a blocked layout and expect the user to reorder the data. Does the plain layout with N = 1 work better than a blocked layout for the same case?

The same question applies to the int4 case.

vpirogov: This change will force copy-B kernel only when plain layout was selected for N == 1. It does not affect cases with blocked layout. Here's before/after comparision:
```
# Before
onednn_verbose,v1,primitive,exec,cpu,matmul,brg_matmul:avx10_1_512_amx,undef,src:bf16::blocked:abdc::f0 wei:bf16:a:blocked:abcd::f0 dst:bf16::blocked:abcd::f0,,,28275x1x1x1024:1x1x1024x1,3.46216
# After
onednn_verbose,v1,primitive,exec,cpu,matmul,brg_matmul:avx10_1_512_amx,undef,src:bf16::blocked:abdc::f0 wei:bf16:a:blocked:abdc::f0 dst:bf16::blocked:abcd::f0,,,28275x1x1x1024:1x1x1024x1,1.54712
```

Same behavior as before #5171 for explicit tags.

`int4` case does not change existing behavior either, as it effectively moves lines 557-558 
to a separate condition and adds comment matching the explicit tag case.

vpirogov: Plain layouts are only picked for 4D shapes, hence the regression affecting only these cases. For 2D and 3D shapes blocked format is selected.

densamoilov: You're right, I somehow didn't connect the dots.

densamoilov (approved):

dzarukin (approved):

PR #5204, opened 2026-05-27
graph: expose scratchpad api

TaoLv:
- Implementation for RFC #5173 
- Add API to query scratchpad size from a compiled partition object: `get_scratchpad_size()`
- Add API to execute a compiled partition with user-managed scratchpad buffer:
   - `compiled_partition::execute(stream, inputs, outputs, scratchpad)`
   - `sycl_interop::execute(compiled_partition, stream, inputs, outputs, scratchpad, deps)`
   - `ocl_interop::execute(compiled_partition, stream, inputs, outputs, scratchpad, deps)`
 - Backend kernel implementations:
   - if `scratchpad_buf == nullptr`, work as before.
   - if `scratchpad_buf != nullptr`, use the user-provided scratchpad buffer for computation.
 - Update verbose log to include the scratchpad information, eg: `scratchpad:user:12345` where `12345` means the size.
 - Update the GQA example to use user provided scratchpad buffer and demonstrate the API usage.
 - Update benchdnn graph driver to use user provided scratchpad buffer. Not test library allocated scratchpad intentionally to simplify the validation. We still have a few legacy examples and gtests which still rely on the old behavior.

### DIFF
--- a/doc/graph/scratchpad.md
+++ b/doc/graph/scratchpad.md
@@ -1,0 +1,95 @@
+Scratchpad Management {#dev_guide_graph_scratchpad}
+===================================================
+
+## Introduction
+
+Some compiled partitions require temporary memory (scratchpad) during execution
+to store the intermediate results. The amount of space required for the
+scratchpad depends on the compiled partition and its actual implementation.
+oneDNN Graph API supports two different scratchpad management modes:
+
+1. **User-managed scratchpad**: The user queries the required scratchpad size,
+   allocates the buffer, and passes it to the execution call. This mode gives
+   the user full control over scratchpad memory lifecycle, enabling buffer reuse
+   across multiple executions and reducing allocation overhead.
+
+2. **Library-managed scratchpad**: The library internally allocates and frees
+   the scratchpad buffer during each execution call. This mode is simple to use
+   but with its own limitations.
+
+## User-Managed Scratchpad
+
+To use user-managed scratchpad, follow these steps:
+
+1. Query the scratchpad logical tensor from the compiled partition.
+2. Check if a scratchpad is required (memory size > 0).
+3. Allocate a tensor with the queried logical tensor descriptor.
+4. Pass the scratchpad tensor to the execution call.
+
+~~~cpp
+// Compile the partition
+dnnl::graph::compiled_partition cp = partition.compile(inputs, outputs, engine);
+
+// Step 1: Query scratchpad requirements
+dnnl::graph::logical_tensor scratchpad_lt = cp.get_scratchpad_logical_tensor();
+
+// Step 2: Check if scratchpad is needed
+dnnl::graph::tensor scratchpad_ts;
+if (scratchpad_lt.get_mem_size() > 0) {
+    // Step 3: Allocate scratchpad tensor
+    void * user_scratchpad = user_allocate_method(scratchpad_lt.get_mem_size());
+    scratchpad_ts = dnnl::graph::tensor(scratchpad_lt, engine, user_scratchpad);
+}
+
+// Step 4: Execute with user-managed scratchpad
+cp.execute(stream, input_tensors, output_tensors, scratchpad_ts);
+~~~
+
+The user-managed scratchpad tensor follows the same rules and limitations as
+input and output tensors. It must be created using the same engine as partition
+compilation unless specified otherwise. If the same compiled partition is
+executed in multiple threads concurrently, a separate scratchpad buffer must be
+used per thread to ensure the thread safety. A user-managed scratchpad buffer
+can be reused across multiple executions of the same or different compiled
+partitions where the size fits.
+
+## Library-Managed Scratchpad
+
+By default, when no scratchpad tensor is provided to the execution call, the
+library allocates the required scratchpad memory internally and frees it after
+execution completes, through the allocator interfaces associated with engine
+object. This mode requires no additional user action.
+
+~~~cpp
+// Compile the partition
+dnnl::graph::compiled_partition cp = partition.compile(inputs, outputs, engine);
+
+// Execute without scratchpad - library manages it internally
+cp.execute(stream, input_tensors, output_tensors);
+~~~
+
+## Work with SYCL Graph Recording Mode
+
+User-managed scratchpad is required when working with SYCL graph recording mode.
+
+SYCL graph recording captures a sequence of kernel submissions into a graph
+object that can be replayed multiple times. During recording, all memory buffers
+accessed by the recorded kernels are bound to the graph. These buffers must
+remain valid across all subsequent replays.
+
+When library-managed scratchpad is used, the library allocates a temporary
+buffer at execution time and frees it immediately after execution completes.
+This is incompatible with SYCL graph recording.
+
+To work correctly with SYCL graph recording mode, users must pass the
+pre-allocated scratchpad tensor to the execute call during recording and keep
+the scratchpad buffer alive across all replay iterations.
+
+## API Reference
+
+| API                                                                 | Description                                                   |
+| :---                                                                | :-----------                                                  |
+| @ref dnnl::graph::compiled_partition::get_scratchpad_logical_tensor | Returns the logical tensor describing the required scratchpad |
+| @ref dnnl::graph::compiled_partition::execute                       | Executes with optional user-managed scratchpad                |
+| @ref dnnl::graph::sycl_interop::execute                             | SYCL interop execute with user-managed scratchpad             |
+| @ref dnnl::graph::ocl_interop::execute                              | OpenCL interop execute with user-managed scratchpad           |
--- a/examples/graph/gqa.cpp
+++ b/examples/graph/gqa.cpp
@@ -192,6 +192,21 @@ void bench_gqa(engine::kind ekind, logical_tensor::data_type dt,
     compiled_partition cp = partitions[0].compile(
             {query, key, scale, mask, value}, {output}, eng);
 
+    // Query scratchpad logical tensor and allocate scratchpad buffer. Here we
+    // leverage the tensor API to allocate the scratchpad buffer for
+    // convenience. In real applications, users can also manage the scratchpad
+    // buffer by:
+    // 1) query the scratchpad logical tensor from a compiled partition.
+    // 2) query the scratchpad size from the scratchpad logical tensor.
+    // 3) allocate a buffer with the required size.
+    // 4) create a tensor object with the scratchpad logical tensor, the buffer,
+    //    and the engine where the buffer is associated.
+    auto scratchpad_lt = cp.get_scratchpad_logical_tensor();
+    tensor scratchpad_ts;
+    if (scratchpad_lt.get_mem_size() > 0) {
+        scratchpad_ts = tensor(scratchpad_lt, eng);
+    }
+
     // Allocate user data.
     std::vector<float> query_data(product(q_sz));
     std::vector<float> key_data(product(kv_sz));
@@ -221,17 +236,17 @@ void bench_gqa(engine::kind ekind, logical_tensor::data_type dt,
     write_to_dnnl_tensor(value_data.data(), ts_value);
 
     // Warmup run.
-    // Execute the compiled partition of mqa.
-    cp.execute(
-            strm, {ts_query, ts_key, ts_scale, ts_mask, ts_value}, {ts_output});
+    // Execute the compiled partition of gqa with user-managed scratchpad.
+    cp.execute(strm, {ts_query, ts_key, ts_scale, ts_mask, ts_value},
+            {ts_output}, scratchpad_ts);
 
     // Wait for the computation to finish.
     strm.wait();
 
     // First run.
     auto start_first = std::chrono::steady_clock::now();
-    cp.execute(
-            strm, {ts_query, ts_key, ts_scale, ts_mask, ts_value}, {ts_output});
+    cp.execute(strm, {ts_query, ts_key, ts_scale, ts_mask, ts_value},
+            {ts_output}, scratchpad_ts);
     strm.wait();
     auto end_first = std::chrono::steady_clock::now();
     std::chrono::duration<double, std::milli> dur_first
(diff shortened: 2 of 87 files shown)


examples/graph/gqa.cpp
```diff
@@ -192,6 +192,21 @@ void bench_gqa(engine::kind ekind, logical_tensor::data_type dt,
     compiled_partition cp = partitions[0].compile(
             {query, key, scale, mask, value}, {output}, eng);
 
+    // Query scratchpad size and allocate scratchpad buffer. Here we leverage
+    // the tensor API to allcoate the scratchpad buffer for convenience. The
```
ElaineBao:
```suggestion
    // the tensor API to allocate the scratchpad buffer for convenience. The
```

rongzha1 (approved):

examples/graph/gqa.cpp
TaoLv: Thanks. Fixed.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

include/oneapi/dnnl/dnnl_graph.h
```diff
@@ -524,6 +524,38 @@ dnnl_status_t DNNL_API dnnl_graph_compiled_partition_execute(
...
+/// @param inputs A list of input tensors.
+/// @param num_outputs The number of output tensors.
+/// @param outputs A non-empty list of output tensors.
+/// @param scratchpad User-provided scratchpad buffer pointer.
+/// @returns #dnnl_success on success or a status describing the error otherwise.
+dnnl_status_t DNNL_API dnnl_graph_compiled_partition_execute_v2(
+        const_dnnl_graph_compiled_partition_t compiled_partition,
+        dnnl_stream_t stream, size_t num_inputs,
+        const_dnnl_graph_tensor_t *inputs, size_t num_outputs,
+        const_dnnl_graph_tensor_t *outputs, void *scratchpad);
```
dzarukin: Could you, please, elaborate the design choice of `void *` over `const_dnnl_graph_tensor_t`?

I can understand the choice of a dedicated argument, that should be fine, but not sure why forcing users to `get_data_handle()` every time to pass the scratchpad...

TaoLv: This is a great question, Dima. Here are what I considered and the problems I met:

1. Scratchpad is neither an input nor an output. So I chose to take it as a separate argument.
2. The designed flow is: the frameworks query the scratchpad size from the library, allocate the buffer from frameworks' infrastructure, then pass the buffer pointer to the execute API. So I assume that there is no `get_data_handle()` every time. This is also based on the fact that frameworks and graph API only support USM.
3. If passing scratchpad as a tensor, it may take additional overhead to construct the tensor around the user buffer. The user will need to either construct a logical tensor by themselves or the library will need to provide an additional query interface (similar to query scratchpad md from a primitive).
4. There are fields in logical tensor which are not very meaningful for scratchpad, e.g. the logical tensor ID. Since it's user provided or accessible, we will need to define what's the value of a scratchpad tensor ID.
5. I agree that one benefit of passing scratchpad as a tensor is that it specifies the associated engine of buffer. With the raw pointer scratchpad, I have to document that the scratchpad buffer should come from the same engine used for the execution.

dzarukin: Maybe then query for a logical tensor for scratchpad and use get_mem_size() method to get the size and allocate a buffer for scratchpad?

I feel we better use `tensor` for scratchpad to avoid potential misuse of host memory when device is involved and for consistency with execute APIs.

TaoLv: Thank you for the suggestions, Dima. I will try to implement that and see how it works. Firstly, I think we will need to design the fields of scratchpad logical tensor. Will the below look good to you?

```
- ID: a magic number (eg. SIZE_MAX).
- ndims = 1
- dtype = u8
- dims = (size,)
- strides = (1,)
```

BTW, could you please help to understand why it can avoid potential misuse of host memory? As the buffer handle is provided by user, they still can mistakenly combine a device engine with a host memory to get a `tensor` object or `memory` object. Thanks!

dzarukin: Is ID really important? I'd think any number not present in a partition should work but if SIZE_MAX is the only guaranteed, so be it. Rest looks OK and aligned with primitives scratchpad definition.

In case when host memory is passed into a tensor meant to run on a device, there's a dedicated check for `mayiuse_system_memory_allocators` (at least SYCL's `dnnl_sycl_interop_memory_create_v2` does that. If tensor interop API doesn't do that, we might need to add that to catch such errors.

ElaineBao (approved):

TaoLv:
I added another commit implementing the tensor-base scratchpad API. You can compare the last two commits to see the differences of the two API styles.  I will rebase and squash the changes later.
@dzarukin  @LuFinch @ElaineBao  @rongzha1 , please take another look. Thanks!

LuFinch:
LGTM! The tensor-base scratchpad API is clear for framework.

TaoLv:
Thank you for the feedback, @LuFinch !

@dzarukin @rongzha1  @ElaineBao  Could you please check if the new APIs look good to you? Thanks!

include/oneapi/dnnl/dnnl_graph_ocl.hpp
```diff
@@ -139,6 +139,44 @@ inline cl_event execute(compiled_partition &c_partition, stream &astream,
...
+/// @param astream Stream object to run over
+/// @param inputs Arguments map.
+/// @param outputs Arguments map.
+/// @param scratchpad User-provided scratchpad tensor. If not provided, the
+///     implementation will allocate scratchpad internally.
+/// @param deps Optional vector with `cl_event` dependencies.
+/// @returns Output event.
+inline cl_event execute(compiled_partition &c_partition, stream &astream,
+        const std::vector<tensor> &inputs, std::vector<tensor> &outputs,
+        const tensor &scratchpad, const std::vector<cl_event> &deps = {}) {
```
dzarukin, line 155: Minor: I think it fits the previous definition with a default argument of tensor same as for regular API. Did you intentionally keep this version separately?

Once scratchpad becomes a requirement (next major release), the default will be stripped off.

src/graph/backend/dnnl/kernels/kernel_base.hpp
```diff
@@ -121,6 +131,11 @@ using FCreateKernel = std::function<kernel_ptr(void)>;
         return #name; \
     }
 
+#define DEF_KERNEL_METHOD_SCRATCHPAD_SIZE() \
+    size_t get_scratchpad_size() const override { \
+        return memory_planner_.total_internal_temporary_size(); \
+    }
```
dzarukin, line 137: Minor: with all the tools in-place (Copilot, clang-format, clang-tidy, coverity, etc.), I feel it's better to avoid using "global" macros and stick to C++ narrative. It seems that `memory_planner_` is a base abstraction and this method altogether can be moved to a base class.

If accessing it is problematic in a common space, memory_planner() method can be introduced to access it from a parent class (if needed). If it can be empty, maybe introduce and extra method that would check its non-emptiness. It would be easier to manage it later.

dzarukin (commented):
LGTM, left two minor comments. Need to craft a doc piece to specify the new ability.

@vpirogov, would we need to document the limitation when used with SYCL graph that user-managed scratchpad is a must?

vpirogov:
> LGTM, left two minor comments. Need to craft a doc piece to specify the new ability.
> 
> @vpirogov, would we need to document the limitation when used with SYCL graph that user-managed scratchpad is a must?

This would be very helpful. I think we can even quip something when we detect recording mode and library-managed scratchpad.

include/oneapi/dnnl/dnnl_graph_ocl.hpp
TaoLv: I intentionally put the argument `scratchpad` near to `inputs` and `outputs` in the argument list. We already had default value for `deps` before this PR. If we want to add `scratchpad` with default value, I think we will have to add it at the end of the argument list, so we don't break existing API calls with `deps` passed explicitly.

```
inline cl_event execute(compiled_partition &c_partition, stream &astream,
        const std::vector<tensor> &inputs, std::vector<tensor> &outputs,
        const std::vector<cl_event> &deps = {}, const tensor &scratchpad = tensor());
``` 

Which one looks better to you?

src/graph/backend/dnnl/kernels/kernel_base.hpp
TaoLv: Good suggestion, Dima. I will think how to do that (probably as a follow-up task).
I noticed two problems when implementing this `get_scratchpad_size()` method in the base class:

- The decomposition kernels actually don't have normal `memory_planner_`. They count memory usage in a different way.
- A base kernel (eg. `sdp_base_t`) may delegate this method to another kernel after compilation (`eg. sdp_primitive_kernel_t` or `larger_partition_kernel_t`).

So I decided to make it pure virtual and let each kernel to implement their own method. The global macro is just a trick to reduce code duplication. Please let me know if any idea to handle this situation. Thanks!

include/oneapi/dnnl/dnnl_graph_ocl.hpp
dzarukin: Let's keep it explicit then, thanks for reminding about API consistency.

src/graph/backend/dnnl/kernels/kernel_base.hpp
dzarukin: Keep it virtual with:
```
if (memory_planner()) return memory_planner_call(...);
return 0;
```

Other kernels where it's not true will override it with what they need. But it's minor, up to your consideration. Feel free to close any time.

TaoLv:
I rebased the commits and added a document for scratchpad. Please check ~

doc/graph/scratchpad.md
```diff
@@ -0,0 +1,94 @@
...
+// Step 4: Execute with user-managed scratchpad
+cp.execute(stream, input_tensors, output_tensors, scratchpad_ts);
+~~~
+
+The user-managed scratchpad tensor can be reused across multiple executions of
+the same or different compiled partitions. If execute the same compiled
+partition in multiple threads concurrently, the user need to guarantee the
+thread safety of the user-managed scratchpad. If the user provides scratchpad
+tensor to a compiled partition, this tensor must be created using the same
+engine that the compiled partition uses.
```
dzarukin: Minor: maybe rephrase that in a way that scratchpad tensor follows same rules and limitations as input and output tensors (if we have them documented, things like same engine for a partition unless specified otherwise, separate buffer if same partition is used in different threads, etc.)?

doc/graph/scratchpad.md
```diff
@@ -0,0 +1,94 @@
...
+the scratchpad buffer alive across all replay iterations.
+
+## API Reference
+
+| API | Description |
+|-----|-------------|
+| @ref dnnl::graph::compiled_partition::get_scratchpad_logical_tensor | Returns the logical tensor describing the required scratchpad |
+| @ref dnnl::graph::compiled_partition::execute | Executes with optional user-managed scratchpad |
+| @ref dnnl::graph::sycl_interop::execute | SYCL interop execute with user-managed scratchpad |
+| @ref dnnl::graph::ocl_interop::execute | OpenCL interop execute with user-managed scratchpad |
```
dzarukin: Nit: we usually try to keep table edges aligned between columns for nicer raw view.

dzarukin (approved):

doc/graph/scratchpad.md
TaoLv: Thanks. Updated.

doc/graph/scratchpad.md
TaoLv: Thanks. Updated.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

ElaineBao (approved):

include/oneapi/dnnl/dnnl_graph_sycl.hpp
```diff
@@ -114,6 +114,41 @@ inline sycl::event execute(compiled_partition &c_partition, stream &astream,
...
+    for (auto &out : outputs) {
+        c_outputs.push_back(out.get());
+    }
+
+    sycl::event sycl_event;
+    error::wrap_c_api(
+            dnnl_graph_sycl_interop_compiled_partition_execute_v2(
+                    c_partition.get(), astream.get(), c_inputs.size(),
+                    c_inputs.data(), c_outputs.size(), c_outputs.data(),
+                    scratchpad.get(), &deps, &sycl_event),
```
dzarukin: My tests with native graph recording feature says this should allow empty scratchpad:
```suggestion
                    scratchpad.get(true), &deps, &sycl_event),
```
If you have a better idea how to handle the situation when scratchpad is not needed, feel free to implement that. It's relevant for all C++ APIs.

dzarukin (commented):
I've also found that batch_norm_bwd didn't pass attributes with user scratchpad resulting in empty scratchpad needed (from md) while it was actually needed. Here's the patch:
```
diff --git a/src/graph/backend/dnnl/executables/batch_norm.cpp b/src/graph/backend/dnnl/executables/batch_norm.cpp
index 8573eb2065..041d0d367a 100644
--- a/src/graph/backend/dnnl/executables/batch_norm.cpp
+++ b/src/graph/backend/dnnl/executables/batch_norm.cpp
@@ -792,11 +792,11 @@ batchnorm_bwd_executable_t::desc_t batchnorm_bwd_executable_t::create_desc(
     }

     auto forward_hints = dnnl::batch_normalization_forward::primitive_desc(
-            p_engine, prop_kind::forward_training, src, src, epsilon, flags);
+            p_engine, prop_kind::forward_training, src, src, epsilon, flags, prm_attr);

     dnnl::batch_normalization_backward::primitive_desc pd(p_engine,
             prop_kind::backward, src, forward_hints.dst_desc(), src, epsilon,
-            flags, forward_hints);
+            flags, forward_hints, prm_attr);

     pd_cache.insert({op.get(), pd});
     return {pd, false};
```

include/oneapi/dnnl/dnnl_graph_sycl.hpp
TaoLv: Good catch, Dima. I set the `allow_empty` flag in the native C++ API, but forgot to do that in SYCL and OCL interop API.

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

PR #5206, opened 2026-05-27
Update third_party/xbyak_aarch64 from v1.2.0 to v1.3.0

henry-gar:
Updates third_party/xbyak_aach64 to include:

- Commit 10c8046: [feat: enable fp16 feature detection](https://github.com/fujitsu/xbyak_aarch64/commit/10c8046d7fb07afb3b9a1249738ff1e8f1ad48e3)
- Commit e7bf12f: [use unique_ptr for CpuInfo to avoid memory leak](https://github.com/fujitsu/xbyak_aarch64/commit/e7bf12f38c488d16abdd4bef3d547df2e54a5575)
- Register manager added in https://github.com/fujitsu/xbyak_aarch64/pull/114

Only files under xbyak_aarch64/src and xbyak_aarch64/xbyak_aarch64 have been added, as per the structure under oneDNN currently.

### DIFF
--- a/third_party/xbyak_aarch64/src/err_impl.h
+++ b/third_party/xbyak_aarch64/src/err_impl.h
@@ -1,6 +1,7 @@
 #pragma once
 /*******************************************************************************
  * Copyright 2020-2023 FUJITSU LIMITED
+ * Copyright 2026 Arm Ltd. and affiliates
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,7 +20,7 @@ Error::Error(int err) : err_(err), msg_("") {
   if (err_ <= 0)
     return;
   fprintf(stderr, "bad err=%d in Xbyak::Error\n", err_);
-  static const char *tbl[32] = {
+  static const char *tbl[] = {
       "none",
       "code is too big",
       "label is redefined",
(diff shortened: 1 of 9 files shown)


jondea:
Looks like you need to change the commit message. But otherwise looks fine.

I would also support the addition of an example usage within a oneDNN primitive (as a separate commit, but in the same PR) just as a test case.

Sqvid (changes requested):
I think it would be better to just do a version bump as a whole.
[`v1.3.0` is available as a hash](https://github.com/fujitsu/xbyak_aarch64/commit/3f8c682b9c6ff562dc008c7d1b8307a683f05d20) so let's just bump to it instead of having a diverging history here. I believe this is [what we did last time too](https://github.com/uxlfoundation/oneDNN/commit/250b01e10c4c9ae63eac0f250d4b025baa3d46e4).
What do you think @jondea?

jondea:
Yes definitely bump to a hash (I think we remove the tests and some other bits) rather than cherry picking. I missed that detail, sorry!

henry-gar:
I've bumped the commit has to v1.3.0 and updated the original commit message. I've discussed with @jondea  about the oneDNN primitive - I don't currently have an example for a oneDNN primitive, and due to time constraints, we agreed to continue without it.

Sqvid:
> Commit 10c8046: [feat: enable fp16 feature detection](https://github.com/fujitsu/xbyak_aarch64/commit/10c8046d7fb07afb3b9a1249738ff1e8f1ad48e3)

I don't actually see those changes reflected here. Could you check again?

henry-gar:
@Sqvid fixed now - thank you for spotting! Now includes both the listed commits (excluding the changes in sample/cpuinfo.cpp).

jondea (approved):

Sqvid (approved):

PR #5207, opened 2026-05-27
cpu: aarch64: fix register guard stack calculation and enable SVE binary injection

Sqvid:
# Description
The stack space calculation for general purpose registers did not take the 16-alignment padding into account. This caused corrupted loads when guarding an odd number of registers.

Fixing this, along with the changes in https://github.com/uxlfoundation/oneDNN/pull/5184, mean we can enable SVE binary post-op injection.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/injectors/injector_utils.cpp
+++ b/src/cpu/aarch64/injectors/injector_utils.cpp
@@ -127,23 +127,27 @@ register_preserve_guard_t<isa>::~register_preserve_guard_t() {
 
 template <cpu_isa_t isa>
 size_t register_preserve_guard_t<isa>::calc_vmm_to_preserve_size_bytes(
-        const std::initializer_list<Xbyak_aarch64::VReg> &vmm_to_preserve)
+        const std::initializer_list<Xbyak_aarch64::VReg> vmm_to_preserve)
         const {
 
-    return std::accumulate(vmm_to_preserve.begin(), vmm_to_preserve.end(),
-            std::size_t(0u),
-            [](std::size_t accum, const Xbyak_aarch64::VReg &vmm) {
-        return accum + simd_bytes(isa);
-    });
+    return vmm_to_preserve.size() * simd_bytes(isa);
 }
 
 template <cpu_isa_t isa>
 size_t register_preserve_guard_t<isa>::stack_space_occupied() const {
-    constexpr static size_t reg64_size = 8;
-    const size_t stack_space_occupied
-            = vmm_to_preserve_size_bytes_ + gpr_regs_.size() * reg64_size;
+    constexpr static size_t reg64_size_bytes = 8;
+    constexpr static size_t stack_alignment_bytes = 16;
 
-    return stack_space_occupied;
+    // If gpr_regs_ need to be backed up then round up to the nearest multiple
+    // of 16 to preserve stack alignment.
+    const size_t gpr_space = utils::rnd_up(
+            gpr_regs_.size() * reg64_size_bytes, stack_alignment_bytes);
+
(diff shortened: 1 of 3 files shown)


src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -127,23 +127,26 @@ register_preserve_guard_t<isa>::~register_preserve_guard_t() {
...
-    });
+    return vmm_to_preserve.size() * simd_bytes(isa);
 }
 
 template <cpu_isa_t isa>
 size_t register_preserve_guard_t<isa>::stack_space_occupied() const {
-    constexpr static size_t reg64_size = 8;
-    const size_t stack_space_occupied
-            = vmm_to_preserve_size_bytes_ + gpr_regs_.size() * reg64_size;
+    constexpr static size_t stack_alignment = 16;
```
jondea: 16 what? Apples? Lemons?! People!?!! Cars!??!!11

src/cpu/aarch64/injectors/injector_utils.cpp
```diff
@@ -127,23 +127,26 @@ register_preserve_guard_t<isa>::~register_preserve_guard_t() {
...
 
-    return stack_space_occupied;
+    // If gpr_regs_ need to be backed up make sure they are 16-aligned
+    const size_t gpr_space = gpr_regs_.empty()
```
jondea: Why do we need to multiply by 2 as well as stack alignment? Also, size should be zero when empty, so we shouldn't need this ternary

jondea (changes requested):
Great stuff, thanks. Just two minor comments

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: I thought Lemons was obvious from the context but fiiiine

Sqvid:
This has an intermittent failure on `test_convtranspose_execute_subgraph_int8.ConvTranspose1d2d3dBinary`.
See Attempt 1 and 2 of: https://github.com/uxlfoundation/oneDNN/actions/runs/26521221485/attempts/1

I need to debug that before proceeding.

src/cpu/aarch64/injectors/injector_utils.cpp
Sqvid: We are dividing by two to because we try to store registers a pair at a time.
Essentially we want this relationship between the number of registers vs stack space:
```
gprs = 0: 0 * 16 = 0
gprs = 1: 1 * 16 = 16
gprs = 2: 1 * 16 = 16
gprs = 3: 2 * 16 = 32
gprs = 4: 2 * 16 = 32
gprs = 5: 3 * 16 = 48
gprs = 6: 3 * 16 = 48
gprs = 7: 4 * 16 = 64
gprs = 8: 4 * 16 = 64
```

jondea: I think that gives us the correct number, but the logic is still expressed in a way that hides the meaning. We are rounding num_gprs * gpr_size (8 bytes) up to the next 16 bytes. I.e.

```
gpr_space = utils::rnd_up(gpr_regs_.size()*8, stack_alignment_bytes);
```

Sqvid: Yeah that would be more clear. Thanks for the suggestion!

Sqvid: Done.

jondea (approved):

renato-arantes (approved):

PR #5211, opened 2026-05-27
[GPU] Refactor required s4 upconversion

Simonsays095:
Addresses [MFDNN-14317](https://jira.devtools.intel.com/browse/MFDNN-14317). Xe3p no longer supports mixed s4/s8 DPAS. The previous solution was to add a few state variables and fix on-the-fly during kernel generation by upconverting s4 to s8. This results in higher register pressure, including several cases where the kernel runs out of registers and fails to compile.

Instead, this PR just changes the internal type to s8 before kernel generation, and then looks for valid kernels in the catalog. Mixed s4/s8 kernels are no longer considered. This way, kernel performance is well-described by the performance model and we skip generation attempts that are likely to fail.

No performance or correctness changes are expected, but I do expect error messages to decrease (part of [MFDNN-15085](https://jira.devtools.intel.com/browse/MFDNN-15085)).

### DIFF
--- a/src/gpu/intel/gemm/jit/generator/pieces/gemm_setup.cxx
+++ b/src/gpu/intel/gemm/jit/generator/pieces/gemm_setup.cxx
@@ -1635,12 +1635,6 @@ bool Generator<hw>::gemmAccumulateCSetup(GEMMProblem &problem, GEMMStrategy &str
     state.repackA |= problem.earlyDequantizeA() && !slmA;
     state.repackB |= problem.earlyDequantizeB() && !slmB;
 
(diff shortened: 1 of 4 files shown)


Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

Simonsays095:
make test perf-gpu
set primitive=gpu:gemm

dyoussif (approved):

h-sadia (approved):

PR #5215, opened 2026-05-28
rfc: appendix to improve functional coverage - implementation proposals

dzarukin:
[Link](https://github.com/uxlfoundation/oneDNN/blob/dzarukin/rfcs/improve_functional_coverage_appendix/rfcs/20260421-improve-functional-coverage/Appendix.md) to a rendered document.

### DIFF
--- a/rfcs/20260421-improve-functional-coverage/Appendix.md
+++ b/rfcs/20260421-improve-functional-coverage/Appendix.md
@@ -1,0 +1,57 @@
+# Introduction
+This RFC is an appendix to the bullet of ~80% optimization coverage.
+Benchdnn case generation tool PoC revealed several issues that require some additional measures that will be discussed below.
+Issue: benchdnn provides a flexible way to skip test cases but doesn’t have instruments to analyze itemized skipped statistics.
+
+Why is it an issue?
+
+When a test case is generated, run and skipped, it’s hard to say what is the exact reason for skipping without digging into which is not an easy task itself –
+dispatch verbose may return too generic messages, or report a multi-return-point function as a single spot without telling which condition exactly failed.
+It is helpful to see the distribution among skipped cases to make sure that a generator is prevented from crafting invalid cases and helped to fine tune the amount of corner cases that are not supported.
+However, it turned out that each driver has lengthy logic about what it supports and what not and adjusting the generator
+to each and one of them and introducing a long-named list of reasons for skipped cases doesn't seem to be the best approach.
+Besides that, that logic can cover bugs unintentionally, e.g. Grouped Matmul with post-ops on GPU return skipped because GPU supports only Source sparse descriptor (that’s what benchdnn says so far).
+Additionally, it doesn't feel right that the library provides API that simply returns unimplemented for a given set of settings, unless documented.
+
+# Proposal
+With the last sentence in mind, even though it is a broad thought, the proposal is to drop most of the logic that makes benchdnn producing nice SKIPPED statuses for certain reasons.
+Such proposal leads to the following consequences:
+* Unimplemented points must be implemented in ref implementations to make the library provide the meaningful results. While it's not possible for all API pieces, it's possible to cover most of them.
+* Exception cases (listed below) would still be skipped on benchdnn side as their implementation on the library side doesn't provide any benefits for end users.
+  Alternatively, it can be restricted from API side (src/common/ directory) for a given backend. However, it would limit the sharing capability of same input files between backends which is opposite to the desired goal.
+* A special place in API is taken by data types.
+  According to the documentation, such support is declared to be hardware-dependent and mustn't be supported in reference implementation if hardware doesn't support it natively.
+  Currently, benchdnn handles this skipping logic via library functions (special holes with exported symbols) to verify if the library supports a given data type or not after a pd returns unimplemented status.
+  Instead, the proposal is to provide a new special status that would indicate that the library doesn't support a given data type,
+  and benchdnn to accept it and skip unconditionally without doing extra verification on its side with the same library functions that report unimplemented when pd is created.
+  This action delegates responsibility to the library which would also provide the user with a clearer message if data types were not supported.
+
+The list of unsupported functionalities:
+* Convolutions:
+    - Winograd algorithm. This algorithm provides tentative advantage on a short number of platforms with low compute capabilities. Suggestion is to unconditionally let this algorithm be skipped if unimplemented.
+    - Depthwise post-op. Its support is limited to old CPU platforms only. Same suggestion for this post-op kind.
+* RNNs:
+    - Support for RNN is sparse and may exclude an algorithm support completely for a backend, or combination clusters of algorithm plus specific data types support.
+      Suggestion is to let RNN be unconditionally skipped as it's a seldomly requested and outdated functionality to be worried about.
+* Matmuls:
+    - `packed` sparse kind is supported for CPU with AMX unit for weights execution argument only. The suggestion is to entirely skip cases with this kind.
+* Reorders:
+    - Runtime dimensions and compensation buffers on GPU are not supported by design. They are skipped now and the suggestion is to continue this practice.
+
+There's a PoC branch which implements all missing pieces in reference:
+`dzarukin/benchdnn_remove_unimpl` in case you'd like to get familiar with all the changes that are triggered by *existing* testing coverage.
+It's hard to say if new things emerge once the coverage going to be extended with a generation tool.
+
+# Summary
+Proposal 1:
+```cpp
+typedef enum {
+    ...
+    // The data type is not supported on a system.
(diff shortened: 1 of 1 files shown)


rfcs/20260421-improve-functional-coverage/Appendix.md
```diff
@@ -0,0 +1,57 @@
...
+to each and one of them and introducing a long-named list of reasons for skipped cases doesn't seem to be the best approach.
+Besides that, that logic can cover bugs unintentionally, e.g. Grouped Matmul with post-ops on GPU return skipped because GPU supports only Source sparse descriptor (that’s what benchdnn says so far).
+Additionally, it doesn't feel right that the library provides API that simply returns unimplemented for a given set of settings, unless documented.
+
+# Proposal
+With the last sentence in mind, even though it is a broad thought, the proposal is to drop most of the logic that makes benchdnn producing nice SKIPPED statuses for certain reasons.
+Such proposal leads to the following consequences:
+* Unimplemented points must be implemented in ref implementations to make the library provide the meaningful results. While it's not possible for all API pieces, it's possible to cover most of them.
+* Exception cases (listed below) would still be skipped on benchdnn side as their implementation on the library side doesn't provide any benefits for end users.
```
rjoursler, line 20: This overall design feels brittle to me.  A few issues that occur to me:

* How do we handle the multi-stage roll-out complex of features?
* How do we ensure the (now more limited) skip logic doesn't become outdated?
* How do we differentiate between cases we expect to have optimized/non-optimized support? These are just as much of an error as missing coverage all together.
* The test generator and benchdnn skip logic need to be kept coherent

I suspect the core issue is that we are trying to do too much in a single test suite. What if we just generate different test kinds. I am thinking something like:

* Optimized Coverage tests - Test that we generate a non-reference primitive for the given problem, unimplemented/reference implementations are errors
* Functional Coverage tests - Tests that we generate primitives for the given problem, unimplemented is an error.
* Optimized Correctness tests: Tests that optimized primitives are functionally correct, unimplemented is fine/ref is fine but skipped.
* Reference Correctness tests: Tests that reference primitives are functionally correct, unimplemented is fine.

There are a few benefits I see to doing this:  first, we remove the skipping logic from benchdnn (almost) entirely and the logic consolidated in the test generator.  Second, this turns the testing into an allow list rather than a skip list so that outdated skip logic does not hide coverage errors. Finally, we can better balance the size of each suite to account for the cost to perform the test.

dzarukin: > How do we handle the multi-stage roll-out complex of features?

I'm not sure what you mean by that but the bar gets higher with reference implementation for both backends, not for CPU only as it usually happened before. Rest remains unchanged.

> How do we ensure the skip logic doesn't become outdated?

By using fundamental statements according to suggestions in the document and not adding more to them. The proposal doesn't change things how it's done now but decreases the size of while lists.

> How do we differentiate between cases we expect to have optimized/non-optimized support? These are just as much of an error as missing coverage all together.

Performance validation is supposed to catch that. It has nothing to do with this proposal per se. The tracking of changing implementation can only be done with previous state and it's not covered, and won't be, in any functional validation that we have.

> The test generator and benchdnn skip logic need to be kept coherent

The only reason to keep them consistent is for generator to be more effective (generate less to test more) but in the abstract ideal world generator should test benchdnn as a black box. And the proposal moves us one step closer to this ideal world where to be efficient the generator shouldn't know more than it should about the library internals.
The skipping logic can't be moved to the generator as there's still an interactive mode which must be user-friendly and explain invalidity of test cases the same way it does for a generator.

echeresh: > How do we differentiate between cases we expect to have optimized/non-optimized support? These are just as much of an error as missing coverage all together.

Do you have any ideas how to improve this? Since our optimized kernel coverage (mainly matmul) is pretty fragemented, implementing regression testing seems like the best ROI. We can track implementation regressions (i.e. jit -> ref) in CI/Nightly, this should be enough to catch dispatch bugs that we care about.

The alternative approach - documenting and maintaining the full support matrix (and keeping it up-to-date with the implementation) would be significantly more expensive and a much bigger effort.

dzarukin: It sounds like a different kind of testing to me where we get lists from the past mark, run them with mode=I and then compare the output of the current branch state and check for regressions which are to be defined (if there are more than 2 impls, it's not obvious what should be marked as degradation and what's is an upgrade).

rjoursler: > Do you have any ideas how to improve this? Since our optimized kernel coverage (mainly matmul) is pretty fragmented, implementing regression testing seems like the best ROI.

I am more thinking about a multi-pronged approach. The fundamental issue with regression tests is that they are very slow: they cannot be parallelized and need to run the workload multiple times to collect sufficient data. If we have a initial test only checking kernel generation, this can run very fast. Essentially this would be a dispatching smoke test before we spend time on a more costly test suite.

> It sounds like a different kind of testing to me where we get lists from the past mark, run them with mode=I and then compare the output of the current branch state and check for regressions which are to be defined 

This would have a similar effect, but I am just thinking it may be simpler to use a generated suite like: `harness_optimized_matmul_gpu_generated` to avoid interactions between oneDNN builds (although if we do enable oneDNN builds to interact, we could speed up functional testing by just skipping the correctness tests when the generated kernel binary is identical).

> The skipping logic can't be moved to the generator as there's still an interactive mode which must be user-friendly and explain invalidity of test cases the same way it does for a generator.

Why not just rely on DNNL_VERBOSE logs to do the explaining? Any explanation logic we put into benchdnn is something that users of the oneDNN API cannot take advantage of.

echeresh: > I am more thinking about a multi-pronged approach. The fundamental issue with regression tests is that they are very slow

I'm thinking about build-to-build comparison. We need three ingredients:

- Persistent logs from one (or a few) recent CI/Nightly runs
- Implementation name logging during regular benchdnn runs (easy)
- Post-processing to compare the current vs older runs
    - If benchdnn lines match but the implementation name is changed from jit to something else - flag this in CI

The comparison will be fast and the methodology is mostly mechanical, there isn't much to invent or think about.

dzarukin: > Why not just rely on DNNL_VERBOSE logs to do the explaining? Any explanation logic we put into benchdnn is something that users of the oneDNN API cannot take advantage of.

Looks like I wasn't quite clear in my point. At this point benchdnn knows (from devs) which unimplemented is expected and which is not. If skip_unimpl is removed, it means it doesn't know it anymore, and the question is what should user think if they run into an unimplemented case in interactive mode (considering that generator would walk around all of such cases)? Messaging about the reasoning is secondary, it can be inspected from verbose if needed.

Secondly, how this approach would work for features implemented for one backend and not for the other and it would make zero value in doing so? I still want to test the feature for the supported backend, and want to have a single list for everything.

rjoursler: > Secondly, how this approach would work for features implemented for one backend and not for the other and it would make zero value in doing so? I still want to test the feature for the supported backend, and want to have a single list for everything.

This makes sense and is related to what I am trying to say, let me see if I can restate what I am seeing in a more clear manner. 

Fundamentally, there are multiple kinds of tests relevant to a development tasks. In my initial comment, I ended up trying to summarize some of the cases that seem useful to me and note that the condition for a failure changes based on the goal of the test. In the current design, we are embedding the failure logic into benchdnn. Architecturally, this makes it difficult to create a test targeting these variable failure conditions, and as a consequence, limits flexibility in testing. What I am proposing as that most of failure logic should be encoded in the generated tests, and for the purposes of pass/fail within benchdnn, we can just rely on the user supplied conditions. This should significantly increase testing flexibility/extensibility while being a negligible change in complexity as were are mostly moving the logic from benchdnn into the test generator.

dzarukin: Did I get you right that you want to control what should be reported as an error? And you say that generated test cases should be in charge of defining that behavior, too?

rjoursler: Yes, as a concrete example, we could add add an `--allow-unimpl` flag (kind of analogous to [--check-ref-impl](https://github.com/uxlfoundation/oneDNN/blob/main/tests/benchdnn/doc/knobs_common.md#--check-ref-impl) ) to benchdnn. Then to implement unsupported RNN functionality as you suggest, the test generator would create cases like:

```
--reset --allow-unimpl --alg=VANILLA_RNN ...
```

The test generator itself would not need customization controls until it is practically useful.

vpirogov (approved):

PR #5216, opened 2026-05-28
Functional fixes

dzarukin:
* Remove sum post-op as allowed from Layer Normalization (ping @Sqvid)
* Remove dead code from cpu simple reorder (this scratchpad is not used anywhere with the move to runtime scales)
* Fix serialization for sparse desc (will be needed in the future)

### DIFF
--- a/src/common/layer_normalization.cpp
+++ b/src/common/layer_normalization.cpp
@@ -193,7 +193,7 @@ status_t layer_normalization_attr_check(const layer_normalization_desc_t &desc,
         if (!attr->post_ops_.has_default_values()) {
             const auto &po = attr->post_ops_;
             using namespace primitive_kind;
-            VCHECK_LNORM_UNIMPL(po.has_default_values({binary, eltwise, sum}),
+            VCHECK_LNORM_UNIMPL(po.has_default_values({binary, eltwise}),
                     VERBOSE_UNSUPPORTED_POSTOP);
 
             // Note: verbose support is inside the call.
--- a/src/common/primitive_serialization.cpp
+++ b/src/common/primitive_serialization.cpp
@@ -58,6 +58,14 @@ status_t serialize_desc(
     return status::success;
 }
 
+void serialize_md_blocking_desc(
+        serialization_stream_t &sstream, const blocking_desc_t &bd, int ndims) {
+    sstream.append_array(ndims, bd.strides);
+    sstream.append(bd.inner_nblks);
+    sstream.append_array(bd.inner_nblks, bd.inner_blks);
+    sstream.append_array(bd.inner_nblks, bd.inner_idxs);
+}
+
 void serialize(serialization_stream_t &sstream, const memory_desc_t &md) {
     sstream.append(md.ndims);
     sstream.append_array(md.ndims, md.dims);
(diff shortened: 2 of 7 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_lnorm
enable benchdnn_reorder

Sqvid:
Thanks for patching this @dzarukin. What about `eltwise` though? It seems only `binary` is mentioned in the docs:
https://github.com/uxlfoundation/oneDNN/blob/08d9bb08719da9d703dba4e0df5a55cc952757eb/doc/primitives/layer_normalization.md?plain=1#L171-L174

So it would seem either the code or the documentation needs a fixup as well.

src/common/primitive_serialization.cpp
```diff
@@ -68,17 +76,32 @@ void serialize(serialization_stream_t &sstream, const memory_desc_t &md) {
...
-        case format_kind::sparse:
         case format_kind::undef:
         case format_kind::any:
         case format_kind::host_scalar: break; // no additional data to serialize
+        case format_kind::sparse:
+            sstream.append(md.format_desc.sparse_desc.encoding);
+            sstream.append(md.format_desc.sparse_desc.nnz);
+            sstream.append_array(sparse_desc_t::max_metadata_types,
+                    md.format_desc.sparse_desc.metadata_types);
+            if (md.format_desc.sparse_desc.encoding
```
densamoilov: Why is this needed for serialization but not needed for hashing?

src/common/layer_normalization.cpp
```diff
@@ -193,7 +193,7 @@ status_t layer_normalization_attr_check(const layer_normalization_desc_t &desc,
         if (!attr->post_ops_.has_default_values()) {
             const auto &po = attr->post_ops_;
             using namespace primitive_kind;
-            VCHECK_LNORM_UNIMPL(po.has_default_values({binary, eltwise, sum}),
+            VCHECK_LNORM_UNIMPL(po.has_default_values({binary, eltwise}),
```
rjoursler, line 196: What is the reasoning behind not supporting `sum` post-ops? To me, post-op support divergence between different primitives generally feels like it is arbitrarily increasing user complexity.

src/common/primitive_serialization.cpp
dzarukin: I've got testing issues with persistent cache returning incorrect pd after adding CSR support on GPU, but not with the primitive cache, though I don't know if it's a coincidence or not. But I can patch it up, too.
I guess, it will get hit too once pushed to OV production (at least grouped_mm piece).

src/common/layer_normalization.cpp
dzarukin: The reason is lack of use cases and lack of existing support.

dzarukin:
> Thanks for patching this @dzarukin. What about `eltwise` though? It seems only `binary` is mentioned in the docs:
> 
> https://github.com/uxlfoundation/oneDNN/blob/08d9bb08719da9d703dba4e0df5a55cc952757eb/doc/primitives/layer_normalization.md?plain=1#L171-L174
> 
> So it would seem either the code or the documentation needs a fixup as well.

Updated, thanks for a reminder.

src/common/primitive_serialization.cpp
densamoilov: The packed descriptor is is not supposed to be initialized by the user but only by the implementation so it's pretty much the same as the `any` format tag. Which means primitive parameters are expected to unambiguously define the packed descriptor and therefore there shouldn't be any problems with it. This is why it's not included in the hashing function.

dzarukin: Thanks for explanation, removed it from serialization.

densamoilov (approved):

mzhukova (approved):

PR #5219, opened 2026-05-29
graph: backend: dnnl: fix empty event propagation

TaoLv:
In some cases, graph backend executable may generate empty (the default) SYCL event which will be propagated to the next executable. This violates SYCL graph capture as empty event can be recognized as an external event. This patch changes the return value from `sycl::event` to `std::optional<sycl::event>`.  The caller will need to check the validness of the return event before putting it into the deps of the next executable.

### DIFF
--- a/src/graph/backend/dnnl/executables/base.hpp
+++ b/src/graph/backend/dnnl/executables/base.hpp
@@ -17,6 +17,7 @@
 #define GRAPH_BACKEND_DNNL_EXECUTABLES_BASE_HPP
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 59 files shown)


TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

ElaineBao (approved):

rongzha1 (approved):

TaoLv:
make test
disable benchdnn_all
enable benchdnn_graph

dzarukin:
Just for my education - instead of changing signatures, why not check if event is empty and pass only non-empty events as dependencies?

TaoLv:
> Just for my education - instead of changing signatures, why not check if event is empty and pass only non-empty events as dependencies?

I tried but did not find a way to check if an SYCL event is empty. There is no such member function in SYCL spec. And `if (!event)` seems not work either. I even failed to compare two default events.

PR #5224, opened 2026-05-29
[GPU] Fix build after GEMM update

Simonsays095:
After a few of my PRs merged yesterday, CI revealed some build errors:
- Since ngen::Product was given a constructor, it's no longer trivially-copyable. I need to cast to void* when copying into `gpu_product_t`.
- I inadvertently changed a `round_up` call to `align_up` which has a more restrictive signature that caused argument narrowing. Reverted that change.

### DIFF
--- a/src/gpu/intel/gemm/jit/dsl/hw.cpp
+++ b/src/gpu/intel/gemm/jit/dsl/hw.cpp
@@ -32,7 +32,8 @@ hw_t::hw_t(const ngen::Product &product, int eu_count, int max_wg_size,
 
 ngen::Product hw_t::product() const {
     ngen::Product product;
-    std::memcpy(&product, &product_, sizeof(product));
+    std::memcpy(static_cast<void *>(&product),
+            static_cast<const void *>(&product_), sizeof(product));
     return product;
 }
 
--- a/src/gpu/intel/gemm/jit/selector/kernel_evaluator.cpp
+++ b/src/gpu/intel/gemm/jit/selector/kernel_evaluator.cpp
@@ -517,7 +517,7 @@ DerivedEvaluateParams getDerivedParams(const kcatalog::Entry &e, const EvaluateP
         auto ldc = static_cast<int>(Tc * (isColMajor(cLayout) ? p.sizes.m : p.sizes.n) * e.driverInfo.cInterleaveChunk(Tc));
         auto &wgCountX = isColMajor(cLayout) ? dp.wgCountM : dp.wgCountN;
         auto &wgCountY = isColMajor(cLayout) ? dp.wgCountN : dp.wgCountM;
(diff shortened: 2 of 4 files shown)


Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

echeresh (approved):

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

src/gpu/intel/gemm/jit/dsl/hw.cpp
```diff
@@ -32,7 +32,8 @@ hw_t::hw_t(const ngen::Product &product, int eu_count, int max_wg_size,
 
 ngen::Product hw_t::product() const {
     ngen::Product product;
-    std::memcpy(&product, &product_, sizeof(product));
+    std::memcpy(static_cast<void *>(&product),
+            static_cast<const void *>(&product_), sizeof(product));
```
rjoursler, line 36: I don't think this is valid fix as it is likely UB, from [cppreference](https://en.cppreference.com/cpp/types/is_trivially_copyable)
> Notes
Objects of trivially-copyable types that are not potentially-overlapping subobjects are the only C++ objects that may be safely copied with std::memcpy or serialized to/from binary files with [std::ofstream::write()](https://en.cppreference.com/cpp/io/basic_ostream/write) / [std::ifstream::read()](https://en.cppreference.com/cpp/io/basic_istream/read).

Using a cast here is just suppressing the compiler warning, not resolving the issue.

echeresh: But both `ngen::Product` and `product_t` are trivially copyable, right? Is anything missing?

Update: just saw this in the PR description:

> Since ngen::Product was given a constructor, it's no longer trivially-copyable

I don't think this is a correct statement, the new constructor doesn't make it non trivially-copyable.

rjoursler: I see, yeah, the issue is that the GCC warning is more strict that necessary: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107361

Using a cast as a workaround is likely fine, but we should add a static_assertion that ngen::Product is trivially copyable so that we maintain some level of compiler diagnostics.

Simonsays095: Added the static assertion - I tried removing the constructors (the other way to clear the error) but in C++11 we can only do aggregate initialization if there are no default values, which opens us up for more issues with uninitialized values. Using memcpy + the static assert to guarantee trivial copyability should be enough safety for us for now.

Simonsays095:
make test
disable test_device_cpu
disable test_device_gpu

rjoursler (approved):

PR #5226, opened 2026-05-29
[GPU] Fix frequency reporting after benchdnn filtering updates

echeresh:
PR is a follow-up for the filtering logic in benchdnn. This is to restore `%freq%` field in the performance template. The frequency is calculated from the time and ticks so the ticks need to be processed during the filtering in the same way as the times.

### DIFF
--- a/tests/benchdnn/utils/timer.cpp
+++ b/tests/benchdnn/utils/timer.cpp
@@ -52,7 +52,7 @@ void timer_t::reset() {
     for (int i = 0; i < n_modes; ++i)
         ms_[i] = 0;
     ms_start_ = 0;
-    ms_vec_.clear();
+    samples_.clear();
 
     start();
 }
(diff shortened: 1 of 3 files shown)


atkassen (approved):

echeresh:
make test

tests/benchdnn/utils/timer.cpp
```diff
@@ -172,30 +177,31 @@ void timer_t::filter_collection() {
...
+        ms_[mode_t::sum] += s.ms;
+        ticks_[mode_t::sum] += s.ticks;
     }
     ms_[mode_t::avg] = ms_[mode_t::sum];
-    times_ = ms_vec_.size();
-    // Frequency is not supported for filtering, explicitly zeroing it to avoid
-    // potential misuse.
-    for (int i = 0; i < n_modes; i++)
-        ticks_[i] = 0;
+    ticks_[mode_t::avg] = ticks_[mode_t::sum];
```
h-sadia, line 203: Just for clarity, what is the purpose of assigning sum to avg mode?

h-sadia (approved):

tests/benchdnn/utils/timer.cpp
dzarukin: This is how average works - it uses sum then divides it by the `times_` at the moment of reporting.

Can we drop avg collection and just rely on sum? Yes, we can.

dzarukin (approved):

tests/benchdnn/utils/timer.cpp
echeresh: Since it's a preexisting convention - we can adjust in a future PR whenever someone touches this code next.

PR #5227, opened 2026-05-30
third_party: ngen: track s0 indirect descriptor dependency for sendg/sendgc

echeresh:
Found and reported while investigating a page fault with OpenVINO on NVL-P.

### DIFF
--- a/third_party/ngen/ngen_auto_swsb.hpp
+++ b/third_party/ngen/ngen_auto_swsb.hpp
@@ -479,6 +479,10 @@ DependencyRegion::DependencyRegion(HW hw_, int esize, RegData rr)
     checkWAW = false;
     rf = rr.getRegFile();
 
+    // Strip out the 0x100 ARFType::s marker so a directly-constructed scalar
+    // register matches the base decoded from an instruction.
+    if (rf == RegFileARF) base &= 0xFF;
+
     int hs = rr.getHS(), vs = rr.getVS();
     int nh = rr.getWidth();
     if (nh == 0) nh = 1;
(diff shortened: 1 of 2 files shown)


atkassen (approved):

third_party/ngen/ngen_auto_swsb.hpp
```diff
@@ -479,6 +479,10 @@ DependencyRegion::DependencyRegion(HW hw_, int esize, RegData rr)
     checkWAW = false;
     rf = rr.getRegFile();
 
+    // Strip out the 0x100 ARFType::s marker so a directly-constructed scalar
+    // register matches the base decoded from an instruction.
+    if (rf == RegFileARF) base &= 0xFF;
```
kealan-barbieri, line 484: should this be handled in `getBase` or do we generally want the ARF register offset present?

kealan-barbieri (approved):

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_deconv
enable benchdnn_pool
enable benchdnn_reorder
enable benchdnn_matmul
enable benchdnn_ip

third_party/ngen/ngen_auto_swsb.hpp
echeresh: Good question. `getBase()` for scalar registers by nGEN design has `0x100` mask to distinguish them from the stack pointer. It's a bit confusing, but I checked - they are encoded similarly in the ISA. That's why we have to do this kind of fixup to match the base extracted from an instruction.

Issue #4299, opened 2025-11-07
Support for long integers (aka int64, s64, long)?


blueberry:
# Summary
Many LLM models take tensors of 64-bit integers at input/output (GPT2, etc). If one uses them from ONNX Runtime or CUDA, both natively support long integers. OneDNN doesn't, even in the most basic way of creating a memory descriptor, or data reordering. 

# Problem statement
The users of oneDNN (usually higher-level libraries) typically rely on DNNL as a base to build tensors on CPUs. Since OneDNN doesn't suport 64-bit integers, such implementations can't ingest data coming form other sources that work with them, such as cuDNN (supports long integers) or ONNX Runtime (also support long tensors).

# Preferred solution
The solution doesn't even to cover most of oneDNN routines, since these would probably be complicated to implement. Here's what I think is not very complicated, and yet will increase the usefulness of oneDNN:

1) Support 64-bits integers as data types in dnnl_memory_desc
2) Support reorder primitive, and other simple-ish primitives, even with sub-optimal performance

Nobody is going to use 64-bit integers as their first choice, but sometimes there's data, and at leas enable oneDNN to recognize this data, and easily convert it into a more optimal data-type (int32, float, or whatever).

vpirogov:
@blueberry, thank you for the suggestion. Couple questions:
* Could you please share a pointer to cuDNN documentation that covers `int64`/`long` data types support?
* Conversion from `int64` to `int32` or `fp32` is fairly straightforward operation in C. What advantages you expect to get from using reorder primitive instead?

blueberry:
1. There's no documentation because there's no current support. I'm suggesting barebones support.
2. The main benefit is *consistent interface*. Supose there's general code that works with a oneDNN tensor, and the code needs to support various data-types that depend on a parameter (say, to accept an onednn tensor and create a cudnn tensor of the same data-type and move the data). Such code would work well for almost all types, but then for int64, it would have to create a completely new branch with custom logic, because there can't be an onednnl tensor of long integers, sice I can't create a dnnl_memory_desc with long integer type.
2a. For the reorder operation: yes, it is straightforward in any language, but then the client of the code has to write custom logic for that case, check what is the type, and call different operation for the reorder, possibly having to implement non-trivial logic if strides are not straightforward, instead of calling the reorder primitive in a general way, for all meaningful data types (64-bit integer is a staple in many languages, such as Java, C#, etc.).
[...]

PR #5228, opened 2026-05-30
rfcs: s64 (int64) data type support

elprofesoriqo:
## Description

RFC proposing minimal s64 (int64) data type support in oneDNN: enable s64 in
memory descriptors and the CPU `reorder` primitive. Compute primitives
(convolution, matmul, etc.) are out of scope.

The `dnnl_s64` token and its internal type plumbing were already added in
commit 4cab67c4f0 ("api: add s64 data type", Sep 2025), but no primitive
accepts the type today. This RFC documents the use cases (ONNX Runtime
interop, OpenVINO ONNX EP conformance, consistent API for language bindings)
and proposes completing the minimal support via `reorder`.

Addresses #4299.

Rendered: https://github.com/elprofesoriqo/oneDNN/blob/rfc/s64-datatype-support/rfcs/20260529-s64-datatype-support/README.md

### DIFF
--- a/rfcs/20260529-s64-datatype-support/README.md
+++ b/rfcs/20260529-s64-datatype-support/README.md
@@ -1,0 +1,203 @@
+# s64 (int64) Data Type Support in oneDNN
+
+## Introduction
+
+The main driver for adding `int64` to oneDNN is interoperability with ONNX.
+ONNX Runtime represents many tensors as `int64`, including token ids,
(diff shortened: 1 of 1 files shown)


dzarukin (commented):
This RFC doesn't explain a motivation and purpose of such reorders. I don't see why these reorders should exist in oneDNN and can't exist on ONNX side to do what they need to do.

PR #5229, opened 2026-05-30
api: restrict int4 odd strides for second-to-innermost dim

dzarukin:
The restriction is moved from matmul to general md creation with tags and strides since these functions are used with `format_kind::any` initialization and should be verified there. Will also affect reorder creation.

The PR extends strides check for tags creation which required additional modifications since existing check didn't account blocked tags properly.

At last, there are some overflows for large sizes in CPU heuristics functions.


dzarukin:
make test

dzarukin:
make test

dzarukin:
make test

dzarukin:
make test
set test_scope=NIGHTLY

dzarukin:
make test
set test_scope=NIGHTLY

mzhukova (approved):

densamoilov (approved):

PR #5234, opened 2026-06-01
common: rnn: validate prop_kind in fwd/bwd desc init

### LINKED ISSUE #4306: dnnl_lstm_backward_primitive_desc_create() can be called with parameter prop_kind=dnnl_forward_training
# Problem statement
dnnl_lstm_backward_primitive_desc_create() can be called with parameter prop_kind=dnnl_forward_training, which is IMO unexpected as for backward primitive there should be used **dnnl_backward**. I have used by mistake dnnl_forward_training and primitive did not worked and it takes for me a "long time" to found the problem ...

# Preferred solution
Check that dnnl_lstm_backward_primitive_desc_create() will not accept dnnl_forward, dnnl_forward_training.

May be there is a reasom for which also backward primitive accepts dnnl_forward* propagation kind I am not aware of ...

mikita12:
This PR adds missing `prop_kind` validation to `rnn_common_fwd_desc_init`
and `rnn_common_bwd_desc_init`.

Previously, all 12 public RNN primitive descriptor creation functions
accepted any `prop_kind` value without validation. Passing
`dnnl_forward_training` to a backward function (or `dnnl_backward` to
a forward function) silently succeeded instead of returning
`invalid_arguments`, making bugs like the one in #4306 very hard to
diagnose.

The fix adds a `VCONDCHECK_RNN` check as the first validation in each
shared init helper, following the same pattern used by
`batch_normalization`, `eltwise`, and `group_normalization`. One check
in each helper covers all 6 forward and all 6 backward public API
functions.

Fixes #4306

## General
- [x] Do all unit and benchdnn tests pass locally for each commit?

## Bug fixes
- [x] Have you included information on how to reproduce the issue?


tests/gtests/api/test_rnn_api.cpp
```diff
@@ -0,0 +1,338 @@
+/*******************************************************************************
```
dzarukin: There's a dedicated RNN test (test_rnn_forward), if you want to add some negative testing, feel free to update that file instead. I think we can keep skipping backward RNN validation in gtests.

dzarukin (commented):
Thanks for adding checks!

tests/gtests/api/test_rnn_api.cpp
mikita12: Done! Removed test_rnn_api.cpp and moved the negative prop_kind tests 
to test_rnn_forward.cpp instead.

dzarukin (approved):

dzarukin:
make test
disable benchdnn_all
enable benchdnn_rnn

mikita12:
@dzarukin 

Could you help me understand what failed in the Intel internal CI? 
I don't have access to the logs.

vpirogov:
> Could you help me understand what failed in the Intel internal CI? I don't have access to the logs.

Completely normal phenomenon. Nothing related to your changes.

rjoursler (approved):

vpirogov (approved):

PR #5086, opened 2026-04-28
clarifying on inconsistencies in quantization guide

mzhukova:
This PR is to clear some inconsistencies on supported quantization as well as an attempt to clarify on group sizes and mask used for setting scales and zero-points (incl. api notes, gtests, quantization guide); Specifically, there were questions on 3D and grouped gemm, therefore adding extra snippets into the doc.

Should address concerns in [MFDNN-15068](https://jira.devtools.intel.com/browse/MFDNN-15068).

// easier to review by commits

PR #5238, opened 2026-06-02
cpu: x64: matmul: Enable full tensor scales and zero points for 3D/4D matmul

inteldimitrius:
# Description
Implements [MFDNN-15067](https://jira.devtools.intel.com/browse/MFDNN-15067).

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul.cpp
+++ b/src/cpu/x64/matmul/brgemm_matmul.cpp
@@ -227,26 +227,16 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             ok = ok && one_of(asc.get_data_type(DNNL_ARG_WEIGHTS), undef, f32);
             ok = ok && one_of(asc.get_data_type(DNNL_ARG_DST), undef, f32);
         }
-        // This impl doesn't support scales over any batch dimensions.
-        if (!asc.has_default_values(DNNL_ARG_WEIGHTS)) {
-            const auto mask = asc.get_mask(DNNL_ARG_WEIGHTS);
-            const int kn_mask = wei_qmask_N() + wei_qmask_K();
-            const bool scale_over_batch = (mask & kn_mask) != mask;
-            if (scale_over_batch && batch() > 1) ok = false;
-        }
         // Implementation has limited support w.r.t. scales groups.
         if (!asc.has_default_values(DNNL_ARG_WEIGHTS)) {
             if (!asc.get(DNNL_ARG_WEIGHTS).has_default_groups()) {
                 // Only grouping over K is supported.
                 ok = ok && asc.get_group(DNNL_ARG_WEIGHTS, 1) == 1;
-                // Only 'per_ocic' mask is supported, but not 'per_tensor' in
-                // benchdnn terms. In numbers, it's '12' is supported while for
-                // 4D '15' is required.
+                // masks 3 per-(k,n) and 7 per-(b,k,n) are supported
                 const int mask = asc.get_mask(DNNL_ARG_WEIGHTS);
-                const int ndims = weights_md_.ndims;
-                const int last_dim = (1 << (ndims - 1));
-                const int prelast_dim = (1 << (ndims - 2));
-                const bool mask_ok = (mask & ~(last_dim | prelast_dim)) == 0;
+                const int kn_mask = wei_qmask_N() + wei_qmask_K();
+                const int batch_mask = full_tensor_mask() & ~kn_mask;
+                const bool mask_ok = (mask & ~(kn_mask | batch_mask)) == 0;
                 ok = ok && mask_ok;
             }
         }
@@ -267,23 +257,20 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             const auto mask = zp.get_mask(DNNL_ARG_SRC);
             if (allow_grouped_src_zp) {
                 const int mk_mask = src_qmask_M() + src_qmask_K();
-                const bool zp_over_batch = (mask & mk_mask) != mask;
-                ok = ok && (mask & ~mk_mask) == 0;
-                ok = ok && !(zp_over_batch && batch() > 1);
+                const int batch_mask = full_tensor_mask() & ~mk_mask;
+                ok = ok && (mask & ~(mk_mask | batch_mask)) == 0;
             } else {
                 ok = ok && mask == 0;
             }
         }
         // Weights zero-points support a common value only, unless multiple
-        // values are allowed; then only grouping over K/N (not batch) is
-        // supported.
+        // values are allowed; then masks 3 and 7 are supported.
         if (!zp.has_default_values(DNNL_ARG_WEIGHTS)) {
             const auto mask = zp.get_mask(DNNL_ARG_WEIGHTS);
             if (allow_multiple_wei_zp) {
                 const auto kn_mask = wei_qmask_N() + wei_qmask_K();
-                const bool zp_over_batch = (mask & kn_mask) != mask;
-                ok = ok && (mask & ~kn_mask) == 0;
-                ok = ok && !(zp_over_batch && batch() > 1);
+                const int batch_mask = full_tensor_mask() & ~kn_mask;
+                ok = ok && (mask & ~(kn_mask | batch_mask)) == 0;
             } else {
                 ok = ok && mask == 0;
             }
@@ -870,8 +857,8 @@ void brgemm_matmul_t<isa>::compute_kernel(
         // Symmetric src/wei zp + 128-shift compensation tile for this
         // brgemm K-block call
         if (bgmmc.with_per_mn_compensation) {
-            fill_per_mn_compensation(brgmm_ctx, ithr, m_blk_idx, n_blk_idx,
-                    A_data_batch_ptr, B_data_batch_ptr,
+            fill_per_mn_compensation(brgmm_ctx, ithr, b_idx, m_blk_idx,
+                    n_blk_idx, A_data_batch_ptr, B_data_batch_ptr,
                     brgmm_ctx.get_M_kernel_size(m_blk_idx),
                     brgmm_ctx.get_N_kernel_size(n_blk_idx), k_blk_idx, is_tail);
         }
@@ -890,13 +877,13 @@ void brgemm_matmul_t<isa>::compute_kernel(
                 return;
             }
             const void *src_scales = bgmmc.is_src_scale_per_k
-                    ? brgmm_ctx.get_src_scales_ptr(m, k)
+                    ? brgmm_ctx.get_src_scales_ptr(b_idx, k, m)
                     : (bgmmc.is_wei_scale_per_k && bgmmc.with_src_scales
-                                      ? brgmm_ctx.get_src_scales_ptr()
+                                      ? brgmm_ctx.get_src_scales_ptr(b_idx)
                                       : nullptr);
             const void *wei_scales = bgmmc.is_wei_scale_per_k
                             && !bgmmc.apply_scales_in_buffer_b
-                    ? brgmm_ctx.get_wei_scales_ptr(n, k)
+                    ? brgmm_ctx.get_wei_scales_ptr(b_idx, k, n)
                     : nullptr;
             void *scratch = is_brg_amx(brg_idx) ? (void *)wsp_tile : nullptr;
             const brgemm_post_ops_data_t post_ops_data {/*bias=*/nullptr,
@@ -918,12 +905,12 @@ void brgemm_matmul_t<isa>::compute_kernel(
         // Final K-block: apply scales, compensation and post-ops. Per-K scales
         // take precedence over the common ones.
         const void *src_scales = bgmmc.is_src_scale_per_k
-                ? brgmm_ctx.get_src_scales_ptr(m, k)
-                : brgmm_ctx.get_src_scales_ptr();
+                ? brgmm_ctx.get_src_scales_ptr(b_idx, k, m)
+                : brgmm_ctx.get_src_scales_ptr(b_idx);
         const void *wei_scales
                 = bgmmc.is_wei_scale_per_k && !bgmmc.apply_scales_in_buffer_b
-                ? brgmm_ctx.get_wei_scales_ptr(n, k)
-                : brgmm_ctx.get_wei_scales_ptr(n, /*k=*/0);
+                ? brgmm_ctx.get_wei_scales_ptr(b_idx, k, n)
+                : brgmm_ctx.get_wei_scales_ptr(b_idx, /*k=*/0, n);
         const void *dst_scales = brgmm_ctx.get_dst_scales_inv_ptr(ithr);
         void *scratch = is_brg_amx(brg_idx)
                 ? (void *)wsp_tile
@@ -1001,8 +988,8 @@ void brgemm_matmul_t<isa>::compute_kernel(
 
 template <cpu_isa_t isa>
 void brgemm_matmul_t<isa>::fill_per_mn_compensation(
-        const brg_matmul_exec_ctx_t &brgmm_ctx, int ithr, int m_blk_idx,
-        int n_blk_idx, const char *A_data_batch_ptr,
+        const brg_matmul_exec_ctx_t &brgmm_ctx, int ithr, int b_idx,
+        int m_blk_idx, int n_blk_idx, const char *A_data_batch_ptr,
         const char *B_data_batch_ptr, int m_kernel_size, int n_kernel_size,
         int k_blk_idx, bool is_tail) const {
     const auto &bgmmc = pd()->get_brgemm_matmul_conf();
@@ -1023,8 +1010,8 @@ void brgemm_matmul_t<isa>::fill_per_mn_compensation(
     // and returns the nibble index so the kernel picks the right nibble.
     int src_zp_nibble_idx = 0;
     const void *src_zp_ptr
-            = brgmm_ctx.get_src_zp_ptr(m, k_start, &src_zp_nibble_idx);
-    const void *wei_zp_ptr = brgmm_ctx.get_wei_zp_ptr(n, k_start);
+            = brgmm_ctx.get_src_zp_ptr(b_idx, k_start, m, &src_zp_nibble_idx);
+    const void *wei_zp_ptr = brgmm_ctx.get_wei_zp_ptr(b_idx, k_start, n);
 
     const bool wei_is_subbyte
             = utils::one_of(bgmmc.orig_wei_dt, data_type::s4, data_type::u4);
@@ -1424,8 +1411,8 @@ void brgemm_matmul_t<isa>::maybe_reduce_partial_results_and_apply_postops(
                             static_cast<const void *>(zp_comp_a),
                             static_cast<const void *>(zp_comp_b),
                             brgmm_ctx.get_zp_c_ptr(), skip_accumulation, 1,
-                            false, false, brgmm_ctx.get_src_scales_ptr(),
-                            brgmm_ctx.get_wei_scales_ptr(n, /*k=*/0),
+                            false, false, brgmm_ctx.get_src_scales_ptr(b),
+                            brgmm_ctx.get_wei_scales_ptr(b, 0, n),
                             brgmm_ctx.get_dst_scales_inv_ptr(ithr)};
 
                     brgemm_kernel_execute_postops(brg_kernel, 0, nullptr,
@@ -1563,9 +1550,9 @@ void brgemm_matmul_t<isa>::copy_b_chunk_in_buffer(
         ctx.current_K_start = k;
         ctx.current_K_iters = k_iters;
         ctx.current_K_pad = brgmm_ctx.get_current_K_pad(k_iters);
-        ctx.src_scales_ptr = brgmm_ctx.get_src_scales_ptr();
-        ctx.wei_scales_ptr = brgmm_ctx.get_wei_scales_ptr(n, k);
-        ctx.zp_b_value_ptr = brgmm_ctx.get_wei_zp_ptr(n, k);
+        ctx.src_scales_ptr = brgmm_ctx.get_src_scales_ptr(b_idx);
+        ctx.wei_scales_ptr = brgmm_ctx.get_wei_scales_ptr(b_idx, k, n);
+        ctx.zp_b_value_ptr = brgmm_ctx.get_wei_zp_ptr(b_idx, k, n);
         if (bgmmc.blocked_B && !bgmmc.is_f16_with_int_wei
                 && isa == avx512_core_fp16) {
             cvt_float16_to_float((float *)ctx.tr_src, (float16_t *)ctx.src,
@@ -2365,9 +2352,19 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
                 + n_blk_local * bgmmc_.s8s8_comp_n_str;
     }
 
-    const void *get_src_scales_ptr(dim_t m = 0, dim_t k = 0) const {
-        if (!bgmmc_.is_src_scale_per_k && !bgmmc_.is_src_scale_per_m)
+    const void *get_src_scales_ptr(
+            dim_t b_idx = 0, dim_t k = 0, dim_t m = 0) const {
+        if (!bgmmc_.is_src_scale_per_k && !bgmmc_.is_src_scale_per_m) {
+            // Pure per-batch addressing (e.g. 3D/4D per_tensor with batch
+            // bits in the scales mask).
+            if (bgmmc_.src_scales_batch_stride > 0) {
+                const dim_t b = get_bb_idx(b_idx, bgmmc_.bcast_A_desc);
+                const auto offset = b * bgmmc_.src_scales_batch_stride
+                        * bgmmc_.src_scales_dt_sz;
+                return (const char *)src_scales_ + offset;
+            }
             return src_scales_;
+        }
 
         const auto k_group_sz = bgmmc_.is_src_scale_per_k
                 ? bgmmc_.src_scales_k_gsize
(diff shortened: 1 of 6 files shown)


src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -2126,7 +2133,7 @@ struct brgemm_matmul_t<isa>::brg_matmul_exec_ctx_t {
 
     // Returns a pointer to the weights scales for the correspondent block based
     // on @p n and @p k.
-    const void *get_wei_scales_ptr(dim_t n, dim_t k = 0) const {
+    const void *get_wei_scales_ptr(dim_t n, dim_t k = 0, int b_idx = 0) const {
```
dzarukin: Minor (feel free to ignore): I'd put them in logical order of appearance: `b_idx, k, n` as they appear in weights desc stripping default values (this interface doesn't seem to be called 100 times to save on some modifications).

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
```diff
@@ -37,3 +37,36 @@
...
+# Covers all four B-broadcast scenarios from the JIRA ticket.
+--reset
+--dt=f32:s4:f32
+--stag=abx --wtag=abx --dtag=abx
+--attr-fpmath=strict:true
+--attr-scales=wei:7:f32:64x1
+4x208x512:4x512x748
+1x208x512:4x512x748
+4x208x512:1x512x748
+1x208x512:1x512x748
```
dzarukin: Let's use `64` instead of `512` for some cases here to make sure a single group is processed correctly as well.
You might want to test wtag=acb and wtag=any to test other copy kernels.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1658,6 +1640,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
     bgmmc.is_runtime_K = is_runtime_value(bgmmc.K);
 
+    if (bgmmc.with_wei_scales && bgmmc.batch > 1) {
+        const int kn_mask = (1 << (bgmmc.ndims - 1)) | (1 << (bgmmc.ndims - 2));
+        const bool has_batch_bits = (wei_scales.get_mask() & ~kn_mask) != 0;
+        if (has_batch_bits) {
+            const dim_t num_k_groups = bgmmc.is_wei_scale_per_k
+                    ? utils::div_up(bgmmc.K, bgmmc.wei_scales_k_gsize)
+                    : 1;
+            bgmmc.wei_scales_batch_stride = num_k_groups * bgmmc.N;
```
dzarukin: There's `quant_entry_t::get_md(...)` interface that returns scales md based on input md (in this case weights), it takes groups into account already and should provide this value through `memory_desc_wrapper(scales_md).strides()[ndims - 2]`. If it works and you find it less error-prone, I suggest to stick to this verions.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -263,6 +263,7 @@ struct brgemm_matmul_conf_t {
     bool is_wei_scale_common = false;
     dim_t wei_scales_k_gsize = 0;
     data_type_t wei_scales_dt = data_type::undef;
+    dim_t wei_scales_batch_stride = 0;
```
dzarukin: You might want to document this is for innermost batch dim regardless of number of batch dimensions just to be clear on the usage.

src/cpu/x64/matmul/brgemm_matmul.cpp
```diff
@@ -251,8 +255,11 @@ status_t brgemm_matmul_t<isa>::pd_t::init(engine_t *engine) {
             if (allow_multiple_wei_zp) {
                 const auto kn_mask = wei_qmask_N() + wei_qmask_K();
                 const bool zp_over_batch = (mask & kn_mask) != mask;
-                const bool mask_ok = (mask & ~kn_mask) == 0;
-                return !(zp_over_batch && batch() > 1) && mask_ok;
+                const bool is_grouped
+                        = !zp.get(DNNL_ARG_WEIGHTS).has_default_groups();
+                const bool mask_ok = is_grouped || (mask & ~kn_mask) == 0;
```
georgen117: **possible BUG**
When `is_grouped` is `true`, `mask_ok` is unconditionally `true` regardless of the mask value. This means a grouped ZP with a batch-only mask passes validation.

At runtime, `is_wei_zp_per_k` and `is_wei_zp_per_n` would both be false for such a mask, so `get_wei_sp_ptr` would apply a batch stride but index into a 1-element-per-batch buffer as if it were N elements wide, silently producing incorrect zp values.

The grouped-scales path in `check_attr_scales` (line ~234) already handles this correctly by requrieing both K and N bits.

The same guard is needed here.

Suggested fix:
```cpp
const bool mask_ok = is_grouped
             ? ((mask & (wei_qmask_K() + wei_qmask_N())) == (wei_qmask_K() + wei_qmask_N()))
             : (mask * ~kn_mask) == 0;
```

Reviewer note: I am pretty sure this is an issue but please check my logic there is still a lot I don't know about the kernel.

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
```diff
@@ -37,3 +37,36 @@
...
+# 3D / 4D per-batch wei zero-points combined with grouped wei scales.
+--reset
+--dt=u8:s8:s32
+--stag=abx --wtag=abx --dtag=abx
+--attr-scales=wei:6:f32:64x1
+--attr-zero-points=wei:per_oc,wei:per_oc:s4
+4x208x512:4x512x748
+1x208x512:4x512x748
+4x208x512:1x512x748
+1x208x512:1x512x748
```
georgen117: The comment indicates this is per-batch however I don't think the ZP or the scales mask include the batch dimention.

`--attr-scales=wir:6:f32:64x1` - mask `6` is `0b110`, which for a 3D weight [batch, K, N] covers bit 1 (K) and 2 (N) only. No batch bit.
`--attr-zero-points=wei:per_oc` - per output channel (N only). No batch bit, and non-grouped

This is a valid and useful test but not the per-batch ZP path.

Suggestion:
1. change comment to `# 3D per-N wei zero-points combined with grouped K+N wei scales.`
2. add new block that actually exercises the per-batch ZP path.
```
# 3D per-batch grouped wei zero-points (mask covers batch+K+N) with grouped wei scales.
--reset
--dt=u8:s8:s32
--stag=abx --wtag=abx --dtag=abx
--attr-scales=wei:7:f32:64x1
--attr-zero-points=wei:7:s8:64x1,wei:7:s4:64x1
4x208x512:4x512x748
1x208x512:4x512x748
4x208x512:1x512x748
1x208x512:1x512x748
```

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Updated!

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
inteldimitrius: Changed couple cases to 64 and acb,any wei-tags.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
inteldimitrius: Reworked using, md wrapper, however it should use `ndims  - 3` index for batch stride. Am I right?

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
inteldimitrius: Added comment for this string as well as for `wei_zp_batch_stride`.

src/cpu/x64/matmul/brgemm_matmul.cpp
inteldimitrius: Updated, thanks!

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
inteldimitrius: Updated comments! Thanks!

inteldimitrius:
make test

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
dzarukin: Yep, should be `-3`.

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -279,6 +282,9 @@ struct brgemm_matmul_conf_t {
     bool is_wei_zp_per_n = false;
     bool is_wei_zp_common = false;
     data_type_t wei_zp_dt = data_type::undef;
+    // This is innermost batch stride for wei scales,
```
dzarukin:
```suggestion
    // This is innermost batch stride for wei zero-points,
```

dzarukin (approved):

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1667,6 +1667,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     bgmmc.is_runtime_N = is_runtime_value(bgmmc.N);
     bgmmc.is_runtime_K = is_runtime_value(bgmmc.K);
 
+    if (bgmmc.with_wei_scales && bgmmc.batch > 1) {
+        const int kn_mask = (1 << (bgmmc.ndims - 1)) | (1 << (bgmmc.ndims - 2));
+        const bool has_batch_bits = (wei_scales.get_mask() & ~kn_mask) != 0;
+        if (has_batch_bits) {
+            memory_desc_t scales_md;
+            wei_scales.get_md(scales_md, weights_md);
+            bgmmc.wei_scales_batch_stride = memory_desc_wrapper(&scales_md)
```
densamoilov: `bgmmc.batch > 1` condition doesn't protect us from `ndims - 3` being 1, which would make accessing the stride undefined behavior. `bgmmc.batch` is the product of all batch dimensions.

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1667,6 +1667,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
...
+    }
+
+    if (has_wei_zp && bgmmc.batch > 1) {
+        const int kn_mask = (1 << (bgmmc.ndims - 1)) | (1 << (bgmmc.ndims - 2));
+        const bool has_batch_bits = (wei_zp.get_mask() & ~kn_mask) != 0;
+        if (has_batch_bits) {
+            memory_desc_t zp_md;
+            wei_zp.get_md(zp_md, weights_md);
+            bgmmc.wei_zp_batch_stride
+                    = memory_desc_wrapper(&zp_md).strides()[bgmmc.ndims - 3];
```
densamoilov: same as above

src/cpu/x64/matmul/brgemm_matmul_utils.cpp
```diff
@@ -1667,6 +1667,28 @@ status_t init_brgemm_matmul_conf(cpu_isa_t isa, brgemm_matmul_conf_t &bgmmc,
     bgmmc.is_runtime_N = is_runtime_value(bgmmc.N);
     bgmmc.is_runtime_K = is_runtime_value(bgmmc.K);
 
+    if (bgmmc.with_wei_scales && bgmmc.batch > 1) {
+        const int kn_mask = (1 << (bgmmc.ndims - 1)) | (1 << (bgmmc.ndims - 2));
```
densamoilov: nit: looks like this line is duplicated in both branches

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
```diff
@@ -37,3 +37,35 @@
...
+1x2x208x512:2x2x512x748
+2x2x208x64:1x2x64x748
+1x1x208x64:1x1x64x748
+
+# 3D / 4D per-batch wei zero-points combined with grouped wei scales.
+--reset
+--dt=u8:s8:s32
+--stag=abx --wtag=abx,acb,any --dtag=abx
+--attr-scales=wei:per_tensor:f32:64x1
+--attr-zero-points=wei:per_oc,wei:per_oc:s4
```
densamoilov: The title says `3D / 4D per-batch wei zero-points` but the policy for ZPs here is `per_oc`. Is it a mistake or am I missing something?

densamoilov: Should it be something like `--dt=bf16:s8:f32 --attr-zero-points=wei:per_tensor:s4:64x1 --attr-fpmath=bf16:true 4x208x512:4x512x748`?

I see an accuracy issue for this case:

```
[ 736][DST][0:0:736] exp_f32:         -45 exp:         -45 got:          10 diff:      55 rdiff: 1.22222
[ 737][DST][0:0:737] exp_f32:         216 exp:         216 got:         157 diff:      59 rdiff:0.273148
[ 738][DST][0:0:738] exp_f32:         -17 exp:         -17 got:          91 diff:     108 rdiff: 6.35294
[ 739][DST][0:0:739] exp_f32:           5 exp:           5 got:          79 diff:      74 rdiff:    14.8
[ 740][DST][0:0:740] exp_f32:        -243 exp:        -243 got:         -46 diff:     197 rdiff:  0.8107
[ 741][DST][0:0:741] exp_f32:         183 exp:         183 got:          80 diff:     103 rdiff:0.562842
[ 742][DST][0:0:742] exp_f32:          11 exp:          11 got:         -17 diff:      28 rdiff: 2.54545
[ 743][DST][0:0:743] exp_f32:         208 exp:         208 got:         379 diff:     171 rdiff:0.822115
[ 744][DST][0:0:744] exp_f32:        -303 exp:        -303 got:         -47 diff:     256 rdiff:0.844885
[ 745][DST][0:0:745] exp_f32:         164 exp:         164 got:         319 diff:     155 rdiff:0.945122
```

inteldimitrius:
make test

src/common/primitive_attr_quant.hpp
```diff
@@ -130,8 +130,14 @@ struct quant_entry_t : public c_compatible {
...
                 /* fill_with_ones = */ true);
         if (!has_default_groups()) {
-            quant_dims[ndims - 2] /= get_group(0);
-            quant_dims[ndims - 1] /= get_group(1);
+            // Note: nstl::max guards are needed because get_group()
+            // returns 0 for out-of-bound access and MSVC raises C4723
+            // warning (potential divide by 0) treated as error.
+            quant_dims[ndims - 2] /= nstl::max(get_group(0), (dim_t)1);
+            quant_dims[ndims - 1] /= nstl::max(get_group(1), (dim_t)1);
+            if (group_ndims_ > 2 && ndims >= 3) {
```
mzhukova: as of now we only allow (by doc and gtests) group size of 2 or 0, see https://github.com/uxlfoundation/oneDNN/pull/5086
if this change is indeed required, we need a rework for doc/interface tests

mzhukova: (also wondering whether this is indeed required and what quantization mechanism is expressed via group size 3..?)

vpirogov: In the discussion in with OpenVINO we agreed that 3D groups are not required. Desired quantization behavior should be expressed by a mask + 2D group. @inteldimitrius, see [3D Batched Matmul](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_quantization.html#example-3d-batched-matmul-weights) example in the documentation.

dzarukin (changes requested):
PR changed significantly since the approval was provided.

src/common/primitive_attr_quant.hpp
inteldimitrius: Hi @vpirogov @mzhukova, thanks for clarifying. 
You right, these changes accidentally leaked from old branch. Fixed and pushed to the remote branch.
The only question is remaining: 
`per_ocic` mask is still present and generates mask `6` for 3D.
`per_tensor` mask generates mask `7` for 3D. 
Currently both are supported. Is this correct? Didn't find anything in doc-page related to `per_ocic` vs. `per_tensor` difference. 
Should I keep `per_ocic`-version mask setting?

tests/benchdnn/inputs/matmul/harness_matmul_3d_bcast
inteldimitrius: Fixed and added in separate commit. 
The fix not directly related to 3D/4D mask. Root cause: uninitialized mask for bf6 with int4 ZP. 
However since the case is 3D a have added regression test to the same place `tests/benchdnn/inputs/matmul/harness_matmul_decompression`. 
Thanks!

inteldimitrius:
make test
disable benchdnn_all
enable benchdnn_matmul

src/common/primitive_attr_quant.hpp
densamoilov: > Currently both are supported.

How do they work?

Should `per_ocic` quantization for a 3D matmul produce quantization parameters with shape 1 × K × N, where the leading dimension of 1 indicates broadcasting across that axis? To me it looks very different from `per_tensor` where each batch has its own KxN quantization parameters.

mzhukova: mask `6` is basically ~we have the same values and recipe (e.g., K-groups) for each item in batch VS `7` is ~we have same recipe but different values for each in batch, right?
I don't know what exactly we're trying to achieve with this PR (`7` or `6`, hence the question earlier), could you clarify maybe with whoever is requested this extension? 

I think names `per_tensor` / `per_ocic` might be a little off, I would focus more on integer values of the masks, since it defines what axis do what..

vpirogov: The policy names are a deprecated benchdnn construct, stick to the numerical masks instead. 

For me looking at relationship between the mask and quantization parameters tensor shape is the most helpful. On top of that groups serve as a modifier for quantization parameter shapes across the last two logical dimensions. You can find examples in discussion around [MFDNN-14588](https://jira.devtools.intel.com/browse/MFDNN-14588).

In short, mask `6` means that every tensor in the batch is quantized the same, mask `7` means that every tensor in the batch has it's unique quantization parameters. Do we need both? Yes, and `6` is already supported if I read Jira description right.

densamoilov: @mzhukova,

> I don't know what exactly we're trying to achieve with this PR (7 or 6, hence the question earlier), could you clarify maybe with whoever is requested this extension?

From this [MFDNN-15067](https://jira.devtools.intel.com/browse/MFDNN-15067)

> per_tensor configurations for both scales and zero-points to be supported.

Given that the deprecated policy `per_tensor` lowers to a mask that has all bits set we get 7. So yes, as Vadim mentioned we should support both and supporting mask 7 is the request.
I couldn't find the case for mask 6 in the [documentation ](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_quantization.html#example-3d-batched-matmul-weights). We seem to only have the case with mask 7 now. Should we add the mask 6 case too?

> For me looking at relationship between the mask and quantization parameters tensor shape is the most helpful. On top of that groups serve as a modifier for quantization parameter shapes across the last two logical dimensions. 

I think this is the perfect way to look at.

mzhukova: Documentation wasn't supposed to mention all possible masks/groups configurations, it gives [general guideline](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_quantization.html#mask-and-groups) on how to interpret and build the mask; as well as some [common examples](https://uxlfoundation.github.io/oneDNN/dev_guide_attributes_quantization.html#common-masks-and-groups-examples); but if `6` is indeed a tricky moment, I could extend it :)

densamoilov: I agree we shouldn't document every mask combination.
My concern is that the difference between 6 and 7 is not only which dimensions have mask bits set. They represent different quantization-parameter broadcasting semantics. I don't think this concept is currently described.

inteldimitrius: Hi guys,
Thanks for sharing info and links.
Let me share my points as a summary:
- Terms `per_ocic`, `per_tensor` are not used on lib level (test inputs still use these terms everywhere)
- > How do they work? 

I can not distinguish these cases and they both go through the same gate with batched idx:  `b_idx x K_group x N`.
So I according to the discussion I need ignore `b_idx` when it's not set  <=> mask=6.
In that sense logic looks clear for me: 7 => b_idx active, 6 => b_idx is not active. 
Am I right?

inteldimitrius: And also I'm wondering if the mask=6 case even expected to support? Going though discussions inside jira-tickets didn't find any explicit mention of it. Looks like OV wants pure [B,K,N] case and mask=7 only.

densamoilov: 3D: mask 6 => quantization parameter shape: 1xKxN => 1 means we broadcast KxN across the batch dimension
3D: mask 7 => quantization parameter shape: BxKxN => each B element has its own quantization parameter KxN

This approach scales up to ND.

> In that sense logic looks clear for me: 7 => b_idx active, 6 => b_idx is not active.

What you're describing here is a way to identify whether the quantization parameter broadcast is needed. Basically the question is, should we use stride 0 for the batch dimensions of the quantization parameter (e.g. 1xKxN) or non-zero (e.g. BxKxN). If the rule `7 => b_idx active, 6 => b_idx is not active` implies that then I think you should be good. Just make sure it works consistently for 4D cases too.

densamoilov: On the other hand, it's not clear to me why it'd be useful to apply the same quantization parameter to each element in the batch. So maybe we just need to support the non-broadcast case and be done with it for now.

inteldimitrius:
make test
disable benchdnn_all
enable benchdnn_matmul

inteldimitrius:
Hi @densamoilov @dzarukin @mzhukova @vpirogov,
I have rebased and updated everything according the discussion in PR.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -371,3 +371,30 @@
 --attr-zero-points=,wei:common:2,wei:per_oc:s4,wei:per_ocic:s4:192x1
 --attr-fpmath=strict:true
 12x4x576:12x576x192
+
+## Batched (3D) per-tensor scales/zero-points
+--reset
+--dt=bf16:s8:bf16
+--wtag=any,abc
```
dzarukin:
```suggestion
--wtag=any,abc,acb
```

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -371,3 +371,30 @@
...
+--attr-scales=wei:per_tensor:bf16:128x1
+--attr-zero-points=,wei:per_tensor:s8:128x1
+--attr-fpmath=bf16:true
+6x5x256:6x256x64
+9x8x384:9x384x64
+
+## Batched (4D) per-tensor scales/zero-points
+--reset
+--dt=bf16:s8:bf16
+--wtag=any,abcd
```
dzarukin: Ditto.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
```diff
@@ -371,3 +371,30 @@
...
+--attr-scales=wei:per_tensor:bf16:128x1
+--attr-zero-points=,wei:per_tensor:s8:128x1
+--attr-fpmath=bf16:true
+2x3x5x256:2x3x256x64
+2x4x6x384:2x4x384x64
+
+## Regression test(bf16 + int4 ZP + 3D shape):
+--reset
+ --dt=bf16:s8:f32
+ --attr-zero-points=wei:per_tensor:s4:64x1
```
dzarukin: There are no int4 zero-points for int8 weights. Precision of zp can't be less than weights precision. If it passes through API checks, those must be tightened.

tests/benchdnn/inputs/matmul/harness_matmul_decompression
inteldimitrius: Added, thanks!

tests/benchdnn/inputs/matmul/harness_matmul_decompression
inteldimitrius: Added: [x64: matmul: int8: tighten conditions for ZP](https://github.com/uxlfoundation/oneDNN/pull/5238/commits/b1d25901b0d17b8d4cac5a434b849aade57f74cb)

src/common/primitive_attr_quant.hpp
vpirogov: @densamoilov, mask 6 represents the case of static quantization when quantization parameters are pre-defined. It's probably not relevant to LLMs.

dzarukin:
Please cherry-pick commits from #5317. This will update matmul ci overage with relevant problems to verify the whole feature completeness.

Then run this command to trigger full CPU validation:
```
make test
disable benchdnn_all
enable benchdnn_matmul
enable arch_cpu_adl
enable arch_cpu_clx
enable arch_cpu_dmr
enable arch_cpu_gnr
enable arch_cpu_hsw
enable arch_cpu_nhm
enable arch_cpu_nvl
enable arch_cpu_skx
enable arch_cpu_snb
enable arch_cpu_spr
enable arch_cpu_srf
```

The job would highlight any remaining issues if any supporting the claim all of existing one gets resolved with this PR.

inteldimitrius:
@dzarukin did it in separate tmp branch: https://github.com/uxlfoundation/oneDNN/pull/5447
the order of cherry picks: main -> fixes https://github.com/uxlfoundation/oneDNN/pull/5446 -> enabling full tensor -> test coverage. 
[Job link](https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/f1748d0a-77cc-f18e-9173-d4f5ef20c6a0)
One early catch [assert](https://github.com/uxlfoundation/oneDNN/blob/9dbc135bdc43ca2c83ea7fba1de7aa881110c2fd/src/cpu/x64/brgemm/jit_brgemm_amx_uker.cpp#L1276) (out of the scope of this PR): [benchdnn: inputs: matmul: exclude asserting case from CI coverage](https://github.com/uxlfoundation/oneDNN/pull/5447/commits/35df5709f9189b18ef68f32c82f4c8a07bcfb1c9)

dzarukin (approved):

densamoilov (approved):

georgen117 (approved):

PR #5245, opened 2026-06-03
[GPU] JIT GEMM: consolidate problem properties in kernel/exec configs

echeresh:
This is the groundwork for https://github.com/uxlfoundation/oneDNN/pull/5181. PR consolidates GEMM problem properties in `kernel_config_t` and `exec_config_t` reducing the scattered swap-AB logic.

The PR reworks how the GPU JIT GEMM primitive descriptor carries problem state. Previously the PD re-derived GEMM state from `desc()`/`attr()`/mds at each call site, recomputing the same facts across kernel selection and execute. Now init captures that state once into a `kernel_config_t` (`cfg`).

**Responsibility split.** Init translates `desc()`/`attr()`/mds into `cfg`. Once `cfg` is initialized, kernel selection and execute read it for problem state, not the descriptor or attributes. Two narrow reads stay on the originals:

- Format setup and support checks run earlier in init, before `cfg` exists
- Exec-time host-scalar scale folding still reads `attr()`

`cfg` wraps a `gemmstone::GEMMProblem` plus the dnnl-side state the problem doesn't carry - dims, leading dims, batch strides, quant mds, post-op/bias metadata.

**Problem is the source of truth.** Inside `cfg`, feature queries (transpose, zero points, sum, sum_ab, ...) derive from the wrapped problem instead of cachingtheir own bools or re-reading attributes.

**Build is two phases.** `init_kernel_config` seeds `cfg` from the descriptor, `finalize_problem` then finalizes it for the target hardware.

### DIFF
--- a/src/gpu/intel/gemm/jit.cpp
+++ b/src/gpu/intel/gemm/jit.cpp
@@ -20,6 +20,7 @@
 #include "gemmstone/driver_info.hpp"
 #include "gpu/intel/compute/utils.hpp"
 #include "gpu/intel/gemm/host_scalars.hpp"
+#include "gpu/intel/gemm/jit/pd.hpp"
 #include "gpu/intel/gemm/jit/walk_orders.hpp"
 #include "gpu/intel/jit/ir/block_2d_utils.hpp"
 #include "gpu/intel/jit/utils/utils.hpp"
@@ -35,6 +36,37 @@ namespace gpu {
 namespace intel {
 namespace gemm {
 
+namespace jit {
+struct exec_args_t {
+    const memory_storage_t *a = nullptr;
+    const memory_storage_t *b = nullptr;
+    const memory_storage_t *c = nullptr;
+
+    const memory_storage_t *ao = nullptr;
+    const memory_storage_t *bo = nullptr;
+    const memory_storage_t *a_scales = nullptr;
+    const memory_storage_t *b_scales = nullptr;
+    const memory_storage_t *c_scales = nullptr;
+    const memory_storage_t *ag = nullptr;
+    const memory_storage_t *bg = nullptr;
+
+    const memory_storage_t *co = nullptr;
+
+    const memory_storage_t *sround_seed = nullptr;
+
+    int16_t ao_host_scalar = 0;
+    int16_t bo_host_scalar = 0;
+    int16_t co_host_scalar = 0;
+
+    float alpha = 1.0f;
+
+    size_t off_a0 = 0;
+    size_t off_b0 = 0;
+    size_t off_c0 = 0;
+    int64_t off_co0 = 0;
+};
+} // namespace jit
+
 bool check_memory_storage(const memory_storage_t *storage, const char *name) {
     if (storage && *storage) return true;
     VERROR(primitive, gpu, "%s,%s: %s", "jit::gemm", "argument is not set",
@@ -44,23 +76,15 @@ bool check_memory_storage(const memory_storage_t *storage, const char *name) {
 
 status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
         intel::stream_t *compute_stream, zero_pool_t *zero_pool,
-        const memory_storage_t &a, const memory_storage_t &b,
-        const memory_storage_t &c, const memory_storage_t *ao,
-        const memory_storage_t *bo, int16_t ao_host_scalar,
-        int16_t bo_host_scalar, const memory_storage_t *a_scales,
-        const memory_storage_t *b_scales, const memory_storage_t *c_scales,
-        const memory_storage_t *ag, const memory_storage_t *bg,
-        const memory_storage_t &co, int16_t co_host_scalar,
-        const memory_storage_t *c_temp,
-        const memory_storage_t *zero_buf_scratchpad,
-        const memory_storage_t *sround_seed, int po_count,
+        const jit::exec_args_t &exec_args, const memory_storage_t *c_temp,
+        const memory_storage_t *zero_buf_scratchpad, int po_count,
         const memory_storage_t **po_srcs, int64_t offset_a, int64_t offset_b,
         int64_t offset_c, int64_t offset_aq, int64_t offset_bq,
-        int64_t offset_co, int64_t *offset_po_src, int32_t lda, int32_t ldb,
-        int32_t ldc, int32_t m, int32_t n, int32_t k, int32_t k0, float alpha,
-        float beta, int32_t cmask, bool last_k_block, bool swap_ab,
+        int64_t offset_co, int64_t *offset_po_src, int32_t m, int32_t n,
+        int32_t k, int32_t k0, float alpha, float beta, bool last_k_block,
         bool disable_hilbert) const {
-    if (pd()->desc()->batch() == 0) return status::success;
+    const auto &cfg = pd()->cfg();
+    if (cfg.batch() == 0) return status::success;
 
     std::unique_ptr<memory_storage_t> zeros;
     const memory_storage_t *zero_buf = nullptr;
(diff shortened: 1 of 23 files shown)


src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
```diff
@@ -183,6 +183,7 @@ struct GEMMProblem : public CommonProblem {
     int aoPtrDims = -1, boPtrDims = -1;             // A/B offset dimensionality (-1: none; 0: scalar; 1: vector, 2: matrix)
     int coPtrDims = -1;                             // C offset dimensionality (-1: none or hostscalar; 0 - others)
     int asPtrDims = -1, bsPtrDims = -1, csPtrDims = -1;           // A/B scale dimensionality (-1: none; 0: scalar; 1: vector, 2: matrix)
+    int agPtrDims = -1, bgPtrDims = -1;             // A/B group-sum dimensionality (-1: none; 0: scalar; 1: vector, 2: matrix)
```
rjoursler: Why is this here if it does not effect code generation at all?

src/gpu/intel/gemm/jit/gen_kernel.hpp
```diff
@@ -154,14 +154,24 @@ struct gen_nocopy_desc_t : public gen_desc_t {
     gemmstone::Scalar beta_;
 };
 
-struct gen_xe_systolic_kernel_desc_t : public gen_desc_t {
+// Runtime problem sizes for kernel selection (not held by GEMMProblem).
+struct select_dims_t {
+    dim_t m = -1, n = -1, k = -1, batch = 1;
+};
```
rjoursler: We likely should move strides into here and the matrix alignment parameter out of the GEMMProblem.

src/gpu/intel/gemm/jit/gen_kernel.hpp
```diff
@@ -154,14 +154,24 @@ struct gen_nocopy_desc_t : public gen_desc_t {
...
+
+// Systolic perf/dispatch inputs not modeled by GEMMProblem. `batched` is the
+// 0/1 with_batch() signal, not in.batchDims (see select_kernel).
+struct systolic_perf_params_t {
+    bool batched = false;
+    bool packed_c = false;
+    int unroll_m = 0, unroll_n = 0;
+    bool alt = false;
+};
```
rjoursler: I am trying to figure out what the purpose of this is. Is this effectively a legacy structure for the pre-XeHPC kernels?

src/gpu/intel/gemm/jit/include/gemmstone/problem.hpp
echeresh: Agree, dropped from here.

src/gpu/intel/gemm/jit/gen_kernel.hpp
echeresh: This structure and `systolic_perf_params_t` was an attempt to make `select_kernel()` signature more readable but now I see that it's more confusing. Suggest to remove both and keep the old longer signature - with explicit m/n/k and lda/ldb/ldc and other parameters.

> the matrix alignment parameter out of the GEMMProblem.

Looks like the alignment is deep inside the matrix addressing structure so I suggest to keep it as is - it's part of the kernel that impacts its generation.

src/gpu/intel/gemm/jit/gen_kernel.hpp
echeresh: This is just to group parameters, as commented above let's drop it for simplicity.

src/gpu/intel/gemm/jit/gen_kernel.hpp
rjoursler: The main thing I was thinking here is that the logic is backwards, see [MFDNN-15136](https://jira.devtools.intel.com/browse/MFDNN-15136). I guess it's not necessarily that we should remove alignment from `GEMMProblem`, but that select kernel is where it needs to be set.

src/gpu/intel/gemm/jit/gen_kernel.hpp
rjoursler: I do think grouping parameters is a good idea, so feel free to keep this. I was more just trying to figure out why there are two groupings.

src/gpu/intel/gemm/jit/pd.hpp
```diff
@@ -192,102 +269,90 @@ struct pd_t : public gemm::pd_t {
...
+void pad_lda(kernel_config_t &cfg, bool swap);
+void swap_fold(kernel_config_t &cfg, bool swap);
+
+status_t transfer_post_ops(
+        gemmstone::GEMMProblem &problem, gpu_post_ops_t &&post_ops);
+
+// Runtime-only exec state (storage pointers, host scalars, offsets, resolved
+// alpha); the finalized config_ is reached by reference via cfg(). A/B storage
+// is swap_ab-routed once in build_exec_config; the config it reaches is folded.
+struct exec_config_t {
```
rjoursler: nit: I would suggest a name change for this structure to `exec_args_t` or something like that. This structure seems like it is effectively wrapper of kernel arguments, and as such it doesn't seem right to me to call it a `config`.

src/gpu/intel/gemm/jit.cpp
```diff
@@ -336,6 +325,120 @@ status_t gen_t::launch_nocopy(const exec_ctx_t &ctx,
...
+static T pick_a(bool s, T ua, T ub) {
+    return s ? ub : ua;
+}
+template <typename T>
+static T pick_b(bool s, T ua, T ub) {
+    return s ? ua : ub;
+}
+
+static status_t build_exec_config(jit::exec_config_t &exec_cfg,
+        const gen_t::pd_t *pd, const exec_ctx_t &ctx) {
```
rjoursler: Could you summarize how this fits in to the kernel creation flow just to make sure I fully understand it? From what I am seeing, I feel like this should be  a member function of `kernel_config_t' and that  the `finalize_problem` function should be unnecessary.

src/gpu/intel/gemm/jit/gen_kernel.hpp
echeresh: Thanks for the link. I certainly lack context here, there is also an update in `finalize()`:

https://github.com/uxlfoundation/oneDNN/blob/a03ad9934930445394c965fc67cf556a9494645c/src/gpu/intel/gemm/jit/gen_kernel.cpp#L106-L110

Suggest to keep it out of scope - I tried to keep the values and flow aligned to the existing version with some exceptions related to the initialization and init checks. The PR is already large so I'd like to limit unnecessary changes.

src/gpu/intel/gemm/jit.cpp
echeresh: I changed names and refactored this a bit - please take a look. Also please check the description for a high-level summary.


This function (which is now `init_exec_args()`) is called from primitive `execute()` to prepare kernel arguments for kernel submission.

> From what I am seeing, I feel like this should be a member function of kernel_config_t'

I kept it a standalone function as it consolidates data from both `kernel_config_t` and `exec_ctx_t` into `exec_args_t` object. Both `kernel_config_t` and `exec_ctx_t` are kind of equal

> that  the finalize_problem` function should be unnecessary.

`finalize_problem()` has some [logic](https://github.com/uxlfoundation/oneDNN/blob/601dd5c3f6bb80930531c07d4e0e0dfeeec7adb6/src/gpu/intel/gemm/jit/pd.cpp#L730) in it, I think it should be kept as is.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip
enable benchdnn_rnn

src/gpu/intel/gemm/jit/gen_kernel.cpp
```diff
@@ -638,14 +643,14 @@ status_t gen_xe_systolic_kernel_desc_t::select_kernel(
...
+    problem_.Ta = problem_.Ta_ext = in.Ta_ext;
+    problem_.Tb = problem_.Tb_ext = in.Tb_ext;
+    problem_.Tc = in.Tc;
+    problem_.Tc_ext = in.Tc_ext;
     problem_.Ts = Type::f32;
-    problem_.Tao = convert_dnnl_to_kernel_type(ao_type);
-    problem_.Tbo = convert_dnnl_to_kernel_type(bo_type);
-    problem_.Tco = convert_dnnl_to_kernel_type(co_type);
+    problem_.Tao = Type::s32;
+    problem_.Tbo = Type::s32;
```
rjoursler, line 652: Why are we hardcoding these now?

PR #5249, opened 2026-06-03
cpu: x64: matmul: fix a guard for merging batch into M

densamoilov:
This PR fixes MFDNN-15137.

We introduced this issue in #4632 when we needed to fix the guard that restricted when the batch dimension could be merged into the M by prohibiting the optimization for the `acbd` format. The problem was that the check relied on format tags, which excluded degenerate shapes such as `59578x1x1x1024`. These are legitimate shapes for the optimization and excluding them led to performance regressions such as the one reported in MFDNN-15137.
```
./benchdnn --mode=P --matmul --dt=bf16:bf16:bf16 --stag=abdc --wtag=abdc --dtag=abcd 59578x1x1x1024:1x1x1024x1
```
I also created a task to migrate brgemm matmul layout decisions from format tags to strides: MFDNN-15148.

Perf results for RGAT on GNR: vs rls-v3.12
image
Perf results for RGAT on GNR: vs main
image

### DIFF
--- a/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
+++ b/src/cpu/x64/matmul/brgemm_matmul_utils.hpp
@@ -502,6 +502,12 @@ void mem_advice_init(brgemm_matmul_conf_t &bgmmc);
 
 bool is_batch_layout_trivial(const memory_desc_wrapper &mdw, const dim_t batch);
 
+// Returns true if logical dimension `inner_dim` is nested immediately inside
+// `outer_dim` in memory, i.e. no other dimension is physically interleaved
+// between them (`stride(outer_dim) == dim(inner_dim) * stride(inner_dim)`).
+bool dims_adjacent(const memory_desc_wrapper &mdw, const int outer_dim,
+        const int inner_dim);
+
 /**
  * Returns the total block size along the K dimension, as the product of
  * the fixed outer block size and the VNNI granularity.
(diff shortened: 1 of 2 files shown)


densamoilov:
make test
disable test_device_gpu

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
```diff
@@ -502,6 +502,12 @@ void mem_advice_init(brgemm_matmul_conf_t &bgmmc);
 
 bool is_batch_layout_trivial(const memory_desc_wrapper &mdw, const dim_t batch);
 
+// Returns true if logical dimension `inner` is nested immediately inside
+// `outer` in memory, i.e. no other dimension is physically interleaved
+// between them (`stride(outer) == dim(inner) * stride(inner)`).
+bool dims_adjacent(
+        const memory_desc_wrapper &mdw, const int outer, const int inner);
```
dzarukin: Nit: 
```suggestion
        const memory_desc_wrapper &mdw, const int outer_dim, const int inner_dim);
```

Maybe move to memory_desc_wrapper?

dzarukin (approved):

src/cpu/x64/matmul/brgemm_matmul_utils.hpp
densamoilov: > Maybe move to memory_desc_wrapper?

I decided to keep it in matmul utils (the only place where it's currently used) for now because we will need to decide on where we should place these kinds of functions as part of MFDNN-15137.

inteldimitrius (approved):

PR #5252, opened 2026-06-04
cpu: aarch64: binary: Avoid parallel overhead on tiny workloads.

renato-arantes:
# Description

This patch improves the AArch64 binary JIT performance for tiny workloads by avoiding multi-threading when the per-thread work is too small.

Previously, the binary JIT used the current runtime thread count even for very small tensors. For these cases, the parallel scheduling overhead can dominate the actual JIT kernel execution time. This patch limits execution to a single thread when:

```cpp
dst.nelems() <= dnnl_get_current_num_threads() * 128
```

This makes the heuristic depend on the amount of work available per thread rather than on a fixed absolute tensor-size threshold.

### Motivation

The original small-shape issue was observed while comparing AArch64 binary JIT against ACL. ACL performs much better for some tiny shapes because it avoids paying excessive parallelisation overhead. However, directly using an ACL-style absolute threshold, such as ```dst.nelems() < 40000```, is too broad for JIT: it serialises shapes where JIT still benefits from parallel execution.

The proposed heuristic is intentionally JIT-specific. It maintains a single-thread path for very small tensors while preserving parallel execution for larger shapes.

For example, with OMP_NUM_THREADS=16, the threshold becomes: 

```
128 * 16 = 2048 elements
```

This keeps the tiny-shape wins while avoiding regressions observed with larger shapes, such as 3264 and 24576 output elements.

### Performance

Benchmarked with benchdnn binary tests, ```mode=P``` and  ```OMP_NUM_THREADS=16```. The numbers below are the sum of ```avg_time across``` 432 binary test cases:

```
+----------------------------+--------------+
| Implementation             | Sum avg time |
+----------------------------+--------------+
| Main JIT                   | 1.92725 ms   |
| New 128 * nthr heuristic   | 1.39529 ms   |
| ACL                        | 3.04138 ms   |
+----------------------------+--------------+
```

The new heuristic is:

1.38x faster than main JIT overall
2.18x faster than ACL overall on this test set

For the smallest bucket, ```<=2048``` output elements:

```
+------------------+--------------+
| Implementation   | Sum avg time |
+------------------+--------------+
| Main JIT         | 1.34566 ms   |
| New heuristic    | 0.801435 ms  |
| ACL              | 0.941339 ms  |
+------------------+--------------+

```

This shows the heuristic preserves the intended tiny-workload improvement while avoiding over-serializing larger shapes.

# Checklist

## General

## Performance improvements

### DIFF
--- a/src/cpu/aarch64/jit_uni_binary.cpp
+++ b/src/cpu/aarch64/jit_uni_binary.cpp
@@ -90,6 +90,13 @@ static bool data_format_supported(
             || (is_superset(isa, asimd) && blk_size == 4);
 }
 
+static int get_exec_num_of_threads(const memory_desc_wrapper &dst_d) {
+    const int max_nthr = dnnl_get_current_num_threads();
+    // minimum approximate work-per-thread we chose from experimental data.
+    const dim_t min_work_per_thread = 128;
+    return dst_d.nelems() <= max_nthr * min_work_per_thread ? 1 : max_nthr;
+}
+
 status_t jit_uni_binary_t::pd_t::init(engine_t *engine) {
     using sm = primitive_attr_t::skip_mask_t;
 
@@ -612,7 +619,7 @@ void jit_uni_binary_t::execute_no_bcast_strategy(const data_t *src0,
         const dim_t nelems0_tail = nelems_per_aligned_dims % simd_w;
         const bool has_tail = nelems0_tail > 0;
 
-        const int nthr = dnnl_get_current_num_threads();
+        const int nthr = get_exec_num_of_threads(dst_d);
         const dim_t thr_per_nelems_group = nstl::min(
                 nstl::max(nthr / batch, (dim_t)1), nelems0_simd + has_tail);
 
@@ -620,8 +627,8 @@ void jit_uni_binary_t::execute_no_bcast_strategy(const data_t *src0,
         // Iterate over batch and over outer dims.
         // Divide number of threads by batch size and limiting it by a number
         // of outer_dims nelems to parallel over it when needed.
-        parallel_nd(
-                batch, thr_per_nelems_group, [&](dim_t b, dim_t nelems_group) {
+        parallel_nd_ext(nthr, batch, thr_per_nelems_group,
+                [&](int, int, dim_t b, dim_t nelems_group) {
             dim_t start = 0, end = 0;
             balance211(nelems0_simd + has_tail, thr_per_nelems_group,
                     nelems_group, start, end);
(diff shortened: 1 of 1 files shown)


Sqvid:
> Benchmarked with benchdnn binary tests, mode=P and OMP_NUM_THREADS=16. The numbers below are the sum of avg_time across 432 binary test cases:

The changes look good but can you please supply separate benchmarks of small and large shapes and how they are affected by your change. Averaging over the entire test set doesn't give a clear enough picture.

jondea:
This sounds broadly reasonable, but I agree with Sid, average numbers hide a lot of sins. For something like this I would expect a systematic parameter search over size and thread count. Probably best visualised with a heatmap with the ratio of runtime between parallelising or not parallelising. Also, what is the current test set? If it is taken from the benchdnn test set, then the shapes are picked more for testing than performance. For heuristics you want to do a grid search.

I'm glad you kept it simple though. We have seen with other primitives that it is best to pick either all threads or just  because of friction at the framework level caused by switching between different (non-trivial) numbers of threads on consecutive layers.

renato-arantes:
> This sounds broadly reasonable, but I agree with Sid, average numbers hide a lot of sins. For something like this I would expect a systematic parameter search over size and thread count. Probably best visualised with a heatmap with the ratio of runtime between parallelising or not parallelising. Also, what is the current test set? If it is taken from the benchdnn test set, then the shapes are picked more for testing than performance. For heuristics you want to do a grid search.
> 
> I'm glad you kept it simple though. We have seen with other primitives that it is best to pick either all threads or just because of friction at the framework level caused by switching between different (non-trivial) numbers of threads on consecutive layers.  

As shown in the plot below, which explores the output element space with a grid search, the heuristic yields substantial improvements and, as expected, becomes more effective as the number of threads increases. 

The border blue/yellow marks the boundary of the heuristic's influence. Blue is within the heuristic's influence, and yellow is where it stops acting and the full number of threads is selected.

image



The table below emphasises the overall improvement achieved by the heuristic. The more threads, the greater the performance improvement.


```text
+---------+-------------------+--------------------+-------------+
| threads | sum_parallel (ms) | sum_heuristic (ms) | sum_speedup |
+---------+-------------------+--------------------+-------------+
| 1       | 0.514608          | 0.513611           | 1.00x       |
| 4       | 0.375263          | 0.345738           | 1.08x       |
| 8       | 0.357177          | 0.306090           | 1.17x       |
| 16      | 0.423946          | 0.305278           | 1.40x       |
| 32      | 0.553979          | 0.312663           | 1.80x       |
| 64      | 5.033399          | 0.380292           | 13.23x      |
+---------+-------------------+--------------------+-------------+
```

Sqvid (approved):
Great results! Thanks for the patch!

jondea (approved):

PR #5258, opened 2026-06-04
common, cpu: move a key to a string to combine data types

dzarukin:
Key notations are described next to a key struct.
The idea is to express supported configs by regex strings to combine those who share identical implementations that support such configs.

### DIFF
--- a/src/common/impl_list_item.cpp
+++ b/src/common/impl_list_item.cpp
@@ -1,0 +1,96 @@
+/*******************************************************************************
+* Copyright 2026 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "common/impl_list_item.hpp"
+
+namespace dnnl {
+namespace impl {
+
+pk_dt_impl_key_t::pk_dt_impl_key_t(prop_kind_t kind, data_type_t src_dt,
+        data_type_t wei_dt, data_type_t dst_dt)
+    : kind_(kind) {
+    using namespace data_type;
+    using namespace prop_kind;
+
+    const bool is_fwd
+            = utils::one_of(kind_, forward_training, forward_inference);
+    const bool is_bwd_d = kind_ == backward_data;
+    const bool is_bwd_w = kind_ == backward_weights;
+
+    // src_dt
+    if (is_fwd) {
+        if (utils::one_of(src_dt, s8, u8)) {
+            s_dt_ += "xi8";
+        } else if (utils::one_of(src_dt, f8_e5m2, f8_e4m3)) {
+            s_dt_ += "xf8";
+        } else {
+            s_dt_ += dnnl_dt2str(src_dt);
+        }
+    } else if (is_bwd_d) {
+        s_dt_ += "*";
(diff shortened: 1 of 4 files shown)


dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_ip

src/common/impl_list_item.hpp
```diff
@@ -21,26 +21,33 @@
...
 
-// This key takes prop_kind and correspondent data_type for src, wei and dst.
+// This key takes prop_kind and correspondent data types for src, wei and dst
+// and forms a regex string that compounds several data type configs.
+// Notation:
+// * To use exact data type, use its name as in include, e.g. f32, bf16, etc.
+// * To use any floating-point type, use xf. Adding a number will limit bits,
+//   e.g., xf8 means any fp8 type.
+// * To use any int8 type, use x8.
+// * To use any data type use *.
```
densamoilov, line 36: It seems that the documentation is split between header and cpp files. Usually the documentation resides in headers at the declaration spot.

src/common/impl_list_item.hpp
```diff
@@ -21,26 +21,33 @@
...
 
 private:
-    size_t value() const {
-        const size_t dtm = data_type::data_type_max;
-        const size_t m1 = static_cast<size_t>(kind) * dtm;
-        const size_t m2 = (m1 + static_cast<size_t>(src_dt)) * dtm;
-        const size_t m3 = (m2 + static_cast<size_t>(wei_dt)) * dtm;
-        return m3 + static_cast<size_t>(dst_dt);
-    }
+    prop_kind_t kind_;
```
densamoilov, line 50: It seems that we have an implicit assumption that we will land in the right data type string slot by using `kind_` (since it's not used in the comparison function anymore). It may happen that we'll miss that slot after adding some combinations in the future.

src/common/impl_list_item.cpp
```diff
@@ -0,0 +1,104 @@
...
+namespace impl {
+
+// This key takes prop_kind and correspondent data types for src, wei and dst
+// and forms a regex string that compounds several data type configs.
+// Notation:
+// * To use exact data type, use its name as in include, e.g. f32, bf16, etc.
+// * To use any floating-point type, use xf. Adding a number will limit bits,
+//   e.g., xf8 means any fp8 type.
+// * To use any int8 type, use x8.
+// * To use any data type use *.
```
densamoilov: > To use any floating-point type, use xf.

It seems that it may not be accurate because it doesn't cover f8 which is always covered by xf8 now (based on how the branching is organized now)


> To use any int8 type, use x8.

Nit. Technically, it means any data type with 8 bits according to your definition of the letters/numbers. Unambiguous definition for int8 (or any other int) could looks something like this xi8, xu8, xs8, etc.

densamoilov (approved):
Looks good to me. The comments are mostly forward looking concerns that don't block this particular PR.

dzarukin:
make test
disable benchdnn_all
enable benchdnn_conv
enable benchdnn_ip

src/common/impl_list_item.hpp
dzarukin: Sure, kept a single version of documentation in a header, thanks.

src/common/impl_list_item.hpp
dzarukin: Discussed offline, updated the comparison function to include `kind_` back.

src/common/impl_list_item.cpp
dzarukin: > It seems that it may not be accurate because it doesn't cover f8 which is always covered by xf8 now (based on how the branching is organized now)

In existing use case it should be fine with fp8 type too. When it stops, there's a syntax like `xf16|f32` to narrow the scope. I didn't want to introduce it without great necessity now since it's C++ and not Python.

> Nit. Technically, it means any data type with 8 bits according to your definition of the letters/numbers. Unambiguous definition for int8 (or any other int) could looks something like this xi8, xu8, xs8, etc.

Renamed to `xi8`, thanks.

mzhukova (approved):

PR #5266, opened 2026-06-05
common: fix name matching with system struct 'user' in llvm-android toolchain

azhai219:
### Problem
Name matching in the affected code path could collide with the system-defined struct user when building with the LLVM Android toolchain.

### Fix
Adjusted the matching and naming logic to avoid the user symbol conflict and keep identifier resolution unambiguous.

### Impact
Restores stable Android LLVM builds with no intended runtime behavior change. Low impact on oneDNN.


dzarukin:
@azhai219, could you, please, post a compilation error message and a link to the conflicting header, or post a piece of it, that declares same-named classes? Thank you.

azhai219:
close the PR since the issue has been fixed. refer to https://github.com/openvinotoolkit/openvino/pull/36294
cc @dzarukin

PR #5267, opened 2026-06-05
cpu: aarch64: Enable s8:s8:s32 jit_int8_matmul

aditew01:
# Description

This enables s8:s8:s32 matmul

Tests already included in `./benchdnn/inputs/matmul/test_matmul_int8`

# Checklist

## General

## Performance improvements

Threads=8, Matmul dims: m=n=k=1024

| gemm:jit avg time (ms) | jit:int8 avg time (ms) | Speedup |
|---:|---:|---:|
| 24.6715 | 0.51798 | **47.63x** |

### New features

- [ ] Was the RFC approved?

### Bug fixes

### DIFF
--- a/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
+++ b/src/cpu/aarch64/matmul/jit_int8_matmul.cpp
@@ -133,12 +133,24 @@ struct jit_int8_matmul_kernel_t : public jit_generator_t {
         const ZReg zp_b_prod0 = ZReg(4);
         const ZReg zp_b_prod1 = ZReg(5);
 
-        for (int a = 0; a < bdb; a++) {
-            for (int b = 0; b < ldb; b++) {
-                if (brg_.is_s8 || brg_.is_u8_s8)
-                    scvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
-                else
-                    ucvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+        // Plain s32 dst stores the accumulators directly
+        // Convert to f32 when zero-point compensation, scales, bias addition, or eltwise post-op
+        // and then convert back to s32 when dst is s32
+        const bool can_store_s32_acc_directly = brg_.dst_dt == s32
+                && brg_.zp_type_a == jit_int8_broadcast_t::none
+                && brg_.zp_type_b == jit_int8_broadcast_t::none
+                && brg_.zp_type_c == jit_int8_broadcast_t::none
+                && !brg_.with_scales && !brg_.is_bias
+                && eltwise_injectors_.empty() && !brg_.with_dst_scales;
+
+        if (!can_store_s32_acc_directly) {
+            for (int a = 0; a < bdb; a++) {
+                for (int b = 0; b < ldb; b++) {
+                    if (brg_.is_s8 || brg_.is_u8_s8)
+                        scvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+                    else
+                        ucvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -870,8 +954,8 @@ status_t jit_int8_matmul_t<isa>::pd_t::init(engine_t *engine) {
         return true;
     };
 
-    const bool problem_dt_correct
-            = (is_s8 || is_u8 || is_u8_s8) && utils::everyone_is(f32, dst_type);
+    const bool problem_dt_correct = (is_s8 || is_u8 || is_u8_s8)
+            && utils::one_of(dst_type, f32, bf16);
```
cfRod: dst type does not include s32?

aditew01: oops. I missed the most imp bit in the rebase. i'll fix it.

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
```diff
@@ -133,12 +133,21 @@ struct jit_int8_matmul_kernel_t : public jit_generator_t {
         const ZReg zp_b_prod0 = ZReg(4);
         const ZReg zp_b_prod1 = ZReg(5);
 
-        for (int a = 0; a < bdb; a++) {
-            for (int b = 0; b < ldb; b++) {
-                if (brg_.is_s8 || brg_.is_u8_s8)
-                    scvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
-                else
-                    ucvtf(acc(a, b).s, P_ALL_ONE, acc(a, b).s);
+        const bool f32_postop = brg_.dst_dt != s32
```
Sqvid: Dont you want to check `brg_.dst_dt == s32` rather than the inverse? You are later using `f32_postop` in `case s32` of the `switch`. And do you want these conditions to be connected with `||` or `&&`? For instance `f32_postop` will currently be true even if just `brg_.zp_type_c != none`, which seems strange

Sqvid (changes requested):

src/cpu/aarch64/matmul/jit_int8_matmul.cpp
aditew01: hmm, i should reword this variable to fp32_epilogue to avoid confusion.

renato-arantes: A bit of comment here explaining what this condition is would be very helpful.

Sqvid: I am not sure I would know what an `fp32_epilogue` was supposed to be either :grimacing:

aditew01: maybe it makes more sense now?

aditew01: added

aditew01: or `needs_f32_post_accumulation`? choosing right variable names is intriguing. 
But hopefully the comment makes sense

michalowski-arm: I think we can potentially flip it and make it something like `can_store_s32_acc_directly`? And then we check for `brg_.dst_dt == s32` and no extra work like postops, scales etc. But yeah I'm not sure what name would work best here, `epilogue` is not really used in onednn though.

aditew01: intuitively, it should work, but if you have postops / scales / zp you'd still want to cast to fp32, compute and cast to s32. I realised it after running the benchdnn tests.

michalowski-arm: yea, but something like:
```
const bool can_store_s32_acc_directly = brg_.dst_dt == s32
        && brg_.zp_type_a == jit_int8_broadcast_t::none
        && brg_.zp_type_b == jit_int8_broadcast_t::none
        && brg_.zp_type_c == jit_int8_broadcast_t::none
        && !brg_.with_scales && !brg_.is_bias
        && eltwise_injectors_.empty() && !brg_.with_dst_scales;
```
should work no? and then we check `if (!can_store_s32_acc_directly) { ... }`

aditew01: to extend this to s8 / u8, you'd need to do a : 

``` 
const bool can_store_s32_acc_directly = (brg_.dst_dt == s32 || brg_.dst_dt == s8) 
        && brg_.zp_type_a == jit_int8_broadcast_t::none
        && brg_.zp_type_b == jit_int8_broadcast_t::none
        && brg_.zp_type_c == jit_int8_broadcast_t::none
        && !brg_.with_scales && !brg_.is_bias
        && eltwise_injectors_.empty() && !brg_.with_dst_scales;
```

so, i feel the existing one is similarly extendible . although I agree that it's a bit ugly.

michalowski-arm: `(brg_.dst_dt == s32 || brg_.dst_dt == s8)` can be expressed as `one_of(brg_.dst_dt, s32, s8)` which I think is fine

michalowski-arm: and I guess we can also simply the broadcast check as:
`everyone_is(jit_int8_broadcast_t::none, brg_.zp_type_a, brg_.zp_type_b, brg_.zp_type_c)`

aditew01: wait, this won't work generically right? ie, a test can have eltwise_injectors_ but not `zp`'s ?

michalowski-arm: but we check for `&& eltwise_injectors_.empty()` too

aditew01: okay, makes sense. I'll try this out

michalowski-arm (approved):
LGTM

Sqvid (approved):
LGTM, thanks. Could you post some performance numbers here please?

PR #5268, opened 2026-06-05
cpu: aarch64: remove X_TRANSLATOR_STACK register

Sqvid:
# Description

This removes a confusing, extra stack-tracking register and frees it up for general-purpose use.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/brgemm/jit_brdgmm_kernel.hpp
+++ b/src/cpu/aarch64/brgemm/jit_brdgmm_kernel.hpp
@@ -20,11 +20,8 @@
 
 #include "common/c_types_map.hpp"
 #include "common/nstl.hpp"
-#include "common/type_helpers.hpp"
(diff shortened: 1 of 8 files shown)


Sqvid:
Passing Nightly correctness here: https://github.com/uxlfoundation/oneDNN/actions/runs/27021089245/job/79752038574

jondea (approved):
Great find, a nice little code quality improvement.

It might be worth checking with @kawakami-k (or @fj-higuchi appears in the blame, but he may be inactive) about the original intent of this variable, although I'm not sure it needs to block it going in.

kawakami-k (approved):

Sqvid:
> Great find, a nice little code quality improvement.

This also seems to lead to a few miscellaneous runtime improvements since we don't allocate an unused 128 KB of memory every preamble.

renato-arantes (approved):

PR #5272, opened 2026-06-06
[GPU] More threads/EU changes

Simonsays095:
My recent Variable Registers per Thread (VRT) PR didn't quite get the entire picture: CRI/NVLP have a few more changes compared to previous architectures:
- NVLP has 10 hardware threads per EU, up from 8 -- affects grf128 kernels
- CRI has 2048 GRF/EU, up from 1024 -- affects maximum workgroup size
 
Both of these are updated in this PR. But most of the lines changed are tangential: I noticed that we do a lot of conversion between `ngen::Product` and `gpu_product_t` in both device_info and dsl hw_t code. Rather than supporting two sets of interfaces for these functions, I removed the ad hoc `gpu_product_t` implementations and replaced them with unique pointers to `ngen::Product` in every case. We use the pointer everywhere we don't want an ngen dependency, and we dereference when we need to, in code where the ngen dependency is OK.

### DIFF
--- a/src/gpu/intel/compute/device_info.cpp
+++ b/src/gpu/intel/compute/device_info.cpp
@@ -247,21 +247,24 @@ size_t device_info_t::max_wg_size(
     return device_max_wg_size;
 }
 
-int device_info_t::grf_per_eu(gpu_arch_t gpu_arch) {
-    switch (gpu_arch) {
-        case gpu::intel::compute::gpu_arch_t::xe_lp: return 896;
-        case gpu::intel::compute::gpu_arch_t::xe_hp:
-        case gpu::intel::compute::gpu_arch_t::xe_hpg:
-        case gpu::intel::compute::gpu_arch_t::xe_hpc:
-        case gpu::intel::compute::gpu_arch_t::xe2:
-        case gpu::intel::compute::gpu_arch_t::xe3p:
-        case gpu::intel::compute::gpu_arch_t::xe3: return 1024;
-        case gpu::intel::compute::gpu_arch_t::unknown: return 896;
+namespace {
+int grf_per_eu(const ngen::Product &product) {
+    ngen::HW hw = ngen::getCore(product.family);
+    switch (hw) {
+        case ngen::HW::XeLP: return 896;
+        case ngen::HW::XeHP:
+        case ngen::HW::XeHPG:
+        case ngen::HW::XeHPC:
+        case ngen::HW::Xe2:
+        case ngen::HW::Xe3: return 1024;
+        case ngen::HW::Xe3p:
+            return product.family == ngen::ProductFamily::CRI ? 2048 : 1024;
+        case ngen::HW::Unknown: return 896;
+        default: gpu_error_not_expected();
     }
     return 1024;
 }
 
-namespace {
 int max_threads_per_eu(const ngen::Product &product) {
     ngen::HW hw = ngen::getCore(product.family);
     gpu_arch_t arch = jit::convert_ngen_arch_to_dnnl(hw);
@@ -285,10 +288,8 @@ int max_threads_per_eu(const ngen::Product &product) {
 int device_info_t::threads_per_eu(
         const ngen::Product &product, int grf_per_thread) {
     gpu_assert(grf_per_thread > 0) << "Invalid GRF per thread";
-    ngen::HW hw = ngen::getCore(product.family);
-    gpu_arch_t arch = jit::convert_ngen_arch_to_dnnl(hw);
     int hw_max = max_threads_per_eu(product);
-    return std::min(hw_max, grf_per_eu(arch) / grf_per_thread);
+    return std::min(hw_max, grf_per_eu(product) / grf_per_thread);
 }
 
 int device_info_t::max_slm_size(const ngen::Product &product) {
--- a/src/gpu/intel/compute/device_info.cpp
+++ b/src/gpu/intel/compute/device_info.cpp
@@ -65,6 +65,8 @@ uint64_t get_future_extensions(
     return extensions;
 }
 
+device_info_t::~device_info_t() = default;
+
 status_t device_info_t::init(
         impl::engine_t *engine, const std::vector<uint8_t> &cache_blob) {
     if (!cache_blob.empty()) {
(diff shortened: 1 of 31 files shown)


src/gpu/intel/gemm/jit/include/gemmstone/dsl/hw.hpp
echeresh: With `unique_ptr` this `operator==()` compares pointers - need a fix to compare values.

src/gpu/intel/compute/device_info.hpp
```diff
@@ -55,11 +51,8 @@ enum class gpu_arch_t {
     xe3p,
 };
 
-// Memory for storing ngen::Product to avoid directly including nGEN because of
-// header dependencies outside of src/gpu/intel.
-struct alignas(int) gpu_product_t {
-    unsigned char data[12];
-};
+// Use gpu_product_t as an opaque handle to an ngen::Product avoid an ngen dependency here
+using gpu_product_t = const ngen::Product *;
```
echeresh: Can we switch `gpu_product_t` usages to `const ngen::Product&`. This way we can avoid pointer/value mix and `ngen::Product` is more readable vs `gpu_product_t`.

src/gpu/intel/compute/device_info.cpp
```diff
@@ -399,7 +417,7 @@ status_t device_info_t::init_from_cache_blob(
     pos += sizeof(expected_type);
 
     DESERIALIZE(gpu_arch_, compute::gpu_arch_t);
-    DESERIALIZE(gpu_product_, compute::gpu_product_t);
+    DESERIALIZE(*product_, ngen::Product);
```
echeresh: Can it be null here? Please check.

src/gpu/intel/gemm/jit/dsl/hw.cpp
```diff
@@ -68,20 +75,42 @@ int hw_t::eus_per_core() const {
...
+int max_threads_per_eu(const ngen::Product product) {
+    auto family = product.family;
+    switch (getCore(family)) {
+        case ngen::HW::XeLP: return 7;
+        case ngen::HW::XeHP:
+        case ngen::HW::XeHPG:
+        case ngen::HW::XeHPC:
+        case ngen::HW::Xe2:
+        case ngen::HW::Xe3: return 8;
+        case ngen::HW::Xe3p: return family == ngen::ProductFamily::CRI ? 10 : 8;
```
echeresh:
```suggestion
        case ngen::HW::Xe3p: return family == ngen::ProductFamily::NVLP ? 10 : 8;
```

echeresh (approved):

src/gpu/intel/jit/utils/type_bridge.hpp
```diff
@@ -155,13 +147,6 @@ inline gemmstone::dsl::type_t to_ir(const data_type_t &dt) {
     return {};
 }
 
-inline compute::gpu_product_t to_gpu_product(const ngen::Product &p) {
-    compute::gpu_product_t result;
-    static_assert(sizeof(ngen::Product) == sizeof(compute::gpu_product_t), "");
-    std::memcpy(&result, &p, sizeof(result));
-    return result;
-}
```
rjoursler, line 163: I am not really a fan of this change. I think it just unnecessarily complicates the device info structure as now we have a non-trivial memory structure.  From offline discussions, it sounds like the main motivation of this is to remove the `to_gpu_product` interface. Why not just modify the interfaces requiring a `gpu_product_t` to take as input an `ngen::Product &` instead? That will be much simpler than this change.

src/gpu/intel/compute/device_info.cpp
```diff
@@ -88,25 +90,16 @@ status_t device_info_t::init(
     return status::success;
 }
 
-const ngen::Product &device_info_t::ngen_product(const gpu_product_t &product) {
-    static_assert(sizeof(ngen::Product) == sizeof(compute::gpu_product_t),
-            "Can't cast gpu_product_t to ngen::Product");
-    return reinterpret_cast<const ngen::Product &>(product);
```
rjoursler, line 94: This is UB as it violate C++ aliasing/lifetime constraints. All uses of this  should be  removed and replaced with `get_ngen_product` if we follow my suggestion.

src/gpu/intel/gemm/jit/include/gemmstone/dsl/hw.hpp
Simonsays095: Fixed

src/gpu/intel/compute/device_info.cpp
Simonsays095: It can (and probably usually is) null here since it's during initialization, thank you for the catch. Took another pass at serialization/deserialization, it should be safe now.

src/gpu/intel/jit/utils/type_bridge.hpp
Simonsays095: Removing `to_gpu_product` (and the inverse, `ngen_product`) are definitely motivating factors - as you point out below, it's been proven to cause issues with UB and compiler warnings recently. This just goes one step further and removes the source of the UB (the proxy struct), using instead the single `ngen::Product` struct in both the interface and memory.

Yes, the unique pointer complicates the memory structure a bit - but the tradeoff is less potential for UB. Maybe some day we can break the transitive dependency chain, or change the implementation to hide ngen a little deeper, but I think this is a step in the right direction in the meantime.

Simonsays095:
make test
disable test_device_cpu
disable build_cpu_runtime_omp
disable build_cpu_runtime_tbb

src/gpu/intel/jit/utils/type_bridge.hpp
rjoursler: What do you think about apply the  pimpl pattern on the device info structure instead of the ngen::Product? This was we can keep the device info structurally simple while still allowing us to avoid exposing unnecessayry details outside of `src/gpu/intel`.

Simonsays095: I thought about that a bit - here's the summary:
- current: `gpu/intel/engine.hpp` has a shared pointer to `compute::device_info_t`. It's initialized in the virtual function `init_device_info` called in `engine_t::init`, which dispatches to the intel/backend engine.
- pimpl: We'd add `gpu/intel/device_info.hpp` that defines a pimpl `gpu::intel::device_info_t`. Then `gpu/intel/engine.hpp` would have this as a direct member, and we'd initialize it from the same virtual `init_device_info` function where we would set the impl as our existing `gpu::intel::compute::device_info_t` (probably renamed to `device_info_impl_t`). We could then store an `ngen::Product` directly and remove the forward declaration - but only after changing any `compute/device_info.hpp` include (even within implementations) to `intel/device_info.hpp`. The implementation lists cannot transitively include ngen, which is where the main crux of this issue is.

This could work, and if we can't break the transitive inclusion of `compute/device_info.hpp` then I think it's the best solution we've got. But it's a pretty big change that I don't want to get into yet. The changes in this PR are conducive to this direction, at least (interfaces directly use an `ngen::Product`, for example).

rjoursler: > We could then store an ngen::Product directly and remove the forward declaration - but only after changing any compute/device_info.hpp include (even within implementations) to intel/device_info.hpp.

I don't see why we need to do this? We can keep the current device info interface and just expose the already declared ngen::Product with:
```
const ngen::Product & product() const;
```
The only real change is that we are moving the structure declaration into device_info.cpp.

Simonsays095: The pimpl pattern uses a unique pointer to store the implementation, so it wouldn't simplify the memory layout more than we we have. With your suggestion (which I interpret as adding a unique pointer to a new `device_info_impl_t` struct in `device_info_t` instead of the `ngen::Product`) we would still have an allocation, and the implementation would just wrap an ngen product.

rjoursler: We already have an allocation [here](https://github.com/uxlfoundation/oneDNN/blob/969e7b39f2881445803b7dd557cb8e867a833399/src/gpu/intel/engine.hpp#L200), so there shouldn't need to be any new allocations. All that is changing is that we are replacing this shared_pointer with a "shared pimpl". Am I missing something?

rjoursler (approved):

PR #5280, opened 2026-06-09
cpu: aarch64: jit: use hardware sp for stack management

Sqvid:
# Description
Switches use of the arbitrary `X_SP` (i.e, `x21`) register over to the hardware stack pointer (i.e, `sp`, `x31`) for stack management. The main advantages are that the hardware `sp` is used by the kernel and other functions while our use of the `X_SP` register is invisible to them. This also frees up x21 for general-purpose use.

# Checklist

## General

### DIFF
--- a/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
+++ b/src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
@@ -107,15 +107,16 @@ void jit_uni_eltwise_injector_t<isa>::injector_preamble(
     assert(preserved_gprs_count == aux_gprs_count());
 
     if (save_state_) {
-        const int reg_size = h->x0.getBit() / 8;
-        if (preserve_p_table_) h->str(x_table, pre_ptr(h->X_SP, -reg_size));
+        // TODO: use register_preserve_guard for better stack management
+        const int reg_size = utils::rnd_up(h->x0.getBit() / 8, 16);
+        if (preserve_p_table_) h->str(x_table, pre_ptr(h->sp, -reg_size));
         for (size_t i = 0; i < preserved_gprs_count; ++i)
-            h->str(XReg(preserved_gpr_idxs[i]), pre_ptr(h->X_SP, -reg_size));
+            h->str(XReg(preserved_gpr_idxs[i]), pre_ptr(h->sp, -reg_size));
 
         if (preserve_vmm_) {
             if (preserved_vecs_count)
-                h->sub_imm(h->X_SP, h->X_SP, preserved_vecs_count * vlen,
-                        h->X_TMP_0);
+                h->sub_imm(
+                        h->sp, h->sp, preserved_vecs_count * vlen, h->X_TMP_0);
             for (size_t i = 0; i < preserved_vecs_count; ++i)
                 store_preserved_vec(i, preserved_vec_idxs[i]);
         }
@@ -135,7 +136,7 @@ void jit_uni_eltwise_injector_t<isa>::injector_preamble_tail(
     const int idx_off = vecs_to_preserve - tail_vecs_to_preserve;
 
     if (save_state_) {
-        if (idx_off) h->add_imm(h->X_SP, h->X_SP, idx_off * vlen, h->X_TMP_0);
+        if (idx_off) h->add_imm(h->sp, h->sp, idx_off * vlen, h->X_TMP_0);
 
         for (size_t i = 0; i < tail_vecs_to_preserve; ++i)
             load_preserved_vec(i, preserved_vec_idxs[idx_off + i]);
@@ -148,7 +149,7 @@ void jit_uni_eltwise_injector_t<isa>::injector_preamble_tail(
         for (size_t i = 0; i < tail_vecs_to_preserve; ++i)
             store_preserved_vec(i, preserved_vec_idxs[idx_off + i]);
 
-        if (idx_off) h->sub_imm(h->X_SP, h->X_SP, idx_off * vlen, h->X_TMP_0);
+        if (idx_off) h->sub_imm(h->sp, h->sp, idx_off * vlen, h->X_TMP_0);
     }
 
     assign_regs();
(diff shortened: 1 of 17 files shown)


Sqvid:
Passing Nightly: https://github.com/uxlfoundation/oneDNN/actions/runs/27220461524

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
```diff
@@ -158,21 +159,20 @@ void jit_uni_eltwise_injector_t<isa>::injector_preamble_tail(
 template <cpu_isa_t isa>
 void jit_uni_eltwise_injector_t<isa>::injector_postamble() {
     using namespace Xbyak_aarch64::util;
-    const int reg_size = h->x0.getBit() / 8;
+    const int reg_size = 16;
```
jondea: Might be worth being a bit more specific here, both WRT which kind of register and that this is bytes

jondea: This is a bit of nit, but I actually think there was some merit to the previous code. Specifically, is there something like
```
XReg::size_bytes()
```
that we can use directly in these situations rather than defining a temp variable?

src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
```diff
@@ -1680,27 +1680,27 @@ void jit_uni_reorder_kernel_f32_t::create_loops(const simple_impl_desc_t &desc,
...
 
         if (curr_node_has_tail) {
-            const size_t reg_bytes = X_TMP_0.getBit() / 8;
+            const size_t reg_bytes = 16;
             if (prb_.nodes[curr_node_id].is_parent_empty()) {
                 mov(reg_loop_cnt, tail_size);
                 // Put info that node is being processed with tail.
                 mov(X_TMP_0, with_tail_info_);
-                str(X_TMP_0, pre_ptr(X_SP, -static_cast<int>(reg_bytes)));
+                str(X_TMP_0, pre_ptr(sp, -static_cast<int>(reg_bytes)));
```
jondea, line 1688: Is there a push/pop stack function?

src/cpu/aarch64/jit_uni_binary_kernel.cpp
```diff
@@ -382,11 +382,11 @@ void jit_uni_binary_kernel_t<isa>::compute_bcast(bool tail) {
 
 template <cpu_isa_t isa>
 void jit_uni_binary_kernel_t<isa>::push(const Xbyak_aarch64::XReg &reg) {
-    str(reg, pre_ptr(X_SP, -static_cast<int>(reg.getBit() / 8)));
+    str(reg, pre_ptr(sp, -16));
```
jondea, line 385: We are now using more of the stack with this method right? The benefit of having two levels of stack pointers is that we can pack our XRegs densely. Do you know how much of an impact this has on perf/stack usage?

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
Sqvid: No we'd have to add it. It is also very annoying that the current function is non-static and badly named. `getBit()` does not clearly reflect that you are going to get the _size_ in bits. What about `rnd_up(x0.getBit() / 8, 16)`?

src/cpu/aarch64/reorder/jit_uni_reorder_kernel.cpp
Sqvid: Long term we probably want to use [`register_preserve_guard`](https://github.com/uxlfoundation/oneDNN/blob/8ff533867c8a8decad26e4cd0f9d7a36d4edc749/src/cpu/aarch64/injectors/injector_utils.hpp#L54). The reason being that if you push and pop individual registers you lose RAII, and you always waste half your space (or more) in padding (i.e, a single `XReg` still needs to move the stack by `16` whereas `register_preserve_guard` can group pairs of stores to avoid unnecessary padding).

src/cpu/aarch64/jit_uni_binary_kernel.cpp
Sqvid: We can still pack densely as we migrate to `register_preserve_guard`. If we do that, we will need at most 8 extra bytes of padding for any number of backed up `XRegs` so it would be negligible. I just want to avoid coupling that migration here. Additionally, there are other patterns that allow dense stack usage. Even in this patch you can see dense stack storage here:
https://github.com/uxlfoundation/oneDNN/blob/79b7e425c80130e61a4a37f8184462519b273d86/src/cpu/aarch64/brgemm/jit_brdgmm_kernel.cpp#L96-L99

Sqvid: [Also note that in my previous patch]( https://github.com/uxlfoundation/oneDNN/commit/6a8e7bb5bdd375d5939472df554e98a04a33670c#diff-fd7b34f41e09fc13dce4e18355fc0e543d31a5b700987ba7270de245724f9ce4L190) in this line of work I removed an extra `128 KB` of stack space. So we are still tremendously more compact than before.

jondea: Seems like a reasonable trade off :)

jondea (approved):

src/cpu/aarch64/injectors/jit_uni_eltwise_injector.cpp
jondea: That function is horrible enough that I don't have strong feelings. But maybe better to use the horrible function until we add a nicer one

Sqvid: Done. changed it to `rnd_up(x0.getBit() / 8, 16)`.

src/cpu/aarch64/jit_generator.hpp
```diff
@@ -172,7 +172,6 @@ class jit_generator_t : public Xbyak_aarch64::CodeGenerator,
     const Xbyak_aarch64::XReg X_TMP_3 = x26;
     const Xbyak_aarch64::XReg X_TMP_4 = x27;
     const Xbyak_aarch64::XReg X_DEFAULT_ADDR = x28;
-    const Xbyak_aarch64::XReg X_SP = x21;
```
renato-arantes, line 175: It would be easier to just set X_SP = sp, no? :)

renato-arantes (approved):

src/cpu/aarch64/jit_generator.hpp
Sqvid: I did that to test the patch, but it will be very confusing for a future reader why we are using it instead of just `sp`.

PR #5286, opened 2026-06-09
build: introduced functions to handle build knobs behavior

vpirogov:
This PR consolidates logic related to managing oneDNN build knobs in two functions: `onednn_option` and `custom_option`. Additional behavior changes:
* Added messages when a build knob is set to non-standard value.
* Reversed the direction of `DNNL_` options mapping. Now `ONEDNN_` is the default as indicated in the documentation. `DNNL_` continues to be supported and generates a compatibility message.
* `ONEDNN_CODE_COVERAGE` default value changed from `OFF` to `NONE` to align with other non-boolean knobs.


cmake/options.cmake
```diff
@@ -24,8 +24,51 @@ set(options_cmake_included true)
...
+function(custom_option name default docstring)
+    if(${ARGC} GREATER 3)
+        set(value "${ARGV3}")
+    else()
+        set(value "${default}")
+    endif()
+
+    if("${default}" MATCHES "^(ON|OFF|TRUE|FALSE|YES|NO|0|1)$")
+        set(${name} ${value} CACHE BOOL "${docstring}")
+        if((NOT ${${name}} AND ${default}) OR (${${name}} AND NOT ${default}))
```
dzarukin, line 46: What does first OR predicate cover?

cmake/options.cmake
```diff
@@ -24,8 +24,51 @@ set(options_cmake_included true)
...
+        endif()
+    else()
+        set(${name} "${value}" CACHE STRING "${docstring}")
+        if(NOT "${${name}}" STREQUAL "${default}")
+            message(STATUS "${name}: ${${name}} (default: ${default})")
+        endif()
+    endif()
+endfunction()
+
+function(onednn_option name default docstring)
```
dzarukin, line 59: Maybe add a comment that this one must be used if the build option is supposed to start with `ONEDNN_`?

dzarukin (approved):

cmake/options.cmake
vpirogov: This construct is a XOR. In CMake all variables are strings, so `${name} NOT STREQUAL ${default}` will not work for things like `ON` and `TRUE`.

vpirogov: Added a comment on why it's needed.

cmake/options.cmake
vpirogov: Yeees.

cmake/options.cmake
```diff
@@ -355,52 +387,47 @@ if(NOT "${DNNL_GPU_RUNTIME}" MATCHES "^(OCL|NONE|DPCPP|SYCL|ZE)$")
     message(FATAL_ERROR "Unsupported GPU runtime: ${DNNL_GPU_RUNTIME}")
 endif()
 
-set(DNNL_GPU_VENDOR "NONE" CACHE STRING
+onednn_option(GPU_VENDOR "INTEL"
```
mzhukova, line 390: so seems like `NONE` is no longer available (even for CPU-only, expected?), maybe change "INTEL (default when DNNL_GPU_RUNTIME is not NONE" to simply "INTEL (default)" in this case?

vpirogov: This is intended. `NONE` is not a valid option when `DNNL_GPU_RUNTIME` is set resulting in weird logic of changing the default value. Updated the description.

mzhukova (approved):

PR #5288, opened 2026-06-09
Fixes and improvements for generic SYCL kernels

dzarukin:
This PR intends to unblock the KernelFoundry team in their endeavor to benchmark their kernels against oneDNN.
They have a dedicated library that has a single entry into their kernels.
The idea is to patch generic SYCL implementations with a piece of code that would check whether a KernelFoundry kernel was requested (by searching for the library in runtime (preloaded with LD_PRELOAD), or by a dedicated env variable) and dispatch into that call with arguments from execute(...).

To have an ability for performance comparison out-of-the-box, the library should be built against ONEDNN_GPU_VENDOR=INTEL w/ generic SYCL kernels (which a new build option targets, otherwise, they are excluded by default).

Also, to compile and have an ability to skip undesired implementations some changes in names and proper checks are needed.

Lastly, matmul kernel arguments size exceeded reasonable amount for any problem which required a radical cut for `md_t` which didn't use any blocks in kernels.

I also copied SYCL profiling capabilities for generic backend to have apples-to-apples comparison.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -436,3 +436,7 @@ onednn_option(AMD_SYCL_KERNELS_TARGET_ARCH ""
     Warning: This option is temporary and will be removed as soon as the compiler
     stops to require specifying the target architecture. After removing the option
     the generic SYCL kernels will always be enabled for AMD vendor.") # internal
+
+onednn_option(INTERNAL_ENABLE_GENERIC_SYCL_KERNELS OFF
+    "Enables implementations with generic SYCL kernels when DNNL_GPU_VENDOR is
+    not GENERIC.")
--- a/src/gpu/generic/sycl/rnn/ref_rnn.hpp
+++ b/src/gpu/generic/sycl/rnn/ref_rnn.hpp
@@ -176,7 +176,7 @@ struct ref_rnn_fwd_t : ref_rnn_common_base_t {
 
         pd_t(const pd_t &other) = default;
 
-        DECLARE_COMMON_PD_T("ref:any", ref_rnn_fwd_t);
+        DECLARE_COMMON_PD_T("sycl:ref:any", ref_rnn_fwd_t);
 
         status_t init(impl::engine_t *engine);
 
@@ -263,7 +263,7 @@ struct ref_rnn_bwd_t : ref_rnn_common_base_t {
 
         pd_t(const pd_t &other) = default;
 
-        DECLARE_COMMON_PD_T("ref:any", ref_rnn_bwd_t);
+        DECLARE_COMMON_PD_T("sycl:ref:any", ref_rnn_bwd_t);
 
         status_t init(impl::engine_t *engine);
 
--- a/src/gpu/generic/sycl/sycl_gpu_kernel.cpp
+++ b/src/gpu/generic/sycl/sycl_gpu_kernel.cpp
@@ -14,11 +14,10 @@
 * limitations under the License.
 *******************************************************************************/
 
-#include "common/stream.hpp"
-
 #include "xpu/sycl/stream_impl.hpp"
 
 #include "gpu/generic/sycl/sycl_gpu_kernel.hpp"
+#include "gpu/gpu_stream.hpp"
 
 namespace dnnl {
 namespace impl {
@@ -38,6 +37,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
         cgh.use_kernel_bundle(*kernel_bundle_);
         cgf(cgh);
     });
+
+    if (stream.is_profiling_enabled()) {
+        auto sycl_event = utils::make_unique<xpu::sycl::event_t>(
+                std::vector<::sycl::event> {event});
+        auto *gpu_stream = utils::downcast<gpu::stream_t *>(&stream);
+        gpu_stream->profiler().register_event(std::move(sycl_event));
+    }
 
     deps = {event};
     return status::success;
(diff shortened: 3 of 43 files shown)


dzarukin:
make test
build_vendor_generic

src/gpu/generic/sycl/sycl_gpu_kernel.cpp
```diff
@@ -39,6 +39,13 @@ status_t kernel_t::parallel_for(impl::stream_t &stream,
         cgf(cgh);
     });
 
+    if (stream.is_profiling_enabled()) {
+        auto sycl_event = utils::make_unique<xpu::sycl::event_t>(
+                std::vector<::sycl::event> {event});
+        auto *gpu_stream = utils::downcast<gpu::stream_t *>(&stream);
+        gpu_stream->profiler().register_event(std::move(sycl_event));
+    }
+
```
avmanerikar, line 47: (nit):
`stream_t::copy()` and `stream_t::fill()` may also need similar modifications to account for profiling:
https://github.com/uxlfoundation/oneDNN/blob/0ecf484bb116949751accd309b451c99fec8745d/src/xpu/sycl/stream_impl.cpp#L139
https://github.com/uxlfoundation/oneDNN/blob/0ecf484bb116949751accd309b451c99fec8745d/src/xpu/sycl/stream_impl.cpp#L185

cmake/options.cmake
```diff
@@ -400,13 +400,6 @@ option(DNNL_DISABLE_GPU_REF_KERNELS
 # Other GPU options
 # -----------------
 
-set(DNNL_AMD_SYCL_KERNELS_TARGET_ARCH "" CACHE STRING
```
densamoilov: Are we dropping the generic sycl kernels support for amd?

dzarukin: I removed everything related to this option.
If sycl kernels can't compile for AMD without those compilation options, I can restore them conditionally under a new build option/macro.
Users can switch to a new build option to enable sycl kernels in terms of support enabling.
However, with the transition of ICX to open source and stripping off Nvidia and AMD sycl compilation targets support, I don't think it's a big deal, if it's a deal at all...

densamoilov: This option was (is) required to let the SYCL compiler know what arch it should compile the kernels for. It was (is) required for AMD GPUs, otherwise it fails to compile. For NVIDIA backend we specify a generic arch option when compiling the generic SYCL kernels.

This change essentially disables generic SYCL kernels for AMD GPUs.

src/gpu/generic/sycl/sycl_gpu_kernel.cpp
dzarukin: Added a profiler pointer in fill and copy, thanks for the catch!

cmake/options.cmake
dzarukin: Agreed to restore the option and let it retire later. Thanks.

dzarukin:
make test
build_vendor_generic

src/xpu/sycl/types.hpp
```diff
@@ -118,9 +118,14 @@ using in_memory_arg_t = memory_arg_t<::sycl::access::mode::read>;
...
+// is used inside conf classes which are passed to kernels as arguments
+// directly. There's a hard limit on arguments size and the less is used, the
+// more can be put into those conf classes.
+//
+// Note: if, for whatever reason, blocking descriptor will be needed, a new
+// abstraction should be introduced and used with those memory descriptors that
+// would utilize it.
 struct md_t {
     // There is a limitation on total size of kernel arguments hence using
     // reduced number of supported dimensions and int32_t for dimensions.
```
rjoursler, line 131: Comment: This restriction seems like it could be removed due to the amount of data deleted from this structure. At least for the intel kernels, we need large buffer support, so this divergence could be relevant to performance comparisons.

src/gpu/generic/sycl/rnn/ref_rnn.hpp
```diff
@@ -263,7 +263,7 @@ struct ref_rnn_bwd_t : ref_rnn_common_base_t {
 
         pd_t(const pd_t &other) = default;
 
-        DECLARE_COMMON_PD_T("ref:any", ref_rnn_bwd_t);
+        DECLARE_COMMON_PD_T("dpcpp:ref:any", ref_rnn_bwd_t);
```
rjoursler: nit here and other similar places: Use `sycl:ref:any` as we are targeting the SYCL language.

cmake/options.cmake
```diff
@@ -436,3 +436,7 @@ onednn_option(AMD_SYCL_KERNELS_TARGET_ARCH ""
     Warning: This option is temporary and will be removed as soon as the compiler
     stops to require specifying the target architecture. After removing the option
     the generic SYCL kernels will always be enabled for AMD vendor.") # internal
+
+onednn_option(INTERNAL_ENABLE_GENERIC_SYCL_KERNELS OFF
+    "Enables implementations with generic SYCL kernels when DNNL_GPU_VENDOR is
+    not GENERIC.")
```
rjoursler, line 442: How do we plan to test this so that we don't accidentally break the feature?

rjoursler (approved):

src/xpu/sycl/types.hpp
dzarukin: We probably can, but what for? The smaller it stays - the better.

cmake/options.cmake
dzarukin: Same as we test AMD's SYCL kernels - we basically don't.
KernelFoundry team will report any problems if any are encountered but I don't see what should break here, it uses regular infrastructure that's been there for several years.

src/xpu/sycl/types.hpp
rjoursler: No clue, this is really just due to the stream of integer overflow bugs I have been squashing over the past year. :scream_cat:

src/gpu/generic/sycl/rnn/ref_rnn.hpp
dzarukin: Agreed, updated all names with sycl over dpcpp, thanks.

densamoilov (approved):

PR #5289, opened 2026-06-10
[GPU] Migrate inner product to matmul

echeresh:
Part of https://github.com/uxlfoundation/oneDNN/pull/5181.


echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

src/gpu/intel/matmul/utils.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "common/c_types_map.hpp"
+#include "common/matmul_pd.hpp"
+#include "common/primitive_desc_iterator.hpp"
+#include "common/utils.hpp"
+
+namespace dnnl {
+namespace impl {
+
+static inline status_t create_matmul_pd(
```
dzarukin: I can see identical calls in src/cpu/x64/matmul_inner_product.hpp, src/gpu/generic/sycl/rnn/ref_rnn.cpp and src/gpu/intel/gated_mlp/ref.hpp.

I think it's time to put this declaration into common/matmul_pd.hpp and create a .cpp with its body to increase reusability. Skipping CPU piece would be fine not to trigger CPU team approval on this PR.

dzarukin (commented):
Rest changes are LGTM.

src/gpu/intel/matmul/utils.hpp
echeresh: Thanks, good idea, I'll move this to `src/common`. I'm not sure about the CPU usage - it uses a special check function:

https://github.com/uxlfoundation/oneDNN/blob/3c67a1e0959c8d60cbff3d80dd76c2f59ade7666/src/cpu/x64/matmul_inner_product.cpp#L61-L65

We can do:

```c++
create_matmul_pd(..., const std::function<bool(const std::shared_ptr<primitive_desc_t> &)> &accept = nullptr);
```

but I'm reluctant to do this just because of this single CPU usage so my suggestion would be to leave it as is.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

src/common/matmul.cpp
```diff
@@ -710,6 +712,33 @@ status_t matmul_desc_init(matmul_desc_t *matmul_desc,
...
+            engine, (op_desc_t *)&matmul_desc, &matmul_attr, nullptr);
+    if (!it.is_initialized()) return status::out_of_memory;
+
+    while (++it != it.end()) {
+        if (*it) {
+            matmul_pd = *it;
+            return status::success;
+        }
+    }
+    return status::unimplemented;
```
dzarukin, line 740: Maybe VCHECK_MATMUL_UNIMPL?

src/common/primitive_desc.hpp
```diff
@@ -575,6 +576,10 @@ struct primitive_desc_t : public c_compatible {
 };
 // NOLINTEND(google-default-arguments)
 
+inline bool is_ref_impl(const primitive_desc_t *pd) {
+    return pd && strstr(pd->name(), "ref") != nullptr;
```
dzarukin: I recommend to check for "ref:any" based on what the library provides. "ref" can trigger way more patterns...

I also think some documentation for the function wouldn't hurt explaining the "reference" is apriori the slowest implementation and undesired in a caller context.

dzarukin (approved):

src/common/primitive_desc.hpp
echeresh: Thanks, fixed.

echeresh:
make test
set test_scope=NIGHTLY
disable test_device_cpu
disable benchdnn_all
enable benchdnn_matmul
enable benchdnn_ip

src/common/primitive_desc.hpp
```diff
@@ -575,6 +576,13 @@ struct primitive_desc_t : public c_compatible {
 };
 // NOLINTEND(google-default-arguments)
 
+// Reference impls are apriori the slowest and undesired when
+// creating a nested primitive. Match the "ref:any" naming
+// convention.
+inline bool is_ref_impl(const primitive_desc_t *pd) {
+    return pd && strstr(pd->name(), "ref:any") != nullptr;
+}
```
rjoursler, line 584: Did you consider just integrating this check into `create_matmul_pd`? It seems like doing that would generally simplify the users of that interface.

rjoursler (approved):

src/common/primitive_desc.hpp
echeresh: Yes, this was the initial version. If you think it's better to move it to the signature - I can update that.

My reasoning was that `create_matmul_pd()` signature is already somewhat overloaded + having an explicit `is_ref_impl()` check in callers vs hidden in a long argument list would rather help readability

rjoursler: Okay, this version is fine with me.

echeresh:
make test perf-gpu
set primitive=ip matmul rnn

